PR #230, opened 2024-04-09
refactor(service): serialize large binary data as BCS

jpcsmith:
An take on option nr. 2 of #213 that uses a custom extractor and implementation of `IntoResponse` to selectively apply BCS encoding to responses.

Closes #186

PR #241, opened 2024-04-10
refactor: use base-64 blob ID string representation

jpcsmith:
Replaces the Hex encoding with a URL-safe base64 encoding.

Depends on #230

### DIFF
--- a/crates/walrus-core/src/lib.rs
+++ b/crates/walrus-core/src/lib.rs
@@ -3,7 +3,7 @@
 
 //! Core functionality for Walrus.
 use std::{
-    fmt::{self, Debug, Display, LowerHex},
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/server.rs
```diff
@@ -42,7 +40,7 @@ pub const RECOVERY_ENDPOINT: &str = "/v1/blobs/:blobId/slivers/:sliverPairIdx/:s
 /// A blob ID encoded as a hex string designed to be used in URLs.
```
giac-mysten: Nit
```suggestion
/// A blob ID encoded as a Base64 string designed to be used in URLs.
```

PR #205, opened 2024-04-04
feat(core): compute supported source symbols from n_shards

giac-mysten:
PR to start the discussion on the configuration and encoding versioning.

PR #243, opened 2024-04-10
feat(service): integrate sui client into storage client

giac-mysten:


crates/walrus-core/src/metadata.rs
```diff
@@ -142,6 +142,21 @@ impl<const V: bool> BlobMetadataWithId<V> {
     pub fn metadata(&self) -> &BlobMetadata {
         &self.metadata
     }
+
+    /// Returns the encoded length of the blob for which this metadata was generated.
+    ///
+    /// Returns `None` if the blob size cannot be computed.
+    pub fn encoded_blob_length(&self, config: &EncodingConfig) -> Option<u64> {
```
jpcsmith: Suggestion: Move to to `config.encoded_length(unencoded_length)` as the calculations here are more in line with the internal details of encoding.

crates/walrus-core/src/metadata.rs
```diff
@@ -142,6 +142,21 @@ impl<const V: bool> BlobMetadataWithId<V> {
...
+        let n_shards = config.n_shards_as_usize();
+        let symbol_size = config
+            .symbol_size_for_blob(self.metadata.unencoded_length.try_into().ok()?)?
+            .get() as u64;
+        let slivers_size = (config.source_symbols_primary.get() as u64
+            + config.source_symbols_secondary.get() as u64)
+            * symbol_size;
+        let metadata_size = n_shards * DIGEST_LEN * 2 + BlobId::LENGTH;
```
jpcsmith: Suggestion: Add a `EncodingConfig::metadata_length()` function to the config, as this value may be useful elsewhere.

crates/walrus-service/bin/client.rs
```diff
@@ -18,6 +20,12 @@ struct Args {
     /// The configuration file with the committee information.
     #[clap(short, long, default_value = "config.yml")]
     config: PathBuf,
+    /// The path to the wallet config file.
+    #[clap(short, long, default_value = default_client_config().into_os_string())]
```
```suggestion
    #[clap(short, long, default_value_os_t = default_client_config())]
```

crates/walrus-service/bin/client.rs
```diff
@@ -18,6 +20,12 @@ struct Args {
...
+    wallet: PathBuf,
+    /// The gas budget for the transactions.
+    #[clap(short, long, default_value = "1_000_000_000")]
```
```suggestion
    #[clap(short, long, default_value_t = 1_000_000_000)]

```

crates/walrus-service/bin/client.rs
```diff
@@ -42,22 +52,40 @@ enum Commands {
     },
 }
 
+fn default_client_config() -> PathBuf {
+    PathBuf::from(env::var("HOME").expect("environment variable HOME should be set"))
+        .join(".sui")
+        .join("sui_config")
+        .join("client.yaml")
```
jpcsmith: Remark: Can we avoid panicking here? Server users may not have a home directory. 

Perhaps we can check if a config exists in the currently directory, if not try to get home and return that, if not set to the local config (which doesn't exist) and return that. Thoughts?

crates/walrus-service/bin/client.rs
```diff
@@ -42,22 +52,40 @@ enum Commands {
...
+
 /// The client.
 #[tokio::main]
 pub async fn main() -> Result<()> {
     tracing_subscriber::fmt::init();
     let args = Args::parse();
     let config: Config = serde_yaml::from_str(&std::fs::read_to_string(args.config)?)?;
-    let client = Client::new(config);
+    // NOTE(giac): at the moment the wallet is needed for both reading and storing, because is also
+    // configures the RPC. In the future, it could be nice to have a "read only" client.
```
jpcsmith, line 93: Remark: This can probably already be done. 

If you change `trait ContractClient` to `trait ContractClient: ReadClient`, and have it forward the calls to the inner read client. Then the client's current impl block can be `impl<T: ReadClient>` and `reserve_and_store_blob` can go in an `impl<T: ContractClient>` block.

crates/walrus-service/src/client.rs
```diff
@@ -23,59 +28,99 @@ mod config;
...
+pub struct Client<T> {
+    reqwest_client: ReqwestClient,
+    sui_client: T,
     committee: Committee,
     concurrent_requests: usize,
     encoding_config: EncodingConfig,
 }
 
-impl Client {
+impl<T: ContractClient> Client<T> {
```
jpcsmith, line 49: Suggestion: Leave this as `impl<T: ReadClient>`, and instead add `reserve_and_store_blob` into a new `impl<T: ContractClient>`.

crates/walrus-service/src/client.rs
```diff
@@ -23,59 +28,99 @@ mod config;
...
+        let reqwest_client = ClientBuilder::new()
             .timeout(config.connection_timeout)
             .build()?;
-        let encoding_config = config.encoding_config();
+        let committee = sui_client.read_client().current_committee().await?;
+        let encoding_config = EncodingConfig::new(
+            config.source_symbols_primary.get(),
+            config.source_symbols_secondary.get(),
+            committee.total_weight,
+        );
```
jpcsmith, line 65: Question: Just to confirm, this will no longer use the config once #205 lands, correct?

crates/walrus-service/src/client.rs
```diff
@@ -23,59 +28,99 @@ mod config;
...
+    /// storage nodes. Finally, the function aggregates the storage confirmations and posts the
+    /// [`ConfirmationCertificate`] on chain.
+    pub async fn reserve_and_store_blob(&self, blob: &[u8], epochs_ahead: u64) -> Result<Blob> {
         let (pairs, metadata) = self
             .encoding_config
             .get_blob_encoder(blob)?
             .encode_with_metadata();
+        let encoded_length = metadata
+            .encoded_blob_length(&self.encoding_config)
+            .expect("the metadata was computed above");
```
jpcsmith:
```suggestion
            .expect("valid for metadata created from the same config");
```

crates/walrus-service/src/client.rs
```diff
@@ -23,59 +28,99 @@ mod config;
...
+        self.sui_client
+            .certify_blob(&blob_sui_object, &certificate)
+            .await
+            .map_err(|e| anyhow!("blob certification failed: {e}"))
+    }
+
+    /// Stores the already-encoded metadata and sliver pairs for a blob into Walrus, by sending
+    /// sliver pairs to at least 2f+1 shards.
+    ///
+    /// Assumest the blob ID has already been registered, with an appropriate blob size.
```
jpcsmith: Loving the old english.

crates/walrus-service/src/client.rs
```diff
@@ -90,43 +135,78 @@ impl Client {
...
+            }
+        }
+
+        anyhow::ensure!(
+            self.committee.is_quorum(total_weight),
+            "not enough confirmations for the blob id were retrieved"
+        );
+        // The signers must sorted in the `ConfirmationCertificate`.
+        // TODO(giac): remove when #238 is closed.
+        signers.sort_unstable();
```
jpcsmith:
```suggestion
```

crates/walrus-sui/src/types.rs
```diff
@@ -288,6 +289,27 @@ impl Committee {
     pub fn quorum_threshold(&self) -> usize {
         (self.total_weight as usize * 2 + 2) / 3
     }
+
+    /// Return the shards handed by the specified storage node.
```
jpcsmith:
```suggestion
    /// Return the shards handed by the specified storage node, based on its index in the committee list.
```

crates/walrus-sui/src/types.rs
```diff
@@ -288,6 +289,27 @@ impl Committee {
...
+    /// Panic if the committee has no shards.
+    pub fn n_shards(&self) -> NonZeroU16 {
+        let shards = self
+            .members
+            .iter()
+            .map(|node| node.shard_ids.len())
+            .sum::<usize>()
+            .try_into()
+            .expect("should fit into a `u16`");
+        NonZeroU16::new(shards).expect("committee has no shards")
```
jpcsmith:
```suggestion
        NonZeroU16::new(shards).expect("committee to have at least 1 shard")
```

crates/walrus-service/bin/client.rs
giac-mysten: Indeed, I had implemented it this way initially, but after a long discussion with Karl we agreed that for the moment it's not so much worth it.

The main issue is not the trait structure, but the fact that the `WalletContext` is very convenient to configure the RPC and sui environment. So having a read client without the wallet requires having some other config that specifies the rpc.

Otoh, adding an `--rpc` flag to specify it directly is not so much more work. What would your preference be?

crates/walrus-service/src/client.rs
giac-mysten: We will still need the `Config` for parameters such as timeouts and concurrent connections, but yes, not for the number of source symbols.

crates/walrus-service/src/client.rs
giac-mysten: rofl

crates/walrus-service/bin/client.rs
giac-mysten: very good point. I added something in that direction, although I didn't quite catch the point about the local config.
In my edit, it checks if a flag with the path is provided, if not checks `./client.yaml`, if not tries `~/.sui/...`.

crates/walrus-service/bin/client.rs
```diff
@@ -42,22 +52,64 @@ enum Commands {
...
+/// Loads the wallet context from the given path.
+///
+/// If no path is provided, tries to load the configuration first from the local folder, and then
+/// from the home folder.
+fn load_wallet_context(path: &Option<PathBuf>) -> Result<WalletContext> {
+    let path = path
+        .as_ref()
+        .cloned()
+        .or(local_client_config())
+        .or(home_client_config())
```
```suggestion
        .or_else(local_client_config)
        .or_else(home_client_config)
```

crates/walrus-service/bin/client.rs
jpcsmith: Ah, thank you for the clarification, then just leave it as is with the note.

crates/walrus-service/src/client.rs
```diff
@@ -91,43 +138,75 @@ impl Client {
...
+                }
+                Err(e) => tracing::error!(
+                    "storing metadata and pairs to node {} failed with error: {}",
+                    node,
+                    e
+                ),
+            }
+        }
+
+        anyhow::ensure!(
```
jpcsmith: Nit: I think markus created an ensure variant that can be used to generate either anyhow::Errors or custom errors. May be good to use that here.

crates/walrus-service/bin/client.rs
jpcsmith: This works, thanks. To clarify my earlier suggestion, the idea was to keep the argument as a `PathBuf`. However, in the event that we're unable to determine that a local config or a config in `~/.sui` exists (e.g., `exists()` fails for both), then we would return the default value as `~/.client.yaml`. The program would then try to open that file and an error that "~/.client.yaml does not exist" would be raised.

I prefer the `Option<>` approach however, as it makes it clear that the default value is not a static value.

Issue #90, opened 2024-03-11
Move Blob Invalid logic and certification


gdanezis:
Add contract logic:
* For a blob to be declared invalid using a certificate.
* For an event to be fired when a blob is invalid.

PR #245, opened 2024-04-10
feat: Initial invalid ID handling

gdanezis:
- Add an invalid message blob ID message that is certified
- Add docs about its type number
- Emit an event using the system if the certification epoch is the system epoch
- Add tests (Still more TODO)

Related to #90

### DIFF
--- a/contracts/blob_store/sources/tests/invalid_tests.move
+++ b/contracts/blob_store/sources/tests/invalid_tests.move
@@ -117,8 +117,9 @@ module blob_store::invalid_tests {
             0, // Intent version
             5, // Epoch
             6, // Stake support
-            vector[171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171,
-            171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171], // Data
+            // Data
+            vector[171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171,
+            171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171],
 
         );
 
@@ -199,7 +200,8 @@ module blob_store::invalid_tests {
 
 
 
-        let expected_blob_id : u256 = 0xabababababababababababababababababababababababababababababababab;
+        let expected_blob_id : u256
+            = 0xabababababababababababababababababababababababababababababababab;
 
         // Now check this is a invalid blob message
         let blob_id = system::invalidate_blob_id(
@@ -268,8 +270,9 @@ module blob_store::invalid_tests {
             0, // Intent version
             50, // Epoch WRONG
             6, // Stake support
-            vector[171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171,
-            171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171], // Data
+            // Data
+            vector[171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171,
+            171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171],
 
         );
 


contracts/blob_store/sources/system.move
```diff
@@ -311,4 +319,43 @@ module blob_store::system {
         });
     }
 
+    // The logic to register and invalid Blob ID
```
karlwuest:
```suggestion
    // The logic to register an invalid Blob ID
```

contracts/blob_store/sources/system.move
```diff
@@ -311,4 +319,69 @@ module blob_store::system {
...
+    /// nodes will only certify invalid blob ids within their period of validity, and this
+    /// endures we are not flooded with invalid events from past epochs.
+    public fun declare_invalid_blob_idh<WAL>(
+        system: &System<WAL>,
+        message: CertifiedInvalidBlobID,
+    ) {
+
+        // Assert the epoch is correct.
+        let epoch = message.epoch;
+        assert!(epoch == epoch(system), ERROR_SYNC_EPOCH_CHANGE);
```
karlwuest: Is `ERROR_SYNC_EPOCH_CHANGE` the right error to emit here?

contracts/blob_store/sources/system.move
```diff
@@ -311,4 +319,69 @@ module blob_store::system {
...
+        // This output is provided as a service in case anything else needs to rely on
+        // certified invalid blob ID information in the future. But out base design only
+        // uses the event emitted here.
+        CertifiedInvalidBlobID { epoch, blob_id }
+    }
+
+    /// System call to process invalid blob id message. This checks that the epoch
+    /// in which the message was certified is correct, before emitting an event. Correct
+    /// nodes will only certify invalid blob ids within their period of validity, and this
+    /// endures we are not flooded with invalid events from past epochs.
```
karlwuest: Currently, given that the only way to access a committee is either through the system object or by owning a `CommitteeCapHolder`, all certified invalid blobs should be from the current epoch as far as I can tell.

Do we ever expect a scenario, where we do not have a fixed sequence of function calls for invalidating a blob id (i.e. where we need the customizability of PTBs)? 
 
If not, it may be cleaner to not expose all these functions publicly, and instead just provide a single public function that performs all required calls (sidenote: since there is one "common" flow, we should anyway have such a function, similar to the `certify` function in `blob.move`). Then we don't have to perform these kind of checks multiple times.

contracts/blob_store/sources/tests/invalid_tests.move
```diff
@@ -0,0 +1,127 @@
...
+        let mut ctx = tx_context::dummy();
+
+        let blob_id : u256 = 0xabababababababababababababababababababababababababababababababab;
+
+        // BCS confirmation message for epoch 0 and blob id `blob_id` with intents
+        let invalid_message : vector<u8> = vector[2, 0, 3, 5, 0, 0, 0, 0, 0, 0, 0,
+            171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171,
+            171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171, 171];
+
+        // Signature from private key scalar(117) on `confirmation`
```
karlwuest:
```suggestion
        // Signature from private key scalar(117) on `invalid_message`
```

contracts/blob_store/sources/tests/invalid_tests.move
```diff
@@ -0,0 +1,127 @@
...
+        let invalid_message_signature : vector<u8> = vector[
```
karlwuest: Suggestion: Just call this `message_signature`, otherwise it sounds as if the signature is invalid, which confused me for a moment.

contracts/blob_store/sources/tests/invalid_tests.move
```diff
@@ -0,0 +1,127 @@
...
```
karlwuest: See comments above on the other test

contracts/blob_store/sources/tests/invalid_tests.move
```diff
@@ -0,0 +1,127 @@
...
+            string::utf8(b"127.0.0.1"),
+            public_key,
+            vector[0, 1, 2, 3, 4, 5]
+        );
+
+        // Create a new committee
+        let cap = committee::create_committee_cap_for_tests();
+        let committee = committee::create_committee(&cap, 999, vector[storage_node]);
+
+        let certified_message = committee::verify_quorum_in_epoch(
```
karlwuest: Can you add a comment that the test should fail here and remove everything below that doesn't run?

contracts/blob_store/sources/system.move
gdanezis: Oops, nice catch, defined a reasonable error.

contracts/blob_store/sources/tests/invalid_tests.move
gdanezis: Done!

contracts/blob_store/sources/tests/invalid_tests.move
gdanezis: I actually changed this test, to now test the system level invalid ID with the incorrect epoch. Otherwise we are checking the message verification not the epoch check specific to invalid ID. Good catch.

contracts/blob_store/sources/system.move
gdanezis: You are right. I restricted this function to package, and then I created a separate
```
    public fun invalidate_blob_id<WAL>(
        system: &System<WAL>,
        signature: vector<u8>,
        members: vector<u16>,
        message: vector<u8>,
    ) : u256
```
That does everything for you.

Issue #221, opened 2024-04-08
Remove openSSL in orchestrator


asonnino:
The orchestrator can be modified to operate using `rusttls` rather than `ssh2`. Tasos made this change for the orchestrator on the Sui repo, so we can get inspired from there.

PR #247, opened 2024-04-10
feat: Initial orchestrator code

asonnino:
Initial code to setup remote instances, configure nodes, and boot clients.

### DIFF
--- a/crates/walrus-orchestrator/src/monitor.rs
+++ b/crates/walrus-orchestrator/src/monitor.rs
@@ -176,8 +176,8 @@ impl Grafana {
             "sudo apt-get install -y apt-transport-https software-properties-common wget".into(),
             "sudo wget -q -O /etc/apt/keyrings/grafana.key https://apt.grafana.com/gpg.key".into(),
             "(sudo rm /etc/apt/sources.list.d/grafana.list || true)".into(),
-            "echo
-            \"deb [signed-by=/usr/share/keyrings/grafana.key] https://apt.grafana.com stable main\"
+            "echo \
+            \"deb [signed-by=/etc/apt/keyrings/grafana.key] https://apt.grafana.com stable main\" \
             | sudo tee -a /etc/apt/sources.list.d/grafana.list"
                 .into(),
             "sudo apt-get update".into(),
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -25,9 +25,9 @@ mod utils;
 
 pub use self::config::Config;
 use self::{
-    communication::NodeCommunication,
+    communication::{NodeCommunication, NodeResult},
     error::SliverRetrieveError,
-    utils::{WeightedFutures, WeightedResult},
+    utils::WeightedFutures,
 };
 
 /// A client to communicate with Walrus shards and storage nodes.
@@ -84,7 +84,8 @@ impl Client {
                 self.concurrent_requests,
             )
             .await;
-        let results = requests.into_results();
+        let results = requests.take_inner_ok();
+        drop(requests);
         Ok((metadata, results))
     }
 
@@ -128,8 +129,7 @@ impl Client {
             )
             .await;
 
-        let slivers = requests.take_results();
-
+        let slivers = requests.take_inner_ok();
         if let Some((blob, _meta)) = decoder.decode_and_verify(metadata.blob_id(), slivers)? {
             // We have enough to decode the blob.
             Ok(blob)
@@ -145,19 +145,25 @@ impl Client {
     /// sliver it receives.
     async fn decode_sliver_by_sliver<'a, I, Fut, T>(
         &self,
-        requests: &mut WeightedFutures<I, Fut, Sliver<T>, SliverRetrieveError>,
+        requests: &mut WeightedFutures<I, Fut, NodeResult<Sliver<T>, SliverRetrieveError>>,
         decoder: &mut BlobDecoder<'a, T>,
         blob_id: &BlobId,
     ) -> Result<Vec<u8>>
     where
         T: EncodingAxis,
         I: Iterator<Item = Fut>,
-        Fut: Future<Output = WeightedResult<Sliver<T>, SliverRetrieveError>>,
+        Fut: Future<Output = NodeResult<Sliver<T>, SliverRetrieveError>>,
     {
-        while let Some(sliver) = requests.execute_next(self.concurrent_requests).await {
-            let result = decoder.decode_and_verify(blob_id, [sliver])?;
-            if let Some((blob, _meta)) = result {
-                return Ok(blob);
+        while let Some(NodeResult(_, _, _, result)) = requests.next(self.concurrent_requests).await
+        {
+            match result {
+                Ok(sliver) => {
+                    let result = decoder.decode_and_verify(blob_id, [sliver])?;
+                    if let Some((blob, _meta)) = result {
+                        return Ok(blob);
+                    }
+                }
+                Err(_e) => (), // TODO(giac): add tracing
             }
         }
         // We have exhausted all the slivers but were not able to reconstruct the blob.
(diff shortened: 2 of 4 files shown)


crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -0,0 +1,374 @@
...
+#[allow(dead_code)] // TODO(Alberto): Will be used to deploy nodes (#222)
+pub struct Orchestrator<P, N, C> {
+    /// The testbed's settings.
+    settings: Settings,
+    /// The state of the testbed (reflecting accurately the state of the machines).
+    instances: Vec<Instance>,
+    /// The node's configuration parameters.
+    node_config: PhantomData<N>,
+    /// The client's configuration parameters.
+    client_config: PhantomData<C>,
```
mlegner: Question: Why is this `PhantomData`? Will this be changed later?

crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -0,0 +1,374 @@
...
+            client_config: PhantomData,
+            instance_setup_commands,
+            protocol_commands,
+            ssh_manager,
+        }
+    }
+
+    /// Select on which instances of the testbed to run the benchmarks. This function returns two
+    /// vector of instances; the first contains the instances on which to run the load generators
+    /// and the second contains the instances on which to run the nodes.
```
mlegner: Nit:
```suggestion
    /// Returns the instances of the testbed on which to run the benchmarks. 
    /// 
    /// This function returns two vectors of instances; the first contains the instances on which to
    /// run the load generators and the second contains the instances on which to run the nodes.
    /// Additionally returns an optional monitoring instance.
```

crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -0,0 +1,374 @@
...
+        &self,
+        parameters: &BenchmarkParameters,
+    ) -> TestbedResult<(Vec<Instance>, Vec<Instance>, Option<Instance>)> {
+        // Ensure there are enough active instances.
+        let available_instances: Vec<_> = self.instances.iter().filter(|x| x.is_active()).collect();
+        let minimum_instances = if self.settings.monitoring {
+            parameters.nodes + self.settings.dedicated_clients + 1
+        } else {
+            parameters.nodes + self.settings.dedicated_clients
+        };
```
mlegner: Nit to deduplicate:
```suggestion
        let minimum_instances = parameters.nodes
            + self.settings.dedicated_clients
            + if self.settings.monitoring { 1 } else { 0 };
```

crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -0,0 +1,374 @@
...
+        if self.settings.monitoring {
+            for region in &self.settings.regions {
+                if let Some(regional_instances) = instances_by_regions.get_mut(region) {
+                    if let Some(instance) = regional_instances.pop_front() {
+                        monitoring_instance = Some(instance.clone());
+                    }
+                    break;
+                }
+            }
+        }
```
mlegner: Nit: Possible optimization: Take the monitoring instance from the region with the most instances.

crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -0,0 +1,374 @@
...
+
+        // Select the instances to host exclusively load generators.
+        let mut client_instances = Vec::new();
+        for region in self.settings.regions.iter().cycle() {
```
mlegner, line 95: Suggestion: Store the iterator in a variable. Then we can continue from the last region in the next loop below.

To have a seamless continuation, I would also reverse the order of the two `if` statements in this loop.

crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -0,0 +1,374 @@
...
+            "sudo apt-get update",
+            "sudo apt-get -y upgrade",
+            "sudo apt-get -y autoremove",
+            // Disable "pending kernel upgrade" message.
+            "sudo apt-get -y remove needrestart",
+            // The following dependencies
+            // * build-essential: prevent the error: [error: linker `cc` not found].
+            // * sysstat - for getting disk stats
+            // * iftop - for getting network stats
+            // * libssl-dev - Required to compile the orchestrator, todo remove this dependency
```
mlegner: Question: Why do we want to remove this dependency? And is there a plan for the removal?

crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -0,0 +1,374 @@
...
+        let context = CommandContext::new().with_execute_from_path(repo_name.into());
+        let all = clients.into_iter().chain(nodes);
+        self.ssh_manager.execute(all, command, context).await?;
+
+        display::done();
+        Ok(())
+    }
+
+    /// Cleanup all instances and optionally delete their log files.
+    pub async fn cleanup(&self, cleanup: bool) -> TestbedResult<()> {
```
mlegner: Nit: The parameter is not very expressive (given that the function is already called `cleanup`). Maybe `delete_logs` would be clearer.

crates/walrus-orchestrator/src/orchestrator.rs
asonnino: Ha yes sorry they will be used in the next PR. I remove them for now to make the PR self-contained and add them back to the next one (where they will be useful).

crates/walrus-orchestrator/src/orchestrator.rs
asonnino: it is tracked here https://github.com/MystenLabs/walrus/issues/221
I will update the comment accordingly

asonnino: I am happy with leaving it as it is but people kept complaining about this dependency in the past (early Sui days)

crates/walrus-orchestrator/src/orchestrator.rs
asonnino: good point

crates/walrus-orchestrator/src/orchestrator.rs
asonnino: not sure I follow. You mean something like this?

```rust
let mut regions = self.settings.regions.iter().cycle();
while  client_instances.len() < self.settings.dedicated_clients {
    let region = regions.next().unwrap();
    if let Some(regional_instances) = instances_by_regions.get_mut(region) {
        if let Some(instance) = regional_instances.pop_front() {
            client_instances.push(instance.clone());
        }
    }
}
```

I guess we want to restart the iterator when selecting the nodes to avoid situations where all
clients are located in one region and all nodes are located in another region (eg, if we only operate
with 2 regions in total).

crates/walrus-orchestrator/src/orchestrator.rs
asonnino: ha no actually they are used to make the generics here work: https://github.com/MystenLabs/walrus/blob/1be51f27f0d98a76b346852df46958c9d608eef5/crates/walrus-orchestrator/src/orchestrator.rs#L139

Is there a better way of doing it?

asonnino: never mind, fixed!

crates/walrus-orchestrator/src/orchestrator.rs
mlegner: Yes, that's what I had in mind.

But you're right, it probably makes sense to start again from the same nodes.

crates/walrus-orchestrator/src/orchestrator.rs
mlegner: > it is tracked here #221 I will update the comment accordingly

Ah, sorry, I missed that. No strong feelings, but let's just keep the issue for now.

PR #253, opened 2024-04-11
refactor: move sliver->shard mapping and sliver verification to core

### LINKED ISSUE #169: Protocol versioning: Include the pair -> shard mapping into the core and version it together with the encoding
- [ ] move `verify_sliver` from `walrus-service/src/client/communication.rs` to `sliver.rs` in `walrus-core`

mlegner:
Closes #169

### DIFF
--- a/crates/walrus-core/src/encoding/slivers.rs
+++ b/crates/walrus-core/src/encoding/slivers.rs
@@ -24,7 +24,6 @@ use crate::{
     ensure,
     merkle::{MerkleProof, MerkleTree, Node, DIGEST_LEN},
     metadata::{SliverPairMetadata, VerifiedBlobMetadataWithId},
-    ShardIndex,
     SliverIndex,
     SliverPairIndex,
 };
@@ -85,16 +84,18 @@ impl<T: EncodingAxis> Sliver<T> {
         self
     }
 
-    /// Checks that the provided sliver matches the corresponding hash in the metadata.
+    /// Checks that the provided sliver is authenticated by the metadata.
+    ///
+    /// The checks include verifying that the sliver has the correct number of symbols and symbol
+    /// size, and that the hash in the metadata matches the Merkle root over the sliver's symbols.
     pub fn verify(
         &self,
         encoding_config: &EncodingConfig,
         metadata: &VerifiedBlobMetadataWithId,
-        shard_idx: ShardIndex,
     ) -> Result<(), SliverVerificationError> {
         ensure!(
-            shard_idx.as_usize() < metadata.metadata().hashes.len(),
-            SliverVerificationError::ShardIndexTooLarge
+            self.index.as_usize() < metadata.metadata().hashes.len(),
+            SliverVerificationError::IndexTooLarge
         );
         ensure!(
             self.symbols.len()
@@ -120,8 +121,8 @@ impl<T: EncodingAxis> Sliver<T> {
             .metadata()
             .hashes
             .get(
-                shard_idx
-                    .to_pair_index(encoding_config.n_shards, metadata.blob_id())
+                self.index
+                    .to_pair_index::<T>(encoding_config.n_shards)
                     .as_usize(),
             )
             .expect("n_shards and shard_index < n_shards are checked above");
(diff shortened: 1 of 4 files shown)


crates/walrus-core/src/encoding/mapping.rs
```diff
@@ -94,41 +95,29 @@ fn is_rotation(pairs: &[SliverPair]) -> bool {
...
-/// integer, and then used to compute the offset for the sliver pair index.
-pub fn shard_index_for_pair(
-    pair_idx: SliverPairIndex,
-    n_shards: NonZeroU16,
-    blob_id: &BlobId,
-) -> ShardIndex {
-    ((pair_idx.as_usize() + rotation_offset(n_shards, blob_id)) % usize::from(n_shards.get()))
-        .try_into()
-        .expect("definitely fits into a u16 because `n_shards` is a u16")
+impl SliverPairIndex {
```
giac-mysten, line 98: I like that the impl is here, instead of where the structs are defined!

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -82,6 +85,53 @@ impl<T: EncodingAxis> Sliver<T> {
         self
     }
 
+    /// Checks that the provided sliver matches the corresponding hash in the metadata.
```
giac-mysten: Nit: this is a problem of my original comment 😇 
```suggestion
    /// Checks that the provided sliver is authenticated by the metadata.
    ///
    /// The checks include verifying that the sliver has the correct number of symbols and symbol
    /// size, and that the hash in the metadata matches the Merkle root over the sliver's symbols.
```

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -82,6 +85,53 @@ impl<T: EncodingAxis> Sliver<T> {
...
+    pub fn verify(
```
giac-mysten, line 91: Question: this could also be on `VerifiedBlobMetadataWithId`. Any reason why it's nicer on `Sliver`?

crates/walrus-service/src/client/communication.rs
```diff
@@ -153,7 +151,7 @@ impl<'a> NodeCommunication<'a> {
     {
         let inner = || async {
             let sliver = self.retrieve_sliver::<T>(metadata, shard_idx).await?;
-            self.verify_sliver(metadata, &sliver, shard_idx)?;
+            sliver.verify(self.encoding_config, metadata, shard_idx)?;
```
giac-mysten: related to the simplification
```suggestion
            sliver.verify(self.encoding_config, metadata)?;
```

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -82,6 +85,53 @@ impl<T: EncodingAxis> Sliver<T> {
...
+        &self,
+        encoding_config: &EncodingConfig,
+        metadata: &VerifiedBlobMetadataWithId,
+        shard_idx: ShardIndex,
```
giac-mysten: I realized that this can be simplified
```suggestion
```

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -82,6 +85,53 @@ impl<T: EncodingAxis> Sliver<T> {
...
+        ensure!(
+            self.symbols.symbol_size() == symbol_size_from_metadata,
+            SliverVerificationError::SymbolSizeMismatch
+        );
+        let pair_metadata = metadata
+            .metadata()
+            .hashes
+            .get(
+                shard_idx
+                    .to_pair_index(encoding_config.n_shards, metadata.blob_id())
```
giac-mysten: related to the simplification
```suggestion
                self.index
                    .to_pair_index::<T>(encoding_config.n_shards)
```

crates/walrus-core/src/encoding/slivers.rs
mlegner: Hmm... In my mind you're verifying the sliver, so having a method on `Sliver` that gets some additional input necessary for the verification makes sense to me.

crates/walrus-core/src/encoding/slivers.rs
mlegner: ~~I don't understand the changes to the `to_pair_index` invocation. But the replacement of the `shard_idx` makes sense.~~

Issue #260, opened 2024-04-12
Implement Event streams for Epoch change events


karlwuest:
Related to #552, #553

PR #261, opened 2024-04-12
refactor: switch to single event stream for blob events

### LINKED ISSUE #240: Consider simplifying ReadClient to return system events as an enum stream.
This is kind of funny, because in the read_client we specifically ask for them separately, and now we merge them. If this is the key way we use them, we should probably ask for all of them at once. Lets have a TODO / issue somewhere.

_Originally posted by @gdanezis in https://github.com/MystenLabs/walrus/pull/225#discussion_r1557285349_

karlwuest:
closes #240 
To keep the PR small, this only replaces the current event streams, support for other events (including moving the events to a separate `events` module in the move contracts) is part of #260.

### DIFF
--- a/crates/walrus-service/src/storage.rs
+++ b/crates/walrus-service/src/storage.rs
@@ -58,18 +58,15 @@ pub struct Storage {
     database: Arc<RocksDB>,
     metadata: DBMap<BlobId, BlobMetadata>,
     blob_info: DBMap<BlobId, BlobInfo>,
-    event_cursor: DBMap<EventCursorKey, EventID>,
+    event_cursor: DBMap<String, EventID>,
     shards: HashMap<ShardIndex, ShardStorage>,
 }
-
-/// Type with a single value to be used as key for the event cursor in the DB
-#[derive(Debug, Serialize, Deserialize)]
-struct EventCursorKey;
 
 impl Storage {
     const METADATA_COLUMN_FAMILY_NAME: &'static str = "metadata";
     const BLOBINFO_COLUMN_FAMILY_NAME: &'static str = "blob_info";
     const EVENT_CURSOR_COLUMN_FAMILY_NAME: &'static str = "event_cursor";
+    const EVENT_CURSOR_KEY: &'static str = "event_cursor";
 
     /// Opens the storage database located at the specified path, creating the database if absent.
     pub fn open(path: &Path, metrics_config: MetricConf) -> Result<Self, anyhow::Error> {
@@ -176,7 +173,7 @@ impl Storage {
 
     /// Get the event cursor for `event_type`
     pub fn get_event_cursor(&self) -> Result<Option<EventID>, TypedStoreError> {
-        self.event_cursor.get(&EventCursorKey)
+        self.event_cursor.get(&Self::EVENT_CURSOR_KEY.to_string())
     }
 
     /// Update the blob info for a blob based on the `BlobEvent`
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/storage.rs
```diff
@@ -58,10 +58,14 @@ pub struct Storage {
...
     metadata: DBMap<BlobId, BlobMetadata>,
     blob_info: DBMap<BlobId, BlobInfo>,
-    event_cursor: DBMap<EventType, EventID>,
+    event_cursor: DBMap<EventCursorKey, EventID>,
     shards: HashMap<ShardIndex, ShardStorage>,
 }
 
+/// Type with a single value to be used as key for the event cursor in the DB
+#[derive(Debug, Serialize, Deserialize)]
+struct EventCursorKey;
```
jpcsmith: I think it would be better to just use a const string or array instead of a struct, as it's unclear what its treatment will be in the database.

crates/walrus-service/src/system_events.rs
```diff
@@ -8,30 +8,21 @@ use std::time::Duration;
...
-}
-
 /// A provider of system events to a storage node.
 #[async_trait]
 pub trait SystemEventProvider: std::fmt::Debug + Sync + Send {
     /// Return a new stream over [`BlobEvent`]s starting from those specified by `from`.
     async fn events(
         &self,
-        cursors: SystemEventCursorSet,
+        cursors: Option<EventID>,
```
jpcsmith: Nit: use singular

crates/walrus-sui/src/test_utils/mock_clients.rs
```diff
@@ -31,43 +32,31 @@ use crate::{
...
         // A channel capacity of 1024 should be enough capacity to not feel backpressure for testing
-        let (registered_events_channel, _) = broadcast::channel(1024);
-        let (certified_events_channel, _) = broadcast::channel(1024);
+        let (events_channel, _) = broadcast::channel(1024);
         Self {
-            registered_events: Arc::new(Mutex::new(registered_events)),
-            certified_events: Arc::new(Mutex::new(certified_events)),
-            registered_events_channel,
-            certified_events_channel,
+            events: Arc::new(Mutex::new(events)),
```
jpcsmith, line 58: Question: Would just converting the vec to a stream work here?

karlwuest: I don't think so, or at least I don't think we would gain anything from it. 
The purpose of this vec is to make sure that two callers of the `blob_events` function receive the same sequence of events, so we need to be able to add more entries and we always have to clone it when returning a stream. Maybe I'm missing something, but I think if we store a stream here that would get much more complicated and not more efficient.

jpcsmith: No, i meant store a vec, and then in blob_events create and return the stream there.

jpcsmith: Actually, you probably interpreted correct, but the thought was regarding removing the channel logic that's in this.

karlwuest: Ah, that works if we only require a ReadClient, but not if we have a MockContractClient, because in that case, we want the returned stream to also provide events that are added after `blob_events` is called.

For example:
```
let client = MockContractClient::default();
let events = client.read_client().blob_events(...).await?;
client.register_blob(...).await?;
let register_event = events.next().await; // This should return the event emitted by the line above
```

jpcsmith: Cool, I follow now. Thanks for clarifying, then all looks good.

PR #267, opened 2024-04-12
ci: improve CI workflows

mlegner:
- Remove weekly schedule; check dependencies on `main` branch instead.
- Create separate workflow for test coverage; can be triggered by labeling a PR with the `report_coverage` label. This job is no prerequisite for merging a PR.
- Enable incremental compilation for standard tests as we have caching.
- Include E2E tests in standard tests job to keep pipeline simple; may have to be separated out in the future.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -8,9 +8,6 @@ on:
   push:
     branches:
       - main
-  # Run workflow on the main branch every Sunday.
-  schedule:
-    - cron: '14 3 * * 0'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
@@ -19,18 +16,12 @@ concurrency:
 env:
   CARGO_TERM_COLOR: always
   CLICOLOR_FORCE: 1
-  # Disable incremental compilation.
-  #
-  # Incremental compilation is useful as part of an edit-build-test-edit cycle,
-  # as it lets the compiler avoid recompiling code that hasn't changed. However,
-  # on CI, we're not making small edits; we're almost always building the entire
-  # project from scratch. Thus, incremental compilation on CI actually
-  # introduces *additional* overhead to support making future builds
-  # faster...but no future builds will ever occur in any given CI environment.
-  #
-  # See https://matklad.github.io/2021/09/04/fast-rust-builds.html#ci-workflow
-  # for details.
-  CARGO_INCREMENTAL: 0
+  # Incremental compilation is useful as part of an edit-build-test-edit cycle, as it lets the
+  # compiler avoid recompiling code that hasn't changed. The setting does not improve the current
+  # compilation but instead saves additional information to speed up future compilations (see
+  # https://doc.rust-lang.org/cargo/reference/profiles.html#incremental). Thus, this is only useful
+  # in CI if the result is cached, which we only do on the `main` branch.
+  CARGO_INCREMENTAL: ${{ github.ref == 'refs/heads/main' && '1' || '0' }}
   # Allow more retries for network requests in cargo (downloading crates) and
   # rustup (installing toolchains). This should help to reduce flaky CI failures
   # from transient network timeouts or other issues.
@@ -69,11 +60,6 @@ jobs:
               - 'rust-toolchain'
               - 'deny.toml'
               - '.github/workflows/code.yml'
-            relevantForE2eTests:
-              - 'crates/walrus-sui/**'
-              - 'crates/walrus-service/**'
-              - 'contracts/**'
-              - '.github/workflows/code.yml'
 
   dependencies:
     name: Check dependencies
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,79 @@
+name: Test Coverage
+
+on:
+  # Run workflow on every PR.
+  pull_request:
+    types:
+      - labeled
```
mlegner, line 7: Adding this as a trigger ensures that the workflow gets triggered when adding the `report_coverage` label. Unfortunately, this also creates an (empty) workflow whenever _any_ label is added (see https://github.com/MystenLabs/walrus/pull/267#issuecomment-2051868971).

The alternative would be to remove this trigger. However, in that case, the workflow would only be triggered if (i) the label is assigned from the beginning, (ii) the PR is closed+reopened, or (iii) a new commit is pushed.

I tend towards the current version (as we don't add that many labels anyway), but don't have a strong opinion.

.github/workflows/code.yml
```diff
@@ -8,9 +8,6 @@ on:
   push:
     branches:
       - main
-  # Run workflow on the main branch every Sunday.
-  schedule:
-    - cron: '14 3 * * 0'
```
asonnino, line 13: Do we use another flow to detect vulnerability reports?

.github/workflows/code.yml
```diff
@@ -19,18 +16,6 @@ concurrency:
...
-  # Incremental compilation is useful as part of an edit-build-test-edit cycle,
-  # as it lets the compiler avoid recompiling code that hasn't changed. However,
-  # on CI, we're not making small edits; we're almost always building the entire
-  # project from scratch. Thus, incremental compilation on CI actually
-  # introduces *additional* overhead to support making future builds
-  # faster...but no future builds will ever occur in any given CI environment.
-  #
-  # See https://matklad.github.io/2021/09/04/fast-rust-builds.html#ci-workflow
-  # for details.
-  CARGO_INCREMENTAL: 0
```
asonnino, line 33: Why do we remove it?

.github/workflows/code.yml
```diff
@@ -87,76 +67,18 @@ jobs:
           # do not check advisories on PRs to prevent sudden failure due to new announcement
           command: check bans licenses sources
 
-  dependencies-schedule:
+  dependencies-main:
     name: Check dependencies (including vulnerabilities)
-    needs: diff
-    if: ${{ github.event_name == 'schedule' }}
+    if: ${{ github.ref == 'refs/heads/main' }}
```
asonnino, line 78: Does it mean that we trigger this job upon every change on main?

.github/workflows/code.yml
```diff
@@ -191,10 +113,25 @@ jobs:
...
+    needs: diff
+    if: ${{ needs.diff.outputs.isRust == 'true' }}
+    runs-on: ubuntu-ghcloud
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2.7.3
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
+      - name: Run tests
+        run: cargo test -- --include-ignored
```
asonnino, line 135: Do we run ignored tests on purpose, right?

.github/workflows/code.yml
```diff
@@ -191,10 +113,25 @@ jobs:
...
+  test:
+    name: Test Rust code
+    needs: diff
+    if: ${{ needs.diff.outputs.isRust == 'true' }}
+    runs-on: ubuntu-ghcloud
+    steps:
+      - uses: actions/checkout@v4
+      - uses: Swatinem/rust-cache@v2.7.3
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
```
asonnino, line 133: I understand that this means we want to cache the output if the job has been triggered because of a change on main (and not on other branches), is that right?

.github/workflows/coverage.yml
asonnino: no strong opinions. How about allowing this job to be triggered manually (as we used to have for the "update.yml" job?

.github/workflows/code.yml
mlegner: To help with the caching. But now that I think about it some more, we should probably disable it generally and enable on `main`.

.github/workflows/coverage.yml
giac-mysten: manual trigger + new commit trigger could work well

.github/workflows/code.yml
mlegner: Yes, this is now included in every run on `main` (job `dependencies-main`).

.github/workflows/code.yml
mlegner: That's right.

.github/workflows/code.yml
mlegner: Yes, this is intentional. However, if that gets too long (currently doesn't look like it), we can separate the E2E tests.

.github/workflows/code.yml
mlegner: Yes. This is due to the fact that caches from the default branch are available on PRs (see [here](https://github.com/Swatinem/rust-cache/tree/v2.7.3/?tab=readme-ov-file#cache-limits-and-control)). Caching on PRs would be useful for that specific PR, but we would quickly exceed the 10GB limit of caches on GitHub.

.github/workflows/code.yml
asonnino: just to be sure, this is not triggered through a schedule but rather upon actions on our side?

.github/workflows/coverage.yml
mlegner: Using a manual trigger was also my first idea. Unfortunately there doesn't seem to be a straight-forward option to have a manual trigger associated with a PR.

.github/workflows/code.yml
mlegner: That's right. Whenever a PR is merged, this adds a commit to `main`, which triggers the job.

Another option would be to just add a weekly scheduled workflow for both the vulnerability reports and the test coverage (as suggested by @giac-mysten out-of-band).

PR #274, opened 2024-04-15
ci: reorganize workflows

mlegner:
- Add new weekly workflow to report coverage and check dependencies.
- Include coverage report for PRs in standard workflow (run only on
  labeled PRs).
- Replace editorconfig-checker action by manual install.

This avoids lots of emails for canceled workflows.

Follow-up on #267.

### DIFF
--- a/crates/walrus-orchestrator/src/settings.rs
+++ b/crates/walrus-orchestrator/src/settings.rs
@@ -187,7 +187,6 @@ impl Settings {
             .path_segments()
             .expect("Url should already be checked when loading settings")
             .collect::<Vec<_>>()[1]
-            .to_string()
             .split('.')
             .next()
             .unwrap()


.github/workflows/lint.yml
```diff
@@ -26,8 +26,8 @@ jobs:
     name: Check editorconfig
     steps:
       - uses: actions/checkout@v4
-      - uses: editorconfig-checker/action-editorconfig-checker@v2.0.0
-      - run: editorconfig-checker
+      - run: pip install editorconfig-checker=="2.7.3"
+      - run: ec
```
giac-mysten, line 30: Question: could you remind me why we prefer running the tool rather than the action?

mlegner: The action is just there to install the tool. And it's a bit outdated, so I changed this to a manual install.

PR #277, opened 2024-04-15
fix: return correct quorum threshold from Committee

karlwuest:
The previous computation erroneously rounded down in cases where the `total_weight` is not exactly `3f + 1`, e.g., for `total_weight = 3`, the returned threshold was `2`.

### DIFF
--- a/crates/walrus-sui/src/types.rs
+++ b/crates/walrus-sui/src/types.rs
@@ -274,18 +274,18 @@ pub struct Committee {
 }
 
 impl Committee {
-    /// Checks if the number is large enough to reach a quorum (`total_weight - f`)where `f`
+    /// Checks if the number is large enough to reach a quorum (`total_weight - f`) where `f`
     /// is the maximum number of faulty nodes, given `total_weight`.
     #[inline]
     pub fn is_quorum(&self, num: usize) -> bool {
-        num >= self.quorum_threshold()
+        num >= self.min_n_correct_nodes()
     }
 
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/src/types.rs
```diff
@@ -274,16 +274,20 @@ pub struct Committee {
 }
 
 impl Committee {
-    /// Checks if the number is large enough to reach a quorum (2f+1).
+    /// Checks if the number is large enough to reach a quorum (`total_weight - f`)where `f`
```
mlegner: Nit:
```suggestion
    /// Checks if the number is large enough to reach a quorum (`total_weight - f`) where `f`
```

crates/walrus-sui/src/types.rs
```diff
@@ -274,16 +274,20 @@ pub struct Committee {
...
+    /// is the maximum number of faulty nodes, given `total_weight`.
     #[inline]
     pub fn is_quorum(&self, num: usize) -> bool {
         num >= self.quorum_threshold()
     }
 
-    /// Returns the quorum threshold (2f+1).
+    /// Returns the quorum threshold, which is calculated as `total_weight - f` where `f`
+    /// is the maximum number of faulty nodes, given `total_weight`. If `total_weight = 3f+1`
```
mlegner: Nit:
```suggestion
    /// is the maximum number of faulty nodes, given `total_weight`. If `total_weight == 3f+1`
```

PR #280, opened 2024-04-15
feat(service): check the weighted futures threshold with function

giac-mysten:
Changes to `WeightedFutures` such that the limit weight is tested via a provided function, most importantly `is_quorum`.

Relates to #277

### DIFF
--- a/crates/walrus-sui/src/types.rs
+++ b/crates/walrus-sui/src/types.rs
@@ -294,9 +294,9 @@ impl Committee {
     /// is the maximum number of faulty shards, given `total_weight`.
     #[inline]
     pub fn is_quorum(&self, num: usize) -> bool {
-        // This expression checks that `num >= 2g + 1`, assuming `total_weight = 3g + 1` with a
-        // possibly fractionary `g` and `f <= g < f+1`.  This in turn always implies that
-        // `num >= total_weight - f`.
+        // Given `total_weight == 3g + 1` with a possibly fractionary `g`, this expression checks
+        // that `num >= 2g + 1`. This in turn always implies that `num >= total_weight - f`, where
+        // `f == floor(g)` is the maximum number of faulty shards.
         3 * num >= 2 * self.total_weight.get() as usize + 1
     }
 


crates/walrus-service/src/client/utils.rs
```diff
@@ -122,14 +122,18 @@ where
...
-    /// successfully without error.
+    /// Executes the futures until the total weight of _successful_ futures gets the `threshold`
+    /// function to return `true`.
     ///
     /// `n_concurrent` is the maximum number of futures that are awaited at any one time to produce
     /// results.
-    pub async fn execute_weight(&mut self, threshold: usize, n_concurrent: usize) {
+    pub async fn execute_weight(
+        &mut self,
+        threshold: &impl Fn(usize) -> bool,
```
karlwuest, line 132: Question: Why do you use a reference to the closure here and not a closure directly? 
Just curious, since I haven't seen this yet.

crates/walrus-sui/src/types.rs
```diff
@@ -278,16 +278,7 @@ impl Committee {
     /// is the maximum number of faulty nodes, given `total_weight`.
```
mlegner: Nit: I didn't notice this in @karlwuest's PR, but this should actually say "shards" instead of "nodes".

crates/walrus-service/src/client/utils.rs
giac-mysten: the problem I had with a simple `impl Fn` is that, in that case, the `threshold` function gets moved in the `while let` loop, because it is passed on to `next_threshold` at every iteration.

this was the simplest solution, with the added benefit that afaik it does not require moving the state captured by the closure.

Do you see a better way of achieving the same?

mlegner: I think the current approach with a reference makes a lot of sense. 👍

crates/walrus-sui/src/types.rs
```diff
@@ -275,19 +275,13 @@ pub struct Committee {
...
-    /// Returns the quorum threshold, which is calculated as `total_weight - f` where `f`
-    /// is the maximum number of faulty nodes, given `total_weight`. If `total_weight == 3f+1`
-    /// for some `f`, this is exactly `2f+1`, otherwise it will be slightly (<=2) larger.
-    #[inline]
-    pub fn min_n_correct_nodes(&self) -> usize {
-        let total_weight = self.total_weight as usize;
-        total_weight - (total_weight - 1) / 3
+        // This expression checks that `num >= 2g + 1`, assuming `total_weight = 3g + 1` with a
+        // possibly fractionary `g` and `f <= g < f+1`.  This in turn always implies that
+        // `num >= total_weight - f`.
```
mlegner: That sounds correct to me, but I found it a bit hard to understand. How about this version?
```suggestion
        // Given `total_weight == 3g + 1` with a possibly fractionary `g`, this expression checks
        // that `num >= 2g + 1`. This in turn always implies that `num >= total_weight - f`, where
        // `f == floor(g)` is the maximum number of faulty shards.
```

crates/walrus-service/src/client/utils.rs
karlwuest: >Do you see a better way of achieving the same?

No, I think that's fine. As I said, I was just curious.

crates/walrus-sui/src/types.rs
karlwuest: After thinking about this more, I would prefer to make the computation and check explicit to improve the readability. We can then make the check similarly explicit in the smart contract. 
I.e. something like:
```
let f = (self.total_weight as usize - 1) / 3;
num >= self.total_weight as usize - f;
```

mlegner: I can do this as part of #285.

PR #281, opened 2024-04-15
refactor: use `NonZeroU16` for number of shards in `Committee`

### LINKED ISSUE #242: Define reasonable default value for `Committee`
Currently, there is a derived `Default` implementation on `walrus_sui::types::Committee`. This is not an allowed instance, because we generally assume a non-zero number of shards.

We should change the type of `total_weight` to `NonZeroU16` and create a custom default implementation.

cc @karlwuest

karlwuest:
closes #242 

- Changes `Committee` to use `NonZeroU16` for the number of shards
- Ensures in the contracts that we cannot have a committee without any shards
- Removes `Default` for `Committee`, `MockSuiReadClient`, and `MockContractClient` since there is no longer a clear default value for the committee.

### DIFF
--- a/crates/walrus-core/src/encoding/config.rs
+++ b/crates/walrus-core/src/encoding/config.rs
@@ -70,35 +70,13 @@ impl EncodingConfig {
     ///
     /// [rfc6330s5.6]: https://datatracker.ietf.org/doc/html/rfc6330#section-5.6
     pub fn new(source_symbols_primary: u16, source_symbols_secondary: u16, n_shards: u16) -> Self {
-        assert!(
-            source_symbols_primary < MAX_SOURCE_SYMBOLS_PER_BLOCK
-                && source_symbols_secondary < MAX_SOURCE_SYMBOLS_PER_BLOCK,
-            "the number of source symbols can be at most `MAX_SOURCE_SYMBOLS_PER_BLOCK`"
-        );
-        assert!(
-            3 * source_symbols_secondary < 2 * n_shards,
-            "the secondary encoding can be at most a 2/3 encoding"
-        );
-        assert!(
-            3 * source_symbols_primary < n_shards,
-            "the primary encoding can be at most a 1/3 encoding"
-        );
-
         let source_symbols_primary = NonZeroU16::new(source_symbols_primary)
             .expect("the number of source symbols must not be 0");
         let source_symbols_secondary = NonZeroU16::new(source_symbols_secondary)
             .expect("the number of source symbols must not be 0");
         let n_shards = NonZeroU16::new(n_shards).expect("implied by previous checks");
 
-        Self {
-            source_symbols_primary,
-            source_symbols_secondary,
-            n_shards,
-            encoding_plan_primary: SourceBlockEncodingPlan::generate(source_symbols_primary.get()),
-            encoding_plan_secondary: SourceBlockEncodingPlan::generate(
-                source_symbols_secondary.get(),
-            ),
-        }
+        Self::new_from_nonzero(source_symbols_primary, source_symbols_secondary, n_shards)
     }
 
     /// See [Self::new].
@@ -107,18 +85,19 @@ impl EncodingConfig {
         source_symbols_secondary: NonZeroU16,
         n_shards: NonZeroU16,
     ) -> Self {
+        let f = (n_shards.get() - 1) / 3;
         assert!(
             source_symbols_primary.get() < MAX_SOURCE_SYMBOLS_PER_BLOCK
                 && source_symbols_secondary.get() < MAX_SOURCE_SYMBOLS_PER_BLOCK,
             "the number of source symbols can be at most `MAX_SOURCE_SYMBOLS_PER_BLOCK`"
         );
         assert!(
-            3 * (source_symbols_secondary.get() - 1) <= 2 * (n_shards.get() - 1),
-            "the secondary encoding can be at most a 2f+1 encoding"
+            source_symbols_secondary.get() <= n_shards.get() - f,
+            "the secondary encoding can be at most a n-f encoding"
         );
         assert!(
-            3 * (source_symbols_primary.get() - 1) <= (n_shards.get() - 1),
-            "the primary encoding can be at most an f+1 encoding"
+            source_symbols_primary.get() <= n_shards.get() - 2 * f,
+            "the primary encoding can be at most an n-2f encoding"
         );
 
         Self {


crates/walrus-service/src/client/config.rs
```diff
@@ -51,7 +51,7 @@ impl LocalCommitteeConfig {
         EncodingConfig::new(
             self.source_symbols_primary.get(),
             self.source_symbols_secondary.get(),
-            self.committee.total_weight,
+            self.committee.n_shards().get(),
         )
```
mlegner: Remark: We should probably add a separate `EncodingConfig::new_from_nonzero` function to avoid back-and-forth conversions. Can be done separately.

crates/walrus-sui/src/types.rs
```diff
@@ -263,31 +263,48 @@ impl AssociatedContractStruct for StorageNode {
...
     pub fn is_quorum(&self, num: usize) -> bool {
-        num >= self.min_n_correct_nodes()
+        num >= self.min_n_correct_nodes().get() as usize
     }
 
-    /// Returns the quorum threshold, which is calculated as `total_weight - f` where `f`
-    /// is the maximum number of faulty nodes, given `total_weight`. If `total_weight == 3f+1`
-    /// for some `f`, this is exactly `2f+1`, otherwise it will be slightly (<=2) larger.
+    /// Returns the minimum number of correct nodes, which is calculated as
+    /// (`n_shards - f`) where `f` is the maximum number of faulty nodes,
```
mlegner: Nit (missed this before):
```suggestion
    /// (`n_shards - f`) where `f` is the maximum number of faulty shards,
```

crates/walrus-sui/src/types.rs
```diff
@@ -263,31 +263,48 @@ impl AssociatedContractStruct for StorageNode {
...
+    /// given [`n_shards`][Self::n_shards]. If `n_shards == 3f+1` for some `f`, this is
+    /// exactly `2f+1`, otherwise it will be slightly (<=2) larger.
     #[inline]
-    pub fn min_n_correct_nodes(&self) -> usize {
-        let total_weight = self.total_weight as usize;
-        total_weight - (total_weight - 1) / 3
+    pub fn min_n_correct_nodes(&self) -> NonZeroU16 {
+        // unwrap safe, because the right side of the subtraction is always strictly smaller
+        NonZeroU16::new(self.total_weight.get() - (self.total_weight.get() - 1) / 3).unwrap()
```
mlegner: Suggestion: Can we move the comment to the body of an `expect`?

crates/walrus-sui/src/types.rs
```diff
@@ -263,31 +263,48 @@ impl AssociatedContractStruct for StorageNode {
...
+        let total_weight = NonZeroU16::new(total_weight).ok_or(InvalidCommittee::EmptyCommittee)?;
+        Ok(Self {
+            members,
+            epoch,
+            total_weight,
+        })
+    }
+
+    /// Checks if the number is large enough to reach a quorum (`n_shards - f`) where `f`
+    /// is the maximum number of faulty nodes, given [`n_shards`][Self::n_shards].
```
mlegner: Nit (missed this before):
```suggestion
    /// is the maximum number of faulty shards, given [`n_shards`][Self::n_shards].
```

crates/walrus-sui/src/test_utils/mock_clients.rs
```diff
@@ -131,15 +126,6 @@ pub struct MockContractClient {
...
-    pub fn new(current_epoch: Epoch) -> Self {
-        let read_client = MockSuiReadClient::default();
-        Self {
-            read_client,
-            current_epoch,
-        }
-    }
-
     /// Construct a [`MockContractClient`] with a provided [`MockSuiReadClient`].
     pub fn new_with_read_client(current_epoch: Epoch, read_client: MockSuiReadClient) -> Self {
```
mlegner: Nit: Given that you deleted the `new` function, can we call this one `new`?

crates/walrus-service/src/client/config.rs
karlwuest: I agree. I can add it to this one or we can do it separately. 
Related question: I noticed the following checks in `EncodingConfig`. Is there a reason for not allowing an encoding where e.g. `source_symbols_secondary == 2f+1`? 
Should we update these checks as well, especially since with #205 we can have a safety limit of 0?

```
  assert!(
      3 * source_symbols_secondary < 2 * n_shards,
      "the secondary encoding can be at most a 2/3 encoding"
  );
  assert!(
      3 * source_symbols_primary < n_shards,
      "the primary encoding can be at most a 1/3 encoding"
  );
```

karlwuest: Done in ba2466b56fa038dc72f5ccbff5b24bb27faffb3d. Please have a quick look (I also fixed the asserts to be compatible with my understanding) and resolve this conversation if you're happy with it.

crates/walrus-core/src/encoding/config.rs
```diff
@@ -101,6 +101,37 @@ impl EncodingConfig {
...
+            "the number of source symbols can be at most `MAX_SOURCE_SYMBOLS_PER_BLOCK`"
+        );
+        assert!(
+            3 * (source_symbols_secondary.get() - 1) <= 2 * (n_shards.get() - 1),
+            "the secondary encoding can be at most a 2f+1 encoding"
+        );
+        assert!(
+            3 * (source_symbols_primary.get() - 1) <= (n_shards.get() - 1),
+            "the primary encoding can be at most an f+1 encoding"
+        );
```
mlegner: Important: These checks should be the same between this function and `new`.

Ideally, the `new` function would just do the non-zero checks and then call this one.

crates/walrus-core/src/encoding/config.rs
```diff
@@ -101,6 +101,37 @@ impl EncodingConfig {
...
+    ) -> Self {
+        assert!(
+            source_symbols_primary.get() < MAX_SOURCE_SYMBOLS_PER_BLOCK
+                && source_symbols_secondary.get() < MAX_SOURCE_SYMBOLS_PER_BLOCK,
+            "the number of source symbols can be at most `MAX_SOURCE_SYMBOLS_PER_BLOCK`"
+        );
+        assert!(
+            3 * (source_symbols_secondary.get() - 1) <= 2 * (n_shards.get() - 1),
+            "the secondary encoding can be at most a 2f+1 encoding"
+        );
```
mlegner: Remark: Here we have the same question again that we discussed yesterday: Should we take `n-f` or `2f+1`. As we settled on `n-f` for the quorum, we should probably do the same here.

Suggestion: Consolidate all BFT computations in a separate PR.

crates/walrus-service/src/client/config.rs
mlegner: Thanks for the changes, they look good. We just need to make the two `new` functions consistent.

crates/walrus-core/src/encoding/config.rs
karlwuest: Ahh, good catch. I wanted to have new call this one, but for some reason I forgot to actually change it before committing 🤦‍♂️

crates/walrus-core/src/encoding/config.rs
karlwuest: I changed it in this PR now, since the previous PR already changed the computation in `min_n_correct_shards`, so it will be consistent after this PR.

PR #282, opened 2024-04-15
feat: Initial root hash logic

gdanezis:
- Check that the BlobID is the hash of the meta-data
- Modify the move contracts
- Modify the rust sdk / clients
- Add tests / fix tests

### DIFF
--- a/crates/walrus-sui/tests/test_walrus_sui.rs
+++ b/crates/walrus-sui/tests/test_walrus_sui.rs
@@ -43,9 +43,6 @@ async fn test_register_certify_blob() -> anyhow::Result<()> {
     ];
 
     let blob_id = BlobId::from_metadata(Node::from(root_hash), EncodingType::RedStuff, size);
-    println!("root_hash: {:?}", root_hash);
-    println!("blob_id: {:?}", blob_id.0);
-
     let blob_obj = walrus_client
         .register_blob(
             &storage_resource,
@@ -77,6 +74,7 @@ async fn test_register_certify_blob() -> anyhow::Result<()> {
     let certificate = get_default_blob_certificate(blob_id, 0);
 
     // Values printed here should match move test `test_blob_certify_single_function`
+    // Note: we keep these commented in case we ever need to recompute them to update tests.
     //
     // println!("certificate signature: {:?}", certificate.signature.as_bytes());
     // let bytes : Vec<u8> = certificate.confirmation.clone().into();


crates/walrus-sui/tests/test_walrus_sui.rs
```diff
@@ -35,14 +35,25 @@ async fn test_register_blob() -> anyhow::Result<()> {
...
         1, 2, 3, 4, 5, 6, 7, 8,
         1, 2, 3, 4, 5, 6, 7, 8,
         1, 2, 3, 4, 5, 6, 7, 8,
         1, 2, 3, 4, 5, 6, 7, 8,
-    ]);
+    ];
+
+    let blob_id = BlobId::from_metadata(Node::from(root_hash), EncodingType::RedStuff, size);
+    println!("root_hash: {:?}", root_hash);
+    println!("blob_id: {:?}", blob_id.0);
```
giac-mysten: Nit: @jpcsmith 's suggestion for these kind of cases:
```suggestion
    dbg!(root_hash);
    dbg!(blob_id);
```

crates/walrus-sui/tests/test_walrus_sui.rs
```diff
@@ -63,9 +74,15 @@ async fn test_register_blob() -> anyhow::Result<()> {
...
-    let blob_obj = walrus_client
-        .certify_blob(&blob_obj, &get_default_blob_certificate(blob_id, 0))
-        .await?;
+    let certificate = get_default_blob_certificate(blob_id, 0);
+
+    // Values printed here should match move test `test_blob_certify_single_function`
+    //
+    // println!("certificate signature: {:?}", certificate.signature.as_bytes());
+    // let bytes : Vec<u8> = certificate.confirmation.clone().into();
+    // println!("certificate message: {:?}", bytes);
```
giac-mysten: Nit: remove if not needed?

contracts/blob_store/sources/blob.move
```diff
@@ -103,8 +127,11 @@ module blob_store::blob {
         assert!(stored_epoch < end_epoch(&storage), ERROR_RESOURCE_BOUNDS);
         assert!(size <= storage_size(&storage), ERROR_RESOURCE_SIZE);
 
-        // TODO(#42): cryptographically verify that the Blob ID authenticates
+        // Cryptographically verify that the Blob ID authenticates
         // both the size and fe_type.
+        assert!(derive_blob_id(root_hash, erasure_code_type, size) == blob_id,
+            ERROR_INVALID_BLOB_ID);
```
giac-mysten, line 133: General question: the advantage of having this check in the contract is that, for a comparatively low increase in contract complexity, we can avoid registering invalid blob IDs. This in turn reduces the load on the storage nodes, that don't need to receive metadata to just discard it as invalid. Correct? Is there any other advantage I'm missing?

gdanezis: I think that the advantage is that for a single BlobID there is a 1:1 matching with its size, encoding and contents, that is known and checked at the time of registration. So there cannot be another instance of the same Blob ID with other params ever.

crates/walrus-sui/tests/test_walrus_sui.rs
gdanezis: I will actually remove these.

crates/walrus-sui/tests/test_walrus_sui.rs
gdanezis: I want to keep these in case we need to recompute the values in the test

PR #283, opened 2024-04-15
feat: benchmark facility

### LINKED ISSUE #222: Implement deploying (and benchmarking facilities) for the orchestrator
The orchestrator can currently only deploy AWS or Vultr machines, without doing anything with them.

asonnino:
Complete benchmark facility and general Orchestrator code

closes #222

### DIFF
--- a/crates/walrus-orchestrator/src/measurements.rs
+++ b/crates/walrus-orchestrator/src/measurements.rs
@@ -33,7 +33,7 @@ pub struct Measurement {
     /// Total number of finalized transactions
     count: usize,
     /// Sum of the squares of the latencies of all finalized transactions
-    squared_sum: Duration,
+    squared_sum: f64,
 }
 
 impl Measurement {
@@ -78,7 +78,7 @@ impl Measurement {
                 }
                 x if x == M::LATENCY_SQUARED_SUM => {
                     measurement.squared_sum = match sample.value {
-                        prometheus_parse::Value::Counter(value) => Duration::from_secs_f64(value),
+                        prometheus_parse::Value::Counter(value) => value,
                         _ => panic!("Unexpected scraped value"),
                     };
                 }
@@ -115,7 +115,7 @@ impl Measurement {
         let first_term = if self.count == 0 {
             return Duration::from_secs(0);
         } else {
-            self.squared_sum.as_secs_f64() / self.count as f64
+            self.squared_sum / self.count as f64
         };
 
         // Compute `avg^2`.
@@ -287,7 +287,7 @@ mod test {
             buckets: HashMap::new(),
             sum: Duration::from_secs(2),
             count: 100,
-            squared_sum: Duration::from_secs(0),
+            squared_sum: 0.0,
         };
 
         assert_eq!(data.average_latency(), Duration::from_millis(20));
@@ -300,14 +300,11 @@ mod test {
             buckets: HashMap::new(),
             sum: Duration::from_secs(50),
             count: 100,
-            squared_sum: Duration::from_secs(75),
+            squared_sum: 75.0,
         };
 
         // squared_sum / count
-        assert_eq!(
-            data.squared_sum.checked_div(data.count as u32),
-            Some(Duration::from_secs_f64(0.75))
-        );
+        assert_eq!(data.squared_sum / data.count as f64, 0.75);
         // avg^2
         assert_eq!(data.average_latency().as_secs_f64().powf(2.0), 0.25);
         // sqrt( squared_sum / count - avg^2 )
@@ -413,7 +410,7 @@ mod test {
                 assert_eq!(data.sum.as_secs(), 1265);
                 assert_eq!(data.count, 1860);
                 assert_eq!(data.timestamp.as_secs(), 30);
-                assert_eq!(data.squared_sum.as_secs(), 952);
+                assert_eq!(data.squared_sum as u64, 952);
             }
         }
     }


crates/walrus-orchestrator/Cargo.toml
```diff
@@ -25,6 +25,7 @@ ssh2 = "0.9.4"
 thiserror = { workspace = true }
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 walrus-core = { workspace = true }
+prometheus-parse = { git = "https://github.com/asonnino/prometheus-parser.git", rev = "75334db" }
```
asonnino: This no ideal. #284 should fix that

crates/walrus-orchestrator/src/main.rs
```diff
@@ -163,9 +199,64 @@ async fn run<C: ServerProviderClient>(
...
+                .with_retries(settings.ssh_retries);
+
+            let instances = testbed.instances();
+
+            let setup_commands = testbed
+                .setup_commands()
+                .await
+                .wrap_err("Failed to load testbed setup commands")?;
+
+            let protocol_commands = Protocol {};
```
asonnino, line 224: This will become more meaningful when closing #284

crates/walrus-orchestrator/Cargo.toml
asonnino: mmh it seems that the CI doesn't like that neither. What would be the best solution?

asonnino: (Interestingly there are no failed checks in my pre-commit hooks.)

mlegner: > (Interestingly there are no failed checks in my pre-commit hooks.)

That is weird... The hooks should run on all changes to any `Cargo.toml`... 🤔

I've fixed the errors in d089b4122360545d1d613b05e3c63f97248a82a8 and added a TODO.

crates/walrus-orchestrator/src/main.rs
```diff
@@ -59,7 +69,35 @@ pub enum Operation {
         action: TestbedAction,
     },
     /// Deploy nodes and run a benchmark on the specified testbed.
-    Benchmark,
+    Benchmark {
+        /// The committee size to deploy.
+        #[clap(long, value_name = "INT", global = true)]
+        committee: usize,
```
mlegner: Important: When trying to run the binary, I get the following error:
```
$ cargo run --bin orchestrator benchmark -h
    Finished dev [unoptimized + debuginfo] target(s) in 1.20s
     Running `target/debug/orchestrator benchmark`
The application panicked (crashed).
Message:  Command benchmark: Global arguments cannot be required.

        'committee' is marked as both global and required
```

crates/walrus-orchestrator/src/main.rs
```diff
@@ -59,7 +69,35 @@ pub enum Operation {
...
+
+        /// The set of loads to submit to the system (tx/s). Setting this parameter to zero
+        /// will not deploy any benchmark clients. This is useful to boot testbeds designed
+        /// to run with external clients and load generators.
+        #[clap(long, value_name = "[INT]", default_value = "200", global = true)]
+        loads: Vec<usize>,
```
mlegner: Question: What is the meaning of the different values?

crates/walrus-orchestrator/src/logs.rs
```diff
@@ -0,0 +1,64 @@
...
+use crate::display;
+
+/// A simple log analyzer counting the number of errors and panics.
+#[derive(Default)]
+pub struct LogsAnalyzer {
+    /// The number of errors in the nodes' log files.
+    pub node_errors: usize,
+    /// Whether a node panicked.
+    pub node_panic: bool,
+    /// The number of errors int he clients' log files.
```
mlegner: Typo:
```suggestion
    /// The number of errors in the clients' log files.
```

mlegner: Question: The code below and in `orchestrator.rs` suggests that this is the maximum of client errors of any single client. Is that the intention?

crates/walrus-orchestrator/src/logs.rs
```diff
@@ -0,0 +1,64 @@
...
+            if counter.node_panic || counter.client_panic {
+                return counter;
+            } else if counter.client_errors > highest.client_errors
+                || counter.node_errors > highest.node_errors
+            {
+                highest = counter;
+            }
+        }
+        highest
+    }
```
mlegner: Remark: This seems to be less an aggregator function than a `max` function.

Also: The semantics of the comparison are a bit unclear. For example, if one counter has more client errors but the other has more node errors, either one of them can be selected depending on their order in the input vector.

Suggestion: Implement `Ord` for `LogsAnalyzer`, which lets you call `max` on an iterator.

crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -48,9 +60,26 @@ impl<P> Orchestrator<P> {
...
             protocol_commands,
             ssh_manager,
+            skip_testbed_update: false,
+            skip_testbed_configuration: false,
         }
     }
 
+    /// Skip the testbed update.
+    pub fn skip_testbed_update(mut self, skip_testbed_update: bool) -> Self {
+        display::warn("Skipping testbed update! Use with care!");
```
mlegner: This should only be printed if `skip_testbed_update` is `true`. Same below.

crates/walrus-orchestrator/src/orchestrator.rs
```diff
@@ -361,4 +394,214 @@ impl<P: ProtocolCommands + ProtocolMetrics> Orchestrator<P> {
...
+        let start = Instant::now();
+        loop {
+            tokio::select! {
+                // Scrape metrics.
+                now = metrics_interval.tick() => {
+                    let elapsed = now.duration_since(start).as_secs_f64().ceil() as u64;
+                    display::status(format!("{elapsed}s"));
+
+                    let mut instances = metrics_commands.clone();
+                    instances.retain(|(instance, _)| !killed_nodes.contains(instance));
```
mlegner, line 437: Remark: Assuming we only have a few killed instances, this is fine. If we have more, then iterating over the vector repeatedly is not ideal; a hash set for `killed_nodes` would be better in that case.

crates/walrus-orchestrator/src/measurements.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Duration since the beginning of the benchmark.
+    timestamp: Duration,
+    /// Latency buckets.
+    buckets: HashMap<BucketId, usize>,
+    /// Sum of the latencies of all finalized transactions.
+    sum: Duration,
+    /// Total number of finalized transactions
+    count: usize,
+    /// Square of the latencies of all finalized transactions.
+    squared_sum: Duration,
```
mlegner: I'm not sure I understand the meaning of this, the units don't work out here. Maybe just store that as an `f64`? Also, it's probably the "Sum of the squares of the latencies of all finalized transactions".

crates/walrus-orchestrator/src/measurements.rs
```diff
@@ -0,0 +1,433 @@
...
+        let mut measurements = HashMap::new();
+        for sample in &parsed.samples {
+            let label = sample
+                .labels
+                .values()
+                .cloned()
+                .collect::<Vec<_>>()
+                .join(",");
+
+            if sample.metric == M::LATENCY_BUCKETS {
```
mlegner: Suggestion: Replace this if, else if, ... chain by a match.

crates/walrus-orchestrator/src/measurements.rs
```diff
@@ -0,0 +1,433 @@
...
+            0.0
+        } else {
+            self.squared_sum.as_secs_f64() / self.count as f64
+        };
+
+        // Compute `avg^2`.
+        let squared_avg = self.average_latency().as_secs_f64().powf(2.0);
+
+        // Compute `squared_sum / count - avg^2`.
+        let variance = if squared_avg > first_term {
```
mlegner, line 125: This case should not occur unless there is an error somewhere.

crates/walrus-orchestrator/src/measurements.rs
```diff
@@ -0,0 +1,433 @@
...
+    /// Compute the average latency.
+    pub fn average_latency(&self) -> Duration {
+        self.sum.checked_div(self.count as u32).unwrap_or_default()
+    }
+
+    /// Compute the standard deviation from the sum of squared latencies:
+    /// `stdev = sqrt( squared_sum / count - avg^2 )`
+    pub fn stdev_latency(&self) -> Duration {
+        // Compute `squared_sum / count`.
+        let first_term = if self.count == 0 {
```
mlegner, line 115: If `self.count == 0`, we can return early as there won't be a meaningful stdev anyway.

crates/walrus-orchestrator/src/measurements.rs
```diff
@@ -0,0 +1,433 @@
...
+            0.0
+        } else {
+            self.squared_sum.as_secs_f64() / self.count as f64
+        };
+
+        // Compute `avg^2`.
+        let squared_avg = self.average_latency().as_secs_f64().powf(2.0);
```
mlegner: Question: Why not `.powi(2)`?

crates/walrus-orchestrator/src/measurements.rs
```diff
@@ -0,0 +1,433 @@
...
+            .filter_map(|x| x.last())
+            .map(|x| x.timestamp)
+            .max()
+            .unwrap_or_default();
+        self.all_measurements(label)
+            .iter()
+            .filter_map(|x| x.last())
+            .map(|x| x.tps(&duration))
+            .max()
+            .unwrap_or_default()
```
mlegner: In essence, what we are doing here is computing `max_count / max_timestamp`, right? Could we maybe define function like the following?
```rust
    fn max_result<T: Default + Ord>(
        &self,
        label: &Label,
        function: impl Fn(&Measurement) -> T,
    ) -> T {
        self.all_measurements(label)
            .iter()
            .filter_map(|x| x.last())
            .map(function)
            .max()
            .unwrap_or_default()
    }
```
Then this could become
```rust
        self.max_result(label, |x| x.count)
            .checked_div(self.max_result(label, |x| x.timestamp.as_secs_f64() as usize))
            .unwrap_or_default() as u64
```

crates/walrus-orchestrator/src/measurements.rs
```diff
@@ -0,0 +1,433 @@
...
+
+    /// Aggregate the stdev latency of multiple data points by taking the max.
+    pub fn aggregate_stdev_latency(&self, label: &Label) -> Duration {
+        self.all_measurements(label)
+            .iter()
+            .filter_map(|x| x.last())
+            .map(|x| x.stdev_latency())
+            .max()
+            .unwrap_or_default()
+    }
```
mlegner: Why are we taking the maximum for the aggregated stdev? AFAIU, standard deviations can be aggregated, but the process is a bit more complicated, see [here](https://math.stackexchange.com/questions/1547141/aggregating-standard-deviation-to-a-summary-point) for example.

crates/walrus-orchestrator/src/main.rs
asonnino: oops. Removing the global

crates/walrus-orchestrator/src/main.rs
asonnino: It is all the loads that the bencharm clients will submit to the system. It allows to run multiple benchmarks through a single command. This is better than making a bash script to call the orchestrator binary multiple times because it allows to skip some lengthy steps such as testbed updates and reconfigurations. I will update the comment to make it clear

crates/walrus-orchestrator/src/logs.rs
asonnino: yes it is the intention. Some clients may not see any error while others may

crates/walrus-orchestrator/src/orchestrator.rs
asonnino: good catch!

crates/walrus-orchestrator/src/orchestrator.rs
asonnino: I think we can have at the very worst 30-40 killed nodes. would it be worthwhile to have an hash set?

crates/walrus-orchestrator/src/measurements.rs
asonnino: >  "Sum of the squares of the latencies of all finalized transactions".

indeed !

crates/walrus-orchestrator/src/measurements.rs
asonnino: yes the process was more complicated and felt there was no need for it. Especially because the longer term plan is to remove this module entirely and directly query a prometheus instance https://github.com/MystenLabs/walrus/issues/284

asonnino: Do you think it is worthwhile to compute a better aggregate or go straight with https://github.com/MystenLabs/walrus/issues/284 ?

crates/walrus-orchestrator/src/orchestrator.rs
mlegner: Not sure, but probably not. 🙂

crates/walrus-orchestrator/src/measurements.rs
mlegner: Ah, I see. No, in this case let's keep it simple. Maybe just call this `max_stdev_latency` to make it clear what this means.

crates/walrus-orchestrator/src/measurements.rs
asonnino: doesn't seem easy:
```
associated consts cannot be referenced in patterns
```
Do you see a workaround?

crates/walrus-orchestrator/src/measurements.rs
asonnino: it was happening when count==0, but now we return early so it shouldn't be necessary.

asonnino: mmh actually it may happen if count==1 and there is a rounding error between 'squared_sum' (reported by Prometheus) and our computation ` self.average_latency().as_secs_f64().powi(2_i32);`

crates/walrus-orchestrator/src/measurements.rs
mlegner: How about something like this?
```rust

            let measurement = measurements.entry(label).or_insert_with(Self::default);
            match &sample.metric {
                x if x == M::LATENCY_BUCKETS => match sample.value {
                    prometheus_parse::Value::Histogram(values) => {
                        for value in values {
                            let bucket_id = value.less_than.to_string();
                            let count = value.count as usize;
                            measurement.buckets.insert(bucket_id, count);
                        }
                    }
                    _ => panic!("Unexpected scraped value"),
                },
                x if x == M::LATENCY_SUM => {
                    measurement.sum = match sample.value {
                        prometheus_parse::Value::Untyped(value) => Duration::from_secs_f64(value),
                        _ => panic!("Unexpected scraped value"),
                    };
                }
                x if x == M::TOTAL_TRANSACTIONS => {
                    measurement.count = match sample.value {
                        prometheus_parse::Value::Untyped(value) => value as usize,
                        _ => panic!("Unexpected scraped value"),
                    };
                }
                x if x == M::LATENCY_SQUARED_SUM => {
                    measurement.squared_sum = match sample.value {
                        prometheus_parse::Value::Counter(value) => Duration::from_secs_f64(value),
                        _ => panic!("Unexpected scraped value"),
                    };
                }
                _ => (),
            }
        }
```

crates/walrus-orchestrator/src/measurements.rs
asonnino: nice!

crates/walrus-orchestrator/src/measurements.rs
mlegner: Ah, ok. I guess it doesn't hurt, so keeping it for such corner cases is probably good.

crates/walrus-orchestrator/src/measurements.rs
asonnino: Nice!

crates/walrus-orchestrator/src/logs.rs
```diff
@@ -6,18 +6,38 @@ use std::cmp::max;
...
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for LogsAnalyzer {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        if self.node_panic
+            || self.client_panic
+            || self.client_errors > other.client_errors
+            || self.node_errors > other.node_errors
```
mlegner, line 32: Remark: This allows for some instances `a > b` and simultaneously `b > a`. But I guess it doesn't matter too much for our use case.

crates/walrus-orchestrator/src/measurements.rs
```diff
@@ -32,7 +32,7 @@ pub struct Measurement {
     sum: Duration,
     /// Total number of finalized transactions
     count: usize,
-    /// Square of the latencies of all finalized transactions.
+    /// Sum of the squares of the latencies of all finalized transactions
     squared_sum: Duration,
```
mlegner: Remark: I still think this shouldn't be a `Duration`: The square of a time isn't a time anymore. This wouldn't change the behavior but makes sure that the physical units make sense.

crates/walrus-orchestrator/src/logs.rs
asonnino: it is not clear in my mind how to order them; we would anyway throw a big error message if any error appears

crates/walrus-orchestrator/src/measurements.rs
asonnino: oops I forgot this one

PR #286, opened 2024-04-16
feat: implement command to initialize system object and sui configs

### LINKED ISSUE #278: CLI command to initialize Walrus objects on chain and create Sui config YAML file

karlwuest:
closes #278

### DIFF
--- a/crates/walrus-sui/src/system_setup.rs
+++ b/crates/walrus-sui/src/system_setup.rs
@@ -97,7 +97,8 @@ pub async fn publish_package(
 
     ensure!(
         transaction_response.status_ok() == Some(true),
-        "Status not ok"
+        "Error during transaction execution: {:?}",
+        transaction_response.errors
     );
 
     // get package id and CommitteeCapHolder id
--- a/crates/walrus-sui/src/test_utils/system_setup.rs
+++ b/crates/walrus-sui/src/test_utils/system_setup.rs
@@ -15,12 +15,12 @@ use crate::{
     types::{Committee, NetworkAddress, StorageNode},
 };
 
-const DEFAULT_GAS_BUDGET: u64 = 10000000000;
-const DEFAULT_CAPACITY: u64 = 1000000000;
+const DEFAULT_GAS_BUDGET: u64 = 1_000_000_000;
+const DEFAULT_CAPACITY: u64 = 1_000_000_000;
 const DEFAULT_PRICE: u64 = 10;
 
 fn contract_path_for_testing(contract: &str) -> anyhow::Result<PathBuf> {
-    Ok(std::env::current_dir()?
+    Ok(PathBuf::from_str(env!("CARGO_MANIFEST_DIR"))?
         .parent()
         .unwrap()
         .parent()
--- a/crates/walrus-sui/src/utils.rs
+++ b/crates/walrus-sui/src/utils.rs
@@ -336,17 +336,14 @@ pub async fn request_sui_from_faucet(
 
     tracing::info!("waiting to receive tokens from faucet");
     // check if we have received a new coin from the faucet
-    loop {
-        let new_coin = handle_pagination(|cursor| {
-            sui_client
-                .coin_read_api()
-                .get_coins(address.to_owned(), None, cursor, None)
-        })
-        .await?
-        .any(|coin| !coins.contains(&coin.coin_object_id));
-        if new_coin {
-            break;
-        }
+    while handle_pagination(|cursor| {
+        sui_client
+            .coin_read_api()
+            .get_coins(address.to_owned(), None, cursor, None)
+    })
+    .await?
+    .all(|coin| coins.contains(&coin.coin_object_id))
+    {
         tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
     }
     tracing::info!("received tokens from faucet");


crates/walrus-sui/src/test_utils/system_setup.rs
```diff
@@ -3,76 +3,23 @@
...
-    let build_config = BuildConfig::new_for_testing();
-    let compiled_package = build_config
-        .build(package_path)
-        .expect("Building package failed");
-    let compiled_modules = compiled_package.get_package_bytes(false);
-    (compiled_package.dependency_ids, compiled_modules)
-}
-
-fn contract_path(contract: &str) -> anyhow::Result<PathBuf> {
+fn contract_path_for_testing(contract: &str) -> anyhow::Result<PathBuf> {
```
giac-mysten, line 22: Question: Can this fail depending on where the test is run?
Suggestion: Maybe using the environment variable `CARGO_MANIFEST_DIR` is more robust?

crates/walrus-sui/src/utils.rs
```diff
@@ -197,6 +217,144 @@ pub(crate) fn blob_id_from_u256(input: U256) -> BlobId {
...
+/// Enum for the different sui networks.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum SuiNetwork {
+    /// Local sui network
+    Localnet,
+    /// Sui Devnet
+    Devnet,
+    /// Sui Testnet
+    Testnet,
+}
```
giac-mysten, line 236: Question: do we need support for custom endpoints? Something like `Custom(SocketAddr)`?

crates/walrus-service/bin/main.rs
```diff
@@ -217,16 +223,19 @@ fn main_with_args(args: Args) -> anyhow::Result<()> {
...
             committee_size,
             n_shards,
-            n_symbols_primary,
-            n_symbols_secondary,
+            sui_network,
+            contract_path,
+            gas_budget,
         } => {
-            generate_dry_run_configs(
+            tracing_subscriber::fmt::init();
```
giac-mysten, line 230: Question: why initialize the subscriber here and not for all commands?

crates/walrus-sui/src/utils.rs
```diff
@@ -197,6 +217,144 @@ pub(crate) fn blob_id_from_u256(input: U256) -> BlobId {
...
+                .coin_read_api()
+                .get_coins(address.to_owned(), None, cursor, None)
+        })
+        .await?
+        .any(|coin| !coins.contains(&coin.coin_object_id));
+        if new_coin {
+            break;
+        }
+        tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
+    }
```
giac-mysten: Nit: very minor simplification
```suggestion
    while let None = handle_pagination(|cursor| {
        sui_client
            .coin_read_api()
            .get_coins(address.to_owned(), None, cursor, None)
    })
    .await?
    .find(|coin| !coins.contains(&coin.coin_object_id))
    {
        tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
    }
```

crates/walrus-sui/src/system_setup.rs
```diff
@@ -0,0 +1,238 @@
...
+
+    // Get a signed transaction
+    let transaction = wallet.sign_transaction(&publish_tx);
+
+    // Submit the transaction
+    let transaction_response = wallet.execute_transaction_may_fail(transaction).await?;
+
+    ensure!(
+        transaction_response.status_ok() == Some(true),
+        "Status not ok"
```
giac-mysten: Nit: improve error message?

crates/walrus-sui/src/test_utils/system_setup.rs
```diff
@@ -3,76 +3,23 @@
...
 use crate::{
-    contracts::{self, StructTag},
+    system_setup::{create_system_object, publish_package, SystemParameters},
     types::{Committee, NetworkAddress, StorageNode},
-    utils::get_created_sui_object_ids_by_type,
 };
 
 const DEFAULT_GAS_BUDGET: u64 = 10000000000;
 const DEFAULT_CAPACITY: u64 = 1000000000;
 const DEFAULT_PRICE: u64 = 10;
```
giac-mysten:
```suggestion
const DEFAULT_GAS_BUDGET: u64 = 10_000_000_000;
const DEFAULT_CAPACITY: u64 = 1_000_000_000;
const DEFAULT_PRICE: u64 = 10;
```

crates/walrus-sui/src/test_utils/system_setup.rs
karlwuest: >Question: Can this fail depending on where the test is run?

Yes, if it would be used in a binary (this would require shipping the contracts with the binary or embedding them, and that's probably not worth doing), which is why I renamed it.

>Maybe using the environment variable CARGO_MANIFEST_DIR is more robust?

Thanks, that's a good suggestion.
I think these should behave the same within our workspace (the working directory when running tests [is always the package root](https://doc.rust-lang.org/cargo/commands/cargo-test.html#working-directory-of-tests)), but `CARGO_MANIFEST_DIR` should be more robust if an external crate would use this for testing.

crates/walrus-sui/src/utils.rs
karlwuest: I would say we add that when/if we need it.

crates/walrus-service/bin/main.rs
karlwuest: Because other commands use a different logging subscriber, in particular `run` uses a `MetricsAndLoggingRuntime`.

PR #288, opened 2024-04-16
refactor: automatically compute number of primary and secondary symbols

### LINKED ISSUE #287: Clean up use of `source_symbols_primary` and `source_symbols_secondary`
With #205 landed, we don't need to specify these values anymore.

karlwuest:
closes #287

### DIFF
--- a/crates/walrus-core/benches/basic_encoding.rs
+++ b/crates/walrus-core/benches/basic_encoding.rs
@@ -10,12 +10,14 @@ use raptorq::SourceBlockEncodingPlan;
 use walrus_core::encoding::{Decoder, DecodingSymbol, Encoder, Primary};
 use walrus_test_utils::{random_data, random_subset};
 
-mod constants;
+const N_SHARDS: u16 = 1000;
+// Likely values for the number of source symbols for the primary and secondary encoding.
+// These values are consistent with BFT and are supported source-block sizes that do not require
+// padding, see https://datatracker.ietf.org/doc/html/rfc6330#section-5.6.
+const SOURCE_SYMBOLS_PRIMARY: u16 = 329;
+const SOURCE_SYMBOLS_SECONDARY: u16 = 662;
 
-const SYMBOL_COUNTS: [u16; 2] = [
-    constants::SOURCE_SYMBOLS_PRIMARY,
-    constants::SOURCE_SYMBOLS_SECONDARY,
-];
+const SYMBOL_COUNTS: [u16; 2] = [SOURCE_SYMBOLS_PRIMARY, SOURCE_SYMBOLS_SECONDARY];
 // Can be at most `u16::MAX`.
 const SYMBOL_SIZES: [u16; 5] = [1, 16, 256, 4096, u16::MAX];
 
@@ -42,7 +44,7 @@ fn basic_encoding(c: &mut Criterion) {
                         let encoder = Encoder::new(
                             data,
                             (*symbol_count).try_into().unwrap(),
-                            constants::N_SHARDS.try_into().unwrap(),
+                            N_SHARDS.try_into().unwrap(),
                             &encoding_plan,
                         )
                         .unwrap();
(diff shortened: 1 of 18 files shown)


crates/walrus-core/src/encoding/config.rs
```diff
@@ -170,10 +161,16 @@ impl EncodingConfig {
...
     /// [`decoding_safety_limit`] and [`source_symbols_for_n_shards`] for more details.
-    #[allow(dead_code)]
-    pub fn new_for_n_shards(n_shards: NonZeroU16) -> Self {
+    ///
+    /// # Panics
+    ///
+    /// Panics if the number of shards causes the number of primary or secondary source symbols
+    /// to be larger than [`MAX_SOURCE_SYMBOLS_PER_BLOCK`].
+    ///
+    /// [rfc6330s5.6]: https://datatracker.ietf.org/doc/html/rfc6330#section-5.6
```
giac-mysten: Suggestion: instead of citing the RFC here, cite it in the docstring for `MAX_SOURCE_SYMBOLS_PER_BLOCK`. 
Also maybe cite `K'_max` in this section https://datatracker.ietf.org/doc/html/rfc6330#section-5.1.2, instead of the table.

crates/walrus-core/benches/blob_encoding.rs
```diff
@@ -23,11 +23,7 @@ const BLOB_SIZES: [(usize, &str); 6] = [
 ];
 
 fn encoding_config() -> EncodingConfig {
-    EncodingConfig::new(
-        constants::SOURCE_SYMBOLS_PRIMARY,
-        constants::SOURCE_SYMBOLS_SECONDARY,
-        constants::N_SHARDS,
-    )
+    EncodingConfig::new(NonZeroU16::new(constants::N_SHARDS).unwrap())
```
mlegner: Suggestion: Remove the constants if we don't use them anymore.

karlwuest: The constants are still used in the `basic_encoding` benchmark.

karlwuest: I'll move `N_SHARDS` to `benches/blob_encoding.rs` and the other two to `benches/basic_encoding.rs` and then remove the `constants` module.

karlwuest: Ok, disregard my previous comment. 
`benches/basic_encoding.rs` requires all three constants. Should we just have `N_SHARDS` defined in both benches or should I keep the `constants` module just for `N_SHARDS` and move the other two to `benches/basic_encoding.rs`?

crates/walrus-core/src/encoding/common.rs
```diff
@@ -5,6 +5,7 @@
 pub const MAX_SYMBOL_SIZE: usize = u16::MAX as usize;
 
 /// The maximum number of source symbols per block for RaptorQ.
+/// See `K'_max` in [rfc6330s5.1.2](<https://datatracker.ietf.org/doc/html/rfc6330#section-5.1.2>)
```
mlegner: Nit: 
```suggestion
/// See `K'_max` in [RFC 6330, Section 5.1.2](<https://datatracker.ietf.org/doc/html/rfc6330#section-5.1.2>)
```

crates/walrus-core/benches/blob_encoding.rs
mlegner: I think the new way is probably simpler. The benchmarks don't necessarily have to use the same values.

crates/walrus-core/benches/basic_encoding.rs
```diff
@@ -10,12 +10,14 @@ use raptorq::SourceBlockEncodingPlan;
...
-const SYMBOL_COUNTS: [u16; 2] = [
-    constants::SOURCE_SYMBOLS_PRIMARY,
-    constants::SOURCE_SYMBOLS_SECONDARY,
-];
+const N_SHARDS: u16 = 1000;
+// Likely values for the number of source symbols for the primary and secondary encoding.
+// These values are consistent with BFT and are supported source-block sizes that do not require
+// padding, see https://datatracker.ietf.org/doc/html/rfc6330#section-5.6.
+const SOURCE_SYMBOLS_PRIMARY: u16 = 324;
+const SOURCE_SYMBOLS_SECONDARY: u16 = 648;
```
mlegner: Suggestion: Replace these by the actual values we would choose for `n == 1000`, namely `329` and `662`.

PR #291, opened 2024-04-17
test: fix the local testbed setup

### LINKED ISSUE #290: Make the local testbed great again
Currently, the local testbed is not functioning due to some issues that were introduced in past refactorings and problems with the Sui connection.

I believe that the main issue is that there is only one coin in the various wallets, but we need two: One for paying and one for gas.

mlegner:
- Call the faucet twice to ensure that two coins are available (one for payment and one for gas).
- Sleep between registering a blob and sending data to storage nodes.
- Fix incorrect commands and instructions in testbed script.
- Fail early in testbed script.

Closes #290

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -7,7 +7,7 @@ use anyhow::{anyhow, Result};
 use fastcrypto::{bls12381::min_pk::BLS12381AggregateSignature, traits::AggregateAuthenticator};
 use futures::Future;
 use reqwest::{Client as ReqwestClient, ClientBuilder};
-use tokio::time::Duration;
+use tokio::time::{sleep, Duration};
 use tracing::Instrument;
 use walrus_core::{
     encoding::{BlobDecoder, EncodingAxis, EncodingConfig, Sliver, SliverPair},
@@ -98,6 +98,10 @@ impl<T: ContractClient> Client<T> {
                 metadata.metadata().encoding_type,
             )
             .await?;
+
+        // We need to wait to be sure that the storage nodes received the registration event.
+        sleep(Duration::from_secs(1)).await;
+
         let certificate = self.store_metadata_and_pairs(&metadata, pairs).await?;
         self.sui_client
             .certify_blob(&blob_sui_object, &certificate)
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/testbed.rs
```diff
@@ -111,10 +111,16 @@ pub async fn testbed_configs(
...
+    let mut faucet_requests = Vec::with_capacity(4);
+    for wallet in [&mut admin_wallet, &mut client_wallet] {
+        for _ in 0..2 {
+            faucet_requests.push(request_sui_from_faucet(
+                wallet.active_address()?,
+                sui_network,
+                &sui_client,
+            ))
+        }
+    }
```
karlwuest, line 123:
```suggestion
    let faucet_requests = [
        admin_wallet.active_address()?,
        client_wallet.active_address()?,
    ]
    .into_iter()
    .flat_map(|address| {
        (0..2).map(move |_| request_sui_from_faucet(address, sui_network, sui_client))
    });
```
Not much shorter, but I liked it 🙂 Feel free to ignore.

crates/walrus-service/src/client.rs
```diff
@@ -98,6 +98,10 @@ impl<T: ContractClient> Client<T> {
                 metadata.metadata().encoding_type,
             )
             .await?;
+
+        // We need to wait to be sure that the storage nodes received the registration event.
+        sleep(Duration::from_secs(1)).await;
```
karlwuest, line 103: Remark: Instead of sleeping here for an arbitrary amount of time, we could also wait until we receive the associated event. This would require a function to get events in reverse order to get the newest, so I don't think this should be part of this PR, but I added #293 to keep track of this.

mlegner: Thanks for creating the issue. 👍

crates/walrus-service/src/testbed.rs
mlegner: Unfortunately, that version doesn't compile. I'll just keep my original version. 😁

karlwuest: >Unfortunately, that version doesn't compile. I'll just keep my original version. 😁

Ahh, I forgot to also mention that the assignment of `sui_client` needs to be changed to 
`let sui_client = &admin_wallet.get_client().await?;`

Issue #50, opened 2024-02-29
Describe RedStuff algorithm in detail in docstring


mlegner:
This should be described in the module docstring in crates/walrus-core/src/encoding.rs.

PR #292, opened 2024-04-17
docs(encoding): add the explanation of the RedStuff encoding

giac-mysten:
Closes: #50

### DIFF
--- a/docs/red-stuff.md
+++ b/docs/red-stuff.md
@@ -347,7 +347,6 @@ For example, then, we have the following settings:
 |      300 |  99 |        5 |        97 |         196 |
 |     1000 | 333 |        5 |       329 |         662 |
 
-
 ## Blob Size Limits
 
 In RaptorQ, the size of a symbol is encoded as a 16-bit integer. Therefore, the maximum size of a blob that can be encoded is $2^{16} - 1 = 65535$
@@ -356,14 +355,14 @@ bytes. As a minimum, a symbol must be at least 1 byte.
 Since the blob is encoded in the rectangular message matrix, the blob size is upperbound by `source_symbols_primary * source_symbols_secondary *
 u16::MAX` and lowerbound by `source_symbols_primary * source_symbols_secondary`. A few examples for the same configurations as above:
 
-| N shards | Min blob size | Max blob size |
-|---------:|--------------:|--------------:|
-|        7 |        15.0 B |        983 KB |
-|       10 |        28.0 B |       1.83 MB |
-|       31 |         171 B |       11.2 MB |
-|      100 |       1.80 KB |        118 MB |
-|      300 |       19.0 KB |       1.25 GB |
-|     1000 |        218 KB |       14.3 GB |
+| N shards | Min blob size | Max blob size | Min encoded blob size | Max encoded blob size |
+|---------:|--------------:|--------------:|----------------------:|----------------------:|
+|        7 |        15.0 B |        983 KB |                56.0 B |               3.67 MB |
+|       10 |        28.0 B |       1.83 MB |                 110 B |               7.21 MB |
+|       31 |         171 B |       11.2 MB |                 868 B |               56.9 MB |
+|      100 |       1.80 KB |        118 MB |               9.10 KB |                596 MB |
+|      300 |       19.0 KB |       1.25 GB |               87.9 KB |               5.76 GB |
+|     1000 |        218 KB |       14.3 GB |                991 KB |               64.9 GB |
 
 ## Sliver Authentication, Blob Metadata, and the Blob ID
 
@@ -388,7 +387,6 @@ Therefore, to prove that a sliver is indeed part of a blob, it is sufficient to 
 
 Finally, the encoding type tag (representing the RedStuff version, or alternative encoding), the length of the blob _before_ the encoding, and the
 Merkle root of the tree over the slivers are hashed together to obtain the _blob ID_.
-
 
 ### Metadata Overhead
 
@@ -417,14 +415,14 @@ However, recall that the number of shards is set and constant, while the number 
 lowering the overhead on the system.  We therefore show here the ratio between the size of the hashes stored on the system to the minimum and maximum
 blob sizes, for `N=1000` shards and different number of nodes (1 node, floor(N/floor(log2(N))) = 111, 1000).
 
-| N=1000                       | Total metadata size | Factor min blob | Factor max blob |
-|------------------------------|--------------------:|----------------:|----------------:|
-| Single node                  |             64.0 KB |           0.294 |        4.48e-06 |
-| floor(N/floor(log2(N)) nodes |             7.10 MB |            32.6 |        0.000498 |
-| N nodes                      |             64.0 MB |             294 |         0.00448 |
+| N = 1000                     | Total metadata size | Factor min blob | Factor max blob | Factor min encoded blob | Factor max encoded blob |
+|------------------------------|--------------------:|----------------:|----------------:|------------------------:|------------------------:|
+| Single node                  |             64.0 KB |           0.294 |        4.48e-06 |                  0.0646 |                9.85e-07 |
+| floor(N/floor(log2(N)) nodes |             7.10 MB |            32.6 |        0.000498 |                    7.17 |                0.000109 |
+| N nodes                      |             64.0 MB |             294 |         0.00448 |                    64.6 |                0.000985 |
 
 We see that for realistic node counts and small blob sizes, the total metadata overhead goes can be multiple times the size of the initial unencoded
-blob.
+blob. However, for larger blob sizes, the overhead is negligible.
 
 [^twincode]: K. V. Rashmi, N. B. Shah and P. V . Kumar, "Enabling node repair in any erasure code for distributed storage," 2011 IEEE International
     Symposium on Information Theory Proceedings, St. Petersburg, Russia, 2011, pp. 1235-1239, doi: 10.1109/ISIT.2011.6033732.


docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+| 7        | 448                      | 3136                   | 20906.667     | 0.31901528    |
+| 10       | 640                      | 6400                   | 22857.143     | 0.34877764    |
+| 11       | 704                      | 7744                   | 19360.        | 0.29541466    |
+| 15       | 960                      | 14400                  | 18701.299     | 0.28536353    |
+| 16       | 1024                     | 16384                  | 32768.        | 0.50000763    |
+| 31       | 1984                     | 61504                  | 35967.251     | 0.54882508    |
+| 50       | 3200                     | 160000                 | 34408.602     | 0.52504161    |
+| 51       | 3264                     | 166464                 | 32512.5       | 0.49610895    |
+| 100      | 6400                     | 640000                 | 35595.106     | 0.54314650    |
+| 1000     | 64000                    | 64000000               | 29385.026     | 0.44838676    |
```
giac-mysten: I had not realized that storing a min-sized blob (217KB) on a 1000-shard committee incurs a ~300x overhead for just the metadata... Maybe triple check 😬

crates/walrus-core/src/encoding.rs
```diff
@@ -1,7 +1,9 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//! TODO(mlegner): Describe encoding algorithm in detail (#50).
+//! Blob encoding functionality, using the RedStuff algorithm.
+//!
+//! For more details on the RedStuff algorithm, see the [relevant docs][../../docs/red-stuff.md].
```
giac-mysten: Links to external `md` files don't work with rustdoc. Any ideas on how to include this in the docs? A script that pastes this on top if the markdown changed?

mlegner: This should work:
```rust
//! Blob encoding functionality, using the RedStuff algorithm.

#![doc = include_str!("../../../docs/red-stuff.md")]
```

docs/red-stuff.md
mlegner: I think we need to be careful with the overall metadata size: We only store this once per *node* and not once per *shard*. Maybe we can assume a relationship like `n_nodes ~ n_shards / log_2(n_shards)` to estimate the overhead.

crates/walrus-core/src/encoding.rs
giac-mysten: That works, very nice! I regret using latex now though 😬

docs/red-stuff.md
giac-mysten: well this is clearly the worst case overhead, but it's not impossible

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+
+The goal of the Walrus system is to provide a distributed storage infrastructure, where a decentralized set of entities—the _storage
+nodes_—collaborate to store and serve files ("blobs" of data). When it comes to storage properties, Walrus has a three key goals:
+
+1. To support extremely high availability and durability of the data;
+2. to have low storage overhead _compared to full replication_, i.e., we don't want to store each blob on every storage node;
+3. to gracefully support node failures, and in particular to allow for _efficient node recovery_ (more on this later).
+
+Given these requirements, one good option is to _erasure encode_ the blobs across the storage nodes. At a high level, erasure encoding (or erasure
+coding) allows to encode the data into $N$ parts, such that the aggregate size of the $N$ blobs is a small multiple of the original blob size, and a
```
mlegner, line 15: Important: We either need to actually make this work with LaTeX (see [here](https://stackoverflow.com/questions/46495063/how-to-write-math-formulas-for-rust-documentation) for some ideas) or replace the "$" by "`".

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+_subset_ $k$ of these parts is sufficient to recover the original blob. We formalize these concepts in the next section, but now note that erasure
+coding already allows us to achieve goals (1) and (2) above, because:
+
+1. Erasure coding allows to recover a blob even if $N - k$ storage nodes fail, providing high availability and durability to the data;
+2. the overall storage overhead is much smaller than for full replication, as for a blob of size $S$, the total storage used in the system is $S * c$
+  instead of $S * N$, where $c << N$ is a small constant (4.5 in Walrus's case).
```
mlegner: If these are actually rendered by LaTeX, I would replace `*` by `\cdot` and `<<` by `\ll`.

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+To achieve the third requirement, however, simple erasure coding is insufficient: A failed node that wants to reconstruct its part of the encoding
+needs to first fetch at least $k$ other parts, reconstruct the blob, and then re-encode its own part. Therefore, the communication overhead for
+recovery is on the order of the size of the whole blob, $S$. With RedStuff, we are instead capable of reconstructing the encoded part of a failed node
+by fetching only $O(S/N)$ data, i.e., only in the order of the size of the lost part. Thus, we can achieve (3).
+
+## Background
+
+We provide here the essential background on the coding schemes used in RedStuff.
+
+### Erasure Codes, Fountain Codes, & RaptoQ
```
mlegner: Remark: I've never seen the combination of an ampersand with the Oxford comma before. 😁 

Nit:
```suggestion
### Erasure Codes, Fountain Codes, and RaptorQ
```

(Same RaptorQ typo in other places as well.)

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+
+Erasure coding considers the problem of error correction in the case of _bit erasures_, i.e., some bits in the message are lost, as in the case of a
+lossy channel. In general, an erasure code divides a blob (or message) of $S$ bytes into $k$ _symbols_ (bitstrings of fixed length $~S/k$), which are
+then encoded to form a longer message of $N$ symbols, such that the original blob can be recovered from any subset $k'$ of the $N$ symbols. $k/N$ is
+called _code rate_.
+
+Fountain codes are a class of erasure codes. The key property of fountain codes is that the encoding process is _rateless_, i.e., the encoder can
+produce an arbitrary number of encoded parts on the fly, without knowing the total number of parts that will be produced. Further, fountain codes are
```
mlegner: Remark: While this is true, this is not really relevant for us as we *do* know the total number of symbols. What is nice, however, is that we can choose this freely.

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+extremely efficient as they typically require XOR operations to encode and decode the data.
+
+RedStuff is based on the RaptoQ fountain code. RaptoQ is one of the most fast and efficient fountain codes, and has the following properties:
+1. It is _systematic_, i.e., the firs $k$ symbols of the encoded message correspond to the original message;
+2. it is a _linear_ code, i.e., the encoding process is a linear transformation of the input symbols, or in other words, the encoded symbols are
+   linear combinations of the input symbols;
+3. it is _almost optimal_, meaning that the $k' ~ k$. Specifically, the probability of decoding failure for $k' = k + H$ symbols received is
+   $O(256^{-(H+1)})$
```
mlegner: Remark: As constants matter in these cases where `H` is relatively small, we could say that it is `< 1 / 256^{H+1}`.

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+
+### The Twin-Code Framework
+
+An old idea in distributed storage is to use an erasure code to encode blobs of data across multiple storage nodes. Thus, by using $k/N$ rate erasure
+code, for $N$ nodes and $k$ source symbols, the system can tolerate $N - k$ node failures, with just an $N/k$ factor of storage overhead.  However, in
```
mlegner: Question: Are the double spaces here and in other places intentional?

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+
+First of all, the framework divides the storage nodes in _Type 0_ and _Type 1_ nodes. Type 0 nodes recover their share of the data by querying Type 1
+nodes, and vice versa.
+
+The framework utilizes two linear erasure codes (not necessarily distinct), $C_0, C_1$, with rates $k/N_0, k/N_1$; column $l$ in the matrix $C_i$ is
+denoted as $g_{il}$ (a linear code can always be represented as a matrix, and encoding is a multiplication between the vector of symbols and the code
+matrix). A blob of size $S$ is subdivided into $k^2$ symbols (by padding it if necessary), and arranged into a square _message matrix_ $M_0$. $M_1 =
+M_O^t$ is the transpose of the message matrix. Encoding the columns of $M_i$ by $C_i$ results in the data stored on Type $i$ nodes. Specifically, the
+$l$ th storage node of type $i$ will store the vector of encoded symbols $M_i g_{il}$.
+
```
mlegner: Remark: I think this description of the node types may be more confusing than helpful.

Suggestion: Jump directly into our system with the different sliver types and just mention in parentheses or a footnote that the original design had different node types instead.

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+We consider a scenario in which a blob of data is encoded and stored across $N$ _shards_—multiple shards can be mapped to the same storage node—in a
+Byzantine setting. Thus, we assume that up to $f$ of the shards can be corrupted by an adversary, with $f < 1/3 N$, and that the remaining $N - f$
+shards are honest.
+
+The key modifications in RedStuff, compared to the Twin-Code framework, are the following:
+
+- RedStuff uses the RaptoQ fountain code for both the Type 0 and Type 1 encoding. The rates are about $(N-2f)/N$ and $(N-f)/N$ respectively. We later
+  discuss more in detail the exact rates and the decoding probability.
+- however, the blob is not laid out in a square message matrix, but in a rectangular one. As we explain in the following, this is an optimization for
```
mlegner: Nit:
```suggestion
- The blob is not laid out in a square message matrix, but in a rectangular one. As we explain in the following, this is an optimization for
```

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+  the specific BFT setting we are considering.
+- Both "Type 0" and "Type 1" encodings are stored on each shard. In our terminology these are called "slivers", and the two together form a "sliver
+  pair".
+
+### Worked Example
+
+#### Encoding
+
+Consider a Walrus instance with $N = 7 = 3f + 1$ shards. This entails that the number of primary source symbols is $N - 2f = 3$, and secondary
+$N - f = 5$. A blob of size $S = 15 * s$ can therefore be divided into 15 symbols of size $s$, and arranged in the matrix as follows.
```
mlegner: Important: The terms "primary" and "secondary" haven't been introduced yet.

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+r_{3,0} & r_{3,1} & r_{3,2} & r_{3,3} & r_{3,4} & \textcolor{blue}{r_{3,5}} & \textcolor{blue}{r_{3,6}} \\
+r_{4,0} & r_{4,1} & r_{4,2} & r_{4,3} & r_{4,4} & \textcolor{blue}{r_{4,5}} & \textcolor{blue}{r_{4,6}} \\
+r_{5,0} & r_{5,1} & r_{5,2} & r_{5,3} & r_{5,4} & \textcolor{blue}{r_{5,5}} & \textcolor{blue}{r_{5,6}} \\
+r_{6,0} & r_{6,1} & r_{6,2} & r_{6,3} & r_{6,4} & \textcolor{blue}{r_{6,5}} & \textcolor{blue}{r_{6,6}} \\
+\end{array}
+\right]
+$$
+
+Note again that these symbols do not need to be stored on any node because they can always be recomputed by expanding either a primary or secondary
+symbol. E.g, `s45` can be obtained by:
```
mlegner: Nit:
```suggestion
symbol. For example, $r_{4,5}$ can be obtained by:
```

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+$$
+\left[
+\begin{array}{c|c|c|c|c}
+s_{0,0} & s_{0,1} & s_{0,2} & s_{0,3} & s_{0,4} \\
+s_{1,0} & s_{1,1} & s_{1,2} & s_{1,3} & s_{1,4} \\
+s_{2,0} & s_{2,1} & s_{2,2} & s_{2,3} & s_{2,4} \\
+\textcolor{blue}{r_{3,0}} & \textcolor{blue}{r_{3,1}} & \textcolor{blue}{r_{3,2}} & \textcolor{blue}{r_{3,3}} & \textcolor{blue}{r_{3,4}} \\
+\textcolor{blue}{r_{4,0}} & \textcolor{blue}{r_{4,1}} & \textcolor{blue}{r_{4,2}} & \textcolor{blue}{r_{4,3}} & \textcolor{blue}{r_{4,4}} \\
+\textcolor{blue}{r_{5,0}} & \textcolor{blue}{r_{5,1}} & \textcolor{blue}{r_{5,2}} & \textcolor{blue}{r_{5,3}} & \textcolor{blue}{r_{5,4}} \\
+\textcolor{blue}{r_{6,0}} & \textcolor{blue}{r_{6,1}} & \textcolor{blue}{r_{6,2}} & \textcolor{blue}{r_{6,3}} & \textcolor{blue}{r_{6,4}} \\
```
mlegner, line 136: Question: Why call these symbols `r` instead of `s`? `r` for "repair"?

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+s_{1,0} & s_{1,1} & s_{1,2} & s_{1,3} & s_{1,4} & \textcolor{blue}{r_{1,5}} & \textcolor{blue}{r_{1,6}} \\
+\hline
+s_{2,0} & s_{2,1} & s_{2,2} & s_{2,3} & s_{2,4} & \textcolor{blue}{r_{2,5}} & \textcolor{blue}{r_{2,6}} \\
+\end{array}
+\right]
+$$
+
+Each of the _columns_ of this column expansion is a _secondary sliver_. E.g., $[r_{0,6}, r_{1,6}, r_{2,6}]$.
+
+The $i$ th _sliver pair_ is composed of the $i$ th primary and $i$ th secondary slivers. For simplicity, consider that the $i$ th sliver pair is
```
mlegner, line 159: Nit: Use `$i$th` instead of `$i$ th`.

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+\left[
+\begin{array}{ccccc|cc}
+s_{0,0} & s_{0,1} & s_{0,2} & \textcolor{red}{s_{0,3}} & s_{0,4} & r_{0,5} & r_{0,6} \\
+s_{1,0} & s_{1,1} & s_{1,2} & \textcolor{red}{s_{1,3}} & s_{1,4} & r_{1,5} & r_{1,6} \\
+s_{2,0} & s_{2,1} & s_{2,2} & \textcolor{red}{s_{2,3}} & s_{2,4} & r_{2,5} & r_{2,6} \\
+\hline
+\textcolor{red}{r_{3,0}} & \textcolor{red}{r_{3,1}} & \textcolor{red}{r_{3,2}} & \textcolor{red}{r_{3,3}} & \textcolor{red}{r_{3,4}} & & \\
+r_{4,0} & r_{4,1} & r_{4,2} & \textcolor{red}{r_{4,3}} & r_{4,4} & & \\
+r_{5,0} & r_{5,1} & r_{5,2} & \textcolor{red}{r_{5,3}} & r_{5,4} & & \\
+r_{6,0} & r_{6,1} & r_{6,2} & \textcolor{red}{r_{6,3}} & r_{6,4} & & \\
```
mlegner: Important: These symbols aren't actually stored on the shard.
```suggestion
r_{4,0} & r_{4,1} & r_{4,2} & r_{4,3} & r_{4,4} & & \\
r_{5,0} & r_{5,1} & r_{5,2} & r_{5,3} & r_{5,4} & & \\
r_{6,0} & r_{6,1} & r_{6,2} & r_{6,3} & r_{6,4} & & \\
```

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+### Decoding Probability and Decoding Safety Limit
+
+As mentioned above, the reconstruction failure probability of the RaptoQ code is $O(256^{-(H+1)})$, where $H$ is the number of extra symbols received.
+Therefore, it is beneficial that in a system with $f$ Byzantine shards, the number of source symbols for the primary encoding is slightly below
+$N-2f$, and for the secondary encoding slightly below $N-f$. This ensures that whenever a validity or quorum threshold of messages is received, there
+is always a positive $H$ for a low failure probability.
+
+We therefore set the following parameters in our encoding configuration:
+
+- $f$, the maximum number of byzantine shards is $\lfloor (N-1) / 3 \rfloor$.
```
mlegner: Typo:
```suggestion
- $f$, the maximum number of Byzantine shards is $\lfloor (N-1) / 3 \rfloor$.
```

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+| 7        | 2   | 0        | 3         | 5           |
+| 10       | 3   | 0        | 4         | 7           |
+| 11       | 3   | 0        | 5         | 8           |
+| 15       | 4   | 0        | 7         | 11          |
+| 16       | 5   | 1        | 5         | 10          |
+| 31       | 10  | 2        | 9         | 19          |
+| 50       | 16  | 3        | 15        | 31          |
+| 51       | 16  | 3        | 16        | 32          |
+| 100      | 33  | 5        | 29        | 62          |
+| 1000     | 333 | 5        | 329       | 662         |
```
mlegner: I see that these chosen values are interesting from a theoretical perspective. From a practical perspective, however, we are most certainly in the range of 100-10k, so it would be interesting to have a few examples there and only 3-4 below 100.

docs/red-stuff.md
```diff
@@ -0,0 +1,405 @@
...
+| 7        | 448                      | 3136                   | 20906.667     | 0.31901528    |
+| 10       | 640                      | 6400                   | 22857.143     | 0.34877764    |
+| 11       | 704                      | 7744                   | 19360.        | 0.29541466    |
+| 15       | 960                      | 14400                  | 18701.299     | 0.28536353    |
+| 16       | 1024                     | 16384                  | 32768.        | 0.50000763    |
+| 31       | 1984                     | 61504                  | 35967.251     | 0.54882508    |
+| 50       | 3200                     | 160000                 | 34408.602     | 0.52504161    |
+| 51       | 3264                     | 166464                 | 32512.5       | 0.49610895    |
+| 100      | 6400                     | 640000                 | 35595.106     | 0.54314650    |
+| 1000     | 64000                    | 64000000               | 29385.026     | 0.44838676    |
```
mlegner: Some suggestions for this table (applies to some of the others as well):
- Use human-readable formatting. For example `64M` instead of `64000000`.
- Reduce the precision to 2-3 significant digits.
- Replace the % by a factor.
- Make the columns right-aligned.

Additional point: The metadata has additional contents beyond the hashes, which are not included here. However, as this is negligible, especially for larger `n_shards`, it's probably fine to keep it as it is, but maybe mention it?

docs/red-stuff.md
mlegner: Yes, that's true. But this makes it seem much less than what can be expected in practice. I think we need a realistic scenario, if not instead of then at least in addition to the worst-case one.

docs/red-stuff.md
mlegner, line 1: General suggestion: Add crosslinks between our modules and this general documentation. I'm not really sure though, if this works when having the text in an external file. In any case, this could be done in a follow-up PR.

docs/red-stuff.md
giac-mysten: They are a relic of the typography of old that still finds it's way in emacs' formatter 😬 will remove

docs/red-stuff.md
giac-mysten: so you would just skip the explanation of why twin code works?

docs/red-stuff.md
giac-mysten: yes, I have used `s` for source, and `r` for "recovery". Will specify.

docs/red-stuff.md
giac-mysten: unfortunately that seems to break github's markdown parsing.

docs/red-stuff.md
giac-mysten: ah, you are right indeed.

docs/red-stuff.md
mlegner: I would still mention that we based our design on it and use a modified version. You could also keep the list of differences. But I would emphasize *our* design instead of  the original.

docs/red-stuff.md
```diff
@@ -29,78 +29,87 @@ by fetching only $O(S/N)$ data, i.e., only in the order of the size of the lost
...
-  needs to reconstruct.
+re-encode its own part. Therefore, the communication overhead for recovery is on the order of the size of the whole blob, $S$.
 
-## RedStuff Encoding
-
-The RedStuff encoding algorithm is an adaptation of the Twin-Code framework, which allows for efficient node recovery in erasure-coded storage
-systems.
+The Twin-Code Framework aims to solve this issue by allowing for efficient node recovery. We briefly describe here how the framework is used in
+RedStuff, and refer to the original paper for the specific details.  The RedStuff encoding algorithm is an adaptation of the Twin-Code framework,
+which allows for efficient node recovery in erasure-coded storage systems.
```
mlegner, line 65: Nit: These sentences feel a bit redundant. Also, while the Twin-Code Framework is mentioned at the very top, it still feels that it comes a bit out of nowhere here.

docs/red-stuff.md
```diff
@@ -29,78 +29,87 @@ by fetching only $O(S/N)$ data, i.e., only in the order of the size of the lost
...
+  [section on sliver-to-shard mapping](#sliver-pair-to-shard-mapping) for more details).
+- The fundamental property we achieve with this construction, and thanks to the linearity of RaptorQ, is that encoding the primary slivers (as rows)
+  with the secondary encoding and the secondary slivers (as columns) with the primary encoding results in the same $N \times N$ expanded message
+  matrix. This property can then be used for lost sliver recovery:
+  - If a shard is trying reconstruct a lost primary sliver, it can request $N-f$ symbols from the encodings of the secondary slivers of other shards,
+    because, as mentioned above, the primary encoding of secondary slivers results in the symbols for primary slivers. Then, since the secondary
+    encoding has $n_C$ source symbols, and $n_C \leq N-2f$, we can decode the original primary sliver from the obtained recovery symbols with high
+    probability (see the later [discussion on recovery probability](#decoding-probability-and-decoding-safety-limit)).
+  - The reconstruction of secondary slivers is identical, but inverting the role of primary and secondary slivers and encodings.
```
mlegner, line 103: Praise: Very nice! While still somewhat complicated, I feel this is much more accessible than the Twin-Code-Framework paper. 🙂

docs/red-stuff.md
```diff
@@ -339,26 +357,20 @@ Since the blob is encoded in the rectangular message matrix, the blob size is up
...
-| ...      | ...           | ...           |
-| 1000     | 217.798 KB    | 14.273392 GB  |
-
+|---------:|--------------:|--------------:|
+|        7 |        15.0 B |      983.0 KB |
+|       10 |        28.0 B |       1.83 MB |
+|       31 |       171.0 B |       11.2 MB |
+|      100 |        1.8 KB |      118.0 MB |
+|      300 |       19.0 KB |       1.25 GB |
+|     1000 |      218.0 KB |       14.3 GB |
```
mlegner: Nit: The significant digits are not fully correct everywhere. For example, it should be either `118 MB` or `117.8 MB` and `218 KB` or `217.8 KB`.

docs/red-stuff.md
```diff
@@ -380,26 +392,40 @@ Merkle root of the tree over the slivers are hashed together to obtain the _blob
...
+cumulative size of the hashes stored on the system, depending on the number of nodes and the number of shards.
+
+| N shards | One node | N/log(N) nodes |  N nodes |
+|---------:|---------:|---------------:|---------:|
+|        7 |  448.0 B |        3.71 KB |  3.14 KB |
+|       10 |  640.0 B |         6.4 KB |   6.4 KB |
+|       31 |  1.98 KB |        41.2 KB |  61.5 KB |
+|      100 |   6.4 KB |       320.0 KB | 640.0 KB |
+|      300 |  19.2 KB |        2.33 MB |  5.76 MB |
+|     1000 |  64.0 KB |        21.3 MB |  64.0 MB |
```
mlegner: Remark: In my original suggestion, I meant `log_2` instead of `log_10` (probably forgot to specify). Also, having fractional values for the number of nodes doesn't make sense. In particular, the value for 7 shards is a bit weird... 😅

docs/red-stuff.md
```diff
@@ -380,26 +392,40 @@ Merkle root of the tree over the slivers are hashed together to obtain the _blob
...
+
+| N=1000             | Total metadata size | Factor min blob | Factor max blob |
+|--------------------|--------------------:|----------------:|----------------:|
+| Single node        |             64.0 KB |           0.294 |        4.48e-06 |
+| N / log10(N) nodes |             21.3 MB |            98.0 |         0.00149 |
+| N nodes            |             64.0 MB |           294.0 |         0.00448 |
+
+
+We see that for realistic node counts and small blob sizes, the total metadata overhead goes can be multiple times the size of the initial unencoded
+blob.
```
mlegner: Nit: There is a "goes" too much.

Also: AFAIU, these factors are relative to the unencoded blob size. However, we already have an encoding factor of ~4.5, which we don't take into account here. Can we include this somehow in the table and/or computation?

Also: Can we formulate this a bit more positively? For example by mentioning that for larger blob sizes, the overhead is negligible?

docs/red-stuff.md
giac-mysten: yeah, this is just because of the way python handles floating point numbers (i.e., it always appends `.0`). Will fix

docs/red-stuff.md
giac-mysten: I added columns for the ratios to the encoded sizes

PR #295, opened 2024-04-18
feat: various client CLI improvements and cleanup

### LINKED ISSUE #210: UX improvements to the client CLI
- [ ] Print useful error messages.
- [ ] After storing a blob successfully, print the blob ID.
- [ ] Allow printing reads to stdout instead of writing to a file.
- [ ] Allow reading blobs without wallet.
- [ ] Maybe: Allow reading a blob from stdin when storing.

Related to #175

mlegner:
- Print more beautiful and useful error messages.
- Add usage documentation.
- After storing a blob successfully, print the blob ID.
- Allow printing reads to stdout instead of writing to a file.
- Allow reading blobs without wallet.
- Adjust instructions in testbed script.
- Include path to wallet config in Walrus config.
- Attempt reading config from standard locations.
- Cleanup storage-node configuration from local config file.

Closes #210

### DIFF
--- a/crates/walrus-sui/src/utils.rs
+++ b/crates/walrus-sui/src/utils.rs
@@ -288,10 +288,12 @@ pub fn create_wallet(
         .parent()
         .unwrap_or(&sui_config_dir()?)
         .join(keystore_filename.unwrap_or(SUI_KEYSTORE_FILENAME));
-    let mut keystore = Keystore::from(FileBasedKeystore::new(&keystore_path)?);
-
+    let mut keystore = FileBasedKeystore::new(&keystore_path)?;
     let (new_address, _phrase, _scheme) =
         keystore.generate_and_add_new_key(SignatureScheme::ED25519, None, None, None)?;
+    keystore.set_path(&keystore_path.canonicalize()?);
+    let keystore = Keystore::from(keystore);
+
     let alias = env.alias.clone();
     SuiClientConfig {
         keystore,


crates/walrus-service/bin/client.rs
```diff
@@ -47,8 +48,16 @@ enum Commands {
...
         blob_id: BlobId,
         /// The file path where to write the blob.
+        ///
+        /// If unset, prints the blob to stdout.
         #[clap(long)]
-        out: PathBuf,
+        out: Option<PathBuf>,
+        /// The URL of the Sui RPC node to use.
+        ///
+        /// If unset, the wallet configuration is applied (if set), or the default [`DEVNET_RPC`]
```
gdanezis: This will becomes the `--help` string, should we actually spell out what DEVNET_RPC is? Is this expanded automatically?

crates/walrus-service/bin/client.rs
```diff
@@ -83,34 +93,98 @@ fn home_client_config() -> Option<PathBuf> {
...
+                Some(path) => {
+                    std::fs::write(&path, blob)?;
+                    println!(
+                        "{} Blob {} reconstructed from Walrus and written to {}.",
+                        success(),
+                        blob_id,
+                        path.display()
+                    )
+                }
+                None => println!("{}", std::str::from_utf8(&blob)?),
```
gdanezis: I am confused by the `from_utf8` . Is blob not bytes? When we push to stdout do we not push bytes? Where is the utf8 coming into play?

scripts/local-testbed.sh
```diff
@@ -45,9 +45,17 @@ done
...
 
 # Instructions to run a client
-echo "\nTo store a file (e.g., the README.md) on the testbed, use the following command:"
-echo "$ cargo run --bin client --\
-    --config working_dir/client_config.yaml --wallet working_dir/sui_client.yaml store README.md 1"
+cat << EOF
+
+To store a file (e.g., the README.md) on the testbed, use the following command:
+$ cargo run --bin client -- \
+--config working_dir/client_config.yaml --wallet working_dir/sui_client.yaml store README.md 1
```
gdanezis: Should we have some default locations we look for config files like the user directory?

crates/walrus-service/bin/client.rs
mlegner: Ah, whoops. You are absolutely right, I wanted to fix this and forgot in the end. Will make this more useful.

crates/walrus-service/bin/client.rs
mlegner: I guess this is something we can discuss. If we want to print the blob to the console, just getting a bunch of numbers between 0 and 255 is probably not what we want. My idea was to simply interpret the bytes as a string instead.

Maybe a good solution would be to have a flag for the output format, with options like `raw` (printing the vector of bytes), `utf8` (the current behavior), and maybe some others. WDYT?

scripts/local-testbed.sh
mlegner: Currently, the default location is `./config.yml`, see [here](https://github.com/MystenLabs/walrus/blob/ml/client-cli/crates/walrus-service/bin/client.rs#L24). We could extend this to also look in some other places as we already do with the wallet configuration (see [here](https://github.com/MystenLabs/walrus/blob/ml/client-cli/crates/walrus-service/bin/client.rs#L64-L94)). To be consistent with Sui, we could use `~/.walrus/walrus_config/client.yml`. WDYT?

crates/walrus-service/bin/client.rs
karlwuest: For playing around with it, I like the string output, because it provides meaningful output with text files, but I think I would generally expect the default behaviour to provide raw output (which allows piping it into a file etc.).

Instead of a flag, I would suggest providing a separate subcommand `cat` that provides the string output and have `read`return the raw output.

crates/walrus-service/bin/client.rs
```diff
@@ -5,25 +5,28 @@
...
 struct Args {
-    // TODO(giac): this will eventually be pulled from the chain.
-    /// The configuration file with the committee information.
+    /// The path to the configuration file with the system information ([`Config`]).
     #[clap(short, long, default_value = "config.yml")]
     config: PathBuf,
-    /// The path to the wallet config file.
+    /// The path to the wallet configuration file.
     #[clap(short, long, default_value = None)]
     wallet: Option<PathBuf>,
```
karlwuest, line 47: Remark: I think we should add the wallet path also to the client config, s.t. we don't have to specify two configs. This doesn't have to be part of this PR, though.

crates/walrus-service/bin/main.rs
```diff
@@ -169,16 +162,6 @@ fn main_with_args(args: Args) -> anyhow::Result<()> {
                     let handled_shards = committee.shards_for_node(storage_node_index);
                     (encoding_config, handled_shards)
                 }
-                CommitteeConfig::FromLocalConfig {
```
karlwuest, line 172: Suggestion: Also remove the `example-local-config.yml` from the repo.

crates/walrus-service/bin/client.rs
giac-mysten: I guess this depends on how we think this command may be used.
If it's to read a book stored on walrus, `utf8` makes sense. If it's to pipe it into another command, bytes make sense.

Maybe having a `read` and a `raw` command is the best option, yes.

crates/walrus-service/bin/client.rs
mlegner: Thanks for the suggestion, that makes sense. I can include this here.

mlegner: One question here: Should we allow relative paths there? If yes, is this relative to the current working directory or relative to the Walrus config file?

Edit: The latter is probably not easily possible, as the config doesn't necessarily have to be read from a file.

Suggestion: Allow absolute or relative paths. Describe behavior of relative paths and recommend usage of absolute ones.

karlwuest: In the config, I think we should just have absolute paths (I think we should also still allow overwriting the value from the config with a command line argument).
I actually just noticed that within the sui client config that we create, a relative path is used for the keystore, we should change this to an absolute path as well, otherwise the sui operations will fail if we call the walrus client from the wrong directory. I created a quick PR to fix this (#297).

crates/walrus-service/bin/client.rs
mlegner: Actually, I realized, I can probably just call `std::io::stdout().write_all(&blob)`, which should cover both use cases.

crates/walrus-service/bin/client.rs
mlegner: > In the config, I think we should just have absolute paths

Sounds good. 👍 

> I think we should also still allow overwriting the value from the config with a command line argument

Yes, I was planning to keep that with the precedence CLI option > Walrus config > current directory > standard Sui config. And I will clearly document this behavior.

crates/walrus-service/bin/client.rs
gdanezis: Nice one!

scripts/local-testbed.sh
gdanezis: I would like that eventually. Fine to do in a separate PR.

mlegner: I placed it at `~/.walrus/config.yml` for now, see [here](https://github.com/MystenLabs/walrus/pull/295/files#diff-2b46ea17b409b004f810f8e93ccc3ba2ca5ef99b0f091a97f86a5fc9a5d30d4eR53).

PR #298, opened 2024-04-19
fix: consistently handle inputs into decoders

### LINKED ISSUE #88: Handle input into decoders consistently
Suggestion: Soft fail with logging; i.e., drop inputs that have an incorrect format but log a warning.

Original discussion in https://github.com/MystenLabs/walrus/pull/80#discussion_r1517933935.

mlegner:
Drop all inputs that don't have the expected length and log a warning in these cases.

Closes #88

### DIFF
--- a/crates/walrus-core/src/lib.rs
+++ b/crates/walrus-core/src/lib.rs
@@ -267,7 +267,7 @@ impl ShardIndex {
     ///
     /// ```
     /// # use walrus_core::ShardIndex;
-    ///
+    /// #
     /// assert!(ShardIndex::range(0..3).eq([ShardIndex(0), ShardIndex(1), ShardIndex(2)]));
     /// assert!(ShardIndex::range(0..3).eq(ShardIndex::range(..3)));
     /// assert!(ShardIndex::range(0..3).eq(ShardIndex::range(..=2)));
--- a/crates/walrus-core/src/utils.rs
+++ b/crates/walrus-core/src/utils.rs
@@ -115,6 +115,24 @@ macro_rules! wrapped_uint {
 
 /// Returns a string that either contains the full `data` slice or its first few values separated by
 /// commas.
+///
+/// # Examples
+///
+/// ```
+/// # use walrus_core::utils::data_prefix_string;
+/// #
+/// assert_eq!(data_prefix_string(&Vec::<u8>::new(), 0), "data: []");
+/// assert_eq!(data_prefix_string(&Vec::<u8>::new(), 1), "data: []");
+/// assert_eq!(data_prefix_string(&Vec::<String>::new(), 0), "data: []");
+/// assert_eq!(data_prefix_string(&[1], 1), "data: [1]");
+/// assert_eq!(data_prefix_string(&[1], 2), "data: [1]");
+/// assert_eq!(data_prefix_string(&[1, 2, 3, 4], 4), "data: [1, 2, 3, 4]");
+/// assert_eq!(data_prefix_string(&[1, 2, 3, 4], 10), "data: [1, 2, 3, 4]");
+/// assert_eq!(data_prefix_string(&["1", "2", "3", "x"], 10), "data: [1, 2, 3, x]");
+/// assert_eq!(data_prefix_string(&[1, 2, 3, 4], 1), "data_prefix: [1, ...]");
+/// assert_eq!(data_prefix_string(&[1, 2, 3, 4], 3), "data_prefix: [1, 2, 3, ...]");
+/// assert_eq!(data_prefix_string(&["x", "y", "z"], 1), "data_prefix: [x, ...]");
+/// ```
 #[inline]
 pub fn data_prefix_string<T: ToString>(data: &[T], max_values_printed: usize) -> String {
     let data_items = data[..data.len().min(max_values_printed)]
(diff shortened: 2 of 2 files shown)


crates/walrus-core/src/encoding/basic_encoding.rs
```diff
@@ -146,6 +146,7 @@ impl Encoder {
 #[derive(Debug, Clone)]
 pub struct Decoder {
     raptorq_decoder: SourceBlockDecoder,
+    symbol_size: NonZeroU16,
```
giac-mysten, line 149: Remark: a bit annoying that we cannot get this from the `SourceBlockDecoder` 🤷

crates/walrus-core/src/encoding/basic_encoding.rs
```diff
@@ -174,10 +176,24 @@ impl Decoder {
...
-            .map(DecodingSymbol::<U, V>::into_encoding_packet);
+        let expected_symbol_size = self.symbol_size.get() as usize;
+        let packets = symbols.into_iter().filter_map(|symbol| {
+            let actual_symbol_size = symbol.len();
+            if actual_symbol_size == expected_symbol_size {
+                Some(symbol.into_encoding_packet())
+            } else {
+                // Drop symbols of incorrect length and log a warning.
+                tracing::warn!(
+                    ?symbol,
```
giac-mysten: Question: do we want to print the whole symbol? it may be quite a large vector in the logs.
Suggestion: maybe implement display  for `DecodingSymbol` to have it show something like `index: 13; data_prefix: [128, 31, 101, ...]; axis: Primary; with_proof: true;`

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -427,18 +427,29 @@ impl<'a, T: EncodingAxis> BlobDecoder<'a, T> {
     ///
     /// If decoding failed due to an insufficient number of provided slivers, it can be continued
     /// by additional calls to [`decode`][Self::decode] providing more slivers.
-    pub fn decode(&mut self, slivers: impl IntoIterator<Item = Sliver<T>>) -> Option<Vec<u8>> {
+    pub fn decode<S>(&mut self, slivers: S) -> Option<Vec<u8>>
+    where
+        S: IntoIterator<Item = Sliver<T>>,
```
giac-mysten, line 432: Question: why change from `impl` to the generic trait version?

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -427,18 +427,29 @@ impl<'a, T: EncodingAxis> BlobDecoder<'a, T> {
...
+            let expected_len = self.decoders.len();
+            let expected_symbol_size = self.symbol_size;
+            if sliver.symbols.len() != expected_len
+                || sliver.symbols.symbol_size() != expected_symbol_size
             {
-                // Ignore slivers of incorrect length or incorrect symbol size.
-                // Question(mlegner): Should we return an error instead? Or at least log this?
+                // Drop slivers of incorrect length or incorrect symbol size and log a warning.
+                tracing::warn!(
+                    ?sliver,
```
giac-mysten: Same here, the sliver may be very long?

crates/walrus-core/src/encoding/symbols.rs
```diff
@@ -300,6 +300,16 @@ impl<T: EncodingAxis, U> DecodingSymbol<T, U> {
...
+
+    /// Returns the symbol size in bytes.
+    pub fn len(&self) -> usize {
+        self.data.len()
+    }
+
+    /// Returns true iff the symbol size is 0.
+    pub fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
```
giac-mysten, line 316: Remark: this is not used afaics.

mlegner: That's true. However, clippy emits a [warning](https://rust-lang.github.io/rust-clippy/rust-1.77.0/index.html#/len_without_is_empty) when you implement `len` but not `is_empty`. I actually wonder why there isn't a trait for that...

crates/walrus-core/src/encoding/basic_encoding.rs
mlegner: Good point, will change.

crates/walrus-core/src/encoding/basic_encoding.rs
mlegner: Agreed. But storing an additional two bytes somewhere won't kill us. 😁

crates/walrus-core/src/encoding/symbols.rs
giac-mysten: very interesting, didn't know that!

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: In general, for function arguments, the `impl trait` version is just [syntactic sugar for an explicit generic type parameter](https://doc.rust-lang.org/reference/types/impl-trait.html#anonymous-type-parameters). Here I had to add the `T: EncodingAxis` to be able to print the sliver, so the `impl trait` version no longer worked (maybe I just did it wrong?).

giac-mysten: > Here I had to add the T: EncodingAxis to be able to print the sliver, so the impl trait version no longer worked (maybe I just did it wrong?).

Odd, the following signature compiles for me

```
fn decode<T: EncodingAxis>(slivers: impl IntoIterator<Item = Sliver<T>>) -> Option<Vec<u8>>
{
    todo!()
}
```

mlegner: Yes, that works. But then we have the trait constraints split between the generic parameter and the `impl trait` clause. So i find it clearer to just remove the syntactic sugar altogether and specify the constraints with the `where` clause.

crates/walrus-core/src/encoding/basic_encoding.rs
mlegner: Done in f051ba4ad59355e9042b2b1d26035e5e6e97a46f.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: Done in f051ba4ad59355e9042b2b1d26035e5e6e97a46f.

PR #300, opened 2024-04-21
ci: fix test-coverage job

mlegner:
Fixes an error introduced in #274.

We need to checkout the code before running a local action.

### DIFF
--- a/.github/actions/run_tarpaulin/action.yml
+++ b/.github/actions/run_tarpaulin/action.yml
@@ -4,7 +4,6 @@ description: Run all tests and report coverage
 runs:
   using: "composite"
   steps:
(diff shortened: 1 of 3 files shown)


.github/actions/run_tarpaulin/action.yml
```diff
@@ -4,7 +4,6 @@ description: Run all tests and report coverage
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@v4
```
giac-mysten, line 7: Question: why do we not need the checkout here?

mlegner: This action is being run after the checkout already happened, so we don't need to do it again.

PR #305, opened 2024-04-22
fix: use the unencoded blob size when registering a blob

### LINKED ISSUE #105: Change size limit enforcement in smart contracts
Currently, the storage size is enforced when [buying a storage resource](https://github.com/MystenLabs/walrus/blob/a66c9e24dc3ed89862555e6ffda166252a6ed96a/contracts/blob_store/sources/system.move#L217). To potentially enable multiple encoding types in the future, and to account for the cost of storing metadata, we decided that the storage size of the resource should be equal to the total storage required in Walrus (i.e. sum of the storage on all shards, counting the metadata once per shard).

Thus, the maximum storage size depends on 1) the number of shards and 2) the encoding type and should be enforced when registering a blob instead of when buying a storage resource.

### LINKED ISSUE #303: Fix the computation of the blob id to make it consistent between contracts and walrus_core
There is currently a disconnect between the way we handle/use the blob size between the computation of the blob id and when interacting with the contracts. 

At some point, we had discussed that for the reservation of space, we should use the encoded size and therefore that's what's passed to the contracts. However, in walrus_core , we use the unencoded size for the computation of the blob Id. After #282 , the computation of the blob id is checked in the contracts, but uses the encoded size. Because of that, storing blobs now no longer works in the testbed. There are a few options to resolve this:

1. Use the encoded size in the computation of the blob Id.
2. Use the unencoded size to reserve space for blobs and compute the encoded size in the contract to reserve space.
3. Use the encoded size to reserve space, but the unencoded size to register a blob (also requires computing the encoded size in the contract to verify that it is within the reserved space).

Option 2 only works if we either reserve space for a specific encoding type or only have a single encoding type, but both 1. and 3. seem viable.

karlwuest:
closes #303 
also closes #105, since this is no longer necessary

### DIFF
--- a/contracts/blob_store/sources/redstuff.move
+++ b/contracts/blob_store/sources/redstuff.move
@@ -16,6 +16,8 @@ module blob_store::redstuff {
     /// Computes the encoded length of a blob for the Red Stuff encoding, given its
     /// unencoded size and the number of shards. The output length includes the
     /// size of the metadata hashes and the blob ID.
+    /// This computation is the same as done by the function of the same name in
+    /// `crates/walrus_core/encoding/config.rs` and should be kept in sync.
     public(package) fun encoded_blob_length(unencoded_length: u64, n_shards: u16) : u64 {
         let slivers_size = ((source_symbols_primary(n_shards) as u64)
             + (source_symbols_secondary(n_shards) as u64))
@@ -83,7 +85,9 @@ module blob_store::redstuff {
     }
 
     #[test]
-    fun test_encoded_size_() {
+    /// These tests replicate the tests for `encoded_blob_length` in
+    /// `crates/walrus_core/encoding/config.rs` and should be kept in sync.
+    fun test_encoded_size() {
         assert_encoded_size(1, 10, 10*((4+7) + 10*2*32 + 32));
         assert_encoded_size(1, 1000, 1000*((329+662) + 1000*2*32 + 32));
         assert_encoded_size((4*7)*100, 10, 10*((4+7)*100 + 10*2*32 + 32));
--- a/contracts/blob_store/Move.lock
+++ b/contracts/blob_store/Move.lock
@@ -21,6 +21,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.23.0"
-edition = "2024.beta"
+compiler-version = "1.22.0"
+edition = "legacy"
 flavor = "sui"
--- a/crates/walrus-core/src/encoding/config.rs
+++ b/crates/walrus-core/src/encoding/config.rs
@@ -259,6 +259,9 @@ impl EncodingConfig {
     ///
     /// The output length includes the metadata and the blob ID sizes.
     /// Returns `None` if the blob size cannot be computed.
+    ///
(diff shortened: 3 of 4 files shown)


contracts/blob_store/sources/redstuff.move
mlegner, line 1: This seems to be a reimplementation of the Rust code. As far as I can tell, it behaves identically, and I see that you added the exact same test cases to both.

However, it would be ideal to have some tests to check that the two implementations are really the same. Is there any straight-forward way of doing that?

Maybe a minimal solution would be to add comments to both files that any change needs to be reflected by the other one.

contracts/blob_store/sources/storage_resource.move
```diff
@@ -90,11 +76,6 @@ module blob_store::storage_resource {
         split_size: u64,
         ctx: &mut TxContext,
     ) : Storage {
-        assert!(
-            split_size >= MIN_STORAGE_AMOUNT
-            && storage.storage_size - split_size >= MIN_STORAGE_AMOUNT,
-            EInvalidAmount
-        );
```
mlegner, line 97: Question: Wouldn't it still be nice to have a minimal size for storage resources as there is a minimum encoded blob size?

contracts/blob_store/sources/redstuff.move
karlwuest: >However, it would be ideal to have some tests to check that the two implementations are really the same. Is there any straight-forward way of doing that?

I don't think there is a really good way of ensuring this (which is why I added the same test cases to both). I think adding a comment in both files is a good suggestion. I will also change the integration tests to use the computed encoding size for reserving the storage resources (currently they just reserve a large amount of space). That should at least catch (some) cases where the calculation in the contract calculates the size as larger than it is, which would make the contract call fail.

contracts/blob_store/sources/storage_resource.move
karlwuest: I thought about this, but I don't really see any benefits to it. Do you have anything in mind?

Also, if we keep a minimum size, it should be independent of the encoding size in my opinion.

mlegner: One would be a defense against malicious splitting, where one splits off a part such that the rest is completely unusable. But then again, this could be handled separately. Also, being able to fuse resources can help with this. So maybe we can indeed remove the minimum size for now and revisit that as soon as we are looking more closely at the sale of the resources?

karlwuest: >One would be a defense against malicious splitting, where one splits off a part such that the rest is completely unusable.

Splitting can only be done by the owner, so no one else would be affected. The owner can anyway make the resource unusable, so I don't think this is an attack vector. 

The only possible attack in that direction, which I can see, would be that there is just enough storage space left to store a minimum sized blob and the attacker then just buys a small piece (e.g. a single byte) to make the rest unusable. However, that rest could still be bought and fused with other assets. To prevent such an attack, the minimum size would need to be enforced on the free remaining space (which would then need to either be `>= min_size` or `0`) instead of the resource that is bought, though.

In a secondary market, other attacks may, of course, be possible, but those should be addressed there if possible.

PR #306, opened 2024-04-22
feat: add walrus-sdk with a StorageNode client

jpcsmith:


crates/walrus-sdk/src/node_response.rs
```diff
@@ -0,0 +1,112 @@
...
+            is_content_type_octet_stream(self.headers()),
+            NodeError::from(Kind::InvalidContentType)
+        );
+
+        let body = self.bytes().await.map_err(Kind::Reqwest)?;
+
+        Ok(bcs::from_bytes(&body).map_err(Kind::Bcs)?)
+    }
+
+    #[inline]
```
jpcsmith: Suggestion: Remove

crates/walrus-sdk/src/node_response.rs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            ServiceResponse::Success { data, .. } => Ok(data),
+            ServiceResponse::Error { message, code } => {
+                Err(Kind::ErrorInNonErrorMessage { message, code }.into())
+            }
+        }
+    }
+}
+
+// TODO(jsmith): Remove from `extract` module
```
jpcsmith: Suggestion: Remove

crates/walrus-core/src/messages/storage_confirmation.rs
```diff
@@ -27,6 +33,56 @@ pub struct SignedStorageConfirmation {
...
+        );
+
+        // Perform the asymmetric verification last, as it is the most expensive operation.
+        public_key.verify(&self.confirmation, &self.signature)?;
+
+        Ok(intent)
+    }
+}
+
+/// Error raised by [`SignedStorageConfirmation::verify`] when unable to verify the
```
giac-mysten: Nit: to verify what? 😂

crates/walrus-service/src/client/communication.rs
```diff
@@ -67,10 +50,11 @@ impl<T, E> WeightedResult for NodeResult<T, E> {
 pub(crate) struct NodeCommunication<'a> {
     pub node_index: NodeIndex,
     pub epoch: Epoch,
-    pub client: &'a ReqwestClient,
     pub node: &'a StorageNode,
     pub encoding_config: &'a EncodingConfig,
     pub span: Span,
+
```
giac-mysten:
```suggestion
```

crates/walrus-service/src/client/communication.rs
```diff
@@ -159,41 +109,19 @@ impl<'a> NodeCommunication<'a> {
...
-        &self,
-        metadata: &VerifiedBlobMetadataWithId,
-        shard_idx: ShardIndex,
-    ) -> Result<Sliver<T>, SliverRetrieveError>
-    where
-        Sliver<T>: TryFrom<SliverEnum>,
-    {
-        let response = self
+        let blob_id = metadata.blob_id();
+        let sliver_pair_index = shard_idx.to_pair_index(self.n_shards(), blob_id);
```
giac-mysten:
```suggestion
        let sliver_pair_index = shard_idx.to_pair_index(self.n_shards(), metadata.blob_id());
```

crates/walrus-sdk/src/storage_node.rs
giac-mysten, line 1: Suggestion: I would call this `client.rs` or `storage_client.rs`

crates/walrus-service/src/client.rs
giac-mysten, line 1: Suggestion: update the docs on line 183 (`verify_confirmation` does not exist anymore)

crates/walrus-sdk/src/error.rs
```diff
@@ -0,0 +1,41 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Errors that may be encountered while interacting with a storage node.
+
+/// Error raised during communication with a node.
+#[derive(Debug, thiserror::Error)]
+#[error(transparent)]
+pub struct NodeError {
+    #[from]
+    kind: Kind,
```
giac-mysten, line 11: Question: why the separation of `NodeError` and kind?

crates/walrus-sdk/src/node_response.rs
```diff
@@ -0,0 +1,112 @@
...
+        );
+
+        match non_error_response
+            .json::<ServiceResponse<T>>()
+            .await
+            .map_err(Kind::Reqwest)?
+        {
+            ServiceResponse::Success { data, .. } => Ok(data),
+            ServiceResponse::Error { message, code } => {
+                Err(Kind::ErrorInNonErrorMessage { message, code }.into())
```
giac-mysten, line 71: Question: A correct node should never return this, right? (i.e., there is never such message created our code)

crates/walrus-sdk/src/storage_node.rs
```diff
@@ -0,0 +1,285 @@
...
+        }
+    }
+
+    /// Requests the metadata for a blob ID from the node.
+    pub async fn get_metadata(
+        &self,
+        blob_id: &BlobId,
+    ) -> Result<UnverifiedBlobMetadataWithId, NodeError> {
+        let url = self.endpoints.metadata(blob_id);
+        let response = self.inner.get(url).send().await.map_err(Kind::Reqwest)?;
```
giac-mysten: Question: do you generally prefer this over the following? fewer lines?
```
        let response = self
            .inner
            .get(self.endpoints.confirmation(blob_id))
            .send()
            .await
            .map_err(Kind::Reqwest)?;
```

crates/walrus-service/src/server/extract.rs
```diff
@@ -45,6 +45,14 @@ impl IntoResponse for BcsRejection {
 #[must_use]
 pub struct Bcs<T>(pub T);
 
+impl<T: DeserializeOwned> Bcs<T> {
+    /// Construct a `Bcs<T>` from a byte slice. Prefer the use the `FromRequest` impl but special
+    /// cases may require extracting a Request into Bytes then optionally constructing a `Bcs<T>`.
```
giac-mysten:
```suggestion
    /// Construct a `Bcs<T>` from a byte slice. Prefer the use of the `FromRequest` impl, but special
    /// cases may require extracting a Request into Bytes then optionally constructing a `Bcs<T>`.
```

crates/walrus-service/tests/test_client.rs
```diff
@@ -18,10 +19,13 @@ use walrus_sui::{
     test_utils::{MockContractClient, MockSuiReadClient},
     types::{BlobCertified, BlobRegistered, Committee, StorageNode as SuiStorageNode},
 };
+use walrus_test_utils::Result as TestResult;
 
 #[tokio::test]
 #[ignore = "ignore E2E tests by default"]
-async fn test_store_and_read_blob() {
+async fn test_store_and_read_blob() -> TestResult {
+    let _ = tracing_subscriber::fmt::try_init();
```
giac-mysten, line 27: Question: do we have a problem with conflicting subscribers when running tests in parallel?

crates/walrus-sdk/src/storage_node/service.rs
```diff
@@ -0,0 +1,37 @@
...
+use walrus_core::{metadata::UnverifiedBlobMetadataWithId, BlobId};
+
+use super::Client;
+use crate::error::NodeError;
+
+/// Requests the unverified metadata corresponding to a Blob ID.
+#[derive(Debug)]
+pub struct UnverifiedMetadataRequest(pub BlobId);
+
+impl Service<UnverifiedMetadataRequest> for Client {
```
giac-mysten: Question: why is this in the current PR?

crates/walrus-sdk/src/storage_node.rs
jpcsmith: What's the issue with `storage_node::Client`?

crates/walrus-sdk/src/error.rs
jpcsmith: The error is now opaque (`Kind` does not need to be an error, but it was just more convenient that way), as consumers of the SDK should not depend on which underlying libraries we're using for bcs, HTTP requests, etc.

crates/walrus-sdk/src/node_response.rs
jpcsmith: That's correct.

crates/walrus-sdk/src/storage_node.rs
jpcsmith: I find the former a tad nicer to look at, but it's just a nit from my side why I split it.

crates/walrus-sdk/src/storage_node/service.rs
jpcsmith: Because it was already in the repo :) Removed.

crates/walrus-service/tests/test_client.rs
jpcsmith: Not with `try_init()`. If it fails because the global state is already configured, it returns an Err() which we drop.

crates/walrus-sdk/Cargo.toml
mlegner, line 1: Nit: Since https://github.com/MystenLabs/walrus/pull/296, we use a slightly different notation.

crates/walrus-sdk/src/storage_node.rs
giac-mysten: that I expect the storage node to be in the `storage_node` file, not its client :) but fine either way.

jpcsmith: Renamed to client. If we add any other clients to the walrus-sdk (such as merging with the walrus-sui crate), then we can move the existing modules under storage/storage-node, as they all relate to that.

PR #311, opened 2024-04-23
refactor: use walrus-sdk in rest-api tests

jpcsmith:
Additionally, this speeds up the server tests (at least on Mac) by disabling proxy configuration in the reqwest client.

### DIFF
--- a/crates/walrus-service/src/server.rs
+++ b/crates/walrus-service/src/server.rs
@@ -360,6 +360,8 @@ mod test {
     pub struct MockServiceState;
 
     impl ServiceState for MockServiceState {
+        /// Returns a valid response only for blob IDs with the first byte 0, None for those
+        /// starting with 1, and otherwise an error.
         fn retrieve_metadata(
             &self,
             blob_id: &BlobId,
@@ -391,6 +393,8 @@ mod test {
             Ok(Some(walrus_core::test_utils::sliver()))
         }
 
+        /// Returns a valid response only for the pair index 0, otherwise, returns
+        /// an internal error.
         fn retrieve_recovery_symbol(
             &self,
             _blob_id: &BlobId,
@@ -405,6 +409,7 @@ mod test {
             }
         }
 
+        /// Successful only for the pair index 0, otherwise, returns an internal error.
         fn store_sliver(
             &self,
             _blob_id: &BlobId,
@@ -418,6 +423,8 @@ mod test {
             }
         }
 
+        /// Returns a confirmation for blob ID starting with zero, None when starting with 1,
+        /// and otherwise an error.
         async fn compute_storage_confirmation(
             &self,
             blob_id: &BlobId,


crates/walrus-service/src/server.rs
```diff
@@ -679,7 +654,7 @@ mod test {
         let url = format!("http://{}{path}", config.inner.rest_api_address);
         // TODO(lef): Extract the path creation into a function with optional arguments
 
-        let client = reqwest::Client::new();
+        let client = storage_node_client(config.as_ref()).into_inner();
```
giac-mysten, line 664: Question: Are you planning to add a `get_decoding_symbol` function to the storage client?

crates/walrus-service/src/server.rs
```diff
@@ -515,7 +535,7 @@ mod test {
         let path = METADATA_ENDPOINT.replace(":blobId", &blob_id);
         let url = format!("http://{}{path}", config.as_ref().rest_api_address);
 
-        let client = reqwest::Client::new();
+        let client = storage_node_client(config.as_ref()).into_inner();
```
giac-mysten, line 545: Question/Suggestion (for future PRs): `into_inner` is only used here, and to get the client in the `get_decoding_symbol` tests. In the latter, it can eventually be replaced by a client `get_symbol..` function. Here however we need the inner client to ensure that we have a `CREATED` rather than a "already stored" `OK` status code.

I was thinking this could be useful in general to know on the client side, but the current implementation of `storage_node::Client::store_metadata` does not expose this. This may be useful for example if the client restarts and tries to store slivers again. If it finds that it is already stored on some nodes, it may try to figure out which nodes already have it before re-transmitting everything...

crates/walrus-service/src/server.rs
```diff
@@ -451,57 +453,75 @@ mod test {
...
 
-        assert_eq!(
-            res.headers().get(reqwest::header::CONTENT_TYPE),
-            Some(&reqwest::header::HeaderValue::from_static(
-                "application/octet-stream"
-            ))
-        );
+    fn blob_id_for_internal_server_error() -> BlobId {
+        let mut blob_id = walrus_core::test_utils::random_blob_id();
+        blob_id.0[0] = 2; // Triggers an internal server error.
```
giac-mysten, line 487: Suggestion: this trigger system is very neat, but it is not documented for any of the `MockServiceState` functions. Maybe a bit of documentation could be helpful in the future? Or is it obvious?

crates/walrus-service/src/server.rs
jpcsmith: I agree, but I am reluctant to add functionality until there's a code path that concretely uses it.

crates/walrus-service/src/server.rs
jpcsmith: Eventually, yes.

crates/walrus-service/src/server.rs
jpcsmith: Added documentation.

Issue #257, opened 2024-04-12
Major Epic: Invalid Blob ID handling


gdanezis:
- [x] Extract a proof of invalid blob ID. (#304)
- [x] Share the proof of invalid blob ID with other storage nodes. (#331, #332)
- [x] Node returns a signed statement about an invalid blob ID. (#331, #332)
- [x] Node returns a statement that a blob is invalid to readers with reference to the event (#418 , #449)
- [x] Node deletes metadata and slivers from invalid blobs from the DB (#432 )
- [x] Certificates of invalid blob ID can be uploaded on chain. (#90)
- [x] Nodes record events about invalid blob IDs. (#314)
- [x] ~~Nodes can respond to challenges using the events about invalid blob IDs. (Now tracked as part of #256)~~

PR #314, opened 2024-04-23
feat: implement sui bindings to invalidate blob id

### LINKED ISSUE #310: Implement rust bindings to submit proof of invalid blobs to contracts
Related to #257 and #304

karlwuest:
closes #310 
related to #257

### DIFF
--- a/crates/walrus-service/src/storage/blob_info.rs
+++ b/crates/walrus-service/src/storage/blob_info.rs
@@ -4,16 +4,46 @@
 //! Blob status for the walrus shard storage
 //!
 
+use std::cmp::Ordering;
+
 use serde::{Deserialize, Serialize};
 use walrus_core::{Epoch, ShardIndex};
 use walrus_sui::types::{BlobCertified, BlobEvent, BlobRegistered, InvalidBlobID};
 
-#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone, Copy, PartialOrd, Ord)]
+#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone, Copy)]
 #[repr(u8)]
 pub enum BlobCertificationStatus {
     Registered = 0,
     Certified = 1,
     Invalid = 2,
+}
+
+impl PartialOrd for BlobCertificationStatus {
+    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+/// Explicitly implement `Ord` to prevent issues from extending or otherwise
+/// changing `BlobCertificationStatus` in the future.
+impl Ord for BlobCertificationStatus {
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self {
+            BlobCertificationStatus::Invalid => match other {
+                BlobCertificationStatus::Invalid => Ordering::Equal,
+                _ => Ordering::Greater,
+            },
+            BlobCertificationStatus::Certified => match other {
+                BlobCertificationStatus::Invalid => Ordering::Less,
+                BlobCertificationStatus::Certified => Ordering::Equal,
+                BlobCertificationStatus::Registered => Ordering::Greater,
+            },
+            BlobCertificationStatus::Registered => match other {
+                BlobCertificationStatus::Registered => Ordering::Equal,
+                _ => Ordering::Less,
+            },
+        }
+    }
 }
 
 #[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone, Copy)]
--- a/crates/walrus-sui/src/types.rs
+++ b/crates/walrus-sui/src/types.rs
@@ -16,7 +16,7 @@ use anyhow::{anyhow, bail};
 use fastcrypto::traits::ToFromBytes;
 use move_core_types::u256::U256;
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
+use serde_json::{Map, Value};
 use sui_sdk::rpc_types::{SuiEvent, SuiMoveStruct, SuiMoveValue};
 use sui_types::{base_types::ObjectID, event::EventID};
 use thiserror::Error;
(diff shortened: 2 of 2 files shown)


crates/walrus-service/src/storage.rs
```diff
@@ -278,10 +278,11 @@ fn merge_blob_info(
         tracing::debug!("Updating {current_val:?} with {new_val:?}");
 
         let val = current_val.unwrap_or(new_val);
-        let status = if val.is_certified() || new_val.is_certified() {
-            BlobCertificationStatus::Certified
+        let status = val.status.max(new_val.status);
```
mlegner, line 281: Important: This is quite elegant but also feels pretty dangerous given that `Ord` is automatically derived. What if we add another status that is not so easily comparable semantically?

Suggestion: Manually implement `Ord` for `BlobCertificationStatus` to make the semantics explicit and decrease the probability of messing this up in the future.

crates/walrus-sui/src/client.rs
```diff
@@ -308,6 +315,28 @@ impl ContractClient for SuiContractClient {
...
     }
 
+    async fn invalidate_blob_id(
+        &self,
+        certificate: &InvalidBlobCertificate,
+    ) -> SuiClientResult<()> {
+        // Sort the list of signers, since the move contract requires them to be in
+        // ascending order (see `blob_store::bls_aggregate::verify_certificate`)
+        let mut signers = certificate.signers.clone();
+        signers.sort_unstable();
```
mlegner, line 325: Remark: It seems to me that this could already be part of the interface of `InvalidBlobCertificate`. However, this would mean that the fields would have to be private and a `new` function would sort the signers.

crates/walrus-sui/src/types.rs
```diff
@@ -553,18 +553,41 @@ impl TryFrom<SuiEvent> for BlobCertified {
...
+    type Error = anyhow::Error;
+
+    fn try_from(sui_event: SuiEvent) -> Result<Self, Self::Error> {
+        let Value::Object(object) = sui_event.parsed_json else {
+            return Err(anyhow!("Event is not of type object"));
+        };
+        let epoch = get_u64_field_from_event!(object, "epoch")?;
+        let blob_id =
+            blob_id_from_u256(get_field_from_event!(object, "blob_id", Value::String)?.parse()?);
+        let event_id = sui_event.id;
```
mlegner: Suggestion: We have this code now three times for the different event types. Can we factor this out into a separate function?

crates/walrus-sui/src/types.rs
```diff
@@ -553,18 +553,41 @@ impl TryFrom<SuiEvent> for BlobCertified {
...
+        Ok(Self {
+            epoch,
+            blob_id,
+            event_id,
+        })
+    }
```
mlegner: Question: Where do we check that this is the correct event type?

karlwuest: This is generally checked before calling `.into()`, but I will add a check to the `try_from` functions.

crates/walrus-sui/src/types.rs
karlwuest: I don't think abstracting this into a separate function makes sense. 
The fields (except for the event_id) are not inherently the same, so changes in one of the events are unlikely to be reflected in another.

I will extract the check from the first three lines into a separate function together with the check for the event type and I'll just assign the `event_id` when creating the object. That will reduce the duplicated code to the assignment of the `epoch` and the `blob_id`.

crates/walrus-service/src/storage.rs
karlwuest: Sounds good 👍

crates/walrus-sui/src/client.rs
karlwuest: The main argument to handle it here is that the we sort the list to fulfil a precondition of the smart contract and, if possible, changes there should not cause changes in `walrus-core`.
However, I don't feel strongly about this and am happy to move this to an invariant for the `InvalidBlobCertificate`. I would then also change it in the `ConfirmationCertificate` to make sure that we handle them the same.

mlegner: Hmm... I agree that this probably shouldn't be in `walrus-core`. Overall, I don't have a strong opinion, so let's just keep it as it is.

crates/walrus-sui/src/types.rs
mlegner: > The fields (except for the event_id) are not inherently the same, so changes in one of the events are unlikely to be reflected in another.

Makes sense.

> I will extract the check from the first three lines into a separate function together with the check for the event type and I'll just assign the event_id when creating the object.

Sounds good.

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -4,18 +4,48 @@
...
+            BlobCertificationStatus::Certified => match other {
+                BlobCertificationStatus::Invalid => Ordering::Less,
+                BlobCertificationStatus::Certified => Ordering::Equal,
+                BlobCertificationStatus::Registered => Ordering::Greater,
+            },
+            BlobCertificationStatus::Registered => match other {
+                BlobCertificationStatus::Registered => Ordering::Equal,
+                _ => Ordering::Less,
+            },
+        }
```
mlegner, line 45: Remark: Looks good. I would have a slightly different version, but probably a matter of taste:
```suggestion
        // Check for equality first.
        if self == other {
            return Ordering::Equal;
        }
        // Define all inequalities.
        match (self, other) {
            (Self::Invalid, _) => Ordering::Greater,
            (Self::Certified, Self::Invalid) => Ordering::Less,
            (Self::Certified, _) => Ordering::Greater,
            (Self::Registered, _) => Ordering::Less,
        }
```

karlwuest: I had almost the same version initially, but then decided for the more verbose version to make it more explicit and clearer that some case is not erroneously included in one of the match arms.

PR #315, opened 2024-04-23
feat: add CommitteeService and -Factory trait

jpcsmith:
This adds the base traits for getting services to interact with committee members at the StorageNode, and uses this to determine both the epoch and the encoding config.

Still to-do:

- [x] Default to using the read-client if no factory is provided.
- [x] Refactor main to not load the committee from chain.

Contributes to #279

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -232,24 +232,20 @@ impl StorageNodeHandleBuilder {
         // Get the shards assigned to this storage node, since we now manage the committee, if the
         // node will be present in the committee, it must have at least one shard assigned to it.
         let shard_assignment = storage.shards_present();
-        let is_in_committee = shard_assignment.is_empty();
+        let is_in_committee = !shard_assignment.is_empty();
 
         // Generate key and address parameters for the node.
         let node_info = StorageNodeTestConfig::new(shard_assignment);
         let public_key = node_info.key_pair.as_ref().public().clone();
 
-        // Create a list of the committee members, that contains one or two nodes.
-        let committee_members = [
-            if is_in_committee {
-                Some(node_info.to_storage_node_info("node-under-test"))
-            } else {
-                None
-            },
-            committee_partner(&node_info).map(|info| info.to_storage_node_info("other-node")),
-        ];
-        debug_assert!(committee_members[0].is_some() || committee_members[1].is_some());
-
         let committee_service_factory = self.committee_service_factory.unwrap_or_else(|| {
+            // Create a list of the committee members, that contains one or two nodes.
+            let committee_members = [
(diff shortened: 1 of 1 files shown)


crates/walrus-service/bin/main.rs
jpcsmith, line 1: Remark: The changes to this file were mostly refactoring. I renamed command functions to the same as their command name, and moved them to an internal `commands` module (i.e., `commands::run`, `commands::keygen`, `commands::generate_dry_run_configs`); and removed `CommitteeConfig` as it's not necessary anymore.

crates/walrus-service/src/committee.rs
```diff
@@ -0,0 +1,95 @@
...
+    /// Returns the number of shards in the committee.
+    fn get_shard_count(&self) -> NonZeroU16;
+
+    /// Excludes a member from calls made to the committee.
+    ///
+    /// An excluded member will not be contacted when making calls against the committee. Returns
+    /// false if the identified member does not exist in the committee.
+    #[must_use]
+    fn exclude_member(&mut self, identity: &PublicKey) -> bool;
+}
```
jpcsmith, line 38: Remark: This will have further aggregate API calls over the committee, such as `get_and_verify_metadata(..)`.

karlwuest: To clarify: this means that this is not just an abstraction over the ReadClient/Committee, but will also be used to interact with storage nodes?

jpcsmith: Correct, the committee api is a side-effect of it anyway storing the committee.

crates/walrus-service/src/node.rs
```diff
@@ -193,50 +205,86 @@ impl StorageNodeBuilder {
...
 
+async fn create_read_client(
+    config: &StorageNodeConfig,
+) -> Result<(SuiReadClient, &SuiConfig), anyhow::Error> {
+    let sui_config @ SuiConfig {
+        rpc,
+        pkg_id,
+        system_object,
+        ..
+    } = config
```
karlwuest, line 269: Fancy! I think this is the first time I've seen the `@` operator outside of a `match`.

crates/walrus-service/src/test_utils.rs
```diff
@@ -199,39 +217,62 @@ impl StorageNodeHandleBuilder {
...
+        // Create a list of the committee members, that contains one or two nodes.
+        let committee_members = [
+            if is_in_committee {
+                Some(node_info.to_storage_node_info("node-under-test"))
+            } else {
+                None
+            },
+            committee_partner(&node_info).map(|info| info.to_storage_node_info("other-node")),
+        ];
+        debug_assert!(committee_members[0].is_some() || committee_members[1].is_some());
```
karlwuest: Suggestion: move this into the `unwrap_or_else` closure below.

Issue #316, opened 2024-04-24
Create client tests with faulty committee


giac-mysten:
Currently the `test_store_and_read_blob` uses only correct nodes.
Add crashed/slow nodes to test the client's operation in Byzantine conditions.

Issue #312, opened 2024-04-23
Create fully end-to-end integration test


karlwuest:
Either by replacing the mock client in the [walrus client test](https://github.com/MystenLabs/walrus/blob/main/crates/walrus-service/tests/test_client.rs) with a `SuiContractClient` running on a sui test cluster or by creating a separate test.

PR #317, opened 2024-04-24
test(service): ensure the client can store and retrieve blob with <f+1 crashed nodes

giac-mysten:
Contributes to #316

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -17,14 +17,7 @@ use tempfile::TempDir;
 use tokio_stream::Stream;
 use tokio_util::sync::CancellationToken;
 use typed_store::rocks::MetricConf;
-use walrus_core::{
-    encoding::EncodingConfig,
-    test_utils,
-    Epoch,
-    ProtocolKeyPair,
-    PublicKey,
-    ShardIndex,
-};
+use walrus_core::{test_utils, Epoch, ProtocolKeyPair, PublicKey, ShardIndex};
 use walrus_sui::types::{
     BlobEvent,
     Committee,
--- a/crates/walrus-service/tests/test_client.rs
+++ b/crates/walrus-service/tests/test_client.rs
@@ -1,10 +1,11 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{num::NonZeroU16, time::Duration};
+use std::{num::NonZeroU16, sync::OnceLock, time::Duration};
 
 use anyhow::anyhow;
 use sui_types::{base_types::ObjectID, digests::TransactionDigest, event::EventID};
+use tokio::sync::Mutex;
 use walrus_core::{
     encoding::{EncodingConfig, Primary},
     BlobId,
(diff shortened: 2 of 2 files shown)


crates/walrus-service/src/test_utils.rs
```diff
@@ -302,6 +302,31 @@ impl TestCluster {
...
+    }
+
+    /// Stops the storage node with index `idx` by cancelling its task.
+    pub fn cancel_node(&mut self, idx: usize) {
+        assert!(
+            idx < self.nodes.len(),
+            "the index of the node to be dropped must be within the node vector"
+        );
+        self.nodes[idx].cancel.cancel();
+    }
```
mlegner, line 462: Remark: This could be implemented in a non-panicking way:
```suggestion
    pub fn cancel_node(&mut self, idx: usize) -> Option<()> {
        Some(self.nodes.get(idx)?.cancel.cancel())
    }
```
OTOH, as this is a test function, we may actually *prefer* it to panic. 🤷

crates/walrus-service/tests/test_client.rs
```diff
@@ -3,30 +3,75 @@
...
+    // HACK: This index is needed to ensure that there is no race condition between the
+    // initialization of the `DBMetrics` in the typed-store.
+    unique_idx_to_avoid_race_condition: u64,
+) {
+    // HACK: Sleep according to the unique index to avoid race conditions. The 20ms increments were
+    // selected to ensure that the race condition did not occur even with different sleep times.
+    tokio::time::sleep(Duration::from_millis(
+        unique_idx_to_avoid_race_condition * 20,
+    ))
+    .await;
```
mlegner: I believe there is a nicer way to do this: You can add a `lock` function like
```rust
use tokio::sync::Mutex;

// Prevent tests running simultaneously to avoid interferences or race conditions.
fn lock() -> &'static Mutex<()> {
    static LOCK: OnceLock<Mutex<()>> = OnceLock::new();
    LOCK.get_or_init(|| Mutex::default())
}
```
then this can be simplified to
```suggestion
) {
    let _lock = lock().lock().await;
```

crates/walrus-service/tests/test_client.rs
```diff
@@ -36,49 +81,33 @@ async fn test_store_and_read_blob() -> TestResult {
...
-            shard_ids: shard_ids.iter().map(ShardIndex::from).collect(),
-        })
-        .collect();
-    let committee = Committee::new(members, 0)?;
-    tokio::time::sleep(Duration::from_millis(50)).await;
-
     let sui_contract_client = MockContractClient::new(
         0,
-        MockSuiReadClient::new_with_blob_ids([blob_id], Some(committee)),
+        MockSuiReadClient::new_with_blob_ids([blob_id], Some(cluster.committee()?)),
```
karlwuest, line 92: Remark: 
Providing the `blob_id` for the `MockSuiReadClient` is not necessary, since the events will be created automatically when calling `register_blob` and `certify_blob` on the `MockContractClient`. Similarly, if a clone of the `MockSuiReadClient` that is used in the contract client is also used as an event provider for the storage nodes, the same events will be provided there. I will add this explanation to the documentation of the mock clients, since this was not very clear so far.

This doesn't need to be changed here, since we anyway want to switch to using a sui test cluster for the integration tests (#312), but this may be relevant when adding unit tests to the client at some later time.

crates/walrus-service/tests/test_client.rs
jpcsmith: Suggestion: If you choose to go this way, I'd suggest using it around only `TestClusterBuilder::build` call, as that's where the race condition is triggered. That should allow the tests to still run in parallel.

crates/walrus-service/src/test_utils.rs
giac-mysten: > OTOH, as this is a test function, we may actually prefer it to panic. 🤷

that was my thinking. If you don't have a strong preference, I would have it panic.

crates/walrus-service/tests/test_client.rs
giac-mysten: This is a great idea, sorry for not thinking of this earlier!

crates/walrus-service/tests/test_client.rs
giac-mysten: Thanks, will have a look at #312 later

PR #321, opened 2024-04-26
test: enable reuse of nodes in testbed script

### LINKED ISSUE #308: Add parameter to local testbed to prevent regenerating configs and resources
Currently, whenever one runs the `local-testbed.sh` script, all configs, keys, and on-chain resources are regenerated. This is not necessary in most cases, so we should have a mode for running a testbed with an existing configuration.

karlwuest:
closes #308 
Also allows running the testbed with a local sui network.

Note: currently storing blobs using the testbed doesn't work, see #322 . 
However, I also tested the script (with minor modifications, since starting a node requires the node index and a subcommand) with the previous working version, where everything works (including the following workflow: storing a blob, then shutting down the committee, restarting the existing committee, and then retrieving the blob).

### DIFF
--- a/scripts/local-testbed.sh
+++ b/scripts/local-testbed.sh
@@ -12,12 +12,14 @@ function ctrl_c() {
 
 (tmux kill-server || true) 2>/dev/null
 
+nets=("devnet", "testnet", "localnet")
+
 function usage() {
     echo "Usage: $0 [OPTIONS]"
     echo "OPTIONS:"
-    echo "  -c <committee_size>   Specify number of storage nodes (default: 4)"
-    echo "  -s <n_shards>         Specify number of shards (default: 10)"
-    echo "  -n <network>          Specify network to generate configs for (default: devnet)"
+    echo "  -c <committee_size>   Number of storage nodes (default: 4)"
+    echo "  -s <n_shards>         Number of shards (default: 10)"
+    echo "  -n <network>          Network (${nets[@]}; default: devnet) to generate configs for"
     echo "  -e                    Use existing config"
     echo "  -h                    Print this usage message"
 }
@@ -38,7 +40,7 @@ committee_size=4 # Default value of 4 if no argument is provided
 shards=10 # Default value of 4 if no argument is provided
 network=devnet
 
-while getopts ":n:c:she" arg; do
+while getopts "n:c:s:eh" arg; do
     case "${arg}" in
         n)
             network=${OPTARG}
@@ -74,9 +76,8 @@ if ! [ "$shards" -ge "$committee_size" ] 2>/dev/null; then
     exit 1
 fi
 
-networks=("devnet", "testnet", "localnet")
-if ! [[ ${networks[@]} =~ $network ]]; then
-    echo "Invalid argument: $network is not a valid network (${networks[@]})."
+if ! [[ ${nets[@]} =~ $network ]]; then
+    echo "Invalid argument: $network is not a valid network (${nets[@]})."
     usage
     exit 1
 fi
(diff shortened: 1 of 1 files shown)


scripts/local-testbed.sh
```diff
@@ -12,36 +12,104 @@ function ctrl_c() {
...
 
-committee_size=${1:-4} # Default value of 4 if no argument is provided
+function usage() {
+    echo "Usage: $0 [OPTIONS]"
+    echo "OPTIONS:"
+    echo "  -c <committee_size>   Specify number of storage nodes (default: 4)"
+    echo "  -s <n_shards>         Specify number of shards (default: 10)"
+    echo "  -n <network>          Specify network to generate configs for (default: devnet)"
+    echo "  -e                    Use existing config"
+    echo "  -h                    Print this usage message"
```
mlegner: Nit: I generally prefer heredocs in such cases to avoid the "echo" boilerplate. They are a bit annoying with indentation, however. We would have to use tabs for indentation (would require setting `indent_style = unset` for bash scripts in `.editorconfig`):
```suggestion
	cat <<-EOF
	Usage: $0 [OPTIONS]
	OPTIONS:
	  -c <committee_size>   Specify number of storage nodes (default: 4)
	  -s <n_shards>         Specify number of shards (default: 10)
	  -n <network>          Specify network to generate configs for (default: devnet)
	  -e                    Use existing config
	  -h                    Print this usage message
	EOF
```

scripts/local-testbed.sh
```diff
@@ -12,36 +12,104 @@ function ctrl_c() {
...
+    echo $cmd
+    tmux new -d -s "$1" "$cmd"
+}
+
+existing=false
+committee_size=4 # Default value of 4 if no argument is provided
+shards=10 # Default value of 4 if no argument is provided
+network=devnet
+
+while getopts ":n:c:she" arg; do
```
mlegner: Important: I think this should be
```suggestion
while getopts "n:c:s:he" arg; do
```

scripts/local-testbed.sh
mlegner: Suggestion: Can we also list the valid options for `network`?

scripts/local-testbed.sh
```diff
@@ -12,36 +12,104 @@ function ctrl_c() {
...
+if ! $existing; then
+    # Generate configs
+    echo Generating configuration...
+    cargo run --bin walrus-node -- generate-dry-run-configs \
+    --working-dir $working_dir --committee-size $committee_size --n-shards $shards \
+    --sui-network $network
+    # Spawn nodes
+    for i in $(seq -w 0 $((committee_size-1))); do
+        run_node "dryrun-node-$i" "--cleanup-storage"
+    done
```
mlegner: Nit: This is covered by the more general node instantiation below. Maybe remove this and move the other one outside of the `else`?

karlwuest: We run it here with `--cleanup-storage`, so we anyway need a case distinction. It also ensures that we start exactly the nodes that we just created (instead of also starting additional nodes for which we have a config in `working_dir`).
I can change this to just removing config files and setting a variable here with `--cleanup-storage` that I pass below if you prefer that.

scripts/local-testbed.sh
karlwuest: >They are a bit annoying with indentation, however.

Yeah, that's why I decided against it. I think I'd prefer keeping spaces for indentation for everything, but if you have a strong preference, I'm happy to change it.

scripts/local-testbed.sh
mlegner: Ah, right, my mistake. No, let's just keep it as it is. 👍

scripts/local-testbed.sh
karlwuest: Does the 👍 signify "yes, I have a strong preference" or "keep it as is"? 😅

mlegner: Haha. I'm happy to keep it as it is (only slight preference). 😁

scripts/local-testbed.sh
```diff
@@ -12,12 +12,14 @@ function ctrl_c() {
...
 
+networks=("devnet", "testnet", "localnet")
+
 function usage() {
     echo "Usage: $0 [OPTIONS]"
     echo "OPTIONS:"
     echo "  -c <committee_size>   Specify number of storage nodes (default: 4)"
     echo "  -s <n_shards>         Specify number of shards (default: 10)"
-    echo "  -n <network>          Specify network to generate configs for (default: devnet)"
+    echo "  -n <network>          Specify network (options: ${networks[@]}) to generate configs for"
```
mlegner: Nit:
```suggestion
    echo "  -n <network>          Specify network (options: ${networks[@]}; default: devnet) to generate configs for"
```

scripts/local-testbed.sh
```diff
@@ -85,30 +86,32 @@ fi
...
+for config in $( ls $working_dir/dryrun-node-*.yaml ); do
+    node_name=$(basename -- "$config")
+    node_name="${node_name%.*}"
+    run_node $node_name $cleanup
+    ((i++))
+done
+
+echo "\nSpawned $i existing nodes in separate tmux sessions."
+
+
```
mlegner: Nit:
```suggestion
echo "\nSpawned $i nodes in separate tmux sessions."
```

PR #323, opened 2024-04-26
docs: improve documentation of mock clients

karlwuest:
See [this comment](https://github.com/MystenLabs/walrus/pull/317#discussion_r1580725081) for context

### DIFF
--- a/crates/walrus-sui/src/test_utils/mock_clients.rs
+++ b/crates/walrus-sui/src/test_utils/mock_clients.rs
@@ -39,6 +39,9 @@ use crate::{
 };
 
 /// Mock `ReadClient` for testing.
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/test_utils/mock_clients.rs
```diff
@@ -39,6 +39,9 @@ use crate::{
 };
 
 /// Mock `ReadClient` for testing.
+/// All events added to a `MockSuiReadClient` are provided in all event streams returned
```
mlegner, line 42: Nit: I would add an empty line to separate the details from the summary. Otherwise everything is included in the summary:
image

```suggestion

/// All events added to a `MockSuiReadClient` are provided in all event streams returned
```

Same for the other docstrings below.

PR #326, opened 2024-04-26
test: implement end to end test

### LINKED ISSUE #312: Create fully end-to-end integration test

karlwuest:
closes #312 

~Happy to rebase after #317 is merged (currently, there are not a lot of conflicts, though).~ Rebase is done.

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -495,6 +495,11 @@ impl TestClusterBuilder {
         Self::default()
     }
 
+    /// Returns a reference to the storage node test configs of the builder.
+    pub fn storage_node_test_configs(&self) -> &Vec<StorageNodeTestConfig> {
+        &self.storage_node_configs
+    }
+
     /// Sets the number of storage nodes and their shard assignments from a sequence of the shards
     /// assignmened to each storage.
     ///
--- a/crates/walrus-service/tests/test_client.rs
+++ b/crates/walrus-service/tests/test_client.rs
@@ -11,7 +11,7 @@ use walrus_service::{
     client::{Client, Config},
     committee::SuiCommitteeServiceFactory,
     system_events::SuiSystemEventProvider,
-    test_utils::{storage_node_test_configs_from_shard_assignment, TestCluster},
+    test_utils::TestCluster,
 };
 use walrus_sui::{
     client::{SuiContractClient, SuiReadClient},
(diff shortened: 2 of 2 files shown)


crates/walrus-service/src/test_utils.rs
```diff
@@ -239,15 +252,13 @@ impl StorageNodeHandleBuilder {
             .storage
             .unwrap_or_else(|| empty_storage_with_shards(&[]));
 
+        let node_info = self
+            .test_config
+            .unwrap_or_else(|| StorageNodeTestConfig::new(storage.shards_present()));
         // Get the shards assigned to this storage node, since we now manage the committee, if the
         // node will be present in the committee, it must have at least one shard assigned to it.
```
```suggestion
        // To be in in the committee the node must have at least one shard assigned to it.
```

crates/walrus-service/src/test_utils.rs
```diff
@@ -456,34 +468,27 @@ impl TestClusterBuilder {
...
-                !shards.is_empty(),
-                "shard assignments to nodes must be non-empty"
-            );
-            shard_assignment.push(shards);
-        }
-        assert!(
-            !shard_assignment.is_empty(),
-            "assignments for at least 1 node must be specified"
-        );
+        self.event_providers = configs.iter().map(|_| None).collect();
```
jpcsmith, line 515: Important: You probably want to reset committee_factories as well here and in with_test_configs below, to ensure that the lengths are the same, storing them as a tuple/struct instead of a vector is also an option.

crates/walrus-service/src/test_utils.rs
```diff
@@ -495,37 +500,60 @@ impl TestClusterBuilder {
...
             .map(|_| Some(Box::new(event_provider.clone()) as _))
             .collect();
         self
     }
 
+    /// Sets the [`CommitteeServiceFactory`] used for each node.
+    ///
+    /// If not provided, defaults to [`StubCommitteeServiceFactory`] created with the committee
+    /// of the cluster.
+    pub fn with_committee_service_factories(
```
jpcsmith: Important: Use the above approach of specifying one then cloning. If instead you really do want to accept a list, then the length needs to be checked against the committee list.

crates/walrus-service/src/test_utils.rs
```diff
@@ -495,37 +500,60 @@ impl TestClusterBuilder {
...
+                .map(|_| {
+                    let factory: Box<dyn CommitteeServiceFactory> = Box::new(
+                        StubCommitteeServiceFactory::from_members(committee_members.clone()),
+                    );
+                    factory
+                })
+                .collect::<Vec<_>>()
+        } else {
+            self.committee_factories
+        };
```
jpcsmith: Suggestion: Handle this like the event_providers case in the iterator below: maintain an equal-sized list of Nones/event providers, and then zip and handle Some/None cases for the factory.

crates/walrus-service/tests/test_client.rs
```diff
@@ -58,39 +58,66 @@ async fn test_store_and_read_blob_with_crash_failures(
...
-
-    let config = Config {
-        concurrent_requests: 10,
-        connection_timeout: Duration::from_secs(10),
-        system_pkg: ObjectID::random(),
-        system_object: ObjectID::random(),
-        wallet_config: None,
-    };
+    // Set up the committee
+    let assignment: Vec<&[u16]> = vec![&[0], &[1, 2], &[3, 4, 5], &[6, 7, 8], &[9, 10, 11, 12]];
```
giac-mysten: Remark: this is the same assignment as the `TestClusterBuilder::default` assignment. If you want to ensure that the two in sync, you can get the assignment from there, and then avoid passing the storage node configs.

karlwuest: Thanks for the suggestion! I added a function to the builder now to return the configs (which are needed to build the committee on chain) and use the default builder.

PR #329, opened 2024-04-29
feat(core): implement creation and verification of inconsistency proofs

### LINKED ISSUE #304: Implement functionality to generate and verify inconsistency proofs
Related to #257

### LINKED ISSUE #325: Add `symbol_size` parameter to `recover_sliver` function
See discussion starting at https://github.com/MystenLabs/walrus/pull/298#issuecomment-2077059794.

mlegner:
Open tasks:
- [x] Add tests
- [x] Improve types for recovery symbols. -> 782cbc5726fff5593eb2153eba5ff36fcc740507
- [x] Fix recovery-symbol verification. -> #349 

Contributes to #257 
Closes #304 
Closes #325

### DIFF
--- a/crates/walrus-core/src/encoding/slivers.rs
+++ b/crates/walrus-core/src/encoding/slivers.rs
@@ -10,14 +10,16 @@ use super::{
     errors::{SliverRecoveryError, SliverVerificationError},
     symbols,
     DecodingSymbol,
-    DecodingSymbolPair,
     EncodeError,
     Encoder,
     EncodingAxis,
     EncodingConfig,
     Primary,
+    RecoverySymbol,
     RecoverySymbolError,
+    RecoverySymbolPair,
     Secondary,
+    SliverRecoveryOrVerificationError,
     Symbols,
 };
 use crate::{
@@ -173,7 +175,7 @@ impl<T: EncodingAxis> Sliver<T> {
             .expect("the encoder should always be able to produce an encoding symbol"))
     }
 
-    /// Gets the recovery symbol for a specific target sliver starting from the current sliver.
+    /// Gets the decoding symbol for a specific target sliver starting from the current sliver.
     ///
     /// A [`Primary`] sliver is used to reconstruct symbols for a [`Secondary`] sliver, and
     /// vice-versa. The `target_pair_index` is the index of the [`SliverPair`] to which the sliver
@@ -184,7 +186,7 @@ impl<T: EncodingAxis> Sliver<T> {
     /// Returns a [`RecoverySymbolError::EncodeError`] if the sliver cannot be encoded. See
     /// [`Encoder::new`] for further details about the returned errors. Returns a
     /// [`RecoverySymbolError::IndexTooLarge`] error if `target_pair_index >= n_shards`.
-    pub fn recovery_symbol_for_sliver(
+    pub fn decoding_symbol_for_sliver(
         &self,
         target_pair_index: SliverPairIndex,
         config: &EncodingConfig,
@@ -201,8 +203,9 @@ impl<T: EncodingAxis> Sliver<T> {
         ))
     }
 
-    /// Gets the recovery symbol for a specific target sliver starting from the current sliver,
-    /// together with the Merkle proof computed over the `n_shards` symbols of the [`Sliver`].
+    /// Gets the recovery symbol for a specific target sliver starting from the current sliver.
+    ///
+    /// This contains the Merkle proof computed over the `n_shards` symbols of the [`Sliver`].
     ///
     /// The `target_pair_index` is the index of the [`SliverPair`] to which the sliver to be
     /// recovered belongs.
@@ -212,13 +215,11 @@ impl<T: EncodingAxis> Sliver<T> {
     /// Returns a [`RecoverySymbolError::EncodeError`] if the sliver cannot be encoded. See
     /// [`Encoder::new`] for further details about the returned errors. Returns a
     /// [`RecoverySymbolError::IndexTooLarge`] error if `target_pair_index >= n_shards`.
-    pub fn recovery_symbol_for_sliver_with_proof<U>(
+    pub fn recovery_symbol_for_sliver(
         &self,
         target_pair_index: SliverPairIndex,
         config: &EncodingConfig,
-    ) -> Result<DecodingSymbol<T::OrthogonalAxis, MerkleProof<U>>, RecoverySymbolError>
-    where
-        U: HashFunction<DIGEST_LEN>,
+    ) -> Result<RecoverySymbol<T::OrthogonalAxis, MerkleProof<Blake2b256>>, RecoverySymbolError>
     {
         Self::check_index(target_pair_index.into(), config.n_shards)?;
         let recovery_symbols = self.recovery_symbols(config)?;
@@ -228,7 +229,7 @@ impl<T: EncodingAxis> Sliver<T> {
             .decoding_symbol_at(target_sliver_index.as_usize(), self.index.into())
             .expect("we have exactly `n_shards` symbols and the bound was checked")
             .with_proof(
-                MerkleTree::<U>::build(recovery_symbols.to_symbols())
+                MerkleTree::<Blake2b256>::build(recovery_symbols.to_symbols())
                     .get_proof(target_sliver_index.as_usize())
                     .expect("bound already checked above"),
             ))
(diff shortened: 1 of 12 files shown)


crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -246,49 +243,55 @@ impl<T: EncodingAxis> Sliver<T> {
...
+        I::IntoIter: Iterator<Item = DecodingSymbol<T, V>>,
+        V: MerkleAuth,
+    {
+        let symbol_size = config
+            .symbol_size_for_blob(metadata.unencoded_length())
+            .expect("TODO(mlegner): check");
+        Self::recover_sliver(
+            recovery_symbols
+                .into_iter()
+                .filter(|s| s.verify_as_recovery_symbol(metadata)),
```
giac-mysten: Remark: we should probably log or otherwise propagate the information on recovery symbols that do not verify. We cannot error out to maintain bft resistance.

crates/walrus-core/src/encoding/symbols.rs
```diff
@@ -283,6 +285,9 @@ pub struct DecodingSymbol<T: EncodingAxis, U = ()> {
     ///
     /// This is equal to the ESI as defined in [RFC 6330][rfc6330s5.3.1].
     ///
+    /// In case this is used for sliver recovery, the index represents the *source* index (the
+    /// sliver it was created from), not the *target* sliver (the sliver being recovered).
+    ///
```
giac-mysten: Remark: I have the growing feeling that it would be better to separate the function of decoding symbols from the function of recovery symbols into two distinct types (eg `DecodingSymbol` and `RecoverySymbol` (where possibly the `RecoverySymbol` is an opaque wrapper around `DecodingSymbol`). The reason is this awkward context-dependent interpretation of the `index`, as well as the `_axis` (which, at least, should have an updated docstring, because under recovery the orthogonal of the current docstring is true).

crates/walrus-core/src/encoding/symbols.rs
```diff
@@ -361,6 +366,25 @@ impl<T: EncodingAxis, U: MerkleAuth> DecodingSymbol<T, U> {
...
+    /// The symbol's index is interpreted as the index of the *source* sliver from which is was
+    /// created. If the index is out of range or any other check fails, this returns `false`.
+    pub fn verify_as_recovery_symbol(&self, metadata: &BlobMetadata) -> bool {
+        let inner = || {
+            let n_shards = metadata.n_shards()?;
+            if self.index >= n_shards.get() {
+                return None;
+            }
+            Some(self.verify_proof(metadata.get_sliver_hash(
+                SliverIndex(self.index).to_pair_index::<T::OrthogonalAxis>(n_shards),
```
giac-mysten: Remark: see comment above about the axis.

crates/walrus-core/src/inconsistency.rs
```diff
@@ -0,0 +1,117 @@
...
+pub enum InconsistencyVerificationError {
+    /// No sliver can be decoded from the authentic recovery symbols.
+    #[error("no sliver can be decoded from the recovery symbols")]
+    RecoveryFailure,
+    /// An error occurred during the verification of the target sliver.
+    #[error(transparent)]
+    VerificationError(#[from] SliverVerificationError),
+    /// The recovered sliver is consistent with the metadata.
+    #[error("the target sliver is authentic")]
+    AuthenticTargetSliver,
```
giac-mysten: Nit: this error sounds like it's not an error 😬 The problem is that someone tried to prove that an actually valid sliver is inconsistent: `SliverNotInconsistent` ? or something like this?

crates/walrus-core/src/metadata.rs
```diff
@@ -202,6 +202,25 @@ impl BlobMetadata {
...
+        NonZeroU16::new(self.hashes.len().try_into().ok()?)
+    }
+
+    /// Returns the blob's unencoded length as a [`usize`].
+    ///
+    /// # Panics
+    ///
+    /// Panics if the unencoded length does not fit into a [`usize`]. This should not happen in
+    /// practice.
+    pub fn unencoded_length(&self) -> usize {
```
giac-mysten: Question: can we avoid the panic by having this return `u64`? `symbol_size_for_blob` could also be turned to have `u64` as input.

mlegner: I was actually thinking about this. I don't think we can avoid conversions between `usize` and `u32` or `u64`. And it is inconvenient returning errors in these cases.

Maybe we should just clearly state assumptions we make about the underlying architecture and potentially other parameters. We could check those assumptions when executing a binary and return an error immediately if they are not fulfilled.

That way we could rely on those assumptions everywhere else and confidently perform conversions or unwraps. WDYT?

crates/walrus-core/src/encoding/slivers.rs
mlegner: Yes, I realized this as well, will add logging here.

When actually recovering slivers, we may have the verification within the requests though such that we are sure that we have a sufficient number of recovery symbols.

crates/walrus-core/src/encoding/symbols.rs
mlegner: Good point. I will see if I can find a clear and ergonomic solution.

crates/walrus-core/src/metadata.rs
giac-mysten: > I don't think we can avoid conversions between usize and u32 or u64.

that's unfortunate. `u32` I can see why, but are you sure this is the case for `u64` too? This is mostly used for blob size, which is a `usize` in the beginning (`blob.len()`), but thereafter is mostly used as `u64`.

If this is the case, we could then just check when creating the blob (which in any case must fit in the machines' memory) and then we should be fine.

> That way we could rely on those assumptions everywhere else and confidently perform conversions or unwraps. WDYT?

In case I am mistaken above, then yes, this is the net best thing 👍

crates/walrus-core/src/encoding/slivers.rs
giac-mysten: > we may have the verification within the requests though such that we are sure that we have a sufficient number of recovery symbols

makes sense!

crates/walrus-core/src/metadata.rs
mlegner: > but are you sure this is the case for `u64` too?

I will take a look.

mlegner: I've created #337 to resolve this and will rebase this PR.

crates/walrus-core/src/inconsistency.rs
```diff
@@ -0,0 +1,117 @@
...
+        metadata: &BlobMetadata,
+        encoding_config: &EncodingConfig,
+    ) -> Result<(), InconsistencyVerificationError> {
+        let sliver = Sliver::recover_sliver_with_verification(
+            self.recovery_symbols,
+            self.target_sliver_index,
+            metadata,
+            encoding_config,
+        )
+        .ok_or(InconsistencyVerificationError::RecoveryFailure)?;
```
karlwuest: Question: Under which conditions can this fail?
In particular, could this also fail if the encoding of the symbols is inconsistent (as opposed to, e.g., a merkle proof verification fails, wrong symbol size, not enough symbols)?
If it's the first, then everything is fine (but it would be good to document this on the called function), but if the second is also a possibility, then we need to return more granular failure information from the callee and make a case distinction (since an inconsistent encoding should cause the proof to be accepted).

mlegner: > could this also fail if the encoding of the symbols is inconsistent [...]?

This is a good question, which I recently asked myself. It turns out that the RaptorQ decoder returns decoded data even if there are more than `n_source_symbols` provided input symbols that are inconsistent with each other. 

I'm not exactly sure *what* that encoded data is exactly, though. One thing I need to check is whether this behavior is specific for this implementation or if this is part of the standard.

crates/walrus-core/src/encoding/symbols.rs
mlegner: Actually, I think this is fine. We verify the proof for the source sliver (with which this was created), not the target sliver. A recovery symbol for a primary sliver is created from a secondary sliver, so we need to use `T::OrthogonalAxis` here.

crates/walrus-core/src/inconsistency.rs
mlegner: I've created a follow-up issue for this: #351.

crates/walrus-core/src/encoding/symbols.rs
mlegner: I've attempted a refactor in 782cbc5726fff5593eb2153eba5ff36fcc740507. Let me know what you think.

crates/walrus-core/src/encoding/symbols.rs
giac-mysten: ah you're right, sorry

PR #334, opened 2024-04-30
feat(client): add multiple configuration parameters for concurrent requests

### LINKED ISSUE #263: Client config: Multiple settings for the batch size, depending on the task performed
Storing slivers can have a larger batch size than collecting slivers, because transferring slivers to nodes is helpful, while requesting slivers that are not the used for reconstruction is a waste of storage nodes' resources.

giac-mysten:
A small change to more flexibly control the number of parallel requests the client makes.
Introduces 3 separate settings:
 
- `concurrent_writes`:  The maximum number of _nodes_ to contact in parallel when writing slivers and metadata. Since writing is never wasted work, this setting defaults to `n-f`
- `concurrent_sliver_reads`: The maximum number of shards to contact in parallel when reading slivers. Since reading can be wasted work, the default is `n-2f`
- `concurrent_metadata_reads`:  The maximum number of shards to contact in parallel when reading metadata. Since reading can be wasted work and 1 correct response is sufficient, the default is `3`.

Closes #263

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -10,7 +10,6 @@ use reqwest::{Client as ReqwestClient, ClientBuilder};
 use tokio::time::{sleep, Duration};
 use tracing::Instrument;
 use walrus_core::{
-    bft,
     encoding::{BlobDecoder, EncodingAxis, EncodingConfig, Sliver, SliverPair},
     ensure,
     messages::{Confirmation, ConfirmationCertificate},
@@ -34,6 +33,8 @@ mod utils;
 use communication::{NodeCommunication, NodeResult};
 use error::StoreError;
 use utils::WeightedFutures;
+
+use self::config::default;
 
 /// A client to communicate with Walrus shards and storage nodes.
 #[derive(Debug)]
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/client.rs
```diff
@@ -56,13 +62,22 @@ impl Client<()> {
...
         let encoding_config = EncodingConfig::new(committee.n_shards());
-
+        // Try to store on n-f nodes concurrently, as the work to store is never wasted.
+        let concurrent_writes = config.concurrent_writes.unwrap_or(
+            (committee.n_shards().get() - bft::max_n_faulty(committee.n_shards())) as usize,
+        );
+        // Read n-2f slivers concurrently to avoid wasted work on the storage nodes.
+        let concurrent_sliver_reads = config.concurrent_writes.unwrap_or(
+            (committee.n_shards().get() - 2 * bft::max_n_faulty(committee.n_shards())) as usize,
+        );
```
mlegner: I like the choices. 

Suggestion: Could we keep all the defaults in a single location? For example, add two functions taking `n_shards` as an input to `config::defaults`?

crates/walrus-service/src/client.rs
```diff
@@ -72,14 +87,18 @@ impl Client<()> {
             reqwest_client,
             sui_client: _,
             committee,
-            concurrent_requests,
+            concurrent_writes: concurrent_requests,
```
mlegner: Nit: Change the variable to `concurrent_writes`.

crates/walrus-service/src/client/config.rs
```diff
@@ -11,9 +11,16 @@ use crate::config::LoadConfig;
...
+    /// The maximum number of storage nodes the client contacts in parallel to write slivers and
+    /// metadata. If `None`, the value is set by the client to `n - f`, depending on the number of
+    /// shards `n`.
+    pub concurrent_writes: Option<usize>,
+    /// The maximum number of slivers the client requests in parallel. If `None`, the value is set
+    /// by the client to `n - 2f`, depending on the number of shards `n`.
+    pub concurrent_sliver_reads: Option<usize>,
+    /// The maximum number of nodes the client contacts to get the blob metadata in parallel.
+    #[serde(default = "defaults::default_concurrent_metadata_reads")]
+    pub concurrent_metadata_reads: usize,
```
mlegner: Important: Can you reflect these changes in `example-client-config.yml`?

crates/walrus-service/src/client.rs
giac-mysten: ah excellent point

PR #337, opened 2024-04-30
refactor: improve integer types and conversions + various minor improvements

mlegner:
This refactoring reduces the use of `usize` in the encoding code and clarifies our assumptions on the underlying architecture. This was triggered by this discussion: https://github.com/MystenLabs/walrus/pull/329#discussion_r1583245447.

This also replaces conversion between integer types using `as` by `from`, `try_from`, `into`, and `try_into` to clearly state why we know or expect the conversion to succeed and be lossless.

The PR also includes a few other minor improvements:
- Relax trait bounds on `walrus_test_utils::random_subset*`.
- Improve some error types.
- Add/improve some docstrings.
- Change metadata input to `Sliver::verify`.
- Add a few convenience functions.

### DIFF
--- a/crates/walrus-core/src/encoding/blob_encoding.rs
+++ b/crates/walrus-core/src/encoding/blob_encoding.rs
@@ -228,6 +228,8 @@ impl<'a> BlobEncoder<'a> {
     }
 }
 
+/// The representation of the expanded message matrix.
+///
 /// The expanded message matrix is represented as vector of rows, where each row is a [`Symbols`]
 /// object. This choice simplifies indexing, and the rows of [`Symbols`] can then be directly
 /// truncated into primary slivers.
--- a/crates/walrus-core/src/encoding/config.rs
+++ b/crates/walrus-core/src/encoding/config.rs
@@ -41,8 +41,10 @@ pub struct EncodingConfig {
 }
 
 impl EncodingConfig {
-    /// Creates a new encoding config with the appropriate number of primary and secondary source
-    /// symbols for the given number of shards.
+    /// Creates a new encoding config, given the number of shards.
+    ///
+    /// The number of shards determines the the appropriate number of primary and secondary source
+    /// symbols.
     ///
     /// The decoding probability is given by the [`decoding_safety_limit`]. See the documentation of
     /// [`decoding_safety_limit`] and [`source_symbols_for_n_shards`] for more details.
@@ -58,17 +60,11 @@ impl EncodingConfig {
 
     /// Creates a new encoding configuration for the provided system parameters.
     ///
-    /// # Arguments
-    ///
-    /// * `source_symbols_primary` - The number of source symbols for the primary encoding. This
-    ///   should be equal to or below `n_shards - 2f`, where `f` is the Byzantine parameter.
-    /// * `source_symbols_secondary` - The number of source symbols for the secondary encoding. This
-    ///   should be equal to or below `n_shards - f`.
-    /// * `n_shards` - The total number of shards.
-    ///
-    /// # Returns
-    ///
-    /// The encoding configuration.
+    /// In a setup with `n_shards` total shards -- among which `f` are Byzantine, and
+    /// `f < n_shards / 3` -- `source_symbols_primary` is the number of source symbols for the
+    /// primary encoding (must be equal to or below `n_shards - 2f`), and `source_symbols_secondary`
+    /// is the number of source symbols for the secondary encoding (must be equal to or below
+    /// `n_shards - f`).
     ///
     /// # Panics
     ///
@@ -270,13 +266,11 @@ impl EncodingConfig {
             .expect("this always fits into a `u64`")
     }
 
-    /// Returns an [`Encoder`] to perform a single primary or secondary encoding of the provided
-    /// data.
+    /// Returns an [`Encoder`] to perform a single primary or secondary encoding of the data.
     ///
-    /// # Arguments
-    ///
-    /// * `encoding_axis` - Sets the encoding parameters for the primary or secondary encoding.
-    /// * `data` - The data to be encoded. This *does not* have to be aligned/padded.
+    /// The `data` to be encoded _does not_ have to be aligned/padded. The `encoding_axis` specifies
+    /// which encoding parameters the encoder uses, i.e., the parameters for either the primary or
+    /// the secondary encoding.
     ///
     /// # Errors
     ///
@@ -291,8 +285,9 @@ impl EncodingConfig {
         )
     }
 
-    /// Returns a [`BlobEncoder`] to encode the provided blob into
-    /// [`SliverPair`s][super::SliverPair].
+    /// Returns a [`BlobEncoder`] to encode a blob into [`SliverPair`s][super::SliverPair].
+    ///
+    /// The `blob` to be encoded does not have to be padded.
     ///
     /// # Errors
     ///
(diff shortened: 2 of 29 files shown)


README.md
```diff
@@ -2,6 +2,13 @@
 
 A decentralized blob store using [Sui](https://github.com/MystenLabs/sui) for coordination and governance.
 
+## Hardware requirements
+
+- We assume that this code is executed on at least 32-bit hardware; concretely, we assume that a `u32` can be converted
+  safely into a `usize`.
+- Servers are assumed to use a 64-bit architecture (or higher); concretely, `usize` has at least 64 bits.
+- When a client is executed on a 32-bit architecture, it may fail for blobs above a certain size.
```
giac-mysten: Question: Can we give an safety lower bound?

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -398,15 +398,14 @@ impl<'a, T: EncodingAxis> BlobDecoder<'a, T> {
...
-            return Err(EncodeError::EmptyData);
-        };
+    /// Returns an [`InvalidDataSizeError::DataTooLarge`] if the `blob_size` is too large to be
+    /// decoded. Returns an [`InvalidDataSizeError::EmptyData`] if `blob_size == 0`.
+    pub fn new(config: &'a EncodingConfig, blob_size: u64) -> Result<Self, InvalidDataSizeError> {
+        let symbol_size = config.symbol_size_for_blob(blob_size)?;
+        let blob_size = NonZeroU64::new(blob_size)
+            .expect("already checked when computing the symbol size")
+            .try_into()
+            .map_err(|_| InvalidDataSizeError::DataTooLarge)?;
```
giac-mysten: Question: Why have a `NonZeroU64` and then try-into `NonZeroUsize`? Can we change `blob_size` to `NonZeroU64`?

crates/walrus-core/src/encoding/config.rs
```diff
@@ -184,47 +185,72 @@ impl EncodingConfig {
...
+    pub fn symbol_size_for_blob(&self, blob_size: u64) -> Result<NonZeroU16, InvalidDataSizeError> {
         utils::compute_symbol_size(blob_size, self.source_symbols_per_blob())
     }
 
+    /// The symbol size when encoding a blob of size `blob_size`.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`InvalidDataSizeError::EmptyData`] if `data_length == 0` and an
+    /// [`InvalidDataSizeError::DataTooLarge`] if the data_length cannot be converted to a `u64` or
```
giac-mysten, line 216: Nit:
```suggestion
    /// [`InvalidDataSizeError::DataTooLarge`] if the `data_length` cannot be converted to a `u64` or
```

crates/walrus-core/benches/blob_encoding.rs
```diff
@@ -81,7 +81,9 @@ fn blob_decoding(c: &mut Criterion) {
                 b.iter_batched(
                     || slivers.clone(),
                     |slivers| {
-                        let mut decoder = config.get_blob_decoder::<Primary>(*blob_size).unwrap();
+                        let mut decoder = config
+                            .get_blob_decoder::<Primary>(*blob_size as u64)
```
giac-mysten: Nit: the `blob_size` clould be defined as `u64` in `BLOB_SIZES` above.

crates/walrus-core/src/encoding/config.rs
```diff
@@ -184,47 +185,72 @@ impl EncodingConfig {
...
     /// Computes the length of a blob of given `unencoded_length`, once encoded.
     ///
-    /// The output length includes the metadata and the blob ID sizes.
-    /// Returns `None` if the blob size cannot be computed.
+    /// The output length includes the metadata and the blob ID sizes. Returns `None` if the blob
+    /// size cannot be computed.
     ///
     /// This computation is the same as done by the function of the same name in
     /// `contracts/blob_store/redstuff.move` and should be kept in sync.
     pub fn encoded_blob_length(&self, unencoded_length: usize) -> Option<u64> {
```
giac-mysten: Question/suggestion: why not change `unencoded_length` to usize? Do you expect this called only from eg `blob.len()`.

README.md
mlegner: Hmm... Given the expansion factor of ~4.5, anything beyond 1GB should definitely fail. I expect sizes up to around 500MB to be fine, but I'm not 100% sure.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: Hmm... The idea here is to fail early in case the `u64` we get from the metadata fits into a `usize` which we use for truncating later.

One thing we could do in addition is actually changing the `unencoded_size` in the metadata to a `NonZeroU64`.

crates/walrus-core/src/encoding/config.rs
mlegner: > why not change unencoded_length to usize?

Do you mean `u64`?

Indeed, I expect this function to be used mainly to compute the encoded length of a given blob, where we get the `usize` more easily than the `u64`.

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -55,14 +56,14 @@ pub struct BlobEncoder<'a> {
...
-        let Some(symbol_size) =
-            utils::compute_symbol_size(blob.len(), config.source_symbols_per_blob())
-        else {
-            return Err(EncodeError::DataTooLarge);
-        };
+    /// The actual encoding can be performed with the [`encode()`][Self::encode] method.
+    ///
+    /// # Errors
+    ///
+    /// Returns an [`InvalidDataSizeError`] if the blob is empty or too large to be encoded.
```
jpcsmith: Question: When is the data too large to be encoded?

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -398,15 +404,16 @@ impl<'a, T: EncodingAxis> BlobDecoder<'a, T> {
...
+    /// Returns an [`InvalidDataSizeError::DataTooLarge`] if the `blob_size` is too large to be
+    /// decoded.
+    pub fn new(
+        config: &'a EncodingConfig,
+        blob_size: NonZeroU64,
+    ) -> Result<Self, InvalidDataSizeError> {
+        let symbol_size = config.symbol_size_for_blob(blob_size.get())?;
+        let blob_size = blob_size
+            .try_into()
+            .map_err(|_| InvalidDataSizeError::DataTooLarge)?;
```
jpcsmith, line 437: Remark: If I understand correctly, `blob_size` is persisted as usize because we flatten it into one vector when decoding, and we want to truncate that to the correct size. However, the maximum size for a vector is `isize::MAX` so the flatten in `decode` will still panic. Even if using `isize::MAX` for `blob_size`, I think it would still panic if the `blob_size + padding > isize::MAX`.

In any case, if we want to make this work for very large file sizes, we'd need to change to a streaming interface, as we probably don't want to be allocating vecs which are PiBytes in size.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: The main reason is that it doesn't fit into the message matrix; i.e., the required symbol size is larger than `1<<16` bytes. There is another possibility on 32-bit architectures where the blob simply doesn't fit into memory. I can add some details to the error and/or docstring.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: Hmm... That's a good point. In fact, the actual maximal blob size for 32-bit architectures is even smaller as we (temporarily) allocate more memory when decoding a blob. Do you think it makes sense to implement a better check for the maximum blob size that takes that into account?

jpcsmith: Probably not, but it would be good to unify the errors for data that is too large (such as panicking instead of returning an error), or adding a panic section in the documentation that clarifies when it will still panic.

mlegner: I've added documentation about possible panics on 32-bit architectures to the blob encoding and decoding functions in be603f7feb39ee8364e0afefa7f26db3190412fe.

PR #339, opened 2024-05-01
refactor(client): improve higher-level errors in client

### LINKED ISSUE #338: Improve error propagation from within the client
As discussed with JP, create an opaque error type on which to match in tests

giac-mysten:
Replaces the `anyhow` errors in the higher levels of the client with a specific `ClientError`, to enable easier matching.

Depends on #337 
Closes #338

### DIFF
--- a/crates/walrus-service/src/client/error.rs
+++ b/crates/walrus-service/src/client/error.rs
@@ -33,7 +33,7 @@ pub struct SliverStoreError {
 
 /// Error raised by a client interacting with the storage system.
 #[derive(Debug, thiserror::Error)]
-#[error("client execution failed with error: {kind}")]
+#[error(transparent)]
 pub struct ClientError {
     /// The inner kind of the error.
     #[from]
--- a/crates/walrus-service/tests/test_client.rs
+++ b/crates/walrus-service/tests/test_client.rs
@@ -28,39 +28,36 @@ use walrus_test_utils::async_param_test;
 
 async_param_test! {
     test_store_and_read_blob_with_crash_failures : [
-        #[ignore = "ignore E2E tests by default"] #[tokio::test] no_failures: (&[], &[], Ok(())),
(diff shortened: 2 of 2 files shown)


crates/walrus-service/src/client/error.rs
```diff
@@ -29,3 +30,50 @@ pub struct SliverStoreError {
     pub sliver_type: SliverType,
     pub error: NodeError,
 }
+
+/// Error raised by a client interacting with the storage system.
+#[derive(Debug, thiserror::Error)]
+#[error("client execution failed with error: {kind}")]
```
jpcsmith: Hint: You can also use `#[error(transparent)]` here if you want.

crates/walrus-service/tests/test_client.rs
```diff
@@ -23,39 +28,58 @@ use walrus_test_utils::async_param_test;
...
+                Err(NotEnoughConfirmations(0, 9).into())
             ),
+        #[ignore = "ignore E2E tests by default"] #[tokio::test] f_plus_one_read_failures:
+            (&[], &[0, 4], Ok(())),
+        #[ignore = "ignore E2E tests by default"] #[tokio::test] two_f_plus_one_read_failures:
+            (&[], &[1, 2, 4], Err(NotEnoughSlivers.into())),
+        #[ignore = "ignore E2E tests by default"] #[tokio::test] all_read_failures:
+            (&[], &[0, 1, 2, 3, 4], Err(NoMetadataReceived.into())),
+        #[ignore = "ignore E2E tests by default"] #[tokio::test] read_and_write_overlap_failures:
+            (&[4], &[2, 3], Err(NotEnoughSlivers.into())),
```
jpcsmith: Hint: The only time you're actually passing information here is if it's a kind, which could be simplified by taking `Option<ClientErrorKind>`

PR #340, opened 2024-05-01
docs: replace lists of args with prose

### LINKED ISSUE #196: Harmonize docstrings for functions, replacing argument lists by prose
See also https://github.com/MystenLabs/walrus/pull/194#discussion_r1549304548.

giac-mysten:
Closes #196

### DIFF
--- a/crates/walrus-core/src/encoding/config.rs
+++ b/crates/walrus-core/src/encoding/config.rs
@@ -40,8 +40,10 @@ pub struct EncodingConfig {
 }
 
 impl EncodingConfig {
-    /// Creates a new encoding config with the appropriate number of primary and secondary source
-    /// symbols for the given number of shards.
+    /// Creates a new encoding config, given the number of shards.
+    ///
+    /// The number of shards determines the the appropriate number of primary and secondary source
+    /// symbols.
     ///
     /// The decoding probability is given by the [`decoding_safety_limit`]. See the documentation of
     /// [`decoding_safety_limit`] and [`source_symbols_for_n_shards`] for more details.
@@ -62,10 +64,6 @@ impl EncodingConfig {
     /// primary encoding (must be equal to or below `n_shards - 2f`), and `source_symbols_secondary`
     /// is the number of source symbols for the secondary encoding (must be equal to or below
     /// `n_shards - f`).
-    ///
-    /// # Returns
-    ///
-    /// The encoding configuration.
     ///
     /// # Panics
     ///
@@ -231,8 +229,7 @@ impl EncodingConfig {
         (self.n_shards_as_usize() * DIGEST_LEN * 2 + BlobId::LENGTH) as u64
     }
 
-    /// Returns an [`Encoder`] to perform a single primary or secondary encoding of the provided
-    /// data.
+    /// Returns an [`Encoder`] to perform a single primary or secondary encoding of the data.
     ///
     /// The `data` to be encoded _does not_ have to be aligned/padded. The `encoding_axis` specifies
     /// which encoding parameters the encoder uses, i.e., the parameters for either the primary or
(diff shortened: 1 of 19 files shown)


crates/walrus-core/src/encoding/config.rs
```diff
@@ -57,13 +57,10 @@ impl EncodingConfig {
...
     /// Creates a new encoding configuration for the provided system parameters.
     ///
-    /// # Arguments
-    ///
-    /// * `source_symbols_primary` - The number of source symbols for the primary encoding. This
-    ///   should be equal to or below `n_shards - 2f`, where `f` is the Byzantine parameter.
-    /// * `source_symbols_secondary` - The number of source symbols for the secondary encoding. This
-    ///   should be equal to or below `n_shards - f`.
-    /// * `n_shards` - The total number of shards.
+    /// In a setup with `n_shards` total shards -- among which `f` are Byzantine --
```
mlegner: Nit: Maybe replace "are" with "can be" or add "up to"?

crates/walrus-core/src/encoding/mapping.rs
```diff
@@ -21,19 +21,16 @@ pub enum SliverAssignmentError {
...
 ///
-/// # Arguments
-///
-/// * `pairs` - The slice of sliver pairs to rotate in place. This is assumed to be already of
-/// length equal to the number of shards.
-/// * `blob_id` - The Blob ID that produced the sliver pairs. It is interpreted as a big-endian
-/// unsigned integer, and the used to compute the amount for which to rotate the slice. The rotation
-/// is such that the last `blob_id % slice.len()` elements of the slice move to the front.
+/// The `blob_id` -- which is typically the blob ID of the blob that produced the sliver pairs -- is
+/// interpreted as a big-endian unsigned integer, and the used to compute the amount for which to
```
mlegner: Typo (pre-existing):
```suggestion
/// interpreted as a big-endian unsigned integer, and is then used to compute the amount for which to
```

crates/walrus-core/src/encoding/mapping.rs
```diff
@@ -124,20 +113,17 @@ fn rotation_offset(n_shards: NonZeroU16, blob_id: &BlobId) -> usize {
...
-/// Rotate the input slice in place, based on the rotation specified by the byte array.
-///
-/// # Arguments
+/// Rotate the input `slice` in place, based on the rotation specified by the byte array.
 ///
-/// * `slice` - The slice to rotate in place.
-/// * `rotation` - Interpreted as a big-endian unsigned integer, this is the amount for which to
-/// rotate the slice. The rotation is such that the last `rotation % slice.len()` elements of the
-/// slice move to the front.
+/// The `rotation` slice is interpreted as a big-endian unsigned integer, as it is the amount for
```
mlegner: Question: How/why is the "`rotation` slice interpreted as a big-endian unsigned integer"?

crates/walrus-core/src/encoding/config.rs
giac-mysten: this depends on the definition. Afaik the common one is that `f` is the number of byzantine nodes, and `n_shards >= 3f+1`... maybe i'll just add `where f is at most (n_shards-1)/3`

crates/walrus-core/src/encoding/mapping.rs
giac-mysten: not sure I understand the question. The idea is that a byte array is supplied, and this is interpreted as a number. This is because we want to use the blob ID to specify the rotation amount, which is `[u8; 32]`. I added a bit more detail.

mlegner: Ah, I think I misread the description. Now reading it again, it makes sense. Sorry about that.

crates/walrus-core/src/encoding/config.rs
mlegner: Hmm... My understanding is that the number of Byzantine nodes (or in this case shards) can vary over time and is not really known. So it doesn't really make sense to define the actual number of Byzantine nodes. However, we have to deal with them somehow, which is why we define an upper limit of them which is smaller than `n/3`. 

I guess in the end it doesn't matter in this case. Your suggestion sounds good to me.

PR #342, opened 2024-05-02
feat: recover metadata from other storage nodes

jpcsmith:
This PR adds fetching of metadata for blobs when encountering certified events. Still todo in this PR

- [x] Move helpers out of committee.rs
- [x] Move request logic into its own task
- [x] Enable test that the metadata is successfully fetched 
- [x] cleanup touched files 

Suggestion: Reviewing the PR commit-by-commit may be easiest.

### DIFF
--- a/crates/walrus-service/src/committee.rs
+++ b/crates/walrus-service/src/committee.rs
@@ -54,8 +54,7 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
     /// Returns the committee used by the service.
     fn committee(&self) -> &Committee;
 
-    // TODO(jsmith): Consider moving this to the service factory method, so that it's guaranteed to
-    // be called
+    // TODO(jsmith): Move to the service factory method, so that it's guaranteed to be called
     /// Excludes a member from calls made to the committee.
     ///
     /// An excluded member will not be contacted when making calls against the committee.
@@ -295,7 +294,8 @@ mod tests {
             let n_nodes = n_not_storing + n_byzantine_and_storing + n_honest_and_storing;
             let committee = test_utils::test_committee(&vec![1; n_nodes]);
 
-            let mut node_clients: Vec<_> = committee.members()
+            let mut node_clients: Vec<_> = committee
+                .members()
                 .iter()
                 .map(|member| (member.public_key.clone(), MockNodeClient::new()))
                 .collect();
@@ -346,7 +346,8 @@ mod tests {
             let n_nodes = 10;
             let committee = test_utils::test_committee(&vec![1; n_nodes]);
 
-            let mut node_clients: Vec<_> = committee.members()
+            let mut node_clients: Vec<_> = committee
+                .members()
                 .iter()
                 .map(|member| (member.public_key.clone(), MockNodeClient::new()))
                 .collect();
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -403,8 +403,8 @@ impl StorageNode {
             .expect("info to be present as it was just updated");
 
         if !blob_info.is_all_stored() {
-            // TODO(jsmith): Do not spawn if there is already a worker for this blob id
-            // TODO(jsmith): Handle cancellation.
+            // TODO(jsmith): Do not spawn if there is already a worker for this blob id (#366)
+            // TODO(jsmith): Handle cancellation. (#366)
             tokio::spawn(BlobSynchronizer::new(blob_id, blob_info, self).sync());
         }
 
@@ -422,11 +422,10 @@ struct BlobSynchronizer {
 
 impl BlobSynchronizer {
     fn new(blob_id: BlobId, blob_info: BlobInfo, node: &StorageNode) -> Self {
-        // TODO(jsmith): Use the methods from ServiceState for storing metadata etc.
         Self {
             blob_id,
             latest_state: blob_info,
-            // TODO(jsmith): Make storage cheaper to clone.
+            // TODO(jsmith): Make storage node cheaper to clone once we have epoch migration (#367)
             storage: node.storage.clone(),
             committee_service: node.committee_service.clone(),
             encoding_config: node.encoding_config.clone(),
--- a/crates/walrus-service/src/utils.rs
+++ b/crates/walrus-service/src/utils.rs
@@ -19,7 +19,7 @@ pub(crate) struct ExponentialBackoff<R> {
     rng: R,
 }
 
-impl ExponentialBackoff<()> {
+impl ExponentialBackoff<StdRng> {
     /// Maximum number of milliseconds to randomly add to the delay time.
     const MAX_RAND_OFFSET_MS: u64 = 1000;
 
@@ -73,7 +73,7 @@ impl<R: Rng> ExponentialBackoff<R> {
     fn random_offset(&mut self) -> Duration {
         Duration::from_millis(
             self.rng
-                .gen_range(0..=ExponentialBackoff::DEFAULT_RAND_OFFSET_MS),
+                .gen_range(0..=ExponentialBackoff::MAX_RAND_OFFSET_MS),
         )
     }
 }
@@ -209,7 +209,7 @@ mod tests {
             for (expected, actual) in expected.iter().zip(actual) {
                 let expected_min = *expected;
                 let expected_max =
-                    *expected + Duration::from_millis(ExponentialBackoff::DEFAULT_RAND_OFFSET_MS);
+                    *expected + Duration::from_millis(ExponentialBackoff::MAX_RAND_OFFSET_MS);
 
                 assert!(actual >= expected_min, "{actual:?} >= {expected_min:?}");
                 assert!(actual <= expected_max);


crates/walrus-service/src/committee.rs
```diff
@@ -68,35 +97,300 @@ where
...
+                    Ok(Ok(metadata)) => {
+                        tracing::debug!("metadata retrieved successfully");
+                        return metadata;
+                    }
+                    Ok(Err(err)) => tracing::debug!(?err, "metadata request failed"),
+                    Err(_) => tracing::debug!("request to node timed out"),
+                }
+            }
+
+            tracing::debug!("failed to retrieve metadata from an entire quorum, backing off");
```
karlwuest: Remark/Suggestion: I understand not logging every failure to request metadata in the inner loop as error, but if we fail to request it from an entire quorum, I think it would make sense to log it as an error.

crates/walrus-service/src/committee.rs
```diff
@@ -68,35 +97,300 @@ where
...
+            n_byzantine_and_storing: usize,
+            n_not_storing: usize,
+        ) -> TestResult {
+            let n_nodes = n_not_storing + n_byzantine_and_storing + n_honest_and_storing;
+            let committee = test_utils::test_committee(&vec![1; n_nodes]);
+
+            let mut node_clients: Vec<_> = (0..n_nodes)
+                .zip(committee.members())
+                .map(|(_, member)| (member.public_key.clone(), MockNodeClient::new()))
+                .collect();
```
karlwuest:
```suggestion
            let mut node_clients: Vec<_> = committee.members()
                .iter()
                .map(|member| (member.public_key.clone(), MockNodeClient::new()))
                .collect();
```

crates/walrus-service/src/committee.rs
```diff
@@ -68,35 +97,300 @@ where
...
+        #[tokio::test(start_paused = true)]
+        async fn does_not_block_indefinitely_on_requests() -> TestResult {
+            let n_honest_and_storing = 4;
+            let n_nodes = 10;
+            let committee = test_utils::test_committee(&vec![1; n_nodes]);
+
+            let mut node_clients: Vec<_> = (0..n_nodes)
+                .zip(committee.members())
+                .map(|(_, member)| (member.public_key.clone(), MockNodeClient::new()))
+                .collect();
```
karlwuest:
```suggestion
            let mut node_clients: Vec<_> = committee.members()
                .iter()
                .map(|member| (member.public_key.clone(), MockNodeClient::new()))
                .collect();
```

crates/walrus-service/src/node.rs
```diff
@@ -351,19 +365,89 @@ impl StorageNode {
...
+            .update_blob_info(&BlobEvent::Certified(event))?;
+
+        let blob_info = self
+            .storage
+            .get_blob_info(&blob_id)?
+            .expect("info to be present as it was just updated");
+
+        if !blob_info.is_all_stored() {
+            // TODO(jsmith): Do not spawn if there is already a worker for this blob id
+            // TODO(jsmith): Handle cancellation.
```
karlwuest: Suggestion: Add github issues for these to keep track of them, unless you're already working on this. Same applies to other `TODO`'s below.

crates/walrus-service/src/node.rs
```diff
@@ -351,19 +365,89 @@ impl StorageNode {
...
+}
+
+impl BlobSynchronizer {
+    fn new(blob_id: BlobId, blob_info: BlobInfo, node: &StorageNode) -> Self {
+        // TODO(jsmith): Use the methods from ServiceState for storing metadata etc.
+        Self {
+            blob_id,
+            latest_state: blob_info,
+            // TODO(jsmith): Make storage cheaper to clone.
+            storage: node.storage.clone(),
```
karlwuest: Suggestion: As far as I can see, we only ever need shared references to storage, except when creating the shards. Would it be an option to store an `Arc<Storage>` in `StorageNode` and here instead of cloning?

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -7,79 +7,181 @@
...
 pub enum BlobCertificationStatus {
-    Registered = 0,
-    Certified = 1,
-    Invalid = 2,
+    // INV: The indices of these must not change after release.
+    // BCS uses the index in encoding enums. Changing the index will change
+    // the database representation.
+    Registered,
+    Certified,
+    Invalid,
```
karlwuest, line 28: Suggestion: Keep the explicit assignment of the `u8` values, since the database representation relies on this.

crates/walrus-service/src/test_utils.rs
```diff
@@ -402,12 +427,14 @@ impl CommitteeService for StubCommitteeService {
...
-            .iter()
-            .any(|info| info.public_key == *identity)
+    fn exclude_member(&mut self, _identity: &PublicKey) {}
+
+    async fn get_and_verify_metadata(
+        &self,
+        _blob_id: &BlobId,
+        _encoding_config: &EncodingConfig,
+    ) -> VerifiedBlobMetadataWithId {
+        std::future::pending().await
```
karlwuest, line 437: Question/Suggestion: Does anything rely on this behaviour or is this function from the stub just not used anywhere? If it's the second, use `unimplemented!` instead.

crates/walrus-service/src/utils.rs
```diff
@@ -0,0 +1,324 @@
...
+pub(crate) struct ExponentialBackoff<R> {
+    min_backoff: Duration,
+    max_backoff: Duration,
+    sequence_index: u32,
+    rng: R,
+}
+
+impl ExponentialBackoff<()> {
+    /// Default amount of milliseconds to randomly add to the delay time.
+    const DEFAULT_RAND_OFFSET_MS: u64 = 1000;
```
karlwuest:
```suggestion
    /// Maximum number of milliseconds to randomly add to the delay time.
    const MAX_RAND_OFFSET_MS: u64 = 1000;
```

crates/walrus-service/src/utils.rs
```diff
@@ -0,0 +1,324 @@
...
+
+    fn next(&mut self) -> Option<Self::Item> {
+        Some(self.next_delay())
+    }
+}
+
+/// Randomly samples from the committee.
+///
+/// Iteration keeps an exclusive reference to the committee to allow reusing
+/// allocated buffers across multiple samples.
```
karlwuest: Question: What do you mean with this comment?

crates/walrus-service/src/utils.rs
```diff
@@ -0,0 +1,324 @@
...
+                .sample_quorum(&mut rng)
+                .collect();
+
+            assert!(!sample.is_empty());
+            assert!(*sample.iter().max().unwrap() < n_nodes);
+            assert_all_unique!(sample);
+        }
+
+        #[test]
+        fn sample_qorum_randomizes_indices_each_iteration() {
```
karlwuest:
```suggestion
        fn sample_quorum_randomizes_indices_each_iteration() {
```

crates/walrus-service/src/committee.rs
jpcsmith: Good point, will do.

crates/walrus-service/src/node.rs
jpcsmith: So `Arc<Storage>` would make it cheaper to clone. At epoch change, however, we will once again need exclusive references to the storage, I figured that this could be resolved then.

crates/walrus-service/src/storage/blob_info.rs
jpcsmith: Unfortunately it does not by default, with C-style enums BCS will only use the index, as documented in the last line [here](https://docs.rs/bcs/latest/bcs/#externally-tagged-enumerations). I also ran a quick test to confirm the above interpretation.

The `u8` representation would probably be more explicit/stable than the index representation, and we could add manual implementations for it, but this is not unique to the status enum: every type that we serialize to the database essentially makes that type's BCS representation part of our public API.  

My current thinking is that we probably need to add tests that the types representation does not change across releases in a non-backwards compatible way.

crates/walrus-service/src/test_utils.rs
jpcsmith: This is the intended behaviour for the stub: its requests never complete.

crates/walrus-service/src/utils.rs
jpcsmith: Thanks, should be removed, it's not longer valid.

crates/walrus-service/src/storage/blob_info.rs
karlwuest: >Unfortunately it does not by default, with C-style enums BCS will only use the index, as documented in the last line [here](https://docs.rs/bcs/latest/bcs/#externally-tagged-enumerations). I also ran a quick test to confirm the above interpretation.

I was not aware of that. In that case not tagging is preferable 👍

crates/walrus-service/src/committee.rs
jpcsmith:
```suggestion
            tracing::error!("failed to retrieve metadata from an entire quorum, backing off");
```

crates/walrus-service/src/utils.rs
jpcsmith:
```suggestion
```

PR #343, opened 2024-05-02
chore: update pre-commit hooks and improve update CI workflow

mlegner:
Effect of improvements is visible in https://github.com/MystenLabs/walrus/pull/345.

### DIFF
--- a/.github/workflows/updates.yml
+++ b/.github/workflows/updates.yml
@@ -29,7 +29,6 @@ jobs:
           --repo https://github.com/pre-commit/pre-commit-hooks
           --repo https://github.com/editorconfig-checker/editorconfig-checker.python
           --repo https://github.com/crate-ci/typos
-          --repo https://github.com/DevinR528/cargo-sort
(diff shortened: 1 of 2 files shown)


.github/workflows/updates.yml
```diff
@@ -44,9 +43,23 @@ jobs:
...
       - run: pre-commit run --all-files
       - uses: peter-evans/create-pull-request@v6.0.5
+        if: ${{ success() }}
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           branch: chore/update-pre-commit-hooks
           title: "chore: update pre-commit hooks"
           commit-message: "chore: update pre-commit hooks"
           body: Update pre-commit hooks to latest version.
+      - uses: peter-evans/create-pull-request@v6.0.5
```
asonnino, line 53: What is this new addition doing?

mlegner: The idea is that we still create the PR if there were errors during the update but add a warning to the PR body. Otherwise, we just don't notice that the workflow ran and failed.

PR #346, opened 2024-05-02
feat(client): better defaults for client communication

### LINKED ISSUE #265: Tune the Reqwest timeouts in the client
Currently the client just has simple timeouts on Reqwest connections, through the [timeout](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.timeout) parameter in the Request client.

giac-mysten:
I had another look at the Reqwest client configuration parameters, and these are the ones that I think make most sense to use and configure in the client. Please take a look at the defaults I selected and see if they make sense to you too :)

Closes #265

### DIFF
--- a/crates/walrus-service/example-client-config.yml
+++ b/crates/walrus-service/example-client-config.yml
@@ -1,19 +1,23 @@
+# Example configuration for the storage client.
+#
+# Default optional values are commented out.
+
 system_pkg: 0x134c017126a49714aa1de0a6f8809f743eaf53e8841f19cdf31c90f9fa222676
 system_object: 0x519cdbd1db849e9e8681e8550a1b903a090a6203f4dad4b9e201f9b54f28de99
-wallet_config: null
-communication_config:
-  concurrent_writes: null
-  concurrent_sliver_reads: null
-  concurrent_metadata_reads: 3
-  reqwest_config:
-    total_timeout:
-      secs: 10
-      nanos: 0
-    pool_idle_timeout: null
-    http2_keep_alive_timeout:
-      secs: 5
-      nanos: 0
-    http2_keep_alive_interval:
-      secs: 30
-      nanos: 0
-    http2_keep_alive_while_idle: true
+# wallet_config: null
+# communication_config:
+#   concurrent_writes: null
+#   concurrent_sliver_reads: null
+#   concurrent_metadata_reads: 3
+#   reqwest_config:
+#     total_timeout:
+#       secs: 10
+#       nanos: 0
+#     pool_idle_timeout: null
+#     http2_keep_alive_timeout:
+#       secs: 5
+#       nanos: 0
+#     http2_keep_alive_interval:
+#       secs: 30
+#       nanos: 0
+#     http2_keep_alive_while_idle: true
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -6,7 +6,7 @@ use std::{collections::HashMap, time::Instant};
 use anyhow::{anyhow, Result};
 use fastcrypto::{bls12381::min_pk::BLS12381AggregateSignature, traits::AggregateAuthenticator};
 use futures::Future;
-use reqwest::Client as ReqwestClient;
+use reqwest::{Client as ReqwestClient, ClientBuilder};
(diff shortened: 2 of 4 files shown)


crates/walrus-service/src/client/config.rs
```diff
@@ -33,10 +21,78 @@ pub struct Config {
...
+    #[serde(default = "ReqwestConfig::default")]
+    pub reqwest_config: ReqwestConfig,
+}
+
+impl Default for ClientCommunicationConfig {
+    fn default() -> Self {
+        // Enforce consistency with the serde defaults.
+        serde_yaml::from_str("").expect("can always deserialize default")
+    }
+}
```
giac-mysten: Is there a better way to ensure that the `Default` implementation is the same as the default serde deserialization?

crates/walrus-service/src/client/config.rs
```diff
@@ -33,10 +21,78 @@ pub struct Config {
     /// If set, this MUST be an absolute path.
     #[serde(deserialize_with = "deserialize_wallet_config")]
     pub wallet_config: Option<PathBuf>,
+    /// Configuration for the client's network communication.
+    #[serde(default = "ClientCommunicationConfig::default")]
```
jpcsmith:
```suggestion
    #[serde(default)]
```

crates/walrus-service/example-client-config.yml
jpcsmith, line 1: Remark: Could we have an example config where optional arguments are commented out? My issue is that if someone copies and uses this config, they unintentionally lock themselves to these parameters when we change the defaults.

crates/walrus-service/src/client/config.rs
```diff
@@ -33,10 +21,78 @@ pub struct Config {
...
+    /// shards `n`.
+    pub concurrent_writes: Option<usize>,
+    /// The maximum number of slivers the client requests in parallel. If `None`, the value is set
+    /// by the client to `n - 2f`, depending on the number of shards `n`.
+    pub concurrent_sliver_reads: Option<usize>,
+    /// The maximum number of nodes the client contacts to get the blob metadata in parallel.
+    #[serde(default = "default::concurrent_metadata_reads")]
+    pub concurrent_metadata_reads: usize,
+    /// The configuration for the `reqwest` client.
+    #[serde(default = "ReqwestConfig::default")]
```
jpcsmith:
```suggestion
    #[serde(default)]
```

crates/walrus-service/src/client.rs
```diff
@@ -58,17 +58,22 @@ impl Client<()> {
...
         sui_read_client: &impl ReadClient,
     ) -> Result<Self> {
-        let reqwest_client = ClientBuilder::new()
-            .timeout(config.connection_timeout)
+        tracing::debug!(?config, "running client");
+        let reqwest_client = config
+            .communication_config
+            .reqwest_config
+            .to_client_builder()
             .build()?;
```
jpcsmith: Nit: What about `reqwest_config.apply(&mut builder)` instead of generating the builder from the config?

crates/walrus-service/src/client/config.rs
jpcsmith: Yes, there is the container attribute [`#[serde(default)]`](https://serde.rs/container-attrs.html#default), which will take the default values for the field from the container's implementation of Default.

giac-mysten: Uhm i'm not sure I understand: seems to me that `#[serde(default)]` is used when deserializing. Here I want to implement the default trait for the struct in such a way that I can also use it in code, and have the same exact default as when deserializing.

Edit: maybe I get it, you're saying I should to the other way around. Makes sense.

jpcsmith: Yea, my apologies for not being clearer. Implement the Default trait with the values you would like to have as Default, then use that attribute on the containing struct, instead of marking fields with `#[serde(default = func)]`.

crates/walrus-service/src/client.rs
giac-mysten: Unfortunately the configuration consumes the builder, so it cannot be passed as `&mut`. But I changed to `apply` anyways.

PR #348, opened 2024-05-02
feat: implement API to submit inconsistency proofs

### LINKED ISSUE #331: Add API endpoint to submit inconsistency proof and return signature
Part of #257

karlwuest:
closes #331 
depends on #329

### DIFF
--- a/crates/walrus-core/Cargo.toml
+++ b/crates/walrus-core/Cargo.toml
@@ -7,7 +7,7 @@ edition.workspace = true
 license.workspace = true
 
 [features]
-test-utils = []
+test-utils = ["walrus-test-utils"]
 
 [dependencies]
 base64 = "0.22.0"
@@ -19,7 +19,7 @@ serde.workspace = true
 serde_with = { workspace = true, features = ["base64"] }
 thiserror.workspace = true
 tracing.workspace = true
-walrus-test-utils.workspace = true
+walrus-test-utils = { workspace = true, optional = true }
 
 [lints]
 workspace = true
--- a/crates/walrus-core/src/lib.rs
+++ b/crates/walrus-core/src/lib.rs
@@ -539,9 +539,9 @@ impl TryFrom<u8> for EncodingType {
 /// from their respective recovery symbols.
 #[derive(Debug)]
 pub enum InconsistencyProof<T: MerkleAuth = MerkleProof> {
-    /// Inconsistency Proof with symbols along the primary axis
+    /// Inconsistency proof for an encoding on the primary axis.
     Primary(PrimaryInconsistencyProof<T>),
-    /// Inconsistency Proof with symbols along the secondary axis
+    /// Inconsistency proof for an encoding on the secondary axis.
     Secondary(SecondaryInconsistencyProof<T>),
 }
 
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -9,7 +9,7 @@ use walrus_core::{
     encoding::{EncodingAxis, EncodingConfig, Sliver, SliverPair},
     inconsistency::InconsistencyProof,
     merkle::MerkleAuth,
-    messages::StorageConfirmation,
+    messages::{InvalidBlobIdAttestation, StorageConfirmation},
     metadata::{UnverifiedBlobMetadataWithId, VerifiedBlobMetadataWithId},
     BlobId,
     Epoch,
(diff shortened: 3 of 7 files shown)


crates/walrus-core/src/inconsistency.rs
```diff
@@ -73,10 +77,23 @@ pub enum InconsistencyVerificationError {
     SliverNotInconsistent,
 }
 
+/// An inconsistency proof for an encoding on the primary axis.
+/// The symbols on this proof originate from the secondary axis.
+pub type PrimaryInconsistencyProof<U> = InconsistencyProof<Secondary, U>;
```
karlwuest: @mlegner is my understanding here correct? 
In particular that the axis of the `InconsistencyProof` is the axis of the source slivers, not the target sliver?

mlegner: Currently, the axis of an inconsistency proof is the axis of the recovery symbols, which is the axis of the target sliver.

karlwuest: Ah, ok, then I misunderstood this. Will change it accordingly.

crates/walrus-core/Cargo.toml
```diff
@@ -19,6 +19,7 @@ serde.workspace = true
 serde_with = { workspace = true, features = ["base64"] }
 thiserror.workspace = true
 tracing.workspace = true
+walrus-test-utils.workspace = true
```
mlegner: Suggestion: Mark this as "optional" and add it as a dependency of the `test-utils` feature:
```suggestion
walrus-test-utils = { workspace = true, optional = true }
```
```rust
[features]
test-utils = ["walrus-test-utils"]
```

crates/walrus-core/src/lib.rs
```diff
@@ -524,6 +530,38 @@ impl TryFrom<u8> for EncodingType {
...
+///
+/// Can be either a [`PrimaryInconsistencyProof`] or a [`SecondaryInconsistencyProof`],
+/// proving that either a [`PrimarySliver`] or a [`SecondarySliver`] cannot be recovered
+/// from their respective recovery symbols.
+#[derive(Debug)]
+pub enum InconsistencyProof<T: MerkleAuth = MerkleProof> {
+    /// Inconsistency Proof with symbols along the primary axis
+    Primary(PrimaryInconsistencyProof<T>),
+    /// Inconsistency Proof with symbols along the secondary axis
+    Secondary(SecondaryInconsistencyProof<T>),
```
mlegner: I believe the docstrings here are not consistent with the rest.

crates/walrus-service/src/node.rs
```diff
@@ -100,6 +109,16 @@ pub enum StoreSliverError {
     Internal(#[from] anyhow::Error),
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum InconsistencyProofError {
+    #[error("Missing metadata for {0:?}")]
```
mlegner: Nit: I think we could use the `Display` format of the blob ID here. (This also applies to a bunch of other errors where we currently use the `Debug` format.

crates/walrus-service/src/server.rs
```diff
@@ -320,6 +324,44 @@ impl<S: ServiceState + Send + Sync + 'static> UserServer<S> {
...
+                );
+                ServiceResponse::success(StatusCode::OK, attestation)
+            }
+            Err(InconsistencyProofError::MissingMetadata(_)) => {
+                tracing::debug!("No metadata found for {blob_id:?}");
+                ServiceResponse::not_found()
+            }
+            Err(InconsistencyProofError::ProofVerificationError(err)) => {
+                tracing::error!("Error when verifying inconsistency proof: {err}");
+                ServiceResponse::error(
```
mlegner, line 353: Suggestion: This is a somewhat expected "error", so I suggest to reduce this to a `warn`. I can also do this as part of #358 (which I will rebase on your branch).

crates/walrus-sdk/src/client.rs
```diff
@@ -237,6 +247,43 @@ impl Client {
...
+        let encoded_proof =
+            bcs::to_bytes(&inconsistency_proof).expect("internal types to be bcs encodable");
+
+        let request = self.inner.put(url).body(encoded_proof);
+        request
+            .send()
+            .await
+            .map_err(Kind::Reqwest)?
+            .response_error_for_status()
+            .await?;
```
mlegner: Suggestion: Return the signed inconsistency message obtained from the other node.

crates/walrus-service/src/node.rs
```diff
@@ -136,6 +155,13 @@ pub trait ServiceState {
         blob_id: &BlobId,
     ) -> impl Future<Output = Result<Option<StorageConfirmation>, anyhow::Error>> + Send;
 
+    /// Verify an inconsistency proof and provide a signed attestation for it, if valid.
```
mlegner: Nit: We generally use the indicative, not the imperative for docstrings.

crates/walrus-core/src/lib.rs
karlwuest: Do you just mean the missing dots at the end (which I will add) or something more fundamental?

mlegner: I meant that the symbols are computed from the orthogonal slivers. However, I may just have misunderstood the description ("along the {} axis").

karlwuest: Ok, this was meant to refer to the reconstruction axis (i.e. the axis of the target sliver). I'll clarify this 👍

crates/walrus-sdk/src/client.rs
karlwuest: Good catch, no idea how I forgot to return this 🤦‍♂️

PR #358, opened 2024-05-06
fix: clean up logging messages and include relevant metadata

### LINKED ISSUE #357: Reduce log level for "expected" problems
On the client, we log errors when individual storage nodes are not reachable. As this is an expected problem, we should reduce this to warning or even info level and only log errors if more than `f` nodes are unreachable.

mlegner:
Closes #357.

### DIFF
--- a/crates/walrus-service/src/client/communication.rs
+++ b/crates/walrus-service/src/client/communication.rs
@@ -119,7 +119,7 @@ impl<'a> NodeCommunication<'a> {
     where
         Sliver<T>: TryFrom<SliverEnum>,
     {
-        tracing::debug!(blob_id = %metadata.blob_id(), "retrieving verified sliver");
+        tracing::debug!(%shard_index, sliver_type = T::NAME, "retrieving verified sliver");
         let sliver_pair_index = shard_index.to_pair_index(self.n_shards(), metadata.blob_id());
         let sliver = self
             .client
@@ -142,7 +142,7 @@ impl<'a> NodeCommunication<'a> {
         metadata: &VerifiedBlobMetadataWithId,
         pairs: Vec<SliverPair>,
     ) -> NodeResult<SignedStorageConfirmation, StoreError> {
-        tracing::debug!(blob_id = %metadata.blob_id(), "storing metadata and sliver pairs",);
+        tracing::debug!("storing metadata and sliver pairs",);
 
         let result = async {
             // TODO(giac): add retry for metadata.
--- a/crates/walrus-service/src/server.rs
+++ b/crates/walrus-service/src/server.rs
@@ -313,7 +313,7 @@ impl<S: ServiceState + Send + Sync + 'static> UserServer<S> {
 
         match state.store_sliver(&blob_id, sliver_pair_index, &sliver) {
             Ok(()) => {
-                tracing::debug!(%blob_id, %sliver_type, "stored sliver");
+                tracing::debug!(%blob_id, %sliver_type, %sliver_pair_index, "stored sliver");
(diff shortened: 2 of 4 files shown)


crates/walrus-service/src/client/communication.rs
```diff
@@ -100,7 +100,7 @@ impl<'a> NodeCommunication<'a> {
         &self,
         blob_id: &BlobId,
     ) -> NodeResult<VerifiedBlobMetadataWithId, NodeError> {
-        tracing::debug!("retrieving metadata");
```
giac-mysten, line 103: Important: the logging of the blob id (specifically, blob id prefix) is already instantiated in the span of the calling functions (read blob and store blob). Therefore, all the calls in `communication.rs` are already annotated with info on the blob id. 

Example output:

```
2024-05-06T14:24:29.326029Z DEBUG read_blob{blob_id_prefix="-7_nNXvF..."}:node{index=1 epoch=0 pk_prefix="o5dg+Pfj..."}: walrus_service::client::communication: retrieving metadata
```

This is the result of a quite long back-and-forth with @jpcsmith on the internals of `tracing` 😬 if you have other suggestions i'd be interested to hear them, but for the moment I would not log the blob ID here.

mlegner: Ah, sorry, my bad. Thanks for the comment. I can actually do something similar in the server.

mlegner: I've made some further improvements to the tracing events and spans in 10c51bc22d593ca8c1adc20468497d13c52ac2a7. Concretely, I've made them more consistent and also tried to include relevant metadata wherever possible and, where applicable, using the `Display` format.

giac-mysten: uhm quick question again here: you are still logging the blob id it seems, although the blob ID prefix is still added to the trace from the parent span. The result is e.g., the following:
```
2024-05-08T14:17:59.243921Z DEBUG read_blob{blob_id_prefix="-7_nNXvF..."}:node{index=3 epoch=0 pk_prefix="qfs81DGg..."}: walrus_service::client::communication: retrieving verified sliver blob_id=-7_nNXvFU_3qZVu3OH1yycRG7LZlyn1_UxEDCDDqGGU
```
so the blob id is logged twice.

Is this wanted?

crates/walrus-service/src/server.rs
```diff
@@ -153,26 +154,28 @@ impl<S: ServiceState + Send + Sync + 'static> UserServer<S> {
             .await
     }
 
+    #[tracing::instrument(level = Level::ERROR, skip_all, fields(blob_id = %blob_id))]
```
giac-mysten, line 157: Thanks for adding this, it is definitely an improvement.

Question: this means that the blob ID will always be logged in full in every log message. Do we want this?

Suggestion (maybe for followup): what about logging the prefix as in the client? and the full blob id only on errors?

mlegner: Hmm... I don't think it makes sense to differentiate: Under normal conditions, we anyway only log messages of info level or higher. OTOH, when debugging, we are not really worried about the amount of logging output. Generally more is better in that case.

crates/walrus-service/src/client/communication.rs
mlegner: This is not intentional, I will fix this. 😞

mlegner: Hmm... I can't reproduce this specific message (I don't have a `blob_id_prefix` anymore) but I found and cleaned up some other duplicated fields.

Issue #360, opened 2024-05-06
Implement blocksites from walrus


giac-mysten:
- [x] Implement the aggregator
- [x] Add blob id to smart contracts, remove the bytes of the resource
- [x] Change the service worker to support fetching the aggregator
- [x] Change the site builder to publish the new/updated resources to walrus
- [x] Add back SuiNS publishing
- [ ] UX: improve the site builder defaults
- [ ] UX: improve the blocksite error messages
- [ ] UX: improve the site builder output
- [ ] UX: add "dry run" of the blocksite creation price 
- [ ] UX: add confirmation for the transaction

PR #361, opened 2024-05-06
feat: implement the aggregator

### LINKED ISSUE #350: Implement the aggregator (cache)

giac-mysten:
The aggregator simply adds an HTTP endpoint in front of a walrus client.

Closes #350 
Contributes to #360

### DIFF
--- a/crates/walrus-service/src/aggregator.rs
+++ b/crates/walrus-service/src/aggregator.rs
@@ -17,7 +17,7 @@ use tower_http::trace::TraceLayer;
 use walrus_core::encoding::Primary;
 
 use crate::{
-    client::{string_prefix, Client},
+    client::{Client, ClientErrorKind},
     server::BlobIdString,
 };
 
@@ -60,10 +60,7 @@ impl<T: Send + Sync + 'static> AggregatorServer<T> {
     ) -> Response {
         match client.read_blob::<Primary>(&blob_id).await {
             Ok(blob) => {
-                tracing::debug!(
-                    blob_id_prefix=?string_prefix(&blob_id),
-                    "successfully retrieved blob"
-                );
+                tracing::debug!(?blob_id, "successfully retrieved blob");
                 let mut response = (StatusCode::OK, blob).into_response();
                 // Allow requests from any origin, s.t. content can be loaded in browsers.
                 response
@@ -71,13 +68,22 @@ impl<T: Send + Sync + 'static> AggregatorServer<T> {
                     .insert(ACCESS_CONTROL_ALLOW_ORIGIN, HeaderValue::from_static("*"));
                 response
             }
-            Err(e) => {
-                tracing::error!(
-                    error=?e,
-                    blob_id_prefix=?string_prefix(&blob_id),
-                    "error retrieving blob"
-                );
-                (StatusCode::NOT_FOUND, "Not found").into_response()
+            Err(error) => {
+                match error.kind() {
+                    // TODO(giac): once issues #362 and #363 are resolved, this logging can be
+                    // further improved, and distinguish network errors from missing metadata.
+                    ClientErrorKind::NoMetadataReceived => {
+                        tracing::info!(
+                            ?blob_id,
+                            "could not retrieve the metadata; the blob may not exist"
+                        );
+                        StatusCode::NOT_FOUND.into_response()
+                    }
+                    _ => {
+                        tracing::error!(error=?error, ?blob_id, "error retrieving blob");
+                        StatusCode::INTERNAL_SERVER_ERROR.into_response()
+                    }
+                }
             }
         }
     }
--- a/crates/walrus-service/bin/aggregator.rs
+++ b/crates/walrus-service/bin/aggregator.rs
@@ -3,21 +3,17 @@
 
 //! The aggregator binary.
 
-mod cli_utils;
-
 use std::{net::SocketAddr, path::PathBuf, sync::Arc};
 
 use anyhow::{Context, Result};
 use clap::Parser;
-use cli_utils::load_configuration;
 use sui_sdk::SuiClientBuilder;
 use walrus_service::{
     aggregator::AggregatorServer,
+    cli_utils::{error, load_configuration},
     client::{Client, Config},
 };
 use walrus_sui::client::SuiReadClient;
-
-use crate::cli_utils::error;
 
 #[derive(Parser, Debug, Clone)]
 #[clap(rename_all = "kebab-case")]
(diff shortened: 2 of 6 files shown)


crates/walrus-service/src/client.rs
```diff
@@ -280,8 +280,10 @@ impl<T> Client<T> {
         let comms = self.node_communications();
         // Create requests to get all slivers from all nodes.
         let futures = comms.iter().flat_map(|n| {
-            n.node.shard_ids.iter().map(|s| {
-                n.retrieve_verified_sliver::<U>(metadata, *s)
+            // NOTE: the cloned here is needed because otherwise the compiler complains about the
+            // lifetimes of `s`.
```
mlegner, line 301: Remark: Really cryptic compiler error without the `cloned`... 🤷

crates/walrus-service/bin/cli_utils.rs
mlegner, line 1: Question: Why have this here instead of under `src`? There, the `#[allow(dead_code)]` would also not be necessary.

crates/walrus-service/src/aggregator.rs
```diff
@@ -0,0 +1,84 @@
...
+    client::{string_prefix, Client},
+    server::BlobIdString,
+};
+
+/// The path to get the blob at the given blob Id.
+pub const BLOB_ENDPOINT: &str = "/v1/:blobId";
+
+/// The aggregator server
+///
+/// Exposes and HTTP endpoint to request blobs by blob ID.
```
mlegner, line 29: Nit:
```suggestion
/// The aggregator server.
///
/// Exposes an HTTP endpoint to request blobs by blob ID.
```

crates/walrus-service/src/aggregator.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    async fn retrieve_blob(
+        State(client): State<Arc<Client<T>>>,
+        Path(BlobIdString(blob_id)): Path<BlobIdString>,
+    ) -> Response {
+        match client.read_blob::<Primary>(&blob_id).await {
+            Ok(blob) => {
+                tracing::debug!(
+                    blob_id_prefix=?string_prefix(&blob_id),
```
mlegner: Suggestion: As this is just a debug message, why not print the full blob ID? It's not that long. (Same below.)

crates/walrus-service/src/aggregator.rs
```diff
@@ -0,0 +1,84 @@
...
+                );
+                let mut response = (StatusCode::OK, blob).into_response();
+                // Allow requests from any origin, s.t. content can be loaded in browsers.
+                response
+                    .headers_mut()
+                    .insert(ACCESS_CONTROL_ALLOW_ORIGIN, HeaderValue::from_static("*"));
+                response
+            }
+            Err(e) => {
+                tracing::error!(
```
mlegner: Suggestion: Only log an "error" if there is something wrong with the internals. If the blob simply doesn't exist (which is a more likely case), just log an info or warning.

crates/walrus-service/src/client.rs
giac-mysten: yeah, I couldn't figure out what was the problem. I read somewhere that these errors often arise from compiler bugs/limitations, but couldn't get to the bottom of it 😞

crates/walrus-service/bin/cli_utils.rs
giac-mysten: uhm I thought that since these are utils for the binaries they would fit better in bin. I can move it though.

crates/walrus-service/src/aggregator.rs
giac-mysten: good point. I will then first merge #339 , so that I have information on the error.

One broader point is: how do we know if a blob does not exist? I guess the only signal we have the number of nodes (weighted by stake) that do not reply with the metadata.

In steady state, if `f+1` nodes return "not found", then we know that the blob ID does not exists.
However, if the blob has been recently certified by `2f+1`, `f` of which are malicious, and the other `f` honest nodes have not synced yet, this is not sufficient and we have to check on `2f+1` stake.

I opened  #362 to track the issue of exiting early in the client. 
Thoughts?

crates/walrus-service/bin/cli_utils.rs
jpcsmith, line 1: Suggestion: Move under src/, as this approach compiles cli_utils separately into binaries that use it, instead of linking against it in the crate.

mlegner: Hmm... I guess the advantage is that we can selectively include it in some binaries but not others. OTOH, the same would be true for the tool-specific code in the library (`src/aggregator.rs`, `src/server.rs`, etc.).

I have a slight preference for keeping the utils in `src`, but no strong one. We can also move it later if we want, so not blocking.

crates/walrus-service/src/aggregator.rs
mlegner: That's a good point. I think your reasoning is correct if we only have the blob ID without any additional information. However, in many cases we will have the on-chain data available, so we know the blob's state and also when we expect all nodes to have it (we still have to define the precise grace period though).

jpcsmith: If using the not-found like this, we will probably need to disambiguate generic not-found errors from the server with those that indicate the endpoint was correct, but the blob-id was not found.

crates/walrus-service/src/client.rs
jpcsmith: Suggestion: use `copied()` instead of `cloned()`

crates/walrus-service/src/aggregator.rs
giac-mysten: > However, in many cases we will have the on-chain data available, so we know the blob's state and also when we expect all nodes to have it (we still have to define the precise grace period though).

That's true. I was assuming we don't want the aggregator to perform RPC requests on behalf of the client to keep it light-weight. Should we add RPC checks at some point?

Also the path Sui resource object -> blob ID is trivial, but blob ID -> sui object may be quite a complex resolution, right? In the worst case it may require listening to all events and keeping a list.

> If using the not-found like this, we will probably need to disambiguate generic not-found errors from the server with those that indicate the endpoint was correct, but the blob-id was not found.

you mean on the walrus node server right? I agree. Opened #363

crates/walrus-service/bin/cli_utils.rs
giac-mysten: ack, will do

crates/walrus-service/src/aggregator.rs
mlegner: > > However, in many cases we will have the on-chain data available, so we know the blob's state and also when we expect all nodes to have it (we still have to define the precise grace period though).
> 
> That's true. I was assuming we don't want the aggregator to perform RPC requests on behalf of the client to keep it light-weight. Should we add RPC checks at some point?
> 
> Also the path Sui resource object -> blob ID is trivial, but blob ID -> sui object may be quite a complex resolution, right? In the worst case it may require listening to all events and keeping a list.

Agreed. I think for now we should work under the assumption that we don't know anything about the blob and use a threshold of `2f+1` not-found responses to conclude the blob does not exist.

PR #364, opened 2024-05-07
refactor: remove `std` from `walrus-core`

### LINKED ISSUE #359: Remove `std` dependence in `walrus-core`
The motivation for this is to make this compilable for WASM (`wasm32-unknown-unknown` target).

mlegner:
Most code changes are simply changing an `std::` import to a `core::` or `alloc::` import.

The only "real" change (though also without any functional effects) is moving the `keys` module and associated functions to `walrus-service`.

We still use `std` for `std::error::Error`, so this doesn't make `walrus-core` a true `no_std` crate. However, `std::error::Error` is supported on WASM, which is the reason for this change.

Also add lints to encourage importing resources from `alloc` and `core` instead of `std` (if available).

Closes #359.


crates/walrus-core/src/lib.rs
```diff
@@ -41,15 +48,10 @@ pub mod bft;
 pub mod encoding;
 pub mod inconsistency;
 pub mod merkle;
+pub mod messages;
 pub mod metadata;
 pub mod utils;
 
-pub mod keys;
-pub use keys::ProtocolKeyPair;
```
jpcsmith, line 48: Suggestion: How about just allowing these to be behind a feature which is on by default, but can be excluded with `--no-default-features`? 

Any tooling that wants to load or manipulate keys/configs will need `keys`.

mlegner: Yes, that works.

jpcsmith: I am also fine with the change since you already made it, but we may eventually need to move it somewhere accessible by the SDK.

mlegner: I've moved it back to core for now. 👍 

We can add the feature when we actually compile for WASM.

Issue #353, opened 2024-05-05
Make end-to-end test with crashed nodes more robust


mlegner:
The full end-to-end test with crashed nodes (`test_store_and_read_blob_with_crash_failures `) fails sometimes. See, for example, this CI job: https://github.com/MystenLabs/walrus/actions/runs/8955352006/job/24595957829

We should figure out why and fix it.

PR #365, opened 2024-05-07
fix: prevent duplicate assignment of local socket addresses

karlwuest:
contributes to #353 

The assignment of local socket addresses could be the source of the test referenced in #353 randomly failing.
In particular, we only bind to the port, but don't connect. This seems to work on Mac to put the address into `TIME-WAIT`, but on Linux this is not sufficient (see [this CI run](https://github.com/MystenLabs/walrus/actions/runs/8989262284/job/24691989183)). Instead, we have to first create a connection and then drop it.
This may have caused a port to be used both by a sui validator and a walrus storage node (because we first created the walrus storage node addresses and then the sui test cluster), s.t. the validator was not able to bind to it and crashed, which then made the consensus fail.

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -450,10 +450,10 @@ impl CommitteeService for StubCommitteeService {
 
 /// Returns a socket address that is not currently in use on the system.
 pub fn unused_socket_address() -> SocketAddr {
-    fallible_unused_socket_address().expect("unused socket address to be available")
+    try_unused_socket_address().expect("unused socket address to be available")
 }
 
-fn fallible_unused_socket_address() -> anyhow::Result<SocketAddr> {
+fn try_unused_socket_address() -> anyhow::Result<SocketAddr> {
     let listener = std::net::TcpListener::bind("127.0.0.1:0")?;
     let address = listener.local_addr()?;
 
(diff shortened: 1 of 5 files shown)


crates/walrus-service/src/test_utils.rs
```diff
@@ -417,8 +422,17 @@ impl CommitteeService for StubCommitteeService {
 
 /// Returns a socket address that is not currently in use on the system.
 pub fn unused_socket_address() -> SocketAddr {
-    let listener = std::net::TcpListener::bind("127.0.0.1:0").unwrap();
-    listener.local_addr().unwrap()
+    fallible_unused_socket_address().expect("unused socket address to be available")
+}
+
+fn fallible_unused_socket_address() -> anyhow::Result<SocketAddr> {
```
```suggestion
fn try_unused_socket_address() -> anyhow::Result<SocketAddr> {
```

crates/walrus-sui/src/test_utils.rs
```diff
@@ -55,6 +56,11 @@ pub fn get_default_invalid_certificate(blob_id: BlobId, epoch: Epoch) -> Invalid
     }
 }
 
+/// Creates and returns a sui test cluster.
+pub async fn create_sui_test_cluster() -> TestCluster {
```
jpcsmith:
```suggestion
pub async fn sui_test_cluster() -> TestCluster {
```

PR #370, opened 2024-05-08
feat(client): propagate information about inexistent blobs IDs

giac-mysten:
Other changes:

- The order in which nodes are contacted is randomized;
- the client exits early when retrieving metadtata;
- the client exits early when retrieving slivers;

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -68,10 +68,19 @@ impl Client<()> {
             .apply(ClientBuilder::new())
             .build()
             .map_err(ClientError::other)?;
+
+        // Get the committee, and check that there is at least one shard per node.
         let committee = sui_read_client
             .current_committee()
             .await
             .map_err(ClientError::other)?;
+        for node in committee.members() {
+            ensure!(
+                !node.shard_ids.is_empty(),
+                ClientErrorKind::InvalidConfig.into(),
+            );
+        }
+
         let encoding_config = EncodingConfig::new(committee.n_shards());
         // Try to store on n-f nodes concurrently, as the work to store is never wasted.
         let concurrent_writes = config
@@ -192,7 +201,7 @@ impl<T> Client<T> {
         pairs: Vec<SliverPair>,
     ) -> Result<ConfirmationCertificate, ClientError> {
         let mut pairs_per_node = self.pairs_per_node(metadata.blob_id(), pairs);
-        let comms = self.node_communications();
+        let comms = self.node_communications()?;
         let mut requests = WeightedFutures::new(comms.iter().map(|n| {
             n.store_metadata_and_pairs(
                 metadata,
@@ -299,7 +308,7 @@ impl<T> Client<T> {
     {
         // TODO(giac): optimize by reading first from the shards that have the systematic part of
         // the encoding. Currently the read order is randomized.
-        let comms = self.node_communications();
+        let comms = self.node_communications()?;
         // Create requests to get all slivers from all nodes.
         let futures = comms.iter().flat_map(|n| {
             // NOTE: the cloned here is needed because otherwise the compiler complains about the
@@ -435,7 +444,7 @@ impl<T> Client<T> {
         &self,
         blob_id: &BlobId,
     ) -> Result<VerifiedBlobMetadataWithId, ClientError> {
-        let comms = self.node_communications_quorum();
+        let comms = self.node_communications_quorum()?;
         let futures = comms.iter().map(|n| {
             n.retrieve_verified_metadata(blob_id)
                 .instrument(n.span.clone())
@@ -469,7 +478,7 @@ impl<T> Client<T> {
         &'a self,
         index: usize,
         node: &'a StorageNode,
-    ) -> NodeCommunication {
+    ) -> Result<NodeCommunication, ClientError> {
         NodeCommunication::new(
             index,
             self.committee.epoch,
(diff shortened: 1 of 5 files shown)


crates/walrus-service/src/client/communication.rs
```diff
@@ -88,6 +88,18 @@ impl<'a> NodeCommunication<'a> {
...
+    /// Returns the number of shards owned by the node.
+    pub fn n_owned_shards(&self) -> NonZeroU16 {
+        NonZeroU16::new(
+            self.node
+                .shard_ids
+                .len()
+                .try_into()
+                .expect("the number of shards is capped"),
+        )
+        .expect("each node has >0 shards")
```
mlegner, line 109: Remark: AFAICT, this is currently only enforced in the [Move code](https://github.com/MystenLabs/walrus/blob/gg/client/retrieve-metadata-improved/contracts/blob_store/sources/bls_aggregate.move#L34). It might make sense to annotate the `StorageNode` struct with this invariant and check it in the constructors.

crates/walrus-service/src/client/error.rs
```diff
@@ -70,7 +70,13 @@ pub enum ClientErrorKind {
     /// The client could not retrieve enough slivers to reconstruct the blob.
     #[error("could not retrieve enough slivers to reconstruct the blob")]
     NotEnoughSlivers,
+    /// The client received enough "not found" messages to confirm that the blob ID does not exist.
+    #[error("the blob ID does not exist")]
+    BlobIdDoesNotExist,
     /// The client could not retrieve the metadata from the storage nodes.
+    ///
+    ///This error differs from the [`ClientErrorKind::BlobIdDoesNotExist`] version in the fact that
```
mlegner: Nit:
```suggestion
    /// This error differs from the [`ClientErrorKind::BlobIdDoesNotExist`] version in the fact that
```

crates/walrus-sui/src/types.rs
```diff
@@ -267,8 +267,9 @@ pub struct Committee {
...
 
 impl Committee {
-    /// Create a new committee for `epoch` consisting of `members`. `members` must contain at
-    /// least one storage node holding at least one shard.
+    /// Create a new committee for `epoch` consisting of `members`.
+    ///
+    /// `members` must contain at least one storage node holding at least one shard.
     pub fn new(members: Vec<StorageNode>, epoch: Epoch) -> Result<Self, InvalidCommittee> {
         let n_shards = members
             .iter()
```
mlegner: Related to my other comment: We could check here that all members have at least one shard.

crates/walrus-service/src/client.rs
```diff
@@ -394,11 +448,22 @@ impl<T> Client<T> {
...
-        Ok(metadata)
+
+        let mut n_not_found = 0;
+        for NodeResult(_, weight, _, result) in requests.into_results() {
+            match result {
+                Ok(metadata) => return Ok(metadata),
+                Err(error) => {
+                    if error.is_status_not_found() && {
+                        n_not_found += weight;
+                        self.committee.is_quorum(n_not_found)
```
mlegner, line 466: Remark: I don't want to start a long discussion about the BFT computations again (😇), but this is slightly inconsistent with the docstring above: This checks if `n_not_found` is at least `n-f`, not `2f+1`.

crates/walrus-service/src/client.rs
```diff
@@ -416,34 +481,63 @@ impl<T> Client<T> {
...
-            self.committee
-                .members()
-                .iter()
-                .map(|n| Vec::with_capacity(n.shard_ids.len())),
-        );
+    fn pairs_per_node(
+        &self,
+        blob_id: &BlobId,
+        pairs: Vec<SliverPair>,
+    ) -> HashMap<usize, Vec<SliverPair>> {
```
mlegner, line 525: Question: Is my understanding correct, that this change is necessary because of the shuffling of the nodes?

giac-mysten: Yes indeed. But in general I also think it improves the code a bit.

crates/walrus-service/src/client.rs
giac-mysten: good catch!

giac-mysten: in this case I do need `2f+1` and not `n-f`, right?

mlegner: So we have a confirmation from `n-f` nodes, so `n-2f` honest ones. To be sure to contact at least one of those, we have to contact `n-(n-2f)+1 = 2f+1` nodes. So yes, we want `2f+1` here.

giac-mysten: and how do we call the function that checks this? 😬 `2f+1` would be the most usual "safety quorum", while `n-f` is the one we use in walrus because of the encoding 🤔 suggestions?

giac-mysten: i would call the check of `2f+1` to `is_quorum`, and rename the `n-f` to `is_above_min_honest`?

mlegner: > i would call the check of `2f+1` to `is_quorum`, and rename the `n-f` to `is_above_min_honest`?

Maybe `is_at_least_min_honest`? Otherwise sounds good. 👍

giac-mysten: ah nice, I had just changed to that exact formulation ahah.
changed in 7a4edb3 , ptal

crates/walrus-service/src/client/communication.rs
mlegner: Thanks for the added check in the Sui types. There still seems to be a disconnect, for example, there is no check or other restriction in the `NodeCommunication::new` function nor the `StorageNode` struct. Maybe the `NodeCommunication::new` function would be a good place as we cannot easily enforce anything on `StorageNode` with its public fields.

giac-mysten: good point. I added two checks now. One in `NodeCommunication::new` and one in `Client::new`, so we check that nodes have `>0` shards immediately when the committee is loaded from chain.

PR #378, opened 2024-05-10
feat(orchestrator): Merge local and remote testbed utilities

### LINKED ISSUE #375: Simplify storage node CLI
There is a lot of code repetition between the dry run genesis used to create a local and remote testbed. Also ensure all nodes use the same ports except when running a local testbed (this makes PE live easier)

asonnino:
* Remove code duplication to create genesis
* Book all storage nodes with the same port when deploying remote testbeds

closes #375

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,7 +69,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.15",
+ "getrandom 0.2.14",
  "once_cell",
  "version_check",
 ]
@@ -158,15 +158,15 @@ dependencies = [
  "rand 0.8.5",
  "rcgen",
  "ring 0.16.20",
- "rustls 0.21.12",
+ "rustls 0.21.11",
  "rustls-webpki 0.101.7",
  "serde",
  "serde_json",
- "socket2 0.5.7",
+ "socket2 0.5.6",
  "tap",
  "thiserror",
  "tokio",
- "tokio-util 0.7.11",
+ "tokio-util 0.7.10 (registry+https://github.com/rust-lang/crates.io-index)",
  "tower",
  "tracing",
  "x509-parser",
@@ -178,7 +178,7 @@ version = "0.0.0"
 source = "git+https://github.com/mystenlabs/anemo.git?rev=26d415eb9aa6a2417be3c03c57d6e93c30bd1ad7#26d415eb9aa6a2417be3c03c57d6e93c30bd1ad7"
 dependencies = [
  "prettyplease 0.1.25",
- "proc-macro2 1.0.82",
+ "proc-macro2 1.0.81",
  "quote 1.0.36",
  "syn 1.0.109",
 ]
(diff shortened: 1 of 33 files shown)


Cargo.lock
mlegner, line 1: Important: This should not be deleted.

crates/walrus-service/bin/main.rs
```diff
@@ -70,11 +76,11 @@ enum Commands {
...
 
-    /// Generate the configuration files to run a local testbed of storage nodes.
-    GenerateDryRunLocalConfigs(GenerateDryRunLocalConfigsArgs),
+    /// Deploy the Walrus system contract on the Sui network.
+    DeploySystemContract(DeploySystemContractArgs),
 
-    /// Generate the configuration files to run a remote testbed of storage nodes.
-    GenerateDryRunRemoteConfigs(GenerateDryRunRemoteConfigsArgs),
+    /// Generate the configuration files to run a testbed of storage nodes.
+    GenerateDryRunConfigs(GenerateDryRunConfigsArgs),
```
mlegner, line 83: Remark: Now that I think about this some more, I feel that those don't actually belong in the storage-node CLI. They would fit better into a separate "admin" CLI. Could be changed in a separate PR.

crates/walrus-service/src/testbed.rs
```diff
@@ -39,146 +40,60 @@ pub fn node_config_name_prefix(node_index: u16, committee_size: NonZeroU16) -> S
...
-/// Deterministically generate storage node configurations without a `SuiConfig`.
-///
-/// The storage node configurations are created for a total of `committee_size` nodes and
-/// `n_shards`.
-pub fn create_local_storage_node_configs(
-    working_dir: &Path,
-    committee_size: NonZeroU16,
-    n_shards: NonZeroU16,
-) -> (Vec<StorageNodeConfig>, Vec<SuiStorageNode>) {
+/// Generate deterministic keypairs for the benchmark purposes.
```
mlegner: Nit: We generally use indicative instead of imperative mood, aligned with [RFC 505](https://github.com/rust-lang/rfcs/blob/master/text/0505-api-comment-conventions.md#formatting).

crates/walrus-service/src/testbed.rs
```diff
@@ -39,146 +40,60 @@ pub fn node_config_name_prefix(node_index: u16, committee_size: NonZeroU16) -> S
...
+    node_index: Option<u16>,
+    committee_size: Option<u16>,
+) -> SocketAddr {
+    let port = if let (Some(node_index), Some(committee_size)) = (node_index, committee_size) {
+        port + committee_size + node_index
+    } else {
+        port
+    };
+    SocketAddr::new(IpAddr::V4(ip), port)
+}
```
mlegner: Remark: It is not fully clear from the documentation that this should be called with the same port as `get_metrics_address` when going for the adjustment version.
Maybe it would be better to have two separate functions instead of the version with the `Option<_>` parameters? (Same for the `get_metrics_address` above.)

Also, there is a slight simplification:
```suggestion
/// Get the REST API address for a node. If both the node index and the committee size is provided,
/// the port is adjusted to ensure uniqueness across nodes.
pub fn get_rest_api_address(
    ip: Ipv4Addr,
    mut port: u16,
    node_index: Option<u16>,
    committee_size: Option<u16>,
) -> SocketAddr {
    if let (Some(node_index), Some(committee_size)) = (node_index, committee_size) {
        port += committee_size + node_index
    };
    SocketAddr::new(IpAddr::V4(ip), port)
}
```

crates/walrus-service/src/testbed.rs
```diff
@@ -39,146 +40,60 @@ pub fn node_config_name_prefix(node_index: u16, committee_size: NonZeroU16) -> S
...
+/// Generate deterministic and even shard allocation for the benchmark purposes.
+pub fn even_shards_allocation(
+    n_shards: NonZeroU16,
+    committee_size: NonZeroU16,
+) -> Vec<Vec<ShardIndex>> {
     let shards_per_node = n_shards.get() / committee_size.get();
     let remainder_shards = n_shards.get() % committee_size.get();
     let mut start = 0;
-    let mut sui_storage_node_configs = Vec::with_capacity(committee_size.get().into());
+    let mut shards_information = Vec::with_capacity(committee_size.get() as usize);
```
mlegner: Nit: I generally dislike the `as` conversions with integer types as it is unclear if this is a (potentially) lossy or lossless conversion without looking at the types.

crates/walrus-service/src/testbed.rs
```diff
@@ -187,14 +102,22 @@ pub async fn deploy_walrus_contract(
...
-    let mut keypairs = benchmark_keypairs(committee_size);
+
+    // Check whether the testbed collocates the storage nodes on the same machine
+    // (that is, local testbed).
+    let ip_set = ips.iter().collect::<HashSet<_>>();
+    let collocated = ip_set.len() != ips.len();
+
+    // Build one Sui storage node config for each storage node.
+    let committee_size = ips.len() as u16;
+    let mut keypairs = benchmark_keypairs(committee_size as usize);
```
mlegner, line 118: Important: This way of implicitly defining the deployment mode (local vs. distributed) feels a bit dangerous to me. For example, if one accidentally has a duplicate IP address in the list, we would conclude that we are in the collocated scenario. Do we expect a scenario where some nodes have the same IP address but not all?

Suggestion: Distinguish between a local scenario with options `committee_size` and `ip_address` and a distributed one with `ip_addresses`, which are then expected and checked to be unique.

If we keep the current behavior, I would add this to the docstring for `ips` in `main.rs`.

Other remark: This code is duplicated with `create_storage_node_configs` below.

crates/walrus-service/src/testbed.rs
asonnino: good point I forgot about that

crates/walrus-service/src/testbed.rs
asonnino: same, I removed the conversion as it is not really needed here

crates/walrus-service/src/testbed.rs
asonnino: Good point, I made it clearer by indicating that the function "formats" the address

crates/walrus-service/src/testbed.rs
asonnino: Good point, I suggest to fix it in a different PR to not overly modify this branch (as it is currently deployed and used by Giacomo). I will make the PR immediatly

asonnino: [#399](https://github.com/MystenLabs/walrus/pull/399)

Cargo.lock
mlegner, line 1: Remark: I'm not opposed to upgrading our indirect dependencies, but I'd suggest doing this in a separate PR.

asonnino: I don't know what happened to cargo lock, I didn't meant to change it

asonnino: I can try to remove it, hopefully it won't break anything

PR #382, opened 2024-05-13
refactor(client): minor changes to the interfaces to facilitate blocksite builder

giac-mysten:
- Pass sliver pairs by reference instead of moving (walrus client);
- split out function to reserve the space for the blob (walrus client);
- expose the wallet in the contract client to allow for external calls (sui contract client).

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -573,7 +573,7 @@ impl<T> Client<T> {
             .map_err(ClientError::other)
     }
 
-    /// Resets the request client inside the Walrus client.
+    /// Resets the reqwest client inside the Walrus client.
     ///
     /// Useful to ensure that the client cannot communicate with storage nodes through connections
     /// that are being kept alive.


crates/walrus-service/src/client/communication.rs
```diff
@@ -161,7 +161,7 @@ impl<'a> NodeCommunication<'a> {
     pub async fn store_metadata_and_pairs(
         &self,
         metadata: &VerifiedBlobMetadataWithId,
-        pairs: Vec<SliverPair>,
+        pairs: impl IntoIterator<Item = &SliverPair>,
```
giac-mysten, line 164: Maybe a plain `impl Iterator` is better here?

jpcsmith: `IntoIterator` is good, as it allows passing slices directly.

crates/walrus-service/src/client/communication.rs
```diff
@@ -203,35 +203,33 @@ impl<'a> NodeCommunication<'a> {
...
+        let mut primary_futures = Vec::with_capacity(self.n_owned_shards().get().into());
+        let mut secondary_futures = Vec::with_capacity(self.n_owned_shards().get().into());
+        for pair in pairs {
+            primary_futures.push(self.store_sliver(blob_id, &pair.primary, pair.index()));
+            secondary_futures.push(self.store_sliver(blob_id, &pair.secondary, pair.index()));
+        }
+        let (mut primary_results, secondary_results) =
+            join!(join_all(primary_futures), join_all(secondary_futures));
+        primary_results.extend(secondary_results);
+        primary_results
```
jpcsmith: Hint: You can the approach you had before by using `futures::Either` in the `flat_map`.

crates/walrus-service/src/client.rs
```diff
@@ -518,35 +529,37 @@ impl<T> Client<T> {
...
+                (
+                    idx,
+                    pairs.iter().filter(|pair| {
+                        node.shard_ids.contains(
+                            &pair
+                                .index()
+                                .to_shard_index(self.committee.n_shards(), blob_id),
+                        )
+                    }),
                 )
```
jpcsmith: Hint: You don't seem to use idx here, if you move it after the map you can avoid needing to accept and return it from the function.

crates/walrus-service/src/client/communication.rs
giac-mysten: ah nice, this works. It's a bit confusing however, this does not seem to be the use case for `Either`... but well.

crates/walrus-service/src/client.rs
giac-mysten: very nice suggestion, thanks!

crates/walrus-service/src/client/communication.rs
jpcsmith: Hmm, how so?

giac-mysten: afaiu `Either` is meant to be use as "either this future or that future", while we use it for "both". imo the better way would be using `Pin<Box<dyn ...` but the type annotations become insanely long

crates/walrus-service/src/client.rs
```diff
@@ -149,42 +149,53 @@ impl<T: ContractClient> Client<T> {
             .get_blob_encoder(blob)
             .map_err(ClientError::other)?
             .encode_with_metadata();
-        tracing::Span::current().record("blob_id", metadata.blob_id().to_string());
```
mlegner, line 152: Question: Why remove this `record` call? This means that the `blob_id` in the current span will not be filled.

giac-mysten: ah my mistake. I wanted to remove the redundancy on the next line, b/c the blob id is already in the span.

Issue #396, opened 2024-05-15
Use a local client for sync


jpcsmith:
The storage node currently syncs with itself using an HTTP client. It would be more efficient to use a local client.

amosehiguese:
Hi @jpcsmith , I will love to take this up. If yes, I can send in a pr for review.

Issue #395, opened 2024-05-15
Track syncs across reboot


jpcsmith:
We need to store the IDs that we're currently syncing to a table, to continue after a reboot.

PR #397, opened 2024-05-15
feat: add sliver recovery flow

### LINKED ISSUE #279: Download missing shard from others within an epoch

jpcsmith:
This PR adds sliver recovery for a full node, this includes syncing the metadata, primary sliver, and secondary sliver. Closes #279.

*Not* included in this PR is
- Resuming sync after a restart (#395).
- Using a local client (instead of an HTTP client) for syncing from its own shards (#396)

### DIFF
--- a/crates/walrus-service/src/committee.rs
+++ b/crates/walrus-service/src/committee.rs
@@ -164,7 +164,7 @@ where
             .await
             .context("unable to create a committee service for the current epoch")?;
 
-        let service = NodeCommitteeService::new(committee, local_identity).await?;
+        let service = NodeCommitteeService::new(committee, local_identity)?;
 
         Ok(Box::new(service))
     }
@@ -178,7 +178,7 @@ pub struct NodeCommitteeService {
 
 impl NodeCommitteeService {
     /// Creates a new `NodeCommitteeService`.
-    pub async fn new(
+    pub fn new(
         committee: Committee,
         local_identity: Option<&PublicKey>,
     ) -> Result<Self, anyhow::Error> {
@@ -326,8 +326,9 @@ where
             .map(|(node, shard_id)| {
                 let public_key = node.public_key();
 
-                // Convert to a sliver pair id. This does not exclude any pair IDs because we're
-                // making orthogonal requests: the node itself already store the orthogonal sliver.
+                // Convert to a sliver pair id. It may be the case that this pair ID is the same as
+                // sliver_id, however, this is fine since we're requesting a symbol from the
+                // orthogonal sliver which may be already stored at this node.
                 let sliver_pair_at_remote = shard_id.to_pair_index(config.n_shards(), blob_id);
                 let simultaneous_requests = &simultaneous_requests;
                 let retry_strategy = default_retry_strategy(self.rng.lock().unwrap().gen());
@@ -439,7 +440,7 @@ where
             .local_identity
             .as_ref()
             .map(|local_identity| self.public_key() == local_identity)
-            .unwrap_or_default()
+            .unwrap_or(false)
     }
 }
 
--- a/crates/walrus-service/src/utils.rs
+++ b/crates/walrus-service/src/utils.rs
@@ -68,16 +68,12 @@ impl<R: Rng> ExponentialBackoff<R> {
         let next_delay_value = self
             .min_backoff
             .saturating_mul(Saturating(2u32).pow(self.sequence_index).0)
+            .saturating_add(self.random_offset())
             .min(self.max_backoff);
 
         self.sequence_index = self.sequence_index.saturating_add(1);
 
-        // Only add the random delay if we've not yet hit the maximum
-        if next_delay_value < self.max_backoff {
-            next_delay_value.saturating_add(self.random_offset())
-        } else {
-            next_delay_value
-        }
+        next_delay_value
     }
 
     fn random_offset(&mut self) -> Duration {
@@ -149,26 +145,21 @@ pub(crate) trait FutureHelpers: Future {
 
 impl<T: Future> FutureHelpers for T {}
 
-pub(crate) async fn retry<R, F, T, Fut>(strategy: ExponentialBackoff<R>, mut func: F) -> T
+pub(crate) async fn retry<R, F, T, Fut>(mut strategy: ExponentialBackoff<R>, mut func: F) -> T
 where
     F: FnMut() -> Fut,
     Fut: Future<Output = Option<T>>,
     R: rand::RngCore,
 {
-    for wait_time in std::iter::once(Duration::ZERO).chain(strategy) {
-        if wait_time != Duration::ZERO {
-            tracing::debug!("waiting {:?} before next attempt", wait_time);
-            tokio::time::sleep(wait_time).await;
-        }
-
+    loop {
         if let Some(value) = func().await {
             return value;
-        } else {
-            tracing::debug!("attempt failed, retrying");
         }
-    }
 
-    unreachable!("ExponentialBackoff is an infinite iterator");
+        let delay = strategy.next().expect("infinite iterator");
+        tracing::debug!(?delay, "attempt failed, waiting before retrying");
+        tokio::time::sleep(delay).await;
+    }
 }
 
 #[cfg(test)]
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -440,9 +440,10 @@ impl CommitteeServiceFactory for StubCommitteeServiceFactory<NodeCommitteeServic
         &self,
         local_identity: Option<&PublicKey>,
     ) -> Result<Box<dyn CommitteeService>, anyhow::Error> {
-        Ok(Box::new(
-            NodeCommitteeService::new(self.committee.clone(), local_identity).await?,
-        ))
+        Ok(Box::new(NodeCommitteeService::new(
+            self.committee.clone(),
+            local_identity,
+        )?))
     }
 }
 


crates/walrus-sdk/src/client.rs
```diff
@@ -174,6 +214,53 @@ impl Client {
         Ok(sliver)
     }
 
+    /// Gets the recovery symbol for a primary or secondary sliver.
+    ///
+    /// The symbol is identified by the (A, sliver_pair_index, target_pair_index) tuple.
```
karlwuest:
```suggestion
    /// The symbol is identified by the (A, sliver_pair_at_remote, intersecting_pair_index) tuple.
```

crates/walrus-sdk/src/client.rs
```diff
@@ -174,6 +214,53 @@ impl Client {
...
+            sliver_pair_at_remote,
+            intersecting_pair_index,
+        );
+        let response = self.inner.get(url).send().await.map_err(Kind::Reqwest)?;
+        response.response_error_for_status().await?.bcs().await
+    }
+
+    /// Gets the recovery symbol for a primary or secondary sliver.
+    ///
+    /// The symbol is identified by the
```
karlwuest:
```suggestion
    /// The symbol is identified by the (A, sliver_pair_at_remote, intersecting_pair_index) tuple.
```

crates/walrus-service/src/committee.rs
```diff
@@ -54,18 +91,46 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
...
+
+    /// Recovers a primary sliver from symbols stored by the committee.
+    async fn recover_primary_sliver(
+        &self,
+        metadata: &VerifiedBlobMetadataWithId,
+        sliver_id: SliverPairIndex,
+        encoding_config: &EncodingConfig,
+    ) -> Result<PrimarySliver, InconsistencyProof<Primary, MerkleProof>>;
+
+    /// Recovers a primary sliver from symbols stored by the committee.
```
karlwuest:
```suggestion
    /// Recovers a secondary sliver from symbols stored by the committee.
```

crates/walrus-service/src/committee.rs
```diff
@@ -89,53 +154,61 @@ impl<T> CommitteeServiceFactory for SuiCommitteeServiceFactory<T>
...
+/// Provides details of the committee, as well as an API to perform requests against its members.
 #[derive(Debug)]
 pub struct NodeCommitteeService {
     inner: NodeCommitteeServiceInner<StorageNodeClient>,
 }
 
 impl NodeCommitteeService {
     /// Creates a new `NodeCommitteeService`.
-    pub async fn new(committee: Committee) -> Result<Self, anyhow::Error> {
+    pub async fn new(
```
karlwuest:
```suggestion
    pub fn new(
```

crates/walrus-service/src/committee.rs
```diff
@@ -160,82 +231,224 @@ impl NodeCommitteeServiceInner<StorageNodeClient> {
...
+        let simultaneous_requests = Arc::new(Semaphore::new(min_symbols_required));
+
+        let mut symbol_requests = self
+            .shuffle_nodes_with_clients(&mut rng)
+            .flat_map(|node| iter::repeat(node).zip(node.shard_ids()))
+            .map(|(node, shard_id)| {
+                let public_key = node.public_key();
+
+                // Convert to a sliver pair id. This does not exclude any pair IDs because we're
+                // making orthogonal requests: the node itself already store the orthogonal sliver.
```
karlwuest: Question: Can you clarify what you mean with this comment? I don't really understand it.

crates/walrus-service/src/committee.rs
```diff
@@ -160,82 +231,224 @@ impl NodeCommitteeServiceInner<StorageNodeClient> {
...
+        &self.as_member().shard_ids
+    }
+
+    /// Returns true if this node corresponds to the local storage node.
+    fn is_local(&self) -> bool {
+        self.inner
+            .local_identity
+            .as_ref()
+            .map(|local_identity| self.public_key() == local_identity)
+            .unwrap_or_default()
```
karlwuest: Nit: I would prefer `unwrap_or(false)` for readability, but up to you.

crates/walrus-service/src/committee.rs
```diff
@@ -352,36 +578,37 @@ mod tests {
...
+                        Some(walrus_core::test_utils::verified_blob_metadata())
+                    })
                 });
-            (&mut iterator).for_each(|(_, client)| {
+            });
+            other.iter_mut().for_each(|client| {
                 client.expect_get_and_verify_metadata().returning(|_, _| {
                     Box::pin(async {
                         tracing::debug!("sleeping for 5 minutes");
                         tokio::time::sleep(Duration::from_secs(3600)).await;
```
karlwuest:
```suggestion
                        tracing::debug!("sleeping for 1 hour");
                        tokio::time::sleep(Duration::from_secs(3600)).await;
```

crates/walrus-service/src/node.rs
```diff
@@ -414,51 +413,133 @@ impl StorageNode {
...
+                shard_storage.put_sliver(&self.blob_id, &sliver)?;
+                tracing::debug!("sliver successfully synced");
+                Ok(None)
+            }
+            Err(proof) => Ok(Some(proof)),
+        }
+    }
+}
+
+async fn recover_sliver<A: EncodingAxis>(
```
karlwuest, line 525: Suggestion: Make this a method on `CommitteeService` with this as default implementation.

crates/walrus-service/src/node.rs
```diff
@@ -881,53 +967,371 @@ mod tests {
...
+            primary: (SliverType::Primary),
+            secondary: (SliverType::Secondary),
+        ]
+    }
+    async fn recovers_sliver_from_only_symbols_of_one_type(
+        sliver_type_to_store: SliverType,
+    ) -> TestResult {
+        let shards: &[&[u16]] = &[&[0], &[1, 2, 3, 4, 5, 6]];
+
+        // Store only a few primary slivers.
```
karlwuest:
```suggestion
        // Store only slivers of type `sliver_type_to_store`.
```

crates/walrus-service/src/node.rs
```diff
@@ -881,53 +967,371 @@ mod tests {
...
         Ok(())
     }
+
+    #[tokio::test]
+    #[ignore = "syncing all symbols takes several seconds"]
+    async fn recovers_sliver_from_a_small_set() -> TestResult {
+        let shards: &[&[u16]] = &[&[0], &(1..=6).collect::<Vec<_>>()];
+        let store_secondary_at: Vec<_> = ShardIndex::range(0..5).collect();
+
+        // Store only a few primary slivers.
```
karlwuest:
```suggestion
        // Store only a few secondary slivers.
```

crates/walrus-service/src/utils.rs
```diff
@@ -3,28 +3,38 @@
...
+
+/// An iterator over exponential wait durations.
 ///
-/// Use [`next_delay()`][Self::next_delay] to get the values or [`wait()`][Self::wait] to
-/// asynchronously pause for the expected delay.
+/// Returns the wait duration for an exponential backoff with a multiplicative factor of 2, and
+/// where each duration includes a random positive offset.
+///
+/// For the `i`-th iterator element and bounds `min_backoff` and `max_backoff`, this returns the
+/// sequence `min(max_backoff, 2^i * min_backoff + rand_i)`.
```
karlwuest, line 23: Nit: This comment is not quite correct, since the random offset is currently added after checking if we exceed `max_backoff`, so if `2^i * min_backoff < max_backoff`, it returns `2^i * min_backoff + rand_i` even if that is larger than `max_backoff`. You can make the implementation consistent by just moving the `min(self.max_backoff)` to after the if/else (if you want to avoid calling `random_offset` in the case where the delay is > max_backoff).

Alternatively, and this might be preferable to keep some jitter in the request timing even with a maximum delay, we could always add the random offset to get `min(max_backoff, 2^i * min_backoff) + rand_i`

crates/walrus-service/src/utils.rs
```diff
@@ -88,104 +97,83 @@ impl<R: Rng> Iterator for ExponentialBackoff<R> {
...
-        let weight = self.committee.members()[*next_index].shard_ids.len();
-        self.total_weight = self.total_weight.saturating_add(weight);
-
-        Some(*next_index)
+        if let Some(value) = func().await {
+            return value;
+        } else {
+            tracing::debug!("attempt failed, retrying");
+        }
     }
```
karlwuest: Suggestion:

```
   loop {
        if let Some(value) = func().await {
            return value;
        }
        tracing::debug!("attempt failed, retrying");
        let wait_time = strategy.next();
        tracing::debug!("waiting {:?} before next attempt", wait_time);
        tokio::time::sleep(wait_time).await;
   }
```

crates/walrus-service/src/node.rs
jpcsmith: I cannot, if I do, then CommitteeService is no longer object safe.

crates/walrus-service/src/utils.rs
jpcsmith: Thanks, that's cleaner.

crates/walrus-service/src/utils.rs
jpcsmith: I changed the code to maintain that the max remains the max. Thanks for catching this.

crates/walrus-service/src/committee.rs
jpcsmith: I updated the comment to clarify. The remote_sliver_id may be equal to the sliver_id, but that's okay because we can potentially fetch the symbol from that orthogonal sliver on our own node.

PR #399, opened 2024-05-15
feat(orchestrator): Explicit CLI different between local and remote testbed

asonnino:
CLI argument takes a special parameter to indicate it is running a remote or local testbed

### DIFF
--- a/scripts/local-testbed.sh
+++ b/scripts/local-testbed.sh
@@ -104,13 +104,13 @@ if ! $existing; then
     # Deploy system contract
     echo Deploying system contract...
     cargo run --bin walrus-node -- deploy-system-contract \
-    --working-dir $working_dir --sui-network $network --n-shards $shards --ips $ips
+    --working-dir $working_dir --sui-network $network --n-shards $shards --ips $ips --local-testbed
 
     # Generate configs
     echo Generating configuration...
     cargo run --bin walrus-node -- generate-dry-run-configs \
     --working-dir $working_dir --sui-network $network --sui-config-path $sui_config_path \
-    --ips $ips
+    --ips $ips --local-testbed
 fi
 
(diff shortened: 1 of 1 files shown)


crates/walrus-service/bin/main.rs
mlegner, line 1: Thanks for the explicit argument! One further suggestion: For the local testbed, it would be simpler to just specify a single IP address. This could be achieved by clap's argument groups and requirements.

One possibility would be to have two arguments `local-ip` (single value) and `remote-ips` (number of storage nodes) that [conflict with each other](https://stackoverflow.com/a/76315811) (instead of the current boolean `local_testbed` and the `ips`).

asonnino: but then we need an additional argument called "committee-size" for local testbeds?

asonnino: what do we do about this? Do I add the extra cli argument "committee-size" to make the distinction between local and remote ips, or do you see a better way?

mlegner: Oh, sorry, @asonnino, I missed your comment here. Yes, we would need another argument.

In my opinion, the combination of `committee-size` and `ip` is still nicer, but I don't have a very strong opinion there. We can also just leave it as it is now. 🙂

asonnino: then maybe let's park this for the moment until we have a Devent (as changing the CLI may further delay the PE team), and then fix it for the next iteration?

Also I suspect we can generally improve the CLI, it is a bit ugly that we need to provide twice the Ip addresses (once to generate the contract and once to generate the configs).

mlegner: Sounds good to me.

PR #401, opened 2024-05-16
refactor(client): change some configurations and add retries

giac-mysten:
- Increases the total timeout for Reqwest, which may otherwise time out when transferring large blobs.
- Adds retries to storing slivers
- Adds more logging to see the progress of storing slivers


crates/walrus-service/src/server.rs
```diff
@@ -146,6 +152,10 @@ impl<S: ServiceState + Send + Sync + 'static> UserServer<S> {
             .route(RECOVERY_ENDPOINT, get(Self::retrieve_recovery_symbol))
             .route(INCONSISTENCY_PROOF_ENDPOINT, put(Self::inconsistency_proof))
             .with_state(self.state.clone())
+            .layer(DefaultBodyLimit::max(
+                usize::from(self.state.n_secondary_source_symbols().get()) * usize::from(u16::MAX)
+                    + HEADROOM,
+            ))
```
giac-mysten: Question: to have an exact calculation of the maximum body size here I added a function to get the `n_secondary_source_symbols` from the `ServiceState`. Is this fine? Or should I just export the encoding config?

crates/walrus-service/src/client/communication.rs
```diff
@@ -195,6 +197,40 @@ impl<'a> NodeCommunication<'a> {
...
+                    .into_iter()
+                    .filter(|p| failed_ids.contains(&p.index()))
+                    .collect();
+            } else {
+                // We have stored all the slivers.
+                break;
+            }
+        }
+        results
+    }
```
jpcsmith: Suggestion: Do the retries at the granularity of the individual sliver, i.e., in `store_sliver`, as this would avoid all the logic for checking which/how-many succeed and retrying the entire chain of computations here.

crates/walrus-service/src/server.rs
jpcsmith: Perhaps since this can all be computed from n_shards, add an API to return that instead?

jpcsmith: Suggestion: This could also be applied with a different value for each route, or only increased for the route that needs the higher value.

crates/walrus-service/src/client/communication.rs
```diff
@@ -203,15 +239,29 @@ impl<'a> NodeCommunication<'a> {
...
+        let mut results = Vec::with_capacity(n_slivers);
+        while let Some(result) = futures.next().await {
+            tracing::trace!(
+                ?result,
+                progress = format!("{} / {}", results.len() + 1, n_slivers),
+                "sliver stored"
+            );
+            results.push(result);
+        }
+        results
```
jpcsmith: Suggestion:  You can also directly collect the stream into a vector, with something like the following which uses methods from the futures `StreamExt` trait (untested):

```suggestion
            .collect::<FuturesUnordered<_>>()
            .inspect(|result| {
                tracing::trace!(
                    ?result,
                    progress = format!("{} / {}", results.len() + 1, n_slivers),
                    "sliver stored"
                );
            })
            .collect::<Vec<_>>()
            .await
```

crates/walrus-service/src/client/communication.rs
```diff
@@ -203,15 +239,29 @@ impl<'a> NodeCommunication<'a> {
...
+                ]
+            })
+            .collect::<FuturesUnordered<_>>();
+
+        let mut results = Vec::with_capacity(n_slivers);
+        while let Some(result) = futures.next().await {
+            tracing::trace!(
+                ?result,
+                progress = format!("{} / {}", results.len() + 1, n_slivers),
+                "sliver stored"
```
jpcsmith, line 252: Remark: The message is "sliver stored" but the sliver may not be stored if the result was an error.

crates/walrus-service/src/client/communication.rs
giac-mysten: 👍 fixed in #403

crates/walrus-service/src/client/communication.rs
giac-mysten: changed in #403

crates/walrus-service/src/client/communication.rs
giac-mysten: that's very nice. However, in #403 i not only log but also break out of the function if i get an error, and I could not find a good way to do the same within the iterator

crates/walrus-service/src/server.rs
giac-mysten: good point, trying to understand which of the different `layer` or `route_layer` options are best.

crates/walrus-service/src/server.rs
```diff
@@ -133,11 +139,23 @@ impl<S: ServiceState + Send + Sync + 'static> UserServer<S> {
...
             .route(
                 SLIVER_ENDPOINT,
-                get(Self::retrieve_sliver).put(Self::store_sliver),
+                get(Self::retrieve_sliver)
+                    .put(Self::store_sliver)
+                    .route_layer(DefaultBodyLimit::max(
+                        usize::try_from(self.state.max_sliver_size())
+                            .expect("running on 64bit arch (see hardware requirements)")
+                            + HEADROOM,
+                    )),
```
giac-mysten: @jpcsmith I added here specific limits for the enpoints. I used the `route_layer`, and not just `layer`, as afaics it only applies if the route matches, which is waht we want. Also I applied them to the `MethodRouter` and not the `Router`, so that afaiu they are separate.

crates/walrus-service/src/node.rs
```diff
@@ -181,8 +181,11 @@ pub trait ServiceState {
         target_pair_index: SliverPairIndex,
     ) -> Result<RecoverySymbol<MerkleProof>, RetrieveSymbolError>;
 
-    /// Returns the number of secondary source symbols.
-    fn n_secondary_source_symbols(&self) -> NonZeroU16;
+    /// Returns the maximum sliver size for the service configuration.
+    fn max_sliver_size(&self) -> u64;
+
+    /// Returns the metadata size for the service configuration.
+    fn metadata_length(&self) -> u64;
```
jpcsmith: Remark: I would prefer if these were not placed on the trait + node. They have nothing to do with the node and are encoding specific. It seems that the computations could be stand-alone functions that take the number of shards.

crates/walrus-service/src/server.rs
jpcsmith: Remark: I think the ordering should be put -> route_layer -> get, otherwise this adds the increased limit to the get request as well.

crates/walrus-service/src/node.rs
giac-mysten: mm I thought it is better to have these as they are the sizes of things the node may store, so in that sense they have more to do than the number of nodes. However, I get your point. will change.

PR #403, opened 2024-05-20
feat(client): implement retries and rate limiting of writes

giac-mysten:
Follow up to #401 

- Fixes `ENOBUFS` errors when writing large blobs
  - Limits the number of concurrent connections per node
  - Limits the number of total connections
- Implements retries with exponential backoff when writing slivers

### DIFF
--- a/crates/walrus-service/src/client/communication.rs
+++ b/crates/walrus-service/src/client/communication.rs
@@ -244,14 +244,13 @@ impl<'a> NodeCommunication<'a> {
                     "could not store sliver after retrying; stopping storing on the node"
                 );
                 return Err(error);
-            } else {
-                tracing::trace!(
-                    node_permits=?self.node_connection_limit.available_permits(),
-                    global_permits=?self.global_connection_limit.available_permits(),
-                    progress = format!("{}/{}", n_slivers - requests.len(), n_slivers),
-                    "sliver stored"
-                );
             }
+            tracing::trace!(
+                node_permits=?self.node_connection_limit.available_permits(),
+                global_permits=?self.global_connection_limit.available_permits(),
+                progress = format!("{}/{}", n_slivers - requests.len(), n_slivers),
+                "sliver stored"
+            );
         }
         Ok(())
     }


crates/walrus-service/src/utils.rs
jpcsmith, line 1: Suggestion: Reduce the code by merging the infinite and non-infinite iterators and storing a `max_retries: None` for the infinite variant.

crates/walrus-service/src/utils.rs
```diff
@@ -111,6 +167,18 @@ impl<T> SuccessOrFailure for Option<T> {
 
 pub(crate) trait FutureHelpers: Future {
     async fn limit(self, permits: Arc<Semaphore>) -> Self::Output
```
jpcsmith: Remark: @karlwuest  has a PR open that *maybe* uses the previous `limit()` implementation, whoever ends up rebasing needs to correctly rename the additional usage. Otherwise, changing `limit` to `batch_limit` could help avoiding the issues during rebase.

crates/walrus-service/src/utils.rs
```diff
@@ -145,20 +213,30 @@ pub(crate) trait FutureHelpers: Future {
...
+        match strategy.next_delay() {
+            Some(delay) => {
+                tracing::debug!(?delay, "attempt failed, waiting before retrying");
+                tokio::time::sleep(delay).await;
+            }
+            None => {
+                tracing::debug!("last attempt failed, returning last failure value");
+                return value;
+            }
+        }
```
jpcsmith: Nit: `if let Some(delay) = strategy.next_delay() {` would be shorter.

crates/walrus-service/src/client/config.rs
```diff
@@ -30,33 +30,57 @@ impl LoadConfig for Config {}
...
-            retries: 0,
+            node_config: NodeConfig::default(),
+        }
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(default)]
+/// Configuration for retries towards the storage nodes.
+pub struct NodeConfig {
```
jpcsmith: Suggestion: How about `RequestRateConfig`? `NodeConfig` is awfully confusing.

crates/walrus-service/src/client/communication.rs
```diff
@@ -197,71 +195,72 @@ impl<'a> NodeCommunication<'a> {
...
+                    progress = format!("{}/{}", n_slivers - requests.len(), n_slivers),
+                    "sliver stored"
+                ),
+                Err(error) => {
+                    tracing::warn!(
+                        node_permits=?node_connection_limit.available_permits(),
+                        global_permits=?self.global_connection_limit.available_permits(),
+                        ?error,
+                        %self.config.max_retries,
+                        "could not store sliver after retrying; closing connection to the node"
```
jpcsmith: Nit: the connection isn't actually being closed though...

crates/walrus-service/src/client/communication.rs
```diff
@@ -197,71 +195,72 @@ impl<'a> NodeCommunication<'a> {
...
+        let n_slivers = requests.len();
+        while let Some(result) = requests.next().await {
+            match result {
+                Ok(()) => tracing::trace!(
+                    node_permits=?node_connection_limit.available_permits(),
+                    global_permits=?self.global_connection_limit.available_permits(),
+                    progress = format!("{}/{}", n_slivers - requests.len(), n_slivers),
+                    "sliver stored"
+                ),
+                Err(error) => {
```
jpcsmith: Nit: `if let Err(error) = ...` would be more succint.

crates/walrus-service/src/client/communication.rs
```diff
@@ -197,71 +195,72 @@ impl<'a> NodeCommunication<'a> {
...
         &self,
         blob_id: &BlobId,
-        pairs: &[&SliverPair],
-    ) -> Vec<Result<(), SliverStoreError>> {
-        let n_slivers = 2 * pairs.len();
-        let mut futures = pairs
-            .iter()
+        pairs: impl IntoIterator<Item = &SliverPair>,
+    ) -> Result<(), SliverStoreError> {
+        let node_connection_limit = Arc::new(Semaphore::new(self.config.max_node_connections));
```
jpcsmith: Remark: This should probably be struct-wide, across all the operations, as the `max_node_connections` limit would not be observed if two invocations of `store_pairs` are run concurrently.

crates/walrus-service/src/client/communication.rs
```diff
@@ -197,71 +195,72 @@ impl<'a> NodeCommunication<'a> {
...
-        }
-        results
+        metadata: &VerifiedBlobMetadataWithId,
+    ) -> Result<(), NodeError> {
+        utils::retry(self.backoff_strategy(), || {
+            self.client
+                .store_metadata(metadata)
+                // TODO(giac): consider adding timeouts and replace the Reqwest timeout.
+                .limit(self.global_connection_limit.clone())
+        })
```
jpcsmith: Question: Should this also be obeying the max connections limit?

crates/walrus-service/src/client.rs
```diff
@@ -226,23 +234,22 @@ impl<T> Client<T> {
...
             elapsed_time = ?start.elapsed(), "stored metadata and slivers onto a quorum of nodes"
         );
-        // Double the execution time, with a minimum of 100 ms. This gives the client time to
+        // Add 10% of the execution time, with a minimum of 100 ms. This gives the client time to
         // collect more storage confirmations.
         let completed_reason = requests
             .execute_time(
-                start.elapsed() + Duration::from_millis(100),
-                self.concurrent_writes,
+                start.elapsed() / 10 + Duration::from_millis(100),
```
jpcsmith, line 254:
```suggestion
                cmp::max(start.elapsed() / 10, Duration::from_millis(100)),
```

Or update the comment to reflect 10% of elapsed + 100ms

crates/walrus-service/src/utils.rs
karlwuest: I'm not using `limit` in my PR.

giac-mysten: i did the renaming anyways.
`limit` stays `limit`, and the new version is called `batch_limit`

crates/walrus-service/src/client/communication.rs
giac-mysten: yes, will change this, although probably the current way this would be called is by creating a separate `NodeCommunication`.

crates/walrus-service/src/client/communication.rs
giac-mysten: now that the limit is global, yes.
before it did not matter, as each `NodeCommunication` only stores one pair of metadata to the node.

one bigger point it is that maybe the client could keep a pool on `NodeCommunication` objects alive, so that in the aggregator these can be reused. Opened an issue #417

crates/walrus-service/src/utils.rs
giac-mysten: @karlwuest I think your PR uses `ExponentialBackoff`, right? I changed it here to take another `max_retries` parameter, so you may need to rebase after this is merged.

crates/walrus-service/src/client/communication.rs
```diff
@@ -218,46 +222,35 @@ impl<'a> NodeCommunication<'a> {
...
+            if let Err(error) = result {
+                tracing::warn!(
+                    node_permits=?self.node_connection_limit.available_permits(),
+                    global_permits=?self.global_connection_limit.available_permits(),
+                    ?error,
+                    ?self.config.max_retries,
+                    "could not store sliver after retrying; stopping storing on the node"
+                );
+                return Err(error);
+            } else {
```
jpcsmith: Nit: unecessary else after return.

PR #408, opened 2024-05-22
feat: sync after restart

jpcsmith:
Modifies the tracking of cursors to only advance the cursor when all previous events have been fully handled. As a result, an in-progress blob sync that fails to complete before a restart will be retried as the event is re-evaluated.

### DIFF
--- a/crates/walrus-service/src/storage/event_sequencer.rs
+++ b/crates/walrus-service/src/storage/event_sequencer.rs
@@ -17,7 +17,7 @@ pub(super) struct EventSequencer {
     // next_required_sequence_num.
     head: Option<EventID>,
     next_required_sequence_num: usize,
-    queue: BinaryHeap<Reverse<SequencedEventId>>,
+    queue: BinaryHeap<Reverse<Sequenced<EventID>>>,
 }
 
 impl EventSequencer {
@@ -26,6 +26,8 @@ impl EventSequencer {
     }
 
     /// Adds the provided (sequence_number, EventID) pair to those observed.
+    ///
+    /// Added sequence numbers must be unique over those observed up to this point.
     pub fn add(&mut self, sequence_number: usize, event_id: EventID) {
         match sequence_number.cmp(&self.next_required_sequence_num) {
             Ordering::Equal => {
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/storage/event_sequencer.rs
```diff
@@ -0,0 +1,86 @@
...
+impl PartialOrd for SequencedEventId {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for SequencedEventId {
+    fn cmp(&self, other: &Self) -> Ordering {
+        let this = (self.0, self.1.tx_digest, self.1.event_seq);
+        let other = (other.0, other.1.tx_digest, other.1.event_seq);
```
karlwuest: Remark: It should never happen that we have the same sequence number and a different event ID. I'm wondering if it would be better to just implement `PartialOrd` and then error/panic in `add`, if two events are not comparable (i.e. same sequence number but different event id).

crates/walrus-service/src/node.rs
```diff
@@ -1279,6 +1295,69 @@ mod tests {
...
+    async fn does_not_advance_cursor_past_incomplete_blobs() -> TestResult {
+        let _ = tracing_subscriber::fmt()
+            .json()
+            .with_env_filter(tracing_subscriber::EnvFilter::from_default_env())
+            .try_init();
+
+        let shards: &[&[u16]] = &[&[1, 6], &[0, 2, 3, 4, 5]];
+        let own_shards = [ShardIndex(1), ShardIndex(6)];
+
+        let blob1 = BLOB;
```
karlwuest: Suggestion: Don't use `BLOB` here or add a comment somewhere that makes clear that `BLOB` and `BLOB_ID` are unrelated. Otherwise, it reads as if they correspond to each other, and makes it hard to understand why the event cursor should not progress.

jpcsmith: Replaced with 3 explicit blobs and removed the use of BLOB and BLOB_ID in the test case.

crates/walrus-service/src/storage/event_sequencer.rs
jpcsmith: Okay, I changed this so that the `PartialOrd` and `PartialEq` implementations only use the sequence number, and added checks for repeated sequence numbers to the `EventSequencer`. If we by mistake pass the same sequence number twice to the class, then we should get an error in debug mode.

PR #411, opened 2024-05-22
feat: implement sending inconsistency certificates to other nodes

### LINKED ISSUE #332: Add functionality to share an inconsistency proof with other storage nodes, collect a certificate, and post this on chain
Part of #257

karlwuest:
closes  #332

### DIFF
--- a/crates/walrus-core/src/messages.rs
+++ b/crates/walrus-core/src/messages.rs
@@ -84,7 +84,7 @@ impl<T> SignedMessage<T> {
 
     /// Verifies that the signature on this message is valid for the specified public key and
     /// that the message contains the expected contents.
-    pub fn verify_contents<I>(
+    pub fn verify_signature_and_contents<I>(
         &self,
         public_key: &PublicKey,
         epoch: Epoch,
@@ -210,6 +210,18 @@ mod tests {
 
     const EPOCH: Epoch = 21;
     const BLOB_ID: BlobId = BlobId([7; 32]);
+
+    impl AsMut<ProtocolMessage<BlobId>> for Confirmation {
+        fn as_mut(&mut self) -> &mut ProtocolMessage<BlobId> {
+            &mut self.0
+        }
+    }
+
+    impl AsMut<ProtocolMessage<BlobId>> for InvalidBlobIdMsg {
+        fn as_mut(&mut self) -> &mut ProtocolMessage<BlobId> {
+            &mut self.0
+        }
+    }
 
     param_test! {
         decoding_fails_for_incorrect_message_type -> TestResult: [
--- a/crates/walrus-core/src/messages/invalid_blob_id.rs
+++ b/crates/walrus-core/src/messages/invalid_blob_id.rs
@@ -9,7 +9,7 @@ use crate::{messages::IntentType, BlobId, Epoch, PublicKey};
 /// A message stating that a Blob Id is invalid.
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(try_from = "ProtocolMessage<BlobId>")]
-pub struct InvalidBlobIdMsg(ProtocolMessage<BlobId>);
+pub struct InvalidBlobIdMsg(pub(crate) ProtocolMessage<BlobId>);
 
 impl InvalidBlobIdMsg {
     const INTENT: Intent = Intent::storage(IntentType::INVALID_BLOB_ID_MSG);
@@ -44,12 +44,6 @@ impl AsRef<ProtocolMessage<BlobId>> for InvalidBlobIdMsg {
     }
 }
 
-impl AsMut<ProtocolMessage<BlobId>> for InvalidBlobIdMsg {
-    fn as_mut(&mut self) -> &mut ProtocolMessage<BlobId> {
-        &mut self.0
-    }
-}
-
 /// A signed [`InvalidBlobIdMsg`] from a storage node.
 pub type InvalidBlobIdAttestation = SignedMessage<InvalidBlobIdMsg>;
 
@@ -62,7 +56,7 @@ impl InvalidBlobIdAttestation {
         epoch: Epoch,
         blob_id: &BlobId,
     ) -> Result<InvalidBlobIdMsg, MessageVerificationError> {
-        self.verify_contents(public_key, epoch, blob_id)
+        self.verify_signature_and_contents(public_key, epoch, blob_id)
     }
 }
 
--- a/crates/walrus-core/src/messages/storage_confirmation.rs
+++ b/crates/walrus-core/src/messages/storage_confirmation.rs
@@ -17,7 +17,7 @@ pub enum StorageConfirmation {
 /// of a blob in their shards.
 #[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(try_from = "ProtocolMessage<BlobId>")]
-pub struct Confirmation(ProtocolMessage<BlobId>);
+pub struct Confirmation(pub(crate) ProtocolMessage<BlobId>);
 
 impl Confirmation {
     const INTENT: Intent = Intent::storage(IntentType::BLOB_CERT_MSG);
@@ -52,12 +52,6 @@ impl AsRef<ProtocolMessage<BlobId>> for Confirmation {
     }
 }
 
-impl AsMut<ProtocolMessage<BlobId>> for Confirmation {
-    fn as_mut(&mut self) -> &mut ProtocolMessage<BlobId> {
-        &mut self.0
-    }
-}
-
 /// A signed [`Confirmation`] from a storage node.
 pub type SignedStorageConfirmation = SignedMessage<Confirmation>;
 
@@ -69,7 +63,7 @@ impl SignedStorageConfirmation {
         epoch: Epoch,
         blob_id: &BlobId,
     ) -> Result<Confirmation, MessageVerificationError> {
-        self.verify_contents(public_key, epoch, blob_id)
+        self.verify_signature_and_contents(public_key, epoch, blob_id)
     }
 }
 


crates/walrus-core/src/messages.rs
```diff
@@ -3,28 +3,33 @@
...
 mod invalid_blob_id;
 pub use invalid_blob_id::{InvalidBlobIdAttestation, InvalidBlobIdMsg};
 
 mod certificate;
-pub use certificate::{ConfirmationCertificate, InvalidBlobCertificate};
+pub use certificate::{CertificateError, ConfirmationCertificate, InvalidBlobCertificate};
 
 /// Trait implemented by system messages.
-pub trait ProtocolMessage: Serialize + for<'de> Deserialize<'de> {}
+pub trait ProtocolMessage: Serialize + for<'de> Deserialize<'de> + Eq + Debug {
```
jpcsmith:
```suggestion
pub trait ProtocolMessage: Serialize + DeserializeOwned + Eq + Debug {
```

crates/walrus-core/src/messages.rs
```diff
@@ -48,7 +53,72 @@ impl<T> SignedMessage<T> {
...
+    if unverified_intent.app_id != IntentAppId::STORAGE {
+        return Err(format!(
+            "invalid App ID for message: {:?}",
+            unverified_intent.app_id
+        ));
+    }
+    Ok(())
+}
+
+use crate::{ensure, wrapped_uint, Epoch, PublicKey};
```
jpcsmith: Suggestion: Move to top of file.

crates/walrus-core/src/messages.rs
```diff
@@ -3,28 +3,33 @@
...
 
 /// Trait implemented by system messages.
-pub trait ProtocolMessage: Serialize + for<'de> Deserialize<'de> {}
+pub trait ProtocolMessage: Serialize + for<'de> Deserialize<'de> + Eq + Debug {
+    /// The intent used when signing a [`ProtocolMessage`] of this type.
+    const INTENT: Intent;
+
+    /// The epoch with which this message is associated.
+    fn epoch(&self) -> Epoch;
+}
```
jpcsmith: Suggestion: Replace `trait ProtocolMessage` with `struct ProtocolMessage<T> { intent, epoch, inner: T }`. This would allow us to unify the decoding and verification of the intent, app-id, and version.

`InvalidBlobIdMsg` and `Certification` would then just be `InvalidBlobIdMsg(ProtocolMessage<BlobId>)` and `Certification(ProtocolMessage<BlobId>)` respectively.

`SignedMessage` below would remain the same, but the bound could be `T: AsRef<ProtocolMessage<U>>`

crates/walrus-sdk/src/client.rs
```diff
@@ -372,6 +373,24 @@ impl Client {
...
+        public_key: &PublicKey,
+    ) -> Result<InvalidBlobIdAttestation, NodeError> {
+        let attestation = self
+            .send_inconsistency_proof(blob_id, inconsistency_proof)
+            .await?;
+        let _ = attestation
+            .verify(public_key, epoch, blob_id)
+            .map_err(NodeError::other)?;
+        Ok(attestation)
+    }
```
jpcsmith: Suggestion: Replace the generic argument M with the concrete type that we're using on this method and those above, as our system does not work for any generic proof at this point.

crates/walrus-service/tests/test_client.rs
```diff
@@ -79,9 +86,106 @@ async fn run_store_and_read_with_crash_failures(
...
+    })
+    .await
+    .map_err(anyhow::Error::new)
+}
+
+async fn default_setup() -> anyhow::Result<(
+    SuiTestCluster,
+    TestCluster,
+    WithTempDir<Client<SuiContractClient>>,
+)> {
```
jpcsmith, line 209: Remark: It would be good at some point to have a single test cluster for all tests, albeit with different wallets and walrus objects. This would remove the need to spin up a new cluster for each test case in a single integration test file.

crates/walrus-test-utils/src/lib.rs
```diff
@@ -185,6 +187,45 @@ impl<T> WithTempDir<T> {
             temp_dir: self.temp_dir,
         }
     }
+
+    /// Convert a `WithTempDir<T>` to a `WithTempDir<U>` by applying the provided
+    /// fallible function to the inner value, while maintaining the temporary directory.
+    pub fn try_map<U, F, E>(self, f: F) -> std::result::Result<WithTempDir<U>, E>
```
jpcsmith: Suggestion: Rename to `and_then` as the similar fallible function on `Resutlt`.

crates/walrus-test-utils/src/lib.rs
```diff
@@ -185,6 +187,45 @@ impl<T> WithTempDir<T> {
...
+        WithTempDir {
+            inner: f(self.inner).await,
+            temp_dir: self.temp_dir,
+        }
+    }
+
+    /// Convert a `WithTempDir<T>` to a `WithTempDir<U>` by applying the provided
+    /// fallible async function to the inner value, while maintaining the temporary
+    /// directory.
+    pub async fn try_map_async<U, F, Fut, E>(self, f: F) -> std::result::Result<WithTempDir<U>, E>
```
jpcsmith: Suggestion: Rename to `and_then_async` as the similar fallible function on `Resutlt`.

crates/walrus-core/src/messages/certificate.rs
```diff
@@ -1,29 +1,88 @@
...
     /// The indices of the signing nodes
     pub signers: Vec<u16>,
-    /// The BCS-encoded [`super::invalid_blob_id::InvalidBlobIdMsg`].
-    pub invalid_blob_id_msg: Vec<u8>,
+    /// The BCS-encoded message.
+    pub serialized_message: Vec<u8>,
     /// The aggregate signature over the BCS encoded confirmation.
     pub signature: BLS12381AggregateSignature,
+    #[serde(skip)]
+    message_type: PhantomData<T>,
```
jpcsmith, line 31: Question: Why not wrap a `SignedMessage`? It seems to have all of these fields already.

karlwuest: In the `SignedMessage`, we have a `BLS12381Signature`, but here we want a `BLS12381AggregateSignature`.

crates/walrus-core/src/messages/certificate.rs
```diff
@@ -1,29 +1,88 @@
...
+            .into_iter()
+            .map(|message| (message.signature, message.serialized_message))
+            .unzip();
+        ensure!(
+            serialized_msgs.len() == 1,
+            CertificateError::MessageMismatch
+        );
+        let serialized_message = serialized_msgs
+            .pop_first()
+            .expect("should contain exactly one message, checked above");
```
jpcsmith: Not sure if it's better, but 

```suggestion
        let mut signed_messages = signed_messages.into_iter();

        let Some(first) = signed_messages.next() else {
            return Err(CertificateError::MessageMismatch);
        };

        let signatures: Vec<_> = core::iter::once(Ok(first.signature))
            .chain(signed_messages.map(|message| {
                (first.serialized_message == message.serialized_message)
                    .then_some(message.signature)
                    .ok_or(CertificateError::MessageMismatch)
            }))
            .collect::<Result<_, _>>()?;
```

crates/walrus-core/src/messages.rs
```diff
@@ -3,28 +3,83 @@
...
+
+        if unverified_intent.app_id != IntentAppId::STORAGE {
+            return Err(D::Error::custom(format!(
+                "invalid App ID for message: {:?}",
+                unverified_intent.app_id
+            )));
+        }
+        Ok(unverified_intent)
+    }
+}
```
jpcsmith: Remark: With the TryFrom implementation, we would no longer have this check, instead just have in the try-from implementation:
```

impl TryFrom<ProtocolMessage<BlobId>> for Certification {

...

    if protocol_message.intent == Self::INTENT {
      Ok(protocol_message)
    } else {
      Err(InvalidIntent { expected, actual })
    }
}
```

Then we could remove the MessageInner trait and the deserialise bounds on ProtocolMessage, and instead rely on standard traits for the check.

crates/walrus-sdk/src/client.rs
```diff
@@ -372,6 +372,24 @@ impl Client {
             }
         }
     }
+
+    /// Sends an inconsistency proof to a node and verifies the returned invalid blob id
+    /// attestation.
+    pub async fn get_and_verify_invalid_blob_attestation(
```
jpcsmith, line 378: Nit: Should this be `send_and_verify`?

crates/walrus-core/src/messages/certificate.rs
karlwuest: This does not do the same thing. Collecting into a `BTreeSet` and then checking that its length is one ensures that all messages are the same. The suggested version just checks that there is at least one message.

crates/walrus-service/src/committee.rs
```diff
@@ -405,6 +425,78 @@ where
...
+        let weight_threshold = bft::min_n_correct(n_shards).get().into();
+        let mut total_weight = 0;
+        let (signed_messages, signer_indices): (Vec<_>, Vec<u16>) = signature_requests
+            .take_while(|(_, node)| {
+                future::ready({
+                    let take = total_weight < weight_threshold;
+                    total_weight += node.shard_ids().len();
+                    take
+                })
+            })
```
jpcsmith: Hint: I believe the following should also work. It's a sync function with a single statement: an async block
```suggestion
            .take_while(|(_, node)| async {
                let take = total_weight < weight_threshold;
                total_weight += node.shard_ids().len();
                take
            })
```

crates/walrus-service/src/committee.rs
```diff
@@ -405,6 +425,78 @@ where
...
+            .collect::<FuturesUnordered<_>>();
+
+        // get quorum and assemble certificate
+        let weight_threshold = bft::min_n_correct(n_shards).get().into();
+        let mut total_weight = 0;
+        let (signed_messages, signer_indices): (Vec<_>, Vec<u16>) = signature_requests
+            .take_while(|(_, node)| {
+                future::ready({
+                    let take = total_weight < weight_threshold;
+                    total_weight += node.shard_ids().len();
```
jpcsmith: Hint: Feel free to add helpers for `n_shards()` and `index_in_committee() -> u16` to the node class above.

crates/walrus-service/src/committee.rs
```diff
@@ -405,6 +425,78 @@ where
...
+                })
+                .instrument(tracing::info_span!("node", ?public_key))
+            })
+            .collect::<FuturesUnordered<_>>();
+
+        // get quorum and assemble certificate
+        let weight_threshold = bft::min_n_correct(n_shards).get().into();
+        let mut total_weight = 0;
+        let (signed_messages, signer_indices): (Vec<_>, Vec<u16>) = signature_requests
+            .take_while(|(_, node)| {
```
jpcsmith: Hint: The `take_while + map` could be simplified with [`StreamExt::scan`](https://docs.rs/futures/latest/futures/prelude/stream/trait.StreamExt.html#method.scan).

Important: This approach (irrespective of the use of scan) always waits until at least 1 additional future completes, as false/None is returned after the future has resolved. To resolve this, either a manual loop is necessary or the use of [`StreamExt::take_until`](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.take_until). With `take_until`, however, it's unclear to me what the correct ordering of operations would be, either

```rust
.for_each(|(_, node)| {
    total_weight += node.shard_ids().len();
})
.take_until(future::poll_fn(|_cx| {
    if total_weight < weight_threshold {
        Poll::Pending
    } else {
        Poll::Ready(())
    }
})
```
or
```rust
.take_until(future::poll_fn(|_cx| {
    if total_weight < weight_threshold {
        Poll::Pending
    } else {
        Poll::Ready(())
    }
})
.for_each(|(_, node)| {
    total_weight += node.shard_ids().len();
})
```
I suspect the former may end prematurely.

crates/walrus-core/src/messages/certificate.rs
jpcsmith: Hmm, perhaps I misunderstood something, but for the sequence `[(sig1, A), (sig2, A), (sig3, A), (sig4, B), (sig5, A)]`, we would get `[Ok(sig1)].chain([Ok(sig2), Ok(sig3), Err(_), ..])` which collects into an `Err(_)`.

karlwuest: Ah sorry, I misread your version and somehow missed the equality check for the messages 🤦‍♂️.

crates/walrus-service/src/committee.rs
karlwuest: That doesn't quite work, but the following does:
```
.take_while(|(_, node)| {
                let take = total_weight < weight_threshold;
                total_weight += node.shard_ids().len();
                async move { take }
            })
```

crates/walrus-sdk/src/client.rs
karlwuest: The `get` here refers to getting the attestation. I didn't want to name it `send_inconsistency_proof_and_get_and_verify_invalid_blob_attestation` 😅.
I could do `send_inconsistency_proof_and_verify_invalid_blob_attestation` but it's still pretty long. Wdyt?

crates/walrus-service/src/committee.rs
karlwuest: The following seems to work, but I think at this point a manual loop may be cleaner, wdyt?

```
        let weight_threshold = &bft::min_n_correct(n_shards).get();
        let total_weight = Mutex::new(0);

        let (signed_messages, signer_indices): (Vec<_>, Vec<u16>) = signature_requests
            .take_until(future::poll_fn(|_cx| {
                if total_weight.lock().unwrap().lt(weight_threshold) {
                    Poll::Pending
                } else {
                    Poll::Ready(())
                }
            }))
            .map(|(signed_msg, node)| {
                *total_weight.lock().unwrap() += node.n_shards();
                (signed_msg, node.index_in_committee())
            })
            .unzip()
            .await;
```

crates/walrus-sdk/src/client.rs
jpcsmith: It's fine as is.

crates/walrus-service/tests/test_client.rs
karlwuest: Opened #429 for this.

crates/walrus-service/src/committee.rs
jpcsmith: Either works, but if going with this approach you just use an atomic instead of a mutex. The `lt()` could be replaced with a < in either case (for the mutex, you may need to dereference the guard, i.e., `*total_weight.lock().unwrap() < weight_threshold`.

karlwuest: I changed this to a loop now, it's much more readable in my opinion.

crates/walrus-core/src/messages.rs
```diff
@@ -48,7 +67,86 @@ impl<T> SignedMessage<T> {
...
+    {
+        let message: ProtocolMessage<T> = bcs::from_bytes(&self.serialized_message)?;
+        public_key.verify(&self.serialized_message, &self.signature)?;
+
+        Ok(message)
+    }
+
+    /// Verifies that the signature on this message is valid for the specified public key and
+    /// that the message contains the expected contents.
+    pub fn verify_contents<I>(
```
jpcsmith: Nit: Sounds orthogonal to `verify_signature`, perhaps `verify_signature_and_contents`.

crates/walrus-core/src/messages/invalid_blob_id.rs
```diff
@@ -3,34 +3,68 @@
...
+    fn as_ref(&self) -> &ProtocolMessage<BlobId> {
+        &self.0
+    }
+}
+
+impl AsMut<ProtocolMessage<BlobId>> for InvalidBlobIdMsg {
+    fn as_mut(&mut self) -> &mut ProtocolMessage<BlobId> {
+        &mut self.0
+    }
+}
```
jpcsmith: Remark: This breaks the intent invariant as you can use this to change the intent. For example: `std::mem::swap(invalid_blob_msg.as_mut(), certification.as_mut())`. It seems only to be used for testing, so I would suggest moving the implementations into the test module that is using them for the test, so that this is only ever compiled in when testing.

crates/walrus-service/src/committee.rs
```diff
@@ -405,6 +425,73 @@ where
...
+                })
+                .instrument(tracing::info_span!("node", ?public_key))
+            })
+            .collect::<FuturesUnordered<_>>();
+
+        // get quorum and assemble certificate
+        let weight_threshold = bft::min_n_correct(n_shards).get();
+        let mut total_weight = 0;
+        let mut signed_messages = Vec::with_capacity(self.committee.n_members());
+        let mut signer_indices = Vec::with_capacity(self.committee.n_members());
```
jpcsmith, line 470: Nit: A tighter upper bound on the capacity would be `weight_threshold`.

karlwuest: No, the `weight_threshold` is based on the number of shards, but this is bounded by the number of nodes.

PR #428, opened 2024-05-24
feat(client): consolidate client and aggregator in one `walrus` binary

### LINKED ISSUE #423: Consolidate `walrus-client` and `walrus-aggregator`
Run the aggregator as, for example, `walrus-client aggregator`.

Related to #414.

mlegner:
Also includes the following changes:
- Change default RPC node from devnet to testnet.
- Fail also when wallet defined in config file is unavailable.

Closes #423

### DIFF
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -101,7 +101,8 @@ async fn client() -> Result<()> {
     let args = Args::parse();
     let config: Config = load_configuration(&args.config)?;
     tracing::debug!(?args, ?config, "initializing the client");
-    let wallet = load_wallet_context(&args.wallet.clone().or(config.wallet_config.clone()));
+    let wallet_path = args.wallet.clone().or(config.wallet_config.clone());
+    let wallet = load_wallet_context(&wallet_path);
 
     match args.command {
         Commands::Store { file, epochs } => {
@@ -132,7 +133,7 @@ async fn client() -> Result<()> {
             out,
             rpc_url,
         } => {
-            let client = get_read_client(config, rpc_url, wallet, args.wallet.is_none()).await?;
+            let client = get_read_client(config, rpc_url, wallet, wallet_path.is_none()).await?;
 
             let blob = client.read_blob::<Primary>(&blob_id).await?;
             match out {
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/cli_utils.rs
```diff
@@ -7,9 +7,15 @@ use std::path::PathBuf;
...
+use sui_sdk::{wallet_context::WalletContext, SuiClient, SuiClientBuilder};
+use walrus_sui::client::SuiReadClient;
 
-use crate::client::{default_configuration_paths, Config};
+use crate::client::{default_configuration_paths, Client, Config};
+
+/// Default URL of the devnet RPC node.
+pub const DEVNET_RPC: &str = "https://fullnode.devnet.sui.io:443";
+/// Default RPC URL to connect to if none is specified explicitly or in the wallet config.
+pub const DEFAULT_RPC_URL: &str = DEVNET_RPC;
```
giac-mysten: Suggestion: Maybe change the default RPC to testnet?

crates/walrus-service/bin/client.rs
```diff
@@ -174,6 +148,15 @@ async fn client() -> Result<()> {
                 None => std::io::stdout().write_all(&blob)?,
             }
         }
+        Commands::Aggregator {
+            rpc_url,
+            bind_address,
+        } => {
+            tracing::debug!(?rpc_url, "attempting to run the Walrus aggregator");
+            let client = get_read_client(config, rpc_url, wallet, args.wallet.is_none()).await?;
```
giac-mysten: `args.wallet.is_none()`

Question: is the idea that if the wallet is set, we don't want to fallback to the default wallet?

crates/walrus-service/src/cli_utils.rs
mlegner: Good point, changed this. Now we have `testnet` as a default everywhere except for the local testbed, which has `devnet`.

crates/walrus-service/bin/client.rs
mlegner: Yes. The thought is that if the user explicitly set a wallet config and that is inaccessible, we should assume they actually wanted that wallet and should not silently fall back to the default RPC node.

Now that I reflect on this, we should maybe also return an error if the wallet was set in the configuration file. WDYT?

giac-mysten: >Now that I reflect on this, we should maybe also return an error if the wallet was set in the configuration file. WDYT?

yes, good idea!

mlegner: Done in 8fa4f57a099279b2e5310be78dd8dc57af8975a8.

PR #433, opened 2024-05-24
feat(client): various improvements to the client

### LINKED ISSUE #416: Improve client output and logging
Several possible improvements (can be extended):

- [ ] Provide better output in all realistic failure scenarios. Concretely, we should get more useful output in case the contract or system object are not found. Something along the lines of "make sure you have configured this parameter correctly and are working with the correct Sui network (localnet/devnet/testnet)".
- [ ] Provide more context for DEBUG entries like
  ```
  2024-05-22T21:09:32.692127Z  WARN reserve_and_store_blob{blob_id="MYeuzUPD0_4WoXhDi-zoolrUrzsoKovXiDVOzDusqs0"}:store_metadata_and_pairs: walrus_service::client: storing metadata and pairs failed node=3 error=error sending request for url (http://149.28.12.241:9185/v1/blobs/MYeuzUPD0_4WoXhDi-zoolrUrzsoKovXiDVOzDusqs0/metadata)
  ```
- [ ] We should not log (even in DEBUG) messages that imply success when the operation wasn't successful. In particular, the following messages:
  ```
  2024-05-22T21:19:58.296362Z DEBUG reserve_and_store_blob{blob_id="MYeuzUPD0_4WoXhDi-zoolrUrzsoKovXiDVOzDusqs0"}:store_metadata_and_pairs: walrus_service::client: stored metadata and slivers onto a quorum of nodes elapsed_time=314.378859ms
  2024-05-22T21:19:58.296442Z DEBUG reserve_and_store_blob{blob_id="MYeuzUPD0_4WoXhDi-zoolrUrzsoKovXiDVOzDusqs0"}:store_metadata_and_pairs: walrus_service::client: stored metadata and slivers onto additional nodes elapsed_time=314.468065ms completed_reason=all futures consumed
  ```
- [x] Typo: "one ore more slivers could not be stored on the node"

mlegner:
- check for existence of contract and system objects at the beginning
- return early when too many errors are returned while communicating with storage nodes during a store operation
- make quorum checks more consistent
- improve logging and error messages
- make wallet config field optional in Walrus config
- improve output of client CLI, including a human-readable formatting of the size

Closes #416

### DIFF
--- a/crates/walrus-service/src/client/config.rs
+++ b/crates/walrus-service/src/client/config.rs
@@ -19,7 +19,7 @@ pub struct Config {
     /// Path to the wallet configuration.
     ///
     /// If set, this MUST be an absolute path.
-    #[serde(deserialize_with = "deserialize_wallet_config")]
+    #[serde(default, deserialize_with = "deserialize_wallet_config")]
     pub wallet_config: Option<PathBuf>,
     /// Configuration for the client's network communication.
     #[serde(default)]


crates/walrus-sui/src/client/read_client.rs
```diff
@@ -206,6 +207,27 @@ impl fmt::Debug for SuiReadClient {
     }
 }
 
+/// Checks if the Walrus package and system object exist on chain.
+async fn check_that_objects_exist(
```
giac-mysten: Nit: `Checks that the ...` or `check_if_objects_exist`? :)

crates/walrus-service/bin/client.rs
```diff
@@ -123,9 +123,12 @@ async fn client() -> Result<()> {
                 .reserve_and_store_blob(&std::fs::read(file)?, epochs)
                 .await?;
             println!(
-                "{} Blob stored successfully.\nBlob ID: {}",
+                "{} Blob stored successfully.\n\
+                Unencoded size: {} bytes\nBlob ID: {}\nSui object ID: {}",
                 success(),
-                blob.blob_id
+                blob.size,
```
giac-mysten: Suggestion: can we have an approximate `B, KB, MB, GB` measure here,  instead of the full-precision number of bytes?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -206,6 +207,27 @@ impl fmt::Debug for SuiReadClient {
...
+    sui_client: &SuiClient,
+    system_pkg_id: ObjectID,
+    system_object_id: ObjectID,
+) -> SuiClientResult<()> {
+    if !matches!(
+        sui_client
+            .read_api()
+            .get_object_with_options(system_pkg_id, SuiObjectDataOptions::default().with_type())
+            .await,
+        Ok(SuiObjectResponse { error: None, .. })
```
giac-mysten: Question: you need to use the read api funciton because we don't have an associated contract type for packages?

crates/walrus-service/bin/client.rs
mlegner: Yes, I think this would be nice. I will see if some crate provides a utility function for this or write one myself.

crates/walrus-sui/src/client/read_client.rs
mlegner: Yes, that was my thinking. However, I'm happy to change this if there is a better way. Is there one, @karlwuest? 🙂

crates/walrus-service/bin/client.rs
karlwuest: You could adapt the [function that we used in the bandwidth market for displaying bandwidth](https://github.com/MystenLabs/scion-bandwidth-market/blob/0d786594369ffc38a55a0ab0c724fea522b99984/crates/bandwidth-market/src/types.rs#L156).

crates/walrus-sui/src/client/read_client.rs
karlwuest: I can look at this on Monday for the best way to do it, but I think it would be better to not request the package at all. Instead we can just request the system object with type information, get the package id from its type and compare it to `system_pkg_id`. That not only checks that some object with that object id exists, but that its actually a package, and that the system object belongs to that package.

crates/walrus-sui/src/client/read_client.rs
karlwuest: Or maybe something like `check_system_deployment`? 
Otherwise it sounds like a check for arbitrary objects.

crates/walrus-service/bin/client.rs
mlegner: I've added this functionality in 4398d2d6af07566cbb6cf39f6fda1d6c83fee389, taking inspiration from the linked function.

karlwuest: Looks great!

crates/walrus-sui/src/client/read_client.rs
mlegner: Done in e8995ba81e8ab953e9944fba2f5e74cf07d61d2e.

crates/walrus-sui/src/client/read_client.rs
mlegner: I agree that this would be the ideal solution. However, I couldn't find a simple way of getting an object's package. 🤔

Given that this check is just an optimization for the CLI and not safety/security-relevant, we can also leave it "imperfect" as it is now IMO. The chance that a user _accidentally_ uses an incorrect config that passes this check is virtually zero.

karlwuest: I'm about to send a suggestion, just a sec.

karlwuest: See 9f3c02ccb77878d3522e85fca0abc6335f829e16

karlwuest: Could be further optimized to combine it with the `get_type_parameters` to just have one rpc call, but since we only do this on client initialization, it's not necessary imo.

mlegner: Cool, thanks @karlwuest. I've cherry-picked your commit.

I didn't find any documentation about some of the functions used (e.g., `address`) but it seems to work fine... 🤷

karlwuest: >I didn't find any documentation about some of the functions used (e.g., address) but it seems to work fine... 🤷

That's pretty much the standard with the sui rust sdk 😅 Usually, I end up looking at the implementations.

PR #439, opened 2024-05-27
feat(server): skip storing metadata and slivers if already stored

### LINKED ISSUE #384: On storage nodes, check if metadata and slivers are already stored before verifying and storing them
For metadata, we already have a case for already-stored metadata (https://github.com/MystenLabs/walrus/blob/main/crates/walrus-service/src/server.rs#L193), but that is never triggered as the inner function never returns the corresponding error type.

For slivers, we have neither the corresponding error type nor do we check that case.

halfprice:
Says by the title. Resolves #384. Tested in unit tests.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -690,13 +690,11 @@ impl ServiceState for StorageNode {
                 })?;
 
         if match sliver {
-            Sliver::Primary(_) => shard_storage
-                .is_sliver_stored::<Primary>(blob_id)
-                .context("unable to check sliver existence")?,
-            Sliver::Secondary(_) => shard_storage
-                .is_sliver_stored::<Secondary>(blob_id)
-                .context("unable to check sliver existence")?,
-        } {
+            Sliver::Primary(_) => shard_storage.is_sliver_stored::<Primary>(blob_id),
+            Sliver::Secondary(_) => shard_storage.is_sliver_stored::<Secondary>(blob_id),
+        }
+        .context("unable to check sliver existence")?
+        {
             return Err(StoreSliverError::AlreadyStored(sliver.r#type(), *blob_id));
         }
 
@@ -1495,13 +1493,10 @@ mod tests {
         last_result.expect("function to have completed at least once")
     }
 
-    // Tests StorageNode::store_metadata skip storing metadata if it is already stored.
     #[tokio::test]
-    async fn metadata_already_stored() -> TestResult {
-        let shards: &[&[u16]] = &[&[0], &[1, 2, 3, 4, 5, 6]];
-
+    async fn skip_storing_metadata_if_already_stored() -> TestResult {
         let (cluster, _, blob) =
-            cluster_with_partially_stored_blob(shards, BLOB, |_, _| true).await?;
+            cluster_with_partially_stored_blob(&[&[0]], BLOB, |_, _| true).await?;
 
         assert!(matches!(
             cluster.nodes[0]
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/server.rs
```diff
@@ -333,7 +333,15 @@ impl<S: ServiceState + Send + Sync + 'static> UserServer<S> {
...
+                ServiceResponse::success(StatusCode::OK, "Sliver stored".to_string())
+            }
+            Err(StoreSliverError::AlreadyStored(sliver_type, blob_id)) => {
+                tracing::debug!(%blob_id, %sliver_type, %sliver_pair_index,
+                    "sliver already stored");
+                ServiceResponse::success(
+                    StatusCode::OK,
+                    StoreSliverError::AlreadyStored(sliver_type, blob_id).to_string(),
+                )
             }
```
mlegner: Suggestion: This could be slightly simplified:
```suggestion
            Err(e @ StoreSliverError::AlreadyStored(sliver_type, blob_id)) => {
                tracing::debug!(%blob_id, %sliver_type, %sliver_pair_index,
                    "sliver already stored");
                ServiceResponse::success(StatusCode::OK, e.to_string())
            }
```

crates/walrus-service/src/node.rs
```diff
@@ -682,6 +689,17 @@ impl ServiceState for StorageNode {
...
                 })?;
 
+        if match sliver {
+            Sliver::Primary(_) => shard_storage
+                .is_sliver_stored::<Primary>(blob_id)
+                .context("unable to check sliver existence")?,
+            Sliver::Secondary(_) => shard_storage
+                .is_sliver_stored::<Secondary>(blob_id)
+                .context("unable to check sliver existence")?,
+        } {
```
mlegner: Nit: Minor deduplication:
```suggestion
        if match sliver {
            Sliver::Primary(_) => shard_storage.is_sliver_stored::<Primary>(blob_id),
            Sliver::Secondary(_) => shard_storage.is_sliver_stored::<Secondary>(blob_id),
        }
        .context("unable to check sliver existence")?
        {
```

crates/walrus-service/src/node.rs
```diff
@@ -1475,4 +1494,46 @@ mod tests {
 
         last_result.expect("function to have completed at least once")
     }
+
+    // Tests StorageNode::store_metadata skip storing metadata if it is already stored.
```
mlegner: Nit: I'd remove this comment and optionally extend the function name to something like `skip_storing_metadata_if_already_stored`. Then we get this info also in output in case the test fails. Same for the other function below.

crates/walrus-service/src/node.rs
```diff
@@ -1475,4 +1494,46 @@ mod tests {
...
+    #[tokio::test]
+    async fn metadata_already_stored() -> TestResult {
+        let shards: &[&[u16]] = &[&[0], &[1, 2, 3, 4, 5, 6]];
+
+        let (cluster, _, blob) =
+            cluster_with_partially_stored_blob(shards, BLOB, |_, _| true).await?;
```
mlegner: Suggestion: Given that we test a node-local property, we don't necessarily have to build a full test cluster. Or if we do (because we can reuse the helper functions), we could just create a cluster with a single node and single shard:
```suggestion
        let (cluster, _, blob) =
            cluster_with_partially_stored_blob(&[&[0]], BLOB, |_, _| true).await?;
```
Same below.

crates/walrus-service/src/node.rs
```diff
@@ -1475,4 +1494,46 @@ mod tests {
...
+            cluster.nodes[0]
+                .storage_node
+                .store_sliver(
+                    blob.blob_id(),
+                    assigned_sliver_pair.index(),
+                    &Sliver::Primary(assigned_sliver_pair.primary.clone())
+                )
+                .unwrap_err(),
+            StoreSliverError::AlreadyStored(sliver_type, blob_id)
+                if sliver_type == SliverType::Primary && blob_id == *blob.blob_id()
```
mlegner: Nit: The sliver type could already be included in the error:
```suggestion
            StoreSliverError::AlreadyStored(SliverType::Primary, blob_id)
                if blob_id == *blob.blob_id()
```

crates/walrus-service/src/server.rs
halfprice: Applied

crates/walrus-service/src/node.rs
halfprice: Applied

crates/walrus-service/src/node.rs
halfprice: Thanks for the info! Good to know that even a single shard cluster is doable. I wasn't sure if there is some minimum number of shards required (I should probably test it out), so I just copied the test setup from another test.

crates/walrus-service/src/node.rs
halfprice: Good catch!

PR #440, opened 2024-05-27
feat: add additional endpoints to client daemon

### LINKED ISSUE #424: Expose additional client functionality in the HTTP API
Related to #414.

Follow-up on #422 and #423.

mlegner:
- `walrus publisher` exposes a simple publisher (without access control)
- `walrus daemon` exposes both the aggregator and publisher interfaces

Also fix some CLI help texts and add an alias "write" for "store".

Extend `README.md` with further info about the Walrus client.

Closes #424

### DIFF
--- a/crates/walrus-service/src/daemon.rs
+++ b/crates/walrus-service/src/daemon.rs
@@ -82,10 +82,10 @@ impl<T: Send + Sync + 'static> ClientDaemon<T> {
 impl<T: ContractClient + 'static> ClientDaemon<T> {
     /// Specifies that the daemon should expose the publisher interface (store blobs).
     pub fn with_publisher(mut self, max_body_limit: usize) -> Self {
-        self.router = self
-            .router
-            .route(BLOB_PUT_ENDPOINT, put(store_blob))
-            .route_layer(DefaultBodyLimit::max(max_body_limit));
+        self.router = self.router.route(
+            BLOB_PUT_ENDPOINT,
+            put(store_blob).route_layer(DefaultBodyLimit::max(max_body_limit)),
+        );
         self
     }
 }


crates/walrus-service/src/client/error.rs
```diff
@@ -61,9 +60,6 @@ impl ClientError {
 #[derive(Debug, thiserror::Error)]
 #[error(transparent)]
 pub enum ClientErrorKind {
-    /// The certification of the blob failed.
-    #[error("blob certification failed: {0}")]
-    CertificationFailed(SuiClientError),
```
giac-mysten: Question: I don't recall where this was used. Why is it not needed anymore?

crates/walrus-service/src/daemon.rs
```diff
@@ -0,0 +1,151 @@
...
+                StatusCode::INTERNAL_SERVER_ERROR.into_response()
+            }
+        },
+    }
+}
+
+#[tracing::instrument(level = Level::ERROR, skip_all, fields(epochs))]
+async fn store_blob<T: ContractClient>(
+    State(client): State<Arc<Client<T>>>,
+    Query(PublisherQuery { epochs }): Query<PublisherQuery>,
```
giac-mysten, line 125: Nice, I was wondering how the number of epochs was set, reading from the top :)
Suggestion: Can we have documentation on how the publisher and aggregator urls look like? Maybe with `curl` examples?

crates/walrus-service/src/daemon.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    /// Runs the daemon.
+    pub async fn run(self) -> Result<(), std::io::Error> {
+        let listener = tokio::net::TcpListener::bind(self.network_address).await?;
+        tracing::info!(address = %self.network_address, "the client daemon is starting");
+        axum::serve(
+            listener,
+            self.router
+                .with_state(self.client)
+                .layer(TraceLayer::new_for_http()),
```
giac-mysten, line 73: Suggestion/Important: Running the Axum server like this results in the `DefaultBodyLimit` of `2MB` being applied, which may be too small for the publisher. Can either have (i) a way to set the body limit when calling the cli, or (ii) documentation here and in the cli that specifies what hard limit the publisher will offer?

crates/walrus-service/bin/client.rs
```diff
@@ -159,14 +169,28 @@ async fn client() -> Result<()> {
...
-            aggregator.run(&bind_address).await?;
+            let aggregator = ClientDaemon::new(client, bind_address).with_aggregator();
+            aggregator.run().await?;
+        }
+        Commands::Daemon { bind_address } => {
+            tracing::debug!("attempting to run the Walrus daemon");
+            let client = get_contract_client(config, wallet, args.gas_budget).await?;
+            let publisher = ClientDaemon::new(client, bind_address)
+                .with_aggregator()
+                .with_publisher();
```
giac-mysten: Quesiton: Could it make sense to have a publisher-only option? i.e., if we want to run it on separate machines?

crates/walrus-service/bin/client.rs
```diff
@@ -31,27 +31,27 @@ struct Args {
...
     ///
     /// 1. From this configuration parameter, if set.
-    /// 1. From the path specified in the Walrus configuration, if set.
-    /// 1. From `./client.yaml`.
-    /// 1. From `./sui_config.yaml`.
-    /// 1. From `~/.sui/sui_config/client.yaml`.
+    /// 2. From the path specified in the Walrus configuration, if set.
+    /// 3. From `./client.yaml`.
+    /// 4. From `./sui_config.yaml`.
+    /// 5. From `~/.sui/sui_config/client.yaml`.
```
giac-mysten, line 49: Question: I thought you wanted the `1`s? 😬

crates/walrus-service/src/daemon.rs
mlegner: Thanks, very good point. I will add a CLI option to set this limit.

crates/walrus-service/src/daemon.rs
mlegner: Yes, good idea. I will add this to the long helptexts. Should I also add this to the README or something like that? Or the module docstring?

crates/walrus-service/bin/client.rs
mlegner: I don't quite follow. Do you mean something different from the `Publisher` command above?

crates/walrus-service/bin/client.rs
mlegner: I generally prefer the `1`s in markdown. However, `clap` doesn't deal with this very well, that's why I changed them and added the `verbatim` above.

crates/walrus-service/src/client/error.rs
mlegner: I don't really know, I just noticed that this isn't used anywhere.

But I can look into this further.

crates/walrus-service/bin/client.rs
giac-mysten: uhm my bad, I was reading the wrong line 😬 this works!

crates/walrus-service/src/daemon.rs
giac-mysten: hm... the readme? no strong preference though, the docstring also works

crates/walrus-service/src/client/error.rs
giac-mysten: I think [here](https://github.com/MystenLabs/walrus/blob/main/crates/walrus-service/src/client.rs#L180-L183) is where it should have been. Up to you if `Other` is good enough there :)

crates/walrus-service/src/daemon.rs
mlegner: Done in 663d929974b3cf2a5413c5d478a9dfe3b899c365.

crates/walrus-service/src/client/error.rs
mlegner: Cool, thanks. I re-added the error type and changed the relevant code to
```rust
            .map_err(|e| ClientErrorKind::CertificationFailed(e).into())
```

crates/walrus-service/src/daemon.rs
```diff
@@ -81,8 +81,11 @@ impl<T: Send + Sync + 'static> ClientDaemon<T> {
 
 impl<T: ContractClient + 'static> ClientDaemon<T> {
     /// Specifies that the daemon should expose the publisher interface (store blobs).
-    pub fn with_publisher(mut self) -> Self {
-        self.router = self.router.route(BLOB_PUT_ENDPOINT, put(store_blob));
+    pub fn with_publisher(mut self, max_body_limit: usize) -> Self {
+        self.router = self
+            .router
+            .route(BLOB_PUT_ENDPOINT, put(store_blob))
+            .route_layer(DefaultBodyLimit::max(max_body_limit));
```
giac-mysten: Remark: the limit could be added directly to the `put`... I don't think there is a big difference though.

crates/walrus-service/src/daemon.rs
mlegner: I will add a very simple description to the docstring for now (to get this in the clap helptext). Let's add a more thorough client documentation in a separate PR. We can then include the internal Notion documentation you wrote.

mlegner: Done in 958fee3ae9f0e641a618def89a24e2056a631ac8.

crates/walrus-service/src/daemon.rs
mlegner: Thanks, done in 7fbcb2c37eb208afe69a180dfe0a501752892a5c.

PR #441, opened 2024-05-28
refactor: consolidate sliver verification between node and client

### LINKED ISSUE #220: Consolidate the storage node and storage client sliver verification errors
There is some de-duplication that may be possible in the code that verifies if the received sliver is valid.

See [comment](https://github.com/MystenLabs/walrus/pull/197#discussion_r1555454533) and [comment](https://github.com/MystenLabs/walrus/pull/197#discussion_r1555456350).

mlegner:
Closes #220

### DIFF
--- a/crates/walrus-core/src/encoding/slivers.rs
+++ b/crates/walrus-core/src/encoding/slivers.rs
@@ -111,10 +111,7 @@ impl<T: EncodingAxis> Sliver<T> {
             SliverVerificationError::SliverSizeMismatch
         );
         ensure!(
-            self.symbols.symbol_size()
-                == metadata
-                    .symbol_size(encoding_config)
-                    .expect("the symbol size is checked in `UnverifiedBlobMetadataWithId::verify`"),
+            Ok(self.symbols.symbol_size()) == metadata.symbol_size(encoding_config),
             SliverVerificationError::SymbolSizeMismatch
         );
         let pair_metadata = metadata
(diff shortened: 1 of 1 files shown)


crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -103,19 +107,14 @@ impl<T: EncodingAxis> Sliver<T> {
...
         );
-        let symbol_size_from_metadata = metadata
-            .symbol_size(encoding_config)
-            .expect("the symbol size is checked in `UnverifiedBlobMetadataWithId::verify`");
         ensure!(
-            self.symbols.symbol_size() == symbol_size_from_metadata,
+            self.symbols.symbol_size()
+                == metadata
+                    .symbol_size(encoding_config)
+                    .expect("the symbol size is checked in `UnverifiedBlobMetadataWithId::verify`"),
```
jpcsmith: Question: Given that this relies on the verification of the BlobMetadata, why not accept the verified variant?

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -133,6 +132,21 @@ impl<T: EncodingAxis> Sliver<T> {
         Ok(())
     }
 
+    /// Returns true iff the sliver has the length expected based on the encoding configuration and
+    /// blob size.
+    pub fn has_correct_length(&self, config: &EncodingConfig, blob_size: NonZeroU64) -> bool {
```
jpcsmith: Suggestion: Make private if there is no public use of this.

crates/walrus-core/src/encoding/slivers.rs
mlegner: Good point. However, as we don't need the verified variant for anything else, I've changed this to
```rust
            Ok(self.symbols.symbol_size()) == metadata.symbol_size(encoding_config),
```

PR #443, opened 2024-05-28
feat: implement blob status endpoint

### LINKED ISSUE #418: Add a blob status endpoint to storage nodes
- Add the event id (of the most recent event that has the current certification status) to the `BlobInfo`
- Add an endpoint to the storage node that returns the status and the corresponding event ID
- This allows a client to then independently verify the event if needed

karlwuest:
closes #418

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -135,7 +135,12 @@ impl Client {
     pub async fn get_blob_status(&self, blob_id: &BlobId) -> Result<BlobStatus, NodeError> {
         let url = self.endpoints.blob_status(blob_id);
         let response = self.inner.get(url).send().await.map_err(Kind::Reqwest)?;
-        let blob_status: BlobStatus = response.response_error_for_status().await?.bcs().await?;
+        let blob_status: BlobStatus = response
+            .response_error_for_status()
+            .await?
+            .json()
+            .await
+            .map_err(Kind::Json)?;
 
         Ok(blob_status)
     }
--- a/crates/walrus-service/src/server.rs
+++ b/crates/walrus-service/src/server.rs
@@ -388,7 +388,7 @@ impl<S: ServiceState + Send + Sync + 'static> UserServer<S> {
         match state.blob_status(&blob_id) {
             Ok(Some(status)) => {
                 tracing::debug!("successfully retrieved blob status");
-                (StatusCode::OK, Bcs(status)).into_response()
+                (StatusCode::OK, Json(status)).into_response()
             }
             Ok(None) => {
                 tracing::debug!("blob not found");
--- a/crates/walrus-service/src/storage/blob_info.rs
+++ b/crates/walrus-service/src/storage/blob_info.rs
@@ -54,9 +54,9 @@ impl From<BlobCertificationStatus> for SdkBlobCertificationStatus {
     fn from(value: BlobCertificationStatus) -> Self {
         use BlobCertificationStatus::*;
         match value {
-            Registered => SdkBlobCertificationStatus::Registered,
-            Certified => SdkBlobCertificationStatus::Certified,
-            Invalid => SdkBlobCertificationStatus::Invalid,
+            Registered => Self::Registered,
+            Certified => Self::Certified,
+            Invalid => Self::Invalid,
         }
     }
 }
--- a/crates/walrus-sdk/src/error.rs
+++ b/crates/walrus-sdk/src/error.rs
@@ -42,6 +42,8 @@ impl NodeError {
 pub(crate) enum Kind {
     #[error("failed to decode the response body as BCS")]
     Bcs(#[from] bcs::Error),
+    #[error("failed to decode the response body as JSON")]
+    Json(reqwest::Error),
     #[error(transparent)]
     Reqwest(#[from] reqwest::Error),
     #[error("{inner}: {message}")]


crates/walrus-sdk/Cargo.toml
```diff
@@ -15,6 +15,7 @@ mime = "0.3.17"
 reqwest = { workspace = true, features = ["json"] }
 serde.workspace = true
 serde_json = "1.0.117"
+sui-types.workspace = true
```
jpcsmith, line 18: Question: This seems to pulled in only for the Event ID, which is an int and byte array, could we avoid pulling this in by just creating a simple type here with the same representation?

crates/walrus-sdk/src/client.rs
```diff
@@ -126,6 +131,15 @@ impl Client {
         Ok(metadata)
     }
 
+    /// Requests the status of a blob ID from the node.
+    pub async fn get_blob_status(&self, blob_id: &BlobId) -> Result<BlobStatus, NodeError> {
+        let url = self.endpoints.blob_status(blob_id);
+        let response = self.inner.get(url).send().await.map_err(Kind::Reqwest)?;
+        let blob_status: BlobStatus = response.response_error_for_status().await?.bcs().await?;
```
jpcsmith: Suggestion: This is an endpoint that could nicely be served as JSON, use JSON instead of BCS here.

crates/walrus-service/src/server.rs
```diff
@@ -377,6 +380,27 @@ impl<S: ServiceState + Send + Sync + 'static> UserServer<S> {
...
+            Ok(None) => {
+                tracing::debug!("blob not found");
+                ServiceResponse::<()>::not_found().into_response()
+            }
+            Err(message) => {
+                tracing::error!(%message, "internal server error");
+                ServiceResponse::<()>::internal_error().into_response()
+            }
+        }
+    }
```
jpcsmith: Suggestion: Return a JSON object instead of BCS for the blob status, as it's already pretty human friendly.

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -28,29 +30,57 @@ pub enum BlobCertificationStatus {
...
     }
 }
 
+impl From<BlobCertificationStatus> for SdkBlobCertificationStatus {
+    fn from(value: BlobCertificationStatus) -> Self {
+        use BlobCertificationStatus::*;
+        match value {
+            Registered => SdkBlobCertificationStatus::Registered,
+            Certified => SdkBlobCertificationStatus::Certified,
+            Invalid => SdkBlobCertificationStatus::Invalid,
```
jpcsmith:
```suggestion
            Registered => Self::Registered,
            Certified => Self::Certified,
            Invalid => Self::Invalid,
```

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -28,29 +30,57 @@ pub enum BlobCertificationStatus {
...
-impl BlobCertificationStatus {
-    pub(crate) fn update_status(self, new_status: Self) -> Self {
+impl PartialOrd for BlobCertificationStatus {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        Some(self.cmp(other))
+    }
+}
+
+impl Ord for BlobCertificationStatus {
+    fn cmp(&self, other: &Self) -> Ordering {
```
jpcsmith, line 40: Question: Why the revert to the `Ord` implementation? Or did you not feel that the use of the named function was clearer than a mapping of "higher" or "lower" to what a status could be updated to?

crates/walrus-sdk/Cargo.toml
karlwuest: Sure! Will do that.

crates/walrus-service/src/storage/blob_info.rs
karlwuest: I think `Ord` is pretty clear, since we do have an ordering on the status. However, I'm happy to change to a different name and not have it as an ordering if you prefer. I just don't really know what too call this kind of comparison other than ordering, so it made sense to implement `Ord`. 
The main reason for this change is that we don't just need an update function for the certification status, we need a way to compare them, because the update of the status is not independent from the change of the event id and the end epoch.

crates/walrus-sdk/Cargo.toml
karlwuest: Hmmm... if we want to do that and have nice outputs that provide the same display for the transaction id, I'll need to pull in fastcrypto instead for the base58 encoding. 
Also, if we ever want to do anything with the event id programmatically, we will anyway have to pull in some parts of sui. My inclination would now be to leave it as is. Let me know if you are strongly opposed to this.

jpcsmith: I'm not strongly opposed, I think we'll eventually want to merge both the HTTP and SUI endpoints into one crate anyways.

crates/walrus-service/src/storage/blob_info.rs
jpcsmith: Why doesnt update_status work for that use case ```if self.update_status(new_status) == new_status { // then handle change }```?

karlwuest: Sure, that would work, but to me it seems like an extra step/more cumbersome. Ordering seems more natural to me, since there is a clear hierarchy between the statuses.

karlwuest: If you feel strongly about this, I can revert to update + doing the `self.update_status(new_status) == new_status` check. Lmk.

karlwuest: We then actually have to make the following check, to which I prefer the one with the ordering:

```
if self.status != status && self.status.update(&status) == status
    || self.status == status && self.end_epoch < end_epoch
{
    self.status = status;
    self.current_status_event = status_event;
    self.end_epoch = end_epoch;
}
```

This is equivalent to the current version, but not to the earlier version. I should have probably mentioned before that the business logic is changed compared to before. The intention is to provide values that are consistent with the event that produced the status.

More context for the change:
Previously, it was possible that a certified blob is extended simply by registering the same blob again. Now, it's only extended if the blob is also certified again. At the time when I first implemented the BlobInfo, this seemed more convenient, but we lose the clear connection from event to the Blob properties. If we want clients to be able to verify that a blob is indeed certified, then what we return (and guarantee) should be consistent with the events on chain, i.e. if the storage for a blob is extended by some time, there should exist an event on chain that attests to this.

PR #444, opened 2024-05-28
feat(client): add system information cli subcommand

giac-mysten:
Adds the `info` subcommand, to pretty-print information on the running Walrus system.

### DIFF
--- a/crates/walrus-service/src/cli_utils.rs
+++ b/crates/walrus-service/src/cli_utils.rs
@@ -245,8 +245,9 @@ impl Display for HumanReadableMist {
             let with_separator = thousands_separator(value);
             return write!(f, "{with_separator} MIST");
         }
+        let digits = if value < 10_000_000 { 4 } else { 3 };
         let sui = mist_to_sui(value);
-        write!(f, "{sui:.3} SUI")
+        write!(f, "{sui:.digits$} SUI",)
     }
 }
 
@@ -278,26 +279,6 @@ fn thousands_separator(num: u64) -> String {
         .join(",")
 }
 
-/// Calculates the minimum number of nodes that are necessary to get above the threshold of
-/// Byzantine shards `f`.
-fn min_nodes_above_f(committee: &Committee) -> (usize, usize) {
-    let mut shards_per_node: Vec<_> = committee
-        .members()
-        .iter()
-        .map(|node| node.shard_ids.len())
-        .collect();
-    shards_per_node.sort_unstable();
-    let mut total = 0;
-    let threshold = bft::max_n_faulty(committee.n_shards()).into();
-    for (idx, count) in shards_per_node.iter().rev().enumerate() {
-        total += count;
-        if total > threshold {
-            return (idx + 1, total);
-        }
-    }
-    unreachable!("threshold < n_shards")
-}
-
 /// Pretty-prints information on the running Walrus system.
 pub fn print_walrus_info(committee: &Committee, price_per_unit_size: u64, dev: bool) {
     let n_shards = committee.n_shards();
@@ -323,10 +304,10 @@ pub fn print_walrus_info(committee: &Committee, price_per_unit_size: u64, dev: b
         Maximum blob size: {hr_max_blob} ({max_blob_size_sep} B)
 
         {price_heading}
-        Price per encoded Byte: {price_per_unit_size}
+        Price per encoded Byte: {price_per_unit_size} MIST
         Price to store metadata: {metadata_price}
         Marginal price per additional 1 MiB (w/o metadata): {price_per_mib_input}
-        Total price per max blob ({hr_max_blob} + metadata): {price_max_blob}
+        Total price per max blob ({hr_max_blob}): {price_max_blob}
         ",
         top_heading = "Walrus system information".bold(),
         storage_heading = "Storage nodes".bold().green(),
@@ -354,7 +335,7 @@ pub fn print_walrus_info(committee: &Committee, price_per_unit_size: u64, dev: b
         encoded_blob_length_for_n_shards(n_shards, max_blob_size_for_n_shards(n_shards))
             .expect("we can compute the encoded length of the max blob size");
     let f = bft::max_n_faulty(n_shards);
-    let (min_nodes_above, shards_above) = min_nodes_above_f(committee);
+    let (min_nodes_above, shards_above) = committee.min_nodes_above_f();
 
     printdoc!(
         "
@@ -484,5 +465,18 @@ mod tests {
     }
     fn test_thousands_separator(num: u64, expected: &str) {
         assert_eq!(thousands_separator(num), expected);
+    }
+
+    param_test! {
+        test_human_readable_mist: [
+            ten: (10, "10 MIST"),
+            ten_thousand: (10_000, "10,000 MIST"),
+            million: (1_000_000, "0.0010 SUI"),
+            nine_million: (9_123_456, "0.0091 SUI"),
+            ten_million: (10_123_456, "0.010 SUI"),
+        ]
+    }
+    fn test_human_readable_mist(mist: u64, expected: &str) {
+        assert_eq!(&format!("{}", HumanReadableMist(mist)), expected,)
     }
 }
--- a/crates/walrus-sui/src/types.rs
+++ b/crates/walrus-sui/src/types.rs
@@ -363,6 +363,26 @@ impl Committee {
     pub fn n_members(&self) -> usize {
         self.members.len()
     }
+
+    /// Computes the minimum number of nodes that are necessary to get above the threshold of
+    /// Byzantine shards `f`.
+    pub fn min_nodes_above_f(&self) -> (usize, usize) {
+        let mut shards_per_node: Vec<_> = self
+            .members()
+            .iter()
+            .map(|node| node.shard_ids.len())
+            .collect();
+        shards_per_node.sort_unstable();
+        let mut total = 0;
+        let threshold = bft::max_n_faulty(self.n_shards()).into();
+        for (idx, count) in shards_per_node.iter().rev().enumerate() {
+            total += count;
+            if total > threshold {
+                return (idx + 1, total);
+            }
+        }
+        unreachable!("threshold < n_shards")
+    }
 }
 
 impl TryFrom<&SuiMoveStruct> for Committee {


crates/walrus-core/src/encoding/config.rs
```diff
@@ -411,18 +406,67 @@ pub fn metadata_length_for_n_shards(n_shards: NonZeroU16) -> u64 {
...
+///
+/// Note that on 32-bit architectures, the actual limit can be smaller than that due to the
+/// limited address space.
+#[inline]
+pub fn max_blob_size_for_n_shards(n_shards: NonZeroU16) -> u64 {
+    u64::from(source_symbols_per_blob_for_n_shards(n_shards).get()) * u64::from(MAX_SYMBOL_SIZE)
+}
+
+#[inline]
+pub fn source_symbols_per_blob_for_n_shards(n_shards: NonZeroU16) -> NonZeroU32 {
```
mlegner: Remark: This function is `pub` but not further exported. Could this be made private?
```suggestion
fn source_symbols_per_blob_for_n_shards(n_shards: NonZeroU16) -> NonZeroU32 {
```

crates/walrus-service/bin/client.rs
```diff
@@ -101,6 +103,19 @@ enum Commands {
...
     },
+    /// Print information about the Walrus storage system this client is connected to.
+    Info {
+        /// The URL of the Sui RPC node to use.
+        ///
+        /// If unset, the wallet configuration is applied (if set), or the fullnode at
+        /// `fullnode.testnet.sui.io:443` is used.
+        // NB: Keep this in sync with `walrus_service::cli_utils`.
+        #[clap(short, long)]
+        rpc_url: Option<String>,
```
mlegner: Given that this is now the third instance of the exact same option, it might make sense to deduplicate this. A possible solution would be something like 41c6e7379cfef2df2229488407c045a4ac5b3efb.

crates/walrus-service/src/cli_utils.rs
```diff
@@ -191,6 +206,135 @@ impl std::fmt::Display for HumanReadableBytes {
...
+            HumanReadableBytes(max_encoded_blob_size),
+            max_encoded_blob_size,
+        );
+
+        let f = bft::max_n_faulty(n_shards);
+        println!("\n{}", "(dev) BFT system information".bold().yellow());
+        println!("Tolerated faults (f): {}", f);
+        println!("Quorum threshold (2f+1): {}", 2 * f + 1);
+        println!(
+            "Minimum number of correct nodes (n-f): {}",
```
mlegner: Nit:
```suggestion
            "Minimum number of correct shards (n-f): {}",
```

crates/walrus-service/src/cli_utils.rs
```diff
@@ -191,6 +206,135 @@ impl std::fmt::Display for HumanReadableBytes {
...
+            bft::min_n_correct(n_shards)
+        );
+
```
mlegner: Suggestion: We could also calculate and print the minimum number of nodes that own more than `f` shards.

crates/walrus-service/src/cli_utils.rs
```diff
@@ -191,6 +206,135 @@ impl std::fmt::Display for HumanReadableBytes {
     }
 }
 
+/// Pretty-prints information on the running Walrus system.
+pub fn print_walrus_info(committee: &Committee, price_per_unit_size: u64, dev: bool) {
```
mlegner: Remark/hint: Given the number of `println!` statements in this function, it might make sense to use something like [`printdoc`](https://docs.rs/indoc/latest/indoc/macro.printdoc.html). The [`indoc` crate](https://github.com/dtolnay/indoc) is maintained by David Tolnay and doesn't have any dependencies.

crates/walrus-service/src/cli_utils.rs
```diff
@@ -191,6 +206,135 @@ impl std::fmt::Display for HumanReadableBytes {
...
+
+    println!("\n{}", "Blob size".bold().green());
+    let max_blob_size = max_blob_size_for_n_shards(n_shards);
+    println!(
+        "Maximum blob size: {} ({} B)",
+        HumanReadableBytes(max_blob_size),
+        max_blob_size
+    );
+
+    println!("\n{}", "Current storage price".bold().green());
```
mlegner: Suggestion:
```suggestion
    println!("\n{}", "Current storage price per epoch".bold().green());
```

crates/walrus-service/src/cli_utils.rs
```diff
@@ -191,6 +206,135 @@ impl std::fmt::Display for HumanReadableBytes {
...
+    println!("Price per encoded Byte: {} MIST", price_per_unit_size);
```
mlegner: Nit: This also needs to be kept in sync with the price unit.

crates/walrus-service/src/cli_utils.rs
```diff
@@ -191,6 +206,135 @@ impl std::fmt::Display for HumanReadableBytes {
...
+            .expect("we can encode 1 MiB") as f64
+            / 1e9
+    );
+    println!(
+        "Price per max blob ({}): {:.3} SUI",
+        HumanReadableBytes(max_blob_size),
+        mist_price_per_blob_size(max_blob_size, n_shards, price_per_unit_size)
+            .expect("we can encode the max blob size") as f64
+            / 1e9
+    );
```
mlegner: Remark: The formatting works well for the current prices and limits. However, this might be different in the future. For example, if the price is lowered to 1 MIST / byte, the price per input MiB would be written as `0.010 SUI`, which doesn't have that much precision anymore.

Not quite sure what to do about this. We could create something like `HumanReadableSuiAmount` with formatting similar to the bytes. Or we just leave it as it is and revisit if the parameters change significantly.

crates/walrus-service/src/cli_utils.rs
giac-mysten: 🤦

crates/walrus-service/src/cli_utils.rs
giac-mysten: good point, but I would say the relevant "precision" is given by the USD/SUI exchange. I.e., i don't care if the storage price is `0.010 SUI` or `0.01009 SUI`, if `USD/SUI = 1`. If however `USD/SUI >> 1`, I start to care 😅 So I would keep it as is, for the moment. 

I can add another digit maybe? `{:.4}`

mlegner: Adding another digit for the per MiB price might be a good idea, in particular if we reduce the prices. 👍

crates/walrus-service/src/cli_utils.rs
```diff
@@ -191,6 +206,135 @@ impl std::fmt::Display for HumanReadableBytes {
...
+        max_blob_size
+    );
+
+    println!("\n{}", "Current storage price".bold().green());
+    println!("Price per encoded Byte: {} MIST", price_per_unit_size);
+    println!(
+        "Price per input MiB: {:.3} SUI",
+        mist_price_per_blob_size(1 << 20, n_shards, price_per_unit_size)
+            .expect("we can encode 1 MiB") as f64
+            / 1e9
```
mlegner: I just noticed that the "per MiB" is a bit misleading: This is the price for a single 1 MiB blob, including the metadata. The marginal price per 1 MiB is actually much smaller. This also explains the apparent inconsistency between this value and the max blob below.

Maybe we can split this into a fixed price for the metadata and a marginal price per MiB?

mlegner: Also: can we add a `sui_price_for_blob_size` function for this computation (which is repeated below)?

crates/walrus-service/src/cli_utils.rs
giac-mysten: Actually, I changed my mind and added `HumanReadableMist`, which turns the print into SUI with 3 decimal digits if the amount goes above 1M MIST

crates/walrus-service/src/cli_utils.rs
giac-mysten: Tried to add a clearer price and description, unsure if it improved.

crates/walrus-service/src/cli_utils.rs
```diff
@@ -203,6 +227,201 @@ impl std::fmt::Display for HumanReadableBytes {
...
+
+impl Display for HumanReadableMist {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let value = self.0;
+        if value < 1_000_000 {
+            let with_separator = thousands_separator(value);
+            return write!(f, "{with_separator} MIST");
+        }
+        let sui = mist_to_sui(value);
+        write!(f, "{sui:.3} SUI")
```
mlegner: This has very low precision between `1_000_000` and roughly `10_000_000` as we only have a single significant digit in those cases.
Suggestion: Increase the precision to `4` in that case.

crates/walrus-service/src/cli_utils.rs
```diff
@@ -203,6 +227,201 @@ impl std::fmt::Display for HumanReadableBytes {
...
+
+        {storage_heading}
+        Number of nodes: {n_nodes}
+        Number of shards: {n_shards}
+
+        {size_heading}
+        Maximum blob size: {hr_max_blob} ({max_blob_size_sep} B)
+
+        {price_heading}
+        Price per encoded Byte: {price_per_unit_size}
```
mlegner: Nit:
```suggestion
        Price per encoded Byte: {price_per_unit_size} MIST
```
Alternatively, use the `HumanReadableMist`.

crates/walrus-service/src/cli_utils.rs
```diff
@@ -203,6 +227,201 @@ impl std::fmt::Display for HumanReadableBytes {
...
+        Price to store metadata: {metadata_price}
+        Marginal price per additional 1 MiB (w/o metadata): {price_per_mib_input}
+        Total price per max blob ({hr_max_blob} + metadata): {price_max_blob}
```
mlegner: Nit: The `{hr_max_blob} + metadata` may be a bit misleading. As you already say "total price", we can maybe just remove the " + metadata".

crates/walrus-service/src/cli_utils.rs
```diff
@@ -203,6 +227,201 @@ impl std::fmt::Display for HumanReadableBytes {
...
+    let mut total = 0;
+    let threshold = bft::max_n_faulty(committee.n_shards()).into();
+    for (idx, count) in shards_per_node.iter().rev().enumerate() {
+        total += count;
+        if total > threshold {
+            return (idx + 1, total);
+        }
+    }
+    unreachable!("threshold < n_shards")
+}
```
mlegner: Remark: This function could be useful for other purposes as well.
Suggestion: Make this a method for `Committee`.

crates/walrus-service/bin/client.rs
giac-mysten: done in 41fbac8

crates/walrus-service/src/cli_utils.rs
giac-mysten: good point, done

PR #446, opened 2024-05-28
feat: implement deletion of slivers and metadata of invalid blobs

### LINKED ISSUE #432: Delete slivers and metadata of invalid blobs from the database
part of #257

karlwuest:
closes #432

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -453,6 +453,7 @@ impl StorageNode {
         Ok(())
     }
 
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -451,6 +452,18 @@ impl StorageNode {
 
         Ok(())
     }
+
+    fn on_blob_invalid(
```
jpcsmith, line 457: Suggestion: Add a TODO to handle in-progress syncs for the blob. We'll need to ensure that by the end of this function, the blob is not stored and will not be stored.

PR #448, opened 2024-05-29
feat: return errors from server routes and add openapi spec

jpcsmith:
Included in this PR:

#### Maps node error types to HTTP error codes

The errors returned by the storage node's `ServiceState` interface have been moved to the `node::errors` module. They are directly converted to the appropriate HTTP error response in `server::responses`. To facilitate that, functions which previously returned an `Option<T>` now return an error indicating that the value could not be found.

#### Move routes to their own module

The rest-api's routes are now located in `server::routes`, they are now free standing functions as opposed to static class methods to facilitate the openapi annotations.

#### Consistent instrumentation of routes

Routes are now all instrumented in a consistent manner, and automatically log any errors that are returned. Attributes added to spans are now namespaced under `walrus` (in accordance with moving towards opentelemetry semantics for tracing), example `walrus.sliver.pair_index`.

#### Initial openapi documentations

`server::openapi` and `server::routes` now contains the initial openapi spec. This can be tentatively viewed at `/v1/api` on storage nodes (such as which run using the local testbed).

#### Todo

- [x] Rebase

#### Upcoming PRs

- Provide schema for successful responses (#460).
- Create traces on incoming requests.
- Instrument storage and node.
- Reorganise the node-related modules to allow multiple hierarchies for aggregator and publisher.

### DIFF
--- a/crates/walrus-service/src/server/openapi.rs
+++ b/crates/walrus-service/src/server/openapi.rs
@@ -18,13 +18,14 @@ pub(super) const GROUP_RECOVERY: &str = "Recovery";
         routes::put_sliver,
         routes::get_storage_confirmation,
         routes::get_recovery_symbol,
-        routes::inconsistency_proof
+        routes::inconsistency_proof,
+        routes::get_blob_status
     ),
     components(schemas(BlobIdString, SliverTypeSchema, SliverPairIndexSchema,))
 )]
 pub(super) struct RestApiDoc;
 
-/// Index identifying one of the blob's sliver pairs. AS blobs are encoded into as many pairs of
+/// Index identifying one of the blob's sliver pairs. As blobs are encoded into as many pairs of
 /// slivers as there are shards in the committee, this value must be from 0 to the number of shards
 /// (exclusive).
 #[derive(utoipa::ToSchema)]
--- a/crates/walrus-service/src/server/responses.rs
+++ b/crates/walrus-service/src/server/responses.rs
@@ -119,6 +119,10 @@ macro_rules! rest_api_error {
 
         impl IntoResponse for $enum {
             fn into_response(self) -> Response {
+                // TODO(jsmith): Unify with the errors being attached to traces (#463).
+                if self.status() == StatusCode::INTERNAL_SERVER_ERROR {
+                    tracing::error!(error = ?self, "internal error");
+                }
                 self.to_response()
             }
         }
--- a/crates/walrus-service/src/server/routes.rs
+++ b/crates/walrus-service/src/server/routes.rs
@@ -343,7 +343,7 @@ pub async fn inconsistency_proof<S: SyncServiceState>(
     ),
     tag = openapi::GROUP_READING_BLOBS
 )]
-pub async fn blob_status<S: SyncServiceState>(
+pub async fn get_blob_status<S: SyncServiceState>(
     State(state): State<Arc<S>>,
     Path(BlobIdString(blob_id)): Path<BlobIdString>,
 ) -> Result<ApiSuccess<BlobStatus>, BlobStatusError> {
--- a/crates/walrus-service/src/server.rs
+++ b/crates/walrus-service/src/server.rs
@@ -78,7 +78,7 @@ where
                 routes::INCONSISTENCY_PROOF_ENDPOINT,
                 put(routes::inconsistency_proof),
             )
-            .route(routes::STATUS_ENDPOINT, get(routes::blob_status))
+            .route(routes::STATUS_ENDPOINT, get(routes::get_blob_status))
             .with_state(self.state.clone())
             .layer(TraceLayer::new_for_http().make_span_with(RestApiSpans {
                 address: *network_address,


crates/walrus-service/src/node.rs
```diff
@@ -591,26 +541,27 @@ impl ServiceState for StorageNode {
         self.storage
             .put_verified_metadata(&verified_metadata_with_id)
             .context("unable to store metadata")?;
-        Ok(())
+
+        Ok(true)
```
mlegner, line 630: Question: When does this ever return `false`?
Suggestion: Add check if the metadata is already stored.

crates/walrus-service/src/node.rs
```diff
@@ -421,6 +357,22 @@ impl StorageNode {
 
         Ok(())
     }
+
+    fn is_sliver_index_valid(&self, index: SliverPairIndex) -> bool {
+        index.get() < self.committee_service.get_shard_count().get()
```
mlegner: Question: Why use the `committee_service` and not the `encoding_config`?

crates/walrus-service/src/node.rs
```diff
@@ -619,40 +570,21 @@ impl ServiceState for StorageNode {
...
-        let shard = sliver_pair_index.to_shard_index(self.encoding_config.n_shards(), blob_id);
-        let shard_storage =
-            self.storage
-                .shard_storage(shard)
-                .ok_or_else(|| StoreSliverError::InvalidShard {
-                    shard,
-                    epoch: self.current_epoch(),
-                })?;
+        ensure!(
+            sliver_pair_index.get() < self.encoding_config.n_shards().get(),
```
mlegner: Nit: This could be replaced by
```suggestion
            self.is_sliver_index_valid(sliver_pair_index),
```

crates/walrus-service/src/node.rs
```diff
@@ -421,6 +357,22 @@ impl StorageNode {
 
         Ok(())
     }
+
+    fn is_sliver_index_valid(&self, index: SliverPairIndex) -> bool {
+        index.get() < self.committee_service.get_shard_count().get()
+    }
```
mlegner: Suggestion: Return a `Result<(), IndexOutOfRange>` instead of a boolean and add `#[from]` attributes to the relevant `*OutOfRange` errors. Then this can be called simply as `self.is_sliver_index_valid()?;`.

crates/walrus-service/src/node.rs
```diff
@@ -25,10 +25,10 @@ use walrus_core::{
         SignedMessage,
         StorageConfirmation,
     },
-    metadata::{UnverifiedBlobMetadataWithId, VerificationError, VerifiedBlobMetadataWithId},
+    metadata::{UnverifiedBlobMetadataWithId, VerifiedBlobMetadataWithId},
     BlobId,
     Epoch,
-    InconsistencyProof,
+    InconsistencyProof as PrimaryOrSecondaryInconsistencyProof,
```
mlegner: Question: Why is this renaming necessary?

crates/walrus-service/src/server/openapi.rs
```diff
@@ -0,0 +1,60 @@
...
+        routes::get_storage_confirmation,
+        routes::get_recovery_symbol,
+        routes::inconsistency_proof
+    ),
+    components(schemas(BlobIdString, SliverTypeSchema, SliverPairIndexSchema,))
+)]
+pub(super) struct RestApiDoc;
+
+/// Index identifying to which of the encoded slivers *pairs* constituting the encoding of the blob
+/// the request corresponds.
```
mlegner: Remark: This description is very difficult to parse and understand. Can we reformulate this?

crates/walrus-service/src/server/responses.rs
```diff
@@ -0,0 +1,300 @@
...
+};
+
+/// Successful API response body as JSON.
+///
+/// Contains the HTTP code as well as a message or response object.
+#[derive(Debug, Serialize, Clone, Copy)]
+#[serde(rename_all = "lowercase")]
+pub(super) enum ApiSuccess<T> {
+    Success { code: u16, data: T },
+}
```
mlegner, line 44: Remark: This response structure is currently not present in the OpenAPI spec. For example, for the storage confirmation:
```json
{
  "success": {
    "code": 200,
    "data": {
      "Signed": {
        "serialized_message": [3,1,4],
        "signature": "XXX"
      }
    }
  }
}
```
This could be added in a separate PR though.

Edit: I see that you have a TODO for this in `routes.rs`. Can you add an issue as well?

Also visible here: We may want to add a `#[serde(rename_all = "lowercase")]` to `StorageConfirmation`.

crates/walrus-service/src/server/responses.rs
```diff
@@ -0,0 +1,300 @@
...
+    };
+}
+
+macro_rules! error_description {
+    ($code:ident, @canonical) => {
+        StatusCode::$code
+            .canonical_reason()
+            .expect("code must have a canonical reason")
+            .to_string()
+    };
```
mlegner, line 146: Nit/Suggestion: If we have this special case, we could add something like `@to_string`, which just calls `to_string` on the specific error. Or, add a special case to `rest_api_error` where if the description is missing, it takes the `error.to_string()` by default.

crates/walrus-service/src/server/responses.rs
mlegner, line 1: Nit: There is a slight inconsistency: All descriptions of parameters and successful responses start with an uppercase letter; all error descriptions with a lowercase letter.

mlegner, line 1: Praise: I like your choice of HTTP status codes.

crates/walrus-service/src/server/routes.rs
```diff
@@ -0,0 +1,337 @@
...
+    Ok(ApiSuccess::ok(attestation))
+}
+
+#[tracing::instrument(level = Level::TRACE, skip_all, fields(walrus.blob_id = %blob_id), err)]
+pub async fn blob_status<S: SyncServiceState>(
+    State(state): State<Arc<S>>,
+    Path(BlobIdString(blob_id)): Path<BlobIdString>,
+) -> Result<ApiSuccess<BlobStatus>, BlobStatusError> {
+    Ok(ApiSuccess::ok(state.blob_status(&blob_id)?))
+}
```
mlegner: Remark: This is currently missing the OpenAPI spec.

crates/walrus-service/src/server/routes.rs
```diff
@@ -0,0 +1,337 @@
...
+
+    // TODO(jsmith): Change to CREATED
+    Ok(ApiSuccess::ok("sliver stored successfully"))
+}
+
+/// Get storage confirmation.
+///
+/// Gets a signed storage confirmation from this storage node, indicating that all shards assigned
+/// to this storage node for the current epoch have stored their respective slivers.
+#[tracing::instrument(level = Level::TRACE, skip_all, fields(walrus.blob_id = %blob_id), err)]
```
mlegner: Important: With the current version I have log entries like the following
```
2024-05-29T12:22:24.602347Z ERROR rest-api{address=0.0.0.0:9189}: walrus_service::server::routes: error=the required slivers are not all stored
```

Why did you choose to set the span level to `TRACE`? IMO, this might be interesting at least at the `DEBUG` level, but possibly at all levels (as we had before).

Also: By default, errors are logged as `ERROR`. However, most errors (like the one shown above) are due to a bad request by the user and needlessly pollute the logs. IMO, only internal errors should be logged as `ERROR`, the rest as `WARN` or even just `DEBUG`.

This applies to all the `instrument` attributes in this file.

crates/walrus-service/src/server/routes.rs
```diff
@@ -0,0 +1,337 @@
...
+pub async fn get_metadata<S: SyncServiceState>(
+    State(state): State<Arc<S>>,
+    Path(BlobIdString(blob_id)): Path<BlobIdString>,
+) -> Result<Bcs<VerifiedBlobMetadataWithId>, RetrieveMetadataError> {
+    Ok(Bcs(state.retrieve_metadata(&blob_id)?))
+}
+
+/// Store blob metadata.
+///
+/// Stores the metadata associated with a registered Walrus blob this storage node. This is a
```
mlegner: Nit:
```suggestion
/// Stores the metadata associated with a registered Walrus blob at this storage node. This is a
```

crates/walrus-service/src/server/routes.rs
```diff
@@ -0,0 +1,337 @@
...
+/// pre-requisite for storing the encoded slivers of the blob. The ID of the blob must first be
+/// registered on Sui, after which storing the metadata becomes possible.
+///
+/// This endpoint may return an error if it has not yet been notified from the chain of the blob
+/// being stored.
```
mlegner: Nit: Minor reformulation:
```suggestion
/// This endpoint may return an error if the node has not yet received the registration event from the chain.
```

crates/walrus-service/src/server/routes.rs
```diff
@@ -0,0 +1,337 @@
...
+        RecoverySymbol::Primary(inner) => Ok(Bcs(inner).into_response()),
+        RecoverySymbol::Secondary(inner) => Ok(Bcs(inner).into_response()),
+    }
+}
+
+/// Verify blob inconsistency.
+///
+/// Accepts an inconsistency proof from other storage nodes, verifies it, and returns an attestation
+/// that the specified blob is inconsistent.
+// TODO(jsmith): This endpoint should be a POST endpoint, not a PUT endpoint.
```
mlegner: Agreed. 👍
Could also be renamed to `post_inconsistency_proof` to be consistent with the other functions.

Suggestion: Can you make the change here or create a follow-up issue?

crates/walrus-service/src/node.rs
jpcsmith: I believe this was before the rebase onto Zhe's commit that added that logic. It should now be fixed.

mlegner: Ah, indeed. 👍

crates/walrus-service/src/node.rs
jpcsmith: I was just having a hard time while refactoring remembering whether that was the enum or the encoding variant, so I had renamed it within the file. Reverting.

crates/walrus-service/src/server/openapi.rs
jpcsmith: How's [this](https://github.com/MystenLabs/walrus/blob/e4d36cefe4eb141e28030b4d399d51d2e200b37c/crates/walrus-service/src/server/openapi.rs#L27)?

mlegner: Yes, thank you. Just one more typo here: `slivers` -> `sliver`.

crates/walrus-service/src/server/responses.rs
jpcsmith: The issue here is that if I match the pattern using a `$($variant:tt)*`, then it globs too much, but matches as a `$variant:pat` it cannot then be used to create the `to_string()` expression. I suspect that modifying the macro to use the initial pattern as both the pattern and the enum variant may be "wordy".

crates/walrus-service/src/server/responses.rs
jpcsmith: Yeah, error messages are [typically lowercase](https://doc.rust-lang.org/std/error/trait.Error.html), so any descriptions arising from them are also lowercase. I programatically make them sentence case now.

crates/walrus-service/src/server/routes.rs
jpcsmith: Thanks for catching that, I went in the wrong direction.

The traces are now at the default INFO level, which gives some flexibility for turning them on/off when running from the command line. And should always be on when running in production (which would be run at INFO or higher verbosity).

The errors by default are now logged as DEBUG.

crates/walrus-service/src/server/routes.rs
jpcsmith: Added as #461

crates/walrus-service/src/server/responses.rs
jpcsmith: Added as #460

crates/walrus-service/src/node.rs
```diff
@@ -1041,9 +1038,10 @@ mod tests {
             let metadata = UnverifiedBlobMetadataWithId::new(blob_id, metadata);
 
             // create valid inconsistency proof
-            let inconsistency_proof = PrimaryOrSecondaryInconsistencyProof::Primary(
-                PrimaryInconsistencyProof::new(index, recovery_symbols),
-            );
+            let inconsistency_proof = InconsistencyProof::Primary(PrimaryInconsistencyProof::new(
+                index,
+                recovery_symbols,
+            ));
```
mlegner: Remark: Funny that by shortening the struct name, we actually get one additional line... 😅 🤷

crates/walrus-service/src/server/openapi.rs
```diff
@@ -24,8 +24,9 @@ pub(super) const GROUP_RECOVERY: &str = "Recovery";
 )]
 pub(super) struct RestApiDoc;
 
-/// Index identifying to which of the encoded slivers *pairs* constituting the encoding of the blob
-/// the request corresponds.
+/// Index identifying one of the blob's sliver pairs. AS blobs are encoded into as many pairs of
```
mlegner: Typo:
```suggestion
/// Index identifying one of the blob's sliver pairs. As blobs are encoded into as many pairs of
```

crates/walrus-service/src/server/openapi.rs
```diff
@@ -0,0 +1,61 @@
...
+#[derive(utoipa::OpenApi)]
+#[openapi(
+    paths(
+        routes::get_metadata,
+        routes::put_metadata,
+        routes::get_sliver,
+        routes::put_sliver,
+        routes::get_storage_confirmation,
+        routes::get_recovery_symbol,
+        routes::inconsistency_proof
```
mlegner: Important:
```suggestion
        routes::inconsistency_proof,
        routes::blob_status,
```

crates/walrus-service/src/server/routes.rs
mlegner: Thanks for the change, this makes sense.

Suggestion: log an ERROR for all 500 responses. Maybe by adding a tracing statement [here](https://github.com/MystenLabs/walrus/blob/refactor/server-response/crates/walrus-service/src/server/responses.rs#L120)?

PR #451, opened 2024-05-29
feat: load generator

### LINKED ISSUE #128: Create a simple load generator based on our client implementation

### LINKED ISSUE #374: Expose performance metrics
The orchestrator operates by reading specific metrics exposed by the storage node. Implementing them would enable benchmarks (useful for the paper)

asonnino:
* [x] Ensure walrus nodes expose metrics
* [x] Ensure clients expose metrics
* [x] Ensure we can sustain an end-to-end test run for over 5min
* [x] Provide minimal Grafana dashboard (template/example)
* [x] Cut orchestrator running time by calling binaries directly (instead of `cargo run`) 

closes #128 
closes #374

### DIFF
--- a/crates/walrus-stress/src/generator.rs
+++ b/crates/walrus-stress/src/generator.rs
@@ -148,12 +148,8 @@ async fn new_client(
 ) -> anyhow::Result<WithTempDir<Client<SuiContractClient>>> {
     // Create the client with a separate wallet
     let wallet = wallet_for_testing_from_faucet(network).await?;
-    let sui_read_client = SuiReadClient::new(
-        wallet.as_ref().get_client().await?,
-        config.system_pkg,
-        config.system_object,
-    )
-    .await?;
+    let sui_read_client =
+        SuiReadClient::new(wallet.as_ref().get_client().await?, config.system_object).await?;
     let sui_contract_client = wallet.and_then(|wallet| {
         SuiContractClient::new_with_read_client(wallet, gas_budget, sui_read_client)
     })?;


crates/walrus-stress/src/lib.rs
karlwuest, line 1: Suggestion: delete the file if it's no longer needed.

crates/walrus-stress/src/main.rs
```diff
@@ -0,0 +1,73 @@
...
+    blob_size: NonZeroUsize,
+}
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+    let _ = tracing_subscriber::fmt::try_init();
+
+    let config = Config::load(args.config_path).context("Failed to load client config")?;
+    //let percentage_writes = stress_parameters.load_type.min(100);
```
karlwuest, line 53:
```suggestion
```

crates/walrus-stress/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+    register_histogram_vec_with_registry,
+    register_int_counter_with_registry,
+    CounterVec,
+    HistogramVec,
+    IntCounter,
+    Registry,
+};
+
+const LATENCY_SEC_BUCKETS: &[f64] = &[
+    0.1, 0.15, 0.20, 0.25, 0.5, 0.75, 1., 1.25, 1.5, 1.75, 2., 2.5, 5., 10., 20., 40., 80., 160.,
```
karlwuest: Suggestion:
We probably want to change the buckets a bit. I extended them to 160 seconds, but in my tests, the latencies were mostly bunched up in the 5-10 second bucket. We can probably have one bucket for everything below 2.5 seconds, but then have a bit more granularity above.

crates/walrus-stress/src/metrics.rs
```diff
@@ -0,0 +1,96 @@
...
+];
+
+/// The workload types for the client.
+pub const WRITE_WORKLOAD: &str = "write";
+
+#[derive(Clone)]
+pub struct ClientMetrics {
```
karlwuest, line 26: Remark: Adding the average latency here would be great as well.

crates/walrus-service/src/client.rs
```diff
@@ -634,8 +634,9 @@ impl<T> Client<T> {
     }
 
     /// Returns the inner sui client.
-    pub fn sui_client(&self) -> &T {
-        &self.sui_client
+    // TODO: Ger rid of mut #478
+    pub fn sui_client(&mut self) -> &mut T {
+        &mut self.sui_client
```
karlwuest: Question: why do we need the mutability here?

```suggestion
    // TODO: Get rid of mut #478
    pub fn sui_client(&mut self) -> &mut T {
        &mut self.sui_client
```

crates/walrus-orchestrator/src/protocol/target.rs
```diff
@@ -255,14 +267,42 @@ impl ProtocolCommands for TargetProtocol {
...
-        // Todo: Implement once we have a load generator (#128)
-        vec![]
+        let clients: Vec<_> = instances.into_iter().collect();
+        let load_per_client = (parameters.load / clients.len()).max(1);
+        clients
+            .into_iter()
+            .map(|instance| {
+                let working_dir = &parameters.settings.working_dir;
+                let client_config_path = working_dir.clone().join("client_config.yaml");
+                // let stress_parameters_path = working_dir.clone().join("stress_parameters.yaml");
```
karlwuest:
```suggestion
```

crates/walrus-orchestrator/src/protocol/target.rs
```diff
@@ -255,14 +267,42 @@ impl ProtocolCommands for TargetProtocol {
...
+
+                let run_command = [
+                    &format!("./{BINARY_PATH}/walrus-stress"),
+                    &format!("--target-load {load_per_client}"),
+                    &format!("--config-path {}", client_config_path.display()),
+                    "--n-clients 10",
```
karlwuest: Suggestion: make the number of clients dependent on the `load_per_client`. With 10 clients per stress client, we can currently expect to support at most about 2 writes per second.

crates/walrus-orchestrator/src/protocol/target.rs
```diff
@@ -190,9 +190,20 @@ impl ProtocolCommands for TargetProtocol {
...
+        let serialized_stress_parameters = serde_yaml::to_string(&parameters.client_parameters)
+            .expect("Failed to serialize stress parameters");
+        let stress_parameters_path = parameters
+            .settings
+            .working_dir
+            .join("stress_parameters.yaml");
+        let upload_stress_parameters_command = format!(
+            "echo -e '{serialized_stress_parameters}' > {}",
+            stress_parameters_path.display()
+        );
```
karlwuest: Is this still needed?

```suggestion

```

crates/walrus-orchestrator/src/protocol/target.rs
```diff
@@ -108,24 +108,24 @@ impl ProtocolParameters for ProtocolNodeParameters {}
...
+            blob_size: 10_000,
+            metrics_port: 9584,
         }
     }
 }
 
 impl Display for ProtocolClientParameters {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(f, "{}% Writes", self.load_type)
+        write!(f, "load: {} tx/s", self.target_load)
```
karlwuest:
```suggestion
        write!(f, "load: {} writes/s", self.target_load)
```

crates/walrus-orchestrator/src/protocol/target.rs
asonnino: mm seems I forgot to push the latest commit with cleanups, doing it now

crates/walrus-stress/src/metrics.rs
asonnino: we can compute the average from the count and the sum, right? 
The following metric: `pub latency_s: HistogramVec` always exposes the buckets, the total count, and the total latency

crates/walrus-service/src/client.rs
asonnino: it was to get coins from the faucet but it is no longer needed due to your changes in the test-utils, removing it

asonnino: this was because the later call needed a mutable reference to call a function in the Sui SDK

crates/walrus-stress/src/metrics.rs
karlwuest: Doesn't this only record the buckets? If that also records the sum of the latencies, then we don't need to add anything.

asonnino: prometheus is a bit confusing but whenever it records bucket it also records the sum of latency and the total number of transactions

crates/walrus-service/tests/test_client.rs
```diff
@@ -186,14 +186,14 @@ async fn test_inconsistency(failed_shards: &[usize]) -> anyhow::Result<()> {
         .for_each(|&idx| cluster.cancel_node(idx));
 
     client
-        .as_ref()
+        .as_mut()
```
karlwuest: Nit: I think this can be changed back to `as_ref()` (same below).

crates/walrus-service/tests/test_client.rs
```diff
@@ -145,7 +145,7 @@ async_param_test! {
 async fn test_inconsistency(failed_shards: &[usize]) -> anyhow::Result<()> {
     let _ = tracing_subscriber::fmt::try_init();
 
-    let (_sui_cluster_handle, mut cluster, client) = default_setup().await?;
+    let (_sui_cluster_handle, mut cluster, mut client) = default_setup().await?;
```
karlwuest:
```suggestion
    let (_sui_cluster_handle, mut cluster, client) = default_setup().await?;
```

crates/walrus-sui/src/client.rs
```diff
@@ -234,8 +234,9 @@ impl SuiContractClient {
     }
 
     /// Returns the wallet context used by the client.
-    pub fn wallet(&self) -> &WalletContext {
-        &self.wallet
+    // TODO: get rid of mut #478
+    pub fn wallet(&mut self) -> &mut WalletContext {
```
karlwuest: Can this also be changed back to `&self` ?

PR #455, opened 2024-05-29
test: use global sui test cluster for integration tests

### LINKED ISSUE #429: Change tests to reuse the same sui test cluster
See comment from @jpcsmith  [here](https://github.com/MystenLabs/walrus/pull/411#discussion_r1611388682)

>  It would be good at some point to have a single test cluster for all tests, albeit with different wallets and walrus objects. This would remove the need to spin up a new cluster for each test case in a single integration test file.

karlwuest:
closes #429

### DIFF
--- a/crates/walrus-sui/src/test_utils.rs
+++ b/crates/walrus-sui/src/test_utils.rs
@@ -8,7 +8,6 @@ pub mod system_setup;
 
 use std::{
     collections::BTreeSet,
-    future,
     path::PathBuf,
     sync::{mpsc, OnceLock},
     thread,
@@ -27,7 +26,10 @@ use sui_types::{
     programmable_transaction_builder::ProgrammableTransactionBuilder,
 };
 use test_cluster::{TestCluster, TestClusterBuilder};
-use tokio::{runtime::Builder, sync::Mutex};
+use tokio::{
+    runtime::{Builder, Runtime},
+    sync::Mutex,
+};
 use walrus_core::{
     messages::{Confirmation, ConfirmationCertificate, InvalidBlobCertificate, InvalidBlobIdMsg},
     BlobId,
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/src/test_utils.rs
```diff
@@ -63,6 +70,40 @@ pub fn get_default_invalid_certificate(blob_id: BlobId, epoch: Epoch) -> Invalid
     InvalidBlobCertificate::new(vec![0], invalid_blob_id_msg, signature)
 }
 
+/// Returns a global sui test cluster, after initializing it if it doesn't exist yet.
```
jpcsmith:
```suggestion
/// Returns the global instance of a Sui test cluster and the path to the wallet config. 
///
/// Initialises the test cluster it if it doesn't exist yet.
```

crates/walrus-sui/src/test_utils.rs
```diff
@@ -63,6 +70,40 @@ pub fn get_default_invalid_certificate(blob_id: BlobId, epoch: Epoch) -> Invalid
...
+                .build()
+                .expect("should be able to build runtime")
+                .block_on(async {
+                    let mut test_cluster = sui_test_cluster().await;
+                    let wallet_path = test_cluster.wallet().config.path().to_path_buf();
+                    tx.send((test_cluster, wallet_path))
+                        .expect("can send test cluster");
+                    let () = future::pending().await;
+                })
+        });
```
jpcsmith: Suggestion: Use the thread to create the test cluster, but then return the runtime, cluster, and path and store them in the global variable. That way, we can control proper cleanup of these items.

In the resulting tuple or struct, the ordering of the elements would be important: the runtime must be after the test cluster so that the test cluster is dropped before the runtime is dropped. I am not sure what the ordering would be for a tuple.

crates/walrus-sui/tests/test_walrus_sui.rs
```diff
@@ -162,15 +180,18 @@ async fn test_register_certify_blob() -> anyhow::Result<()> {
...
-    let (package_id, system_object) = publish_with_default_system(&mut wallet).await?;
-    let walrus_client =
-        SuiContractClient::new(wallet, package_id, system_object, 10000000000).await?;
+    let mut wallet = new_wallet_on_global_test_cluster().await?;
+    let (package_id, system_object) = publish_with_default_system(&mut wallet.inner).await?;
+    let walrus_client = wallet
+        .and_then_async(|wallet| {
+            SuiContractClient::new(wallet, package_id, system_object, GAS_BUDGET)
+        })
+        .await?;
```
jpcsmith: Suggestion: Wrap these three in a helper function in this file, since they seem to be always called together.

crates/walrus-sui/src/test_utils.rs
karlwuest: >Use the thread to create the test cluster, but then return the runtime, cluster, and path and store them in the global variable.

I'm not sure if I understand what you mean. Since the runtime blocks, I'm not sure how I would return it from the thread. I could create a cancellation token and `await` on that instead of on `future::pending()`, though. Maybe I'm missing something? Can you clarify what you mean?

karlwuest: Ignore my previous comment, I think I get what you mean now. I was under the impression that we always need a `block_on` to start a runtime (since the documentation states that "This is the runtime’s entry point").

Issue #422, opened 2024-05-23
Client: add functionality to obtain the status of a blob


mlegner:
Based on either the full blob or the blob ID.

Depends on #418.

PR #456, opened 2024-05-29
feat(aggregator): add cache-friendly headers

### LINKED ISSUE #454: Aggregator: add caching-friendly headers

giac-mysten:
Added headers:

- `Cache-control: public, max-age=86400, stale-while-revalidate=3600`
- `ETag: <BLOB_ID>`

Rationale:

Originally I had considered adding `Cache-control: immutable,...`.
This header ensures that the resource is revalidated (queried from the
server) only if it is stale (`max-age` is expired). However, this is
not great for blob invalidation.

Therefore, I opted for not having `immutable`, so the browser checks
if the resource has changed by sending the get request with
the `ETag`. However, this requires the aggregator to respond 
`304 not modified` in case the blob has not been invalidated.

Therefore, this PR currently depends on #422 . Once that is done,
the aggregator can check if there is an `Etag` header in the request,
query the status of the blob, and respond `304 not modified` if it
has not been invalidated.

`stale-while-revalidate` ensures that expired resources can be
revalidated asynchronously.

Closes #454

### DIFF
--- a/crates/walrus-service/src/client/daemon/routes.rs
+++ b/crates/walrus-service/src/client/daemon/routes.rs
@@ -15,7 +15,9 @@ use reqwest::header::{
     ACCESS_CONTROL_ALLOW_METHODS,
     ACCESS_CONTROL_ALLOW_ORIGIN,
     ACCESS_CONTROL_MAX_AGE,
+    CACHE_CONTROL,
     CONTENT_TYPE,
+    ETAG,
     X_CONTENT_TYPE_OPTIONS,
 };
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/daemon.rs
```diff
@@ -106,6 +112,22 @@ async fn retrieve_blob<T: Send + Sync>(
             headers.insert(ACCESS_CONTROL_ALLOW_ORIGIN, HeaderValue::from_static("*"));
             // Prevent the browser from trying to guess the MIME type to avoid dangerous inferences.
             headers.insert(X_CONTENT_TYPE_OPTIONS, HeaderValue::from_static("nosniff"));
+            // Insert headers that help caches distribute Walrus blobs.
+            //
+            // Cache for 1 day, and allow refreshig on the client side. Refreshes use the ETag to
```
gdanezis: One day may even be on the short side for prod, but fine for test. Should we have this as a config?

giac-mysten: Very good point. This PR is parked for the moment, as I wanted to double check all these parameters and headers with PE before merging. Will revisit the default duration with them.

mlegner: How about we merge this as it is now and modify it later if needed?

crates/walrus-service/src/daemon.rs
```diff
@@ -116,6 +118,22 @@ async fn retrieve_blob<T: Send + Sync>(
...
+            // check if the content has changed. This allows invalidated blobs to be removed from
+            // caches. `stale-while-revalidate` allows stale content to be served for 1 hour while
+            // the browser tries to validate it (async revalidation).
+            headers.insert(
+                CACHE_CONTROL,
+                HeaderValue::from_static("public, max-age=86400, stale-while-revalidate=3600"),
```
quepcompute: @giac-mysten any particular significance to the revalidation period of 1hr?

giac-mysten: Sorry @quepcompute , had missed this question. It's just an attempt to strike a balance between cache friendliness and "blockability".

PR #466, opened 2024-05-30
refactor: slightly improve `NodeCommunication`s

mlegner:
- Avoid creating `NodeCommunication`s for all nodes when only a subset is required.
- Only return an error if there is a good reason and the case is actually possible.

### DIFF
--- a/crates/walrus-sui/src/types.rs
+++ b/crates/walrus-sui/src/types.rs
@@ -269,6 +269,8 @@ impl AssociatedContractStruct for StorageNode {
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub struct Committee {
     /// The members of the committee
+    // INV: `!members.is_empty()`
+    // INV: `members.iter().all(|m| !m.shard_ids.is_empty())`
     members: Vec<StorageNode>,
     /// The current epoch
     pub epoch: Epoch,


crates/walrus-service/src/client/communication.rs
```diff
@@ -78,19 +77,20 @@ impl<'a> NodeCommunication<'a> {
...
             %node_index,
             %config.max_node_connections,
             "initializing communication with node"
         );
-        ensure!(
-            !node.shard_ids.is_empty(),
-            ClientErrorKind::InvalidConfig.into()
-        );
-        Ok(Self {
+        Some(Self {
```
giac-mysten, line 93: Important question: IIRC we had decided to go with errors because, for a valid config, there should never be a node without shards. Now this is relaxed. Do you foresee cases in which this is useful?

mlegner: Hmm... I don't have a strong opinion on this. In practice, there shouldn't be a difference in behavior anyway as we check for empty nodes when constructing the `Committee`: https://github.com/MystenLabs/walrus/blob/ml/refactor-node-comms/crates/walrus-sui/src/types.rs#L287

For me then, the errors just make it a bit hard to follow when errors are returned and which (because of the opaque `NodeError` type). Given invariants on `Committee` I would also be open to just panic in some cases.

mlegner: I explicitly added a comment about the invariants on `Committee` in e9b945ba96cfaba5d9b40359d4506390558b94b1.

giac-mysten: works for me!

PR #467, opened 2024-05-30
fix: Make delete blob an atomic op

sadhansood:
This PR makes the two delete operations i.e. `delete_metadata` and `delete_slivers` atomic. While i don't think this is a big issue as these operations would be idempotent if there was a server restart between the first and second delete because the event is going to get reprocessed but it could still lead to some potential confusing state where the slivers exist but its metadata doesn't (or even more confusing would be entry in blob_info cf exist but not in metadata cf)

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -405,8 +405,7 @@ impl StorageNode {
         event_sequence_number: usize,
         event: InvalidBlobID,
     ) -> anyhow::Result<()> {
-        self.storage.delete_metadata(&event.blob_id)?;
-        self.storage.delete_slivers(&event.blob_id)?;
+        self.storage.delete_blob(&event.blob_id)?;
         self.storage
             .maybe_advance_event_cursor(event_sequence_number, &event.event_id)?;
         Ok(())
--- a/crates/walrus-service/src/storage.rs
+++ b/crates/walrus-service/src/storage.rs
@@ -19,6 +19,7 @@ use typed_store::{
     rocks::{
         self,
         errors::typed_store_err_from_bcs_err,
+        DBBatch,
         DBMap,
         MetricConf,
         ReadWriteOptions,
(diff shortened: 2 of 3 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -405,8 +405,7 @@ impl StorageNode {
         event_sequence_number: usize,
         event: InvalidBlobID,
     ) -> anyhow::Result<()> {
-        self.storage.delete_metadata(&event.blob_id)?;
-        self.storage.delete_slivers(&event.blob_id)?;
```
jpcsmith, line 409: Suggestion: Make these functions private if they're no long used publicly.

crates/walrus-service/src/storage.rs
```diff
@@ -205,6 +205,19 @@ impl Storage {
...
+        &self,
+        wb: &mut DBBatch,
+        blob_id: &BlobId,
+        operation: BlobInfoMergeOperand,
+    ) -> Result<(), TypedStoreError> {
+        tracing::debug!(?operation, %blob_id, "updating blob info");
+        wb.partial_merge_batch(&self.blob_info, [(blob_id, operation.to_bytes())])?;
+        Ok(())
+    }
+
```
jpcsmith: Suggestion: Merge with the above merge_update_blob_info, accepting an `Option<DBBatch>` and constructing one when `None`.

crates/walrus-service/src/storage.rs
```diff
@@ -205,6 +205,19 @@ impl Storage {
         batch.write()
     }
 
+    /// Update the blob info for `blob_id` to `new_state` using the merge operation and provided
+    /// write batch
+    fn merge_update_blob_info_using_wb(
+        &self,
+        wb: &mut DBBatch,
```
jpcsmith: Suggestion: Prefer descriptive variable names. Rename all uses of `wb` to `batch`.
```suggestion
        batch: &mut DBBatch,
```

crates/walrus-service/src/storage.rs
```diff
@@ -359,8 +393,8 @@ fn merge_blob_info(
 }
 
 fn deserialize_from_db<'de, T>(val: &'de [u8]) -> Option<T>
-where
-    T: Deserialize<'de>,
+    where
+        T: Deserialize<'de>,
```
jpcsmith: Important: Ensure that your editor is using the correct version of cargo-fmt, or the CI will block your PRs for edits like this (see `rust-toolchain.toml`).

mlegner: I've also added some additional docs on this in #468.

sadhansood: Thanks, the contributing guide was super useful. I think I am all set now.

PR #470, opened 2024-05-31
chore: check if blob is valid before storing metadata

### LINKED ISSUE #449: Return an error for invalid blobs
Add an error in the REST Api for requests for invalid blobs that returns the event id.
Depends on #448

karlwuest:
closes #449 

This PR only returns a specialized error when trying to store metadata, but not for other requests.

Originally, my intention with https://github.com/MystenLabs/walrus/issues/449 was to return such an error for all requests if the blob is invalid, but I realized that we don't do this for non-registered or expired blobs either. The only place where we actually need to check if the blob is invalid is when storing the metadata, so this is also where we now return a specialized error. 

Since the client anyway only gets an http status and error message and doesn't really do anything based on the "type" of error (i.e. the error message), I don't think we need to return a specialized error in most cases. If the client wants more information after getting a `ClientErrorKind::BlobIdDoesNotExist`, it could always check the blob status then.

Do we want to do this for all requests on invalid blobs? 
Adding it everywhere would require checking the blob info for each request and then it would make sense to also return specialized errors for expired/non-registered blobs.

### DIFF
--- a/crates/walrus-service/src/server/responses.rs
+++ b/crates/walrus-service/src/server/responses.rs
@@ -202,7 +202,7 @@ rest_api_error! {
     StoreMetadataError: [
         (NotRegistered, CONFLICT, Self::NotRegistered.to_string()),
         (InvalidMetadata(_), BAD_REQUEST, "the provided metadata cannot be verified"),
-        (InvalidBlob(_), CONFLICT, @canonical),
+        (InvalidBlob(_), CONFLICT, "the blob for the provided metadata is invalid"),
         (BlobExpired, GONE, Self::BlobExpired.to_string()),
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/server/responses.rs
```diff
@@ -202,6 +202,7 @@ rest_api_error! {
     StoreMetadataError: [
         (NotRegistered, CONFLICT, Self::NotRegistered.to_string()),
         (InvalidMetadata(_), BAD_REQUEST, "the provided metadata cannot be verified"),
+        (InvalidBlob(_), CONFLICT, @canonical),
```
jpcsmith: Suggestion: Provide a string description.

PR #471, opened 2024-05-31
feat(client): add option to print output as JSON

mlegner:


crates/walrus-service/bin/client.rs
```diff
@@ -378,9 +384,8 @@ impl BlobIdOutput {
 async fn client() -> Result<()> {
     tracing_subscriber::fmt::init();
     let mut app = App::parse();
-    let mut json = false;
 
-    if let Commands::Json { command_string } = app.command {
+    while let Commands::Json { command_string } = app.command {
```
giac-mysten, line 388: Question: i like this, but do you see any use for nested useless json? 😄

mlegner: Haha, not really. But I didn't see any downside of it and otherwise I would have had to forbid using the `--json` flag with JSON mode.

giac-mysten: >  would have had to forbid using the --json flag with JSON mode.

why forbid it? isn't it overwritten regardless once json mode is enabled?

mlegner: Hmm... You're right. There was some reason why I thought I needed this... 🤔 Maybe I just confused myself there...

giac-mysten: 👍 in any case it's nice to unpack!

PR #474, opened 2024-06-03
feat: add a health_info endpoint in storage server

### LINKED ISSUE #249: Add liveness endpoint to StorageNode, and use in tests to reduce waiting.
No, since we spawn tasks to handle the node loop and REST API. Spawning a task does not yield the execution context, and tests are by default single threaded. A yield here also would not work, as it's not guaranteed that those APIs (specifically the rest APIs) will make sufficient headway before we are rescheduled on the thread, in which case, one or multiple servers may not be ready to receive requests.

If we want move ~it~ the wait into the build(..) of `StorageNodeHandleBuilder`, we can add a new endpoint to the server to check for liveness, and await it coming online.

_Originally posted by @jpcsmith in https://github.com/MystenLabs/walrus/pull/225#discussion_r1557865185_

halfprice:
The main usage of this end point right now is to wait for storage nodes' HTTP server to be up and running before making requests to it.

Closes #249

### DIFF
--- a/crates/walrus-sdk/src/api.rs
+++ b/crates/walrus-sdk/src/api.rs
@@ -134,9 +134,12 @@ impl Ord for BlobStatus {
 #[derive(Debug, Deserialize, Serialize, utoipa::ToSchema)]
 pub struct ServiceHealthInfo {
     /// The uptime of the service.
+    #[schema(value_type = Object)]
     pub uptime: Duration,
     /// The epoch of the storage node.
+    #[schema(value_type = u64)]
     pub epoch: Epoch,
     /// The public key of the storage node.
+    #[schema(value_type = [u8], format = "Base58")]
     pub public_key: PublicKey,
 }
--- a/crates/walrus-service/src/server/routes.rs
+++ b/crates/walrus-service/src/server/routes.rs
@@ -360,7 +360,7 @@ pub async fn get_blob_status<S: SyncServiceState>(
     responses(
         (status = 200, description = "Server is running", body = ApiSuccessServiceHealthInfo),
     ),
-    tag = openapi::GROUP_RECOVERY
+    tag = openapi::GROUP_STATUS
 )]
 pub async fn health_info<S: SyncServiceState>(
     State(state): State<Arc<S>>,
--- a/crates/walrus-service/src/server/openapi.rs
+++ b/crates/walrus-service/src/server/openapi.rs
@@ -22,6 +22,7 @@ use crate::server::responses::ApiSuccess;
 pub(super) const GROUP_STORING_BLOBS: &str = "Writing Blobs";
 pub(super) const GROUP_READING_BLOBS: &str = "Reading Blobs";
 pub(super) const GROUP_RECOVERY: &str = "Recovery";
+pub(super) const GROUP_STATUS: &str = "Status";
 
 #[derive(utoipa::OpenApi)]
 #[openapi(


crates/walrus-service/src/test_utils.rs
```diff
@@ -394,12 +406,34 @@ impl Default for StorageNodeHandleBuilder {
...
+                    return Ok(());
+                }
+                Err(err) => {
+                    info!(
+                        "Ping endpoint {:?} received error {:?}. Retrying in 1s.",
+                        client, err
+                    );
+                }
+            }
+            sleep(Duration::from_secs(1)).await;
```
mlegner: Suggestion: Can we reduce this sleep duration to something like 100 or 200 ms?

crates/walrus-service/src/test_utils.rs
```diff
@@ -394,12 +406,34 @@ impl Default for StorageNodeHandleBuilder {
...
             run_rest_api: Default::default(),
             run_node: Default::default(),
+            wait_until_node_ready: Default::default(),
             contract_service: None,
             test_config: None,
         }
     }
 }
 
+/// Waits until the node is ready by pinging the node's using the node client.
```
mlegner: Typo:
```suggestion
/// Waits until the node is ready by pinging the node using the node client.
```
Or:
```suggestion
/// Waits until the node is ready by querying the node's ping endpoint using the node client.
```

crates/walrus-sdk/src/client.rs
```diff
@@ -418,29 +434,36 @@ mod tests {
...
     where
         F: FnOnce(UrlEndpoints) -> Url,
     {
         let endpoints = UrlEndpoints(Url::parse("http://node.com").unwrap());
         let url = url_fn(endpoints);
-        let expected = format!("http://node.com/v1/blobs/{BLOB_ID}/{expected_path}");
+        let expected = format!("http://node.com/{expected_path}");
 
         assert_eq!(url.to_string(), expected);
     }
```
jpcsmith: Nit: Instead of making the the test more generic and to cover this one case, add a new test case for the health check endpoint. Alternatively, add a helper function that can be called as, for example `blob_path("slivers/1/primary/2")`, and place the format calls in that.

crates/walrus-service/src/node.rs
```diff
@@ -1173,13 +1173,11 @@ mod tests {
             TestCluster::builder()
                 .with_shard_assignment(assignment)
                 .with_system_event_providers(events.clone())
+                .with_wait_until_all_nodes_ready(true)
```
jpcsmith: Suggestion: Remove this argument and just make it the default.

crates/walrus-sdk/src/client.rs
```diff
@@ -408,6 +412,18 @@ impl Client {
...
+
+    /// Sends a ping request to the node.
+    pub async fn ping(&self) -> Result<(), NodeError> {
+        let response = self
+            .inner
+            .get(self.endpoints.ping())
+            .send()
+            .await
+            .map_err(Kind::Reqwest)?;
+        assert!(matches!(response.status(), reqwest::StatusCode::OK));
```
mlegner: Important: Can we replace this `assert` with an `ensure` returning a `NodeErrorKind::StatusWithMessage`? I don't think we should panic when we receive an unexpected reply here.

crates/walrus-service/src/test_utils.rs
```diff
@@ -260,6 +262,12 @@ impl StorageNodeHandleBuilder {
         self
     }
 
+    /// Sets whether to wait until the node is ready.
+    pub fn wait_until_node_ready(mut self, wait_until_node_ready: bool) -> Self {
+        self.wait_until_node_ready = wait_until_node_ready;
+        self
+    }
+
```
jpcsmith: Suggestion: Remove the option to configure this and make it the default if the rest API is started, as I doubt in our tests there will be a use-case where we want the rest api without it being started.

crates/walrus-service/src/test_utils.rs
```diff
@@ -394,12 +406,34 @@ impl Default for StorageNodeHandleBuilder {
...
+        loop {
+            match client.ping().await {
+                Ok(()) => {
+                    return Ok(());
+                }
+                Err(err) => {
+                    info!(
+                        "Ping endpoint {:?} received error {:?}. Retrying in 1s.",
+                        client, err
+                    );
```
jpcsmith: Suggestion: Do not log this, or if you do, log it with trace level. The expectation is that this will repeatedly fail until the endpoint comes online.

crates/walrus-service/src/test_utils.rs
jpcsmith: Suggestion: This can be simplified using 
```rust
while let Err(err) = client.ping().await {
    tokio::time::sleep(..).await;
}
```

jpcsmith: Nit: When importing functions, import their parent module and use `::` notation, for example `time::sleep(..)` or `tokio::time::sleep(..)`.

jpcsmith: Suggestion: Reduce from 1 second to 10 ms or 20 ms.

halfprice: Done.

crates/walrus-service/src/test_utils.rs
halfprice: Updated.

crates/walrus-sdk/src/client.rs
halfprice: The code is updated, and the assert is no longer needed. However, for future reference, the only possible outcome from response is StatusCode::OK. If something else returns, isn't it indicating something wildly wrong happened?

crates/walrus-sdk/src/client.rs
halfprice: New test added.

crates/walrus-service/src/node.rs

crates/walrus-service/src/test_utils.rs
halfprice: Option removed, and we always wait for rest api to start.

crates/walrus-service/src/test_utils.rs
halfprice: Move to `trace` level. Decide to still log it since otherwise the error will be silently ignored.

crates/walrus-service/src/test_utils.rs
halfprice: Done

halfprice: Done

halfprice: Done

crates/walrus-sdk/src/client.rs
halfprice: Look at it again, I realize that it's client side code. I agree that we shouldn't panic no matter what.

crates/walrus-service/src/server/routes.rs
```diff
@@ -348,3 +349,21 @@ pub async fn get_blob_status<S: SyncServiceState>(
...
+/// Get storage health information.
+///
+/// Gets the storage node's health information and basic running stats.
+#[tracing::instrument(skip_all)]
+#[utoipa::path(
+    get,
+    path = openapi::rewrite_route(HEALTH_ENDPOINT),
+    responses(
+        (status = 200, description = "Server is running")
+    ),
```
jpcsmith: Important: This requires the open API response spec associated with it.

crates/walrus-service/src/server/routes.rs
```diff
@@ -348,3 +349,21 @@ pub async fn get_blob_status<S: SyncServiceState>(
 ) -> Result<ApiSuccess<BlobStatus>, BlobStatusError> {
     Ok(ApiSuccess::ok(state.blob_status(&blob_id)?))
 }
+
+/// Get storage health information.
+///
+/// Gets the storage node's health information and basic running stats.
+#[tracing::instrument(skip_all)]
```
jpcsmith, line 356: Important: Instrument this the same as the other endpoints.

crates/walrus-service/src/test_utils.rs
```diff
@@ -400,6 +405,23 @@ impl Default for StorageNodeHandleBuilder {
...
+/// Waits until the node is ready by querying the node's health info endpoint using the node
+/// client.
+async fn wait_for_node_ready(client: &Client) -> anyhow::Result<()> {
+    tokio::time::timeout(Duration::from_secs(10), async {
+        while let Err(err) = client.get_server_health_info().await {
+            tracing::trace!(
+                "Node {:?} is not ready yet. Error: {:?}. Retrying...",
+                client,
+                err
+            );
```
jpcsmith: Suggestion:  Use the client address or node's public key or some other tag, instead of the full debug output of the client.
```suggestion
            tracing::trace!(%err, ?client, "node is not ready yet, retrying...");
```

crates/walrus-sdk/src/api.rs
```diff
@@ -128,3 +129,26 @@ impl Ord for BlobStatus {
...
+pub enum HealthStatus {
+    /// The health status is unknown.
+    #[default]
+    Unknown,
+    /// The service is healthy.
+    Healthy,
+}
+
+/// Represents information about the health of the storage node service.
+#[derive(Debug, Default, Deserialize, Serialize)]
```
jpcsmith:
```suggestion
#[derive(Debug, Default, Deserialize, Serialize, utopia::ToSchema)]
```

crates/walrus-sdk/src/api.rs
```diff
@@ -128,3 +129,26 @@ impl Ord for BlobStatus {
...
+pub struct ServiceHealthInfo {
+    /// The status of the service's health.
+    pub status: HealthStatus,
```
jpcsmith: Question: When is this ever Unknown?

crates/walrus-sdk/src/api.rs
```diff
@@ -128,3 +129,26 @@ impl Ord for BlobStatus {
...
+    /// The uptime of the service.
+    pub uptime: Option<Duration>,
+    /// The epoch of the storage node.
+    pub epoch: Option<Epoch>,
+    /// The public key of the storage node.
+    pub public_key: Option<PublicKey>,
```
jpcsmith: Question: Why options?

crates/walrus-service/src/test_utils.rs
```diff
@@ -400,6 +405,23 @@ impl Default for StorageNodeHandleBuilder {
...
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
jpcsmith: Suggestion: Reduce to 25 ms. 

Remark: This sets a minimum runtime for tests as, if the first lookup fails (which is most likely will), the test must wait at least 100 ms before it's able to make any progress. While some tests required a long a wait due to the number of tasks and futures being created, running in CI, etc. this is not always the case. Setting a value of 100 ms forces tests using the cluster to take at least 100 ms.

Alternatively, for our unit/integration tests we can add something like the following to `server.rs` which will allow using `wait_until_ready()` on an instance of the http server to simply block until the server has bound to a port and is listening.

```diff
diff --git a/crates/walrus-service/src/server.rs b/crates/walrus-service/src/server.rs
index cfdcf13..7aca1e0 100644
--- a/crates/walrus-service/src/server.rs
+++ b/crates/walrus-service/src/server.rs
@@ -37,6 +37,8 @@ const HEADROOM: usize = 128;
 pub struct UserServer<S> {
     state: Arc<S>,
     cancel_token: CancellationToken,
+    #[cfg(any(test, feature = "test-utils"))]
+    ready: tokio::sync::Semaphore,
 }

 impl<S> UserServer<S>
@@ -48,6 +50,8 @@ where
         Self {
             state,
             cancel_token,
+            #[cfg(any(test, feature = "test-utils"))]
+            ready: tokio::sync::Semaphore::new(0),
         }
     }

@@ -86,10 +90,22 @@ where
             }));

         let listener = tokio::net::TcpListener::bind(network_address).await?;
+        #[cfg(any(test, feature = "test-utils"))]
+        self.ready.add_permits(1);
         axum::serve(listener, app)
             .with_graceful_shutdown(self.cancel_token.clone().cancelled_owned())
             .await
     }
+
+    /// Wait until this server begins listening for connections.
+    #[cfg(any(test, feature = "test-utils"))]
+    pub async fn wait_until_ready(&self) {
+        let _ = self
+            .ready
+            .acquire()
+            .await
+            .expect("semaphore cannot be dropped before self");
+    }
 }
```

crates/walrus-service/src/server/routes.rs
halfprice: Added `body` field in responses. Is that what the comment is about? (Sorry, not very familiar with open API, and not sure what does `response spec` mean).

crates/walrus-service/src/server/routes.rs
halfprice: It currently already is. This endpoint doesn't take any other arguments, and it returns an ApiSuccess instead of a Result.

crates/walrus-sdk/src/api.rs
halfprice: Right now, the MockServiceState use it as Unknown since this is the default value. It is created to indicate the uninitialized default value. We could just have `Healthy` as the default value for now, but that will overload `Healthy` to also indicate uninitialized.

crates/walrus-sdk/src/api.rs
halfprice: Optional fields are those that don't have obvious default value, so that None indicates unset. For API messages, optional fields also give us the flexibility to change messages structs later.

crates/walrus-service/src/test_utils.rs
halfprice: I changed the timeout to 25ms.

The reason that I didn't pay too much attention to the duration is that, as this is a test util for tokio tests, I just have the mindset that any sleep duration doesn't matter in terms of test run time. Tokio can simulate time and skip idle duration, and we probably should leverage that (by adding `start_paused = true` to tokio tests).

The wait_until_ready() is also a good approach, but it limited to the case where test body has access to the server. I could do that if that's preferred, but I suspect something similar to the client side solution is still required when we isolate server nodes to their own runtime.

jpcsmith: > The reason that I didn't pay too much attention to the duration is that, as this is a test util for tokio tests, I just have the mindset that any sleep duration doesn't matter in terms of test run time. Tokio can simulate time and skip idle duration, and we probably should leverage that (by adding start_paused = true to tokio tests).

Sure, but currently we use non-emulated networking with backoffs affected by time, so I'm not sure what effect skipping all our timeouts will have. Perhaps add an issue for that?

> The wait_until_ready() is also a good approach, but it limited to the case where test body has access to the server.

It is limited to that, but I was primarily concerned with the runtime of the tests that we already have.

> I could do that if that's preferred, but I suspect something similar to the client side solution is still required when we isolate server nodes to their own runtime.

25 ms should be fine for now.

crates/walrus-sdk/src/api.rs
jpcsmith: Suggestion: Just remove the field completely since it doesn't carry any meaning in production.

crates/walrus-sdk/src/api.rs
jpcsmith: My apologies for not being clearer. From what I can see there is no need for these fields to be options, since their values can currently always be set by the server instance. For the mock server, it can return whatever it wants for those fields, as it's only for testing.

Suggestion: replace the options with their inner type.

crates/walrus-service/src/test_utils.rs
jpcsmith: Suggestion: This will dump the entire debug representation of the client into the logs, use just the address or private key instead.

crates/walrus-service/src/server/routes.rs
jpcsmith: You're right, my apologies for the oversight.

crates/walrus-service/src/server/routes.rs
jpcsmith: Sorry, I meant the specification of the response. In addition to the body call that you added, there also associated code that needs to be added in `server::openapi`. See the code references of `ApiSuccessBlobStatus` for an example of what needs to be done.

Alternatively, add a follow-up issue and assign me to it.

crates/walrus-service/src/test_utils.rs
halfprice: Sorry, missed detailed comment earlier. Just blindly applied the suggestion.

I removed client because it doesn't provide additional information than error. Now the logged line looks like:
`node is not ready yet, retrying... err=error sending request for url (http://127.0.0.1:63533/v1/health)`

crates/walrus-service/src/server/routes.rs
halfprice: I tried following ApiSuccessBlobStatus. Please take a look and see if this is what needed. Happy to create a issue if it sill misses things.

crates/walrus-sdk/src/api.rs
halfprice: Done.

crates/walrus-sdk/src/api.rs
halfprice: I removed option for now. As a note, I think all the required fields may hurt extensibility/compatibility along the way, since this struct is used in an API.

crates/walrus-service/src/test_utils.rs
halfprice: I tried in `retrieves_metadata_from_other_nodes_on_certified_blob_event`.

With simulated time, 10s retry delay in the function caused the test to finish in 380ms.
Without simulated time, 25ms retry delay caused the test to finish in 420ms.

Both hit the retry once.

I can take a stab to update all the tests to use simulated clock.

halfprice: Took a quick look at the existing tests. I saw that they are already leveraging the tokio simulated time pretty good. Sent out https://github.com/MystenLabs/walrus/pull/494 to just enable another test.

crates/walrus-sdk/src/api.rs
jpcsmith: Yea, noted, none of the APIs are final though as this is only devnet...

crates/walrus-service/src/server/routes.rs
jpcsmith: The following should be sufficient to get the api to render. You can test that it renders by running the test `server::openapi::tests::test_openapi_generation_does_not_panic` and attempting to open the `/tmp/api.html` in your browser (automated check pending for #495).

```diff
diff --git a/crates/walrus-sdk/src/api.rs b/crates/walrus-sdk/src/api.rs
index 1cf361b..cc30161 100644
--- a/crates/walrus-sdk/src/api.rs
+++ b/crates/walrus-sdk/src/api.rs
@@ -134,9 +134,12 @@ impl Ord for BlobStatus {
 #[derive(Debug, Deserialize, Serialize, utoipa::ToSchema)]
 pub struct ServiceHealthInfo {
     /// The uptime of the service.
+    #[schema(value_type = Object)]
     pub uptime: Duration,
     /// The epoch of the storage node.
+    #[schema(value_type = u64)]
     pub epoch: Epoch,
     /// The public key of the storage node.
+    #[schema(value_type = [u8], format = "Base58")]
     pub public_key: PublicKey,
 }
diff --git a/crates/walrus-service/src/server/openapi.rs b/crates/walrus-service/src/server/openapi.rs
index 76b55a0..1e5c518 100644
--- a/crates/walrus-service/src/server/openapi.rs
+++ b/crates/walrus-service/src/server/openapi.rs
@@ -22,6 +22,7 @@ use crate::server::responses::ApiSuccess;
 pub(super) const GROUP_STORING_BLOBS: &str = "Writing Blobs";
 pub(super) const GROUP_READING_BLOBS: &str = "Reading Blobs";
 pub(super) const GROUP_RECOVERY: &str = "Recovery";
+pub(super) const GROUP_STATUS: &str = "Status";

 #[derive(utoipa::OpenApi)]
 #[openapi(
diff --git a/crates/walrus-service/src/server/routes.rs b/crates/walrus-service/src/server/routes.rs
index eca1e31..42e73dc 100644
--- a/crates/walrus-service/src/server/routes.rs
+++ b/crates/walrus-service/src/server/routes.rs
@@ -360,7 +360,7 @@ pub async fn get_blob_status<S: SyncServiceState>(
     responses(
         (status = 200, description = "Server is running", body = ApiSuccessServiceHealthInfo),
     ),
-    tag = openapi::GROUP_RECOVERY
+    tag = openapi::GROUP_STATUS
 )]
 pub async fn health_info<S: SyncServiceState>(
     State(state): State<Arc<S>>,
```

halfprice: Thanks for checking. This is what it looks like after open:

Screenshot 2024-06-09 at 9 03 37 PM

Seems ok.

PR #475, opened 2024-06-03
feat(client): add command to obtain a blob's status

### LINKED ISSUE #422: Client: add functionality to obtain the status of a blob

mlegner:
General approach:
- Ask a set of storage nodes that contains at least one correct node for the blob status.
- Construct a vector of distinct responses.
- Go through this vector from "later" statuses (invalid) to "earlier" (nonexistent) and return the first that verifies with the corresponding Sui event.

The CLI command accepts either a blob ID or a file, in which case the blob ID is first computed.

The PR also contains some refactoring commits that can be reviewed separately (and could also be factored out to a separate PR).

Closes #422

### DIFF
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -9,6 +9,7 @@ use std::{
     net::SocketAddr,
     num::{NonZeroU16, NonZeroU64},
     path::{Path, PathBuf},
+    time::Duration,
 };
 
 use anyhow::{anyhow, Result};
@@ -119,9 +120,22 @@ enum Commands {
         rpc_arg: RpcArg,
     },
     /// Get the status of a blob.
+    ///
+    /// This queries multiple storage nodes representing more than a third of the shards for the
+    /// blob status and return the "latest" status (in the life-cycle of a blob) that can be
+    /// verified with an on-chain event.
+    ///
+    /// This does not take into account any transient states. For example, for invalid blobs, there
+    /// is a short period in which some of the storage nodes are aware of the inconsistency before
+    /// this is posted on chain. During this time period, this command would still return a
+    /// "verified" status.
     BlobStatus {
         #[clap(flatten)]
         file_or_blob_id: FileOrBlobId,
+        /// Timeout for status requests to storage nodes.
+        #[clap(short, long, value_parser = humantime::parse_duration, default_value = "1s")]
+        #[serde(default = "default::status_timeout")]
+        timeout: Duration,
         #[clap(flatten)]
         #[serde(default)]
         rpc_arg: RpcArg,
@@ -244,7 +258,33 @@ struct FileOrBlobId {
     blob_id: Option<BlobId>,
 }
 
+impl FileOrBlobId {
+    fn get_or_compute_blob_id(self, encoding_config: &EncodingConfig) -> Result<BlobId> {
+        Ok(match self {
+            FileOrBlobId {
+                blob_id: Some(blob_id),
+                ..
+            } => blob_id,
+            FileOrBlobId {
+                file: Some(file), ..
+            } => {
+                tracing::debug!(
+                    file = %file.display(),
+                    "checking status of blob read from the filesystem"
+                );
+                *encoding_config
+                    .get_blob_encoder(&std::fs::read(&file)?)?
+                    .compute_metadata()
+                    .blob_id()
+            }
+            _ => unreachable!("the CLI enforces exactly one of the options to be present"),
+        })
+    }
+}
+
 mod default {
+    use std::time::Duration;
+
     pub(crate) fn gas_budget() -> u64 {
         500_000_000
     }
@@ -255,6 +295,10 @@ mod default {
 
     pub(crate) fn max_body_size_kib() -> usize {
         10_240
+    }
+
+    pub(crate) fn status_timeout() -> Duration {
+        Duration::from_secs(1)
     }
 }
 
@@ -509,6 +553,7 @@ async fn run_app(app: App) -> Result<()> {
         }
         Commands::BlobStatus {
             file_or_blob_id,
+            timeout,
             rpc_arg: RpcArg { rpc_url },
         } => {
             tracing::debug!(?file_or_blob_id, "getting blob status");
@@ -522,29 +567,10 @@ async fn run_app(app: App) -> Result<()> {
             .await?;
             let client = Client::new_read_client(config, &sui_read_client).await?;
             let file = file_or_blob_id.file.clone();
-            let blob_id = match file_or_blob_id {
-                FileOrBlobId {
-                    blob_id: Some(blob_id),
-                    ..
-                } => blob_id,
-                FileOrBlobId {
-                    file: Some(file), ..
-                } => {
-                    tracing::debug!(
-                        file = %file.display(),
-                        "checking status of blob read from the filesystem"
-                    );
-                    *client
-                        .encoding_config()
-                        .get_blob_encoder(&std::fs::read(&file)?)?
-                        .compute_metadata()
-                        .blob_id()
-                }
-                _ => unreachable!(),
-            };
+            let blob_id = file_or_blob_id.get_or_compute_blob_id(client.encoding_config())?;
 
             let status = client
-                .get_verified_blob_status(&blob_id, &sui_read_client)
+                .get_verified_blob_status(&blob_id, &sui_read_client, timeout)
                 .await?;
             println!(
                 "{}",
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -235,9 +235,6 @@ impl<T: ContractClient> Client<T> {
 }
 
 impl<T> Client<T> {
-    // TODO(mlegner): make configurable.
-    const STATUS_TIMEOUT: Duration = Duration::from_secs(5);
-
     /// Stores the already-encoded metadata and sliver pairs for a blob into Walrus, by sending
     /// sliver pairs to at least 2f+1 shards.
     ///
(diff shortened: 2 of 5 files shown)


crates/walrus-service/src/client.rs
```diff
@@ -524,6 +543,46 @@ impl<T> Client<T> {
...
+        &self,
+        blob_id: &BlobId,
+        read_client: &U,
+    ) -> Result<BlobStatus, ClientError> {
+        let comms = self.node_communications_with_correct_node();
+        let futures = comms
+            .iter()
+            .map(|n| n.get_blob_status(blob_id).instrument(n.span.clone()));
+        let mut requests = WeightedFutures::new(futures);
+        // TODO(mlegner): Instead add more NodeCommunications and do an execute_weight?
```
mlegner: @giac-mysten: WDYT about this?

crates/walrus-service/src/client.rs
```diff
@@ -219,6 +235,9 @@ impl<T: ContractClient> Client<T> {
 }
 
 impl<T> Client<T> {
+    // TODO(mlegner): make configurable.
+    const STATUS_TIMEOUT: Duration = Duration::from_secs(5);
```
mlegner: @giac-mysten: This is only needed if we go the timeout route. Also, maybe a shorter timeout would be sufficient (as correct nodes probably answer within less than a second).

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -193,6 +200,22 @@ impl ReadClient for SuiReadClient {
...
+        let mut events = self
+            .sui_client
+            .event_api()
+            .get_events(event_id.tx_digest)
+            .await?;
+        let index: usize = event_id.event_seq.try_into().map_err(|_| {
+            SuiClientError::Internal(anyhow!("conversion of event sequence number failed"))
+        })?;
+        if index < events.len() {
+            Ok(BlobEvent::try_from(events.swap_remove(index))?)
```
giac-mysten: Question: is the `event_seq` always the index of the event in the array of events returned by `get_events`? Could not find the relevant documentation.

crates/walrus-service/src/client.rs
```diff
@@ -524,6 +543,46 @@ impl<T> Client<T> {
...
+            .execute_time(Self::STATUS_TIMEOUT, self.max_concurrent_status_reads)
+            .await;
+
+        let mut statuses: Vec<_> =
+            // Using `HashSet` to get unique status responses.
+            HashSet::<BlobStatus>::from_iter(requests.take_inner_ok().into_iter())
+                .into_iter()
+                .collect();
+
+        // Going through statuses from later (invalid) to earlier (nonexistent), see implementation
```
giac-mysten, line 567: very nice!

crates/walrus-service/src/client.rs
giac-mysten: question: what is the exact goal here? do we want to be sure that if the blob is invalid, the client gets to know?

If yes, then i think we may need to contact more nodes.
Consider the scenario:

* a blob is invalid, and an `InvalidBlobCertificate` is created.
* this requires `2f+1` signatures (`n-f`), of which `f` can be faulty. This means that `f+1` honest nodes know of the invalid blob.
* if we ask only `f+1` nodes for status, then the only correct one may not have been of the `f+1` honest that know of the invalid blob (previous step).
* Therefore, for quorum intersection, we need to ask `2f+1` nodes for the blob status to be sure we find 1 honest that knows it is invalid.

(did I miss anything?)

So it may be necessary to load all `NodeCommunications`, and `execute_weight` with a `2f+1` weight.

Otherwise, if we don't want timeliness, I think both your approaches work. Wdyt?

crates/walrus-service/bin/client.rs
```diff
@@ -446,6 +507,57 @@ async fn run_app(app: App) -> Result<()> {
...
+                        "checking status of blob read from the filesystem"
+                    );
+                    *client
+                        .encoding_config()
+                        .get_blob_encoder(&std::fs::read(&file)?)?
+                        .compute_metadata()
+                        .blob_id()
+                }
+                _ => unreachable!(),
+            };
```
giac-mysten: Nit: can this be a method on `FileOrBlobId`

crates/walrus-service/bin/client.rs
```diff
@@ -222,6 +230,20 @@ struct RpcArg {
     rpc_url: Option<String>,
 }
 
+#[serde_as]
+#[derive(Debug, Clone, Args, Deserialize)]
+#[serde(rename_all = "snake_case")]
+#[group(required = true, multiple = false)]
```
giac-mysten, line 250: very cool, didn't know about this!

crates/walrus-service/bin/client.rs
```diff
@@ -446,6 +507,57 @@ async fn run_app(app: App) -> Result<()> {
...
+                        file = %file.display(),
+                        "checking status of blob read from the filesystem"
+                    );
+                    *client
+                        .encoding_config()
+                        .get_blob_encoder(&std::fs::read(&file)?)?
+                        .compute_metadata()
+                        .blob_id()
+                }
+                _ => unreachable!(),
```
giac-mysten: Nit:
```suggestion
                _ => unreachable!("the CLI requires exactly one of the options to be present"),
```

crates/walrus-service/src/client.rs
```diff
@@ -524,6 +543,46 @@ impl<T> Client<T> {
         Err(ClientErrorKind::NoMetadataReceived.into())
     }
 
+    /// Gets the blob status from multiple nodes and returns the latest status that can be verified.
+    ///
+    /// The nodes are selected such that at least one correct node is contacted.
```
giac-mysten, line 545: See conversation later about requirements.

crates/walrus-service/src/client.rs
```diff
@@ -657,3 +715,38 @@ impl<T> Client<T> {
...
+
+    let (event_blob_id, event_end_epoch) = match (status, blob_event) {
+        (BlobCertificationStatus::Registered, BlobEvent::Registered(event)) => {
+            (event.blob_id, event.end_epoch)
+        }
+        (BlobCertificationStatus::Certified, BlobEvent::Certified(event)) => {
+            (event.blob_id, event.end_epoch)
+        }
+        (BlobCertificationStatus::Invalid, BlobEvent::InvalidBlobID(event)) => {
+            (event.blob_id, end_epoch)
```
giac-mysten: Question: the "end epoch" for invalidity here is only returned to have the check pass below, correct?

crates/walrus-sui/src/client/read_client.rs
mlegner: Same here. It seems to work in practice... 😅  

@karlwuest do you have any insights/pointers?

crates/walrus-service/src/client.rs
mlegner: Yes that's right. It's a bit hacky, maybe it would be nicer to just include the epoch check in the other two cases and not return the `end_epoch`?

crates/walrus-service/src/client.rs
mlegner: Hmm... I think you are right that we should specify the guarantees that this command provides.

My idea was that we can/should ignore transient states such as the one you describe and just rely on states that are certified on chain. There is only a short time where a blob is known to be invalid without this being posted on chain. And after it is posted on chain, all nodes know about it. WDYT?

I can definitely document this behavior in docstrings/helptexts.

crates/walrus-service/src/client.rs
mlegner: Agreed. Let's document what exact guarantees we provide.

crates/walrus-service/src/client.rs
giac-mysten: up to you, I think it's also fine as is :)

crates/walrus-service/src/client.rs
giac-mysten: >There is only a short time where a blob is known to be invalid without this being posted on chain. And after it is posted on chain, all nodes know about it. WDYT?

I see, that makes sense. I would add this to the docstring/helptext then.


Regarding the distinction between execution by weight or by time, i guess the execution by weight may be a bit more robust? but in any case I still think they re fine

crates/walrus-sui/src/client/read_client.rs
karlwuest: I'm pretty sure that this should be the case, but the documentation is not very good for events. Might be better to just do something like:

```
self.sui_client
    .event_api()
    .get_events(event_id.tx_digest)
    .await?
    .into_iter()
    .find(|event| event.id == event_id).ok_or_else(
            SuiClientError::NoCorrespondingBlobEvent(event_id)
     )?
```
(edit: formatting + into_iter)

mlegner: Ah, very good idea, will do. 🙂

mlegner: This is my final version:
```rust

        self.sui_client
            .event_api()
            .get_events(event_id.tx_digest)
            .await?
            .into_iter()
            .find(|e| e == event_id)
            .and_then(|e| e.try_into().ok())
            .ok_or(SuiClientError::NoCorrespondingBlobEvent(event_id))
```

crates/walrus-service/src/client.rs
mlegner: I think they both have pros and cons. I could add an execution by time *and* weight to get the best from both worlds. The problem with just executing by weight is that this discards any error responses, which we can simply ignore with the time-based execution.

I kept the time-based execution for now, made the timeout configurable, and reduced the default timeout.

crates/walrus-service/src/client.rs
mlegner:
```
    /// Get the status of a blob.
    ///
    /// This queries multiple storage nodes representing more than a third of the shards for the
    /// blob status and return the "latest" status (in the life-cycle of a blob) that can be
    /// verified with an on-chain event.
    ///
    /// This does not take into account any transient states. For example, for invalid blobs, there
    /// is a short period in which some of the storage nodes are aware of the inconsistency before
    /// this is posted on chain. During this time period, this command would still return a
    /// "verified" status.
```

PR #477, opened 2024-06-03
feat(api)!: add success responses and use camelCase in JSON

### LINKED ISSUE #460: Add openapi schema for successful responses

jpcsmith:
This adds the successful responses in the API docs and uses camelCase in JSON instead of snake_case as the former is more standard.

Closes #460

### DIFF
--- a/crates/walrus-core/src/messages.rs
+++ b/crates/walrus-core/src/messages.rs
@@ -49,11 +49,16 @@ impl<T> ProtocolMessage<T> {
 #[serde_with::serde_as]
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
 #[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 pub struct SignedMessage<T> {
     /// The BCS-encoded message.
+    ///
+    /// This is serialized as a base64 string in human-readable encoding formats such as JSON.
     #[serde_as(as = "serde_with::IfIsHumanReadable<serde_with::base64::Base64>")]
+    #[cfg_attr(feature = "utoipa", schema(format = Byte))]
     pub serialized_message: Vec<u8>,
     /// The signature over the BCS encoded message.
+    #[cfg_attr(feature = "utoipa", schema(format = Byte, value_type = [u8]))]
     pub signature: BLS12381Signature,
     #[serde(skip)]
     message_type: PhantomData<T>,
--- a/crates/walrus-core/src/messages/storage_confirmation.rs
+++ b/crates/walrus-core/src/messages/storage_confirmation.rs
@@ -8,9 +8,11 @@ use crate::{messages::IntentType, BlobId, Epoch, PublicKey};
 
 /// Confirmation from a storage node that it has stored the sliver pairs for a given blob.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
-#[serde(rename_all = "lowercase")]
+#[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 pub enum StorageConfirmation {
     /// Confirmation based on the storage node's signature.
+    #[cfg_attr(feature = "utoipa", schema(value_type = SignedMessage::<()>))]
     Signed(SignedStorageConfirmation),
 }
 
--- a/crates/walrus-sdk/Cargo.toml
+++ b/crates/walrus-sdk/Cargo.toml
@@ -20,7 +20,7 @@ thiserror.workspace = true
 tokio.workspace = true
 tower-service = { version = "0.3.2", optional = true }
 tracing = "0.1.40"
-utoipa = "4.2.3"
+utoipa.workspace = true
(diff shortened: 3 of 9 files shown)


crates/walrus-service/src/server/openapi.rs
jpcsmith, line 1: Remark: This file contains several schemas for types that are located in other crates, for example `EventID` from sui and `SliverType`, `SliverPairIndex` and `SignedMessage` from walrus-core.

Question: Should I instead add utoipa as a dependency to walrus-core and have the types natively implement the schema?

crates/walrus-service/src/server/openapi.rs
```diff
@@ -32,24 +53,75 @@ pub(super) struct RestApiDoc;
...
+#[derive(ToSchema)]
+#[schema(as = EventID, rename_all = "camelCase")]
+struct EventIdSchema {
+    #[schema(format = Byte)]
+    tx_digest: Vec<u8>,
+    // u64 represented as a string
+    #[schema(value_type = String)]
+    event_seq: u64,
+}
+
```
jpcsmith, line 64:
```suggestion
// Creates `ToSchema` API implementations and type aliases for `ApiSuccess<T>`. 
// This is required as utoipa's current method for handling generics in schemas is not 
// working for enums. See https://github.com/juhaku/utoipa/issues/835.
```

crates/walrus-service/src/server/openapi.rs
mlegner: Hmm... Good question. How about we add a feature like `web-api` to `walrus-core` and only enable `utoipa` with that feature? Then clients who don't need these derivations can simple keep the feature disabled and don't pull in the additional dependencies.

crates/walrus-sdk/Cargo.toml
```diff
@@ -20,6 +20,7 @@ thiserror.workspace = true
 tokio.workspace = true
 tower-service = { version = "0.3.2", optional = true }
 tracing = "0.1.40"
+utoipa = "4.2.3"
```
mlegner: Nit: Can we add this to the workspace's `Cargo.toml`?

crates/walrus-core/src/messages/storage_confirmation.rs
```diff
@@ -8,6 +8,7 @@ use crate::{messages::IntentType, BlobId, Epoch, PublicKey};
 
 /// Confirmation from a storage node that it has stored the sliver pairs for a given blob.
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "lowercase")]
```
mlegner: Question/nit: Why `lowercase` here and not `camelCase`?

crates/walrus-core/src/messages.rs
```diff
@@ -46,9 +46,12 @@ impl<T> ProtocolMessage<T> {
 }
 
 /// A signed message from a storage node.
+#[serde_with::serde_as]
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
 pub struct SignedMessage<T> {
     /// The BCS-encoded message.
+    #[serde_as(as = "serde_with::IfIsHumanReadable<serde_with::base64::Base64>")]
```
mlegner: Suggestion: Mention in the docstring that this is additionally base64-encoded in some cases.

crates/walrus-service/src/server/openapi.rs
```diff
@@ -32,24 +53,75 @@ pub(super) struct RestApiDoc;
...
+            pub(super) type $alias = ApiSuccess<$name>;
+
+            impl<'r> ToSchema<'r> for $alias {
+                fn schema() -> (&'r str, RefOr<Schema>) {
+                    (stringify!($alias), api_success_alias!(@schema $method $name))
+                }
+            }
+        )*
+    };
+}
```
mlegner: Two questions/suggestions about this macro:
1. It seems to combine two different functionalities. Why not split this into two macros for clarity?
2. Why have the macro handle multiple aliases instead of just having it handle one and then calling it multiple times?

crates/walrus-core/src/messages/storage_confirmation.rs
jpcsmith: No reason, changed.

crates/walrus-service/src/server/openapi.rs
jpcsmith: The former was due to [this](https://danielkeep.github.io/tlborm/book/pat-internal-rules.html) article, though perhaps misapplied since we're not exporting anything here. The latter had no reason, besides it looked a bit less cluttered. Changed both as per the suggestions.

crates/walrus-service/src/server/openapi.rs
jpcsmith: Added utopia as an optional dependency, and derived the Schema based on its presence.

crates/walrus-service/src/server/openapi.rs
mlegner: Ah, very interesting. I hadn't thought about it this way and it definitely makes sense for public macros.

PR #479, opened 2024-06-03
feat: change storage unit to 1KiB

### LINKED ISSUE #445: Change the price in contracts to be per KB/KiB instead of per Byte
This allows to lower the prices further than 1SUI/GB.
This change affects also the `info` command on the storage client (#444).

karlwuest:
closes #445 
also allows specifying price & capacity when deploying a contract using the cli

### DIFF
--- a/crates/walrus-service/src/cli_utils.rs
+++ b/crates/walrus-service/src/cli_utils.rs
@@ -304,7 +304,7 @@ pub fn print_walrus_info(committee: &Committee, price_per_unit_size: u64, dev: b
         Maximum blob size: {hr_max_blob} ({max_blob_size_sep} B)
 
         {price_heading}
-        Price per encoded storage unit: {price_per_unit_size} MIST
+        Price per encoded storage unit: {price_per_unit_size} MIST/KiB
         Price to store metadata: {metadata_price}
         Marginal price per additional 1 MiB (w/o metadata): {price_per_mib_input}
         Total price per max blob ({hr_max_blob}): {price_max_blob}


contracts/blob_store/sources/system.move
```diff
@@ -36,6 +36,10 @@ module blob_store::system {
     /// and therefore 2 x 52 weeks = 2 years.
     const MAX_PERIODS_AHEAD : u64 = 104;
 
+
+    // Keep in sync with the same constant in `crates/walrus-sui/utils.rs`.
+    const BYTES_PER_UNIT_SIZE : u64 = 1_000;
```
mlegner: Question: What was the motivation to go for an SI instead of a binary prefix? AFAIU, binary prefixes are more standard for data while SI prefixes are generally used for bandwidth.

So far we use binary prefixes in the client CLI. But the Red Stuff docs use SI prefixes (cc @giac-mysten). Might be good to use one or the other consistently.

crates/walrus-service/bin/node.rs
```diff
@@ -117,6 +117,12 @@ struct DeploySystemContractArgs {
     /// [default: <WORKING_DIR>/testbed_config.yaml]
     #[clap(long)]
     testbed_config_path: Option<PathBuf>,
+    /// The price to set per unit of storage and epoch.
```
mlegner: Suggestion: Mention where one can find the storage unit.

crates/walrus-service/bin/node.rs
```diff
@@ -117,6 +117,12 @@ struct DeploySystemContractArgs {
...
+    #[arg(long, default_value_t = 10)]
+    price_per_unit: u64,
+    /// The storage capacity to deploy the system with.
```
mlegner: Nit:
```suggestion
    /// The storage capacity in bytes to deploy the system with.
```

crates/walrus-service/src/cli_utils.rs
```diff
@@ -304,7 +304,7 @@ pub fn print_walrus_info(committee: &Committee, price_per_unit_size: u64, dev: b
         Maximum blob size: {hr_max_blob} ({max_blob_size_sep} B)
 
         {price_heading}
-        Price per encoded Byte: {price_per_unit_size} MIST
+        Price per encoded storage unit: {price_per_unit_size} MIST
```
mlegner: Suggestion: Also print the storage unit.

contracts/blob_store/sources/system.move
karlwuest: The choice was mostly arbitrary. I'll change it to 1024 bytes to be more in line with the rest of the cli.

crates/walrus-service/src/cli_utils.rs
mlegner: I don't see any change here. Did you decide against it or just forget to include the change?

karlwuest: Hmm... that's weird, I'm sure that I changed this. Let me check.

karlwuest: I forgot to save the file before committing...

crates/walrus-orchestrator/src/protocol/target.rs
```diff
@@ -166,15 +167,17 @@ impl ProtocolCommands for TargetProtocol {
...
+        let testbed_config = deploy_walrus_contract(DeployTestbedContractParameters {
+            working_dir: parameters.settings.working_dir.as_path(),
+            sui_network: parameters.node_parameters.sui_network,
+            contract_path: parameters.node_parameters.contract_path.clone(),
+            gas_budget: parameters.client_parameters.gas_budget,
+            shards_information: shards,
+            ips: ips.clone(),
+            rest_api_port: parameters.node_parameters.rest_api_port,
+            storage_capacity: 1_000_000_000_000,
+            price_per_unit: 1,
```
asonnino, line 179: @karlwuest what does this price represents, 1 SUI per GB?

mlegner: This is 1 MIST per KiB of encoded data. So 1 SUI per TiB encoded data if I compute this correctly.

karlwuest: That's 1 MIST per storage unit (KiB)

asonnino: perfect!

PR #480, opened 2024-06-03
feat: add prometheus metrics to the storage node

jpcsmith:
This PR adds walrus-specific metrics to the storage node.

TODO:
- [x] HTTP endpoint metrics
- [x] More metrics: errors and latency (may already be available from span latencies)
- [ ] Grafana dashboards

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -408,22 +408,38 @@ impl StorageNode {
         event_sequence_number: usize,
         event: BlobCertified,
     ) -> anyhow::Result<()> {
+        let start = tokio::time::Instant::now();
+        let histogram_set = self.inner.metrics.recover_blob_duration_seconds.clone();
+
         if self.inner.storage.is_stored_at_all_shards(&event.blob_id)? {
             self.inner
                 .mark_event_completed(event_sequence_number, &event.event_id)?;
+
+            histogram_set
+                .with_label_values(&[metrics::STATUS_SKIPPED])
+                .observe(start.elapsed().as_secs_f64());
+
             return Ok(());
         }
 
-        // Slivers, and possible metadata, are not stored.
+        // Slivers, and possibly metadata, are not stored.
         // TODO(jsmith): Do not spawn if there is already a worker for this blob id (#366)
         // TODO(jsmith): Handle cancellation. (#366)
         // TODO(kwuest): Handle epoch change. (#405)
         let synchronizer = BlobSynchronizer::new(event, event_sequence_number, self.inner.clone());
         tokio::spawn(
-            async {
-                if let Err(err) = synchronizer.sync().await {
-                    tracing::error!(?err, "blob synchronizer failed")
-                }
+            async move {
+                // TODO(jsmith): Add a case for recording cancellation (#366).
+                let label = if let Err(err) = synchronizer.run().await {
+                    tracing::error!(?err, "blob synchronizer failed");
+                    metrics::STATUS_FAILURE
+                } else {
+                    metrics::STATUS_SUCCESS
+                };
+
+                histogram_set
+                    .with_label_values(&[label])
+                    .observe(start.elapsed().as_secs_f64());
             }
             .in_current_span(),
         );
--- a/crates/walrus-service/src/node/metrics.rs
+++ b/crates/walrus-service/src/node/metrics.rs
@@ -10,6 +10,12 @@ use prometheus::{
     Registry,
 };
 
+pub(crate) const STATUS_FAILURE: &str = "failure";
+pub(crate) const STATUS_SUCCESS: &str = "success";
+pub(crate) const STATUS_ABORTED: &str = "aborted";
+pub(crate) const STATUS_SKIPPED: &str = "skip";
+pub(crate) const STATUS_INCONSISTENT: &str = "inconsistent";
+
 macro_rules! with_label {
     ($metric:expr, $label:expr) => {
         $metric.with_label_values(&[$label.as_ref()])
@@ -36,7 +42,7 @@ macro_rules! define_node_metric_set {
         ),+ $(,)?
     ) => {
         #[derive(Debug)]
-        pub(super) struct NodeMetricSet {
+        pub(crate) struct NodeMetricSet {
             $($( pub $metric: $metric_type ),*),*
         }
 
@@ -82,5 +88,15 @@ define_node_metric_set! {
             "Time (in seconds) spent processing events",
             &["event_type"]
         ),
+        (
+            recover_blob_duration_seconds,
+            "Time (in seconds) spent recovering blobs",
+            &["status"]
+        ),
+        (
+            recover_blob_part_duration_seconds,
+            "Time (in seconds) spent recovering metadata or slivers of blobs",
+            &["part", "status"]
+        )
     ]
 }
--- a/crates/walrus-service/src/utils.rs
+++ b/crates/walrus-service/src/utils.rs
@@ -3,16 +3,26 @@
 
 //! Utility functions used internally in the crate.
 
-use std::{fmt::Debug, future::Future, num::Saturating, sync::Arc, time::Duration};
+use std::{
+    fmt::Debug,
+    future::Future,
+    num::Saturating,
+    pin::Pin,
+    sync::Arc,
+    task::{ready, Context, Poll},
+    time::Duration,
+};
 
-use futures::FutureExt;
+use futures::{future::FusedFuture, FutureExt};
+use pin_project::pin_project;
+use prometheus::HistogramVec;
 use rand::{
     distributions::{DistIter, Uniform},
     rngs::StdRng,
     Rng,
     SeedableRng,
 };
-use tokio::sync::Semaphore;
+use tokio::{sync::Semaphore, time::Instant};
 
 /// The representation of a backoff strategy.
 pub trait BackoffStrategy {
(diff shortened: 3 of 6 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -350,12 +358,18 @@ impl StorageNode {
 
             match event {
                 BlobEvent::Certified(event) => {
+                    metrics::increment!(self, BlobEventsTotal, label = "certified");
```
gdanezis: I see you use increment quite a bit on blob events below. Here is the issue: as part of ensuring nodes are up to date we might want to check that they have processed the same events / number of events. BUT in case of crash my understanding is that the counters restart at zero, no? So should we use a gauge here and track the total number of events ever processed successfully?

This all ties to the fact that that sequence of events is the common reference to keep the nodes in sync.

jpcsmith: Yea, these are currently useful for getting the rate at which things are being processed, but not the absolute global number. For events, we can store a counter for each event type in the database and convert these metrics to guages, as this will be little overhead. I will convert these.

crates/walrus-service/src/storage/event_cursor_table.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+
+    /// Get the event cursor for `event_type`
+    pub fn get_event_cursor(&self) -> Result<Option<EventID>, TypedStoreError> {
+        if let Some((_, cursor)) = self.inner.get(&CURSOR_KEY)? {
+            Ok(Some(cursor))
+        } else {
+            Ok(None)
+        }
+    }
```
mlegner: Nit: Maybe a bit more concise?
```suggestion
    pub fn get_sequentially_processed_event_count(&self) -> Result<u64, TypedStoreError> {
        Ok(self
            .inner
            .get(&CURSOR_KEY)?
            .map(|(count, _)| count)
            .unwrap_or_default())
    }

    /// Get the event cursor for `event_type`
    pub fn get_event_cursor(&self) -> Result<Option<EventID>, TypedStoreError> {
        Ok(self.inner.get(&CURSOR_KEY)?.map(|(_, cursor)| cursor))
    }
```

crates/walrus-service/src/storage/event_sequencer.rs
```diff
@@ -3,21 +3,18 @@
...
-    collections::BinaryHeap,
+    collections::{hash_map::Entry, BinaryHeap, HashMap},
     sync::{Arc, Mutex},
 };
 
 use sui_types::event::EventID;
 
-/// On receiving a sequence of (sequence_number, EventID) pairs, tracks the highest sequentially
-/// observed sequence_number.
+/// On receiving a sequence of (sequence_number, EventID, BlobEventType) pairs, tracks the highest
```
mlegner: Nit:
```suggestion
/// On receiving a sequence of (sequence_number, EventID, BlobEventType) triples, tracks the highest
```

mlegner: Question: The `sequence_number` is an internal value, right? Because there is also a sequence number internal to Sui events.

Also, I'm not sure I fully understand the invariants enforced by this struct. Is the `next_required_sequence_num` the largest sequence number up to which all have been handled + 1? And do we assume that the caller ensures that no duplicates are inserted?

crates/walrus-service/src/storage/event_sequencer.rs
```diff
@@ -29,79 +26,38 @@ impl EventSequencer {
...
-        self.head.as_ref()
-    }
-
-    /// Returns the furthest EventID that has been observed in a contiguous sequence, and removes
-    /// it from the sequence.
-    pub fn pop(&mut self) -> Option<EventID> {
-        self.head.take()
+        if self.queue.insert(sequence_number, event_id).is_some() {
+            panic!("queued sequence number repeated")
+        }
```
mlegner: Question: Why is the first type of duplicate just logged and the second causes a panic?

crates/walrus-service/src/storage/event_cursor_table.rs
```diff
@@ -0,0 +1,144 @@
...
+            Ok(EventProgress {
+                persisted: count,
+                pending: event_queue.remaining(),
+            })
+        } else {
+            Ok(EventProgress {
+                persisted: 0,
+                pending: event_queue.remaining(),
+            })
+        }
```
mlegner: Nit: This could be deduplicated:
```suggestion
        }

        Ok(EventProgress {
            persisted: count,
            pending: event_queue.remaining(),
        })
```

crates/walrus-service/src/storage/event_cursor_table.rs
```diff
@@ -0,0 +1,144 @@
...
+    for operand_bytes in operands {
+        let (cursor, progress_increment) = bcs::from_bytes::<ProgressMergeOperand>(operand_bytes)
+            .expect("merge operand to be decodable");
+        tracing::debug!("updating {current_val:?} with {cursor:?} (+{progress_increment})");
+
+        let updated_progress = if let Some((progress, _)) = current_val {
+            progress + progress_increment
+        } else {
+            progress_increment
+        };
```
mlegner: Nit: Maybe shorter?
```suggestion
        let updated_progress = current_val
            .map(|(progress, _)| progress)
            .unwrap_or_default()
            + progress_increment;
```

crates/walrus-service/src/node/metrics.rs
```diff
@@ -0,0 +1,103 @@
...
+
+macro_rules! set {
+    ($metric:expr, label = $label:expr, $amount:expr) => {
+        $metric.with_label_values(&[$label.as_ref()]).set($amount)
+    };
+}
+
+pub(super) use add;
+pub(super) use inc;
+pub(super) use set;
```
mlegner: Question: Do we really need these macros? They don't seem to simplify the code much and make it harder to see what is actually happening.

If we want to avoid having to write `.with_label_values(&[some_label.as_ref()])`, we could just have a single macro `with_label!`.

crates/walrus-service/src/node/metrics.rs
```diff
@@ -0,0 +1,103 @@
...
+            .expect("metrics defined at compile time must be valid")
+    }};
+    ($metric_type:ty, $registry:ident, $opts:expr, $label_names:expr) => {{
+        let metric = <$metric_type>::new($opts.into(), $label_names).unwrap();
+        $registry
+            .register(Box::new(metric.clone()))
+            .map(|()| metric)
+            .expect("metrics defined at compile time must be valid")
+    }};
+}
```
mlegner: Nit: How about moving the registration to the other macro below?
```suggestion
macro_rules! node_metric {
    ($metric_type:ty, $registry:ident, $opts:expr) => {{
        <$metric_type>::with_opts($opts).unwrap()
    }};
    ($metric_type:ty, $registry:ident, $opts:expr, $label_names:expr) => {{
        <$metric_type>::new($opts.into(), $label_names).unwrap()
    }};
}
```

```rust
                    $metric: {
                        let metric = node_metric!(
                            $metric_type,
                            registry,
                            Opts::new(stringify!($metric), $descr).namespace("walrus")
                            $(, $labels)?
                        );
                        registry
                        .register(Box::new(metric.clone()))
                        .expect("metrics defined at compile time must be valid");
                        metric
                    }
```

crates/walrus-service/src/storage/event_sequencer.rs
jpcsmith: Yes, I updated the comments and variable names to `index`, let me know if it's clearer now and better differentiates.

crates/walrus-service/src/storage/event_sequencer.rs
jpcsmith: This was due to the latter resulting in an event being lost, but the same could perhaps be said about the former case, so I've unified it now.

crates/walrus-service/src/node/metrics.rs
jpcsmith: It was an attempt to create a unified interface for the operations, such that's its obvious which lines of code are just metrics. I've reverted to the suggestion and directly calling the methods.

crates/walrus-service/src/utils.rs
```diff
@@ -180,10 +190,107 @@ pub(crate) trait FutureHelpers: Future {
...
+        F: FnOnce(Option<&<Self as Future>::Output>) -> [&'static str; N],
+    {
+        Observe::new(self, histograms, get_labels)
+    }
 }
 
 impl<T: Future> FutureHelpers for T {}
 
+#[pin_project(PinnedDrop)]
+pub(crate) struct Observe<Fut, F, const N: usize>
```
mlegner, line 210: Praise: I like this approach with a fused future.

crates/walrus-service/src/utils.rs
```diff
@@ -180,10 +190,107 @@ pub(crate) trait FutureHelpers: Future {
...
+        Self: Sized,
+        F: FnOnce(Option<&<Self as Future>::Output>) -> [&'static str; N],
+    {
+        Observe::new(self, histograms, get_labels)
+    }
 }
 
 impl<T: Future> FutureHelpers for T {}
 
+#[pin_project(PinnedDrop)]
```
mlegner, line 209: Question: Can you briefly explain the use of `pin_project` (on this PR or in a code comment)? I'm not very familiar with the whole topic.

jpcsmith: In vanilla rust, when working with a pinned struct (such as needed for futures), you need to write unsafe code to get a mutable reference to its fields, as the usage of the mutable reference could violate properties of the `Pin` guarantee. `pin_project` provides a way of working with pinned structs that maintains the guarantees (provided that you do not write additional unsafe code). 

In this case, the pinning of the struct only guarantees only that the inner future is pinned. Since we never hand out a `Pin<T>` of any of the other fields, we never guaranteed that they will not move.

Here we also needed a drop implementation. But implementing `Drop` for a struct that was previously handed out as `Pin` provides an `&mut` reference to that struct, which can be used to invalidate the `Pin` guarantee (without unsafe code). `pin_project` prevents you from implementing `Drop` for the struct, and instead provides and delegates to the `PinnedDrop` trait which forces you to maintain the pin guarantee even when implementing the logic for `Drop`.

mlegner: Thanks a lot for the details, @jpcsmith. ❤️

PR #481, opened 2024-06-04
chore: also set client config paths when specifying target path in cli

karlwuest:


crates/walrus-service/src/testbed.rs
```diff
@@ -262,7 +275,7 @@ pub async fn create_storage_node_configs(
     working_dir: &Path,
     testbed_config: TestbedConfig,
     metrics_port: u16,
-    set_config_dir: Option<PathBuf>,
+    set_config_dir: Option<&PathBuf>,
```
giac-mysten: Nit
```suggestion
    set_config_dir: Option<&Path>,
```

crates/walrus-service/bin/node.rs
```diff
@@ -141,8 +141,8 @@ struct GenerateDryRunConfigsArgs {
     metrics_port: u16,
     /// Path of the directory in which the config files will be stored on deployed nodes.
     ///
-    /// If specified, the working directory in the paths contained in the node config and
-    /// the wallet config will be replaced with this directory.
+    /// If specified, the working directory in the paths contained in the node and client configs
+    /// and the wallet configs will be replaced with this directory.
```
giac-mysten:
```suggestion
    /// If specified, the working directory in the paths contained in the node, client,
    /// and wallet configs will be replaced with this directory.
```

crates/walrus-service/src/testbed.rs
```diff
@@ -220,14 +220,15 @@ pub async fn create_client_config(
     system_object: ObjectID,
     working_dir: &Path,
     sui_network: SuiNetwork,
+    set_config_dir: Option<&PathBuf>,
```
giac-mysten:
```suggestion
    set_config_dir: Option<&Path>,
```

karlwuest: That doesn't allow us to then just provide `set_config_dir.as_ref()` as an argument.

karlwuest: Nevermind, I can pass `set_config_dir.as_deref()`

PR #482, opened 2024-06-04
chore: change log level to warn for rpc polling error

karlwuest:


crates/walrus-sui/src/client/read_client.rs
```diff
@@ -294,7 +294,8 @@ where
                 // We retry here, since this error generally (only?)
                 // occurs if the cursor could not be found, but this is
                 // resolved quickly after retrying.
-                tracing::error!("RPC error for otherwise valid RPC call: {}", e);
+                tracing::warn!("RPC error for otherwise valid RPC call: {}", e);
```
mlegner: Suggestion: Mention the backoff and the cursor as the likely cause in the log message.

PR #486, opened 2024-06-04
fix(client): ensure that the binary has the correct exit code

giac-mysten:
Necessary for downstream applications using the JSON mode of the CLI to detect if a command succeded.

### DIFF
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -9,6 +9,7 @@ use std::{
     net::SocketAddr,
     num::{NonZeroU16, NonZeroU64},
     path::{Path, PathBuf},
+    process::ExitCode,
     time::Duration,
 };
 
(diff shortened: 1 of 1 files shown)


crates/walrus-service/bin/client.rs
```diff
@@ -666,9 +667,11 @@ async fn run_app(app: App) -> Result<()> {
...
-pub async fn main() {
+pub async fn main() -> ExitCode {
     if let Err(e) = client().await {
         // Print any error in a (relatively) user-friendly way.
-        eprintln!("{} {:#}", error(), e)
+        eprintln!("{} {:#}", error(), e);
+        return ExitCode::FAILURE;
     }
+    ExitCode::SUCCESS
 }
```
jpcsmith, line 677: Remark: The following would also achieve the same thing I believe, since errors implement the Termination trait. `context(..)` could also be used, but then you may lose the `{:#}` formatting (i think).
```rust
#[tokio::main]
pub async fn main() -> anyhow::Result<()> {
    client().await.map_err(|e| anyhow!("{} {:#}", error(), e)
}
```

giac-mysten: Cool. should I open another pr?

jpcsmith: No need, was just more of FYI/fun-fact.

PR #489, opened 2024-06-05
feat: Make rocksdb options tunable as config

sadhansood:
This PR enables setting different db options for column families as config. Other important reason to do this is to be able to benchmark storage by tuning these parameters easily.

### DIFF
--- a/crates/walrus-service/src/config.rs
+++ b/crates/walrus-service/src/config.rs
@@ -23,6 +23,8 @@ use sui_sdk::types::base_types::ObjectID;
 use walrus_core::keys::{ProtocolKeyPair, ProtocolKeyPairParseError};
 use walrus_sui::{types::NetworkAddress, utils::SuiNetwork};
 
+use crate::storage::DatabaseConfig;
+
 /// Trait for loading configuration from a YAML file.
 pub trait LoadConfig: DeserializeOwned {
     /// Load the configuration from a YAML file located at the provided path.
@@ -43,6 +45,8 @@ pub trait LoadConfig: DeserializeOwned {
 pub struct StorageNodeConfig {
     /// Directory in which to persist the database
     pub storage_path: PathBuf,
+    /// Option config to tune storage db
+    pub db_config: Option<DatabaseConfig>,
     /// Key pair used in Walrus protocol messages
     #[serde_as(as = "PathOrInPlace<Base64>")]
     pub protocol_key_pair: PathOrInPlace<ProtocolKeyPair>,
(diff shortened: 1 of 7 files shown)


crates/walrus-service/src/config.rs
```diff
@@ -23,6 +23,8 @@ use sui_sdk::types::base_types::ObjectID;
 use walrus_core::keys::{ProtocolKeyPair, ProtocolKeyPairParseError};
 use walrus_sui::utils::SuiNetwork;
 
+use crate::storage::DBConfig;
```
jpcsmith: Suggestion: rename to DatabaseConfig

crates/walrus-service/src/storage.rs
```diff
@@ -52,6 +61,152 @@ mod event_sequencer;
 mod shard;
 pub use shard::ShardStorage;
 
+/// Options for configuring a column family.
+#[serde_with::serde_as]
+#[skip_serializing_none]
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct CFOptions {
```
jpcsmith: Suggestion: Rename to DatabaseTableOptions

crates/walrus-service/src/storage/shard.rs
```diff
@@ -168,10 +171,13 @@ impl ShardStorage {
     }
 
     /// Returns the name and options for the column family for a shard with the specified index.
-    pub(crate) fn slivers_column_family_options(id: ShardIndex) -> (String, Options) {
+    pub(crate) fn slivers_column_family_options(
+        id: ShardIndex,
+        db_config: &DBConfig,
+    ) -> (String, Options) {
         // TODO(jsmith): Optimize for sliver storage (#65).
```
jpcsmith:
```suggestion
```

crates/walrus-service/src/storage.rs
```diff
@@ -52,6 +61,152 @@ mod event_sequencer;
...
+            blob_garbage_collection_age_cutoff: Some(0.5),
+            blob_garbage_collection_force_threshold: Some(0.5),
+            blob_compaction_read_ahead_size: Some(10 << 20),
+            write_buffer_size: Some(256 << 20),
+            target_file_size_base: Some(4 << 20),
+            max_bytes_for_level_base: Some(32 << 20),
+        }
+    }
+
+    pub fn configure_options(&self, options: &mut Options) {
```
jpcsmith: Suggestion: Replace with `to_options(&self) -> Options`, as we use Options::default() everywhere with the exception of when adding merge operators.

Then places where we don't need to modify this, we can just return `config.field_name.to_options()`. In the case we need to make modifications, we can store the options to a variable and modify it afterwards.

PR #490, opened 2024-06-05
feat: allow specifying db path when creating configs

### LINKED ISSUE #488: Allow specifying db path when creating testbed configs
See [here](https://github.com/MystenLabs/sui-operations/pull/3820#discussion_r1626336303)

karlwuest:
closes #488 

See [here](https://github.com/MystenLabs/sui-operations/pull/3820#discussion_r1625257684) for context

Also makes previously added cli options nicer to use (no `"` needed around list of ips, deterministic keys as flag).

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -293,12 +293,10 @@ mod commands {
             listening_ips,
             metrics_port,
             set_config_dir.as_deref(),
+            set_db_path.as_deref(),
         )
         .await?;
-        for (i, mut storage_node_config) in storage_node_configs.into_iter().enumerate() {
-            if let Some(ref db_path) = set_db_path {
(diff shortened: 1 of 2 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -288,7 +295,10 @@ mod commands {
             set_config_dir.as_deref(),
         )
         .await?;
-        for (i, storage_node_config) in storage_node_configs.into_iter().enumerate() {
+        for (i, mut storage_node_config) in storage_node_configs.into_iter().enumerate() {
+            if let Some(ref db_path) = set_db_path {
+                storage_node_config.storage_path.clone_from(db_path);
+            }
```
mlegner: Suggestion: Move this into the `create_storage_node_configs` function to have all configs in the same place.

PR #494, opened 2024-06-06
test: re-enable recovers_sliver_from_a_small_set

halfprice:
By setting start_paused = true. The test finishes in 0.5s.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1362,8 +1362,7 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test]
-    #[ignore = "syncing all symbols takes several seconds"]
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1496,14 +1495,11 @@ mod tests {
 
         let _ = tokio::time::timeout(duration, async {
             loop {
-                let result = func_to_retry().await;
-                if result.is_ok() {
-                    last_result = Some(func_to_retry().await);
```
halfprice, line 1501: I'm guessing calling func_to_retry() even if the previous call was successful is not intended?

jpcsmith: You're right, that wasn't intentional.

PR #502, opened 2024-06-07
feat(client)!: check status before storing and improve output

### LINKED ISSUE #383: Client: check if a blob already exists in Walrus before reserving space and sending data to nodes
The client currently doesn't check if a blob already exists before reserving space and sending metadata and slivers to storage nodes.

mlegner:
- When attempting to store a blob, check the blob status first and only store it if necessary.
- Add `force` option to disable this optimization.
- Make input and output in JSON mode "camelCase".
- Extend the output for storing blobs in the JSON and HTTP APIs.

Closes #383

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,7 +120,7 @@ cargo run --bin walrus -- json \
         "config": "working_dir/client_config.yaml",
         "command": {
             "read": {
-                "blob_id": "4BKcDC0Ih5RJ8R0tFMz3MZVNZV8b2goT6_JiEEwNHQo"
+                "blobId": "4BKcDC0Ih5RJ8R0tFMz3MZVNZV8b2goT6_JiEEwNHQo"
             }
         }
     }'
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -18,7 +18,6 @@ use clap::{Args, Parser, Subcommand};
 use colored::Colorize;
 use serde::{Deserialize, Serialize};
 use serde_with::{base64::Base64, serde_as, DisplayFromStr};
-use sui_types::base_types::ObjectID;
 use tracing_subscriber::{util::SubscriberInitExt, EnvFilter};
 use walrus_core::{
     encoding::{EncodingConfig, Primary},
@@ -29,6 +28,7 @@ use walrus_sdk::api::BlobStatus;
 use walrus_service::{
     cli_utils::{
         error,
+        format_event_id,
         get_contract_client,
         get_read_client,
         get_sui_read_client_from_rpc_node_or_wallet,
@@ -38,15 +38,15 @@ use walrus_service::{
         success,
         HumanReadableBytes,
     },
-    client::Client,
+    client::{BlobStoreResult, Client},
     daemon::ClientDaemon,
 };
-use walrus_sui::{client::ReadClient, types::Blob};
+use walrus_sui::client::ReadClient;
 
 #[derive(Parser, Debug, Clone, Deserialize)]
 #[command(author, version, about = "Walrus client", long_about = None)]
 #[clap(rename_all = "kebab-case")]
-#[serde(rename_all = "snake_case")]
+#[serde(rename_all = "camelCase")]
 struct App {
     /// The path to the wallet configuration file.
     ///
@@ -94,7 +94,7 @@ struct App {
 #[serde_as]
 #[derive(Subcommand, Debug, Clone, Deserialize)]
 #[clap(rename_all = "kebab-case")]
-#[serde(rename_all = "snake_case")]
+#[serde(rename_all = "camelCase", rename_all_fields = "camelCase")]
 enum Commands {
     /// Store a new blob into Walrus.
     #[clap(alias("write"))]
@@ -105,6 +105,13 @@ enum Commands {
         #[clap(short, long, default_value_t = default::epochs())]
         #[serde(default = "default::epochs")]
         epochs: u64,
+        /// Do not check for the blob status before storing it.
+        ///
+        /// This will create a new blob even if the blob is already certified for a sufficient
+        /// duration.
+        #[clap(long, action)]
+        #[serde(default)]
+        force: bool,
     },
     /// Read a blob from Walrus, given the blob ID.
     Read {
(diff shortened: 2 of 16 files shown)


README.md
```diff
@@ -120,7 +120,7 @@ cargo run --bin walrus -- json \
         "config": "working_dir/client_config.yaml",
         "command": {
             "read": {
-                "blob_id": "4BKcDC0Ih5RJ8R0tFMz3MZVNZV8b2goT6_JiEEwNHQo"
+                "blobId": "4BKcDC0Ih5RJ8R0tFMz3MZVNZV8b2goT6_JiEEwNHQo"
```
giac-mysten, line 123: Question: Why the switch to camel case?

crates/walrus-service/bin/client.rs
```diff
@@ -94,7 +94,7 @@ struct App {
 #[serde_as]
 #[derive(Subcommand, Debug, Clone, Deserialize)]
 #[clap(rename_all = "kebab-case")]
-#[serde(rename_all = "snake_case")]
+#[serde(rename_all = "camelCase", rename_all_fields = "camelCase")]
```
giac-mysten, line 97: Question: do I understand correctly that `rename_all_fields` is necessary for the names of the struct variants?

crates/walrus-service/bin/client.rs
```diff
@@ -105,6 +105,10 @@ enum Commands {
         #[clap(short, long, default_value_t = default::epochs())]
         #[serde(default = "default::epochs")]
         epochs: u64,
+        /// Create a new blob even if the blob is already certified or marked as invalid.
```
giac-mysten: Question: I'm a bit unsure about the `invalid` part here. What is the result of forcing an invalid blob ID? It becomes certified again, if the slivers are correct this time?
Suggestion: Clarify the expected behavior.

crates/walrus-service/src/client.rs
```diff
@@ -170,22 +174,73 @@ impl<T: ContractClient> Client<T> {
...
+                    status: BlobCertificationStatus::Invalid,
+                    status_event,
+                    ..
+                } => {
+                    tracing::debug!("blob is marked as invalid");
+                    return Ok(BlobStoreResult::MarkedInvalid {
+                        blob_id,
+                        event: status_event,
+                    });
+                }
```
giac-mysten, line 237: Remark: The `Registered` status is not considered here, but it could be used to skip the registration phase (if the registration is appropriate) and directly upload the slivers?

crates/walrus-sui/src/client.rs
```diff
@@ -318,7 +318,7 @@ impl ContractClient for SuiContractClient {
 
     async fn certify_blob(
         &self,
-        blob: &Blob,
+        blob: Blob,
```
giac-mysten, line 321: Question: Consume it to get rid of the old blob?

README.md
mlegner: Mainly to be consistent with the formats of our internal REST APIs. Also, camelCase seems to be a bit more common for JSON. See also #477 (cc @jpcsmith).

I don't have any objections to snake_case, but I think it would be good to be consistent.

giac-mysten: sounds good, just curious!

crates/walrus-service/bin/client.rs
mlegner: Yes, that's right. The `rename_all` only renames the variant names themselves (e.g., `Read` -> `read`) but not the fields of the variants (so `blob_id` would remain `blob_id`). This wasn't an issue with the `snake_case` variant, because the fields were already in snake_case.

crates/walrus-service/bin/client.rs
mlegner: I guess I can also change the docstring to "Do not check the blob status before storing the blob.".

crates/walrus-sui/src/client.rs
mlegner: This is mainly to prevent bugs where we certify the blob but then reuse the old and now outdated object. If one really wants the old object, one can clone it before calling this method.

crates/walrus-service/src/client.rs
mlegner: Unfortunately, that doesn't help as one would be unable to certify the blob afterwards. But I can mention this in a comment.

crates/walrus-service/bin/client.rs
giac-mysten: That works. Forward-looking question: what will be the status of "2f+1 nodes don't want to store this anymore" blobs? We may need to distinguish it from invalid.

mlegner: > I'm a bit unsure about the invalid part here. What is the result of forcing an invalid blob ID?

Actually, thinking about it, this should never actually happen in practice as our client correctly encodes the blob. So I can actually remove this part.

> Forward-looking question: what will be the status of "2f+1 nodes don't want to store this anymore" blobs? We may need to distinguish it from invalid.

Do you mean when they add it to their blocklist?

giac-mysten: >Actually, thinking about it, this should never actually happen in practice as our client correctly encodes the blob. So I can actually remove this part.

that was my initial thought, but that's not a good idea. Imagine an adversary that wants to prevent storing `README.md`. It can create the correct blob ID for it, and then upload an invalid blob for it. So forcing the store of a blob id that was deemed invalid is actually useful.

>Do you mean when they add it to their blocklist?
yep

mlegner: > It can create the correct blob ID for it, and then upload an invalid blob for it.

That wouldn't work: The consistency is actually a property of the blob ID not the data sent to storage nodes. If an adversary computes a correct blob ID, it can try storing invalid metadata or slivers. However, either the certification will already fail (invalid metadata, too many invalid slivers), or all nodes will eventually have correct slivers.

crates/walrus-service/src/client.rs
mlegner: Added a comment to the code.

crates/walrus-service/bin/client.rs
mlegner: Modified the docstring to
```rust
        /// Do not check for the blob status before storing it.
        ///
        /// This will create a new blob even if the blob is already certified for a sufficient
        /// duration.
```

giac-mysten: Ah you're right. thanks for the change!

PR #509, opened 2024-06-10
chore: remove system pkg from configs and constructors

### LINKED ISSUE #491: Remove contract object from config and obtain via system object
Currently, we configure both the contract object and the system object. However, the contract can be found easily based on the system object, so we can reduce the config to just the system object.

karlwuest:
closes #491

### DIFF
--- a/crates/walrus-service/src/client/config.rs
+++ b/crates/walrus-service/src/client/config.rs
@@ -12,7 +12,7 @@ use crate::config::LoadConfig;
 /// Config for the client.
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct Config {
-    /// The walrus system object id.
+    /// The Walrus system object ID.
     pub system_object: ObjectID,
     /// Path to the wallet configuration.
     #[serde(default)]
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/client/config.rs
```diff
@@ -12,9 +12,7 @@ use crate::config::LoadConfig;
 /// Config for the client.
 #[derive(Debug, Serialize, Deserialize, Clone)]
 pub struct Config {
-    /// The walrus package id.
-    pub system_pkg: ObjectID,
-    /// The system walrus system object id.
+    /// The walrus system object id.
```
mlegner: Nit (although pre-existing):
```suggestion
    /// The Walrus system object ID.
```

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -251,10 +245,7 @@ async fn check_system_deployment(
 
     let object_pkg_id = get_package_id_from_object_response(&response)
         .map_err(|_| SuiClientError::WalrusSystemObjectDoesNotExist(system_object_id))?;
-    if object_pkg_id != system_pkg_id {
-        return Err(SuiClientError::WalrusPackageDoesNotExist(system_pkg_id));
```
mlegner, line 255: Suggestion: Also remove this error variant.

PR #511, opened 2024-06-11
fix: Make reqwest not require openssl

gdanezis:
Specialise the options of the `reqwests` dependency to not use the default openssl on linux systems.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ futures = "0.3.30"
 move-core-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.26.0" }
 mysten-metrics = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.26.0" }
 rand = "0.8.5"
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -17,7 +17,7 @@ futures = "0.3.30"
 move-core-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.26.0" }
 mysten-metrics = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.26.0" }
 rand = "0.8.5"
-reqwest = "0.12.4"
+reqwest = { version = "0.12.4", default-features = false, features = ["http2", "blocking", "json", "rustls-tls"] }
```
mlegner: Question: How did you select the features to include? The [default features](https://github.com/seanmonstar/reqwest/blob/695bc0463726bb243e235f17c7f8833974835ec8/Cargo.toml#L30) include `charset` and `macos-system-configuration` but, for example, not `blocking`.

gdanezis: I did it at random, until the thing compiled :) LOL. Now selecting the minimum.

PR #513, opened 2024-06-11
chore(deps): minor cleanup of Cargo.toml files

mlegner:


crates/walrus-stress/Cargo.toml
```diff
@@ -10,12 +10,12 @@ anyhow.workspace = true
...
 futures.workspace = true
 mysten-metrics.workspace = true
-prometheus = "0.13.4"
+prometheus.workspace = true
 rand.workspace = true
 serde.workspace = true
 tokio.workspace = true
 tracing.workspace = true
-tracing-subscriber = { version = "0.3.18", features = ["env-filter"] }
+tracing-subscriber = { workspace = true, features = ["env-filter"] }
```
karlwuest: I think the `env-filter` feature is not actually used.
```suggestion
tracing-subscriber.workspace = true
```

PR #516, opened 2024-06-11
refactor(client)!: change the default config name for the client

giac-mysten:
Since we always share it as `client_config.yaml`, it makes sense to use that name as a default and avoid confusing users.

It's also more convenient b/c there is no need to rename it.


crates/walrus-service/bin/client.rs
mlegner, line 1: Suggestion: While we're at it, can we also remove the `./client.yaml` as a default for the Sui config? This is otherwise somewhat confusing.

PR #517, opened 2024-06-11
action to tag main commits when CI passes

quepcompute:
Tagging the commits on `main` when CI completes.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -1,0 +1,48 @@
+name: Tag
+
+on:
+  repository_dispatch:
+    types: [tag-walrus-main]
+
+  workflow_call:
+    inputs:
+      walrus_commit:
+        description: 'Walrus repo commit to tag'
+        type: string
+        required: true
+      tag_name:
+        description: 'Tag Name'
+        type: string
+        required: true
+  workflow_dispatch:
+    inputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/tag.yml
```diff
@@ -0,0 +1,48 @@
...
+  TAG_NAME: "${{ github.event.client_payload.tag_name || inputs.tag_name }}"
+
+jobs:
+  tag:
+    name: Tag
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11  # Pin v4.1.1
```
mlegner, line 39: Question: Do you recommend we generally pin actions?

quepcompute: My understanding is that its generally good practice to pin to a stable version, not subject to breaking changes of future releases.

quepcompute: If tagging every commit is unnecessary or premature I can remove it from the CI workflows and close this PR, since the repository dispatch GH action isn't defined in this repo, nothing has happened so far.

mlegner: I currently don't see the need for the workflow, but @johnjmartin and you are the experts on this topic. 🙂

PR #519, opened 2024-06-12
fix(deps): bump all Sui deps to testnet-v1.27.0

mlegner:
- Have to downgrade rocksdb again as this was downgraded in typed-store.
- Temporary workaround for removed `read_dynamic_field_value` method, see https://github.com/MystenLabs/sui/pull/18193.


crates/walrus-sui/src/utils.rs
```diff
@@ -451,7 +451,13 @@ macro_rules! match_for_correct_type {
...
     ($struct:expr, $field_name:expr, $field_type:path $({ $var:ident })*) => {
         match_for_correct_type!(
-            $struct.read_dynamic_field_value($field_name),
+            // TODO(mlegner): Change this back to $struct.field_value($field_name) when bumping Sui
+            // to a version that includes https://github.com/MystenLabs/sui/pull/18193.
+            match &$struct {
+                SuiMoveStruct::WithFields(fields) => fields.get($field_name).cloned(),
+                SuiMoveStruct::WithTypes { type_: _, fields } => fields.get($field_name).cloned(),
+                _ => None,
+            },
```
karlwuest, line 460: Remark: An alternative would be to convert the SuiMoveStruct to a json `Value` and then we could unify the macros for events and structs which might be cleaner anyway because we can reduce duplication a bit. Nothing to do for this PR, but I can look into this after devnet.

karlwuest: Created an issue for this: https://github.com/MystenLabs/walrus/issues/522

PR #528, opened 2024-06-13
feat: implement cancellation of blob syncs on invalid blob events

### LINKED ISSUE #366: Improve management blob sync worker lifecycle
Blob sync workers should be unique for each blob id and support cancellation

karlwuest:
closes #366

### DIFF
--- a/crates/walrus-service/src/node/blob_sync.rs
+++ b/crates/walrus-service/src/node/blob_sync.rs
@@ -116,38 +116,35 @@ impl BlobSyncHandler {
     ) -> Result<Option<(usize, EventID)>, TypedStoreError> {
         let node = &blob_synchronizer.node;
         let histogram_set = node.metrics.recover_blob_duration_seconds.clone();
-        let mut cancelled = false;
-        let label;
-        select! {
+        let (is_cancelled, label) = select! {
             _ = blob_synchronizer.cancel_token.cancelled() => {
                 tracing::info!("cancelled blob sync");
-                cancelled = true;
-                label = metrics::STATUS_CANCELLED;
+                (true, metrics::STATUS_CANCELLED)
             }
             sync_result = blob_synchronizer.run() => {
-                label = if sync_result.is_err() {
+                let sync_result = sync_result.and_then(|_| Ok(node.mark_event_completed(
+                    blob_synchronizer.event_sequence_number,
+                    &blob_synchronizer.event_id,
+                )?));
+                if sync_result.is_err() {
                     tracing::error!(?sync_result, "blob synchronizer failed");
-                    metrics::STATUS_FAILURE
+                    (false, metrics::STATUS_FAILURE)
                 } else {
-                    metrics::STATUS_SUCCESS
-                };
+                    (false, metrics::STATUS_SUCCESS)
+                }
             }
-        }
+        };
 
+        self.remove_sync_handle(&blob_synchronizer.blob_id).await;
         histogram_set
             .with_label_values(&[label])
             .observe(start.elapsed().as_secs_f64());
-        self.remove_sync_handle(&blob_synchronizer.blob_id).await;
-        if cancelled {
+        if is_cancelled {
             Ok(Some((
                 blob_synchronizer.event_sequence_number,
                 blob_synchronizer.event_id,
             )))
         } else {
-            node.mark_event_completed(
-                blob_synchronizer.event_sequence_number,
-                &blob_synchronizer.event_id,
-            )?;
             Ok(None)
         }
     }
@@ -161,7 +158,10 @@ impl BlobSyncHandler {
             .filter_map(|(_, sync)| sync.cancel())
             .collect();
 
-        try_join_all(join_handles).await?;
+        try_join_all(join_handles)
+            .await?
+            .into_iter()
+            .collect::<Result<Vec<_>, _>>()?;
         Ok(())
     }
 }


crates/walrus-service/src/node.rs
```diff
@@ -403,56 +406,55 @@ impl StorageNode {
...
         &self,
         event_sequence_number: usize,
         event: BlobCertified,
     ) -> anyhow::Result<()> {
         let start = tokio::time::Instant::now();
         let histogram_set = self.inner.metrics.recover_blob_duration_seconds.clone();
 
-        if self.inner.storage.is_stored_at_all_shards(&event.blob_id)? {
+        if self.inner.storage.is_stored_at_all_shards(&event.blob_id)?
+            || self
```
jpcsmith: Suggestion: add a method on the storage to check if a blob id is invalid.

crates/walrus-service/src/node.rs
```diff
@@ -403,56 +406,55 @@ impl StorageNode {
...
                 .with_label_values(&[metrics::STATUS_SKIPPED])
                 .observe(start.elapsed().as_secs_f64());
-
             return Ok(());
         }
 
-        // Slivers, and possibly metadata, are not stored.
-        // TODO(jsmith): Do not spawn if there is already a worker for this blob id (#366)
-        // TODO(jsmith): Handle cancellation. (#366)
+        // Slivers, and possible metadata, are not stored.
```
jpcsmith: Nit: should be "possibly"

crates/walrus-service/src/node.rs
```diff
@@ -403,56 +406,55 @@ impl StorageNode {
...
         event_sequence_number: usize,
         event: InvalidBlobId,
     ) -> anyhow::Result<()> {
+        if let Some((event_seq_num, event_id)) = self
+            .inner
+            .blob_sync_handler
+            .cancel_sync(&event.blob_id)
+            .await?
+        {
+            // Advance the event cursor with the event of the cancelled sync
```
jpcsmith: Suggestion: extend this description to clarify that since the blob is invalid, the event is complete without a sync.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -29,6 +34,79 @@ use crate::{
...
+    blob_syncs_in_progress: Arc<Mutex<HashMap<BlobId, InProgressSyncHandle>>>,
+}
+
+impl BlobSyncHandler {
+    pub async fn cancel_sync(&self, blob_id: &BlobId) -> anyhow::Result<Option<(usize, EventID)>> {
+        let mut guard = self.blob_syncs_in_progress.lock().await;
+        let Some(current_sync) = guard.remove(blob_id) else {
+            return Ok(None);
+        };
+        // drop the lock to prevent dead locks due to
```
jpcsmith: Nit: 
```suggestion
        // drop the lock to prevent deadlocks
```

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -29,6 +34,79 @@ use crate::{
...
+        event: BlobCertified,
+        event_sequence_number: usize,
+        node: Arc<StorageNodeInner>,
+    ) -> Result<(), TypedStoreError> {
+        let blob_id = event.blob_id;
+        let mut blob_syncs_guard = self.blob_syncs_in_progress.lock().await;
+        if let Entry::Vacant(entry) = blob_syncs_guard.entry(blob_id) {
+            let event_id = event.event_id;
+            let synchronizer = BlobSynchronizer::new(event, event_sequence_number, node);
+            let cancel_token = synchronizer.cancel_token.clone();
```
jpcsmith: Important: When the storage node shuts down in `node.rs's StorageNode::run`, we also need to cancel all in progress syncs.

Suggestion: Accept a cancel token from the storage node and create children tokens here to provide to the BlobSynchronizers.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -29,6 +34,79 @@ use crate::{
...
+            }
+            _ => None,
+        }
+    }
+
+    pub async fn start_sync(
+        &self,
+        event: BlobCertified,
+        event_sequence_number: usize,
+        node: Arc<StorageNodeInner>,
```
jpcsmith: Suggestion: Provide node as a struct parameter, this gives the feeling that a single synchronizer may manage tasks from different nodes.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -78,8 +160,28 @@ impl BlobSynchronizer {
...
+                tracing::debug!("cancelled blob sync");
+                return Ok(Some((self.event_sequence_number, self.event_id)));
+            }
+            _= self.run() => (),
+        }
+
+        self.node
+            .blob_sync_handler
+            .remove_sync_handle(&self.blob_id, self.event_id)
+            .await;
```
jpcsmith: Remark: This feels a bit unclean that the synchronizer itself manages its presence in the container class.
Suggestion: Move this entire function as is (`sync`) into `BlobSyncHandler`, and have it accept `synchronizer: BlobSynhronizer` as an argument instead.

crates/walrus-service/src/node.rs
```diff
@@ -403,56 +406,55 @@ impl StorageNode {
...
-                    tracing::error!(?err, "blob synchronizer failed");
-                    metrics::STATUS_FAILURE
-                } else {
-                    metrics::STATUS_SUCCESS
-                };
-
-                histogram_set
-                    .with_label_values(&[label])
-                    .observe(start.elapsed().as_secs_f64());
-            }
```
jpcsmith, line 443: Important: The handling of the events seem to be missing in blob_sync.

crates/walrus-service/src/node/blob_sync.rs
karlwuest: The problem with that is that the BlobSyncHandler is a field on the `StorageNodeInner`. I originally had this as a `Weak<StorageNodeInner>`, but then it needs to be set after creating it, which I didn't find ideal either. 
I'm happy to change it back to that, but maybe I'm overlooking a better solution. Do you have a suggestion?

karlwuest: I now just moved the `blob_sync_handler` from `StorageNodeInner` to `StorageNode` (see ea927d77d544a07624612d5687213c184df63a74).

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -91,6 +97,40 @@ impl BlobSyncHandler {
...
+        )?;
+
+        Ok(None)
+    }
+
+    pub async fn cancel_all(&self) -> anyhow::Result<()> {
+        let mut guard = self.blob_syncs_in_progress.lock().await;
+        let syncs: Vec<_> = guard.drain().collect();
+        // drop the lock to prevent deadlocks
+        drop(guard);
```
jpcsmith: Nit: Before there was a nice invariant on this struct that if handle is in `blob_syncs_in_progress`, then it is not yet complete, since it's only removed on completion. It would be nice if we could keep this invariant by instead iterating over the map and canceling the blob syncs, "cloning" the join handles to await them after you've released the guard (such as with [`FuturesExt::shared()`][shared] or `std::mem::take` on an `Option` wrapping the handle), and then awaiting all of the handles.

If you go this route, then you can also unify the branches in the `select!` above and only perform the `histogram_set` once.

[shared]: https://docs.rs/futures/latest/futures/future/trait.FutureExt.html#method.shared

jpcsmith: Hint:
```suggestion
        let syncs: Vec<_> = async {
            let mut guard = self.blob_syncs_in_progress.lock().await;
            guard.drain().collect()
        }.await;
```

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -91,6 +97,40 @@ impl BlobSyncHandler {
...
+        try_join_all(syncs.into_iter().map(|(_, sync)| sync.cancel())).await?;
```
jpcsmith: Important: The futures are returning `Result<Result<_, anyhow::Error>, JoinError>` errors, this propagates the first but discards the second.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -101,13 +102,32 @@ impl BlobSyncHandler {
     #[tracing::instrument(skip_all, fields(blob_id = %blob_synchronizer.blob_id), err)]
     pub async fn sync(
         blob_synchronizer: BlobSynchronizer,
+        start: tokio::time::Instant,
     ) -> anyhow::Result<Option<(usize, EventID)>> {
+        let histogram_set = blob_synchronizer
+            .node
+            .metrics
+            .recover_blob_duration_seconds
+            .clone();
```
jpcsmith:
```suggestion
        let node = &blob_synchronizer.node;
        let histogram_set = node
            .metrics
            .recover_blob_duration_seconds
            .clone();
```
And then use the reference to the node elsewhere in this function as well.

crates/walrus-service/src/node/blob_sync.rs
karlwuest: I'm not sure I follow. Can you clarify what you mean?

crates/walrus-service/src/node/blob_sync.rs
karlwuest: Made some changes in that direction in ca2cc931651b5f55d7774281c7a38177d9cc6e09. Do you want to have another quick look before merging?

`Shared` sadly doesn't seem to work for `JoinHandle` (the type can be instantiated, but it doesn't implement Future), because `JoinError` is not `Clone`. But I think it works quite nicely with the `Option` and makes a few other things cleaner.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -116,64 +113,71 @@ impl BlobSyncHandler {
...
-    ) -> anyhow::Result<Option<(usize, EventID)>> {
-        let histogram_set = blob_synchronizer
-            .node
-            .metrics
-            .recover_blob_duration_seconds
-            .clone();
+    ) -> Result<Option<(usize, EventID)>, TypedStoreError> {
+        let node = &blob_synchronizer.node;
+        let histogram_set = node.metrics.recover_blob_duration_seconds.clone();
+        let mut cancelled = false;
```
jpcsmith: Nit: Rename `is_cancelled`

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -116,64 +113,71 @@ impl BlobSyncHandler {
...
                     tracing::error!(?sync_result, "blob synchronizer failed");
                     metrics::STATUS_FAILURE
                 } else {
                     metrics::STATUS_SUCCESS
                 };
-                histogram_set
-                    .with_label_values(&[label])
-                    .observe(start.elapsed().as_secs_f64());
             }
         }
```
jpcsmith: Suggestion: return `label` and `is_cancelled` from the select:
```rust
let (is_cancelled, label) = select! {
    ...
};
```

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -116,64 +113,71 @@ impl BlobSyncHandler {
...
+        if cancelled {
+            Ok(Some((
+                blob_synchronizer.event_sequence_number,
+                blob_synchronizer.event_id,
+            )))
+        } else {
+            node.mark_event_completed(
+                blob_synchronizer.event_sequence_number,
+                &blob_synchronizer.event_id,
+            )?;
```
jpcsmith: Question: Could this be moved to the block of ` sync_result = blob_synchronizer.run() => {`?

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -116,64 +113,71 @@ impl BlobSyncHandler {
...
-        blob_synchronizer.node.mark_event_completed(
-            blob_synchronizer.event_sequence_number,
-            &blob_synchronizer.event_id,
-        )?;
-
-        Ok(None)
+        histogram_set
+            .with_label_values(&[label])
+            .observe(start.elapsed().as_secs_f64());
+        self.remove_sync_handle(&blob_synchronizer.blob_id).await;
```
jpcsmith: Suggestion: Swap the order of these two statements.

crates/walrus-service/src/node/blob_sync.rs
jpcsmith: Question: What's the return type of the expression `try_join_all(syncs.into_iter().map(|(_, sync)| sync.cancel())).await;`? It seems to me that it's `Result<Vec<Result<_, anyhow::Error>>, JoinError>` which means that only first error is returned using `?`, and any errors in the vec are discarded.

karlwuest: Ok, now I get it. I misunderstood it as `try_join_all` returning a `Result<Result<_, _>, _>` and didn't realize that I'm getting a Vec of errors here. Thanks for the clarification!
I'll handle the errors contained in the Vector.

PR #529, opened 2024-06-13
feat(client): add default bind address

mlegner:
Also clean up other arguments and make sure the behavior of CLI and JSON modes is equivalent.


crates/walrus-service/bin/client.rs
```diff
@@ -312,6 +329,12 @@ mod default {
     pub(crate) fn status_timeout() -> Duration {
         Duration::from_secs(1)
     }
+
+    pub(crate) fn bind_address() -> SocketAddr {
+        "127.0.0.1:31415"
```
giac-mysten, line 334: crowded port? 😂

PR #530, opened 2024-06-13
feat: Add crash recovery testing

sadhansood:
The script works in configurable number of runs. In each run a cluster is brought up (by copying data dir of previous run), some storage nodes are crashed and a file is stored. In the end, we attempt to read all blobs stored and compare.
Run it as:
```
 ./scripts/crash_recovery.sh  
```
from the walrus directory

### DIFF
--- a/scripts/crash_recovery.sh
+++ b/scripts/crash_recovery.sh
@@ -1,0 +1,243 @@
+# Copyright (c) Mysten Labs, Inc.
+# SPDX-License-Identifier: Apache-2.0
+#!/bin/bash
+set -euo pipefail
+
+RUN=0
+RUNS=2
+COMMITTEE_SIZE=4
+SHARDS=20
+NUM_SERVERS_TO_CRASH=1
+NETWORK=devnet
+TIMESTAMP=$(date +"%Y-%m-%d_%H-%M-%S")
+WORKING_DIR=./working_dir/crash_recovery_testing
+RUN_DIR=$WORKING_DIR/$TIMESTAMP
+SUI_CLI="sui client"
+WALRUS_CLI="cargo run --bin walrus"
+WALRUS_NODE="cargo run --bin walrus-node"
+BLOBS_PER_RUN=5
+BLOB_IDS=()
+
+function log(){
+    echo "[$(date +'%T')] $1"
+}
+
+# Generate a random file of size $1 MB
+function generate() {
+    local size=$1
+    local filename=$2
+
+    if [[ -z "$size" || -z "$filename" ]]; then
+        log "cannot generate a file for unspecified size/filename"
+        return 1
+    fi
+    log "Generating file of size $size MB..."
+    dd if=/dev/urandom of="$filename" bs=1M count="$size" iflag=fullblock > "${RUN_DIR}"/"${RUN}"/log/err 2>&1
+}
+
+# Run a node
+function run_node() {
+    cmd="${WALRUS_NODE} run --config-path $1"
+    log "$cmd"
+    $cmd > "${RUN_DIR}"/"${RUN}"/log/err 2>&1 &
+    PID=$!
+    sleep 5
+    log "ensuring that storage node with pid ${PID} is running..."
+    ps -p ${PID} > "${RUN_DIR}"/"${RUN}"/log/err
+}
+
+# Start storage nodes
+function start_storage_nodes() {
+    for i in $(seq 0 $((COMMITTEE_SIZE - 1))); do
+        sed 's|storage_path.*|storage_path: \
+        '"${RUN_DIR}/${RUN}/data/storage-node-${i}"'|' \
+        "${RUN_DIR}/dryrun-node-${i}.yaml" > "${RUN_DIR}/${RUN}/storage-node-${i}.yaml"
+        run_node "${RUN_DIR}/${RUN}/storage-node-${i}.yaml"
+    done
+    log "servers are up and running"
+}
+
+# Prepare client, get gas from faucet, etc
+function prepare_client() {
+    cmd="${SUI_CLI} --client.config ${RUN_DIR}/sui_client.yaml faucet \
+    --url https://faucet.devnet.sui.io/v1/gas"
+    $cmd > "${RUN_DIR}"/"${RUN}"/log/err 2>&1
+    log "waiting for coin"
+    sleep 10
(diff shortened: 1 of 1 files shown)


scripts/crash_recovery.sh
```diff
@@ -0,0 +1,240 @@
+# Copyright (c) Mysten Labs, Inc.
+# SPDX-License-Identifier: Apache-2.0
+#!/bin/bash
+
```
jpcsmith: Suggestion: Set bash ["strict mode" ](https://gist.github.com/robin-a-meade/58d60124b88b60816e8349d1e3938615) to make it clear when something breaks while running the script, or partially set it (for example if you want to handle every error manually).
```suggestion
set -euo pipefail
```

scripts/crash_recovery.sh
```diff
@@ -0,0 +1,240 @@
...
+    store_file "$RUN_DIR/${RUN}/data/blob"
+}
+
+function read_files() {
+    # Read all the files back and compare
+    log "Reading all files back and comparing..."
+    run=1
+    for blob_id in "${BLOB_IDS[@]}"; do
+        log "Getting blob status for $blob_id ..."
+        cmd="./target/debug/walrus --config $RUN_DIR/client_config.yaml \
```
jpcsmith: Suggestion: Use cargo to run the binary. Not everyone may be using the default cargo directory structure.
Suggestion: Define a and use a global WALRUS_CLI="cargo run --bin walrus ..." with the common prefix of these commands.

scripts/crash_recovery.sh
```diff
@@ -0,0 +1,240 @@
...
+            echo "Error running node $i"
+            exit 1
+        fi
+    done
+    log "servers are up and running"
+}
+
+# Prepare client, get gas from faucet, etc
+function prepare_client() {
+    cmd="sui client --client.config ${RUN_DIR}/sui_client.yaml faucet \
```
jpcsmith: Suggestion: Define a gloabl SUI_CLI variable with the common prefix for running the sui command.

scripts/crash_recovery.sh
```diff
@@ -0,0 +1,240 @@
...
+    size=${#array[*]}
+    max=$(( 32768 / size * size ))
+
+    for ((i=size-1; i>0; i--)); do
+        while (( (rand=$RANDOM) >= max )); do :; done
+        rand=$(( rand % (i+1) ))
+        tmp=${array[i]} array[i]=${array[rand]} array[rand]=$tmp
+    done
+    echo "${array[@]}"
+}
```
jpcsmith: Remark: There is also `sort -R` which can be used to randomly sort.

scripts/crash_recovery.sh
```diff
@@ -0,0 +1,280 @@
...
+    # Start actual crash recovery testing
+    log "Starting crash recovery testing iterations..."
+    for i in $(seq 1 ${RUNS}); do
+        RUN=$i
+        log "====== RUN #${RUN} ======"
+        crash_and_store
+        if [ $? -ne 0 ]; then
+            echo "Error running crash recovery iteration $i"
+            exit 1
+        fi
```
jpcsmith: Suggestion: Remove such lines, as with errexit set (i.e., `set -e`) the script will exit on any non zero error code, such checks are now no-ops. If you want to print something on a failure, then enable `set -o errtrace` and `trap` the errors (see, [here](https://stackoverflow.com/questions/25378845/what-does-set-o-errtrace-do-in-a-shell-script) for an example).

Given that you're logging when you do any operation, a trap message that says "TEST FAILED!" will likely be sufficient to identify where it failed.

scripts/crash_recovery.sh
```diff
@@ -0,0 +1,280 @@
...
+
+# Generate a random file of size $1 MB
+function generate() {
+    local size=$1
+    local filename=$2
+
+    if [[ -z "$size" || -z "$filename" ]]; then
+        return 1
+    fi
+    dd if=/dev/urandom of="$filename" bs=1M count="$size" iflag=fullblock
```
jpcsmith:
```suggestion
    if [[ -z "$size" || -z "$filename" ]]; then
        log "cannot generate a file for unspecified size/filename"
        exit 1
    fi
    dd if=/dev/urandom of="$filename" bs=1M count="$size" iflag=fullblock
```

scripts/crash_recovery.sh
```diff
@@ -0,0 +1,280 @@
...
+function run_node() {
+    cmd="${WALRUS_NODE} run --config-path $1"
+    log "$cmd"
+    $cmd > "${RUN_DIR}"/"${RUN}"/log/err 2>&1 &
+    PID=$!
+    sleep 5
+    if [ "$(ps -ef | grep ${PID} | grep -v grep)" == "" ]; then
+        log "server failed to start, pid ${PID} not found"
+        return 1
+    fi
```
jpcsmith: Suggestion: The following, if it works, may be simpler.

```suggestion
    log "ensuring that storage node ${PID} is running..."
    ps -p ${PID}
```

scripts/crash_recovery.sh
```diff
@@ -0,0 +1,280 @@
...
+function start_storage_nodes() {
+    for i in $(seq 0 $((COMMITTEE_SIZE - 1))); do
+        sed 's|storage_path.*|storage_path: \
+        '"${RUN_DIR}/${RUN}/data/storage-node-${i}"'|' \
+        "${RUN_DIR}/dryrun-node-${i}.yaml" > "${RUN_DIR}/${RUN}/storage-node-${i}.yaml"
+        run_node "${RUN_DIR}/${RUN}/storage-node-${i}.yaml"
+        if [ $? -ne 0 ]; then
+            echo "Error running node $i"
+            exit 1
+        fi
```
jpcsmith: Remark: `run_node` will already exit on failure.
```suggestion
```

scripts/crash_recovery.sh
```diff
@@ -0,0 +1,280 @@
...
+}
+
+function store_files() {
+    ips=" "
+    for i in $(seq 1 $COMMITTEE_SIZE); do
+        ips+="127.0.0.1 "
+    done
+
+    # Deploy system contract
+    echo Deploying system contract...
```
jpcsmith:
```suggestion
    log "Deploying system contract..."
```

scripts/crash_recovery.sh
sadhansood: This worked, thanks!

PR #533, opened 2024-06-14
feat: add reads to stress client

### LINKED ISSUE #506: Read transactions for load generator
Our current load generator only supports write transactions, we could update it to submit a mix of read and write requests

### LINKED ISSUE #527: Add read load to stress client

karlwuest:
- closes #506
- closes #527 
- additionally continues running the stress client when a write fails

### DIFF
--- a/crates/walrus-orchestrator/src/benchmark.rs
+++ b/crates/walrus-orchestrator/src/benchmark.rs
@@ -40,7 +40,7 @@ impl<N, C> Display for BenchmarkParametersGeneric<N, C> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         write!(
             f,
-            "{} nodes ({}) - {} tx/s",
+            "{} nodes ({}) - {} ops/min",
             self.nodes, self.settings.faults, self.load
         )
     }
--- a/crates/walrus-orchestrator/src/main.rs
+++ b/crates/walrus-orchestrator/src/main.rs
@@ -74,8 +74,8 @@ pub enum Operation {
         #[clap(long, value_name = "INT", default_value_t = 4, global = true)]
         committee: usize,
 
-        /// The set of loads to submit to the system (tx/s). Each load triggers a separate
-        /// benchmark run. Setting a load to zero will not deploy any benchmark clients
+        /// The set of loads to submit to the system (reads/writes per minunte). Each load triggers
+        /// a separate benchmark run. Setting a load to zero will not deploy any benchmark clients
         /// (useful to boot testbeds designed to run with external clients and load generators).
         #[clap(long, value_name = "[INT]", default_value = "200", global = true)]
         loads: Vec<usize>,
--- a/crates/walrus-orchestrator/src/measurements.rs
+++ b/crates/walrus-orchestrator/src/measurements.rs
@@ -253,7 +253,7 @@ impl MeasurementsCollection {
         table.add_row(row![bH2->""]);
         table.add_row(row![b->"Nodes:", self.parameters.nodes]);
         table.add_row(row![b->"Faults:", self.parameters.settings.faults]);
-        table.add_row(row![b->"Load:", format!("{} tx/s", self.parameters.load)]);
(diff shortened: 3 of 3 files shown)


crates/walrus-orchestrator/src/protocol/target.rs
```diff
@@ -264,7 +264,7 @@ impl ProtocolCommands for TargetProtocol {
     {
         let clients: Vec<_> = instances.into_iter().collect();
         let load_per_client = (parameters.load / clients.len()).max(1);
-        let number_of_tasks = load_per_client * 5;
+        let number_of_tasks = (load_per_client / 10).max(1);
```
asonnino, line 269: should we add a comment behind this formula?

crates/walrus-stress/src/main.rs
```diff
@@ -23,14 +23,18 @@ mod metrics;
...
+    /// The target write load to submit to the system (writes/minute).
     /// The actual load may be limited by the number of clients.
-    #[clap(long, default_value = "2")]
-    target_load: NonZeroU64,
+    #[clap(long, default_value = "60")]
+    write_load: NonZeroU64,
+    /// The target read load to submit to the system (reads/minute).
+    /// The actual load may be limited by the number of clients.
+    #[clap(long, default_value = "60")]
+    read_load: u64,
```
asonnino: The way the orchestrator is built is by expecting the load generator to define a percentage of write load (eg, number from 0 to 100). It would then use this percentage to determine how many write and read transactions ti should send. This is similar to how the Sui load generator work. I have nothing against having two separate `write_load` and `read_load` but then we need to (1) adapt the settings of the orchestrator to take this into account, and (2) perhaps give a bit of an intuition about what it means to have `write_load` set to 60% and `read_load` set to 80%.

asonnino:
```suggestion
    /// The target read load to submit to the system (reads/minute).
    /// The actual load may be limited by the number of clients.
    #[clap(long, default_value = "60")]
    read_load: NonZeroU64,
```

crates/walrus-stress/src/generator.rs
```diff
@@ -23,106 +25,183 @@ use walrus_test_utils::WithTempDir;
...
 
-                    for _ in 0..tx_per_burst {
-                        let Some((client, mut blob)) = self.client_pool.pop_front() else {
+                    for _ in 0..writes_per_burst {
+                        let Some((client, mut blob)) = self.write_client_pool.pop_front() else {
                             tracing::warn!("No client available to submit write");
                             break;
                         };
                         write_finished.push(tokio::spawn(async move {
                             blob.refresh();
```
asonnino, line 132: ok so this is what ensures that all write requests we submit have a different blob. (I missed that for a moment)

crates/walrus-stress/src/main.rs
karlwuest: >The way the orchestrator is built is by expecting the load generator to define a percentage of write load (eg, number from 0 to 100). It would then use this percentage to determine how many write and read transactions ti should send. This is similar to how the Sui load generator work. I have nothing against having two separate write_load and read_load but then we need to (1) adapt the settings of the orchestrator to take this into account, and (2) perhaps give a bit of an intuition about what it means to have write_load set to 60% and read_load set to 80%.

I think for manually running the stress client it is much more understandable if the load is specified through the actual load values instead of percentages, because it's not clear what 100% would be. Could you adapt the orchestrator as a follow up?

Edit: I'm not sure what to adapt in the orchestrator. I currently don't see where it is set up with percentage loads. IIUC, it was previously just set up for load per second. 
However, I just noticed that I didn't change tx/s to ops/min in some spots, so I corrected that.

crates/walrus-stress/src/generator.rs
karlwuest: I'll add a comment

crates/walrus-stress/src/main.rs
asonnino: yes let's do that, I can change the orchestrator later

PR #539, opened 2024-06-17
feat(client): add blocklist feature for daemon

### LINKED ISSUE #505: Add blocklist to the aggregator

mlegner:
The blocklist is simply a list of blob IDs in YAML syntax and can be set with an optional argument. This is currently only read once during startup. Continuous updates could be added in a follow-up PR.

Both aggregator and publisher return a 403 (Forbidden) message for blocked IDs.

The functionality is added to the `Client`, so it would also work for the CLI `store` and `read`. However, I don't really see a need for this and thus haven't implemented it.

Depends on #529 
Closes #505

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -40,6 +40,8 @@ use walrus_sui::{
     types::{Blob, BlobEvent, Committee, StorageNode},
 };
 
+mod blocklist;
+pub use blocklist::Blocklist;
 mod communication;
 mod config;
 pub use config::{default_configuration_paths, ClientCommunicationConfig, Config};
@@ -73,6 +75,7 @@ pub struct Client<T> {
     max_concurrent_status_reads: usize,
     encoding_config: EncodingConfig,
     global_write_limit: Arc<Semaphore>,
+    blocklist: Option<Blocklist>,
 }
 
 impl Client<()> {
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/client.rs
```diff
@@ -819,6 +804,49 @@ impl<T> Client<T> {
...
+    /// Returns the blob ID.
+    pub fn blob_id(&self) -> &BlobId {
+        match self {
+            Self::AlreadyCertified { blob_id, .. } => blob_id,
+            Self::MarkedInvalid { blob_id, .. } => blob_id,
+            Self::NewlyCreated(Blob { blob_id, .. }) => blob_id,
+        }
+    }
+}
+
```
mlegner, line 849: Remark: This part is just moved because it was between the different `impl Client` blocks.

crates/walrus-service/src/client/blocklist.rs
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    /// Removes a blob ID from the blocklist.
+    ///
+    /// Returns whether the ID was previously blocked.
+    #[inline]
+    pub fn remove(&mut self, blob_id: &BlobId) -> bool {
+        self.blocked_blobs.remove(blob_id)
+    }
+}
```
giac-mysten, line 68: Question: where are these used?

mlegner: Currently, they aren't. 😅 

Should I remove them?

giac-mysten: No, it's ok to leave them I guess.

PR #540, opened 2024-06-17
fix: limit the number of parallel blob syncs

jpcsmith:
This adds configuration to the SuiNodeCommitteeService, as well as the BlobSyncHandler to allow modifying the number of syncs in parallel as well as timeouts.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -44,7 +44,7 @@ use crate::{
     committee::{CommitteeService, CommitteeServiceFactory, SuiCommitteeServiceFactory},
     config::{StorageNodeConfig, SuiConfig},
     contract_service::{SuiSystemContractService, SystemContractService},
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/config.rs
```diff
@@ -58,10 +58,66 @@ pub struct StorageNodeConfig {
...
+    pub max_concurrent_blob_syncs: usize,
+    /// Configuration of the committee service timeouts and retries
+    #[serde(flatten)]
+    pub committee_service_config: CommitteeServiceConfig,
+}
+
+impl Default for BlobRecoveryConfig {
+    fn default() -> Self {
+        Self {
+            max_concurrent_blob_syncs: 10_000,
```
karlwuest, line 83: Nit: 10k seems high as a default, if we suspect that we had problems due to 100.

jpcsmith: The goal is to have this be the same as it was before the PR, and then modify it from the config. Hence the use of an arbitrarily large value.

PR #547, opened 2024-06-18
fix(walrus-service)!: update defaults for blob recovery and change format

jpcsmith:


crates/walrus-service/src/config.rs
```diff
@@ -128,7 +139,10 @@ pub struct SuiConfig {
     pub system_object: ObjectID,
     /// Interval with which events are polled, in milliseconds.
     #[serde_as(as = "serde_with::DurationMilliSeconds")]
-    #[serde(default = "defaults::polling_interval")]
+    #[serde(
+        rename = "event_polling_interval_ms",
```
karlwuest: Nit: just to keep aligned with the usual naming in `Duration`
```suggestion
        rename = "event_polling_interval_millis",
```

Issue #356, opened 2024-05-06
Improve and extend integration/E2E tests with the simtest framework


mlegner:
https://github.com/MystenLabs/mysten-sim allows us to create more deterministic and thorough end-to-end tests. Integrating this requires the following tasks:
- [ ] Add necessary plumbing to our `TestCluster`.
- [ ] Define relevant failpoints throughout our code.
- [ ] Define interesting test cases and failure scenarios.
- [ ] Implement individual tests.

PR #550, opened 2024-06-18
test: deterministic testing for Walrus

halfprice:
This PR introduces a deterministic testing framework, known as `simtest`, to Walrus, enhancing its capability to identify bugs in distributed systems. The core concept is to run Tokio actors within an msim-based runtime that deterministically schedules actor execution.

Key additions and functionalities of this PR include:

- Integrating the basic dependency to MSIM.
- Creating a simple test that performs the following:
  - Sets up a Sui cluster.
  - Establishes a Walrus cluster using the Sui cluster.
  - Writes a random blob.
  - Reads the blob back.
  - Verifies deterministic execution by setting `check_determinism` flag in the test. It runs the test twice and comparing RNG values throughout to ensure consistency in random number generation.
 
The next phase involves adding crash recovery tests using simtest. To achieve this, each Walrus node needs to be initiated within a dedicated simulator node, allowing for easy shutdown and recovery testing.

Related to #356

### DIFF
--- a/crates/walrus-service/src/lib.rs
+++ b/crates/walrus-service/src/lib.rs
@@ -15,7 +15,6 @@ pub mod testbed;
 pub mod utils;
 
 mod node;
-/// A module for creating a test Walrus cluster in process.
 #[cfg(any(test, feature = "test-utils"))]
 pub mod test_cluster;
 pub use node::{StorageNode, StorageNodeBuilder};
--- a/crates/walrus-service/src/test_cluster.rs
+++ b/crates/walrus-service/src/test_cluster.rs
@@ -1,6 +1,8 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+//! A module for creating a test Walrus cluster in process.
+
 use std::{
     sync::{Arc, OnceLock},
     time::Duration,


Cargo.toml
```diff
@@ -14,36 +22,48 @@ bcs = "0.1.6"
...
 serde_yaml = "0.9"
-sui-keys = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
-sui-move-build = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
-sui-sdk = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
-sui-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
-sui-config = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
-telemetry-subscribers = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
+# TODO(zhewu): Right now, the pinned sui commit is testnet-v1.27.2 + latest MSIM. This is to allow simtest to work
+# before 1.28. We can switch back to official release after 1.28 since that will contain the latest MSIM change.
+sui-config = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
```
halfprice, line 41: I spent quite sometime trying to make cargo patch work with latest sui branch including a msim update. However, due to multiple sui version are used, the patch can only be applied to one version. Therefore, I created a custom commit in sui, which is just the testnet-v1.27.2 + latest MISM change. We can revert to use tags when we move forward to testnet 1.28.0

Cargo.toml
```diff
@@ -14,36 +22,48 @@ bcs = "0.1.6"
...
+sui-sdk = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
+sui-simulator = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
+sui-types = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
+telemetry-subscribers = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
 tempfile = "3.10.1"
-test-cluster = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
+test-cluster = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
 thiserror = "1.0.61"
-tokio = "1.38.0"
+tokio = "=1.36.0"
```
halfprice, line 54: Since MSIM uses tokio 1.36, we have to use 1.36 here. We should update tokio version at MSIM, Sui, and Walrus all together.

mlegner: Makes sense! Can we then add these to [Dependabot's ignore list](https://github.com/MystenLabs/walrus/blob/main/.github/dependabot.yml#L20)?

Cargo.toml
```diff
@@ -14,36 +22,48 @@ bcs = "0.1.6"
 clap = { version = "4.5.7", features = ["derive"] }
 fastcrypto = "0.1"
 futures = "0.3.30"
+futures-timer = "=3.0.2" # required for MSIM
 mime = "0.3.17"
-move-core-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
-mysten-metrics = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.27.2" }
+move-core-types = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
+msim = { git = "https://github.com/MystenLabs/mysten-sim.git", rev = "862e52287eea423d132d462879631dad8e346147", package = "msim" }
```
mlegner: Question: Is the `package = "msim"` actually necessary?

Cargo.toml
mlegner: Does this include `futures-timer`?

Cargo.toml
mlegner: Question: Do all these dependencies have to be full dependencies? Or could they be "optional" or dev dependencies?

crates/walrus-simtest/README.md
```diff
@@ -0,0 +1,11 @@
...
+
+## Install simtest
+
+https://github.com/MystenLabs/sui/tree/main/crates/sui-simulator#how-to-run-sim-tests
+
+## Run simtest
+
+    $ cargo simtest <test_filter>
+
+TODO(zhewu): describe how to use simtest in detail.
```
mlegner: Remark: Open TODO: Either resolve it or create a follow-up issue.

sui_programmability/examples/basics/README.md
mlegner, line 1: Question: What is the purpose of these contracts?

crates/walrus-simtest/README.md
```diff
@@ -0,0 +1,11 @@
+# Walrus Simtest
+
+## Install simtest
+
+https://github.com/MystenLabs/sui/tree/main/crates/sui-simulator#how-to-run-sim-tests
```
mlegner, line 5: This points to a `./scripts/simtest/install.sh` file, which is not included here. Is there anything special about the `cargo-simtest` script here or could we take it from the `MystenLabs/sui` repo?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -0,0 +1,53 @@
...
+    else {
+        panic!("expect newly stored blob")
+    };
+
+    // Read the blob using primary slivers.
+    let read_blob = client
+        .as_ref()
+        .read_blob::<Primary>(&blob_confirmation.blob_id)
+        .await
+        .expect("Read blob failed");
```
mlegner: Nit: Can we reformulate this as an expectation (also below)?
```suggestion
        .expect("should be able to read blob we just stored");
```

scripts/simtest/cargo-simtest
```diff
@@ -0,0 +1,108 @@
...
+    --config "patch.crates-io.tokio.path = \"$LOCAL_MSIM_PATH/msim-tokio\""
+    --config "patch.'https://github.com/MystenLabs/mysten-sim'.msim.path = \"$LOCAL_MSIM_PATH/msim\""
+    --config "patch.crates-io.futures-timer.path = \"$LOCAL_MSIM_PATH/mocked-crates/futures-timer\""
+  )
+else
+  cargo_patch_args=(
+    --config 'patch.crates-io.tokio.git = "https://github.com/MystenLabs/mysten-sim.git"'
+    --config 'patch.crates-io.tokio.rev = "862e52287eea423d132d462879631dad8e346147"'
+    --config 'patch.crates-io.futures-timer.git = "https://github.com/MystenLabs/mysten-sim.git"'
+    --config 'patch.crates-io.futures-timer.rev = "862e52287eea423d132d462879631dad8e346147"'
```
mlegner, line 59: AFAICT, these are the only differences to the script in the MystenLabs/sui repo. What is the reason for this? Could we not just take the script from the sui repo?

crates/walrus-service/src/lib.rs
```diff
@@ -15,6 +15,8 @@ pub mod testbed;
 pub(crate) mod utils;
 
 mod node;
+/// A module for creating a test Walrus cluster in process.
+pub mod test_cluster;
```
mlegner: Suggestion: Annotate this with `#[cfg(any(test, feature = "test-utils"))]` or move to the `test_utils` module.

Cargo.toml
halfprice: yes, `futures-timer` is added for the same reason.

crates/walrus-service/src/lib.rs
halfprice: Done

crates/walrus-simtest/README.md
halfprice: After running the ./scripts/simtest/install.sh, it installs an `cargo simtest` command that calls the local cargo-simtest script. That's why we need to copy the simtest script here.

crates/walrus-simtest/tests/simtest.rs
halfprice: Done.

scripts/simtest/cargo-simtest
halfprice: Please see above reply. The `cargo simtest` command installation search for the local cargo-simtest script.

sui_programmability/examples/basics/README.md
halfprice: In sui, there are a few static initialization, one of them needs to use an example contract in this specific location:
https://github.com/MystenLabs/sui/blob/681b0ad45f6ee4eb3a36c90a11fc961835bd2cf6/crates/sui-proc-macros/src/lib.rs#L61

It's a bit nuance so I hope people don't need to worry too much about it.

Cargo.toml
halfprice: In fact, both are implicitly dependend since we are depending on sui-simulator. Removed both dependency entry.

halfprice: These are removed. Can't define dev-dependency in workspace toml file, but in each package, they are `[target.'cfg(msim)'.dependencies]`, so it won't be used to compile if not running in simulation.

crates/walrus-simtest/README.md
halfprice: Done

Cargo.toml
mlegner: Given the delay in merging this PR (my fault, sorry about that), we now have `testnet-v1.28.1` available. Do you want to change these tags in this PR or do you prefer a separate PR?

Cargo.toml
mlegner: It's a bit unfortunate that we have to use an old version of Tokio just because of our testing setup. Is there any way around this?

crates/walrus-simtest/README.md
mlegner: I see. Can we include the install script here as well? Otherwise, it's a bit complicated to install `cargo-simtest`. (This can also be done as part of #568.)

sui_programmability/examples/basics/README.md
mlegner: Hmm... That's unfortunate that we need an additional top-level directory just for this. I guess there is no way around this?

crates/walrus-service/src/lib.rs
```diff
@@ -16,6 +16,7 @@ pub(crate) mod utils;
 
 mod node;
 /// A module for creating a test Walrus cluster in process.
```
mlegner: Nit: Normally we document modules inside them with a `//!` docstring.

Cargo.toml
halfprice: Probably a separate PR. There was an issue with the MSIM framework, and I need to make sure that 1.28.1 contains that fix. I'll put it on my list to update Walrus to 1.28.1

Cargo.toml
halfprice: We can update MSIM to use 1.38. But this needs to be done in all MSIM, Sui, and Walrus. I can try to update it after this PR.

sui_programmability/examples/basics/README.md
halfprice: We can make `sim_test` macro to take a sample sui contract path. Again, this can be done in a separate PR.

crates/walrus-simtest/README.md
halfprice: Another option is to move the installation repo to MSIM repo. I updated #568 for now

crates/walrus-service/src/lib.rs
halfprice: Done

PR #555, opened 2024-06-19
feat(telemetry): add more instrumentation and distributed tracing

### LINKED ISSUE #463: Add HTTP server traces
Add HTTP server traces conforming to opentelemetry spec, including attaching error information to the trace

jpcsmith:
Add additional trace instrumentation and distributed tracing.

This instrumentation applies several semantic conventions based on the [OpenTelemetry spec](https://opentelemetry.io/docs/specs/semconv/general/trace/) to traces in `walrus-service` and `walrus-sdk`; and cleans up the trace instrumentation in the storage node. Once complete and merged in, this will allow traces started at a client or a storage node to be linked with those at another storage node. 

For example, the attached image which shows the trace for the synchronisation of a blob.

TODO:
- ~[ ] Add [retry counts](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-request-retries-and-redirects) to HTTP requests.~ (deferred to #573)
- [x] Refactor and complete the creation of traces in `walrus-sdk`. (#566)
- ~[ ] Ensure that traces are properly exported for walrus-service.~ (deferred to #574)
- [x] Report trace IDs in errors from the server.
- [x] Cleanup TODOs

Screenshot 2024-06-19 at 19 05 46

Closes  #463, depends on #566

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,10 +27,10 @@ futures-timer = "=3.0.2" # required for MSIM
 mime = "0.3.17"
 move-core-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.29.2" }
 mysten-metrics = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.29.2" }
-opentelemetry = "^0.23.0"
-opentelemetry-http = "^0.12.0"
-opentelemetry-otlp = "^0.16"
-opentelemetry_sdk = "^0.23.0"
+opentelemetry = "=0.23.0"
+opentelemetry-http = "=0.12.0"
+opentelemetry-otlp = "=0.16"
+opentelemetry_sdk = "=0.23.0"
 prometheus = "0.13.4"
 rand = "0.8.5"
 reqwest = { version = "0.12.5", default-features = false, features = ["http2", "json", "rustls-tls"] }


Cargo.toml
```diff
@@ -27,6 +27,7 @@ mime = "0.3.17"
 move-core-types = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
 mysten-metrics = { git = "https://github.com/MystenLabs/sui", rev = "6c3a6ba165be8483fec3026394477a01bf48acda" }
 opentelemetry = "^0.23.0"
+opentelemetry-http = "^0.12.0"
 opentelemetry-otlp = "^0.16"
 opentelemetry_sdk = "^0.23.0"
```
mlegner: Nit: The caret is not necessary as this is the default anyway.
```suggestion
opentelemetry = "0.23.0"
opentelemetry-http = "0.12.0"
opentelemetry-otlp = "0.16"
opentelemetry_sdk = "0.23.0"
```

docs/telemetry-attributes.csv
```diff
@@ -0,0 +1,14 @@
...
+walrus.shard_index;int;Index of a shard within the Walrus committee.
+walrus.sliver.pair_index;int;Index of a sliver pair within the encoding of a blob.
+walrus.sliver.remote_pair_index;int;Index of a sliver pair located at a remote node.
+walrus.sliver.type;string;Type of the encoded sliver.;Example `primary` or `secondary`.
+walrus.recovery.symbol_type;string;Type of the recovery symbol, either `primary` or `secondary`.
+walrus.event.tx_digest;string;Digest of the Sui transaction that generated the event.
+walrus.event.event_seq;int;Sequence number of the event in the Sui transaction that generated it.
+walrus.event.kind;string;Type of the Walrus event;One of `certified`, `registered`, `invalid-blob`.
+walrus.event.index;int;Index of the Walrus event in the stream of all events.
+# TODO(jsmith): Change string repr of shard index
```
mlegner, line 14: Question: What is the plan / intended change here?

docs/telemetry-attributes.csv
```diff
@@ -0,0 +1,14 @@
+id;type;brief;note
+walrus.blob_id;string;Url-safe base64 encoded blob ID.
+walrus.epoch;int;Walrus epoch.
+walrus.node.public_key;string;Base64 encoded public key of a walrus node.
+walrus.shard_index;int;Index of a shard within the Walrus committee.
+walrus.sliver.pair_index;int;Index of a sliver pair within the encoding of a blob.
+walrus.sliver.remote_pair_index;int;Index of a sliver pair located at a remote node.
+walrus.sliver.type;string;Type of the encoded sliver.;Example `primary` or `secondary`.
+walrus.recovery.symbol_type;string;Type of the recovery symbol, either `primary` or `secondary`.
```
mlegner: Nit:
```suggestion
walrus.sliver.type;string;Type of the encoded sliver.;One of `primary`, `secondary`.
walrus.recovery.symbol_type;string;Type of the recovery symbol.;One of `primary`, `secondary`.
```

crates/walrus-service/src/telemetry.rs
```diff
@@ -0,0 +1,222 @@
...
+impl MakeHttpSpan {
+    pub fn new() -> MakeHttpSpan {
+        Self
+    }
+
+    fn make_span<B>(&mut self, request: &Request<B>) -> Span {
+        let route = self.get_route(request);
+        let span = tracing::info_span!(
+            "rest_api",
+            // Override the exported span name to "{http.request.method} {http.route}"
```
mlegner: Question: How does that override happen?

crates/walrus-service/src/telemetry.rs
```diff
@@ -0,0 +1,222 @@
...
+    fn record_client_address<B>(
+        &self,
+        _request: &Request<B>,
+        span: &Span,
+        peer_ip: Option<IpAddr>,
+    ) {
+        // Check the standard Forwarded header for the client's address, and fallback to the
+        // de-facto but non-standardized X-Forwarded-For header, before accepting that the peer is
+        // also the client.
+        // TODO(jsmith): Parse the forwarded headers to correctly report the client behind proxies
```
mlegner: Remark: Open TODO.

crates/walrus-service/src/telemetry.rs
```diff
@@ -0,0 +1,222 @@
...
+
+    /// Record the server's address as the host identified in the header, or forwarded by a proxy.
+    fn record_server_address<B>(&self, request: &Request<B>, span: &Span) {
+        // Set the server address based on https://opentelemetry.io/docs/specs/semconv/http/http-spans/
+        // Should be the first of:
+        // - The original host which may be passed by the reverse proxy in the Forwarded#host,
+        //   X-Forwarded-Host, or a similar header.
+        // - The :authority pseudo-header in case of HTTP/2 or HTTP/3
+        // - The Host header.
+        // TODO(jsmith): Handle forwarded hosts and HTTP/2 authority
```
mlegner: Remark: Open TODO.

crates/walrus-service/src/telemetry.rs
```diff
@@ -0,0 +1,222 @@
...
+            (http::version::Version::HTTP_10, "1.0"),
+            (http::version::Version::HTTP_11, "1.1"),
+            (http::version::Version::HTTP_2, "2.0"),
+            (http::version::Version::HTTP_3, "3.0"),
+        ]
+        .iter()
+        .find(|(version, _)| *version == request.version())
+        {
+            span.record("network.protocol.version", value);
+        }
```
mlegner: Nit: This feels a bit convoluted to me. How about something like this?
```suggestion
        let version = match request.version() {
            http::version::Version::HTTP_09 => "0.9",
            http::version::Version::HTTP_10 => "1.0",
            http::version::Version::HTTP_11 => "1.1",
            http::version::Version::HTTP_2 => "2.0",
            http::version::Version::HTTP_3 => "3.0",
            _ => {
                return;
            }
        };
        span.record("network.protocol.version", version);
```

crates/walrus-service/src/committee.rs
```diff
@@ -404,12 +425,15 @@ where
...
                 "attempting to recover sliver using collected symbols"
             );
-            let result = Sliver::<A>::recover_sliver_or_generate_inconsistency_proof(
-                recovery_symbols.clone(),
-                sliver_id.to_sliver_index::<A>(config.n_shards()),
-                metadata.as_ref(),
-                config,
-            );
+
+            let result = tracing::info_span!("reconstruct_sliver").in_scope(|| {
```
mlegner: Nit:
```suggestion
            let result = tracing::info_span!("recover_sliver").in_scope(|| {
```

Cargo.toml
jpcsmith: The goal was to make it really explicit, since everything breaks the second one of these is out of sync with the others, but I can change it if you think it's unecessary.

docs/telemetry-attributes.csv
jpcsmith: Remove the display prefix. If an unambiguous representation is required, then we have the debug representation `ShardIndex(_)`. Currently the prefix in the string representation prevents it from being meaningfully used, as we cannot format the prefix for use in any pleasant human output, and in other cases (such as logging) it's associated with a label regardless of the display representation.

Cargo.toml
mlegner: Ah, I see. Then maybe use an `=` instead?

crates/walrus-service/src/telemetry.rs
jpcsmith: It's a [special field](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/#special-fields) handled by `tracing_opentelemetry` which we use for export. Replaced with with word 'Overrides'.

crates/walrus-service/src/telemetry.rs
jpcsmith: Yep, that's nicer, thanks.

crates/walrus-service/src/committee.rs
jpcsmith: `recover_sliver` is used for the function name of 2 other spans already, hence the change of name. But I can name it after the function which is being called: `recover_sliver_or_generate_inconsistency_proof`.

PR #558, opened 2024-06-20
feat(stress): Write clients generator

asonnino:
Add a write clients generator to periodically replace the ones we initially created. This is useful for continous load generation when we need to avoid that the clients eventually run out of SUI to pay for gas.

### DIFF
--- a/crates/walrus-service/src/client/error.rs
+++ b/crates/walrus-service/src/client/error.rs
@@ -58,15 +58,18 @@ impl ClientError {
 
     /// Whether the error is an out-of-gas error.
     pub fn is_out_of_coin_error(&self) -> bool {
-        matches!(&self.kind, ClientErrorKind::NotEnoughGas(..))
+        matches!(
+            &self.kind,
+            ClientErrorKind::NoCompatiblePaymentCoin | ClientErrorKind::NoCompatibleGasCoin(..)
+        )
     }
 }
 
 impl From<SuiClientError> for ClientError {
     fn from(value: SuiClientError) -> Self {
         let kind = match value {
-            SuiClientError::NoCompatiblePaymentCoin => ClientErrorKind::NotEnoughGas(None),
-            SuiClientError::NoCompatibleGasCoin(e) => ClientErrorKind::NotEnoughGas(Some(e)),
+            SuiClientError::NoCompatiblePaymentCoin => ClientErrorKind::NoCompatiblePaymentCoin,
+            SuiClientError::NoCompatibleGasCoin(e) => ClientErrorKind::NoCompatibleGasCoin(e),
             error => ClientErrorKind::Other(error.into()),
         };
         Self { kind }
@@ -104,9 +107,12 @@ pub enum ClientErrorKind {
     /// The blob ID is blocked.
     #[error("the blob ID {0} is blocked")]
     BlobIdBlocked(BlobId),
-    /// The client does not have enough gas coins to interact with the blockchain.
-    #[error("the client is out of gas")]
-    NotEnoughGas(Option<anyhow::Error>),
+    #[error("no compatible payment coin found")]
+    /// No matching payment coin found for the transaction.
+    NoCompatiblePaymentCoin,
+    #[error("no compatible gas coin found: {0}")]
+    /// No matching gas coin found for the transaction.
+    NoCompatibleGasCoin(anyhow::Error),
     /// A failure internal to the node.
     #[error("client internal error: {0}")]
     Other(Box<dyn std::error::Error + Send + Sync + 'static>),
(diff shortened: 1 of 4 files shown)


crates/walrus-stress/src/main.rs
```diff
@@ -46,6 +49,10 @@ struct Args {
     /// The blob size to use for the load generation.
     #[clap(long, default_value = "10000")]
     blob_size: NonZeroUsize,
+    /// The period in seconds for the gas refill. This option is useful for continuous load testing
+    /// where the gas budget need to be refilled periodically.
+    #[clap(long, default_value = "None")]
+    gas_refill_period: Option<NonZeroU64>,
```
giac-mysten: Question: do I understand correctly that this is not really a "gas refill", but more a "switch to a new client with fresh gas"? If yes, maybe mention it here.

asonnino: yes, the way we implement this "gas refill" (in spirit) is by injecting new clients full of gas

karlwuest: I haven't had time to go through the PR in detail yet, but why do we go through the effort of creating new clients? Why don't we just keep a list of addresses and periodically request funds for those addresses from the faucet? This could just run in an independent thread.

crates/walrus-stress/src/generator.rs
```diff
@@ -185,11 +190,12 @@ impl LoadGenerator {
                         } else if let Some(entry) = blob_ids.choose_mut(&mut thread_rng()) {
                             blob_result.blob_id().clone_into(entry);
                         }
+                        self.write_client_pool.push_back((client, blob));
                     } else {
+                        // This may happen if the client runs out of gas.
```
karlwuest: Remark: This can also happen for other reasons. We should probably match on the `ClientErrorKind` contained in the error that is returned and also provide the reason as part of the output.

crates/walrus-stress/src/generator.rs
```diff
@@ -200,6 +206,9 @@ impl LoadGenerator {
                         metrics.observe_error("failed to read blob");
                     }
                 },
+                Some(new_clients) = rx_new_clients.recv() => {
+                    self.write_client_pool.extend(new_clients);
+                }
```
karlwuest: Remark: I think it's better to keep the number of clients constant to better control what's happening. By adding more clients, we may just increase the error rate (e.g. due to too many open files).

Suggestion: As mentioned in my other comment, I think an approach where we just add more gas objects to the already existing clients would be better. This should be fairly straight forward and can run in an independent task. We only need the addresses to periodically request gas from the faucet. Even better would be to also use a Sui client to periodically check if we still have enough gas and then just refresh it if that's not the case.

crates/walrus-stress/src/generator.rs
```diff
@@ -241,3 +250,34 @@ fn burst_load(load: u64) -> (u64, Interval) {
...
+            interval.tick().await;
+            let clients = try_join_all(
+                (0..n_write_clients).map(|_| new_client(&client_config, network, gas_budget)),
+            )
+            .await?;
+            let blobs = (0..n_write_clients).map(|_| {
+                BlobData::random(
+                    StdRng::from_rng(thread_rng()).expect("rng should be seedable from thread_rng"),
+                    blob_size,
+                )
```
karlwuest: Remark: Regenerating new random blobs becomes quite expensive for large blob sizes, and for large blob sizes we need to refresh more often, since we run out of gas more quickly.

crates/walrus-stress/src/generator.rs
giac-mysten: for this I would suggest adding a function `is_out_of_coin_error` to the `ClientError`, which matches internally.

crates/walrus-service/src/client/error.rs
```diff
@@ -55,6 +55,22 @@ impl ClientError {
...
+    pub fn is_out_of_coin_error(&self) -> bool {
+        matches!(&self.kind, ClientErrorKind::NotEnoughGas(..))
+    }
+}
+
+impl From<SuiClientError> for ClientError {
+    fn from(value: SuiClientError) -> Self {
+        let kind = match value {
+            SuiClientError::NoCompatiblePaymentCoin => ClientErrorKind::NotEnoughGas(None),
+            SuiClientError::NoCompatibleGasCoin(e) => ClientErrorKind::NotEnoughGas(Some(e)),
```
karlwuest: Suggestion: Keep the difference between these two also in the `ClientErrorKind`, since the payment coin and the gas coin won't necessarily have the same type in the future.

crates/walrus-stress/src/generator.rs
```diff
@@ -251,33 +274,30 @@ fn burst_load(load: u64) -> (u64, Interval) {
...
         loop {
             interval.tick().await;
-            let clients = try_join_all(
-                (0..n_write_clients).map(|_| new_client(&client_config, network, gas_budget)),
+
+            try_join_all(
+                addresses
+                    .iter()
+                    .cloned()
+                    .map(|address| request_sui_from_faucet(address, network, &sui_client)),
```
karlwuest: Suggestion: I don't think we need to wait for the coins to be received in each request, so instead of using `request_sui_from_faucet` here, we could just make `send_faucet_request` public and use that.

PR #561, opened 2024-06-20
fix(client): limit the number of connections dependent on the blob size

### LINKED ISSUE #556: Tune client parameters for large blobs
When storing largo blobs (100s of MiB) on the Walrus devnet, there are lots of errors like the following:

> 2024-06-19T15:54:47.598998Z DEBUG reserve_and_store_blob{blob_id="4uaZ2AHAyDlEou7E-17QmM3-OmBxhOvO5rHEWF-ZMvE"}:store_metadata_and_pairs:store_sliver_by_axis: walrus_sdk::client: error=error sending request for url (http://lax-dnt-sto-00.devnet.sui.io:9185/v1/blobs/4uaZ2AHAyDlEou7E-17QmM3-OmBxhOvO5rHEWF-ZMvE/slivers/96/primary)

While the store operation succeeds in the end, this is still not ideal. My suspicion is that when storing many large slivers in parallel (here one primary sliver is 10 MiB) some of those requests fail due to timeouts or something like this.

One possibility would be to adapt the number of parallel connections based on the blob size.

mlegner:
This adds a added a `max_data_in_flight` config parameter with a default of 100 Mb = 12.5 MB and limits the number of connections for storing and reading slivers such that their total amount of data does not exceed that maximum.

It also decreases the default timeout to limit the impact of Byzantine nodes.

Requires some documentation changes. There we can also emphasize that the communication config may not work well for large blobs with very bad networking.

Closes #556

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -6,7 +6,8 @@
 use std::{collections::HashMap, sync::Arc, time::Instant};
 
 use anyhow::anyhow;
-use communication::{NodeCommunication, NodeResult};
+use communication::{NodeCommunication, NodeReadCommunication, NodeResult, NodeWriteCommunication};
+use config::CommunicationLimits;
 use error::StoreError;
 use fastcrypto::{bls12381::min_pk::BLS12381AggregateSignature, traits::AggregateAuthenticator};
 use futures::Future;
@@ -43,7 +44,6 @@ use walrus_sui::{
     utils::price_for_unencoded_length,
 };
 
-use self::config::default;
 use crate::client::utils::CompletedReasonWeight;
 
 mod blocklist;
@@ -77,14 +77,7 @@ pub struct Client<T> {
     // INV: committee.n_shards > 0
     committee: Committee,
     storage_price_per_unit_size: u64,
-    // The maximum number of nodes to contact in parallel when writing.
-    max_concurrent_writes: usize,
-    // The maximum number of shards to contact in parallel when reading slivers.
-    max_concurrent_sliver_reads: usize,
-    // The maximum number of nodes to contact in parallel when reading metadata.
-    max_concurrent_metadata_reads: usize,
-    // The maximum number of nodes to contact in parallel when getting the blob status.
-    max_concurrent_status_reads: usize,
+    communication_limits: CommunicationLimits,
     encoding_config: EncodingConfig,
     global_write_limit: Arc<Semaphore>,
     blocklist: Option<Blocklist>,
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/client/config.rs
```diff
@@ -75,9 +69,94 @@ impl ClientCommunicationConfig {
...
+            max_concurrent_sliver_reads,
+            max_concurrent_metadata_reads,
+            max_concurrent_status_reads,
+            max_data_in_flight,
+        }
+    }
+
+    fn max_connections_for_request_and_blob_size(
+        &self,
+        request_size: usize,
```
giac-mysten: Suggestion: Use `NonZeroUsize` to avoid division by 0?

crates/walrus-service/src/client/config.rs
```diff
@@ -75,9 +69,94 @@ impl ClientCommunicationConfig {
...
+    }
+
+    fn sliver_size_for_blob(
+        &self,
+        blob_size: NonZeroU64,
+        encoding_config: &EncodingConfig,
+    ) -> usize {
+        encoding_config
+            .sliver_size_for_blob::<Primary>(blob_size.get())
+            .expect("blob should not be too large to be encoded")
```
giac-mysten: Question: How do we enforce this? Since `CommunicationLimits` is `pub`, `blob_size` could be anything?

crates/walrus-service/src/client/config.rs
```diff
@@ -75,9 +69,94 @@ impl ClientCommunicationConfig {
...
+            .max(1)
+    }
+
+    fn sliver_size_for_blob(
+        &self,
+        blob_size: NonZeroU64,
+        encoding_config: &EncodingConfig,
+    ) -> usize {
+        encoding_config
+            .sliver_size_for_blob::<Primary>(blob_size.get())
```
giac-mysten, line 127: Question/remark: Seems we are over-approximating (by 33%?) the data in transfer by only using the size of primary slivers, right? I think it's fair.

crates/walrus-service/src/client.rs
```diff
@@ -499,7 +476,14 @@ impl<T> Client<T> {
         let enough_source_symbols =
             |weight| weight >= self.encoding_config.n_source_symbols::<U>().get().into();
         requests
-            .execute_weight(&enough_source_symbols, self.max_concurrent_sliver_reads)
+            .execute_weight(
+                &enough_source_symbols,
+                self.communication_limits
+                    .max_concurrent_writes_for_blob_size(
```
giac-mysten: Question: this is a bit confusing. We are in the `request_slivers_and_decode` function, and are requesting with a `max_concurrent_writes_for_blob_size` limit?
did you mean `reads`?

crates/walrus-service/src/client.rs
```diff
@@ -112,34 +105,19 @@ impl Client<()> {
...
-            config.communication_config.max_concurrent_metadata_reads;
-        let max_concurrent_status_reads = config
-            .communication_config
-            .max_concurrent_status_reads
-            .unwrap_or(default::max_concurrent_status_reads(committee.n_shards()));
-        let global_write_limit = Arc::new(Semaphore::new(max_concurrent_writes));
+        let communication_limits =
+            CommunicationLimits::new(&config.communication_config, encoding_config.n_shards());
+        let global_write_limit =
+            Arc::new(Semaphore::new(communication_limits.max_concurrent_writes));
```
giac-mysten, line 115: Question: Is the write limit per blob never applied? See line 348, there we don't limit the number of storage nodes, and the semaphore here uses the generic value?

crates/walrus-service/src/client/config.rs
mlegner: Good point. Let me make the struct (more) private and emphasize that these functions panic if they are not called with valid blob sizes.

crates/walrus-service/src/client.rs
mlegner: Yes, this was an oversight.

crates/walrus-service/src/client.rs
mlegner: Hmm... You are right. But why did the behavior change then...? 🤔 
I will invistigate.

mlegner: It seems the change in behavior was just due to different networking conditions, so my "fix" didn't actually change anything. 🤦

I will try to apply a blob-specific connection limit.

crates/walrus-service/src/client/config.rs
mlegner: Yes, that's true. I didn't find a convenient way of making this more precise. But as the default value is anyway somewhat arbitrary, I don't think this discrepancy matters too much.

crates/walrus-service/src/client.rs
mlegner: OK, I've now fixed the behavior for the store operation by introducing yet another semaphore... 😅 
This seems to work, but makes the setup a bit more complicated. Let me know what you think.

PR #563, opened 2024-06-21
feat: resolve `~` to home directory in paths

mlegner:
So far, this only worked in direct CLI options where it is expanded by the shell. With this change, it also works for the JSON API and for paths in config files.

For example, one can now set a wallet config path in the `client_config.yaml` as follows:

```yaml
wallet_config: ~/.sui/sui_config/client.yaml
```

Also, the following JSON API example now works, which didn't before:

```sh
walrus json '{
    "config": "~/.walrus/client_config.yaml",
    "command": {"blobId": {"file": "~/.walrus/client_config.yaml"}}
}'
```

The `~` is only replaced if it occurs as a directory at the beginning of paths.

### DIFF
--- a/crates/walrus-service/src/utils.rs
+++ b/crates/walrus-service/src/utils.rs
@@ -1,18 +1,20 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Utility functions used internally in the crate.
+//! Utility functions for the Walrus service.
 
 use std::{
     fmt::Debug,
     future::Future,
     num::Saturating,
+    path::PathBuf,
     pin::Pin,
     sync::Arc,
     task::{ready, Context, Poll},
     time::Duration,
 };
 
+use anyhow::{Context as _, Result};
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/utils.rs
```diff
@@ -315,6 +318,41 @@ where
...
+        let home = home::home_dir().context("unable to resolve home directory")?;
+        return Ok(home.join(
+            path.strip_prefix("~")
+                .expect("we just checked for this prefix"),
+        ));
+    } else {
+        Ok(path)
+    }
+}
+
```
giac-mysten, line 355: Question: did you consider [std::fs::canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html)?

mlegner: AFAIU, `canonicalize` does not replace the `~` with the home directory. I can of course call this in addition, but I'm not sure how much this helps.

I'm merging this in the current state, but please let me know if you would like to see a follow-up. 🙂

PR #566, opened 2024-06-21
feat: add http tracing to the sdk client

jpcsmith:
Add HTTP tracing to the client and propagate spans from the client to servers, so that calls made using the client can be associated with traces at the server.

Additionally, rename functions in `walrus-sdk::client` such that generic methods have no suffix and enum methods have a `_by_type` suffix.

Depends on #562

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -53,16 +53,25 @@ impl UrlEndpoints {
         self.0.join(&format!("/v1/blobs/{blob_id}/")).unwrap()
     }
 
-    fn metadata(&self, blob_id: &BlobId) -> Url {
-        self.blob_resource(blob_id).join("metadata").unwrap()
+    fn metadata(&self, blob_id: &BlobId) -> (Url, &'static str) {
+        (
+            self.blob_resource(blob_id).join("metadata").unwrap(),
+            METADATA_URL_TEMPLATE,
+        )
     }
 
-    fn confirmation(&self, blob_id: &BlobId) -> Url {
-        self.blob_resource(blob_id).join("confirmation").unwrap()
+    fn confirmation(&self, blob_id: &BlobId) -> (Url, &'static str) {
+        (
+            self.blob_resource(blob_id).join("confirmation").unwrap(),
+            STORAGE_CONFIRMATION_URL_TEMPLATE,
+        )
     }
 
-    fn blob_status(&self, blob_id: &BlobId) -> Url {
-        self.blob_resource(blob_id).join("status").unwrap()
+    fn blob_status(&self, blob_id: &BlobId) -> (Url, &'static str) {
+        (
+            self.blob_resource(blob_id).join("status").unwrap(),
+            STATUS_URL_TEMPLATE,
+        )
     }
 
     fn recovery_symbol<A: EncodingAxis>(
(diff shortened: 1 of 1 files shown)


crates/walrus-sdk/src/client.rs
```diff
@@ -30,10 +32,19 @@ use walrus_core::{
...
 };
 
+const METADATA_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/metadata";
+const SLIVER_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/slivers/:sliver_pair_index/:sliver_type";
+const STORAGE_CONFIRMATION_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/confirmation";
+const RECOVERY_URL_TEMPLATE: &str =
+    "/v1/blobs/:blob_id/slivers/:sliver_pair_index/:sliver_type/:target_pair_index";
+const INCONSISTENCY_PROOF_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/inconsistent/:sliver_type";
+const STATUS_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/status";
+const HEALTH_URL_TEMPLATE: &str = "/v1/health";
```
giac-mysten, line 46: Question/suggestion: are these "duplicates" of the data in `UrlEndpoints`? May be worth using these and `replace`?

crates/walrus-sdk/src/client.rs
```diff
@@ -422,16 +442,114 @@ impl Client {
...
+    /// The HTTP span ends after the parsing of the headers, since the response may be streamed.
+    async fn send_request(
+        &self,
+        mut request: Request,
+        url_template: &'static str,
+    ) -> Result<(Response, Span), NodeError> {
+        let url = request.url();
+        let span = tracing::info_span!(
+            "http_request",
+            otel.name = format!("{} {}", request.method().as_str(), url_template),
```
giac-mysten, line 480: Question: the method also reported in `otel.method`. Is the duplication necessary/mandated?

crates/walrus-sdk/src/client.rs
```diff
@@ -239,181 +271,169 @@ impl Client {
...
-            .expect("successfully bcs encodes within limit defaults");
-
-        let request = self.inner.put(url).body(encoded_metadata);
-        request
-            .send()
-            .await
-            .map_err(Kind::Reqwest)?
-            .response_error_for_status()
+        let request = self.create_request_with_payload(Method::PUT, url, metadata.as_ref())?;
+        self.request_json::<String>(request, METADATA_URL_TEMPLATE)
```
giac-mysten: (maybe a)Suggestion: the `self.endpoints...` functions could return the appropriate url, so that it does not have to be selected manually?

crates/walrus-sdk/src/client.rs
```diff
@@ -422,16 +442,114 @@ impl Client {
...
+                span.record("otel.status_code", "ERROR");
+                span.record("otel.status_message", err.to_string());
+                span.record("error.type", "reqwest::Error");
+
+                Err(NodeError::reqwest(err))
+            }
+        }
+    }
+
+    async fn request_bcs<T: DeserializeOwned>(
```
giac-mysten: Nit: I was confused initially b/c all requests (with payload) are in bcs. Here bcs is the content of the response. Though all descriptive names are much longer.

crates/walrus-sdk/src/client.rs
```diff
@@ -422,16 +442,114 @@ impl Client {
...
+            .map_err(NodeError::reqwest)
+    }
+}
+
+// opentelemetry_http currently uses too low a version of the http crate,
+// so we reimplement the injector here.
+struct HeaderInjector<'a>(pub &'a mut HeaderMap);
+
+impl<'a> Injector for HeaderInjector<'a> {
+    fn set(&mut self, key: &str, value: String) {
```
giac-mysten, line 565: Question: we ignore headers if they don't match `HeaderName` and `HeaderValue`

crates/walrus-service/src/storage/event_cursor_table.rs
```diff
@@ -48,10 +48,16 @@ impl EventCursorTable {
...
-            event_queue: Arc::new(Mutex::new(EventSequencer::default())),
-        })
+            event_queue: Arc::default(),
+        };
+
+        let next_index = this.get_sequentially_processed_event_count()?;
+        *this.event_queue.lock().unwrap() =
+            EventSequencer::continue_from(next_index.try_into().expect("64-bit architecture"));
+
+        Ok(this)
```
giac-mysten: Question: why is this necessary?

crates/walrus-sdk/src/client.rs
jpcsmith: Do you mean `http.request.method`? Yes, both of these are as per the otel [semantic specification for HTTP](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#name). `otel.name` replaces "http_request" as the name of the span when it's exported to the telemetry collectors, and the `http.request.method` is the HTTP method. These are separate, searchable fields in things like grafana.

crates/walrus-sdk/src/client.rs
jpcsmith: Changed to `send_and_parse_bcs_response`

crates/walrus-sdk/src/client.rs
jpcsmith: Do you mean we ignore the header if they are not valid header names and values? Then yes, but this is only used by the propagator which should provide valid strings.

crates/walrus-service/src/storage/event_cursor_table.rs
jpcsmith: My apologies, this was not part of this PR, and was instead part of #562 .

crates/walrus-sdk/src/client.rs
jpcsmith: Each call to replace performs a new allocation, which seems a bit unfortunate for URLs with multiple replacements, and format strings would also require duplicating these strings...

PR #572, opened 2024-06-25
feat(stress): improve stability of stress client

karlwuest:
Several small improvements:
- limits number of concurrent reads which reduces the error rate
- checks the gas objects for each address before requesting gas from the faucet, which allows checking more often
- allows specifying a write or read load of 0

### DIFF
--- a/crates/walrus-stress/src/generator.rs
+++ b/crates/walrus-stress/src/generator.rs
@@ -21,8 +21,8 @@ use tokio::{
     task::JoinHandle,
     time::{Interval, MissedTickBehavior},
 };
-use walrus_core::encoding::Primary;
-use walrus_service::client::{Client, Config};
+use walrus_core::{encoding::Primary, BlobId};
+use walrus_service::client::{Client, ClientError, Config};
 use walrus_sui::{
     client::{SuiContractClient, SuiReadClient},
     test_utils::wallet_for_testing_from_faucet,
@@ -33,7 +33,12 @@ use walrus_test_utils::WithTempDir;
 use crate::{metrics, metrics::ClientMetrics};
 
 const DEFAULT_GAS_BUDGET: u64 = 100_000_000;
+
+// If a node has less than `MIN_NUM_COINS` without at least `MIN_COIN_VALUE`,
+// we need to request additional coins from the faucet.
 const MIN_COIN_VALUE: u64 = 500_000_000;
+// We need at least a payment coin and a gas coin.
+const MIN_NUM_COINS: usize = 2;
 
 /// Minimum burst duration.
 const MIN_BURST_DURATION: Duration = Duration::from_millis(100);
@@ -144,24 +149,7 @@ impl LoadGenerator {
                 write_interval.period().as_millis()
             );
         } else {
-            // Do a single write, s.t. we have a blob to read.
-            let (client, blob) = self
-                .write_client_pool
-                .pop_front()
-                .expect("write client should be available");
-            loop {
-                let result = client
-                    .as_ref()
-                    .reserve_and_store_blob(blob.as_ref(), 1, true)
-                    .await;
-                if let Ok(blob_result) = result {
-                    blob_ids.push(blob_result.blob_id().to_owned());
-                    break;
-                } else {
-                    tracing::error!("initial write failed, retrying");
-                }
-            }
-            self.write_client_pool.push_back((client, blob));
+            self.single_write().await?;
         }
 
         let (reads_per_burst, read_interval) = burst_load(read_load);
@@ -281,6 +269,18 @@ impl LoadGenerator {
         }
         Ok(())
     }
+
+    async fn single_write(&self) -> Result<BlobId, ClientError> {
+        let (client, blob) = self
+            .write_client_pool
+            .front()
+            .expect("write client should be available");
+        client
+            .as_ref()
+            .reserve_and_store_blob(blob.as_ref(), 1, true)
+            .await
+            .map(|blob_result| blob_result.blob_id().to_owned())
+    }
 }
 
 async fn new_client(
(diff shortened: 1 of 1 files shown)


crates/walrus-stress/src/generator.rs
```diff
@@ -110,26 +128,51 @@ impl LoadGenerator {
...
+                let result = client
+                    .as_ref()
+                    .reserve_and_store_blob(blob.as_ref(), 1, true)
+                    .await;
+                if let Ok(blob_result) = result {
+                    blob_ids.push(blob_result.blob_id().to_owned());
+                    break;
+                } else {
+                    tracing::error!("initial write failed, retrying");
+                }
```
giac-mysten: Suggestion: this loop may be a bit aggressive.. back off or wait before retrying?

crates/walrus-sui/src/utils.rs
```diff
@@ -414,15 +414,15 @@ pub async fn request_sui_from_faucet(
         let _ = send_faucet_request(address, network)
             .await
             .inspect_err(|e| tracing::warn!(error = ?e, "faucet request failed, retrying"))
-            .inspect(|_| tracing::info!("waiting to receive tokens from faucet"));
-        tracing::info!("sleeping for {backoff:?}");
+            .inspect(|_| tracing::debug!("waiting to receive tokens from faucet"));
+        tracing::debug!("sleeping for {backoff:?}");
```
asonnino, line 418: good idea!

crates/walrus-stress/src/generator.rs
```diff
@@ -110,26 +128,51 @@ impl LoadGenerator {
...
+            }
+            self.write_client_pool.push_back((client, blob));
+        }
```
asonnino: Should we isolate this into a function (its goal seem to only ensure we make at least one write)?

crates/walrus-stress/src/generator.rs
```diff
@@ -273,24 +326,35 @@ pub fn refill_gas(
...
+                    .get_coins_stream(address, None)
+                    .filter(|coin| future::ready(coin.balance >= MIN_COIN_VALUE))
+                    .take(2)
+                    .collect::<Vec<_>>()
+                    .await
+                    .len()
+                    < 2
+                {
+                    tracing::debug!("Clients gas coins refilled");
+                    metrics.observe_gas_refill();
```
giac-mysten, line 352: Question/suggestion: move these after the actual request

crates/walrus-stress/src/generator.rs
```diff
@@ -217,14 +263,18 @@ impl LoadGenerator {
...
+                            self.metrics.observe_latency(metrics::READ_WORKLOAD, elapsed);
+                        }
+                        Err(e) => {
+                            tracing::error!(error=?e, "failed to read blob");
+                            self.metrics.observe_error("failed to read blob");
+                        }
                     }
+                    self.read_client_pool.push_back(client);
                 },
                 else => break
```
asonnino, line 267: I am wondering whether we shouldn't have two different tasks, one to submit the writes and one to submit the reads. But this can be done through another PR

crates/walrus-stress/src/generator.rs
```diff
@@ -273,24 +326,35 @@ pub fn refill_gas(
...
+                    .len()
+                    < 2
+                {
+                    tracing::debug!("Clients gas coins refilled");
+                    metrics.observe_gas_refill();
+                    send_faucet_request(address, network).await
+                } else {
+                    Ok(())
+                }
+            }))
```
asonnino: maybe `iter().filter()` can help simplify this expression

crates/walrus-stress/src/generator.rs
```diff
@@ -273,24 +326,35 @@ pub fn refill_gas(
...
-                    .map(|address| send_faucet_request(address, network)),
-            )
+            let sui_client = &sui_client;
+            let metrics = &metrics;
+            try_join_all(addresses.iter().cloned().map(|address| async move {
+                if sui_client
+                    .coin_read_api()
+                    .get_coins_stream(address, None)
+                    .filter(|coin| future::ready(coin.balance >= MIN_COIN_VALUE))
+                    .take(2)
```
asonnino: Why take(2)? Should we provide some comment?

crates/walrus-stress/src/generator.rs
```diff
@@ -273,24 +326,35 @@ pub fn refill_gas(
...
+                    .collect::<Vec<_>>()
+                    .await
+                    .len()
+                    < 2
```
asonnino: Why `2` ?

crates/walrus-stress/src/generator.rs
karlwuest: This is because we need 2 coin objects (one for the payment, one for the gas). I'll add a comment.

crates/walrus-stress/src/generator.rs
karlwuest: Same as above, I'll clarify

crates/walrus-stress/src/generator.rs
karlwuest: Do you mean two separate methods on the generator instead of one `run` function?

crates/walrus-stress/src/generator.rs
karlwuest: I'll actually change this to just do a write and return early on error. I don't think it's necessary to keep retrying.

crates/walrus-stress/src/generator.rs
karlwuest: Do you mean to replace the `if`? 
Yes, that would be nicer, but since we need to do an async call to get the coins, that requires turning the addresses into a stream first, and then it quickly becomes less readable.

crates/walrus-stress/src/generator.rs
asonnino: I mean two different tokio tasks (so two different reactor loops), one for reads and one for writes.

karlwuest: Ok, I agree. I guess this will become even more crowded when we add inconsistent writes. Let's clean this up in another PR.

PR #588, opened 2024-07-08
feat(client): add dry-run option for store

### LINKED ISSUE #564: Add store dry-run command to the Walrus CLI
This should do the following:
- encode the blob,
- determine the unencoded and encoded size,
- get the current status, and
- determine the storage cost.

mlegner:
Also fix a minor bug in the tracing-subscriber initialization.

Example output: 
image

Closes #564

### DIFF
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -21,7 +21,7 @@ use serde::{Deserialize, Serialize};
 use serde_with::{base64::Base64, serde_as, DisplayFromStr};
 use tracing_subscriber::{layer::SubscriberExt as _, util::SubscriberInitExt, EnvFilter, Layer};
 use walrus_core::{
-    encoding::{EncodingConfig, Primary},
+    encoding::{encoded_blob_length_for_n_shards, EncodingConfig, Primary},
     metadata::VerifiedBlobMetadataWithId,
     BlobId,
 };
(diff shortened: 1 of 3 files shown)


crates/walrus-service/bin/client.rs
```diff
@@ -632,17 +669,47 @@ async fn run_app(app: App) -> Result<()> {
...
+                let encoded_size =
+                    encoded_blob_length_for_n_shards(encoding_config.n_shards(), unencoded_size)
+                        .expect("must be valid as the encoding succeeded");
+                let price_per_unit_size = client
+                    .sui_client()
+                    .read_client()
+                    .price_per_unit_size()
+                    .await?;
+                let storage_cost =
+                    price_for_encoded_length(encoded_size, price_per_unit_size, epochs);
```
karlwuest, line 693: Suggestion: Move this into a function (e.g. a method on `Client`) that returns a `DryRunOutput`.

mlegner: Yes, that's a good point. I was just about to do that but then I thought that it makes more sense to do a more holistic refactoring of the client binary, in particular the 200-lines `run_app` function... 😇

So I'll merge this in but will create a follow-up refactoring PR.

mlegner: -> #589

Issue #581, opened 2024-07-01
Client: check if quorum of storage nodes is reachable and report an error otherwise


mlegner:
Currently, if the user has an old system object configured after a redeployment, they will get a somewhat unhelpful error message. Ideally, we would print a more specific/useful error message if we cannot reach a quorum of storage nodes and suggest users to update their config and client.

One option for this would be a health check that is run when starting the client (by default) and which reports a specific error. This could also be made available as a separate command and/or as part of the info command.

Issue #565, opened 2024-06-21
(How) Should we support empty blobs on Walrus?


mlegner:
Currently, Walrus does not support empty blobs. This is annoying when using Walrus Sites, as there are legitimate reasons to have empty files in websites.

So we definitely should support empty files in some way. We can do this in multiple ways:
1. Have a special case in Walrus Sites on the Sui side.
1. Add a special case in the Walrus client. For example, use blob ID `0x0` for empty files.
1. Simply treat it as any other blob and adjust the encoding.

PR #590, opened 2024-07-10
feat(client): make reserve and register atomic

### LINKED ISSUE #587: Use PTB to combine getting a storage resource and registering a blob
Currently, the client obtains a storage resource in one transaction and then registers the blob in a second one. This can be an issue when the two requests are sent to different RPC nodes, which is possible when using an RPC proxy such as [Suibase](https://suibase.io/how-to/proxy.html).

The problem can be resolved by performing both actions in a single PTB.

mlegner:
The main change is the new
`walrus_sui::client::ContractClient::reserve_and_register_blob` method, which is used in `walrus_service::client`. The implementation of this in `SuiContractClient` is accompanied with some refactoring of the contract calls.

Besides this main change, this commit contains the following changes:
- more useful error messages for empty blobs and missing gas coin
- move price computations from `walrus_service::cli_utils` to `walrus_sui::utils`
- ignore simtest by default
- improve some docstrings

Closes #587

Related to #565, #581

### DIFF
--- a/crates/walrus-service/src/client/error.rs
+++ b/crates/walrus-service/src/client/error.rs
@@ -60,7 +60,7 @@ impl ClientError {
     pub fn is_out_of_coin_error(&self) -> bool {
         matches!(
             &self.kind,
-            ClientErrorKind::NoCompatiblePaymentCoin | ClientErrorKind::NoCompatibleGasCoin(..)
+            ClientErrorKind::NoCompatiblePaymentCoin | ClientErrorKind::NoCompatibleGasCoin
         )
     }
 }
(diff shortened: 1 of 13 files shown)


crates/walrus-service/src/client/error.rs
```diff
@@ -107,12 +113,12 @@ pub enum ClientErrorKind {
...
     #[error("the blob ID {0} is blocked")]
     BlobIdBlocked(BlobId),
-    #[error("no compatible payment coin found")]
     /// No matching payment coin found for the transaction.
+    #[error("no compatible payment coin found")]
     NoCompatiblePaymentCoin,
-    #[error("no compatible gas coin found: {0}")]
     /// No matching gas coin found for the transaction.
-    NoCompatibleGasCoin(anyhow::Error),
+    #[error("no compatible gas coin found; two separate coins are required to reserve storage")]
```
karlwuest: Nit: The cause of this error is not necessarily that there is only one coin available.

mlegner: Good point, although this is the most likely cause for the error. I've slightly reformulated this and added "with sufficient funds" to make it clearer.

PR #593, opened 2024-07-11
ci: use cargo nextest in pre-commit hook and CI

### LINKED ISSUE #393: Start using nextest to run tests in CI and with pre-commit
See https://nexte.st and the [usage in the Sui repo](https://github.com/search?q=repo:MystenLabs/sui%20nextest&type=code) for inspiration.

halfprice:
Resolve #393

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -121,7 +121,9 @@ jobs:
     if: ${{ needs.diff.outputs.isRust == 'true' }}
     runs-on: ubuntu-ghcloud
     steps:
-      - uses: taiki-e/install-action@nextest
+      - uses: taiki-e/install-action@v2.41.14
+        with:
+          tool: cargo-nextest
       - uses: actions/checkout@v4
       - uses: Swatinem/rust-cache@v2.7.3
         with:


CONTRIBUTING.md
```diff
@@ -64,13 +64,13 @@ for example Clippy.
 ## Tests
```
mlegner, line 65: Suggestion: Can you mention that it's necessary to install `cargo-nextest` here and above in the "Pre-commit hooks" section?

.github/workflows/code.yml
```diff
@@ -121,12 +121,13 @@ jobs:
     if: ${{ needs.diff.outputs.isRust == 'true' }}
     runs-on: ubuntu-ghcloud
     steps:
+      - uses: taiki-e/install-action@nextest
```
mlegner: Suggestion: Can we pin the version here?
```suggestion
      - uses: taiki-e/install-action@v2.41.14
        with:
          tool: cargo-nextest
```

halfprice: Why do we want to pin install-action version?

CONTRIBUTING.md
halfprice: Done

.github/workflows/code.yml
mlegner: I generally prefer pinned versions so that they don't break at some random point when a new version is released. Instead, they are upgraded monthly in a more controlled manner by Dependabot. 

Admittedly, this may not be very important in this specific case. But given that we have automatic updates through Dependabot I also don't see a disadvantage.

halfprice: I see. As long as we don't need to manually update the version. This sounds good. This is also more explicit if something goes wrong in install-action future version.

mlegner: For the automatic updates, see the [Dependabot config](https://github.com/MystenLabs/walrus/blob/main/.github/dependabot.yml) and https://github.com/MystenLabs/walrus/pull/582 as an example PR.

Issue #262, opened 2024-04-12
Provide more customization in Rust SDK for contract interactions


karlwuest:
- [ ] Allow specifying which payment coins to use
- [x] Merge payment coins before buying a storage resource in a PTB if necessary (#1211)
- [ ] Clean up coins with zero balance
- [x] Provide functionality to get all owned storage resources / blob objects -> #739 
- [x] Make sui bindings more ptb friendly & allow multiple operations in one transaction (#1135 , #1211)

PR #594, opened 2024-07-11
feat(client): optimize gas and payment coin management

### LINKED ISSUE #524: Add mechanisms to split SUI coins if necessary
Currently, we need at least two coins: one to pay for the storage and one to pay for gas. It would be nice if the client would automatically split a coin if only one is available.

mlegner:
- Allow the use of multiple gas coins that have sufficient total balance.
- For `register_space` transactions, distinguish between `WAL == SUI` and `WAL != SUI`:
  - For `WAL != SUI`, keep previous behavior.
  - For `WAL == SUI` use the gas coin(s) also for payment by splitting off a sufficient amount.
- This contains some refactoring of the `SuiContractClient` and the `SuiReadClient`.

Example transaction block: [A8HHL84HuUWaASXKYsRPJ4uYjbEaeFySRH7G3wtCdzmz](https://testnet.suivision.xyz/txblock/A8HHL84HuUWaASXKYsRPJ4uYjbEaeFySRH7G3wtCdzmz).

Closes #524
Related to #262

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -41,7 +41,6 @@ use crate::{
     contracts::{self, FunctionTag},
     types::{Blob, StorageResource},
     utils::{
-        call_args_to_object_ids,
         get_created_sui_object_ids_by_type,
         get_sui_object,
         price_for_encoded_length,
@@ -200,11 +199,7 @@ impl SuiContractClient {
             pt_builder.transfer_arg(self.wallet_address, Argument::NestedResult(result_index, i));
         }
 
-        self.sign_and_send_ptb(
-            pt_builder.finish(),
-            self.get_compatible_gas_coins(call_args).await?,
-        )
-        .await
+        self.sign_and_send_ptb(pt_builder.finish(), None).await
     }
 
     fn add_move_call_to_ptb(
@@ -227,24 +222,14 @@ impl SuiContractClient {
 
     async fn get_compatible_gas_coins(
         &self,
-        call_args: Vec<CallArg>,
-    ) -> SuiClientResult<Vec<ObjectRef>> {
-        self.get_compatible_gas_coins_with_balance(call_args, self.gas_budget)
-            .await
-    }
-
-    async fn get_compatible_gas_coins_with_balance(
-        &self,
-        call_args: Vec<CallArg>,
-        balance: u64,
+        min_balance: Option<u64>,
     ) -> SuiClientResult<Vec<ObjectRef>> {
         Ok(self
             .read_client
             .get_coins_with_total_balance(
                 self.wallet_address,
                 None,
-                balance,
-                call_args_to_object_ids(call_args),
+                min_balance.unwrap_or(self.gas_budget),
             )
             .await
             .map_err(SuiClientError::NoCompatibleGasCoins)?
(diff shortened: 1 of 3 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -282,13 +296,126 @@ impl SuiContractClient {
...
+                    encoded_size,
+                    epochs_ahead,
+                    price,
+                    system_object_arg,
+                    pt_builder,
+                )
+                .await?,
+                self.gas_budget,
+            )
+        };
```
karlwuest: Suggestion: Instead of having two functions `reserve_arguments_with_gas_coin` and `reserve_arguments_with_separate_coin`, just use a single function and make the case distinction there. See proposed function below. With that function, we still need an `if` here for the gas budget, but it still seems cleaner. Alternatively, the function could be modified to also return the value needed for the gas coins.

crates/walrus-sui/src/client.rs
```diff
@@ -282,13 +296,126 @@ impl SuiContractClient {
...
+        let payment_coin = self.get_payment_coin(price).await?;
+        Ok([
+            system_object_arg.clone(),
+            encoded_size.into(),
+            epochs_ahead.into(),
+            payment_coin.object_ref().into(),
+        ]
+        .iter()
+        .map(|arg| pt_builder.input(arg.to_owned()))
+        .collect::<Result<Vec<_>>>()?)
```
karlwuest: Suggestion: Combine these two functions.

```suggestion
    /// Prepares and returns the arguments for the `reserve_space` function to buy a storage using
    /// either the gas coin if the payment coin type is Sui, adding the necessary transaction to the PTB,
    /// or a separate payment coin otherwise.
    async fn reserve_arguments(
        &self,
        encoded_size: u64,
        epochs_ahead: u64,
        price: u64,
        system_object_arg: CallArg,
        pt_builder: &mut ProgrammableTransactionBuilder,
    ) -> SuiClientResult<Vec<Argument>> {
        let payment_coin_arg = if self.read_client.uses_sui_coin() {
            let price_argument = pt_builder.input(call_arg_pure!(&price))?;
            let Argument::Result(split_result_index) =
                pt_builder.command(Command::SplitCoins(Argument::GasCoin, vec![price_argument]))
            else {
                unreachable!("this always returns an `Argument::Result`")
            };
            Argument::NestedResult(split_result_index, 0)
        } else {
            pt_builder.input(self.get_payment_coin(price).await?.object_ref().into())?
        };

        Ok(vec![
            pt_builder.input(system_object_arg.clone())?,
            pt_builder.input(encoded_size.into())?,
            pt_builder.input(epochs_ahead.into())?,
            payment_coin_arg,
        ])
```

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -155,20 +158,71 @@ impl SuiReadClient {
...
+    ///
+    /// Filters out any coin objects included in the `forbidden_objects` set.
+    ///
+    /// Returns `None` if no coins of sufficient total balance are found.
+    pub(crate) async fn get_coins_with_total_balance(
+        &self,
+        owner_address: SuiAddress,
+        coin_type: Option<String>,
+        balance: u64,
+        forbidden_objects: BTreeSet<ObjectID>,
```
karlwuest: Remark: Now that we split the gas coin if the payment coin type is Sui, I don't think we necessarily need forbidden coins anymore.

crates/walrus-sui/src/client.rs
mlegner: Sounds good, will make that change.

crates/walrus-sui/src/client/read_client.rs
mlegner: Yes, good point. I kept it because I wasn't sure if this will be needed in the future, but probably not. I'll remove it, shouldn't be too difficult to add it again later.

mlegner: I removed the `forbidden_objects` feature and some code that was no longer used in 28b4823ccba2c12f094ef77315379fe363ad0024.

crates/walrus-sui/src/client.rs
mlegner: Thanks for the suggestion. I used your code but integrated it directly into the `add_reserve_transaction_and_get_gas_coins` function (renamed to `add_reserve_transaction`) in 28b4823ccba2c12f094ef77315379fe363ad0024.

PR #599, opened 2024-07-15
chore(formatting): format existing move code

damirka:
- Runs prettier with dev-version of Move plugin
- Renames `ERROR_` to conventional `ESomethingHappened`
- Uses `sui-1.29` which is going out this week (and contributes macros to stdlib)
- Uses receiver syntax and index syntax from Move 2024


contracts/blob_store/Move.toml
```diff
@@ -3,7 +3,7 @@ name = "blob_store"
 edition = "2024.beta"
 
 [dependencies]
-Sui = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-framework/packages/sui-framework", rev = "testnet-v1.28.3" }
+Sui = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-framework/packages/sui-framework", rev = "releases/sui-v1.29.0-release" }
```
karlwuest: Remark/question: We generally try to keep all sui dependencies at the same version, is there a specific reason to go with `releases/sui-v1.29.0-release` instead of `testnet-v1.29.0`?

contracts/blob_store/sources/blob.move
```diff
@@ -76,18 +77,13 @@ module blob_store::blob {
...
     /// Derive the blob_id for a blob given the root_hash, erasure_code_type and size.
-    public fun derive_blob_id(root_hash: u256, erasure_code_type: u8, size : u64) : u256 {
-
-        let blob_id_struct = BlobIdDerivation {
-            erasure_code_type,
-            size,
-            root_hash,
-        };
+    public fun derive_blob_id(root_hash: u256, erasure_code_type: u8, size: u64): u256 {
+        let blob_id_struct = BlobIdDerivation { erasure_code_type, size, root_hash };
```
karlwuest: Question/Suggestion:
Is this the result of the autoformatter? 
If so, can the formatter be set to write large(-ish) struct instantiations on multiple lines? I think having them on a single line hurts readability.

contracts/blob_store/sources/redstuff.move
```diff
@@ -12,16 +11,15 @@ module blob_store::redstuff {
...
     /// This computation is the same as done by the function of the same name in
     /// `crates/walrus_core/encoding/config.rs` and should be kept in sync.
-    public(package) fun encoded_blob_length(unencoded_length: u64, n_shards: u16) : u64 {
-        let slivers_size = ((source_symbols_primary(n_shards) as u64)
-            + (source_symbols_secondary(n_shards) as u64))
-            * (symbol_size(unencoded_length, n_shards) as u64);
+    public(package) fun encoded_blob_length(unencoded_length: u64, n_shards: u16): u64 {
+        let slivers_size = (
+            (source_symbols_primary(n_shards) as u64) + (source_symbols_secondary(n_shards) as u64),
+        ) * (symbol_size(unencoded_length, n_shards) as u64);
```
karlwuest: Question: Is this from the autoformatter? 

Suggestion: Since the brackets here are used to group mathematical expressions and not for function calls, I would expect the formatting to be something like the following (most importantly without the comma):

```suggestion
        let slivers_size =
            ((source_symbols_primary(n_shards) as u64) + (source_symbols_secondary(n_shards) as u64))
        * (symbol_size(unencoded_length, n_shards) as u64);
```

contracts/blob_store/sources/tests/blob_tests.move
```diff
@@ -116,26 +122,82 @@ module blob_store::blob_tests {
...
-
+    public fun test_blob_certify_single_function(): system::System<TESTWAL> {
         let mut ctx = tx_context::dummy();
 
         // Derive blob ID and root_hash from bytes
         let root_hash_vec = vector[
-            1, 2, 3, 4, 5, 6, 7, 8,
-            1, 2, 3, 4, 5, 6, 7, 8,
-            1, 2, 3, 4, 5, 6, 7, 8,
-            1, 2, 3, 4, 5, 6, 7, 8,
```
karlwuest: Question/Suggestion: Can we tell the autoformatter to keep the existing formatting here, something equivalent to `#[rustfmt::skip]`, s.t. we don't get one line per entry?
(Obviously applies to all following cases as well)

contracts/blob_store/Move.toml
damirka: Good catch! I wasn't sure if testnet branch was created for today's testnet release. Updated the version to match the release.

PR #600, opened 2024-07-17
feat(blob info)!: add registered/certified/invalidated epoch info in blob_info

halfprice:
The main purpose for adding `certified_epoch` is that at beginning of a epoch change, a storage node can get the list of blobs to fetch/send shards.

`registered_epoch` can be useful for cleaning up too old registered but not yet certified blobs.

Also versioning the BlobInfo data structure in blob_info table for upgradability.

BREAKING CHANGE: this changes the on disk data format of the `blob_info` table.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -44,7 +44,7 @@ use crate::{
     committee::{CommitteeService, CommitteeServiceFactory, SuiCommitteeServiceFactory},
     config::{StorageNodeConfig, SuiConfig},
     contract_service::{SuiSystemContractService, SystemContractService},
-    storage::{blob_info::BlobInfoAPI, EventProgress, ShardStorage, Storage},
+    storage::{blob_info::BlobInfoApi, EventProgress, ShardStorage, Storage},
     system_events::{SuiSystemEventProvider, SystemEventProvider},
 };
 
--- a/crates/walrus-service/src/storage.rs
+++ b/crates/walrus-service/src/storage.rs
@@ -55,7 +55,7 @@ use self::{
     event_cursor_table::EventCursorTable,
     event_sequencer::EventSequencer,
 };
-use crate::storage::blob_info::{BlobInfoAPI, Mergeable as _};
+use crate::storage::blob_info::{BlobInfoApi, Mergeable as _};
 
 pub(crate) mod blob_info;
 mod event_cursor_table;
@@ -398,7 +398,8 @@ impl Storage {
     pub fn has_metadata(&self, blob_id: &BlobId) -> Result<bool, TypedStoreError> {
         Ok(self
             .get_blob_info(blob_id)?
-            .map(|info| info.is_metadata_stored())
+            .as_ref()
+            .map(BlobInfo::is_metadata_stored)
             .unwrap_or_default())
     }
 
@@ -695,15 +696,17 @@ pub(crate) mod tests {
 
     async_param_test! {
         update_blob_info -> TestResult: [
-            contains_register: (false),
-            skip_register: (true),
+            in_order: (false, false),
+            skip_register: (true, false),
+            skip_certify: (false, true),
         ]
     }
-    async fn update_blob_info(skip_register: bool) -> TestResult {
+    async fn update_blob_info(skip_register: bool, skip_certify: bool) -> TestResult {
         let storage = empty_storage();
         let storage = storage.as_ref();
         let blob_id = BLOB_ID;
 
+        let registered_epoch = if skip_register { None } else { Some(1) };
         if !skip_register {
             let state0 = BlobInfo::new(
                 1,
@@ -723,33 +726,34 @@ pub(crate) mod tests {
             assert_eq!(storage.get_blob_info(&blob_id)?, Some(state0));
         }
 
-        let registered_epoch = if skip_register { None } else { Some(1) };
-
-        let state1 = BlobInfo::new_for_testing(
-            42,
-            BlobCertificationStatus::Certified,
-            event_id_for_testing(),
-            registered_epoch,
-            Some(2),
-            None,
-        );
-        storage.merge_update_blob_info(
-            &blob_id,
-            BlobInfoMergeOperand::ChangeStatus {
-                status_changing_epoch: 2,
-                end_epoch: 42,
-                status: BlobCertificationStatus::Certified,
-                status_event: state1.current_status_event(),
-            },
-        )?;
-        assert_eq!(storage.get_blob_info(&blob_id)?, Some(state1));
+        let certified_epoch = if skip_certify { None } else { Some(2) };
+        if !skip_certify {
+            let state1 = BlobInfo::new_for_testing(
+                42,
+                BlobCertificationStatus::Certified,
+                event_id_for_testing(),
+                registered_epoch,
+                certified_epoch,
+                None,
+            );
+            storage.merge_update_blob_info(
+                &blob_id,
+                BlobInfoMergeOperand::ChangeStatus {
+                    status_changing_epoch: 2,
+                    end_epoch: 42,
+                    status: BlobCertificationStatus::Certified,
+                    status_event: state1.current_status_event(),
+                },
+            )?;
+            assert_eq!(storage.get_blob_info(&blob_id)?, Some(state1));
+        }
 
         let state2 = BlobInfo::new_for_testing(
             42,
             BlobCertificationStatus::Invalid,
             event_id_for_testing(),
             registered_epoch,
-            Some(2),
+            certified_epoch,
             Some(3),
         );
         storage.merge_update_blob_info(
--- a/crates/walrus-service/src/storage/blob_info.rs
+++ b/crates/walrus-service/src/storage/blob_info.rs
@@ -63,8 +63,8 @@ impl From<BlobCertificationStatus> for SdkBlobCertificationStatus {
 }
 
 #[enum_dispatch]
-/// The `BlobInfoAPI` trait defines methods for retrieving information about a blob.
-pub trait BlobInfoAPI {
+/// The `BlobInfoApi` trait defines methods for retrieving information about a blob.
+pub trait BlobInfoApi {
     /// Returns the end epoch of the blob.
     fn end_epoch(&self) -> Epoch;
 
@@ -173,7 +173,7 @@ impl BlobInfoV1 {
     }
 }
 
-impl BlobInfoAPI for BlobInfoV1 {
+impl BlobInfoApi for BlobInfoV1 {
     fn end_epoch(&self) -> Epoch {
         self.end_epoch
     }
@@ -235,7 +235,7 @@ impl Mergeable for BlobInfoV1 {
 
 /// Internal representation of the BlobInfo for use in the database etc.
 /// Use [`BlobStatus`] for anything public facing (e.g. communication to the client).
-#[enum_dispatch(BlobInfoAPI)]
+#[enum_dispatch(BlobInfoApi)]
 #[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
 pub enum BlobInfo {
     V1(BlobInfoV1),


Cargo.toml
```diff
@@ -20,6 +20,7 @@ license = "Apache-2.0"
 anyhow = "1.0.86"
 bcs = "0.1.6"
 clap = { version = "4.5.9", features = ["derive"] }
+enum_dispatch = "^0.3"
```
mlegner: Remark: We have a few other locations in the code where this could be useful.

Also, nit:
```suggestion
enum_dispatch = "0.3"
```

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -61,75 +62,157 @@ impl From<BlobCertificationStatus> for SdkBlobCertificationStatus {
     }
 }
 
+#[enum_dispatch]
+pub trait BlobInfoAPI {
```
mlegner: Suggestion: Add a default implementation

```rust
    fn into_blob_status(&self) -> BlobStatus {
        BlobStatus::Existent {
            end_epoch: self.end_epoch(),
            status: self.status().into(),
            status_event: self.current_status_event(),
        }
    }
```

Then we can also add

```rust
impl From<BlobInfo> for BlobStatus {
    fn from(value: BlobInfo) -> Self {
        value.into_blob_status()
    }
}
```

and keep the `.map(BlobStatus::from)` in `node.rs`.

crates/walrus-service/src/node.rs
```diff
@@ -1443,20 +1449,22 @@ mod tests {
...
 
         let node_client = cluster.client(0);
         let config = &blob1_details.config;
 
         // Send events that some unobserved blob has been certified.
-        let blob2_details = EncodedBlob::new(&blob2, config.clone());
-        let blob2_registered_event = BlobRegistered::for_testing(*blob2_details.blob_id());
+        // let blob2_details = EncodedBlob::new(&blob2, config.clone());
+        // let blob2_registered_event = BlobRegistered::for_testing(*blob2_details.blob_id());
+        let blob2_registered_event = BlobRegistered::for_testing(BLOB_ID);
```
mlegner: Question: What is the purpose of these changes?

crates/walrus-service/src/storage.rs
```diff
@@ -518,21 +518,39 @@ fn merge_blob_info(
...
+                } => Some(BlobInfo::new(
+                    registered_epoch,
+                    end_epoch,
+                    status,
+                    status_event,
+                )),
+                BlobInfoMergeOperand::CertifyBlob { .. } => {
+                    let blob_id = bcs::from_bytes::<BlobId>(key);
+                    panic!("Certifying an unknown blob {:?}", blob_id);
+                }
```
mlegner: Important: We need to make sure this is compatible with the new event stream. cc @sadhansood

crates/walrus-service/src/storage.rs
```diff
@@ -518,21 +518,39 @@ fn merge_blob_info(
...
+                BlobInfoMergeOperand::InvalidateBlob { .. } => {
+                    let blob_id = bcs::from_bytes::<BlobId>(key);
+                    panic!("Certifying an unknown blob {:?}", blob_id);
```
mlegner: Nit:
```suggestion
                    panic!("Invalidating an unknown blob {:?}", blob_id);
```

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -139,9 +222,76 @@ impl Mergeable for BlobInfo {
...
+        certified_epoch: Epoch,
+        end_epoch: Epoch,
+        status: BlobCertificationStatus,
+        status_event: EventID,
+    },
+    InvalidateBlob {
+        invalidated_epoch: Epoch,
         end_epoch: Epoch,
         status: BlobCertificationStatus,
         status_event: EventID,
```
mlegner: Question: What is the benefit of this refactoring? We could also just add an `change_epoch: Epoch` field and otherwise match on the `status` field.

crates/walrus-service/src/storage.rs
halfprice: Can you clarify which part needs compatibility with event stream? The operand is created inside the node, so as long as the event stream delivers a certified blob event, we can always convert it into an internal operand.

mlegner: So the current design of the event stream contains implicitly defined blobs that are never explicitly registered but will still be certified in some form. So we could get the case where we have a so far unknown blob ID for which we get a "certified" event.

We don't necessarily have to solve this in the current PR, but we have to keep it in mind for the future.

halfprice: Got it. I think you mentioned this yesterday, and I plan to change the logic here so that the certify blob doesn't need a register blob event (not sure about invalidate blob without a certify blob event though)

sadhansood: My understanding is like @halfprice, that event stream creates the blob certified event for event blobs just like any other blob. @mlegner - Were you thinking that event blob certification should emit a special event type for event blobs (I don't think we would need it)?

sadhansood: @mlegner - see your comment now, it makes sense. Definitely event blobs will not have registered event emitted.

mlegner: > Were you thinking that event blob certification should emit a special event type for event blobs (I don't think we would need it)?

No, not necessarily. What I wanted to highlight is that the code currently doesn't handle "certified" events without seeing an appropriate "registered" event first. We either need to change this behavior or emit a registered event as well in the event stream.

Cargo.toml
halfprice: Done

crates/walrus-service/src/storage/blob_info.rs
halfprice: Great! Thanks for the suggestion.

crates/walrus-service/src/node.rs
halfprice: It used to be that we can't certify the blob before it is registered.

This is updated since certify blob no longer requires registration.

crates/walrus-service/src/storage.rs
karlwuest: I think we should just accept certified events without seeing a registered event before.

halfprice: PR updated and ready for review. Now, certifying blob no longer requires registration. @mlegner @sadhansood PTAL!

crates/walrus-service/src/storage.rs
halfprice: No longer relevent.

crates/walrus-service/src/storage/blob_info.rs
halfprice: Yeah, good point. This was created due to that initially, I have different events with different type of fields. This is no longer relevant, so having a single ChangeStatus makes more sense now. And PR is updated reflecting this.

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -61,75 +62,168 @@ impl From<BlobCertificationStatus> for SdkBlobCertificationStatus {
     }
 }
 
-/// Internal representation of the BlobInfo for use in the database etc.
-/// Use [`BlobStatus`] for anything public facing (e.g. communication to the client).
+#[enum_dispatch]
+/// The `BlobInfoAPI` trait defines methods for retrieving information about a blob.
+pub trait BlobInfoAPI {
```
mlegner: Nit: I guess the [canonical name](https://rust-lang.github.io/api-guidelines/naming.html#casing-conforms-to-rfc-430-c-case) would be `BlobInfoApi`. (Sorry, didn't notice in my first pass.)

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -61,75 +62,168 @@ impl From<BlobCertificationStatus> for SdkBlobCertificationStatus {
...
                 status,
                 status_event,
             } => {
-                if self.status < status || self.status == status && self.end_epoch < end_epoch {
-                    self.status = status;
-                    self.current_status_event = status_event;
-                    self.end_epoch = end_epoch;
-                }
+                self.update_status_changing_epoch(status, status_changing_epoch);
+                self.update_status(end_epoch, status, status_event);
```
mlegner, line 226: Important: If the new status is "lower" than the old one, I don't think we should update the status-changing epoch. Maybe we can move the call to `update_status_changing_epoch` into the if clause in the `update_status` method?

crates/walrus-service/src/storage.rs
```diff
@@ -398,7 +398,7 @@ impl Storage {
     pub fn has_metadata(&self, blob_id: &BlobId) -> Result<bool, TypedStoreError> {
         Ok(self
             .get_blob_info(blob_id)?
-            .map(|info| info.is_metadata_stored)
+            .map(|info| info.is_metadata_stored())
```
mlegner: Hint/nit: Now that this is a method, we could also write this as follows (no change necessary, just noticed this):
```suggestion
            .map(BlobInfo::is_metadata_stored)
```

crates/walrus-service/src/storage.rs
```diff
@@ -682,69 +693,108 @@ pub(crate) mod tests {
...
 
-    #[tokio::test]
-    async fn update_blob_info() -> TestResult {
+    async_param_test! {
+        update_blob_info -> TestResult: [
+            contains_register: (false),
+            skip_register: (true),
+        ]
+    }
+    async fn update_blob_info(skip_register: bool) -> TestResult {
```
mlegner: Suggestion: Can we add an additional `skip_certify` parameter to also test for invalidation without certification?

crates/walrus-service/src/storage/blob_info.rs
halfprice: Done

crates/walrus-service/src/storage.rs
halfprice: Yeah, didn't do it earlier because we need to convert the Option to a reference first.

crates/walrus-service/src/storage.rs
halfprice: Done

crates/walrus-service/src/storage/blob_info.rs
halfprice: I was thinking about this too. The reason I chose to apply the status changing epoch even though the status is lower is because the old implementation (the one I moved to update_status) seems to indicate that out of order events is somewhat expected (receiving certify before register). If so, I think it makes sense to still record the old status changing epoch instead of simply discard it (register event is somehow processed after certify event).

We also discussed that we really shouldn't see out of order events. In such case, instead of silently ignoring the lower events, we probably should return an error. I think this may be something we can keep in mind and see if we want to change to this behavior.

mlegner: Hmm... So I believe we can really assume that events arrive in order. Then we only have the issue of "special" system-internal blobs that don't require registration. If we want to be strict but still allow these internal blobs, we could add a flag to the certified event to indicate that this is a non-registered blob (cc @karlwuest).

If we don't want to assume in-order events, I understand that we may want to later record previously missed status changes. However, even in that case, we shouldn't just overwrite the status changing epochs. In the current implementation, if we get a new `BlobCertified` event, we would overwrite a previously recorded `certified_epoch`.

So the minimum changes I would make are something like the following:

```rust
/// Updates the epoch where the status changed.
fn update_status_changing_epoch(
    &mut self,
    status: BlobCertificationStatus,
    status_changing_epoch: Epoch,
) {
    match status {
        BlobCertificationStatus::Registered => {
            self.registered_epoch = Some(
                self.registered_epoch
                    .unwrap_or_default()
                    .min(status_changing_epoch),
            );
        }
        BlobCertificationStatus::Certified => {
            self.certified_epoch = Some(
                self.certified_epoch
                    .unwrap_or_default()
                    .min(status_changing_epoch),
            );
        }
        BlobCertificationStatus::Invalid => {
            self.invalidated_epoch = Some(
                self.invalidated_epoch
                    .unwrap_or_default()
                    .min(status_changing_epoch),
            );
        }
    }
}
```

halfprice: Summarize the discussion offline with @mlegner . The suggestion makes sense. What I wanted to point out was just that we either want to allow applying lower status info, or we should trigger an error. We shouldn't just silently ignore it. However, since we don't know exactly what the semantic is once we integrate all the components, what we will do here is to allow applying lower status info (for execution freedom) and log errors, for now. We will re-evaluate it once we have integrate all the components.

Thanks for the comments @mlegner !

Issue #589, opened 2024-07-08
Reorganize the client CLI code


mlegner:
Currently, the client binary code is quite unstructured with everything in one file and all the logic in a single 200-line `run_app` function. It would be easier to navigate if we split the file into multiple modules and the function into multiple smaller functions (maybe even de-duplicating some code in the process)

- Place the logic for each command in a separate function in a commands module (similar to the walrus-node binary).
- Put all the arguments definitions into a separate module.
- Maybe a better structure for `client::cli_output` and `client::responses`. The original idea for the separation was to use the responses also in the daemon, but it may be a bit nicer to have the two a bit closer together.

Issue #462, opened 2024-05-30
Make JSON formats of client daemon more consistent and generate OpenAPI specification


mlegner:
We have an OpenAPI specification for the storage nodes since #448. We should do the same for the client daemon.

PR #601, opened 2024-07-17
feat(daemon): add endpoint with OpenAPI specification

mlegner:
This contains quite some refactoring changes, so it's best reviewed commit-by-commit.

It intentionally does not contain any breaking changes, all response and output formats remain unchanged. The format of HTTP responses (in particular errors) will be changed in a separate PR.

Contributes to #462 
Contributes to #589

### DIFF
--- a/crates/walrus-service/src/client/daemon/openapi.rs
+++ b/crates/walrus-service/src/client/daemon/openapi.rs
@@ -65,8 +65,8 @@ mod tests {
     #[test]
     fn test_openapi_generation_does_not_panic() {
         std::fs::write(
-            // Can also be used to view the api.
-            Path::new("/tmp/api.html"),
+            // Can also be used to view the API.
+            Path::new("/tmp/api-daemon.html"),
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/client/daemon/openapi.rs
```diff
@@ -0,0 +1,74 @@
...
+    use std::path::Path;
+
+    use utoipa::OpenApi as _;
+    use utoipa_redoc::Redoc;
+
+    use super::*;
+
+    #[test]
+    fn test_openapi_generation_does_not_panic() {
+        std::fs::write(
```
giac-mysten, line 67: Question: why do we need to write the generated api? Can we just `Redoc::new(..`?

mlegner: Yes, I guess so. I copied this from JP's tests and I assume his idea was that the written file can be used to inspect the generated docs.

However, I just noticed that I should at least use different filenames... 😅

PR #602, opened 2024-07-17
feat: staking contract skeleton

damirka:


contracts/walrus/sources/staked_wal.move
```diff
@@ -0,0 +1,63 @@
...
+module walrus::staked_wal {
+    /// Represents a staked WAL, does not store the `Balance` inside, but uses
+    /// `u64` to represent the staked amount. Behaves similarly to `Balance` and
+    /// `Coin` providing methods to `split` and `join`.
+    public struct StakedWal has key, store {
+        id: UID,
+        /// ID of the staking pool.
+        pool_id: ID,
+        /// The staked amount.
+        principal: u64,
```
karlwuest: We also need to add the epoch here.

contracts/walrus/sources/staked_wal.move
```diff
@@ -0,0 +1,63 @@
...
+            pool_id: staked_wal.pool_id,
+            principal: amount,
+        }
+    }
+
+    /// Joins the staked WAL with another staked WAL.
+    /// Both staked WALs must belong to the same staking pool.
+    public fun join(staked_wal: &mut StakedWal, other: StakedWal, ctx: &mut TxContext) {
+        let StakedWal { id, pool_id, principal } = other;
+
```
karlwuest: Add a TODO to handle the case where the staking epochs are not the same. I think to handle this case, we will need to pass a reference to the system object, s.t. we can do the conversion of the value from one epoch to the other.

contracts/walrus/sources/staking_pool.move
```diff
@@ -0,0 +1,49 @@
...
+
+    /// Stake the given amount of WAL in the pool.
+    public fun stake<T>(
+        pool: &mut StakingPool<T>,
+        to_stake: Coin<T>,
+        ctx: &mut TxContext,
+    ): StakedWal {
+        abort ENotImplemented
+    }
+
```
karlwuest: Suggestion: add a skeleton for `request_withdrawal` as well.

contracts/walrus/sources/staking_pool.move
```diff
@@ -0,0 +1,49 @@
...
+    use sui::coin::Coin;
+    use walrus::staked_wal::StakedWal;
+
+    /// TODO: remove this once the module is implemented.
+    const ENotImplemented: u64 = 0;
+
+    /// Represents a single staking pool for a token.
+    public struct StakingPool<phantom T> has store {
+        id: ID,
+        total_staked: Balance<T>,
```
karlwuest: Suggestion: split this into `active_stake` and `stake_to_withdraw` (feel free to choose a better name)

contracts/walrus/sources/storage_node.move
```diff
@@ -0,0 +1,86 @@
...
+        storage_price: u64,
+        write_price: u64,
+        storage_capacity: u64,
+    ) {}
+
+    public fun vote_end() {}
+
+    public fun epoch_sync_done() {}
+
+    public fun shart_transfer_failed() {}
```
karlwuest:
```suggestion
    public fun shard_transfer_failed() {}
```

contracts/walrus/sources/system.move
```diff
@@ -0,0 +1,130 @@
...
+    public fun vote_for_next_epoch<T>(
+        system: &mut System<T>,
+        cap: &StorageNodeCap,
+        storage_price: u64,
+        write_price: u64,
+        storage_capacity: u64,
+    ) {
+        assert!(system.version == VERSION);
+        system.inner_mut().vote_for_next_epoch(cap, storage_price, write_price, storage_capacity)
+    }
```
karlwuest: Let's maybe change this to individual `set_storage_price`, `set_write_price`, `set_storage_capacity` functions.

contracts/walrus/sources/system.move
```diff
@@ -0,0 +1,130 @@
...
+
+    public fun voting_end<T>(system: &mut System<T>, cap: &StorageNodeCap) {
```
karlwuest: I think we can make this callable by anyone, so we can remove the `cap`

contracts/walrus/sources/system.move
```diff
@@ -0,0 +1,130 @@
...
+        system
+            .inner_mut()
+            .shard_transfer_failed(
+                cap,
+                node_identity,
+                shard_ids,
+            )
+    }
+
+    public fun invalidate_blob_id<T>(
```
karlwuest: This doesn't need to take a `StorageNodeCap`, but has to take a certificate as argument, i.e., the same signature as the function in the current version:
```
        system: &System<T>,
        signature: vector<u8>,
        members: vector<u16>,
        message: vector<u8>,
```

contracts/walrus/sources/storage_node.move
```diff
@@ -0,0 +1,86 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#[allow(unused_field, unused_function, unused_variable)]
+module walrus::storage_node {
```
karlwuest: I think we don't need most of the public functions in this module since they are public functions in the system module.

damirka: This one is a leftover from initial layout, I'll patch that.

contracts/walrus/sources/system.move
damirka: Can we keep them together for the time being? Util we settled, I'd rather not have to maintain 5 more functions.

damirka: I'll add a TODO to get back to it.

karlwuest: Sure 👍

contracts/walrus/sources/staked_wal.move
damirka: Added a note.

contracts/walrus/sources/staked_wal.move
damirka: Currently removed implementation of this one until we figure out the epoch setting in general.

contracts/walrus/sources/staking_pool.move
damirka: Done!

contracts/walrus/sources/staking_pool.move
damirka: Done.

contracts/blob_store/Move.lock
```diff
@@ -2,19 +2,19 @@
 
 [move]
 version = 2
-manifest_digest = "537E35587055E526E85CD04C400536AFFEC5CC0BE833EB4B5FAC245850C34C96"
+manifest_digest = "04BC1483EBABEFA6569B42BDD39FD47C20DE62C3DFDD585117C55C51BC96B422"
```
damirka: Rolling this buddy back right now.

contracts/walrus/tests/staking_tests.move
karlwuest, line 1: Suggestion: remove this file as it's not used. We can add it once we need it.

scripts/move_coverage.sh
```diff
@@ -31,6 +31,13 @@ fi
 # 2) Check coverage and print summaries
 error=0
 for dir in contracts/*; do
+
+    # TODO: remove me when system is in a better shape
```
karlwuest, line 35: Can you create an issue for this and reference this here?

Issue #606, opened 2024-07-19
Determine storage node's behavior when seeing blob events out of order


halfprice:
This may not be possible once we integrate all the components. If that's the case, we need to make sure proper errors are generated.

PR #607, opened 2024-07-19
feat(blob info): when receiving duplicate blob events, preserve the lower status changing epoch

halfprice:
Also added error log line when out of order event is observed. We need to revisit once we have integrated with reliable event stream and whether we should turn this into an invariant violation, or need to deal with it.

Related to #606

### DIFF
--- a/crates/walrus-service/src/storage/blob_info.rs
+++ b/crates/walrus-service/src/storage/blob_info.rs
@@ -133,7 +133,7 @@ impl BlobInfoV1 {
 
     /// Updates the status of the blob.
     /// If the new status is higher than the current status, the status is updated.
-    /// If the new status is the same as the current status, the status with the higher epoch is kept.
+    /// If the new status is the same as the current status, the status with the higher end epoch is kept.
     fn update_status(
         &mut self,
         blob_id: &BlobId,
@@ -142,7 +142,7 @@ impl BlobInfoV1 {
         status_changing_epoch: Epoch,
         status_event: EventID,
     ) {
-        self.update_status_changing_epoch(blob_id, status, status_changing_epoch);
+        self.update_status_changing_epoch_to_earliest(blob_id, status, status_changing_epoch);
 
         if self.status < status || self.status == status && self.end_epoch < end_epoch {
             self.status = status;
@@ -152,7 +152,8 @@ impl BlobInfoV1 {
     }
 
     /// Updates the epoch where the status changed.
-    fn update_status_changing_epoch(
+    /// We keep the earliest epoch for each status.
+    fn update_status_changing_epoch_to_earliest(
         &mut self,
         blob_id: &BlobId,
         status: BlobCertificationStatus,
@@ -160,7 +161,7 @@ impl BlobInfoV1 {
     ) {
         if let Some(current_status_epoch) = self.status_changing_epoch(status) {
             if current_status_epoch <= status_changing_epoch {
-                // The status epoch is already set to a higher epoch.
+                // The status epoch is already set to a lower epoch.
                 return;
             }
 
@@ -276,7 +277,7 @@ impl BlobInfo {
         ));
 
         let BlobInfo::V1(inner) = &mut blob_info;
-        inner.update_status_changing_epoch(&blob_id, status, status_changing_epoch);
+        inner.update_status_changing_epoch_to_earliest(&blob_id, status, status_changing_epoch);
 
         blob_info
     }


crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -146,6 +140,16 @@ impl BlobInfoV1 {
...
         status_event: EventID,
     ) {
+        if self.status >= status {
+            // TODO(zhewu): revisit once we have a better understanding of contract behavior and
+            // whether out of order events are possible. (#606)
+            tracing::error!(
+                "Status already set to the same or a higher value: {:?} vs {:?}",
+                self.status,
+                status
+            );
```
karlwuest: Question: Why should we never see a lower status here? Do we check and filter this somewhere else? 

Remark 1: We get all events in order, but we can see registration and certification events for the same blob Id arbitrarily often. This means that we can easily have a registration event that follows a certification event for the same blob.

Remark 2: With blob deletion, we will anyway need to keep track of not just the latest status, but actually need to keep more fine-grained information about events (see [here](https://www.notion.so/mystenlabs/Blob-Lifecycle-Improvements-520bc6445e09465f92c50007b673d0a9?pvs=4#1a8aef734af94d11a6c93d3ed5ab4862)).

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -160,15 +164,55 @@ impl BlobInfoV1 {
...
-            }
-            BlobCertificationStatus::Invalid => {
-                self.invalidated_epoch = Some(status_changing_epoch);
-            }
+            BlobCertificationStatus::Registered => match self.registered_epoch {
+                Some(epoch) => {
+                    if epoch != status_changing_epoch {
+                        // TODO(zhewu): revisit once we have a better understanding of contract behavior and
+                        // whether duplicated events are possible. (#606)
+                        tracing::error!(
```
karlwuest: Suggestion: don't trace this on the error level, see comment above for context.

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -160,15 +164,55 @@ impl BlobInfoV1 {
...
+                None => {
+                    self.registered_epoch = Some(status_changing_epoch);
+                }
+            },
+            BlobCertificationStatus::Certified => match self.certified_epoch {
+                Some(epoch) => {
+                    if epoch != status_changing_epoch {
+                        // TODO(zhewu): revisit once we have a better understanding of contract behavior and
+                        // whether duplicated events are possible. (#606)
+                        tracing::error!(
```
karlwuest: Suggestion: Don't trace this on the error level, see comment above for context.

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -396,7 +436,7 @@ mod tests {
                         &initial_blob_info,
                         end_epoch,
                         *status, // Higher status
-                        7,
+                        1,
```
karlwuest, line 424: Question: Just out of curiosity, is there a reason for this change or is this arbitrary?

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -160,15 +164,55 @@ impl BlobInfoV1 {
...
+                            status_changing_epoch
+                        );
+                    }
+                    self.invalidated_epoch = Some(cmp::min(epoch, status_changing_epoch));
+                }
+                None => {
+                    self.invalidated_epoch = Some(status_changing_epoch);
+                }
+            },
         }
```
mlegner: I generally agree with @karlwuest's comments.

Also, nit: I believe this could be written a bit more concisely: By adding a `set_status_changing_epoch` trait method
```rust
fn set_status_changing_epoch(&mut self, status: BlobCertificationStatus, epoch: Epoch) {
    match status {
        BlobCertificationStatus::Registered => self.registered_epoch = Some(epoch),
        BlobCertificationStatus::Certified => self.certified_epoch = Some(epoch),
        BlobCertificationStatus::Invalid => self.invalidated_epoch = Some(epoch),
    };
}
```
this could be written as
```rust
        if let Some(epoch) = self.status_changing_epoch(status) {
            // TODO(zhewu): revisit once we have a better understanding of contract behavior and
            // whether duplicated events are possible. (#606)
            tracing::debug!(
                ?status,
                "epoch for status already set to a different value: {} (before) vs {} (after)",
                epoch,
                status_changing_epoch
            );
            self.set_status_changing_epoch(status, status_changing_epoch.min(epoch));
        } else {
            self.set_status_changing_epoch(status, status_changing_epoch);
        }
```

crates/walrus-service/src/storage/blob_info.rs
halfprice: See the context for this PR here: https://github.com/MystenLabs/walrus/pull/600#discussion_r1682414500

I think there is a big misunderstanding on my part, where I thought `one Blob object on chain corresponding to one blob in storage` (that's why I added all of these error messages that we shouldn't expect lower status). Looking at the contract and the doc again, I realize the blob in storage is ref-counted data that can be referenced by multiple Blob objects on chain. 

I think taking the min of all register/certify/invalid epoch of all the events is also incorrect here, since those events are for on chain Blob object, rather than for the blob data in storage, isn't it? Ideally, we want to keep track of such event per on chain Blob object, but we are currently missing the object id in the information from the event?

mlegner: > one Blob object on chain corresponding to one blob in storage

I believe we will actually switch to that. I.e., we keep a status per Sui `Blob` object and not per blob ID. So your code will probably make more sense then.

halfprice: Discussed with @mlegner earlier. I was thinking maybe I can implement sui object ref counting before the contract lands, but on a second thought, I think it's probably better to not adjust the code to an anticipated future, just in case the future may still change.

Now I know that seeing a lower status is expect, I have updated the PR to accommodate @karlwuest and @mlegner 's comments. I still added one warning log since although it is expected to see lower status, it's not expected for the same status, we see lower epoch in later events. That is what the warning is about.

crates/walrus-service/src/storage/blob_info.rs
halfprice: Revised.

crates/walrus-service/src/storage/blob_info.rs
halfprice: Revised

crates/walrus-service/src/storage/blob_info.rs
halfprice: Yes, there is a reason for this change. This is because before this PR, we always update the epoch according to the latest event. In this PR, the behavior is changed to always set the epoch to the smallest seen.

crates/walrus-service/src/storage/blob_info.rs
halfprice: Thanks for the suggestion @mlegner ! Updated the code similarly.

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -156,9 +150,26 @@ impl BlobInfoV1 {
     /// Updates the epoch where the status changed.
     fn update_status_changing_epoch(
         &mut self,
+        blob_id: &BlobId,
         status: BlobCertificationStatus,
         status_changing_epoch: Epoch,
     ) {
+        if let Some(current_status_epoch) = self.status_changing_epoch(status) {
+            if current_status_epoch <= status_changing_epoch {
+                // The status epoch is already set to a higher epoch.
```
karlwuest:
```suggestion
                // The status epoch is already set to a lower epoch.
```

crates/walrus-service/src/storage/blob_info.rs
```diff
@@ -217,12 +224,13 @@ impl Mergeable for BlobInfoV1 {
...
             BlobInfoMergeOperand::ChangeStatus {
+                blob_id,
                 status_changing_epoch,
                 end_epoch,
                 status,
                 status_event,
             } => {
-                self.update_status_changing_epoch(status, status_changing_epoch);
+                self.update_status_changing_epoch(&blob_id, status, status_changing_epoch);
                 self.update_status(end_epoch, status, status_event);
```
karlwuest: Question: Is this really the behaviour that we want, namely updating the status changing epoch and the status independently?

As an example: We could get an event in epoch 1 that certifies a blob until epoch 3, then in epoch 2 we get an event that the same blob is certified until epoch 4. In this case, the status is updated (with the new end epoch), but the the `status_changing_epoch` remains at 1.

This might be what we want, but from the name I would expect this value to be updated iff the status changes. What is the (intended future) use of the `status_changing_epoch`?

halfprice: Good call. Calling `update_status_changing_epoch` inside `update_status` now. I can see why it was confusing.

The reason I separated them out before was that I thought seeing lower status event is unexpected. So `update_status_changing_epoch` was created for dealing with "unexpected" lower status case. Ultimately, I mixed up blob with sui blob object. Now I know that it's expected, it changes how I see these events.

In the future, each blob will track "status_changing_event" per Sui blob object instead of per blob. There will be
an update_blob_object_status (such as register/certify events) as well as update_blob_status (such as invalidate event).

crates/walrus-service/src/storage/blob_info.rs
halfprice: Done

crates/walrus-service/src/storage/blob_info.rs
karlwuest: >In the future, each blob will track "status_changing_event" per Sui blob object instead of per blob. There will be
an update_blob_object_status (such as register/certify events) as well as update_blob_status (such as invalidate event).

My question was more directed at what the `status_changing_epoch` will be used for. I.e., is the intention that this tracks when the status was last changed (in my example, should it be updated to 2), or when the status was first set (current behaviour). In the former case, it would be good to rename this (e.g. `status_first_set`), in the latter, the logic to set this needs to be changed.

~The example with certification still applies even if we track this per sui blob object, because we will get certifications for extended blobs as well.~

PR #610, opened 2024-07-22
feat(client): improved default config paths

### LINKED ISSUE #534: Change default location(s) of Walrus configuration?
We currently look for the Walrus client config at `./client_config.yaml` and `~/.walrus/client_config.yaml`. However, arguably it would be better to use `$XDG_CONFIG_HOME/walrus/client_config.yaml` (with fallback `~/.config/walrus/client_config.yaml`) instead (or in addition), see the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html). Also, we could look for `.yml` in addition to `.yaml`.

If we decide to change this, we need to adjust the docs, examples, and CI workflows in the walrus-docs repository.

mlegner:
Look for the client configuration file also in
`$XDG_CONFIG_HOME/walrus/` (if the variable is set) and `~/.config/walrus/`.

Also support `client_config.yml` in addition to `client_config.yaml`.

Closes #534

### DIFF
--- a/crates/walrus-core/src/lib.rs
+++ b/crates/walrus-core/src/lib.rs
@@ -67,7 +67,7 @@ pub type Epoch = u64;
 // Schema definition for the type alias used in OpenAPI schemas.
 #[derive(Debug)]
 #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
-#[schema(as = Epoch)]
+#[cfg_attr(feature = "utoipa", schema(as = Epoch))]
 #[allow(dead_code)]
 pub struct EpochSchema(u64);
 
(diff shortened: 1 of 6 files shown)


crates/walrus-core/src/lib.rs
```diff
@@ -67,7 +67,7 @@ pub type Epoch = u64;
 // Schema definition for the type alias used in OpenAPI schemas.
 #[derive(Debug)]
 #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
-#[schema(as = Epoch)]
+#[cfg_attr(feature = "utoipa", schema(as = Epoch))]
```
giac-mysten, line 70: Question: This was a bug, correct?

mlegner: Yes, that's correct. 😇

Issue #259, opened 2024-04-12
Major Epic: Implement Shard recovery


gdanezis:
In same epoch (write path) (Devnet): 
- [x] when a blob ID is certified download missing slivers to node from others. (#279)

Across epochs (Testnet): 
- [x] Get a full list of live certified blob IDs from past events.
- [x] Download all missing metadata (if necessary) and slivers. 
- [x] Make available sync done signed message.
- [x] Post sync done message on chain.

PR #611, opened 2024-07-23
feat(storage node): create sync_shard storage node endpoint

halfprice:
The main goal of this PR is to create a new `sync_shard` api in storage node with proper authentication.

The authorization is done via including the sender's public key in the authorization header, and signs the SyncShardRequest. Only storage nodes that are part of the walrus storage nodes (currently, only the committee members) can send sync shard request from other storage nodes.

The return value of the API is currently set to nothing, and will be implemented in the follow up PR.

Once this PR is in, the sender and receiver logic can be implemented in parallel next.

#259

### DIFF
--- a/crates/walrus-core/src/messages.rs
+++ b/crates/walrus-core/src/messages.rs
@@ -79,12 +79,6 @@ impl<T> SignedMessage<T> {
 }
 
 impl<T> SignedMessage<T> {
-    /// Verifies the signature on this message under `public_key`.
-    pub fn verify_signature(&self, public_key: &PublicKey) -> Result<(), MessageVerificationError> {
-        public_key.verify(&self.serialized_message, &self.signature)?;
-        Ok(())
-    }
-
     /// Extracts the underlying message and verifies the signature on this message under `public_key`.
     pub fn verify_signature_and_get_message<I>(
         &self,
@@ -94,7 +88,7 @@ impl<T> SignedMessage<T> {
         T: AsRef<ProtocolMessage<I>> + DeserializeOwned,
     {
         let message: T = bcs::from_bytes(&self.serialized_message)?;
-        self.verify_signature(public_key)?;
+        public_key.verify(&self.serialized_message, &self.signature)?;
 
         Ok(message)
     }
--- a/crates/walrus-core/src/messages/sync_shard.rs
+++ b/crates/walrus-core/src/messages/sync_shard.rs
@@ -66,12 +66,12 @@ impl SyncShardRequest {
 pub struct SyncShardMsg(pub(crate) ProtocolMessage<SyncShardRequest>);
 
 impl SyncShardMsg {
-    const INTENT: Intent = Intent::storage(IntentType::INVALID_BLOB_ID_MSG);
+    const INTENT: Intent = Intent::storage(IntentType::SYNC_SHARD_MSG);
 
     /// Creates a new InvalidBlobIdMsg message for the provided blob ID.
     pub fn new(epoch: Epoch, request: SyncShardRequest) -> Self {
         Self(ProtocolMessage {
-            intent: Intent::storage(IntentType::INVALID_BLOB_ID_MSG),
+            intent: Intent::storage(IntentType::SYNC_SHARD_MSG),
             epoch,
             message_contents: request,
         })
--- a/crates/walrus-service/src/committee.rs
+++ b/crates/walrus-service/src/committee.rs
@@ -130,8 +130,9 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
     ) -> InvalidBlobCertificate;
 
     /// Checks if the given public key belongs to a Walrus storage node.
-    /// Note that the node may not be part of the current committee, since
-    /// it can be future or past committee member.
+    /// TODO (#629): once node catching up is implemented, we need to make sure that the node
+    /// may not be part of the current committee (node from past committee in the previous epoch
+    /// or will be come new committee in the future) can still communicate with each other.
     fn is_walrus_storage_node(&self, public_key: &PublicKey) -> bool;
 }
 


crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use serde::{Deserialize, Serialize};
+
+use super::SignedMessage;
+use crate::{BlobId, Epoch, ShardIndex};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SyncShardRequestV1 {
+    /// The index of the shard to sync.
+    shard_index: ShardIndex,
```
jpcsmith, line 14: Question: Is this the id of the shard that needs to be recovered or the ID of the shard from which we are requesting data?

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    /// Whether the shard is for a primary sliver.
+    primary_sliver: bool,
```
jpcsmith: Question: Could you clarify/rephrease for me what `primary_sliver` here signifies?

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,56 @@
...
+    /// The ID of the blob to start syncing from.
+    starting_blob_id: BlobId,
+
+    /// The number of blobs to sync starting from `starting_blob_id`.
+    /// Note that only blobs certified at the moment of epoch change are synced.
+    num_blobs: u64,
+
+    /// The epoch in which the blobs are certified. In the context of
+    /// an epoch change, this is the previous epoch.
+    epoch: Epoch,
```
jpcsmith: Question: Do you mean "epoch up until which blobs were certified"? This currently reads as if we're only interested in the blobs certified in the previous epoch, and are not interested in blobs certified before that.

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,56 @@
...
+
+    /// Whether the shard is for a primary sliver.
+    primary_sliver: bool,
+
+    /// The ID of the blob to start syncing from.
+    starting_blob_id: BlobId,
+
+    /// The number of blobs to sync starting from `starting_blob_id`.
+    /// Note that only blobs certified at the moment of epoch change are synced.
+    num_blobs: u64,
```
jpcsmith: Question: How do we envision this count being used?

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,56 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SyncShardRequestV1 {
+    /// The index of the shard to sync.
+    shard_index: ShardIndex,
+
+    /// Whether the shard is for a primary sliver.
+    primary_sliver: bool,
+
+    /// The ID of the blob to start syncing from.
+    starting_blob_id: BlobId,
```
jpcsmith, line 22: Question: Is the idea here that if both the requester and server have reached the epoch change event, that their view of the live blobs will be the same (at least until the server has started inserting entries for the next epoch)? 

Remark: This may add a bit of complexity and I am not certain that it saves much over simply listing the blob IDs or blob ID prefixes.

crates/walrus-core/src/messages/sync_shard.rs
halfprice: It is the id of the shard needs to be entirely transferred from one storage node to another, due to the change of shard-storage-node assignment.

crates/walrus-core/src/messages/sync_shard.rs
halfprice: Comments updated. PTAL

crates/walrus-core/src/messages/sync_shard.rs
halfprice: Correct. Comments updated.

crates/walrus-core/src/messages/sync_shard.rs
halfprice: Discussed this in person. Also updated the comments.

crates/walrus-core/src/messages/sync_shard.rs
halfprice: Discussed in person.

crates/walrus-core/src/messages.rs
```diff
@@ -77,7 +80,14 @@ impl<T> SignedMessage<T> {
 
 impl<T> SignedMessage<T> {
     /// Verifies the signature on this message under `public_key`.
-    pub fn verify_signature<I>(
+    pub fn verify_signature(&self, public_key: &PublicKey) -> Result<(), MessageVerificationError> {
```
jpcsmith: Suggestion: Call this method in `verify_signature_get_message` below.

crates/walrus-core/src/messages.rs
```diff
@@ -77,7 +80,14 @@ impl<T> SignedMessage<T> {
...
+        public_key.verify(&self.serialized_message, &self.signature)?;
+        Ok(())
+    }
+
+    /// Verifies the signature on this message under `public_key`, and extracts the underlying message
+    /// upon successful verification.
+    pub fn verify_signature_get_message<I>(
```
jpcsmith: Nit: Rename to `verify_signature_and_get_message` or `verify_signature_and_decode`. It may also be worth adding to the documentation that this is equivalent to checking that the message decodes before attempting to verify the signature.

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use serde::{Deserialize, Serialize};
+
+use super::SignedMessage;
+use crate::{BlobId, Epoch, ShardIndex};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SyncShardRequestV1 {
```
jpcsmith: Suggestion: Add a docstring describing the purpose of the request.

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// The shard index that is requested to be synced.
+    shard_index: ShardIndex,
+
+    /// Whether the sync is for the primary sliver or the secondary sliver in the shard.
+    /// Note that storage node stores primary and secondary slivers in separate
+    /// RocksDB column family, so it's more efficient to transfer them separately.
+    primary_sliver: bool,
```
jpcsmith, line 19: Important: This can be removed as there is only a single column family. Both DBMaps are views into the same column family. Keys are (bool, blob_id).

crates/walrus-core/src/messages/sync_shard.rs
jpcsmith: Thank you for the discussion, I agree that your approach is cleaner than my suggestion since the goal is to implement an RPC for an iterator over the table.

crates/walrus-sdk/src/client.rs
```diff
@@ -466,6 +480,44 @@ impl Client {
...
     }
 
+    /// Syncs a shard from the storage node.
+    #[tracing::instrument(
+        skip_all,
+        fields(
+            walrus.sync.shard_index = %shard_index,
+            walrus.sync.epoch = epoch,
+            walrus.sync.starting_blob_id = %starting_blob_id,
+            walrus.sync.num_blobs = %num_blobs,
```
jpcsmith:
```suggestion
            walrus.shard_index = %shard_index,
            walrus.epoch = epoch,
            walrus.blob_id = %starting_blob_id,
            walrus.sync.num_blobs = %num_blobs,
```

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,62 @@
...
+    /// The ID of the blob to start syncing from.
+    starting_blob_id: BlobId,
+
+    /// The number of blobs to sync starting from `starting_blob_id`.
+    /// Since the blobs are stored in RocksDB ordered by the key, the sync basically
+    /// scans the RocksDB from `starting_blob_id` and reads `num_blobs` blobs for
+    /// efficient scanning.
+    ///
+    /// Note that only blobs certified at the moment of epoch change are synced.
+    num_blobs: u64,
```
jpcsmith: Suggestion: Rename this to `sliver_count` as, as mentioned above, the iterator is over primary and secondary slivers.

crates/walrus-sdk/src/client.rs
```diff
@@ -44,6 +53,7 @@ const RECOVERY_URL_TEMPLATE: &str =
 const INCONSISTENCY_PROOF_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/inconsistent/:sliver_type";
 const STATUS_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/status";
 const HEALTH_URL_TEMPLATE: &str = "/v1/health";
+const SYNC_SHARD_TEMPLATE: &str = "/v1/sync_shard";
```
jpcsmith: Suggestion: How about `/v1/migrate/shard`?

crates/walrus-service/src/committee.rs
```diff
@@ -128,6 +128,11 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
         inconsistency_proof: &InconsistencyProofEnum,
         n_shards: NonZeroU16,
     ) -> InvalidBlobCertificate;
+
+    /// Checks if the given public key belongs to a Walrus storage node.
+    /// Note that the node may not be part of the current committee, since
+    /// it can be future or past committee member.
+    fn is_walrus_storage_node(&self, public_key: &PublicKey) -> bool;
```
jpcsmith: Suggestion: Clarify that this checks if the key belongs to a node of the current committee.

crates/walrus-core/src/messages.rs
jpcsmith: Important: The constraint `T: AsRef<ProtocolMessage<I>> + DeserializeOwned` has security implications. All messages signed by the public key should be `ProtocolMessage`s, as this adds the required headers to ensure that a signed messaged cannot be reused for a different purpose. 

Instead of adding this function, ensure that your message is a proper `ProtocolMessage`.

crates/walrus-service/src/node.rs
```diff
@@ -833,6 +850,23 @@ impl ServiceState for StorageNodeInner {
...
+        &self,
+        public_key: PublicKey,
+        signed_request: SignedMessage<SyncShardRequest>,
+    ) -> Result<(), SyncShardError> {
+        if !self.committee_service.is_walrus_storage_node(&public_key) {
+            return Err(SyncShardError::Unauthorized);
+        }
+        signed_request.verify_signature(&public_key)?;
+        let SyncShardRequest::V1(request) =
+            bcs::from_bytes(&signed_request.serialized_message).unwrap();
```
jpcsmith: Suggestion: Use the function that decodes and verifies. Decoding is cheaper than verifying and should be done first, so that we don't spent cycles verifying messages which cannot decode.

crates/walrus-core/src/messages.rs
halfprice: Done

halfprice: I made the suggested change, but since this message won't be used as message submitted on chain, I just want to confirm it is ok to use ProtocolMessage to just communicate between nodes?

crates/walrus-core/src/messages.rs
halfprice: Done.

crates/walrus-core/src/messages/sync_shard.rs
halfprice: Done.

crates/walrus-core/src/messages/sync_shard.rs
halfprice: As discussed offline, we probably want to separate the two slivers into two column families. So keep it for now.

crates/walrus-core/src/messages/sync_shard.rs
halfprice: Done.

crates/walrus-sdk/src/client.rs
halfprice: Sounds good.

crates/walrus-sdk/src/client.rs
halfprice: Done.

crates/walrus-service/src/committee.rs
halfprice: We discussed in the doc that the sync api can be used not only amount the current committee, but any registered storage nodes, so that old committee members can send data to new committee, and we also want to allow future committee to be able to sync data early before it becomes a committee.

crates/walrus-service/src/node.rs

crates/walrus-service/src/committee.rs
jpcsmith: Okay, but as it stands, the documentation and the implementation does not agree. Could we add a TODO to the implementation to make it clear that the implementation is currently incomplete?

Suggestion: Could you also clarify in the docstring the actual bounds: how far in advance/how far in the past will be checked? epoch i-1 and epoch i+1?

crates/walrus-core/src/messages.rs
jpcsmith: It is. The ProtocolMessage just adds framing that prevents the attack mentioned above.

crates/walrus-core/src/messages.rs
```diff
@@ -77,15 +80,21 @@ impl<T> SignedMessage<T> {
 
 impl<T> SignedMessage<T> {
     /// Verifies the signature on this message under `public_key`.
-    pub fn verify_signature<I>(
+    pub fn verify_signature(&self, public_key: &PublicKey) -> Result<(), MessageVerificationError> {
```
jpcsmith: Suggestion: Revert the changes here since the message now inherits from ProtocolMessage.

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,102 @@
...
+    }
+}
+
+/// A message stating that a Blob Id is invalid.
+#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(try_from = "ProtocolMessage<SyncShardRequest>")]
+pub struct SyncShardMsg(pub(crate) ProtocolMessage<SyncShardRequest>);
+
+impl SyncShardMsg {
+    const INTENT: Intent = Intent::storage(IntentType::INVALID_BLOB_ID_MSG);
```
jpcsmith:
```suggestion
    const INTENT: Intent = Intent::storage(IntentType::SYNC_SHARD_MSG);
```

crates/walrus-core/src/messages/sync_shard.rs
```diff
@@ -0,0 +1,102 @@
...
+
+    /// Creates a new InvalidBlobIdMsg message for the provided blob ID.
+    pub fn new(epoch: Epoch, request: SyncShardRequest) -> Self {
+        Self(ProtocolMessage {
+            intent: Intent::storage(IntentType::INVALID_BLOB_ID_MSG),
```
jpcsmith:
```suggestion
            intent: Intent::storage(IntentType::SYNC_SHARD_MSG),
```

crates/walrus-core/src/messages/sync_shard.rs
halfprice: Good catch!

crates/walrus-core/src/messages/sync_shard.rs
halfprice: Done

crates/walrus-core/src/messages.rs
halfprice: I removed verify_signature and kept the new function name. There was also an issue with the old function as it should return Result<T,..> instead of Result<ProtocolMessage<T>, ..>

crates/walrus-service/src/committee.rs
halfprice: Added todo #629

PR #613, opened 2024-07-23
feat(client): support 'info' command in JSON mode

### LINKED ISSUE #465: Add the option to call the `info` command in JSON mode as well
Currently the `info` command only works for CLI mode.

mlegner:
Closes #465

### DIFF
--- a/crates/walrus-service/src/cli_utils/cli_output.rs
+++ b/crates/walrus-service/src/cli_utils/cli_output.rs
@@ -9,7 +9,6 @@ use indoc::printdoc;
 use prettytable::{format, row, Table};
 use serde::Serialize;
 use walrus_sdk::api::BlobStatus;
-use walrus_sui::utils::price_for_unencoded_length;
 
 use crate::{
     cli_utils::{
@@ -26,6 +25,7 @@ use crate::{
         BlobStatusOutput,
         BlobStoreResult,
         DryRunOutput,
+        ExampleBlobInfo,
         InfoDevOutput,
         InfoOutput,
         ReadOutput,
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/cli_utils/cli_output.rs
```diff
@@ -1,46 +1,62 @@
...
 pub trait CliOutput: Serialize {
-    /// Returns the output as a human-readable string for the CLI.
-    fn cli_output(&self) -> String;
+    /// Writes the output to stdout as a human-readable string for the CLI.
+    fn print_cli_output(&self);
 
-    /// Returns the output as a string formatted depending on the output mode.
-    fn output_string(&self, json: bool) -> Result<String> {
+    /// Writes the output to stdout formatted depending on the output mode.
+    fn print_output(&self, json: bool) -> Result<()> {
```
giac-mysten, line 41: Question: printing directly is an instance of YAGNI?

crates/walrus-service/src/client/responses.rs
```diff
@@ -144,3 +162,137 @@ pub struct BlobStatusOutput {
     /// The blob's status.
     pub status: BlobStatus,
 }
+
+/// The output of the `info` command.
+#[serde_as]
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct InfoOutput {
```
giac-mysten, line 172: Remark: IIUC, the information content of the JSON command is lower than the text output (some fields are missing). I think it's perfect -- only the examples are missing.

Question: should we write this in the CLI help string?

crates/walrus-service/src/cli_utils/cli_output.rs
mlegner: Essentially three different reasons:
- YAGNI. 😁 
- Potentially more efficient as we don't create strings that we immediately write.
- Allows using formatting on the printed table, which is not possible when converting to a string.

crates/walrus-service/src/client/responses.rs
mlegner: Yes, I can mention this in the help string. Do you think I should add the examples to the JSON output as well?

mlegner: I actually added the example blobs to the `InfoOutput` in c9c3e9e9a2f52651f4d90943d0003f36acf61bda.

giac-mysten: Perfect. Looks great!

PR #614, opened 2024-07-23
feat: support empty blobs

### LINKED ISSUE #565: (How) Should we support empty blobs on Walrus?

mlegner:
Closes #565

### DIFF
--- a/crates/walrus-core/src/encoding/utils.rs
+++ b/crates/walrus-core/src/encoding/utils.rs
@@ -2,55 +2,39 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use alloc::vec::Vec;
-use core::num::{NonZeroU16, NonZeroU32, NonZeroU64};
+use core::num::{NonZeroU16, NonZeroU32};
 
 use raptorq::{EncodingPacket, ObjectTransmissionInformation};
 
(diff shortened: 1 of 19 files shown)


crates/walrus-core/src/encoding/utils.rs
```diff
@@ -2,55 +2,39 @@
...
 #[inline]
-pub fn compute_symbol_size_from_nonzero(
-    data_length: NonZeroU64,
+pub fn compute_symbol_size(
+    data_length: u64,
     n_symbols: NonZeroU32,
 ) -> Result<NonZeroU16, DataTooLargeError> {
-    u16::try_from((data_length.get() - 1) / u64::from(n_symbols.get()) + 1)
+    // Use a 1-byte symbol size for the empty blob.
+    let data_length = data_length.max(1);
```
giac-mysten, line 23: Remark: A-HA!

PR #615, opened 2024-07-24
chore: refactor skeleton to separate system and staking objects

karlwuest:


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -37,6 +41,49 @@ module walrus::staking_inner {
         pool_id
     }
 
+        public(package) fun register_candidate(
```
damirka: indentation is off

PR #616, opened 2024-07-24
feat: checkpoint based event processing

sadhansood:
This PR adds the following new functionalities:
1. Use checkpoints to read walrus events
2. Verify checkpoints using chain of reference and committee
3. Store events in a local store
4. Prune events once committed
5. Store local state on disk for restart functionality

What is not added yet:
1. Ability to catch up from walrus events stored on walrus
2. Ability to start writing walrus event blobs
3. Switching the current event provider interface to start using new framework
4. Tests

### DIFF
--- a/crates/walrus-event/src/checkpoint_processor.rs
+++ b/crates/walrus-event/src/checkpoint_processor.rs
@@ -1,0 +1,543 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use std::{path::PathBuf, sync::Arc, time::Duration};
+
+use anyhow::{anyhow, bail, Result};
+use async_trait::async_trait;
+use move_core_types::annotated_value::{MoveDatatypeLayout, MoveTypeLayout};
+use rocksdb::Options;
+use serde::{Deserialize, Serialize};
+use sui_config::genesis::Genesis;
+use sui_package_resolver::{error::Error as PackageResolverError, Package, PackageStore, Resolver};
+use sui_rest_api::Client;
+use sui_sdk::rpc_types::SuiEvent;
+use sui_storage::verify_checkpoint_with_committee;
+use sui_types::{
+    base_types::ObjectID,
+    committee::Committee,
+    full_checkpoint_content::CheckpointData,
+    messages_checkpoint::{CheckpointSequenceNumber, TrustedCheckpoint, VerifiedCheckpoint},
+    object::Object,
+};
+use tokio::time::{sleep, Instant};
+use tokio_util::sync::CancellationToken;
+use typed_store::{
+    rocks,
+    rocks::{errors::typed_store_err_from_rocks_err, DBMap, MetricConf, ReadWriteOptions},
+    Map,
+};
+
+use crate::EventConfig;
+
+#[allow(dead_code)]
+const CHECKPOINT_STORE: &str = "checkpoint_store";
+#[allow(dead_code)]
+const WALRUS_PACKAGE_STORE: &str = "walrus_package_store";
+#[allow(dead_code)]
+const COMMITTEE_STORE: &str = "committee_store";
+#[allow(dead_code)]
+const EVENT_STORE: &str = "event_store";
+const MAX_TIMEOUT: Duration = Duration::from_secs(60);
+const RETRY_DELAY: Duration = Duration::from_secs(5);
+
+#[derive(Eq, PartialEq, Clone, Debug, Serialize, Deserialize)]
+pub struct CheckpointOrderedEventID {
+    sequence: CheckpointSequenceNumber,
+    counter: u64,
+}
+
+impl CheckpointOrderedEventID {
+    fn new(sequence: CheckpointSequenceNumber, counter: u64) -> Self {
+        CheckpointOrderedEventID { sequence, counter }
+    }
+}
+
+#[derive(Eq, PartialEq, Clone, Debug, Serialize, Deserialize)]
+pub struct CheckpointSuiEvent {
+    pub serialized_event: Vec<u8>,
+    pub checkpoint_sequence_number: CheckpointOrderedEventID,
+}
+
+#[derive(Clone)]
+pub struct CheckpointProcessor {
+    /// Full node REST client.
+    client: Client,
+    /// The address of the Walrus system package.
+    system_pkg_id: ObjectID,
+    /// Event to read next.
+    event_store_next_cursor: CheckpointOrderedEventID,
+    /// Store which only stores the latest checkpoint.
+    checkpoint_store: DBMap<(), TrustedCheckpoint>,
+    /// Store which only stores the latest Walrus package.
+    walrus_package_store: DBMap<(), Object>,
+    /// Store which only stores the latest Sui committee.
+    committee_store: DBMap<(), Committee>,
+    /// Store which only stores all uncommitted events.
+    event_store: DBMap<CheckpointOrderedEventID, CheckpointSuiEvent>,
+}
+#[allow(dead_code)]
+impl CheckpointProcessor {
+    async fn poll(&mut self, timeout: Duration) -> Result<Vec<CheckpointSuiEvent>> {
+        let mut events = vec![];
+        let mut iter = self.event_store.unbounded_iter();
+        iter = iter.skip_to(&self.event_store_next_cursor)?;
+        let start_time = std::time::Instant::now();
+        while start_time.elapsed() < timeout {
+            if let Some((checkpoint_event_id, event)) = iter.next() {
+                events.push(event.clone());
+                self.event_store_next_cursor = CheckpointOrderedEventID::new(
+                    checkpoint_event_id.sequence,
+                    checkpoint_event_id.counter.saturating_add(1),
+                );
+            } else {
+                break;
+            }
+        }
+        Ok(events)
+    }
+
+    async fn commit(&mut self, committed_event_id: CheckpointOrderedEventID) -> Result<()> {
+        let iter = self.event_store.unbounded_iter();
+        let rev_iter = iter.skip_to(&committed_event_id)?.reverse();
+        let mut write_batch = self.event_store.batch();
+        for (event_id, _) in rev_iter {
+            write_batch
+                .delete_batch(&self.event_store, std::iter::once(event_id))
+                .map_err(|e| anyhow!("Failed to remove event from event store: {}", e))?;
+        }
+        write_batch
+            .write()
+            .map_err(|e| anyhow!("Failed to remove event from event store: {}", e))?;
+        Ok(())
+    }
+
+    pub async fn start(&self) -> Result<CancellationToken> {
+        let cancellation_token = CancellationToken::new();
+        tokio::task::spawn(Self::start_tailing_checkpoints(
+            self.clone(),
+            cancellation_token.clone(),
+        ));
+        Ok(cancellation_token)
+    }
+
+    async fn get_full_checkpoint(
+        &self,
+        sequence_number: CheckpointSequenceNumber,
+    ) -> Result<CheckpointData> {
+        let start_time = Instant::now();
+
+        loop {
+            match self.client.get_full_checkpoint(sequence_number).await {
+                Ok(checkpoint) => return Ok(checkpoint),
+                Err(inner) => {
+                    let elapsed = start_time.elapsed();
+                    if elapsed >= MAX_TIMEOUT {
+                        return Err(inner);
+                    }
+                    let delay = RETRY_DELAY.min(MAX_TIMEOUT - elapsed);
+                    sleep(delay).await;
+                }
+            }
+        }
+    }
+
+    async fn start_tailing_checkpoints(self, cancel_token: CancellationToken) -> Result<()> {
+        while !cancel_token.is_cancelled() {
+            let Some(prev_checkpoint) = self.checkpoint_store.get(&())? else {
+                bail!("No checkpoint found in the checkpoint store");
+            };
+            let next_checkpoint = prev_checkpoint.inner().sequence_number().saturating_add(1);
+            let Ok(checkpoint) = self.get_full_checkpoint(next_checkpoint).await else {
+                // TODO:
+                // 1. Read walrus events from event blobs s we've fallen behind the first unpruned
+                // checkpoint on the fullnode or the fullnode doesn't have the new checkpoint yet
+                // 2. Reset the committee and previous checkpoint using a light client
+                // based approach so we can continue processing events from full node
+                // 3. Reset the walrus package store with the latest walrus package object from fullnode
+                bail!("Failed to get checkpoint {}", next_checkpoint);
+            };
+            let Some(committee) = self.committee_store.get(&())? else {
+                bail!("No committee found in the committee store");
+            };
+            let verified_checkpoint = verify_checkpoint_with_committee(
+                Arc::new(committee.clone()),
+                &VerifiedCheckpoint::new_from_verified(prev_checkpoint.into_inner()),
+                checkpoint.checkpoint_summary.clone(),
+            )
+            .map_err(|checkpoint| {
+                anyhow!(
+                    "Failed to verify sui checkpoint: {}",
+                    checkpoint.sequence_number
+                )
+            })?;
+            let resolver = Resolver::new(self.clone());
+            let mut write_batch = self.event_store.batch();
+            for tx in checkpoint.transactions.into_iter() {
+                for object in tx.output_objects.into_iter() {
+                    if object.id() == self.system_pkg_id {
+                        write_batch
+                            .insert_batch(&self.walrus_package_store, std::iter::once(((), object)))
+                            .map_err(|e| {
+                                anyhow!("Failed to insert object into walrus package store: {}", e)
+                            })?;
+                    }
+                }
+                let tx_events = tx.events.unwrap();
+                for (seq, tx_event) in tx_events
+                    .data
+                    .into_iter()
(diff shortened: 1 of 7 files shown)


crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    pub(crate) async fn start(&self) -> Result<oneshot::Sender<()>> {
+        let (sender, receiver) = oneshot::channel();
+        tokio::task::spawn(Self::start_tailing_checkpoints(self.clone(), receiver));
+        Ok(sender)
+    }
+
+    async fn start_tailing_checkpoints(self, mut receiver: oneshot::Receiver<()>) -> Result<()> {
+        while receiver.try_recv().is_err() {
```
karlwuest: Question: Do we want to continue this loop even if the Sender gets dropped/the channel is closed? 
Suggestion: Use a cancellation token or match on the error type and return an Error if `try_recv` returns a `TryRecvError::Closed`.

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+            false,
+        )?;
+        let event_store = DBMap::reopen(
+            &database,
+            Some(EVENT_STORE),
+            &ReadWriteOptions::default(),
+            false,
+        )?;
+        if !checkpoint_store.is_empty() {
+            // This is a fresh start as there is no prev disk state
```
karlwuest: Question: Shouldn't the fresh start happen if the store is empty?

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+            committee_store.schedule_delete_all()?;
+            event_store.schedule_delete_all()?;
+            walrus_package_store.schedule_delete_all()?;
+            let genesis = Genesis::load(config.sui_genesis_path)?;
+            let genesis_committee = genesis.committee()?;
+            let checkpoint = genesis.checkpoint();
+            committee_store.insert(&(), &genesis_committee)?;
+            checkpoint_store.insert(&(), checkpoint.serializable_ref())?;
+        }
+        let iter = event_store.unbounded_iter();
```
karlwuest: Suggestion: Use a more descriptive name, e.g. `event_iter`

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+                    .collect();
+                let committee = Committee::new(
+                    checkpoint.checkpoint_summary.epoch().saturating_add(1),
+                    next_committee,
+                );
+                self.committee_store.insert(&(), &committee)?;
+            }
+        }
+        Ok(())
+    }
```
karlwuest:
```suggestion
    }
    
```

crates/walrus-service/src/config.rs
```diff
@@ -88,6 +88,21 @@ impl Default for BlobRecoveryConfig {
     }
 }
 
+/// Configuration for event processing.
+
+#[serde_with::serde_as]
```
mlegner: Nit: Where do we need this?
```suggestion
```

crates/walrus-service/src/config.rs
```diff
@@ -88,6 +88,21 @@ impl Default for BlobRecoveryConfig {
...
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct EventConfig {
+    /// The rest url of the fullnode
+    pub rest_url: String,
+    /// The url of the checkpoint client to connect to.
+    pub path: PathBuf,
+    /// The path to the sui genesis file
+    pub sui_genesis_path: PathBuf,
+    /// ObjectID of the walrus system object
+    pub system_object_id: ObjectID,
```
mlegner: Nit: We generally use standard capitalization and punctuation in docstrings. This also applies to other struct definitions.
```suggestion
    /// The REST URL of the full node.
    pub rest_url: String,
    /// The URL of the checkpoint client to connect to.
    pub path: PathBuf,
    /// The path to the Sui genesis file.
    pub sui_genesis_path: PathBuf,
    /// ObjectID of the Walrus system object.
    pub system_object_id: ObjectID,
```

Also, question: Does the "REST URL" refer to the "standard" RPC URL?

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+
+#[derive(Clone, Serialize, Deserialize)]
+pub struct CheckpointSuiEvent {
+    pub event: SuiEvent,
+    pub checkpoint_sequence_number: CheckpointOrderedEventID,
+}
+
+#[derive(Clone)]
+pub struct CheckpointProcessor {
+    // Full node rest client
```
mlegner: Nit: These could be docstrings (`///` instead of `//`).

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+impl CheckpointProcessor {
+    async fn poll(&mut self, timeout: Duration) -> Result<Vec<CheckpointSuiEvent>> {
+        let mut events = vec![];
+        let mut iter = self.event_store.unbounded_iter();
+        iter = iter.skip_to(&self.event_store_next_cursor)?;
+        let start_time = std::time::Instant::now();
+        while start_time.elapsed() < timeout {
+            if let Some((checkpoint_event_id, event)) = iter.next() {
+                events.push(event.clone());
+                self.event_store_next_cursor = checkpoint_event_id;
```
mlegner: Question/important: Doesn't this mean that we handle that specific event twice if this is the last one processed in this poll? Once now and once when we continue from that event in the next poll?

Related to that, could you add some tests to make sure this works as intended?

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+            } else {
+                break;
+            }
+        }
+        Ok(events)
```
mlegner: Nit: Probably a matter of taste, but to me looping over the iterator and breaking for the timeout feels more natural than the other way round:
```suggestion
        let mut events = vec![];
        let iter = self
            .event_store
            .unbounded_iter()
            .skip_to(&self.event_store_next_cursor)?;
        let start_time = std::time::Instant::now();

        for (checkpoint_event_id, event) in iter {
            events.push(event);
            self.event_store_next_cursor = checkpoint_event_id;
            if start_time.elapsed() >= timeout {
                break;
            }
        }

        Ok(events)
```

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+        let checkpoint_processor = CheckpointProcessor {
+            client,
+            walrus_package_store,
+            checkpoint_store,
+            committee_store,
+            event_store,
+            system_pkg_id: config.system_object_id,
+            event_store_next_cursor: cursor,
+        };
+        Ok(checkpoint_processor)
```
mlegner: Nit: This could be shortened:
```suggestion
        Ok(CheckpointProcessor {
            client,
            walrus_package_store,
            checkpoint_store,
            committee_store,
            event_store,
            system_pkg_id: config.system_object_id,
            event_store_next_cursor: cursor,
        })
```

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+    pub event: SuiEvent,
+    pub checkpoint_sequence_number: CheckpointOrderedEventID,
+}
+
+#[derive(Clone)]
+pub struct CheckpointProcessor {
+    // Full node rest client
+    client: Client,
+    // The address of the walrus system package
+    system_pkg_id: ObjectID,
```
mlegner: Important: Is this the object ID of the Walrus _package_ or the _system object_? The comment and uses of this field seem to imply the _package_, but the `new` function below assigns the _system object_.

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+
+    async fn start_tailing_checkpoints(self, mut receiver: oneshot::Receiver<()>) -> Result<()> {
+        while receiver.try_recv().is_err() {
+            let Some(prev_checkpoint) = self.checkpoint_store.get(&())? else {
+                bail!("No checkpoint found in the checkpoint store");
+            };
+            let next_checkpoint = prev_checkpoint.inner().sequence_number().saturating_add(1);
+            let Ok(checkpoint) = self.client.get_full_checkpoint(next_checkpoint).await else {
+                // TODO:
```
mlegner: Suggestion: Can you create an issue for this follow-up work?

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+use std::{sync::Arc, time::Duration};
+
+use anyhow::{anyhow, bail, Result};
```
mlegner: Suggestion: Given that this is an internal package, we may want to programmatically process returned errors. In this case an explicit error enum type (using `thiserror`) would be helpful. This can also be done in a follow-up PR.

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+                        ))
+                        .await?;
+                    let move_datatype_layout = match move_type_layout {
+                        MoveTypeLayout::Struct(s) => Some(MoveDatatypeLayout::Struct(s)),
+                        MoveTypeLayout::Enum(e) => Some(MoveDatatypeLayout::Enum(e)),
+                        _ => None,
+                    }
+                    .ok_or(anyhow!("Failed to get move datatype layout"))?;
+                    let sui_event = SuiEvent::try_from(
+                        tx_event.clone(),
```
mlegner: Remark: There are quite a few uses of `clone()` in this function. Some of them could be avoided by just using `into_iter()` instead of `iter()`.

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+        Ok(sender)
+    }
+
+    async fn start_tailing_checkpoints(self, mut receiver: oneshot::Receiver<()>) -> Result<()> {
+        while receiver.try_recv().is_err() {
+            let Some(prev_checkpoint) = self.checkpoint_store.get(&())? else {
+                bail!("No checkpoint found in the checkpoint store");
+            };
+            let next_checkpoint = prev_checkpoint.inner().sequence_number().saturating_add(1);
+            let Ok(checkpoint) = self.client.get_full_checkpoint(next_checkpoint).await else {
```
mlegner: Question: What happens if that checkpoint does not exist yet?

crates/walrus-service/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,295 @@
...
+            } else {
+                break;
+            }
+        }
+        Ok(events)
+    }
+
+    async fn commit(&mut self, committed_event_id: CheckpointOrderedEventID) -> Result<()> {
+        let iter = self.event_store.unbounded_iter();
+        let rev_iter = iter.skip_to(&committed_event_id)?.reverse();
```
mlegner: Question: Why do we remove events in reverse order?

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: fixed

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: fixed

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: done

crates/walrus-service/src/config.rs
sadhansood: removed

crates/walrus-service/src/config.rs
sadhansood: no, it is actually a different endpoint which returns full checkpoint data

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: fixed

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: test added, also you were right and hence fixed

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: ah yeah it should be the system package id, fixed the name

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: ok, will do

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: You are absolutely right, i will do this in a follow up PR

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: done, removed some of the clones

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: If the checkpoint doesn't exist even after several retries (default set to 1 minute), that means this full node is stuck and in that scenario as well we will start the catch up from event blobs

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: It was for the ease of iterating until there is no value returned by the iter, but actually i changed the code to delete all elements in a single atomic batch

crates/walrus-service/src/checkpoint_processor.rs
mlegner: Thanks for the added tests. Maybe I missed it, but I didn't see where you test the case where we poll multiple times and make sure we don't get any overlap.

sadhansood: ah ok, let me add the test for that.

crates/walrus-service/src/checkpoint_processor.rs
mlegner: Can you create an issue for this?

crates/walrus-service/src/checkpoint_processor.rs
mlegner: I believe the `tx_event.clone()` could also be changed to `tx_event` if you change `tx.events.as_ref().unwrap()` to `tx.events.unwrap()` and use `tx_events.data.into_iter()`.

crates/walrus-service/src/checkpoint_processor.rs
mlegner: Ah, cool. I'm wondering about the performance of first finding the `committed_event_id` and then going through the events backwards adding them to the batch. Wouldn't it be faster to go in order until hitting the `committed_event_id`? Or does it not matter because the `skip_to` function is very efficient? (I don't know the details of how these database operations work.)

crates/walrus-event/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,537 @@
...
+                    checkpoint.checkpoint_summary.epoch().saturating_add(1),
+                    next_committee,
+                );
+                write_batch
+                    .insert_batch(&self.committee_store, std::iter::once(((), committee)))
+                    .map_err(|e| {
+                        anyhow!("Failed to insert committee into committee store: {}", e)
+                    })?;
+            }
+            write_batch.write()?;
```
halfprice, line 248: Should the `checkpoint_store` also be updated here before the writes?

crates/walrus-event/src/checkpoint_processor.rs
```diff
@@ -0,0 +1,537 @@
...
+            }
+        }
+        Ok(events)
+    }
+
+    async fn commit(&mut self, committed_event_id: CheckpointOrderedEventID) -> Result<()> {
+        let iter = self.event_store.unbounded_iter();
+        let rev_iter = iter.skip_to(&committed_event_id)?.reverse();
+        let mut write_batch = self.event_store.batch();
+        for (event_id, _) in rev_iter {
```
halfprice, line 103: Nit: felt like a candidate for delete range for better performance? (no need to address it in this PR)

crates/walrus-event/src/checkpoint_processor.rs
sadhansood: great catch, thanks @halfprice - fixed it

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: done, thanks!

crates/walrus-service/src/checkpoint_processor.rs
sadhansood: Actually, looping in the reverse is slightly more efficient because we don;t have to compare every iterated element with `committed_event_id ` and makes it possible to write the for loop like:
```
for (event_id, _) in rev_iter {
}
```
It might be possible to achieve the same going forward with a bounded_iter (where it upfront limit how far the iterator can go i.e. `committed_event_id` but i will check that api and can make this change later)

PR #620, opened 2024-07-25
fix(client): make blob status check more robust

### LINKED ISSUE #518: Make blob-status check more robust in the client
See https://github.com/MystenLabs/walrus/pull/514#issuecomment-2161098687 for context:

> [There is a problem] if no valid status is obtained before the timeout.
>
> I guess increasing the timeout is a short-term fix, but a more fundamental solution would be to wait for either a timeout or a threshold of responses as discussed https://github.com/MystenLabs/walrus/pull/475#discussion_r1623970199.

mlegner:
Add a function to `WeightedFutures` that combines waiting for a timeout and a threshold and use this in the status check.

Also add some tracing and improve default values for the communication config.

Closes #518

### DIFF
--- a/crates/walrus-service/client_config_example.yaml
+++ b/crates/walrus-service/client_config_example.yaml
@@ -5,7 +5,6 @@
 system_object: 0xd6f0cacdb8b365cdda24ab954a32ebc86f8b8a1707322b4cc0ba54dcbbbc5772
 # wallet_config: null
 # communication_config:
-#   max_concurrent_writes: null
 #   max_concurrent_sliver_reads: null
 #   max_concurrent_metadata_reads: null
 #   max_concurrent_status_reads: null
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -81,7 +81,6 @@ pub struct Client<T> {
     storage_price_per_unit_size: u64,
     communication_limits: CommunicationLimits,
     encoding_config: EncodingConfig,
-    global_write_limit: Arc<Semaphore>,
     blocklist: Option<Blocklist>,
 }
 
@@ -113,8 +112,6 @@ impl Client<()> {
         let encoding_config = EncodingConfig::new(committee.n_shards());
         let communication_limits =
             CommunicationLimits::new(&config.communication_config, encoding_config.n_shards());
-        let global_write_limit =
-            Arc::new(Semaphore::new(communication_limits.max_concurrent_writes));
 
         Ok(Self {
             config,
@@ -124,7 +121,6 @@ impl Client<()> {
             storage_price_per_unit_size,
             encoding_config,
             communication_limits,
-            global_write_limit,
             blocklist: None,
         })
     }
@@ -139,7 +135,6 @@ impl Client<()> {
             storage_price_per_unit_size,
             encoding_config,
             communication_limits,
-            global_write_limit,
             blocklist,
         } = self;
         Client::<T> {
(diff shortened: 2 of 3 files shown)


crates/walrus-service/bin/client.rs
```diff
@@ -357,7 +357,7 @@ mod default {
     }
 
     pub(crate) fn status_timeout() -> Duration {
-        Duration::from_secs(1)
+        Duration::from_secs(10)
```
giac-mysten, line 356: Question: curious what prompted this change and PR in general. Was the status command hanging or timing out?

crates/walrus-service/src/client/config.rs
```diff
@@ -274,15 +272,18 @@ pub(crate) mod default {
     use walrus_core::bft;
 
     pub fn max_concurrent_writes(n_shards: NonZeroU16) -> usize {
-        (n_shards.get() - bft::max_n_faulty(n_shards)).into()
+        // No limit as we anyway want to store as many slivers as possible.
+        n_shards.get().into()
     }
```
giac-mysten, line 277: Important/question: this is not the number of writes, but the number of _concurrent_ writes. Don't we want to limit that? This value then goes into the `global_write_limit` semaphore. I would even argue that it should be lowered from the default value we had (`n-f`)?

crates/walrus-service/src/client/utils.rs
```diff
@@ -231,25 +255,63 @@ where
...
+
+/// Represents the reason why the [`WeightedFutures::execute_until`] completed.
+#[derive(Debug, Clone, Copy)]
+pub enum CompletedReason {
+    ThresholdReached,
+    /// Contains the total weight of successful futures.
+    Timeout(usize),
+    /// Contains the total weight of successful futures.
+    FuturesConsumed(usize),
+}
```
giac-mysten, line 291: Question/suggestion: would it make sense to unify both `CompletedReasonWeight` and `...Time`, and use just `CompletedReason`? This would require returning the total weight of successful futures from `execute_all`, but not a big deal. 

I can also do this in a followup if you think it makes sense.

crates/walrus-service/bin/client.rs
mlegner: Before, we had to wait until the timeout was reached if just a single storage node didn't reply, such that it was important not to have a too long timeout. Now we return early as soon as we have a sufficient number of responses, so increasing the timeout is possible to cover corner cases with very slow network without slowing down the common case.

crates/walrus-service/src/client/config.rs
mlegner: Ah, good point. I think the previous value that depended on the number of shards didn't make too much sense as the write futures are per node, not per shard.

I agree that the unlimited value doesn't make too much sense. However, we already limit the number of concurrent writes through the `max_data_in_flight` parameter added in #561. So maybe we could just remove this parameter altogether?

crates/walrus-service/src/client/utils.rs
mlegner: Hmm... I generally prefer return types that are as specific as possible; i.e., having enum types where all variants can actually occur. But I don't have a very strong opinion on this.

giac-mysten: sounds good to me.

crates/walrus-service/src/client/config.rs
giac-mysten: mmm yep, I think that makes sense.

mlegner: OK, let me update this then.

mlegner: Done in d6c8218cfef4e92add7d0d08550073339b88b01e.

PR #622, opened 2024-07-26
feat(epoch change)!: add shard status in ShardStorage

halfprice:
- Add shard_status singleton table in ShardStorage to track shard status.
- Implement initial logic to reject client write when a shard is locked and about to be moved out.

### DIFF
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -71,3 +71,15 @@ sui:
 #     write_buffer_size: 268435456
 #     target_file_size_base: 4194304
 #     max_bytes_for_level_base: 33554432
+#   shard_status:
+#     enable_blob_files: false
+#     min_blob_size: 0
+#     blob_file_size: 0
+#     blob_compression_type: none
+#     enable_blob_garbage_collection: false
+#     blob_garbage_collection_age_cutoff: 0.0
+#     blob_garbage_collection_force_threshold: 0.0
+#     blob_compaction_read_ahead_size: 0
+#     write_buffer_size: 67108864
+#     target_file_size_base: 67108864
+#     max_bytes_for_level_base: 536870912


crates/walrus-service/src/storage/shard.rs
```diff
@@ -46,9 +46,24 @@ impl<const P: bool, T: Borrow<BlobId>> From<T> for SliverKey<P> {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub enum ShardStatus {
+    Active,
+    ActiveSync,
+    ActiveRecover,
+    LockedToMove,
+}
```
mlegner: Nit: I think I understand the statuses, but it would still be nice to have a descriptive docstring for each variant.

crates/walrus-service/src/node.rs
```diff
@@ -723,6 +723,14 @@ impl ServiceState for StorageNodeInner {
 
         let shard_storage = self.get_shard_for_sliver_pair(sliver_pair_index, blob_id)?;
 
+        let shard_status = shard_storage
+            .status()
+            .context("Unable to retrieve shard status")?;
+
+        if !shard_status.is_owned_by_node() {
+            return Err(ShardNotAssigned(shard_storage.id(), self.current_epoch()).into());
+        }
```
mlegner, line 761: Remark: If all shards this node has are locked, we should also stop responding to other write requests.

halfprice: I was actually thinking about the opposite direction: even though a storage doesn't own any shard, it should still try it best to keep track of metadata and blob_info. This is because we have the assumption that all storage nodes should have the same view of existing blobs (by keeping track of blob info and blob metadata).

crates/walrus-service/src/storage/shard.rs
halfprice: I document the current status in the PR.

I actually removed `ActiveSync` and `ActiveRecover`, and plan to add them back only when implement related functionalities. I personally don't like to put too much speculative code in the repo.

crates/walrus-service/src/node.rs
mlegner: Hmm... Good point. This probably depends on whether or not the storage node wants to remain a storage node or leave the system. Either way, we can address this question in a follow-up.

crates/walrus-service/src/storage/shard.rs
mlegner: > I personally don't like to put too much speculative code in the repo.

I fully agree. 🙂

crates/walrus-service/src/node/storage.rs
```diff
@@ -170,13 +171,19 @@ pub struct DatabaseConfig {
     blob_info: DatabaseTableOptions,
     event_cursor: DatabaseTableOptions,
     shard: DatabaseTableOptions,
+    shard_status: DatabaseTableOptions,
```
mlegner, line 174: Suggestion: Add this to the [example config file](https://github.com/MystenLabs/walrus/blob/main/crates/walrus-service/node_config_example.yaml).

halfprice: Done

PR #623, opened 2024-07-26
feat: add network key to smart contracts

jpcsmith:
Depends on #628

TODO:
- [x] Add network key in rust. 
- [x] Exchange network key between rust and smart contract.

### DIFF
--- a/crates/walrus-sui/Cargo.toml
+++ b/crates/walrus-sui/Cargo.toml
@@ -32,7 +32,7 @@ tokio = { workspace = true, features = ["sync"] }
 tokio-stream.workspace = true
 tracing.workspace = true
 utoipa = { workspace = true, optional = true }
-walrus-core.workspace = true
+walrus-core = { workspace = true, features = ["test-utils"] }
 walrus-test-utils = { workspace = true, optional = true }
 
 [lints]


contracts/blob_store/sources/storage_node.move
```diff
@@ -23,15 +24,27 @@ module blob_store::storage_node {
...
+        network_public_key: vector<u8>,
         shard_ids: vector<u16>,
     ): StorageNodeInfo {
-        StorageNodeInfo { name, network_address, public_key: g1_from_bytes(&public_key), shard_ids }
+        StorageNodeInfo {
+            name,
+            network_address,
+            public_key: g1_from_bytes(&public_key),
+            // TODO(jsmith): Any validation or conversion we should do here?
+            network_public_key,
```
jpcsmith: Question: @karlwuest Would you suggest checking the length of the key here or any other conversion or validation?

karlwuest: A `TODO` for things like this is sufficient, since a lot of the implementations will anyway change, and we will just integrate this in the new set of contracts for testnet (in `contracts/walrus`). 
I'm guessing we'll be using ECDSA keys? We can do a check for the length, but I don't think any more than that is needed. There are no attacks similar to what we need to protect against for BLS aggregate signatures, worst case is that a node submitting a bogus key cannot authenticate itself, which can also happen if they throw away the key.

Issue #407, opened 2024-05-21
Improve configuration files


mlegner:
We have different configuration files for clients and storage nodes. This makes sense given that they require different parameters.

However, they differ even in parts that could be the same:

- The Sui config of the client and storage node could be the same (given that storage nodes will have to interact with the chain as well). Currently the config structure is different as is the name of some parameters (e.g., `system_pkg` vs. `pkg_id`).
- The way `Duration`s are deserialized differs between `ClientCommunicationConfig` and `SuiConfig`.

Furthermore, it would be nice to have templates/examples for both types of configuration files in the same location. The one for the SN is outdated and in the root directory; the one for the client is in `crates/walrus-service`. The template for `walrus-orchestrator` is in `crates/walrus-orchestrator/assets/settings-template.yaml`.

Related to this, we should make sure that default parameters of different components are consistent with each other (e.g., the Sui network).

Some other related issues:
- #373 
- #862
- #904 
- #960
- #1002

PR #625, opened 2024-07-26
refactor!: reorganize `walrus-service` crate

### LINKED ISSUE #400: Move admin tasks like generating configs or deploying contracts into a separate "admin" CLI
They don't actually belong in the storage-node CLI. They would fit better into a separate "admin" CLI.

_Originally posted by @mlegner in https://github.com/MystenLabs/walrus/pull/378#discussion_r1597945722_

### LINKED ISSUE #621: Split and reorganize the walrus-service crate
Our crates are generally well set up, but the walrus-service crate is a bit unorganized with a mixture of general, client-specific, and node-specific code in a somewhat inconsistent hierarchy.

Suggestion: Move the code specific to the Walrus client and storage node to separate crates walrus-client and walrus-node, respectively, and use a consistent hierarchy within those crates.

mlegner:
This reorganizes the `walrus-service` modules and adds new Cargo features.

We now have five top-level modules activated with different features:
- `common` contains functionality shared by client and storage node. Essentially some API definitions, telemetry settings, and other utilities. This is activated with either of the `client` and `node` features.
- `client` contains all client-related code (including the client daemon) and is activated with the `client` feature.
- `node` contains all storage-node-related code and is activated with the `node` feature.
- `testbed` contains all functionality for creating configuration and deploying the system contract and is activated with the `deploy` feature.
- `test_utils` contains various test utilities and is activated with the `test-utils` feature.

In addition, this moves the admin functionality from the `walrus-node` CLI to a new `walrus-deploy` CLI and moves the end-to-end tests to a new `walrus-e2e-tests` crate. The latter is mainly to make sure all necessary features are activated for the tests.

Finally, this contains a few other changes:
- Change visibility of some structs and functions, required by the new crate structure. This required also some new docstrings.
- Some changes to the `api_success_alias` and `rest_api_error` macros.
- Move some definitions into `walrus_service::utils`.
- Clean up some dependencies.
- Clean up some imports, linter exceptions, and docstrings.
- Keep all dependency versions except for the orchestrator in the workspace's `Cargo.toml` file.
- Create a new example for the node configuration file.
- Clean up some TODOs.

This PR represents a breaking change in the sense that the `walrus-node` binary no longer provides the administrative functionality. Scripts and CI workflows that use that functionality need to migrate to `walrus-deploy`.

Closes #400 
Contributes to #407 
Closes #621

### DIFF
--- a/crates/walrus-service/Cargo.toml
+++ b/crates/walrus-service/Cargo.toml
@@ -23,30 +23,22 @@ required-features = ["deploy"]
 
 [features]
 client = [
-    "dep:bcs",
     "dep:colored",
     "dep:humantime",
     "dep:indoc",
     "dep:itertools",
-    "dep:prettytable",
-    "dep:telemetry-subscribers",
-    "dep:utoipa-redoc",
-    "dep:walrus-sdk"
+    "dep:prettytable"
 ]
 deploy = ["client", "node"]
 node = [
     "dep:async-trait",
-    "dep:bcs",
     "dep:enum_dispatch",
     "dep:mime",
     "dep:mysten-metrics",
     "dep:rocksdb",
-    "dep:telemetry-subscribers",
     "dep:tokio-stream",
     "dep:tokio-util",
-    "dep:typed-store",
-    "dep:utoipa-redoc",
-    "dep:walrus-sdk"
+    "dep:typed-store"
 ]
 test-utils = [
     "client",
@@ -60,7 +52,7 @@ test-utils = [
 anyhow.workspace = true
 async-trait = { workspace = true, optional = true }
 axum.workspace = true
-bcs = { workspace = true, optional = true }
+bcs.workspace = true
 clap.workspace = true
 colored = { workspace = true, optional = true }
 enum_dispatch = { workspace = true, optional = true }
@@ -87,7 +79,7 @@ serde_with.workspace = true
 serde_yaml.workspace = true
 sui-sdk.workspace = true
 sui-types.workspace = true
-telemetry-subscribers = { workspace = true, optional = true }
+telemetry-subscribers.workspace = true
 tempfile = { workspace = true, optional = true }
 thiserror.workspace = true
 tokio.workspace = true
@@ -99,9 +91,9 @@ tracing-opentelemetry.workspace = true
 tracing-subscriber.workspace = true
 typed-store = { workspace = true, optional = true }
 utoipa = { workspace = true, features = ["axum_extras"] }
-utoipa-redoc = { workspace = true, optional = true }
+utoipa-redoc.workspace = true
 walrus-core = { workspace = true, features = ["utoipa"] }
-walrus-sdk = { workspace = true, optional = true }
+walrus-sdk.workspace = true
 walrus-sui = { workspace = true, features = ["utoipa"] }
 walrus-test-utils = { workspace = true, optional = true }
 
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -5,6 +5,7 @@
 
 storage_path: /path/to/node/storage
 protocol_key_pair: /path/to/keypair
+network_key_pair: /path/to/network_keypair
 sui:
   rpc: https://fullnode.testnet.sui.io:443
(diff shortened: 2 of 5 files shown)


crates/walrus-service/Cargo.toml
```diff
@@ -6,80 +6,113 @@ version.workspace = true
...
+    "dep:mime",
+    "dep:mysten-metrics",
+    "dep:rocksdb",
+    "dep:telemetry-subscribers",
+    "dep:tokio-stream",
+    "dep:tokio-util",
+    "dep:typed-store",
+    "dep:utoipa-redoc",
+    "dep:walrus-sdk"
+]
```
jpcsmith: Question: Why make the dependencies used by both the client and node optional? These should safely always be dependencies.

crates/walrus-sui/Cargo.toml
```diff
@@ -8,7 +8,12 @@ license.workspace = true
 
 [features]
 default = ["test-utils"]
-test-utils = ["dep:test-cluster", "dep:walrus-test-utils", "dep:tempfile"]
+test-utils = [
+    "dep:test-cluster",
+    "dep:walrus-test-utils",
+    "dep:tempfile",
+    "walrus-core/test-utils"
```
jpcsmith, line 15: Question: Why a path here?

crates/walrus-service/src/common/utils.rs
```diff
@@ -24,11 +24,51 @@ use rand::{
...
+        let reader = std::fs::File::open(path)
+            .with_context(|| format!("Unable to load config from {}", path.display()))?;
+
+        Ok(serde_yaml::from_reader(reader)?)
+    }
+}
+
 /// The representation of a backoff strategy.
-pub(crate) trait BackoffStrategy {
+pub trait BackoffStrategy {
```
jpcsmith: Question: Why the increase in visibility here?

crates/walrus-service/src/common/telemetry.rs
```diff
@@ -159,7 +166,7 @@ impl MakeHttpSpan {
     }
 }
 
-struct HeaderExtractor<'a>(pub &'a http::HeaderMap);
+struct HeaderExtractor<'a>(pub(crate) &'a http::HeaderMap);
```
jpcsmith: Question: Given that this is `pub(self)` scoped, why change the visibility of the field?

crates/walrus-service/node_config_example.yaml
```diff
@@ -0,0 +1,72 @@
+# Example configuration for a Walrus storage node.
+#
+# Only the uncommented parameters are required.
+# Default optional values are commented out.
+
+storage_path: /path/to/node/storage
+protocol_key_pair: /path/to/keypair
```
jpcsmith, line 7:
```suggestion
protocol_key_pair: /path/to/keypair
network_key_pair: /path/to/network_keypair
```

crates/walrus-service/Cargo.toml
mlegner: Ah, good point. Originally I had the `common` module public such that the crate made sense without any of the features enabled. But after making it `pub(crate)`, we can also unify these.

crates/walrus-sui/Cargo.toml
mlegner: According to my understanding, that's the way to enable features of dependencies (see [here](https://doc.rust-lang.org/cargo/reference/features.html#dependency-features)). However, I forgot to disable the feature in the default case, which I will change.

crates/walrus-service/src/common/utils.rs
mlegner: That's an artifact of multiple stages of refactoring... 😅

crates/walrus-service/node_config_example.yaml
mlegner: Oh, thank you, I missed that change.

crates/walrus-service/src/common/telemetry.rs
mlegner: No reason, will change back.

crates/walrus-service/node_config_example.yaml
jpcsmith: So did I when implementing it :)

PR #627, opened 2024-07-29
fix(client): merge the returned coin back into the gas

giac-mysten:
Ensure that the coins that are split off to pay for storage are merged back into the gas coin. 
Avoids coins with zero balance in the user's wallet.

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -345,11 +345,20 @@ impl SuiContractClient {
                 .with_type_params(&[self.read_client.coin_type.clone()]),
             reserve_arguments,
         )?;
-        // Transfer back payment coin.
-        pt_builder.transfer_arg(
-            self.wallet_address,
-            Argument::NestedResult(reserve_result_index, 1),
-        );
+
+        let returned_coin = Argument::NestedResult(reserve_result_index, 1);
+
+        if self.read_client.uses_sui_coin() {
+            // Join the return coin back into the gas coin. This ensures that there are no
+            // zero-balance coins left in the user's wallet.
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -345,11 +345,20 @@ impl SuiContractClient {
...
+        if self.read_client.uses_sui_coin() {
+            // Join the return coin back into the gas coin. This ensures that there are no
+            // zero-balance coins left in the user's wallet.
+            pt_builder.command(Command::MergeCoins(Argument::GasCoin, vec![returned_coin]));
+        } else {
+            // Transfer back payment coin.
+            pt_builder.transfer_arg(
+                self.wallet_address,
+                Argument::NestedResult(reserve_result_index, 1),
+            );
```
mlegner, line 360: Nit: If we're at it, we could also check for a zero balance and destroy the coin in that case with [`destroy_zero`](https://github.com/MystenLabs/sui/blob/eb55624e7a5decda7f4946fe86d2b96b3aed6fb1/crates/sui-framework/packages/sui-framework/sources/coin.move#L205).

giac-mysten: I thought about this. What would we gain from this, except the cleaner semantics? I the gas price different?

mlegner: I meant the case where the payment coin is not SUI. In this case we could also end up with a zero balance on the payment coin, and then it might make sense to destroy it.

mlegner: @giac-mysten How about we merge the PR as it is now and follow up on this? This is currently not relevant anyway and even later it's just a corner case.

mlegner: I've added this to #262.

PR #628, opened 2024-07-29
feat: add a key-pair for authenticating network communication

jpcsmith:
Defines a key-pair for authenticating network communication and add it to the config.

This is a breaking change as prior configs lacking the network key pair cannot be used.

### DIFF
--- a/crates/walrus-core/src/keys.rs
+++ b/crates/walrus-core/src/keys.rs
@@ -171,6 +171,7 @@ impl<T: SupportedKeyPair> SerializeAs<TaggedKeyPair<T>> for SerdeWithBase64 {
     }
 }
 
+/// Serde seserialization visitor for writing an array of known length into a pre-allocated buffer.
 struct BinaryBufferVisitor<const N: usize> {
     length: usize,
 }


crates/walrus-core/src/keys.rs
```diff
@@ -4,163 +4,239 @@
...
+impl<T: SupportedKeyPair> SerializeAs<TaggedKeyPair<T>> for SerdeWithBase64 {
+    fn serialize_as<S>(source: &TaggedKeyPair<T>, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        SerializeAsWrap::<Vec<u8>, SerdeWithBase64>::new(&Vec::from(source)).serialize(serializer)
+    }
+}
+
+struct BinaryBufferVisitor<const N: usize> {
```
jpcsmith, line 175:
```suggestion
/// Serde seserialization visitor for writing an array of known length into a pre-allocated buffer.
struct BinaryBufferVisitor<const N: usize> {
```

crates/walrus-core/src/keys.rs
```diff
@@ -4,163 +4,239 @@
...
     {
-        let decoded_bytes: [u8; Self::ENCODED_LENGTH] =
-            <[serde_with::Same; Self::ENCODED_LENGTH]>::deserialize_as(deserializer)?;
-
-        if decoded_bytes[0] == SignatureScheme::BLS12381.to_u8() {
-            let keypair =
-                BLS12381KeyPair::from_bytes(&decoded_bytes[1..]).map_err(D::Error::custom)?;
+        let decoded_bytes: [u8; ENCODING_BUFFER_LENGTH] = deserializer.deserialize_tuple(
+            Self::ENCODED_LENGTH,
+            BinaryBufferVisitor::new(Self::ENCODED_LENGTH),
```
karlwuest, line 214: Question (out of curiosity): Why is a custom visitor for the buffer needed here? Why does the previous approach not work anymore?

jpcsmith: The compiler complains that we cannot use `Self::ENCODED_LENGTH` in a constant expression (such as in the type `let decoded_bytes: [u8; Self::ENCODED_LENGTH] = ...`) because it's dependent upon a generic parameter. As a result, the previous approach which used a generic to decode into an array of the exact length does not work.

karlwuest: I see, thanks for the context!

PR #630, opened 2024-07-30
feat(storage)!: split primary and secondary slivers into separate column families

halfprice:
This PR creates primary and secondary slivers in different column families in ShardStorage.

Previously, they are using the same column family, which makes range iterator unintuitive. e.g. primary_sliver.iter().begin() points to the beginning of second sliver, and secondary_sliver.iter().end() points to the end of primary sliver.

This PR also changes the DBMap key from SliverKey=(bool, BlobId) to just BlobId, since primary and secondary slivers
are now stored in different column families.

BREAKING: this PR changes on disk data layout, which is incompatible with previous layout.

### DIFF
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -213,14 +213,12 @@ fn id_from_column_family_name(name: &str) -> Option<(ShardIndex, SliverType)> {
             tracing::warn!(%name, "ignoring shard-like column family with an ID out of range");
             return None;
         };
-        let sliver_type = captures.get(2)?.as_str();
-        if sliver_type == "primary" {
-            Some((ShardIndex(id), SliverType::Primary))
-        } else if sliver_type == "secondary" {
-            Some((ShardIndex(id), SliverType::Secondary))
-        } else {
-            panic!("Invalid sliver type in regex capture");
-        }
+        let sliver_type = match captures.get(2)?.as_str() {
+            "primary" => SliverType::Primary,
+            "secondary" => SliverType::Secondary,
+            _ => panic!("Invalid sliver type in regex capture"),
+        };
+        Some((ShardIndex(id), sliver_type))
     })
 }
 
@@ -246,7 +244,7 @@ mod tests {
         ShardIndex,
         SliverType,
     };
-    use walrus_test_utils::{async_param_test, Result as TestResult};
+    use walrus_test_utils::{async_param_test, param_test, Result as TestResult};
 
     use super::id_from_column_family_name;
     use crate::{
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/storage/shard.rs
```diff
@@ -173,43 +154,71 @@ impl ShardStorage {
...
+    /// with the specified index.
     pub(crate) fn slivers_column_family_options(
         id: ShardIndex,
         db_config: &DatabaseConfig,
-    ) -> (String, Options) {
-        (
-            slivers_column_family_name(id),
-            db_config.shard().to_options(),
-        )
+    ) -> HashMap<SliverType, (String, Options)> {
```
jpcsmith: Nit: This could just be an array of 2 elements.

crates/walrus-service/src/storage/shard.rs
```diff
@@ -173,43 +154,71 @@ impl ShardStorage {
...
+            return None;
+        };
+        let sliver_type = captures.get(2)?.as_str();
+        if sliver_type == "primary" {
+            Some((ShardIndex(id), SliverType::Primary))
+        } else if sliver_type == "secondary" {
+            Some((ShardIndex(id), SliverType::Secondary))
+        } else {
+            panic!("Invalid sliver type in regex capture");
+        }
```
jpcsmith: Nit:
```suggestion
        let sliver_type = match captures.get(2)?.as_str() {
            "primary" => SliverType::Primary,
            "secondary" => SliverType::Secondary,
            _ => panic!("Invalid sliver type in regex capture");
        }
        Some((ShardIndex(id), sliver_type))
```

crates/walrus-service/src/storage/shard.rs
```diff
@@ -387,4 +403,25 @@ mod tests {
...
+        );
+
+        assert_eq!(id_from_column_family_name("shard-a/primary-slivers"), None);
+
+        assert_eq!(id_from_column_family_name("shard-20/random-slivers"), None);
+
+        assert_eq!(id_from_column_family_name("shard-20/slivers"), None);
+
+        Ok(())
+    }
```
jpcsmith: Nit: Better as a parameterised test, see the `param_test!` macro.

crates/walrus-service/src/storage/shard.rs
halfprice: Done

crates/walrus-service/src/storage/shard.rs
halfprice: Done

crates/walrus-service/src/storage/shard.rs
halfprice: I personally like to make this a bit more explicit so that when holding a return value, getting the data per type doesn't implicitly rely on the order of the output.

PR #635, opened 2024-07-31
refactor: use macro for version to prevent unnecessary rebuilds

### LINKED ISSUE #632: Check for unnecessary rebuilds
The placement of the GIT_REVISION in a library file may be causing unnecessary rebuilds
https://github.com/MystenLabs/walrus/blob/c93c971c5e226f3b691ce6a5dcc15ffabaf7aa43/crates/walrus-service/src/cli_utils.rs#L24. See [here](https://github.com/MystenLabs/sui/commit/e66106986c) for the issue in Sui. In Sui this was fixed by moving this to a macro.

mlegner:
Closes #632

### DIFF
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -39,7 +39,7 @@ use walrus_service::{
         Client,
         ClientDaemon,
     },
-    version,
+    utils::version,
 };
 use walrus_sui::{
     client::{ContractClient, ReadClient},
--- a/crates/walrus-service/bin/deploy.rs
+++ b/crates/walrus-service/bin/deploy.rs
@@ -16,7 +16,7 @@ use tokio::sync::oneshot;
 use walrus_service::{
     node::config::defaults::{METRICS_PORT, REST_API_PORT},
     testbed,
-    version,
+    utils::version,
 };
 use walrus_sui::utils::SuiNetwork;
 
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -25,8 +25,7 @@ use tokio_util::sync::CancellationToken;
 use walrus_core::keys::ProtocolKeyPair;
 use walrus_service::{
     node::{config::StorageNodeConfig, server::UserServer, StorageNode},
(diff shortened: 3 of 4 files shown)


crates/walrus-service/bin/client.rs
```diff
@@ -39,13 +39,15 @@ use walrus_service::{
         Client,
         ClientDaemon,
     },
-    utils::VERSION,
+    version,
```
jpcsmith: Question: Is macro imported here a re-export from `utils`?

crates/walrus-service/bin/client.rs
```diff
@@ -39,13 +39,15 @@ use walrus_service::{
...
 };
 use walrus_sui::{
     client::{ContractClient, ReadClient},
     utils::price_for_encoded_length,
 };
 
+version!(VERSION);
```
```suggestion
const VERSION: &str = utils::version!();
```
The current implementation gives the impression that the version is being passed to the macro, instead of it being the return value of the macro.

crates/walrus-service/bin/client.rs
mlegner: AFAIU, exported macros always live at the top level of a crate. Not sure if that can be changed... 🤔

crates/walrus-service/bin/client.rs
mlegner: I slightly changed the macro in 56ea32c56418697656fb1f2fc2ba100d080e4bdc to make the behavior a bit more intuitive.

crates/walrus-service/bin/client.rs
jpcsmith: No need to "force" it, but elsewhere it was imported via `utils` so I found it strange that it was imported via the root here. If importing via utils also works here, it would be perhaps be clearer to do so as that's where it's implemented.

mlegner: I think it worked before because I had a `pub use version` in the `utils` module. Without this, it cannot be imported from `utils` but only from the root.

I re-added the `pub use` statement and changed the imports to the `utils` module in bc911610ed9bf8f2700ccc8bb40fa3f8915ac7c8.

PR #638, opened 2024-08-01
feat(staking): implements most of the methods in the skeleton

damirka:


contracts/walrus/sources/staking.move
```diff
@@ -126,9 +126,12 @@ module walrus::staking {
         staking.inner_mut().request_withdrawal(staked_wal, amount)
     }
 
+    #[allow(lint(self_transfer))]
     /// Withdraws the staked amount from the staking pool.
     public fun withdraw_stake(staking: &mut Staking, staked_wal: StakedWal, ctx: &mut TxContext) {
-        staking.inner_mut().withdraw_stake(staked_wal, ctx)
+        let coin = staking.inner_mut().withdraw_stake(staked_wal, ctx);
+        // TODO: we shouldn't return a Coin here
```
tnowacki: Should the TODO be on `withdraw_stake`

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -1,46 +1,52 @@
...
+        pool_id,
+        principal,
+        activation_epoch,
     }
+}
 
-    /// TODO: cycle back on this
-    public fun split(_self: &mut StakedWal): StakedWal { abort 0 }
+/// Burns the staked WAL and returns the `principal`.
+public(package) fun burn(staked_wal: StakedWal): u64 {
```
tnowacki: ramping up here, so forgive my stupid questions. 

Do we want a proof of burn receipt here?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,145 @@
...
+
+use sui::{clock::Clock, coin::Coin, object_table::{Self, ObjectTable}, sui::SUI};
+use walrus::{
+    staked_wal::{Self, StakedWal},
+    staking_pool::{Self, StakingPool},
+    storage_node::StorageNodeCap
+};
+
+/// TODO: remove this once the module is implemented.
+const ENotImplemented: u64 = 0;
```
tnowacki:
```suggestion
/// TODO: remove this once the module is implemented.
#[error]
const ENotImplemented: vector<u8> = b"NOT YET IMPLEMENTED";
```

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,145 @@
...
+
+/// The inner object for the staking part of the system.
+public struct StakingInnerV1 has store {
+    /// Stored staking pools, each identified by a unique `ID` and contains
+    /// the `StakingPool` object.
+    ///
+    /// TODO: consider using Object Table
+    pools: ObjectTable<ID, StakingPool>,
+    /// The current epoch of the Walrus system.
+    current_epoch: u64,
```
tnowacki: This is different than the sui epoch?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,145 @@
...
+/// Stakes the given amount of `T` with the pool.
+public(package) fun stake_with_pool(
+    self: &mut StakingInnerV1,
+    to_stake: Coin<SUI>,
+    pool_id: ID,
+    ctx: &mut TxContext,
+): StakedWal {
+    let pool = &mut self.pools[pool_id];
+    let activation_epoch = self.current_epoch + 1;
+    staking_pool::stake(pool, to_stake, activation_epoch, ctx)
```
tnowacki:
```suggestion
    let activation_epoch = self.current_epoch + 1;
    self.pools[pool_id].stake(to_stake, activation_epoch, ctx)
```

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -1,125 +1,194 @@
...
+}
 
-        destroy(pool);
+#[allow(unused_variable)]
+/// Small helper to match the value. Rust says hi!
+macro fun matches<$T>($x: &$T, $p: $T): bool {
+    let p = $p;
+    match ($x) {
+        p => true,
+        _ => false,
```
tnowacki: This is not doing what you think it is doing.

This 
```suggestion
matches!(&pool.state, PoolState::New)
```
expands to (I will enumerate `p` for shadowing to show what is going on more clearly)
```
let p#0 = PoolState::new;
match ($x) {
    p#1 => true,
    _ => false, // unreachable, you should get a compiler warning about this
}
```

So it will always evaluate to true, i.e 
```
assert!(matches!(&PoolState::New, PoolState::Withdrawing))
assert!(matches!(&PoolState::New, PoolState::Active))
assert!(matches!(&PoolState::New, PoolState::New))
```

contracts/walrus/sources/staking/storage_node.move
```diff
@@ -3,64 +3,74 @@
...
 
-    #[test_only]
-    /// Create a storage node capability for testing.
-    public fun new_cap_for_testing(pool_id: ID, ctx: &mut TxContext): StorageNodeCap {
-        StorageNodeCap { id: object::new(ctx), pool_id }
+#[test_only]
+/// Create a storage node with dummy name & address
+public fun new_for_testing(public_key: vector<u8>, weight: u16): StorageNodeInfo {
+    let mut shard_ids = vector[];
+    weight.do!(|i| shard_ids.push_back(i));
```
tnowacki, line 64: I know it is guaranteed here but should we assert
`assert!(shard_ids.length() <= 0xFFFF)` (probably want a constant but you get the idea)

contracts/walrus/sources/staking/staked_wal.move
karlwuest: I don't think we need it. 
@damirka maybe `unwrap` is a better name than `burn`?

contracts/walrus/sources/staking/staking_inner.move
karlwuest: Yes, Walrus has much longer epochs.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,145 @@
...
 
-    // === Staking ===
+// === Staking ===
 
-    /// Blocks staking for the pool, marks it as "withdrawing".
-    public(package) fun set_withdrawing(self: &mut StakingInnerV1, pool_id: ID) {
-        self.pools[pool_id].set_withdrawing();
-    }
+/// Blocks staking for the pool, marks it as "withdrawing".
+/// TODO: Is this action instant or should it be processed in the next epoch?
```
karlwuest, line 113: This should exclude the pool immediately from committee selection, but it doesn't include the storage node from being part of a committee that has already been decided, and it does not allow immediate withdrawal (this requires waiting until all shards have been transferred).

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -1,125 +1,194 @@
...
+    ctx: &mut TxContext,
+): StakedWal {
+    assert!(pool.is_active());
+    assert!(to_stake.value() > 0);
+
+    let amount = to_stake.value();
+    let staked_wal = staked_wal::mint(
+        pool.id.to_inner(),
+        amount,
+        current_epoch + 1, // always the next epoch
```
karlwuest: This should be the next epoch as long as we're before the point at which the new committee is decided, but after that it has to be `current_epoch + 2`. 
I will add some more detail (with a figure) about the epoch change process to the notion doc.

contracts/walrus/sources/staking/staked_wal.move
damirka: It's a common language that we use everywhere and probably stems from NFTs. Don't have a strong preference, not a fan of `mint` and `burn` either.

karlwuest: `burn` to me implies that the object with its contents is destroyed, but here we're returning the contents. I have no strong feelings about this, though, so feel free to keep it as is.

damirka: We do the same with Coin/Balance iirc. Usually it means that the object is destroyed but this logic doesn't normally apply to the return type.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,184 @@
...
+    pools: ObjectTable<ID, StakingPool>,
+    /// The current epoch of the Walrus system. The epochs are not the same as
+    /// the Sui epochs, not to be mistaken with `ctx.epoch()`.
+    current_epoch: u64,
+    /// The vote for the next epoch. Used to update the `storage_price`,
+    /// `write_price` and `storage_capacity`. If the vote is not changed, the
+    /// current values are used as a vote for the next epoch.
+    ///
+    /// TODO: Revisit and consider a Table approach to prevent hitting
+    /// the object size limit. Current solution is a temporary workaround.
```
karlwuest: Remark: We could also store the voting parameters within each staking pool, then we only have one dynamic field access per staking pool when we create the committee.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,184 @@
...
+}
 
-    public(package) fun new(ctx: &mut TxContext): StakingInnerV1 {
-        StakingInnerV1 { pools: object_table::new(ctx) }
-    }
+#[allow(unused_field)]
+public struct VotingParams has store, copy, drop {
+    storage_price: u64,
+    write_price: u64,
+    storage_capacity: u64,
```
karlwuest: Suggestion: change this to `node_capacity`

Context: In Lisbon, Nihar and I (+ others) discussed changing this from an explicit vote for the `storage_capacity` to each node specifying how much capacity they have (in total), then during voting this counts as a vote for `node_capacity/n_shards_assigned_to_node*n_total_shards`. This makes it easier for storage nodes to decide how to vote for this parameter, because when specifying the vote, they cannot be sure how many shards they will receive.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,184 @@
...
+
+    let pool_id = object::id(&pool);
+    self.pools.add(pool_id, pool);
+    pool_id
+}
 
-    public(package) fun withdraw_node(self: &mut StakingInnerV1, cap: StorageNodeCap) {
-        abort ENotImplemented
-    }
+public(package) fun register_candidate(
```
karlwuest, line 70: Remark: Here, we will also need to pass the node parameters including initial votes (does not need to be included in this PR).

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,184 @@
...
-        self: &mut StakingInnerV1,
-        pool_id: ID,
-        ctx: &mut TxContext,
-    ): StorageNodeCap {
-        abort ENotImplemented
-    }
+// === Storage Node ===
+
+/// Creates a new staking pool with the given `commission_rate`.
+public(package) fun create_pool(
```
karlwuest, line 48: Question/Remark: Where do you plan on storing the StorageNodeInfo? 
If we store this in the pool, then we need to pass this here as well. We could then call this function from the `register_candidate` function to which we can pass all node parameters.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,184 @@
...
+    let wctx = &self.new_walrus_context();
+    self.pools[pool_id].stake(to_stake, wctx, ctx)
+}
+
+/// Requests withdrawal of the given amount from the staked `T`, the withdraw is
+/// not immediate and will be processed in the next epoch.
+public(package) fun request_withdraw_stake(
+    self: &mut StakingInnerV1,
+    staked_wal: &mut StakedWal,
+    amount: u64,
```
karlwuest: I think we can remove the amount here. If we have `split` for `StakedWal`, then this function can just request withdrawal for the full amount.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -2,128 +2,184 @@
...
+    ctx: &mut TxContext,
+) {
+    let wctx = &self.new_walrus_context();
+    self.pools[staked_wal.pool_id()].request_withdraw_stake(staked_wal, wctx, ctx)
+}
+
+/// Similar to the `request_withdrawal` but takes the full value of the `StakedWal`.
```
karlwuest: The `request_withdrawal` should mark the `StakedWal` as "to withdraw" and then this function should actually perform the withdrawal once the stake can be withdrawn.

contracts/walrus/sources/staking.move
```diff
@@ -3,214 +3,223 @@
...
+    node_cap
+}
+
+/// Blocks staking for the nodes staking pool
+/// Marks node as "withdrawing",
+/// - excludes it from the next committee selection
+/// - still has to remain active while it is part of the committee and until all shards have
+///     been transferred to its successor
+/// - The staking pool is deleted once the last funds have been withdrawn from it by its stakers
+public fun withdraw_node(staking: &mut Staking, cap: StorageNodeCap) {
```
karlwuest:
```suggestion
public fun withdraw_node(staking: &mut Staking, cap: &mut StorageNodeCap) {
```
We still need the StorageNodeCap to, e.g., withdraw rewards after the last epoch.

contracts/walrus/sources/staking.move
```diff
@@ -3,214 +3,223 @@
...
+    abort ENotImplemented
+}
+
+/// Checks if the node should either have received the specified shards from the specified node
+/// or vice-versa.
+///
+/// - also checks that for the provided shards, this function has not been called before
+/// - if so, slashes both nodes and emits an event that allows the receiving node to start
+///     shard recovery
+/// TODO(kwuest): decide whether to keep this in `staking` or move to `system`
```
karlwuest: Let's keep this in staking, I think this will be cleaner because then we can do all of the epoch progress handling here.

contracts/walrus/sources/staking.move
```diff
@@ -3,214 +3,223 @@
...
+    staking.inner_mut().request_withdraw_stake(staked_wal, amount, ctx);
+}
+
+#[allow(lint(self_transfer))]
+/// Withdraws the staked amount from the staking pool.
+public fun withdraw_stake(staking: &mut Staking, staked_wal: StakedWal, ctx: &mut TxContext) {
+    let coin = staking.inner_mut().withdraw_stake(staked_wal, ctx);
+    // TODO: we shouldn't return a Coin here
+    transfer::public_transfer(coin, ctx.sender())
+}
```
karlwuest: Question: Is there a reason for not just doing the following?

```suggestion
public fun withdraw_stake(staking: &mut Staking, staked_wal: StakedWal, ctx: &mut TxContext): Coin<SUI> {
    staking.inner_mut().withdraw_stake(staked_wal, ctx)
}
```

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -1,46 +1,127 @@
...
+    Withdrawing,
+}
 
-    /// Burns the staked WAL and returns the `pool_id` and the `principal`.
-    public(package) fun burn(staked_wal: StakedWal): (ID, u64) {
-        let StakedWal { id, pool_id, principal } = staked_wal;
-        id.delete();
-        (pool_id, principal)
+/// Represents a staked WAL, does not store the `Balance` inside, but uses
+/// `u64` to represent the staked amount. Behaves similarly to `Balance` and
```
karlwuest, line 24: Remark: As mentioned, we can/should change this now to store the Balance (no need to include it in this PR).

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -1,46 +1,127 @@
...
-        StakedWal {
-            id: object::new(ctx),
-            pool_id,
-            principal,
-        }
-    }
+/// The state of the staked WAL.
+public enum StakedWalState has store, copy, drop {
+    Staked,
+    Withdrawing,
```
karlwuest: Suggestion: When set to `Withdrawing`, we also need to add the epoch in which it can be withdrawn.

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -1,46 +1,127 @@
...
+}
+
+/// Burns the staked WAL and returns the `principal`.
+public(package) fun burn(sw: StakedWal): u64 {
+    let StakedWal { id, principal, .. } = sw;
+    id.delete();
+    principal
+}
+
+/// Sets the staked WAL state to `Withdrawing` and updates the `activation_epoch`
```
karlwuest: Question: Why do we need to update the activation epoch here?

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -1,125 +1,268 @@
...
+
+use sui::{balance::{Self, Balance}, coin::Coin, sui::SUI, vec_map::{Self, VecMap}};
+use walrus::{staked_wal::{Self, StakedWal}, walrus_context::WalrusContext};
+
+/// Represents the state of the staking pool.
+///
+/// TODO: revisit the state machine.
+public enum PoolState has store, copy, drop {
+    Active,
+    Withdrawing,
```
karlwuest: Remark: Similar to `StakedWal`, we also need to add the epoch here at which the pool can actually withdraw, and then add another state "Withdrawn" that allows all stake from this pool to be immediately unstaked.

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -1,125 +1,268 @@
...
+    };
+
+    staked_wal
+}
+
+/// Request withdrawal of the given amount from the staked WAL.
+/// Marks the `StakedWal` as withdrawing and updates the activation epoch.
+///
+/// TODO: rewards calculation.
+/// TODO: if pool is out and is withdrawing, we can perform the withdrawal immediately
```
karlwuest, line 150: Only if the pool is already withdrawn.

contracts/walrus/tests/staking/staked_wal_tests.move
```diff
@@ -0,0 +1,90 @@
...
+    // test that zero can be destroyed
+    let zero = staked_wal.split(0, ctx);
+    zero.destroy_zero();
+
+    // test that the staked WAL can be burned
+    let principal = staked_wal.burn();
+    assert!(principal == 100);
+}
+
+#[test, expected_failure]
```
karlwuest, line 45: Suggestion: add expected error code, otherwise this test will also succeed if the split doesn't cause an error.

contracts/walrus/tests/staking/staked_wal_tests.move
```diff
@@ -0,0 +1,90 @@
...
+    let ctx = &mut tx_context::dummy();
+    let pool_id = ctx.fresh_object_address().to_id();
+    let mut staked_wal = staked_wal::mint(pool_id, 100, 1, ctx);
+
+    let _other = staked_wal.split(101, ctx);
+
+    abort 1337
+}
+
+#[test, expected_failure]
```
karlwuest, line 57: Suggestion: add expected error code

contracts/walrus/tests/staking/staked_wal_tests.move
```diff
@@ -0,0 +1,90 @@
...
+    let pool_id = ctx.fresh_object_address().to_id();
+    let mut sw1 = staked_wal::mint(pool_id, 100, 1, ctx);
+    let sw2 = staked_wal::mint(pool_id, 100, 2, ctx);
+
+    sw1.join(sw2);
+
+    abort 1337
+}
+
+#[test, expected_failure]
```
karlwuest, line 70: Suggestion: add expected error code

contracts/walrus/tests/staking/staked_wal_tests.move
```diff
@@ -0,0 +1,90 @@
...
+    let pool_id2 = ctx.fresh_object_address().to_id();
+    let mut sw1 = staked_wal::mint(pool_id1, 100, 1, ctx);
+    let sw2 = staked_wal::mint(pool_id2, 100, 1, ctx);
+
+    sw1.join(sw2);
+
+    abort 1337
+}
+
+#[test, expected_failure]
```
karlwuest, line 84: Suggestion: add expected error code

contracts/walrus/tests/staking/staked_wal_tests.move
damirka: We don't have error codes just yet, this and `assert!()` calls will have defined error constants in follow ups.

damirka: Tracking issue: #672

contracts/walrus/sources/staking/staking_inner.move
damirka: Good catch! Forgot to patch the comment, implementation is different.

contracts/walrus/sources/staking.move
damirka: Nice! But the function signature should be changed to return `Coin<SUI>` to be more composable.

contracts/walrus/sources/staking/staked_wal.move
damirka: We reuse `activation_epoch` for it, see the implementation.

contracts/walrus/sources/staking/staked_wal.move
damirka: To not introduce a new field, and then to still support broader APIs like "split" and "join" in the way they're right now. Activation epoch becomes the epoch when the "withdrawing" is possible.

contracts/walrus/sources/staking/staking_pool.move
damirka: Good one, will do.

contracts/walrus/sources/staking/staked_wal.move
karlwuest: Let's not overload `activation_epoch` for this. I think it's better if we keep separate fields for separate purposes. We may still need to access the `activation_epoch` to properly calculate slashing/rewards (because the rewards from the last epoch still need to be accounted for).

contracts/walrus/sources/staking/staked_wal.move
damirka: Done, will push in a bit.

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -50,16 +55,15 @@ public(package) fun mint(
...
 
 /// Burns the staked WAL and returns the `principal`.
-public(package) fun burn(sw: StakedWal): u64 {
+public(package) fun unwrap(sw: StakedWal): Balance<SUI> {
     let StakedWal { id, principal, .. } = sw;
     id.delete();
     principal
 }
 
 /// Sets the staked WAL state to `Withdrawing` and updates the `activation_epoch`
```
karlwuest:
```suggestion
/// Sets the staked WAL state to `Withdrawing`
```

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -11,17 +11,20 @@ use walrus::{staked_wal::{Self, StakedWal}, walrus_context::WalrusContext};
 ///
 /// TODO: revisit the state machine.
 public enum PoolState has store, copy, drop {
+    /// The pool is new and awaits the stake to be added.
+    New,
+    /// The pool is active and can accept stakes.
     Active,
+    /// The pool awaits the stake to be withdrawn.
     Withdrawing,
```
karlwuest:
```suggestion
    /// The pool awaits the stake to be withdrawn. The value inside the 
    /// variant is the epoch in which the pool will be withdrawn.
    Withdrawing(u64),
```

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -9,10 +9,15 @@
...
 
+use sui::{balance::Balance, sui::SUI};
+
 /// The state of the staked WAL.
 public enum StakedWalState has store, copy, drop {
+    /// Default state of the staked WAL - it is staked in the staking pool.
     Staked,
-    Withdrawing,
+    /// The staked WAL is in the process of withdrawing. The value inside the
+    /// invariant is the epoch when the staked WAL can be withdrawn.
```
karlwuest:
```suggestion
    /// The staked WAL is in the process of withdrawing. The value inside the
    /// variant is the epoch when the staked WAL can be withdrawn.
```

contracts/walrus/sources/staking/staking_inner.move
damirka: Tracking issue #674,

contracts/walrus/sources/staking/staking_inner.move
damirka: Done, thanks!

contracts/walrus/sources/staking/staked_wal.move
damirka: Done, thanks. Now stores in the enum.

contracts/walrus/sources/staking/staking_inner.move
damirka: Added an issue #675

PR #647, opened 2024-08-07
feat(storage): server side implementation of fetching shard

halfprice:
This PR implements handling `SyncShardRequest` in the server side (source storage node of a sync shard operation).

Basically, the storage locates the beginning of the range of blobs to fetch, filters out blobs that are not certified, and
returns the slivers of the shards.

Related #259

### DIFF
--- a/crates/walrus-service/src/node/errors.rs
+++ b/crates/walrus-service/src/node/errors.rs
@@ -122,8 +122,8 @@ pub enum SyncShardError {
     Unauthorized,
     #[error(transparent)]
     MessageVerificationError(#[from] MessageVerificationError),
-    #[error("Requested shard {0} not found")]
-    ShardNotFound(ShardIndex),
+    #[error(transparent)]
+    ShardNotAssigned(#[from] ShardNotAssigned),
     #[error("The request came from an epoch that is too old: {0}. Current epoch is {1}")]
     EpochTooOld(Epoch, Epoch),
     #[error(transparent)]
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -333,6 +333,7 @@ fn shard_status_column_family_name(id: ShardIndex) -> String {
 mod tests {
     use std::collections::HashMap;
 
+    use typed_store::TypedStoreError;
     use walrus_core::{
         encoding::{Primary, Secondary},
         BlobId,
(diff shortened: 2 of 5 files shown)


crates/walrus-service/src/node/errors.rs
```diff
@@ -122,4 +122,10 @@ pub enum SyncShardError {
     Unauthorized,
     #[error(transparent)]
     MessageVerificationError(#[from] MessageVerificationError),
+    #[error("Requested shard {0} not found")]
+    ShardNotFound(ShardIndex),
```
jpcsmith: Is this analogous to the `ShardNotAssigned` error used above (i.e., the requested shard is not valid for the storage node). If so, reuse the error.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -482,4 +519,90 @@ mod tests {
...
+        assert!(shard.fetch_slivers(sliver_type, &[blob_ids[1]])?.is_empty());
+
+        // Tests fetching a mix of existent and non-existent slivers.
+        assert_eq!(
+            shard.fetch_slivers(sliver_type, &blob_ids)?,
+            vec![
+                (blob_ids[0], data[&blob_ids[0]][&sliver_type].clone()),
+                (blob_ids[2], data[&blob_ids[2]][&sliver_type].clone())
+            ]
+        );
```
jpcsmith: Suggestion: These are 4 separate tests in a single function, which makes debugging hard when something breaks. Split them out into individual tests.

crates/walrus-service/src/test_utils.rs
```diff
@@ -646,6 +648,7 @@ pub struct TestClusterBuilder {
     event_providers: Vec<Option<Box<dyn SystemEventProvider>>>,
     committee_factories: Vec<Option<Box<dyn CommitteeServiceFactory>>>,
     contract_services: Vec<Option<Box<dyn SystemContractService>>>,
+    initial_epoch: Option<Epoch>,
```
jpcsmith, line 651: Suggestion: Use 0 instead of option for the initial_epoch if unset.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -250,6 +250,39 @@ impl ShardStorage {
             .map(|s| s.unwrap_or(ShardStatus::Active))
     }
 
+    /// Fetches the slivers with `sliver_type` for the provided blob IDs.
+    pub(crate) fn fetch_slivers(
+        &self,
+        sliver_type: SliverType,
+        slivers_to_fetch: &[BlobId],
+    ) -> Result<Vec<(BlobId, Sliver)>, TypedStoreError> {
```
jpcsmith, line 258: Remark: We may want to return the raw slivers instead of the decoded slivers.

crates/walrus-service/src/node/errors.rs
halfprice: Yes, good suggestion. Done.

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/test_utils.rs
halfprice: I think Option is better in terms of readability since it is an optional input for TestClusterBuilder. If it is not set, the default value is indeed set to 0.

crates/walrus-service/src/node/storage/shard.rs
halfprice: The idea here is that we want to do some kind of verification of the sliver on the receiver side. So I use the decoded version here.

However, I think your suggestion is still valid if the encoded bytes is smaller (is that true?), so that we can safe on bandwidth. If so, I think maybe we also want to change `get_sliver` to return raw slivers as well?

halfprice: @jpcsmith I merged the PR for now, but let me know what do you think of this

PR #659, opened 2024-08-12
feat(sui-client): add configurable sui network

giac-mysten:
Add the option to pass a custom Sui netowrk for deployment, in the format
`<rpc_url>;<faucet_url>`

### DIFF
--- a/crates/walrus-stress/src/generator.rs
+++ b/crates/walrus-stress/src/generator.rs
@@ -114,7 +114,7 @@ impl LoadGenerator {
         tracing::info!("Spawning gas refill task...");
         let gas_refill_handle = refill_gas(
             addresses,
-            network.clone(),
+            network,
             gas_refill_period,
             metrics.clone(),
             sui_client,


crates/walrus-sui/src/utils.rs
```diff
@@ -299,15 +318,22 @@ impl SuiNetwork {
...
-    pub fn r#type(&self) -> &str {
+    pub fn r#type(&self) -> String {
         match self {
-            SuiNetwork::Localnet => "localnet",
-            SuiNetwork::Devnet => "devnet",
-            SuiNetwork::Testnet => "testnet",
+            SuiNetwork::Localnet => "localnet".to_owned(),
+            SuiNetwork::Devnet => "devnet".to_owned(),
+            SuiNetwork::Testnet => "testnet".to_owned(),
+            SuiNetwork::Custom { rpc, faucet } => format!("custom({};{})", rpc, faucet),
```
karlwuest:
```suggestion
            SuiNetwork::Custom { rpc, faucet } => format!("{};{}", rpc, faucet),
```
Nit: I would remove the `custom()` to make sure that the `FromStr` implementation is compatible with this output.

crates/walrus-stress/src/generator.rs
```diff
@@ -365,7 +370,7 @@ pub fn refill_gas(
                     .len()
                     < MIN_NUM_COINS
                 {
-                    let result = send_faucet_request(address, network).await;
+                    let result = send_faucet_request(address, network_ref.clone()).await;
```
karlwuest: Suggestion: instead of cloning for all faucet requests, `send_faucet_request` and `request_sui_from_faucet` could just take the `SuiNetwork` as a reference.

giac-mysten: ah good point! nicer now.

PR #662, opened 2024-08-13
chore: ensure that reqwest client uses http2

karlwuest:


crates/walrus-sdk/src/client.rs
```diff
@@ -558,6 +558,9 @@ impl Client {
 
         match result {
             Ok(response) => {
+                // Check that the response is using HTTP 2 when running tests.
```
jpcsmith:
```suggestion
                // Check that the response is using HTTP 2 when not in release mode.
```

PR #670, opened 2024-08-14
fix(publisher): enforce sequential Sui transactions

### LINKED ISSUE #577: Publisher cannot handle concurrent requests
The publisher currently uses a single `SuiContractClient` for all requests. 
If two clients try to upload a blob simultaneously, the gas or payment object can get locked which makes the request fail.

mlegner:
Verified this locally by sending 100 simultaneous requests. Before this change, this caused many errors like the following:
```
2024-08-14T15:55:42.428142Z ERROR put_blob{epochs=1}: walrus_service::client::daemon::routes: error storing blob error=client internal error: RPC call failed: ErrorObject { code: ServerError(-32002), message: "Failed to sign transaction by a quorum of validators because of locked objects. Retried a conflicting transaction Some(TransactionDigest(PBbxoZKbhRTnUi1UBfMgWGAvuEzmizFwWGAMrKkXsPt)), success: Some(true)", data: Some(RawValue({"PBbxoZKbhRTnUi1UBfMgWGAvuEzmizFwWGAMrKkXsPt":[["0x2ec6a9f00d3492f889902c6fc926524f2b7539cb9975b6a1f67733eb06ca1376",212,"B1SAu9o26hxTwbDcTofnFF9xzjwDpVVykDmQC7UAer5S"],["0x2ec6a9f00d3492f889902c6fc926524f2b7539cb9975b6a1f67733eb06ca1376",212,"B1SAu9o26hxTwbDcTofnFF9xzjwDpVVykDmQC7UAer5S"],["0x2ec6a9f00d3492f889902c6fc926524f2b7539cb9975b6a1f67733eb06ca1376",212,"B1SAu9o26hxTwbDcTofnFF9xzjwDpVVykDmQC7UAer5S"],["0x2ec6a9f00d3492f889902c6fc926524f2b7539cb9975b6a1f67733eb06ca1376",212,"B1SAu9o26hxTwbDcTofnFF9xzjwDpVVykDmQC7UAer5S"]]})) }
```

With the fix, all 100 requests are handled successfully.

Closes #577

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -260,9 +260,8 @@ impl SuiContractClient {
     }
 
     /// Returns the active address of the client.
-    pub fn address(&mut self) -> anyhow::Result<SuiAddress> {
-        let mut wallet = self.wallet.blocking_lock();
-        wallet.active_address()
+    pub fn address(&self) -> SuiAddress {
+        self.wallet_address
     }
 
     async fn price_for_encoded_length(
--- a/crates/walrus-stress/src/generator.rs
+++ b/crates/walrus-stress/src/generator.rs
@@ -99,11 +99,7 @@ impl LoadGenerator {
 
         let addresses = write_clients
             .iter_mut()
-            .map(|client| {
-                client
-                    .address()
-                    .expect("clients are created with an have an active address")
-            })
+            .map(|client| client.address())
             .collect();
         for write_client in write_clients {
             write_client_pool_tx.send(write_client).await?;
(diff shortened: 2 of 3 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -264,9 +259,10 @@ impl SuiContractClient {
         }
     }
 
-    /// Returns the wallet context used by the client.
-    pub fn wallet(&mut self) -> &mut WalletContext {
-        &mut self.wallet
+    /// Returns the active address of the client.
+    pub fn address(&mut self) -> anyhow::Result<SuiAddress> {
+        let mut wallet = self.wallet.blocking_lock();
+        wallet.active_address()
```
karlwuest:
```suggestion
    pub fn address(&self) -> anyhow::Result<SuiAddress> {
        self.wallet_address
```
Important: `blocking_lock` on a `tokio::sync::Mutex` will panic if called from an async context, so we shouldn't use this in any function that may be called from a tokio task.

PR #671, opened 2024-08-14
feat: Add ability to write and store event blob

sadhansood:
This PR adds functionalities for:

1. Defining the format of an event blob. This can be found in `event_blob.rs`
2. Writing an event blob and storing its slivers locally. This can eb found in `blob_writer.rs`

In the next PR, we will integrate `checkpoint_processor.rs`, `event_blob.rs` and `blob_writer.rs` with the current existing code such that we can start reading events from checkpoints and writing them as blobs and storing them as slivers

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,7 +79,7 @@ version = "0.7.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "891477e0c6a8957309ee5c45a6368af3ae14bb510732d2684ffa19af310920f9"
 dependencies = [
- "getrandom 0.2.14",
+ "getrandom 0.2.15",
  "once_cell",
  "version_check",
 ]
@@ -168,15 +168,15 @@ dependencies = [
  "rand 0.8.5",
  "rcgen",
  "ring 0.17.8",
- "rustls 0.23.10",
- "rustls-webpki 0.102.4",
+ "rustls 0.23.12",
+ "rustls-webpki 0.102.6",
  "serde",
  "serde_json",
- "socket2 0.5.6",
+ "socket2 0.5.7",
  "tap",
  "thiserror",
  "tokio",
- "tokio-util 0.7.10 (registry+https://github.com/rust-lang/crates.io-index)",
+ "tokio-util 0.7.11",
  "tower",
  "tracing",
  "x509-parser",
@@ -188,9 +188,9 @@ version = "0.0.0"
 source = "git+https://github.com/mystenlabs/anemo.git?rev=dbb5a074c2d25660525ab5d36d65ff0cb8051949#dbb5a074c2d25660525ab5d36d65ff0cb8051949"
 dependencies = [
  "prettyplease",
- "proc-macro2 1.0.81",
+ "proc-macro2 1.0.86",
  "quote 1.0.36",
- "syn 2.0.60",
+ "syn 2.0.75",
 ]
 
 [[package]]
@@ -208,7 +208,7 @@ dependencies = [
  "tokio",
  "tower",
  "tracing",
- "uuid 1.8.0",
+ "uuid 1.10.0",
 ]
 
 [[package]]
@@ -219,15 +219,16 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
-version = "0.6.13"
+version = "0.6.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d96bd03f33fe50a863e394ee9718a706f988b9079b20c3784fb726e7678b62fb"
+checksum = "64e15c1ab1f89faffbf04a634d5e1962e9074f2741eef6d97f3c4e322426d526"
 dependencies = [
  "anstyle",
  "anstyle-parse",
  "anstyle-query",
  "anstyle-wincon",
  "colorchoice",
+ "is_terminal_polyfill",
  "utf8parse",
 ]
 
@@ -239,27 +240,27 @@ checksum = "1bec1de6f59aedf83baf9ff929c98f2ad654b97c9510f4e70cf6f661d49fd5b1"
 
 [[package]]
 name = "anstyle-parse"
-version = "0.2.3"
+version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c75ac65da39e5fe5ab759307499ddad880d724eed2f6ce5b5e8a26f4f387928c"
+checksum = "eb47de1e80c2b463c735db5b217a0ddc39d612e7ac9e2e96a5aed1f57616c1cb"
 dependencies = [
  "utf8parse",
 ]
 
 [[package]]
 name = "anstyle-query"
-version = "1.0.2"
+version = "1.1.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e28923312444cdd728e4738b3f9c9cac739500909bb3d3c94b43551b16517648"
+checksum = "6d36fc52c7f6c869915e99412912f22093507da8d9e942ceaf66fe4b7c14422a"
 dependencies = [
  "windows-sys 0.52.0",
 ]
 
 [[package]]
 name = "anstyle-wincon"
-version = "3.0.2"
+version = "3.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1cd54b81ec8d6180e24654d0b371ad22fc3dd083b6ff8ba325b72e00c87660a7"
+checksum = "5bf74e1b6e971609db8ca7a9ce79fd5768ab6ae46441c572e46cf596f59e57f8"
 dependencies = [
  "anstyle",
  "windows-sys 0.52.0",
@@ -354,7 +355,7 @@ dependencies = [
  "derivative",
  "digest 0.10.7",
  "itertools 0.10.5",
- "num-bigint 0.4.4",
+ "num-bigint 0.4.6",
  "num-traits",
  "paste",
  "rustc_version",
@@ -377,9 +378,9 @@ version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7abe79b0e4288889c4574159ab790824d0033b9fdcb2a112a3182fac2e514565"
 dependencies = [
- "num-bigint 0.4.4",
+ "num-bigint 0.4.6",
  "num-traits",
- "proc-macro2 1.0.81",
+ "proc-macro2 1.0.86",
  "quote 1.0.36",
  "syn 1.0.109",
 ]
@@ -443,7 +444,7 @@ dependencies = [
  "ark-serialize-derive",
  "ark-std",
  "digest 0.10.7",
- "num-bigint 0.4.4",
+ "num-bigint 0.4.6",
 ]
 
 [[package]]
@@ -452,7 +453,7 @@ version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ae3281bc6d0fd7e549af32b52511e1302185bd688fd3359fa36423346ff682ea"
 dependencies = [
- "proc-macro2 1.0.81",
+ "proc-macro2 1.0.86",
  "quote 1.0.36",
  "syn 1.0.109",
 ]
@@ -481,15 +482,15 @@ dependencies = [
 
 [[package]]
 name = "arrayref"
-version = "0.3.7"
+version = "0.3.8"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6b4930d2cb77ce62f89ee5d5289b4ac049559b1c45539271f5ed4fdc7db34545"
+checksum = "9d151e35f61089500b617991b791fc8bfd237ae50cd5950803758a179b41e67a"
 
 [[package]]
 name = "arrayvec"
-version = "0.7.4"
+version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "96d30a06541fbafbc7f82ed10c06164cfbd2c401138f6addd8404629c4b16711"
+checksum = "7c02d123df017efcdfbd739ef81735b36c5ba83ec3c59c80a9d7ecc718f92e50"
 
 [[package]]
 name = "ascii-canvas"
@@ -522,7 +523,7 @@ version = "0.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "726535892e8eae7e70657b4c8ea93d26b8553afb1ce617caee529ef96d7dee6c"
 dependencies = [
- "proc-macro2 1.0.81",
+ "proc-macro2 1.0.86",
  "quote 1.0.36",
  "syn 1.0.109",
  "synstructure",
@@ -534,7 +535,7 @@ version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2777730b2039ac0f95f093556e61b6d26cebed5393ca6f152717777cec3a42ed"
 dependencies = [
- "proc-macro2 1.0.81",
+ "proc-macro2 1.0.86",
  "quote 1.0.36",
  "syn 1.0.109",
 ]
@@ -581,9 +582,9 @@ version = "0.3.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "16e62a023e7c117e27523144c5d2459f4397fcc3cab0085af8e2224f643a0193"
 dependencies = [
- "proc-macro2 1.0.81",
+ "proc-macro2 1.0.86",
  "quote 1.0.36",
- "syn 2.0.60",
+ "syn 2.0.75",
 ]
 
 [[package]]
@@ -597,9 +598,9 @@ version = "0.1.81"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6e0c28dcc82d7c8ead5cb13beb15405b57b8546e93215673ff8ca0349a028107"
 dependencies = [
- "proc-macro2 1.0.81",
+ "proc-macro2 1.0.86",
  "quote 1.0.36",
- "syn 2.0.60",
+ "syn 2.0.75",
 ]
 
 [[package]]
@@ -620,6 +621,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2ce4f10ea3abcd6617873bae9f91d1c5332b4a778bd9ce34d0cd517474c1de82"
 
 [[package]]
+name = "atomic-waker"
+version = "1.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
+
+[[package]]
 name = "atomic_float"
 version = "0.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -631,9 +638,9 @@ version = "1.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3c87f3f15e7794432337fc718554eaa4dc8f04c9677a950ffe366f20a162ae42"
 dependencies = [
- "proc-macro2 1.0.81",
+ "proc-macro2 1.0.86",
  "quote 1.0.36",
- "syn 2.0.60",
+ "syn 2.0.75",
 ]
 
 [[package]]
@@ -644,9 +651,9 @@ checksum = "7460f7dd8e100147b82a63afca1a20eb6c231ee36b90ba7272e14951cb58af59"
 
 [[package]]
 name = "autocfg"
-version = "1.2.0"
+version = "1.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f1fdabc7756949593fe60f30ec81974b613357de856987752631dea1e3394c80"
+checksum = "0c4b4d0bd25bd0b74681c0ad21497610ce1b7c91b1022cd21c80c6fbdd9476b0"
 
 [[package]]
 name = "aws-config"
@@ -711,14 +718,14 @@ dependencies = [
  "percent-encoding",
  "pin-project-lite",
  "tracing",
- "uuid 1.8.0",
+ "uuid 1.10.0",
 ]
 
 [[package]]
 name = "aws-sdk-ec2"
-version = "1.65.0"
+version = "1.67.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1354c70a43d4a3eae02a912bde565853c159cc6af6cb210b82a354f0cf714a33"
+checksum = "80e9e023bd1b9aa490f9495dc26a96b340dae08388d9b41fee269ed64b2f57a9"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -740,9 +747,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-sso"
-version = "1.37.0"
+version = "1.39.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "1074e818fbe4f9169242d78448b15be8916a79daa38ea1231f2e2e10d993fcd2"
+checksum = "11822090cf501c316c6f75711d77b96fba30658e3867a7762e5e2f5d32d31e81"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -762,9 +769,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-ssooidc"
-version = "1.38.0"
+version = "1.40.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "29755c51e33fa3f678598f64324a169cf4b7d3c4865d2709d4308f53366a92a4"
+checksum = "78a2a06ff89176123945d1bbe865603c4d7101bea216a550bb4d2e4e9ba74d74"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -784,9 +791,9 @@ dependencies = [
 
 [[package]]
 name = "aws-sdk-sts"
-version = "1.37.0"
+version = "1.39.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6e52dc3fd7dfa6c01a69cf3903e00aa467261639138a05b06cd92314d2c8fb07"
+checksum = "a20a91795850826a6f456f4a48eff1dfa59a0e69bdbf5b8c50518fd372106574"
 dependencies = [
  "aws-credential-types",
  "aws-runtime",
@@ -880,9 +887,9 @@ dependencies = [
 
 [[package]]
 name = "aws-smithy-runtime"
-version = "1.6.2"
+version = "1.6.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ce87155eba55e11768b8c1afa607f3e864ae82f03caf63258b37455b0ad02537"
+checksum = "0abbf454960d0db2ad12684a1640120e7557294b0ff8e2f11236290a1b293225"
 dependencies = [
  "aws-smithy-async",
  "aws-smithy-http",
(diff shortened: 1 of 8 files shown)


crates/walrus-event/Cargo.toml
mlegner, line 1: Suggestion: Keep the versions in the workspace `Cargo.toml` file and only add `some-dependency.workspace = true` here.

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+pub const EVENT_BLOB_MAGIC: u32 = 0x0000DEAD;
+pub const MAGIC_BYTES_SIZE: usize = std::mem::size_of::<u32>();
+pub const EVENT_BLOB_FORMAT_VERSION: u32 = 1;
+pub const BLOB_FORMAT_BYTES_SIZE: usize = std::mem::size_of::<u32>();
+
+#[derive(Copy, Clone, Debug, Eq, PartialEq, TryFromPrimitive, IntoPrimitive)]
+#[repr(u8)]
+pub enum EntryEncoding {
+    Bcs = 1,
+}
```
mlegner: Remark: So far we've manually implemented the conversions in various places. I'd prefer not adding a new dependency (`num_enum`) just for this.

However, if we already add a new crate, we should use it consistently throughout our codebase.

crates/walrus-sdk/src/error.rs
```diff
@@ -29,6 +29,16 @@ impl NodeError {
...
     }
 
+    /// Returns true if the HTTP error status code associated with the error is
+    /// [`StatusCode::MISDIRECTED_REQUEST`].
+    pub fn is_shard_not_assigned(&self) -> bool {
+        if let Kind::Reqwest(inner) = &self.kind {
+            inner.status() == Some(StatusCode::MISDIRECTED_REQUEST)
+        } else {
+            false
+        }
```
mlegner: Question: Why not this shorter version?
```suggestion
        self.http_status_code() == Some(StatusCode::MISDIRECTED_REQUEST)
```

deny.toml
mlegner, line 1: Remark: These changes should no longer be needed after rebasing on the latest `main`.

crates/walrus-event/src/checkpoint_processor.rs
```diff
@@ -458,6 +500,7 @@ mod tests {
             let event = CheckpointSuiEvent {
                 serialized_event: serde_json::to_vec(&SuiEvent::random_for_testing()).unwrap(),
                 checkpoint_sequence_number: CheckpointOrderedEventID::new(0, i),
+                is_end_of_walrus_epoch_event: false,
             };
```
mlegner: Suggestion: We seem to have the exact same event definition in at least 4 places. Can we factor this out into a separate function?

crates/walrus-event/src/checkpoint_processor.rs
```diff
@@ -214,6 +254,8 @@ impl CheckpointProcessor {
                     let checkpoint_event = CheckpointSuiEvent {
                         serialized_event: serde_json::to_vec(&sui_event)?,
                         checkpoint_sequence_number: checkpoint_event_id.clone(),
+                        // TODO: Set this to true if this is epoch closing event
```
mlegner, line 293: Please add a reference to a GitHub issue for all TODOs that remain when merging the PR. 🙏

crates/walrus-event/src/checkpoint_processor.rs
```diff
@@ -40,22 +47,55 @@ const EVENT_STORE: &str = "event_store";
 const MAX_TIMEOUT: Duration = Duration::from_secs(60);
 const RETRY_DELAY: Duration = Duration::from_secs(5);
 
-#[derive(Eq, PartialEq, Clone, Debug, Serialize, Deserialize)]
+#[derive(Eq, PartialEq, Default, Clone, Debug, Serialize, Deserialize)]
 pub struct CheckpointOrderedEventID {
-    sequence: CheckpointSequenceNumber,
-    counter: u64,
+    pub sequence: CheckpointSequenceNumber,
+    pub counter: u64,
```
mlegner: Nit: Could you describe the meaning of these fields?

Also: The name "sequence" is a bit misleading as it refers to a sequence *number* and not a sequence.

crates/walrus-event/src/checkpoint_processor.rs
```diff
@@ -40,22 +47,55 @@ const EVENT_STORE: &str = "event_store";
...
+                sequence: self.sequence.saturating_sub(1),
+                counter: u64::MAX,
+            }
+        } else {
+            CheckpointOrderedEventID {
+                sequence: self.sequence,
+                counter: self.counter.saturating_sub(1),
+            }
+        }
+    }
```
mlegner: Question: What is the intended semantics of this function?

crates/walrus-event/src/checkpoint_processor.rs
```diff
@@ -40,22 +47,55 @@ const EVENT_STORE: &str = "event_store";
...
+    pub(crate) fn read(rbuf: &mut BufReader<File>) -> Result<CheckpointOrderedEventID> {
+        let sequence = rbuf.read_u64::<BigEndian>()?;
+        let counter = rbuf.read_u64::<BigEndian>()?;
+        Ok(CheckpointOrderedEventID::new(sequence, counter))
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn size() -> usize {
+        std::mem::size_of::<CheckpointSequenceNumber>() + std::mem::size_of::<u64>()
+    }
```
mlegner: Suggestion: Given this is a constant, can we define a constant `SIZE` instead of a method?

crates/walrus-event/src/event_blob.rs
mlegner, line 1: General suggestion: Can you add some docstrings to the various constants, structs, and functions? It is not always immediately obvious to me what the purpose or intended semantics are.

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+        let len = rbuf.read_varint::<u64>()? as usize;
+        if len == 0 {
+            return Err(anyhow!("Invalid object length of 0 in file"));
+        }
+        let encoding = rbuf.read_u8()?;
+        let mut data = vec![0u8; len];
+        rbuf.read_exact(&mut data)?;
+        let entry = BlobEntry {
+            data,
+            encoding: EntryEncoding::try_from(encoding)?,
```
mlegner: Nit: If we move the `try_from(..)?` five lines up, we can fail early (before allocating space for the data) in case of an unknown encoding.

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    pub fn decode(self) -> Result<CheckpointSuiEvent> {
+        let data = match &self.encoding {
+            EntryEncoding::Bcs => self.data,
+        };
+        let res = bcs::from_bytes(&data)?;
+        Ok(res)
+    }
+
+    pub fn read<R: Read>(rbuf: &mut R) -> Result<BlobEntry> {
```
mlegner: Nit:
```suggestion
    pub fn read<R: Read>(rbuf: &mut R) -> Result<Self> {
```

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+use walrus_core::BlobId;
+
+use crate::checkpoint_processor::{CheckpointOrderedEventID, CheckpointSuiEvent};
+
+pub const MAX_VARINT_LENGTH: usize = 10;
+pub const EVENT_ENCODING_BYTES: usize = 1;
+pub const EVENT_BLOB_MAGIC: u32 = 0x0000DEAD;
+pub const MAGIC_BYTES_SIZE: usize = std::mem::size_of::<u32>();
+pub const EVENT_BLOB_FORMAT_VERSION: u32 = 1;
+pub const BLOB_FORMAT_BYTES_SIZE: usize = std::mem::size_of::<u32>();
```
mlegner: Suggestion: As most of these constants are only used in a single struct, can we move them there (and potentially rename them)?

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+        let mut counter = 0;
+        let n = (self.data.len() as u64).encode_var(&mut buf);
+        wbuf.write_all(&buf[0..n])?;
+        counter += n;
+        buf[0] = self.encoding.into();
+        wbuf.write_all(&buf[0..EVENT_ENCODING_BYTES])?;
+        counter += 1;
+        wbuf.write_all(&self.data)?;
+        counter += self.data.len();
+        Ok(counter)
```
mlegner: Suggestion: How about this more compact version without the need of a counter and reducing the number of writes?
```suggestion
        let mut buf = [0u8; MAX_VARINT_LENGTH + EVENT_ENCODING_BYTES];
        let n = (self.data.len() as u64).encode_var(&mut buf);
        buf[n] = self.encoding.into();
        wbuf.write_all(&buf[0..n + EVENT_ENCODING_BYTES])?;
        wbuf.write_all(&self.data)?;
        Ok(n + EVENT_ENCODING_BYTES + self.data.len())
```

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    pub fn size(&self) -> usize {
+        let mut blob_size = self.data.len().required_space();
+        blob_size += EVENT_ENCODING_BYTES;
+        blob_size += self.data.len();
+        blob_size
```
mlegner: Nit: Maybe a matter of taste, but I feel this is clearer without the reassignments:
```suggestion
        self.data.len().required_space() + EVENT_ENCODING_BYTES + self.data.len()
```

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    #[allow(dead_code)]
+    pub(crate) fn from_bytes(bytes: &[u8]) -> Result<CheckpointSuiEvent> {
+        let (encoding, data) = bytes.split_first().ok_or(anyhow!("empty bytes"))?;
+        BlobEntry {
+            data: data.to_vec(),
+            encoding: EntryEncoding::try_from(*encoding)?,
+        }
+        .decode()
+    }
```
mlegner: Suggestion: Remove the `.decode()` here to make this more versatile. Or, alternatively, call this something like `decode_event_from_bytes`.

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+#[allow(dead_code)]
+/// An iterator over blobs in a blob file.
```
mlegner: Question: Should this say "events" and "event" instead of "blobs" and "blob"?
```suggestion
/// An iterator over events in an event file.
```

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,201 @@
...
+
+    #[allow(dead_code)]
+    fn tail_size() -> usize {
+        2 * CheckpointOrderedEventID::size() + BlobId::LENGTH
+    }
+
+    #[allow(dead_code)]
+    fn min_blob_size() -> usize {
+        Self::header_size() + Self::tail_size()
+    }
```
mlegner: Question: Why are these functions and not simply constants?

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,356 @@
...
+            buf_offset: MAGIC_BYTES_SIZE,
+            commit_file_size,
+            prev_blob_id,
+            client,
+            encoding_config,
+            skip_storing_slivers: false,
+        })
+    }
+
+    fn new_for_testing(root_dir_path: PathBuf, commit_file_size: usize) -> Result<Self> {
```
mlegner: Nit: This could be defined only for tests and the `test-utils` feature.
```suggestion
    #[cfg(any(test, feature = "test-utils"))]
    fn new_for_testing(root_dir_path: PathBuf, commit_file_size: usize) -> Result<Self> {
```

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,356 @@
...
+        )?;
+        self.buf_offset = MAGIC_BYTES_SIZE + BLOB_FORMAT_BYTES_SIZE;
+        self.wbuf = BufWriter::new(f);
+        Ok(())
+    }
+
+    async fn store_slivers(&mut self, content: &[u8]) -> Result<()> {
+        let blob_encoder = self.encoding_config.get_blob_encoder(content)?;
+        let blob_metadata = blob_encoder.compute_metadata();
+        let sliver_pairs = blob_encoder.encode();
```
mlegner: Important: This is essentially does the encoding twice. The following does it in one step:
```suggestion
        let (sliver_pairs, blob_metadata) = blob_encoder.encode_with_metadata();
```

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,356 @@
...
+                .or_else(|e| {
+                    if e.is_shard_not_assigned() {
+                        Ok(())
+                    } else {
+                        Err(e)
+                    }
+                })
+        }))
+        .await
+        .map(|_| ())?;
```
mlegner, line 243: Suggestion: Filter the slivers at the beginning and only attempt to store the ones for which the node is responsible.

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,356 @@
...
+            .open(next_file_path)?;
+        f.seek(SeekFrom::Start(n as u64))?;
+        Ok(f)
+    }
+
+    fn finalize(&mut self) -> Result<()> {
+        let start_checkpoint = self.start.as_ref().cloned().unwrap_or_default();
+        let end_checkpoint = self.end.as_ref().cloned().unwrap_or_default();
+        start_checkpoint.write(&mut self.wbuf)?;
+        end_checkpoint.write(&mut self.wbuf)?;
```
mlegner: Suggestion: Implement `Copy` for `CheckpointOrderedEventId` and then shorten this:
```suggestion
        self.start.unwrap_or_default().write(&mut self.wbuf)?;
        self.end.unwrap_or_default().write(&mut self.wbuf)?;
```

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,356 @@
...
+        if self.skip_storing_slivers {
+            // copy file to /tmp for testing purposes
+            let mut tmp_file = File::create("/tmp/event_blob")?;
+            tmp_file.write_all(&file_content)?;
+            tmp_file.flush()?;
+            return Ok(());
+        }
+        self.store_slivers(&file_content).await?;
+        Ok(())
+    }
```
mlegner: Suggestion: Instead of testing-only fields and branches, could we just have two versions of this `cut` function? One for the "standard" behavior and one for the tests?

crates/walrus-event/src/blob_writer.rs
sadhansood: This is a todo (which i added) as i think it'd be easier to do so once we cache the shard assignments from epoch change event

Cargo.lock
mlegner, line 1: Important: This should not be deleted.

Cargo.toml
```diff
@@ -45,6 +48,7 @@ reqwest = { version = "0.12.5", default-features = false, features = [
 ] }
 rocksdb = "0.21.0"
 serde = { version = "1.0.207", features = ["derive"] }
+serde_as = "0.0.1"
```
mlegner: Chore: This should be removed.
```suggestion
```

crates/walrus-event/Cargo.toml
```diff
@@ -9,11 +9,17 @@ license.workspace = true
...
 async-trait.workspace = true
 bcs.workspace = true
+byteorder.workspace = true
+futures-util .workspace = true
+integer-encoding.workspace = true
 move-core-types.workspace = true
+reqwest.workspace = true
 rocksdb.workspace = true
 serde = { workspace = true, features = ["derive"] }
+serde_as.workspace = true
```
mlegner:
```suggestion
```

crates/walrus-event/Cargo.toml
```diff
@@ -9,11 +9,17 @@ license.workspace = true
 anyhow.workspace = true
 async-trait.workspace = true
 bcs.workspace = true
+byteorder.workspace = true
+futures-util .workspace = true
```
mlegner: Nit:
```suggestion
futures-util.workspace = true
```

crates/walrus-event/Cargo.toml
```diff
@@ -26,6 +32,10 @@ tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
 tokio-stream.workspace = true
 tokio-util.workspace = true
 typed-store.workspace = true
+url.workspace = true
+walrus-core.workspace = true
+walrus-sdk.workspace = true
+walrus-service = { workspace = true, features = ["node"] }
```
mlegner, line 37: Question: Why do we need the `walrus-service` crate? And won't this cause circular dependencies once we use the `walrus-event` crate in `walrus-service`?

crates/walrus-event/src/checkpoint_processor.rs
mlegner: Here my idea was something like a `default_event_for_testing()` function.

crates/walrus-event/src/checkpoint_processor.rs
mlegner: Thanks for the change. Can we make this an associated constant for `WalrusEventID`? Like `WalrusEventID::SIZE`?

crates/walrus-event/Cargo.toml
sadhansood: We are using the `StorageNodeConfig` to initialize the node client. This `StorageNodeConfig` is defined in the `walrus-service` crate. If we run into circular dependency, i'll refactor this config out to break the dependency.

crates/walrus-event/src/checkpoint_processor.rs
```diff
@@ -223,6 +301,12 @@ impl CheckpointProcessor {
                         .map_err(|e| anyhow!("Failed to insert event into event store: {}", e))?;
                 }
             }
+            let end_of_checkpoint =
+                WalrusEvent::new_end_of_checkpoint(checkpoint.checkpoint_summary.sequence_number);
```
halfprice, line 305: If the checkpoint does not contain any Walrus events, do we still write this end of checkpoint marker? If so, we may want to consider skip such marker, since with Mysticeti, the checkpoint rate can be in 10~20 checkpoints / second range.

Also, is the purpose of end of checkpoint marker purely for convenient purpose, since we can always find checkpoint boundary by looking at checkpoint number.

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,215 @@
...
+    total_size: u64,
+    prev_blob_id: BlobId,
+    start: CheckpointSequenceNumber,
+    end: CheckpointSequenceNumber,
+    blob_format_version: u32,
+}
+
+impl EventBlob {
+    /// The magic bytes at the start of the blob file.
+    pub const EVENT_BLOB_MAGIC: u32 = 0x0000DEAD;
```
halfprice: Curious what's the purpose of this magic?

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,215 @@
...
+                "Unsupported blob format version",
+            )));
+        }
+        let tail_size = Self::EVENT_BLOB_TAIL_SIZE as i64;
+        total_size -= tail_size as u64;
+        buf_reader.seek(SeekFrom::End(-tail_size))?;
+        let start = buf_reader.read_u64::<byteorder::BigEndian>()?;
+        let end = buf_reader.read_u64::<byteorder::BigEndian>()?;
+        let mut prev_blob_id = [0u8; BlobId::LENGTH];
+        buf_reader.read_exact(&mut prev_blob_id)?;
```
halfprice, line 175: What if the file doesn't contain any blobs, is this read still valid?

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,392 @@
...
+
+    fn finalize(&mut self) -> Result<()> {
+        self.wbuf
+            .write_u64::<BigEndian>(self.start.unwrap_or_default())?;
+        self.wbuf
+            .write_u64::<BigEndian>(self.end.unwrap_or_default())?;
+        self.wbuf.write_all(
+            &self
+                .prev_blob_id
```
halfprice, line 194: Question: what's the purpose of storing previous blob id at the end of the file?

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,392 @@
...
+        )?;
+        self.buf_offset = EventBlob::MAGIC_BYTES_SIZE + EventBlob::BLOB_FORMAT_BYTES_SIZE;
+        self.wbuf = BufWriter::new(f);
+        Ok(())
+    }
+
+    async fn store_slivers(&mut self, content: &[u8]) -> Result<()> {
+        let blob_encoder = self.encoding_config.get_blob_encoder(content)?;
+        let (sliver_pairs, blob_metadata) = blob_encoder.encode_with_metadata();
+        self.client.store_metadata(&blob_metadata).await?;
```
halfprice, line 222: This blob is not required to be registered on chain first?

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,392 @@
...
+        // TODO: Once shard assignment per storage node will be read from walrus
+        // system object at the beginning of the walrus epoch, we can only store the blob for
+        // shards that are locally assigned to this node
```
halfprice, line 225: Can you elaborate what does "only store ... locally" mean? Since this is a regular blob, won't that the slivers for other shards also need to be sent to other nodes?

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,392 @@
...
+        .await
+        .map(|_| ())?;
+        // TODO: Once slivers are stored locally, invoke the new sui contract endpoint to certify the
+        // event blob
+        let event_blob_metadata = EventBlobMetadata {
+            blob_id: *blob_metadata.blob_id(),
+            start: self.start.as_ref().cloned().unwrap_or_default(),
+            end: self.end.as_ref().cloned().unwrap_or_default(),
+        };
+        self.prev_blob_id.insert(&(), &event_blob_metadata)?;
```
halfprice, line 269: Nit: maybe move this part to the end of `cut`, since they are not about `store_slivers`

Also, is the purpose of storing prev_blob_id in a DBMap for crash-recovery purpose?

crates/walrus-event/src/checkpoint_processor.rs
sadhansood: Good question @halfprice, the checkpoint marker is to ensure the liveness of the event blob stream. If we cut the blobs by size the last blob may not get committed for a long time if there are no walrus events. We discussed this problem during our design session and agreed upon committing -  a blob by size, at the end of walrus epoch and at the end of sui epoch. So, instead of doing that since i thought end of sui epoch is still too long a duration, we now commit - at the end of walrus epoch and at the end of every N sui checkpoints. Sure this means some of the event blobs could be empty but this keeps the event blob stream alive and making progress. We can make N large enough to maybe commit once every 10 minutes?

crates/walrus-event/src/event_blob.rs
sadhansood: This is just to quickly verify the event blob file and a basic integrity check that it isn't corrupted.

crates/walrus-event/src/event_blob.rs
sadhansood: Yes, if the file doesn't contain any blob - it'd still contain a prev blob id

crates/walrus-event/src/blob_writer.rs
sadhansood: This is because once we read the latest blob, we can get a reference chain of all event blobs

crates/walrus-event/src/blob_writer.rs
sadhansood: No, the plan is to invoke another sui contract endpoint to certify the event blob directly (without registering it first) and that would endpoint would also take care of updating the walrus system object with latest event blob id

crates/walrus-event/src/blob_writer.rs
sadhansood: Yes, but every node will store the slivers for its locally assigned shard and invoke the new event blob certify contract endpoint (which would send certified events once majority have stored it locally)

crates/walrus-event/src/blob_writer.rs
sadhansood: Yes, it allows the node to start the event stream from the last event in the `prev_blob` after a crash or restart

crates/walrus-event/src/checkpoint_processor.rs
mlegner: I don't see the constant anymore. Has this been lost during the refactoring?

crates/walrus-event/src/blob_writer.rs
```diff
@@ -0,0 +1,392 @@
...
+        let mut metab = [0u8; EventBlob::BLOB_FORMAT_BYTES_SIZE];
+        BigEndian::write_u32(&mut metab, blob_format_version);
+        n += f.write(&metab)?;
+        drop(f);
+        f = OpenOptions::new()
+            .append(true)
+            .read(true)
+            .open(next_file_path)?;
+        f.seek(SeekFrom::Start(n as u64))?;
+        Ok(f)
```
mlegner: This function seems unnecessarily complicated. How about this?
```suggestion
        let next_file_path = dir_path.join("current_blob");
        let mut file = File::create(&next_file_path)?;
        file.write_u32::<BigEndian>(magic_bytes)?;
        file.write_u32::<BigEndian>(blob_format_version)?;
        drop(file);
        file = OpenOptions::new()
            .append(true)
            .read(true)
            .open(next_file_path)?;
        file.seek(SeekFrom::Start(EventBlob::EVENT_BLOB_HEADER_SIZE as u64))?;
        Ok(file)
```

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// The size of the format version in bytes.
+    pub const BLOB_FORMAT_BYTES_SIZE: usize = std::mem::size_of::<u32>();
+    /// The size of the header of the blob file in bytes.
+    pub const EVENT_BLOB_HEADER_SIZE: usize =
+        EventBlob::MAGIC_BYTES_SIZE + EventBlob::BLOB_FORMAT_BYTES_SIZE;
+    /// The size of the tail of the blob file in bytes.
+    pub const EVENT_BLOB_TAIL_SIZE: usize = 2 * std::mem::size_of::<u64>() + BlobId::LENGTH;
+    /// The minimum size of a blob file in bytes.
+    pub const EVENT_BLOB_MIN_SIZE: usize =
+        EventBlob::EVENT_BLOB_HEADER_SIZE + EventBlob::EVENT_BLOB_TAIL_SIZE;
```
mlegner: Nit: Given these constants are associated with the `EventBlob` struct, the `EVENT_BLOB_` prefix is unnecessary.

crates/walrus-event/src/event_blob.rs
```diff
@@ -0,0 +1,215 @@
...
+    pub const MAGIC_BYTES_SIZE: usize = std::mem::size_of::<u32>();
+    /// The format version of the blob file.
+    pub const EVENT_BLOB_FORMAT_VERSION: u32 = 1;
+    /// The size of the format version in bytes.
+    pub const BLOB_FORMAT_BYTES_SIZE: usize = std::mem::size_of::<u32>();
+    /// The size of the header of the blob file in bytes.
+    pub const EVENT_BLOB_HEADER_SIZE: usize =
+        EventBlob::MAGIC_BYTES_SIZE + EventBlob::BLOB_FORMAT_BYTES_SIZE;
+    /// The size of the tail of the blob file in bytes.
+    pub const EVENT_BLOB_TAIL_SIZE: usize = 2 * std::mem::size_of::<u64>() + BlobId::LENGTH;
```
mlegner: Nit: For me, the "opposite" of "header" is "[trailer](https://en.wikipedia.org/wiki/Trailer_(computing))".

Issue #560, opened 2024-06-20
Major Epic: Epoch change


gdanezis:
Related issues (incomplete):
- #940 
- #941

PR #673, opened 2024-08-15
feat(storage): shard sync skeleton in the destination node

halfprice:
The basic framework for fetching blobs in the new shard owner. This PR is more focusing on the framework and only does the bare minimum in doing the actual syncing work (simply scan all the certified blobs in the src node and store it locally).

Suggestions are welcome!

Related to #259 #560

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -332,7 +332,7 @@ impl StorageNode {
         committee_service_factory: Box<dyn CommitteeServiceFactory>,
         contract_service: Box<dyn SystemContractService>,
         registry: &Registry,
-        pre_created_storage: Option<Storage>, // For testing purposes. TODO(#703): consider this test only input.
+        pre_created_storage: Option<Storage>, // For testing purposes. TODO(#703): remove.
     ) -> Result<Self, anyhow::Error> {
         let start_time = Instant::now();
         let committee_service = committee_service_factory
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -59,12 +59,14 @@ impl ShardSyncHandler {
         }
     }
 
-    /// Restarts syncing shards that were previously syncing. This method is used when restarting the node.
+    /// Restarts syncing shards that were previously syncing. This method is used when restarting
+    /// the node.
     pub async fn restart_syncs(&self) -> Result<(), anyhow::Error> {
         for shard_index in self.node.storage.shards() {
             let shard_storage = self.node.storage.shard_storage(shard_index).unwrap();
 
-            // Restart the syncing task for shards that were previously syncing (in ActiveSync status).
+            // Restart the syncing task for shards that were previously syncing (in ActiveSync
+            // status).
             if shard_storage.status()? == ShardStatus::ActiveSync {
                 self.start_shard_sync_impl(shard_storage.clone()).await;
             }
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -335,7 +335,8 @@ impl ShardStorage {
 
         assert_eq!(self.status()?, ShardStatus::ActiveSync);
 
-        // TODO(#705): handle better crash recovery (we shouldn't always restart from beginning to sync).
+        // TODO(#705): handle better crash recovery (we shouldn't always restart from beginning
+        //             to sync).
         // TODO(#705): handle missing individual blobs.
         // TODO(#259): handle non-happy path.
         self.sync_shard_before_epoch_internal(epoch, node.clone(), SliverType::Primary)


crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -283,10 +294,144 @@ impl ShardStorage {
...
+        &self,
+        node: Arc<StorageNodeInner>,
+    ) -> Result<(), SyncShardError> {
+        tracing::info!("Syncing shard to epoch: {}", node.current_epoch());
+        // TODO: make sure that the shard is in a valid status to sync.
+        self.shard_status
+            .insert(&(), &ShardStatus::ActiveSync)
+            .context("Update shard status encountered error")?;
+
+        // TODO: handle crash recovery.
```
halfprice: I put quite a few such TODOs in the code, which should be addressed soon. Let me know if creating issues for them are still prefered.

crates/walrus-service/src/node.rs
```diff
@@ -1919,4 +1950,94 @@ mod tests {
...
+
+    // Tests the basic `sync_shard` API.
+    #[tokio::test]
+    async fn sync_shard_client_success() -> TestResult {
+        telemetry_subscribers::init_for_testing();
+
+        let (cluster, _, blob_details) = cluster_with_initial_epoch_and_certified_blob(
+            &[&[0], &[1]],
+            &[
+                &[1; 32], &[2; 32], &[3; 32], &[4; 32], &[5; 32], &[6; 32], &[7; 32], &[8; 32],
```
halfprice: Sorry, I'm too dumb to make this nicer...

crates/walrus-service/src/node.rs
```diff
@@ -332,20 +327,35 @@ struct StorageNodeInner {
...
-        mut storage: Storage,
         event_provider: Box<dyn SystemEventProvider>,
         committee_service_factory: Box<dyn CommitteeServiceFactory>,
         contract_service: Box<dyn SystemContractService>,
         registry: &Registry,
-        max_concurrent_blob_syncs: usize,
+        pre_created_storage: Option<Storage>, // For testing purposes.
     ) -> Result<Self, anyhow::Error> {
         let start_time = Instant::now();
         let committee_service = committee_service_factory
```
halfprice, line 338: Not sure whether this has been thought before. Right now, committee_service only contains committee for 1 epoch. When there are multiple epochs, do we envision this committee_service to store committees for multiple epochs, or we will have something like Map<epoch, CommitteeService>?

crates/walrus-service/src/node/storage/shard.rs
mlegner: Maybe just an "umbrella" issue that covers all these TODOs?

crates/walrus-sui/src/types.rs
```diff
@@ -414,6 +414,13 @@ impl Committee {
         }
         unreachable!("threshold < n_shards")
     }
+
+    /// Returns the index of the member that holds the specified shard.
+    pub fn member_index_for_shard(&self, shard: ShardIndex) -> Option<usize> {
+        self.members
+            .iter()
+            .position(|node| node.shard_ids.contains(&shard))
+    }
```
mlegner: Important: This should never fail if `shard < n_shards`, right? We should _never_ have an unassigned shard.

Suggestion: Define an invariant to `Committee` that all shards are mapped to exactly one storage node and enforce this in the `Committee::new` function. Then we can just `.expect` here and return a `usize`.

crates/walrus-service/src/node.rs
jpcsmith: Something of the latter, but behind the `CommitteeService` interface. So during the transition a new committee service would be created for the new committee, but these could be referenced through the same interface, which directs reads towards the old and writes towards the new; until the next event that just removes the old committee.

crates/walrus-service/src/node/committee.rs
```diff
@@ -129,6 +138,17 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
         n_shards: NonZeroU16,
     ) -> InvalidBlobCertificate;
 
+    /// Syncs a shard to the given epoch.
+    async fn sync_shard_to_epoch(
```
mlegner: Question: What does "to the epoch" mean here?

crates/walrus-service/src/node.rs
mlegner: Not sure if this is nicer: 🤷

```rust
        let blobs: Vec<[u8; 32]> = (1..24).map(|i| [i; 32]).collect();
        let blobs: Vec<_> = blobs.iter().map(|b| &b[..]).collect();
        let (cluster, _, blob_details) =
            cluster_with_initial_epoch_and_certified_blob(&[&[0], &[1]], &blobs, 0).await?;
```

crates/walrus-service/src/node/committee.rs
```diff
@@ -556,6 +585,32 @@ where
...
+        &self,
+        shard: ShardIndex,
+        starting_blob_id: BlobId,
+        sliver_count: u64,
+        epoch: Epoch,
+        key_pair: &ProtocolKeyPair,
+    ) -> Result<Vec<(BlobId, GeneralSliver)>, anyhow::Error> {
+        let Some(member_index) = self.committee.member_index_for_shard(shard) else {
+            return Err(SyncShardError::NoOwnerForShard(shard).into());
+        };
```
mlegner, line 600: This error should never actually happen, see my other comment.

crates/walrus-service/src/node/committee.rs
```diff
@@ -556,6 +585,32 @@ where
...
+
+        if let Some(client) = node.client() {
+            Ok(client
+                .sync_shard::<A>(shard, starting_blob_id, sliver_count, epoch, key_pair)
+                .await
+                .unwrap_or_default()
+                .into())
+        } else {
+            Err(SyncShardError::NoSyncClient.into())
+        }
```
mlegner, line 612: Question: Why not simply create a client in this case instead of returning an error? cc @jpcsmith

crates/walrus-service/src/node/committee.rs
```diff
@@ -680,6 +735,49 @@ impl CommitteeService for NodeCommitteeService {
...
+        key_pair: &ProtocolKeyPair,
+    ) -> Result<Vec<(BlobId, GeneralSliver)>, anyhow::Error> {
+        tracing::info!(
+            shard = %shard,
+            starting_blob_id = %starting_blob_id,
+            sliver_type = ?sliver_type,
+            sliver_count = sliver_count,
+            epoch = epoch,
+            "syncing shard to epoch"
+        );
```
mlegner: Nit: This could be shortened:
```suggestion
        tracing::info!(
            %shard,
            %starting_blob_id,
            ?sliver_type,
            sliver_count,
            epoch,
            "syncing shard to epoch"
        );
```
Alternatively, we can instrument the method with `#[tracing::instrument(skip(self, key_pair))]` and then have just have `tracing::info!("syncing shard to epoch");` here.

crates/walrus-service/src/node/shard_sync.rs
mlegner, line 1: Question: Could you explain why we have a different structure for the shard syncs compared to the blob syncs with a `BlobSyncHandler` struct which one can use to trigger syncs?

crates/walrus-service/src/node/storage.rs
```diff
@@ -213,8 +213,9 @@ pub struct Storage {
     metadata: DBMap<BlobId, BlobMetadata>,
     blob_info: DBMap<BlobId, BlobInfo>,
     event_cursor: EventCursorTable,
-    shards: HashMap<ShardIndex, ShardStorage>,
+    shards: HashMap<ShardIndex, Arc<ShardStorage>>,
     config: DatabaseConfig,
+    shard_sync_sender: Option<Sender<ShardIndex>>,
```
mlegner: Nit: Could you add a comment what the purpose of this field is?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -283,10 +294,144 @@ impl ShardStorage {
         })
     }
 
+    /// Syncs the shard to the last epoch of from the previous shard owner.
```
mlegner: Question: Does "last" mean "current" here?
```suggestion
    /// Syncs the shard to the current epoch from the previous shard owner.
```
Could we then also call this `sync_shard_to_current_epoch`?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -283,10 +294,144 @@ impl ShardStorage {
...
+    )]
+    async fn sync_shard_to_epoch_internal(
+        &self,
+        node: &Arc<StorageNodeInner>,
+        sliver_type: SliverType,
+    ) -> Result<(), SyncShardError> {
+        let mut starting_blob_id = None;
+        while let Some(next_starting_blob_id) = next_certified_blob_id(
+            &node.storage.blob_info,
+            node.current_epoch(),
```
mlegner: Question: Shouldn't this be the previous epoch? The old shard owner wouldn't get blobs that are written in this epoch.

crates/walrus-service/src/node/shard_sync.rs
halfprice: If the question is about why the struct looks different: the main difference is that blob sync is directly triggered (called) when processing events, where shard sync is triggered by storage adding shards. This is to make sure that the shard is created before the sync happens, so we want the shard to drive the initiation of the sync.

If the question is about why not also do shard sync in BlobSyncHandler: I think we want to differentiate these two type of tasks to control them separately. Generalize BlobSyncHandler to handle both syncs will probably make it more confusing.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Yes. The problem is that currently, we don't have a notion of `previous epoch` in CommitteeService yet. This is currently just a place holder. I'm going to work on this next.

crates/walrus-service/src/node/committee.rs
halfprice: We want to sync up to the end of previous epoch, that's what the epoch is indicated.

For new blobs created in the current epoch, we don't want to sync them.

crates/walrus-service/src/node/committee.rs
halfprice: I didn't read the code too closely to have an understanding of all the cases, but one of the errors that cause client creation fails is a wrong url.

crates/walrus-service/src/node.rs
```diff
@@ -332,20 +327,35 @@ struct StorageNodeInner {
...
     async fn new(
+        config: &StorageNodeConfig,
         key_pair: ProtocolKeyPair,
-        mut storage: Storage,
         event_provider: Box<dyn SystemEventProvider>,
         committee_service_factory: Box<dyn CommitteeServiceFactory>,
         contract_service: Box<dyn SystemContractService>,
         registry: &Registry,
-        max_concurrent_blob_syncs: usize,
+        pre_created_storage: Option<Storage>, // For testing purposes.
```
jpcsmith: Question: In which cases do we not want the storage to already be created?

crates/walrus-service/src/node/committee.rs
mlegner: Ah, I see. I misinterpreted the "to" as indicating some destination.

Maybe "before" would be clearer?

crates/walrus-service/src/node/shard_sync.rs
mlegner: My question was more in the first direction. Maybe to be a bit more specific: For the blob sync we have this handler struct on which we can call sync methods that then spin off a task that performs the actual sync. In the shard sync we have a background task to which we send indexes of shards to sync. I would like to understand the benefit of the perpetual background task (shard sync) over the struct (blob sync).

crates/walrus-service/src/node/committee.rs
mlegner: Ah, I see. But is it clear that there *should* exist a client for every storage node? So the only reason we don't find one is that there was an error before?

crates/walrus-service/src/node/storage/shard.rs
mlegner: OK, makes sense. Can you add a TODO comment there to make this clear?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -283,10 +294,144 @@ impl ShardStorage {
...
+
     #[cfg(test)]
     pub(crate) fn lock_shard_for_epoch_change(&self) -> Result<(), TypedStoreError> {
         self.shard_status.insert(&(), &ShardStatus::LockedToMove)
     }
+
+    #[cfg(test)]
+    pub(crate) fn update_status_in_test(&self, status: ShardStatus) -> Result<(), TypedStoreError> {
+        self.shard_status.insert(&(), &status)
+    }
```
jpcsmith: Suggestion: Please don't expose internal state to tests. It makes them brittle. Use the existing public APIs.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -283,10 +294,144 @@ impl ShardStorage {
...
+    /// Syncs the shard to the last epoch of from the previous shard owner.
+    #[tracing::instrument(
+        skip_all,
+        fields(
+            walrus.node = %node.protocol_key_pair.as_ref().public(),
+            walrus.shard_index = %self.id
+        ),
+        err
+    )]
+    pub async fn sync_shard_to_epoch(
```
jpcsmith: Remark: I would rather keep `Storage` and `Shard` as walrus-specific layers atop the storage, as opposed to having business logic for syncing within them. 

Suggestion: Move the logic for syncing a higher level of abstraction, such as within the storage node or a dedicated service created there. You can then add any functions you need to `Shard` to transition the state as necessary.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -283,10 +294,144 @@ impl ShardStorage {
...
+        err
+    )]
+    async fn sync_shard_to_epoch_internal(
+        &self,
+        node: &Arc<StorageNodeInner>,
+        sliver_type: SliverType,
+    ) -> Result<(), SyncShardError> {
+        let mut starting_blob_id = None;
+        while let Some(next_starting_blob_id) = next_certified_blob_id(
+            &node.storage.blob_info,
```
jpcsmith: Important: Do not expose the internals of the `Storage` to the `Shard`, as it will make it difficult to make any changes to one without breaking the other. Add public APIs as necessary.

crates/walrus-service/src/node/shard_sync.rs
halfprice: I see. There is one challenge to pass a sync handler to `shard storage`. Currently, the objects are created in the following order:

per shard storage -> storage -> node_inner -> sync handler.

The `sync handler` must be created after `node_inner` so that it can call any sync function needed. Since shard sync is initiated by `shard storage`, it requires some code structure change in order for the `shard storage` to have a reference to the `sync handler`. By creating a channel, it makes `shard storage` -> `sync handler` communication easier (in terms of writing code).

If we worry about the overhead by having an extra background thread, we can make `shard_sync_handler` an background async function call instead of in a dedicated thread. But I don't feel the `perpetual` manner of the background task will cause any issue since it'll just be blocked waiting on the channel most of the time.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -283,10 +294,144 @@ impl ShardStorage {
...
+        },
+        Unbounded,
+    ));
+
+    for result in iter {
+        let (blob_id, blob_info) = result?;
+        if blob_info.is_certified()
+            && blob_info
+                .status_changing_epoch(super::blob_info::BlobCertificationStatus::Certified)
+                .unwrap()
```
jpcsmith, line 457: Nit: Could be replaced with a `epoch_certified() -> Option<Epoch>` which would be clearer.

crates/walrus-service/src/node/storage.rs
jpcsmith: Question: Ditto.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -57,7 +68,7 @@ impl ShardStorage {
         database: &Arc<RocksDB>,
         db_config: &DatabaseConfig,
         initial_shard_status: Option<ShardStatus>,
-    ) -> Result<Self, TypedStoreError> {
+    ) -> Result<Arc<Self>, TypedStoreError> {
```
jpcsmith: Suggestion: Wrap in an Arc outside of the constuctor.

Question: Why the need for an `Arc`, isn't `ShardStorage` already `Send + Sync + Clone`? If the goal is to make it cheaper to do so, then it would better to internally wrap the type in an `Arc`.

crates/walrus-service/src/node/shard_sync.rs
jpcsmith: Remark: I would prefer if we make this into it's own class, following the template for the `BlobSyncHandler`. I am not yet seeing why we cannot construct `per shard storage -> storage -> node_inner` then start start a service there there to begin the sync. Additionally, (as I also wrote elsewhere), the `ShardStorage` shouldn't be initiating the sync.

crates/walrus-service/src/node/server/responses.rs
jpcsmith, line 1: Remark: Will revisit in a subsequent revision.

crates/walrus-service/src/node/errors.rs
jpcsmith, line 1: Remark: Will revisit in a subsequent revision.

crates/walrus-service/src/node/committee.rs
jpcsmith: Yea, that's correct. This should ideally never happen, but it's possible that we're unable to create a reqwest client due to the node providing a bad hostname (which does not parse) or an when it tries to initialise things like proxy settings, TLS, DNS resolver.

crates/walrus-service/src/node/committee.rs
```diff
@@ -680,6 +735,49 @@ impl CommitteeService for NodeCommitteeService {
...
+        sliver_count: u64,
+        epoch: Epoch,
+        key_pair: &ProtocolKeyPair,
+    ) -> Result<Vec<(BlobId, GeneralSliver)>, anyhow::Error> {
+        tracing::info!(
+            shard = %shard,
+            starting_blob_id = %starting_blob_id,
+            sliver_type = ?sliver_type,
+            sliver_count = sliver_count,
+            epoch = epoch,
```
jpcsmith: Remark: There are existing names for some of these attributes (see `docs/telemetry-attributes.csv`). Additionally, this should probably be attached to a span elsewhere, as opposed to only being present in an event.

crates/walrus-service/src/node/committee.rs
halfprice: SG. Done.

crates/walrus-service/src/node/committee.rs
halfprice: This error should never happen only when the system behave correctly. If we change this into an assert, in case there is a bug in the contract where some shards are not assigned, this will cause nodes to crash, which may become availability burden. Since the contract and the nodes are relatively independent components, I feel it's better to log some error vs. crashing the node upon invariant violation.

crates/walrus-service/src/node/committee.rs
halfprice: Update to use instrument

crates/walrus-service/src/node/committee.rs
halfprice: Updated to use instrument

crates/walrus-service/src/node/storage/shard.rs
halfprice: Implemented the suggestion. Another reason for an `Arc` is for the sync task to always have a valid `ShardStorage` to operate on.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Can you elaborate? We explicitly don't want these method to be public

crates/walrus-service/src/node/storage/shard.rs
halfprice: Good idea. Will do it in a separate PR since other type of events may also benefit from such refactor.

crates/walrus-sui/src/types.rs
halfprice: Added a TODO.

crates/walrus-service/src/node/storage/shard.rs
halfprice: SG. I'll do that once the PR is ready to merge.

crates/walrus-service/src/node.rs
halfprice: Taken!

crates/walrus-service/src/node/storage/shard.rs
halfprice: Done. And changed the function name to sync_shard_before_epoch as suggested in another comment.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Done.

crates/walrus-service/src/node.rs
halfprice: All tests are created using pre-created storage. This change is just to make the existing test work. But I agree that as an interface, it's probably better not to have such weird input to just for tests. Will try to address it in a follow up PR.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Hi @jpcsmith , after our chat last Friday, I did a few sample implementation to move the sync logic layer up in NodeStorage or Storage, but eventually still moved the implementation back to ShardStorage, because I still think this is the cleanest way to structure the code.

The main issue with moving sync logic in upper layer such as NodeStorage, is that the sync logic is too far away from the states used in the sync logic. Since syncing is per shard, any internal state such as progress checkpointing and missing blobs are naturally stored inside ShardStorage, but is core to the sync logic. We could expose them, but in that case (1) the API that exposes the syncing related states become unintuitive to use, and (2) an outside logic uses ShardStorage's internal state to implement their functionality.

The next PR will implement using sliver recovering to recover missing blobs, and it may make more sense once you see that PR. Let me know if you want to chat more about this.

crates/walrus-service/src/node/shard_sync.rs
halfprice: As discussed, I created a ShardSyncHandler which exposes an API to start background tasks to sync shards. PTAL.

crates/walrus-service/src/node/committee.rs
mlegner: IMO, if our contracts have such a severe bug that we have an unassigned shard, we probably have bigger problems than crashed nodes... 😅 

However, I don't see a real downside of having this error either, so let's leave it as it is. 🙂

crates/walrus-service/src/node/storage/shard.rs
jpcsmith: Hey Zhe, the place for this would be in a `ShardSynchronizer`, see `blob_sync.rs` for the analogous used for syncing blobs. 

While the following is severely oversimplified, the logic within such a service would run something as follows, based on the the existing code and state in the "Walrus Storage Node Epoch Change" document:

```rust
impl ShardSynchronizer {
    fn run(..) {
        let expected_slivers = storage.live_blob_ids(epoch);
        // For simplicity, sync_shard_before_epoch can be be abstracted as an asynchronous ordered stream
        let incoming_slivers = sync_shard_before_epoch_stream(..);
        let mut sliver_batch = vec![];

        for expected_blob_id in expected_slivers {
            // Drop all blob_ids returned in the stream until the next expected.
            if let Some((incoming_blob_id, incoming_sliver)) = advance_until(incoming_slivers, expected_blob_id) {
                debug_assert_eq!(expected_blob_id, incoming_blob_id);
                sliver_batch.push((incoming_blob_id, incoming_sliver));
            } else {
                // The stream advanced past the expected blob ID, the remote didnt send it.
                shard.mark_missing(expected_blob_id, sliver_type);
            }

            if batch.len() == WRITE_BATCH_SIZE {
                shard.put_multiple_slivers(sliver_batch);
                shard.advance_sync_progress(incoming_blob_id, sliver_type);
                sliver_batch.clear();
            }
        }
    }
}
```

This would require several new APIs added to the ShardStorage and Storage, in addition to those for changing the `ShardState` to being in-sync:

```rust
impl Storage {
    /// Returns an iterator over blobs that are alive at epoch Epoch.
    fn live_blob_ids(epoch: Epoch) -> impl Iterator<Item = Result<BlobId, _>>;
}

impl ShardStorage {
    /// For a shard that's being synced, mark the specified
    fn mark_missing(blob_id: BlobId, type_: SliverType);
    /// Store multiple slivers.
    fn put_multiple_slivers(slivers: Vec<Sliver>);
    /// Update the latest successfully synced blob ID for a sliver of a particular type.
    fn advance_sync_progress(blob_id: BlobId, type_: SliverType);
}
```

With the above, however, if the source storage node actually has very few slivers actually stored within the shard, for example, the *very last blob ID* that is alive, then we would need to store every single blob_id (of potentially millions) except the few missing, before moving on to recovery or individual fetching.

This could be avoided, perhaps, with a cursor based approach similar to that which we use for processing the event stream:

```rust
impl ShardSynchronizer {
    fn run(..) {
        let expected_slivers = storage.live_blob_ids(epoch);
        // For simplicity, sync_shard_before_epoch can be be abstracted as an asynchronous ordered stream
        let incoming_slivers = sync_shard_before_epoch_stream(..);

        for (blob_index, expected_blob_id) in expected_slivers.enumerate() {
            // Drop all blob_ids returned in the stream until the next expected.
            if let Some((incoming_blob_id, incoming_sliver)) = advance_until(incoming_slivers, expected_blob_id) {
                debug_assert_eq!(expected_blob_id, incoming_blob_id);
                // Switched to single slivers here for simplicity, but could also be done
                // batched, with batches prematurely needing to be written if we encounter a missing
                // sliver.
                shard.put_sliver(incoming_blob_id, incoming_sliver);
                shard.maybe_advance_sync_progress(live_blob_index, expected_blob_id, sliver_type);
                shard.mark_highest_synced(live_blob_index,expected_blob_id, sliver_type);
            } else {
                // Immmediately start recovering the sliver
                self.recover_sliver(live_blob_index, expected_blob_id, sliver_type);
            }
        }
    }

    fn recover_sliver(blob_index: usize, blob_id: BlobId, ..) {
        // Spawn off a task that does the following:
        // 1. Attempt to fetch solely that sliver repeatedly, with backoff, from the remote shard being synced.
        // 2. Failing that fall back to recovery from the committee.
        // 3. Call shard.maybe_advance_sync_progress with the supplied live_blob_index.
    }
}

impl ShardStorage {
    /// Sequentially update the last blob ID successfully synced for a sliver of a particular type.
    fn maybe_advance_sync_progress(index: usize, blob_id: BlobId, type_: SliverType);
    /// Stores the highest blob ID successfully synced (this function could be merged with the above,
    /// but left explicit for demonstration purposes).
    fn mark_highest_synced(index: usize, blob_id: BlobId, type_: SliverType);
}
```

The above avoids the previous issue by only maintaining the highest blob id and its index into
the list of all live blob IDs. All slivers of a type below this blob ID have been successfully
stored in the shard. Some above this have likely been stored. This can be kept memory and the
aforementioned value updated in the database when the gap is filled (see `EventSequencer` for the
same thing operating on event ids). In this case though, since re-fetching a sliver that has already
been stored is expensive, we can also store the furthest blob_id synced. On a restart, we can start
the bulk sync from that point, and individually check each sliver between the low and high values,
restarting recovery for any that are missing.

In both cases, the logic for the sync lives in its own service, and uses API methods on the shard.

crates/walrus-service/src/node/storage/shard.rs
jpcsmith: Sure, your test right now are white-box tests. They require knowing the implementation detials of the ShardStorage to run the tests (for example, that slivers are stored in a DBMap). Any changes to the internal state of the storage requires modifying multiple tests.

Instead tests should test the API and visible properties of the `ShardStorage`. Using methods that the storage node will need to call to sync the shard.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Hi @jpcsmith , thanks for sketching out the idea, and that's what I understood from our last discussion as well. My argument is that all the new APIs created in ShardStorage is still the `logic for the sync`, because the implementation and usage of those APIs are only serving the purpose of the sync.

Those APIs are essentially the getter/setter methods to access states only used for `sync`, and my main point is that the `sync` related state should be close to where the sync logic runs. If we really want the fetching logic implemented in `ShardSynchronizer`, we should also put sync checkpointing/bookkeeping in `ShardSynchronizer`, as separate tables in `ShardSynchronizer`, not in `ShardStorage`.

The main difference between the implementation patten with the BlobSync is that, BlobSync is `stateless`. The logic of BlobSync doesn't need to persist any state.

Right now, in the proposal from the previous message, we have sync logic running in ShardSynchronizer, but the state of the sync is stored in ShardStorage. I think this separation will continue adding complexity in managing these two classes. If I may speculate a bit, with crash-recovery in mind, we will add more public methods to access/change sync state:

```
impl ShardStorage {
    /// For a shard that's being synced, mark the specified
    fn mark_missing(blob_id: BlobId, type_: SliverType);
    fn get_missing_blobs(/*query for table lookup*/) -> [(BlobId, SliverType)];

    /// We need either a clear_missing_blobs, or a new put_sliver that does the clear
    /// We don't want put_sliver and put_multiple_slivers to pay a cost of missing blob
    /// look up in the common case.
    fn clear_missing_blobs();

    /// Store multiple slivers.
    fn put_multiple_slivers(slivers: Vec<Sliver>);

    /// Update the latest successfully synced blob ID for a sliver of a particular type.
    fn advance_sync_progress(blob_id: BlobId, type_: SliverType);
    /// Gets the last sync progress mark to restart the sync
    fn get_sync_progress() -> (blob_id, type);

    /// Updates shard status to be fully synced.
    fn mark_sync_done()

    /// Gets the finishing status of the last sync to be updated on chain.
    fn get_sync_finish_status()

    ....
}
```
Basically ShardStorage becomes a proxy to store shard sync state, and we have sync logic in one place, and sync logic state in another.

crates/walrus-service/src/node/shard_sync_handler.rs
mlegner, line 1: Nit: Can we name this file `shard_sync.rs`?

crates/walrus-service/src/node/shard_sync_handler.rs
```diff
@@ -0,0 +1,213 @@
...
+    pub async fn restart_syncs(&self) -> Result<(), anyhow::Error> {
+        for shard_index in self.node.storage.shards() {
+            let shard_storage = self.node.storage.shard_storage(shard_index).unwrap();
+
+            // Restart the syncing task for shards that were previously syncing (in ActiveSync status).
+            if shard_storage.status()? != ShardStatus::ActiveSync {
+                continue;
+            }
+
+            self.start_shard_sync_impl(shard_storage.clone()).await;
```
mlegner: Nit: A bit clearer in my opinion:
```suggestion
            // Restart the syncing task for shards that were previously syncing (in ActiveSync status).
            if shard_storage.status()? == ShardStatus::ActiveSync {
                self.start_shard_sync_impl(shard_storage.clone()).await;
            }
```

crates/walrus-service/src/node/shard_sync_handler.rs
```diff
@@ -0,0 +1,213 @@
...
+        let epoch_to_sync = self.node.current_epoch();
+        tracing::info!(
+            "Syncing shard index {} to the end of epoch {}",
+            shard_storage.id(),
+            epoch_to_sync
+        );
+
+        // TODO: implement rate limiting for shard syncs.
+        let mut shard_sync_in_progress = self.shard_sync_in_progress.lock().await;
+        if let Entry::Vacant(entry) = shard_sync_in_progress.entry(shard_storage.id()) {
```
mlegner: Nit: You could reduce the indentation level by using a `let`/`else` construction:
```rust

        let Entry::Vacant(entry) = shard_sync_in_progress.entry(shard_storage.id()) else {
            tracing::info!(
                "Shard index: {} is already being synced. Skipping starting new sync task.",
                shard_storage.id()
            );
            return;
        };
```

crates/walrus-service/src/node/storage/shard.rs
mlegner: Question: Are you suggesting to move the `next_certified_blob_id` (or a similar function) to the `Storage`?

crates/walrus-service/src/node/storage/shard.rs
mlegner: Thanks for the in-depth discussion on the shard-sync design, @halfprice and @jpcsmith. I have a bit of an outside view on this topic, as I haven't really worked on the storage code myself, so I spent quite some time looking through the code and considering the trade-offs.

While from an architecture perspective, I agree with @jpcsmith's suggestion of having the shard sync in a separate service, I feel in this case the practical benefits of putting it into the `ShardStorage` outweigh the more theoretical disadvantages:

As it is unclear if we want to support multiple types/structures of storage in the future and, if so, how their shard sync processes would differ, I would apply the YAGNI principle here and skip all the API boilerplate until it is clear that we actually need it.

My suggestion therefore is to go with @halfprice's proposal of having the shard-sync logic integrated in the `ShardStorage`.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Done.

crates/walrus-service/src/node/storage/shard.rs
halfprice: I see. I think you are more talking about `primary_slivers` and `secondary_slivers` above. It makes sense and I updated to a `sliver_count` method.

crates/walrus-service/src/node/storage/shard.rs
halfprice: I think what @jpcsmith meant was the direct usage of `blob_info` in `node.storage` should be avoided.

I created a new function `blob_info_iter(range)` in storage that returns an iterator with range to go though blob info.

crates/walrus-service/src/node/shard_sync_handler.rs
halfprice: Done.

There is also a sync_shard.rs, which stores the messages used for syncing shards. I'll rename that to avoid confusion.

crates/walrus-service/src/node/shard_sync_handler.rs
halfprice: Done

crates/walrus-service/src/node/shard_sync_handler.rs
halfprice: Done.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Thanks @mlegner for the detailed review! I don't think it is the end of the day if we go to one way or the other. I just feel that the sync logic is able to directly access the sync logic state vs. using APIs to access sync logic state is more readable and easier to maintain. I would also suggest let's go with the current PR if that's ok with @jpcsmith.

I think I've addressed all the comments in this PR and ready for another round of review.

crates/walrus-service/src/node/storage/shard.rs
jpcsmith: @mlegner Zhe's comment is correct. The state within these classes should be private w.r.t, one another, and my comment was against accessing the DBMap directly.

crates/walrus-service/src/node/storage/shard.rs
jpcsmith: Thank's for weighing in and unblocking this @mlegner. 

@halfprice you have @mlegner's approval so you can merge it in once comments are resolved.

PR #677, opened 2024-08-16
feat: add write payments, clean up code

karlwuest:


contracts/walrus/sources/staking/storage_node.move
```diff
@@ -8,12 +8,17 @@ module walrus::storage_node;
 use std::string::String;
 use sui::{bls12381::{G1, g1_from_bytes}, group_ops::Element};
 
+// Error codes
+const EInvalidNetworkPublicKey: u64 = 1;
+
+
```
damirka: Nit: I believe there should be just 1 line in between, formatter should fix it.

contracts/walrus/sources/system/messages.move
```diff
@@ -0,0 +1,175 @@
...
+use sui::bcs;
+
+const APP_ID: u8 = 3;
+
+// Message Types
+const BLOB_CERT_MSG_TYPE: u8 = 1;
+const INVALID_BLOB_ID_MSG_TYPE: u8 = 2;
+
+// Errors
+const EIncorrectAppId: u64 = 0;
```
damirka, line 15: Eventually we'd want to replace all numeric error codes with error messages, as @tnowacki suggested in my other PR:

```
#[error]
const EIncorrectAppId: vector<u8> = b"Application ID is invalid";
```

contracts/walrus/sources/system/blob.move
```diff
@@ -145,80 +144,38 @@ public(package) fun register(
...
-    assert!(message.epoch == sys.epoch(), EWrongEpoch);
+    assert!(message.certified_epoch() == current_epoch, EWrongEpoch);
 
     // Check that the storage in the blob is still valid
-    assert!(message.epoch < blob.storage.end_epoch(), EResourceBounds);
+    assert!(message.certified_epoch() < blob.storage.end_epoch(), EResourceBounds);
 
     // Mark the blob as certified
-    blob.certified_epoch = option::some(message.epoch);
+    blob.certified_epoch = option::some(message.certified_epoch());
```
damirka: Nit: I always feel a little uncomfortable when option gets rewritten directly; and believe that using `fill`, `swap` and `extract` is a better strategy for both safety (they prevent filling "is_some", and extracting "is_none") and readability - you always know the expected state of an Option.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -68,6 +66,42 @@ public(package) fun shard_transfer_failed(
     abort ENotImplemented
 }
 
+/// Update epoch to next epoch, and update the committee, price and capacity.
+///
+/// Called by the epoch change function that connects `Staking` and `System`. Returns
+/// the balance of the rewards from the previous epoch.
+public(package) fun next_epoch(
```
damirka: We're using `advance_epoch` in staking right now. I'm not sold on either but eventually we should make a decision on naming for consistency.

contracts/walrus/sources/system/messages.move
karlwuest: Yes, that's a great suggestion! Let's do that in a follow up and replace all of them.

contracts/walrus/sources/system/blob.move
karlwuest: Changed to `fill`.

contracts/walrus/sources/system/system_state_inner.move
karlwuest: I changed this to `advance_epoch` now 👍

PR #680, opened 2024-08-16
feat!: use HTTPS for communication with storage nodes

jpcsmith:
Add HTTPS to connections served by the server, and only accept TLS connections at the client. Specifically, this PR adds

- Securing communication with TLS certificates loaded from files.
- Securing communication with self-signed TLS certificates.
- Allows disabling TLS entirely (so as to offload it to a middlebox, the client *only* works with TLS).
- Changes network keys to P-256 instead of Ed25519, since the latter is not yet supported by Let's Encrypt.
- Use a separate reqwest client instance for each storage node client. This ensures that a request to a client for node A is not authenticated with a certificate from node B.
- Allows non-graceful shutdown of the REST api if specified in the configuration file. As a result, I removed the resetting of http connections in the client for tests.
- Unifies storage node configs used in tests and the testbed so that they inherit from a single test config, and overwrite needed values.

### DIFF
--- a/crates/walrus-event/src/blob_writer.rs
+++ b/crates/walrus-event/src/blob_writer.rs
@@ -149,12 +149,12 @@ impl EventBlobWriter {
         )?;
         let file = Self::next_file(&root_dir_path, EventBlob::MAGIC, EventBlob::FORMAT_VERSION)?;
         let encoding_config = EncodingConfig::new(100.try_into().unwrap());
-        let yaml = "---\n\
-        storage_path: target/storage\n\
-        protocol_key_pair:\n  BBlm7tRefoPuaKoVoxVtnUBBDCfy+BGPREM8B6oSkOEj\n\
-        network_key_pair:\n  AFzrKKEebJ56OhX4QHH9NQshlGKEQuwdD3HlE5d3n0jm";
 
-        let node_config: StorageNodeConfig = serde_yaml::from_str(yaml)?;
+        let node_config = StorageNodeConfig {
+            storage_path: "target/storage".into(),
+            ..walrus_service::test_utils::storage_node_config().inner
+        };
+
         let client = Client::for_storage_node(
             &node_config.rest_api_address.ip().to_string(),
             node_config.rest_api_address.port(),
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -139,7 +139,7 @@ mod tests {
             .unwrap()
     }
 
-    #[tokio::test(start_paused = true)]
+    #[tokio::test(start_paused = false)]
     async fn test_restart_syncs() {
         let cluster = create_test_cluster(&[&[0, 1, 2]]).await;
         for i in [0, 2] {
@@ -173,7 +173,7 @@ mod tests {
             .contains_key(&ShardIndex(2)));
     }
 
-    #[tokio::test(start_paused = true)]
+    #[tokio::test(start_paused = false)]
     async fn test_start_new_shard_sync() {
         let cluster = create_test_cluster(&[&[0]]).await;
         let shard_sync_handler = ShardSyncHandler::new(cluster.nodes[0].storage_node.inner.clone());


Cargo.toml
```diff
@@ -31,19 +32,17 @@ mockall = "0.12.1"
...
+rcgen = "0.13.1"
 regex = "1"
-reqwest = { version = "0.12.5", default-features = false, features = [
-  "http2",
-  "json",
-  "rustls-tls",
-] }
+reqwest = { version = "0.12.5", default-features = false, features = [ "http2", "json", "rustls-tls", ] }
 rocksdb = "0.21.0"
+rustls = { version = "0.23.12", default-features = false, features = ["ring", "tls12", "logging"] }
```
giac-mysten, line 49: Curiosity: why the `ring` crypto provider instead of the default? Also, iiuc `tls12` is the default.

crates/walrus-core/src/lib.rs
```diff
@@ -655,3 +656,17 @@ macro_rules! ensure {
         }
     };
 }
+
+/// Returns a string `<first-4-bytes-as-hex>.network.walrus.alt` corresponding to the public key.
+pub fn server_name_from_public_key(public_key: &NetworkPublicKey) -> String {
```
giac-mysten: Question: why add this to `walrus-core`? The naming of the servers seems more of a `service` or `sdk` concern. Afaics this is only used in `sdk`.

other question: do we expect all storage nodes to respect this naming scheme?

praise: love that you used `alt`!

crates/walrus-sdk/src/client.rs
```diff
@@ -132,6 +138,99 @@ impl UrlEndpoints {
...
+    /// to [`build()`][Self::build]. Calling this method instead authenticates the storage node solely on the
+    /// basis that it is able to establish the TLS connection with the private key corresponding to
+    /// the provided public key.
+    pub fn authenticate_with_public_key(mut self, public_key: NetworkPublicKey) -> Self {
+        self.server_public_key = Some(public_key);
+        self
+    }
+
+    /// Clears proxy settings in the client, and disables fetching proxy settings from the operating
+    /// system. On some systems, this can speed up the construction of the client.
```
giac-mysten:
```suggestion
    /// Clears proxy settings in the client, and disables fetching proxy settings from the OS.
    ///
    /// On some systems, this can speed up the construction of the client.

```

giac-mysten: Maybe mention that this is only used for the simulation tests?

crates/walrus-sdk/src/client.rs
```diff
@@ -140,12 +239,25 @@ pub struct Client {
...
+    pub fn builder() -> ClientBuilder {
+        ClientBuilder::default()
+    }
+
+    /// Creates a new client for the storage node identified by the DNS name or socket address,
+    /// and authenticated with the provided public key.
+    ///
+    /// This method ensures that the storage node is authenticated on the basis that the storage
+    /// node is able to establish the connection on the basis of certificate with the provided
+    /// identity and public key.
```
giac-mysten:
```suggestion
    /// Creates a new client for the storage node.
    ///
    /// The storage node is identified by the DNS name or socket address, and authenticated with
    /// the provided public key.
    ///
    /// This method ensures that the storage node is authenticated: Only the storage node can
    /// establish the connection using the self-signed certificate corresponding to the provided
    /// identity and public key.

```

crates/walrus-sdk/src/client.rs
```diff
@@ -132,6 +138,99 @@ impl UrlEndpoints {
...
+
+        let endpoints = UrlEndpoints(url);
+
+        let mut builder = self
+            .inner
+            .https_only(true)
+            .http2_prior_knowledge()
+            .use_rustls_tls()
+            .tls_built_in_root_certs(false)
+            .tls_built_in_webpki_certs(false);
```
giac-mysten: Question: do I understand correctly that we only want this client to authenticate the self-signed certificate with the network key, and nothing else?

Afaics from the docs,  `tls_built_in_webpki_certs` is available on crate feature `rustls-tls-webpki-roots` only, and  `tls_built_in_native_certs` is available on crate feature `rustls-tls-native-roots` only. Remove?

```suggestion
            .tls_built_in_root_certs(false)
```

crates/walrus-sdk/src/tls.rs
```diff
@@ -0,0 +1,112 @@
...
+    time::{Time, Validity},
+    Version,
+};
+
+/// Create a unsigned, self-signed TLS certificate for a public key.
+///
+/// The subject (and issuer) is the storage node identified by
+/// `walrus_core::server_name_from_public_key(public_key)`, whose alternative name is provided by
+/// `subject_alt_name`. The alternative name should correspond to the DNS name or IP-address under
+/// to which the connection, so that the certificate can be used to verify the connection.
```
giac-mysten:
```suggestion
/// `subject_alt_name`. The alternative name should correspond to the DNS name or IP-address to
///  which the connection is established, so that the certificate can be used to verify the connection.
```

crates/walrus-sdk/src/tls.rs
```diff
@@ -0,0 +1,112 @@
...
+///
+/// When using the Rustls TLS backend, a certificate generated in this way can be used to establish
+/// a root of trust for securing TLS connections, as Rustls assumes that the certificate's
+/// signature, expiry, etc. has already been verified.
```
giac-mysten, line 53: Praise: very nice!

crates/walrus-service/src/node.rs
```diff
@@ -1530,7 +1530,7 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test(start_paused = true)]
+    #[tokio::test(start_paused = false)]
```
giac-mysten, line 1619:
```suggestion
    #[tokio::test(start_paused = false)]
```
Nit: can remove?

scripts/crash_recovery.sh
```diff
@@ -185,7 +185,7 @@ function get_blob_confirmation() {
     blob_id=$1
     host_address=$2
     #/v1/blobs/{blob_id}/confirmations
-    cmd="curl -s -X GET http://${host_address}/v1/blobs/${blob_id}/confirmation"
+    cmd="curl -s -X GET https://${host_address}/v1/blobs/${blob_id}/confirmation"
```
giac-mysten: Question: this will fail bc the host does not have a tls certificate. Ignore SSL errors?
```suggestion
    cmd="curl -s -X GET https://${host_address}/v1/blobs/${blob_id}/confirmation" --insecure
```

crates/walrus-service/src/node/server.rs
```diff
@@ -690,8 +937,80 @@ mod test {
...
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn client_accepts_self_signed_certificate() -> TestResult {
+        let mut config = test_utils::storage_node_config();
+
+        // Enable self-signed certificates by enabling TLS but removing any certificate.
+        config.as_mut().tls.disable_tls = false;
+        config.as_mut().tls.pem_files = None;
```
giac-mysten: Question: isn't this already done in `build()`?

crates/walrus-service/src/node/config.rs
```diff
@@ -53,11 +53,47 @@ pub struct StorageNodeConfig {
     /// Socket address on which the REST API listens.
     #[serde(default = "defaults::rest_api_address")]
     pub rest_api_address: SocketAddr,
+    /// Duration for which to wait for connections to close before shutting down.
+    ///
+    /// Set explicitly to None to wait indefinitely.
+    #[serde(
+        default = "defaults::rest_graceful_shutdown_period_secs",
+        skip_serializing_if = "Option::is_none",
+        with = "::serde_with::rust::double_option"
```
giac-mysten: Question: Why the leading double colon?

crates/walrus-service/src/node/server.rs
```diff
@@ -36,30 +49,253 @@ mod routes;
...
+        /// The server name (DNS name or IP address) to be placed in the certificate.
+        server_name: String,
+        /// The network key pair used to create the self-signed certificate.
+        network_key_pair: NetworkKeyPair,
+    },
+}
+
 /// Represents a user server.
 #[derive(Debug)]
 pub struct UserServer<S> {
```
giac-mysten, line 137: Nit: `UserServer` is an odd name, given that we only have one server :)

crates/walrus-service/src/node/server.rs
```diff
@@ -36,30 +49,253 @@ mod routes;
...
+}
+
+/// Source of the TLS private key and certificate.
+#[derive(Debug)]
+pub enum TlsCertificateSource {
+    /// Load PEM encoded x509 certificate and a PKCS8 encoded private key from the specified paths.
+    ///
+    /// These ideally should be certificates issued to by a public CA such as Let's Encrypt,
+    /// but can also be self-signed certificates.
+    // TODO(jsmith): Do we need to occasionally reload this?
```
giac-mysten: Reply: good question, I guess the certs will expire sooner or later right?

giac-mysten: For now we can just reboot the node?

crates/walrus-service/src/client.rs
```diff
@@ -672,72 +679,59 @@ impl<T> Client<T> {
...
         &'a self,
-        constructor: impl Fn((usize, &'a StorageNode)) -> Option<NodeCommunication<'a, W>>,
-    ) -> Vec<NodeCommunication<'a, W>> {
-        let mut comms: Vec<_> = self
-            .committee
-            .members()
-            .iter()
-            .enumerate()
-            .filter_map(constructor)
+        constructor: impl Fn(usize) -> Result<Option<NodeCommunication<'a, W>>, ClientBuildError>,
```
giac-mysten, line 684: cool :)

Cargo.toml
jpcsmith: It's what's used in Sui, but to disable aws-lc-rs and enable ring I need to disable default features and re-enable all the features that were part of the default feature.

crates/walrus-core/src/lib.rs
jpcsmith: > Question: why add this to walrus-core? The naming of the servers seems more of a service or sdk concern. Afaics this is only used in sdk.

Good point, I should put it into walrus-sdk, since that's imported by service. I will move it there.

> other question: do we expect all storage nodes to respect this naming scheme?

Only for self-signed certificates; but now I realise that since I didn't test with a non-self signed certificate, my implementation doesn't support it, since the only installed trust root is the remote storage node itself 🫤.

crates/walrus-sdk/src/client.rs
jpcsmith: I don't follow, it doesn't need to be. It's a valid setting in any client code (as it is with the underlying reqwest library).

crates/walrus-service/src/node/server.rs
jpcsmith:
```suggestion
    // TODO(jsmith): Reload the certificate on change (#709)
```

jpcsmith: Yea, sounds good. Will add an issue to track.

crates/walrus-core/src/lib.rs
jpcsmith: I will need to refactor the code to additionally use a custom verifier, so that we can accept other trust roots while still pinning the public key.

crates/walrus-service/src/node.rs
jpcsmith: Do you mean just to have `#[tokio::test]`? If so, I would keep it explicit since someone may come along and re-add the `start_paused = true` thinking it was overlooked.

crates/walrus-service/src/node/config.rs
jpcsmith: It's from a copy and paste, will remove.
```suggestion
        with = "serde_with::rust::double_option"
```

crates/walrus-service/src/node/server.rs
jpcsmith, line 137:
```suggestion
// TODO(jsmith):  Rename to something more appropriate (#710)
pub struct UserServer<S> {
```

jpcsmith: Added an issue.

crates/walrus-service/src/node/server.rs
jpcsmith: Could you rephrase?

Cargo.toml
giac-mysten: Interesting, thanks!

crates/walrus-core/src/lib.rs
giac-mysten: just to make sure: the problem here is if a storage node wants to use their normal domain instead of the `alt` domain?

crates/walrus-sdk/src/client.rs
giac-mysten: It seemed to me that the proxy is always gated by the `msim` feature.

crates/walrus-service/src/node.rs
giac-mysten: sounds good

scripts/crash_recovery.sh
giac-mysten: (to be clear, i'm not 100% sure this is the case, maybe doublecheck 😬 )

crates/walrus-service/src/node/server.rs
giac-mysten: Sorry, i considering that this config is used to build the storage node client, and therefore the exclusive use of self-signed certificates was already given. However, this also is used to configure the server, so it makes sense 🙏

crates/walrus-core/src/lib.rs
jpcsmith: Not only. As the only root CA is that of the remote storage node itself, any other issuer, such as from the PKI would be rejected, even if they use the `alt` name when requesting from Let's Encrypt

giac-mysten: mmm i would assume that let'sencrypt (and other CAs) rejects requests for `alt`, as it cannot perform the ACME checks?

jpcsmith: I can't say, but the common [subject alternative name](https://en.wikipedia.org/wiki/Subject_Alternative_Name) extension allows you to add multiple names, including email addresses, company names, etc, so it doesn't have to be the only name.

crates/walrus-sdk/src/tls.rs
```diff
@@ -0,0 +1,254 @@
+// Copyright (c) Mysten Labs, Inc.
```
bmwill, line 1: One possible option would be to add the p256 support to sui-tls crate so that there is one less place were we need to manage tls code

bmwill: that and it will make it easier to hook into sui-proxy cc @johnjmartin

PR #681, opened 2024-08-17
feat: make stress client write different blob sizes

karlwuest:
Additionally:
- adds a tag `TESTBLOB` to the start of each blob
- prevents `Interval::tick` from panicking if the read or write rate is set to 0
- creates write clients sequentially to prevent RPC timeouts due to the task not being scheduled when creating many clients with large blobs

### DIFF
--- a/crates/walrus-stress/src/generator/blob.rs
+++ b/crates/walrus-stress/src/generator/blob.rs
@@ -43,10 +43,17 @@ impl BlobData {
         self.bytes[index] = self.bytes[index].wrapping_add(1);
     }
 
+    /// Returns a slice of the blob with a size `2^x`, where `x` is chosen uniformly at random
+    /// between `min_size_log2` and `max_size_log2`.
     pub fn random_size_slice(&self) -> &[u8] {
         let blob_size_log2 = thread_rng().gen_range(self.min_size_log2..=self.max_size_log2);
         let blob_size = 2_usize.pow(blob_size_log2 as u32);
         &self.bytes[..blob_size]
+    }
+
+    pub fn refresh_and_get_random_slice(&mut self) -> &[u8] {
+        self.refresh();
+        self.random_size_slice()
     }
 }
 
--- a/crates/walrus-stress/src/generator/write_client.rs
+++ b/crates/walrus-stress/src/generator/write_client.rs
@@ -50,12 +50,12 @@ impl WriteClient {
 
     /// Stores a fresh consistent blob and returns the blob id and elapsed time.
     pub async fn write_fresh_blob(&mut self) -> Result<(BlobId, Duration), ClientError> {
-        self.blob.refresh();
+        let blob = self.blob.refresh_and_get_random_slice();
         let now = Instant::now();
         let blob_id = self
             .client
             .as_ref()
-            .reserve_and_store_blob(self.blob.random_size_slice(), 1, true)
+            .reserve_and_store_blob(blob, 1, true)
             .await?
             .blob_id()
             .to_owned();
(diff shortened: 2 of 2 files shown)


crates/walrus-stress/src/generator/blob.rs
```diff
@@ -1,30 +1,53 @@
...
-            .await
-            .expect("should be able to join spawned task");
+        let size = 2_usize.pow(max_size_log2 as u32);
+        let n_additional_bytes = size - TAG.len();
+        let bytes = tokio::spawn(async move {
+            TAG.iter()
+                .cloned()
+                .chain((0..n_additional_bytes).map(|_| new_rng.gen::<u8>()))
+                .collect()
+        })
```
mlegner, line 27: Question: Why is this done in a separate task?

crates/walrus-stress/src/generator/write_client.rs
```diff
@@ -80,7 +82,7 @@ impl WriteClient {
     /// node, if the shards are distributed equally and assigned sequentially.
     async fn write_inconsistent_blob(&self) -> Result<Blob, ClientError> {
```
mlegner: Question: Why don't we have to call `.refresh()` on the blob here? Is the assumption that we anyway send fewer inconsistent blobs than normal ones so we refresh the blob often enough?

crates/walrus-stress/src/generator/blob.rs
```diff
@@ -1,30 +1,53 @@
...
     }
 
-    /// Changes the blob by incrementing (wrapping) a randomly chosen byte.
+    /// Changes the blob by incrementing (wrapping) a randomly chosen byte excluding
+    /// the `TESTBLOB` tag at the start of the blob data.
     pub fn refresh(&mut self) {
-        let index = self.rng.gen_range(0..self.bytes.len());
+        let index = self.rng.gen_range(TAG.len()..self.bytes.len());
         self.bytes[index] = self.bytes[index].wrapping_add(1);
     }
```
mlegner, line 44: Suggestion: We could have a convenience function `refresh_and_get_random_slice` such that one cannot forget to call `refresh`.

crates/walrus-stress/src/generator/blob.rs
karlwuest: This can take some time depending on the size of the blob that is created and does not yield while it's creating the blob, so it's spawned in a separate task to not block the current thread. With the changes in this PR (making the creation of the write clients sequential), this may not be as important anymore, but previously just running it in the same task could cause issues with rpc requests timing out because the respective futures did not get scheduled frequently enough.

crates/walrus-stress/src/generator/write_client.rs
karlwuest: `refresh()` is called in the caller of this function (namely `write_fresh_inconsistent_blob()`).

crates/walrus-stress/src/generator/blob.rs
karlwuest: Added this, but only using it in `write_fresh_blob` and not `write_fresh_inconsistent` blob, because refreshing requires a mutable reference, so making the two calls separately is easier.

crates/walrus-stress/src/generator/write_client.rs
karlwuest: I renamed the `write_inconsistent_blob` function and changed the signature to take a blob as argument. Now `write_fresh_blob` and `write_fresh_inconsistent_blob` are a bit more similar, and the names are hopefully less confusing.

Issue #687, opened 2024-08-19
Clean up move tests


karlwuest:
- [ ] Replace byte vectors in tests with more concise hex strings to improve readability with the auto-formatter
- [ ] Reduce duplicate code in tests
- [ ] Generate signatures in move tests instead of using constants

PR #689, opened 2024-08-19
test: add test_only functions for bls signing and pk derivation

karlwuest:
Adding bls signing functions to be able to clean up the tests in a follow up (see #687 )

### DIFF
--- a/contracts/walrus/tests/bls_test_utils.move
+++ b/contracts/walrus/tests/bls_test_utils.move
@@ -1,0 +1,43 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+module walrus::bls_test_utils;
+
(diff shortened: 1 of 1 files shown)


contracts/walrus/tests/bls_test_utils.move
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+#[test_only]
```
damirka: All files in the `tests/` directory are test / dev only by default, so no annotation is necessary. Similarly, it's not required for its members.

damirka: As a reference, see [test_utils.move](https://github.com/MystenLabs/walrus/blob/main/contracts/walrus/tests/staking/test_utils.move). Already mentioned in the #690, maybe it's time we create shared test utils for system and staking.

PR #690, opened 2024-08-19
feat: implement blob deletion and direct extension

karlwuest:


contracts/walrus/sources/system.move
```diff
@@ -47,11 +42,11 @@ public fun certify_event_blob(system: &mut System, cap: &StorageNodeCap, blob_id
 public fun reserve_space(
     self: &mut System,
     storage_amount: u64,
-    periods_ahead: u64,
+    epochs_ahead: u64,
```
damirka, line 45: Does it make sense to have `u64` here? Maybe nitpicking, but also curious

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -245,6 +220,71 @@ public(package) fun extend_blob_with_resource(
...
+    start_offset: u64,
+    end_offset: u64,
+    payment: &mut Coin<SUI>,
+) {
+    let storage_units = storage_units_from_size(storage_size);
+    let period_payment_due = self.storage_price_per_unit_size * storage_units;
+    let coin_balance = payment.balance_mut();
+
+    let mut i = start_offset;
+    while (i < end_offset) {
```
damirka: Can be macro'd.

```rust 
start_offset.range_do!(end_offset, |i| {
  // ...
});
```

contracts/walrus/tests/system/blob_tests.move
```diff
@@ -668,19 +683,304 @@ public fun test_blob_extend_bad_period(): system::System {
...
     blob1.burn();
     system
 }
+
+#[test]
+public fun test_direct_extend_happy(): system::System {
+    let mut ctx = tx_context::dummy();
+
+    // A test coin.
+    let mut fake_coin = coin::mint_for_testing<SUI>(100000000, &mut ctx);
```
damirka: There's a utility for that in `tests/staking/test_utils.move`. Maybe we could move it higher in the tree so it's used both in staking and system.

damirka: It reduces number of deps in tests and helps refactor in the future.

contracts/walrus/sources/system.move
karlwuest: That's a good question. We've just been using `u64` for epochs everywhere, and I think we should keep it consistent. However, a `u32` should be sufficient as well (at least for the next few million years). I'm not sure how much we gain from changing it, but it also shouldn't hurt. We can maybe quickly discuss this with others in the engineering sync or on slack.

contracts/walrus/tests/system/blob_tests.move
karlwuest: Done! I also integrated `bls_test_utils` into `test_utils`

contracts/walrus/sources/system.move
karlwuest: Started a discussion on slack and created #694 to keep track of this.

contracts/walrus/sources/system/blob.move
```diff
@@ -166,12 +186,24 @@ public(package) fun certify_with_certified_msg(
...
-        message.certified_epoch(),
-        message.certified_blob_id(),
-        blob.storage.end_epoch(),
-    );
+    blob.emit_certified(false);
+}
+
+/// Deletes a deletable blob and returns the contained storage.
+/// Emits a `BlobDeleted` event for the given epoch.
+public(package) fun delete(self: Blob, epoch: u64): Storage {
```
gdanezis, line 196: Here should we not consider:
- That the blob was registered but not certified?
- That the current epoch > the end of storage epoch, and the meta-data may have been deleted from the nodes?
- That the blob may be deleted if it is not deletable but the storage has expired?

contracts/walrus/sources/system/events.move
```diff
@@ -0,0 +1,86 @@
...
+    size: u64,
+    erasure_code_type: u8,
+    end_epoch: u64,
+    deletable: bool,
+    // The object id of the related `Blob` object
+    object_id: ID,
+}
+
+/// Signals that a blob is certified.
+public struct BlobCertified has copy, drop {
```
gdanezis, line 25: Should we add here whether the Blob is deletable, because this is the event that a light-client would authenticate to provide a proof of availability. And this proof needs to certify the deletable property.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -245,6 +220,71 @@ public(package) fun extend_blob_with_resource(
     blob.extend_with_resource(extension, self.epoch());
 }
 
+/// Extend the period of validity of a blob by extending its contained storage resource.
+public(package) fun extend_blob(
+    self: &mut SystemStateInnerV1,
```
gdanezis, line 224: Is there a version of extend that takes a storage resource instead?

contracts/walrus/sources/system/events.move
karlwuest: That's a good point, I will add this!

contracts/walrus/tests/system/blob_tests.move
```diff
@@ -668,19 +684,304 @@ public fun test_blob_extend_bad_period(): system::System {
...
+    assert!(blob1.storage().end_epoch() == 6);
+
+    coin::burn_for_testing<SUI>(fake_coin);
+    blob1.burn();
+    system
+}
+
+#[test, expected_failure(abort_code = blob::ENotCertified)]
+public fun test_direct_extend_not_certified(): system::System {
+    let mut ctx = tx_context::dummy();
```
damirka: Hint: given that we never use context by value and it has `drop`, it's more convenient to write:
```
let ctx = &mut tx_context::dummy();
``` 

Shorter + you don't have to think about `&mut ctx` anywhere else.

contracts/walrus/sources/system/blob.move
karlwuest: Good questions!

>That the blob was registered but not certified?

Do you have a scenario in mind where this would make a difference? I thought about this, but I don't think it should matter. The storage nodes that receive the BlobDeleted event anyway know what the current status of the blob is. If the blob has been registered but not certified yet, the nodes that already stored some information still need to delete it, and if there is nothing stored, then there just isn't anything to do (except updating the status).

>That the current epoch > the end of storage epoch, and the meta-data may have been deleted from the nodes?

My thinking was that the storage nodes are aware of the blob status and if they are not storing it anymore, they can just ignore the event. However, adding a check for this is pretty simple and prevents unnecessary events from being created, so I will do that.

>That the blob may be deleted if it is not deletable but the storage has expired?

This would just be to delete the blob object on Sui, right? In that case, we can just use the `burn` function, since we also don't want to return a storage object. Note that for destroying the blob object, we don't really care if it's expired, so any blob object can be destroyed by its owner (that doesn't delete the blob on Walrus, of course).

contracts/walrus/sources/system/system_state_inner.move
karlwuest: Yes, that's the `extend_blob_with_resource` function.

contracts/walrus/sources/system/blob.move
karlwuest: @gdanezis: can I resolve this conversation?

karlwuest: @gdanezis I will merge this now. Lmk if you think that we should do a follow up on one of the questions that you raised.

mlegner: @karlwuest Related to @gdanezis's question above: Could we maybe add a field to the event with the previous status of the blob? This would make it more efficient to keep track of the blob status on storage nodes.

karlwuest: Adding this shouldn't be a problem, but why would this make it more efficient to keep track of the status?

mlegner: Some context from the offline discussion:

I'm currently investigating multiple options of keeping an aggregate status for a blob ID. The most efficient way in RocksDB is to implement a custom merge operator, which gets two inputs: The current (aggregate) status and the merge operand.

If I want to keep track of the number of registered and certified deletable blobs, the merge operator needs to contain the type of blob that was deleted. And if that isn't in the event, I need to look it up in the DB, which defeats the purpose of the merge operator.

karlwuest: Added #730

PR #697, opened 2024-08-19
feat(service): add endpoint for sliver status

giac-mysten:
Adds a new endpoint, `SLIVER_STATUS_ENDPOINT`, to check whether a sliver is stored with a storage node.

Rationale: 
* we want to avoid the cost of reading the sliver form the database. Therefore, making a `HEAD` request to the sliver endpoint and then dropping the body is not a good option;
* A custom `HEAD` implementation on the `SLIVER_ENDPOINT` would have the wrong semantics, as it won't have the correct headers
* Therefore, we add a new endpoint to check the sliver status.

### DIFF
--- a/crates/walrus-service/src/node/server/routes.rs
+++ b/crates/walrus-service/src/node/server/routes.rs
@@ -231,7 +231,11 @@ pub async fn put_sliver<S: SyncServiceState>(
         ("sliver_type" = SliverType, ),
     ),
     responses(
-        (status = 200, description = "The storage status of the primary or secondary sliver", body=ApiSuccessSliverStatus),
+        (
+            status = 200,
+            description = "The storage status of the primary or secondary sliver",
+            body=ApiSuccessSliverStatus,
+        ),
         RetrieveSliverError,
     ),
     tag = openapi::GROUP_READING_BLOBS


crates/walrus-service/src/node/server/routes.rs
```diff
@@ -208,6 +212,51 @@ pub async fn put_sliver<S: SyncServiceState>(
...
+#[utoipa::path(
+    get,
+    path = api::rewrite_route(SLIVER_STATUS_ENDPOINT),
+    params(
+        ("blob_id" = BlobIdString, ),
+        ("sliver_pair_index" = SliverPairIndex, ),
+        ("sliver_type" = SliverType, ),
+    ),
+    responses(
+        (status = 200, description = "The primary or secondary sliver is stored and available", body=ApiSuccessSliverStatus),
```
jpcsmith: Suggestion: Update the description, this returns 200 also when the sliver is not stored.

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -208,6 +212,51 @@ pub async fn put_sliver<S: SyncServiceState>(
     Ok(ApiSuccess::ok("sliver stored successfully"))
 }
 
+/// Check if the blob slivers are present.
+///
+///
```
mlegner: Nit:
```suggestion
```

PR #698, opened 2024-08-20
feat: add event definitions to new contracts

### LINKED ISSUE #696: Implement all required event types in the contracts

karlwuest:
closes #696

### DIFF
--- a/contracts/walrus/sources/system/events.move
+++ b/contracts/walrus/sources/system/events.move
@@ -48,29 +48,30 @@ public struct InvalidBlobID has copy, drop {
     blob_id: u256,
 }
 
-/// Signals that the epoch change has started.
+/// Signals that epoch `epoch` has started and the epoch change is in progress.
 public struct EpochChangeStart has copy, drop {
     epoch: u64,
 }
 
-/// Signals that nodes with 2f+1 shards have finished the epoch change.
+/// Signals that a set of storage nodes holding at least 2f+1 shards have finished the epoch
+/// change, i.e., received all of their assigned shards.
 public struct EpochChangeDone has copy, drop {
     epoch: u64,
 }
 
-/// Signals that a specific node has received all shards for the new epoch.
-public struct AllShardsReceived has copy, drop {
+/// Signals that a node has received the specified shards for the new epoch.
+public struct ShardsReceived has copy, drop {
     epoch: u64,
-    node_id: ID,
+    shards: vector<u16>,
 }
 
-/// Signals that the committee for the next epoch has been selected.
-public struct CommitteeSelected has copy, drop {
-    epoch: u64,
+/// Signals that the committee and the system parameters for `next_epoch` have been selected.
+public struct EpochParametersSelected has copy, drop {
+    next_epoch: u64,
 }
 
-/// Signals that the happy path shard transfer has failed for the given shards.
-public struct ShardTransferFailed has copy, drop {
+/// Signals that the given shards can be recovered using the shard recovery endpoint.
+public struct ShardRecoveryStart has copy, drop {
     epoch: u64,
     shards: vector<u16>,
 }
@@ -124,14 +125,14 @@ public(package) fun emit_epoch_change_done(epoch: u64) {
     event::emit(EpochChangeDone { epoch })
 }
 
-public(package) fun emit_all_shards_received(epoch: u64, node_id: ID) {
-    event::emit(AllShardsReceived { epoch, node_id })
+public(package) fun emit_shards_received(epoch: u64, shards: vector<u16>) {
+    event::emit(ShardsReceived { epoch, shards })
 }
 
-public(package) fun emit_committee_selected(epoch: u64) {
-    event::emit(CommitteeSelected { epoch })
+public(package) fun emit_epoch_parameters_selected(next_epoch: u64) {
+    event::emit(EpochParametersSelected { next_epoch })
 }
 
-public(package) fun emit_shard_transfer_failed(epoch: u64, shards: vector<u16>) {
-    event::emit(ShardTransferFailed { epoch, shards })
+public(package) fun emit_shard_recovery_start(epoch: u64, shards: vector<u16>) {
+    event::emit(ShardRecoveryStart { epoch, shards })
 }


contracts/walrus/sources/system/events.move
```diff
@@ -48,6 +48,35 @@ public struct InvalidBlobID has copy, drop {
...
+
+/// Signals that a specific node has received all shards for the new epoch.
+public struct AllShardsReceived has copy, drop {
+    epoch: u64,
+    node_id: ID,
+}
+
+/// Signals that the committee for the next epoch has been selected.
+public struct CommitteeSelected has copy, drop {
+    epoch: u64,
```
gdanezis: Should we provide a hint about where to find this committee? Like the object ID, or even inline the committee in the event?

karlwuest: Storage nodes and clients anyway need to have the system object configured, so I'm not sure how much we gain from including its object ID. I thought about inlining the committee, but this might be a bit inflexible. If we ever want to change anything in the internal representation of the committee, we would need a new event version.

contracts/walrus/sources/system/events.move
```diff
@@ -48,6 +48,35 @@ public struct InvalidBlobID has copy, drop {
     blob_id: u256,
 }
 
+/// Signals that the epoch change has started.
+public struct EpochChangeStart has copy, drop {
+    epoch: u64,
+}
```
mlegner: Nit: I assume the `epoch` refers to the *new* epoch. Can we state this explicitly somewhere in a comment?

contracts/walrus/sources/system/events.move
```diff
@@ -48,6 +48,35 @@ public struct InvalidBlobID has copy, drop {
...
+/// Signals that the committee for the next epoch has been selected.
+public struct CommitteeSelected has copy, drop {
+    epoch: u64,
+}
+
+/// Signals that the happy path shard transfer has failed for the given shards.
+public struct ShardTransferFailed has copy, drop {
+    epoch: u64,
+    shards: vector<u16>,
+}
```
giac-mysten: question: is this emitted by the system object after the timeout for the happy path elapses?

contracts/walrus/sources/system/events.move
mlegner: Suggestion: It is a bit unclear if the `epoch` field refers to the current epoch or the epoch for which the committee has been selected. Can we clarify this?

contracts/walrus/sources/system/events.move
```diff
@@ -48,6 +48,35 @@ public struct InvalidBlobID has copy, drop {
...
+
+/// Signals that a specific node has received all shards for the new epoch.
+public struct AllShardsReceived has copy, drop {
+    epoch: u64,
+    node_id: ID,
+}
+
+/// Signals that the committee for the next epoch has been selected.
+public struct CommitteeSelected has copy, drop {
```
mlegner: Question: Does this include all the other system parameters for the next epoch? If so, maybe a different name like `SystemParametersFinalizedForEpoch` would be clearer.

contracts/walrus/sources/system/events.move
```diff
@@ -48,6 +48,35 @@ public struct InvalidBlobID has copy, drop {
...
+/// Signals that nodes with 2f+1 shards have finished the epoch change.
+public struct EpochChangeDone has copy, drop {
+    epoch: u64,
+}
+
+/// Signals that a specific node has received all shards for the new epoch.
+public struct AllShardsReceived has copy, drop {
+    epoch: u64,
+    node_id: ID,
+}
```
mlegner: Question: Where do we need/use this event?

contracts/walrus/sources/system/events.move
```diff
@@ -48,6 +48,35 @@ public struct InvalidBlobID has copy, drop {
     blob_id: u256,
 }
 
+/// Signals that the epoch change has started.
+public struct EpochChangeStart has copy, drop {
+    epoch: u64,
+}
+
+/// Signals that nodes with 2f+1 shards have finished the epoch change.
```
mlegner: Important: There are several possible semantics of this:
1. Nodes who together hold more than 2f shards have all the shards they should have.
2. More than 2f shards are stored at their owners.

The comment is consistent with version 1. However, under some not too unlikely circumstances, this will significantly delay this event:
Consider 4 storage nodes with the same number of shards. In an epoch change, two gain one shard each from the other two. In this case all shards except 2 are stored at their owners, but we still need to wait for at least one of the shard transfers to succeed before we can finish the epoch change.

My suggestion would therefore be to go with version 2.

contracts/walrus/sources/system/events.move
karlwuest: This will be emitted after a node claims that it did not receive a (set of) shards.

contracts/walrus/sources/system/events.move
karlwuest: This allows doing a rolling update of the committee that we read from. @jpcsmith was planning to implement such a rolling update.

contracts/walrus/sources/system/events.move
giac-mysten: What if the node does not emit this? Also, is there no global event for the end of the happy path?

contracts/walrus/sources/system/events.move
mlegner: OK, makes sense (for the storage nodes). However, that would work even better by having a `ShardTransferred` event for each transferred shard (non-transferred shards don't need this). This would be consistent with my [suggestion](https://github.com/MystenLabs/walrus/pull/698/files#r1723449312) for the `EpochChangeDone` event.

contracts/walrus/sources/system/events.move
```diff
@@ -48,6 +48,35 @@ public struct InvalidBlobID has copy, drop {
...
+    node_id: ID,
+}
+
+/// Signals that the committee for the next epoch has been selected.
+public struct CommitteeSelected has copy, drop {
+    epoch: u64,
+}
+
+/// Signals that the happy path shard transfer has failed for the given shards.
+public struct ShardTransferFailed has copy, drop {
```
halfprice: Can you explain how we would use this event?

IIRC during the offsite, we discussed that the happy path to non happy path transition is driven by the node itself as opposite to driving by the contract. If that is still the case, from the contract perspective, it seems that the contract only need to know whether a node is successfully migrated, and if any shards are migrated using non happy path route?

contracts/walrus/sources/system/events.move
karlwuest: Version 1 is the intended meaning. I can clarify that in the contract. I think we will want to use that semantic at least for testnet for simplicity (plus we use the same for blob certificates).
I agree that version 2 is nicer long term, but it also adds more complexity. It requires the storage nodes to send potentially multiple transactions and it requires keeping track of all synced shards in the staking object, where we need to make sure that we don't run into limits for object size or dynamic field lookups.

contracts/walrus/sources/system/events.move
mlegner: AFAIU the idea is that other storage nodes only help with shard recovery once such an event is emitted.

contracts/walrus/sources/system/events.move
karlwuest: There is no global end of the happy path. 

If a storage node did not receive a shard and it doesn't call the function to emit this, then it will not be able to use the shard recovery endpoint (eventually, don't know if this access control will be implemented for testnet, cc @jpcsmith ). It's still open whether we are slashing the receiving node at the end of the epoch if they have called neither the "happy path failed" function nor the "epoch change done" function or whether they will just get slashed through failing challenges + having to pay for the shard recovery endpoint once they do want to recover the shard.

Sidenote: I will change the name of this event (edit: to `ShardRecoveryStart`), because the goal of this is to notify storage nodes that the receiving node can use the shard recovery endpoint. Since we also need shard recovery if, e.g., the hard drive fails, we can use the same event (but a different function to initiate this).

contracts/walrus/sources/system/events.move
karlwuest: Added this to the comment.

contracts/walrus/sources/system/events.move
karlwuest: Renamed to `EpochParametersSelected`

contracts/walrus/sources/system/events.move
karlwuest: Renamed to `next_epoch`

contracts/walrus/sources/system/events.move
giac-mysten: Ok, thanks for the explanation! 

I had missed the discussion on disabling the recovery endpoint during epoch change. Do we foresee that the recovery endpoint will be locked only for old blobs, such that newly-published blobs can be correctly fetched by the new committee?

also, what if the epoch change begins while a node in the old committee is still recovering a blob that it did not receive the slivers for?

contracts/walrus/sources/system/events.move
karlwuest: Yes, what @mlegner said. The idea is that the storage node sends a transaction to the contract that the happy path transfer was unsuccesfull. This will then emit this event, which then let's the other nodes know that they should allow access to the shard recovery endpoint.

I now renamed the event, because we need the same if shard recovery is necessary for benign reasons (e.g. hard drive failure). In that case, the idea is that the node that needs shard recovery can use the contract to initiate shard recovery (either with a payment or by accepting slashing, tbd) which then will also emit this event and allow access to the endpoint.

contracts/walrus/sources/system/events.move
karlwuest: >I had missed the discussion on disabling the recovery endpoint during epoch change. Do we foresee that the recovery endpoint will be locked only for old blobs, such that newly-published blobs can be correctly fetched by the new committee?

My understanding is that the shard recovery endpoint is separate from the blob recovery endpoint (@halfprice and @jpcsmith correct me if that is wrong).

I think we discussed limiting the blob recovery endpoint to recent blobs (i.e. certified in the current or previous epoch) in Lisbon, but I don't recall if we made a decision on this or if this would just be covered by rate limiting (to prevent using the blob recovery endpoint as alternative to the shard recovery endpoint).

contracts/walrus/sources/system/events.move
karlwuest: I changed this now to contain a list of shards instead of `node_id` and renamed to `ShardsReceived`. However, at least initially, as also mentioned in the other comment, this will likely not be emitted incrementally, but just once per node.

contracts/walrus/sources/system/events.move
mlegner: OK, makes sense to me. 

Suggestion: Create an issue for the Mainnet milestone to make this more fine-grained.

contracts/walrus/sources/system/events.move
giac-mysten: Really cool! I had missed this conversation unfortunately. Thanks for going over it again.

contracts/walrus/sources/system/events.move
karlwuest: #706

contracts/walrus/sources/system/events.move
jpcsmith: >  eventually, don't know if this access control will be implemented for testnet, 

Limiting access based on the events may not make it in for testnet, but we'll see.

> My understanding is that the shard recovery endpoint is separate from the blob recovery endpoint (@halfprice and @jpcsmith correct me if that is wrong).

Yea, that's correct, the shard recovery is for bulk transfer.

contracts/walrus/sources/system/events.move
karlwuest: @gdanezis, I'll resolve this now. Please lmk or create an issue if you feel strongly about inlining this or referencing the system object.

PR #702, opened 2024-08-20
feat(move): add `register_thrifty` function that minimizes storage resource use

sblackshear:
`blob::register` allows the storage resource to be "bigger" than the blob being registered:

```
assert!(encoded_size <= storage_size(&storage), EResourceSize);
```

This is harmless from the system's perspective, but potentially troubling for a user who uses the wrong storage resource to register a blob.

In addition to overpaying up front, `blob::extend` looks at the size of the storage resource locked in the blob, so a user who accidentally overpays once will also need to overpay upon extension.

This PR leaves the old `register` functionality in place for now, but adds a new `register_thrifty` function that splits the input storage resource into exactly the right size + returns the remainder to the user. We could also consider changing the behavior of `register` under the hood (likely will change nothing in practice given that the CLI/SDKs currently create a storage resource exactly the right size)

### DIFF
--- a/contracts/blob_store/Move.toml
+++ b/contracts/blob_store/Move.toml
@@ -3,7 +3,8 @@ name = "blob_store"
 edition = "2024.beta"
 
 [dependencies]
-Sui = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-framework/packages/sui-framework", rev = "testnet-v1.31.1" }
+#Sui = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-framework/packages/sui-framework", rev = "testnet-v1.31.1" }
+Sui = { local = "../../../sui/crates/sui-framework/packages/sui-framework" }
 
 [addresses]
 blob_store = "0x0"
--- a/contracts/blob_store/sources/blob.move
+++ b/contracts/blob_store/sources/blob.move
@@ -91,6 +91,55 @@ module blob_store::blob {
         blob_id
     }
 
+    /// Variant of register that refunds leftover storage resources if `size` is smaller than
+    /// `storage.size`
+    public fun register_thrifty<WAL>(
+        sys: &System<WAL>,
+        mut storage: Storage,
+        blob_id: u256,
+        root_hash: u256,
+        size: u64,
+        erasure_code_type: u8,
+        ctx: &mut TxContext,
+    ): (Blob, Option<Storage>) {
+        let storage_size = storage.storage_size();
+        // check that the encoded size is less than the storage size
+        let encoded_size = encoding::encoded_blob_length(
+            size,
+            erasure_code_type,
+            sys.n_shards(),
+        );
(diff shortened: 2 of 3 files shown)


contracts/blob_store/Move.toml
mlegner, line 1: These changes should probably not be committed.

contracts/blob_store/sources/blob.move
mlegner, line 1: The changes should be placed in `contracts/walrus/...`: The `blob_store` contract is the "legacy" contract we use for Devnet, all new contracts for Testnet and beyond are in the `walrus` directory.

contracts/blob_store/sources/blob.move
```diff
@@ -91,6 +91,55 @@ module blob_store::blob {
...
+        root_hash: u256,
+        size: u64,
+        erasure_code_type: u8,
+        ctx: &mut TxContext,
+    ) {
+        let (blob, storage_opt) = register_thrifty(sys, storage, blob_id, root_hash, size, erasure_code_type, ctx);
+        transfer::transfer(blob, ctx.sender());
+        // doesn't work because option::destroy is buggy :(
+        //storage_opt.destroy!(|storage_resource| transfer::transfer(storage_resource, ctx.sender()))
+        if (storage_opt.is_some()) {
```
damirka, line 136:
```suggestion
storage_opt.do!(|opt| transfer::public_transfer(opt, ctx.sender()));
```

contracts/blob_store/Move.toml
damirka: +1

sblackshear: Whoops, apologies for this!

contracts/blob_store/sources/blob.move
sblackshear: Unfortunately, this doesn't work for the same reason my commented-out code doesn't work (`destroy` delegates to `do`). I think `do!` is missing a `else o.destroy_none()` that would fix this.

sblackshear:
```error[E06001]: unused value without 'drop'
    ┌─ ./sources/blob.move:136:76
    │
104 │     ): (Blob, Option<Storage>) {
    │               ---------------
    │               │      │
    │               │      The type 'std::option::Option<blob_store::storage_resource::Storage>' can have the ability 'drop' but the type argument 'blob_store::storage_resource::Storage' does not have the required ability 'drop'
    │               The type 'std::option::Option<blob_store::storage_resource::Storage>' does not have the ability 'drop'
    ·
136 │         storage_opt.do!(|opt| transfer::public_transfer(opt, ctx.sender()));
    │                                                                            ^ Invalid return
    │
    ┌─ ./../../../sui/crates/sui-framework/packages/move-stdlib/sources/option.move:155:13
    │
155 │         let o = $o;
    │             - The local variable 'o' might still contain a value. The value does not have the 'drop' ability and must be consumed before the function returns```

Issue #523, opened 2024-06-13
Improve blob browsing experience


mlegner:
Originally reported by @halfprice offline:

> I wish there could be a better blob management flow. Right now, I have to remember the blob id myself, and if I lost it, it's hard for me to find what files I uploaded. The only way seems to be [through the explorer](https://testnet.suivision.xyz/object/0x015f11922be121c295d760bbe25551f4c62eda4b2907a2e16f34c20460fd3ec7), and the object id there is in decimal format instead of hex.

- [x] It would be nice to be able to get a list of owned blob IDs (in correct base64 encoding) through the CLI. -> This is tracked in #262 
- [x] We should also document how the blob ID in the Sui explorer can be converted to a blob ID in base64. Alternatively, we could also accept numeric values in decimal or hex.

PR #707, opened 2024-08-21
feat(client): provide help for blob IDs in decimal format

mlegner:
- provide helpful error to users trying to use blob IDs in decimal format
- add command to convert blob IDs between decimal and base64 format

Contributes to #523

### DIFF
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -265,10 +265,11 @@ enum Commands {
         #[serde(flatten)]
         rpc_arg: RpcArg,
     },
+    /// Convert a decimal value to the Walrus blob ID (using URL-safe base64 encoding).
     ConvertBlobId {
-        #[clap(flatten)]
-        #[serde(flatten)]
-        blob_id: BlobIdDecimalOrBase64,
+        /// The decimal value to be converted to the Walrus blob ID.
+        #[serde_as(as = "DisplayFromStr")]
+        blob_id_decimal: BlobIdDecimal,
     },
 }
 
@@ -352,23 +353,6 @@ impl FileOrBlobId {
             _ => Err(anyhow!("either the file or blob ID must be defined")),
         }
     }
-}
-
-#[serde_as]
-#[derive(Debug, Clone, Args, Deserialize)]
-#[serde(rename_all = "camelCase")]
-#[group(required = true, multiple = false)]
-struct BlobIdDecimalOrBase64 {
-    /// The blob ID in decimal format.
-    #[clap(short, long)]
-    #[serde_as(as = "Option<DisplayFromStr>")]
-    #[serde(default)]
-    decimal: Option<BlobIdDecimal>,
-    /// The blob ID in URL-safe base64 format.
-    #[clap(short, long, allow_hyphen_values = true)]
-    #[serde_as(as = "Option<DisplayFromStr>")]
-    #[serde(default)]
-    base64: Option<BlobId>,
 }
 
 mod default {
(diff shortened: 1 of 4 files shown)


crates/walrus-service/bin/client.rs
```diff
@@ -625,6 +657,26 @@ async fn run_app(app: App) -> Result<()> {
...
+                    ..
+                } => {
+                    BlobIdConversionOutput::from(blob_id).print_output(app.json)?;
+                }
+                // This case is required for JSON mode where we don't have the clap checking.
+                _ => Err(anyhow!(
+                    "the blob ID must be provided either in decimal or base64 format"
+                ))?,
+            }
+        }
```
karlwuest: Suggestion/Remark: I don't think that we should provide conversion both ways. I think it would be better to only provide conversion from decimal to base64 but not the other way around to reinforce the base64 format as default.

mlegner: That's a good point. I went a bit back and forth on this. My idea was that users can then compare the numeric value to what is shown on the Sui explorers. But then again, they can just take the explorer value and convert it to the standard format. 🤷 

So no strong opinion either way, I'm happy to remove the base64-to-decimal conversion (makes everything a bit simpler anyway).

crates/walrus-service/bin/client.rs
```diff
@@ -340,6 +354,23 @@ impl FileOrBlobId {
     }
 }
 
+#[serde_as]
+#[derive(Debug, Clone, Args, Deserialize)]
+#[serde(rename_all = "camelCase")]
+#[group(required = true, multiple = false)]
+struct BlobIdDecimalOrBase64 {
```
jpcsmith: Nit: Odd that this isn't an enum.

crates/walrus-service/bin/client.rs
jpcsmith: An alternative, in the vein of the current solution, is to refuse to convert it but convert it anyway 😆. If the user provides a base64 value, return an error saying that the value is already a valid BlobId and mention the u256 value in the error. If users actually end up depending on it for conversions, as opposed to just stumbling upon the error while learning walrus, they will likely voice that their desire for the bidirectional conversion.

crates/walrus-service/src/client/cli_utils.rs
```diff
@@ -287,6 +291,77 @@ pub fn format_event_id(event_id: &EventID) -> String {
     format!("(tx: {}, seq: {})", event_id.tx_digest, event_id.event_seq)
 }
 
+/// Error type distinguishing between a valid blob ID in decimal format and any other parse error.
+#[derive(Debug, thiserror::Error)]
+pub enum BlobIdParseError {
+    /// Attempting to parse a decimal value for the blob ID.
+    #[error(
+        "you seem to be using the numeric value for the blob ID (maybe copied from a Sui explorer) \
+        whereas Walrus uses URL-safe base64 encoding;\nthe provided value correctly encoded is {0}"
```
jpcsmith: Suggestion: Make it explicit that the blob ID *is* a URL-safe base64 value.

crates/walrus-service/src/client/cli_utils/cli_output.rs
```diff
@@ -165,6 +166,19 @@ impl CliOutput for BlobStatusOutput {
...
 }
 
+impl CliOutput for BlobIdConversionOutput {
+    fn print_cli_output(&self) {
+        match self {
+            BlobIdConversionOutput::Base64(blob_id) => {
+                println!("Blob ID in standard format: {}", blob_id)
+            }
+            BlobIdConversionOutput::Decimal(blob_id_decimal) => {
+                println!("Blob ID in decimal format: {}", blob_id_decimal)
```
jpcsmith: Nit: I would avoid phrasing this as "formats" as the blob ID. And instead print it as "Blob ID" and "Blob ID as a decimal".

crates/walrus-service/bin/client.rs
mlegner: Unfortunately, this is not supported by clap yet: https://github.com/clap-rs/clap/issues/2621

crates/walrus-service/bin/client.rs
mlegner: > refuse to convert it but convert it anyway

Haha, I love that solution. Will implement that and emphasize that the base64 encoding **is** the blob ID (as suggested in other comments).

crates/walrus-service/bin/client.rs
jpcsmith: Remark: Would accepting it as a string and attempting to parse as a u256, then as a blob id would be cleaner?

mlegner: Now that I'm going to accept only the decimal format anyway, I can simplify this and remove the multiple options.

PR #708, opened 2024-08-21
feat: only handle sync shard request if src node also in the new epoch

halfprice:
Previously, we don't check if the source and destination storage nodes are in the same epoch or not. The idea was that we want the source node to start sending slivers to the destination even if it hasn't received epoch change event. This means that the source node may not see all certified blob up until the epoch boundary yet.

This optimization adds quite a bit complexity, so in this PR, we enforce that the source storage node should also see the epoch change event, and has moved to the new epoch before serving any sync shard request. For honest node, this means that it contains the full list of certified blobs and the bulk shard transfer should just send all certified blobs' shard to the destination node. This also makes bulk shard transfer easy to reason about.

For the optimization mentioned earlier, we can implement it later if there is a larger delay if epoch change events are received by different storage node.

Contribute to #259

### DIFF
--- a/crates/walrus-sdk/src/error.rs
+++ b/crates/walrus-sdk/src/error.rs
@@ -74,7 +74,7 @@ pub(crate) enum Kind {
     Bcs(#[from] bcs::Error),
     #[error(transparent)]
     Reqwest(#[from] reqwest::Error),
-    #[error("{inner}: {message} ({service_error:?})")]
+    #[error("{inner}: {message}")]
     StatusWithMessage {
         inner: reqwest::Error,
         message: String,
@@ -123,7 +123,7 @@ pub enum ServiceError {
     /// The requested epoch is invalid.
     InvalidEpoch {
         /// The epoch client is in.
-        client_epoch: Epoch,
+        request_epoch: Epoch,
         /// The epoch server is in.
         server_epoch: Epoch,
     },
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -957,7 +957,7 @@ impl ServiceState for StorageNodeInner {
         // In a normal scenario, a storage node will never fetch shards from a future epoch.
         if request.epoch() != self.current_epoch() {
             return Err(InvalidEpochError {
-                client_epoch: request.epoch(),
+                request_epoch: request.epoch(),
                 server_epoch: self.current_epoch(),
             }
             .into());
--- a/crates/walrus-service/src/node/errors.rs
+++ b/crates/walrus-service/src/node/errors.rs
@@ -123,9 +123,9 @@ pub enum BlobStatusError {
 
 /// Error returned when the epoch in a request is invalid.
 #[derive(Debug, thiserror::Error, Serialize, Clone)]
-#[error("Invalid epoch. Client epoch: {client_epoch}. Server epoch: {server_epoch}")]
+#[error("Invalid epoch. Client epoch: {request_epoch}. Server epoch: {server_epoch}")]
 pub struct InvalidEpochError {
-    pub client_epoch: Epoch,
+    pub request_epoch: Epoch,
     pub server_epoch: Epoch,
 }
 
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -124,15 +124,15 @@ impl ShardSyncHandler {
 
                     if let SyncShardClientError::RequestError(node_error) = err {
                         if let Some(ServiceError::InvalidEpoch {
-                            client_epoch,
+                            request_epoch,
                             server_epoch,
                         }) = node_error.service_error()
                         {
-                            if client_epoch > server_epoch {
+                            if request_epoch > server_epoch {
                                 tracing::info!(
                                     "Source storage node hasn't reached the epoch yet.
                                 Client epoch: {}, Server epoch: {}",
-                                    client_epoch,
+                                    request_epoch,
                                     server_epoch
                                 );
                                 // Retry the sync after backoff.
--- a/crates/walrus-service/src/node/server/responses.rs
+++ b/crates/walrus-service/src/node/server/responses.rs
@@ -111,8 +111,10 @@ rest_api_error!(
         (MessageVerificationError(_), BAD_REQUEST, None, "Request verification failed"),
         (ShardNotAssigned(_), MISDIRECTED_REQUEST, None,
         "the requested sliver is not stored at a shard assigned to this storage node"),
-        (InvalidEpoch(InvalidEpochError{client_epoch, server_epoch}), BAD_REQUEST,
-        Some(ServiceError::InvalidEpoch{client_epoch: *client_epoch, server_epoch: *server_epoch}),
+        (InvalidEpoch(InvalidEpochError{request_epoch, server_epoch}), BAD_REQUEST,
+        Some(ServiceError::InvalidEpoch{
+            request_epoch: *request_epoch,
+            server_epoch: *server_epoch}),
         "The requested epoch is invalid"),
         (Internal(_), INTERNAL_SERVER_ERROR, None, @canonical),
         (StorageError(_), INTERNAL_SERVER_ERROR, None, "Storage error"),


crates/walrus-service/src/node.rs
```diff
@@ -894,6 +895,13 @@ impl ServiceState for StorageNodeInner {
             ));
         }
 
+        // Wait until the current storage node receives the epoch change event, so that the node
+        // contains the full list of certified blobs to send to the requester.
+        utils::retry(FixedIntervalRetry::new(Duration::from_secs(1)), || async {
+            request.epoch() == self.current_epoch()
+        })
+        .await;
```
jpcsmith: Suggestion: Do not tie up resources on the server waiting for this, just return an error saying that the request is not valid for the current epoch. The caller can retry.

halfprice: Hi @jpcsmith , I made a pass based on the discussion last Friday, that now the server can return a specific WalrusServiceError serialize into json on the server side, deserialize it into a WalrusServiceError on the client side.

Right now, I only put one error, InvalidEpoch, in WalrusServiceError for the purpose of this PR's implementation. Not only it is going to require quite some effort to move the rest of server side error into the new place, but also, I just want to make sure that this approach is actually the right way to do it.

Can you PTAL?

crates/walrus-sdk/src/error.rs
```diff
@@ -73,6 +83,12 @@ pub(crate) enum Kind {
     ErrorInNonErrorMessage { code: u16, message: String },
     #[error("invalid content type in response")]
     InvalidContentType,
+    #[error("{inner}: {message}. Detailed reason: {reason:?}")]
```
jpcsmith:
```suggestion
    #[error("{message} ({reason})")]
```

crates/walrus-sdk/src/error.rs
```diff
@@ -104,3 +120,12 @@ pub(crate) enum BuildErrorKind {
...
 }
+
+/// Defines a more detailed server side reason that can be returned with an error.
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+pub enum ServiceErrorReason {
+    /// The requested epoch is invalid because it is too old.
+    InvalidEpochTooOld(Epoch),
+    /// The requested epoch is invalid because it is too new.
+    InvalidEpochTooNew(Epoch),
+}
```
jpcsmith: Suggestion: Remove the epoch, this duplicates the information that you already have in the error.

Do you need the Epoch? If so, then we'd need to already move towards a more final representation. Consider the `StoreSliverError` and `RetrieveSliverError`, they both have a `ShardNotAssigned` variant:

```rust
#[derive(Debug, thiserror::Error)]
pub enum StoreSliverError {
    #[error(transparent)]
    ShardNotAssigned(#[from] ShardNotAssigned),
    ...
}
#[derive(Debug, thiserror::Error)]
pub enum RetrieveSliverError {
    #[error(transparent)]
    ShardNotAssigned(#[from] ShardNotAssigned),
    ...
}
```
At the client, we could have `ServiceError` (no `Reason`) with
```rust
enum ServiceError {
   ShardNotAssigned,
   InvalidEpoch { server_epoch: Epoch },
   ...
}
```
The goal is then to communicate the error from the server to the client. This would look something like the following in
```json
{ 
   "error": {
       "message": "<error message goes here>",
       "reason": "shardNotAssigned",
       "metadata": null
    }
}
```
The `ServiceError::ShardNotAssigned` could then be reconstructed with by checking the reason, or a `ServiceError::InvalidEpoch` could be reconstructed from 
```json
{ 
   "error": {
       "message": "<error message goes here>",
       "reason": "invalidEpoch",
       "metadata": {
           "server_epoch": "17",
       }
    }
}
```

If you already want to go down this route of detailed errors at the client, then the fastest way would be perhaps to rename the `ServiceErrorReason` enum to a `ServiceError` enum, then cast the errors returned from the axum server from their original type into a `ServiceError` and then serialise to the above format.

By contrast, my original suggestion was just, as a next step in this evolution, to just define the reason that is attached to each error, without the metadata.

halfprice: Yes, the `epoch` information is needed here for the destination storage node to decide which type of shard recovery to use next. If the epoch is too far apart, the node needs to go into using `recovery` instead of `sync` directly. Passing this information more explicitly allows the client to not rely on parsing the message string, which is preferred.

This is the output from the previous revision, I think it's quite close to what you suggested.

```
{
  "error": {
    "code": 400,
    "message": "Requester epoch too old. Server epoch: 10",
    "reason": {
      "InvalidEpochTooOld": 10
    }
  }
}
```

I just made an other update. In the new revision, I renamed ServiceErrorReason with ServiceError, and in the NodeError constructed in the client side, there is a more specific branch where the client can get the passed ServiceError object. So RPC -> NodeError -> ServiceError.

The main difference from this to the new suggestion is that in the error json, currently the `reason` is a serialized ServiceError object instead of breaking ServiceError object into a plain `reason` and serialized `metadata` field. It is going to require some effort to break them apart as we need to define another metadata object to do the ServiceError <-> [Reason, Metadata] conversion which I think it requires its own PR. (also, is it necessary to break them apart? It seems currently the reason field can directly serialize and deserialized to ServerError is quite convenient).

Is the current structure a reasonable state to move forward?

crates/walrus-sdk/src/error.rs
halfprice: Done

crates/walrus-sdk/src/error.rs
jpcsmith: Serde can handle all that for us, the following code achieves this with little to no change to your code (see [the playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6e7454b4811db80eb615a084618c36ad) for it in action)
```rust
use serde::{Serialize, Deserialize};

#[derive(Serialize, Deserialize, Debug)]
struct RestApiErrorResponse {
    code: u16,
    message: String,
    #[serde(flatten)]
    reason: Option<ServiceError>
}

#[derive(Serialize, Deserialize, Debug)]
#[serde(tag = "reason", content = "metadata", rename_all = "camelCase")]
enum ServiceError {
    InvalidEpoch { server_epoch: usize }
}

fn main() {
    let error = ServiceError::InvalidEpoch { server_epoch: 19 };
    let response = RestApiErrorResponse {
        code: 400,
        message: "Something or other".to_string(),
        reason: Some(error),
    };
    
    let serialized = serde_json::to_string_pretty(&response).unwrap();
    println!("serialized = {:#}", serialized);
    
    let deserialized = serde_json::from_str::<RestApiErrorResponse>(&serialized).unwrap();
    println!("deserialized = {:#?}", deserialized);
}
```
outputs
```text
serialized = {
  "code": 400,
  "message": "Something or other",
  "reason": "invalidEpoch",
  "metadata": {
    "server_epoch": 19
  }
}
deserialized = RestApiErrorResponse {
    code: 400,
    message: "Something or other",
    reason: Some(
        InvalidEpoch {
            server_epoch: 19,
        },
    ),
}
```

crates/walrus-sdk/src/error.rs
```diff
@@ -104,3 +120,12 @@ pub(crate) enum BuildErrorKind {
...
     FailedToLoadCerts(#[from] std::io::Error),
 }
+
+/// Defines a more detailed server side reason that can be returned with an error.
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+pub enum ServiceError {
+    /// The requested epoch is invalid because it is too old.
+    InvalidEpochTooOld(Epoch),
+    /// The requested epoch is invalid because it is too new.
+    InvalidEpochTooNew(Epoch),
```
jpcsmith:
```suggestion
    EpochTooOld{ server_epoch: Epoch },
    /// The requested epoch is invalid because it is too new.
    EpochTooNew{ server_epoch: Epoch },
```

crates/walrus-sdk/src/error.rs
```diff
@@ -73,6 +83,12 @@ pub(crate) enum Kind {
     ErrorInNonErrorMessage { code: u16, message: String },
     #[error("invalid content type in response")]
     InvalidContentType,
+    #[error("{message} ({service_error:?})")]
+    StatusWithServiceError {
+        inner: reqwest::Error,
+        message: String,
+        service_error: ServiceError,
+    },
```
jpcsmith: Suggestion: Merge this with `StatusWithMessage` and make the `ServiceError` optional.

crates/walrus-sdk/src/node_response.rs
```diff
@@ -34,11 +34,22 @@ impl NodeResponse for Response {
...
+                message, reason, ..
+            }) => match reason {
+                Some(service_error) => Err(Kind::StatusWithServiceError {
+                    inner,
+                    message,
+                    service_error,
+                }
+                .into()),
+                None => Err(Kind::StatusWithMessage { inner, message }.into()),
+            },
```
jpcsmith: Suggestion: Revert after unifying `Kind::StatusWithServiceError` and `Kind::StatusWithMessage`.

crates/walrus-service/src/common/api.rs
```diff
@@ -310,3 +327,4 @@ pub(crate) fn rewrite_route(path: &str) -> String {
 pub(crate) use api_success_alias;
 #[allow(unused_imports)]
 pub(crate) use rest_api_error;
+use walrus_sdk::error::ServiceError;
```
jpcsmith: Suggestion: Move this and the 2 prior imports to the top of this file.

crates/walrus-service/src/node/errors.rs
```diff
@@ -140,3 +153,17 @@ pub enum SyncShardError {
...
+            Some(ServiceError::InvalidEpochTooOld(epoch)) => {
+                Self::InvalidEpoch(InvalidEpochError::TooOld(epoch))
+            }
+            Some(ServiceError::InvalidEpochTooNew(epoch)) => {
+                Self::InvalidEpoch(InvalidEpochError::TooNew(epoch))
+            }
+            _ => Self::Internal(error.into()),
+        }
+    }
+}
```
jpcsmith: Question: Why is this conversion necessary?

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -96,23 +96,38 @@ impl ShardSyncHandler {
...
+                if let Err(err) = sync_result {
+                    tracing::error!(
+                        "Failed to sync shard index: {} to before epoch: {}. Error: {:?}",
+                        shard_index,
+                        current_epoch,
+                        err
+                    );
+
+                    if let SyncShardError::InvalidEpoch(InvalidEpochError::TooNew(_)) = err {
+                        let retry_interval = Duration::from_secs(1);
```
jpcsmith: Suggestion: Use an exponential backoff. Given that you have no idea how far behind the remote storage node is, even if you back off to one or several minutes, the sync time will still dominates the overall time (retry time + sync time).

crates/walrus-sdk/src/error.rs
halfprice: Cool. Thanks for the pointer! Made an update to incorporate this. This is what the current json response look like in a reply:

```
{
  "error": {
    "code": 400,
    "message": "Invalid epoch. Client epoch: 1. Server epoch: 10",
    "reason": "invalidEpoch",
    "metadata": {
      "client_epoch": 1,
      "server_epoch": 10
    }
  }
}
```

crates/walrus-sdk/src/error.rs
halfprice: Given that we can put more information in the detailed error, I combined these two errors in a single `InvalidEpoch` error with information about client and server epoch.

crates/walrus-sdk/src/node_response.rs
halfprice: Done

crates/walrus-service/src/common/api.rs
halfprice: This was auto added. Probably because of the two `use` above. Not sure why adding the above 2 `pub(crate) use` here. I removed them and the `use walrus_sdk::error::ServiceError;` is now on top.

crates/walrus-service/src/node/errors.rs
halfprice: So we used to use the same `SyncShardError` in both server side and client side. So we had:

SyncShardError(server) -> Response -> NodeError -> SyncShardError(client).

This was necessary due to the last conversion.

But I think the right way to do it is to separate server and client side error, so the usage is clear.

Now, I have

SyncShardServiceError(server) -> Response -> NodeError -> SyncShardClientError

where SyncShardClientError contains some local error and also the NodeError to clearly indicate the error returned via the RPC call.

crates/walrus-service/src/node/shard_sync.rs
halfprice: Great catch! I originally put it as a place holder in a draft, but I never revisited it again... Probably not a good idea to retry every second...

crates/walrus-sdk/src/error.rs
halfprice: Done.

crates/walrus-service/src/node/errors.rs
jpcsmith: I would say it's sufficient to stop at NodeError, from which it can get the reason if relevant. Mirroring all the error types from the server to the client is overkill.

crates/walrus-sdk/src/error.rs
```diff
@@ -64,10 +74,11 @@ pub(crate) enum Kind {
     Bcs(#[from] bcs::Error),
     #[error(transparent)]
     Reqwest(#[from] reqwest::Error),
-    #[error("{inner}: {message}")]
+    #[error("{inner}: {message} ({service_error:?})")]
```
jpcsmith: Nit: Don't use the debug representation in the display. It's unnecessarily verbose.

crates/walrus-sdk/src/error.rs
```diff
@@ -104,3 +115,16 @@ pub(crate) enum BuildErrorKind {
...
 }
+
+/// Defines a more detailed server side error that can be returned to the client.
+#[derive(Debug, Serialize, Deserialize, Clone, Copy)]
+#[serde(tag = "reason", content = "metadata", rename_all = "camelCase")]
+pub enum ServiceError {
+    /// The requested epoch is invalid.
+    InvalidEpoch {
+        /// The epoch client is in.
+        client_epoch: Epoch,
```
jpcsmith:
```suggestion
        request_epoch: Epoch,
```

crates/walrus-service/src/node/errors.rs
halfprice: SyncShardClientError doesn't mirror all server errors. It is just local error (error generated while initiating the task) + NodeError. So it's all relevant to the client.

crates/walrus-sdk/src/error.rs
halfprice: Done

crates/walrus-sdk/src/error.rs
halfprice: Done

PR #711, opened 2024-08-21
chore: refactor bls committee to remove node info and `Committee`

### LINKED ISSUE #132: Remove extra node information from system object
The network adresses and names of the storage nodes do not need to be stored in the system object. To reduce the size of the system object, which gets loaded for most blob operations, we can only store this information in the "staking object" .

karlwuest:
closes #132

### DIFF
--- a/contracts/walrus/sources/system/bls_aggregate.move
+++ b/contracts/walrus/sources/system/bls_aggregate.move
@@ -3,11 +3,9 @@
 
 module walrus::bls_aggregate;
 
-use sui::{bls12381::{Self, bls12381_min_pk_verify, G1}, group_ops};
-use walrus::messages::{Self, CertifiedMessage};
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -3,51 +3,92 @@
 
 module walrus::bls_aggregate;
 
-use sui::{bls12381::{Self, bls12381_min_pk_verify}, group_ops};
+use sui::{bls12381::{Self, bls12381_min_pk_verify, G1}, group_ops};
+use walrus::messages::{Self, CertifiedMessage};
 use walrus::storage_node::StorageNodeInfo;
```
damirka: Nit: probably want to format this file.

PR #712, opened 2024-08-21
feat(staking): committee selection

damirka:
Adds:

- node registration with full information stored in `StakingPool`
- very simple implementation of an `ActiveSet` to keep track of top nodes by stake
- committee selection as a separate step
- better staking methods
- tests to update node / voting parameters
- tests for `staking_inner`

### DIFF
--- a/contracts/walrus/Move.lock
+++ b/contracts/walrus/Move.lock
@@ -21,6 +21,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.32.0"
+compiler-version = "1.30.3"
 edition = "2024.beta"
 flavor = "sui"
--- a/contracts/walrus/sources/staking.move
+++ b/contracts/walrus/sources/staking.move
@@ -10,7 +10,7 @@ use sui::{clock::Clock, coin::Coin, dynamic_field as df, sui::SUI};
 use walrus::{
     staked_wal::StakedWal,
     staking_inner::StakingInnerV1,
-    storage_node::StorageNodeCap,
+    storage_node::{Self, StorageNodeCap},
     system::System
 };
 
@@ -166,7 +166,7 @@ fun inner_mut(staking: &mut Staking): &mut StakingInnerV1 {
 // === Tests ===
 
 #[test_only]
-use walrus::{storage_node, staking_inner};
+use walrus::staking_inner;
 
 #[test_only]
 use sui::{clock, coin};
--- a/contracts/walrus/sources/staking/staked_wal.move
+++ b/contracts/walrus/sources/staking/staked_wal.move
@@ -13,10 +13,10 @@ use sui::{balance::Balance, sui::SUI};
 
 /// The state of the staked WAL.
 public enum StakedWalState has store, copy, drop {
-    /// Default state of the staked WAL - it is staked in the staking pool.
+    // Default state of the staked WAL - it is staked in the staking pool.
     Staked,
-    /// The staked WAL is in the process of withdrawing. The value inside the
-    /// variant is the epoch when the staked WAL can be withdrawn.
+    // The staked WAL is in the process of withdrawing. The value inside the
+    // variant is the epoch when the staked WAL can be withdrawn.
     Withdrawing(u64),
 }
 
--- a/contracts/walrus/sources/staking/staking_pool.move
+++ b/contracts/walrus/sources/staking/staking_pool.move
@@ -16,14 +16,14 @@ use walrus::{
 ///
 /// TODO: revisit the state machine.
 public enum PoolState has store, copy, drop {
-    /// The pool is new and awaits the stake to be added.
+    // The pool is new and awaits the stake to be added.
     New,
-    /// The pool is active and can accept stakes.
+    // The pool is active and can accept stakes.
     Active,
-    /// The pool awaits the stake to be withdrawn. The value inside the
-    /// variant is the epoch in which the pool will be withdrawn.
+    // The pool awaits the stake to be withdrawn. The value inside the
+    // variant is the epoch in which the pool will be withdrawn.
     Withdrawing(u64),
-    /// The pool is empty and can be destroyed.
+    // The pool is empty and can be destroyed.
     Withdrawn,
 }
 
--- a/contracts/walrus/sources/system.move
+++ b/contracts/walrus/sources/system.move
@@ -11,7 +11,7 @@ use walrus::{
     bls_aggregate::BlsCommittee,
     storage_node::StorageNodeCap,
     storage_resource::Storage,
-    system_state_inner::{Self, SystemStateInnerV1}
+    system_state_inner::SystemStateInnerV1,
 };
 
 /// Flag to indicate the version of the system.
@@ -169,6 +169,9 @@ fun inner(system: &System): &SystemStateInnerV1 {
 }
 
 // === Testing ===
+
+#[test_only]
+use walrus::system_state_inner;
 
 #[test_only]
 public(package) fun new_for_testing(ctx: &mut TxContext): System {
--- a/contracts/walrus/sources/system/system_state_inner.move
+++ b/contracts/walrus/sources/system/system_state_inner.move
@@ -9,7 +9,7 @@ use walrus::{
     blob::{Self, Blob},
     bls_aggregate::BlsCommittee,
     events::emit_invalid_blob_id,
-    storage_accounting::{Self, FutureAccountingRingBuffer},
+    storage_accounting::FutureAccountingRingBuffer,
     storage_node::StorageNodeCap,
     storage_resource::{Self, Storage}
 };
@@ -331,7 +331,7 @@ fun storage_units_from_size(size: u64): u64 {
 // == Testing ==
 
 #[test_only]
-use walrus::bls_aggregate;
+use walrus::{bls_aggregate, storage_accounting};
 
 #[test_only]
 public(package) fun new_for_testing(ctx: &mut TxContext): SystemStateInnerV1 {


contracts/walrus/sources/staking/active_set.move
```diff
@@ -0,0 +1,164 @@
...
+    /// determine if a storage node can be added to the active set.
+    min_staked_wal: u64,
+    /// The storage nodes in the active set, sorted by the amount of staked WAL.
+    nodes: VecMap<ID, u64>,
+    // /// The sorted list of storage nodes in the active set by the amount of
+    // /// staked WAL.
+    // sorted_nodes: vector<ID>,
+    /// The total amount of staked WAL in the active set.
+    total_staked_wal: u64,
+    /// Stores sorted indexes of the nodes in the active set. This allows us to
```
karlwuest: Suggestion: Add to the comment what this is sorted by.

contracts/walrus/sources/staking/active_set.move
```diff
@@ -0,0 +1,164 @@
...
+/// to a 1000 nodes. The nodes are sorted by the amount of staked WAL.
+/// Additionally, the active set tracks the total amount of staked WAL to make
+/// the calculation of the rewards and voting power distribution easier.
+public struct ActiveSet has store, copy, drop {
+    /// The maximum number of storage nodes in the active set.
+    /// Potentially remove this field.
+    max_size: u16,
+    /// The minimum amount of staked WAL in the active set. This is used to
+    /// determine if a storage node can be added to the active set.
+    min_staked_wal: u64,
```
karlwuest: Remark: I think we could remove this and just say that, to join the active set, a node has to have a stake of at least `total_staked_wal/max_size` and then `max_size` by default should be `n_shards`. It might make sense to just use `n_shards` directly, since we're currently not planning on having different limits for these.

contracts/walrus/sources/staking/active_set.move
```diff
@@ -0,0 +1,164 @@
...
+    nodes: VecMap<ID, u64>,
+    // /// The sorted list of storage nodes in the active set by the amount of
+    // /// staked WAL.
+    // sorted_nodes: vector<ID>,
+    /// The total amount of staked WAL in the active set.
+    total_staked_wal: u64,
+    /// Stores sorted indexes of the nodes in the active set. This allows us to
+    /// quickly find the index of a node in the sorted list of nodes. Uses `u16`
+    /// to save space, as the active set can only contain up to 1000 nodes.
+    idx_sorted: vector<u16>,
```
karlwuest, line 41: Question: Do we need this sorted for anything other than looking up the node with the smallest stake?

If not, we could use a min heap for this instead of a vector that we're keeping sorted. This would be a bit more efficient. This is just an optimization, though, so low priority, but we can create an issue to optimize the data structures (to keep it more general... we may also want to consider building a simple hash map instead of using a VecMap).

contracts/walrus/sources/staking/active_set.move
```diff
@@ -0,0 +1,164 @@
...
+        nodes: vec_map::empty(),
+        // sorted_nodes: vector[],
+        total_staked_wal: 0,
+        idx_sorted: vector[],
+    }
+}
+
+public(package) fun insert(set: &mut ActiveSet, node_id: ID, staked_amount: u64) {
+    assert!(!set.nodes.contains(&node_id));
+
```
karlwuest, line 85: Suggestion: Add an early return if the staked amount is less than the minimum stake for the active set.

contracts/walrus/sources/staking/active_set.move
```diff
@@ -0,0 +1,164 @@
...
+/// Additionally, the active set tracks the total amount of staked WAL to make
+/// the calculation of the rewards and voting power distribution easier.
+public struct ActiveSet has store, copy, drop {
+    /// The maximum number of storage nodes in the active set.
+    /// Potentially remove this field.
+    max_size: u16,
+    /// The minimum amount of staked WAL in the active set. This is used to
+    /// determine if a storage node can be added to the active set.
+    min_staked_wal: u64,
+    /// The storage nodes in the active set, sorted by the amount of staked WAL.
```
karlwuest:
```suggestion
    /// The amount of staked WAL for each storage node in the active set.
```

contracts/walrus/sources/staking/active_set.move
```diff
@@ -0,0 +1,164 @@
...
+        if (insert_idx.is_some()) {
+            insert_idx.do!(|idx| set.idx_sorted.insert(idx as u16, map_idx as u64));
+        } else {
+            set.idx_sorted.push_back(map_idx);
+        };
+
+        if (set.nodes.size() as u16 == set.max_size) {
+            let (_node_id, stake) = set.nodes.get_entry_by_idx(set.idx_sorted[0] as u64);
+            set.min_staked_wal = *stake;
+        }
```
karlwuest: Suggestion: I don't think we should set the `min_staked_wal` to the lowest stake in the active set, because then we will eventually always have 1000 nodes in the set, even if for most of them the stake is not sufficient to receive a shard. Instead, we should set it to a threshold value based on how much stake is needed to receive at least one shard. I think it makes sense to initially set it to some fixed value and then, once nodes are registered, set it to `total_stake/n_shards`. We can then later also remove nodes from the active set if they are below the threshold and were not selected for the next committee.

contracts/walrus/sources/staking/active_set.move
karlwuest: Just saw that there is a max heap implementation in the sui framework: https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/sui-framework/sources/priority_queue.move

karlwuest: Added #714 to keep track of this.

contracts/walrus/sources/staking/active_set.move
karlwuest: Added #715 to keep track of this

contracts/walrus/sources/staking/active_set.move
karlwuest: Added #715 to keep track of this

contracts/walrus/sources/staking/active_set.move
```diff
@@ -0,0 +1,180 @@
...
+
+/// Contains an active set of storage nodes. The active set is a smart collection
+/// that only stores up to a 1000 nodes. The nodes are sorted by the amount of
+/// staked WAL. Additionally, the active set tracks the total amount of staked
+/// WAL to make the calculation of the rewards and voting power distribution easier.
+///
+/// TODOs:
+/// - consider using a different data structure for the active set
+/// - consider removing `min_stake` field, use threshold from number of
+///   shards + total_staked
```
karlwuest:
```suggestion
/// - consider using a different data structure for the active set (#714)
/// - consider removing `min_stake` field, use threshold from number of
///   shards and total_staked (#715)
```

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -33,27 +51,46 @@ public struct StakingInnerV1 has store {
     /// The current epoch of the Walrus system. The epochs are not the same as
     /// the Sui epochs, not to be mistaken with `ctx.epoch()`.
     current_epoch: u64,
+    /// Stores the active set of storage nodes. Provides automatic sorting and
+    /// tracks the total amount of staked WAL.
+    active_set: ActiveSet,
+    /// The current committee in the system.
+    committee: BlsCommittee,
+    /// The previous committee in the system.
+    previous_committee: BlsCommittee,
```
karlwuest: Remark: I think here we need to store more than just the `BlsCommittee` because we need to know the shard assignment for the current and previous committee. On the other hand, we don't need the BLS keys here, so we can store the committee here differently, with just `node_id` and `shards` in there for each member. Alternatively, we can just store the vector of `node_id`s and then, in each staking pool store `shards_in_epoch` for the previous, the current, and the next epoch (if already set).

Related to this: we also need a `next_committee` (which needs to be set once the committee for the next epoch is decided).

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -12,12 +23,19 @@ use sui::{
...
+    bls_aggregate::{Self, BlsCommittee},
     staked_wal::{Self, StakedWal},
     staking_pool::{Self, StakingPool},
-    storage_node::StorageNodeCap,
+    storage_node::{Self, StorageNodeCap, StorageNodeInfo},
     walrus_context::{Self, WalrusContext}
 };
 
+/// The number of shards in the system.
+const SHARDS: u16 = 1000;
```
karlwuest: Question: Can we just set this as a parameter when deploying the system? Currently, a lot of the tests that we have use a different number of shards and re-integrating the contracts will be easier, if we can just set it accordingly.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -168,10 +221,44 @@ public(package) fun epoch(self: &StakingInnerV1): u64 {
...
+                    shard_idx = shard_idx + 1;
+                    shard_idx
+                });
+
+                pool.assign_shards(shards);
+                info_list.push_back(*pool.node_info());
+            },
+        );
+
+    let committee = bls_aggregate::new_bls_committee(new_epoch, &info_list);
```
karlwuest: Important: 
- This should happen in a different function that finalizes the committee for the next epoch and then set `next_committee` to the created committee.
- Add a TODO that the correct shard assignment algorithm needs to be implemented
- Even for the "dummy assignment", the assignment needs to distribute all shards which doesn't necessarily happen with this algorithm. We can just simplify this for now by just adding all remaining shards to the last node. Please add a warning there as well that this is just a dummy assignment.

contracts/walrus/sources/staking/staking_inner.move
damirka: We now set `shards: u16` parameter in staking inner. Thanks! I was too afraid to ask for it myself, but it drastically simplifies tests.

contracts/walrus/sources/staking/staking_inner.move
damirka: Yeah, a simpler data structure with better access is a must.

contracts/walrus/tests/staking/staking_inner_tests.move
```diff
@@ -0,0 +1,95 @@
...
+
+    // make sure the pools are no longer there
+    assert!(!staking.has_pool(pool_one));
+    assert!(!staking.has_pool(pool_two));
+
+    destroy(staking);
+}
+
+#[test]
+fun test_staking_active_set() {
```
damirka, line 37: @karlwuest let me know where exactly in this scenario the committee selection needs to happen (according to your comment: https://github.com/MystenLabs/walrus/pull/712#discussion_r1726720741)

contracts/walrus/sources/staking/staking_inner.move
damirka: Just pushed a new commit which splits these two.

contracts/walrus/tests/staking/staking_inner_tests.move
karlwuest: The order in the test is correct!

Issue #688, opened 2024-08-19
Before storing slivers and metadata to storage nodes, check if they already have them


giac-mysten:
- [x] Add an endpoint to check if the slivers are stored
- [x]  Use the sliver endpoint to perform the check in the client
- [x] Add an endpoint to check if the metadata are stored
- [x]  Use the metadata endpoint to perform the check in the client

PR #716, opened 2024-08-21
feat(client): add sliver status endpoint to client and sdk

giac-mysten:
* Adds the sliver status endpoint to the storage node `UserServer`.
* Adds the ability to query the sliver status endpoint from the sdk.
* Uses this endpoint in the `NodeCommunication` before storing a sliver, to check if the sliver is already present.

Contributes to #688

### DIFF
--- a/crates/walrus-service/src/client/communication.rs
+++ b/crates/walrus-service/src/client/communication.rs
@@ -34,11 +34,15 @@ use super::{
 use crate::common::utils::{self, ExponentialBackoff, FutureHelpers};
 
 /// Below this threshold, the `NodeCommunication` client will not check if the sliver is present on
-/// the node, but directly try to store it. This threshold is chosen as follows: Assume an MSS of
-/// 1480 Bytes, and an initial congestion window size of 4 packets. If we want the sliver to fit
-/// into the initial congestion window, it has to be smaller than 1480 * 4 = 5920.
-// TODO(giac): the above is quite arbitrary, and disregards TLS. Suggestions?
-const SLIVER_CHECK_THRESHOLD: usize = 5920;
+/// the node, but directly try to store it.
+///
+/// The threshold is chosend in a somewhat arbitrary way, but with the guiding principle that the
+/// direct sliver store should only take 1 RTT, therefore having similar latency to the sliver
+/// status check. To ensure this is the case, we take compute the threshold as follows: take the TCP
+/// payload size (1440 B); multiply it for an initial congestion window of 4 packets (although in
+/// modern systems this is usually 10, there may be other data being sent in this window); and
+/// conservatively subtract 200 B to account for HTTP headers and other overheads.
+const SLIVER_CHECK_THRESHOLD: usize = 5560;
 
 /// Represents the index of the node in the vector of members of the committee.
 pub type NodeIndex = usize;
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client/communication.rs
```diff
@@ -309,25 +317,64 @@ impl<'a> NodeWriteCommunication<'a> {
...
+    /// Stores a sliver on a node, first checking that the sliver is not already stored.
+    ///
+    ///  If the sliver is already stored, the function returns.
+    async fn check_and_store_sliver<A: EncodingAxis>(
+        &self,
+        blob_id: &BlobId,
+        sliver: &SliverData<A>,
+        pair_index: SliverPairIndex,
+    ) -> Result<(), SliverStoreError> {
+        match self.get_sliver_status::<A>(blob_id, pair_index).await? {
```
mlegner: Suggestion: Add a threshold for the sliver size below which it is stored directly without doing the check before. For very small slivers, the check is almost as expensive as directly storing the sliver.

giac-mysten: Uhm great point! I added a suggestion on the threshold in d520621460a6c1d02552ab64f8543f64f0b74775 , could you have another look?

crates/walrus-service/src/client/communication.rs
```diff
@@ -29,6 +33,13 @@ use super::{
 };
 use crate::common::utils::{self, ExponentialBackoff, FutureHelpers};
 
+/// Below this threshold, the `NodeCommunication` client will not check if the sliver is present on
+/// the node, but directly try to store it. This threshold is chosen as follows: Assume an MSS of
+/// 1480 Bytes, and an initial congestion window size of 4 packets. If we want the sliver to fit
```
jpcsmith: Remark: 10 segments has been recommended/default for initial congestion windows for some time (see the [Ubuntu 20 man pages](https://manpages.ubuntu.com/manpages/focal/en/man4/tcp.4freebsd.html) and [rfc6928](https://www.rfc-editor.org/rfc/rfc6928.html)).

crates/walrus-service/src/client/communication.rs
mlegner: I like the systematic approach, but I'm not sure if the initial congestion window is very relevant: In most cases, the client has already an active TCP connection at this point. Maybe a more "robust" limit would be that it fits into a single packet. 🤷 

There, we also have to take the HTTP overhead into account, so the threshold would be lower. According to Wireshark, just the HTTP header is 66 bytes for a PUT _primary_ sliver request and 68 for a _secondary_ (because "secondary" has two more letters...). 

So that would put us at 1500 (Ethernet MTU) - 40 (IPv6) - 20 (TCP) - ~70 (HTTP) = ~1370 as a threshold. This is sensitive to our path formats though. 🙄 (Side note: if we had a PUT sliver pair endpoint, we would save ~70 kB in transfer when storing a blob.)

In comparison, a GET status request is 156 bytes and the response is 319 (on the wire). We could also take that and say that storing the sliver should not be more than 10x (or some other arbitrary number) the size of this additional request/response.

(Side note: my Wireshark investigation showed that for minimally-sized slivers, a sliver pair is typically sent in the same packet.)

crates/walrus-service/src/client/communication.rs
```diff
@@ -309,25 +324,74 @@ impl<'a> NodeWriteCommunication<'a> {
...
+        pair_index: SliverPairIndex,
+    ) -> Result<(), SliverStoreError> {
+        if sliver.len() < SLIVER_CHECK_THRESHOLD {
+            tracing::debug!(
+                ?pair_index,
+                sliver_type=?A::sliver_type(),
+                sliver_len=sliver.len(),
+                "the sliver is sufficiently small not to require a check"
+            );
+            return self.store_sliver(blob_id, sliver, pair_index).await;
```
jpcsmith: Nit: Unify the two `store_sliver` calls by performing the checking within the condition.

crates/walrus-service/src/client/communication.rs
giac-mysten: >In most cases, the client has already an active TCP connection at this point. Maybe a more "robust" limit would be that it fits into a single packet. 🤷

that was my initial thought: since the connection is likely active, only in the very worst case we have to deal with the congestion window. _But_ we may have to consider that case.

> There, we also have to take the HTTP overhead into account...

Yes, you're right. I was a bit too optimistic there. There is also some TLS overhead possibly, but well. 

> We could also take that and say that storing the sliver should not be more than 10x (or some other arbitrary number) the size of this additional request/response.

An arbitrary threshold is what I wanted to avoid. My goal was that even in the worst case, we only use 1 RTT for storing the sliver, so that the latency is ~the same as the GET status request.

What about 1300 * 4 as a threshold? 1300 to be conservative?

mlegner: > My goal was that even in the worst case, we only use 1 RTT for storing the sliver, so that the latency is ~the same as the GET status request.

Works for me.

> What about 1300 * 4 as a threshold? 1300 to be conservative?

Works for me. However, I think the computation would be more like 1440 * 4 - ~200 (conservative value for HTTP and other overheads) = 5560.

jpcsmith: > (Side note: if we had a PUT sliver pair endpoint, we would save ~70 kB in transfer when storing a blob.)

@mlegner How so?

mlegner: We're sending ~1000 sliver pairs. If we had an endpoint for a sliver pair, we could save 1  HTTP header per sliver pair (assuming the header is roughly the same size), which is roughly 70 bytes, so 1000 * 70 bytes = 70 kB in total.

Then again, by just changing our URL paths, we would save some data; for example, replacing "secondary" by "s" would save 8kB per blob... 😅

PR #720, opened 2024-08-22
feat: add metadata status endpoint

### LINKED ISSUE #688: Before storing slivers and metadata to storage nodes, check if they already have them

giac-mysten:
The metadata status endpoint allows to check whether the metadata for a blob is stored on a node. 
The storage client uses this endpoint to avoid sending the metadata again when storing a blob.

Also adds retries for checking and storing metadata.

Note that there is no size threshold for the metadata check, as for any practical number of shards (>=100) the size of metadata is already large enough to warrant a check. 

Closes #688

### DIFF
--- a/crates/walrus-service/src/client/communication.rs
+++ b/crates/walrus-service/src/client/communication.rs
@@ -369,29 +369,26 @@ impl<'a> NodeWriteCommunication<'a> {
         sliver: &SliverData<A>,
         pair_index: SliverPairIndex,
     ) -> Result<(), SliverStoreError> {
-        if metdadata_status == &StoredOnNodeStatus::Nonexistent {
+        let print_debug = |message| {
             tracing::debug!(
                 ?pair_index,
                 sliver_type=?A::sliver_type(),
                 sliver_len=sliver.len(),
-                "the metadata has just been stored on the node; storing the sliver directly"
+                message
+            );
+        };
+        if metdadata_status == &StoredOnNodeStatus::Nonexistent {
+            print_debug(
+                "the metadata has just been stored on the node; storing the sliver directly",
             );
         } else if sliver.len() < SLIVER_CHECK_THRESHOLD {
-            tracing::debug!(
-                ?pair_index,
-                sliver_type=?A::sliver_type(),
-                sliver_len=sliver.len(),
+            print_debug(
                 "the sliver is sufficiently small not to require a status check; storing the sliver"
             );
         } else if self.get_sliver_status::<A>(blob_id, pair_index).await?
             == StoredOnNodeStatus::Nonexistent
         {
-            tracing::debug!(
-                ?pair_index,
-                sliver_type=?A::sliver_type(),
-                sliver_len=sliver.len(),
-                "the sliver is not stored on the node; storing the sliver"
-            );
+            print_debug("the sliver is not stored on the node; storing the sliver");
         } else {
             tracing::debug!(
                 ?pair_index,


crates/walrus-sdk/src/api.rs
```diff
@@ -130,16 +130,17 @@ impl Ord for BlobStatus {
     }
 }
 
-/// Contains the status of a sliver.
+/// Contains the storage status of status of a sliver or metadata.
```
mlegner: Nit:
```suggestion
/// Contains the storage status of a sliver or metadata.
```

crates/walrus-sdk/src/api.rs
```diff
@@ -130,16 +130,17 @@ impl Ord for BlobStatus {
...
 #[derive(
     Debug, Deserialize, Serialize, PartialEq, Eq, Clone, Copy, Default, Hash, utoipa::ToSchema,
 )]
 #[serde(rename_all = "camelCase", rename_all_fields = "camelCase")]
-pub enum SliverStatus {
-    /// The sliver does not exist on the storage node.
+// TODO(giac): Is this fine as a name? Do we want to separate sliver and metadata statuses?
```
mlegner: This is fine for me. I don't think we need different structs.

crates/walrus-service/src/client/communication.rs
```diff
@@ -272,11 +272,19 @@ impl<'a> NodeWriteCommunication<'a> {
...
-            // TODO(giac): consider adding timeouts and replace the Reqwest timeout.
-        })
-        .await
+        match self
+            .retry_with_limits_and_backoff(|| self.client.get_metadata_status(metadata.blob_id()))
+            .await?
+        {
+            StoredOnNodeStatus::Stored => {
+                tracing::debug!("the metadata is already stored on the node");
+                Ok(())
```
mlegner: Remark: I thought about the following optimization: If the metadata is *not* stored yet, this implies that none of the slivers are stored. So we don't need to check the sliver status before sending them.

OTOH, if we want to use the sliver status check to enable parallel uploads, this is no longer the case. 🤷

crates/walrus-service/src/node.rs
```diff
@@ -751,6 +764,22 @@ impl ServiceState for StorageNodeInner {
...
+    ) -> Result<StoredOnNodeStatus, RetrieveMetadataError> {
+        if let Some(blob_info) = self
+            .storage
+            .get_blob_info(blob_id)
+            .context("could not retrieve blob info")?
+        {
+            if blob_info.is_metadata_stored() {
+                return Ok(StoredOnNodeStatus::Stored);
+            }
+        }
```
mlegner: Hint: There is a `Storage::has_metadata` function that does what we need here.

Also, important: IIUC, your code results in an internal server error if the node doesn't know about the requested blob ID yet, which it should definitely not do.

giac-mysten: whops, you're right, thanks for the hint. Should be fixed: d27e497456698f45f2934624502c2877b63d0d58

crates/walrus-service/src/client/communication.rs
giac-mysten: mmm interesting optimization. I imagine fully parallel uploads (where many clients try to upload a blob at the same time without knowing of each other) may be a rare occurrence?

I added an implementation of this optimization: 8e6838f8ed9bbb562f7d9bc3eebe78fa8d7032fe . Lmk what you think!

crates/walrus-service/src/client/communication.rs
```diff
@@ -338,14 +358,25 @@ impl<'a> NodeWriteCommunication<'a> {
...
         pair_index: SliverPairIndex,
     ) -> Result<(), SliverStoreError> {
-        if sliver.len() < SLIVER_CHECK_THRESHOLD {
+        if metdadata_status == &StoredOnNodeStatus::Nonexistent {
+            tracing::debug!(
+                ?pair_index,
+                sliver_type=?A::sliver_type(),
+                sliver_len=sliver.len(),
+                "the metadata has just been stored on the node; storing the sliver directly"
+            );
```
mlegner: Nit: These debug statements have a lot of common elements. Can we either instrument the function and add those there? Or, if we don't want to do this, define a local closure?
```rust
        let print_debug = |message| {
            tracing::debug!(
                ?pair_index,
                sliver_type=?A::sliver_type(),
                sliver_len=sliver.len(),
                message,
            );
        };
```

crates/walrus-service/src/client/communication.rs
mlegner: I like it! 💯 

Also, praise: I like the detailed documentation!

crates/walrus-service/src/client/communication.rs
giac-mysten: 🤦 I was thinking about this, and somehow always forget closures...

Issue #705, opened 2024-08-20
Make basic shard sync robust


halfprice:
Follow up todos after #673:
- [ ] verify sliver validity
- [ ] rate limiting for shard syncs
- [ ] retries for retriable errors
- [ ] more tests

PR #721, opened 2024-08-23
feat(storage node): checkpointing and recovering shard sync task

halfprice:
This PR implements handling crash recovery of happy path shard sync task.

During the process of a shard sync tasks, the task keeps track of syncing progress by checkpointing the current
sync location in shard_sync_progress table, from which the shard can resume the sync task upon restarting.

Introduce the usage of failpoints to simulate shard restarting in tests. Running the test requires adding
`--features failure_injection` feature.

Contribute to #705

### DIFF
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -454,7 +454,8 @@ impl ShardStorage {
         mut last_synced_blob_id: Option<BlobId>,
     ) -> Result<(), SyncShardError> {
         // Helper to track the number of scanned blobs to test recovery. Not used in production.
-        let mut _scan_count = 0;
+        #[cfg(feature = "failure_injection")]
+        let mut scan_count = 0;
         while let Some(next_starting_blob_id) =
             next_certified_blob_id_before_epoch(&node.storage, epoch, last_synced_blob_id)
                 .context("Scanning certified blobs encountered error")?
@@ -501,7 +502,7 @@ impl ShardStorage {
                 #[cfg(feature = "failure_injection")]
                 {
                     (|| -> Result<(), anyhow::Error> {
-                        _scan_count += 1;
+                        scan_count += 1;
                         fail::fail_point!("fail_point_fetch_sliver", |arg| {
                             if let Some(arg) = arg {
                                 let parts: Vec<&str> = arg.split(',').collect();
@@ -512,12 +513,12 @@ impl ShardStorage {
                                     arg = ?arg,
                                     if_trigger_in_primary = ?trigger_in_primary,
                                     trigger_index = ?trigger_at,
-                                    blob_count = ?_scan_count
+                                    blob_count = ?scan_count
                                 );
                                 if ((trigger_in_primary && sliver_type == SliverType::Primary)
                                     || (!trigger_in_primary
                                         && sliver_type == SliverType::Secondary))
-                                    && trigger_at == _scan_count
+                                    && trigger_at == scan_count
                                 {
                                     return Err(anyhow!("fetch_sliver simulated sync failure"));
                                 }


crates/walrus-service/src/node.rs
```diff
@@ -2095,4 +2107,134 @@ mod tests {
 
         Ok(())
     }
+
+    #[cfg(feature = "failure_injection")]
+    mod failure_injection_tests {
```
mlegner, line 2116: Suggestion: Move all failpoint tests to the `walrus-e2e-tests` crate and add the `failure_injection` feature to the `walrus-service` dev dependency. Then, one doesn't have to enable features manually to run the tests.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -406,9 +486,47 @@ impl ShardStorage {
                         batch.insert_batch(&self.secondary_slivers, [(blob.0, secondary)])?;
                     }
                 }
-                starting_blob_id = Some(blob.0);
+
+                // Inject a failure point to simulate a sync failure.
+                if fail::has_failpoints() {
```
mlegner: Question: Will the compiler optimize this away statically? To be sure, we could change this as follows (also for the variable definition `let mut _scan_count = 0;` above):
```suggestion
                // Inject a failure point to simulate a sync failure.
                #[cfg(feature = "failure_injection")]
                {
```

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -124,6 +124,11 @@ impl ShardSyncHandler {
         });
         entry.insert(shard_sync_task);
     }
+
+    #[cfg(test)]
```
mlegner, line 128: Nit:
```suggestion
    #[cfg(all(test, feature = "failure_injection"))]
```

crates/walrus-service/src/node.rs
```diff
@@ -2095,4 +2107,134 @@ mod tests {
...
+                assert_eq!(
+                    shard_storage_src
+                        .get_sliver(&blob_id, SliverType::Secondary)
+                        .unwrap()
+                        .unwrap(),
+                    shard_storage_dst
+                        .get_sliver(&blob_id, SliverType::Secondary)
+                        .unwrap()
+                        .unwrap()
+                );
```
mlegner: Nit: This could be shortened a bit:
```suggestion
                for sliver_type in [SliverType::Primary, SliverType::Secondary] {
                    assert_eq!(
                        shard_storage_src.get_sliver(&blob_id, sliver_type),
                        shard_storage_dst.get_sliver(&blob_id, sliver_type),
                    );
                }
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -71,12 +73,35 @@ impl ShardStatus {
     }
 }
 
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+struct ShardSyncProgressV1 {
+    last_synced_blob_id: BlobId,
+    sliver_type: SliverType,
```
mlegner, line 82: Nit: Can you describe the meaning of these fields, in particular the `sliver_type`?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -363,8 +441,10 @@ impl ShardStorage {
         epoch: Epoch,
         node: Arc<StorageNodeInner>,
         sliver_type: SliverType,
+        mut starting_blob_id: Option<BlobId>,
```
mlegner: Nit: The name implies that this particular blob is the first to sync. However, this has already been synced.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -406,9 +486,47 @@ impl ShardStorage {
...
+                                }
+                            }
+                            Ok(())
+                        });
+                        Ok(())
+                    })()?;
+                }
+            }
+
+            starting_blob_id = fetched_slivers.last().map(|(id, _)| *id);
```
mlegner: Important: What happens when `fetched_slivers.is_empty()`? AFAIU, we get a `None` here and start from the beginning in the next loop iteration. This should definitely not happen, so we either need to use the previous version where we leave the `starting_blob_id` unchanged and send the same request again, or have some other special handling.

halfprice: Great catch!

I explicitly tested if there is nothing to send between src and dst, but that won't trigger the bug if src and dst have the same certified blob list (in which line 451 will just return None and quit the loop).

I added a test scenario where the dst node expects some blobs, but the src node sends back nothing. This will trigger the bug.

To fix it, we just need to break out the loop if `starting_blob_id` == None. We can't reuse the same `starting_blob_id` and restart the loop since line 451 again will expect some blobs and enter the loop.

crates/walrus-service/src/node.rs

crates/walrus-service/src/node/storage/shard.rs
mlegner: Thanks for the quick fix! Probably we want some more elaborate handling for cases where we don't get back what we expect (maybe add a TODO?), but this should work for now. 👍

crates/walrus-service/src/node.rs
```diff
@@ -2212,29 +2198,48 @@ mod tests {
...
-                        .unwrap()
-                        .unwrap()
-                );
+                for sliver_type in [SliverType::Primary, SliverType::Secondary] {
+                    assert_eq!(
+                        shard_storage_src.get_sliver(&blob_id, sliver_type),
+                        shard_storage_dst.get_sliver(&blob_id, sliver_type),
+                    );
+                }
             });
```
mlegner: Nit: I just realized that we have the exact same code twice. This could be factored out into a separate function.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Done. 

It should compile away, that's why I have to use `_scan_count` instead of `scan_count`. But your suggestion is more explicit.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Make sense. Updated it `last_synced_blob_id`

crates/walrus-service/src/node/storage/shard.rs
halfprice: This is what TODO(#705) about (also noted above for handling missing blobs).

crates/walrus-service/src/node.rs
halfprice: Done. Yeah, probably should have just done that...

crates/walrus-service/src/node/shard_sync.rs
halfprice: I'm not sure if we want to add `feature = "failure_injection"` there. The purpose of using attributes is to restrict the API to not be available in prod, which I think `cfg(test)` already does the job.

crates/walrus-service/src/node.rs
halfprice: I think this needs more work than I thought.

The original goal of using fail point to inject failure is from the convenience that simtest uses failpoint to simulate failures. By enabling it here, I hope it'll be easier to use it in simtest. But I just realized that Sui uses it's own failpoint implementation instead of the `fail` crate.

I think what we should do is to also use the Sui failpoints implementation in these tests as well, which only available under `cfg(msim). So we should move these tests to walrus-simtest, and use simtest setup for these tests.

This going to require quite some work, so I created a TODO and an issue #726 to track it. I hope to get back to this after finishing non-happy path shard recovery.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -443,12 +451,12 @@ impl ShardStorage {
         epoch: Epoch,
         node: Arc<StorageNodeInner>,
         sliver_type: SliverType,
-        mut starting_blob_id: Option<BlobId>,
+        mut last_synced_blob_id: Option<BlobId>,
     ) -> Result<(), SyncShardError> {
         // Helper to track the number of scanned blobs to test recovery. Not used in production.
         let mut _scan_count = 0;
```
mlegner: Nit:
```suggestion
        #[cfg(feature = "failure_injection")]
        let mut _scan_count = 0;
```
Then we can actually rename this to `scan_count`.

crates/walrus-service/src/node/storage/shard.rs
mlegner: For future reference: The `has_failpoints` evaluates to `cfg!(feature = "failpoints")`, see [here](https://docs.rs/fail/latest/src/fail/lib.rs.html#619-621).

According to [these docs](https://doc.rust-lang.org/rust-by-example/attribute/cfg.html), the `cfg!` macro simply evaluates to a literal `true` or `false`. So it doesn't disable the code, but the compiler still likely optimizes all the code away.

In contrast, the `#[cfg()]` construct actually disables the code, so we don't have to rely on compiler optimizations.

crates/walrus-service/src/node/storage/shard.rs

PR #729, opened 2024-08-26
feat(staking): committee now stores shard ids

damirka:
Stores `shard_ids` instead of their count in previous / current / next committees.

### DIFF
--- a/contracts/walrus/sources/staking/committee.move
+++ b/contracts/walrus/sources/staking/committee.move
@@ -44,34 +44,38 @@ public(package) fun initialize(assigned_number: VecMap<ID, u16>): Committee {
 ///
 /// This assumes that the number of shards in the new committee is equal to the
 /// number of shards in the current committee. Check for this is not performed.
-public(package) fun transition(cmt: &Committee, assigned_number: VecMap<ID, u16>): Committee {
+public(package) fun transition(cmt: &Committee, mut new_assignments: VecMap<ID, u16>): Committee {
     let mut new_cmt = vec_map::empty();
     let mut to_move = vector[];
-    let mut revisit = assigned_number.keys();
     let size = cmt.0.size();
 
     size.do!(
         |idx| {
             let (node_id, prev_shards) = cmt.0.get_entry_by_idx(idx);
             let node_id = *node_id;
+            let assigned_len = new_assignments
+                .get_idx_opt(&node_id)
+                .map!(
+                    |idx| {
+                        let (_, value) = new_assignments.remove_entry_by_idx(idx);
+                        value as u64
+                    },
+                );
 
             // if the node is not in the new committee, remove all shards, make
             // them available for reassignment
-            if (!assigned_number.contains(&node_id)) {
+            if (assigned_len.is_none() || assigned_len.borrow() == &0) {
                 let shards = cmt.0.get(&node_id);
                 to_move.append(*shards);
                 return
             };
 
             let curr_len = prev_shards.length();
-            let assigned_len = *assigned_number.get(&node_id) as u64;
+            let assigned_len = assigned_len.destroy_some();
 
             // node stays the same, we copy the shards over, best scenario
             if (curr_len == assigned_len) {
                 new_cmt.insert(node_id, *prev_shards);
-                revisit
-                    .find_index!(|e| e == &node_id)
-                    .do!(|found_idx| { revisit.remove(found_idx); })
             };
 
             // if the node is in the new committee, check if the number of shards
@@ -81,9 +85,6 @@ public(package) fun transition(cmt: &Committee, assigned_number: VecMap<ID, u16>
                 let mut node_shards = *prev_shards;
                 (curr_len - assigned_len).do!(|_| to_move.push_back(node_shards.pop_back()));
                 new_cmt.insert(node_id, node_shards);
-                revisit
-                    .find_index!(|e| e == &node_id)
-                    .do!(|found_idx| { revisit.remove(found_idx); })
             };
 
             // if the node is in the new committee, and we already freed enough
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/staking/committee.move
```diff
@@ -0,0 +1,237 @@
...
+            // if the node is not in the new committee, remove all shards, make
+            // them available for reassignment
+            if (!assigned_number.contains(&node_id)) {
+                let shards = cmt.0.get(&node_id);
+                to_move.append(*shards);
+                return
+            };
+
+            let curr_len = prev_shards.length();
+            let assigned_len = *assigned_number.get(&node_id) as u64;
```
karlwuest: Suggestion: Currently, we may be iterating through the `VecMap` twice, once for checking if the value is contained and then once for getting the value. Instead, we can do something like the following to only iterate through it once:

```suggestion
            let assigned_len = assigned_number.try_get(&node_id);
            // if the node is not in the new committee, remove all shards, make
            // them available for reassignment
            if (assigned_len.is_none()) {
                let shards = cmt.0.get(&node_id);
                to_move.append(*shards);
                return
            };

            let curr_len = prev_shards.length();
            let assigned_len = assigned_len.extract() as u64;
```

contracts/walrus/sources/staking/committee.move
```diff
@@ -0,0 +1,237 @@
...
+
+            // node stays the same, we copy the shards over, best scenario
+            if (curr_len == assigned_len) {
+                new_cmt.insert(node_id, *prev_shards);
+                revisit
+                    .find_index!(|e| e == &node_id)
+                    .do!(|found_idx| { revisit.remove(found_idx); })
```
karlwuest: Question: Is there a reason to start with a filled `revisit` and remove nodes, instead of starting with an empty one and then adding nodes that need more shards? The current way iterates over all indices in revisit quite often. 

Remark: Using `swap_remove` here would save quite a bit of computation already.

contracts/walrus/sources/staking/committee.move
```diff
@@ -0,0 +1,237 @@
...
+        |idx| {
+            let (node_id, prev_shards) = cmt.0.get_entry_by_idx(idx);
+            let node_id = *node_id;
+
+            // if the node is not in the new committee, remove all shards, make
+            // them available for reassignment
+            if (!assigned_number.contains(&node_id)) {
+                let shards = cmt.0.get(&node_id);
+                to_move.append(*shards);
+                return
```
karlwuest, line 70: Question: Since "closures" are a bit different in move than in rust: Does this return only return from the "closure"? Or does this return from the `transition` function?

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -181,8 +185,15 @@ public(package) fun request_withdraw_stake(
...
     assert!(staked_wal.activation_epoch() <= wctx.epoch());
 
-    // depend on the committee selection + whether a node is active / has been active
-    staked_wal.set_withdrawing(wctx.epoch() + 1);
+    let node_in_committee = wctx.committee().contains(pool.id.as_inner());
+
+    // If the node is in the committee, the stake will be withdrawn in E+2,
+    // otherwise in E+1.
+    if (node_in_committee && wctx.committee_selected()) {
+        staked_wal.set_withdrawing(wctx.epoch() + 2);
```
damirka, line 193: We could do a mini hack here to avoid conditional, and for initial state assign all shards to `0x0` or some other address.

contracts/walrus/sources/staking/committee.move
```diff
@@ -0,0 +1,237 @@
...
+                }
+
+                // else, node remains in the revisit list, we need to free up more
+                // shards to assign to this node
+            };
+        },
+    );
+
+    // In the `revisit` set we have either nodes that need more shards or nodes
+    // that were not in the new committee. Last step is to assign the remaining
```
damirka: In the old committee.
```suggestion
    // that were not in the old committee. Last step is to assign the remaining
```

contracts/walrus/sources/staking/committee.move
```diff
@@ -0,0 +1,237 @@
...
+    // shards to the nodes that need them
+    revisit.destroy!(
+        |node_id| {
+            let mut node_shards = cmt.0.try_get(&node_id).destroy_or!(vector[]);
+            let assigned_len = *assigned_number.get(&node_id) as u64;
+            let diff = assigned_len - node_shards.length();
+            diff.do!(|_| node_shards.push_back(to_move.pop_back()));
+            new_cmt.insert(node_id, node_shards);
```
karlwuest: Important: We need a check here if `node_shards` is empty. In that case, we don't want to add the node to the committee.

Remark: If we add nodes to (an initially empty) `revisit` instead of removing nodes from it, this is solved automatically.

contracts/walrus/sources/staking/committee.move
damirka: Given that we first go through current set, new nodes are left out. Illustrated in a silly sample below:
```
curr: [ 1 ] [ 2 ] [ 3 ]
new:        [ 2 ] [ 3 ] [ 4 ]
```
I haven't figured a way to track new entries to the committee and making sure they get added. Currently, we iterate through `curr`.

contracts/walrus/sources/staking/committee.move
damirka: They return from the closure.

contracts/walrus/sources/staking/committee.move
karlwuest: Ok, let's just do a `swap_remove` then for now, s.t. we don't have to shift all entries around.

PR #732, opened 2024-08-27
feat: implement epoch change in staking

### LINKED ISSUE #735: Implement initialization of staking and system object

karlwuest:
closes #735

### DIFF
--- a/contracts/walrus/sources/init.move
+++ b/contracts/walrus/sources/init.move
@@ -1,0 +1,42 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+module walrus::init;
+
+use sui::clock::Clock;
+use walrus::{staking, system};
+
+/// Must only be created by `init`.
+public struct InitCap has key, store {
+    id: UID,
+}
+
+/// Init function, creates an init cap and transfers it to the sender.
+/// This allows the sender to call the function to actually initialize the system
+/// with the corresponding parameters. Once that function is called, the cap is destroyed.
+fun init(ctx: &mut TxContext) {
+    let id = object::new(ctx);
+    let init_cap = InitCap { id };
+    transfer::transfer(init_cap, ctx.sender());
+}
+
+/// Function to initialize walrus and share the system and staking objects.
+/// This can only be called once, after which the `InitCap` is destroyed.
+/// TODO: decide what to add as system parameters instead of constants.
+public fun initialize_walrus(cap: InitCap, n_shards: u16, clock: &Clock, ctx: &mut TxContext) {
+    system::create_empty(ctx);
+    staking::create(n_shards, clock, ctx);
+    cap.destroy();
+}
+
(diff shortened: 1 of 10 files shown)


Cargo.lock
damirka, line 1: Is it intentional?

karlwuest: No, I'll exclude it. For some reason, it was replaced on my machine by the pre-commit script.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -284,24 +361,64 @@ public(package) fun select_committee(self: &mut StakingInnerV1, ctx: &mut TxCont
...
+
+    // Add any leftover rewards to the rewards to distribute.
+    let leftover_value = self.leftover_rewards.value();
+    rewards.join(self.leftover_rewards.split(leftover_value));
+    let rewards_per_shard = rewards.value() / (self.n_shards as u64);
+    let (node_ids, shard_assignments) = (*self.previous_committee.inner()).into_keys_values();
+    node_ids.zip_do!(
+        shard_assignments,
+        |node_id, shards| self
+            .pools[node_id]
```
damirka, line 417: I would add a TODO here: accessing pools for the second time may be expensive (not sure about the limit).

contracts/walrus/sources/system/epoch_parameters.move
```diff
@@ -0,0 +1,55 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+module walrus::epoch_parameters;
+
+/// The epoch parameters for the system.
+public struct EpochParams has store, copy, drop {
```
damirka, line 7: We already have these fields in `staking_pool::PoolParams`. Perhaps, worth using this struct there.

karlwuest: They are slightly different, in particular the capacity and their meaning is different, so I would prefer using different types. The `PoolParams` also include a commission.

contracts/walrus/sources/staking/staking_inner.move
karlwuest: Will add it. Related question: Do we actually need the update on the active set that is done a bit above or could we just do these updates lazily (e.g. next time someone stakes or wants to withdraw stake)?

damirka: By triggering `advance_epoch` on each pool, we make sure their voting / node parameters are in sync, which is a requirement for voting, I assume.

karlwuest: Ok, let's discuss this tomorrow. I added the TODO.

contracts/walrus/sources/init.move
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Mysten Labs, Inc.
```
damirka, line 1: Love this!

PR #734, opened 2024-08-28
feat(storage node): shard sync recovery path using BlobSyncHandler

halfprice:
- During shard sync, store missing blobs in a `pending_recover_blobs` table in shard.
- After shard sync, using blob recover to recover blobs in `pending_recover_blobs`.
- Non happy path currently do not preempt happy path currently.

Contributes to #259 #560 #705

### DIFF
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -253,7 +253,10 @@ impl<'a> BlobInfoIter<'a> {
 
 impl Debug for BlobInfoIter<'_> {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        f.debug_struct("BlobInfoIter").finish()
+        f.debug_struct("BlobInfoIter")
+            .field("before_epoch", &self.before_epoch)
+            .field("only_certified", &self.only_certified)
+            .finish()
     }
 }
 
@@ -270,7 +273,7 @@ impl Iterator for BlobInfoIter<'_> {
             if blob_info.is_certified()
                 && blob_info
                     .status_changing_epoch(blob_info::BlobCertificationStatus::Certified)
-                    .unwrap()
+                    .expect("We just checked that the blob is certified")
                     < self.before_epoch
             {
                 return item;
@@ -1362,13 +1365,13 @@ pub(crate) mod tests {
         let new_epoch = 3;
 
         let blob_ids = [
-            BlobId([1; 32]), // Not certified.
-            BlobId([2; 32]), // Not exist.
-            BlobId([3; 32]), // Certified within epoch 2
-            BlobId([4; 32]), // Certified after epoch 2
-            BlobId([5; 32]), // Invalid
-            BlobId([6; 32]), // Certified within epoch 2
-            BlobId([7; 32]), // Not exist.
+            BlobId([0; 32]), // Not certified.
+            BlobId([1; 32]), // Not exist.
+            BlobId([2; 32]), // Certified within epoch 2
+            BlobId([3; 32]), // Certified after epoch 2
+            BlobId([4; 32]), // Invalid
+            BlobId([5; 32]), // Certified within epoch 2
+            BlobId([6; 32]), // Not exist.
         ];
 
         let blob_info_map = HashMap::from([
@@ -1382,6 +1385,11 @@ pub(crate) mod tests {
         let mut batch = blob_info.batch();
         batch.insert_batch(&blob_info, blob_info_map.iter())?;
         batch.write()?;
+
+        assert_eq!(
+            all_certified_blob_ids(&storage, None, new_epoch)?,
+            vec![blob_ids[2], blob_ids[5]]
+        );
 
         for blob_id in blob_ids.iter().take(2) {
             assert_eq!(
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -20,7 +20,7 @@ use rocksdb::{Options, DB};
 use serde::{Deserialize, Serialize};
 use typed_store::{
     rocks::{
-        be_fix_int_ser as ToRocksDbKey,
+        be_fix_int_ser as to_rocks_db_key,
         errors::typed_store_err_from_rocks_err,
         DBBatch,
         DBMap,
@@ -110,6 +110,17 @@ impl ShardSyncProgress {
     }
 }
 
+// Represents the last synced status of the shard after restart.
+enum ShardLastSyncStatus {
+    // The last synced blob ID for the primary slivers.
+    // If the last_synced_blob_id is None, it means the shard has not synced any primary slivers.
+    Primary { last_synced_blob_id: Option<BlobId> },
+    // The last synced blob ID for the secondary slivers.
+    Secondary { last_synced_blob_id: Option<BlobId> },
+    // The shard is in recovery mode.
+    Recovery,
+}
+
 #[derive(Debug, Clone)]
 pub struct ShardStorage {
     id: ShardIndex,
@@ -130,6 +141,7 @@ impl ShardStorage {
     ) -> Result<Self, TypedStoreError> {
         let rw_options = ReadWriteOptions::default();
 
+        // TODO: cleanup all the column faimily initiation as they are highly repetitive.
         let shard_cf_options = Self::slivers_column_family_options(id, db_config);
         for (_, (cf_name, options)) in shard_cf_options.iter() {
             if database.cf_handle(cf_name).is_none() {
@@ -431,82 +443,40 @@ impl ShardStorage {
             shard_status == ShardStatus::ActiveSync || shard_status == ShardStatus::ActiveRecover
         );
 
-        // Checks if the shard was previously syncing and resumes syncing from the last synced blob.
-        let (mut last_synced_blob_id, skip_primary, skip_secondary) =
-            if shard_status == ShardStatus::ActiveRecover {
-                // We are in recovery mode. Skip happy path syncing and directly jump to recover
-                // missing blobs.
-                (None, true, true)
-            } else {
-                match self.shard_sync_progress.get(&())? {
-                    Some(ShardSyncProgress::V1(ShardSyncProgressV1 {
-                        last_synced_blob_id,
-                        sliver_type,
-                    })) => {
-                        tracing::info!(
-                            "Resuming shard sync from blob id: {}, sliver type: {}",
-                            last_synced_blob_id,
-                            sliver_type
-                        );
-                        (
-                            Some(last_synced_blob_id),
-                            sliver_type == SliverType::Secondary,
-                            false,
-                        )
-                    }
-                    None => (None, false, false),
-                }
-            };
-
-        if !skip_primary {
-            self.sync_shard_before_epoch_internal(
-                epoch,
-                node.clone(),
-                SliverType::Primary,
+        match self.get_last_sync_status(&shard_status)? {
+            ShardLastSyncStatus::Primary {
                 last_synced_blob_id,
-            )
-            .await?;
-            last_synced_blob_id = None;
-        }
-
-        if !skip_secondary {
-            self.sync_shard_before_epoch_internal(
-                epoch,
-                node.clone(),
-                SliverType::Secondary,
-                last_synced_blob_id,
-            )
-            .await?;
-        }
-
-        if !self.pending_recover_slivers.is_empty() {
-            tracing::info!(
-                "Shard sync is done. Still has missing blobs. Shard enters recovery mode.",
-            );
-            self.shard_status.insert(&(), &ShardStatus::ActiveRecover)?;
-
-            fail::fail_point!("fail_point_after_start_recovery", |_| {
-                Err(SyncShardError::Internal(anyhow!(
-                    "sync shard simulated sync failure after start recovery"
-                )))
-            });
-
-            loop {
-                self.recover_missing_blobs(blob_sync_handler.clone())
-                    .await?;
-
-                if self.pending_recover_slivers.is_empty() {
-                    // TODO: check recovered blob consistency.
-                    break;
-                }
-                tracing::warn!(
-                    "Recovering missing blobs for shard: {} still misses blobs.
-                    Retrying in 60 seconds.",
-                    self.id
-                );
-                tokio::time::sleep(Duration::from_secs(60)).await;
+            } => {
+                self.sync_shard_before_epoch_internal(
+                    epoch,
+                    node.clone(),
+                    SliverType::Primary,
+                    last_synced_blob_id,
+                )
+                .await?;
+                self.sync_shard_before_epoch_internal(
+                    epoch,
+                    node.clone(),
+                    SliverType::Secondary,
+                    None,
+                )
+                .await?;
             }
+            ShardLastSyncStatus::Secondary {
+                last_synced_blob_id,
+            } => {
+                self.sync_shard_before_epoch_internal(
+                    epoch,
+                    node.clone(),
+                    SliverType::Secondary,
+                    last_synced_blob_id,
+                )
+                .await?;
+            }
+            ShardLastSyncStatus::Recovery => {}
         }
+
+        self.recovery_any_missing_slivers(blob_sync_handler).await?;
 
         let mut batch = self.shard_status.batch();
         batch.insert_batch(&self.shard_status, [((), ShardStatus::Active)])?;
@@ -514,6 +484,43 @@ impl ShardStorage {
         batch.write()?;
 
         Ok(())
+    }
+
+    /// Returns the last sync status for the shard.
+    fn get_last_sync_status(
+        &self,
+        shard_status: &ShardStatus,
+    ) -> Result<ShardLastSyncStatus, TypedStoreError> {
+        let last_sync_status = if shard_status == &ShardStatus::ActiveRecover {
+            // We are in recovery mode. Skip happy path syncing and directly jump to recover
+            // missing blobs.
+            ShardLastSyncStatus::Recovery
+        } else {
+            match self.shard_sync_progress.get(&())? {
+                Some(ShardSyncProgress::V1(ShardSyncProgressV1 {
+                    last_synced_blob_id,
+                    sliver_type,
+                })) => {
+                    tracing::info!(
+                        "Resuming shard sync from blob id: {}, sliver type: {}",
+                        last_synced_blob_id,
+                        sliver_type
+                    );
+                    match sliver_type {
+                        SliverType::Primary => ShardLastSyncStatus::Primary {
+                            last_synced_blob_id: Some(last_synced_blob_id),
+                        },
+                        SliverType::Secondary => ShardLastSyncStatus::Secondary {
+                            last_synced_blob_id: Some(last_synced_blob_id),
+                        },
+                    }
+                }
+                None => ShardLastSyncStatus::Primary {
+                    last_synced_blob_id: None,
+                },
+            }
+        };
+        Ok(last_sync_status)
     }
 
     #[tracing::instrument(
@@ -532,15 +539,14 @@ impl ShardStorage {
     ) -> Result<(), SyncShardClientError> {
         // Helper to track the number of scanned blobs to test recovery. Not used in production.
         #[cfg(feature = "failure_injection")]
-        let mut scan_count = 0;
+        let mut scan_count: u64 = 0;
 
         let mut blob_info_iter = node
             .storage
             .certified_blob_info_iter_before_epoch(epoch, last_synced_blob_id);
 
         let mut next_blob_info = blob_info_iter.next().transpose()?;
-        while next_blob_info.is_some() {
-            let next_starting_blob_id = next_blob_info.as_ref().unwrap().0;
+        while let Some((next_starting_blob_id, _)) = next_blob_info {
             tracing::debug!(
                 "Syncing shard to before epoch: {}. Starting blob id: {}",
                 epoch,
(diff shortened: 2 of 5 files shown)


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -107,19 +126,31 @@ impl BlobSyncHandler {
...
             let cancel_token = CancellationToken::new();
             let synchronizer =
-                BlobSynchronizer::new(event, event_index, self.node.clone(), cancel_token.clone());
-
-            let sync_handle = tokio::spawn(
-                self.clone()
-                    .sync(synchronizer, start, self.semaphore.clone())
+                BlobSynchronizer::new(event.blob_id, self.node.clone(), cancel_token.clone());
+
+            let handler_clone = self.clone();
```
mlegner: Question: Is my understanding correct that we can easily clone the `BlobSyncHandler` here because it only contains `Arc`s? If that's the case, could we add a comment to the struct definition that this has to remain cheaply cloneable? (This seems to have been the case already before, so not related to your changes.)

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -134,12 +165,64 @@ impl BlobSyncHandler {
...
+                    })
+                    .instrument(spawned_trace)
+                    .await;
+                handler_clone.blob_syncs_notify_reads.notify(&blob_id, &());
+                result
+            });
+            entry.insert(InProgressSyncHandle {
+                cancel_token,
+                blob_sync_handle: Some(sync_handle),
+            });
```
mlegner: Remark: This code is almost exactly the same as in the `start_sync` function; this could be deduplicated.

crates/walrus-service/src/node/blob_sync.rs
halfprice: Yes, that's correct. Added a struct level doc.

crates/walrus-service/src/node/blob_sync.rs
halfprice: Extracted it out

crates/walrus-service/src/test_utils.rs
```diff
@@ -714,6 +714,19 @@ impl TestClusterBuilder {
         self
     }
 
+    /// Sets the individual event providers for each storage node.
```
mlegner, line 718: Nit: Can we document that the length of the slice should be the same as the number of storage nodes?

Also, in many cases the caller could move ownership of the providers such that we wouldn't have to clone them. But as this is a test function, that probably doesn't matter too much.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -140,6 +154,16 @@ impl ShardStorage {
...
+        let pending_recover_slivers_cf_name = pending_recover_slivers_column_family_name(id);
+        if database
+            .cf_handle(&pending_recover_slivers_cf_name)
+            .is_none()
+        {
+            let (_, options) = Self::pending_recover_slivers_column_family_options(id, db_config);
+            database
+                .create_cf(&pending_recover_slivers_cf_name, &options)
+                .map_err(typed_store_err_from_rocks_err)?;
+        }
```
mlegner, line 177: Remark: I feel we have more and more code duplication in this function. Maybe at some point (not this PR), we can deduplicate these operations through a macro or so.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -387,29 +429,38 @@ impl ShardStorage {
...
-                    sliver_type == SliverType::Secondary,
-                )
-            }
-            None => (None, false),
-        };
+        let (mut last_synced_blob_id, skip_primary, skip_secondary) =
+            if shard_status == ShardStatus::ActiveRecover {
+                // We are in recovery mode. Skip happy path syncing and directly jump to recover
+                // missing blobs.
+                (None, true, true)
```
mlegner: Suggestion: Replace this triple by an enum to make the current status more explicit:
```rust
enum ShardSyncStatus {
    Recovery,
    Primary {
        last_synced_blob_id: Option<BlobId>,
    },
    Secondary {
        last_synced_blob_id: Option<BlobId>,
    },
}
```
Then you can do matches like 
```rust
if let ShardSyncStatus::Primary{last_synced_blob_id}) = shard_sync_status { ... }
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -421,13 +472,44 @@ impl ShardStorage {
...
+                    "sync shard simulated sync failure after start recovery"
+                )))
+            });
+
+            loop {
+                self.recover_missing_blobs(blob_sync_handler.clone())
+                    .await?;
+
+                if self.pending_recover_slivers.is_empty() {
+                    // TODO: check recovered blob consistency.
```
mlegner: Question: Isn't this handled by the blob sync internally?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -421,13 +472,44 @@ impl ShardStorage {
...
+                    break;
+                }
+                tracing::warn!(
+                    "Recovering missing blobs for shard: {} still misses blobs.
+                    Retrying in 60 seconds.",
+                    self.id
```
mlegner: Nit: The shard ID is already included in the span and could be removed here.

crates/walrus-service/src/node/storage.rs
```diff
@@ -178,6 +178,7 @@ pub struct DatabaseConfig {
     shard: DatabaseTableOptions,
     shard_status: DatabaseTableOptions,
     shard_sync_progress: DatabaseTableOptions,
+    pending_recover_slivers: DatabaseTableOptions,
```
mlegner, line 180: Remark: We now have so many tables, it becomes very tedious to change the config as this has to be done for every table separately. It would be nice if in the config file one could specify custom "default" and "blob-optimized" options, which would apply to all corresponding tables that don't have a more specific config.

crates/walrus-service/src/node/storage.rs
```diff
@@ -224,6 +231,57 @@ pub struct Storage {
...
+            iter,
+            before_epoch,
+            only_certified,
+        }
+    }
+}
+
+impl Debug for BlobInfoIter<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("BlobInfoIter").finish()
```
mlegner: Suggestion: Include the `before_epoch` and `only_certified` in the debug output.

crates/walrus-service/src/node/storage.rs
```diff
@@ -224,6 +231,57 @@ pub struct Storage {
...
+        let mut item = self.iter.next();
+        if !self.only_certified {
+            return item;
+        }
+
+        while let Some(Ok((_, ref blob_info))) = item {
+            if blob_info.is_certified()
+                && blob_info
+                    .status_changing_epoch(blob_info::BlobCertificationStatus::Certified)
+                    .unwrap()
```
mlegner: Suggestion: Can you replace this with an `expect` describing why this can never panic?

crates/walrus-service/src/node/storage.rs
```diff
@@ -358,14 +416,21 @@ impl Storage {
...
-        match range {
-            Some(range) => self.blob_info.safe_range_iter(range),
-            None => self.blob_info.safe_iter(),
-        }
+        before_epoch: Epoch,
+        last_synced_blob_id: Option<BlobId>,
+    ) -> BlobInfoIter {
+        BlobInfoIter::new(
+            Box::new(match last_synced_blob_id {
+                Some(starting_blob_id) => self
```
mlegner, line 428: Nit: Can we use the same name `last_synced_blob_id` here?

crates/walrus-service/src/node/storage.rs
```diff
@@ -224,6 +231,57 @@ pub struct Storage {
     config: DatabaseConfig,
 }
 
+/// A iterator over the blob info table.
+pub struct BlobInfoIter<'a> {
+    iter: Box<dyn Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send + 'a>,
```
mlegner, line 235: Question: Do we actually need the trait object here? AFAICT, we always have the same iterator type.

crates/walrus-service/src/node/storage.rs
```diff
@@ -358,14 +416,21 @@ impl Storage {
...
-        }
+        before_epoch: Epoch,
+        last_synced_blob_id: Option<BlobId>,
+    ) -> BlobInfoIter {
+        BlobInfoIter::new(
+            Box::new(match last_synced_blob_id {
+                Some(starting_blob_id) => self
+                    .blob_info
+                    .safe_range_iter((Excluded(starting_blob_id), Unbounded)),
+                None => self.blob_info.safe_iter(),
```
mlegner, line 431: Question: What happens to these iterators when data is written to the DB?

crates/walrus-service/src/node/storage.rs
```diff
@@ -1245,4 +1310,96 @@ pub(crate) mod tests {
...
+        let new_epoch = 3;
+
+        let blob_ids = [
+            BlobId([1; 32]), // Not certified.
+            BlobId([2; 32]), // Not exist.
+            BlobId([3; 32]), // Certified within epoch 2
+            BlobId([4; 32]), // Certified after epoch 2
+            BlobId([5; 32]), // Invalid
+            BlobId([6; 32]), // Certified within epoch 2
+            BlobId([7; 32]), // Not exist.
```
mlegner: Nit: Can we shift the values by 1 so that `blob_ids[x] == BlobId([x;32])`? This would help with following the code.

crates/walrus-service/src/node/storage.rs
```diff
@@ -1245,4 +1310,96 @@ pub(crate) mod tests {
...
+            (blob_ids[3], certified_blob_info(3)),
+            (blob_ids[4], invalid_blob_info(2)),
+            (blob_ids[5], certified_blob_info(2)),
+        ]);
+
+        let mut batch = blob_info.batch();
+        batch.insert_batch(&blob_info, blob_info_map.iter())?;
+        batch.write()?;
+
+        for blob_id in blob_ids.iter().take(2) {
```
mlegner, line 1394: Nit: Can we also test the `None` case?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -6,20 +6,28 @@
...
 use anyhow::Context;
 use fastcrypto::traits::KeyPair;
+use futures::{stream::FuturesUnordered, StreamExt};
 use regex::Regex;
 use rocksdb::{Options, DB};
 use serde::{Deserialize, Serialize};
 use typed_store::{
-    rocks::{errors::typed_store_err_from_rocks_err, DBBatch, DBMap, ReadWriteOptions, RocksDB},
+    rocks::{
+        be_fix_int_ser as ToRocksDbKey,
```
mlegner: Nit: Seems a bit weird to me to rename this looking like a type. Can we use `to_rocks_db_key` instead?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -456,10 +538,14 @@ impl ShardStorage {
...
-                .context("Scanning certified blobs encountered error")?
-        {
+
+        let mut blob_info_iter = node
+            .storage
+            .certified_blob_info_iter_before_epoch(epoch, last_synced_blob_id);
+
+        let mut next_blob_info = blob_info_iter.next().transpose()?;
+        while next_blob_info.is_some() {
+            let next_starting_blob_id = next_blob_info.as_ref().unwrap().0;
```
mlegner: Nit: Instead of the `is_some` + `unwrap` construct, we could use a `while let`:
```suggestion
        while let Some((next_starting_blob_id, _)) = &next_blob_info {
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -456,10 +538,14 @@ impl ShardStorage {
...
-            next_certified_blob_id_before_epoch(&node.storage, epoch, last_synced_blob_id)
-                .context("Scanning certified blobs encountered error")?
-        {
+
+        let mut blob_info_iter = node
+            .storage
+            .certified_blob_info_iter_before_epoch(epoch, last_synced_blob_id);
+
+        let mut next_blob_info = blob_info_iter.next().transpose()?;
+        while next_blob_info.is_some() {
```
mlegner: Important: This loop now has almost 100 lines. I think it is time to refactor this a bit to keep it readable.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -482,8 +568,7 @@ impl ShardStorage {
                 )
                 .await?;
             for blob in fetched_slivers.iter() {
-                tracing::debug!("Synced blob id: {} to before epoch: {}.", blob.0, epoch,);
-                //TODO(#705): Track missing blobs.
+                tracing::info!("Synced blob id: {} to before epoch: {}.", blob.0, epoch,);
```
mlegner: Question: What is the motivation of increasing this to info level? There will be a lot of these messages...

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -536,13 +629,129 @@ impl ShardStorage {
...
         }
 
+        if next_blob_info.is_none() {
+            return Ok(());
+        }
+
+        let mut batch = match sliver_type {
+            SliverType::Primary => self.primary_slivers.batch(),
+            SliverType::Secondary => self.secondary_slivers.batch(),
+        };
```
mlegner: Question: Why do you create the batch from these tables?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -536,13 +629,129 @@ impl ShardStorage {
...
+        batch.write()?;
+
+        Ok(())
+    }
+
+    // Given a iterator over blob info table that is currently pointing to `next_blob_info`,
+    // and `fetched_blob_id` returned from the remote storage node, checks if there are any
+    // missing blobs between `next_blob_info` and `fetched_blob_id`.
+    // Returns the next blob info that is greater than `fetched_blob_id` or None if there are no.
+    fn check_missing_blobs(
```
mlegner: Nit: Can we rename this a bit? AFAIU, the purpose of this function is to add the missing blobs to the DB to be recovered later.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -536,13 +629,129 @@ impl ShardStorage {
...
+        &self,
+        blob_info_iter: &mut BlobInfoIter,
+        mut next_blob_info: Option<(BlobId, BlobInfo)>,
+        fetched_blob_id: BlobId,
+        sliver_type: SliverType,
+        db_batch: &mut DBBatch,
+    ) -> Result<Option<(BlobId, BlobInfo)>, TypedStoreError> {
+        let Some((mut next_blob_id, _)) = next_blob_info else {
+            assert!(blob_info_iter.next().transpose()?.is_none());
```
mlegner: Important: Do we really want to panic here? Can we change this to either an error or a `debug_assert`? Also, maybe we could just change the type of that parameter to `(BlobId, BlobInfo)>`.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -536,13 +629,129 @@ impl ShardStorage {
...
+        }
+        batch.write()?;
+
+        Ok(())
+    }
+
+    // Given a iterator over blob info table that is currently pointing to `next_blob_info`,
+    // and `fetched_blob_id` returned from the remote storage node, checks if there are any
+    // missing blobs between `next_blob_info` and `fetched_blob_id`.
+    // Returns the next blob info that is greater than `fetched_blob_id` or None if there are no.
```
mlegner: Nit: Can we make these docstrings instead of normal comments? (Also applies to other functions.)

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -536,13 +629,129 @@ impl ShardStorage {
...
+            "Start recovering missing blob {} in shard {}",
+            blob_id,
+            self.id
+        );
+        // Note that `blob_sync_handler` tries to recover ALL missing shards for the given blob
+        // id. `start_sync_for_blob_recovery` can be called multiple times, but there will
+        // be only one active recovery task for the given blob id.
+        blob_sync_handler
+            .start_sync_for_blob_recovery(blob_id, start_time)
+            .await;
```
mlegner, line 776: Important: Essentially this means that for each blob missing after shard sync, we go through all shard storages and check if the corresponding slivers are stored. This seems a massive waste of resources if we just want to recover one sliver pair, especially for large storage nodes that can have tens of shards.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -536,13 +629,129 @@ impl ShardStorage {
...
+                    (SliverType::Primary, blob_id),
+                    (SliverType::Secondary, blob_id),
+                ],
+            )?;
+            batch.write()?;
+        }
+
+        // If sliver is not recovered, it won't be removed from the `pending_recover_slivers` table.
+        // Later retry will revisit this blob again.
         Ok(())
```
mlegner, line 791: Important: AFAIU, the blob sync spawns a new task, so we will not have the sliver stored after starting the blob sync. OTOH, we probably don't have to implement retries here as the blob sync itself takes care of that.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -989,110 +1199,97 @@ mod tests {
...
     }
 
-    #[tokio::test]
-    async fn test_next_certified_blob_id_before_epoch() -> TestResult {
+    pub fn random_blob_id() -> BlobId {
+        let mut bytes = [0; 32];
+        rand::thread_rng().fill_bytes(&mut bytes);
+        BlobId(bytes)
+    }
```
mlegner: Hint: There already exists such a function: `walrus_core::test_utils::random_blob_id()`.

crates/walrus-service/src/node/storage.rs
halfprice: Yes. The issue is that DBMap::SafeIter is a private type.

crates/walrus-service/src/node/storage.rs
halfprice: Done

crates/walrus-service/src/node/storage.rs
halfprice: Done

crates/walrus-service/src/node/storage.rs
halfprice: Done

crates/walrus-service/src/node/storage.rs
halfprice: Iterator operates on a snapshot of a DB. So if there are new writes after the iterator is created, they won't be observed.

crates/walrus-service/src/node/storage.rs
halfprice: Done

crates/walrus-service/src/node/storage.rs
halfprice: Done

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs
halfprice: Added a TODO

crates/walrus-service/src/node/storage/shard.rs
halfprice: This is not blob consistency, but that the synced blob is consistent with what the shard wants to sync: it has received all blobs. 

This meant as an invariant check only in tests.

Revised the TODO.

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs
halfprice: Reverted. Was used in debugging...

crates/walrus-service/src/node/storage/shard.rs
halfprice: Because I blindly accepted a code suggestion... Changed.

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs
halfprice: I think so because next_blob_info should be the last item read from blob_info_iter, and it this is not none, it's a invariant violation. This check is meant for catching implementation bugs. Changed to debug_assert and added more comments.

crates/walrus-service/src/node/storage/shard.rs
halfprice: This is not a TODO, but a comment.

`recover_missing_blobs` waits until all blobs are synced. If there is any missing blobs, it'll restart recovering again. The comment here is for that case.

crates/walrus-service/src/test_utils.rs
halfprice: Done

crates/walrus-service/src/node/storage/shard.rs
halfprice: Yeah, I should have pointed it out. The issue with that one is that it always returns the same blob id since it always uses the same seed. I moved `random()` into BlobID, so that it's easier to find when next time people need it. Maybe we can replace all the random_blob_id() with this one.

crates/walrus-service/src/node/storage/shard.rs
halfprice: update it to `check_and_record_missing_blobs`

crates/walrus-service/src/node/storage/shard.rs
halfprice: Done. Break it into several helper functions.

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs
mlegner: Ah, I see, thanks for the clarification.

crates/walrus-service/src/node/storage.rs
mlegner: Clarification: This is not something for this PR but for a follow-up.

crates/walrus-service/src/node/storage.rs
mlegner: Hmm... But couldn't we make the `BlobInfoIter` generic over an inner iterator implementing `Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>`? Something like this:

```rust
/// A iterator over the blob info table.
pub struct BlobInfoIter<I> {
    iter: I,
    before_epoch: Epoch,
    only_certified: bool,
}

impl<I> BlobInfoIter<I>
where
    I: Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>>,
{
    pub fn new(iter: I, before_epoch: Epoch, only_certified: bool) -> Self {
        Self {
            iter,
            before_epoch,
            only_certified,
        }
    }
}
```

crates/walrus-service/src/node/storage.rs
mlegner: OK, thanks for the insight! Just for the future: We need to make sure that the shard sync and blob sync still work properly if blobs are deleted.

crates/walrus-service/src/node/storage/shard.rs
mlegner: Ah, good point. Yes, maybe you could just replace the body of the `random_blob_id` with this one?

crates/walrus-service/src/node/storage/shard.rs
mlegner: Hmm... And as the `start_sync_for_blob_recovery` doesn't actually start multiple tasks for the same blob, we just call it again when we check later?

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -134,12 +148,76 @@ impl BlobSyncHandler {
         Ok(())
     }
 
+    /// An entry point for blob sync initiated by the shard sync process.
+    #[tracing::instrument(skip_all, fields(otel.kind = "PRODUCER"))]
+    pub fn start_sync_for_blob_recovery(
+        &self,
+        blob_id: BlobId,
+        start: tokio::time::Instant,
+    ) -> Registration<BlobId, ()> {
```
mlegner, line 157: Question: What is the purpose of the return value? We actually immediately drop it in the currently only place where we call this function.

halfprice: This is for the caller to be able wait on the task associated with the BlobID. It was await at the caller so it's not dropped immediately.

The new version of the code doesn't involve this.

crates/walrus-service/src/node/storage/shard.rs
halfprice: No longer relevent since we are moving with #746

crates/walrus-service/src/node/storage/shard.rs
halfprice: Done in the #746

PR #737, opened 2024-08-28
fix(client): improve output for incompatible system object

mlegner:


Cargo.lock
giac-mysten, line 1: Question: many changes to the cargo lock. Are these expected?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -297,13 +297,22 @@ async fn get_system_package_id(
...
-        .map_err(|_| SuiClientError::WalrusSystemObjectDoesNotExist(system_object_id))?;
-
-    get_sui_object_from_object_response::<SystemObject>(&response, system_object_id)
-        .map_err(|_| SuiClientError::WalrusSystemObjectDoesNotExist(system_object_id))?;
-
-    let object_pkg_id = get_package_id_from_object_response(&response)
-        .map_err(|_| SuiClientError::WalrusSystemObjectDoesNotExist(system_object_id))?;
+        .map_err(|error| {
+            tracing::debug!(%error, "unable to get the Walrus system object");
+            SuiClientError::WalrusSystemObjectDoesNotExist(system_object_id)
```
giac-mysten, line 302: Remark: i guess the differentiation may be done with different error types. Calling everything `WalrusSystemObjectDoesNotExist` is a bit misleading. But good for now IMO.

Cargo.lock
mlegner: I believe that is due to a mistake in #731 where I ran the simtests before committing.

crates/walrus-sui/src/client/read_client.rs
mlegner: My idea there is to provide that extra info just in the debug messages as a normal user probably doesn't care too much about the differences.

Issue #535, opened 2024-06-14
Additional client optimizations


mlegner:
- [x]  When a blob is not certified but registered by the active wallet, do not register it again. -> #741 
- [x] #688
- [x] When a blob is certified but not for a sufficient duration, don't send any metadata and slivers at all. -> #741 

Both of these help when resuming a failed store operation.

PR #739, opened 2024-08-28
feat(client): list all registered blobs

giac-mysten:
Adds a new `list-blobs` command to the client, that shows a table of all registered and certified blobs for the active wallet.

Contributes to #262, #535

### DIFF
--- a/crates/walrus-service/src/client/cli_utils/cli_output.rs
+++ b/crates/walrus-service/src/client/cli_utils/cli_output.rs
@@ -302,24 +302,18 @@ impl CliOutput for Vec<Blob> {
         table.set_format(default_table_format());
         table.set_titles(row![
             b->"Blob ID",
-            b->"Unencoded Size",
-            b->"Stored Epoch",
-            b->"Certified Epoch",
+            bc->"Unencoded size",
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client/cli_utils/cli_output.rs
```diff
@@ -295,6 +296,37 @@ impl CliOutput for InfoOutput {
...
+    fn print_cli_output(&self) {
+        let mut table = Table::new();
+        table.set_format(default_table_format());
+        table.set_titles(row![
+            b->"Blob ID",
+            b->"Unencoded Size",
+            b->"Stored Epoch",
+            b->"Certified Epoch",
+            b->"Object ID",
+            // TODO(giac): too much info? too little? add storage?
```
karlwuest: Suggestion: I think it would be good to add the expiration epoch, but we can probably remove either the certified or registered epoch.

mlegner: I agree, the expiration epoch would be great. We could actually remove both registered and certified epochs and instead have a boolean "certified" column.

PR #741, opened 2024-08-28
feat(client): check if we own an already-registered blob object for the blob id

### LINKED ISSUE #535: Additional client optimizations

giac-mysten:
Before storing a new blob, the client checks if it owns a Blob object already registered for the same blob ID (for a sufficient duration). If it does, it uses this blob object instead of registering a new one.

Closes #535

### DIFF
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -272,7 +272,12 @@ enum Commands {
         blob_id_decimal: BlobIdDecimal,
     },
     /// List all registered blobs for the current wallet.
-    ListBlobs,
+    ListBlobs {
+        #[clap(long, action)]
+        #[serde(default)]
+        /// The output list of blobs will include expired blobs.
+        include_expired: bool,
+    },
 }
 
 #[derive(Debug, Clone, Args, Deserialize)]
@@ -646,10 +651,10 @@ async fn run_app(app: App) -> Result<()> {
         Commands::ConvertBlobId { blob_id_decimal } => {
             BlobIdConversionOutput::from(blob_id_decimal).print_output(app.json)?;
         }
-        Commands::ListBlobs => {
+        Commands::ListBlobs { include_expired } => {
             let contract_client =
                 SuiContractClient::new(wallet?, config?.system_object, app.gas_budget).await?;
-            let blobs = contract_client.owned_blobs().await?;
+            let blobs = contract_client.owned_blobs(include_expired).await?;
             blobs.print_output(app.json)?;
         }
     }
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -312,7 +312,7 @@ impl<T: ContractClient> Client<T> {
     ) -> ClientResult<Option<Blob>> {
         Ok(self
             .sui_client
-            .owned_blobs()
+            .owned_blobs(false)
             .await?
             .into_iter()
             .find(|blob| {
--- a/crates/walrus-sui/src/test_utils/mock_clients.rs
+++ b/crates/walrus-sui/src/test_utils/mock_clients.rs
@@ -270,7 +270,7 @@ impl ContractClient for MockContractClient {
         &self.read_client
     }
 
-    async fn owned_blobs(&self) -> SuiClientResult<Vec<Blob>> {
+    async fn owned_blobs(&self, include_expired: bool) -> SuiClientResult<Vec<Blob>> {
         let blob = Blob {
             id: ObjectID::random(),
             stored_epoch: self.current_epoch,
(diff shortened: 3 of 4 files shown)


crates/walrus-service/src/client.rs
```diff
@@ -277,18 +278,48 @@ impl<T: ContractClient> Client<T> {
         })
     }
 
-    /// Reserves the space for the blob on chain and registers it.
+    /// Check if the blob is registered; if not, reserves the space for the blob and registers it.
```
```suggestion
    /// Checks if the blob is registered; if not, reserves the space for the blob and registers it.
```

crates/walrus-service/src/client.rs
```diff
@@ -277,18 +278,48 @@ impl<T: ContractClient> Client<T> {
...
+
+    /// Checks if the blob is registered by the active wallet for a sufficient duration.
+    async fn is_blob_registered_in_wallet(
+        &self,
+        blob_id: &BlobId,
+        epochs_ahead: u64,
+    ) -> ClientResult<Option<Blob>> {
+        Ok(self
+            .sui_client
+            .owned_blobs()
```
mlegner: Question: Is there any way of finding a matching object besides getting all of them and then filtering them locally? (I guess not, but would be nice...)

crates/walrus-sui/src/test_utils/mock_clients.rs
```diff
@@ -269,6 +269,24 @@ impl ContractClient for MockContractClient {
...
+            erasure_code_type: EncodingType::RedStuff,
+            certified_epoch: None,
+            storage: StorageResource {
+                id: ObjectID::random(),
+                start_epoch: self.current_epoch,
+                end_epoch: self.current_epoch + 1,
+                storage_size: 50_000,
+            },
+        };
+        Ok(vec![blob])
```
mlegner: Suggestion: Can we have multiple blobs here? Maybe just the one blob cloned?

crates/walrus-service/bin/client.rs
```diff
@@ -644,6 +646,12 @@ async fn run_app(app: App) -> Result<()> {
         Commands::ConvertBlobId { blob_id_decimal } => {
             BlobIdConversionOutput::from(blob_id_decimal).print_output(app.json)?;
         }
+        Commands::ListBlobs => {
+            let contract_client =
+                SuiContractClient::new(wallet?, config?.system_object, app.gas_budget).await?;
+            let blobs = contract_client.owned_blobs().await?;
+            blobs.print_output(app.json)?;
+        }
```
mlegner: Question: What do we do with expired blobs? Can we mark them in some way or filter them out unless a flag `include-expired` is set?

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -222,3 +222,62 @@ fn error_kind_matches(actual: &ClientErrorKind, expected: &ClientErrorKind) -> b
...
+
+    // Now ask the client to store again.
+    let blob_store = client
+        .inner
+        .reserve_and_store_blob(&blob, epochs_ahead_required, false)
+        .await?;
+
+    if let BlobStoreResult::NewlyCreated { blob_object, .. } = blob_store {
+        // Check if the storage object used is the same.
+        let should_be_same_object = epochs_ahead_registered >= epochs_ahead_required;
```
mlegner: Suggestion: Can we add this as a function parameter to make the expectation more explicit?

crates/walrus-service/src/client.rs
giac-mysten: I don't think so, but maybe @karlwuest knows more? the filters I see are `Package, MoveModule, AddressOwner, ObjectOwner, ObjectId, ObjectIds`. Maybe with GraphQL (at some point in the future?)

crates/walrus-e2e-tests/tests/test_client.rs
giac-mysten: I had it this way at the beginning, then thought it's nice to have it automatic. Will revert.

crates/walrus-service/bin/client.rs
giac-mysten: nice feature. this is a bit tangential to this pr, but added it.

crates/walrus-sui/src/test_utils/mock_clients.rs
giac-mysten: I added two blobs, one of which expired and is returned only if `include_expired`.

PR #745, opened 2024-08-29
feat: Integrate event processor with existing walrus code

sadhansood:
The PR integrates new checkpoint based event processor with existing walrus code. I am working on adding tests but PR is ready for review (and it has gotten big already), so wanted to get some signal early on. 
The biggest underlying change is to replace `usize` with `u128` for tracking event cursor index in the current codebase. This is required because with the new event processor, we assign an event sequence number (u128) to each event which is comprised of two u64 parts: `CheckpointSequenceNumber` (which is the sequence number of the checkpoint the event is part of) and `counter` (index of the event in the sequence of emitted events in a checkpoint). This event sequence number will be further used to track the event index up to which event blobs have been certified by the committee in the new endpoint `certify_event_blob` in the walrus smart contract which we are yet to add


crates/walrus-sui/src/types.rs
```diff
@@ -705,14 +705,16 @@ impl AssociatedSuiEvent for InvalidBlobId {
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub enum BlobEvent {
     /// A registration event.
     Registered(BlobRegistered),
     /// A certification event.
     Certified(BlobCertified),
     /// An invalid blob ID event.
     InvalidBlobID(InvalidBlobId),
+    /// An end of checkpoint event.
+    EndOfCheckpoint(),
```
mlegner: Important: I don't think this should be added here: Currently, the different variants all correspond to actual Sui events of the Walrus contract, and, AFAIU, the `EndOfCheckpoint` is just an internal "dummy" event, which should be filtered out by the event processor.

This also leads to some clumsy code like the panics below.

My suggestion would be to instead change the `IndexedBlobEvent` to take a `blob_or_dummy_event: BlobOrDummyEvent` where
```rust
enum BlobOrDummyEvent {
    BlobEvent(BlobEvent),
    EndOfCheckpoint,
}
```

crates/walrus-event/src/lib.rs
```diff
@@ -6,19 +6,21 @@ use std::path::PathBuf;
...
     pub rest_url: String,
-    /// The URL of the checkpoint client to connect to.
+    /// The path where local dbs are stored.
     pub path: PathBuf,
     /// The path to the Sui genesis file.
     pub sui_genesis_path: PathBuf,
     /// ObjectID of the Walrus system package.
     pub system_pkg_id: ObjectID,
+    /// Number of checkpoints per blob.
+    pub checkpoints_per_blob: u64,
```
mlegner: Important: AFAIU, this needs to be the same among all storage nodes, right? In this case we shouldn't make this configurable through a config file. Either this should be a constant defined in the code or, maybe better, stored in the on-chain system object (cc @karlwuest).

crates/walrus-service/src/node/config.rs
```diff
@@ -70,6 +71,9 @@ pub struct StorageNodeConfig {
     /// Configuration for TLS of the rest API.
     #[serde(default)]
     pub tls: TlsConfig,
+    /// Configuration for running checkpoint processor
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub event_processor_config: Option<EventProcessorConfig>,
```
mlegner, line 79: Remark: This contains a field (`system_pkg_id`) that is redundant with other parts of the config (in this case the `SuiConfig`).

crates/walrus-service/src/node/storage/event_sequencer.rs
```diff
@@ -53,6 +58,11 @@ impl EventSequencer {
     pub fn advance(&mut self) {
         while self.queue.remove(&self.head_index).is_some() {
             self.head_index += 1;
+            // Move cursor over a gap if it exists
+            // Handle sequence gaps
```
mlegner: Question: IIUC, we are now talking about two different kinds of gaps: The docstring of the function ("Advances the queue to the next gap in the sequence") talks about gaps in the sequence of handled events. This comment talks about gaps of unassigned event indices. Is that right?

Suggestion: If this is the case, this should be clarified.

crates/walrus-service/src/node/storage/event_sequencer.rs
```diff
@@ -12,8 +12,9 @@ use sui_types::event::EventID;
 /// provides access to the `EventID`s sequentially starting from 0.
 #[derive(Debug, Default)]
 pub(super) struct EventSequencer {
-    head_index: usize,
-    queue: HashMap<usize, EventID>,
+    head_index: u128,
+    queue: HashMap<u128, EventID>,
+    event_range: HashMap<u128, u128>,
```
mlegner: Suggestion: Can you describe what is stored in the `event_range`?

Also: The struct's docstring doesn't seem to be fully representative if we no longer have sequential indices.

crates/walrus-service/src/node/storage/event_sequencer.rs
```diff
@@ -23,7 +24,7 @@ impl EventSequencer {
     }
 
     /// Creates a new `EventSequencer`, that continues sequencing from the provided next index.
-    pub fn continue_from(next_index: usize) -> Self {
+    pub fn continue_from(next_index: u128) -> Self {
         Self {
             head_index: next_index,
             ..Self::default()
```
mlegner: Question: Is this function still correct when we no longer have sequential indices?

crates/walrus-service/src/node/metrics.rs
```diff
@@ -86,7 +86,7 @@ define_node_metric_set! {
         (slivers_stored_total, "The total number of slivers stored", &["sliver_type"]),
         (slivers_retrieved_total, "Total number of sliver instances returned", &["sliver_type"])
     ],
-    GenericGaugeVec<AtomicU64>: [
+    GenericGaugeVec<AtomicF64>: [
```
mlegner: Question: Why does a `u64` not suffice here?

crates/walrus-event/src/blob_writer.rs
```diff
@@ -59,20 +62,24 @@ struct EventBlobWriter {
     /// Client to store the slivers and metadata of the event blob.
     client: Client,
```
mlegner: This is not related to the changes of this PR, but now that I went through the code again, I realized that we are using the public API of the storage node. I'm not sure this is the best way to store data related to these event blobs:
- We will run into complications with storing metadata and slivers for blobs that are not registered first.
- This creates unnecessary overhead compared to an internal non-HTTP API.
- Under heavy load, the HTTP API may be congested by normal clients.

I think it would be good to investigate if we could have an internal interface through which we can store this data.

crates/walrus-event/src/event_processor.rs
```diff
@@ -54,106 +61,138 @@ const MAX_TIMEOUT: Duration = Duration::from_secs(60);
 const RETRY_DELAY: Duration = Duration::from_secs(5);
 
 #[derive(Eq, PartialEq, Default, Clone, Debug, Serialize, Deserialize)]
```
mlegner: Suggestion: Also derive `Copy` here.

crates/walrus-event/src/event_processor.rs
```diff
@@ -417,22 +470,26 @@ impl CheckpointProcessor {
             .seek_to_first()
             .next()
             .map(|item| item.0)
-            .unwrap_or(WalrusEventID::new(0, 0));
-        let checkpoint_processor = CheckpointProcessor {
+            .unwrap_or(EventSequenceNumber::new(0, 0));
+        let checkpoint_processor = EventProcessor {
```
mlegner: Nit: Given that `CheckpointProcessor` was renamed to `EventProcessor`, we should also change variable names and docstrings accordingly.

crates/walrus-event/src/event_processor.rs
```diff
@@ -54,106 +61,138 @@ const MAX_TIMEOUT: Duration = Duration::from_secs(60);
...
+impl EventSequenceNumber {
     pub fn new(sequence_number: CheckpointSequenceNumber, counter: u64) -> Self {
-        WalrusEventID {
-            sequence_number,
+        Self {
+            checkpoint_sequence_number: sequence_number,
             counter,
         }
     }
+    pub fn from(index: u128) -> Self {
```
mlegner: Nit: The function name is a bit too generic. Maybe `from_index` instead?

crates/walrus-service/src/node.rs
```diff
@@ -324,6 +328,7 @@ pub struct StorageNodeInner {
     storage: Storage,
     encoding_config: Arc<EncodingConfig>,
     event_provider: Box<dyn SystemEventProvider>,
+    event_processor: Option<EventProcessor>,
```
mlegner: Question: What is the intended relationship between the event provider and event processor? My understanding is that we will use one or the other, is that right?

Question: Could the `EventProcessor` simply implement the `SystemEventProvider` trait and we would just instantiate the `StorageNode` with either the `SuiSystemEventProvider` or the `EventProcessor`?

crates/walrus-service/src/node.rs
```diff
@@ -407,6 +414,43 @@ impl StorageNode {
         StorageNodeBuilder::default()
     }
 
+    fn create_client(config: &StorageNodeConfig) -> anyhow::Result<Client> {
```
mlegner: Nit: Can we call this `create_sdk_client_for_self` or something like that to clarify what kind of client this is?

crates/walrus-event/src/lib.rs
karlwuest: Hmmm... can we just define it as a function of the number of shards in the system? This determines the maximum blob size, so it can inform what batch size makes sense for the event blob.

Question (I didn't look at the code): since the name is `checkpoints_per_blob`, do we create event blobs purely on the number of Sui checkpoints, independent on how many events these contain? Why do we not base this on the number of events (+ e.g., at least one every Sui epoch to ensure that we can retrieve all events even if all full nodes heavily prune)?

crates/walrus-event/src/event_processor.rs
```diff
@@ -54,106 +61,138 @@ const MAX_TIMEOUT: Duration = Duration::from_secs(60);
...
+        Self {
+            checkpoint_sequence_number: (index >> 64) as u64,
+            counter: index as u64,
+        }
+    }
```
jpcsmith: Question: Why the need to represent this as a u128 as opposed to just a totally ordered pair of u64s?

crates/walrus-sui/src/types.rs
jpcsmith: Suggestion: Remove the `()` at the end of the enum variant.

crates/walrus-service/bin/node.rs
```diff
@@ -183,6 +183,7 @@ impl MetricsAndLoggingRuntime {
 struct StorageNodeRuntime {
     walrus_node_handle: JoinHandle<anyhow::Result<()>>,
     rest_api_handle: JoinHandle<Result<(), io::Error>>,
+    event_processor_handle: JoinHandle<anyhow::Result<()>>,
```
jpcsmith: Suggestion: Move to inside the storage node.

crates/walrus-service/src/node.rs
```diff
@@ -427,12 +471,117 @@ impl StorageNode {
     }
 
     async fn process_events(&self) -> anyhow::Result<()> {
+        if let Some(event_processor) = self.inner.event_processor.as_ref() {
+            self.process_events_using_verified_checkpoints(event_processor.clone())
+                .await
+        } else {
+            self.process_events_using_events_api().await
+        }
```
jpcsmith: Suggestion: Have the processor implement the `SystemEventProvider` trait and provide it during construction of the storage node.

Afterwards, unify the `process_events_using_*` methods.

crates/walrus-service/src/node/storage/event_sequencer.rs
jpcsmith, line 1: Question: Could we keep the event sequence number as globally unique, sequential integer? That way it does not require any additional information such as the distribution of events across checkpoints.. We could of course also store the checkpoint number (when known) with the event sequence number.

It would also be common to both sources of events (verified and unverified).

crates/walrus-service/src/node.rs
jpcsmith: Suggestion: Remove this function. If the event processor needs a client to be constructed, that should be provided before constructing it and passing it as a dependency to the storage node.

crates/walrus-event/src/blob_writer.rs
gdanezis: Are you thinking here of each strorage node just storing locally the slivers without going through the usual registration and certification process, or requiring storage resources? @mlegner

mlegner: Yes, this is the plan. There will be on-chain state, but the mechanism will be slightly different. I believe the plan is to have a contract to "vote" for a new event blob and as soon as 2/3 of the storage nodes have called that with the same blob ID, that blob ID is simultaneously registered and certified. (Am I describing this correctly, @sadhansood?)

And every storage node can compute its own slivers locally, so there is no need to go through any public APIs to store metadata and slivers.

crates/walrus-service/src/node/config.rs
sadhansood: yeah, removed it from event processor config

crates/walrus-event/src/blob_writer.rs
sadhansood: Yes, you summarized it pretty well @mlegner. I realized that going through public apis would lead to failure if the blob is not pre-registered and hence we now directly invoke the node's apis to store metadata and slivers.

crates/walrus-service/src/node.rs
sadhansood: Now, we only have a single abstraction `EventManager` which both `SuiSystemEventProvider` and `EventProcessor` implement. But more or less, we are keeping everything as is and adding very little new things.

crates/walrus-service/bin/node.rs
```diff
@@ -233,6 +234,16 @@ impl StorageNodeRuntime {
...
+        let walrus_node_clone = walrus_node.clone();
+        let walrus_node_cancel_token = cancel_token.child_token();
+        let event_manager_handle = tokio::spawn(async move {
+            let cancel_token = walrus_node_cancel_token.clone();
+            let result = walrus_node_clone.start_event_manager(cancel_token).await;
+            if let Err(ref error) = result {
+                tracing::error!(?error, "event manager exited with an error");
+            }
+            result
+        });
```
jpcsmith: Suggestion: If this is running as part of the storage node, it should be a dependent service, managed by the storage node. Do not start it here.

crates/walrus-service/src/node/system_events.rs
```diff
@@ -49,19 +48,113 @@ impl SuiSystemEventProvider {
...
+        cursor: EventStreamCursor,
+    ) -> Result<Box<dyn Stream<Item = IndexedStreamElement> + Send + Sync + 'life0>, anyhow::Error>;
+}
+
+/// A manager for event retention. This is used to drop events that are no longer needed.
+#[async_trait]
+pub trait EventRetentionManager: std::fmt::Debug + Sync + Send {
+    /// Remove events before the specified cursor.
+    async fn drop_events_before(&self, cursor: EventStreamCursor) -> Result<(), anyhow::Error>;
+}
```
jpcsmith, line 69: Question: How would calls to this work if the event manager is a service being shared across multiple instances of a storage node?

crates/walrus-service/src/node/system_events.rs
```diff
@@ -49,19 +48,113 @@ impl SuiSystemEventProvider {
...
+            while let Ok(events) = cloned_self.poll(element_index).await {
+                for event in events {
+                    tx_event.send(event).await?;
+                    element_index += 1;
+                }
+                sleep(cloned_self.event_polling_interval).await;
+            }
+            Err::<(), anyhow::Error>(anyhow!("Failed to poll events"))
+        });
+        Ok(Box::new(ReceiverStream::new(rx_event)))
```
jpcsmith: Suggestion: Rewrite using `unfold` instead of spawning a task. If you're spawning a task, then you will need to ensure that you can cancel the task during shutdown.

```suggestion
        let mut interval = tokio::time::interval(self.event_polling_interval);
        interval.set_missed_tick_behavior(MissedTickBehavior::Skip);

        let event_stream = stream::unfold(
            (interval, cursor.element_index),
            move |(mut interval, element_index)| async move {
                interval.tick().await;

                let events = self
                    .poll(element_index)
                    .await
                    .inspect_err(|error| tracing::error!(?error, "failed to poll event stream"))
                    .ok()?;

                // Update the index such that the next future continues the sequence.
                let n_events = u64::try_from(events.len()).expect("number of events is within u64");
                Some((stream::iter(events), (interval, element_index + n_events)))
            },
        )
        .flatten();

        Ok(Box::new(event_stream))
```

crates/walrus-service/src/node/system_events.rs
```diff
@@ -49,19 +48,113 @@ impl SuiSystemEventProvider {
...
+        tokio::spawn(async move {
+            let mut element_index = cursor.element_index;
+            while let Ok(events) = cloned_self.poll(element_index).await {
+                for event in events {
+                    tx_event.send(event).await?;
+                    element_index += 1;
+                }
+                sleep(cloned_self.event_polling_interval).await;
+            }
+            Err::<(), anyhow::Error>(anyhow!("Failed to poll events"))
```
jpcsmith: Remark: This would need to be logged. Since the result of the task is never retrieved, this error is simply discarded. However, see below for an alternative that does not spawn a task.

crates/walrus-service/src/node/storage/event_blob_writer.rs
jpcsmith, line 1: Suggestion: Keep this logic external to the storage node.

Remark: One of the tasks moving from testnet to mainnet will be to distribute a single storage node across multiple machines within the data centre. This will likely take the form of them sharing the `EventProcessor`. By moving this file into the storage node, you couple it's implementation to the private API of the storage node and make such a transition harder. It would therefore be beneficial to continue to use the public API of the storage node for storing these blobs. This would then compose naturally with load balances and a distributed storage node. 

If/when performance of storing the events becomes a concern, we can explore alternatives for for single-host deployments (we will likely be adding a `Client` that has the exact same API but delegates to functions as opposed to HTTP).

crates/walrus-service/bin/node.rs
sadhansood: done

crates/walrus-service/src/node/system_events.rs
sadhansood: The same way it is today, with the difference of `drop_events_before` implemented as an rest call to event manager from storage node (instead of being implemented as updating in memory state today)

crates/walrus-service/src/node/system_events.rs
sadhansood: great suggestion, done and thank you

crates/walrus-service/src/node/system_events.rs
sadhansood: done

crates/walrus-service/src/node/storage/event_blob_writer.rs
sadhansood: `EventBlobWriter` needs access to storage node's internal api to save blob metadata and slivers without doing some error checking which is not possible with public api (which btw is what i was using before until i realized this). This is because we don't register event blobs but directly store slivers in local shards and invoke `certify_event_blob` smart contract endpoint. We most certainly will need to fix this if one event processor is going to get shared across multiple nodes. But it is something we can do later once we get close to that point

crates/walrus-service/src/node/system_events.rs
jpcsmith: And how does it know when "all" the instances have called it? My feeling is that we know we will need a single interface to a storage node, clustered or other, and so it would be better to have all sharding hidden behind that. So any information that needs to flow from one or more storage node instances should just go through that interface.

crates/walrus-service/src/node/storage/event_blob_writer.rs
jpcsmith: I don't see why this necessitates using the private API of the type. If we provide a `register_event_blob` call which, once triggered by enough nodes, emits an event for the registration of the blob, then we could store it through the normal APIs. Similarly, sufficient `certify_event_blob` calls would trigger a certification event for the blob.

crates/walrus-service/src/test_utils.rs
```diff
@@ -561,10 +600,16 @@ fn try_unused_socket_address() -> anyhow::Result<SocketAddr> {
...
     async fn events(
         &self,
-        _cursor: Option<EventID>,
-    ) -> Result<Box<dyn Stream<Item = BlobEvent> + Send + Sync + 'life0>, anyhow::Error> {
+        _cursor: EventStreamCursor,
+    ) -> Result<Box<dyn Stream<Item = IndexedStreamElement> + Send + Sync + 'life0>, anyhow::Error>
+    {
         Ok(Box::new(
-            tokio_stream::iter(self.clone()).chain(tokio_stream::pending()),
+            tokio_stream::iter(
```
jpcsmith, line 610: Question: Does this work? Elsewhere we use `futures::stream`

crates/walrus-service/src/node/system_events.rs
sadhansood: If you look into the implementation of this call, it just updates the index upto which it is safe to prune events. Event manager will know and keep track of that index per sharded storage node, and hence the min of that index is what it will prune events upto.

crates/walrus-service/src/node/storage/event_blob_writer.rs
sadhansood: Yes, the plan is to invoke `certify_event_blob` by enough participant to emit the blob certification event but we decided to skip on registering the event blob. However your point is valid, that we could also register an event blob by directly invoking a smart contract endpoint which would help utilize the public endpoints. I will look into this more. Thanks!

crates/walrus-service/src/test_utils.rs
sadhansood: Yes, i think it is being invoked in tests and it works

PR #746, opened 2024-08-29
feat(storage node): shard sync recovery path using service committee

halfprice:
Alternative implementation of #734 , in which the sliver recovery directly uses service committee instead of using BlobSyncHandler.

- During shard sync, store missing blobs in a `pending_recover_blobs` table in shard.
- After shard sync, using blob recover to recover blobs in `pending_recover_blobs`.
- Non happy path currently do not preempt happy path currently.

Contributes to #259 #560 #705

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1497,10 +1497,11 @@ mod tests {
         Ok((cluster, events, details))
     }
 
-    // Creates a test cluster with custom initial epoch and blobs that are partially stored
-    // in shard 0.
-    // The function is created for testing shard syncing/recovery. So for blobs that are
-    // not stored in shard 0, it also won't receive a certified event.
+    /// Creates a test cluster with custom initial epoch and blobs that are partially stored
+    /// in shard 0.
+    ///
+    /// The function is created for testing shard syncing/recovery. So for blobs that are
+    /// not stored in shard 0, it also won't receive a certified event.
     async fn cluster_with_partially_stored_blobs_in_shard_0<'a, F>(
         assignment: &[&[u16]],
         blobs: &[&'a [u8]],
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -231,18 +231,20 @@ pub struct Storage {
 }
 
 /// A iterator over the blob info table.
-pub struct BlobInfoIter<'a> {
-    iter: Box<dyn Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send + 'a>,
+pub struct BlobInfoIter<I: ?Sized>
+where
+    I: Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send,
+{
+    iter: Box<I>,
     before_epoch: Epoch,
     only_certified: bool,
 }
 
-impl<'a> BlobInfoIter<'a> {
-    pub fn new(
-        iter: Box<dyn Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send + 'a>,
-        before_epoch: Epoch,
-        only_certified: bool,
-    ) -> Self {
+impl<I: ?Sized> BlobInfoIter<I>
+where
+    I: Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send,
+{
+    pub fn new(iter: Box<I>, before_epoch: Epoch, only_certified: bool) -> Self {
         Self {
             iter,
             before_epoch,
@@ -251,7 +253,10 @@ impl<'a> BlobInfoIter<'a> {
     }
 }
 
-impl Debug for BlobInfoIter<'_> {
+impl<I: ?Sized> Debug for BlobInfoIter<I>
+where
+    I: Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send,
+{
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("BlobInfoIter")
             .field("before_epoch", &self.before_epoch)
@@ -260,7 +265,10 @@ impl Debug for BlobInfoIter<'_> {
     }
 }
 
-impl Iterator for BlobInfoIter<'_> {
+impl<I: ?Sized> Iterator for BlobInfoIter<I>
+where
+    I: Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send,
+{
     type Item = Result<(BlobId, BlobInfo), TypedStoreError>;
 
     fn next(&mut self) -> Option<Self::Item> {
@@ -283,6 +291,9 @@ impl Iterator for BlobInfoIter<'_> {
         item
     }
 }
+
+pub type BlobInfoIterator<'a> =
+    BlobInfoIter<dyn Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send + 'a>;
 
 impl Storage {
     const METADATA_COLUMN_FAMILY_NAME: &'static str = "metadata";
@@ -422,7 +433,7 @@ impl Storage {
         &self,
         before_epoch: Epoch,
         last_synced_blob_id: Option<BlobId>,
-    ) -> BlobInfoIter {
+    ) -> BlobInfoIterator {
         BlobInfoIter::new(
             Box::new(match last_synced_blob_id {
                 Some(starting_blob_id) => self
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -40,7 +40,7 @@ use walrus_core::{
     SliverType,
 };
 
-use super::{blob_info::BlobInfo, BlobInfoIter, DatabaseConfig};
+use super::{blob_info::BlobInfo, BlobInfoIterator, DatabaseConfig};
 use crate::node::{blob_sync::recover_sliver, errors::SyncShardClientError, StorageNodeInner};
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
@@ -142,7 +142,7 @@ impl ShardStorage {
     ) -> Result<Self, TypedStoreError> {
         let rw_options = ReadWriteOptions::default();
 
-        // TODO: cleanup all the column faimily initiation as they are highly repetitive.
+        // TODO(#766): cleanup all the column family initiation as they are highly repetitive
         let shard_cf_options = Self::slivers_column_family_options(id, db_config);
         for (_, (cf_name, options)) in shard_cf_options.iter() {
             if database.cf_handle(cf_name).is_none() {
@@ -502,7 +502,7 @@ impl ShardStorage {
                     sliver_type,
                 })) => {
                     tracing::info!(
-                        "Resuming shard sync from blob id: {}, sliver type: {}",
+                        "resuming shard sync from blob id: {}, sliver type: {}",
                         last_synced_blob_id,
                         sliver_type
                     );
@@ -548,7 +548,7 @@ impl ShardStorage {
         let mut next_blob_info = blob_info_iter.next().transpose()?;
         while let Some((next_starting_blob_id, _)) = next_blob_info {
             tracing::debug!(
-                "Syncing shard to before epoch: {}. Starting blob id: {}",
+                "syncing shard to before epoch: {}. Starting blob id: {}",
                 epoch,
                 next_starting_blob_id,
             );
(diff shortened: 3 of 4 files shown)


crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -475,71 +568,275 @@ impl ShardStorage {
...
+            tracing::warn!("Recovering missing blobs still misses blobs. Retrying in 60 seconds.",);
+            tokio::time::sleep(Duration::from_secs(60)).await;
+        }
+
+        Ok(())
+    }
+
+    /// Recovers missing blob slivers stored in `pending_recover_slivers` table, using
+    /// `blob_sync_handler`.
+    async fn recover_missing_blobs(
```
halfprice, line 739: Only the code path from this function differ from #734 .

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -475,71 +568,275 @@ impl ShardStorage {
...
+                    sliver_id,
+                    &node.encoding_config,
+                )
+                .await
+            }
+        };
+
+        match result {
+            Ok(sliver) => self.put_sliver(&blob_id, &sliver)?,
+            Err(inconsistency_proof) => {
```
halfprice, line 811: Say if a blob is certified, but wrong encoded sliver was lost in the old storage for some reason, it is the new storage node's responsibility to submit inconsistency proof, is that correct?

Also, besides in handling blob sync, is there any other way to discover inconsistency? Just want to get an understanding of what's the consequence if we miss submit invalid blob here.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -740,55 +739,102 @@ impl ShardStorage {
...
-            )?;
-            batch.write()?;
+
+        let metadata = node.storage.get_metadata(&blob_id)?;
+        if metadata.is_none() {
+            tracing::warn!(
+                "Blob {} is missing in the metadata table. For certified blob, Blob sync task should
+                recover the metadata. Skip recovering it for now.",
+                blob_id
+            );
```
mlegner: Question: How would we handle this for a completely new node that has to obtain the metadata for all blobs?

crates/walrus-service/src/node/storage/shard.rs
mlegner: 1. If something got lost in the old node, the new node needs to take care of it anyway, so I'd say yes.
2. Currently, yes. There are some other ways to detect (certain types of) inconsistencies, but those are a bit less relevant and currently not implemented.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -740,55 +739,102 @@ impl ShardStorage {
...
+
+        let mut batch = self.pending_recover_slivers.batch();
+        batch.delete_batch(
+            &self.pending_recover_slivers,
+            [
+                (SliverType::Primary, blob_id),
+                (SliverType::Secondary, blob_id),
+            ],
+        )?;
+        batch.write()
```
mlegner: Important: We should only delete the type of sliver we have actually recovered.

crates/walrus-service/src/node/storage/shard.rs
halfprice: > 2. Currently, yes.

Do you mean `currently, no`?

crates/walrus-service/src/node/storage/shard.rs
halfprice: As discussed in the meeting, during event catch up, the node needs to sync metadata.

When epoch change starts, the new node should have all the previously certified blobs' metadata in the node.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Good catch!

crates/walrus-service/src/node/storage/shard.rs
mlegner: > > 2. Currently, yes.
> 
> Do you mean `currently, no`?

Yes... 😅

halfprice: Thanks for the info!

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -116,6 +142,7 @@ impl ShardStorage {
     ) -> Result<Self, TypedStoreError> {
         let rw_options = ReadWriteOptions::default();
 
+        // TODO: cleanup all the column faimily initiation as they are highly repetitive.
```
mlegner: I created an issue for this:
```suggestion
        // TODO: cleanup all the column family initiation as they are highly repetitive (#766).
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -475,71 +568,267 @@ impl ShardStorage {
...
+            next_blob_info = blob_info_iter.next().transpose()?;
+        }
+        batch.write()?;
+
+        Ok(())
+    }
+
+    // Helper function to add fetched slivers to the db batch and check for missing blobs.
+    // Advance blob_info_iter to the next blob that is greater than the last fetched blob id,
+    // which is the next expected blob to fetch, and return the next expected blob.
```
mlegner: Nit: Can we make this a docstring?

```suggestion
    /// Helper function to add fetched slivers to the DB batch and check for missing blobs.
    ///
    /// Advances `blob_info_iter` to the next blob that is greater than the last fetched blob ID,
    /// which is the next expected blob to fetch, and returns the next expected blob.
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -475,71 +568,267 @@ impl ShardStorage {
...
             }
+            tracing::warn!("Recovering missing blobs still misses blobs. Retrying in 60 seconds.",);
+            tokio::time::sleep(Duration::from_secs(60)).await;
+        }
+
+        Ok(())
+    }
+
+    /// Recovers missing blob slivers stored in `pending_recover_slivers` table, using
+    /// `blob_sync_handler`.
```
mlegner: Nit: This is no longer correct:
```suggestion
    /// Recovers missing blob slivers stored in the `pending_recover_slivers` table.
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -475,71 +568,267 @@ impl ShardStorage {
...
+
+        loop {
+            self.recover_missing_blobs(node.clone()).await?;
+
+            if self.pending_recover_slivers.is_empty() {
+                // TODO: in test, check that we have recovered all the certified blobs.
                 break;
             }
+            tracing::warn!("Recovering missing blobs still misses blobs. Retrying in 60 seconds.",);
+            tokio::time::sleep(Duration::from_secs(60)).await;
```
mlegner: Question: Can this actually happen in practice?

I think it's good to have this, just in case, but if it's not expected to be ever triggered, we could add a comment about this.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -475,71 +568,267 @@ impl ShardStorage {
...
+        Ok(())
+    }
+
+    /// Recovers missing blob slivers stored in `pending_recover_slivers` table, using
+    /// `blob_sync_handler`.
+    async fn recover_missing_blobs(
+        &self,
+        node: Arc<StorageNodeInner>,
+    ) -> Result<(), TypedStoreError> {
+        let semaphore = Semaphore::new(5); // TODO: make this configurable.
```
mlegner: Suggestion: Link this to an existing or new issue.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -475,71 +568,267 @@ impl ShardStorage {
...
+                )
+                .await
+            }
+        };
+
+        match result {
+            Ok(sliver) => self.put_sliver(&blob_id, &sliver)?,
+            Err(inconsistency_proof) => {
+                // TODO: once committee service supports multi-epoch. This needs to use the
+                // committee from the latest epoch.
```
mlegner: Suggestion: Link this to an existing or new issue.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -475,71 +568,267 @@ impl ShardStorage {
...
+        let _guard = semaphore.acquire().await;
+        tracing::info!(
+            "Start recovering missing blob {} in shard {}",
+            blob_id,
+            self.id
+        );
+
+        let metadata = node.storage.get_metadata(&blob_id)?;
+        if metadata.is_none() {
+            tracing::warn!(
```
mlegner, line 776: Suggestion: As this is expected to never happen, can we make this an error?
```suggestion
            tracing::error!(
```

Also, nit: Normally we start log and error messages with lowercase letters. This applies to many messages in this PR.

crates/walrus-service/src/node/storage.rs
```diff
@@ -223,6 +230,60 @@ pub struct Storage {
     config: DatabaseConfig,
 }
 
+/// A iterator over the blob info table.
+pub struct BlobInfoIter<'a> {
+    iter: Box<dyn Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send + 'a>,
```
mlegner: Copying my question/suggestion from [xxx](https://github.com/MystenLabs/walrus/pull/734#discussion_r1738336951) regarding the trait object here:

Couldn't we make the `BlobInfoIter` generic over an inner iterator implementing `Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>`? Something like this:

```rust
/// A iterator over the blob info table.
pub struct BlobInfoIter<I> {
    iter: I,
    before_epoch: Epoch,
    only_certified: bool,
}

impl<I> BlobInfoIter<I>
where
    I: Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>>,
{
    pub fn new(iter: I, before_epoch: Epoch, only_certified: bool) -> Self {
        Self {
            iter,
            before_epoch,
            only_certified,
        }
    }
}
```

crates/walrus-service/src/node.rs
```diff
@@ -1500,6 +1497,69 @@ mod tests {
         Ok((cluster, events, details))
     }
 
+    // Creates a test cluster with custom initial epoch and blobs that are partially stored
+    // in shard 0.
+    // The function is created for testing shard syncing/recovery. So for blobs that are
+    // not stored in shard 0, it also won't receive a certified event.
```
mlegner: Nit: Can we make this a docstring?
```suggestion
    /// Creates a test cluster with custom initial epoch and blobs that are partially stored
    /// in shard 0.
    ///
    /// The function is created for testing shard syncing/recovery. So for blobs that are
    /// not stored in shard 0, it also won't receive a certified event.
```

crates/walrus-service/src/node/storage/shard.rs
halfprice: > Can this actually happen in practice?

I would say yes. One example is network issue where the node can't talk to other nodes.

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs
halfprice: I updated all the logging to start with lowercase letter.

For this log level, I still think it should be a warning because it's possible that blob sync is in the process of recovering the blob metadata, so it is not necessarily an error situatiion.

crates/walrus-service/src/node/storage.rs
halfprice: Sorry for missing the reply in the old PR! Suggestion taken and updated.

I did switched to the suggestion in the last revision, but one thing I found was that I have to use `BlobInfoIter<Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>>>` everywhere instead of just `BlobInfoIter`, and since we only meant to use it by taking `Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>>` only, it maybe a bit confusing that BlobInfoIter is not meant to take any generic iterator.

Looking at this again today, I did make this work by creating a type that takes a specific iterator

`pub type BlobInfoIterator<'a> =
    BlobInfoIter<dyn Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send + 'a>;`

and use it everywhere else in the code. I think it's cleaner now.

crates/walrus-service/src/node.rs

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs

PR #750, opened 2024-08-30
feat(staking): Adds setting for first epoch duration

damirka:
Note: not a fan of current testing base; perhaps, this is something I could take while you're away @karlwuest. E2E tests are doomed to be large, and we need to be smart about it.

### DIFF
--- a/contracts/walrus/tests/e2e_tests.move
+++ b/contracts/walrus/tests/e2e_tests.move
@@ -86,12 +86,6 @@ fun test_init_and_first_epoch_change(): (staking::Staking, system::System) {
     scenario.next_tx(nodes[0].sui_address());
     {
         staking.voting_end(&clock);
-    };
-
-    // === advance clock and change epoch ===
-    clock.increment_for_testing(EPOCH_ZERO_DURATION);
-    scenario.next_tx(nodes[0].sui_address());
-    {
         staking.initiate_epoch_change(&mut system, &clock);
     };
 


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -369,8 +378,12 @@ public(package) fun initiate_epoch_change(
         EpochState::NextParamsSelected(last_epoch_change) => last_epoch_change,
         _ => abort EWrongEpochState,
     };
+
     let now = clock.timestamp_ms();
-    assert!(now >= last_epoch_change + EPOCH_DURATION, EWrongEpochState);
+
+    if (self.epoch == 0) assert!(now >= self.first_epoch_start, EWrongEpochState)
```
karlwuest, line 391: Important: We also need to add a similar case distinction to `voting_end`. I think for starting the first epoch, we don't need a gap between when the committee is decided and when epoch change happens (since there are no shards to transfer yet), so we can just do the same check there if we're in epoch 0.

damirka: Just pushed. Voting end can be performed at the same timestamp as the epoch advancement.

contracts/walrus/tests/e2e_tests.move
```diff
@@ -82,14 +82,14 @@ fun test_init_and_first_epoch_change(): (staking::Staking, system::System) {
...
-    clock.increment_for_testing(PARAM_SELECTION_DELTA);
+    clock.increment_for_testing(EPOCH_ZERO_DURATION);
     scenario.next_tx(nodes[0].sui_address());
     {
         staking.voting_end(&clock);
     };
 
     // === advance clock and change epoch ===
-    clock.increment_for_testing(FIRST_EPOCH_START);
+    clock.increment_for_testing(EPOCH_ZERO_DURATION);
```
karlwuest: Nit: we can remove this increment, since the clock should have been sufficiently incremented above.

damirka: Nice catch, optimized!

Issue #496, opened 2024-06-07
Expose logs and metrics in the Walrus daemon


giac-mysten:
- [x] export metrics and logs for the daemon -> #751 
- [x] split the runtimes in the client binary for daemon and metrics -> #767

PR #751, opened 2024-08-30
feat(client): export HTTP request metrics and logs

giac-mysten:
Contributes to #496

### DIFF
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -18,6 +18,7 @@ use clap::{Args, Parser, Subcommand};
 use prometheus::Registry;
 use serde::Deserialize;
 use serde_with::{serde_as, DisplayFromStr};
+use telemetry_subscribers::{TelemetryGuards, TracingHandle};
 use tracing_subscriber::{layer::SubscriberExt as _, util::SubscriberInitExt, EnvFilter, Layer};
 use walrus_core::{
     encoding::{encoded_blob_length_for_n_shards, EncodingConfig, Primary},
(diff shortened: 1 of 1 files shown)


crates/walrus-service/bin/client.rs
```diff
@@ -309,9 +310,13 @@ struct RpcArg {
 #[serde(rename_all = "camelCase")]
 struct DaemonArgs {
     /// The address to which to bind the service.
-    #[clap(short, long, default_value_t = default::bind_address())]
+    #[clap(long, default_value_t = default::bind_address())]
```
mlegner: Important: In the docs we currently use the short `-b` in a few places. These need to be updated if we remove the short option.

giac-mysten: thanks for the catch @mlegner . Let me see what the best option is.

giac-mysten: I changed the `short` for `--metrics-address` to `-a`, and reintroduced `-b` for `--bind-address`.

crates/walrus-service/bin/client.rs
```diff
@@ -467,6 +478,19 @@ fn init_tracing_subscriber() -> Result<()> {
     Ok(())
 }
 
+// Starts the Prometheus server and returns the registry.
+fn run_registry_and_telemetry(mut metrics_address: SocketAddr) -> Registry {
+    metrics_address.set_ip(IpAddr::V4(Ipv4Addr::UNSPECIFIED));
+    let registry_service = mysten_metrics::start_prometheus_server(metrics_address);
+    let registry = registry_service.default_registry();
+    let (_telemetry_guards, _tracing_handle) = telemetry_subscribers::TelemetryConfig::new()
```
jpcsmith: Important: The guards cannot be dropped, they must be kept alive.

giac-mysten: whops. did not consider that, sorry. should be fixed here: 4f8ed461fb2b34549b276a0c49b165bfc3e2d402

PR #763, opened 2024-09-02
refactor: deserialize move structs and events from bcs

### LINKED ISSUE #522: Convert move structs and events from BCS instead of reading the json values
Old context: [here](https://github.com/MystenLabs/walrus/pull/519#discussion_r1636211508). However, converting the BCS encoding of move structs should be much simpler than what we are currently doing.

karlwuest:
closes #522 

This is mainly in preparation for the reintegration of the new contracts.

### DIFF
--- a/crates/walrus-sui/src/contracts.rs
+++ b/crates/walrus-sui/src/contracts.rs
@@ -4,7 +4,9 @@
 //! Walrus contract bindings. Provides an interface for looking up contract function,
 //! modules, and type names.
 
-use anyhow::{anyhow, Context, Result};
+use core::fmt;
+
+use anyhow::{Context, Result};
 use move_core_types::{identifier::Identifier, language_storage::StructTag as MoveStructTag};
 use serde::de::DeserializeOwned;
 use sui_sdk::{
@@ -12,8 +14,31 @@ use sui_sdk::{
     types::base_types::ObjectID,
 };
 use sui_types::TypeTag;
+use thiserror::Error;
 use tracing::instrument;
 use walrus_core::ensure;
+
+/// Error returned when converting a Sui object or event to a rust struct.
+#[derive(Debug, Error)]
+pub enum MoveConversionError {
+    #[error("object data does not contain bcs")]
+    /// Error if the object data we are trying to convert does not contain bcs.
+    NoBcs,
+    #[error("not a move object")]
+    /// Error if the object data is not a move object.
+    NotMoveObject,
+    /// Error resulting if the object or event does not have the expected type.
+    #[error("the move struct {actual} does not match the expected type {expected}")]
+    TypeMismatch {
+        /// Expected type of the struct.
+        expected: String,
+        /// Actual type of the struct.
+        actual: String,
+    },
+    #[error(transparent)]
+    /// Error during BCS deserialization.
+    Bcs(#[from] bcs::Error),
+}
 
 /// A trait for types that correspond to a contract type.
 ///
@@ -25,7 +50,7 @@ pub trait AssociatedContractStruct: DeserializeOwned {
 
     /// Converts a [`SuiObjectData`] to [`Self`].
     #[instrument(err, skip_all)]
-    fn try_from_object_data(sui_object_data: &SuiObjectData) -> Result<Self, anyhow::Error> {
+    fn try_from_object_data(sui_object_data: &SuiObjectData) -> Result<Self, MoveConversionError> {
         tracing::debug!(
             "converting move object to rust struct {:?}",
             Self::CONTRACT_STRUCT,
@@ -33,17 +58,19 @@ pub trait AssociatedContractStruct: DeserializeOwned {
         let raw = sui_object_data
             .bcs
             .as_ref()
-            .ok_or_else(|| anyhow!("object data does not contain bcs"))?;
+            .ok_or(MoveConversionError::NoBcs)?;
         let raw = raw
             .try_as_move()
-            .ok_or_else(|| anyhow!("object data is a package, not a move object"))?;
+            .ok_or(MoveConversionError::NotMoveObject)?;
         ensure!(
             raw.type_.name.as_str() == Self::CONTRACT_STRUCT.name
                 && raw.type_.module.as_str() == Self::CONTRACT_STRUCT.module,
-            "object is not of type {:?}",
-            Self::CONTRACT_STRUCT
+            MoveConversionError::TypeMismatch {
+                expected: Self::CONTRACT_STRUCT.to_string(),
+                actual: format!("{}::{}", raw.type_.module.as_str(), raw.type_.name.as_str()),
+            }
         );
-        raw.deserialize()
+        Ok(bcs::from_bytes(&raw.bcs_bytes)?)
     }
 }
 
@@ -114,6 +141,12 @@ impl<'a> From<&'a MoveStructTag> for StructTag<'a> {
             name: value.name.as_str(),
             module: value.module.as_str(),
         }
+    }
+}
+
+impl<'a> fmt::Display for StructTag<'a> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}::{}", self.module, self.name)
     }
 }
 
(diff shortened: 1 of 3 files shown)


crates/walrus-sui/src/client/read_client.rs
```diff
@@ -270,7 +270,11 @@ impl ReadClient for SuiReadClient {
     }
 
     async fn current_committee(&self) -> SuiClientResult<Committee> {
-        Ok(self.get_system_object().await?.current_committee)
+        Ok(self
+            .get_system_object()
+            .await?
+            .current_committee
+            .ok_or_else(|| anyhow!("current committee is not set"))?)
```
giac-mysten, line 277: Question: if it is expected that the committee may not be set, why not return the `Option`?

crates/walrus-sui/src/types.rs
```diff
@@ -229,70 +136,58 @@ pub struct StorageNode {
     /// Name of the storage node.
     pub name: String,
     /// The network address of the storage node.
+    #[serde(deserialize_with = "deserialize_network_address")]
     pub network_address: NetworkAddress,
     /// The public key of the storage node.
+    #[serde(deserialize_with = "deserialize_bls_key")]
     pub public_key: PublicKey,
     /// The network key of the storage node.
+    #[serde(deserialize_with = "deserialize_network_key")]
```
giac-mysten: Question: here we only specify the deserialization. Is the default serialization working out of the box?

crates/walrus-sui/src/types.rs
```diff
@@ -95,7 +59,7 @@ pub struct Blob {
     /// The epoch in which the blob has been registered.
     pub stored_epoch: Epoch,
     /// The blob ID.
-    #[serde_as(as = "DisplayFromStr")]
+    #[serde(serialize_with = "serialize_blob_id")]
```
giac-mysten, line 69: Question: how is this deserialized from the integer representation?

crates/walrus-sui/src/utils.rs
```diff
@@ -181,8 +160,13 @@ pub(crate) fn get_sui_object_from_object_response<U>(
 where
     U: AssociatedContractStruct,
 {
-    let obj_struct = get_struct_from_object_response(object_response)?;
-    U::try_from(obj_struct).map_err(|_e| {
+    U::try_from_object_data(
+        object_response
+            .data
+            .as_ref()
+            .ok_or_else(|| anyhow!("Response does not contain object data"))?,
```
giac-mysten: Nit:

```suggestion
            .ok_or_else(|| anyhow!("response does not contain object data"))?,
```

crates/walrus-sui/src/utils.rs
```diff
@@ -503,28 +450,31 @@ where
...
-) -> Result<impl Iterator<Item = Result<SuiMoveStruct>> + 'a> {
+) -> Result<impl Iterator<Item = Result<SuiObjectData>> + 'a> {
     let struct_tag = object_type.to_move_struct_tag(package_id, type_args)?;
     Ok(handle_pagination(move |cursor| {
         sui_client.read_api().get_owned_objects(
             owner,
             Some(SuiObjectResponseQuery {
                 filter: Some(SuiObjectDataFilter::StructType(struct_tag.clone())),
-                options: Some(SuiObjectDataOptions::full_content()),
+                options: Some(SuiObjectDataOptions::full_content().with_bcs()),
```
giac-mysten: Nit: i guess we don't need the full content anymore?
```suggestion
                options: Some(SuiObjectDataOptions::new().with_bcs()),
```
or even
```suggestion
                options: Some(SuiObjectDataOptions::bcs_lossless()),
```

crates/walrus-sui/src/contracts.rs
```diff
@@ -4,21 +4,47 @@
...
+    #[instrument(err, skip_all)]
+    fn try_from_object_data(sui_object_data: &SuiObjectData) -> Result<Self, anyhow::Error> {
+        tracing::debug!(
+            "converting move object to rust struct {:?}",
+            Self::CONTRACT_STRUCT,
+        );
+        let raw = sui_object_data
+            .bcs
+            .as_ref()
+            .ok_or_else(|| anyhow!("object data does not contain bcs"))?;
```
mlegner: Question/suggestion: Could we convert all these anyhow errors to matchable thiserror errors? Not necessarily in this PR.

crates/walrus-sui/src/client/read_client.rs
karlwuest: It is expected that the committee is always set. In the new contracts it will not be an Option anymore. I didn't want to do an `expect` here, because the field is public, and I didn't refactor more of the code since it will anyway change with the integration of the new contracts.

crates/walrus-sui/src/types.rs
karlwuest: I don't think we need to serialize it anywhere. Let me check if I can just remove the derivation of `Serialize`. I would expect the serialization to have the same mismatch as the deserialization, namely in the move type, there are no constant size arrays, so the bcs always prepends the length, but in Rust, the keys use fixed size arrays.

crates/walrus-sui/src/types.rs
karlwuest: It's not. We only need to deserialize from bcs, which works directly with the derived implementation.

crates/walrus-sui/src/types.rs
```diff
@@ -229,70 +136,58 @@ pub struct StorageNode {
...
-impl TryFrom<SuiMoveStruct> for StorageNode {
-    type Error = anyhow::Error;
+#[instrument(err, skip_all)]
+fn deserialize_network_address<'de, D>(deserializer: D) -> Result<NetworkAddress, D::Error>
+where
+    D: Deserializer<'de>,
+{
+    let addr_string: String = Deserialize::deserialize(deserializer)?;
+    NetworkAddress::from_str(&addr_string).map_err(D::Error::custom)
+}
```
mlegner: Question: Why is this custom function necessary instead of a `#[serde_as(as = "DisplayFromStr")]` attribute?

crates/walrus-sui/src/types.rs
giac-mysten: Cool. If we don't serialize then let's remove the derivation, to avoid having the wrong derivation.

crates/walrus-sui/src/utils.rs
karlwuest: That's true, we do need the type as well, though. I'll change to `.with_bcs().with_type()`.

crates/walrus-sui/src/types.rs
giac-mysten: ah very nice!

crates/walrus-sui/src/utils.rs
giac-mysten: where do we need the type? I thought we were trusting that the full node will return the correct type only?

karlwuest: We check the type when we deserialize. I think that's important to avoid benign errors and to make it easier to find bugs if they occur.

crates/walrus-sui/src/types.rs
```diff
@@ -598,42 +452,39 @@ pub struct BlobRegistered {
...
-            size,
-            erasure_code_type,
-            end_epoch,
+            epoch: inner.epoch,
+            blob_id: inner.blob_id,
+            size: inner.size,
+            erasure_code_type: inner.erasure_code_type,
+            end_epoch: inner.end_epoch,
             event_id: sui_event.id,
         })
```
mlegner, line 490: Nit: It feels like this construction should be generalizable, maybe with a macro, so that it doesn't have to be duplicated for all the events.

crates/walrus-sui/src/types.rs
karlwuest: You're right, that should work!

crates/walrus-sui/src/types.rs
karlwuest: I agree that it would be nice to generalize this, but I'm not sure what the best way to do this is yet. 
One option would also be that we change from having an enum over all event types to an `Event` struct that contains the event id + event data, which could then be an enum for the different event types. Since the contract integration also adds additional events, I will consider this as part of that, or, if it can be separated nicely, as a follow up.

crates/walrus-sui/src/contracts.rs
karlwuest: Done.

Issue #717, opened 2024-08-22
Node: Implement per-object status tracking


mlegner:
The blob deletion feature makes it more complicated to keep track of which blobs need to be stored on storage nodes. It is no longer sufficient to keep track of the latest expiration epoch.

Consider a blob that is registered (and certified) multiple times with different `Blob` objects. At some point, one of the owners triggers a delete. The storage nodes need to determine if/when they can actually delete the data. As long as one certified and not expired `Blob` object exists, they need to retain the data, so they need to keep track of the number of live objects for each blob ID.

The solution is to keep track of `Blob` objects instead of just blob IDs. This means that for the same blob ID, we may keep track of multiple statuses.

(This is not a DoS attack vector as all `Blob` objects come with costs; in fact, creating multiple objects for the same blob ID reduces the resource requirements for the system compared to using them for different blob IDs as the actual data only has to be stored once.)

Depends on the new event structures added in #690.

Issue #718, opened 2024-08-22
Node: Delete blob data when blobs expire or are deleted


mlegner:
When deletable blobs are explicitly deleted, the data is already removed assuming there are no other related `Blob` objects. Data is also deleted if a blob is marked as invalid.

What is missing:

- [ ] Delete metadata and slivers on epoch change when blobs expire.
- [X] Clean up the blob info table on epoch change. --> [WAL-473](https://linear.app/mysten-labs/issue/WAL-473/update-aggregate-blob-info-on-epoch-change-based-on-per-object-status)
- [ ] At some point, delete the blob info for invalid blobs?

Expiry is currently handled based on the blob info: If there is no valid certified blob, slivers and metadata are not served, etc. There is an open issue to correctly track expiry of deletable blobs (#1005).

Open question: Should we have a "grace period" during which blobs are not accessible but data is kept and blobs can be extended?

PR #764, opened 2024-09-02
feat(node): more sophisticated blob status tracking

mlegner:
This PR prepares the status tracking of nodes for deletable blobs.

### Why is this required?

The current status tracking works (mostly) for permanent blobs that we have so far. However, the blob deletion feature makes it more complicated to keep track of which blobs need to be stored on storage nodes. 

The main difference between deletable and permanent blobs for status tracking is that deletable blobs are not deleted "in order" of their expiration epoch. So a deletable blob with later expiration epoch may be deleted before one with an earlier expiration epoch. This means that, for deletable blobs, it is not sufficient/possible to keep track of the latest expiration epoch: What should we do if that blob is deleted but there remain other objects for that blob ID?

### Goals and principles

- All public endpoints should require loading a constant amount of status-related data from the database. This implies that we need an aggregate blob status and cannot simply keep track of each `Blob` object separately as we would then need to iterate over all of those that correspond to a given blob ID to handle a request.
- Corollary of the previous: Updating the aggregate status should be possible without looking through all per-object data in our database.
- Our status should still contain as fine-grained information as possible. In particular, it should be possible to infer whether a blob ID has related *permanent* `Blob` objects and what their status and expiration is. This allows clients to not store blobs that are already permanently stored for a sufficient period.

### Design

The proposed design tracks permanent and deletable blobs in a different way:
- Permanent blobs are tracked similarly to the previous approach. There is a slight change to solve a bug with the previous approach when both a *registered* and *certified* blob exist for a blob ID and the *registered* expires after the *certified*.
- For deletable `Blob` objects, we only track the count of *registered* and *certified* objects.

### Scope

This does not yet add a table for per-object tracking (which is needed to update the status when blobs expire) but prepares the aggregate status to be used with deletable blobs.

### Open tasks

- [x] Rebase after #746 is merged
- [x] Clean up and add tests

Contributes to #717

### DIFF
--- a/crates/walrus-sdk/src/api.rs
+++ b/crates/walrus-sdk/src/api.rs
@@ -6,7 +6,7 @@
 use std::cmp::Ordering;
 
 use serde::{Deserialize, Serialize};
-use sui_types::{base_types::ObjectID, event::EventID};
+use sui_types::event::EventID;
 use tokio::time::Duration;
 use walrus_core::{Epoch, PublicKey};
 
@@ -59,10 +59,7 @@ pub enum BlobStatus {
         end_epoch: Epoch,
         /// Whether the blob is certified (true) or only registered (false).
         is_certified: bool,
-        /// The Sui object ID of a permanent `Blob` object with the above certification status and
-        /// `end_epoch`.
-        object_id: ObjectID,
-        /// The ID of the last Sui event related to the `Blob` object.
+        /// The ID of the Sui event that caused the status with the given `end_epoch`.
         status_event: EventID,
         /// Counts of deletable `Blob` objects.
         deletable_status: DeletableStatus,
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -8,7 +8,7 @@ use std::num::NonZeroU32;
 use enum_dispatch::enum_dispatch;
 use rocksdb::MergeOperands;
 use serde::{Deserialize, Serialize};
-use sui_types::{base_types::ObjectID, event::EventID};
+use sui_types::event::EventID;
 use tracing::Level;
 use walrus_core::Epoch;
 use walrus_sdk::api::{BlobStatus, DeletableStatus};
@@ -80,7 +80,6 @@ pub(super) enum BlobInfoMergeOperand {
 
 #[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
 pub(super) struct BlobStatusChangeInfo {
-    pub(super) object_id: ObjectID,
     pub(super) deletable: bool,
     pub(super) epoch: Epoch,
     pub(super) end_epoch: Epoch,
@@ -111,7 +110,6 @@ impl BlobInfoMergeOperand {
         Self::ChangeStatus {
             change_type,
             change_info: BlobStatusChangeInfo {
-                object_id: walrus_sui::test_utils::object_id_for_testing(),
                 deletable,
                 epoch,
                 end_epoch,
@@ -131,8 +129,7 @@ impl From<&BlobRegistered> for BlobInfoMergeOperand {
         } = value;
         Self::ChangeStatus {
             change_info: BlobStatusChangeInfo {
-                object_id: ObjectID::ZERO, // TODO(mlegner): update with new event structs (#762).
-                deletable: false,          // TODO(mlegner): update with new event structs (#762).
+                deletable: false, // TODO(mlegner): update with new event structs (#762).
                 epoch: *epoch,
                 end_epoch: *end_epoch,
                 status_event: *event_id,
@@ -152,8 +149,7 @@ impl From<&BlobCertified> for BlobInfoMergeOperand {
         } = value;
         Self::ChangeStatus {
             change_info: BlobStatusChangeInfo {
-                object_id: ObjectID::ZERO, // TODO(mlegner): update with new event structs (#762).
-                deletable: false,          // TODO(mlegner): update with new event structs (#762).
+                deletable: false, // TODO(mlegner): update with new event structs (#762).
                 epoch: *epoch,
                 end_epoch: *end_epoch,
                 status_event: *event_id,
@@ -228,16 +224,12 @@ impl ValidBlobInfoV1 {
             count_deletable_certified: *count_deletable_certified,
         };
         if let Some(PermanentBlobInfoV1 {
-            end_epoch,
-            object_id,
-            event,
-            ..
+            end_epoch, event, ..
         }) = permanent_certified.as_ref().or(permanent_total.as_ref())
         {
             BlobStatus::Permanent {
                 end_epoch: *end_epoch,
                 is_certified: permanent_certified.is_some(),
-                object_id: *object_id,
                 status_event: *event,
                 deletable_status,
             }
@@ -465,14 +457,7 @@ pub(crate) struct PermanentBlobInfoV1 {
     pub count: NonZeroU32,
     /// The latest expiration epoch among these objects.
     pub end_epoch: u64,
-    /// The ID of the object with the latest expiration epoch.
-    ///
-    /// If there are multiple, this contains the object with the highest status ('certified' >
-    /// 'registered') and the earliest `status_changing_epoch`.
-    pub object_id: ObjectID,
-    /// The epoch in which that object attained its current status.
-    pub status_changing_epoch: Epoch,
-    /// The ID of the latest blob event of that object.
+    /// The ID of the first blob event that led to the status with the given `end_epoch`.
     pub event: EventID,
 }
 
@@ -490,8 +475,6 @@ impl PermanentBlobInfoV1 {
             *self = PermanentBlobInfoV1 {
                 count: self.count,
                 end_epoch: change_info.end_epoch,
-                object_id: change_info.object_id,
-                status_changing_epoch: change_info.epoch,
                 event: change_info.status_event,
             };
         }
@@ -504,8 +487,7 @@ impl PermanentBlobInfoV1 {
     /// Panics if the change info has `deletable == true`.
     fn update_optional(existing_info: &mut Option<Self>, change_info: &BlobStatusChangeInfo) {
         let BlobStatusChangeInfo {
-            object_id: new_object_id,
-            epoch: new_epoch,
+            epoch: _,
             end_epoch: new_end_epoch,
             status_event: new_status_event,
             deletable,
@@ -517,8 +499,6 @@ impl PermanentBlobInfoV1 {
                 *existing_info = Some(PermanentBlobInfoV1 {
                     count: NonZeroU32::new(1).unwrap(),
                     end_epoch: *new_end_epoch,
-                    object_id: *new_object_id,
-                    status_changing_epoch: *new_epoch,
                     event: *new_status_event,
                 })
             }
@@ -527,14 +507,12 @@ impl PermanentBlobInfoV1 {
     }
 
     #[cfg(test)]
-    fn new_for_testing(count: u32, status_changing_epoch: Epoch, end_epoch: Epoch) -> Self {
-        use walrus_sui::test_utils::{fixed_event_id_for_testing, object_id_for_testing};
+    fn new_for_testing(count: u32, end_epoch: Epoch) -> Self {
+        use walrus_sui::test_utils::fixed_event_id_for_testing;
 
         Self {
             count: NonZeroU32::new(count).unwrap(),
             end_epoch,
-            status_changing_epoch,
-            object_id: object_id_for_testing(),
             event: fixed_event_id_for_testing(),
         }
     }
@@ -626,9 +604,8 @@ impl Mergeable for BlobInfoV1 {
             change_type: BlobStatusChangeType::Register,
             change_info:
                 BlobStatusChangeInfo {
-                    object_id,
                     deletable,
-                    epoch,
+                    epoch: _,
                     end_epoch,
                     status_event,
                 },
@@ -649,8 +626,6 @@ impl Mergeable for BlobInfoV1 {
                 Some(PermanentBlobInfoV1 {
                     count: NonZeroU32::new(1).unwrap(),
                     end_epoch,
-                    object_id,
-                    status_changing_epoch: epoch,
                     event: status_event,
                 }),
             )
@@ -720,7 +695,7 @@ impl BlobInfo {
         end_epoch: Epoch,
         status: BlobCertificationStatus,
         current_status_event: EventID,
-        registered_epoch: Option<Epoch>,
+        _registered_epoch: Option<Epoch>,
         certified_epoch: Option<Epoch>,
         invalidated_epoch: Option<Epoch>,
     ) -> Self {
@@ -734,16 +709,10 @@ impl BlobInfo {
                 let permanent_total = PermanentBlobInfoV1 {
                     count: NonZeroU32::new(1).unwrap(),
                     end_epoch,
-                    object_id: walrus_sui::test_utils::object_id_for_testing(),
-                    status_changing_epoch: registered_epoch
-                        .unwrap_or_else(|| certified_epoch.unwrap()),
                     event: current_status_event,
                 };
                 let permanent_certified = matches!(status, BlobCertificationStatus::Certified)
-                    .then(|| PermanentBlobInfoV1 {
-                        status_changing_epoch: certified_epoch.unwrap(),
-                        ..permanent_total
-                    });
+                    .then(|| permanent_total.clone());
                 ValidBlobInfoV1 {
                     permanent_total: Some(permanent_total),
                     permanent_certified,
@@ -939,12 +908,12 @@ mod tests {
                 ..Default::default()
             }),
             permanent: (ValidBlobInfoV1{
-                permanent_total: Some(PermanentBlobInfoV1::new_for_testing(2, 1, 3)),
+                permanent_total: Some(PermanentBlobInfoV1::new_for_testing(2, 3)),
                 ..Default::default()
             }),
             permanent_certified: (ValidBlobInfoV1{
-                permanent_total: Some(PermanentBlobInfoV1::new_for_testing(2, 0, 3)),
-                permanent_certified: Some(PermanentBlobInfoV1::new_for_testing(1, 1, 2)),
+                permanent_total: Some(PermanentBlobInfoV1::new_for_testing(2, 3)),
+                permanent_certified: Some(PermanentBlobInfoV1::new_for_testing(1,  2)),
                 initial_certified_epoch: Some(1),
                 ..Default::default()
             }),
@@ -977,12 +946,12 @@ mod tests {
                 ..Default::default()
             }),
             permanent: (ValidBlobInfoV1{
-                permanent_total: Some(PermanentBlobInfoV1::new_for_testing(2, 1, 3)),
+                permanent_total: Some(PermanentBlobInfoV1::new_for_testing(2,  3)),
                 ..Default::default()
             }),
             permanent_certified: (ValidBlobInfoV1{
-                permanent_total: Some(PermanentBlobInfoV1::new_for_testing(2, 0, 3)),
-                permanent_certified: Some(PermanentBlobInfoV1::new_for_testing(1, 1, 2)),
+                permanent_total: Some(PermanentBlobInfoV1::new_for_testing(2, 3)),
+                permanent_certified: Some(PermanentBlobInfoV1::new_for_testing(1, 2)),
                 initial_certified_epoch: Some(1),
                 ..Default::default()
             }),
@@ -1081,7 +1050,7 @@ mod tests {
                     BlobStatusChangeType::Register, false, 1, 2, fixed_event_id_for_testing()
                 ),
                 ValidBlobInfoV1{
-                    permanent_total: Some(PermanentBlobInfoV1::new_for_testing(1, 1, 2)),
+                    permanent_total: Some(PermanentBlobInfoV1::new_for_testing(1, 2)),
                     ..Default::default()
                 },
             ),
--- a/crates/walrus-service/src/node/server.rs
+++ b/crates/walrus-service/src/node/server.rs
@@ -404,7 +404,7 @@ mod tests {
         api::{BlobStatus, DeletableStatus, ServiceHealthInfo, StoredOnNodeStatus},
         client::{Client, ClientBuilder},
     };
-    use walrus_sui::test_utils::{event_id_for_testing, object_id_for_testing};
+    use walrus_sui::test_utils::event_id_for_testing;
     use walrus_test_utils::{async_param_test, Result as TestResult, WithTempDir};
 
     use super::*;
@@ -526,7 +526,6 @@ mod tests {
                     end_epoch: 3,
                     status_event: event_id_for_testing(),
                     is_certified: true,
-                    object_id: object_id_for_testing(),
                     deletable_status: DeletableStatus {
                         count_deletable_total: 0,
                         count_deletable_certified: 0,


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,722 @@
...
+            permanent_certified,
+            initial_certified_epoch,
+        } = &mut self
+        else {
+            return self;
+        };
+
+        if change_info.deletable {
+            match change_type {
+                BlobStatusChangeType::Register => *count_deletable_total += 1,
```
halfprice: Does deletable blob not need to track end epoch, etc.?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,722 @@
...
+                    Self::certify_permanent(permanent_total, permanent_certified, &change_info);
+                }
+                BlobStatusChangeType::Extend => {
+                    let Some(permanent_certified) = permanent_certified else {
+                        tracing::error!("attempt to extend a permanent blob when none is tracked");
+                        return self;
+                    };
+                    Self::extend_permanent(permanent_certified, &change_info);
+                }
+                BlobStatusChangeType::Delete { was_certified } => {
```
halfprice, line 287: Who sends this change? There is a BlobDeleted event but I think that is for deletable blobs.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,722 @@
...
+            }
+            _ => (),
+        }
+    }
 
-        if self.status < status || self.status == status && self.end_epoch < end_epoch {
-            self.status = status;
-            self.current_status_event = status_event;
-            self.end_epoch = end_epoch;
+    fn maybe_unset_initial_certified_epoch(&mut self) {
```
halfprice, line 319: Say a blob object is certified in epoch 1 and expires at epoch 5, another object mapping to the same blob is certified in epoch 4 and expires at epoch 10. At epoch 5, do we need to update initial_certified_epoch to 4 for this blob?

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: The problem is that this is not doable in the aggregate status because it's impossible to update on deletion. The plan is to *also* track each individual `Blob` object, but this will be added separately.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: The plan is to do this during the cleanup at the end of epochs. There is a separate issue for this: #718

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: IMO, no. My idea of the field is that it tracks the first time since which the blob was continuously certified in any way. This only gets unset when there is no longer *any* certified `Blob` object.

I can describe this in a bit more detail in the docstring.

crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -0,0 +1,176 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use rocksdb::{DBCompressionType, Options};
+use serde::{Deserialize, Serialize};
+use serde_with::skip_serializing_none;
```
halfprice, line 6:
```suggestion
use serde_with::skip_serializing_none;

```

halfprice: Also, just to confirm, the change in this code is a complete copy from storage.rs?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct BlobInfoV1Inner {
+    /// The total number of `Blob` objects for that blob ID with the given status.
+    pub count: NonZeroU32,
+    /// The latest expiration epoch among these objects.
+    pub end_epoch: u64,
+    /// The ID of the object with the latest expiration epoch.
+    ///
+    /// If there are multiple, this contains the object with the highest status ('certified' >
+    /// 'registered') and the earliest `status_changing_epoch`.
```
halfprice: > ...the highest status ('certified' > 'registered')...

Question: is this still needed? We are using this data structure to only track single status now.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+            return self;
+        };
+
+        if change_info.deletable {
+            match change_type {
+                BlobStatusChangeType::Register => *count_deletable_total += 1,
+                BlobStatusChangeType::Certify => {
+                    if count_deletable_total <= count_deletable_certified {
+                        tracing::error!(
+                            "attempt to certify a deletable blob before corresponding register"
```
halfprice, line 253: suggestion: can we log the blob id or object id in the message?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+        new_certified_epoch: Epoch,
     ) {
-        self.update_status_changing_epoch_to_earliest(blob_id, status, status_changing_epoch);
+        match current_initial_certified_epoch {
+            None => *current_initial_certified_epoch = Some(new_certified_epoch),
+            Some(current_certified_epoch) if *current_certified_epoch > new_certified_epoch => {
+                *current_initial_certified_epoch = Some(new_certified_epoch)
+            }
+            _ => (),
+        }
```
halfprice: Nit: what about

```
if current_initial_certified_epoch
    .map_or(true, |existing_epoch| existing_epoch > new_certified_epoch)
{
    *current_initial_certified_epoch = Some(new_certified_epoch);
}
```

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+    // This should not happen if the system works properly, but one never knows...
+    fn decrement_deletable_counter_on_deletion(counter: &mut u32) {
+        *counter = counter.checked_sub(1).unwrap_or_else(|| {
+            tracing::error!("attempt to delete blob when count was already 0");
+            0
+        });
+    }
+
+    /// Processes a register status change on the [`Option<BlobInfoV1Inner>`] object representing
+    /// all  permanent blobs.
```
halfprice:
```suggestion
    /// all permanent blobs.
```

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+        change_info: &BlobStatusChangeInfo,
+    ) {
+        let BlobStatusChangeInfo {
+            object_id: new_object_id,
+            epoch: new_epoch,
+            end_epoch: new_end_epoch,
+            status_event: new_status_event,
+            deletable,
+        } = change_info;
+        debug_assert!(!deletable);
```
halfprice: Nit: since checking this variable is not expensive, I would just use `assert`

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+    /// The total number of `Blob` objects for that blob ID with the given status.
+    pub count: NonZeroU32,
+    /// The latest expiration epoch among these objects.
+    pub end_epoch: u64,
+    /// The ID of the object with the latest expiration epoch.
+    ///
+    /// If there are multiple, this contains the object with the highest status ('certified' >
+    /// 'registered') and the earliest `status_changing_epoch`.
+    pub object_id: ObjectID,
+    /// The epoch in which that object attained its current status.
```
halfprice: Suggestion: can we add the information more explicit here that it is corresponding to the latest update to the `end_epoch` field? Every certified event updates the `count`, but not this field.

Also, after reading the PR, it seems this field is no longer needed. Maybe we can just remove it.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+    ) {
+        debug_assert!(!change_info.deletable);
+
+        blob_info_inner.count = blob_info_inner.count.saturating_add(1);
+        if change_info.end_epoch > blob_info_inner.end_epoch {
+            *blob_info_inner = BlobInfoV1Inner {
+                count: blob_info_inner.count,
+                end_epoch: change_info.end_epoch,
+                object_id: change_info.object_id,
+                status_changing_epoch: change_info.epoch,
```
halfprice: Question: based on the old behavior, this field is set to min(existing_value, chenge_info.epoch). Is the meaning of the field changed here?

Also, as mentioned above, we may not need it anymore.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+        } else {
+            match change_type {
+                BlobStatusChangeType::Register => {
+                    Self::register_permanent(permanent_total, &change_info);
+                }
+                BlobStatusChangeType::Certify => {
+                    let Some(permanent_total) = permanent_total else {
+                        tracing::error!("attempt to certify a permanent blob when none is tracked");
+                        return self;
+                    };
```
halfprice: Suggestion: move this check inside `certify_permanent` so that all the validity checks are in one place.

Another option would be to create a `validate_register_state`, and then directly call `update_optional_blob_info_inner(permanaent_certified, change_info)`.

crates/walrus-sdk/src/api.rs
```diff
@@ -109,27 +87,66 @@ impl PartialOrd for BlobStatus {
...
+                )
+                    .cmp(&(
+                        is_certified_other,
+                        end_epoch_other,
+                        count_deletable_certified_other,
+                        count_deletable_total_other,
+                    ))
             }
         }
     }
```
halfprice: Nice job coming up with this detailed ordering. However, as we can see, it becomes very complex (for the reader), so I wonder if there is any way we can avoid it.

Looking at the usage, it is only used [here](https://github.com/MystenLabs/walrus/blob/359c0ed45b366005fd54970b931c05bc37356128/crates/walrus-service/src/client.rs#L685-L696). Is there a way we can modify the callsite so that it doesn't rely on the ordering?

Also, I'm not sure if the new ordering may change the assumption of the usage. Looks like now even if two statuses both are valid, if they have different `count_deletable_total`, are they considered different status?

crates/walrus-service/src/client/cli_utils/cli_output.rs
```diff
@@ -151,18 +151,42 @@ impl CliOutput for BlobStatusOutput {
...
+                ..
+            } => {
+                let status = (if is_certified {
+                    "certified"
+                } else {
+                    "registered"
+                })
+                .bold();
+                println!(
+                    "There is a {status} permanent Blob object for  blob ID {blob_str}.\n\
```
halfprice:
```suggestion
                    "There is a {status} permanent Blob object for blob ID {blob_str}.\n\
```

crates/walrus-service/src/node/storage/database_config.rs
mlegner: > Also, just to confirm, the change in this code is a complete copy from storage.rs?

Yes, except for the added `pub(super)` visibility of struct fields.

crates/walrus-sdk/src/api.rs
```diff
@@ -88,16 +47,35 @@ pub enum BlobStatus {
...
+        count_deletable_certified: u32,
+    },
+    /// The blob exists within Walrus; but there is no related permanent object, so it may be
+    /// deleted at any time.
+    Deletable {
+        /// Total number of active deletable blobs for the given blob ID.
+        count_deletable_total: u32,
+        /// Number of certified deletable blobs for the given blob ID.
+        count_deletable_certified: u32,
     },
```
jpcsmith: Remark: Since these field are used here and in the `Permanent` status, perhaps a struct `DeletableStatus` could be used in both places. It would also allow defining its own ordering.

crates/walrus-service/src/client/cli_utils/cli_output.rs
```diff
@@ -151,18 +151,42 @@ impl CliOutput for BlobStatusOutput {
...
             } => println!(
-                "Status for blob ID {blob_str}: {}\n\
-                    End epoch: {}\n\
-                    Related event: {}",
-                status.to_string().bold(),
-                end_epoch,
-                format_event_id(&status_event),
+                "Blob ID {blob_str} is registered on Walrus but only as deletable Blob objects:\n\
+                    Total number of objects: {count_deletable_total}\n\
+                    Number of certified objects: {count_deletable_certified}"
```
jpcsmith: Suggestion: Could we reword this a bit to focus on what's stored on walrus? For example 
```raw
Total stored: X (of Y registered)
```

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+
+    /// Updates the (optionally) existing blob info with the provided merge operand and returns the
+    /// result.
+    ///
+    /// Returns the preexisting value if the merge fails. An error is logged in this case.
     #[must_use]
-    fn merge(self, operand: Self::MergeOperand) -> Self;
+    fn merge(existing_val: Option<Self>, operand: Self::MergeOperand) -> Option<Self> {
+        match existing_val {
+            Some(existing_val) => Some(Self::merge_preexisting(existing_val, operand)),
```
jpcsmith:
```suggestion
            Some(existing_val) => Some(existing_val.merge_preexisting(operand)),
```

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
 
-pub(crate) trait Mergeable {
+pub(crate) trait Mergeable: Sized {
     type MergeOperand;
 
+    /// Updates the existing blob info with the provided merge operand and returns the result.
+    ///
+    /// Returns the preexisting value if the merge fails. An error is logged in this case.
+    #[must_use]
+    fn merge_preexisting(self, operand: Self::MergeOperand) -> Self;
```
jpcsmith: Nit:
```suggestion
    fn merge_with(self, operand: Self::MergeOperand) -> Self;
```
Since `self.merge_preexisting(operand)` reads as if operand is what is pre-existing.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+    ///
+    /// Returns the preexisting value if the merge fails. An error is logged in this case.
     #[must_use]
-    fn merge(self, operand: Self::MergeOperand) -> Self;
+    fn merge(existing_val: Option<Self>, operand: Self::MergeOperand) -> Option<Self> {
+        match existing_val {
+            Some(existing_val) => Some(Self::merge_preexisting(existing_val, operand)),
+            None => Self::merge_new(operand),
+        }
+    }
```
jpcsmith: Praise: 😎

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+    fn initial_certified_epoch(&self) -> Option<Epoch>;
+    /// Returns the event through which this blob was marked invalid.
+    ///
+    /// Returns `None` if it isn't invalid.
+    fn invalidation_event(&self) -> Option<EventID>;
+
+    /// Converts the blob information to a `BlobStatus` object.
+    fn to_blob_status(&self) -> BlobStatus;
+
+    // Question(mlegner): Do we actually need these functions? Currently, they are unused.
```
jpcsmith: Remark: I'm in favour of removing things :), if someone needs it, they can add it back.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+    /// Returns true iff there is any corresponding registered or certified blob object.
+    #[allow(dead_code)]
+    fn is_registered(&self) -> bool;
 }
 
-impl PartialOrd for BlobCertificationStatus {
-    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
-        Some(self.cmp(other))
+#[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
+pub(crate) enum BlobInfoMergeOperand {
```
jpcsmith: Question: Does this (and subsequent structs) need to be crate visible, or can they be `pub(super)`?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+        permanent_total: Option<BlobInfoV1Inner>,
+        permanent_certified: Option<BlobInfoV1Inner>,
+        initial_certified_epoch: Option<Epoch>,
+    },
+}
 
-    /// Returns the certification status of the blob.
-    fn status(&self) -> BlobCertificationStatus;
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct BlobInfoV1Inner {
```
jpcsmith: Nit: Since this is specific to permanent blobs, perhaps `PermanentBlobInfoV1`? I was confused as to why a `BlobInfo` would have multiple inners.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+                        status_event: *event,
+                        count_deletable_total: *count_deletable_total,
+                        count_deletable_certified: *count_deletable_certified,
+                    }
+                } else {
+                    BlobStatus::Deletable {
+                        count_deletable_total: *count_deletable_total,
+                        count_deletable_certified: *count_deletable_certified,
+                    }
+                }
```
jpcsmith: Suggestion: Perhaps the logic for handling the permanent case could be simplified through some combination of 

```rust
let is_certified = permanent_certified.is_some();
if let Some(xxxx) = permanent_certified.or(permanent_total) { // may need some as_refs()s
```

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: You mean we should remove the object ID altogether? Sure, I can do that. However, it's also in the client-facing `walrus_sdk::api::BlobStatus`, so that would change as well.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: This should already be contained in the span through the `#[tracing::instrument]` on the function.

I'm not completely sure if/how this gets logged, however, as this is only called in the RocksDB merge operator. Do you know more about this (or @jpcsmith)?

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Nice, done.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Yes, this is changed: "The epoch in which that object attained its current status."

However, I'm open to removing this.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+    // INV: initial_certified_epoch.is_some()
+    //      <=> count_deletable_certified > 0 || permanent_certified.is_some()
+    Valid {
+        is_metadata_stored: bool,
+        count_deletable_total: u32,
+        count_deletable_certified: u32,
+        permanent_total: Option<BlobInfoV1Inner>,
+        permanent_certified: Option<BlobInfoV1Inner>,
+        initial_certified_epoch: Option<Epoch>,
+    },
```
jpcsmith: Nit: This variant has a lot of fields, and quite a bit of logic attached to it when updating. Perhaps refactoring to a `Valid(Valid)` struct would help simplify some of the code during the update.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
-    fn update_status_changing_epoch_to_earliest(
-        &mut self,
-        blob_id: &BlobId,
-        status: BlobCertificationStatus,
-        status_changing_epoch: Epoch,
+    /// Decrements a counter on blob deletion.
+    ///
+    /// Logs an error if the counter is 0.
+    // Question(mlegner): Do we actually have to handle the cases where those counts are already 0?
+    // This should not happen if the system works properly, but one never knows...
```
jpcsmith: Remark: I would be happy with `debug_assert` and a log in production. That would (hopefully) help us to spot any such errors as soon as they happen.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1,288 +1,753 @@
...
+            Some(blob_info_inner) => Self::update_blob_info_inner(blob_info_inner, change_info),
+        }
     }
 
-    fn is_metadata_stored(&self) -> bool {
-        self.is_metadata_stored
+    fn update_blob_info_inner(
+        blob_info_inner: &mut BlobInfoV1Inner,
+        change_info: &BlobStatusChangeInfo,
+    ) {
```
jpcsmith: Remark: This feels like it should be defined on BlobInfoV1Inner.

crates/walrus-sdk/src/api.rs
mlegner: > Looking at the usage, it is only used [here](https://github.com/MystenLabs/walrus/blob/359c0ed45b366005fd54970b931c05bc37356128/crates/walrus-service/src/client.rs#L685-L696). Is there a way we can modify the callsite so that it doesn't rely on the ordering?

I think not when we want to preserve exactly the current behavior. However, I could sort the statuses according to the "weight" of storage nodes that returned them. Only in very rare cases where a few storage nodes have already seen a new event that others haven't would this change the behavior.

> Looks like now even if two statuses both are valid, if they have different count_deletable_total, are they considered different status?

Yes, that is true.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Good point. I've reduced the visibility of all definitions as much as possible.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: I've moved the check into `certify_permanent` and also the one below into `extend_permanent`.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Good suggestion, this made a few things much nicer in a32395e52f5e74928278b6c5b7425db5949799d1

crates/walrus-sdk/src/api.rs
mlegner: This got a bit nicer after implementing JP's suggestion [here](https://github.com/MystenLabs/walrus/pull/764#discussion_r1746793626).

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Removed the object ID and status-changing epoch in 8b708dd02ffb6b28e7b0b41d85347373c93bd5a8.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Removed the object ID and status-changing epoch in 8b708dd02ffb6b28e7b0b41d85347373c93bd5a8.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Removed the object ID and status-changing epoch in 8b708dd02ffb6b28e7b0b41d85347373c93bd5a8.

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: Given that merge operator is executed lazily by RocksDb, we can assume we may only have `merge_blob_info` as the highest level instruction. In there, we do log the key that the merge operator is operating on, but it's just logging bytes. We may want to convert it to an actual readable blob id (at some point).

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: I just wanted to say that that part of the doc, `the highest status ('certified' > 'registered')`, is still needed or not, since now, each data structure is only used to track one type of status. No longer relevant since it is removed.

crates/walrus-sdk/src/api.rs
halfprice: LGTM. Maybe we can revisit this after evolving BlobStatus, and see if managing the ordering may become too complicated.

PR #767, opened 2024-09-03
feat(client): separate the metrics runtime from the daemon runtime

### LINKED ISSUE #496: Expose logs and metrics in the Walrus daemon

### LINKED ISSUE #589: Reorganize the client CLI code

giac-mysten:
* The client separates functionality between `cli` commands -- simple interactive commands -- and `daemon` commands -- that spin up a server and keep running until stopped.
* For `daemon` commands, the client now has two separate runtimes, one for metrics and logging and the metrics server, and one for running the daemon itself
* Most of the binary logic, setup, and args has been moved to `src` (under @jpcsmith 's suggestion)
* The logic for each command is separated into the `ClientCommandRunner` struct

Closes #589 
Closes #496

### DIFF
--- a/crates/walrus-service/src/client/cli/utils.rs
+++ b/crates/walrus-service/src/client/cli/utils.rs
@@ -7,48 +7,55 @@ use std::env;
 
 use anyhow::{anyhow, Result};
 use tracing::subscriber::DefaultGuard;
-use tracing_subscriber::{layer::SubscriberExt as _, util::SubscriberInitExt, EnvFilter, Layer};
+use tracing_subscriber::{
+    filter::Filtered,
+    layer::{Layered, SubscriberExt as _},
+    util::SubscriberInitExt,
+    EnvFilter,
+    Layer,
+    Registry,
+};
 
-// Prepare the tracing subscriber based on the environment variables.
-macro_rules! prepare_subscriber {
-    () => {{
-        // Use INFO level by default.
-        let directive = format!(
-            "info,{}",
-            env::var(EnvFilter::DEFAULT_ENV).unwrap_or_default()
-        );
-        let layer = tracing_subscriber::fmt::layer().with_writer(std::io::stderr);
+type PrepareResult = Result<
+    Layered<Filtered<Box<dyn Layer<Registry> + Send + Sync>, EnvFilter, Registry>, Registry>,
+>;
 
-        // Control output format based on `LOG_FORMAT` env variable.
-        let format = env::var("LOG_FORMAT").ok();
-        let layer = if let Some(format) = &format {
-            match format.to_lowercase().as_str() {
-                "default" => layer.boxed(),
-                "compact" => layer.compact().boxed(),
-                "pretty" => layer.pretty().boxed(),
-                "json" => layer.json().boxed(),
-                s => Err(anyhow!("LOG_FORMAT '{}' is not supported", s))?,
-            }
-        } else {
-            layer.boxed()
-        };
+/// Prepare the tracing subscriber based on the environment variables.
+fn prepare_subscriber() -> PrepareResult {
+    // Use INFO level by default.
+    let directive = format!(
+        "info,{}",
+        env::var(EnvFilter::DEFAULT_ENV).unwrap_or_default()
+    );
+    let layer = tracing_subscriber::fmt::layer().with_writer(std::io::stderr);
 
-        tracing_subscriber::registry().with(layer.with_filter(EnvFilter::new(directive.clone())))
-    }};
+    // Control output format based on `LOG_FORMAT` env variable.
+    let format = env::var("LOG_FORMAT").ok();
+    let layer = if let Some(format) = &format {
+        match format.to_lowercase().as_str() {
+            "default" => layer.boxed(),
+            "compact" => layer.compact().boxed(),
+            "pretty" => layer.pretty().boxed(),
+            "json" => layer.json().boxed(),
+            s => Err(anyhow!("LOG_FORMAT '{}' is not supported", s))?,
+        }
+    } else {
+        layer.boxed()
+    };
+
+    Ok(tracing_subscriber::registry().with(layer.with_filter(EnvFilter::new(directive.clone()))))
 }
 
 /// Initializes the logger and tracing subscriber as the global subscriber.
 pub fn init_tracing_subscriber() -> Result<()> {
-    let subscriber = prepare_subscriber!();
-    subscriber.init();
+    prepare_subscriber()?.init();
     tracing::debug!("initialized global tracing subscriber");
     Ok(())
 }
 
 /// Initializes the logger and tracing subscriber as the subscriber for the current scope.
 pub fn init_scoped_tracing_subscriber() -> Result<DefaultGuard> {
-    let subscriber = prepare_subscriber!();
-    let guard = subscriber.set_default();
+    let guard = prepare_subscriber()?.set_default();
     tracing::debug!("initialized scoped tracing subscriber");
     Ok(guard)
 }


crates/walrus-service/src/client/cli_utils/args.rs
```diff
@@ -0,0 +1,418 @@
...
+use anyhow::{anyhow, Result};
+use clap::{Args, Parser, Subcommand};
+use serde::Deserialize;
+use serde_with::{serde_as, DisplayFromStr};
+use walrus_core::{encoding::EncodingConfig, BlobId};
+
+use super::{parse_blob_id, read_blob_from_file, BlobIdDecimal, HumanReadableBytes};
+
+/// The version of the Walrus client.
+pub const VERSION: &str = crate::utils::version!();
```
mlegner: Important: This should be in the binary. See the comment at the macro definition.

crates/walrus-service/bin/client.rs
```diff
@@ -3,716 +3,65 @@
...
-        #[serde(default)]
-        /// The output list of blobs will include expired blobs.
-        include_expired: bool,
-    },
-}
-
-#[derive(Debug, Clone, Args, Deserialize)]
-#[serde(rename_all = "camelCase")]
-struct PublisherArgs {
+pub struct ClientArgs {
```
jpcsmith, line 33: Nit: No reason to change the name, you can also call it app since it's just a wrapper around `cli_utils::App`.

crates/walrus-service/bin/client.rs
```diff
@@ -3,716 +3,65 @@
...
-                SuiContractClient::new(wallet?, config?.system_object, app.gas_budget).await?;
-            let blobs = contract_client.owned_blobs(include_expired).await?;
-            blobs.print_output(app.json)?;
-        }
+#[tokio::main]
+async fn run_app(app: App, metrics_runtime: Option<MetricsAndLoggingRuntime>) -> Result<()> {
+    match (app.is_daemon_command(), metrics_runtime) {
+        (true, Some(runtime)) => run_daemon_app(app, runtime).await,
+        (false, None) => run_cli_app(app).await,
+        _ => unreachable!("if the command is a daemon command, then metrics_runtime must be Some"),
```
```suggestion
async fn run_app(app: App, metrics_runtime: Option<MetricsAndLoggingRuntime>) -> Result<()> {
    if app.is_daemon_command() {
        let runtime = metrics_runtime.expect("metrics runtime must be set for daemon commands");
        run_daemon_app(app, runtime).await
    } else {
        run_cli_app(app).await
    }
```

crates/walrus-service/src/client/cli_utils.rs
jpcsmith, line 1: Nit: The name `client::cli` would also be appropriate, since it embodies the cli entrypoint.

crates/walrus-service/src/client/cli_utils/args.rs
jpcsmith: *i.e., should be in the `bin/` directory

crates/walrus-service/src/client/cli_utils/args.rs
```diff
@@ -0,0 +1,418 @@
...
+    pub fn is_daemon_command(&self) -> bool {
+        self.command.is_daemon_command()
+    }
+
+    /// Gets the metrics address from the commands that support it.
+    pub fn get_metrics_address(&self) -> Option<SocketAddr> {
+        match &self.command {
+            Commands::Publisher { args } => Some(args.daemon_args.metrics_address),
+            Commands::Aggregator { daemon_args, .. } => Some(daemon_args.metrics_address),
+            Commands::Daemon { args } => Some(args.daemon_args.metrics_address),
```
jpcsmith:
```suggestion
            Commands::Publisher { args } | Commands::Daemon { args } => Some(args.daemon_args.metrics_address),
            Commands::Aggregator { daemon_args, .. } => Some(daemon_args.metrics_address),
```

crates/walrus-service/src/client/cli/utils.rs
```diff
@@ -0,0 +1,108 @@
...
+        .with(layer.with_filter(EnvFilter::new(directive.clone())))
+        .init();
+    tracing::debug!(%directive, ?format, "initialized tracing subscriber");
+
+    Ok(())
+}
+
+/// Runs the binary commands in "cli" mode (i.e., without running a server).
+#[tokio::main]
+pub async fn run_cli_app(runner: ClientCommandRunner, command: CliCommands) -> Result<()> {
```
mlegner: Question: Is the `#[tokio::main]` attribute the reason to have these as standalone functions instead of methods on `ClientCommandRunner`?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -0,0 +1,434 @@
...
+    Cli(CliCommands),
+    /// Commands to run the binary in daemon mode.
+    #[clap(flatten)]
+    Daemon(DaemonCommands),
+    /// Run the client by specifying the arguments in a JSON string; CLI options are ignored.
+    Json {
+        /// The JSON-encoded args for the Walrus CLI; if not present, the args are read from stdin.
+        ///
+        /// The JSON structure follows the CLI arguments, containing global options and a "command"
+        /// object at the root level. The "command" object itself contains the command (e.g.,k
```
```suggestion
        /// object at the root level. The "command" object itself contains the command (e.g.,
```

crates/walrus-service/bin/client.rs
```diff
@@ -3,716 +3,61 @@
...
-    let config = load_configuration(&app.config);
-    tracing::debug!(?app, ?config, "initializing the client");
-    let wallet_path = app.wallet.clone().or(config
-        .as_ref()
-        .ok()
-        .and_then(|conf| conf.wallet_config.clone()));
-    let wallet = load_wallet_context(&wallet_path);
+fn client() -> Result<()> {
+    let mut app = ClientArgs::parse().inner;
+    app.extract_json_command()?;
```
mlegner, line 41: Important: Currently we don't have a tracing subscriber for this part, which does have some useful output sometimes.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -0,0 +1,434 @@
...
+        ///           "out": "some_output_file"
+        ///         }
+        ///       }
+        ///     }
+        ///
+        /// Important: If the "read" command does not have an "out" file specified, the output JSON
+        /// string will contain the full bytes of the blob, encoded as a Base64 string.
+        #[clap(verbatim_doc_comment)]
+        command_string: Option<String>,
+    },
```
mlegner: Important: The current setup changes the expected format of the JSON input. The following resolves that:

```suggestion
    /// Run the client by specifying the arguments in a JSON string; CLI options are ignored.
    Json {
        /// The JSON-encoded args for the Walrus CLI; if not present, the args are read from stdin.
        ///
        /// The JSON structure follows the CLI arguments, containing global options and a "command"
        /// object at the root level. The "command" object itself contains the command (e.g.,
        /// "store", "read", "publisher", "blobStatus", ...) with an object containing the command
        /// options.
        ///
        /// Note that where CLI options are in "kebab-case", the respective JSON strings are in
        /// "camelCase".
        ///
        /// For example, to read a blob and write it to "some_output_file" using a specific
        /// configuration file, you can use the following JSON input:
        ///
        ///     {
        ///       "config": "path/to/client_config.yaml",
        ///       "command": {
        ///         "read": {
        ///           "blobId": "4BKcDC0Ih5RJ8R0tFMz3MZVNZV8b2goT6_JiEEwNHQo",
        ///           "out": "some_output_file"
        ///         }
        ///       }
        ///     }
        ///
        /// Important: If the "read" command does not have an "out" file specified, the output JSON
        /// string will contain the full bytes of the blob, encoded as a Base64 string.
        #[clap(verbatim_doc_comment)]
        command_string: Option<String>,
    },
    /// Commands to run the binary in CLI mode.
    #[clap(flatten)]
    #[serde(untagged)]
    Cli(CliCommands),
    /// Commands to run the binary in daemon mode.
    #[clap(flatten)]
    #[serde(untagged)]
    Daemon(DaemonCommands),
```

crates/walrus-service/src/client/cli/utils.rs
giac-mysten: good question: for some reason I thought `tokio::main` required a standalone function. But actually, it does not. Fixed.

crates/walrus-service/bin/client.rs
giac-mysten: Yep, unfortunately i'm not too sure how to fix this problem. Should I create a temporary subscriber just for this?

crates/walrus-service/src/client/cli/args.rs
giac-mysten: great catch, thanks!

crates/walrus-service/bin/client.rs
mlegner: Hmm... Not sure. If that's not too complicated, I'd say yes. I don't really have a better solution right now.

crates/walrus-service/src/client/cli/args.rs
mlegner: Potential follow-up: We could add some tests with example input strings for the JSON mode (for example from our docs or the walrus-sites) and make sure they are parsed successfully.

giac-mysten: let me add that

crates/walrus-service/bin/client.rs
giac-mysten: I added a temporary scoped subscribed here: d46e8524b650b99d50d459fc4e5efea6343f598e

The only thing i'm not too happy with is the macro `prepare_subscriber!()`, which I introduced to deduplicate code and avoid dealing with the hideous signature 😬 lmk if there is a better way.

crates/walrus-service/src/client/cli/args.rs
giac-mysten: Added here 606020f8b05238f3b61b923eb3db4956aba53680

crates/walrus-service/bin/client.rs
mlegner: Thanks, this works great.

No strong opinion, but I'd have a slight preference for using a function here instead of a macro, despite the somewhat ugly return type `Result<Layered<Filtered<Box<dyn Layer<Registry> + Send + Sync>, EnvFilter, Registry>, Registry>>`. 🤷

PR #768, opened 2024-09-03
fix: clean up all `unwrap`s in production code

### LINKED ISSUE #765: Check all `unwrap`s in the codebase
Something that came up during the incident review. 
There seem to be `unwrap`s in the codebase that are not commented/marked.

Check if this is true, remove them, or turn them to `expect`s.

mlegner:
This PR removes all instances of `unwrap` in production code. This excludes tests, benchmarks, and the `walrus-orchestrator` crate.

Some cases are replaced by an `expect` explaining the reason why this can never panic. In some cases a slight rewrite of the code allowed removing the unwrapping altogether.

There are two cases where the previous behavior was potentially problematic:
- In `crates/walrus-event/src/checkpoint_processor.rs`
- The authorization header in `crates/walrus-service/src/node/server/extract.rs` and `crates/walrus-service/src/node/server/routes.rs`.

Closes #765

### DIFF
--- a/crates/walrus-service/src/node/server/extract.rs
+++ b/crates/walrus-service/src/node/server/extract.rs
@@ -8,8 +8,10 @@ use axum::{
     http::{header, request::Parts, HeaderMap, HeaderValue, StatusCode},
     response::{IntoResponse, Response},
 };
+use fastcrypto::traits::EncodeDecodeBase64 as _;
 use reqwest::header::AUTHORIZATION;
 use serde::{de::DeserializeOwned, Deserialize, Serialize};
+use walrus_core::PublicKey;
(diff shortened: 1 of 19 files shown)


crates/walrus-service/src/node/server/extract.rs
```diff
@@ -126,7 +128,7 @@ where
                     StatusCode::INTERNAL_SERVER_ERROR,
                     StatusCode::INTERNAL_SERVER_ERROR
                         .canonical_reason()
-                        .unwrap(),
+                        .expect("this definitely has a canonical reason"),
```
giac-mysten, line 131: 😂

PR #775, opened 2024-09-04
chore: add uptime metric in walrus node

halfprice:
To monitor how long the process runs

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -101,6 +101,16 @@ mod commands {
         }
 
         let metrics_runtime = MetricsAndLoggingRuntime::start(config.metrics_address)?;
+        let registry_clone = metrics_runtime.registry.clone();
+        metrics_runtime.runtime.spawn(async move {
+            registry_clone
+                .register(mysten_metrics::uptime_metric(
+                    "walrus_node",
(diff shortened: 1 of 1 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -101,6 +101,16 @@ mod commands {
         }
 
         let metrics_runtime = MetricsAndLoggingRuntime::start(config.metrics_address)?;
+        let registry_clone = metrics_runtime.registry.clone();
+        metrics_runtime.runtime.spawn(async move {
+            registry_clone
+                .register(mysten_metrics::uptime_metric(
+                    "walrus_node",
+                    VERSION,
+                    "walrus",
```
jpcsmith, line 110: Question: Do we share the "walrus" network identifier across devnet, testnet, and mainnet? Or should this be something else? Does it matter since the metrics are tagged by grafana agent?

halfprice: I don't think it matters in terms of mixing up metrics across different network, If metrics are already grouped by the network tag (taking a quick look I think walrus does this already).

In Sui, this is set to be the digest of genesis checkpoint. We could do similar thing (maybe package id of the contract). But not sure worth the effort at the moment.

PR #776, opened 2024-09-05
feat: create ShardSyncConfig in StorageNodeConfig

halfprice:
To allow customize shard sync related configs.

Contribute to #705

### DIFF
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -554,7 +554,7 @@ impl ShardStorage {
                     self.id(),
                     next_starting_blob_id,
                     sliver_type,
-                    config.sliver_count_per_sync_request as u64,
+                    config.sliver_count_per_sync_request,
                     epoch,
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -544,7 +554,7 @@ impl ShardStorage {
                     self.id(),
                     next_starting_blob_id,
                     sliver_type,
-                    10, // TODO(#705): make this configurable.
+                    config.sliver_count_per_sync_request as u64,
```
mlegner: Suggestion: As this is the only location where we use this field, we could make it a `u64` from the start.

crates/walrus-service/src/node.rs
mlegner, line 1: Remark: There is a related TODO left on line 2400 in this file.

halfprice: Yes, that was intentional. The goal of that TODO is to setup the test using different config, which this PR hasn't address yet.

Issue #782, opened 2024-09-06
Add funding wallet to the stress client


giac-mysten:
Given the new rate limits on the faucet, the stress client keeps running out of SUI.

We should add a sui config to the stress client, configured with a wallet with funds.
These funds should then be used by the stress client to fund the wallets of the walrus clients it uses.

PR #788, opened 2024-09-09
feat(stress): allow taking funds from local wallet

giac-mysten:
The stress clients are now able to get funds from a local wallet (which should be pre-provisioned with a lot of SUI).
The funds are taken from this wallet both for initialization and when the clients run out.

To use a wallet as a "gas station", pass the wallet config _explicitly_ when running the stress client.
`cargo run  ... --wallet-path working_dir/sui_client.yaml`

If the config is not specified, the stress client will use the standard faucet.

Closes: #782

### DIFF
--- a/crates/walrus-stress/src/refill.rs
+++ b/crates/walrus-stress/src/refill.rs
@@ -1,0 +1,239 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Utilities to refill gas for the stress clients.
+
+use std::{collections::BTreeSet, path::PathBuf, sync::Arc, time::Duration};
+
+use anyhow::Result;
+use futures::{
+    future::{self, try_join_all},
+    StreamExt,
+};
+use sui_sdk::{
+    rpc_types::SuiTransactionBlockResponse,
+    types::{
+        base_types::SuiAddress,
+        programmable_transaction_builder::ProgrammableTransactionBuilder,
+    },
+    wallet_context::WalletContext,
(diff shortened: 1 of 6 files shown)


crates/walrus-stress/src/refill.rs
```diff
@@ -0,0 +1,239 @@
...
+use crate::metrics::ClientMetrics;
+
+// If a node has less than `MIN_NUM_COINS` without at least `MIN_COIN_VALUE`,
+// we need to request additional coins from the faucet.
+const MIN_COIN_VALUE: u64 = 500_000_000;
+// We need at least a payment coin and a gas coin.
+const MIN_NUM_COINS: usize = 2;
+/// The amount in MIST that is transferred from the wallet refill account to the stress clients at
+/// each request.
+const WALLET_REFILL_AMOUNT: u64 = 1_000_000_000;
```
asonnino, line 35: Are we certain enough about these values that we don't need to add them as CLI parameters?

giac-mysten: The only value I added is `WALLET_REFILL_AMOUNT`, which is the amount that the wallet sends to the client each time it gets a request.

I chose `1 SUI` b/c it's the same the faucet would provide (and we cannot configure the faucet) -- I can add a configuration option if you think it's useful. The other two values are from the previous implementation, and I think are fine as they are.

asonnino: yeah looks good then

asonnino: let's land

PR #796, opened 2024-09-11
feat(move): apportionment using D'Hondt

### LINKED ISSUE #619: Implement shard assignment
During epoch change, we need to re-distribute shards to represent the stake distribution. There is code for voting-power assignment in Sui that we can potentially reuse for that: https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/sui-system/sources/voting_power.move

Part of #560

tnowacki:
Closes #619

### DIFF
--- a/contracts/walrus/sources/staking/staking_inner.move
+++ b/contracts/walrus/sources/staking/staking_inner.move
@@ -408,108 +408,95 @@ fun dhondt(
     // initial price guess following Pukelsheim
     let mut price = from_rational(total_stake, n_shards + (n_nodes / 2));
     if (n_nodes == 0) return (price, vector[]);
-    let mut needs_one_more = vector<u64>[];
-    let mut needs_one_less = vector<u64>[];
-    loop {
-        let mut shards = stake.map_ref!(|s| u64_div(*s, price));
-        needs_one_more.do!(|i| *&mut shards[i] = shards[i] + 1);
-        needs_one_less.do!(|i| *&mut shards[i] = shards[i] - 1);
-        let n_shards_distributed = shards.fold!(0, |acc, x| acc + x);
-
-        // if all shards are distributed, we are done
-        if (n_shards_distributed == n_shards) break (price, shards.map!(|s| s as u16));
-
-        needs_one_more = vector[];
-        needs_one_less = vector[];
-        if (n_shards_distributed < n_shards) {
-            let mut with_max = vector[];
-            let mut max = from_raw(0);
-            let new_prices = stake.zip_map_ref!(&shards, |s, m| from_rational(*s, *m + 1));
-            let mut i = 0;
-            new_prices.do!(|p| {
-                if (p.to_raw() > max.to_raw()) {
-                    max = p;
-                    with_max = vector[i];
-                } else if (p == max) {
-                    with_max.push_back(i);
-                };
-                i = i + 1;
-            });
-            let n_max = with_max.length();
-            let adjusted_distribution = n_shards_distributed + n_max;
-            if (n_max > 1 && adjusted_distribution > n_shards) {
-                // If there is more than one node with the same max price, we need to manually
-                // adjust enough of them so that `n_shards_distributed == n_shards`. In which case,
-                // there is no need to adjust the price.
-                let to_give = n_shards - n_shards_distributed;
-                // slightly optimized
-                // needs_one_more = ranking.filter!(|i| with_max.contains(i)).take!(to_give)
-                'update_needs_one_more: {
-                    let n = ranking.length();
-                    // iterate over the ranking, giving shards from the nodes with the highest
-                    // rank first
-                    n.do!(|i| {
-                        let idx = &ranking[i];
-                        if (with_max.contains(idx)) {
-                            needs_one_more.push_back(*idx);
-                            if (needs_one_more.length() == to_give) return 'update_needs_one_more;
-                        }
-                    })
-                }
-            } else {
-                // decrease the price such that one node gets an additional shard
-                price = max
-            }
-        } else {
-            let mut with_min = vector[];
-            // TODO use std::u64::max_value!()
-            let mut min = from_raw(0xFFFF_FFFF_FFFF_FFFF);
-            let new_prices = stake.zip_map_ref!(&shards, |s, m| {
-                let m = *m;
-                if (m == 0) option::none()
-                else option::some(from_rational(*s, m))
-            });
-            let mut i = 0;
-            new_prices.do!(|p_opt| {
-                p_opt.do!(|p| {
-                    if (p.to_raw() < min.to_raw()) {
-                        min = p;
-                        with_min = vector[i];
-                    } else if (p == min) {
-                        with_min.push_back(i);
-                    }
+    let mut shards = stake.map_ref!(|s| u64_div(*s, price));
+    let mut n_shards_distributed = shards.fold!(0, |acc, x| acc + x);
+    // loop until all shards are distributed
+    while (n_shards_distributed != n_shards) {
+        n_shards_distributed = if (n_shards_distributed < n_shards) {
+                // We decrease the price slightly such that some nodes get an additional shard.
+                price = from_raw(0);
+                let mut at_threshold = vector[];
+                let mut i = 0;
+                stake.zip_do_ref!(&shards, |s, m| {
+                    let threshold_price = from_rational(*s, *m + 1);
+                    if (threshold_price.to_raw() > price.to_raw()) {
+                        price = threshold_price;
+                        at_threshold = vector[i];
+                    } else if (threshold_price == price) {
+                        at_threshold.push_back(i);
+                    };
+                    i = i + 1;
                 });
-                i = i + 1;
-            });
-            let n_min = with_min.length();
-            let adjusted_distribution = n_shards_distributed.max(n_min) - n_min;
-            if (n_min > 1 && adjusted_distribution < n_shards) {
-                // If there is more than one node with the same min price, we need to manually
-                // adjust enough of them so that `n_shards_distributed == n_shards`. In which case,
-                // there is no need to adjust the price.
-                let to_take = n_shards_distributed - n_shards;
-                // slightly optimized
-                // needs_one_less =
-                //     ranking.reverse().filter!(|i| with_min.contains(i)).take!(to_take)
-                'update_needs_one_less: {
-                    let n = ranking.length();
-                    // reverse iterate over the ranking, taking shards from the nodes with the
-                    // lowest rank first
-                    n.do!(|i| {
-                        let idx = &ranking[n - 1 - i];
-                        if (with_min.contains(idx)) {
-                            needs_one_less.push_back(*idx);
-                            if (needs_one_less.length() == to_take) return 'update_needs_one_less;
+
+                let adjusted_distribution = n_shards_distributed + at_threshold.length();
+                if (adjusted_distribution <= n_shards) {
+                    // We give one additional shard to all nodes with the same threshold price.
+                    at_threshold.do!(|n| *&mut shards[n] = shards[n] + 1);
+                    adjusted_distribution
+                } else {
+                    // If there are more nodes with the same threshold price than additional shards
+                    // to distribute, we only give an additional shard to a subset according to
+                    // their rank so that `n_shards_distributed == n_shards`.
+                    let mut to_give = n_shards - n_shards_distributed;
+                    // Iterate over the ranking, giving shards from the nodes with the highest
+                    // rank first.
+                    ranking.length().do!(|i| {
+                        if (to_give == 0) return;
+                        let idx = &ranking[i];
+                        if (at_threshold.contains(idx)) {
+                            *&mut shards[*idx] = shards[*idx] + 1;
+                            to_give = to_give - 1;
                         }
-                    })
+                    });
+                    n_shards
                 }
             } else {
-                // increase the price such that one node loses one shard
-                // min + (1 / 4000000000)
-                price = min.add(from_raw(1))
+                // We increase the price slightly such that some nodes get one fewer shard.
+                price = from_raw(0xFFFF_FFFF_FFFF_FFFF); // TODO: use std::u64::max_value!()
+                let mut at_threshold = vector[];
+                let mut i = 0;
+                stake.zip_do_ref!(&shards, |s, m| {
+                    let m = *m;
+                    if (m != 0) {
+                        let threshold_price = from_rational(*s, m);
+                        if (threshold_price.to_raw() < price.to_raw()) {
+                            price = threshold_price;
+                            at_threshold = vector[i];
+                        } else if (threshold_price == price) {
+                            at_threshold.push_back(i);
+                        }
+                    };
+                    i = i + 1;
+                });
+                // `at_threshold.length() <= n_shards_distributed` due to the check `m != 0` above.
+                let adjusted_distribution = n_shards_distributed - at_threshold.length();
+                if (adjusted_distribution >= n_shards) {
+                    // We increase the price slightly above the threshold such that all nodes at
+                    // the threshold lose one shard.
+                    // price += 1 / 4000000000
+                    price = price.add(from_raw(1));
+                    at_threshold.do!(|n| *&mut shards[n] = shards[n] - 1);
+                    adjusted_distribution
+                } else {
+                    // If there are more nodes with the same threshold price than shards to take
+                    // away, we only remove a shard from a subset according to their rank so that
+                    // `n_shards_distributed == n_shards`. In this case, the price remains at the
+                    // threshold.
+                    let mut to_take = n_shards_distributed - n_shards;
+                    let n = ranking.length();
+                    n.do!(|i| {
+                        if (to_take == 0) return;
+                        let idx = &ranking[n - 1 - i];
+                        if (at_threshold.contains(idx)) {
+                            *&mut shards[*idx] = shards[*idx] - 1;
+                            to_take = to_take - 1;
+                        }
+                    });
+                    n_shards
+                }
             }
-        }
-    }
+    };
+    (price, shards.map!(|s| s as u16))
 }
 
 use fun fp_add as FixedPoint32.add;


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +360,131 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+    };
+
+    let n_nodes = n_nodes as u64;
+    let n_shards = n_shards as u64;
+    assert!(total_stake > 0, ENoStake);
+    assert!(n_nodes == stake.length(), EMismatchedNodes);
+
+    // initial price guess following Pukelsheim
+    // total_stake / (n_shards + (count_nodes / 2)
+    let mut price = from_rational(total_stake, n_shards + (n_nodes / 2));
```
mlegner, line 409: Nit: I believe the `/2` should not be an integer division. However, in our cases it's probably fine as the difference is likely minimal (and it's only an initial guess).

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +360,131 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+    let active_ids = self.active_set.active_ids();
+    let stake = active_ids.map_ref!(|node_id| self.active_set[node_id]);
+    let (_price, shards) =
+        dhondt(self.active_set.total_stake(), self.active_set.size(), self.n_shards, stake);
+    shards
+}
+
+fun dhondt(
+    total_stake: u64,
+    n_nodes: u16,
```
mlegner: Question: Why are these explicit inputs to the functions? Those can be computed from the `stake` vector.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +360,131 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+                } else if (p == max) {
+                    with_max.push_back(i);
+                };
+                i = i + 1;
+            });
+            if (with_max.length() > 1) {
+                // if there is more than one node with the same price, we need to manually adjust
+                // just one of them. And no need to adjust the price.
+                // TODO choose the node in a more meaningful way
+                needs_one_more = option::some(with_max[0]);
```
mlegner: Important: This is not sufficient, there could be multiple nodes that need one more. For example, the test case `dhondt_case(7, stake, vector[2, 2, 2, 1])` fails with the current code.

mlegner: Also, there is an optimization: If `n_shards_distributed + with_max.length() <= n_shards`, then we *can* adjust the price and don't have to break ties.

contracts/walrus/sources/staking/staking_inner.move
tnowacki: Was just grabbing data in the active set, I'll instead rely on the stake directly

contracts/walrus/sources/staking/staking_inner.move
tnowacki: > Important: This is not sufficient, there could be multiple nodes that need one more. For example, the test case `dhondt_case(7, stake, vector[2, 2, 2, 1])` fails with the current code.

What is `stake` in your test case?

contracts/walrus/sources/staking/staking_inner.move
tnowacki: It is a bit tricky to use a rational here instead of `n_nodes / 2`. I want to make sure the denominator here is as large as possible to deal with the integer upper bounds of `from_rational`

contracts/walrus/sources/staking/staking_inner.move
tnowacki: I got a case that wasn't solved by this and ended up using the optimization. Let me know what you think!

contracts/walrus/sources/staking/staking_inner.move
mlegner: OK, makes sense. As I said, it doesn't matter much anyway, so let's just keep it like this.

contracts/walrus/sources/staking/staking_inner.move
mlegner: > What is `stake` in your test case?

Whoops, apologies. That would be four identical values like `vector[25000, 25000, 25000, 25000]`.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,173 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+                    max = p;
+                    with_max = vector[i];
+                } else if (p == max) {
+                    with_max.push_back(i);
+                };
+                i = i + 1;
+            });
+            let n_max = with_max.length();
+            let adjusted_distribution = n_shards_distributed + n_max;
+            if (n_max > 1 && adjusted_distribution > n_shards) {
```
mlegner: Nit: I think the `n_max > 1` is implied by the second clause.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,173 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+                // if there is more than one node with the same price, we need to manually adjust
+                // just one of them. And no need to adjust the price.
```
mlegner: Nit: This comment is no longer 100% accurate.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,173 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+    shards
+}
+
+// Implementation of the D'Hondt method (aka Jefferson method) for apportionment.
+// TODO: currently because of Fixedpoint32, there is a "low" limit on the `total_stake`.
+// Assuming 1000 shards (and ignoring the number of nodes), the limit for the total stake is
+// 4,000,000,000,000 (4e12).
+fun dhondt(
+    // a ranking of the nodes by index in the `stake` vector. The lower the index, the higher the
+    // rank.
```
mlegner: AFAIU, this ranking says which nodes get additional shards in case of ties, is that right? So the first index in this vector is the node that has the highest precedence for additional shards. And this ranking is provided as input to be able to define this ranking based on any node parameters we want.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,173 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+        v
+    };
+    let (_price, shards) = dhondt(ranking, self.n_shards, stake);
+    shards
+}
+
+// Implementation of the D'Hondt method (aka Jefferson method) for apportionment.
+// TODO: currently because of Fixedpoint32, there is a "low" limit on the `total_stake`.
+// Assuming 1000 shards (and ignoring the number of nodes), the limit for the total stake is
+// 4,000,000,000,000 (4e12).
```
mlegner, line 387: Question: What happens if this is exceeded? Is it then simply not very precise or does it fail completely? (I assume the latter because `4e12/1000 ~ 2e32` gives the maximum value of the `FixedPoint32`.)

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,173 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+                    let n = ranking.length();
+                    n.do!(|i| {
+                        let idx = &ranking[i];
+                        if (with_max.contains(idx)) {
+                            v.push_back(*idx);
+                            if (v.length() == to_give) return 'needs_one_more v;
+                        }
+                    });
+                    v
+                };
```
mlegner: Question: What is the benefit of working on this temporary vector instead of directly modifying the `needs_one_more` vector?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,173 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+                            v.push_back(*idx);
+                            if (v.length() == to_take) return 'needs_one_less v;
+                        }
+                    });
+                    v
+                };
+            } else {
+                // increase the price such that one node loses one shard
+                // min + (1 / 4000000000)
+                price = min.add(from_raw(1))
```
mlegner: I like this! 👍

contracts/walrus/tests/staking/staking_inner_tests.move
```diff
@@ -176,3 +177,74 @@ fun test_epoch_sync_done_duplicate() {
...
+    let (_price, allocation) = dhondt(shards, stake);
+    assert_eq!(allocation, expected);
+    assert_eq!(allocation.sum!(), shards);
+}
+
+#[test]
+fun test_dhondt_basic() {
+    // even
+    let stake = vector[25000, 25000, 25000, 25000];
+    dhondt_case(4, stake, vector[1, 1, 1, 1]);
```
mlegner, line 196: Suggestion: Can we add some test cases with multiple `needs_one_more` or `needs_one_less`?
```move
    dhondt_case(7, stake, vector[2, 2, 2, 1]);
    dhondt_case(6, stake, vector[2, 2, 1, 1]);
```

contracts/walrus/sources/staking/staking_inner.move
tnowacki: That is a much better way to explain what is going on. I'll update the comment

contracts/walrus/sources/staking/staking_inner.move
tnowacki: It will abort

contracts/walrus/sources/staking/staking_inner.move
tnowacki: I could do that... but with the strange control flow, it might be a bit harder to follow? I'll give it a shot though

contracts/walrus/tests/staking/staking_inner_tests.move
tnowacki: Some of those are covered below, in test_dhondt_edge_case. Maybe I should break that out into a test just for ties

contracts/walrus/sources/staking/staking_inner.move
tnowacki: Created a follow up task tracking this #835 as requested.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,178 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
     assert!(self.next_committee.is_none());
 
     let shard_threshold = self.active_set.total_stake() / (self.n_shards as u64);
```
karlwuest:
```suggestion
```

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,178 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+
+// Implementation of the D'Hondt method (aka Jefferson method) for apportionment.
+// TODO: currently because of Fixedpoint32, there is a "low" limit on the `total_stake`.
+// Assuming 1000 shards (and ignoring the number of nodes), the limit for the total stake is
+// 4,000,000,000,000 (4e12).
+fun dhondt(
+    // A ranking of the nodes by index in the `stake` vector. The lower the index, the higher the
+    // rank. So the first index in this vector is the node that has the highest precedence for
+    // additional shards.
+    ranking: vector<u64>,
```
karlwuest, line 392: Question: Is this just for tie-breaking?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,178 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+
+    // initial price guess following Pukelsheim
+    let mut price = from_rational(total_stake, n_shards + (n_nodes / 2));
+    if (n_nodes == 0) return (price, vector[]);
+    let mut needs_one_more = vector<u64>[];
+    let mut needs_one_less = vector<u64>[];
+    loop {
+        let mut shards = stake.map_ref!(|s| u64_div(*s, price));
+        needs_one_more.do!(|i| *&mut shards[i] = shards[i] + 1);
+        needs_one_less.do!(|i| *&mut shards[i] = shards[i] - 1);
```
karlwuest: Suggestion: Currently, either the price is updated or one of `needs_one_more`/`needs_one_less`. It seems that it should be possible to always update either the `needs_one_more`/`needs_one_less` and only do the shard distribution based on the current price at initialization (outside the loop). This should be a bit more efficient since we save the iteration over the stake vector. Am I missing something?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,178 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+                p_opt.do!(|p| {
+                    if (p.to_raw() < min.to_raw()) {
+                        min = p;
+                        with_min = vector[i];
+                    } else if (p == min) {
+                        with_min.push_back(i);
+                    }
+                });
+                i = i + 1;
+            });
```
karlwuest: See above (same code for max).

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -350,35 +356,178 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
+            let mut i = 0;
+            new_prices.do!(|p| {
+                if (p.to_raw() > max.to_raw()) {
+                    max = p;
+                    with_max = vector[i];
+                } else if (p == max) {
+                    with_max.push_back(i);
+                };
+                i = i + 1;
+            });
```
karlwuest: Suggestion/Nit: The `zip_map` + `do` is quite nice for readability, but it may be more efficient here to just do a single loop that does all of this, either with an actual loop or with `zip_do`.

contracts/walrus/sources/staking/staking_inner.move
mlegner: Yes.

contracts/walrus/sources/staking/staking_inner.move
mlegner: So the `needs_one_more` and `needs_one_less` are only needed at the end in case of ties. The `shards` have to be recomputed in the loop, however, instead of recomputing them completely, we could just update the vector by increasing the shards for all nodes in `with_max` or decreasing them for the nodes in `with_min`.

The special tie-breaking could happen outside of the loop as this only occurs at most once at the end.

karlwuest: >instead of recomputing them completely, we could just update the vector by increasing the shards for all nodes in with_max or decreasing them for the nodes in with_min

Yes, this is what I meant. That would essentially be the same as just updating the `needs_one_more`/`needs_one_less`, wouldn't it?

mlegner: OK, I have a version that does not time out and passes the tests. Will clean it up and share the code here.

mlegner: Here's my slightly optimized code (also addresses Karl's comment about the `zip_map` below):
```move
{
    use std::fixed_point32::{
        divide_u64 as u64_div,
        create_from_rational as from_rational,
        create_from_raw_value as from_raw,
        get_raw_value as to_raw,
    };

    let total_stake = stake.fold!(0, |acc, x| acc + x);
    let n_nodes = stake.length();
    let n_shards = n_shards as u64;
    assert!(total_stake > 0, ENoStake);

    // initial price guess following Pukelsheim
    let mut price = from_rational(total_stake, n_shards + (n_nodes / 2));
    if (n_nodes == 0) return (price, vector[]);
    let mut shards = stake.map_ref!(|s| u64_div(*s, price));
    let mut n_shards_distributed = shards.fold!(0, |acc, x| acc + x);
    loop {
        // if all shards are distributed, we are done
        if (n_shards_distributed == n_shards) break (price, shards.map!(|s| s as u16));

        if (n_shards_distributed < n_shards) {
            // We decrease the price slightly such that some nodes get an additional shard.
            price = from_raw(0);
            let mut at_threshold = vector[];
            let mut i = 0;
            stake.zip_do_ref!(&shards, |s, m| {
                let threshold_price = from_rational(*s, *m + 1);
                if (threshold_price.to_raw() > price.to_raw()) {
                    price = threshold_price;
                    at_threshold = vector[i];
                } else if (threshold_price == price) {
                    at_threshold.push_back(i);
                };
                i = i + 1;
            });

            let adjusted_distribution = n_shards_distributed + at_threshold.length();
            if (adjusted_distribution <= n_shards) {
                // We give one additional shard to all nodes with the same threshold price.
                at_threshold.do!(|n| {
                    *&mut shards[n] = shards[n] + 1;
                });
                n_shards_distributed = adjusted_distribution;
            } else {
                // If there are more nodes with the same threshold price than additional shards to
                // distribute, we only give an additional shard to a subset according to their rank
                // so that `n_shards_distributed == n_shards`.
                let mut to_give = n_shards - n_shards_distributed;
                // Iterate over the ranking, giving shards from the nodes with the highest
                // rank first.
                ranking.length().do!(|i| {
                    if (to_give == 0) return;
                    let idx = &ranking[i];
                    if (at_threshold.contains(idx)) {
                        *&mut shards[*idx] = shards[*idx] + 1;
                        to_give = to_give - 1;
                    }
                });
                n_shards_distributed = n_shards;
            }
        } else {
            // We increase the price slightly such that some nodes get one fewer shard.
            price = from_raw(0xFFFF_FFFF_FFFF_FFFF);
            let mut at_threshold = vector[];
            let mut i = 0;
            stake.zip_do_ref!(&shards, |s, m| {
                let m = *m;
                if (m != 0) {
                    let threshold_price = from_rational(*s, m);
                    if (threshold_price.to_raw() < price.to_raw()) {
                        price = threshold_price;
                        at_threshold = vector[i];
                    } else if (threshold_price == price) {
                        at_threshold.push_back(i);
                    }
                };
                i = i + 1;
            });
            // `at_threshold.length() <= n_shards_distributed` due to the check `m != 0` above.
            let adjusted_distribution = n_shards_distributed - at_threshold.length();
            if (adjusted_distribution >= n_shards) {
                // We increase the price slightly above the threshold such that all nodes at the
                // threshold lose one shard.
                // price + (1 / 4000000000)
                price = price.add(from_raw(1));
                at_threshold.do!(|n| {
                    *&mut shards[n] = shards[n] - 1;
                });
                n_shards_distributed = adjusted_distribution;
            } else {
                // If there are more nodes with the same threshold price than shards to take away,
                // we only remove a shard from a subset according to their rank so that
                // `n_shards_distributed == n_shards`. In this case, the price remains at the
                // threshold.
                let mut to_take = n_shards_distributed - n_shards;
                let n = ranking.length();
                n.do!(|i| {
                    if (to_take == 0) return;
                    let idx = &ranking[n - 1 - i];
                    if (at_threshold.contains(idx)) {
                        *&mut shards[*idx] = shards[*idx] - 1;
                        to_take = to_take - 1;
                    }
                });
                n_shards_distributed = n_shards;
            }
        }
    }
}
```

tnowacki: I'll need to look over this in a bit. But the timeout for tests was just a bad cleanup from addressing a previous round of comments (updating the wrong variable caused an infinite loop)

PR #797, opened 2024-09-12
feat: epoch change events

halfprice:
- Storage node receiving and processing epoch change events

contribute to #560

### DIFF
--- a/crates/walrus-event/src/lib.rs
+++ b/crates/walrus-event/src/lib.rs
@@ -11,7 +11,7 @@ use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use serde::{Deserialize, Serialize};
 use sui_types::{event::EventID, messages_checkpoint::CheckpointSequenceNumber};
 use walrus_core::BlobId;
-use walrus_sui::types::ContractEvent;
+use walrus_sui::types::{BlobEvent, ContractEvent};
 
 pub mod event_processor;
 
@@ -83,17 +83,9 @@ impl EventStreamElement {
         }
     }
 
-    pub fn blob_event(&self) -> Option<&ContractEvent> {
+    pub fn blob_event(&self) -> Option<&BlobEvent> {
         match self {
-            EventStreamElement::ContractEvent(event @ ContractEvent::BlobRegistered(_)) => {
-                Some(event)
-            }
-            EventStreamElement::ContractEvent(event @ ContractEvent::BlobCertified(_)) => {
-                Some(event)
-            }
-            EventStreamElement::ContractEvent(event @ ContractEvent::InvalidBlobID(_)) => {
-                Some(event)
-            }
+            EventStreamElement::ContractEvent(ContractEvent::BlobEvent(event)) => Some(event),
             _ => None,
         }
     }
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -47,7 +47,7 @@ use walrus_event::event_processor::EventProcessor;
 use walrus_sdk::api::{BlobStatus, ServiceHealthInfo, StoredOnNodeStatus};
 use walrus_sui::{
     client::SuiReadClient,
-    types::{BlobCertified, ContractEvent, InvalidBlobId},
+    types::{BlobCertified, BlobEvent, ContractEvent, EpochChangeEvent, InvalidBlobId},
 };
 
 use self::{
@@ -502,31 +502,41 @@ impl StorageNode {
                     storage.update_blob_info(blob_event)?;
                 }
                 match stream_element.element {
-                    EventStreamElement::ContractEvent(ContractEvent::BlobCertified(event)) => {
+                    EventStreamElement::ContractEvent(ContractEvent::BlobEvent(
+                        BlobEvent::Certified(event),
+                    )) => {
                         self.process_blob_certified_event(element_index, event)
                             .await?;
                     }
-                    EventStreamElement::ContractEvent(ContractEvent::InvalidBlobID(event)) => {
+                    EventStreamElement::ContractEvent(ContractEvent::BlobEvent(
+                        BlobEvent::InvalidBlobID(event),
+                    )) => {
                         self.process_blob_invalid_event(element_index, event)
                             .await?;
                     }
-                    EventStreamElement::ContractEvent(ContractEvent::BlobRegistered(event)) => {
+                    EventStreamElement::ContractEvent(ContractEvent::BlobEvent(
+                        BlobEvent::Registered(event),
+                    )) => {
                         self.inner
                             .mark_event_completed(element_index, &event.event_id)?;
                     }
-                    EventStreamElement::ContractEvent(ContractEvent::EpochParametersSelected(
-                        event,
+                    EventStreamElement::ContractEvent(ContractEvent::EpochChangeEvent(
+                        EpochChangeEvent::EpochParametersSelected(event),
                     )) => {
                         tracing::info!("EpochParametersSelected event received: {:?}", event,);
                         self.inner
                             .mark_event_completed(element_index, &event.event_id)?;
                     }
-                    EventStreamElement::ContractEvent(ContractEvent::EpochChangeStart(event)) => {
+                    EventStreamElement::ContractEvent(ContractEvent::EpochChangeEvent(
+                        EpochChangeEvent::EpochChangeStart(event),
+                    )) => {
                         tracing::info!("EpochChangeStart event received: {:?}", event);
                         self.inner
                             .mark_event_completed(element_index, &event.event_id)?;
                     }
-                    EventStreamElement::ContractEvent(ContractEvent::EpochChangeDone(event)) => {
+                    EventStreamElement::ContractEvent(ContractEvent::EpochChangeEvent(
+                        EpochChangeEvent::EpochChangeDone(event),
+                    )) => {
                         tracing::info!("EpochChangeDone event received: {:?}", event);
                         self.inner
                             .mark_event_completed(element_index, &event.event_id)?;
@@ -1173,9 +1183,7 @@ mod tests {
     async fn services_slivers_for_shards_managed_according_to_committee() -> TestResult {
         let shard_for_node = ShardIndex(0);
         let node = StorageNodeHandle::builder()
-            .with_system_event_provider(vec![ContractEvent::BlobRegistered(
-                BlobRegistered::for_testing(BLOB_ID),
-            )])
+            .with_system_event_provider(vec![BlobRegistered::for_testing(BLOB_ID).into()])
             .with_shard_assignment(&[shard_for_node])
             .with_node_started(true)
             .build()
@@ -1208,9 +1216,7 @@ mod tests {
 
         tokio::time::sleep(Duration::from_millis(50)).await;
         assert!(storage.is_stored_at_all_shards(&BLOB_ID)?);
-        events.send(ContractEvent::InvalidBlobID(InvalidBlobId::for_testing(
-            BLOB_ID,
-        )))?;
+        events.send(InvalidBlobId::for_testing(BLOB_ID).into())?;
 
         tokio::time::sleep(Duration::from_millis(100)).await;
         assert!(!storage.is_stored_at_all_shards(&BLOB_ID)?);
@@ -1221,7 +1227,7 @@ mod tests {
     async fn returns_correct_blob_status() -> TestResult {
         let blob_event = BlobRegistered::for_testing(BLOB_ID);
         let node = StorageNodeHandle::builder()
-            .with_system_event_provider(vec![ContractEvent::BlobRegistered(blob_event.clone())])
+            .with_system_event_provider(vec![blob_event.clone().into()])
             .with_shard_assignment(&[ShardIndex(0)])
             .with_node_started(true)
             .build()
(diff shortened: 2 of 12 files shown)


crates/walrus-service/src/node/committee.rs
```diff
@@ -105,6 +105,9 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
     /// Returns the committee used by the service.
     fn committee(&self) -> &Committee;
 
+    /// Returns the previous committee used by the service.
+    fn previous_committee(&self) -> &Committee;
```
halfprice: This is a temporary change just to be able to execute initial epoch change. This entire struct and logic probably will be replaced by @jpcsmith 's change.

crates/walrus-service/src/node/config.rs
```diff
@@ -221,6 +221,8 @@ pub struct SuiConfig {
     pub system_object: ObjectID,
     /// Object ID of walrus staking object.
     pub staking_object: ObjectID,
+    /// Object ID of walrus storage node capability object.
+    pub storage_node_capability_object: Option<ObjectID>,
```
halfprice: we need this object to be able issue storage node related transaction on chain. This object is unique per storage node (owned by the wallet that registers the node)

crates/walrus-service/src/node/storage.rs
```diff
@@ -58,7 +58,7 @@ pub struct Storage {
     metadata: DBMap<BlobId, BlobMetadata>,
     blob_info: DBMap<BlobId, BlobInfo>,
     event_cursor: EventCursorTable,
-    shards: HashMap<ShardIndex, Arc<ShardStorage>>,
+    shards: DashMap<ShardIndex, Arc<ShardStorage>>,
```
halfprice: After using DashMap, we can have storage node still holds Arc<StorageNodeInner> instead of Arc<RwLock<StorageNodeInner>> even now we want to modify the shard map.

crates/walrus-event/src/lib.rs
```diff
@@ -102,12 +106,19 @@ pub struct IndexedStreamElement {
...
+    pub fn new(contract_event: ContractEvent, event_sequence_number: EventSequenceNumber) -> Self {
+        match contract_event {
+            ContractEvent::BlobEvent(blob_event) => Self {
+                element: EventStreamElement::BlobEvent(blob_event),
+                global_sequence_number: event_sequence_number,
+            },
+            ContractEvent::EpochChangeEvent(epoch_event) => Self {
+                element: EventStreamElement::EpochChangeEvent(epoch_event),
+                global_sequence_number: event_sequence_number,
+            },
```
jpcsmith: Suggestion: Just rename `EpochStreamElement::BlobEvent -> EpochStreamElement::ContractEvent` and have it store a `ContractEvent` instead of a `BlobEvent`.

crates/walrus-sui/src/types.rs
```diff
@@ -185,15 +195,6 @@ impl Committee {
...
 
-    /// Return the shards handed by the specified storage node,
-    /// based on its index in the committee list.
-    pub fn shards_for_node(&self, node_id: usize) -> Vec<ShardIndex> {
-        self.members
-            .get(node_id)
-            .map(|node| node.shard_ids.clone())
-            .unwrap_or_default()
-    }
-
```
jpcsmith, line 196: Question: Is this no longer used?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -83,6 +86,17 @@ pub trait ReadClient: Send + Sync {
...
+    /// Returns the current and previous committee.
+    fn current_and_previous_committee(
+        &self,
+    ) -> impl Future<Output = SuiClientResult<(Committee, Committee)>> + Send;
+
+    /// Returns the storage nodes and the number of shards in the committee.
+    fn get_committee_shard_assignment(
+        &self,
+        shard_assignment: &CommitteeShardAssignment,
+    ) -> impl Future<Output = SuiClientResult<(Vec<StorageNode>, u16)>> + Send;
```
jpcsmith: Suggestion: Make internal, this does not need to be a part of the trait.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -83,6 +86,17 @@ pub trait ReadClient: Send + Sync {
 
     /// Returns the current committee.
     fn current_committee(&self) -> impl Future<Output = SuiClientResult<Committee>> + Send;
+
+    /// Returns the current and previous committee.
+    fn current_and_previous_committee(
+        &self,
+    ) -> impl Future<Output = SuiClientResult<(Committee, Committee)>> + Send;
```
jpcsmith: Remark: We also need the upcoming committee.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -352,7 +369,44 @@ impl ReadClient for SuiReadClient {
...
+        )
+        .map_err(|err| SuiClientError::Internal(err.into()))?;
+
+        let (previous_committee_member, previous_n_shard) = self
+            .get_committee_shard_assignment(&staking_object.inner.previous_committee)
+            .await?;
+        let previous_committee = Committee::new(
+            previous_committee_member,
+            staking_object.inner.epoch - 1,
+            NonZeroU16::new(previous_n_shard).unwrap_or(NonZeroU16::new(1).unwrap()),
```
jpcsmith: Remark: The number of shards cannot change, so you should be able to use the number of shards from the staking object.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -105,13 +120,15 @@ impl SuiReadClient {
...
         staking_object_id: ObjectID,
+        storage_node_cap_id: Option<ObjectID>,
     ) -> SuiClientResult<Self> {
         let system_pkg_id = get_system_package_id(&sui_client, system_object_id).await?;
         Ok(Self {
             system_pkg_id,
             sui_client,
             system_object_id,
             staking_object_id,
+            storage_node_cap_id,
```
jpcsmith: Suggestion: Move to the contract client, as this is not used in the read client.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -105,4 +109,28 @@ where
...
+        utils::retry(backoff, || async {
+            self.contract_client
+                .lock()
+                .await
+                .epoch_sync_done()
+                .await
+                .inspect_err(|error| {
+                    tracing::error!(
+                        ?error,
+                        "submitting invalidity certificate to contract failed"
```
jpcsmith: Remark: Invalid error message.

crates/walrus-service/src/node/storage.rs
jpcsmith: Question: Doesn't this open us up to race conditions where the map is queried while it is being modified? It would perhaps be better to use a RwLock, as that would avoid any weird errors where it's being accessed while being edited for epoch change. The map will only be modified around once an epoch.

crates/walrus-service/src/node.rs
```diff
@@ -614,11 +625,57 @@ impl StorageNode {
...
+                // Initiate move shard!
+                todo!();
+            }
+        }
+
+        self.inner
+            .committee_service
+            .store(Arc::new(committee_service));
+
+        self.inner.contract_service.epoch_sync_done().await;
```
jpcsmith: Question: Shouldn't this be performed only after the sync is complete?

crates/walrus-service/src/node/storage.rs
halfprice: DashMap is implemented using RwLock<HashMap<>>, but more performant (from the [doc](https://docs.rs/dashmap/latest/dashmap/struct.DashMap.html), DashMap tries to be very simple to use and to be a direct replacement for RwLock<HashMap<K, V>>). Any access to it requires internal locking, so it should provide the same property as using RwLock<HashMap>.

Another thing is that to change shards to RwLock<HashMap>, we need to change StorageInner and Storage from Arc<> to Arc<RwLock<>> as well, and that change itself can be huge (and we don't gain much).

crates/walrus-event/src/lib.rs
halfprice: SG. I was debating about this, and I saw that there are places where we only want to take BlobEvent, so I didn't do it in the first run. But I think your suggestion is cleaner.

crates/walrus-sui/src/types.rs
halfprice: Yes, not used anywhere even before this PR.

crates/walrus-sui/src/client/read_client.rs
halfprice: I didn't add it because this PR doesn't handle sync shard yet. Will be adding one when we need it.

crates/walrus-sui/src/client/read_client.rs
halfprice: This field is removed, and in the new revision, we get the storage capability object by querying the chain.

crates/walrus-service/src/node.rs
halfprice: This PR doesn't handle sync yet. For that to work, we need the committee service to be able to talk to multiple epochs committee.

This PR just try to finish the epoch change from 0 -> 1, in which epoch 0 doesn't have any shards assigned.

crates/walrus-sui/src/types/events.rs
jpcsmith, line 1: Suggestion: It would be worth keeping the distinction between Blob and Epoch change events at this level. So a 
```rust
enum ContractEvent {
    EpochChange(EpochChangeEvent),
    Blob(BlobEvent)
}
```
The `IndexedStreamElement` just was not the place for the division.

I see now where the misunderstanding arose, could you perhaps revert this unification by checking back out the earlier version from your local reflog?

crates/walrus-event/src/lib.rs
jpcsmith: There is a misunderstanding, the `ContractEvent::BlobEvent` and `ContractEvent::EpochChangeEvent` are good and should remain as is. But you do not need to a *also* have a `EventStreamElement::EpochChangeEvent` and a `EventStreamElement::BlobEvent`, you can just have a single `EventStreamElement::ContractEvent`.

Then, someone can match `EventStreamElement::ContractEvent(ContractEvent::BlobEvent(_))` or `EventStreamElement::ContractEvent(ContractEvent::EpochChangeEvent(_))` if they care about the difference, or `EventStreamElement::ContractEvent(_)` if they do no not.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -67,6 +67,7 @@ pub(crate) trait BlobInfoApi {
 
 #[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
 pub(super) enum BlobInfoMergeOperand {
+    None,
```
jpcsmith: Suggestion: In cases like this, implementing `TryFrom` is better than adding a `None` variant. But see the other comment regarding the misunderstanding.

crates/walrus-service/src/node/storage.rs
jpcsmith: The concern is not so much DashMap vs RwLock, it's where the locking happens. It's the difference between
```rust
{
  let shards = self.shards.write().unwrap();
  shards.insert(ShardIndex(0), ...);
  shards.insert(ShardIndex(1), ...);
  shards.insert(ShardIndex(2), ...);
}
```
and 
```rust
shards.write().unwrap().insert(ShardIndex(0), ...);  // or DashMap::insert(ShardIndex(0), ...)
// (1) Code from other threads could do shards.read() or DashMap::get here
shards.write().unwrap().insert(ShardIndex(1), ...);  // or DashMap::insert(ShardIndex(1), ...)
// (2) Code from other threads could do shards.read() or DashMap::get here
shards.write().unwrap().insert(ShardIndex(2), ...);  // or DashMap::insert(ShardIndex(2), ...)
```
In the former, while our shard map is in "not-fully-complete" state, it is not accessible. In the latter, it's possible that other operations interleave with our changes to the shard map, resulting in behaviour such as a caller iterating over the shards map at point `(1)` and thinking that it is only responsible for a single shard. See, for example, `compute_storage_confirmation` which calls `is_stored_at_all_shards`. That function could incorrectly return a storage confirmation signature due the above race condition.

Therefore, it would be good if we could ensure that updates are handled all together, and that the node moves from one complete view of its shards to another complete view, without the risk of race conditions. If DashMap can provide something akin to the first example and solves any other problems with the implementation, then we can pull in the dependency, otherwise, clear synchronisation would be preferred. 

We can, however, discuss this further on the follow up PR.

crates/walrus-sui/src/types/events.rs
jpcsmith, line 1: Remark: Something seems to have removed many spaces from this file.

crates/walrus-service/src/node/storage.rs
halfprice: Excellent point and I totally agree with the concern. I was assuming that dashmap should be able to provide atomic inserts of multiple key values (since Sui also uses dashmap and there was a place where atomic inserts are required). However, reading its implementation a bit, I'm not sure if it can actually provide that. It appears that it has an internal sharding, and only provide atomic inserts into the same shard. So let me dig into this a bit.

With that being said, I think my previous replies are also not clear. The original choice of using DashMap is that we can still use Arc<StorageNodeInner> in StorageNode struct. If we want to remain to use HashMap, we have to change Arc<StorageNodeInner> to Arc<RwLock<StorageNodeInner>> and that change itself is a lot. I'm currently trying to see if we can have Arc<RwLock<DashMap<>>> to have a clear atomic update. Will update you in the follow up PR.

crates/walrus-event/src/lib.rs
halfprice: Done.

crates/walrus-sui/src/types/events.rs
halfprice: Good catch. Probably due to my copying from an earlier commit...

PR #799, opened 2024-09-12
feat(client): reuse existing storage resources

### LINKED ISSUE #754: Support using existing `Storage` resources to store a new blob.

giac-mysten:
Allows the client to reuse storage resources that are unused in its wallet.

When storing a blob, the client checks if any unused storage resources are present. 
If there are resources with enough storage and valid for enough epochs are found, the one with the _smallest end epoch_ among the ones with the _smallest storage size_ is chosen, in order to get the tightest possible fit to the desired store operation.

In other words, the client:
* filters all storage resources for which `storage_size >= target_size` and `end_epoch >= target_end_epoch`
* sorts by `storage_size` (increasing), breaking ties by looking at `end_epoch` (smallest wins)
* picks the first in the sorting order.

Closes #754

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -404,17 +404,6 @@ impl SuiContractClient {
 
         Ok(reserve_result_index)
     }
-
-    /// Computes the epoch for which to filter a list of objects.
-    ///
-    /// If `include_expired` is `true`, returns `0` to include all objects.
-    async fn epoch_for_filter(&self, include_expired: bool) -> Result<Epoch> {
-        Ok(if !include_expired {
-            self.read_client.current_committee().await?.epoch
-        } else {
-            0
-        })
-    }
 }
 
 impl ContractClient for SuiContractClient {
@@ -462,13 +451,12 @@ impl ContractClient for SuiContractClient {
         let price = self
             .write_price_for_encoded_length(storage.storage_size)
             .await?;
-        let asdf = self.read_client.get_object_ref(storage.id).await?.into();
         let res = self
             .move_call_and_transfer(
                 contracts::system::register_blob,
                 vec![
                     self.read_client.call_arg_from_system_obj(true).await?,
-                    asdf,
+                    self.read_client.get_object_ref(storage.id).await?.into(),
                     call_arg_pure!(&blob_id),
                     call_arg_pure!(&root_digest),
                     blob_size.into(),
@@ -730,8 +718,7 @@ impl ContractClient for SuiContractClient {
     }
 
     async fn owned_blobs(&self, include_expired: bool) -> SuiClientResult<Vec<Blob>> {
-        let epoch_limit = self.epoch_for_filter(include_expired).await?;
-
+        let current_epoch = self.read_client.current_committee().await?.epoch;
         Ok(get_owned_objects::<Blob>(
             &self.read_client.sui_client,
             self.wallet_address,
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -430,12 +462,13 @@ impl ContractClient for SuiContractClient {
         let price = self
             .write_price_for_encoded_length(storage.storage_size)
             .await?;
+        let asdf = self.read_client.get_object_ref(storage.id).await?.into();
```
jpcsmith: Question: What's asdf?

crates/walrus-sui/src/client.rs
```diff
@@ -710,9 +739,50 @@ impl ContractClient for SuiContractClient {
             &[],
         )
         .await?
-        .filter(|blob| include_expired || blob.storage.end_epoch > current_epoch)
+        .filter(|blob| include_expired || blob.storage.end_epoch > epoch_limit)
```
jpcsmith: Remark: Since epoch_limit is at least zero, is end_epoch always at least 1? (I suppose yes given our special zero-th epoch?) Would it be worth it to define a constant for epoch zero (like `GENESIS_EPOCH`?) in the code to make the fact that it's a special case clear?

crates/walrus-sui/src/client.rs
```diff
@@ -710,9 +739,50 @@ impl ContractClient for SuiContractClient {
...
+            // Sort in descending order, so we can `pop` the best fit as the last element.
+            let size_cmp = b.storage_size.cmp(&a.storage_size);
+            // Break ties by comparing the end epoch, and take the one that is the closest to
+            // `end_epoch`. NOTE: we are already sure that these values are above the minimum.
+            if size_cmp == std::cmp::Ordering::Equal {
+                b.end_epoch.cmp(&a.end_epoch)
+            } else {
+                size_cmp
+            }
+        });
```
jpcsmith: Nit: This could be done with `reduce` instead of `collect + sort`.

crates/walrus-sui/src/client.rs
giac-mysten: 😬 forgot to rename the variable sorry.

crates/walrus-sui/src/client.rs
giac-mysten: >Since epoch_limit is at least zero, is end_epoch always at least 1?

yes, because it is the exclusive index of the end epoch (so a blob stored for 1 epoch at epoch 2 ends at epoch 3).

>Would it be worth it to define a constant for epoch zero (like GENESIS_EPOCH?) in the code to make the fact that it's a special case clear?

Probably, but is it needed here?

jpcsmith: No, but I did not understand why this removed all storage resources for epoch zero, as that would be a valid storage resource for a blob being registered in epoch zero. I then understood that because of how we do genesis, that's not possible. But this nuance is not visible in code.

crates/walrus-sui/src/client.rs
giac-mysten: amazing, let me add that!

crates/walrus-sui/src/client.rs
giac-mysten: I see.

I think there is a more subtle point I think. Returning `0` is not really elegant, because `include_expired` is anyway true and then short circuits. I'll think if there is a better way to avoid the confusing `0`.

crates/walrus-sui/src/client.rs
giac-mysten: actually, @amnn just suggested `min_by_key`. even nicer :)

crates/walrus-sui/src/client.rs
giac-mysten: Ah i realize now this is useless. Completely removed.

PR #803, opened 2024-09-13
feat: implement multi-epoch committee service

jpcsmith:
Modifies the `CommitteeService` in `node/committee/` to restart requests on epoch change, and removes the `CommitteeServiceFactory`.

TODO:
- [x] Add tests

### DIFF
--- a/crates/walrus-service/src/node/committee.rs
+++ b/crates/walrus-service/src/node/committee.rs
@@ -46,13 +46,13 @@ use super::errors::SyncShardClientError;
 pub(crate) type DefaultRecoverySymbol = walrus_core::RecoverySymbol<MerkleProof>;
 
 /// Service used to query the current, prior, and upcoming committees.
-#[async_trait::async_trait]
+#[async_trait]
 pub(crate) trait CommitteeLookupService: Send + Sync + std::fmt::Debug {
     /// Returns the active committees, which are possibly already transitioning.
     async fn get_active_committees(&self) -> Result<ActiveCommittees, anyhow::Error>;
 }
 
-#[async_trait::async_trait]
+#[async_trait]
 impl<T: ReadClient + std::fmt::Debug> CommitteeLookupService for T {
     async fn get_active_committees(&self) -> Result<ActiveCommittees, anyhow::Error> {
         let committee = self
@@ -137,7 +137,7 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
 }
 
 /// Interface for creating new [`NodeService`]s, such as during epoch change.
-#[async_trait::async_trait]
+#[async_trait]
 pub(crate) trait NodeServiceFactory: Send {
     type Service: NodeService;
 
@@ -148,7 +148,7 @@ pub(crate) trait NodeServiceFactory: Send {
     ) -> Result<Self::Service, ClientBuildError>;
 }
 
-#[async_trait::async_trait]
+#[async_trait]
 impl<F, S, Fut> NodeServiceFactory for F
 where
     F: for<'a> FnMut(&'a StorageNode, &'a Arc<EncodingConfig>) -> Fut + Send,
--- a/crates/walrus-service/src/node/committee/request_futures.rs
+++ b/crates/walrus-service/src/node/committee/request_futures.rs
@@ -598,7 +598,14 @@ where
     }
 }
 
-type StoredFuture<'a> = BoxFuture<'a, (usize, u16, Option<InvalidBlobIdAttestation>)>;
+type RequestWeight = u16;
+type NodeIndexInCommittee = usize;
+type AttestationWithWeight = (
+    NodeIndexInCommittee,
+    RequestWeight,
+    Option<InvalidBlobIdAttestation>,
+);
+type StoredFuture<'a> = BoxFuture<'a, AttestationWithWeight>;
 
 #[pin_project::pin_project]
 struct PendingInvalidBlobAttestations<'fut, 'iter, T> {
@@ -724,7 +731,7 @@ where
         }
     }
 
-    fn next_request(&mut self) -> Option<(u16, StoredFuture<'fut>)> {
+    fn next_request(&mut self) -> Option<(RequestWeight, StoredFuture<'fut>)> {
         for &index in self.nodes.by_ref() {
             let committee_epoch = self.committee.epoch;
             let node_info = &self.committee.members()[index];


crates/walrus-test-utils/src/lib.rs
```diff
@@ -163,6 +163,15 @@ macro_rules! async_param_test {
         async_param_test!(
             $func_name -> $return_ty: [ $( #[tokio::test] $case_name: ( $($args),+ ) ),* ]
         );
+    };
+    (#[tokio::test(start_paused = true)] $func_name:ident -> $return_ty:ty: [
```
giac-mysten, line 217: Question: why isn't the generic `#[$outer:meta]` arm above enough.

Cargo.toml
```diff
@@ -74,7 +74,7 @@ thiserror = "1.0.63"
 tokio = "=1.36.0"
 tokio-stream = "0.1.16"
 tokio-util = "0.7.12"
-tower = "0.5.1"
+tower = "0.5"
```
giac-mysten, line 76: Question: is the downgrade wanted?

crates/walrus-service/src/node/committee/active_committees.rs
```diff
@@ -0,0 +1,294 @@
...
+    ///
+    /// The epoch of the system corresponds to this committee.
+    current_committee: Arc<Committee>,
+
+    /// The prior committee.
+    ///
+    /// When `is_transitioning` is true, this committee continues to serve reads. It is only `None`
+    /// when the current epoch is zero.
+    // INV!: current_committee.epoch == prior_committee.epoch + 1
+    // INV!: current_committee.epoch == 0 || prior_committee.is_some()
```
giac-mysten: Nit: in the rest of the codebase we use `INV` without `!`. 

Question: is this notation just for develpers?

crates/walrus-service/src/node/committee/active_committees.rs
```diff
@@ -0,0 +1,294 @@
...
+/// [`ActiveCommittees::begin_committee_change`].
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub(crate) enum BeginCommitteeChangeError {
+    /// Error returned when attempting to start a committee change while one is already in progress.
+    #[error("the committees are already changing")]
+    ChangeInProgress,
+    /// Error returned if the expected next committee has an epoch that is not 1 greater than the
+    /// current epoch.
+    #[error("the epoch of the new committee is invalid")]
+    InvalidEpoch,
```
giac-mysten: Nit: would it make sense to add the epochs?
```suggestion
    InvalidEpoch { current: u64, new: u64 },
```

crates/walrus-service/src/node/committee/active_committees.rs
```diff
@@ -0,0 +1,294 @@
...
+    CommitteeMismatch,
+}
+
+/// Errors returned when completing a committee change with
+/// [`ActiveCommittees::end_committee_change`].
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub(crate) enum EndCommitteeChangeError {
+    /// The current committee change that is in progress was not for the specified epoch.
+    #[error("the epoch provided does not match the new epoch")]
+    InvalidEpoch,
```
giac-mysten: Nit: add the epoch here and afterwards too?

crates/walrus-service/src/node/committee/active_committees.rs
```diff
@@ -0,0 +1,294 @@
...
+            }
+            Ordering::Equal => Some(&self.current_committee),
+            Ordering::Greater => None,
+        }
+    }
+
+    /// Returns the committee for the specified epoch.
+    ///
+    /// If the epoch is not the current, prior, or known next epoch, then None is returned.
+    pub fn committee_for_epoch(&self, epoch: Epoch) -> Option<&Arc<Committee>> {
```
giac-mysten, line 172: Nit: maybe
```suggestion
    pub fn committee_for_neighboring_epoch(&self, epoch: Epoch) -> Option<&Arc<Committee>> {
```

crates/walrus-service/src/node/committee/active_committees.rs
```diff
@@ -0,0 +1,294 @@
...
+        self.check_invariants();
+        Ok(())
+    }
+
+    /// Begins the transition of committees that occurs during epoch change.
+    ///
+    /// If the upcoming committee has not been set, then the expected committee defines the upcoming
+    /// committee. Otherwise, the previously set upcoming committee is compared with the expected
+    /// committee. If they are the same, then the change can be initiated, otherwise an error is
+    /// returned.
```
giac-mysten, line 217: General question: is there any reasonable instance in which they may not be the same? And what can we do in that case?

crates/walrus-service/src/node/committee/active_committees.rs
```diff
@@ -0,0 +1,294 @@
...
+    AlreadySet,
+    /// Error returned when the epoch of the new committee is not 1 greater than that of the
+    /// current committee.
+    #[error("the epoch of the new committee is invalid")]
+    InvalidEpoch,
+}
+
+/// The current, prior, and next committees in the system.
+#[derive(Debug, Clone)]
+pub(crate) struct ActiveCommittees {
```
giac-mysten, line 53: General comment: A subset of the functionality of `ActiveCommittees` will also be needed on the storage client side (importantly, the committee to read and the committee to write blobs). Maybe in a follow up we could separate this functionality out in a way that it usable from the client as well?

crates/walrus-service/src/node/committee.rs
```diff
@@ -155,742 +136,35 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
...
-                    )
-                    .await
-            }
-        }
+    async fn make_service(
+        &mut self,
+        info: &StorageNode,
+        encoding_config: &Arc<EncodingConfig>,
+    ) -> Result<Self::Service, ClientBuildError> {
+        (self)(info, encoding_config).await
```
giac-mysten, line 165: Took me a while to parse this 😅

crates/walrus-service/src/node/committee/node_service.rs
```diff
@@ -0,0 +1,277 @@
...
+    InvalidBlobAttestation(InvalidBlobIdAttestation),
+    ShardSlivers(Vec<(BlobId, Sliver)>),
+}
+
+impl Response {
+    /// Convert a response to its inner type and panic if there is a mismatch.
+    ///
+    /// As responses correspond to the request, this should never panic except in the case of
+    /// programmer error.
+    pub fn into_value<T>(self) -> T
```
giac-mysten, line 93: Question: why both this `into_value` function and the `TryFrom` implementations below?

crates/walrus-service/src/node/committee/node_service.rs
```diff
@@ -0,0 +1,277 @@
...
+
+impl_response_conversion!(VerifiedBlobMetadataWithId, Response::VerifiedMetadata);
+impl_response_conversion!(DefaultRecoverySymbol, Response::VerifiedRecoverySymbol);
+impl_response_conversion!(InvalidBlobIdAttestation, Response::InvalidBlobAttestation);
+impl_response_conversion!(Vec<(BlobId, Sliver)>, Response::ShardSlivers);
+
+/// Marker trait for types implementing the [`tower::Service`] signature expected of services
+/// used for communication with the committee.
+pub(crate) trait NodeService:
+    for<'a> Service<Request<'a>, Response = Response, Error = NodeError> + Clone
```
giac-mysten: didn't know about HRTB either 😳

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -0,0 +1,778 @@
...
+        let mut committee_listener = self.shared.subscribe_to_committee_changes();
+
+        loop {
+            let (n_members, weak_committee) = {
+                let active_committees = committee_listener.borrow_and_update();
+                let committee = active_committees
+                    .read_committee(self.epoch_certified)
+                    .expect("epoch must not be in the future");
+
+                (committee.n_members(), Arc::downgrade(committee))
```
giac-mysten, line 93: Question: why the downgrade? in general, what is the use of `Weak` pointers? Is it to avoid cyclic Rc?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -0,0 +1,778 @@
...
+                shared.config.retry_interval_max,
+            ),
+            shared,
+        }
+    }
+
+    pub async fn run(mut self) -> Result<Sliver, InconsistencyProofEnum> {
+        // Since recovery currently consumes the symbols, rather than copy the symbols in every
+        // case to handle the rare cases when we fail to recovery the sliver with the default number
+        // of symbols, we instead retry the entire process with an increased symbol count.
```
giac-mysten: Very interesting!
Question: is the idea that the memory overhead is not worth it in the average case? The network overhead for each failure is pretty annoying though? also, one byzantine storage node can force a full refetch?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -0,0 +1,778 @@
...
+                    panic!("checked above by pre-condition")
+                }
+                SliverVerificationError::SliverSizeMismatch
+                | SliverVerificationError::SymbolSizeMismatch => panic!(
+                    "should not occur since symbols were verified and sliver constructed here"
+                ),
+                SliverVerificationError::MerkleRootMismatch => {
+                    panic!("should have been converted to an inconsistency proof")
+                }
+                SliverVerificationError::RecoveryFailed(_) => todo!("what generates this?"),
```
giac-mysten, line 469: https://github.com/MystenLabs/walrus/blob/78b2c7b8b8432b9e351ece60f5142447993d962a/crates/walrus-core/src/encoding/slivers.rs#L126

this line IIUC

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -0,0 +1,778 @@
...
+                                walrus.node.public_key = %node_info.public_key
+                            ));
+                    Some((request, node_weight))
+                })
+                .fuse();
+
+            let mut pending_requests = FuturesUnordered::default();
+            let mut pending_weight: u16 = 0;
+
+            'refill_pending_requests: loop {
```
giac-mysten: Labelled loops 🤯

crates/walrus-test-utils/src/lib.rs
jpcsmith: Because that is per test-case, as opposed to being applied to all cases. Since I started adding `start_paused = true` to all of my test cases, I added this as a shortcut to make things easier.

In either case, I added a more general implementation in #828, and will rebase onto that after the review.

Cargo.toml
jpcsmith: Yes, the relaxation of the patch release was intentional, as I felt there was no need to over-specify the version (which makes it harder for cargo to find dependencies that satisfy all packages).

crates/walrus-service/src/node/committee/active_committees.rs
jpcsmith: Yes, it's for developers to not violate the invariants on the internal state when making changes to the struct.

crates/walrus-service/src/node/committee/active_committees.rs
jpcsmith: Yes, because we fetch the current committee from chain when we process an event signalling that epoch change has started, but we may be late in processing that event.

With our epochs of 1 month this may not occur in practice; however, with an epoch of an hour or so during testing, we may have the case where a storage node comes online after being offline long enough to miss multiple epoch changes.

The `NodeCommitteeService` has a method that will just replace the `ActiveCommittee` altogether in the event that that occurs.

crates/walrus-service/src/node/committee/node_service.rs
jpcsmith: The `into_value` panics, avoiding the need to `expect` everywhere. Since responses should always match the requests (as this is checked in the client), it is generally safe to unwrap here.

crates/walrus-service/src/node/committee/node_service.rs
jpcsmith: Then crawl back into ignorance with me. I removed them as they were complicating things with implementations of associated types that implement `NodeService`. Now `Request` just keeps a copy of all it's arguments.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: It's to add an additional signal. In the event that the committee is dropped it becomes impossible to create a request to members of the committee, which saves some requests which would likely fail anyways.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: I updated this comment to read 
```suggestion
        // Since recovery currently consumes the symbols, rather than copy the symbols in every
        // case to handle the rare cases when we fail to *decode* the sliver despite collecting the
        // required number of symbols, we instead retry the entire process with an increased amount.
```
This only occurs if the the number of symbols required for recovery, including the safety limit* that you added of a few symbols, fails the RaptorQ decoding. In this case we try again with 1 more symbol. Network failures and byzantine nodes are dealt with the inner retry mechanisms.

*I assume that `encoding::symbols::min_symbols_for_recovery` accounts for this.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: That seems to arrive at IndexTooLarge (which is not possible) or EncodeError which is MisalignedData or InvalidDataSize. 

The MisalignedData I assume should not happen since we verify the data, and InvalidDataSize happens on too large or empty data, both which shouldn't occur I believe. I will change this to a panic.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: Thankfully removed now.

crates/walrus-service/src/node/committee/request_futures.rs
giac-mysten: mmm I see, makes sense. Looking at `min_symbols_for_recovery` though, there seems to be a small problem in there?

https://github.com/MystenLabs/walrus/blob/62600f806469352b6c8aa9ee7bbbf9d38d5204b8/crates/walrus-core/src/encoding/symbols.rs#L474-L482

Here the computation is using the `f+1, 2f+1` limits, while we use `n-2f,n-f` for the encoding.
I guess this function should internally call `encoding::config::source_symbols_for_n_shards` to get the right minimum amount of symbols? in practice these numbers were always the same, so we didn't see any problems. However, for `n>3f+1` we would eat into the safety margin, which could cause failures in recovery for small number of shards.

@mlegner a quick look here? I made a quick fix PR in case you agree: #830.

crates/walrus-service/src/node/committee/node_service.rs
giac-mysten: 😬 I see

crates/walrus-service/src/node/committee/active_committees.rs
giac-mysten: makes sense, thanks!

crates/walrus-service/src/node/committee/request_futures.rs
giac-mysten: >The MisalignedData I assume should not happen since we verify the data

I think you're right, verified slivers should not be misaligned. This works!

crates/walrus-service/src/node/committee.rs
```diff
@@ -51,43 +22,58 @@ use walrus_core::{
...
-    config: &CommitteeServiceConfig,
-) -> ExponentialBackoff<StdRng> {
-    ExponentialBackoff::new_with_seed(
-        config.retry_interval_min,
-        config.retry_interval_max,
-        None,
-        seed,
-    )
+/// Service used to query the current, prior, and upcoming committees.
+#[async_trait::async_trait]
```
giac-mysten: Nit: you are already importing `async_trait`
```suggestion
#[async_trait]
```

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -0,0 +1,901 @@
...
+                panic!("attestations must be verified beforehand: {:?}", err)
+            }
+            Err(CertificateError::MessageMismatch) => {
+                panic!("messages must be verified against the same epoch and blob id")
+            }
+        }
+    }
+}
+
+type StoredFuture<'a> = BoxFuture<'a, (usize, u16, Option<InvalidBlobIdAttestation>)>;
```
giac-mysten: Question: what are the implications of using a `BoxFuture`? I guess we are not creating many of these?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -653,6 +592,177 @@ where
...
+    #[pin]
+    pending_requests: FuturesUnordered<StoredFuture<'fut>>,
+}
+
+impl<'fut, 'iter, T> std::future::Future for PendingInvalidBlobAttestations<'fut, 'iter, T>
+where
+    T: NodeService + 'fut,
+{
+    type Output =
+        Result<HashMap<usize, InvalidBlobIdAttestation>, HashMap<usize, InvalidBlobIdAttestation>>;
```
giac-mysten, line 640: Interesting, using a non error type as error return 🤔

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -653,6 +592,177 @@ where
...
+        };
+        this.refill_pending_requests();
+        this
+    }
+
+    fn refill_pending_requests(&mut self) {
+        while self.pending_weight < self.required_weight {
+            let Some((weight, request)) = self.next_request() else {
+                tracing::trace!("no more requests to dispatch");
+                break;
```
giac-mysten, line 727: Question: In this case, does this not mean that we will never reach the required weight? Should this error out so that we can also terminate the caller quickly?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -653,6 +592,177 @@ where
     }
 }
 
+type StoredFuture<'a> = BoxFuture<'a, (usize, u16, Option<InvalidBlobIdAttestation>)>;
```
giac-mysten: Nit: type-alias the index and weight?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -653,6 +592,177 @@ where
...
+
+#[pin_project::pin_project]
+struct PendingInvalidBlobAttestations<'fut, 'iter, T> {
```
giac-mysten, line 611: I like this!

crates/walrus-service/src/node/committee/active_committees.rs
```diff
@@ -97,6 +97,24 @@ impl ActiveCommittees {
...
     }
 
+    /// Construct a new set of `ActiveCommittees`, which are transitioning to the current epoch.
+    ///
+    /// # Panics
+    ///
+    /// Panics if the prior committee's epoch does not precede that of the current committees, or if
+    /// they have a different number of shards.
+    #[cfg(test)]
+    pub fn new_transitioning(current_committee: Committee, prior_committee: Committee) -> Self {
```
giac-mysten, line 107: question: does this need to be in the struct impl?

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -0,0 +1,632 @@
...
+
+/// Default committee service used for communicating between nodes.
+///
+/// Requests the current committee state using a [`CommitteeLookupService`].
+pub(crate) struct NodeCommitteeService<T = RemoteStorageNode> {
+    inner: NodeCommitteeServiceInner<T>,
+    committee_lookup: Box<dyn super::CommitteeLookupService>,
+}
+
+impl NodeCommitteeService<RemoteStorageNode> {
```
giac-mysten, line 149: Nit: equivalent? or do you want it there to make it explicit?
```suggestion
impl NodeCommitteeService {
```

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -0,0 +1,632 @@
...
+    pub fn builder() -> NodeCommitteeServiceBuilder<RemoteStorageNode> {
```
giac-mysten, line 150: Nit: same as above?
```suggestion
    pub fn builder() -> NodeCommitteeServiceBuilder {
```

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -305,7 +378,7 @@ pub(super) struct NodeCommitteeServiceInner<T> {
     /// The set of active committees, which can be observed for changes.
     pub active_committees: watch::Sender<ActiveCommittees>,
     /// Services for members of the active read and write committees.
-    pub services: SyncRwLock<HashMap<PublicKey, T>>,
+    pub services: SyncMutex<HashMap<PublicKey, T>>,
```
giac-mysten, line 381: Question: why the change?

crates/walrus-service/src/node/committee/test_committee_service.rs
```diff
@@ -0,0 +1,371 @@
...
+        Ok(tower::service_fn(|_request| std::future::pending()).boxed_clone())
+    }
+}
+
+#[async_trait::async_trait]
+impl CommitteeLookupService for ActiveCommittees {
+    async fn get_active_committees(&self) -> Result<ActiveCommittees, anyhow::Error> {
+        Ok(self.clone())
+    }
+}
```
giac-mysten: Question: is this only used in tests?

crates/walrus-service/src/node/committee/test_committee_service.rs
```diff
@@ -0,0 +1,371 @@
...
+        &mut pending_request,
+        "must timeout since the new committee is unavailable for reads"
+    );
+
+    handle.finish_transition();
+    committee_service.end_committee_change(NEW_EPOCH)?;
+
+    let returned_metadata = time::timeout(Duration::from_secs(60), &mut pending_request)
+        .await
+        .expect("request must succeed since new committee has metadata");
```
giac-mysten, line 307: Nice!

crates/walrus-service/src/node/committee/test_committee_service.rs
```diff
@@ -0,0 +1,687 @@
...
+    let mut pending_request = std::pin::pin!(pending_request);
+
+    assert_timeout!(
+        &mut pending_request,
+        "must timeout since insufficient services in the current committee respond"
+    );
+
+    committee_handle.begin_transition_to(next_committee);
+    committee_service.begin_committee_change().await?;
+
```
giac-mysten, line 615: Question: how do we see that the collection has restarted here, since we have a timeout?

crates/walrus-service/src/node/committee/test_committee_service.rs
```diff
@@ -0,0 +1,687 @@
...
+    assert_timeout!(
+        &mut pending_request,
+        "must timeout since insufficient services in the current committee respond"
+    );
+
+    Ok(())
+}
+
+#[tokio::test(start_paused = true)]
+async fn collects_inconsistency_proof_despite_epoch_change() -> TestResult {
```
giac-mysten, line 625: Remark: i guess some code could be deduplicated between this and the previous test.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: For a single request possibly 2f+1 or more if each storage node has 1 shard. It's an allocation and indirection, but the overhead is small and dwarfed by the actual request. If/when rust allows naming `async` futures, we can get rid of it.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: No, it just means we've sent a request to every member of the committee. Those outstanding requests may still succeed with enough attestations.

crates/walrus-service/src/node/committee/active_committees.rs
jpcsmith: We need to be able to create a new committee that is transitioning when we fetch the state from the chain, as the committee may currently be transitioning.

crates/walrus-service/src/node/committee/committee_service.rs
jpcsmith: The desire was to make it explicit.

crates/walrus-service/src/node/committee/committee_service.rs
jpcsmith: A `Mutex<T: Send>` is `Sync`, it's contained value is now safe to be accessed simultaneously on multiple threads. A `RwLock<T>` is only `Sync` for `T: Sync`, since it allows handing out multiple (shared) references to the inner type simultaneous. 

To use a RwLock here, `NodeService` would need to be required to be `Sync`. I did not want to restrict it further until we observe a bottleneck or assure ourselves that it won't cause issues when we start layering T.

crates/walrus-service/src/node/committee/test_committee_service.rs
jpcsmith: No, it is/will be also implemented by every ReadClient. But as for many of the service dependencies that we have, there is one type that we will likely, consistently implement this with.

crates/walrus-service/src/node/committee/test_committee_service.rs
jpcsmith: Yes, otherwise the combined attestations from nodes in both committees would be sufficient to satisfy the request.

Issue #752, opened 2024-08-30
Client support for blob deletion


mlegner:
Our client should have some support for deletable blobs:
- [x] Flag for `store` to create a deletable blob.
- [x] New `delete` command.

PR #804, opened 2024-09-13
feat(client): add blob deletion

giac-mysten:
Adds the ability to delete owned deletable blobs with the `delete` command.

This command takes as input:
- a blob ID or a file: The client will delete all deletable and owned blob objects with matching blob ID.
- an object ID: The client will delete the blob object of that object ID, if it is deletable.

(also adds the `delete_blob` function to `ContractClient`).

Closes: #752

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -374,7 +374,6 @@ impl<T: ContractClient> Client<T> {
             .owned_blobs(false)
             .await?
             .into_iter()
-            // TODO(giac): do we have to filter non-registered blobs as well? Leaning towards not.
             .filter(|blob| blob.blob_id == *blob_id && blob.deletable))
     }
 


crates/walrus-e2e-tests/tests/test_client.rs
jpcsmith, line 1: Remark: Given the use of multiple booleans in these arguments, which become rather difficult to follow, it may be worthwhile to employ [the recommendation](https://rust-lang.github.io/api-guidelines/type-safety.html#arguments-convey-meaning-through-types-not-bool-or-option-c-custom-type) from the Rust API guidelines to instead use custom types. These could be `Persistence::Permanent/Deletable` and `Force::Yes/No` or similar.

crates/walrus-service/src/client.rs
```diff
@@ -317,8 +358,43 @@ impl<T: ContractClient> Client<T> {
...
+    pub async fn deletable_blobs_by_id<'a>(
+        &self,
+        blob_id: &'a BlobId,
+    ) -> ClientResult<impl Iterator<Item = Blob> + 'a> {
+        Ok(self
+            .sui_client
+            .owned_blobs(false)
+            .await?
+            .into_iter()
+            // TODO(giac): do we have to filter non-registered blobs as well? Leaning towards not.
```
jpcsmith: Question: Do you mean non-certified?

crates/walrus-sui/src/client.rs
```diff
@@ -430,12 +468,13 @@ impl ContractClient for SuiContractClient {
         let price = self
             .write_price_for_encoded_length(storage.storage_size)
             .await?;
+        let asdf = self.read_client.get_object_ref(storage.id).await?.into();
```
jpcsmith: Remark: You and your `asdf`s...

crates/walrus-service/src/client/responses.rs
```diff
@@ -362,3 +364,33 @@ impl InfoOutput {
         })
     }
 }
+
+#[serde_as]
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub(crate) struct DeleteOutput {
+    #[serde(skip_serializing_if = "std::option::Option::is_none")]
+    #[serde_as(as = "NoneAsEmptyString")]
```
mlegner: Question: Do we actually need this if we anyway skip it in this case?

crates/walrus-service/src/client/responses.rs
```diff
@@ -362,3 +364,33 @@ impl InfoOutput {
...
+        deleted_blobs: Vec<Blob>,
+    ) -> Self {
+        Self {
+            blob_id,
+            file,
+            object_id,
+            deleted_blobs,
+        }
+    }
+}
```
mlegner: Suggestion: Remove this constructor as all the fields have the same visibility as the overall type, and we don't enforce any invariants.

crates/walrus-service/src/client.rs
```diff
@@ -358,6 +361,40 @@ impl<T: ContractClient> Client<T> {
                     && blob.deletable == deletable
             }))
     }
+
+    // Blob deletion
+
+    /// Returns an iterator over the list of lobs that can be deleted, based on the Blob ID.
```
```suggestion
    /// Returns an iterator over the list of blobs that can be deleted, based on the blob ID.
```

crates/walrus-service/src/client.rs
mlegner: If you mean non-certified, I'd say we don't have to filter them out.

crates/walrus-service/src/client.rs
```diff
@@ -358,6 +361,40 @@ impl<T: ContractClient> Client<T> {
...
+            .sui_client
+            .owned_blobs(false)
+            .await?
+            .into_iter()
+            // TODO(giac): do we have to filter non-registered blobs as well? Leaning towards not.
+            .filter(|blob| blob.blob_id == *blob_id && blob.deletable))
+    }
+
+    #[tracing::instrument(skip_all, fields(blob_id))]
+    /// Deletes all owned blobs that match the provided Blob ID.
```
mlegner: Nit: Can you document the return value?
```suggestion
    /// Deletes all owned blobs that match the provided Blob ID and returns the number of deleted objects.
```

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -363,3 +390,11 @@ fn default_table_format() -> format::TableFormat {
         .padding(1, 1)
         .build()
 }
+
+fn blob_and_file_str(blob_id: &BlobId, file: Option<PathBuf>) -> String {
```
mlegner: Nit: this could take a reference, then you don't have to clone the file when calling this function.
```suggestion
fn blob_and_file_str(blob_id: &BlobId, file: &Option<PathBuf>) -> String {
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -403,6 +419,77 @@ impl FileOrBlobId {
...
+    ///
+    /// This command deletes _all_ owned blob objects matching the provided blob ID.
+    #[clap(short, long, allow_hyphen_values = true, value_parser = parse_blob_id)]
+    #[serde_as(as = "Option<DisplayFromStr>")]
+    #[serde(default)]
+    pub(crate) blob_id: Option<BlobId>,
+    /// The object ID of the blob object to be deleted.
+    ///
+    /// This command deletes only the blob object with the given object ID.
+    #[clap(short, long, allow_hyphen_values = true, value_parser = parse_blob_id)]
```
mlegner: Important+nit: We shouldn't have the `value_parser` and don't need the `allow_hyphen_values`.
```suggestion
    #[clap(short, long)]
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -403,6 +419,77 @@ impl FileOrBlobId {
...
+                );
+                Ok(Some(
+                    *encoding_config
+                        .get_blob_encoder(&read_blob_from_file(file)?)?
+                        .compute_metadata()
+                        .blob_id(),
+                ))
+            }
+            // This case is required for JSON mode where we don't have the clap checking.
+            _ => Ok(None),
```
mlegner: Remark: This also covers the case where an object ID is defined.

crates/walrus-sui/src/client.rs
giac-mysten: lol, I just hadn't rebased yet

crates/walrus-service/src/client.rs
giac-mysten: indeed.

crates/walrus-service/src/client/responses.rs
giac-mysten: Yes, because this way we ensure that when the `Option` is `Some`, the `BlobId` is serialized as with the `Display` implementation. Otherwise, it would be serialized as an optional byte string.

crates/walrus-service/src/client/cli/args.rs
giac-mysten: added to the comment!

crates/walrus-service/src/client/cli/args.rs
giac-mysten: whops, you are right, this is a copy mistake. Thanks.

crates/walrus-service/src/client/responses.rs
giac-mysten: I guess I could just use `Option<DisplayFromStr>`...

mlegner: Ah, makes sense. No strong opinion between the options as long as both work.

crates/walrus-service/src/client.rs
mlegner: Can you then remove the comment?

crates/walrus-e2e-tests/tests/test_client.rs
giac-mysten: This is a nice addition. I had started implementing this, but then the changes started going deeper than expected, and maybe it is better to split them in a separate PR.

One question about this: should we propagate these custom types all the way to the structs? (e.g., the `Blob` now has a `deletable: bool`. Should it become `permanence: Permanence` too?) @mlegner @jpcsmith

mlegner: Hmm... I'm leaning towards not propagating this to structs for two reasons:
1. In contrast to function parameters, we usually have the field name to provide context for structs.
2. It makes (de-)serializing more complicated. 

Specifically, for the `Blob` struct, this should have a 1:1 correspondence with the Move type.

giac-mysten: >Specifically, for the Blob struct, this should have a 1:1 correspondence with the Move type.

yep, we _could_ implement the serialization and deserialization to boolean, but I agree that that would be more confusing than anything. 

Will create a followup for this (#821)

PR #807, opened 2024-09-14
test: make testbed work with new contract

### LINKED ISSUE #794: Finish contract integration in the testbed setup
In local testbed setup, after contract is published, and system&staking objects are created, make each storage node to register itself and finish epoch zero.

See https://github.com/MystenLabs/walrus/pull/783#discussion_r1752217722 for context.

halfprice:
Primarily focus on making `local-testbed.sh` work.

`cargo run --bin walrus-deploy -- deploy-system-contract` deploys the contract and create node config
`cargo run --bin walrus-deploy -- generate-dry-run-configs` register nodes in the contract and finish epoch zero.

close #794

### DIFF
--- a/crates/walrus-service/src/testbed.rs
+++ b/crates/walrus-service/src/testbed.rs
@@ -295,7 +295,10 @@ pub async fn deploy_walrus_contract(
 
     // TODO(#814): make epoch duration in test configurable. Currently hardcoded to 1 hour.
     let system_ctx = create_and_init_system(&mut admin_wallet, n_shards, 0, 3600000).await?;
-    println!("Contract context: {:?}", system_ctx);
+    println!(
+        "Walrus contract created:\npackage id: {:?}\nsystem object: {:?}\nstaking object: {:?}",
+        system_ctx.package_id, system_ctx.system_obj_id, system_ctx.staking_obj_id
+    );
 
     Ok(TestbedConfig {
         sui_network,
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/testbed.rs
```diff
@@ -268,25 +295,22 @@ pub async fn deploy_walrus_contract(
 
     // TODO(#814): make epoch duration in test configurable. Currently hardcoded to 1 hour.
     let system_ctx = create_and_init_system(&mut admin_wallet, n_shards, 0, 3600000).await?;
-
-    // TODO(#794): Contract is published, and system&staking objects are created. However, the
-    // system currently is not moving forward due to that storage nodes are not registered.
+    println!("Contract context: {:?}", system_ctx);
```
giac-mysten: Question: is this information essential when running the testbed?

Suggestion: if it is important for the testbed user, format it in a more readable manner and print it. Otherwise log it with tracing.

crates/walrus-service/src/testbed.rs
```diff
@@ -423,6 +468,36 @@ pub async fn create_storage_node_configs(
...
+    register_committee_and_stake(
+        admin_wallet,
+        &testbed_config.system_ctx,
+        &node_params,
+        &contract_clients.iter().collect::<Vec<_>>(),
+        &amounts_to_stake,
+    )
+    .await?;
+
+    end_epoch_zero(contract_clients.first().unwrap()).await?;
```
mlegner: Nit: Can we replace the `unwrap` by an `expect`?

crates/walrus-service/src/testbed.rs
halfprice: I think these are useful information so that we can check activity on chain. I formatted as follows:

```
Walrus contract created:
package id: 0xaaa
system object: 0xbbb
staking object: 0xccc
```

crates/walrus-service/src/testbed.rs
halfprice: Done

PR #808, opened 2024-09-15
feat: Update smart contract to certify event blobs

sadhansood:
This PR updates walrus smart contracts to also start certifying event blobs. Each node will invoke the contract endpoint and upon reaching majority, the blob will be certified and a certified event emitted.

### DIFF
--- a/contracts/walrus/sources/staking/storage_node.move
+++ b/contracts/walrus/sources/staking/storage_node.move
@@ -5,7 +5,9 @@
 module walrus::storage_node;
 
 use std::string::String;
-use sui::{bls12381::{G1, g1_from_bytes}, group_ops::Element};
+use sui::bls12381::{G1, g1_from_bytes};
+use sui::group_ops::Element;
+use walrus::event_blob::EventBlobAttestation;
 
 // Error codes
 const EInvalidNetworkPublicKey: u64 = 1;
@@ -25,6 +27,7 @@ public struct StorageNodeCap has key, store {
     id: UID,
     node_id: ID,
     last_epoch_sync_done: u32,
+    last_event_blob_attestation: Option<EventBlobAttestation>,
 }
 
 /// A public constructor for the StorageNodeInfo.
@@ -51,11 +54,14 @@ public(package) fun new_cap(node_id: ID, ctx: &mut TxContext): StorageNodeCap {
         id: object::new(ctx),
         node_id,
         last_epoch_sync_done: 0,
+        last_event_blob_attestation: option::none(),
     }
 }
 
 /// Return the public key of the storage node.
-public(package) fun public_key(self: &StorageNodeInfo): &Element<G1> { &self.public_key }
+public(package) fun public_key(self: &StorageNodeInfo): &Element<G1> {
+    &self.public_key
+}
 
 /// Return the node ID of the storage node.
 public fun id(cap: &StorageNodeInfo): ID { cap.node_id }
@@ -63,12 +69,30 @@ public fun id(cap: &StorageNodeInfo): ID { cap.node_id }
 /// Return the pool ID of the storage node.
 public fun node_id(cap: &StorageNodeCap): ID { cap.node_id }
 
-/// Return the last epoch in which the storage node attested that it has finished syncing.
-public fun last_epoch_sync_done(cap: &StorageNodeCap): u32 { cap.last_epoch_sync_done }
+/// Return the last epoch in which the storage node attested that it has
+/// finished syncing.
+public fun last_epoch_sync_done(cap: &StorageNodeCap): u32 {
+    cap.last_epoch_sync_done
+}
 
-/// Set the last epoch in which the storage node attested that it has finished syncing.
+/// Set the last epoch in which the storage node attested that it has finished
+/// syncing.
 public(package) fun set_last_epoch_sync_done(self: &mut StorageNodeCap, epoch: u32) {
     self.last_epoch_sync_done = epoch;
+}
+
+/// Return the latest event blob attestion.
+public fun last_event_blob_attestation(cap: &mut StorageNodeCap): Option<EventBlobAttestation> {
+    cap.last_event_blob_attestation
+}
+
+/// Set the last epoch in which the storage node attested that it has finished
+/// syncing.
+public(package) fun set_last_event_blob_attestation(
+    self: &mut StorageNodeCap,
+    attestation: EventBlobAttestation,
+) {
+    self.last_event_blob_attestation = option::some(attestation);
 }
 
 #[test_only]
--- a/contracts/walrus/sources/system/bls_aggregate.move
+++ b/contracts/walrus/sources/system/bls_aggregate.move
@@ -3,11 +3,9 @@
 
 module walrus::bls_aggregate;
 
-use sui::{
-    bls12381::{Self, bls12381_min_pk_verify, G1},
-    group_ops::{Self, Element},
-    vec_map::{Self, VecMap}
-};
+use sui::bls12381::{Self, bls12381_min_pk_verify, G1};
+use sui::group_ops::{Self, Element};
+use sui::vec_map::{Self, VecMap};
 use walrus::messages::{Self, CertifiedMessage};
 
 // Error codes
@@ -66,6 +64,13 @@ public(package) fun new_bls_committee_member(
     }
 }
 
+// === Accessors for BlsCommitteeMember ===
+
+/// Get the node id of the committee member.
+public(package) fun node_id(self: &BlsCommitteeMember): sui::object::ID {
+    self.node_id
+}
+
 // === Accessors for BlsCommittee ===
 
 /// Get the epoch of the committee.
@@ -78,23 +83,46 @@ public(package) fun n_shards(self: &BlsCommittee): u16 {
     self.n_shards
 }
 
+/// Returns the member at given index
+public(package) fun get_idx(self: &BlsCommittee, idx: u64): &BlsCommitteeMember {
+    self.members.borrow(idx)
+}
+
 /// Checks if the committee contains a given node.
 public(package) fun contains(self: &BlsCommittee, node_id: &ID): bool {
-    self.members.find_index!(|member| member.node_id == *node_id).is_some()
+    self.find_index(node_id).is_some()
+}
+
+/// Returns the member weight if it is part of the committee or 0 otherwise
+public(package) fun get_member_weight(self: &BlsCommittee, node_id: &ID): u16 {
+    self.find_index(node_id).and!(|idx| {
+        let member = &self.members[idx];
+        option::some(member.weight)
+    }).get_with_default(0)
+}
+
+/// Finds the index of the member by node_id
+public(package) fun find_index(self: &BlsCommittee, node_id: &ID): std::option::Option<u64> {
+    self.members.find_index!(|member| &member.node_id == node_id)
 }
 
 /// Returns the members of the committee with their weights.
 public(package) fun to_vec_map(self: &BlsCommittee): VecMap<ID, u16> {
     let mut result = vec_map::empty();
-    self.members.do_ref!(|member| result.insert(member.node_id, member.weight));
+    self.members.do_ref!(|member| {
+        result.insert(member.node_id, member.weight)
+    });
     result
 }
 
 /// Verifies that a message is signed by a quorum of the members of a committee.
 ///
-/// The signers are listed as indices into the `members` vector of the committee in increasing
-/// order and with no repetitions. The total weight of the signers (i.e. total number of shards)
-/// is returned, but if a quorum is not reached the function aborts with an error.
+/// The signers are listed as indices into the `members` vector of the committee
+/// in increasing
+/// order and with no repetitions. The total weight of the signers (i.e. total
+/// number of shards)
+/// is returned, but if a quorum is not reached the function aborts with an
+/// error.
 public(package) fun verify_quorum_in_epoch(
     self: &BlsCommittee,
     signature: vector<u8>,
@@ -110,9 +138,17 @@ public(package) fun verify_quorum_in_epoch(
     messages::new_certified_message(message, self.epoch, stake_support)
 }
 
-/// Verify an aggregate BLS signature is a certificate in the epoch, and return the type of
-/// certificate and the bytes certified. The `signers` vector is an increasing list of indexes
-/// into the `members` vector of the committee. If there is a certificate, the function
+/// Returns true if the weight is more than the aggregate weight of quorum members of a committee.
+public(package) fun verify_quorum(self: &BlsCommittee, weight: u16): bool {
+    3 * (weight as u64) >= 2 * (self.n_shards as u64) + 1
+}
+
+/// Verify an aggregate BLS signature is a certificate in the epoch, and return
+/// the type of
+/// certificate and the bytes certified. The `signers` vector is an increasing
+/// list of indexes
+/// into the `members` vector of the committee. If there is a certificate, the
+/// function
 /// returns the total stake. Otherwise, it aborts.
 public(package) fun verify_certificate(
     self: &BlsCommittee,
@@ -122,7 +158,8 @@ public(package) fun verify_certificate(
 ): u16 {
     // Use the signers flags to construct the key and the weights.
 
-    // Lower bound for the next `member_index` to ensure they are monotonically increasing
+    // Lower bound for the next `member_index` to ensure they are monotonically
+    // increasing
     let mut min_next_member_index = 0;
     let mut aggregate_key = bls12381::g1_identity();
     let mut aggregate_weight = 0;
@@ -144,7 +181,7 @@ public(package) fun verify_certificate(
     // The expression below is the solution to the inequality:
     // n_shards = 3 f + 1
     // stake >= 2f + 1
-    assert!(3 * (aggregate_weight as u64) >= 2 * (self.n_shards as u64) + 1, ENotEnoughStake);
+    assert!(verify_quorum(self, aggregate_weight), ENotEnoughStake);
 
     // Verify the signature
     let pub_key_bytes = group_ops::bytes(&aggregate_key);
--- a/contracts/walrus/sources/system/event_blob.move
+++ b/contracts/walrus/sources/system/event_blob.move
@@ -1,0 +1,159 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// Module to certify event blobs.
+module walrus::event_blob;
+
+use sui::vec_map::VecMap;
+
+// === Definitions related to event blob certification ===
+
+/// Event blob index which was attested by a storage node.
+public struct EventBlobAttestation has store, copy, drop {
+    checkpoint_sequence_num: u64,
+    epoch: u32,
+}
+
+/// State of a certified event blob.
+public struct EventBlob has copy, store, drop {
+    /// Blob id of the certified event blob.
+    blob_id: u256,
+    /// Ending sui checkpoint of the certified event blob.
+    ending_checkpoint_sequence_number: u64,
+}
+
+/// State of event blob stream.
+#[allow(unused_field)]
+public struct EventBlobCertificationState has key, store {
+    id: UID,
+    /// Latest certified event blob.
+    latest_certified_blob: Option<EventBlob>,
+    /// Current event blob being attested.
+    aggregate_weight_per_blob: VecMap<u256, u16>,
+}
+
+// === Accessors related to event blob attestation ===
+
+public(package) fun new_attestation(
+    checkpoint_sequence_num: u64,
+    epoch: u32,
+): EventBlobAttestation {
+    EventBlobAttestation {
+        checkpoint_sequence_num,
+        epoch,
+    }
+}
+
+public(package) fun last_attested_event_blob_checkpoint_seq_num(self: &EventBlobAttestation): u64 {
+    self.checkpoint_sequence_num
+}
+
+public(package) fun last_attested_event_blob_epoch(self: &EventBlobAttestation): u32 { self.epoch }
+
+// === Accessors for EventBlob ===
+
+public(package) fun new_event_blob(
+    ending_checkpoint_sequence_number: u64,
+    blob_id: u256,
+): EventBlob {
+    EventBlob {
+        blob_id,
+        ending_checkpoint_sequence_number,
+    }
+}
+
+/// Returns the blob id of the event blob
+public(package) fun blob_id(self: &EventBlob): u256 {
+    self.blob_id
+}
+
+/// Returns the ending checkpoint sequence number of the event blob
+public(package) fun ending_checkpoint_sequence_number(self: &EventBlob): u64 {
+    self.ending_checkpoint_sequence_number
+}
+
+// === Accessors for EventBlobCertificationState ===
+
+/// Creates a blob state with no signers and no last checkpoint sequence number
+public(package) fun create_with_empty_state(ctx: &mut TxContext): EventBlobCertificationState {
+    let id = object::new(ctx);
+    EventBlobCertificationState {
+        id,
+        latest_certified_blob: option::none(),
+        aggregate_weight_per_blob: sui::vec_map::empty(),
+    }
+}
+
+/// Returns the blob id of the latest certified event blob
+public(package) fun get_latest_certified_blob_id(self: &EventBlobCertificationState): Option<u256> {
+    self.latest_certified_blob.map!(|state| state.blob_id())
+}
+
+/// Returns the checkpoint sequence number of the latest certified event
+/// blob
+public(package) fun get_latest_certified_checkpoint_sequence_number(
+    self: &EventBlobCertificationState,
+): Option<u64> {
+    self.latest_certified_blob.map!(|state| state.ending_checkpoint_sequence_number())
+}
+
+/// Returns true if a blob is already certified or false otherwise
+public(package) fun is_blob_already_certified(
+    self: &EventBlobCertificationState,
+    ending_checkpoint_sequence_num: u64,
+): bool {
+    self
+        .get_latest_certified_checkpoint_sequence_number()
+        .map!(
+            |
+                latest_certified_sequence_num,
+            | latest_certified_sequence_num >= ending_checkpoint_sequence_num,
+        )
+        .get_with_default(false)
+}
+
+/// Updates the latest certified event blob
+public(package) fun update_latest_certified_event_blob(
+    self: &mut EventBlobCertificationState,
+    checkpoint_sequence_number: u64,
+    blob_id: u256,
+) {
+    self.get_latest_certified_checkpoint_sequence_number().do!(|latest_certified_sequence_num| {
+        assert!(checkpoint_sequence_number > latest_certified_sequence_num);
+    });
+    self.latest_certified_blob =
+        option::some(
+            new_event_blob(checkpoint_sequence_number, blob_id),
+        );
+}
+
+/// Update the aggregate weight of an event blob
+public(package) fun update_aggregate_weight(
+    self: &mut EventBlobCertificationState,
+    blob_id: u256,
+    weight: u16,
+): u16 {
+    let agg_weight = self.aggregate_weight_per_blob.get_mut(&blob_id);
+    *agg_weight = *agg_weight + weight;
+    *agg_weight
+}
+
+/// Start tracking which nodes are signing the event blob with given id for
+/// event blob certification
+public(package) fun start_tracking_blob(self: &mut EventBlobCertificationState, blob_id: u256) {
+    if (!self.aggregate_weight_per_blob.contains(&blob_id)) {
+        self.aggregate_weight_per_blob.insert(blob_id, 0);
+    };
+}
+
+/// Stop tracking nodes for the given blob id
+public(package) fun stop_tracking_blob(self: &mut EventBlobCertificationState, blob_id: u256) {
+    if (self.aggregate_weight_per_blob.contains(&blob_id)) {
+        self.aggregate_weight_per_blob.remove(&blob_id);
+    };
+}
+
+/// Reset blob certification state upon epoch change
+public(package) fun reset(self: &mut EventBlobCertificationState) {
+    self.aggregate_weight_per_blob = sui::vec_map::empty();
+}
--- a/contracts/walrus/sources/system/system_state_inner.move
+++ b/contracts/walrus/sources/system/system_state_inner.move
@@ -4,17 +4,19 @@
 #[allow(unused_variable, unused_mut_parameter, unused_field)]
 module walrus::system_state_inner;
 
-use sui::{balance::Balance, coin::Coin};
+use sui::balance::Balance;
+use sui::coin::Coin;
 use wal::wal::WAL;
-use walrus::{
-    blob::{Self, Blob},
-    bls_aggregate::{Self, BlsCommittee},
-    epoch_parameters::EpochParams,
-    events::emit_invalid_blob_id,
-    storage_accounting::{Self, FutureAccountingRingBuffer},
-    storage_node::StorageNodeCap,
-    storage_resource::{Self, Storage}
-};
+use walrus::blob::{Self, Blob};
+use walrus::bls_aggregate::{Self, BlsCommittee};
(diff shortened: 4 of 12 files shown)


contracts/walrus/tests/system/blob_tests.move
```diff
@@ -13,6 +13,7 @@ use walrus::{
     storage_resource::{Self, split_by_epoch, destroy, Storage},
     system::{Self, System},
     system_state_inner,
+
```
mlegner: Nit: This change could be reverted.

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,98 @@
...
+
+/// Module to certify event blobs.
+module walrus::event_blob;
+
+use sui::vec_map::VecMap;
+
+// === Definitions related to event blob certification ===
+
+/// State of a certified event blob.
+public struct CertifiedEventBlobState has copy, store, drop {
```
mlegner: Nit: IMO "state" isn't really a fitting term for this struct. Maybe something like `EventBlobMetadata` or simply `EventBlob` would be better.

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,98 @@
...
+    /// Blob id of the certified event blob.
+    blob_id: u256,
+    /// Ending sui checkpoint of the certified event blob.
+    checkpoint_sequence_number: u64,
```
mlegner: Nit: Can we call this `ending_checkpoint_sequence_number` or `last_checkpoint_sequence_number`?

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,98 @@
...
+
+/// State of event blob certification.
+#[allow(unused_field)]
+public struct EventBlobCertificationState has key, store {
+    id: UID,
+    /// Latest certified event blob.
+    latest_certified_blob_state: Option<CertifiedEventBlobState>,
+    /// Nodes which have signed an event blob undergoing certification
+    /// at a particular epoch.
+    blob_signers: VecMap<u256, VecMap<u32, vector<u16>>>,
```
mlegner: Question: Is this a mapping `blob_id -> (epoch -> [node_id])`? 

Suggestion: Only track this for the current epoch and clean it up on epoch change.

Also, suggestion: In addition to the vector of signers, can we also store their current total weight to simplify the quorum check?

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,98 @@
...
+
+/// Returns the ending checkpoint sequence number of the event blob
+public(package) fun checkpoint_sequence_number(self: &CertifiedEventBlobState): u64 {
+        self.checkpoint_sequence_number
+}
+
+// === Accessors for EventBlobCertificationState ===
+
+/// Creates a blob state with no signers and no last checkpoint sequence number
+public(package) fun create_empty(ctx: &mut TxContext): EventBlobCertificationState {
```
mlegner: Nit: Can we make the function name a bit more descriptive? It is not clear what type of "empty" object is created from the name.

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -78,15 +86,27 @@ public(package) fun n_shards(self: &BlsCommittee): u16 {
...
 }
 
+/// Returns the member at given index
+public(package) fun get_idx(self: &BlsCommittee, idx: u64): &BlsCommitteeMember {
+    self.members.borrow(idx)
+}
+
 /// Checks if the committee contains a given node.
 public(package) fun contains(self: &BlsCommittee, node_id: &ID): bool {
     self.members.find_index!(|member| member.node_id == *node_id).is_some()
```
mlegner: Nit: Given the new function `find_index` this could be simplified:
```suggestion
    self.find_index(node_id).is_some()
```

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -296,9 +301,74 @@ public(package) fun certify_event_blob(
...
+        let last_certified_sequence_number
+            = self.event_blob_certification_state
+                .get_latest_certified_checkpoint_sequence_number().destroy_some();
+        ending_checkpoint_sequence_num > last_certified_sequence_number
+    } else {
+        true
+    };
+    if (certify_blob) {
+        let epoch = self.committee().epoch();
+        let node_index = self.committee().find_index(&cap.node_id()).destroy_some() as u16;
```
mlegner: Suggestion: Can we have a specific error code for the case where the node is not part of the current committee?

contracts/walrus/sources/system/event_blob.move
mlegner: Another question: Is there a danger this object might get too large if a malicious storage node "certifies" many invalid blobs? How can we avoid this or clean it up?

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -110,6 +130,33 @@ public(package) fun verify_quorum_in_epoch(
...
+        // Bounds check happens here
+        let member = &self.members[member_index];
+        let key = &member.public_key;
+        let weight = member.weight;
+
+        aggregate_key = bls12381::g1_add(&aggregate_key, key);
+        aggregate_weight = aggregate_weight + weight;
+    });
+    3 * (aggregate_weight as u64) >= 2 * (self.n_shards as u64) + 1
+}
```
mlegner: Important: I don't think this is the best way of checking that the event blob was certified by a quorum of nodes. Specifically, we don't need any keys, we just have to sum up the weight of the nodes. There is also no requirement of ordering.

Suggestion: Define methods `min_weight_for_quorum(self: &BlsCommittee) -> u16` and/or `is_quorum(self: &BlsCommittee, weight: u16) -> bool` and use this in the verification.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -296,9 +301,74 @@ public(package) fun certify_event_blob(
...
+        if (blob_certified) {
+            let storage = storage_resource::create_storage(
+                self.committee.epoch(),
+                self.committee.epoch() + MAX_EPOCHS_AHEAD, encoded_blob_length(
+                    size,
+                    encoding_type,
+                    self.n_shards()
+                ),
+                ctx
+            );
```
mlegner: Question: Is my understanding correct that this is then outside of the storage accounting? 

Suggestion: Keep this within the storage-accounting framework but allow it to exceed the `total_capacity_size`. This could be a function similar to `system_state_inner::reserve_space` but without payment and without the capacity check. Then we at least keep track of the space of the storage capacity used by the event blobs even though we cannot guarantee that we stay below the max capacity.

Maybe others have different opinions on this though.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -296,9 +301,74 @@ public(package) fun certify_event_blob(
...
+                ctx,
+            );
+            let certified_blob_msg = messages::certified_event_blob_message(self.epoch(),blob_id);
+            blob.certify_with_certified_msg(self.epoch(), certified_blob_msg);
+            self.event_blob_certification_state
+                .update_latest_certified_checkpoint_sequence_number(
+                    ending_checkpoint_sequence_num,
+                    blob_id
+                );
+            self.event_blob_certification_state.stop_tracking_blob(blob_id);
```
mlegner: Important: We also have to do some garbage collection there in other cases to defend against malicious nodes adding data to the state.

contracts/walrus/sources/system/event_blob.move
karlwuest: >In addition to the vector of signers, can we also store their current total weight to simplify the quorum check?

My suggestion would be to handle this similarly to the "epoch change done", i.e. we store the highest voted for sequence number in the StorageNodeCap and only store the aggregated weight in the shared object. Then we only need to keep track of the BlobId "candidates" per sequence number that has not received a quorum yet, but we don't need to keep a list of signers around. In principle, we could even say that the event blob at sequence number i+1 can only be voted for once the event blob at sequence number i has been certified, i.e. keeping track of just a map of candidate blob id to weight (cc @sadhansood). 

(Disclaimer: I'm only skimming the PR/comments, so I didn't look at any of the code in detail yet. I can do a detailed review once I'm back.)

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -78,15 +86,27 @@ public(package) fun n_shards(self: &BlsCommittee): u16 {
...
+
 /// Checks if the committee contains a given node.
 public(package) fun contains(self: &BlsCommittee, node_id: &ID): bool {
     self.members.find_index!(|member| member.node_id == *node_id).is_some()
 }
 
+/// Finds the index of the member by node_id
+public(package) fun find_index(self: &BlsCommittee, node_id: &ID): std::option::Option<u64> {
+    self.members.find_index!(|member| member.node_id == *node_id)
```
damirka: nit, I believe here we can do:
```suggestion
    self.members.find_index!(|member| &member.node_id == node_id)
```

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,98 @@
...
+    blob_id: u256): CertifiedEventBlobState {
+    CertifiedEventBlobState {
+        blob_id,
+        checkpoint_sequence_number,
+    }
+}
+
+/// Returns the blob id of the event blob
+public(package) fun blob_id(self: &CertifiedEventBlobState): u256 {
+        self.blob_id
```
damirka: I would suggest to run a formatter on the codebase.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -296,9 +301,74 @@ public(package) fun certify_event_blob(
...
+            std::option::some(blob)
+        } else {
+            std::option::none()
```
damirka: Option is always imported, so you can do (everywhere):
```suggestion
            option::none()
```

contracts/walrus/tests/system/event_blob_tests.move
```diff
@@ -0,0 +1,186 @@
...
+            assert!(state.get_latest_certified_checkpoint_sequence_number().is_none());
+        } else {
+            // 7th node signing the blob triggers blob certification
+            assert!(state.get_latest_certified_checkpoint_sequence_number().is_some());
+        };
+        index = index + 1
+    };
+    blob.destroy!(|b| b.burn());
+    nodes.destroy!(|node| node.destroy());
+    system
```
damirka: You don't want to return anything from test functions. Better use 
```move
use sui::test_utils::destroy;
```

```suggestion
    destroy(system);
```

contracts/walrus/tests/system/event_blob_tests.move
```diff
@@ -0,0 +1,186 @@
...
+public fun test_event_blob_certify_change_epoch(): system::System {
+    let ctx = &mut tx_context::dummy();
+    let mut system: system::System = system::new_for_testing_with_multiple_members(ctx);
+    // Total of 10 nodes
+    assert!(system.committee().to_vec_map().size() == 10);
+    let mut nodes = test_nodes();
+    set_storage_node_caps(&system, ctx, &mut nodes);
+    let blob_id = blob::derive_blob_id(ROOT_HASH, RED_STUFF, SIZE);
+    let mut index = 0;
+    let mut blob: vector<Blob> = vector::empty();
```
damirka:
```suggestion
    let mut blob: vector<Blob> = vector[];
```

contracts/walrus/tests/system/event_blob_tests.move
```diff
@@ -0,0 +1,186 @@
...
+    // the blob in previous epoch
+    system.committee_mut().increment_epoch_for_testing();
+    system.certify_event_blob(
+        nodes.borrow(index).cap(), blob_id, ROOT_HASH, SIZE, RED_STUFF, 100, ctx)
+            .destroy!(|b| blob.push_back(b));
+    let state = system.inner().get_event_blob_certification_state();
+    assert!(state.get_latest_certified_checkpoint_sequence_number().is_none());
+    index = 0;
+    // All nodes sign the blob in current epoch
+    while (index < 10) {
```
damirka, line 197: Not necessary, but slightly increases readability:
```suggestion
10.do!(|index| {
```

contracts/walrus/tests/test_utils.move
```diff
@@ -256,6 +256,19 @@ public fun new_bls_committee_for_testing(epoch: u32): bls_aggregate::BlsCommitte
...
 }
 
+/// Test committee with 10 committee member and 100 shards, using
+/// `test_utils::bls_sk_for_testing()` as secret key.
+public fun new_bls_committee_with_multiple_members_for_testing(epoch: u32, tx_context: &mut TxContext): bls_aggregate::BlsCommittee {
+    let keys = bls_secret_keys_for_testing();
+    let members = keys.map!(|sk| {
+        let pub_key = bls12381::g1_from_bytes(&bls_min_pk_from_sk(&sk));
+        let node_id = tx_context.fresh_object_address().to_id();
+        let member = bls_aggregate::new_bls_committee_member(pub_key, 100, node_id);
```
damirka: nit: can return without assignment
```suggestion
        bls_aggregate::new_bls_committee_member(pub_key, 100, node_id)
```

contracts/walrus/sources/system/system_state_inner.move
sadhansood: @mlegner can you elaborate how a malicious node can add data to the state. Or maybe this comment is not relevant anymore.

contracts/walrus/tests/system/event_blob_tests.move
sadhansood: This did not work, or threw a compiler error for some reason

contracts/walrus/sources/system/event_blob.move
mlegner: I believe the formatter reformatted also some existing code. @damirka What formatter (and version) are we using for the code?

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -47,12 +48,16 @@ public struct SystemStateInnerV1 has key, store {
     write_price_per_unit_size: u64,
     /// Accounting ring buffer for future epochs.
     future_accounting: FutureAccountingRingBuffer,
+    /// event blob certification state
+    event_blob_certification_state: EventBlobCertificationState,
```
mlegner, line 55: Important: This addition requires a corresponding change to the Rust code in `crates/walrus-sui/src/types/move_structs.rs`.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -133,6 +141,35 @@ public(package) fun reserve_space(
     )
 }
 
+/// Allow buying a storage reservation for a given period of epochs without payment.
+/// Only to be used for event blobs.
+fun reserve_space_without_payment(
```
mlegner, line 140: Suggestion: With this new function, we can simplify the `reserve_space` function above.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -296,9 +339,93 @@ public(package) fun certify_event_blob(
...
     blob_id: u256,
+    root_hash: u256,
     size: u64,
-) {
-    abort ENotImplemented
+    encoding_type: u8,
+    ending_checkpoint_sequence_num: u64,
+    ctx: &mut TxContext,
+): Option<Blob> {
+    assert!(self.committee().contains(&cap.node_id()), EIncorrectCommittee);
```
mlegner: Nit: Not sure if this is the correct error type. Maybe a new one would be more appropriate?

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -296,9 +339,93 @@ public(package) fun certify_event_blob(
...
+        .is_blob_already_certified(
+            ending_checkpoint_sequence_num,
+        );
+
+    let blob_signed = self
+        .event_blob_certification_state
+        .is_blob_already_signed(
+            &cap.node_id(),
+            ending_checkpoint_sequence_num,
+        );
```
mlegner: Suggestion: Instead of this check and the various updates below, how about a function `event_blob::maybe_update_latest_signed_sequence_number` (replacing `update_latest_signed_sequence_number`) that increases the node's latest signed sequence number to the input (if it is larger)? It can also return whether or not the number was increased.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -296,9 +339,93 @@ public(package) fun certify_event_blob(
...
+                MAX_EPOCHS_AHEAD,
+                ctx,
+            );
+            let mut blob = blob::new(
+                storage,
+                blob_id,
+                root_hash,
+                size,
+                encoding_type,
+                true,
```
mlegner: Question: Why is this marked as deletable?

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,251 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// Module to certify event blobs.
+module walrus::event_blob;
+
+use sui::vec_map::VecMap;
+use walrus::bls_aggregate::BlsCommittee;
+
+const MAX_SIGNED_EVENT_BLOBS: u64 = 104;
```
mlegner: Question: How is this value chosen?

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,251 @@
...
+    /// There is an upper limit on the number of signed but
+    /// uncertified blobs per node
+    signed_blob_ids_per_node: VecMap<ID, sui::vec_set::VecSet<u256>>,
+    /// Aggregate weight of signed but uncertified blob ids.
+    aggregate_weight_per_blob: VecMap<u256, u16>,
+}
+
+// === Accessors for EventBlob ===
+
+public(package) fun new_certified_blob_state(
```
mlegner: Nit: Now that the struct has been renamed, this function should probably be aligned.

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,251 @@
...
+
+/// Returns the ending checkpoint sequence number of the event blob
+public(package) fun ending_checkpoint_sequence_number(self: &EventBlob): u64 {
+    self.ending_checkpoint_sequence_number
+}
+
+// === Accessors for EventBlobCertificationState ===
+
+/// Creates a blob state with no signers and no last checkpoint sequence number
+public(package) fun create_with_no_blobs(ctx: &mut TxContext): EventBlobCertificationState {
```
mlegner: Nit: The function name doesn't tell you *what* is being created. Not an event blob but the state.

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,251 @@
...
+public struct EventBlobCertificationState has key, store {
+    id: UID,
+    /// Latest certified event blob.
+    latest_certified_blob: Option<EventBlob>,
+    /// Latest checkpoint sequence number signed by a node.
+    latest_signed_sequence_num_per_node: VecMap<ID, u64>,
+    /// All signed but uncertified blob ids per node.
+    /// There is an upper limit on the number of signed but
+    /// uncertified blobs per node
+    signed_blob_ids_per_node: VecMap<ID, sui::vec_set::VecSet<u256>>,
```
mlegner: Question: Is there a limit to the size of this map?

Also, AFAIU @karlwuest's comment, he suggested to put this and the `latest_signed_sequence_num_per_node` into the `StorageNodeCap` instead of having a map in the global state. That would also simplify the cleanup on epoch change.

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,251 @@
...
+    /// Aggregate weight of signed but uncertified blob ids.
+    aggregate_weight_per_blob: VecMap<u256, u16>,
```
mlegner, line 32: Assuming 1000 shards and thus at most 333 malicious nodes, this map could potentially contain something like `333*104 ~ 34k` entries. Given that operations on a `VecMap` have O(n) complexity, does the code still work in this worst case?

contracts/walrus/sources/system/system_state_inner.move
mlegner: This is no longer relevant with the changes.

contracts/walrus/sources/system/event_blob.move
mlegner: OK, it turns out that we use some non-default settings:
- `printWidth = 100` 
- `autoGroupImports = 'package'`
- `wrapComments = false`

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,251 @@
...
+/// Module to certify event blobs.
+module walrus::event_blob;
+
+use sui::vec_map::VecMap;
+use walrus::bls_aggregate::BlsCommittee;
+
+const MAX_SIGNED_EVENT_BLOBS: u64 = 104;
+
+// error codes
+const ESignedEventBlobLimitReached: u64 = 100;
```
karlwuest: Nit: Replace this with a useful error message, e.g.
```
#[error]
const ESignedEventBlobLimitReached: vector<u8> = b"Limit for signed event blobs reached.";
```

contracts/walrus/sources/system/event_blob.move
karlwuest: With 34k entries, the object holding this will exceed the maximum object size.

contracts/walrus/sources/system/system_state_inner.move
karlwuest: Yes, this should be a separate error type, something like `ENotCommitteeMember`.

contracts/walrus/sources/system/event_blob.move
sadhansood: This comment is not relevant anymore I believe

contracts/walrus/sources/system/system_state_inner.move
sadhansood: Thanks, made the relevenat change.

contracts/walrus/sources/system/system_state_inner.move
sadhansood: done

contracts/walrus/sources/staking/storage_node.move
```diff
@@ -19,12 +20,18 @@ public struct StorageNodeInfo has store, copy, drop {
...
+    epoch: u32,
+}
+
 /// A Capability which represents a storage node and authorizes the holder to
 /// perform operations on the storage node.
 public struct StorageNodeCap has key, store {
     id: UID,
     node_id: ID,
     last_epoch_sync_done: u32,
+    last_event_blob_attestation: Option<EventBlobAttestation>,
```
karlwuest, line 30: Nit: I think we could even get away without the `Option` and using (0, 0) as the initial value, since we know that in epoch 0 no event blobs are created.

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -78,23 +83,52 @@ public(package) fun n_shards(self: &BlsCommittee): u16 {
...
 ///
-/// The signers are listed as indices into the `members` vector of the committee in increasing
-/// order and with no repetitions. The total weight of the signers (i.e. total number of shards)
-/// is returned, but if a quorum is not reached the function aborts with an error.
+/// The signers are listed as indices into the `members` vector of the committee
+/// in increasing
+/// order and with no repetitions. The total weight of the signers (i.e. total
+/// number of shards)
+/// is returned, but if a quorum is not reached the function aborts with an
+/// error.
```
karlwuest, line 125: Nit: This looks like the result of the formatter with a line width of 80, but we're generally using a line width of 100. Can you change this in your formatter settings and fix these comments?

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -144,7 +190,10 @@ public(package) fun verify_certificate(
     // The expression below is the solution to the inequality:
     // n_shards = 3 f + 1
     // stake >= 2f + 1
-    assert!(3 * (aggregate_weight as u64) >= 2 * (self.n_shards as u64) + 1, ENotEnoughStake);
+    assert!(
+        3 * (aggregate_weight as u64) >= 2 * (self.n_shards as u64) + 1,
```
karlwuest: Nit: replace this with a call to the new `verify_quorum` function.

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,149 @@
...
+use sui::vec_map::VecMap;
+
+// === Definitions related to event blob certification ===
+
+/// State of a certified event blob.
+public struct EventBlob has copy, store, drop {
+    /// Blob id of the certified event blob.
+    blob_id: u256,
+    /// Ending sui checkpoint of the certified event blob.
+    ending_checkpoint_sequence_number: u64,
```
karlwuest, line 22: Question: what does this sequence number actually refer to and what assumptions do we make about this?

contracts/walrus/sources/system/event_blob.move
```diff
@@ -0,0 +1,149 @@
...
+        .map!(
+            |
+                latest_certified_sequence_num,
+            | latest_certified_sequence_num >= ending_checkpoint_sequence_num,
+        )
+        .get_with_default(false)
+}
+
+/// Updates the latest certified event blob
+public(package) fun update_latest_certified_checkpoint_sequence_number(
```
karlwuest: Suggestion: update name to more accurately reflect what the function does

```suggestion
public(package) fun update_latest_certified_event_blob(
```

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -294,11 +348,102 @@ fun process_storage_payments(
...
+                false,
+                self.epoch(),
+                self.n_shards(),
+                ctx,
+            );
+            let certified_blob_msg = messages::certified_event_blob_message(
+                self.epoch(),
+                blob_id,
+            );
+            blob.certify_with_certified_msg(self.epoch(), certified_blob_msg);
```
karlwuest: Suggestion: This can be simplified by just directly emitting the necessary event. All that `certify_with_certified_msg` does additionally is to perform some checks that will succeed by construction here. Then we also don't need the `messages::certified_event_blob_message` function anymore.

```suggestion
            blob.emit_certified(false);
```

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -294,11 +348,102 @@ fun process_storage_payments(
...
     blob_id: u256,
+    root_hash: u256,
     size: u64,
-) {
-    abort ENotImplemented
+    encoding_type: u8,
+    ending_checkpoint_sequence_num: u64,
+    epoch: u32,
+    ctx: &mut TxContext,
+): Option<Blob> {
```
karlwuest: Important: We should not return a blob object here in any case, since there is no "owner" for the event blob. I would suggest to just burn the blob after certifying it.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -294,11 +348,102 @@ fun process_storage_payments(
...
+
+    let attestation = new_attestation(ending_checkpoint_sequence_num, epoch);
+    cap.set_last_event_blob_attestation(attestation);
+
+    let blob_certified = self
+        .event_blob_certification_state
+        .is_blob_already_certified(
+            ending_checkpoint_sequence_num,
+        );
+    if (!blob_certified) {
```
karlwuest: Suggestion: For the nested `if` statements here (line 399 and 406), flip the conditions and add early returns to improve readability.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -294,11 +348,102 @@ fun process_storage_payments(
...
+                    attestation.last_attested_event_blob_checkpoint_seq_num()
+                        <= certified_checkpoint_seq_num,
+                EIncorrectAttestation,
+            );
+        } else {
+            assert!(
+                attestation.last_attested_event_blob_epoch() < self.epoch(),
+                EIncorrectAttestation,
+            );
+        }
```
karlwuest, line 374: Important: These checks only prevent a node from voting again for a lower checkpoint sequence number. However, they do not prevent a node from repeatedly voting for new blobs with incrementally higher sequence numbers, e.g., a malicious node can just fill up the `VecMap` in the `EventBlobCertificationState`, which can prevent others from voting since the transactions will (at some point) start aborting. To fix this, we can add an explicit sequence number for the event blobs that is incremented by exactly one for each blob (see comment on `EventBlobAttestation`) and then add a check that the index we are voting for is one higher than the latest certified. This will ensure that each node can only vote for one single blob at a time.

contracts/walrus/sources/staking/storage_node.move
```diff
@@ -19,12 +20,18 @@ public struct StorageNodeInfo has store, copy, drop {
     network_public_key: vector<u8>,
 }
 
+public struct EventBlobAttestation has store, copy, drop {
```
karlwuest: Suggestion: Move this to `event_blob.move`, change its name to something like `EventBlobIndex` and also add a sequence number per event blob (this can be per-epoch). Then we can also use this `EventBlobIndex` in the `EventBlob` to ensure that each node can only have one active vote at a time. See comment on `system_state_inner::certify_event_blob` for context.

contracts/walrus/sources/system/system_state_inner.move
sadhansood: @karlwuest - I think we already prevent that attack. In this code block:
```assert!(
                attestation.last_attested_event_blob_epoch() < self.epoch() ||
                    attestation.last_attested_event_blob_checkpoint_seq_num()
                        <= certified_checkpoint_seq_num,
                EIncorrectAttestation,
            );
```
If the node's previous blob `attestation` is in the current epoch, then we assert that the checkpoint sequence number of that previous `attestation` has to be lower or equal to the certified checkpoint sequence number. So, for example: latest `certified_checkpoint_seq_num` = 100, malicious node `M` attests next blob with sequence num 200 and tries to attest another blob with sequence num 300 but would fail (since its previous attestaion of blob with seq num 200 hasn't gotten certified yet). There is a test which tests something similar `test_multiple_event_blobs_in_flight`

contracts/walrus/sources/system/system_state_inner.move
sadhansood: Both `BlobRegistered` and `BlobCeritifed` events contain this object id, and if we burn this blob those events refer to a non existent object. It is probably fine, but do we benefit in any way by keeping it and being consistent in our handling of those events?

karlwuest: There should not be any issues with having events that refer to a non-existing blob, because we anyway cannot rely on the blob object existing (or being accessible, could be wrapped) for any of the blobs.

contracts/walrus/sources/system/system_state_inner.move
karlwuest: You are right. I misread that check, sorry.

contracts/walrus/sources/system/bls_aggregate.move
sadhansood: fixed it

contracts/walrus/sources/system/event_blob.move
sadhansood: done

contracts/walrus/sources/system/system_state_inner.move
sadhansood: ok, burned the blob

contracts/walrus/sources/staking/storage_node.move
sadhansood: done

contracts/walrus/sources/system/event_blob.move
sadhansood: not relevant

contracts/walrus/sources/system/bls_aggregate.move
sadhansood: fixed

contracts/walrus/sources/system/system_state_inner.move
sadhansood: fixed

PR #810, opened 2024-09-16
test: use Mysten failpoints instead of the fail crate for failure injection tests

### LINKED ISSUE #726: Migrate the usage of fail crate to sui internal fail_point
So that we can use the same failpoint implementation for both simtest and failure injection test.

halfprice:
This PR replaces the use of fail/failpoints with Mysten's fail point implementation in place.

The TODO also mentioned to move these tests to walrus-simtest (or walrus-e2e-tests), and it turns out that it is very complicated to do so. These tests are not high level integration tests, and they currently use many crate-only functions. 

Once epoch change is implemented, we can add more high level integration tests in walrus-simtest uses the same failpoints.

close #726

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -2476,7 +2476,7 @@ mod tests {
         // places to break the sync process.
         // TODO(#705): make shard sync parameters configurable.
         simtest_param_test! {
-            simtest_sync_shard_start_from_progress: [
+            simtest_sync_shard_start_from_progress -> TestResult: [
                 primary1: (1, SliverType::Primary),
                 primary5: (5, SliverType::Primary),
                 primary10: (10, SliverType::Primary),
@@ -2491,12 +2491,14 @@ mod tests {
                 secondary23: (23, SliverType::Secondary),
             ]
         }
-        async fn simtest_sync_shard_start_from_progress(break_index: u64, sliver_type: SliverType) {
+        async fn simtest_sync_shard_start_from_progress(
+            break_index: u64,
+            sliver_type: SliverType,
+        ) -> TestResult {
             telemetry_subscribers::init_for_testing();
 
-            let (cluster, blob_details, shard_storage_dst) = setup_cluster_for_shard_sync_tests()
-                .await
-                .expect("Failed to setup cluster");
+            let (cluster, blob_details, shard_storage_dst) =
+                setup_cluster_for_shard_sync_tests().await?;
 
             register_fail_point_arg(
                 "fail_point_fetch_sliver",
@@ -2509,13 +2511,10 @@ mod tests {
                 .storage_node
                 ._shard_sync_handler
                 .start_new_shard_sync(ShardIndex(0))
-                .await
-                .expect("Failed to start shard sync");
+                .await?;
 
             // Waits for the shard sync process to stop.
-            wait_until_no_sync_tasks(&cluster.nodes[1].storage_node._shard_sync_handler)
-                .await
-                .expect("Failed to wait for shard sync");
+            wait_until_no_sync_tasks(&cluster.nodes[1].storage_node._shard_sync_handler).await?;
 
             // Check that shard sync process is not finished.
             let shard_storage_src = cluster.nodes[0]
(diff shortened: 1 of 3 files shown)


crates/walrus-test-utils/src/lib.rs
```diff
@@ -166,6 +166,43 @@ macro_rules! async_param_test {
...
+        simtest_param_test!( $func_name -> (): [ $( #[sim_test] $case_name: ($($args),+) ),* ] );
+    };
+    ($func_name:ident -> $return_ty:ty: [
+        $( $case_name:ident: ( $($args:expr),+ ) ),+$(,)?
+    ]) => {
+        simtest_param_test!(
+            $func_name -> $return_ty: [ $( #[sim_test] $case_name: ( $($args),+ ) ),* ]
+        );
+    }
+}
```
mlegner: AFAICT, the only difference between this and the `async_param_test` is the default `#[sim_test]` instead of `#[tokio::test]` attribute.

Suggestion: 
```suggestion
/// Macro for creating parametrized *simtest* tests.
///
/// This is identical to the [`async_param_test`], except that it annotates the tests with
/// `#[sim_test]` instead of `#[tokio::test]`.
#[macro_export]
macro_rules! simtest_param_test {
    ($func_name:ident: [
        $( $case_name:ident: ( $($args:expr),+ ) ),+$(,)?
    ]) => {
        $crate::async_param_test!(
            $func_name -> (): [ $( #[sim_test] $case_name: ($($args),+) ),* ]
        );
    };
    ($func_name:ident -> $return_ty:ty: [
        $( $case_name:ident: ( $($args:expr),+ ) ),+$(,)?
    ]) => {
        $crate::async_param_test!(
            $func_name -> $return_ty: [ $( #[sim_test] $case_name: ( $($args),+ ) ),* ]
        );
    }
}
```

crates/walrus-service/src/node.rs
mlegner, line 1: Important: We currently run simtests with the command `cargo simtest --run-ignored all simtest`, see [here](https://github.com/MystenLabs/walrus/blob/78b2c7b8b8432b9e351ece60f5142447993d962a/.github/workflows/code.yml#L162) and [here](https://github.com/MystenLabs/walrus/blob/78b2c7b8b8432b9e351ece60f5142447993d962a/CONTRIBUTING.md#simtests).

This only includes tests that have "simtest" somewhere in the name. So we either need to change that or make sure that all simtests are named according to that scheme.

crates/walrus-service/src/node.rs
```diff
@@ -2457,8 +2460,8 @@ mod tests {
         // Note that currently, each sync batch contains 10 blobs. So testing various interesting
         // places to break the sync process.
         // TODO(#705): make shard sync parameters configurable.
-        async_param_test! {
-            sync_shard_start_from_progress -> TestResult: [
+        simtest_param_test! {
+            sync_shard_start_from_progress: [
```
mlegner: Question: What is the reason to remove the return value? This seems to cause quite a lot of additional "expect"s in the code.

crates/walrus-test-utils/src/lib.rs
halfprice: Applied

crates/walrus-service/src/node.rs
halfprice: Thanks for the info! I added `simtest` to the test names affected in this PR, but I suspect that this may get overlooked again.

I think `cargo simtest --run-ignored all` is probably just fine?

crates/walrus-service/src/node.rs
halfprice: What I encountered originall was that, simtest requires function output to be a `send` (https://github.com/MystenLabs/mysten-sim/blob/b320996d8dfb99b273fe31c0222c659332283c99/msim/src/sim/runtime/mod.rs#L461). I tried to make `TestResult` a `send`, but it still has error that `ProtocolKeyPair::from_str` is not send.

It turns out that all I need to do is to add `send` to KeyPairParseError (see the change in `crates/walrus-core/src/keys.rs`). How simtest can also return TestResult.

crates/walrus-service/src/node.rs
mlegner: Unfortunately without the filter I get errors like "not implemented: this test can only be run in tokio" for many test cases... 🤷

halfprice: I see. Yeah, simtest automatically adds a `ignore` to non-sim async test. But by running with `--run-ignored all`, we re-enable them... That's why I tried `cargo simtest`, and it was just fine.

halfprice: created https://github.com/MystenLabs/walrus/pull/823

PR #816, opened 2024-09-16
feat: separate the ExponentialBackoff struct from its state

jpcsmith:
This enables using a temporary borrow of an Rng to advance the state, as rng instances are quite large (~350 B).

### DIFF
--- a/crates/walrus-service/src/common/utils.rs
+++ b/crates/walrus-service/src/common/utils.rs
@@ -19,12 +19,7 @@ use anyhow::{Context as _, Result};
 use futures::{future::FusedFuture, FutureExt};
 use pin_project::pin_project;
 use prometheus::{HistogramVec, Registry};
-use rand::{
-    distributions::{DistIter, Uniform},
-    rngs::StdRng,
-    Rng,
-    SeedableRng,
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/common/utils.rs
```diff
@@ -87,6 +82,51 @@ pub(crate) trait BackoffStrategy {
...
+    pub fn new(min_backoff: Duration, max_backoff: Duration, max_retries: Option<u32>) -> Self {
+        Self {
+            min_backoff,
+            max_backoff,
+            sequence_index: 0,
+            max_retries,
+        }
+    }
+
+    pub fn next_delay<R: Rng + ?Sized>(&mut self, rng: &mut R) -> Option<Duration> {
```
giac-mysten, line 106: Question: Why is the `?Sized` widening bound required here? The borrow checker does not complain if I remove it.

jpcsmith: It's not required, it's makes the interface more permissive. I added it as it's default on a lot of functions in the `rand` crate and I was not sure when it may be useful.

PR #823, opened 2024-09-18
test: create `walrus_simtest` proc macros to append `simtest` prefix to all walrus simtest

halfprice:
Currently, all integration tests are marked as ignored, but CI still run them by setting `--run-ignored` to true. This means that for non-simtest that `[sim_test]` attributes sets to ignore, it will be run and then fail.

This PR adds a `[walrus_simtest]` attributes which appends a `simtest_` prefix to all the simtests (which we already manually do today), and then invoke the `[sim_test]` attributes.

Also updates the cargo-simtest script that `cargo nextest` only runs tests containing `simtest` in its name. Now, we can do

`cargo simtest` to run all the cheap simtests and `cargo simtest --run-ignored all` to run all the integration simtest.

### DIFF
--- a/crates/walrus-test-utils/src/lib.rs
+++ b/crates/walrus-test-utils/src/lib.rs
@@ -198,7 +198,7 @@ macro_rules! async_param_test {
 
 /// Macro for creating parametrized *simtest* tests.
 ///
-/// Note that this macro reuses [`async_param_test`] macro, but use `#[sim_test]` instead of
+/// Note that this macro reuses [`async_param_test`] macro, but use `#[walrus_simtest]` instead of
 /// `#[tokio::test]` attribute to each test case.
 #[macro_export]
 macro_rules! simtest_param_test {
@@ -409,6 +409,7 @@ mod tests {
         ]
     }
     async fn async_sum_no_return_with_shared_meta(lhs: usize, rhs: usize, total: usize) {
+        #[cfg(not(msim))]
         tokio::time::resume(); // Panics if not paused.
         assert_eq!(lhs + rhs, total);
     }
@@ -425,6 +426,7 @@ mod tests {
         rhs: usize,
         total: usize,
     ) -> Result<(), Box<dyn Error>> {
+        #[cfg(not(msim))]
         tokio::time::resume(); // Panics if not paused.
         assert_eq!(lhs + rhs, total);
         Ok(())
@@ -457,6 +459,7 @@ mod tests {
         rhs: usize,
         total: usize,
     ) -> Result<(), Box<dyn Error>> {
+        #[cfg(not(msim))]
         tokio::time::resume(); // Panics if not paused.
         assert_eq!(lhs + rhs, total);
         Ok(())


CONTRIBUTING.md
mlegner, line 1: Nit: Line 103 is no longer applicable, can you remove that?

Also, lines 100-101 are no longer fully true and could be slightly changed.

crates/walrus-service/Cargo.toml
```diff
@@ -107,6 +107,7 @@ utoipa = { workspace = true, features = ["axum_extras"] }
 utoipa-redoc.workspace = true
 walrus-core = { workspace = true, features = ["utoipa"] }
 walrus-event.workspace = true
+walrus-proc-macros.workspace = true
```
mlegner: Question: Does that have to be a full dependency?

crates/walrus-test-utils/src/lib.rs
```diff
@@ -175,13 +175,15 @@ macro_rules! simtest_param_test {
     ($func_name:ident: [
         $( $case_name:ident: ( $($args:expr),+ ) ),+$(,)?
     ]) => {
-        async_param_test!( $func_name -> (): [ $( #[sim_test] $case_name: ($($args),+) ),* ] );
+        async_param_test!(
+            $func_name -> (): [ $( #[walrus_simtest] $case_name: ($($args),+) ),* ]
```
mlegner, line 209: Nit: The docstring above still mentions `sim_test` instead of `walrus_simtest`.

crates/walrus-proc-macros/src/lib.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    // Extract the function name
+    let fn_name = &input.sig.ident;
+
+    // Create the new name by adding "simtest_" prefix
+    let new_fn_name = syn::Ident::new(&format!("simtest_{}", fn_name), fn_name.span());
+
+    input.sig.ident = new_fn_name;
+    let output = quote! {
+        #[sim_test(#(#args)*)]
```
mlegner: Suggestion: Can we use `sui_macros::sim_test` here? Then we don't have to explicitly import `sim_test` anywhere and can remove the imports in `crates/walrus-service/src/node.rs` and `crates/walrus-simtest/tests/simtest.rs`.

crates/walrus-proc-macros/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+publish = false
+authors.workspace = true
+edition.workspace = true
+license.workspace = true
+version.workspace = true
+
+[dependencies]
+quote.workspace = true
+syn.workspace = true
+
```
mlegner, line 13: Suggestion: Add `sui-macros` here:
```suggestion
sui-macros.workspace = true

```

CONTRIBUTING.md
halfprice: Updated.

crates/walrus-proc-macros/src/lib.rs
halfprice: Good suggestion. Done.

crates/walrus-service/Cargo.toml
halfprice: Move to dev-dependency.

crates/walrus-test-utils/src/lib.rs
halfprice: Updated

PR #825, opened 2024-09-19
feat(node): update blob-status tracking for new events

### LINKED ISSUE #762: Connect new blob status tracking with new events
Follow-up on #717.

mlegner:
- Add `BlobDeleted` event.
- Include blob extensions and deletions into status tracking.
- Clean up after blob deletion if there are no more registered or certified blobs.

Manual testing and added E2E test are successful.

Closes #762 
Contributes to #718

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -474,6 +474,9 @@ impl StorageNode {
         let next_index: usize = from_element_index.try_into().expect("64-bit architecture");
         let index_stream = stream::iter(next_index..);
         let mut indexed_element_stream = index_stream.zip(event_stream);
+        // Important: Events must be handled consecutively and in order to prevent (intermittent)
+        // invariant violations and interference between different events. See, for example,
+        // `Self::cancel_sync_and_mark_certified_event_completed`.
         while let Some((element_index, stream_element)) = indexed_element_stream.next().await {
             let span = tracing::info_span!(
                 parent: None,
@@ -640,9 +643,14 @@ impl StorageNode {
                 self.cancel_sync_and_mark_certified_event_completed(&blob_id)
                     .await?;
             }
+            // Note that this function is called *after* the blob info has already been updated with
+            // the event. So it can happen that the only registered blob was deleted and the blob is
+            // now no longer registered.
             if !blob_info.is_registered() {
                 self.inner.storage.delete_blob(&event.blob_id, true)?;
             }
+        } else {
+            tracing::warn!(%blob_id, "handling `BlobDeleted` event for untracked blob");
         }
 
         self.inner
@@ -668,8 +676,9 @@ impl StorageNode {
     /// Cancels any existing blob syncs for the provided `blob_id` and marks the corresponding
     /// events as completed.
     ///
-    /// To avoid race conditions, this function must not be run concurrently with any logic for
-    /// processing events from the stream, so a cancelled event cannot be restarted.
+    /// To avoid interference with later events (e.g., cancelling a sync initiated by a later event
+    /// or immediately restarting the sync that is cancelled here), this function should be called
+    /// before any later events are being handled.
     async fn cancel_sync_and_mark_certified_event_completed(
         &self,
         blob_id: &BlobId,
@@ -679,7 +688,6 @@ impl StorageNode {
         {
             // Advance the event cursor with the event of the cancelled sync. Since the blob is
             // invalid the associated blob certified event is completed without a sync.
-            //
             self.inner
                 .mark_event_completed(cancelled_event_index, &cancelled_event_id)?;
         }
@@ -1246,11 +1254,16 @@ mod tests {
 
     async_param_test! {
         deletes_blob_data_on_event -> TestResult: [
-            invalid_blob_event: (InvalidBlobId::for_testing(BLOB_ID).into()),
-            blob_deleted_event: (BlobDeleted::for_testing(BLOB_ID).into()),
+            invalid_blob_event_registered: (InvalidBlobId::for_testing(BLOB_ID).into(), false),
+            blob_deleted_event_registered: (
+                BlobDeleted{was_certified: false, ..BlobDeleted::for_testing(BLOB_ID)}.into(),
+                false
+            ),
+            invalid_blob_event_certified: (InvalidBlobId::for_testing(BLOB_ID).into(), true),
+            blob_deleted_event_certified: (BlobDeleted::for_testing(BLOB_ID).into(), true),
         ]
     }
-    async fn deletes_blob_data_on_event(event: BlobEvent) -> TestResult {
+    async fn deletes_blob_data_on_event(event: BlobEvent, is_certified: bool) -> TestResult {
         let events = Sender::new(48);
         let node = StorageNodeHandle::builder()
             .with_storage(populated_storage(&[
@@ -1265,6 +1278,7 @@ mod tests {
 
         tokio::time::sleep(Duration::from_millis(50)).await;
 
+        assert!(storage.is_stored_at_all_shards(&BLOB_ID)?);
         events.send(
             BlobRegistered {
                 deletable: true,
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -618,27 +631,62 @@ impl StorageNode {
...
 
+    #[tracing::instrument(skip_all)]
+    async fn process_blob_deleted_event(
+        &self,
+        event_index: usize,
+        event: BlobDeleted,
+    ) -> anyhow::Result<()> {
+        let blob_id = event.blob_id;
+
+        if let Some(blob_info) = self.inner.storage.get_blob_info(&blob_id)? {
```
halfprice, line 641: Suggestion: Is it normal if `get_blob_info` returns None? Should we log something here?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -533,6 +578,28 @@ impl BlobInfoApi for BlobInfoV1 {
...
 
+    fn is_registered(&self) -> bool {
+        matches!(
+            self,
+            Self::Valid(ValidBlobInfoV1 {
+                count_deletable_total,
+                permanent_total: Some(_),
+                ..
+            }) if *count_deletable_total > 0
+        )
```
halfprice: Looks like the logic here is `registered deletable blob count > 0` AND `has permanent registered`. But what you want is OR?

Same for `is_certified` below.

crates/walrus-service/src/node.rs
```diff
@@ -618,27 +631,62 @@ impl StorageNode {
...
+            if !blob_info.is_certified() {
+                self.cancel_sync_and_mark_certified_event_completed(&blob_id)
+                    .await?;
+            }
+            if !blob_info.is_registered() {
```
halfprice, line 649: Question: is it possible to receive blob deleted event without a registered event? If we check `register after delete` here, should we also consider dealing with `certify after delete` case?

crates/walrus-service/src/node.rs
```diff
@@ -618,27 +631,62 @@ impl StorageNode {
...
+        self.inner
+            .mark_event_completed(event_index, &event.event_id)?;
+        Ok(())
+    }
+
+    /// Cancels any existing blob syncs for the provided `blob_id` and marks the corresponding
+    /// events as completed.
+    ///
+    /// To avoid race conditions, this function must not be run concurrently with any logic for
+    /// processing events from the stream, so a cancelled event cannot be restarted.
```
halfprice: Suggestion: consider using a `Mutex` to guarantee this property.

crates/walrus-service/src/node.rs
```diff
@@ -1215,8 +1268,24 @@ mod tests {
...
+
+        events.send(
+            BlobRegistered {
+                deletable: true,
+                ..BlobRegistered::for_testing(BLOB_ID)
+            }
+            .into(),
+        )?;
+        events.send(
+            BlobCertified {
```
halfprice: Thought: consider adding a registered only test case.

crates/walrus-service/src/node.rs
```diff
@@ -618,27 +631,62 @@ impl StorageNode {
...
+
     #[tracing::instrument(skip_all)]
     async fn process_blob_invalid_event(
         &self,
         event_index: usize,
         event: InvalidBlobId,
+    ) -> anyhow::Result<()> {
+        self.cancel_sync_and_mark_certified_event_completed(&event.blob_id)
+            .await?;
+        self.inner.storage.delete_blob(&event.blob_id, false)?;
```
halfprice, line 670: It seems that delete an invalid blob is a no-op at the moment?

crates/walrus-service/src/node.rs
```diff
@@ -618,27 +631,62 @@ impl StorageNode {
...
+            self.blob_sync_handler.cancel_sync(blob_id).await?
         {
             // Advance the event cursor with the event of the cancelled sync. Since the blob is
             // invalid the associated blob certified event is completed without a sync.
             //
-            // Race condition is avoided here by the fact that process_blob_invalid_event is not
-            // run concurrently with any logic for processing events from the stream, so a
-            // cancelled event cannot be restarted.
             self.inner
                 .mark_event_completed(cancelled_event_index, &cancelled_event_id)?;
```
halfprice, line 692: Nit (the issue exists before this PR): I feel this is better to be done inside `cancel_sync`, since there is no point to not mark event completed if the sync task is cancelled.

crates/walrus-service/src/node.rs
mlegner: Good point. Indeed, we should always have the blob info if we see a `BlobDeleted` event. I will add a warning.

crates/walrus-service/src/node.rs
mlegner: No, we should always have a `BlobRegistered` event before a `BlobDeleted` event.

However, here we are checking the blob info *after* it has been updated with the deletion event. At this point it could have happened that the last registered blob has been removed so the blob is no longer registered.

I can add a comment to the code here.

crates/walrus-service/src/node.rs
mlegner: Can you elaborate on this observation? The blob data (metadata and slivers) should be removed for invalid blobs.

crates/walrus-service/src/node.rs
mlegner: Good point, I'll see if this is feasible.

crates/walrus-service/src/node.rs
mlegner: Hmm... Are we sure that there won't be cases where we want to cancel the sync for other reasons?

cc @jpcsmith

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Indeed, thanks a lot for pointing this out! 🤦

crates/walrus-service/src/node.rs
jpcsmith: So a clean shutdown cancels pending tasks. In that case we should not mark the event as completed.

mlegner: Alright, thanks for the context. In this case I suggest to leave it as it is now.

crates/walrus-service/src/node.rs
mlegner: Actually, I think this would be quite complicated to do properly. I went for more details in the docstring and added a comment to the `process_events` function to make sure we keep handling events consecutively in the future.

However, there is potentially another issue: What happens if during shard sync/recovery we attempt to sync or recover a sliver and during that process the corresponding blob is either marked as invalid or deleted?

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: I think the updated version should be what we want:
```rust
        matches!(
            self,
            Self::Valid(ValidBlobInfoV1 {
                count_deletable_total,
                permanent_total,
                ..
            }) if *count_deletable_total > 0 || permanent_total.is_some()
        )
```

crates/walrus-service/src/node.rs
halfprice: It doesn't look like we call `cancel_sync` when doing a clean shutdown. It seems that the only places it is called are when processing invalid blob and processing deleted blob.

The reason I think it's better done inside `cancel_sync` is because BlobSyncHandler task is strongly tight to an event. The current style requires the future usage of `cancel_sync` to also ***must*** mark event complete. Any future usage of `cancel_sync` without marking event complete stops moving cursor forward (since there is no mechanism to restart the sync after calling `cancel_sync`).

crates/walrus-service/src/node.rs
halfprice: Thanks for the explanation! This makes sense now.

crates/walrus-service/src/node.rs
halfprice: The question was due to that I missed the subtly that we process the blob_info before calling `process_blob_XXX_event`... I thought we do nothing when deleting an invalid blob... All good now.

crates/walrus-service/src/node.rs
halfprice: Yeah, I can imagine there are places to make this tricky to implement.

Not sure if I follow the potential issue though. For blob sync, I don't think we should hold the mutex during the entire sync task.

PR #826, opened 2024-09-19
fix(events): more thorough verification of events in sui checkpoints

sadhansood:
When reading downloaded checkpoint for events, we ensure that it is the right checkpoint summary (by verifying it is signed by the right committee and points to the previous verified checkpoint) but do not check checkpoint's content and events. This PR fixes this issue and adds the missing event verification.

### DIFF
--- a/crates/walrus-event/src/event_processor.rs
+++ b/crates/walrus-event/src/event_processor.rs
@@ -30,6 +30,9 @@ use sui_storage::verify_checkpoint_with_committee;
 use sui_types::{
     base_types::ObjectID,
     committee::Committee,
+    effects::TransactionEffectsAPI,
+    full_checkpoint_content::CheckpointData,
+    message_envelope::Message,
     messages_checkpoint::{TrustedCheckpoint, VerifiedCheckpoint},
     object::Object,
     SYSTEM_PACKAGE_ADDRESSES,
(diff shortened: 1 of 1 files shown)


crates/walrus-event/src/event_processor.rs
```diff
@@ -154,6 +156,64 @@ impl EventProcessor {
...
+
+            if let Some((events_digest, events)) = transaction
+                .effects
+                .events_digest()
+                .zip(transaction.events.as_ref())
+            {
+                if *events_digest != events.digest() {
+                    bail!("Events digest does not match");
+                }
+            }
```
halfprice, line 285: Is this part still necessary if the event digest match?

crates/walrus-event/src/event_processor.rs
```diff
@@ -154,6 +156,64 @@ impl EventProcessor {
...
+        )
+        .map_err(|checkpoint| {
+            anyhow!(
+                "Failed to verify sui checkpoint: {}",
+                checkpoint.sequence_number
+            )
+        })?;
+
+        // Verify that checkpoint summary matches the content
+        if verified_checkpoint.content_digest != *checkpoint.checkpoint_contents.digest() {
```
halfprice, line 255: There is also a `verify_with_contents` in CertifiedCheckpointSummary

crates/walrus-event/src/event_processor.rs
sadhansood: Yes good question, once the effects digest match, we compare the events digest to ensure they match with the events in the `CheckpointData` struct.

crates/walrus-event/src/event_processor.rs
sadhansood: It is pretty much the same thing but we have `TrustedCheckpoint` in here whereas that function takes `CertifiedCheckpoint`, so a little bit of extra conversion would be required and hence this one liner makes sense

PR #828, opened 2024-09-20
refactor: allow specifying attributes for all param tests

jpcsmith:


crates/walrus-test-utils/src/lib.rs
```diff
@@ -124,46 +124,76 @@ macro_rules! param_test {
...
+///     Ok(())
+/// }
+/// ```
 ///
 /// See [`param_test`] for more information and examples.
 #[macro_export]
 macro_rules! async_param_test {
-    ($func_name:ident -> $return_ty:ty: [
-        $( $(#[$outer:meta])+ $case_name:ident: ( $($args:expr),+ ) ),+$(,)?
+    // Macro uses 'internal rules' to avoid any difficulty with imports. Additionally, we use
```
giac-mysten, line 156: Question: what do you mean by "difficulty with imports"?

jpcsmith: If I split this into multiple macros, they all need to be public, even though they are not used by themselves. Seems [there is a hit to compile time](https://veykril.github.io/tlborm/decl-macros/patterns/internal-rules.html) for doing things the way I did here. We can always split it if need be.

PR #834, opened 2024-09-20
fix(client): avoid reading `Blob` object after certification

### LINKED ISSUE #728: Make blob certification more robust against fullnode load balancing
Similar to #587.

See https://github.com/MystenLabs/walrus-sites/issues/178#issuecomment-2309606160 for context.

mlegner:
The only purpose of this was to check that the `certified_epoch` was updated correctly. However, we *know* that this must have happened if the `certify_blob` contract completed without error. We can therefore simply modify the `Blob` object locally and safe one RPC request.

This speeds up the write process and avoids spurious errors in case of load-balancing between Sui full nodes.

Closes #728

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -15,10 +15,7 @@ use futures::Future;
 use rand::{seq::SliceRandom, thread_rng};
 use reqwest::Client as ReqwestClient;
 use sui_types::base_types::ObjectID;
-use tokio::{
-    sync::Semaphore,
-    time::{sleep, Duration},
-};
+use tokio::{sync::Semaphore, time::Duration};
 use tracing::{Instrument, Level};
 use walrus_core::{
     encoding::{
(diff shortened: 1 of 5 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -164,11 +164,13 @@ pub trait ContractClient: Send + Sync {
 
     /// Certifies the specified blob on Sui, given a certificate that confirms its storage and
     /// returns the certified blob.
+    // NB: This intentionally takes an owned `Blob` object even though it is not required, as the
+    // corresponding object on Sui will be changed in the process.
```
giac-mysten, line 168: Nice idea, I like that we are "destroying" the outdated object this way.

crates/walrus-service/src/client.rs
```diff
@@ -272,30 +269,30 @@ impl<T: ContractClient> Client<T> {
...
+        // We do not need to wait explicitly as we anyway retry all requests to storage nodes.
+        let certificate = self
+            .send_blob_data_and_get_certificate(&metadata, &pairs)
+            .await?;
+        self.sui_client
+            .certify_blob(blob.clone(), &certificate)
             .await
             .map_err(|e| ClientError::from(ClientErrorKind::CertificationFailed(e)))?;
 
+        blob.certified_epoch = Some(self.committee.epoch);
```
giac-mysten, line 286: Remark: with epoch change this may be problematic, for example if the retry is within `send_blob_data_and_get_certificate`. Should keep this in mind.

mlegner: Yes, that's a good point...

I'll add a comment about this.

PR #836, opened 2024-09-20
fix: Make sure duplicate events are not processed or written in blob

sadhansood:
Thanks to @halfprice, we discovered a couple issues which are now being fixed in this PR:
1. event_blob_writer was deleting the local dir on startup
2. event_sequencer was being fed events which it may have already committed. This could happen when event blob writer is behind event sequencer and upon restart, we start from the event index which has already been committed by the sequencer.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -15,7 +15,7 @@ use shard_sync::ShardSyncHandler;
 use sui_types::{digests::TransactionDigest, event::EventID};
 use tokio::{select, time::Instant};
 use tokio_util::sync::CancellationToken;
-use tracing::{field, Instrument};
+use tracing::{field, info, Instrument};
 use typed_store::{rocks::MetricConf, DBMetrics, TypedStoreError};
 use walrus_core::{
     encoding::{EncodingAxis, EncodingConfig, RecoverySymbolError},
@@ -492,6 +492,7 @@ impl StorageNode {
                 "error.type" = field::Empty,
             );
             let cloned_stream_element = stream_element.clone();
+            info!("processing event: {}", element_index);
             async move {
                 self.process_stream_element(storage, element_index, cloned_stream_element)
                     .await
@@ -534,6 +535,7 @@ impl StorageNode {
             // if event blob writer is behind event sequencer
             return Ok(());
         }
+        info!("stream processing event: {:?}", stream_element);
         let _timer_guard = &self
             .inner
             .metrics
@@ -1117,7 +1119,7 @@ where
 mod tests {
 
     use std::{sync::OnceLock, time::Duration};
-
+    use clap::builder::TypedValueParser;
     use storage::{
         tests::{populated_storage, WhichSlivers, BLOB_ID, OTHER_SHARD_INDEX, SHARD_INDEX},
         ShardStatus,
@@ -1220,7 +1222,7 @@ mod tests {
     async fn services_slivers_for_shards_managed_according_to_committee() -> TestResult {
         let shard_for_node = ShardIndex(0);
         let node = StorageNodeHandle::builder()
-            .with_system_event_provider(vec![BlobRegistered::for_testing(BLOB_ID).into()])
+            .with_system_event_manager(vec![BlobRegistered::for_testing(BLOB_ID).into()])
             .with_shard_assignment(&[shard_for_node])
             .with_node_started(true)
             .build()
@@ -1245,7 +1247,7 @@ mod tests {
                 (SHARD_INDEX, vec![(BLOB_ID, WhichSlivers::Both)]),
                 (OTHER_SHARD_INDEX, vec![(BLOB_ID, WhichSlivers::Both)]),
             ])?)
-            .with_system_event_provider(events.clone())
+            .with_system_event_manager(events.clone())
             .with_node_started(true)
             .build()
             .await?;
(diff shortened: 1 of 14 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -554,34 +503,122 @@ impl StorageNode {
...
+        stream_element: IndexedStreamElement,
+    ) -> Result<(), anyhow::Error> {
+        let next_sequencer_index: usize = self
+            .inner
+            .storage
+            .get_sequentially_processed_event_count()?
+            .try_into()
+            .expect("64-bit architecture");
+        if element_index < next_sequencer_index {
+            // It is possible to receive committed events upon restart
```
halfprice, line 534: I think the issue is not receiving committed events, but rather the index is off.

When we initialize the event stream [here](https://github.com/MystenLabs/walrus/blob/32cf705d6088a27fb65982b8d44c68f55f3783c0/crates/walrus-service/src/node.rs#L470), we always starts from the last recorded event in the EventCursorTable. So it actually won't receive unprocessed (marked as complete) event.

Say we have event stream `e0(done processing), e1(done processing), e2(done processing), [crash], e3`. Upon restart, `e3` is always going to be the first event that will be sent out, and we want it to use index `3`. However, if we do min(last cursor index, blob writer index), we may assign index smaller than 3 to e3.


So I think the actual fix might be that in addition to the skipping logic here, we also want
```
let event_stream = Box::into_pin(self.inner.event_manager.events(event_cursor).await?);
```
to also take blob writer's last commit index into account when initializing the stream, so that we can start from the correct events matching the index.

crates/walrus-service/src/node/storage/event_blob_writer.rs
```diff
@@ -321,13 +324,13 @@ impl EventBlobWriter {
...
-            .latest_committed_event_index()
-            .map_or(false, |i| i >= element_index)
+            .latest_event_index
+            .or(self.latest_committed_event_index)
+            .map_or(false, |i| element_index <= i)
         {
-            // It is possible to receive committed events upon restart
-            return Ok(());
+            // This ensures we do not write the same event twice.
+            panic!("index repeated in blob writer: ({element_index})");
```
halfprice, line 332: I think this panic may still get hit? Say the event cursor is not moving due to a long running BlobSync task, but blob writer has written following events. Upon restart, we will start from the index of the BlobSync task, which is smaller than the blob writer's latest index.

crates/walrus-service/src/node.rs
```diff
@@ -492,60 +493,8 @@ impl StorageNode {
             );
             let cloned_stream_element = stream_element.clone();
             async move {
```
jpcsmith, line 496: Nit: The `async move` block is no longer necessary, as the `inspect` and `instrument` calls can be done directly on the future returned by `process_stream_element` .

crates/walrus-service/src/node.rs
```diff
@@ -554,34 +503,122 @@ impl StorageNode {
...
+    }
+
+    #[tracing::instrument(skip_all)]
+    fn next_event_index(&self, event_blob_writer: &EventBlobWriter) -> anyhow::Result<u64> {
         let storage = &self.inner.storage;
-        let last_committed_sequencer_index: Option<u64> =
-            storage.get_event_cursor()?.map(|(index, _)| index);
-        let last_committed_event_blob_writer_index: Option<u64> =
-            event_blob_writer.latest_committed_event_index();
+        let next_sequencer_index: Option<u64> = storage.get_event_cursor()?.map(|(index, _)| index);
```
jpcsmith, line 617: Question: When i + 1 for the blob writer but only i for the sequencer?

crates/walrus-service/src/node.rs
jpcsmith: Before the addition of the BlobWriter, the system had the invariant that the event cursor stored in the database was the last event that was completely processed, and so we could alway start from that event. It would be nice if we could maintain this invariant and avoid having each subsystem needing to re-process different segments of the event stream on restart.

My suggestion would be to have the BlobWriter process events before the event is marked as complete, and ensure that it is idempotent for repeated calls (to handle failures between processing and commit of the event).

This would also require that once the BlobWriter has successfully processed an event, its internal state is persisted so that failures after the event is committed to the database do not require the blobwriter to process the event again, as is currently the case.

crates/walrus-service/src/node.rs
```diff
@@ -554,34 +503,122 @@ impl StorageNode {
...
+        element_index: usize,
+        stream_element: IndexedStreamElement,
+    ) -> Result<(), anyhow::Error> {
+        if event_blob_writer
+            .latest_committed_event_index()
+            .map_or(false, |i| i >= element_index as u64)
+        {
+            // It is possible to receive committed events upon restart
+            return Ok(());
+        }
```
jpcsmith, line 607: Suggestion: Move this and the similar check in `process_stream_event` to the outer loop and only check against the event at start

That is, instead of the following pseudocode which is roughly equivalent to the current approach
```rust
loop {
   ..
   let most_recent_blob_writer_event = ...;
   if current_event > most_recent_blob_writer_event {
       store_event_in_blob(..)
   } 

   let most_recent_processed_event = ...;
   if current_event > most_recent_processed_event {
      process_stream_element(..)
   } 
````
we would have
```rust
let last_blob_writer_event_before_restart = ...;
let last_general_event_before_restart = ...;
loop {
   ..
   if current_event > last_blob_writer_event_before_restart {
       store_event_in_blob(..)
   } 
   if current_event > last_general_event_before_restart {
      process_stream_element(..)
   } 
````

PR #837, opened 2024-09-22
test: deterministic simulation test with simulated walrus node crash and restart

halfprice:
The main goal of this PR is to make each Walrus node run in a dedicated simulated runtime, so that we can simulate
various individual node crash and recovery scenarios.

Added one simple test with one single node failure to illustrate the usage.

contribute #356

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -124,7 +124,7 @@ pub trait StorageNodeHandleTrait {
     /// Builds a new storage node handle.
     fn build(
         builder: StorageNodeHandleBuilder,
-        sui_cluster: Option<Arc<Mutex<PathBuf>>>,
+        sui_cluster_wallet_path: Option<Arc<Mutex<PathBuf>>>,
         system_context: Option<SystemContext>,
         storage_dir: TempDir,
     ) -> impl std::future::Future<Output = anyhow::Result<Self>> + Send
@@ -180,7 +180,7 @@ impl StorageNodeHandleTrait for StorageNodeHandle {
 
     async fn build(
         builder: StorageNodeHandleBuilder,
-        _sui_cluster: Option<Arc<Mutex<PathBuf>>>,
+        _sui_cluster_wallet_path: Option<Arc<Mutex<PathBuf>>>,
         _system_context: Option<SystemContext>,
         _storage_dir: TempDir,
     ) -> anyhow::Result<Self> {
@@ -244,7 +244,7 @@ impl StorageNodeHandleTrait for SimStorageNodeHandle {
 
     async fn build(
         builder: StorageNodeHandleBuilder,
-        sui_cluster: Option<Arc<Mutex<PathBuf>>>,
+        sui_cluster_wallet_path: Option<Arc<Mutex<PathBuf>>>,
         system_context: Option<SystemContext>,
         storage_dir: TempDir,
     ) -> anyhow::Result<Self>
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/test_utils.rs
```diff
@@ -393,6 +529,197 @@ impl StorageNodeHandleBuilder {
...
+        let startup_sender = Arc::new(startup_sender);
+        let handle = sui_simulator::runtime::Handle::current();
+        let builder = handle.create_node();
+        let node_handle = builder
+            .ip(ip)
+            .name(&format!(
+                "{:?}",
+                config.protocol_key_pair.get().unwrap().public()
+            ))
+            .init(move || {
```
halfprice, line 607: This is the main entrance of the sim node handle. A lot of weird looking changes in this PR is just to make sure that `whatever passed in here doesn't contain any initialized objects, e.g. no initialized event manager or storage`. Restarting a node invokes this function again, so we only want any `node starting input` here to simulate proper starting/restarting.

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -1,58 +1,163 @@
...
+                "batch-write-before",
+                "batch-write-after",
+                "put-cf-before",
+                "put-cf-after",
+                "delete-cf-before",
+                "delete-cf-after",
+            ],
+            move || {
+                handle_failpoint(do_not_fail_nodes.clone(), fail_triggered.clone(), 0.01);
+            },
```
halfprice, line 118: Here is an example of simulated crash/recovery. Those fail points already exist in DBMap in Sui (I believe added by @sadhansood). Basically what this means is that around DB updates, there is 1% of chance we will crash this node, and restarting after a random period, and we will only do this once in this test.

We can later add other failpoints anywhere in our codebase to trigger specific crash point.

crates/walrus-sui/src/test_utils.rs
```diff
@@ -256,6 +283,20 @@ pub async fn new_wallet_on_sui_test_cluster(
     Ok(wallet)
 }
 
+/// Returns a wallet on the global Sui test cluster as well as a handle to the cluster.
+pub async fn new_wallet_on_sui_test_cluster_from_path(
```
halfprice, line 287: These are the things I'll cleanup tomorrow...

crates/walrus-service/src/test_utils.rs
jpcsmith: Suggestion: Move the code used to startup the storage node from `bin/node.rs` into `src/node/cli.rs` and have both `bin/node.rs` and the `test_utils.rs` use the same code. Otherwise, this will go out of sync and our tests will stop testing our actual starting/restarting logic.

crates/walrus-service/src/test_utils.rs
```diff
@@ -135,6 +205,72 @@ impl AsRef<StorageNode> for StorageNodeHandle {
     }
 }
 
+/// A test configuration for a storage node.
+#[derive(Debug)]
+pub struct SimStorageNodeHandle {
```
jpcsmith, line 210: Question: Given that this stores pretty much the same thing as `StorageNodeHandle`, why do we need a separate struct?

crates/walrus-service/src/test_utils.rs
halfprice: Yes, that's a top clean up task. But looking at the code, I'm not sure if it should be done in this PR though. It is going to require quite some changes.

crates/walrus-service/src/test_utils.rs
halfprice: The difference in the StorageNodeHandleTrait implementation for these two struct is the main reason. Basically SimStorageNodeHandle removes `per instance objects` in StorageNodeHandle, such as `storage`, `rest api server`, `client` (each restarts changes these objects in Simulation), and added a sim task handle.

I had one version of the implementation that just uses StorageNodeHandle, and it was very messy and hard to read.

Another subtle issue is that we want to reuse StorageNodeHandle in simtest that `does not require per node failure (e.g. all walrus nodes run in the same runtime)`, so because of this, I also ruled out using cfg to select StorageNodeHandle members during compilation.

Issue #753, opened 2024-08-30
Client support for epoch change


giac-mysten:
Signals to the client that epoch change is under way:

* too many shards/nodes do not respond, or sliver stores fail with `ShardNotAssigned`; or
* the certification of the blob fails.

Then, the client can update the committee and try storing again from scratch. 

Since, when storing metadata and slivers, 

* we check which ones are already present; and 
* we re-use available storage and registered blob objects (#754);

we can simply restart the the storing process with the new committee.

PR #841, opened 2024-09-23
feat(sui): expose prev and next committee in read client

giac-mysten:
* Exposes the previous and next committees in the read client.
* Removes the unused epoch status.
* Exposes the `EpochState` in the read client.

Contributes to #753

### DIFF
--- a/crates/walrus-sui/src/client/read_client.rs
+++ b/crates/walrus-sui/src/client/read_client.rs
@@ -57,6 +57,15 @@ use crate::{
 const EVENT_MODULE: &str = "events";
 const MULTI_GET_OBJ_LIMIT: usize = 50;
 
+/// The current, previous, and next committee, and the current epoch state.
+#[derive(Debug)]
+pub struct CommitteesAndState {
+    pub current: Committee,
+    pub previous: Committee,
+    pub next: Option<Committee>,
+    pub epoch_state: EpochState,
+}
+
 /// Trait to read system state information and events from chain.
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/client/read_client.rs
```diff
@@ -84,6 +86,19 @@ pub trait ReadClient: Send + Sync {
...
+    fn previous_committee(&self) -> impl Future<Output = SuiClientResult<Committee>> + Send;
+
+    /// Returns the committee that will become active in the next epoch.
+    ///
+    /// This committee is `None` until known.
+    // INV: upcoming_committee.epoch == current_committee.epoch + 1
+    fn next_committee(&self) -> impl Future<Output = SuiClientResult<Option<Committee>>> + Send;
+
+    /// Returns the current epoch state.
+    fn epoch_state(&self) -> impl Future<Output = SuiClientResult<EpochState>> + Send;
```
jpcsmith, line 110: Remark: This results in 4 separate calls to the chain, could we have an interface that just returns all four at once?

giac-mysten: Yup, adding it, just discussed with @karlwuest offline :)

>This results in 4 separate calls to the chain
there are actually more calls to the chain, because of the translation object id -> storage node data. But yes, with one call to get the committees and status, we save 3 calls.

More importantly, we get an atomic view of the system status, which is nicer and avoids (unlikely) epoch transition effects.

PR #842, opened 2024-09-23
refactor: separate ActiveCommittees for the client

giac-mysten:
- Extracts the `ActiveCommittees` struct from the node and exposes it in `common` for use with the client.
- Adds a function to obtain an instance of `ActiveCommittees` from `CommitteesAndState`.
- The walrus `Client` is updated to use `ActiveCommittees`, but without any change to track committee change just yet.
- Refactors the names of the committees, bringing them all to `previous`, `current`, `next`, to match the contracts.

Contributes to #753

### DIFF
--- a/crates/walrus-service/src/common/active_committees.rs
+++ b/crates/walrus-service/src/common/active_committees.rs
@@ -153,7 +153,7 @@ impl ActiveCommittees {
     pub fn from_committees_and_state(committees_and_state: CommitteesAndState) -> Self {
         Self::new_with_next(
             committees_and_state.current,
-            Some(committees_and_state.previous),
+            committees_and_state.previous,
             committees_and_state.next,
             committees_and_state.epoch_state.is_transitioning(),
         )
@@ -200,11 +200,7 @@ impl ActiveCommittees {
         }
 
         match certified_epoch.cmp(&self.current_committee.epoch) {
-            Ordering::Less => {
-                Some(self.previous_committee.as_ref().expect(
-                    "the committee exists as the current committee does not have epoch zero",
-                ))
-            }
+            Ordering::Less => self.previous_committee.as_ref(),
             Ordering::Equal => Some(&self.current_committee),
             Ordering::Greater => None,
         }
@@ -390,5 +386,11 @@ impl CommitteeTracker {
 
         self.0.check_invariants();
         Ok(previous_committee)
+    }
+}
+
+impl From<ActiveCommittees> for CommitteeTracker {
+    fn from(active_committees: ActiveCommittees) -> Self {
+        Self::new(active_committees)
     }
 }
--- a/crates/walrus-service/src/node/committee/committee_service.rs
+++ b/crates/walrus-service/src/node/committee/committee_service.rs
@@ -122,8 +122,8 @@ where
         S: CommitteeLookupService + std::fmt::Debug + 'static,
     {
         // TODO(jsmith): Allow setting the local service factory.
-        let committee_tracker =
-            CommitteeTracker::new(lookup_service.get_active_committees().await?);
+        let committee_tracker: CommitteeTracker =
+            lookup_service.get_active_committees().await?.into();
         let encoding_config = Arc::new(EncodingConfig::new(
             committee_tracker
                 .committees()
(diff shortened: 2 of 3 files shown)


crates/walrus-service/src/common/active_committees.rs
```diff
@@ -48,7 +51,47 @@ pub(crate) enum InvalidNextCommittee {
     InvalidEpoch { expected: Epoch, actual: Epoch },
 }
 
+/// Helper trait to separate the committee transition logic from the active committee state.
+pub(crate) trait CommitteeTransitions {
```
jpcsmith: Suggestion: Since this is storage node specific, keep it at its original location under the node path.

Suggestion: Instead of an extension trait and splitting the definition from the implementation, I would be in favour of just have a struct which stores an active committees and operates on it. This feels a bit like an abuse of traits.

Nit: `CommitteeTracker` would be a good name as well, and sounds like like the enum representing the possible transitions that a committee can go through.

crates/walrus-service/src/common/active_committees.rs
```diff
@@ -115,6 +158,52 @@ impl ActiveCommittees {
...
+            upcoming_committee: upcoming_committee.map(Arc::new),
+            is_transitioning,
+        };
+        this.check_invariants();
+        this
+    }
+
+    /// Construct a new set of `ActiveCommittees`, fetching the status from Sui.
+    #[tracing::instrument(skip_all)]
+    pub async fn from_sui(read_client: &impl ReadClient) -> Result<Self, SuiClientError> {
```
jpcsmith: Remark: `from_sui` sounds like it's being converted from an object of type `sui`. Perhaps something along the lines of `query_sui` or something of that ilk.

crates/walrus-service/src/common/active_committees.rs
```diff
@@ -115,6 +158,52 @@ impl ActiveCommittees {
...
+        let epoch_state = read_client.epoch_state().await?;
+        let current_committee = read_client.current_committee().await?;
+        let prior_committee = read_client.previous_committee().await?;
+        let upcoming_committee = if let EpochState::NextParamsSelected(_) = epoch_state {
+            // This is the only case where the upcoming committee is known.
+            read_client.next_committee().await?
+        } else {
+            None
+        };
```
jpcsmith: Hint: If keeping as separate calls, this may be a good place for a [`join!`](https://docs.rs/futures/latest/futures/macro.join.html).

giac-mysten: Changing to a single call.

crates/walrus-service/src/common/active_committees.rs
giac-mysten: >Suggestion: Instead of an extension trait and splitting the definition from the implementation, I would be in favour of just have a struct which stores an active committees and operates on it. This feels a bit like an abuse of traits.

question on this: do you mean that you would prefer that all these functions would go into the `impl ActiveCommittees` as before? Happy to make the change.

crates/walrus-service/src/common/active_committees.rs
giac-mysten: Makes sense. However, since i know have a single call on the read client that returns a struct `CommitteesAndState`, I'll just implement `from_committees_and_state`.

crates/walrus-service/src/common/active_committees.rs
jpcsmith: No, I meant
```rust
struct CommitteeTransitions(ActiveCommittees);

impl CommitteeTransitions {

}
```

crates/walrus-service/src/common/active_committees.rs
```diff
@@ -200,11 +200,7 @@ impl ActiveCommittees {
         }
 
         match certified_epoch.cmp(&self.current_committee.epoch) {
-            Ordering::Less => {
-                Some(self.previous_committee.as_ref().expect(
-                    "the committee exists as the current committee does not have epoch zero",
-                ))
-            }
+            Ordering::Less => self.previous_committee.as_ref(),
```
giac-mysten, line 203: @jpcsmith did we need the `expect` here?
if `certified_epoch = 0` and the current `epoch = 1`, we would get an empty committee, but still `Some`.
Does this break any of the invariants?

jpcsmith: Well, the implicit condition that we would in theory be able to request data from that committee, which is impossible for committee zero. But the expect probably isn't necessary.

PR #847, opened 2024-09-24
feat: storage node epoch change state transition

jpcsmith:
This PR implements the main storage node epoch change logic:
 - monitor epoch change start event
 - trigger service committee change to update the new/previous/next committee
 - start any shard sync if neeeded.

contribute to #560

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -759,24 +759,10 @@ impl StorageNode {
                 .storage
                 .create_storage_for_shards(&shard_diff.gained)?;
 
-            if event.epoch == GENESIS_EPOCH {
-                // TODO(jsmith): Create the shard in sync mode and have it immediately progress to
-                // being completed. i.e., merge the two cases.
-                for shard in &shard_diff.gained {
-                    self.inner
-                        .storage
-                        .shard_storage(*shard)
-                        .expect("we just created it")
-                        .set_active_status()?;
-                }
-                self.inner
-                    .contract_service
-                    .epoch_sync_done(self.inner.node_object_id)
-                    .await;
-            } else {
-                for shard in &shard_diff.gained {
-                    self.shard_sync_handler.start_new_shard_sync(*shard).await?;
-                }
+            // There shouldn't be an epoch change event for the genesis epoch.
+            assert!(event.epoch != GENESIS_EPOCH);
+            for shard in &shard_diff.gained {
+                self.shard_sync_handler.start_new_shard_sync(*shard).await?;
             }
         }
 
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -248,7 +248,7 @@ mod tests {
             .storage
             .shard_storage(ShardIndex(0))
             .expect("Failed to get shard storage")
-            .set_active_status()
+            .update_status_in_test(ShardStatus::Active)
             .expect("Failed to update shard status");
 
         assert!(matches!(
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -367,10 +367,6 @@ impl ShardStorage {
             .map(|s| s.unwrap_or(ShardStatus::None))
     }
 
-    pub(crate) fn set_active_status(&self) -> Result<(), TypedStoreError> {
-        self.shard_status.insert(&(), &ShardStatus::Active)
-    }
-
     pub(crate) fn set_start_sync_status(&self) -> Result<(), TypedStoreError> {
         self.shard_status.insert(&(), &ShardStatus::ActiveSync)
     }
@@ -550,6 +546,10 @@ impl ShardStorage {
             .certified_blob_info_iter_before_epoch(epoch, last_synced_blob_id);
 
         let mut next_blob_info = blob_info_iter.next().transpose()?;
+        // For transitioning from GENESIS epoch to epoch 1, since GENESIS epoch does not have
+        // any committees and should not receive any user blobs, there shouldn't be any certified
+        // blobs. In case, the shard sync should finish immediately and transition to Active state.
+        assert!(epoch != 0 && (epoch > 1 || next_blob_info.is_none()));
         while let Some((next_starting_blob_id, _)) = next_blob_info {
             tracing::debug!(
                 "syncing shard to before epoch: {}. Starting blob id: {}",


crates/walrus-service/src/common/active_committees.rs
```diff
@@ -189,8 +189,9 @@ impl ActiveCommittees {
     }
 
     fn check_invariants(&self) {
+        // Both epoch 0 and epoch 1 do not have previous committee.
         assert!(
-            self.current_committee.epoch == 0 || self.previous_committee.is_some(),
+            self.current_committee.epoch <= 1 || self.previous_committee.is_some(),
```
halfprice: Note: in tests, we don't set prior committee for epoch 1.

crates/walrus-service/src/node.rs
```diff
@@ -378,31 +385,24 @@ impl StorageNode {
...
-        if managed_shards.is_empty() {
-            tracing::info!(epoch = committee.epoch, "node does not manage any shards");
-        }
-
-        for shard in managed_shards {
-            storage
-                .create_storage_for_shard(*shard)
-                .with_context(|| format!("unable to initialize storage for shard {}", shard))?;
-        }
+        // TODO(jsmith): This should be fetched from the wallet.
```
halfprice: This is a bit tricky since we don't know which object is the node object.

crates/walrus-service/src/node.rs
```diff
@@ -441,6 +442,26 @@ impl StorageNode {
...
+            "storage node is lagging by more than 1 epoch, recovery not yet supported"
+        );
+
+        self.start_processing_events(cancel_token).await
+    }
+
+    /// Run the walrus-node logic until cancelled using the provided cancellation token. Skip
+    /// the start-up check for lagging.
+    #[cfg(any(test, feature = "test-utils"))]
+    pub async fn run_skip_start_up_check(
```
halfprice: Many tests has following pattern:

- BuildClusterAndRun
- SendEvent.

Since running the node checks lagging, which requires reading the next event, the above situation basically blocks forever. So create this to work around this issue, but @jpcsmith you may have better way to do it.

crates/walrus-service/src/node.rs
```diff
@@ -663,6 +731,149 @@ impl StorageNode {
...
+            Err(BeginCommitteeChangeError::ChangeAlreadyInProgress) => {
+                // We are likely processing a backlog of events. Since the committee service has a
+                // more recent committee or has already had the current committee marked as
+                // transitioning, our shards have also already been configured for the more
+                // recent committee and there is actual nothing to do.
+                tracing::debug!(
+                    walrus.epoch = epoch,
+                    "skipping epoch change start event for an older epoch"
+                );
+                return Ok(true);
```
halfprice: This case is motivated by test, but may also be useful in prod.

In tests, we move the contract to epoch 1 before starting the node, which means that when receiving epoch start event, the node already has the new committee, but we still want to execute the epoch change logic.

crates/walrus-service/src/node/committee.rs
```diff
@@ -54,27 +54,23 @@ pub(crate) trait CommitteeLookupService: Send + Sync + std::fmt::Debug {
...
             )
-            .unwrap();
-            Ok(ActiveCommittees::new(
-                committee,
-                Some(fake_previous_committee),
-            ))
-        }
+        };
+
+        Ok(ActiveCommittees::new(committee, previous_committee))
```
halfprice: @jpcsmith PTAL this change

crates/walrus-service/src/node.rs
jpcsmith: I briefly discussed this with Karl, and there should be a capability object in the storage node's wallet from which we can get this ID. If there are multiple, we will need them to set the value in the config.

```suggestion
        // TODO(jsmith): This should be fetched from the wallet (#862).
```

crates/walrus-service/src/common/active_committees.rs
jpcsmith: Remark: Epoch 1 does indeed have a prior committee, it's the empty committee of epoch 0.

crates/walrus-service/src/node.rs
jpcsmith: Fair point, since this seems like it's localised to when we the event is "begin committee change X" and we're already in X, I added a case for this specifically.

crates/walrus-service/src/node/committee.rs
jpcsmith: Refactored to using the `get_committees_and_state()` method of the `ReadClient`.

crates/walrus-service/src/common/active_committees.rs
halfprice: Sure. It's just that we are not setting it in the test: https://github.com/MystenLabs/walrus/blob/afe7bb5fa68f35c8930135c6e1d59bc0c490d875/crates/walrus-service/src/test_utils.rs#L577

crates/walrus-service/src/node.rs
halfprice: Right now, in the epoch change done, we are fetching the capability object, and use the node id to identify which capability object to to use... Kind of a chicken and egg problem...

I also discussed this with @karlwuest last week and was also suggested to fetch the capability object. But there could be multiple (unless we enforce it in the contract??). So maybe putting it in the config is easier/cleaner.

crates/walrus-service/src/common/active_committees.rs
jpcsmith: Yea, but in this case, fixing the code in test utils as opposed to changing the invariant is the preferred approach.

crates/walrus-service/src/node.rs
jpcsmith: I moved the check to the event loop, and execute it only once. That way we can avoid adding code to skip the check.

crates/walrus-service/src/node.rs
```diff
@@ -663,6 +720,143 @@ impl StorageNode {
...
+        } else {
+            self.inner
+                .storage
+                .create_storage_for_shards(&shard_diff.gained)?;
+
+            // There shouldn't be an epoch change event for the genesis epoch.
+            assert!(event.epoch != GENESIS_EPOCH);
+            for shard in &shard_diff.gained {
+                self.shard_sync_handler.start_new_shard_sync(*shard).await?;
+            }
```
halfprice, line 766: I made one final tweak. Before, we were specializing event.epoch == GENESIS_EPOCH, and don't start sync. In fact, we should really specialize event.epoch == `GENESIS_EPOCH + 1`. My curiosity of how could all tests passing made me dig a little bit further, and I realized that when doing epoch 0 -> 1, since there won't be any certified blobs in epoch 0, the sync will just finish immediately. So here, we can actually just call start sync directly and greatly simplify the code.

PR #850, opened 2024-09-24
fix: don't interpret network address in walrus-sui

### LINKED ISSUE #785: Add default port to `FromStr` on `NetworkAddress`
Currently, the implementation expects both a host and port to be in the string.

karlwuest:
closes #785

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -256,23 +256,23 @@ impl ClientBuilder {
 
     /// Convenience function to build the client where the server is identified by a [`SocketAddr`].
     ///
-    /// Equivalent `self.build(&remote.ip().to_string(), remote.port())`
+    /// Equivalent `self.build(&remote.to_string())`
     pub fn build_for_remote_ip(self, remote: SocketAddr) -> Result<Client, ClientBuildError> {
-        self.build(&remote.ip().to_string(), remote.port())
+        self.build(&remote.to_string())
     }
 
     /// Consume the `ClientBuilder` and return a configured [`Client`].
     ///
-    /// This method fails if a valid URL cannot be created with the provided host and port, the
+    /// This method fails if a valid URL cannot be created with the provided address, the
     /// Rustls TLS backend cannot be initialized, or the resolver cannot load the system
     /// configuration.
-    pub fn build(mut self, host: &str, port: u16) -> Result<Client, ClientBuildError> {
+    pub fn build(mut self, address: &str) -> Result<Client, ClientBuildError> {
         #[cfg(msim)]
         {
             self = self.no_proxy();
         }
 
-        let url = Url::parse(&format!("https://{host}:{port}"))
+        let url = Url::parse(&format!("https://{address}"))
             .map_err(|_| BuildErrorKind::InvalidHostOrPort)?;
         // We extract the host from the URL, since the provided host string may have details like a
         // username or password.
@@ -344,13 +344,12 @@ impl Client {
     /// identity and public key.
 
     pub fn for_storage_node(
-        host: &str,
-        port: u16,
+        address: &str,
         public_key: &NetworkPublicKey,
     ) -> Result<Client, ClientBuildError> {
         Self::builder()
             .authenticate_with_public_key(public_key.clone())
-            .build(host, port)
+            .build(address)
     }
 
     /// Converts this to the inner client.
--- a/crates/walrus-sui/src/types.rs
+++ b/crates/walrus-sui/src/types.rs
@@ -6,13 +6,10 @@
 
 use std::{
     fmt::Display,
-    net::{SocketAddr, ToSocketAddrs},
-    num::NonZeroU16,
-    str::FromStr,
-    vec,
+    net::SocketAddr,
+    num::{NonZeroU16, ParseIntError},
 };
 
-use anyhow::bail;
 use sui_types::base_types::ObjectID;
 use thiserror::Error;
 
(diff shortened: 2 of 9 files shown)


crates/walrus-sui/src/types.rs
```diff
@@ -42,45 +42,55 @@ pub use move_structs::{
...
+            Ok(Self {
+                host: host.to_owned(),
+                port,
+            })
+        } else {
+            Ok(Self {
+                host: s.to_owned(),
+                port: DEFAULT_REST_API_PORT,
+            })
+        }
```
jpcsmith: Suggestion: The failed parse should return an error instead of using the default port, as this implies a problem with the info they provided on chain..

karlwuest: If we return an error in this function, the deserialization of the `StorageNodeInfo` will fail, which will currently (probably) crash the storage nodes, since we won't be able to construct the committee. Since the input is untrusted, we want this to succeed even if the info is malformed. Because of this, I log the error, but then return the default. I'm open to other suggestions, if you have any.

mlegner: I agree that we should not crash based on input provided (indirectly) by other storage nodes. I think this was the cause of a Sui testnet outage (a storage node set a UDP address instead of TCP).

Could we instead enforce in the contract that all the addresses are well-formed?

karlwuest: We could but it's more cumbersome. The easiest way would be to require storage nodes to always submit a port (submitting an address as string and then doing validation is not something that move is built to do well). 
However, even then the "well-formedness" is quite limited and we don't know whether we'll be able to connect to the address, e.g. the hostname could be invalid, so we don't really lose anything by just assigning the default port.

jpcsmith: I agree that we should not crash, but my inclination is to handle the error in the caller. If the port is invalid, it's likely the host is invalid as well, and this is being propagated and used as if it was a valid host and port.

If the case is that we always need to successfully parse it into something, then one approach would be to have an enum that either provides a valid parsed or invalid, unparsed network address. The consumer can then react appropriately: things like output and logging would not care, entities attempting to connect would log and not bother connecting since it was invalid.

In any case, I'm not against this merging in, I just think it will probably cause some weird bugs if it's every actually triggered.

karlwuest: >I agree that we should not crash, but my inclination is to handle the error in the caller. If the port is invalid, it's likely the host is invalid as well, and this is being propagated and used as if it was a valid host and port.

This is the same as if we have an invalid host name, though, and we don't know that until we resolve it and connect to it. We are also not doing any other validation checks on the hostname at this point. It could, e.g., contain invalid characters.

>If the case is that we always need to successfully parse it into something, then one approach would be to have an enum that either provides a valid parsed or invalid, unparsed network address. The consumer can then react appropriately: things like output and logging would not care, entities attempting to connect would log and not bother connecting since it was invalid.

I don't think returning an error here is an option, because that would prevent us from getting the committee from chain (or we would need to return an incomplete committee that excludes the node with the malformed address, which would create many more issues). Having an enum for this (or just an `Option` for the port) would be a possibility, though. However, then I think the better option would be to just require the nodes to register host and port separately in the contract. I think this will require a lot less refactoring.

Another possibility that would remove the illusion of having any validation on the address, would be to remove the split into host and port and simply keep it as a string. Then we would do all handling of the address in `walrus-service`.

This is not something that needs to be merged today, so we can discuss this briefly in person tomorrow.

jpcsmith: > Another possibility that would remove the illusion of having any validation on the address, would be to remove the split into host and port and simply keep it as a string. Then we would do all handling of the address in walrus-service.

That would be a good option, I believe the walrus-sdk is, in any case, the only consumer of it.

crates/walrus-sdk/src/client.rs
```diff
@@ -256,23 +256,24 @@ impl ClientBuilder {
...
     /// configuration.
-    pub fn build(mut self, host: &str, port: u16) -> Result<Client, ClientBuildError> {
+    pub fn build(mut self, address: &str) -> Result<Client, ClientBuildError> {
         #[cfg(msim)]
         {
             self = self.no_proxy();
         }
 
-        let url = Url::parse(&format!("https://{host}:{port}"))
+        // TODO(karl): add default port if none is given?
```
karlwuest: @jpcsmith @mlegner I'm not sure what the best practices are here. Should we use the default Walrus REST API port here if none is given or fall back to the default https port (and only use our default as default when registering a storage node)?

crates/walrus-sui/src/types.rs
karlwuest: Ok, I updated the PR in this direction. I kept the network address as a newtype, mainly to make it more convenient in the testbed setup. There is one open question regarding the default port, namely if we should use the default when connecting to a host without a provided port (or use the default https).

crates/walrus-sdk/src/client.rs
jpcsmith: I would require storage nodes to be explicit in their registration of the port for now. This is used as an https endpoint. A hostname without a port in that case is 443.

karlwuest: Perfect, that was my inclination as well.

crates/walrus-sui/src/types.rs
```diff
@@ -44,47 +41,39 @@ use walrus_core::{bft, ensure, Epoch, NetworkPublicKey, PublicKey, ShardIndex};
...
-            bail!("invalid network address")
-        };
-        let port = port.parse()?;
-        Ok(Self {
-            host: host.to_owned(),
-            port,
-        })
+impl Display for NetworkAddress {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.0)
```
jpcsmith:
```suggestion
        self.0.fmt(f)
```

crates/walrus-sui/src/types.rs
```diff
@@ -44,47 +41,39 @@ use walrus_core::{bft, ensure, Epoch, NetworkPublicKey, PublicKey, ShardIndex};
...
 }
 
-impl ToSocketAddrs for NetworkAddress {
-    type Iter = vec::IntoIter<SocketAddr>;
-
-    fn to_socket_addrs(&self) -> std::io::Result<Self::Iter> {
-        (self.host.as_str(), self.port).to_socket_addrs()
+impl From<SocketAddr> for NetworkAddress {
+    fn from(value: SocketAddr) -> Self {
+        Self(format!("{}:{}", value.ip(), value.port()))
```
jpcsmith: Suggestion: use `to_string()`, as this does not handle IPv6 addresses, which would be formatted using square brackets such as `[f800::0]:443`
```suggestion
        Self(value.to_string())
```

PR #852, opened 2024-09-25
feat(node): expose the initial certified epoch in blob status

giac-mysten:
Adds the initial certified epoch to the `BlobStatus` for permanent and deletable blobs. This change is necessary to establish from which committee to read deletable blobs during epoch change.

Contributes to: #753

### DIFF
--- a/crates/walrus-sdk/src/api.rs
+++ b/crates/walrus-sdk/src/api.rs
@@ -64,7 +64,6 @@ pub enum BlobStatus {
         /// Counts of deletable `Blob` objects.
         deletable_counts: DeletableCounts,
         /// If the blob is certified, contains the epoch where it was initially certified.
-        // INV: certified_epoch.is_some() == is_certified
         initial_certified_epoch: Option<Epoch>,
     },
     /// The blob exists within Walrus; but there is no related permanent object, so it may be


crates/walrus-sdk/src/api.rs
```diff
@@ -78,6 +78,9 @@ pub struct DeletableStatus {
     pub count_deletable_total: u32,
     /// Number of certified deletable `Blob` objects for the given blob ID.
     pub count_deletable_certified: u32,
+    /// If the blob is certified, contains the epoch where it was initially certified.
+    // INV: certified_epoch.is_some() == count_deletable_certified > 0
+    pub initial_certified_epoch: Option<Epoch>,
```
mlegner: Important: The `initial_certified_epoch` refers to both permanent and deletable blobs, so the added invariant is not fully correct. Also, if we want to add this to the `DeletableStatus`, it would be good to rename that struct.

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -154,11 +154,19 @@ impl CliOutput for BlobStatusOutput {
...
                 count_deletable_total,
                 count_deletable_certified,
-            }) => println!(
+                initial_certified_epoch,
+            }) => {
+                let initial_certified_str = if let Some(epoch) = initial_certified_epoch {
+                    format!(", initially certified in epoch {}", epoch)
+                } else {
+                    "".to_string()
+                };
```
mlegner, line 166: Suggestion: Can we add something similar to the `Permanent` case below?

crates/walrus-sdk/src/api.rs
giac-mysten: ah definitely, I misinterpreted the meaning of `DeletableStatus`. Changed this part, and added the `initial_certified_epoch` to the enum struct. I also added it to the `Permanent` version, since we will need it there as well.

Question: should the `initial_certified_epoch` be considered in the sorting order of `BlobStatus`? If so, I assume we want to have the _smaller_ initial certified epoch be preferred (so `Ordering::Greater`), right?

since `is_certified` is anyways the first sorting criterion, we can use the natural sorting of `None < Some(a) < Some(b)` with `a < b`

mlegner: > Question: should the `initial_certified_epoch` be considered in the sorting order of `BlobStatus`? If so, I assume we want to have the _smaller_ initial certified epoch be preferred (so `Ordering::Greater`), right?

Yes, that makes sense. However, I've been discussing with @halfprice whether we can simply remove the ordering as this is becoming a bit complicated.

crates/walrus-sdk/src/api.rs
```diff
@@ -62,31 +62,40 @@ pub enum BlobStatus {
         /// The ID of the Sui event that caused the status with the given `end_epoch`.
         status_event: EventID,
         /// Counts of deletable `Blob` objects.
-        deletable_status: DeletableStatus,
+        deletable_status: DeletableCounts,
+        /// If the blob is certified, contains the epoch where it was initially certified.
+        // INV: certified_epoch.is_some() == is_certified
```
mlegner: Nit: This is actually not true, there could also be a certified deletable blob... 😅

crates/walrus-sdk/src/api.rs
```diff
@@ -62,31 +62,40 @@ pub enum BlobStatus {
...
     },
     /// The blob exists within Walrus; but there is no related permanent object, so it may be
     /// deleted at any time.
-    Deletable(DeletableStatus),
+    Deletable {
+        /// If the blob is certified, contains the epoch where it was initially certified.
+        // INV: certified_epoch.is_some() == count_deletable_certified > 0
+        initial_certified_epoch: Option<Epoch>,
+        /// Counts of deletable `Blob` objects.
+        deletable_status: DeletableCounts,
```
mlegner: Nit: Can we call that `deletable_counts` to be consistent with the struct name?
```suggestion
        deletable_counts: DeletableCounts,
```

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -151,38 +151,54 @@ impl CliOutput for BlobStatusOutput {
...
                 println!(
                     "There is a {status} permanent Blob object for blob ID {blob_str}.\n\
-                        End epoch: {}\n\
-                        Related event: {}",
-                    end_epoch,
-                    format_event_id(&status_event),
+                        End epoch: {end_epoch}\n\
+                        Related event: {}\n\
```
mlegner: Nit: Should the `\n` be part of the `initial_certified_str`?

crates/walrus-sdk/src/api.rs
giac-mysten: but then would it be returned as `Permanent`?

mlegner: Yes, at least currently, although we could debate whether a registered permanent blob is better or worse than a certified deletable one and/or whether the current status struct is the best way to represent the status.

giac-mysten: Ok, I removed the invariant and opened a tracking issue (#864)

PR #855, opened 2024-09-25
refactor: cleanup the error handling for the committee tracker

jpcsmith:
This removes some of the changes from #847 into its own PR.

### DIFF
--- a/crates/walrus-service/src/node/committee/committee_service.rs
+++ b/crates/walrus-service/src/node/committee/committee_service.rs
@@ -251,7 +251,7 @@ where
         let mut modify_result = Ok(());
         let modify_tracker = |tracker: &mut CommitteeTracker| {
             // Guaranteed by the caller.
-            debug_assert_eq!(tracker.next_epoch(), next_committee.epoch);
+            assert_eq!(tracker.next_epoch(), next_committee.epoch);
             if let Err(NextCommitteeAlreadySet(next_committee)) =
                 tracker.set_committee_for_next_epoch(next_committee)
             {


crates/walrus-service/src/common/active_committees.rs
```diff
@@ -284,71 +261,52 @@ impl CommitteeTracker {
...
-    ) -> Result<(), InvalidNextCommittee> {
-        ensure!(
-            self.0.next_committee.is_none(),
-            InvalidNextCommittee::AlreadySet
+    ) -> Result<(), NextCommitteeAlreadySet> {
+        assert_eq!(
+            committee.epoch,
+            self.next_epoch(),
+            "committee's epoch must match the next epoch"
         );
```
halfprice, line 285: Is this invariant true even if a node is lagging behind for several epochs?

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -236,141 +225,105 @@ where
...
             .inner
             .services
             .lock()
             .expect("thread did not panic with mutex");
 
-        let mut result = Ok(());
+        let mut modify_result = Ok(());
+        let modify_tracker = |tracker: &mut CommitteeTracker| {
+            // Guaranteed by the caller.
+            debug_assert_eq!(tracker.next_epoch(), next_committee.epoch);
```
halfprice: nit: since this cheap, may as well make it an assert.

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -236,141 +225,105 @@ where
...
+            }
+        );
+        ensure!(
+            epoch >= current_epoch,
+            EndCommitteeChangeError::ProvidedEpochIsInThePast {
+                provided: epoch,
+                expected: current_epoch,
+            }
+        );
+        debug_assert_eq!(epoch, current_epoch);
```
halfprice, line 307: nit: this is redundant?

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -565,6 +522,50 @@ where
...
+            }
+        );
+        ensure!(
+            new_epoch >= expected_next_epoch,
+            BeginCommitteeChangeError::EpochIsNotGreater {
+                expected: expected_next_epoch,
+                actual: new_epoch,
+            }
+        );
+        debug_assert_eq!(new_epoch, expected_next_epoch);
```
halfprice, line 546: nit: seems not needed.

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -236,141 +225,105 @@ where
...
+            }
+
+            modify_result = tracker.start_change().map_err(|error| match error {
+                StartChangeError::UnknownNextCommittee => unreachable!("committee set above"),
+                StartChangeError::ChangeInProgress => {
+                    BeginCommitteeChangeError::ChangeAlreadyInProgress
+                }
+            });
+
+            modify_result.is_ok()
```
halfprice, line 275: Note/question: trying to understand if following is expected: `set_committee_for_next_epoch` succeeded and tracker is modified, but `start_change` returns error and the watch listener is not notified.

crates/walrus-service/src/common/active_committees.rs
jpcsmith: Yes, this invariant is always true.

crates/walrus-service/src/node/committee/committee_service.rs
jpcsmith:
```suggestion
            assert_eq!(tracker.next_epoch(), next_committee.epoch);
```

crates/walrus-service/src/node/committee/committee_service.rs
jpcsmith: They are, but I add them sometimes for sanity to state that that condition must be true after this point. If the code above gets changed but the assertion doesn't, then this will catch the fact that the code below was relying on the condition.

crates/walrus-service/src/node/committee/committee_service.rs
jpcsmith: Same as above.

crates/walrus-service/src/node/committee/committee_service.rs
jpcsmith: It's not possible (a change cannot be in progress if the next committee was not already set), and setting the next committee actually does not require notifying the listeners, so it would not be an issue.

PR #857, opened 2024-09-25
feat: add diff for shard assignments

jpcsmith:
Slight modification of the code you wrote in #847 and added to a separate PR to further simplify #847.

### DIFF
--- a/crates/walrus-service/src/common/utils.rs
+++ b/crates/walrus-service/src/common/utils.rs
@@ -4,6 +4,7 @@
 //! Utility functions for the Walrus service.
 
 use std::{
+    collections::HashSet,
     fmt::Debug,
     future::Future,
     net::{IpAddr, Ipv4Addr, SocketAddr},
@@ -59,6 +60,9 @@ macro_rules! version {
     }};
 }
 pub use version;
+use walrus_core::{PublicKey, ShardIndex};
+
+use super::active_committees::ActiveCommittees;
 
 /// Trait for loading configuration from a YAML file.
 pub trait LoadConfig: DeserializeOwned {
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/common/utils.rs
```diff
@@ -455,8 +459,48 @@ impl MetricsAndLoggingRuntime {
     }
 }
 
+/// The difference between shard allocations in different epochs.
+#[derive(Debug, Clone, Eq, PartialEq, Default)]
+pub(crate) struct ShardDiff {
+    /// Shards lost from the assignment.
+    pub lost: Vec<ShardIndex>,
+    /// Shards gained in the assignment.
+    pub gained: Vec<ShardIndex>,
```
halfprice, line 468: There are two cases for gained.

One is when a shard is newly created. This only applies to epoch 0 -> 1. In this case, we don't need to initiate any shard sync, and shard can directly start at Active state.

Another case is when shard moving is needed. This applies to all other epoch changes.

halfprice: We can also just check in `process_epoch_change_begin` whether we are transitioning from epoch 0 to 1 (and leave this PR as is). So your call.

jpcsmith: > One is when a shard is newly created. This only applies to epoch 0 -> 1. In this case, we don't need to initiate any shard sync, and shard can directly start at Active state.

Shouldn't that just result in a a shard sync which ends immediately since there is nothing to sync, and we quickly move to the active state? Do we really need to special case it?

halfprice: [Not right now](https://github.com/MystenLabs/walrus/blob/a61b8ee0a0e8ee8ac998af3114eb6773827d9fbd/crates/walrus-service/src/node/committee/committee_service.rs#L205).

So we need to specialize it either in the process event function, or in the sync handler. Even in the handler itself, it needs to specialize epoch 0->1 vs. other epoch changes, since we don't really expect to not find the prior owner of the shard.

PR #863, opened 2024-09-26
fix(deploy): add server name and improve output

mlegner:


crates/walrus-service/src/testbed.rs
```diff
@@ -476,7 +489,12 @@ pub async fn create_storage_node_configs(
             metrics_address,
             rest_api_address,
             sui,
-            ..test_utils::storage_node_config().inner
+            db_config: None,
+            rest_graceful_shutdown_period_secs: None,
+            blob_recovery: Default::default(),
+            tls,
+            shard_sync_config: Default::default(),
+            event_processor_config: None,
```
jpcsmith, line 497: Question: Why not just populate with the defaults? Or are these actually different from the defaults?
```suggestion
            tls,
            ..test_utils::storage_node_config().inner
```

crates/walrus-service/src/testbed.rs
```diff
@@ -563,14 +581,7 @@ async fn create_storage_node_wallets(
         })
         .collect::<Result<Vec<_>, _>>()?;
 
-    if faucet_cooldown.is_some() {
```
jpcsmith, line 566: Question: Is the check for the faucet cooldown no longer necessary?

mlegner: I thought it doesn't hurt to print the wallets in any case.

crates/walrus-service/src/testbed.rs
mlegner: Two reasons:
1. I'm not sure using something from `test_utils` makes sense outside of tests.
2. Doing this feels dangerous as any additions to the config struct may be hidden in this default. I guess this is how the TLS config got forgotten.

jpcsmith: Well... it is the **test**bed 😏, but okay.

mlegner: Haha. Given that we use this for public deployments, we could maybe rename it... 😅

PR #866, opened 2024-09-26
feat(client): track committee changes

giac-mysten:
The client now uses the correct committee when reading and storing blobs.

Main changes:

- Refactored the `NodeCommunicationFactory` into its own module; it now takes a `certified_epoch` parameter when creating the vector of `NodeCommunication` objects to read, so that the correct committee for reading is selected.
- The Walrus `Client` now owns a (Sui) `ReadClient` for reads. This is necessary as, before any read, we need to know when the blob was certified. This is taken from the blob status, which may be read from chain.
- This latter change also required making `ReadClient` a supertrait of `ContractClient`. Otherwise, the "store" client would not be able to read as well.
- Added retries to `get_verified_blob_status`.

Contributes to #753

### DIFF
--- a/crates/walrus-service/src/common/active_committees.rs
+++ b/crates/walrus-service/src/common/active_committees.rs
@@ -239,7 +239,7 @@ impl ActiveCommittees {
     /// Returns the minimum number of correct shards.
     ///
     /// Given the invariants enforced by this struct, the result of this function is the same for
-    /// all members of the committee.
+    /// all committees.
     ///
     /// This is (`n_shards - f`), where `f` is the maximum number of faulty shards, given
     /// `n_shards`. See [walrus_core::bft] for further details.
@@ -252,7 +252,7 @@ impl ActiveCommittees {
     /// Checks if the number is large enough to reach a quorum (`2f + 1`).
     ///
     /// Given the invariants enforced by this struct, the result of this function is the same for
-    /// all members of the committee.
+    /// all committees.
     ///
     /// `f` is the maximum number of faulty shards, given `n_shards`. See [walrus_core::bft] for
     /// further details.
@@ -265,7 +265,7 @@ impl ActiveCommittees {
     ///
     ///
     /// Given the invariants enforced by this struct, the result of this function is the same for
-    /// all members of the committee.
+    /// all committees.
     ///
     /// The validity threshold is `f + 1`, where `f` is the maximum number of faulty shards. See
     /// [walrus_core::bft] for further details.
@@ -281,12 +281,12 @@ impl ActiveCommittees {
     /// the pairs of the members in previous, current, and next, committees.
     #[allow(clippy::mutable_key_type)]
     pub fn unique_node_address_and_key(&self) -> HashSet<(&NetworkAddress, &NetworkPublicKey)> {
-        let mut members = HashSet::from_iter(extract_address_and_pk(&self.current_committee));
+        let mut members = HashSet::from_iter(self.current_committee.network_addresses_and_pks());
         if let Some(previous) = self.previous_committee() {
-            members.extend(extract_address_and_pk(previous));
+            members.extend(previous.network_addresses_and_pks());
         }
         if let Some(next) = self.next_committee() {
-            members.extend(extract_address_and_pk(next));
+            members.extend(next.network_addresses_and_pks());
         }
         members
     }
@@ -418,13 +418,4 @@ impl From<ActiveCommittees> for CommitteeTracker {
     fn from(active_committees: ActiveCommittees) -> Self {
         Self::new(active_committees)
     }
-}
-
-fn extract_address_and_pk(
-    committee: &Committee,
-) -> impl Iterator<Item = (&NetworkAddress, &NetworkPublicKey)> {
-    committee
-        .members()
-        .iter()
-        .map(|member| (&member.network_address, &member.network_public_key))
 }
--- a/crates/walrus-sui/src/types.rs
+++ b/crates/walrus-sui/src/types.rs
@@ -267,4 +267,13 @@ impl Committee {
         self.member_index_for_shard(shard.into())
             .map(|index| &self.members[index])
     }
+
+    /// Returns the network address and network public key of the committee members.
+    pub fn network_addresses_and_pks(
+        &self,
+    ) -> impl Iterator<Item = (&NetworkAddress, &NetworkPublicKey)> {
+        self.members
+            .iter()
+            .map(|member| (&member.network_address, &member.network_public_key))
+    }
 }


crates/walrus-service/src/client/communication/mod.rs
mlegner, line 1: Nit: Can we name this `crates/walrus-service/src/client/communication.rs` to be consistent with our existing convention?

crates/walrus-service/src/common/active_committees.rs
```diff
@@ -181,13 +184,81 @@ impl ActiveCommittees {
...
     ///
     /// Given the invariants enforced by this struct, `n_shards` is the same for all committees.
     pub fn n_shards(&self) -> NonZeroU16 {
         self.current_committee.n_shards()
     }
 
+    /// Checks if the number is larger or equal to the minimum number of correct shards.
+    ///
+    /// Given the invariants enforced by this struct, the result of this function is the same for
+    /// all members of the committee.
```
mlegner: Question: What does "all members of the committee" refer to here and below? Maybe better "all committees"?

crates/walrus-service/src/common/active_committees.rs
```diff
@@ -181,13 +184,81 @@ impl ActiveCommittees {
...
+    // storage nodes once created. Is there a better way?
+    #[allow(clippy::mutable_key_type)]
+    pub fn all_members(&self) -> HashSet<&StorageNode> {
+        let mut members = HashSet::from_iter(self.current_committee.members().iter());
+        if let Some(previous) = self.previous_committee() {
+            members.extend(previous.members().iter());
+        }
+        if let Some(next) = self.next_committee() {
+            members.extend(next.members().iter());
+        }
```
mlegner: Nit: Maybe slightly more elegant?
```suggestion
        self.previous_committee()
            .map(|c| members.extend(c.members().iter()));
        self.next_committee()
            .map(|c| members.extend(c.members().iter()));
```

crates/walrus-service/src/common/active_committees.rs
```diff
@@ -181,13 +184,81 @@ impl ActiveCommittees {
...
+    /// [walrus_core::bft] for further details.
+    #[inline]
+    #[allow(dead_code)]
+    pub fn is_above_validity(&self, num: usize) -> bool {
+        self.current_committee.is_above_validity(num)
+    }
+
+    /// Returns the union of the members in previous, current, and next, committees.
+    // TODO(giac): having the mutable key type here should be fine, since we do not modify the
+    // storage nodes once created. Is there a better way?
```
mlegner: Hmm... The main issue I see is the `shard_ids`: This means that a storage node can occur in the `all_members` result three times with different vectors of shard IDs. Is that intentional?

Maybe we could have a custom `Hash` implementation for `StorageNode` that just takes its public key or some other unique identifier?

crates/walrus-service/src/client.rs
```diff
@@ -30,38 +25,36 @@ use walrus_core::{
...
 
 pub mod cli;
+mod communication;
 pub mod responses;
 
 mod blocklist;
 pub use blocklist::Blocklist;
 
-mod communication;
-
```
mlegner: Nit: I think the ordering was intentional before: First the public modules, then all the private ones.

crates/walrus-service/src/client.rs
```diff
@@ -138,22 +130,26 @@ impl Client<()> {
...
+            committees: committees.clone(),
             storage_price_per_unit_size,
-            encoding_config,
+            encoding_config: encoding_config.clone(),
             communication_limits,
             blocklist: None,
             communication_factory: NodeCommunicationFactory::new(
                 config.communication_config.clone(),
+                committees,
+                encoding_config,
```
mlegner, line 146: Suggestion: As we have the `committees` now in two places and they are not actually tiny, how about we wrap them in an `Arc`?

crates/walrus-service/src/client.rs
```diff
@@ -763,28 +791,88 @@ impl<T> Client<T> {
...
+
+            // Return only if we know that the blob does not exist, or if we have a valid status.
+            if let Err(client_error) = maybe_status {
+                if matches!(client_error.kind(), &ClientErrorKind::BlobIdDoesNotExist) {
+                    return Err(client_error);
+                }
+            } else {
+                // The status is Ok.
+                return maybe_status;
+            }
```
mlegner: Nit: I feel a match would be clearer here:
```suggestion
            match maybe_status {
                Ok(_) => {
                    return maybe_status;
                }
                Err(client_error)
                    if matches!(client_error.kind(), &ClientErrorKind::BlobIdDoesNotExist) =>
                {
                    return Err(client_error)
                }
                Err(_) => (),
            }
```

crates/walrus-service/src/client.rs
```diff
@@ -820,124 +908,13 @@ impl<T> Client<T> {
...
-        .expect("the threshold is below the total number of shards")
-    }
-
     /// Maps the sliver pairs to the node that holds their shard.
     fn pairs_per_node<'a>(
         &'a self,
         blob_id: &'a BlobId,
         pairs: &'a [SliverPair],
     ) -> HashMap<usize, impl Iterator<Item = &SliverPair>> {
+        // TODO!
```
mlegner: Question: What is this about?

crates/walrus-service/src/common/active_committees.rs
giac-mysten: Although I agree, clippy complains, so i'm back to the `if let`.

crates/walrus-service/src/common/active_committees.rs
giac-mysten: good catch! The comment here was referring to the fact that `clippy` complains if a hashset contains [mutable keys.](https://rust-lang.github.io/rust-clippy/master/index.html#/mutable_key_type) 

However, I am starting to think that this is anyways an ill-defined function, as you point out. 

Changing it to return just the unique `(address, key)` pairs, which is anyways what we want here.

crates/walrus-service/src/client.rs
giac-mysten: Forgot to update the committee here 😬

crates/walrus-service/src/common/active_committees.rs
```diff
@@ -410,3 +411,12 @@ impl From<ActiveCommittees> for CommitteeTracker {
...
 }
+
+fn extract_address_and_pk(
+    committee: &Committee,
+) -> impl Iterator<Item = (&NetworkAddress, &NetworkPublicKey)> {
+    committee
+        .members()
+        .iter()
+        .map(|member| (&member.network_address, &member.network_public_key))
+}
```
mlegner: Nit: Can we make this a method on `Committee`?

crates/walrus-service/src/common/active_committees.rs
mlegner: Should also be changed below.

crates/walrus-service/src/client.rs
```diff
@@ -279,7 +306,7 @@ impl<T: ContractClient> Client<T> {
             .map_err(|e| ClientError::from(ClientErrorKind::CertificationFailed(e)))?;
 
         // TODO(mlegner): Make sure this works if the epoch changes. (#753)
-        blob.certified_epoch = Some(self.committees.current_committee().epoch);
+        blob.certified_epoch = Some(self.committees.write_committee().epoch);
```
mlegner, line 309: Question: Does this PR resolve the TODO?

giac-mysten: Unfortunately not yet: this does ensure that the right committee is picked during epoch change. However, if the start of the epoch change occurs e.g., after the slivers have been stored but before the certificate is collected, this is a different committee and it may not have enough shards in common with the old to provide a certificate. Therefore, the client may not be able to get a certificate here during epoch change. The retry is not implemented yet.

crates/walrus-service/src/client.rs
```diff
@@ -763,28 +792,91 @@ impl<T> Client<T> {
...
                 self.communication_limits.max_concurrent_status_reads,
             )
             .await;
 
+        // If 2f+1 nodes return a 404 status, we know the blob does not exist.
+        let n_not_found = requests
+            .inner_err()
+            .iter()
+            .filter(|err| err.is_status_not_found())
+            .count();
```
mlegner, line 874: @giac-mysten I know this is a very old PR, but I just stumbled over this code here: Shouldn't we sum up the *weight* and check if that is a quorum instead of counting the *nodes*?

giac-mysten: I think you are right :( fixing.

PR #882, opened 2024-09-27
fix: disable native certs and proxy in integration tests by default

jpcsmith:
Disables proxy and root certs in tests

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -231,7 +231,6 @@ impl ClientBuilder {
     /// Clears proxy settings in the client, and disables fetching proxy settings from the OS.
     ///
     /// On some systems, this can speed up the construction of the client.
-
     pub fn no_proxy(mut self) -> Self {
         self.inner = self.inner.no_proxy();
         self


crates/walrus-service/src/node/committee/node_service.rs
```diff
@@ -273,18 +272,54 @@ impl Service<Request> for RemoteStorageNode {
...
-    .map(|client| RemoteStorageNode {
-        client,
-        encoding_config: encoding_config.clone(),
-    });
+#[derive(Debug, Clone, PartialEq, Eq, Default)]
+pub(crate) struct DefaultNodeServiceFactory {
+    /// If true, disables the use of proxies.
+    ///
+    /// This speeds up the construction of new instances.
+    pub no_use_proxy: bool,
```
giac-mysten: Maybe?
```suggestion
    pub disable_proxy: bool,
```

crates/walrus-service/src/node/committee/node_service.rs
```diff
@@ -273,18 +272,54 @@ impl Service<Request> for RemoteStorageNode {
...
+
+    /// If true, disables the loading of native certificates.
+    ///
+    /// This speeds up the construction of new instances.
+    pub no_loading_native_certs: bool,
```
giac-mysten:
```suggestion
    pub disable_loading_native_certs: bool,
```

crates/walrus-service/src/node/committee/node_service.rs
```diff
@@ -273,18 +272,54 @@ impl Service<Request> for RemoteStorageNode {
...
+        encoding_config: &Arc<EncodingConfig>,
+    ) -> Result<Self::Service, ClientBuildError> {
+        let mut builder = walrus_sdk::client::Client::builder()
+            .authenticate_with_public_key(member.network_public_key.clone());
+
+        if self.no_loading_native_certs {
+            builder = builder.tls_built_in_root_certs(false);
+        }
+        if self.no_use_proxy {
+            builder = builder.no_proxy();
```
giac-mysten, line 315: Nit: there is an extra blank line [here](https://github.com/MystenLabs/walrus/blob/cda954701f7cbc57be9eee0d36788cfd1f25272d/crates/walrus-sdk/src/client.rs#L234)

crates/walrus-service/src/client/config.rs
```diff
@@ -55,21 +55,27 @@ pub struct ClientCommunicationConfig {
     pub reqwest_config: ReqwestConfig,
     /// The configuration specific to each node connection.
     pub request_rate_config: RequestRateConfig,
+    /// Disable the use of system proxies for communication.
+    pub disable_proxy: bool,
+    /// Disable the use of operating system certificates for authenticating the communication.
+    pub disable_native_certs: bool,
```
giac-mysten, line 61: Question: The `default` value for `bool` is false. This means that these values will be `false` in the client calls, most often, as people won't touch the defaults.

I think disabling certs is fine, as the pk for the node is anyways in the system object. But disabling the proxies may be problematic in general?

jpcsmith: > Question: The default value for bool is false. This means that these values will be false in the client calls, most often, as people won't touch the defaults.

Yes, and since `disable_proxy` is `false`, we use any configured proxy by default, and since `disable_native_certs` is `false`, we load native certs by default.

They should always need to opt-in to disabling proxies or native certs.

PR #883, opened 2024-09-27
feat(staking): adds pool token calculation

damirka:
- Implements pool token / WAL ratio in each staking pool. 
- Improves amount tracking for future epochs via PendingValues type.

### DIFF
--- a/crates/walrus-sui/src/types/move_structs.rs
+++ b/crates/walrus-sui/src/types/move_structs.rs
@@ -160,7 +160,7 @@ pub(crate) struct StakingPool {
     /// Currently active stake.
     active_stake: u64,
     /// Pool token balance.
-    token_balance: u64,
+    pool_token_balance: u64,
     /// Pending withdrawals from the pool token balance indexed by epoch.
     pending_pool_token_withdraw: Vec<(Epoch, u64)>,
     /// The commission rate for the pool.


crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -155,16 +155,24 @@ pub(crate) struct StakingPool {
...
+    /// Pool token balance.
+    token_balance: u64,
+    /// Pending additions to the pool token balance indexed by epoch.
+    pending_pool_token: Vec<(Epoch, u64)>,
+    /// Pending withdrawals from the pool token balance indexed by epoch.
+    pending_withdrawal_token: Vec<(Epoch, u64)>,
     /// The commission rate for the pool.
     commission_rate: u64,
+    /// Exchange rates table ID.
+    exchange_rates: ObjectID,
```
karlwuest, line 170: I think this should fix the issue:

```suggestion
    #[serde(deserialize_with = "deserialize_table")]
    exchange_rates: ObjectID,
```

contracts/walrus/sources/staking/exchange_rate.move
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// A utility module which implements an `ExchangeRate` struct and its methods.
+/// It stores a fixed point exchange rate between the Wal token and pool token.
+module walrus::pool_exchange_rate;
+
+#[allow(unused_const)]
+/// The scaling factor for the exchange rate.
+const SCALING: u128 = 1_000_000_000;
```
karlwuest: Question: what's the indended purpose of this scaling factor?

contracts/walrus/sources/staking/pending.move
```diff
@@ -0,0 +1,14 @@
...
+module walrus::pending;
+
+use sui::vec_map::{Self, VecMap};
+
+/// A value that contains values for E, E+1 and E+2.
+public struct Pending(VecMap<u32, u64>) has store, copy, drop;
+
+/// Create a new rotated value.
+public fun empty(): Pending { Pending(vec_map::empty()) }
+
```
karlwuest: Question: Is this (intended to be) used anywhere or was this replaced with `PendingValues`?

contracts/walrus/sources/staking/pending_values.move
```diff
@@ -0,0 +1,79 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+module walrus::pending_values;
+
+use sui::vec_map::{Self, VecMap};
+
+/// Represents a map of pending values. The key is the epoch when the value is
+/// pending, and the value is the amount of WALs or pool tokens.
+public struct PendingValues(VecMap<u32, u64>) has store, drop, copy;
```
karlwuest, line 10: Question: Do we expect epochs to be added out of order? If not, we could just use a ringbuffer (similar to the FutureAccountingRingBuffer) where we store the values such that they can be directly indexed and avoid iterating through the VecMap. Or do we anyway have an invariant that the VecMap is guaranteed to stay small?

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -151,6 +175,10 @@ public(package) fun stake(
     };
 
     let staked_amount = to_stake.value();
+    let pool_token_amount = pool
+        .exchange_rate_at_epoch(current_epoch)
+        .get_token_amount(staked_amount);
+
```
karlwuest: Important: I think this part and the line (191) below updating the pending pool tokens needs to be done during epoch change, because we need the exchange rate after the rewards from the epoch have been added, i.e. at the point where the stake becomes active.

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -186,18 +213,25 @@ public(package) fun request_withdraw_stake(
...
     // otherwise in E+1.
-    if (node_in_committee && wctx.committee_selected()) {
-        staked_wal.set_withdrawing(wctx.epoch() + 2);
+    // TODO: add a check that the node is in the committee: `node_in_committee &&`
+    // let node_in_committee = wctx.committee().contains(pool.id.as_inner());
+    let withdraw_epoch = if (wctx.committee_selected()) {
+        wctx.epoch() + 2
     } else {
-        staked_wal.set_withdrawing(wctx.epoch() + 1);
+        wctx.epoch() + 1
```
karlwuest, line 219: Comment/Suggestion: I think these withdrawal epochs are only correct if we do additional checks when withdrawing, in particular whether the node has transferred all shards already. Without this check (which is currently not implemented), we need to wait another epoch, because the funds may still be slashed if the transfer is not done. 

I would suggest to a) add an additional check now that the epoch change has been completed when withdrawing, and b) add a TODO comment + a new issue to convert this check into a check that the shards of the node have been transferred (for the mainnet milestone, since we don't have slashing yet anyway).

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -210,17 +244,34 @@ public(package) fun withdraw_stake(
...
     assert!(staked_wal.node_id() == pool.id.to_inner());
+    assert!(staked_wal.withdraw_epoch() >= wctx.epoch());
     assert!(staked_wal.activation_epoch() <= wctx.epoch());
     assert!(staked_wal.is_withdrawing());
 
-    let principal = staked_wal.unwrap();
-    pool.pending_withdrawal_amount = pool.pending_withdrawal_amount - principal.value();
-    principal.into_coin(ctx)
-}
+    let token_amount = staked_wal.value();
```
karlwuest: Important: The `token_amount` needs to be calculated from the past exchange rate. I think it should be:
```suggestion
    let token_amount = pool.exchange_rate_at_epoch(staked_wal.activation_epoch()).get_token_amount(staked_wal.value());
```

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -210,17 +244,34 @@ public(package) fun withdraw_stake(
...
+    let principal_amount = principal.value();
+    let total_amount = latest_exchange_rate.get_wal_amount(token_amount);
+    let rewards_amount = if (total_amount >= principal_amount) {
+        total_amount - principal_amount
+    } else 0;
+
+    // withdraw rewards. due to rounding errors, there's a chance that the
+    // rewards amount is higher than the rewards pool, in this case, we
+    // withdraw the maximum amount possible.
+    // TODO: add an assert that rounding is within the acceptable range.
```
karlwuest: Can you add an issue to keep track of this?

contracts/walrus/tests/staking/staking_pool_tests.move
```diff
@@ -3,109 +3,234 @@
...
+    let mut pool = pool().build(&wctx, ctx);
+    let mut staked_wal_a = pool.stake(mint(1000, ctx), &wctx, ctx);
+
+    assert_eq!(pool.pool_token_at_epoch(wctx.epoch()), 0);
+    assert_eq!(pool.pool_token_at_epoch(wctx.epoch() + 1), 1000);
+
+    // E1: node is in the committee, rewards are distributed, 1000 WAL received
+    let (wctx, ctx) = test.next_epoch();
+
+    pool.advance_epoch(mint(1000, ctx).into_balance(), &wctx);
```
karlwuest: Important: Since Alice's stake was not active in the epoch for which these rewards are paid, Alice should not receive any of these.

We should also have a test for this that is a little more complex and spans several epochs. I think the following test should be correct (but fail with the current implementation):

- Epoch 0: Alice stakes 1000
- Epoch 1: pool receives 100 rewards, Bob stakes 2000
- Epoch 2: pool receives 100 rewards, Alice requests withdrawal before committee selection
- Epoch 3: pool receives 100 rewards, Bob requests withdrawal before committee selection, Alice withdraws
- Epoch 4: Bob withdraws

In this case, Alice will get 1000 pool tokens, Bob will get 1818 (rounded down), then when withdrawing, Alice receives 1135 tokens (rounded down), Bob receives 2165.

contracts/walrus/sources/staking/pending.move
damirka: Oh, it's an artefact, overlooked!

contracts/walrus/sources/staking/exchange_rate.move
damirka: It's to implement more precise calculation later.

contracts/walrus/sources/staking/staking_pool.move
damirka: Just to be clear, are you saying that even in E+1, there's a period when the shard transfer hasn't been performed, and stake is not yet withdrawable?

karlwuest: Yes

contracts/walrus/sources/staking/pending_values.move
damirka: It will stay small, we will probably trigger flushing on stake action too.

damirka: It is currently possible to have some epochs skipped if `advance_epoch` wasn't called on a pool:
```
E1 -> 10
E2 -> 20
E4 -> 30
```

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -226,8 +226,8 @@ public(package) fun set_next_commission(
     cap: &StorageNodeCap,
     commission_rate: u64,
 ) {
-    let wctx = &self.new_walrus_context();
-    self.pools[cap.node_id()].set_next_commission(commission_rate, wctx);
+    let _wctx = &self.new_walrus_context();
```
karlwuest: Nit: remove these assignments if they are not needed.

contracts/walrus/sources/staking/exchange_rate.move
```diff
@@ -0,0 +1,60 @@
...
+    // When either amount is 0, that means we have no stakes with this pool.
+    // The other amount might be non-zero when there's dust left in the pool.
+    if (exchange_rate.wal_amount == 0 || exchange_rate.pool_token_amount == 0) {
+        return token_amount
+    };
+
+    let token_amount = (token_amount as u128) * SCALING;
+    let res = token_amount * exchange_rate.wal_amount / exchange_rate.pool_token_amount;
+
+    (res / SCALING) as u64
```
karlwuest: Nit: I'm not sure if the SCALING makes a difference here. Do you have an example where this matters?

contracts/walrus/sources/staking/pending.move
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Mysten Labs, Inc.
```
karlwuest: Just commenting as a reminder to delete this file.

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -102,9 +137,9 @@ public fun is_withdrawing(sw: &StakedWal): bool {
 /// Aborts if the `node_id` or `activation_epoch` of the staked WALs do not match.
 public fun join(sw: &mut StakedWal, other: StakedWal) {
     let StakedWal { id, state, node_id, activation_epoch, principal } = other;
-    assert!(sw.state == state);
-    assert!(sw.node_id == node_id);
-    assert!(sw.activation_epoch == activation_epoch);
+    assert!(sw.state == state, EMetadataMismatch);
```
karlwuest, line 151: Important: This will prevent almost all withdrawing StakedWal from being joined, except if they have the same token amount. We should either explicitly exclude withdrawing StakedWal from being merged or have a slightly more sophisticated check (and merge) here that allows withdrawing StakedWal with the same withdrawal epoch to be merged.

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -102,9 +137,9 @@ public fun is_withdrawing(sw: &StakedWal): bool {
...
+    assert!(sw.node_id == node_id, EMetadataMismatch);
+    assert!(sw.activation_epoch == activation_epoch, EMetadataMismatch);
     id.delete();
 
     sw.principal.join(principal);
```
karlwuest, line 158: Important: If we join two StakedWal that have the same Withdrawing state, this will cause the pool_token_amount to be too low.

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -153,113 +184,166 @@ public(package) fun stake(
...
+    wctx: &WalrusContext,
+) {
+    // process the pending and withdrawal amounts
+    let current_epoch = wctx.epoch();
+
+    assert!(current_epoch > pool.latest_epoch, EPoolAlreadyUpdated);
+    assert!(rewards.value() == 0 || pool.wal_balance > 0, EIncorrectEpochAdvance);
+
+    // if rewards are calculated only for full epochs, rewards addition should
+    // happen prior to pool token calculation. Othewise we can add then to the
```
karlwuest:
```suggestion
    // happen prior to pool token calculation. Otherwise we can add then to the
```

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -270,64 +354,74 @@ public(package) fun destroy_empty(pool: StakingPool) {
...
 /// Set the state of the pool to `Active`.
 public(package) fun set_is_active(pool: &mut StakingPool) {
     assert!(pool.is_new());
     pool.state = PoolState::Active;
 }
 
-/// Returns the amount stored in the `active_stake`.
-public(package) fun active_stake(pool: &StakingPool): u64 {
-    pool.active_stake
+public(package) fun pool_token_at_epoch(pool: &StakingPool, epoch: u32): u64 {
```
karlwuest: Question/Nit: We cannot query this for any past epoch, right? Can you add a comment explaining what exactly we get from this function and maybe give it a name that fits better?

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -270,64 +354,74 @@ public(package) fun destroy_empty(pool: StakingPool) {
...
-/// Returns the amount stored in the `stake_to_withdraw`.
-public(package) fun stake_to_withdraw_amount(pool: &StakingPool): u64 {
-    pool.stake_to_withdraw.value()
+/// Returns the expected active stake for epoch `E` for the pool. It processes
+/// the pending stake and withdrawal requests from the current epoch to `E`.
+///
+/// Should be the main function to calculate the active stake for the pool at
+/// the given epoch, due to the complexity of the pending stake and withdrawal
+/// requests, and lack of immediate updates.
+public(package) fun wal_balance_at_epoch(pool: &StakingPool, epoch: u32): u64 {
```
karlwuest, line 394: ~Important: I don't think that this function gives us the correct expected stake at `epoch`. The problem is that we are using all pending pool token withdrawals to compute this. I don't think it's solvable without storing additional information. I'll try to write a test case for this.~

Edit: The calculation seems to be correct in the scenario that I thought would fail. I'll add the test case as suggestion in the tests.

contracts/walrus/tests/staking/staking_pool_tests.move
```diff
@@ -1,111 +1,513 @@
...
+    assert_eq!(balance_b.destroy_for_testing(), 1500);
+
+    pool.destroy_empty()
+}
+
+#[test]
+// Scenario:
+// E0: Alice stakes: 1000 WAL;
+// E1: Bob stakes: 2000 WAL;
+// E2: +1000 Rewards; Alice requests withdrawal;
```
karlwuest, line 188: Nit: (this applies to multiple comments) The rewards are paid for the previous epoch and I think it would help to understand why the rewards that Alice receives are correct if this is made clear in the comments.

contracts/walrus/tests/staking/staking_pool_tests.move
```diff
@@ -1,111 +1,594 @@
...
+    let balance_d = pool.withdraw_stake(staked_wal_d, &wctx);
+
+    assert_eq!(balance_a.destroy_for_testing(), 4000);
+    assert_eq!(balance_b.destroy_for_testing(), 3000);
+    assert_eq!(balance_c.destroy_for_testing(), 6000);
+    assert_eq!(balance_d.destroy_for_testing(), 3000);
+
+    pool.destroy_empty()
+}
+
```
karlwuest, line 420: Another test case that we can add (which I initially thought would fail):

```suggestion

#[test]
// Scenario:
// E0: Alice stakes: 1000 WAL;
// E1: Bob stakes: 1000 WAL; Chalie stakes: 1000 WAL;
// E2: +1000 Rewards for E1; Alice requests withdrawal;
// E3: +1000 Rewards for E2; Bob requests withdrawal;
// E4: +1000 Rewards for E3;
// E5: +1000 Rewards for E4;
fun correct_stake_pending_withdrawals() {
    let mut test = context_runner();

    // E0: Alice stakes 1000 WAL
    let (wctx, ctx) = test.current();
    let mut pool = pool().build(&wctx, ctx);
    let mut staked_a = pool.stake(mint_balance(1000), &wctx, ctx);

    // No rewards yet, stake for next committee selection is 1000
    assert_eq!(pool.wal_balance_at_epoch(wctx.epoch() + 1), 1000);

    // E1: Bob stakes 1000 WAL; Chalie stakes 1000 WAL;
    let (wctx, ctx) = test.next_epoch();
    pool.advance_epoch(mint_balance(0), &wctx);
    let mut staked_b = pool.stake(mint_balance(1000), &wctx, ctx);
    let staked_c = pool.stake(mint_balance(1000), &wctx, ctx);

    // No rewards yet, stake for next committee selection is 3000
    assert_eq!(pool.wal_balance_at_epoch(wctx.epoch() + 1), 3000);

    // E2: +1000 Rewards for E1; Alice requests withdrawal
    let (wctx, _) = test.next_epoch();
    pool.advance_epoch(mint_balance(1000), &wctx);
    pool.request_withdraw_stake(&mut staked_a, &wctx);


    // All rewards for the previous epoch go to Alice, Alice's stake does not count anymore.
    // Stake is 2000 (Bob + Chalie)
    assert_eq!(pool.wal_balance_at_epoch(wctx.epoch() + 1), 2000);

    // E3: +1000 Rewards for E2; Bob requests withdrawal
    let (wctx, _) = test.next_epoch();
    pool.advance_epoch(mint_balance(1000), &wctx);
    pool.request_withdraw_stake(&mut staked_b, &wctx);

    // Half of the rewards for the previous epoch go to Alice, a quarter to Bob and Charlie each.
    // Alice's and Bob's stakes does not count anymore.
    // Stake is 1250 (Chalie + rewards on his stake)
    assert_eq!(pool.wal_balance_at_epoch(wctx.epoch() + 1), 1250);

    // E4: +1000 Rewards for E3;
    let (wctx, _) = test.next_epoch();
    pool.advance_epoch(mint_balance(1000), &wctx);

    // Half of the rewards for the previous epoch go to Bob and Charlie each.
    // Alice's and Bob's stakes does not count anymore.
    // Stake is 1750 (Chalie + rewards on his stake)
    assert_eq!(pool.wal_balance_at_epoch(wctx.epoch() + 1), 1750);

    // E5: +1000 Rewards for E4;
    let (wctx, _) = test.next_epoch();
    pool.advance_epoch(mint_balance(1000), &wctx);

    // All rewards go to Charlie
    // Stake is 2750 (Chalie + rewards on his stake)
    assert_eq!(pool.wal_balance_at_epoch(wctx.epoch() + 1), 2750);

    destroy(pool);
    destroy(vector[
        staked_a,
        staked_b,
        staked_c,
    ]);
}

```

contracts/walrus/tests/staking/staking_pool_tests.move
damirka: Added a comment at the top to revisit.

Issue #859, opened 2024-09-26
Set up WAL faucet/exchange for Testnet


mlegner:
We need a simple way for users to get WAL. The idea is to set up a contract that allows users to exchange SUI for WAL at a 1:1 rate.

This should also be added to the `walrus` CLI as a subcommand (like `walrus get-wal`).

- [x] Add contracts. -> #884 
- [ ] Add bindings in Rust.
- [ ] Implement command on the `walrus` binary.
- [ ] Add deployment and funding to the `walrus-deploy` binary.

PR #884, opened 2024-09-27
feat(move): add SUI/WAL exchange contract

mlegner:
Also add metadata to existing contracts.

Contributes to #859

### DIFF
--- a/contracts/wal_exchange/sources/wal_exchange.move
+++ b/contracts/wal_exchange/sources/wal_exchange.move
@@ -15,22 +15,67 @@ const EInsufficientFundsInExchange: vector<u8> =
 const EInsufficientInputBalance: vector<u8> =
     b"The provided coin has insufficient balance for the exchange";
 
+#[error]
+const EUnauthorizedAdminCap: vector<u8> =
+    b"The provided AdminCap is not authorized for this exchange";
+
+#[error]
+const EInvalidExchangeRate: vector<u8> =
+    b"The provided exchange rate is invaled; neither field can be 0.";
+
 /// A public exchange that allows exchanging SUI for WAL at a fixed exchange rate.
 public struct Exchange has key, store {
     id: UID,
     wal: Balance<WAL>,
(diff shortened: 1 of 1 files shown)


contracts/wal_exchange/sources/wal_exchange.move
```diff
@@ -0,0 +1,162 @@
...
+/// A public exchange that allows exchanging SUI for WAL at a fixed exchange rate.
+public struct Exchange has key, store {
+    id: UID,
+    wal: Balance<WAL>,
+    sui: Balance<SUI>,
+    wal_per_sui: u64,
+}
+
+/// Creates a new shared exchange with the given exchange rate.
+public fun new(wal_per_sui: u64, ctx: &mut TxContext) {
```
karlwuest: Suggestion: Return an `AdminCap` here that then allows setting the exchange rate (+ potentially withdrawing the balances).

contracts/wal_exchange/sources/wal_exchange.move
```diff
@@ -0,0 +1,162 @@
...
+#[error]
+const EInsufficientInputBalance: vector<u8> =
+    b"The provided coin has insufficient balance for the exchange";
+
+/// A public exchange that allows exchanging SUI for WAL at a fixed exchange rate.
+public struct Exchange has key, store {
+    id: UID,
+    wal: Balance<WAL>,
+    sui: Balance<SUI>,
+    wal_per_sui: u64,
```
karlwuest: Suggestion: This limits the factor to an integer. Instead we could store the exchange rate as a tuple of two u64 to provide a rational number. E.g. something like the following:

```
public struct ExchangeRate has copy, drop, store {
    wal: u64,
    sui: u64,
}

fun wal_to_sui(self: &ExchangeRate, amount: u64): u64 {
    amount * self.sui / self.wal
}

fun sui_to_wal(self: &ExchangeRate, amount: u64): u64 {
    amount * self.wal / self.sui
}
```

PR #890, opened 2024-09-30
feat: include epoch explicitly when sending "epoch sync done"

### LINKED ISSUE #878: Make epoch for which we attest sync done explicit.
This prevents issues where a node that is resyncing tries to attest to the sync for an older epoch, but having this counted for the current epoch, even though the node has not fully synced up to the current epoch yet.

karlwuest:
closes #878

### DIFF
--- a/contracts/walrus/sources/staking/staking_inner.move
+++ b/contracts/walrus/sources/staking/staking_inner.move
@@ -66,6 +66,9 @@ const EDuplicateSyncDone: vector<u8> = b"Node already attested that sync is done
 #[error]
 const ENoStake: vector<u8> = b"Total stake is zero for apportionment";
 
+#[error]
+const ENotInCommittee: vector<u8> = b"The node is not part of the current committee";
+
 /// The epoch state.
 public enum EpochState has store, copy, drop {
     // Epoch change is currently in progress. Contains the weight of the nodes that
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/staking.move
```diff
@@ -153,8 +153,13 @@ public fun shard_transfer_failed(
 }
 
 /// Signals to the contract that the node has received all its shards for the new epoch.
-public fun epoch_sync_done(staking: &mut Staking, cap: &mut StorageNodeCap, clock: &Clock) {
-    staking.inner_mut().epoch_sync_done(cap, clock);
+public fun epoch_sync_done(
+    staking: &mut Staking,
+    cap: &mut StorageNodeCap,
+    epoch: u32,
```
damirka, line 159: How much do we trust this value?

karlwuest: This value is the epoch for which the storage node intends to signal that it has finished the epoch sync. We don't trust it in the sense that we can assume that this is actually the current epoch.

Basically, the intention of this is that an out-of-sync storage node does not accidentally (or because it is being attacked) certify that it is done with the sync of, e.g., epoch `e + 1`, event though it just finished the sync for epoch `e`.

PR #891, opened 2024-09-30
feat: parsing ShardsReceived and ShardRecoveryStart events

### LINKED ISSUE #881: Add bindings for `ShardsReceived` event
Currently, this causes the following error:
```
2024-09-27T11:52:27.315855Z ERROR poll_for_events:sui-event{event_id=EventID { tx_digest: TransactionDigest(8NJrJwkGbou2TL3SwdtTE3ru9ACBdXVSztSfCStw5EJX), event_seq: 0 } event_type=StructTag { address: 40bd670e1a984bb3701a9767421f49269f7ef5b11805bb0d113f4e6351e4e6bc, module: Identifier("events"), name: Identifier("ShardsReceived"), type_params: [] }}: walrus_sui::client::read_client: could not convert event
```

halfprice:
Parsing ShardsReceived and ShardRecoveryStart events in storage node.

Also convert parse error into an assertion failure so that new events added in the future must be parsed in the storage node.

Close #881

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -552,92 +552,103 @@ impl StorageNode {
         bail!("event stream for blob events stopped")
     }
 
+    #[tracing::instrument(skip_all)]
     async fn process_event(
         &self,
         element_index: usize,
         stream_element: IndexedStreamElement,
     ) -> anyhow::Result<()> {
-        let storage = &self.inner.storage;
         let _timer_guard = &self
             .inner
             .metrics
             .event_process_duration_seconds
             .with_label_values(&[stream_element.element.label()])
             .start_timer();
-        if let Some(blob_event) = stream_element.element.blob_event() {
-            storage.update_blob_info(blob_event)?;
-        }
         match stream_element.element {
-            EventStreamElement::ContractEvent(ContractEvent::BlobEvent(BlobEvent::Registered(
-                event,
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -540,86 +539,107 @@ impl StorageNode {
...
+                    span.record("otel.status_message", field::display(err));
+                })
+                .instrument(span)
+                .await?;
         }
 
         bail!("event stream for blob events stopped")
     }
 
+    async fn process_event(
```
jpcsmith, line 556: Suggestion: Further simplify this by adding methods for each sub-event type, to avoid having a single growing list of match conditions, for example:
```rust
       match stream_element.element {
            EventStreamElement::ContractEvent(ContractEvent::BlobEvent(event)) => process_blob_event(..),
            EventStreamElement::ContractEvent(ContractEvent::EpochChangeEvent(event)) => process_epoch_change_event(...),
            EventStreamElement::CheckpointBoundary => {
                self.inner.mark_element_at_index(element_index)?;
            }
       }

halfprice: SG! Done.

mlegner: Thanks a lot for the suggestion and change, I was just doing the same refactoring in my upcoming PR. 😁

PR #893, opened 2024-10-01
test: various performance improvements for E2E tests

### LINKED ISSUE #865: Use `sui start` for running integration tests
Integration test experience may be improved by using an external binary to provide sui. See `sui start` and the example of its use in CI here: https://github.com/MystenLabs/suins-contracts/blob/main/.github/workflows/sdk_test.yml

jpcsmith:
This adds a number of changes that help to speed up running integration tests. For example, they reduce `cargo test --test test_client -- --ignored` from taking around 235 seconds to just 30 seconds:

#### Using an external sui test cluster

An external test cluster started with `sui start` can be used for tests through the use of the `SUI_TEST_CONFIG_DIR` environment variable. Each test will still continue to publish it's own walrus packages, but they will do so to the external test cluster instead of the in-process test cluster. The cluster needs to already be running.

This removes the time required to build the test cluster, which, for some tests, was a large portion of their runtime. While there was a shared in-process test cluster, running many tests against it seemed to result in errors and retries.

Unfortunately, this does not yet work with `cargo nextest` (despite it being the motivating reason for not just running `sui start` from inside the process), as this would require inter-process sharing of the cluster's wallet config.

#### Use a single transaction to fund wallets in a test

A large amount of time was spent waiting for a wallet to be funded before releasing the lock on the shared admin wallet to be able to fund the next wallet. This change funds multiple wallet simultaneously to avoid the wait.

#### Register storage nodes in parallel

Registers and submits stake for storage nodes in parallel. Reduces each test's runtime by several seconds.

#### Cache compiled move modules

Since the move code does not change across tests, it only needs to be compiled once. This change caches the compiled value in a static variable. This provides speedups when running many tests in parallel (i.e., 1 test will not benefit; 2 will not really benefit since the second test which runs in parallel will block waiting on the package to be compiled by the first test; after a point though, tests are scheduled and start that already see the package compiled).

Unfortunately, this does not benefit `cargo nextest` as that would require caching it to the filesystem and doing synchronisation across the processes.

Closes #865

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -13,8 +13,7 @@ use std::{
     num::NonZeroU16,
     path::PathBuf,
     str::FromStr,
-    sync::{Arc, Mutex, MutexGuard, OnceLock},
-    thread,
+    sync::{Arc, Mutex},
 };
 
 use async_trait::async_trait;
@@ -23,11 +22,7 @@ use futures::{stream::FuturesUnordered, StreamExt};
 use prometheus::Registry;
 use sui_types::base_types::ObjectID;
 use tempfile::TempDir;
-use tokio::{
-    runtime::{Builder, Runtime},
-    task::JoinHandle,
-    time::Duration,
-};
+use tokio::{task::JoinHandle, time::Duration};
 use tokio_stream::{wrappers::BroadcastStream, Stream};
 use tokio_util::sync::CancellationToken;
 use tracing::Instrument;
@@ -337,7 +332,6 @@ impl SimStorageNodeHandle {
 
     /// Builds and runs a storage node with the provided configuration. Returns the handles to the
     /// REST API, the node, and the event processor.
-    #[cfg(msim)]
     async fn build_and_run_node(
         config: StorageNodeConfig,
         cancel_token: CancellationToken,
@@ -832,7 +826,6 @@ impl StorageNodeHandleBuilder {
             rest_api_address: storage_node_config.rest_api_address,
             metrics_address: storage_node_config.metrics_address,
             cancel_token,
-            #[cfg(msim)]
             node_id: handle.id(),
             storage_node_capability: self.storage_node_capability,
         })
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/src/system_setup.rs
```diff
@@ -56,6 +61,28 @@ fn get_pkg_id_from_tx_response(tx_response: &SuiTransactionBlockResponse) -> Res
         .ok_or_else(|| anyhow!("no immutable object was created"))
 }
 
+fn compile_package(package_path: &Path) -> Arc<CompiledPackage> {
+    if cfg!(any(test, feature = "test-utils")) {
```
mlegner: Important: As test-utils is a default feature, this will almost always be true. How about we instead always do the caching but have an additional function parameter force_recompilation or something like this?

jpcsmith: As you pointed out, this compiles using the test-config even in production, due to the feature being activated. Refactoring this requires specifying both the BuildConfig (for tests vs non-tests) and whether to force a rebuild. This then needs to be passed through the entire call chain to the compile package call.

crates/walrus-service/src/test_utils.rs
```diff
@@ -1762,20 +1763,32 @@ pub mod test_cluster {
...
+            inner: wallets,
+            // The on-file representation of the wallets are dropped at the end of scope.
+            temp_dir,
+        } = test_utils::create_and_fund_wallets_on_cluster(sui_cluster.clone(), members.len())
+            .await?;
+
+        // In simtest, storage nodes load sui wallet config from the `tmp_dir`. We need to keep the
+        // directory alive throughout the test.
+        #[cfg(msim)]
+        Box::leak(Box::new(temp_dir));
```
karlwuest, line 1840: Question: Does this not prevent the temp dir from being deleted on disk, since the value is now never dropped?

mlegner: AFAIU, this will remain alive while the process is alive. There shouldn't be an issue removing the directory after the tests have finished.

PR #896, opened 2024-10-01
feat(client): configurable extra time for sliver store

giac-mysten:
* Add the possibility to configure the amount of extra time the client waits after storing the first 2/3 slivers. 
* Increase the defaults to 50% of the original store time + 0.5s
* Create a "test" that writes the example client config to keep it in sync with changes (was outdated).

### DIFF
--- a/crates/walrus-service/src/client/config.rs
+++ b/crates/walrus-service/src/client/config.rs
@@ -346,8 +346,7 @@ impl SliverWriteExtraTime {
     ///
     /// The extra time is computed as `store_time * factor + base`.
     pub fn extra_time(&self, store_time: Duration) -> Duration {
-        let extra_time =
-            Duration::from_millis((store_time.as_millis() as f64 * self.factor) as u64);
+        let extra_time = Duration::from_nanos((store_time.as_nanos() as f64 * self.factor) as u64);
         self.base + extra_time
     }
 }


crates/walrus-service/src/client/config.rs
```diff
@@ -329,3 +331,55 @@ pub(crate) mod default {
...
+}
+
+impl SliverWriteExtraTime {
+    /// Returns the extra time for the given time.
+    ///
+    /// The extra time is computed as `store_time / fraction + base`.
+    pub fn extra_time(&self, store_time: Duration) -> Duration {
+        let extra_time =
+            Duration::from_millis((store_time.as_millis() as f64 * self.factor) as u64);
+        self.base + extra_time
```
giac-mysten: Remark: 2a551ac28b886ff2200bfd44ae970655414f9c38 changes from just being able to provide a "denominator", such that `extra_time  = store_time / denominator` to a factor. However, this requires lossy conversion to `f64` and back. 

I think we are fine with this.

crates/walrus-service/src/client/config.rs
```diff
@@ -329,3 +331,55 @@ pub(crate) mod default {
...
+    /// The multiplication factor for the time it took to store n-f sliver.
+    pub factor: f64,
+    /// The minimum extra time.
+    pub base: Duration,
+}
+
+impl SliverWriteExtraTime {
+    /// Returns the extra time for the given time.
+    ///
+    /// The extra time is computed as `store_time / fraction + base`.
```
mlegner: Nit:
```suggestion
    /// The extra time is computed as `store_time * factor + base`.
```

crates/walrus-service/src/client/config.rs
mlegner: Agreed, I don't think this is an issue.

crates/walrus-service/src/client/config.rs
```diff
@@ -329,3 +331,55 @@ pub(crate) mod default {
...
+    pub fn extra_time(&self, store_time: Duration) -> Duration {
+        let extra_time =
+            Duration::from_millis((store_time.as_millis() as f64 * self.factor) as u64);
```
mlegner: Question: Any reason to use `millis` instead of, e.g., `nanos`?

crates/walrus-service/src/client/config.rs
```diff
@@ -329,3 +331,55 @@ pub(crate) mod default {
...
+
+#[cfg(test)]
+mod tests {
+    use rand::{rngs::StdRng, SeedableRng};
+    use sui_types::base_types::ObjectID;
+
+    /// Serializes a default config to the example file when tests are run.
+    ///
+    /// This test ensures that the `client_config_example.yaml` is kept in sync with the config
+    /// struct in this file.
```
mlegner, line 371: Amazing! We should do something similar with the `node_config`, but this can be done in a separate PR.

As a side note: I was briefly looking into whether we could use the docstrings as comments to explain the meaning of some of the config parameters but didn't find a simple solution.

crates/walrus-service/src/client/config.rs
giac-mysten: No strong reasons. I though that since both `millis` and `nanos` are provided as `u128` and the `millis` resolution is probably enough, the chance of problematic conversions (overflow/large loss in resolution) when casting to `f64` is lower with `millis`. 

Should I change?

crates/walrus-service/src/client/config.rs
giac-mysten: ah that would be cool indeed. The closest thing I have found is the [documented crate](https://github.com/cyqsimon/documented): it would allow to access the docs at runtime, and then I guess manually add them to the yaml.

Also FYI this discussion on [serializing docstrings in serde](https://github.com/serde-rs/serde/issues/1430)

crates/walrus-service/src/client/config.rs
mlegner: No strong opinion. I guess neither precision nor overflows are going to be an issue here: If we have an overflow with nanoseconds in a `u64`, our client is anyway unusable... 🤣

crates/walrus-service/src/client/config.rs
mlegner: Created a separate issue for this: #898.

crates/walrus-service/src/client/config.rs
giac-mysten: agreed. I'll change to `nanos`.

PR #899, opened 2024-10-01
fix(node): idempotent blob info updates for events

### LINKED ISSUE #888: Fix event processing issues
The new implementation of the blob status tracking (#764, #825) made incorrect assumptions about the way events are processed:
Currently, events are dispatched in order but are processed out-of-order and potentially multiple times (in case of a crash/restart). 

We need to fix this in some way. Making the event handling fully sequential is not a good solution.

My suggestion is to just make the *status updates* sequential and only do the update once for each event. This could be achieved by adding the last event index to each `BlobInfo` or, more efficiently, keep a global `last_event_index_handled_for_status` entry in the DB.

mlegner:
Open TODOs:
- [x] Fix existing tests.
- [x] Add test for idempotency.

Closes #888

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -477,26 +477,25 @@ impl StorageNode {
     /// Continues the event stream from the last committed event.
     async fn continue_event_stream(
         &self,
-    ) -> anyhow::Result<Pin<Box<dyn Stream<Item = IndexedStreamElement> + Send + Sync + '_>>> {
+    ) -> anyhow::Result<(
+        Pin<Box<dyn Stream<Item = IndexedStreamElement> + Send + Sync + '_>>,
+        usize,
+    )> {
         let storage = &self.inner.storage;
-        let from_event_id = storage
+        let (from_event_id, next_event_index) = storage
             .get_event_cursor_and_next_index()?
-            .map(|(cursor, _)| cursor);
-        let from_element_index = self.next_event_index()?;
-        let event_cursor = EventStreamCursor::new(from_event_id, from_element_index);
+            .map_or((None, 0), |(cursor, index)| (Some(cursor), index));
+        let event_cursor = EventStreamCursor::new(from_event_id, next_event_index);
 
-        Ok(Box::into_pin(
-            self.inner.event_manager.events(event_cursor).await?,
+        Ok((
+            Box::into_pin(self.inner.event_manager.events(event_cursor).await?),
+            next_event_index.try_into().expect("64-bit architecture"),
         ))
     }
 
     async fn process_events(&self) -> anyhow::Result<()> {
-        let event_stream = self.continue_event_stream().await?;
+        let (event_stream, next_event_index) = self.continue_event_stream().await?;
 
-        let next_event_index: usize = self
-            .next_event_index()?
-            .try_into()
-            .expect("64-bit architecture");
         let index_stream = stream::iter(next_event_index..);
         let mut maybe_epoch_at_start = Some(self.inner.committee_service.get_epoch());
 
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -110,8 +110,7 @@ impl BlobInfoTable {
     }
 
     pub fn delete(&self, batch: &mut DBBatch, blob_id: &BlobId) -> Result<(), TypedStoreError> {
-        batch.delete_batch(&self.blob_info, [blob_id])?;
-        Ok(())
+        batch.delete_batch(&self.blob_info, [blob_id])
     }
 
     /// Returns an iterator over all blobs that were certified before the specified epoch in the
@@ -120,12 +119,12 @@ impl BlobInfoTable {
     pub fn certified_blob_info_iter_before_epoch(
         &self,
         before_epoch: Epoch,
-        starting_blob_id: Bound<BlobId>,
+        starting_blob_id_bound: Bound<BlobId>,
     ) -> BlobInfoIterator {
         BlobInfoIter::new(
             Box::new(
                 self.blob_info
-                    .safe_range_iter((starting_blob_id, Unbounded)),
+                    .safe_range_iter((starting_blob_id_bound, Unbounded)),
             ),
             before_epoch,
             true,
@@ -219,21 +218,20 @@ where
     type Item = Result<(BlobId, BlobInfo), TypedStoreError>;
 
     fn next(&mut self) -> Option<Self::Item> {
-        let mut item = self.iter.next();
-        if !self.only_certified {
-            return item;
-        }
-
-        while let Some(Ok((_, ref blob_info))) = item {
-            if matches!(
-                blob_info.initial_certified_epoch(),
-                Some(initial_certified_epoch) if initial_certified_epoch < self.before_epoch
-            ) {
-                return item;
+        for item in self.iter.by_ref() {
+            let Ok((_, blob_info)) = &item else {
+                return Some(item);
+            };
+            if !self.only_certified
+                || matches!(
+                    blob_info.initial_certified_epoch(),
+                    Some(initial_certified_epoch) if initial_certified_epoch < self.before_epoch
+                )
+            {
+                return Some(item);
             }
-            item = self.iter.next();
         }
-        item
+        None
     }
 }
 


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -3,17 +3,247 @@
...
+    pub fn options(db_config: &DatabaseConfig) -> [(&'static str, Options); 2] {
+        let mut blob_info_options = db_config.blob_info.to_options();
+        blob_info_options.set_merge_operator("merge blob info", merge_blob_info, |_, _, _| None);
+        [
+            (Self::BLOBINFO_COLUMN_FAMILY_NAME, blob_info_options),
+            (
+                Self::EVENT_INDEX_COLUMN_FAMILY_NAME,
+                // Doesn't make sense to have special options for the table containing a single
+                // value.
+                DatabaseTableOptions::default().to_options(),
```
halfprice, line 68: Suggestion: I wonder if it helps with readability if we create a DatabaseConfig::default, and use `db_config.default` here. Otherwise, the function takes a db_config but not really uses here.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -3,17 +3,247 @@
...
+            return Ok(());
+        }
+
+        let operation = BlobInfoMergeOperand::from(event);
+        tracing::debug!(?operation, "updating blob info");
+
+        let mut batch = self.blob_info.batch();
+        self.merge_blob_info_batch(&mut batch, &event.blob_id(), &operation)?;
+        batch.insert_batch(&self.latest_handled_event_index, [(&(), event_index)])?;
+        batch.write()
```
halfprice: Note: this is not an issue currently since we process events one by one, but want to mention that test-and-set on `latest_handled_event_index` is not atomic in this function. Maybe add in the function doc that this function cannot be called concurrently.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -3,17 +3,247 @@
...
+        blob_id: &BlobId,
+        operand: &BlobInfoMergeOperand,
+    ) -> Result<(), TypedStoreError> {
+        batch.partial_merge_batch(&self.blob_info, [(blob_id, operand.to_bytes())])?;
+        Ok(())
+    }
+
+    pub fn delete(&self, batch: &mut DBBatch, blob_id: &BlobId) -> Result<(), TypedStoreError> {
+        batch.delete_batch(&self.blob_info, [blob_id])?;
+        Ok(())
```
halfprice: Nit: this can just be returning `batch.delete_batch(&self.blob_info, [blob_id])`

crates/walrus-service/src/node.rs
```diff
@@ -652,10 +656,13 @@ impl StorageNode {
...
-        let storage = &self.inner.storage;
-        let index: Option<u64> = storage.get_event_cursor()?.map(|(index, _)| index);
-        Ok(index.unwrap_or(0))
+    fn next_event_index(&self) -> anyhow::Result<u64> {
+        Ok(self
+            .inner
+            .storage
+            .get_event_cursor()?
+            // TODO(mlegner): Make sure this is correct and doesn't need a `+ 1`.
+            .map_or(0, |(index, _)| index))
```
halfprice: I think what `EventCursorTable` store is the number of events processed, which is indeed the next event index. If my understanding is correct, maybe change `index` to `count` for consistency.

crates/walrus-service/src/node.rs
```diff
@@ -480,7 +480,7 @@ impl StorageNode {
     ) -> anyhow::Result<Pin<Box<dyn Stream<Item = IndexedStreamElement> + Send + Sync + '_>>> {
         let storage = &self.inner.storage;
         let from_event_id = storage.get_event_cursor()?.map(|(_, cursor)| cursor);
-        let from_element_index = self.get_last_committed_event_index()?;
+        let from_element_index = self.next_event_index()?;
```
halfprice: suggestion: both line 482 and 483 call get_event_cursor, so we may want to combine them to only call it once (the pattern was existed before this PR).

crates/walrus-service/src/node.rs
```diff
@@ -491,9 +491,11 @@ impl StorageNode {
     async fn process_events(&self) -> anyhow::Result<()> {
         let event_stream = self.continue_event_stream().await?;
 
-        let from_element_index = self.get_last_committed_event_index()?;
-        let next_index: usize = from_element_index.try_into().expect("64-bit architecture");
-        let index_stream = stream::iter(next_index..);
+        let next_event_index: usize = self
```
halfprice: suggestion: we can also use the value of `next_event_index()` from continue_event_stream. In this way, they will be consistent.

crates/walrus-service/src/node/storage.rs
```diff
@@ -504,15 +407,12 @@ impl Storage {
...
-            .filter_map(|blob_info| match blob_info {
-                Ok((blob_id, blob_info)) => match blob_info.initial_certified_epoch() {
-                    Some(epoch) if epoch < current_epoch => Some(Ok(blob_id)),
-                    _ => None,
-                },
-                Err(e) => Some(Err(e)),
-            })
+            .certified_blob_info_iter_before_epoch(
+                current_epoch,
+                Included(request.starting_blob_id()),
```
halfprice, line 414: Nice!

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -3,17 +3,247 @@
...
+    }
+
+    /// Returns an iterator over all blobs that were certified before the specified epoch in the
+    /// blob info table starting with the `starting_blob_id` bound.
+    #[tracing::instrument(skip_all)]
+    pub fn certified_blob_info_iter_before_epoch(
+        &self,
+        before_epoch: Epoch,
+        starting_blob_id: Bound<BlobId>,
```
halfprice: Nit: maybe something like `starting_blob_id_bound` to indicate that this is the starting `bound` of the scan?

crates/walrus-service/src/node/storage/blob_info.rs
jpcsmith, line 68: Remark: While it probably does not make sense to make this table configurable via an external config, we will likely want to configure this (and other similar tables) that store a single value for that use case.

mlegner: Yes, good point. I actually thought there already existed an issue for this, but I cannot find it. I'll create a new one.

> the function takes a db_config but not really uses here.

It uses it for the other table.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Good point. I already have in some docstring that the blob-info update must be called sequentially, but I can make this even clearer.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: I wanted to do this, but the batch operations all return a `Result<&mut DBBatch, _>`.

crates/walrus-service/src/node.rs
mlegner: I agree with your assessment, will see if I can make the naming clearer.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -3,17 +3,247 @@
...
+                DatabaseTableOptions::default().to_options(),
+            ),
+        ]
+    }
+
+    /// Updates the blob info for a blob based on the [`BlobEvent`].
+    ///
+    /// Only updates the info if the provided `event_index` hasn't been processed yet.
+    #[tracing::instrument(skip(self))]
+    pub fn update_blob_info(
```
jpcsmith, line 77: Question: Is my understanding correct that all (non-repeated) calls to `update_blob_info` must be sequential? That is, `event_index > latest_handled_event_index + 1` is not allowed? If so, it would be good to assert this as a pre-condition int he method.

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: Other ones do, but `delete_batch` returns `Result<(), TypedStoreError>`

mlegner: Oh, interesting (and weirdly inconsistent 🤷)... I'll apply your suggestion in that case. 🙂

crates/walrus-service/src/node/storage/blob_info.rs
jpcsmith: ~Question: Is my understanding correct that all (non-repeated) calls to update_blob_info must be sequential? That is, event_index > latest_handled_event_index.get() + 1 is not allowed? If so, it would be good to assert this as a pre-condition int he method.~

~Remark: Given the above pre-condition, we have the guarantee that after the test the assertion `event_index == latest_handled_event_index.get() + 1` is true. The above non-atomicity would then only be a problem if the function was called with the same `event_index` twice.~

In general though,  it is a problem that methods on `Sync` structs that take `&self` but are not safe to be called concurrently are present. This results in surprises as the compiler is unable to enforce that the method is only called in a valid way. This happens, I believe, with  `BlobSync` as well. 

In this case, however, it should be sufficient to add a Mutex around the `latest_handled_event_index` map, that would be acquired before the test and released after the `batch.write()`. While we process events sequentially, there is no lock contention and the Mutex would be very cheap. If we get to the point of parallel updates, then the method would continue to be correct even if called with the same `event_index` twice.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -3,17 +3,247 @@
...
+            if matches!(
+                blob_info.initial_certified_epoch(),
+                Some(initial_certified_epoch) if initial_certified_epoch < self.before_epoch
+            ) {
+                return item;
+            }
+            item = self.iter.next();
+        }
+        item
+    }
```
jpcsmith: Nit: The following makes it clear that errors are immediately returned and the condition on which items are returned.
```suggestion
        while let Some(item) = self.iter.next() {
            let Ok((blob_id, blob_info)) = item else {
                return Some(item);
            };

            if !self.only_certified
                || matches!(
                    blob_info.initial_certified_epoch(),
                    Some(initial_certified_epoch) if initial_certified_epoch < self.before_epoch
                )
            {
                return Some(Ok((blob_id, blob_info)));
            }
        }
        None
```

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -3,17 +3,247 @@
...
+}
+
+pub type BlobInfoIterator<'a> =
+    BlobInfoIter<dyn Iterator<Item = Result<(BlobId, BlobInfo), TypedStoreError>> + Send + 'a>;
```
jpcsmith, line 239: Question: why not just name the iterator type?

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: -> #904

crates/walrus-service/src/node.rs
jpcsmith: Remark: It does store the count. The addition of the next-event index to `get_event_cursor` is a source of confusion. Would renaming `get_event_cursor` help, `get_event_cursor_and_next_index`?

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: OK, I will add this to #904.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: No, they don't have to be sequential. In fact, they are not because we now have many events that are not blob events and do not cause an update to the blob info.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: I think @halfprice looked into this and the actual type seems to be hidden in the RocksDB crate.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: > In this case, however, it should be sufficient to add a Mutex around the latest_handled_event_index map, that would be acquired before the test and released after the batch.write(). While we process events sequentially, there is no lock contention and the Mutex would be very cheap. If we get to the point of parallel updates, then the method would continue to be correct even if called with the same event_index twice.

Sounds good.

crates/walrus-service/src/node/storage/blob_info.rs
jpcsmith: Fair point, thank you for the clarification.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Implemented in 91252bad6af1f01f1e70dd652496c3039394b1a9. I used a `std::sync::Mutex` as we never hold it across `await` points.

crates/walrus-service/src/node.rs
mlegner: I renamed the functions in dd24a3e352bb24fd2432d41d29fe2b0d2d1d2f23 and also swapped the order of cursor and index/progress/count to match the function names and descriptions.

This is a breaking change (also over the previous state of this PR), but I guess that is fine at this point.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Thanks for the suggestion. I applied a minor modification of it.

crates/walrus-service/src/node/storage/event_cursor_table.rs
```diff
@@ -65,22 +65,23 @@ impl EventCursorTable {
...
 
-    /// Get the event cursor for `event_type`
-    pub fn get_event_cursor(&self) -> Result<Option<(u64, EventID)>, TypedStoreError> {
-        let entry = self.inner.get(&CURSOR_KEY)?;
-        Ok(entry)
+    /// Returns the current event cursor and the next event index.
+    pub fn get_event_cursor_and_next_index(
+        &self,
+    ) -> Result<Option<(EventID, u64)>, TypedStoreError> {
+        self.inner.get(&CURSOR_KEY)
```
jpcsmith, line 75: Nit: This could return `(Option<EventId>, u64)`.

Issue #835, opened 2024-09-20
Implement new `FixedPoint64` type and use in apportionment algorithm


tnowacki:
Followup to #619

Currently, apportionment algorithm can support a total_stake of roughly `(2^32 - 1) * n_shards`. Assuming 1000 shards, this gives a limit of about 4 trillion. 

Depending in what unit `total_stake` is expressed, this may not be enough and will need to be fixed by implementing a 64 bit fixed point library.

PR #902, opened 2024-10-01
fix(move): downscaling for the total stake in apportionment

tnowacki:
- Scaled down total stake until it is less than the max. Scales all stake amounts by that factor.
- Added more apportionment tests

Contributes to #835

### DIFF
--- a/contracts/walrus/sources/staking/staking_inner.move
+++ b/contracts/walrus/sources/staking/staking_inner.move
@@ -428,10 +428,8 @@ fun dhondt(
     if (n_nodes == 0) return (price, vector[]);
     let mut shards = stake.map_ref!(|s| u64_div(*s, price));
     let mut n_shards_distributed = shards.fold!(0, |acc, x| acc + x);
-    let mut i = 0;
     // loop until all shards are distributed
     while (n_shards_distributed != n_shards) {
-        i = i + 1;
         n_shards_distributed = if (n_shards_distributed < n_shards) {
                 // We decrease the price slightly such that some nodes get an additional shard.
                 price = from_raw(0);


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -408,6 +411,12 @@ fun dhondt(
     };
 
     let total_stake = stake.fold!(0, |acc, x| acc + x);
+
+    // TODO remove this when the FixedPoint32 has a replacement
+    let scaling = DHONDT_TOTAL_STAKE_MAX.max(total_stake) / DHONDT_TOTAL_STAKE_MAX;
+    let total_stake = total_stake / scaling;
```
mlegner: Important: I don't think this works due to the integer division: Say `total_stake == 2* DHONDT_TOTAL_STAKE_MAX - 1`, then `scaling == 1`, which means that no scaling is applied. The following should do the correct upward rounding:
```suggestion
    let scaling = (DHONDT_TOTAL_STAKE_MAX.max(total_stake) + DHONDT_TOTAL_STAKE_MAX - 1) / DHONDT_TOTAL_STAKE_MAX;
    let total_stake = total_stake / scaling;
```

tnowacki: I'll use `divide_and_round_up` to avoid any overflow issues, but good callout!

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -417,8 +428,10 @@ fun dhondt(
     if (n_nodes == 0) return (price, vector[]);
     let mut shards = stake.map_ref!(|s| u64_div(*s, price));
     let mut n_shards_distributed = shards.fold!(0, |acc, x| acc + x);
+    let mut i = 0;
```
karlwuest: Remark: This doesn't seem to be used anywhere and is shadowed by other variables `i` within scopes in the loop.

contracts/walrus/tests/staking/staking_inner_tests.move
```diff
@@ -260,3 +260,75 @@ macro fun sum<$T>($v: vector<$T>): $T {
...
+#[random_test]
+fun test_larger_dhondt_inputs_100_nodes_random_stake(seed: vector<u8>) {
+    random_dhondt_inputs(seed, 100, 10_000_000_000_000_000_000);
+}
+
+// TODO fix dhondt efficiency
+// #[random_test]
+// fun test_larger_dhondt_inputs_1000_nodes_random_stake(seed: vector<u8>) {
+//     random_dhondt_inputs(seed, 1_000, 10_000_000_000_000_000_000);
+// }
```
tnowacki, line 318: @karlwuest, we are not able to handle cases such as this one. 
What is essentially happening is that for certain distributions of stake, the initial `price` guess is _really_ far off. Causing the main loop to iterate for about ~450 times. 

I could try to dream up solutions to this, but I believe you had mentioned better algorithms?

contracts/walrus/sources/staking/staking_inner.move
mlegner: Ah, good point about the overflow! 🙇

contracts/walrus/tests/staking/staking_inner_tests.move
mlegner: I guess a possible approach could be to try to make bigger steps if we're very far off. But I haven't looked into this.

karlwuest: Currently, it seems that the core problem is that we  have to iterate over some vectors, e.g. the stake vector, within the loop, giving us a quadratic worst-case complexity. The reason why we need to do this is that we can both add and remove shards from the assignment. We can instead get an n*log(n) worst-case complexity if we switch to an initial guess that underestimates the number of shards and then assigns one additional shard per iteration, because then we only have to update one quotient per iteration instead of all of them and can use a heap to keep track of them.

I just did a quick basic implementation of what I had in mind and added it to a new branch (#926 ), I will clean this up tomorrow and add tie-breaking. 

This already looks promising:
```
Test Statistics:

┌────────────────────────────────────────────────────────────────────────────────┬────────────┬───────────────────────────┐
│                                   Test Name                                    │    Time    │         Gas Used          │
├────────────────────────────────────────────────────────────────────────────────┼────────────┼───────────────────────────┤
│ walrus::staking_inner_tests::test_larger_dhondt_inputs_1000_nodes_fixed_stake  │   0.359    │          277718           │
├────────────────────────────────────────────────────────────────────────────────┼────────────┼───────────────────────────┤
│ walrus::staking_inner_tests::test_larger_dhondt_inputs_1000_nodes_random_stake │   7.046    │          6766226          │
├────────────────────────────────────────────────────────────────────────────────┼────────────┼───────────────────────────┤
│ walrus::staking_inner_tests::test_larger_dhondt_inputs_100_nodes_fixed_stake   │   0.158    │            662            │
├────────────────────────────────────────────────────────────────────────────────┼────────────┼───────────────────────────┤
│ walrus::staking_inner_tests::test_larger_dhondt_inputs_100_nodes_random_stake  │   2.060    │           56747           │
└────────────────────────────────────────────────────────────────────────────────┴────────────┴───────────────────────────┘

```

One drawback is that it will not work as is if we introduce a `FixedPoint64` since the priority queue requires a `u64`. However, an additional advantage is that the algorithm is much simpler than what we currently have.

karlwuest: I think for the tie-breaking, we could potentially further optimize this by using a custom heap that already takes the ranking into account.

karlwuest: And it looks like the setup takes a majority of the time of the test:
```
│ walrus::staking_inner_tests::test_larger_dhondt_setup_1000_nodes_random_stake  │   5.526    │          4119309          │
```

contracts/walrus/sources/staking/staking_inner.move
tnowacki: Sorry! left over from debugging :)

Issue #811, opened 2024-09-16
Client: Split pre-existing storage resource if it is larger than necessary


mlegner:
#799 enabled the client to use existing owned `Storage` resources to register a new blob and selected the best-fitting one for this purpose. However, we can go further and first split the resource using the [`split_by_epoch` and `split_by_size` functions](<https://github.com/MystenLabs/walrus/blob/1653b1abe2829b6dbab0327d3d2ea5c7302e6321/contracts/walrus/sources/system/storage_resource.move#L49-L78>) to avoid wastage.

kayvank:
@mlegner, if this issue is still required, I would like to start working on it, please.

PR #903, opened 2024-10-02
feat(client): correct cost computation for new blobs

### LINKED ISSUE #820: Correct price outputs for new contracts
- [ ] Use WAL (and whatever the smaller unit of WAL is) instead of SUI/MIST
- [ ] Account for write fee in addition to storage fee
- [ ] Correctly handle different scenarios; e.g., existing `Storage` resource, existing registered but not certified `Blob` object.

giac-mysten:
* Introduces the `ResourceOperation` enum, which informs what operation has been done with storage and blob objects on sui, which is then used to compute the final operation cost f4048925312fd0c3f284ac668617535dffcdb21b
* Implements the correct display of cost with `WAL` values 608e64dd6615aac40ef603872e8b4d73eb581401
* Refactors the operations for `Blob` and `Storage` resource management into a `ResourceManager` struct. This separation isn't very substantial at the moment, but it is in preparation of having more complex resource operations (splitting, merging, extending, etc). c7dacd3a7f7f2a34409f078f46b86b2e0b1fbc1d
* Fixes a bug in storing blobs, whereby a deletable and already certified blob with the same expiration could be picked for certification, and would fail because already certified. 5860582ed5a84b4664c1b176325e5fe667544a31

Closes #820

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -453,3 +453,60 @@ async fn test_storage_nodes_delete_data_for_deleted_blobs() -> TestResult {
 
     Ok(())
 }
+
+/// Tests that storing the same blob multiple times with possibly different end epochs,
+/// persistence, and force-store conditions always works.
+#[ignore = "ignore E2E tests by default"]
+#[tokio::test]
+async fn test_multiple_stores_same_blob() -> TestResult {
+    let _ = tracing_subscriber::fmt::try_init();
+    let (_sui_cluster_handle, _cluster, client) = test_cluster::default_setup().await?;
+    let client = client.as_ref();
+    let blob = walrus_test_utils::random_data(314);
+
+    // NOTE: not in a param_test, because we want to test these store operations in sequence.
+    // If the last `bool` parameter is `true`, the store operation should return a
+    // `BlobStoreResult::AlreadyCertified`. Otherwise, it should return a
+    // `BlobStoreResult::NewlyCreated`.
+    let configurations = vec![
+        (1, StoreWhen::NotStored, BlobPersistence::Deletable, false),
+        (1, StoreWhen::Always, BlobPersistence::Deletable, false),
+        (2, StoreWhen::NotStored, BlobPersistence::Deletable, false),
+        (3, StoreWhen::Always, BlobPersistence::Deletable, false),
+        (1, StoreWhen::NotStored, BlobPersistence::Permanent, false),
+        (1, StoreWhen::NotStored, BlobPersistence::Permanent, true),
+        (1, StoreWhen::Always, BlobPersistence::Permanent, false),
+        (4, StoreWhen::NotStored, BlobPersistence::Permanent, false),
+        (2, StoreWhen::NotStored, BlobPersistence::Permanent, true),
+        (2, StoreWhen::Always, BlobPersistence::Permanent, false),
+        (1, StoreWhen::NotStored, BlobPersistence::Deletable, true),
+        (5, StoreWhen::NotStored, BlobPersistence::Deletable, false),
+    ];
+
+    for (epochs, store_when, persistence, is_already_certified) in configurations {
+        let result = client
+            .reserve_and_store_blob(&blob, epochs, store_when, persistence)
+            .await?;
+
+        println!(
+            "epochs: {}, store_when: {:?}, persistence: {:?}, is_already_certified: {}",
+            epochs, store_when, persistence, is_already_certified
+        );
+        match result {
+            BlobStoreResult::NewlyCreated { .. } => {
+                assert!(!is_already_certified, "the blob should be newly stored");
+            }
+            BlobStoreResult::AlreadyCertified { .. } => {
+                assert!(is_already_certified, "the blob should be already stored");
+            }
+            _ => panic!("we either store the blob, or find it's already created"),
+        }
+    }
+
+    // At the end of all the operations above, count the number of blob objects owned by the
+    // client.
+    let blobs = client.sui_client().owned_blobs(false).await?;
+    assert_eq!(blobs.len(), 9);
+
+    Ok(())
+}
--- a/crates/walrus-service/src/client/resource.rs
+++ b/crates/walrus-service/src/client/resource.rs
@@ -253,3 +253,44 @@ impl<'a, C: ContractClient> ResourceManager<'a, C> {
             }))
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use walrus_test_utils::param_test;
+
+    use super::*;
+
+    param_test! {
+        test_price_computation: [
+            one_epoch: (1024, 1, (1, 1), (2, 1)),
+            two_epochs: (1024, 2, (1, 1), (3, 1)),
+            higher_write: (1024, 1, (1, 2), (3, 2)),
+            larger_blob: (2048, 1, (1, 2), (6, 4)),
+            even_larger_blob: (2049, 1, (1, 2), (9, 6)),
+            more_epochs: (2049, 2, (1, 2), (12, 6)),
+        ]
+    }
+    fn test_price_computation(
+        encoded_length: u64,
+        epochs_ahead: u32,
+        storage_and_write_prices: (u64, u64),
+        scratch_and_reuse_costs: (u64, u64),
+    ) {
+        let (storage_price, write_price) = storage_and_write_prices;
+        let computation = PriceComputation::new(storage_price, write_price);
+        let scratch = RegisterBlobOp::RegisterFromScratch {
+            encoded_length,
+            epochs_ahead,
+        };
+        let storage = RegisterBlobOp::ReuseStorage { encoded_length };
+        let registration = RegisterBlobOp::ReuseRegistration { encoded_length };
+
+        let (expected_cost_scratch, expected_cost_reuse_storage) = scratch_and_reuse_costs;
+        assert_eq!(computation.operation_cost(&scratch), expected_cost_scratch);
+        assert_eq!(
+            computation.operation_cost(&storage),
+            expected_cost_reuse_storage
+        );
+        assert_eq!(computation.operation_cost(&registration), 0);
+    }
+}


crates/walrus-service/src/client/cli.rs
```diff
@@ -249,32 +249,94 @@ impl std::fmt::Display for HumanReadableBytes {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub(crate) struct GeorgieCoin(u64);
+
+/// The human readable representation of the GEORGIE and WAL coins.
+pub(crate) type HumanReadableGeorgie = HumanReadableCoin<GeorgieCoin>;
+
+impl CurrencyForDisplay for GeorgieCoin {
+    const SUPERUNIT_NAME: &'static str = "WAL";
+    const UNIT_NAME: &'static str = "GEORGIE";
+    const SUBUNIT_IN_UNIT: u64 = 3_141_592;
```
giac-mysten: Important: help with naming and value conversion needed! You can start thinking about it while I finish up the rest of the PR @mlegner @jpcsmith @karlwuest @sadhansood @halfprice

crates/walrus-service/src/client/cli.rs
```diff
@@ -249,32 +249,94 @@ impl std::fmt::Display for HumanReadableBytes {
...
-///
-/// 1. If the price is below 1_000_000 MIST, it is printed fully, with thousands separators.
-/// 2. Else, it is printed in SUI with 3 decimal places.
+trait CurrencyForDisplay {
+    /// The name of the main unit of currency
+    const SUPERUNIT_NAME: &'static str;
+    /// The name of the subunit, that divides the main currency.
+    const UNIT_NAME: &'static str;
+    /// The number of subunits in a unit.
+    const SUBUNIT_IN_UNIT: u64;
```
mlegner: I don't think we need quite this generality given that currencies on Sui always have a power of 10 as a factor here, see the [`sui::coin` module](https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/packages/sui-framework/sources/coin.move#L53). That only stores a `decimals: u8` as a relationship, which we could mirror here.

Theoretically, we could even read this from chain, but that may be a bit overkill given that we only handle two coin types.

crates/walrus-service/src/client/cli.rs
```diff
@@ -249,32 +249,94 @@ impl std::fmt::Display for HumanReadableBytes {
...
+    }
+
+    /// Gets the value of the current coin.
+    fn value(&self) -> u64;
+
+    // Creates a new instance of the currency.
+    fn new(value: u64) -> Self;
+}
+
+/// The rerepresentation of a currency in a human readable format.
```
mlegner: Nit:
```suggestion
/// The representation of a currency in a human readable format.
```

crates/walrus-service/src/client/cli.rs
mlegner: For WAL, we will almost certainly have 9 decimals as for SUI.

I would probably name the coins according to their main currency, i.e., `SuiCoin` and `WalCoin`. Regarding the smaller unit of WAL, I don't think we've reached a conclusion, but `GEORGIE` works as a placeholder.

crates/walrus-service/src/client/cli.rs
giac-mysten: mmm I see. I did this because there was talk of having `21 M GEORGIE = 1 WAL` -- idk if it was a joke or not -- in which case we do need the generality.

crates/walrus-service/src/client/cli.rs
giac-mysten: >I would probably name the coins according to their main currency, i.e., SuiCoin and WalCoin.

this was my initial thinking as well, but then it felt weird the the `UNIT_NAME` for `SuiCoin` is `MIST`... anyways, will revert.

crates/walrus-service/src/client/cli.rs
mlegner: Joke or not, I believe the 21M were suggested for the total supply of WAL.

crates/walrus-service/src/client/cli.rs
mlegner: We could also have `UNIT_NAME` and `SUB_UNIT_NAME` to make it more consistent. But I also think the current naming is fine.

crates/walrus-service/src/client/cli.rs
karlwuest: >Joke or not, I believe the 21M were suggested for the total supply of WAL.

Yes, I think so. 21M is the total (maximum) supply of Bitcoin, which is where this is coming from.

>mmm I see. I did this because there was talk of having 21 M GEORGIE = 1 WAL -- idk if it was a joke or not -- in which case we do need the generality.

We are using the `Coin` type, so we anyway can only do decimals.

giac-mysten: ah perfect, will change to decimals then.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -164,9 +164,15 @@ async fn test_inconsistency(failed_shards: &[usize]) -> TestResult {
         SliverPairIndex::new(1).to_shard_index(NonZeroU16::new(13).unwrap(), &blob_id)
     );
     // Register blob.
-    let blob_sui_object = client
+    let (blob_sui_object, _op) = client
```
mlegner: Nit: If the operation does not serve any "guard" functionality, you can also just use `_` instead of a name. (Also in other locations.)

crates/walrus-service/src/client/responses.rs
```diff
@@ -59,10 +75,12 @@ pub enum BlobStoreResult {
...
-        /// The encoded size, including metadata.
-        encoded_size: u64,
+        /// The operation that created the blob.
+        resource_operation: RegisterBlobOp,
         /// The storage cost, excluding gas.
         cost: u64,
+        /// The encoded size, including metadata.
+        encoded_size: u64,
         /// Whether the blob is deletable.
         deletable: bool,
```
mlegner: Nit: Not related to your changes, but I recently noticed the redundancy that the `deletable` is also in the `Blob` object.

crates/walrus-service/src/client/cli.rs
```diff
@@ -249,32 +249,98 @@ impl std::fmt::Display for HumanReadableBytes {
...
-/// follows:
-///
-/// 1. If the price is below 1_000_000 MIST, it is printed fully, with thousands separators.
-/// 2. Else, it is printed in SUI with 3 decimal places.
+trait CurrencyForDisplay {
+    /// The name of the main unit of currency
+    const SUPERUNIT_NAME: &'static str;
+    /// The name of the subunit, that divides the main currency.
+    const UNIT_NAME: &'static str;
+    /// The number of subunits in a unit.
```
mlegner: Nit: This docstring is no longer applicable. Maybe something like
```suggestion
    /// Number of decimal places the coin uses: 1 superunit is equal to 10^decimal units.
```

crates/walrus-service/src/client/cli.rs
```diff
@@ -249,32 +249,98 @@ impl std::fmt::Display for HumanReadableBytes {
...
+    const DECIMALS: u8;
+
+    /// Converts a unit to subunits.
+    fn unit_to_superunit(value: u64) -> f64 {
+        value as f64 / Self::units_in_superunit() as f64
+    }
```
mlegner: The docstring, function name, and implementation are not really consistent.

crates/walrus-service/src/client/cli.rs
```diff
@@ -303,6 +369,18 @@ pub fn format_event_id(event_id: &EventID) -> String {
...
+pub fn certification_event_string(event_or_object: &EventOrObjectId) -> String {
+    match event_or_object {
+        EventOrObjectId::Event(event_id) => {
+            format!("Certification event ID: {}", format_event_id(event_id))
+        }
+        EventOrObjectId::Object(object_id) => {
+            format!("Owned Blob registration object ID: {}", object_id)
+        }
+    }
+}
```
mlegner: Nit: Can we instead implement `CliOutput` for `EventOrObjectId`?

crates/walrus-service/src/client/resource.rs
```diff
@@ -0,0 +1,276 @@
...
+use crate::client::responses::EventOrObjectId;
+
+#[derive(Debug, Clone)]
+pub(crate) struct PriceComputation {
+    write_price_per_unit_size: u64,
+    storage_price_per_unit_size: u64,
+    n_shards: NonZeroU16,
+}
+
+/// Struct to compute the cost of operations with blob and storage resources.
```
mlegner: Nit: I guess this should be at the struct definition.

crates/walrus-service/src/client/resource.rs
```diff
@@ -0,0 +1,276 @@
...
+        store_when: StoreWhen,
+    ) -> ClientResult<(Blob, StoreOp)> {
+        let (blob, op) = self
+            .get_existing_registration(metadata, epochs_ahead, persistence, store_when)
+            .await?;
+
+        // If the blob is deletable and already certified, return early.
+        let store_op = if blob.certified_epoch.is_some() {
+            debug_assert!(
+                blob.deletable && !store_when.is_store_always(),
```
mlegner, line 126: Question: Why do we know that the blob is deletable?

crates/walrus-service/src/client/resource.rs
```diff
@@ -0,0 +1,276 @@
...
+                })?,
+                epochs_ahead + self.write_committee_epoch,
+            )
+            .await?
+        {
+            tracing::debug!(
+                storage_object=%storage_resource.id,
+                "using an existing storage resource to register the blob"
+            );
+            let blob = self
```
mlegner, line 197: Suggestion: Can we add a TODO to split the storage resource if possible/necessary?

crates/walrus-service/src/client/resource.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// - First, it checks if the blob is registered and returns the corresponding [`Blob`];
+    /// - otherwise, it checks if there is an appropriate storage resource (with sufficient space
+    ///   and for a sufficient duration) that can be used to register the blob; or
+    /// - if the above fails, it purchases a new storage resource and registers the blob.
+    ///
+    /// If we are forcing a store ([`StoreWhen::Always`]), the function filters out already
+    /// certified blobs owned by the wallet, such that we always create a new certification
+    /// (possibly reusing storage resources or uncertified but registered blobs).
+    #[tracing::instrument(skip_all, err(level = Level::DEBUG))]
+    pub async fn get_existing_registration(
```
mlegner, line 157: Nit: I find the long `if, else if, else` expression a bit difficult to read. But I don't have a good way to improve this at hand...

crates/walrus-service/src/client/resource.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// registration that would be made now to write a new blob.
+    ///
+    /// If `filter_certified` is `true`, the function filters out already certified blobs owned by
+    /// the wallet.
+    async fn is_blob_registered_in_wallet(
+        &self,
+        blob_id: &BlobId,
+        epochs_ahead: EpochCount,
+        persistence: BlobPersistence,
+        filter_certified: bool,
```
mlegner: Nit: I'd find `include_certified` (so the opposite) more intuitive.

crates/walrus-service/src/client.rs
```diff
@@ -122,10 +125,11 @@ impl Client<()> {
                 .map_err(ClientError::other)?,
         ));
 
-        let storage_price_per_unit_size = sui_read_client
-            .storage_price_per_unit_size()
-            .await
-            .map_err(ClientError::other)?;
+        let price_computation = PriceComputation::new(
+            sui_read_client.write_price_per_unit_size().await?,
+            sui_read_client.storage_price_per_unit_size().await?,
```
mlegner: Question: It seems that we read data from chain, especially the system object, multiple times in separate requests. Is there any simple way to combine those? @karlwuest?

crates/walrus-service/src/client/resource.rs
```diff
@@ -0,0 +1,276 @@
...
+                .storage_fee_for_unencoded_length(*unencoded_length, *epochs_ahead)
+                .map(|storage_fee| {
+                    storage_fee + self.write_fee_for_unencoded_length(*unencoded_length)
+                }),
+            RegisterBlobOp::ReuseStorage { unencoded_length } => {
+                Some(self.write_fee_for_unencoded_length(*unencoded_length))
+            }
+            _ => Some(0), // No cost for reusing registration or no-op.
+        }
+    }
```
mlegner: Remark: If we had the `encoded_length` here, we wouldn't need to recompute it multiple times and also wouldn't need the `Option` result.

crates/walrus-e2e-tests/tests/test_client.rs
giac-mysten: that's fair, I just usually add it for readability (to remind what that is). But removed it.

crates/walrus-service/src/client/responses.rs
giac-mysten: good point, removed it!

crates/walrus-service/src/client/cli.rs
giac-mysten: changed the docstring.

crates/walrus-service/src/client/cli.rs
giac-mysten: uhm `CliOutput` requires printing directly to stout. What about `Display`?

crates/walrus-service/src/client/resource.rs
giac-mysten: If it was permanent and certified, and would be sufficient to register the blob, we would have returned after the status check. However, I agree having this here is not correct.

giac-mysten: I need to think whether it is better to move it back in the main function, where the statement is enforced, or how to enforce it here.

crates/walrus-service/src/client/resource.rs
giac-mysten: I think the storage management can be quite complicated. I'll add a todo, but the overhaul may be quite large.

crates/walrus-service/src/client/resource.rs
giac-mysten: Agreed. This may be solved when we improve the resource management end to end.

crates/walrus-service/src/client.rs
giac-mysten: good point. I added a function to the read client that returns both, querying the system object only once.

crates/walrus-service/src/client/resource.rs
giac-mysten: good point. Somehow I foresaw that the encoded length was actually already computed in the function and therefore it's easy to reuse in here. Changed.

crates/walrus-service/src/client/resource.rs
giac-mysten: i'll solve it in #911

crates/walrus-service/src/client/cli.rs
mlegner: Ah, indeed. Sorry, I forgot that. I think it's fine to leave it like this.

crates/walrus-service/src/client/resource.rs
```diff
@@ -3,86 +3,60 @@
...
-            n_shards,
+            write_price_per_unit_size,
         }
     }
 
     /// Computes the cost of the operation.
     ///
-    /// Returns `None` if `unencoded_length` is invalid for the current encoding and `n_shards`, and
+    /// Returns `None` if `encoded_length` is invalid for the current encoding and `n_shards`, and
     /// therefore the encoded length cannot be computed.
```
mlegner: Nit: The `None` case is no longer there.
```suggestion
```

crates/walrus-service/src/client/resource.rs
```diff
@@ -3,86 +3,60 @@
...
+                encoded_length,
                 epochs_ahead,
-            } => self
-                .storage_fee_for_unencoded_length(*unencoded_length, *epochs_ahead)
-                .map(|storage_fee| {
-                    storage_fee + self.write_fee_for_unencoded_length(*unencoded_length)
-                }),
-            RegisterBlobOp::ReuseStorage { unencoded_length } => {
-                Some(self.write_fee_for_unencoded_length(*unencoded_length))
+            } => self.storage_fee_for_encoded_length(*encoded_length, *epochs_ahead),
```
mlegner: Question/important: Don't we also need the write fee here?

crates/walrus-service/src/client/resource.rs
mlegner: There is also an open issue for this: #811

crates/walrus-service/src/client/resource.rs
giac-mysten: 🤦

PR #910, opened 2024-10-02
feat(staking): implements param voting

damirka:


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -143,13 +144,6 @@ public(package) fun new(
     }
 }
 
-// === Accessors ===
-
-/// Returns the next epoch parameters if set, otherwise aborts with an error.
```
damirka, line 148: Moved to the bottom of the file into `=== accessors ===` section.

contracts/walrus/tests/e2e_tests.move
```diff
@@ -5,6 +5,7 @@
 module walrus::e2e_tests;
```
damirka, line 5: Ideally requires extra tests

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -225,6 +216,49 @@ public(package) fun voting_end(self: &mut StakingInnerV1, clock: &Clock) {
...
+        write_prices.insert(pool.write_price(), weight);
+        storage_prices.insert(pool.storage_price(), weight);
+
+        // for node capacities we want to reverse the order, and pick the highest
+        // value that satisfies 2/3 of the weight
+        node_capacities.insert(pool.node_capacity(), weight);
+    });
+
+    // 2/3 of the weight is the minimum threshold for prices
+    let min_threshold = (self.n_shards * 2 / 3) as u64;
```
karlwuest: This is a bit more subtle, we cannot just use 2/3 (we want a quorum here).
```suggestion
    let min_threshold = self.n_shards - (self.n_shards - 1) / 3 as u64;
```

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -225,6 +216,49 @@ public(package) fun voting_end(self: &mut StakingInnerV1, clock: &Clock) {
...
+    let max_threshold = (self.n_shards as u64) - min_threshold;
+
+    epoch_parameters::new(
+        max_threshold(&mut node_capacities, max_threshold),
```
karlwuest: Here, we cannot just apply the same function with different thresholds, because the function uses `>=` to compare to the threshold, which means that  in the `max_threshold` case, we actually don't necessarily have a quorum that votes for a value less than the threshold.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -225,6 +216,59 @@ public(package) fun voting_end(self: &mut StakingInnerV1, clock: &Clock) {
...
+    let mut capacity_votes = priority_queue::new(vector[]);
+
+    size.do!(|i| {
+        let (node_id, shards) = inner.get_entry_by_idx(i);
+        let pool = &self.pools[*node_id];
+        let weight = shards.length();
+        write_prices.insert(pool.write_price(), weight);
+        storage_prices.insert(pool.storage_price(), weight);
+        // The vote for capacity is determined by the node capacity and number of assigned shards.
+        let capacity_vote = pool.node_capacity() / weight * (self.n_shards as u64);
```
mlegner, line 237: Important: The way we do this currently means that up to 1/3 of the nodes (by weight) have a lower capacity than they need. I believe we can only resolve this by a different approach for the system capacity, but this is something we will have to look into again.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -225,6 +216,59 @@ public(package) fun voting_end(self: &mut StakingInnerV1, clock: &Clock) {
...
+    });
+
+    epoch_parameters::new(
+        quorum_above(&mut capacity_votes, self.n_shards),
+        quorum_below(&mut storage_prices, self.n_shards),
+        quorum_below(&mut write_prices, self.n_shards),
+    )
+}
+
+/// Take the highest value, s.t. a quorum voted for a value larger or equal to this.
```
mlegner: Question: A quorum here is `2f+1`, right?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -225,6 +216,59 @@ public(package) fun voting_end(self: &mut StakingInnerV1, clock: &Clock) {
...
+fun take_threshold_value(vote_queue: &mut PriorityQueue<u64>, threshold: u64): u64 {
+    let mut sum_weight = 0;
+    loop {
+        let (value, weight) = vote_queue.pop_max();
+        sum_weight = sum_weight + weight;
+        if (sum_weight >= threshold) {
+            return value
+        };
+    }
+}
```
mlegner: Nit: Can we rename the threshold to `threshold_weight` and add a comment explaining that the loop will always succeed if the threshold weight is smaller than the total weight?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -652,3 +710,57 @@ public(package) fun pub_dhondt(n_shards: u16, stake: vector<u64>): (FixedPoint32
...
     dhondt(ranking, n_shards, stake)
 }
+
+#[test]
+fun test_quorum_above() {
+    let mut queue = priority_queue::new(vector[]);
+    let votes = vector[1, 2, 3, 4, 5, 6, 7, 8, 9, 10];
+    let weights = vector[5, 5, 4, 6, 3, 7, 2, 8, 1, 9];
+    votes.zip_do!(weights, |vote, weight|
+    queue.insert(vote, weight));
```
mlegner: Nit: I guess this should be indented (similar cases below).

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -225,6 +216,59 @@ public(package) fun voting_end(self: &mut StakingInnerV1, clock: &Clock) {
...
+        storage_prices.insert(pool.storage_price(), weight);
+        // The vote for capacity is determined by the node capacity and number of assigned shards.
+        let capacity_vote = pool.node_capacity() / weight * (self.n_shards as u64);
+        capacity_votes.insert(capacity_vote, weight);
+    });
+
+    epoch_parameters::new(
+        quorum_above(&mut capacity_votes, self.n_shards),
+        quorum_below(&mut storage_prices, self.n_shards),
+        quorum_below(&mut write_prices, self.n_shards),
```
mlegner, line 244: Question: How did we decide whether to take the value at 1/3 or 2/3?

contracts/walrus/sources/staking/staking_inner.move
karlwuest: Yes

contracts/walrus/sources/staking/staking_inner.move
karlwuest: We choose the values, s.t. a quorum is "happy" with the value. This means that for the prices, the chosen value has to be at least as high as the vote for a quorum, and for the capacity, the value must be at most as high as the vote for a quorum.

contracts/walrus/sources/staking/staking_inner.move
karlwuest: I think we cannot resolve this unless we go back to the idea of assigning fewer shards to nodes that do not have enough capacity, because 1/3 can be byzantine. I'm happy to revisit this discussion, but from our last discussion in Portugal, I'm not sure if there is a truly satisfying solution to this. 

The current implementation does what we concluded that discussion with, namely scaling down the capacity vote by the number of assigned shards, since the node does not know a priori how many shards it will receive.

It is important to note here, that these values are all votes, i.e. the value does not have to literally be the capacity of the node. Would it help if we add a `_vote` at the end of the function/variable names?

Is there anything actionable that you see for this PR or do you want to open an issue to revisit the discussion on voting for the system capacity?

mlegner: > Is there anything actionable that you see for this PR or do you want to open an issue to revisit the discussion on voting for the system capacity?

Sorry, I wasn't clear in my comment. I agree that what you implemented is what we discussed. My comment was more that we should revisit this discussion some time before Mainnet.

mlegner: I guess we could tell the node providers to vote for some percent less than their actual capacity.

karlwuest: Added #937 to keep track of this.

Issue #819, opened 2024-09-17
Client: ensure that registering an already certified blobs for more epochs does not incur in storing slivers


giac-mysten:
Currently, calling the `store` command with `--epochs` may incur in storing the slivers again, even if the blob is already stored.

PR #911, opened 2024-10-02
feat(client): check if slivers are already available

giac-mysten:
Ensures that the client sends slivers only if there are no certified blobs on Walrus with the same ID.

To this end:
* Now storing blobs always incurs a blob status check to see if the blob is already registered (slivers available).
* Implements a standalone request to gather certificates without sending slivers.
* Cleans up further the `reserve_and_store_blob` function

Also fixes two todos in `get_blob_status_with_retries`.

Closes: #819

### DIFF
--- a/crates/walrus-service/src/client/resource.rs
+++ b/crates/walrus-service/src/client/resource.rs
@@ -114,10 +114,11 @@ impl<'a, C: ContractClient> ResourceManager<'a, C> {
         }
     }
 
-    /// Returns the appropriate storage operation for the given blob to be stored.
+    /// Returns the appropriate store operation for the given blob.
     ///
-    /// The function considers the requirements for the store operation (epcoh ahead, persistence,
-    /// force store), the status of the blob on chain, and the available resources in the wallet.
+    /// The function considers the requirements given to the store operation (epochs ahead,
+    /// persistence, force store), the status of the blob on chain, and the available resources in
+    /// the wallet.
     pub async fn store_operation_for_blob(
         &self,
         metadata: &VerifiedBlobMetadataWithId,


crates/walrus-service/src/client/resource.rs
```diff
@@ -109,13 +114,30 @@ impl<'a, C: ContractClient> ResourceManager<'a, C> {
...
+    /// the wallet.
+    pub async fn store_operation_for_blob(
         &self,
         metadata: &VerifiedBlobMetadataWithId,
         epochs_ahead: EpochCount,
         persistence: BlobPersistence,
         store_when: StoreWhen,
-    ) -> ClientResult<(Blob, StoreOp)> {
+        blob_status: BlobStatus,
+    ) -> ClientResult<StoreOp> {
```
giac-mysten, line 129: I'm thinking of adding a `ResourceManagerFactory`, and add `metadata, epochs_ahead, persistence, store_when` to the `ResourceManager` struct. This is to make the `ResourceManager` specific to one store operation for a blob.

Then, one would call `ResourceManagerFactory::new(sui_client, write_epoch).build_for_blob(metadata, epochs_ahead, persistence, store_when).store_operation_for_blob(blob_status)`

Too indirect? Thoughts?

mlegner: What would be the main benefit? Being able to remove all the parameters to the functions?

No strong opinions. Conceptually I like it, but given that we only have 3 functions, I don't think it's necessary.

giac-mysten: maybe that in the future, if the computation becomes more complicate, it's easier to pass arguments. On the other hand, YAGNI :)

PR #913, opened 2024-10-02
chore: update sui dependency to testnet-v1.34.2

sadhansood:
Update the dependency to include a flx which disables fullnode pruning in tests.

### DIFF
--- a/contracts/wal_exchange/Move.lock
+++ b/contracts/wal_exchange/Move.lock
@@ -2,7 +2,7 @@
 
 [move]
 version = 2
-manifest_digest = "5AD346CF3FF24168065A7AD2D3669725B79DF80B2547A38E5901055B24D44D17"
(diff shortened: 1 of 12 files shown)


contracts/wal_exchange/Move.lock
```diff
@@ -30,6 +30,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.34.2"
+compiler-version = "1.34.0"
```
halfprice, line 33: should this be 1.34.2?

sadhansood: It's an automatic change, even if i manual revert it - `sui move build` just changes it to 1.34.2

mlegner: Remark: This is just the version of `sui` you have installed.

PR #915, opened 2024-10-03
feat: add logic to end voting for an epoch

jpcsmith:
This adds the `EpochChangeDriver` sub-service, which schedules and calls the contract to end the voting for the parameters for the next epoch.

### DIFF
--- a/crates/walrus-service/src/node/epoch_change_driver.rs
+++ b/crates/walrus-service/src/node/epoch_change_driver.rs
@@ -1,0 +1,550 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Drive voting end and epoch change.
+
+use std::{
+    fmt::Formatter,
+    future::Future,
+    num::NonZero,
+    pin::Pin,
+    sync::{Arc, Mutex},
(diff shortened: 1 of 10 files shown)


crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -0,0 +1,550 @@
...
+
+/// Return a random duration uniformly sampled from the interval `[0..max)`.
+fn jitter(max: Duration, rng: &mut StdRng) -> Duration {
+    rng.gen_range(Duration::ZERO..max)
+}
+
+/// Returns the duration of voting for epoch parameters as function of the epoch duration
+pub fn default_voting_duration(epoch_duration: Duration) -> Duration {
+    epoch_duration / 2
+}
```
karlwuest, line 344: Nit: Since this is part of the system parameters, it may be more fitting to turn this into a method `voting_duration` on `FixedSystemParameters`.

jpcsmith: I had the same thought, but then I would need to get a `FixedSystemParameters` to be able to call it, or have it as a method in addition to a free-function. I therefore opted to just add the free function for now since it was easier to use in the tests.

PR #918, opened 2024-10-03
feat: walrus CLI command to register storage node and add stake with the node's pool

### LINKED ISSUE #876: Provide CLI endpoint to register as candidate
Add an CLI endpoint for storage nodes to register themselves as a candidate to the node's binary.

halfprice:
This PR created two commands for the node operator to start storage node.

*register* that registers the storage node as committee candidate

usage:

```
> walrus-node register --config-path dryrun-node-1.yaml
Successfully registered storage node with capability:
      Capability object ID: 0xc8c932a3a8421796dbf003127a9e335a91349c844272cb52d67edf5456226f3b
      Node ID: 0x2a1f1e7100290187b369c1598462abcc8a0f67a0737979b6a05fcf5c1e20ff7f
```

*stake* which allows an account to stake WAL with a storage node

usage:

```
> walrus stake --node-id 0x2a1f1e7100290187b369c1598462abcc8a0f67a0737979b6a05fcf5c1e20ff7f
Success: Staked WAL successfully.
Staking info:
StakedWal:
    object id: 0x38238269b5ced80033c197c05efc9799e8d006592258d3af9658a0a44cb53039,
    state: Staked,
    node_id: 0x2a1f1e7100290187b369c1598462abcc8a0f67a0737979b6a05fcf5c1e20ff7f,
    principal: 1000000000,
    activation_epoch: 1
```

Both commands use the sui wallet included in the input StorageNodeConfig, since that will be wallet the node later use to interact with the contract.

close #876

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -74,6 +74,10 @@ enum Commands {
         #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
         /// The path to the node's configuration file.
         config_path: PathBuf,
+        #[clap(short, long)]
+        #[serde(default)]
+        /// The name of the node.
+        name: Option<String>,
     },
 }
 
@@ -88,13 +92,15 @@ fn main() -> anyhow::Result<()> {
 
         Commands::KeyGen { out } => commands::keygen(&out)?,
 
-        Commands::Register { config_path } => commands::register_node(config_path)?,
+        Commands::Register { config_path, name } => commands::register_node(config_path, name)?,
     }
     Ok(())
 }
 
 mod commands {
     use std::io;
+
+    use anyhow::anyhow;
 
     use super::*;
 
@@ -178,14 +184,28 @@ mod commands {
     }
 
     #[tokio::main]
-    pub(crate) async fn register_node(config_path: PathBuf) -> anyhow::Result<()> {
+    pub(crate) async fn register_node(
+        config_path: PathBuf,
+        name: Option<String>,
+    ) -> anyhow::Result<()> {
         let storage_config = StorageNodeConfig::load(&config_path)?;
+        let node_name = name.or(storage_config.name.clone()).ok_or(anyhow!(
+            "Name is required to register node. Set it in the config file or provided in the \
+                commandline argument."
+        ))?;
+        let registration_params = storage_config.to_registration_params(node_name);
 
         // Uses the Sui wallet configuration in the storage node config to register the node.
         let contract_client = get_contract_client_from_node_config(&storage_config).await?;
+        let proof_of_possession = walrus_sui::utils::generate_proof_of_possession(
+            storage_config.protocol_key_pair(),
+            &contract_client,
+            &registration_params,
+        )
+        .await?;
 
         let node_capability = contract_client
-            .register_candidate(&storage_config.into())
+            .register_candidate(&registration_params, &proof_of_possession)
             .await?;
 
         println!("Successfully registered storage node with capability:",);
--- a/crates/walrus-service/src/client/cli.rs
+++ b/crates/walrus-service/src/client/cli.rs
@@ -20,7 +20,6 @@ use walrus_core::BlobId;
 use walrus_sui::client::{SuiContractClient, SuiReadClient};
 
 use super::{default_configuration_paths, Blocklist, Client, Config};
-use crate::utils as ServiceUtils;
 
 mod args;
 mod cli_output;
@@ -43,7 +42,7 @@ pub const DEFAULT_RPC_URL: &str = TESTNET_RPC;
 // NB: When making changes to the logic, make sure to update the argument docs in
 // `crates/walrus-service/bin/client.rs`.
 pub fn load_configuration(path: &Option<PathBuf>) -> Result<Config> {
-    let path = ServiceUtils::path_or_defaults_if_exist(path, &default_configuration_paths())
+    let path = crate::utils::path_or_defaults_if_exist(path, &default_configuration_paths())
         .ok_or(anyhow!("could not find a valid Walrus configuration file"))?;
     tracing::info!("using Walrus configuration from '{}'", path.display());
 
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -265,9 +265,9 @@ pub enum CliCommands {
         #[clap(long)]
         /// The object ID of the storage node to stake with.
         node_id: ObjectID,
-        #[clap(short, long, default_value_t = default::staking_amount_wal())]
-        #[serde(default = "default::staking_amount_wal")]
-        /// The amount of WAL to stake with the storage node.
+        #[clap(short, long, default_value_t = default::staking_amount_frost())]
+        #[serde(default = "default::staking_amount_frost")]
+        /// The amount of FROST (smallest unit of WAL token) to stake with the storage node.
         amount: u64,
     },
 }
@@ -534,7 +534,7 @@ mod default {
             .expect("this is a correct socket address")
     }
 
-    pub(crate) fn staking_amount_wal() -> u64 {
+    pub(crate) fn staking_amount_frost() -> u64 {
         1_000_000_000
     }
 }
--- a/crates/walrus-service/src/client/cli/runner.rs
+++ b/crates/walrus-service/src/client/cli/runner.rs
@@ -55,7 +55,7 @@ use crate::{
         Config,
         StoreWhen,
     },
-    utils::{self as ServiceUtils, MetricsAndLoggingRuntime},
+    utils::MetricsAndLoggingRuntime,
 };
 
 /// A helper struct to run commands for the Walrus client.
@@ -86,7 +86,7 @@ impl ClientCommandRunner {
             .as_ref()
             .ok()
             .and_then(|conf| conf.wallet_config.clone()));
-        let wallet = ServiceUtils::load_wallet_context(&wallet_path);
+        let wallet = crate::utils::load_wallet_context(&wallet_path);
 
         Self {
             wallet_path,
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -39,8 +39,6 @@ use crate::common::utils::{self, LoadConfig};
 #[serde_as]
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct StorageNodeConfig {
-    /// Name of the storage node.
-    pub name: String,
     /// Directory in which to persist the database
     #[serde(deserialize_with = "utils::resolve_home_dir")]
     pub storage_path: PathBuf,
@@ -86,6 +84,9 @@ pub struct StorageNodeConfig {
     pub commission_rate: u64,
     /// The parameters for the staking pool.
     pub voting_params: VotingParams,
+    /// Name of the storage node.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub name: Option<String>,
 }
 
 impl StorageNodeConfig {
@@ -110,21 +111,20 @@ impl StorageNodeConfig {
             .get()
             .expect("key pair should already be loaded into memory")
     }
-}
 
-impl From<StorageNodeConfig> for NodeRegistrationParams {
-    fn from(value: StorageNodeConfig) -> Self {
-        let network_key_pair = value.network_key_pair();
-        let protocol_key_pair = value.protocol_key_pair();
-        Self {
-            name: value.name.clone(),
-            network_address: value.rest_api_address.into(),
+    /// Converts the configuration into a registration parameters used for node registration.
+    pub fn to_registration_params(&self, name: String) -> NodeRegistrationParams {
+        let network_key_pair = self.network_key_pair();
+        let protocol_key_pair = self.protocol_key_pair();
+        NodeRegistrationParams {
+            name,
+            network_address: self.rest_api_address.into(),
             public_key: protocol_key_pair.public().clone(),
             network_public_key: network_key_pair.public().clone(),
-            commission_rate: value.commission_rate,
-            storage_price: value.voting_params.storage_price,
-            write_price: value.voting_params.write_price,
-            node_capacity: value.voting_params.node_capacity,
+            commission_rate: self.commission_rate,
+            storage_price: self.voting_params.storage_price,
+            write_price: self.voting_params.write_price,
+            node_capacity: self.voting_params.node_capacity,
         }
     }
 }
--- a/crates/walrus-stress/src/refill.rs
+++ b/crates/walrus-stress/src/refill.rs
@@ -20,7 +20,6 @@ use sui_sdk::{
     SuiClient,
 };
 use tokio::{sync::Mutex, task::JoinHandle, time::MissedTickBehavior};
-use walrus_service::utils as ServiceUtils;
 use walrus_sui::utils::{send_faucet_request, sign_and_send_ptb, SuiNetwork};
 
 use crate::metrics::ClientMetrics;
@@ -137,7 +136,7 @@ impl FaucetOrWallet {
                 "Creating gas refill station from wallet: {:?}",
                 &wallet_path
             );
-            let wallet = ServiceUtils::load_wallet_context(&Some(wallet_path))?;
+            let wallet = walrus_service::utils::load_wallet_context(&Some(wallet_path))?;
             Ok(Self::new_wallet(wallet)?)
         } else {
             tracing::info!("Created gas refill station from faucet: {:?}", &sui_network);
--- a/crates/walrus-core/src/messages/proof_of_possession.rs
+++ b/crates/walrus-core/src/messages/proof_of_possession.rs
@@ -22,7 +22,7 @@ pub struct ProofOfPossessionMsg(pub(crate) ProtocolMessage<ProofOfPossessionBody
 impl ProofOfPossessionMsg {
     const INTENT: Intent = Intent::storage(IntentType::PROOF_OF_POSSESSION_MSG);
 
-    /// Creates a new confirmation message for the provided blob ID.
+    /// Creates a new `ProofOfPossessionMsg` with the given epoch, sui address, and BLS public key.
     pub fn new(epoch: Epoch, sui_address: [u8; 32], bls_public_key: PublicKey) -> Self {
         Self(ProtocolMessage {
             intent: Intent::storage(IntentType::PROOF_OF_POSSESSION_MSG),
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -90,4 +90,3 @@ voting_params:
 #     target_file_size_base: 67108864
 #     max_bytes_for_level_base: 536870912
 # commission_rate: 0
-# node_id: 0x222
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -1252,7 +1252,7 @@ pub fn storage_node_config() -> WithTempDir<StorageNodeConfig> {
     let temp_dir = TempDir::new().expect("able to create a temporary directory");
     WithTempDir {
         inner: StorageNodeConfig {
-            name: "node".to_string(),
+            name: Some("node".to_string()),
             protocol_key_pair: walrus_core::test_utils::protocol_key_pair().into(),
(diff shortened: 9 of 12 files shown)


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -260,6 +260,22 @@ pub enum CliCommands {
         #[serde(flatten)]
         target: FileOrBlobIdOrObjectId,
     },
+    /// Register a new node with the Walrus storage network.
+    RegisterNode {
+        #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
+        /// The path to the node's configuration file.
+        config_path: PathBuf,
+    },
```
mlegner: Suggestion: Put this in the node binary (`walrus-node`) instead of the client binary.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -260,6 +260,22 @@ pub enum CliCommands {
...
+    /// Stake with storage node.
+    Stake {
+        #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
+        /// The path to the node's configuration file.
+        config_path: PathBuf,
+        #[clap(short, long, default_value_t = default::staking_amount())]
+        #[serde(default = "default::staking_amount")]
+        /// The amount to stake with the storage node.
+        amount: u64,
+    },
```
mlegner: Instead of the `config_path`, can we directly use the node ID as a parameter?

crates/walrus-service/src/node/config.rs
```diff
@@ -78,6 +81,14 @@ pub struct StorageNodeConfig {
...
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub event_processor_config: Option<EventProcessorConfig>,
+    /// The commission rate of the storage node.
+    #[serde(default)]
+    pub commission_rate: u64,
+    /// The parameters for the staking pool.
+    pub voting_params: VotingParams,
+    /// The on chain ID of the node. This config can be populated by running `walrus register-node`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub node_id: Option<ObjectID>,
```
mlegner: Question: Do we need this for anything besides (currently) the staking?

crates/walrus-service/src/client/cli/args.rs
halfprice: Done

crates/walrus-service/src/node/config.rs
halfprice: yeah, other node <-> contract interaction requires the capability object, and we use the node id to find it.

mlegner: Hmm... How about we directly use the capability object's ID in this case?

crates/walrus-service/bin/node.rs
```diff
@@ -165,6 +198,62 @@ mod commands {
...
+    pub(crate) async fn register_node(config_path: PathBuf) -> anyhow::Result<()> {
+        let storage_config = StorageNodeConfig::load(&config_path)?;
+        let mut storage_config_store = storage_config.clone();
+
+        // Uses the Sui wallet configuration in the storage node config to register the node.
+        let contract_client = get_contract_client_from_node_config(&storage_config).await?;
+
+        let node_capability = contract_client
+            .register_candidate(&storage_config.into())
+            .await?;
```
mlegner: Important: Can we check if a `node_id` is already defined and return an error in that case?

crates/walrus-service/bin/node.rs
```diff
@@ -66,6 +69,26 @@ enum Commands {
...
+    /// Stake with storage node.
+    Stake {
+        #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
+        /// The path to the node's configuration file.
+        config_path: PathBuf,
+        #[clap(short, long, default_value_t = default_staking_amount())]
+        #[serde(default = "default_staking_amount")]
+        /// The amount to stake with the storage node.
+        amount: u64,
+    },
```
jpcsmith: Suggestion: Move staking to the client binary's as it's primarily targeted to the general audience.

crates/walrus-service/bin/node.rs
```diff
@@ -66,6 +69,26 @@ enum Commands {
...
+    /// Stake with storage node.
+    Stake {
+        #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
+        /// The path to the node's configuration file.
+        config_path: PathBuf,
+        #[clap(short, long, default_value_t = default_staking_amount())]
+        #[serde(default = "default_staking_amount")]
+        /// The amount to stake with the storage node.
+        amount: u64,
```
jpcsmith: Suggestion: Rename as this is unclear as to whether it's WAL or the subdivision of WAL.

crates/walrus-service/bin/node.rs
```diff
@@ -66,6 +69,26 @@ enum Commands {
...
+    RegisterNode {
+        #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
+        /// The path to the node's configuration file.
+        config_path: PathBuf,
+    },
+    /// Stake with storage node.
+    Stake {
+        #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
+        /// The path to the node's configuration file.
+        config_path: PathBuf,
```
jpcsmith, line 76: Suggestion: The argument names, such as `config_path` should have the same name and short and long forms as in the client.

crates/walrus-service/bin/node.rs
```diff
@@ -165,6 +198,62 @@ mod commands {
...
+        let contract_client = get_contract_client_from_node_config(&storage_config).await?;
+
+        let node_capability = contract_client
+            .register_candidate(&storage_config.into())
+            .await?;
+
+        // Updates storage config to include the node ID in it.
+        storage_config_store.node_id = Some(node_capability.node_id);
+        let serialized_storage_node_config = serde_yaml::to_string(&storage_config_store)?;
+        fs::write(config_path, serialized_storage_node_config)?;
```
jpcsmith: Important: Do not write values to the config file. Config files are input sources, most users would be surprised when fields in their config file get overwritten.

Suggestion: Output the node ID during registration, and provide a CLI command to display it from the user's wallet if they want to see it again (given the config pointing to the walrus contract). There should be no need to add it to the config as a node should only have a single capability for a given walrus instance. See [this slack thread](https://mysten-labs.slack.com/archives/C079UUZQSM6/p1727957122983089) for more information.

crates/walrus-service/bin/node.rs
```diff
@@ -165,6 +198,62 @@ mod commands {
...
+        let storage_config = StorageNodeConfig::load(&config_path)?;
+        let Some(node_id) = storage_config.node_id else {
+            bail!("Node ID not found in the config file");
+        };
+
+        // Uses the Sui wallet configuration in the storage node config to register the node.
+        let contract_client = get_contract_client_from_node_config(&storage_config).await?;
+        let staked_wal = contract_client.stake_with_pool(amount, node_id).await?;
+        StakeOutput { staked_wal }.print_output(false)
+    }
```
jpcsmith: Important: This should be in the walrus client, and allow the user to specify the ID of the node with which they would like to stake.

crates/walrus-service/src/client/responses.rs
jpcsmith, line 1: Important: This file is for client response types, not storage node logic. Move relevant responses to `node/cli/responses.rs`.

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -392,6 +394,26 @@ impl CliOutput for DeleteOutput {
...
 }
 
+impl CliOutput for RegisterNodeOutput {
+    fn print_cli_output(&self) {
+        println!("{} Registered storage node with capability:", success());
+        println!("Capability object id: {}", self.node_capability.id);
+        println!("Node id: {}", self.node_capability.node_id);
+        println!("The node id is added to the storage config file.");
+    }
+}
```
jpcsmith: Important: This file is for client logic, not storage node logic. Move the `CliOutput` trait to the `common` module, and the implementation of the trait for storage-node-specific types to `src/node/cli`.

crates/walrus-service/src/node/config.rs
jpcsmith: Remark: We should be able to look this up from the node's wallet, see my other comment on the topic.

crates/walrus-service/bin/node.rs
```diff
@@ -46,8 +48,9 @@ struct Args {
     command: Commands,
 }
 
-#[derive(Subcommand, Debug, Clone)]
+#[derive(Subcommand, Debug, Clone, Deserialize)]
 #[clap(rename_all = "kebab-case")]
+#[serde(rename_all = "camelCase", rename_all_fields = "camelCase")]
```
mlegner: Question: Why do we need these?

crates/walrus-service/bin/node.rs
mlegner: Suggestion (as a follow-up): Add a command to `walrus-deploy` to stake with multiple storage nodes in a single command to simplify the deployment.

crates/walrus-service/bin/node.rs
```diff
@@ -66,6 +69,26 @@ enum Commands {
         #[clap(default_value = "protocol.key")]
         out: PathBuf,
     },
+    /// Register a new node with the Walrus storage network.
+    RegisterNode {
```
mlegner: Nit: Maybe `Register` would be sufficiently clear and shorter.

crates/walrus-service/src/client/cli/cli_output.rs
mlegner, line 1: Important: The node binary should not import from the `client` module. If we want to use the `CliOutput` trait, let's move it to `common`. But as we don't have the JSON mode on `walrus-node` it might not actually be necessary; we can also just implement `Display` instead.

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -392,6 +394,26 @@ impl CliOutput for DeleteOutput {
     }
 }
 
+impl CliOutput for RegisterNodeOutput {
+    fn print_cli_output(&self) {
+        println!("{} Registered storage node with capability:", success());
+        println!("Capability object id: {}", self.node_capability.id);
```
mlegner: Nit: Can we capitalize "ID" here and below?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -392,6 +394,26 @@ impl CliOutput for DeleteOutput {
...
+}
+
+impl CliOutput for StakeOutput {
+    fn print_cli_output(&self) {
+        println!("{} Staked WAL successfully.", success());
+        println!("Node id: {}", self.staked_wal.node_id);
+        println!("Staked WAL object id: {}", self.staked_wal.id);
+        println!("Staked WAL state: {:?}", self.staked_wal.state);
+        println!("Staked principal: {}", self.staked_wal.principal);
+        println!("Activation epoch: {}", self.staked_wal.activation_epoch);
```
mlegner: Suggestion: Implement `Display` for `StakedWal`.

crates/walrus-service/src/node/config.rs
```diff
@@ -78,6 +81,14 @@ pub struct StorageNodeConfig {
     /// Configuration for running checkpoint processor
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub event_processor_config: Option<EventProcessorConfig>,
+    /// The commission rate of the storage node.
+    #[serde(default)]
+    pub commission_rate: u64,
```
jpcsmith, line 84: Question: what does the u64 stands for, as rate implies percentage or fraction?

crates/walrus-service/src/node/config.rs
```diff
@@ -38,6 +39,8 @@ use crate::common::utils::{self, LoadConfig};
 #[serde_as]
 #[derive(Debug, Clone, Deserialize, Serialize)]
 pub struct StorageNodeConfig {
+    /// Name of the storage node.
+    pub name: String,
```
jpcsmith: Suggestion: Remove the name from the config, and instead have it be provided on the command line, since it is only ever used during registration.

crates/walrus-service/src/client/cli/args.rs
halfprice: Yes. In fact, given that this is not an operator only op, we should just use node id.

crates/walrus-service/bin/node.rs
halfprice: created #927

crates/walrus-service/src/client/cli/cli_output.rs
halfprice: Done

crates/walrus-service/src/node/config.rs
halfprice: This field is removed from the config file.

crates/walrus-service/bin/node.rs
halfprice: No longer applicable since we no longer record node_id in the config file.

crates/walrus-service/bin/node.rs
halfprice: Good question. Since parsing config path requires serde, I assumed this is needed. Looking closely, this is only used in the client cli if the input is in json format, and the node binary doesn't support json input, so we don't really need this. Removed.

crates/walrus-service/src/client/cli/cli_output.rs
halfprice: Done. And made sure that node doesn't import client module.

crates/walrus-service/src/client/cli/cli_output.rs
halfprice: Done and used

crates/walrus-service/bin/node.rs
halfprice: Done

crates/walrus-service/bin/node.rs
halfprice: I updated the comment to indicate it is WAL. It'll be printed in the help. feels `amount_wal_to_stake` is a bit mouthful as an input option.

crates/walrus-service/bin/node.rs
halfprice: Done

crates/walrus-service/bin/node.rs
halfprice: Done, and removed writing node id to config file.

crates/walrus-service/src/client/responses.rs
halfprice: As @mlegner mentioned, CliOutput is for json output, and we currently don't support it in walrus-node command, I just directly print the output instead of using CliOuput.

crates/walrus-service/src/node/config.rs
halfprice: The code doesn't mention this anywhere, but I was assuming this is in basis points (added this in the doc), same as in Sui. cc @karlwuest

crates/walrus-service/src/node/config.rs
halfprice: I keep this as is, because `name` is also in the node info on chain, so I feel having a permanent place to store it may be helpful.

crates/walrus-service/bin/node.rs
```diff
@@ -46,7 +47,7 @@ struct Args {
     command: Commands,
 }
 
-#[derive(Subcommand, Debug, Clone)]
+#[derive(Subcommand, Debug, Clone, Deserialize)]
```
mlegner, line 50: Nit: The addition of `Deserialize` here is not really needed I think.

crates/walrus-stress/src/refill.rs
```diff
@@ -20,7 +20,7 @@ use sui_sdk::{
     SuiClient,
 };
 use tokio::{sync::Mutex, task::JoinHandle, time::MissedTickBehavior};
-use walrus_service::client::cli::load_wallet_context;
+use walrus_service::utils as ServiceUtils;
```
mlegner: Nit: The PascalCasing here is a bit misleading. Can we either import this `as service_utils` or simply use `walrus_service::utils::...` below?

crates/walrus-service/src/client/cli.rs
```diff
@@ -20,6 +20,7 @@ use walrus_core::BlobId;
 use walrus_sui::client::{SuiContractClient, SuiReadClient};
 
 use super::{default_configuration_paths, Blocklist, Client, Config};
+use crate::utils as ServiceUtils;
```
mlegner: Nit: The PascalCasing here is a bit misleading. Can we either import this as `service_utils` or simply use `crate::utils::...` below?

crates/walrus-service/bin/node.rs
mlegner: Important: I think this should be `FROST` instead of `WAL`.

crates/walrus-service/src/node/config.rs
karlwuest: Currently, the commission rate doesn't do anything yet, but yes, I also assumed that this is provided in basis points.

crates/walrus-service/src/node/config.rs
jpcsmith: It's just an additional thing that we *have* to have in the config, which is rather annoying. Given that it's only used during registration, it should not be required to set it, and forcing a user to set it will likely cause them to put garbage since it's not used, until it is used.

crates/walrus-service/src/client/cli.rs
jpcsmith: +1 for importing as `utils` or just calling `crate::utils`.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -260,6 +260,16 @@ pub enum CliCommands {
...
     },
+    /// Stake with storage node.
+    Stake {
+        #[clap(long)]
+        /// The object ID of the storage node to stake with.
+        node_id: ObjectID,
+        #[clap(short, long, default_value_t = default::staking_amount_wal())]
+        #[serde(default = "default::staking_amount_wal")]
+        /// The amount of WAL to stake with the storage node.
+        amount: u64,
```
jpcsmith: This may also need to be changed to FROST.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -48,13 +48,14 @@ use crate::{
...
             ReadOutput,
+            StakeOutput,
         },
         Client,
         ClientDaemon,
         Config,
         StoreWhen,
     },
-    utils::MetricsAndLoggingRuntime,
+    utils::{self as ServiceUtils, MetricsAndLoggingRuntime},
```
jpcsmith: Suggestion: Just call crate::utils or use as utils.

crates/walrus-service/bin/node.rs
halfprice: I see. I misunderstood the original comment. I thought FROST is also (part of) WAL token. Comment updated.

crates/walrus-service/bin/node.rs
halfprice: It is required. Otherwise the below serde is not included. See similar issue here: https://users.rust-lang.org/t/cannot-find-attribute-serde-in-this-scope/48112

crates/walrus-stress/src/refill.rs
halfprice: Make sense. There was a similar module util in one of the files, and I was just trying to address the conflict. Using full path SGTM.

crates/walrus-service/src/client/cli.rs
halfprice: Done

crates/walrus-service/src/client/cli/args.rs
halfprice: Do you mean s/amount/frost or just update comment?

crates/walrus-service/src/client/cli/runner.rs
halfprice: done

crates/walrus-service/src/node/config.rs
halfprice: As discussed in chat, changing this field in the config file to an optional field in the config, and an optional input in the command line. The command line arg takes precedence.

crates/walrus-service/src/client/cli/args.rs
halfprice: Keep the arg name `amount`, and update everywhere else `FROST`.

`--frost` as an arg seems unintuitive.

Issue #318, opened 2024-04-25
Ensure that each public key only appears once in the Committee.


jpcsmith:
In addition to #32, we need to ensure that an entity does not register two storage nodes with the same public key, otherwise we'll be unable to use the node's protocol public key as an ID for that node.

PR #923, opened 2024-10-03
feat: fetch capability object assuming that there is at most one per address

halfprice:
Going to use `get_address_capability_object` in #918 

Contract side enforcement will be handled in #318

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -622,7 +622,7 @@ pub struct StubContractService {
 impl SystemContractService for StubContractService {
     async fn invalidate_blob_id(&self, _certificate: &InvalidBlobCertificate) {}
 
-    async fn epoch_sync_done(&self, _node_id: ObjectID, _epoch: Epoch) {}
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -823,15 +823,13 @@ impl ContractClient for SuiContractClient {
...
+    async fn epoch_sync_done(&self, epoch: Epoch) -> SuiClientResult<()> {
+        let node_capability = get_address_capability_object(
             &self.read_client.sui_client,
             self.wallet_address,
             self.read_client.system_pkg_id,
-            &[],
         )
         .await?
-        .find(|cap| cap.node_id == node_id)
         .ok_or(SuiClientError::StorageNodeCapabilityObjectNotSet)?;
```
jpcsmith, line 836: Question: Any object to just caching this in the SuiContractClient? @karlwuest

karlwuest: No, caching it there probably makes sense.

halfprice: Will be done as a follow up

Issue #929, opened 2024-10-04
Improve performance of blob recovery


mlegner:
Currently, blob recovery is very slow on the PTN, sometimes at a rate of just 1-2 blobs per second (for nodes with 100 shards). We need to find the reasons for this and increase the performance substantially.

PR #933, opened 2024-10-04
fix: set connection timeout limit and increase blob sync limits

jpcsmith:
This PR sets a default TCP connect timeout and increase blob sync limits.

#### TCP Connect timeout

Sets the connect timeout (time for the TCP handshake) to storage nodes to 5 seconds by default in the client builder, and 1s in the storage node's config.

The initial retransmission timeout (RTO) is 1 second (3 seconds legacy implementations). If out first SYN packet is lost, then a second will be sent at `t ~= 1s` (`t ~= 3s` legacy), and a third at `t ~= 1s + 2.5s` (`t ~= 3s + 7.5s` legacy), since the RTO is increased by a factor of 2.5 after each failure.

A value of 5 seconds therefore allows 3 SYN attempts assuming an initial RTO of 1 second and 2 attempts for an initial RTO of 3 seconds.

The value of 1s in the storage node config causes us to quickly fail if we are unable to connect to a remote node.

See [RFC6298](https://datatracker.ietf.org/doc/html/rfc6298#section-2) for more details.

#### Blob sync limit increases

~We now allow up to 50,000 slivers to be recovered in parallel. If we assume around 100 MB unencoded blobs, this is around 250 kB per sliver on average, which amounts to around 12.5 GB of memory used by recovery.~

~Since around 1% stake is 10 shards, we would expect nodes to have at most around 30 (50) shards on mainnet (testnet)
With 20 shards, a node must recover 40 slivers per blob, for which the 50,000 sliver limit allows for recovering around 1,250 blobs at once. The limit is therefore set to 1000 for now which accounts for the fact that we have larger shards at the moment and we do not want to spread the sliver request permits across too many blobs.~

The new limit is set such that we do not hammer the remote servers with too many requests. The limit is somewhat low, and requires re-evaluation.

Contributes to #929

### DIFF
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -136,8 +136,8 @@ pub struct BlobRecoveryConfig {
 impl Default for BlobRecoveryConfig {
     fn default() -> Self {
         Self {
-            max_concurrent_blob_syncs: 1000,
-            max_concurrent_sliver_syncs: 50_000,
+            max_concurrent_blob_syncs: 100,
+            max_concurrent_sliver_syncs: 2_000,
             committee_service_config: CommitteeServiceConfig::default(),
         }
     }


crates/walrus-sdk/src/client.rs
```diff
@@ -195,9 +195,22 @@ pub struct ClientBuilder {
...
 
 impl ClientBuilder {
+    /// Default timeout that is configured for connecting to the remote server.
+    ///
+    /// Modern advice is that TCP implementations should have a retry timeout of 1 second
+    /// (previously, it was 3 seconds). In the event of a lossy network, the SYN/SYN-ACK
+    /// packets may be lost necessitating one or several retries until actually connecting to the
+    /// server. We therefore want to allow for several retries.
+    ///
+    /// The default of 5 seconds should allow for around 2-3 SYN attempts before failing.
```
mlegner, line 209: I generally agree with these arguments. However, in our case we should have a TCP connection open to the other storage nodes in most cases already, and we have implicit retries because we send multiple separate requests to other nodes. So I'd say we could reduce this timeout (to 1 second?) to have a better protection against unresponsive nodes.

jpcsmith: How about I add a config variable for the storage node to set this, and make that variable a default of 1s?

crates/walrus-service/src/node/config.rs
```diff
@@ -136,8 +136,8 @@ pub struct BlobRecoveryConfig {
 impl Default for BlobRecoveryConfig {
     fn default() -> Self {
         Self {
-            max_concurrent_blob_syncs: 10,
-            max_concurrent_sliver_syncs: 500,
+            max_concurrent_blob_syncs: 1000,
+            max_concurrent_sliver_syncs: 50_000,
```
mlegner: Actually, thinking about this again: The limits might work with respect to memory, but they generate a huge number of requests: 50k slivers require ~500*50k = 25M requests, which may be a bit much. 🤔

crates/walrus-sdk/src/client.rs
mlegner: That sounds great!

jpcsmith: Added.

crates/walrus-service/src/node/config.rs
mlegner: How about we reduce the sliver value to something like ~2k but keep the value for the blobs unchanged (or only reduce it slightly)?
```suggestion
            max_concurrent_blob_syncs: 200,
            max_concurrent_sliver_syncs: 2000,
```

jpcsmith: We would likely hit the http2 concurrent stream limit (currently set to 200 in axum/hyper) before we hit that. With 10 storage nodes, we can actually only have 2,000 requests (of any type) in flight from a single source at a given time.

Looking at the source code of reqwest, if we exceed this it seems that the request is refused at the server, then retried transparently in the reqwest client, so it's not to our benefit to exceed this by too much.

Shall we set slivers to 2000 and blobs to 100?

```suggestion
            max_concurrent_blob_syncs: 100,
            max_concurrent_sliver_syncs: 2_000,
```

mlegner: Sure, that works. 👍

jpcsmith: Had not refreshed, I now see your comment.

PR #934, opened 2024-10-04
feat: implement proof of possession

### LINKED ISSUE #32: Implement verification of proof of private key possession in public key registration function
When storage nodes register their BLS public key, we need to verify that they are in possession of the corresponding private key to prevent rogue key attacks.

Depends on #131

karlwuest:
closes #32

### DIFF
--- a/contracts/walrus/sources/staking/staking_inner.move
+++ b/contracts/walrus/sources/staking/staking_inner.move
@@ -169,7 +169,7 @@ public(package) fun create_pool(
         network_address,
         public_key,
         network_public_key,
-    proof_of_possession,
+        proof_of_possession,
         commission_rate,
         storage_price,
         write_price,


contracts/walrus/sources/staking.move
```diff
@@ -234,109 +238,6 @@ public(package) fun is_epoch_sync_done(self: &Staking): bool {
...
-fun test_withdraw_node() {
-    let ctx = &mut tx_context::dummy();
-    let clock = clock::create_for_testing(ctx);
-    let mut staking = new_for_testing(ctx);
-    let mut cap = staking.register_candidate(
-        b"node".to_string(),
-        b"127.0.0.1".to_string(),
-        x"820e2b273530a00de66c9727c40f48be985da684286983f398ef7695b8a44677",
-        x"820e2b273530a00de66c9727c40f48be985da684286983f398ef7695b8a44677",
-        0,
```
jpcsmith, line 265: Question: Why the test purge?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -168,6 +169,7 @@ public(package) fun create_pool(
         network_address,
         public_key,
         network_public_key,
+    proof_of_possession,
```
jpcsmith: Nit: indentation.

crates/walrus-core/src/messages/proof_of_possession.rs
```diff
@@ -0,0 +1,88 @@
...
+        Self(ProtocolMessage {
+            intent: Intent::storage(IntentType::PROOF_OF_POSSESSION_MSG),
+            epoch,
+            message_contents: ProofOfPossessionBody {
+                sui_address,
+                bls_public_key,
+            },
+        })
+    }
+}
```
jpcsmith, line 36: Question: Is there any benefit to specifying the epoch?

contracts/walrus/sources/staking.move
karlwuest: These were all test skeletons that Damir added when adding the contract skeleton. They don't actually do anything, so I cleaned them up (we test those functions in dedicated test files).

crates/walrus-core/src/messages/proof_of_possession.rs
karlwuest: It provides some freshness, but more importantly, the definition for our message type includes the epoch (see [here](https://github.com/MystenLabs/walrus/blob/cdc7164e01e4e3d438d7c2dd985a1dd8928cbf99/contracts/walrus/docs/msg_formats.txt#L19)).

contracts/walrus/sources/staking/staking_inner.move
karlwuest: Good catch, I thought I had fixed that before pushing.

crates/walrus-core/src/messages/proof_of_possession.rs
jpcsmith: I know it requires the epoch, but since it's never checked, one could always submit with Epoch 0, so I was wondering if there was any benefit to not doing so by default.

karlwuest: The epoch is checked when verifying the proof.

karlwuest: There is even a test for that in `walrus::e2etests::test_register_invalid_pop_epoch()`.

jpcsmith: Ah, my mistake, I now see that you verify the signature against a new message that you create with the current epoch.

PR #939, opened 2024-10-05
feat: walrus-node command to generate storage node config

### LINKED ISSUE #875: Allow generating a new config for the storage node via CLI
Generate a storage node config and keys. Keys should be placed in key-files with appropriate permission, as opposed to directly in the config.

halfprice:
Create `generate-config` command in `walrus-node` binary to generate a storage node config to be used in the network.

The command requires to load protocol key and network key in key files.

close #875


scripts/generate_node_config.sh
mlegner, line 1: Question: What is this script used for? Is this just an example?

crates/walrus-service/bin/node.rs
```diff
@@ -188,11 +266,23 @@ mod commands {
...
+        let mut storage_config = StorageNodeConfig::load(&config_path)?;
         let node_name = name.or(storage_config.name.clone()).ok_or(anyhow!(
             "Name is required to register node. Set it in the config file or provided in the \
                 commandline argument."
         ))?;
+
+        if storage_config.protocol_key_pair.is_path() {
+            storage_config.protocol_key_pair.load()?;
+        } else {
+            bail!("storage config protocol keys must be loaded from a key file");
```
mlegner, line 320: Question: Is this to force the operators to use key files instead of putting the keys into the configuration?

scripts/generate_node_config.sh
halfprice: Yes, just an example so that people can copy paste and refill quickly. Not sure if this is needed. I added this because it took me a few try to get the command correct.

crates/walrus-service/bin/node.rs
halfprice: yes, that's the idea

scripts/generate_node_config.sh
mlegner: OK, makes sense. I don't think that anyone will look for it there though. So I'd either remove it or put it somewhere into the README file.

halfprice: Remove it for now and will add one to the README file later.

PR #944, opened 2024-10-05
fix(node): various improvements to the node binary

mlegner:
- remove unnecessary serde attributes
- minor improvements to the `key-gen` command
- move options for `generate-config` to new `GenerateConfigArgs` struct
- make votes for storage and write price optional

### DIFF
--- a/crates/walrus-service/bin/deploy.rs
+++ b/crates/walrus-service/bin/deploy.rs
@@ -78,7 +78,7 @@ struct DeploySystemContractArgs {
     testbed_config_path: Option<PathBuf>,
     // Note: The storage unit is set in `crates/walrus-sui/utils.rs`. Change the unit in
     // the doc comment here if it changes.
-    /// The price in FROST to set per unit of storage (1 KiB) and epoch.
+    /// The price in FROST to set per unit of storage (1 KiB) per epoch.
     #[arg(long, default_value_t = config::defaults::storage_price())]
     storage_price: u64,
     /// The price in FROST to set for writing one unit of storage (1 KiB).
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -69,7 +69,10 @@ enum Commands {
         /// [default: ./<KEY_TYPE>.key]
         #[clap(short, long)]
         out: Option<PathBuf>,
-        /// Which type of key to generate.
+        /// Which type of key to generate. Valid options are 'protocol' and 'network'.
+        ///
+        /// The protocol key is used to sign Walrus protocol messages.
+        /// The network key is used to authenticate nodes in network communication.
         #[clap(short, long)]
         key_type: KeyType,
         /// Overwrite existing files.
@@ -124,7 +127,7 @@ struct GenerateConfigArgs {
     /// The path where the Walrus database will be stored.
     storage_path: PathBuf,
     #[clap(long)]
-    /// The path to the key pair used in Walrus protocol messages
+    /// The path to the key pair used in Walrus protocol messages.
     protocol_key_pair_path: PathBuf,
     #[clap(long)]
     /// The path to the key pair used to authenticate nodes in network communication.


crates/walrus-service/bin/deploy.rs
```diff
@@ -75,11 +78,11 @@ struct DeploySystemContractArgs {
     testbed_config_path: Option<PathBuf>,
     // Note: The storage unit is set in `crates/walrus-sui/utils.rs`. Change the unit in
     // the doc comment here if it changes.
-    /// The price to set per unit of storage (1 KiB) and epoch.
-    #[arg(long, default_value_t = 50)]
+    /// The price in FROST to set per unit of storage (1 KiB) and epoch.
```
halfprice:
```suggestion
    /// The price in FROST to set per unit of storage (1 KiB) per epoch.
```

crates/walrus-service/bin/deploy.rs
```diff
@@ -75,11 +78,11 @@ struct DeploySystemContractArgs {
...
+    #[arg(long, default_value_t = config::defaults::storage_price())]
     storage_price: u64,
-    /// The price to set for writing one unit of storage (1 KiB).
-    #[arg(long, default_value_t = 50)]
+    /// The price in FROST to set for writing one unit of storage (1 KiB).
```
halfprice, line 84:
```suggestion
    /// The price in FROST to set for writing one unit of storage (1 KiB) per epoch.
```

crates/walrus-service/bin/node.rs
```diff
@@ -62,92 +62,115 @@ enum Commands {
...
+        force: bool,
     },
 
+    /// Generate a new node configuration.
+    GenerateConfig(Box<GenerateConfigArgs>),
+
     /// Register a new node with the Walrus storage network.
     Register {
         #[clap(short, long)]
-        #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
```
halfprice, line 79: Just give this a try and it is indeed not needed. Does this mean that we can remove `resolve_home_dir` anywhere?

crates/walrus-service/bin/node.rs
```diff
@@ -62,92 +62,115 @@ enum Commands {
...
-        #[clap(default_value = "protocol.key")]
-        out: PathBuf,
-        /// Whether to generate a network key instead of a protocol key.
-        #[clap(short, long, default_value_t = false)]
-        network_key: bool,
+        /// not overwritten and the operation will fail unless the `--force` option is provided.
+        /// [default: ./<KEY_TYPE>.key]
+        #[clap(short, long)]
+        out: Option<PathBuf>,
+        /// Which type of key to generate.
```
halfprice: suggestion: can we print the two options?

crates/walrus-service/bin/node.rs
```diff
@@ -322,36 +361,32 @@ mod commands {
...
+            voting_params: VotingParams {
+                storage_price,
+                write_price,
+                node_capacity,
+            },
+            rest_api_address,
+            metrics_address,
+            name,
+            commission_rate,
+            ..Default::default()
```
halfprice, line 390: Nice! Should have thought about this...

crates/walrus-service/src/node/config.rs
```diff
@@ -367,6 +367,16 @@ pub mod defaults {
         Some(Some(REST_GRACEFUL_SHUTDOWN_PERIOD_SECS))
     }
 
+    /// The default vote for the storage price.
+    pub fn storage_price() -> u64 {
+        5
```
halfprice, line 372: Question: I thought about doing this but don't know how to set the default properly. 5 FROST seems too cheap...

crates/walrus-service/bin/deploy.rs
mlegner: This is a one-time price, so there should not be an epoch.

crates/walrus-service/bin/node.rs
mlegner: The serde attributes are only necessary in the client binary when using the JSON mode. As we don't have that in the node binary, we don't need any serde attributes.

crates/walrus-service/src/node/config.rs
mlegner: Assuming 1 WAL ~ 1 USD, this value of 5 FROST / KiB / epoch corresponds to ~ 5 * 12 * 1M FROST ~ 0.06 USD per GiB (encoded!) per year, which corresponds to ~ 0.3 USD per GiB (unencoded) per year. So this seems quite reasonable to me.

crates/walrus-service/bin/node.rs
halfprice: Good to know!

crates/walrus-service/src/node/config.rs
halfprice: The alpha here is `Assuming 1 WAL ~ 1 USD` :)

PR #946, opened 2024-10-05
feat: start epoch change

### LINKED ISSUE #941: Improve/extend `EpochChangeDriver` on storage nodes
The `EpochChangeDriver` added in #915 now calls the `voting_end` Move function at an appropriate time, see [this transaction](https://testnet.suivision.xyz/txblock/Fo973xVUiqrUL7LyHiomXi6aMGquLGtGQxb5UvKKEQpy).

There are a few things to add/improve:
- [ ] Make sure the driver works if the node is restarted during an epoch before the voting ends.
- [ ] Only call the function if it hasn't been called before by a different node; currently, all nodes call this function (see [here](https://testnet.suivision.xyz/object/0x222c31f446ffaf64092b336462284d991c0b8ee118f142f426165ee367f0368e)).
- [ ] Add corresponding calls to the `initiate_epoch_change` function. (The call on the explorer was done manually.)

jpcsmith:
This starts epoch change, cancels pending scheduled calls on appropriate events, and schedules calls when the node starts-up.

Closes #941. Depends on #945.

### DIFF
--- a/crates/walrus-service/src/node/epoch_change_driver.rs
+++ b/crates/walrus-service/src/node/epoch_change_driver.rs
@@ -16,7 +16,6 @@ use std::{
 use chrono::{DateTime, TimeDelta, Utc};
 use futures::{future::BoxFuture, FutureExt};
 use rand::{rngs::StdRng, Rng, SeedableRng};
-use tracing::Instrument;
 use walrus_core::Epoch;
 use walrus_sui::{
     client::FixedSystemParameters,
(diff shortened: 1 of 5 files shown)


crates/walrus-service/tests/epoch_change.rs
```diff
@@ -0,0 +1,30 @@
...
+
+use tokio::time;
+use walrus_core::Epoch;
+use walrus_service::test_utils::test_cluster;
+use walrus_test_utils::Result as TestResult;
+
+#[ignore = "ignore E2E tests by default"]
+#[tokio::test]
+async fn nodes_drive_epoch_change() -> TestResult {
+    let epoch_duration = Duration::from_secs(10);
```
mlegner, line 14: Question: Are 5 seconds too short for the test to be robust?

crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -224,36 +341,109 @@ impl std::fmt::Debug for EpochChangeDriverInner<'_> {
...
+        }
+    }
+
+    /// Invoke the stored operation after its selected delay, retrying if it fails.
+    #[tracing::instrument(skip_all)]
+    async fn wait_then_call(mut self) {
         let mut backoff = ExponentialBackoffState::new(MIN_BACKOFF, MAX_BACKOFF, None);
 
-        while let Err(error) = self.try_to_end_voting().await {
+        while let Err(error) = self.try_wait_then_call().await {
```
mlegner, line 395: Remark: AFAIU, this adds the jitter in every retry, correct? I'd say this is probably not necessary, but OTOH it also isn't problematic.

crates/walrus-service/tests/epoch_change.rs
jpcsmith: Increased this to 10, in the event that this is too flaky, we can increase it.

crates/walrus-service/src/node/epoch_change_driver.rs
jpcsmith: Correct, this adds jitter even if the duration is zero. One of the benefits of this is that a yet-to-be-processed event may cancel the scheduled operation, so this may avoid transactions that would otherwise fail.

PR #949, opened 2024-10-06
feat(node): add one-step setup command

mlegner:


crates/walrus-service/bin/node.rs
```diff
@@ -387,24 +453,72 @@ mod commands {
...
+        keygen(&network_key_path, KeyType::Network, true)?;
+
+        println!(
+            "Generating Sui wallet for {} at '{}'",
+            sui_network.r#type(),
+            wallet_config.display()
+        );
+        let mut wallet = walrus_sui::utils::create_wallet(&wallet_config, sui_network.env(), None)?;
+
+        println!("Requesting SUI from faucet...");
```
halfprice: Question: what about WAL?

mlegner: The storage nodes don't actually need WAL, but they need SUI for gas fees.

halfprice: Yes, make sense... Just waking up.

Given that we have experienced issues with the testnet faucet (and there may not be a faucet for localnet depending on the setup), can we make this step an optional step? Also the step that generating the wallet as an optional step. I'm guessing that we will probably sponsor the wallet before the operator ready to start node.

crates/walrus-service/bin/node.rs
```diff
@@ -122,16 +164,7 @@ impl KeyType {
...
-    #[clap(long)]
-    /// The path to the key pair used in Walrus protocol messages.
-    protocol_key_pair_path: PathBuf,
-    #[clap(long)]
-    /// The path to the key pair used to authenticate nodes in network communication.
-    network_key_pair_path: PathBuf,
+struct ConfigArgs {
     #[clap(long)]
     /// HTTP URL of the Sui full-node RPC endpoint (including scheme and port).
     sui_rpc: String,
```
halfprice: This seems to be required right now, but `sui_network` is not required. Can we also add the testnet public fullnode as the default?

mlegner: I think we could use the RPC node defined in the wallet as the default. 👍

crates/walrus-service/bin/node.rs
mlegner: > can we make this step an optional step?

I can change this to "Trying to get SUI from faucet" and then just print a warning if it fails.

I'd keep the wallet setup. If the operator wants to do things manually, they can always create the wallet and keys manually and then use the `generate-config` command.

crates/walrus-service/bin/node.rs
mlegner: Done in 7bb714fcb8183e8e3d226e7b48389312e8fabc5e.

crates/walrus-service/bin/node.rs
mlegner: Fixed the faucet calls in 7bb714fcb8183e8e3d226e7b48389312e8fabc5e.

PR #951, opened 2024-10-06
feat: integrate WAL exchange into Rust code

### LINKED ISSUE #859: Set up WAL faucet/exchange for Testnet

mlegner:
Creates Rust bindings for interacting with a WAL exchange.

- some refactoring and renaming
- improvements to the WAL exchange contracts
- add command to exchange SUI for WAL
- create and fund an exchange during the testbed deployment
- tested by running the local testbed; see the created exchange [here](https://devnet.suivision.xyz/object/0x30676e9a972fea6a8ea306b1a9c32e5494b0a4133404a1336e06f595a6829849)

Follow-up tasks (will create new issue): 
- Get and return the amount of WAL received in the exchange.
- Check the exchange rate beforehand?

Closes #859

### DIFF
--- a/contracts/walrus/Move.toml
+++ b/contracts/walrus/Move.toml
@@ -7,6 +7,7 @@ edition = "2024.beta"
 [dependencies]
 Sui = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-framework/packages/sui-framework", rev = "testnet-v1.34.2" }
 WAL = { local = "../wal" }
+# TODO: Remove this after we publish contracts separately (#909).
 WAL_exchange = { local = "../wal_exchange" }
 
 # For remote import, use the `{ git = "...", subdir = "...", rev = "..." }`.
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -547,11 +547,11 @@ mod default {
     }
 
     pub(crate) fn staking_amount_frost() -> u64 {
-        1_000_000_000
+        1_000_000_000 // 1 WAL
     }
 
     pub(crate) fn exchange_amount_mist() -> u64 {
-        1_000_000_000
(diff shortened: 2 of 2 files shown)


contracts/walrus/Move.toml
```diff
@@ -7,6 +7,7 @@ edition = "2024.beta"
 [dependencies]
 Sui = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-framework/packages/sui-framework", rev = "testnet-v1.34.2" }
 WAL = { local = "../wal" }
+WAL_exchange = { local = "../wal_exchange" }
```
karlwuest, line 11: Suggestion: Can you add a comment to remove this as dependency once we change the deployment from deploying as a single package to separate packages?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -537,6 +549,10 @@ mod default {
     pub(crate) fn staking_amount_frost() -> u64 {
         1_000_000_000
     }
+
+    pub(crate) fn exchange_amount_mist() -> u64 {
+        1_000_000_000
```
karlwuest: Suggestion: Reduce this amount, e.g. to 0.5 Sui. Currently the faucet issues exactly 1 Sui coins, so if the client only sends one faucet request before calling the function to get WAL, the transaction will fail.

mlegner: Ah, very good point. Will do.

PR #962, opened 2024-10-07
feat(move): add metadata to blob objects

giac-mysten:
Adds the ability to add metadata to blob objects, in the form of a `VecMap<String, String>` in a dynamic field.

### DIFF
--- a/contracts/walrus/tests/system/blob_tests.move
+++ b/contracts/walrus/tests/system/blob_tests.move
@@ -382,7 +382,7 @@ public fun test_delete_undeletable_blob(): system::System {
 
 #[test]
 public fun test_blob_add_metadata(): system::System {
-    get_blob_and_clean_up!(| blob| {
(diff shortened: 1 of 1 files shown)


contracts/walrus/tests/system/blob_tests.move
```diff
@@ -416,3 +469,20 @@ fun register_default_blob(system: &mut System, storage: Storage, deletable: bool
 fun default_blob_id(): u256 {
     blob::derive_blob_id(ROOT_HASH, RED_STUFF, SIZE)
 }
+
+/// Utiliy macro that calls the given function on a new Blob.
+///
+/// Creates the system, registers the default blob, and calls the given function with the blob.
+/// Finally, it destroys the blob and returns the system.
+macro fun get_blob_and_clean_up($f: |&mut Blob|->()): system::System {
```
karlwuest: Nit: maybe call this `call_function_with_default_blob` or something similar instead.

PR #974, opened 2024-10-08
feat: Enable event processor to bootstrap from non genesis checkpoint

sadhansood:
Today event processor only bootstraps from genesis checkpoint. This is unsustainable for existing networks like testnet and mainnet (fine for private testnet however). This PR allows node bootstrapping from the checkpoint when walrus package was published first.

### DIFF
--- a/crates/walrus-event/src/lib.rs
+++ b/crates/walrus-event/src/lib.rs
@@ -4,12 +4,23 @@
 use std::{
     fs::File,
     io::{BufReader, BufWriter},
-    path::PathBuf,
 };
 
+use anyhow::anyhow;
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use serde::{Deserialize, Serialize};
-use sui_types::{event::EventID, messages_checkpoint::CheckpointSequenceNumber};
+use sui_rest_api::Client;
+use sui_sdk::{
+    rpc_types::{SuiObjectDataOptions, SuiTransactionBlockResponseOptions},
+    SuiClient,
+};
+use sui_types::{
+    base_types::ObjectID,
+    committee::Committee,
+    event::EventID,
+    messages_checkpoint::{CheckpointSequenceNumber, VerifiedCheckpoint},
+    sui_serde::BigInt,
+};
 use walrus_core::BlobId;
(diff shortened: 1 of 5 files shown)


crates/walrus-event/src/lib.rs
```diff
@@ -146,3 +158,44 @@ impl EventStreamCursor {
...
+        .await?;
+    let sui_committee = sui_client
+        .governance_api()
+        .get_committee_info(Some(BigInt::from(checkpoint_data.checkpoint_summary.epoch)))
+        .await?;
+    let committee = Committee::new(
+        sui_committee.epoch,
+        sui_committee.validators.into_iter().collect(),
+    );
+    let verified_checkpoint = VerifiedCheckpoint::new_unchecked(checkpoint_data.checkpoint_summary);
```
jpcsmith, line 196: Question: Since this is no longer rooted in genesis, how is this verified?

crates/walrus-event/src/lib.rs
```diff
@@ -146,3 +158,44 @@ impl EventStreamCursor {
...
+                .with_previous_transaction(),
+        )
+        .await?;
+    let txn = sui_client
+        .read_api()
+        .get_transaction_with_options(
+            object
+                .data
+                .ok_or(anyhow!("No object data"))?
+                .previous_transaction
```
jpcsmith, line 180: Question: Is this not the last transaction to mutate the system object, instead of the one that created the object?

sadhansood: It is, but package objects are immutable afaik, so we should only get the transaction which created the walrus package.

crates/walrus-event/src/lib.rs
sadhansood: It is not today, our plan is to replace most of the FN reads to using light client (instead of using the regular FN client today) which does verified reads by mainnet. But in theory, we could do it by downloading all end of epoch sui checkpoint summaries and verifying the reference chain (light client does the same thing)

crates/walrus-event/src/lib.rs
jpcsmith: Ah, okay if that's the case.

PR #980, opened 2024-10-08
feat: add logs and shard health status

jpcsmith:
This PR allows getting the status of the shards, either in summary (default), or in summary and in detail (with `?detailed=true`), for example:

I have not added the option to get the details to the client, as it's uncertain whether we want to expose this programatically, or just for the storage node admins.

Output of `curl -k https://127.0.0.1:9189/v1/health | jq`:

<details>
<summary>
<i>JSON output...</i>
</summary>

```json
{
  "success": {
    "code": 200,
    "data": {
      "uptime": {
        "secs": 14,
        "nanos": 522612125
      },
      "epoch": 1,
      "publicKey": "uZnPIyJt//yPwJMg7jvQOFvFnQZc1klChzxwuGfDJTWFH0MQdDPr2hU+lZZCWniR",
      "shardSummary": {
        "owned": 3,
        "unknown": 0,
        "ready": 3,
        "inTransfer": 0,
        "inRecovery": 0,
        "readOnly": 0
      }
    }
  }
}
```
</details>

Output of `curl -k 'https://127.0.0.1:9189/v1/health?detailed=true' | jq`:
<details>
<summary>
<i>JSON output...</i>
</summary>

```json
{
  "success": {
    "code": 200,
    "data": {
      "uptime": {
        "secs": 94,
        "nanos": 408460833
      },
      "epoch": 1,
      "publicKey": "uZnPIyJt//yPwJMg7jvQOFvFnQZc1klChzxwuGfDJTWFH0MQdDPr2hU+lZZCWniR",
      "shardSummary": {
        "owned": 3,
        "unknown": 0,
        "ready": 3,
        "inTransfer": 0,
        "inRecovery": 0,
        "readOnly": 0
      },
      "shardDetail": {
        "owned": [
          {
            "shard": 0,
            "status": "ready"
          },
          {
            "shard": 1,
            "status": "ready"
          },
          {
            "shard": 2,
            "status": "ready"
          }
        ],
        "other": []
      }
    }
  }
}
```
</details>

### DIFF
--- a/crates/walrus-sdk/src/api.rs
+++ b/crates/walrus-sdk/src/api.rs
@@ -244,7 +244,7 @@ pub struct ShardStatusSummary {
     pub in_transfer: usize,
     /// The number of owned shards that are being recovered.
     pub in_recovery: usize,
-    /// The number of shards, no longer owned by the node, that are read only, 
+    /// The number of shards, no longer owned by the node, that are read only,
     /// i.e., only serving reads from this node.
     pub read_only: usize,
 }


crates/walrus-sdk/src/api.rs
```diff
@@ -217,4 +218,64 @@ pub struct ServiceHealthInfo {
...
+}
+
+/// Summary of the shard statuses.
+#[derive(Debug, Default, Clone, Deserialize, Serialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ShardStatusDetail {
+    /// Statuses of the shards for which the node is responsible in this epoch.
+    pub owned: Vec<ShardHealthInfo>,
+    /// Statuses of other shards the node currently stores.
+    pub other: Vec<ShardHealthInfo>,
```
mlegner, line 263: Question: What statuses can these `other` shards have? Only `unknown` and `read_only`?

Edit: below it seems those are either counted as `read_only` in the summary or excluded. Maybe we could then split this into `read_only` and `other_unknown`, each of which is a `Vec<ShardIndex>`?

crates/walrus-sdk/src/api.rs
```diff
@@ -217,4 +218,64 @@ pub struct ServiceHealthInfo {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub shard_detail: Option<ShardStatusDetail>,
+}
+
+/// Summary of the shard statuses.
+#[derive(Debug, Default, Clone, Deserialize, Serialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase")]
+pub struct ShardStatusSummary {
+    /// The number of shards, for which this node is responsible.
+    pub owned: usize,
```
mlegner: Suggestion: Mention that the `unknown`, `ready`, `in_transfer`, and `in_recovery` are all subcategories of `owned`.

crates/walrus-service/src/node.rs
```diff
@@ -990,6 +998,88 @@ impl StorageNodeInner {
...
+    ) -> Result<(), anyhow::Error> {
+        let this = self.clone();
+        tokio::task::spawn_blocking(move || this.storage.create_storage_for_shards(&new_shards))
+            .in_current_span()
+            .await??;
+        Ok(())
+    }
+}
+
+fn api_status_from_shard_status(status: ShardStatus) -> ApiShardStatus {
```
mlegner, line 1063: Question: Is there a reason not to have an `impl From<ShardStatus> for ApiShardStatus`?

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -416,22 +416,36 @@ pub async fn get_blob_status<S: SyncServiceState>(
...
 )]
 pub async fn health_info<S: SyncServiceState>(
+    Query(query): Query<HealthInfoQuery>,
     State(state): State<Arc<S>>,
 ) -> ApiSuccess<ServiceHealthInfo> {
-    ApiSuccess::ok(state.health_info())
+    let mut health_info = state.health_info();
+    if !query.detailed {
+        health_info.shard_detail = None;
+    }
```
mlegner: Remark: It feels a bit wasteful to compute the detailed info and then throw it away in the default case.
Suggestion: Add a parameter to the `health_info()` and `shard_health_status()` functions that determines whether to include the details or create separate functions.

crates/walrus-sdk/src/api.rs
jpcsmith: From the perspective of the protocol, we have shards that we're responsible for, plus those that we are sending to another storage node (read-only). Since we currently do not cleanup shards, we may have shards in a variety of states at the moment, but I figured it best not to expose this.

A shard is currently `Unknown` if it's state is set to `ShardStatus` is `None`, we fail to get its status (due to a DB error), or we fail to get a read-lock on the shard-table (such as when many shards are being created)`. We could report the first and second cases as `Unknown`, but as my goal was to only expose nodes held for transfer (which they may not be), and not the other internals of our management of the node (which IMO should only be exposed through metrics), I opted to just exclude them in the report. 

Let me know if you still think they should be reported.

mlegner: Ah, I see. I think I agree with your reasoning. Maybe then we can clarify in the docs that this includes shards that are not in the summary, just to avoid confusion?

crates/walrus-service/src/node.rs
jpcsmith: I did not want the storage node to need to worry about the conversion to an unique API-only representation, and was reluctant to place a from implementation for the storage node outside of the node. Since it is only used in the one place in the code, I opted for a helper function.

mlegner: OK, works for me.

crates/walrus-sdk/src/api.rs
```diff
@@ -226,24 +226,33 @@ pub struct ServiceHealthInfo {
...
+    /// The number of owned shards that are up-to-date for the epoch.
     pub ready: usize,
-    /// The number of shards that are being transferred to the node.
+    /// The number of owned shards that are being transferred to the node.
     pub in_transfer: usize,
-    /// The number of shards that are being recovered.
+    /// The number of owned shards that are being recovered.
     pub in_recovery: usize,
-    /// The number of shards that are read only, i.e., serving reads from this node.
+    /// The number of owned shards that are read only, i.e., serving reads from this node.
```
mlegner: Nit: If I understand correctly, these are "no longer owned", right?

crates/walrus-sdk/src/api.rs
```diff
@@ -226,24 +226,33 @@ pub struct ServiceHealthInfo {
...
+/// being transferred to another storage node.
 #[derive(Debug, Default, Clone, Deserialize, Serialize, utoipa::ToSchema)]
 #[serde(rename_all = "camelCase")]
 pub struct ShardStatusSummary {
     /// The number of shards, for which this node is responsible.
+    ///
+    /// Their statuses are summarized in `unknown`, `ready`, `in_transfer`, and `in_recovery`.
     pub owned: usize,
-    /// The number of shards in an unknown state.
+    /// The number of owned shards shards in an unknown state.
```
mlegner: Nit: Duplicated "shards".

jpcsmith:
```suggestion
    /// The number of owned shards in an unknown state.
```

crates/walrus-sdk/src/api.rs
jpcsmith:
```suggestion
    /// The number of shards, no longer owned by the node, that are read only, 
    /// i.e., only serving reads from this node.
```

crates/walrus-sdk/src/api.rs
```diff
@@ -217,4 +218,74 @@ pub struct ServiceHealthInfo {
...
+    pub owned: usize,
+    /// The number of owned shards in an unknown state.
+    pub unknown: usize,
+    /// The number of owned shards that are up-to-date for the epoch.
+    pub ready: usize,
+    /// The number of owned shards that are being transferred to the node.
+    pub in_transfer: usize,
+    /// The number of owned shards that are being recovered.
+    pub in_recovery: usize,
+    /// The number of shards, no longer owned by the node, that are read only,
```
jpcsmith:
```suggestion
    /// The number of shards, no longer owned by the node, that are read only,
```

PR #981, opened 2024-10-08
fix: do not evaluate sui config dir eagerly

karlwuest:


crates/walrus-sui/src/utils.rs
```diff
@@ -326,10 +326,12 @@ pub fn create_wallet(
...
-    let keystore_path = config_path
-        .parent()
-        .unwrap_or(&sui_config_dir()?)
-        .join(keystore_filename.unwrap_or(SUI_KEYSTORE_FILENAME));
+    let keystore_path = if let Some(config_dir) = config_path.parent() {
+        config_dir.join(keystore_filename.unwrap_or(SUI_KEYSTORE_FILENAME))
+    } else {
+        sui_config_dir()?.join(keystore_filename.unwrap_or(SUI_KEYSTORE_FILENAME))
+    };
+
```
jpcsmith: Hint: This could also be done with [`map_or_else`](https://doc.rust-lang.org/core/option/enum.Option.html#method.map_or_else), something along the lines of 
```rust
    .map_or_else(|| sui_config_dir(), Ok)?
    .join(..)
```

karlwuest: Ah, that's the function I was looking for, thanks! I knew there must be a one-liner 😄

PR #986, opened 2024-10-08
fix: do not send epoch sync done if the node is not part of committee

### LINKED ISSUE #982: Check if node is part of current committee before calling epoch sync done
If a node is registered and started, it otherwise continues to call the function. See e.g. here: https://suiscan.xyz/testnet/account/0x28a48742a8bc98364dd99882ff1f3b18c69a682c7758091aefe0d15f453e3357/activity

halfprice:
Fix #982

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -813,8 +813,8 @@ impl StorageNode {
         }
 
         if shard_diff.gained.is_empty() {
-            let node_in_committee = committees.current_committee().contains(public_key);
-            if node_in_committee && committees.epoch() == event.epoch {
+            let is_node_in_committee = committees.current_committee().contains(public_key);
+            if is_node_in_committee && committees.epoch() == event.epoch {
                 // We are in the current committee, but no shards were gained. Directly signal that
                 // the epoch sync is done.
                 tracing::info!("no shards gained, so signalling that epoch sync is done");
@@ -829,7 +829,7 @@ impl StorageNode {
                 tracing::info!(
                     "skip sending epoch sync done event. \
                     node in committee: {}, committee epoch: {}, event epoch: {}",
-                    node_in_committee,
+                    is_node_in_committee,
                     committees.epoch(),
                     event.epoch
                 );
@@ -3157,7 +3157,12 @@ mod tests {
         contract_service.expect_epoch_sync_done().never();
         contract_service
             .expect_fixed_system_parameters()
-            .returning(|| Ok(FixedSystemParameters::default()));
+            .returning(|| {
+                Ok(FixedSystemParameters {
+                    epoch_duration: Duration::from_secs(600),
+                    epoch_zero_end: Utc::now() + Duration::from_secs(60),
+                })
+            });
         contract_service
             .expect_get_epoch_and_state()
             .returning(move || Ok((0, EpochState::EpochChangeDone(Utc::now()))));
--- a/crates/walrus-service/src/node/committee.rs
+++ b/crates/walrus-service/src/node/committee.rs
@@ -139,7 +139,6 @@ pub enum EndCommitteeChangeError {
 /// with committee members.
 ///
 /// It is associated with a single storage epoch.
-#[cfg_attr(test, mockall::automock)]
 #[async_trait]
 pub trait CommitteeService: std::fmt::Debug + Send + Sync {
     /// Returns the epoch associated with the committee.
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -15,6 +15,7 @@ use std::{
 };
 
 use async_trait::async_trait;
+use chrono::Utc;
 use futures::{stream::FuturesUnordered, StreamExt};
 use prometheus::Registry;
 use sui_types::base_types::ObjectID;
@@ -318,7 +319,10 @@ impl StorageNodeHandleBuilder {
 
         let contract_service = self.contract_service.unwrap_or_else(|| {
             Box::new(StubContractService {
-                system_parameters: FixedSystemParameters::default(),
+                system_parameters: FixedSystemParameters {
+                    epoch_duration: Duration::from_secs(600),
+                    epoch_zero_end: Utc::now() + Duration::from_secs(60),
+                },
             })
         });
 
--- a/crates/walrus-service/tests/epoch_change.rs
+++ b/crates/walrus-service/tests/epoch_change.rs
@@ -17,7 +17,7 @@ async fn nodes_drive_epoch_change() -> TestResult {
         test_cluster::default_setup_with_epoch_duration(epoch_duration).await?;
 
     let target_epoch: Epoch = 4;
-    // Allow time to reach the desired epoch, with an additional 20% for the jitter.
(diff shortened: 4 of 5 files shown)


crates/walrus-service/src/test_utils.rs
```diff
@@ -666,6 +668,25 @@ impl SystemContractService for StubContractService {
...
+        Self {
+            system_parameters: FixedSystemParameters {
+                epoch_duration: Duration::from_secs(600),
+                epoch_zero_end: Utc::now() + Duration::from_secs(60),
+            },
+            epoch_sync_done_called: Arc::new(AtomicU32::new(0)),
+        }
+    }
+}
+
```
jpcsmith: Suggestion: Instead of adding a function to the `StubContractService` to meet the needs of a single test, see instead the usage of `MockSystemContractService` in the `epoch_change_driver` module. You can create an instance that does this count in the test case.

crates/walrus-service/src/node.rs
```diff
@@ -813,12 +813,21 @@ impl StorageNode {
...
+                tracing::info!("no shards gained, so signalling that epoch sync is done");
+                self.inner
+                    .contract_service
+                    .epoch_sync_done(event.epoch)
+                    .await;
+            } else {
+                tracing::info!(
+                    "no shards gained and not in the new committee, finishing epoch change start"
+                );
+            }
```
jpcsmith: Remark: We also schedule operations on startup, so we need to add a check there as well.

crates/walrus-service/src/test_utils.rs
halfprice: Excellent! That is very handy.

I tried to use MockCommitteeService as well, but I only need to mock one call, get_epoch(), in the test. The complexity of adding a mock call for all the functions lead me towards just initialize the StorageNodeHandleBuilder with a different initial epoch.

crates/walrus-service/src/node.rs
halfprice: Do you mean the `epoch_sync_done` call? It is only reactive upon receiving epoch change start event, isn't?

Or do you mean the call scheduled in `schedule_relevant_calls_for_current_epoch`? Those calls seem to be general calls and not specific for the node itself. So not sure exactly what to check there. The committee service is also the most recent committee service fetched upon starting I assume.

crates/walrus-service/tests/epoch_change.rs
```diff
@@ -11,13 +11,14 @@ use walrus_test_utils::Result as TestResult;
...
 async fn nodes_drive_epoch_change() -> TestResult {
+    let _ = tracing_subscriber::fmt::try_init();
     let epoch_duration = Duration::from_secs(10);
     let (_sui, storage_nodes, _) =
         test_cluster::default_setup_with_epoch_duration(epoch_duration).await?;
 
     let target_epoch: Epoch = 4;
     // Allow time to reach the desired epoch, with an additional 20% for the jitter.
-    let time_to_reach_epoch = epoch_duration * target_epoch * 12 / 10;
+    let time_to_reach_epoch = epoch_duration * target_epoch * 15 / 10;
```
halfprice, line 21: Local run never had problem. In CI is very flaky, so increase the timeout to make it more tolerable.

crates/walrus-service/src/node.rs
jpcsmith: ~Only a node in the current committee should be calling events, so a node should not call `schedule_relevant_calls_for_current_epoch()` if, on startup, it is not in the current committee.~

I misunderstood the issue, I thought it was with the recently added scheduling of calls as opposed to the epoch sync done call, you can ignore this comment.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -84,6 +84,15 @@ pub struct FixedSystemParameters {
...
 
+impl Default for FixedSystemParameters {
+    fn default() -> Self {
+        Self {
+            epoch_duration: Duration::from_secs(600),
+            epoch_zero_end: Utc::now() + Duration::from_secs(60),
+        }
+    }
+}
+
```
jpcsmith: Important: Do not implement a default for this as it is determined from the contract. This gives the invalid impression that this could be valid for a user of the `read_client`. If you need one for testing, add a test-specific constructor to one of the `test_utils` modules.

crates/walrus-service/tests/epoch_change.rs
jpcsmith: Okay, but change the doc-comment as well from 20% to 50%.

crates/walrus-service/src/test_utils.rs
```diff
@@ -269,6 +269,14 @@ impl StorageNodeHandleBuilder {
...
     }
 
+    /// Specify the initial epoch for the node.
+    ///
+    /// If specified, the committee service will be created with the provided epoch.
+    pub fn with_initial_epoch(mut self, epoch: Option<Epoch>) -> Self {
+        self.initial_epoch = epoch;
+        self
+    }
+
```
jpcsmith, line 280: Remark: This has the problem that if the node does not also have the appropriate events in the event stream, then there's no way for it to actually advance to match the committee epoch. The workaround I had added (in one of the integration tests) was to manually advance the epoch state and generate the appropriate events.

crates/walrus-service/src/node/committee.rs
```diff
@@ -139,6 +139,7 @@ pub enum EndCommitteeChangeError {
 /// with committee members.
 ///
 /// It is associated with a single storage epoch.
+#[cfg_attr(test, mockall::automock)]
 #[async_trait]
 pub trait CommitteeService: std::fmt::Debug + Send + Sync {
```
jpcsmith: Remark: This does not seem to be used.

crates/walrus-service/src/node.rs
```diff
@@ -813,12 +813,29 @@ impl StorageNode {
         }
 
         if shard_diff.gained.is_empty() {
-            tracing::debug!("no shards gained, so signalling that epoch sync is done");
-            self.inner
-                .contract_service
-                .epoch_sync_done(event.epoch)
-                .await;
+            let node_in_committee = committees.current_committee().contains(public_key);
```
jpcsmith: Nit:
```suggestion
            let is_node_in_committee = committees.current_committee().contains(public_key);
```

crates/walrus-service/src/test_utils.rs
halfprice: I think the workaround is mainly for when using the test cluster, where we have to start from epoch 0 and march to the desired epoch. Here is for the test that tests a single storage node object. For case like this, we use the StubCommitteeService, so it is already not moving by itself even with proper events.

crates/walrus-service/src/node/committee.rs
halfprice: Yes, had a try with it and bailed...

crates/walrus-sui/src/client/read_client.rs
halfprice: Good point. Did not look closely how `FixedSystemParameters` is actually used. I removed the default and use the fixed value where they are needed.

crates/walrus-service/tests/epoch_change.rs
halfprice: Done

PR #989, opened 2024-10-08
refactor: use the default port when registering

jpcsmith:


crates/walrus-service/bin/node.rs
```diff
@@ -396,10 +402,18 @@ mod commands {
...
-        }
+        // If we have an IP address, use a SocketAddr to get the string representation
+        // as IPv6 addresses are enclosed in square brackets.
+        let public_address = if let Ok(ip_addr) = IpAddr::from_str(&public_host) {
+            NetworkAddress(SocketAddr::new(ip_addr, port).to_string())
+        } else {
+            // Do a minor sanity check that the user has not included a port in the hostname
+            ensure!(
+                !public_host.contains(':'),
+                "DNS names must not contain ':', did you include a port?"
```
karlwuest: Nit: Instead of asking "did you include a port?", it may be more helpful to say "to specify a port, use the --port option"

crates/walrus-service/bin/node.rs
```diff
@@ -396,10 +402,18 @@ mod commands {
...
+            );
+            NetworkAddress(public_host)
```
karlwuest: Important: This should include the port.
```suggestion
            NetworkAddress(format!("{public_host}:{port}"))
```

crates/walrus-service/bin/node.rs
jpcsmith:
```suggestion
                "DNS names must not contain ':', to specify a port, use the --port option."
```

PR #990, opened 2024-10-08
feat: Add event processor config to dry run config

sadhansood:
As title says, this PR adds event processor config to the storage node config in dry run config generation

### DIFF
--- a/crates/walrus-service/bin/deploy.rs
+++ b/crates/walrus-service/bin/deploy.rs
@@ -129,6 +129,10 @@ struct GenerateDryRunConfigsArgs {
     /// sending another request.
     #[clap(long)]
     faucet_cooldown: Option<Duration>,
+    /// Enable checkpoint based event processor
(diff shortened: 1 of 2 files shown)


crates/walrus-service/bin/deploy.rs
```diff
@@ -129,6 +129,10 @@ struct GenerateDryRunConfigsArgs {
     /// sending another request.
     #[clap(long)]
     faucet_cooldown: Option<Duration>,
+    /// Enable checkpoint based event processor
+    /// [default: false]
+    #[clap(long, default_value = "false")]
```
halfprice: Nit: default_value_t = false to avoid conversion

mlegner: I think `action` would be even better:
```rust
    /// Enable checkpoint based event processor.
    #[clap(long, action)]
```

PR #991, opened 2024-10-08
feat: add simple human-readable parsing for storage capacity

jpcsmith:


crates/walrus-service/bin/node.rs
```diff
@@ -180,7 +180,11 @@ struct ConfigArgs {
     staking_object: ObjectID,
     #[clap(long)]
     /// Initial storage capacity of this node in bytes.
-    node_capacity: u64,
+    ///
+    /// The value can either by unitless; have suffixes for powers of 1000, such as (B),
+    /// kilobytes (K), etc, or have suffixes for the IEC units such as kibibytes (Ki),
+    /// mebibytes (Mi), etc.
```
karlwuest, line 186: Remark: I would expect most users to specify e.g. "KiB", or "TB" instead of just "Ki" or "B", see comment on `ByteCount`

crates/walrus-service/src/common/utils.rs
```diff
@@ -528,9 +529,61 @@ pub fn load_wallet_context(path: &Option<PathBuf>) -> Result<WalletContext> {
...
+
+impl FromStr for ByteCount {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> std::result::Result<Self, Self::Err> {
+        // ensure!(s.is_ascii(), "only ascii characters allowed in byte sizes");
+        if let Some(value_prefix) = s.strip_suffix("B") {
+            // Bytes cannot have fractional components
+            return Ok(ByteCount(u64::from_str(value_prefix.trim())?));
+        }
```
karlwuest: Important: This will return an error if the input is e.g. `1TB`.
Suggestion: just strip the suffix and continue.

```suggestion
        s.strip_suffix("B");
```

crates/walrus-service/src/common/utils.rs
```diff
@@ -580,4 +633,29 @@ mod tests {
...
+                kibi: ("3.5Ki", (3.5 * 1024.0) as u64),
+                mega: ("1.93M", (1.93 * 1e6) as u64),
+                mebi: ("1478Mi", 1478 * 1024 * 1024),
+                giga: ("21G", (21.0 * 1e9) as u64),
+                gibi: ("21.791Gi", (21.791 * 1024.0 * 1024.0* 1024.0) as u64),
+                tera: ("7.4T", (7.4 * 1e12) as u64),
+                tebi: ("19Ti", 19 * 1024 * 1024* 1024 * 1024),
+                peta: ("21000P", (21000.0 * 1e15) as u64),
+                pebi: ("21.721Pi", (21.721 * 1024.0 * 1024.0 * 1024.0 * 1024.0 * 1024.0) as u64),
+                with_space: ("1.489 M", (1.489 * 1e6) as u64),
```
karlwuest:
```suggestion
                with_space: ("1.489 M", (1.489 * 1e6) as u64),
                kilobyte: ("2KB", 2000),
                kibibyte: ("3.5KiB", (3.5 * 1024.0) as u64),
                megabyte: ("1.93MB", (1.93 * 1e6) as u64),
                mebibyte: ("1478MiB", 1478 * 1024 * 1024),
                gigabyte: ("21GB", (21.0 * 1e9) as u64),
                gibibyte: ("21.791GiB", (21.791 * 1024.0 * 1024.0* 1024.0) as u64),
                terabyte: ("7.4TB", (7.4 * 1e12) as u64),
                tebibyte: ("19TiB", 19 * 1024 * 1024* 1024 * 1024),
                petabyte: ("21000PB", (21000.0 * 1e15) as u64),
                pebibyte: ("21.721PiB", (21.721 * 1024.0 * 1024.0 * 1024.0 * 1024.0 * 1024.0) as u64),
```

crates/walrus-service/bin/node.rs
jpcsmith: I probably should not speak to "most users", but the goal was to match the standard linux filesystem tools, `ls`, `df`, `du`, etc., which use [numfmt](https://www.man7.org/linux/man-pages/man1/numfmt.1.html) and work with 1 or 2 letter suffixes.

But I can add support for the 3-letter suffixes.

karlwuest: I should also not speak for "most users", but I personally would expect it not to return an error if I entered "TB" and would probably enter it this way 😅
Thanks for adding support for this!

jpcsmith: (Was in reference to the "they will certainly use DNS names!" lesson from yesterday). Added support and more tests.

PR #992, opened 2024-10-09
test: simtest with storage node crash and restart

### LINKED ISSUE #356: Improve and extend integration/E2E tests with the simtest framework

halfprice:
This PR adds crash-recovery ability to walrus simtest. To achieve this, the main effort involves making each walrus node
run in a dedicated simulated runtime (node). After this, we can shut down and restart sim node in tests to test
crash and recovery scenarios.

I added one simple test with one single node failure to illustrate the usage.

Sample log showing crash-restart:

```
2022-10-03T19:50:18.052229Z  WARN node{id=10 name="rr8rjxq" ... crates/walrus-simtest/tests/simtest.rs:160: Crashing node Node(10) for 19 seconds
2022-10-03T19:50:18.052229Z  INFO node{id=10 name="rr8rjxq/" ... msim/src/sim/task.rs:128: killing node Node(10). Will restart in 19s
2022-10-03T19:50:37.092046Z  WARN node{id=10 name="rr8rjxq" ... crates/typed-store/src/metrics.rs:989: DBMetrics registry overwritten
2022-10-03T19:50:37.201680Z  INFO node{id=10 name="rr8rjxq" ... crates/walrus-service/src/node/storage.rs:74: Recovers storage from "/var/folders/b2/259h62m53_g6wbtjz_htk0mh0000gn/T/tmp.1dIgUffQp3/.tmpoyrUFQ"
2022-10-03T19:50:37.280030Z  INFO node{id=10 name="rr8rjxq" ... walrus-service/src/node/system_events.rs:84: resuming from event cursor=EventStreamCursor { event_id: Some(EventID { tx_digest: TransactionDigest(H1xTHGzpvRevjJXjFHRru6r1r8XwBC9Sw5hVksmR7pwC), event_seq: 0 }), element_index: 7 }
```

Close #356

### DIFF
--- a/crates/walrus-simtest/tests/simtest.rs
+++ b/crates/walrus-simtest/tests/simtest.rs
@@ -163,6 +163,8 @@ mod tests {
             }
         })
         .await;
+
+        assert!(fail_triggered.load(std::sync::atomic::Ordering::SeqCst));
     }
 
     // Action taken during a various failpoints. There is a chance with `probability` that the


crates/walrus-service/src/test_utils.rs
```diff
@@ -401,6 +538,213 @@ impl StorageNodeHandleBuilder {
...
+    async fn build_and_run_node(
+        config: StorageNodeConfig,
+        cancel_token: CancellationToken,
+    ) -> anyhow::Result<(
+        tokio::task::JoinHandle<Result<(), std::io::Error>>,
+        tokio::task::JoinHandle<Result<(), anyhow::Error>>,
+        tokio::task::JoinHandle<Result<(), anyhow::Error>>,
+    )> {
+        // TODO(#996): extract the common code to start the code, and use it here as well as in
+        // StorageNodeRuntime::start.
```
halfprice: @jpcsmith regarding moving the start of a node to a common function, I spent a long time reorg the code, and it became very messy due to that all the spawned thread in node.rs are currently not in the same place. Some clean up there is needed first. So I added this TODO as a next step.

crates/walrus-service/src/test_utils.rs
```diff
@@ -401,6 +538,213 @@ impl StorageNodeHandleBuilder {
...
+                        &config.storage_path,
+                    )
+                    .await?,
+                )
+            } else {
+                Box::new(crate::node::system_events::SuiSystemEventProvider::new(
+                    sui_read_client.clone(),
+                    Duration::from_millis(100),
+                ))
+            };
```
halfprice: @sadhansood can you double check this part? This is to create event processor based on event_processor_config

crates/walrus-service/src/test_utils.rs
```diff
@@ -161,6 +227,77 @@ impl AsRef<StorageNode> for StorageNodeHandle {
...
+
+impl StorageNodeHandleTrait for SimStorageNodeHandle {
+    fn cancel(&self) {
+        unimplemented!("cannot directly cancel a storage in proactively.")
+    }
+
+    fn client(&self) -> &Client {
+        // Node client may change everytime the node restarts.
+        unimplemented!("simulation test does not have a stable client.")
+    }
```
jpcsmith: Question: What about the client need's to be `Stable`?

crates/walrus-service/src/test_utils.rs
```diff
@@ -161,6 +227,77 @@ impl AsRef<StorageNode> for StorageNodeHandle {
...
+    /// Cancellation token for the node.
+    pub cancel_token: CancellationToken,
+    /// The wrapped simulator node id.
+    #[cfg(msim)]
+    pub node_id: sui_simulator::task::NodeId,
+}
+
+impl StorageNodeHandleTrait for SimStorageNodeHandle {
+    fn cancel(&self) {
+        unimplemented!("cannot directly cancel a storage in proactively.")
```
```suggestion
        unimplemented!("cannot directly cancel a storage proactively.")
```

crates/walrus-service/src/test_utils.rs
```diff
@@ -401,6 +538,213 @@ impl StorageNodeHandleBuilder {
             client,
         })
     }
+
+    /// Starts and running a storage node with the provided configuration.
+    #[allow(unused_variables)]
+    pub async fn start_node(
```
jpcsmith: Question: Should this also be `cfg(msim)` given that it returns a SimStorageNodeHandle?

crates/walrus-service/src/test_utils.rs
```diff
@@ -401,6 +538,213 @@ impl StorageNodeHandleBuilder {
...
+        })
+    }
+
+    // Starts and runs a storage node with the provided configuration in a dedicated simulator
+    // node.
+    #[cfg(msim)]
+    async fn spawn_node(
+        config: StorageNodeConfig,
+        cancel_token: CancellationToken,
+    ) -> sui_simulator::runtime::NodeHandle {
```
jpcsmith: Nit: move these to their own implementation block.
```rust
#[cfg(msim)]
impl StorageNodeHandle {
```

crates/walrus-service/src/test_utils.rs
```diff
@@ -725,28 +1083,24 @@ impl SystemEventProvider for tokio::sync::broadcast::Sender<ContractEvent> {
 
 /// A cluster of [`StorageNodeHandle`]s corresponding to several running storage nodes.
 #[derive(Debug)]
-pub struct TestCluster {
+pub struct TestCluster<T: StorageNodeHandleTrait> {
```
jpcsmith: Remark: Set the default as `T: StorageNodeHandleTrait = StorageNodeHandle`, then remove the generic argument `TestCluster<StorageNodeHandle>` from the legacy code.

crates/walrus-service/src/test_utils.rs
```diff
@@ -1037,27 +1444,6 @@ where
...
-            committee_services: shard_assignment.iter().map(|_| None).collect(),
-            contract_services: shard_assignment.iter().map(|_| None).collect(),
-            storage_node_configs: shard_assignment
-                .into_iter()
-                .map(StorageNodeTestConfig::new)
-                .collect(),
-        }
-    }
-}
-
```
jpcsmith: Suggestion: Keep the default.

crates/walrus-service/tests/epoch_change.rs
```diff
@@ -5,15 +5,16 @@ use std::time::Duration;
...
 use walrus_test_utils::Result as TestResult;
 
 #[ignore = "ignore E2E tests by default"]
 #[tokio::test]
 async fn nodes_drive_epoch_change() -> TestResult {
     let epoch_duration = Duration::from_secs(10);
     let (_sui, storage_nodes, _) =
-        test_cluster::default_setup_with_epoch_duration(epoch_duration).await?;
+        test_cluster::default_setup_with_epoch_duration::<StorageNodeHandle>(epoch_duration)
+            .await?;
```
jpcsmith: Suggestion: Keep the functions `default_setup` and `default_setup_with_epoch_duration` as using `StorageNodeHandle`. Add a function `default_setup_X` that takes a generic argument, and have the above functions delegate to it.

Use that function in test codes that require something other than a `StorageNodeHandle`.

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -1,58 +1,171 @@
...
+            &[
+                "batch-write-before",
+                "batch-write-after",
+                "put-cf-before",
+                "put-cf-after",
+                "delete-cf-before",
+                "delete-cf-after",
+            ],
+            move || {
+                handle_failpoint(do_not_fail_nodes.clone(), fail_triggered.clone(), 0.01);
```
sadhansood: @halfprice is the decision to crash the node evaluated at every time code hits one of the failpoints? If so, do we know at the end of the test if any of the nodes were crashed or not (or how many times)? Since (it looks like) this is controlled via random number seed - does it even make sense to run this test deterministically with a version of the random number seed which we know will cause at least one failure, at least `N` failures, etc?

crates/walrus-service/src/test_utils.rs
```diff
@@ -401,6 +538,213 @@ impl StorageNodeHandleBuilder {
...
+        let startup_sender = Arc::new(startup_sender);
+        let handle = sui_simulator::runtime::Handle::current();
+        let builder = handle.create_node();
+
+        // This is the entry point of node restart.
+        let node_handle = builder
+            .ip(ip)
+            .name(&format!(
+                "{:?}",
+                config.protocol_key_pair.get().unwrap().public()
```
sadhansood: nit (here and in some other places): Maybe add `context` to the unwrap or use `expect`

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -1,58 +1,171 @@
...
+        // Write a random blob.
+        let blob = walrus_test_utils::random_data(data_length);
+        let BlobStoreResult::NewlyCreated {
+            blob_object: blob_confirmation,
+            ..
+        } = client
+            .as_ref()
+            .reserve_and_store_blob(&blob, 1, StoreWhen::Always, BlobPersistence::Permanent)
+            .await
+            .unwrap()
```
sadhansood: nit: `could we use `?` instead of `unwrap` here

crates/walrus-service/src/test_utils.rs
sadhansood: Yes, this looks right to me!

crates/walrus-simtest/tests/simtest.rs
sadhansood: nit: maybe make the `0.01` a const variable, longer term maybe we will even benefit from making this a test variable and run this test with differetn probablilities

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -1,58 +1,171 @@
...
+                "put-cf-after",
+                "delete-cf-before",
+                "delete-cf-after",
+            ],
+            move || {
+                handle_failpoint(do_not_fail_nodes.clone(), fail_triggered.clone(), 0.01);
+            },
+        );
+
+        let _ = tokio::time::timeout(Duration::from_secs(60), async {
```
sadhansood, line 155: suggestion: instead of timeout, would it be better to have a fixed number of test iterations to run? I am saying so because under contention sometime tokio tasks may not get scheduled by the runtime and we may not get complete signal from this test (if it didn't run enough number of iterations)

crates/walrus-service/src/test_utils.rs
```diff
@@ -161,6 +227,77 @@ impl AsRef<StorageNode> for StorageNodeHandle {
...
+    /// The address of the metric service.
+    pub metrics_address: SocketAddr,
+    /// Cancellation token for the node.
+    pub cancel_token: CancellationToken,
+    /// The wrapped simulator node id.
+    #[cfg(msim)]
+    pub node_id: sui_simulator::task::NodeId,
+}
+
+impl StorageNodeHandleTrait for SimStorageNodeHandle {
```
sadhansood, line 420: nit: i think maybe some extra documentation here will help readers understand why these functions are unimplemented. It seems like because nodes are changing due to crash and restart (adding that as a doc string will be nice)

crates/walrus-service/src/test_utils.rs
halfprice: Hmm, looking closely, the client is just created using the node's address, which will remain the same over the time. So I guess we could create a client. However, this exposed client is not used in simtest, so I updated the comment that this is not needed.

crates/walrus-service/src/test_utils.rs
halfprice: Yes. This was missed because the callsite was a `cfg(msim)`, so this was anyway not built in simtest.

I made a pass to make all Sim related code inside `cfg(msim)` including SimStorageNodeHandle itself.

crates/walrus-service/src/test_utils.rs
halfprice: Done.

crates/walrus-service/src/test_utils.rs
halfprice: Done.

crates/walrus-service/src/test_utils.rs
halfprice: I added the default back, and use the default in TestClusterBuilder::new()

crates/walrus-service/tests/epoch_change.rs
halfprice: Good idea. Created `default_setup_with_epoch_duration_generic` for the generic one.

crates/walrus-simtest/tests/simtest.rs
halfprice: > is the decision to crash the node evaluated at every time code hits one of the failpoints

yes

> do we know at the end of the test if any of the nodes were crashed or not (or how many times)

Right now, we only crash one node once. I made some change in the test that we will wait until that crash happen before finishing the test.

> does it even make sense to run this test deterministically with a version of the random number seed which we know will cause at least one failure, at least N failures, etc?

Not sure exactly what do you mean. MSIM_TEST_NUM can be used to change the test seed, which will change the RNG. Also, every time the code is updated, it will produce a different execution process (since tokio scheduler's tasks are different).

crates/walrus-service/src/test_utils.rs
halfprice: Done

crates/walrus-simtest/tests/simtest.rs
halfprice: Done (added a CONST)

crates/walrus-service/src/test_utils.rs
halfprice: I added a bit more doc here. There are more docs in StorageNodeHandleTrait to explain. Let me know if this is still not clear.

crates/walrus-simtest/tests/simtest.rs
halfprice: Done

crates/walrus-simtest/tests/simtest.rs
halfprice: The reason I'm using timeout based here is because we also restart the node after sometime. We don't want the workload to finish too fast. It's just easier to control the timeout here and the timeout of node restart

sadhansood: ok makes sense, could we assert in the end that we had at least one crash. Sorry for being paranoid, but wondering if there's a possibility that the test could pass without actually triggering a crash if the probability is set too low or by chance which could lead to false confidence in the system's resilience.

halfprice: The way the test works right now is, `start workload` -> `wait until crash happens` -> `run the workload for another minute`. So the crash should have happened when test finishes. But there is no harm to add an assertion that the failure is triggered (I have added it).

PR #1001, opened 2024-10-09
fix: add node to active set if stake is added after voting

karlwuest:


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -342,9 +342,14 @@ public(package) fun stake_with_pool(
...
-    // whether the stake was applied to E+1 or E+2.
-    self.active_set.insert_or_update(node_id, pool.wal_balance_at_epoch(wctx.epoch() + 1));
+    // Active set only tracks the stake for the next vote, which either happens for the committee
+    // in wctx.epoch() + 1, or in wctx.epoch() + 2, depending on whether the vote already happened.
+    match (self.epoch_state) {
+        EpochState::NextParamsSelected(_) => self
+            .active_set
+            .insert_or_update(node_id, pool.wal_balance_at_epoch(wctx.epoch() + 2)),
+        _ => self.active_set.insert_or_update(node_id, pool.wal_balance_at_epoch(wctx.epoch() + 1)),
+    };
```
jpcsmith: Nit: This could be made clearer by having the match return either `wctx.epoch() + 2` or `wctx.epoch() + 1`. It would then be obvious that the condition that varies is which epoch we're staking with.

PR #1003, opened 2024-10-09
feat: specify max_epochs_ahead and initial epoch duration in deployment

karlwuest:


crates/walrus-sui/src/system_setup.rs
```diff
@@ -139,25 +139,51 @@ pub async fn publish_coin_and_system_package(
     Ok((walrus_pkg_id, init_cap_id, treasury_cap_id))
 }
 
+/// Parameters used to call the `init_walrus` function in the Walrus contracts.
+#[derive(Debug, Clone, Copy)]
+pub struct InitSystemParams {
+    /// Number of shards in the system.
+    pub n_shards: u16,
```
jpcsmith: Suggestion: Make `NonZero<u16>`.

crates/walrus-sui/src/system_setup.rs
```diff
@@ -139,25 +139,51 @@ pub async fn publish_coin_and_system_package(
...
     init_cap: ObjectID,
-    n_shards: u16,
-    epoch_zero_duration_ms: u64,
-    epoch_duration_ms: u64,
+    InitSystemParams {
+        n_shards,
+        epoch_zero_duration,
+        epoch_duration,
+        max_epochs_ahead,
+    }: InitSystemParams,
```
jpcsmith: Tiny nit: Prefer `params.n_shards`, etc instead of adding everything to the parent's scope.

PR #1011, opened 2024-10-09
feat(client): add command to generate new Sui wallet

### LINKED ISSUE #998: Add `setup-wallet` command to `walrus` binary to simplify setup for users

mlegner:
Closes #998.

### DIFF
--- a/crates/walrus-service/src/client/cli/runner.rs
+++ b/crates/walrus-service/src/client/cli/runner.rs
@@ -154,8 +154,12 @@ impl ClientCommandRunner {
                 self.stake_with_node_pool(node_id, amount).await
             }
 
-            CliCommands::GenerateSuiWallet { path, sui_network } => {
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -154,6 +154,11 @@ impl ClientCommandRunner {
                 self.stake_with_node_pool(node_id, amount).await
             }
 
+            CliCommands::GenerateSuiWallet { path, sui_network } => {
+                generate_sui_wallet(sui_network, &path, Duration::from_secs(20)).await?;
```
karlwuest: Nit (also applies above): 20 seconds is quite short for the faucet, it often takes almost 1 minute on testnet.

mlegner: Good point. I made the timeout configurable and increased the default to 1 minute in 68b4c33747275d64832b8a1ed59ef5cee1ee9144.

PR #1021, opened 2024-10-09
feat: Add metrics to event processor

sadhansood:
Add metrics to track latest downloaded and total number of downloaded checkpoints in event processor.

### DIFF
--- a/crates/walrus-event/src/event_processor.rs
+++ b/crates/walrus-event/src/event_processor.rs
@@ -9,6 +9,13 @@ use move_core_types::{
     account_address::AccountAddress,
     annotated_value::{MoveDatatypeLayout, MoveTypeLayout},
 };
+use prometheus::{
+    register_int_counter_with_registry,
+    register_int_gauge_with_registry,
+    IntCounter,
+    IntGauge,
+    Registry,
+};
 use rocksdb::Options;
 use sui_package_resolver::{
     error::Error as PackageResolverError,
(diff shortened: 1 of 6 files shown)


crates/walrus-event/src/event_processor.rs
```diff
@@ -105,6 +121,27 @@ pub struct EventProcessor {
...
+            .expect("this is a valid metrics registration"),
+            total_downloaded_checkpoints: register_int_counter_with_registry!(
+                "total_downloaded_checkpoints",
+                "Total number of checkpoints downloaded",
+                registry,
+            )
+            .expect("this is a valid metrics registration"),
+        }
+    }
 }
```
jpcsmith: Suggestion: Prefix these metrics with `event_processor_`. So `event_processor_latest_downloaded_checkpoint`, and `event_processor_total_downloaded_checkpoints`, otherwise they will be a pain to find in Grafana.

The name of the struct fields can remain unchanged.

crates/walrus-service/bin/node.rs
```diff
@@ -609,6 +612,7 @@ impl EventProcessorRuntime {
         sui_config: SuiConfig,
         event_processor_config: Option<EventProcessorConfig>,
         db_path: &Path,
+        metrics_registry: Registry,
```
jpcsmith: Question: Why does this need to be owned?

crates/walrus-event/src/event_processor.rs
sadhansood: makes sense

crates/walrus-service/bin/node.rs
sadhansood: you are right, my nad - this could just be a ref

PR #1022, opened 2024-10-09
feat: report committee metrics in daemon

### LINKED ISSUE #921: Export binary version and current epoch and display on Grafana
This applies primarily to the storage nodes but also to aggregator and publisher.

jpcsmith:
Closes #921

### DIFF
--- a/crates/walrus-service/src/node/committee/committee_service.rs
+++ b/crates/walrus-service/src/node/committee/committee_service.rs
@@ -335,7 +335,7 @@ where
                 Ok(outgoing) => {
                     let outgoing = outgoing.clone();
                     self.inner.record_epoch_change_metrics(tracker.committees());
-                    Some((outgoing.clone(), current))
+                    Some((outgoing, current))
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client.rs
```diff
@@ -165,7 +172,31 @@ impl Client<()> {
...
+        if let Ok(committees_guard) = self.committees.try_read() {
+            metrics.current_epoch.set(committees_guard.epoch());
+
+            // NOTE(jsmith): With the current ActiveCommittees we do not track whether voting has
+            // ended, however, these states are sufficient here as they affect the committee used.
+            if committees_guard.is_change_in_progress() {
+                metrics.current_epoch_state.set_change_sync_state();
+            } else {
+                metrics.current_epoch_state.set_change_done_state();
+            }
```
giac-mysten: Question: could we use the presence of `committees.next_committee` as signal that the voting has ended?

PR #1025, opened 2024-10-10
feat: export information about the contract addresses

jpcsmith:


crates/walrus-service/src/common/utils.rs
```diff
@@ -629,6 +630,34 @@ pub fn export_build_info(registry: &Registry, version: &'static str) {
...
+        opts,
+        &["system_object", "staking_object", "active_address"],
+        registry
+    )
+    .expect("static metric is valid");
+
+    metric
+        .get_metric_with_label_values(&[
+            &system_object.to_hex_uncompressed(),
+            &staking_object.to_hex_uncompressed(),
```
giac-mysten, line 652: Question: why uncompressed?

jpcsmith: I didn't see a `to_hex()` option and this seems to just give the full, fixed with hex string.

PR #1032, opened 2024-10-10
fix!: ensure that correct nodes are evicted from ActiveSet

karlwuest:
This fixes several issues with the sorting (wrong indices in `idx_sorted` and indices were not updated if they changed due to a node being removed) in `ActiveSet` that we only use to find the minimum stake by removing the sorting and just finding the minimum when we need it. This probably also makes it a bit more efficient, because it removes a lot of the iterations that the `VecMap` lookups do.

Breaking changes: changes Move struct

### DIFF
--- a/contracts/walrus/sources/staking/active_set.move
+++ b/contracts/walrus/sources/staking/active_set.move
@@ -104,7 +104,7 @@ public(package) fun insert(set: &mut ActiveSet, node_id: ID, staked_amount: u64)
         // If there is such a node, replace it in the list.
         if (min_idx.is_some()) {
             let min_idx = min_idx.extract();
-            set.total_stake = set.total_stake - set.nodes[min_idx].staked_amount + staked_amount;
+            set.total_stake = set.total_stake - min_stake + staked_amount;
             *&mut set.nodes[min_idx] = ActiveSetEntry { node_id, staked_amount };
             true
         } else {
@@ -170,11 +170,11 @@ public(package) fun cur_min_stake(set: &ActiveSet): u64 {
     }
 }
 
-// #[syntax(index)]
-// /// Get the staked amount of the storage node with the given `node_id`.
-// public(package) fun borrow(set: &ActiveSet, node_id: &ID): &u64 {
-//     &set.nodes[node_id]
-// }
+#[test_only]
+public fun stake_for_node(set: &ActiveSet, node_id: ID): u64 {
+    set.nodes.find_index!(|entry| entry.node_id == node_id)
+        .map!(|index| set.nodes[index].staked_amount ).destroy_with_default(0)
+}
 
 // === Test ===
 
@@ -187,8 +187,16 @@ fun test_evict_correct_node_simple() {
     set.insert_or_update(object::id_from_address(@4), 7);
     set.insert_or_update(object::id_from_address(@5), 6);
 
+    let mut total_stake = 10 + 9 + 8 + 7 + 6;
+
+    assert!(set.total_stake == total_stake);
+
     // insert another node which should eject node 5
     set.insert_or_update(object::id_from_address(@6), 11);
+
+    // check if total stake was updated correctly
+    total_stake = total_stake - 6 + 11;
+    assert!(set.total_stake == total_stake);
 
     let active_ids = set.active_ids();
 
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/staking/active_set.move
```diff
@@ -49,33 +47,32 @@ public(package) fun new(max_size: u16, threshold_stake: u64): ActiveSet {
...
+public(package) fun update(set: &mut ActiveSet, node_id: ID, staked_amount: u64): bool {
+    let index = set.nodes.find_index!(|entry| entry.node_id == node_id);
+    if (index.is_none()) {
+        return false
+    };
+    index.do!(|idx| {
+        set.total_stake = set.total_stake - set.nodes[idx].staked_amount + staked_amount;
+        set.nodes[idx].staked_amount = staked_amount;
+    });
+    true
```
mlegner, line 75: Question: Is there some way to replace this with a `match(index)`?

Probably not as the standard `Option` doesn't seem to be an enum (yet).

contracts/walrus/sources/staking/active_set.move
```diff
@@ -84,71 +81,68 @@ public(package) fun update(set: &mut ActiveSet, node_id: ID, staked_amount: u64)
...
+        set.nodes.length().do!(|i| {
+            if (set.nodes[i].staked_amount < min_stake) {
+                min_idx = option::some(i);
+                min_stake = set.nodes[i].staked_amount;
+            }
+        });
+        // If there is such a node, replace it in the list.
+        if (min_idx.is_some()) {
+            let min_idx = min_idx.extract();
+            set.total_stake = set.total_stake - set.nodes[min_idx].staked_amount + staked_amount;
```
mlegner: Nit: Here you could use `min_stake` directly:
```suggestion
            set.total_stake = set.total_stake - min_stake + staked_amount;
```

contracts/walrus/sources/staking/active_set.move
```diff
@@ -159,26 +153,85 @@ public(package) fun total_stake(set: &ActiveSet): u64 { set.total_stake }
...
 
-#[syntax(index)]
-/// Get the staked amount of the storage node with the given `node_id`.
-public(package) fun borrow(set: &ActiveSet, node_id: &ID): &u64 {
-    &set.nodes[node_id]
+// #[syntax(index)]
+// /// Get the staked amount of the storage node with the given `node_id`.
+// public(package) fun borrow(set: &ActiveSet, node_id: &ID): &u64 {
+//     &set.nodes[node_id]
+// }
```
mlegner: Question: Why is this commented out?

contracts/walrus/sources/staking/active_set.move
```diff
@@ -159,26 +153,85 @@ public(package) fun total_stake(set: &ActiveSet): u64 { set.total_stake }
...
+
+// === Test ===
+
+#[test]
+fun test_evict_correct_node_simple() {
```
mlegner, line 182: Suggestion: Can we also test the total stake here?

contracts/walrus/sources/staking/active_set.move
```diff
@@ -159,26 +153,85 @@ public(package) fun total_stake(set: &ActiveSet): u64 { set.total_stake }
...
+
+    // node 5 should not be part of the set
+    assert!(!active_ids.contains(&object::id_from_address(@5)));
+
+    // all other nodes should be
+    assert!(active_ids.contains(&object::id_from_address(@1)));
+    assert!(active_ids.contains(&object::id_from_address(@2)));
+    assert!(active_ids.contains(&object::id_from_address(@3)));
+    assert!(active_ids.contains(&object::id_from_address(@4)));
+    assert!(active_ids.contains(&object::id_from_address(@6)));
```
mlegner: Suggestion: Can we also test that their stake was updated and the total stake?

contracts/walrus/sources/staking/active_set.move
karlwuest: Enums in move can only be matched on in the module that defines them (the variants are private), so sadly, that is not possible.

contracts/walrus/sources/staking/active_set.move
karlwuest: Good catch! I commented this out to see where it's used and then I forgot to remove it afterwards.

PR #1033, opened 2024-10-10
fix: use platform agnostic tmp dir in tests

halfprice:
See if this fix the test failure in windows CI

### DIFF
--- a/crates/walrus-service/src/node/storage/event_blob_writer.rs
+++ b/crates/walrus-service/src/node/storage/event_blob_writer.rs
@@ -223,11 +223,7 @@ impl EventBlobWriter {
         self.wbuf.flush()?;
         self.wbuf.get_ref().sync_data()?;
         let off = self.wbuf.get_ref().stream_position()?;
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/storage/event_blob_writer.rs
```diff
@@ -310,7 +310,7 @@ impl EventBlobWriter {
         file.seek(SeekFrom::Start(0))?;
         let mut file_content = Vec::new();
         file.read_to_end(&mut file_content)?;
-        let mut tmp_file = File::create("/tmp/event_blob")?;
+        let mut tmp_file = File::create(self.root_dir_path.join("event_blob"))?;
```
sadhansood, line 326: nit: We could also use the `tempfile::tempdir` like we use in other tests

halfprice: root_dir_path now is set to tempfile::tempdir

PR #1034, opened 2024-10-10
fix: remove epoch check in process_events

sadhansood:
Remove this check as it seems to be causing lagging nodes to crash.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -521,7 +521,6 @@ impl StorageNode {
         let (event_stream, next_event_index) = self.continue_event_stream().await?;
 
         let index_stream = stream::iter(next_event_index..);
-        let mut maybe_epoch_at_start = Some(self.inner.committee_service.get_epoch());
 
         let mut indexed_element_stream = index_stream.zip(event_stream);
         // Important: Events must be handled consecutively and in order to prevent (intermittent)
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -547,25 +546,6 @@ impl StorageNode {
...
-
-                    // if event.event_epoch() < starting_epoch - 1
-                    if event.event_epoch() + 1 < epoch_at_start {
-                        let error = anyhow!(
-                            "the current epoch ({}) is too far ahead of the event epoch: {}",
-                            epoch_at_start,
-                            event.event_epoch(),
-                        );
-                        tracing::error!(%error);
-                        return Err(error);
```
jpcsmith, line 564: Important: Leave the check, and change this to emitting a warning instead of exiting with an error.

Remark: The check exits lagging nodes. It was intentionally added because we have neither yet implemented recovery for lagging nodes, nor sufficiently tested using the regular event-processing code path for recovery. It would therefore be good to still emit a warning because removing this will likely lead to bugs, and it will be helpful to see the warning in the logs to know that the removal of the check may be the source.

sadhansood: Thanks @jpcsmith , it makes sense to me. I'll close this PR for now and sync up with team once more before making any changes here as it seems critical for preventing lagging nodes get into an erroneous state

PR #1039, opened 2024-10-11
feat: add move errors to walrus-sui, remove clever errors

karlwuest:
This adds the error types but doesn't use them yet for the move calls.


crates/walrus-sui/src/types/move_errors.rs
mlegner, line 1: Question: Is there any way we can ensure this is in sync with the Move errors?

contracts/errors.out
mlegner, line 1: Question: How is this generated and what is it used for?

contracts/modules.out
mlegner, line 1: Question: How is this generated and what is it used for?

contracts/errors.out
karlwuest: With `cat` and `grep` and a little bit of `vim` 😅 
Didn't mean to commit this (I thought `*.out` was in .gitignore). I will remove this (and modules.out).

crates/walrus-sui/src/types/move_errors.rs
karlwuest: I don't know of a good way to do this automatically. I added comments in the move files to keep the rust error codes up to date, but I agree that an automatic way would be best. I'll add an issue for this.

PR #1047, opened 2024-10-11
feat: Ensure experimental rest endpoint is present at startup

sadhansood:
This PR checks if the endpoints which returns full checkpoint is working at startup and fail with an appropriate error message.

### DIFF
--- a/crates/walrus-event/src/lib.rs
+++ b/crates/walrus-event/src/lib.rs
@@ -6,7 +6,7 @@ use std::{
     io::{BufReader, BufWriter},
 };
 
-use anyhow::anyhow;
+use anyhow::{anyhow, bail};
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use serde::{Deserialize, Serialize};
 use sui_rest_api::Client;
@@ -195,4 +195,31 @@ pub async fn get_bootstrap_committee_and_checkpoint(
     );
     let verified_checkpoint = VerifiedCheckpoint::new_unchecked(checkpoint_data.checkpoint_summary);
     Ok((committee, verified_checkpoint))
+}
+
+async fn check_experimental_rest_endpoint_exists(client: Client) -> anyhow::Result<bool> {
+    // TODO: Once full nodes update their openapi.json spec to include the experimental endpoint, we
+    // should update this function to download the openapi.json spec and check if the endpoint
+    // /checkpoints/{checkpoint}/full is present or not.
+    let latest_checkpoint = client.get_latest_checkpoint().await?;
+    let mut total_remaining_attempts = 5;
+    while client
+        .get_full_checkpoint(latest_checkpoint.sequence_number)
+        .await
+        .is_err()
+    {
+        total_remaining_attempts -= 1;
+        if total_remaining_attempts == 0 {
+            return Ok(false);
+        }
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+    }
+    Ok(true)
+}
+
+pub async fn ensure_experimental_rest_endpoint_exists(client: Client) -> anyhow::Result<()> {
+    if !check_experimental_rest_endpoint_exists(client.clone()).await? {
+        bail!("Full node does not support experimental endpoint");
+    }
+    Ok(())
 }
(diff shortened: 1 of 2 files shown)


crates/walrus-event/src/lib.rs
```diff
@@ -196,3 +196,30 @@ pub async fn get_bootstrap_committee_and_checkpoint(
...
+        .get_full_checkpoint(latest_checkpoint.sequence_number)
+        .await
+        .is_err()
+    {
+        total_remaining_attempts -= 1;
+        if total_remaining_attempts == 0 {
+            return Ok(false);
+        }
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+    }
```
halfprice, line 216: Nit: there is a utils::retry() helper you can use. There is currently only one strategy, ExponentialBackoff, but I guess you can either set min and max backoff to be the same value, or create a FixedIntervalBackoff.

crates/walrus-event/src/lib.rs
```diff
@@ -196,3 +196,30 @@ pub async fn get_bootstrap_committee_and_checkpoint(
     let verified_checkpoint = VerifiedCheckpoint::new_unchecked(checkpoint_data.checkpoint_summary);
     Ok((committee, verified_checkpoint))
 }
+
+async fn check_experimental_rest_endpoint_exists(client: Client) -> anyhow::Result<bool> {
+    // TODO: Once full nodes update their openapi.json spec to include the experimental endpoint, we
```
halfprice, line 201: Suggestion: for TODO like this that there isn't an immediate action needed, I think typically we need to create a issue and put the issue number here to track it.

crates/walrus-event/src/lib.rs
```diff
@@ -196,3 +196,30 @@ pub async fn get_bootstrap_committee_and_checkpoint(
...
+    }
+    Ok(true)
+}
+
+pub async fn ensure_experimental_rest_endpoint_exists(client: Client) -> anyhow::Result<()> {
+    if !check_experimental_rest_endpoint_exists(client.clone()).await? {
+        bail!("Full node does not support experimental endpoint");
+    }
+    Ok(())
+}
```
halfprice, line 225: Nit: I feel we can just use `check_experimental_rest_endpoint_exists` in the call site. Adding another layer feels a bit unnecessary.

crates/walrus-event/src/lib.rs
sadhansood: That is a cool api, I just checked. But adding walrus-service to this crate causes cyclic dependency which requires some refactor to break. We should pull out the utils in another `walrus-utils` crate outside, but until then i'll keep this.

crates/walrus-event/src/lib.rs
sadhansood: Done

sadhansood: I updated the todo in a different PR (sorry things got mixed up)

crates/walrus-event/src/lib.rs
sadhansood: updated the comment to use this api later once utils is outside walrus-service

crates/walrus-event/src/lib.rs
sadhansood: added an info log here as well which justifies this function now

crates/walrus-event/src/lib.rs
```diff
@@ -196,3 +196,30 @@ pub async fn get_bootstrap_committee_and_checkpoint(
...
+            return Ok(false);
+        }
+        tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+    }
+    Ok(true)
+}
+
+pub async fn ensure_experimental_rest_endpoint_exists(client: Client) -> anyhow::Result<()> {
+    if !check_experimental_rest_endpoint_exists(client.clone()).await? {
+        bail!("Full node does not support experimental endpoint");
```
mlegner, line 222: I'm a bit late, but one comment here: This error message may not be too helpful for the operator: Which experimental endpoint? What is it needed for?

Can we clarify this a bit?

sadhansood: good call out, i can definitely see adding more details to this error message could be informative and avoid confusions.

mlegner: Resolved in #1057.

PR #1048, opened 2024-10-11
feat: Better error logging on checkpoint failures

sadhansood:
This came up in another PR, and rightly so we want to only log error on checkpoint failure if server pruned the checkpoint or debug otherwise

### DIFF
--- a/crates/walrus-event/src/lib.rs
+++ b/crates/walrus-event/src/lib.rs
@@ -9,7 +9,7 @@ use std::{
 use anyhow::{anyhow, bail};
 use byteorder::{BigEndian, ReadBytesExt, WriteBytesExt};
 use serde::{Deserialize, Serialize};
-use sui_rest_api::Client;
+use sui_rest_api::{client::sdk, Client};
 use sui_sdk::{
     rpc_types::{SuiObjectDataOptions, SuiTransactionBlockResponseOptions},
     SuiClient,
@@ -21,6 +21,7 @@ use sui_types::{
     messages_checkpoint::{CheckpointSequenceNumber, VerifiedCheckpoint},
     sui_serde::BigInt,
 };
+use tracing::{debug, error, info};
 use walrus_core::BlobId;
 use walrus_sui::types::{BlobEvent, ContractEvent};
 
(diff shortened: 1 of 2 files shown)


crates/walrus-event/src/lib.rs
```diff
@@ -196,3 +197,27 @@ pub async fn get_bootstrap_committee_and_checkpoint(
...
+/// If the error is due to a checkpoint that is already present on the server, it is logged as an
+/// error. Otherwise, it is logged as a debug.
+fn handle_checkpoint_error(err: Option<sdk::Error>, next_checkpoint: u64) {
+    let error = err.as_ref().map(|e| e.to_string()).unwrap_or_default();
+    if let Some(checkpoint_height) = err
+        .as_ref()
+        .and_then(|e| e.parts())
+        .and_then(|p| p.checkpoint_height)
+    {
+        if next_checkpoint <= checkpoint_height {
```
halfprice: I kind of feel we should do the other way: log debug!() if next_checkpoint >= checkpoint_height, and use error!() in other cases.

I feel the only case that is not an error is if next_checkpoint >= checkpoint_height. Current log masks all other fetching failures such as fullnode unavailable.

What do you think?

sadhansood: Isn't it what the code is already doing? IIUC, your suggestion of logigng with `debug` on `next_checkpoint > checkpoint_height` and `error` otherwise is what this is also doing or maybe i misunderstood?

halfprice: I think right now, the code logs error!() only if next_checkpoint <= checkpoint_height. For error like `unreachable`, it goes to debug!(), but I think that should also go to error!() since it is also an abnormal state.

sadhansood: ok I think i understand what you were saying i.e. default log as error but debug only if we can check the checkpoint height and find it lower that next_checkpoint

crates/walrus-event/src/lib.rs
```diff
@@ -217,9 +218,35 @@ async fn check_experimental_rest_endpoint_exists(client: Client) -> anyhow::Resu
...
+                "Failed to read next checkpoint: {}, checkpoint_height: {}, error: {}",
+                next_checkpoint, checkpoint_height, error
+            );
+            return;
+        }
+    }
+    error!(
+        "Failed to read next checkpoint: {} with error: {}",
+        next_checkpoint, error
+    );
```
mlegner, line 251: Remark: We generally try to log relevant additional variables as fields instead of integrating them into the string to help with searching:
```rust
debug!(next_checkpoint, checkpoint_height, %error, "Failed to read next checkpoint");
error!(next_checkpoint, %error, "Failed to read next checkpoint");
```

sadhansood: will take care of it

mlegner: Resolved in #1057.

PR #1050, opened 2024-10-12
feat: add status endpoint for LB health check

pei-mysten:
## Summary
as title, we need an endpoint to indicate the server is running for k8s & GKE LB health check.
I'm keeping the impl as simple as possible - just returning an "OK" string

## Test Plan
deployed to `https://aggregator.walrus-testnet.walrus.space/status` and `https://publisher.walrus-testnet.walrus.space/status`

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -25,7 +25,7 @@ max_line_length = 150
 indent_size = unset
 
 # Ignore paths
-[{.git/**/*,**/*.lock,**/Move.toml,LICENSE,docs/devnet-public/glossary.md,crates/walrus-orchestrator/assets/*,.editorconfig}]
+[{.git/**/*,**/*.lock,**/Move.toml,LICENSE,docs/devnet-public/glossary.md,crates/walrus-orchestrator/assets/*,.editorconfig,venv/**/*}]
 charset = unset
 end_of_line = unset
 indent_size = unset
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ build/
 client_config.yaml
 working_dir
 *.log
+venv/


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -172,6 +174,21 @@ pub(super) async fn store_blob_options() -> impl IntoResponse {
...
 
+#[tracing::instrument(level = Level::ERROR, skip_all)]
+#[utoipa::path(
+    get,
+    path = api::rewrite_route(STATUS_ENDPOINT),
+    responses(
+        (status = 200, description = "OK"),
+        (status = 404, description = "Not found"),
+        (status = 500, description = "Internal server error" ),
+        // TODO(mlegner): Improve error responses. (#178, #462)
```
mlegner: Question: Is the idea that the status endpoint *always* responds with a 200 status when the service is running?
```suggestion
        (status = 200, description = "The service is running"),
```

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -31,6 +31,8 @@ use crate::{
     common::api::{self, BlobIdString},
 };
 
+/// status endpoint.
```
mlegner:
```suggestion
/// The status endpoint, which always returns a 200 status when it is available.
```

crates/walrus-service/src/client/daemon/routes.rs
pei-mysten: yes we can be more specific on when to return 200 in the future, for now we can just blindly return 200 for passing LB health check

PR #1057, opened 2024-10-14
feat(node): several minor improvements

### LINKED ISSUE #1051: Support running nodes *before* they are members of a committee
Currently, a storage node fails if it is run without being part of the current committee. Supporting that case would help with deployments and make sure that new nodes can catch up with events before they become part of a committee (#993).

mlegner:
- Do not fail when started in genesis epoch.
- Do not call `voting_end` in genesis epoch.
- More useful error if full node does not provide the REST endpoint for checkpoints.
- Reduce log level for expected unavailable checkpoints to 'trace'.
- Add metadata as fields to log entries for checkpoint processing.

Addresses issue raised in https://github.com/MystenLabs/walrus/pull/1047#discussion_r1798040304.
Addresses issue raised in https://github.com/MystenLabs/walrus/pull/1048#discussion_r1798042910.
Closes #1051

### DIFF
--- a/crates/walrus-service/src/node/committee/committee_service.rs
+++ b/crates/walrus-service/src/node/committee/committee_service.rs
@@ -646,6 +646,13 @@ async fn add_members_from_committee<T: NodeService>(
     committee: &Committee,
     encoding_config: &Arc<EncodingConfig>,
 ) -> Result<(), anyhow::Error> {
+    if committee.epoch == 0 {
+        tracing::info!(
+            "we are in the genesis epoch, not creating any services for committee members"
+        );
+        return Ok(());
+    }
+
     let mut n_created = 0usize;
 
     for member in committee.members() {
--- a/crates/walrus-service/src/node/epoch_change_driver.rs
+++ b/crates/walrus-service/src/node/epoch_change_driver.rs
@@ -110,6 +110,12 @@ impl EpochChangeDriver {
     /// `schedule_*` methods for alternatives that will retry until successful.
     pub async fn schedule_relevant_calls_for_current_epoch(&self) -> Result<(), anyhow::Error> {
         let (epoch, state) = self.contract_service.get_epoch_and_state().await?;
+
+        if epoch == 0 {
+            tracing::info!("not scheduling any epoch-change calls in genesis epoch");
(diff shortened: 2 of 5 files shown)


crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -646,6 +646,13 @@ async fn add_members_from_committee<T: NodeService>(
     committee: &Committee,
     encoding_config: &Arc<EncodingConfig>,
 ) -> Result<(), anyhow::Error> {
+    if committee.epoch == 0 {
+        tracing::info!(
+            "we are in the genesis epoch, not creating any services for committee members"
+        );
+        return Ok(());
+    }
+
```
giac-mysten, line 655: question: the services are then created at epoch change, correct?

crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -110,6 +110,12 @@ impl EpochChangeDriver {
     /// `schedule_*` methods for alternatives that will retry until successful.
     pub async fn schedule_relevant_calls_for_current_epoch(&self) -> Result<(), anyhow::Error> {
         let (epoch, state) = self.contract_service.get_epoch_and_state().await?;
+
+        if epoch == 0 {
```
giac-mysten, line 114: question: any value in having `const GENESIS: Epoch = 0;`, or even `Epoch::is_genesis(&self)` ?

mlegner: I can definitely add that. 👍

crates/walrus-service/src/node/committee/committee_service.rs
mlegner: Yes, that's right. The services for the new epoch are created as a result of processing the epoch change event.

crates/walrus-service/src/node/epoch_change_driver.rs
mlegner: Actually, as `Epoch` is just a type alias, we cannot create a method `is_genesis`, and I'm not sure having a `GENESIS` constant adds much value.

PR #1058, opened 2024-10-14
feat: remove moved out shard in storage nodes after epoch change

### LINKED ISSUE #985: For moved out shard, the storage node can delete the shard at the beginning of the next start epoch change event

halfprice:
This PR implements shard cleanup after a shard is moved out of a storage node.

If a shard is moved out of a storage in epoch x, at the beginning of x+1, the shard is no longer needed in this storage and hence can be deleted from the storage node.

close #985

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1150,17 +1150,6 @@ impl StorageNodeInner {
         Ok(())
     }
 
-    async fn remove_storage_for_shard_in_background(
-        self: &Arc<Self>,
-        shards: ShardIndex,
-    ) -> Result<(), anyhow::Error> {
-        let this = self.clone();
-        tokio::task::spawn_blocking(move || this.storage.remove_storage_for_shards(&[shards]))
-            .in_current_span()
-            .await??;
-        Ok(())
-    }
-
     fn is_blob_registered(&self, blob_id: &BlobId) -> Result<bool, anyhow::Error> {
         Ok(self
             .storage
--- a/crates/walrus-service/src/node/background_shard_remover.rs
+++ b/crates/walrus-service/src/node/background_shard_remover.rs
@@ -52,7 +52,7 @@ impl BackgroundShardRemover {
 
             let result = utils::retry(backoff, || async {
                 remover_clone
-                    .remove_storage_for_shards(event_element, event_clone.clone(), shards.clone())
+                    .remove_storage_for_shards(event_element, event_clone.clone(), &shards.clone())
                     .await
             })
             .await;
@@ -78,13 +78,13 @@ impl BackgroundShardRemover {
         &self,
         event_element: usize,
         event: EpochChangeStart,
-        shards: Vec<ShardIndex>,
+        shards: &[ShardIndex],
     ) -> anyhow::Result<()> {
         for shard in shards {
             tracing::info!(shard = %shard, "start removing shard from storage");
             self.node
-                .remove_storage_for_shard_in_background(shard)
-                .await
+                .storage
+                .remove_storage_for_shards(&[*shard])
                 .map_err(|err| {
                     tracing::error!(
                         epoch = %event.epoch,


crates/walrus-service/src/common/utils.rs
```diff
@@ -481,33 +481,49 @@ pub(crate) struct ShardDiff {
...
 }
 
 impl ShardDiff {
     /// Returns a new `ShardDiff` when moving from the allocation in
     /// `committees.previous_committee()` to `committees.current_committee()` for the node
     /// identified by the provided public key.
-    pub fn diff_previous(committees: &ActiveCommittees, id: &PublicKey) -> ShardDiff {
+    pub fn diff_previous(
+        committees: &ActiveCommittees,
+        exist: &[ShardIndex],
```
mlegner, line 496: Nit: Can you describe what this parameter is?

crates/walrus-service/src/common/utils.rs
```diff
@@ -822,4 +838,64 @@ mod tests {
         test_parse_various!(peta, "P", 1e15);
         test_parse_various!(pebi, "Pi", 1024 * 1024 * 1024 * 1024 * 1024u64);
     }
+
+    #[test]
+    fn test_shard_diff() {
```
mlegner: How about the following version:
```rust
    use walrus_test_utils::assert_unordered_eq;

    param_test! {
        test_shard_diff: [
            only_gain: (
                ShardDiff::diff(
                    &[],
                    &[ShardIndex(1), ShardIndex(2), ShardIndex(3), ShardIndex(4)],
                    &[],
                ),
                ShardDiff {
                    gained: vec![ShardIndex(1), ShardIndex(2), ShardIndex(3), ShardIndex(4)],
                    ..Default::default()
                },
            ),
            gained_and_lost: (
                ShardDiff::diff(
                    &[ShardIndex(1), ShardIndex(2), ShardIndex(3), ShardIndex(4)],
                    &[ShardIndex(2), ShardIndex(3), ShardIndex(4), ShardIndex(5)],
                    &[],
                ),
                ShardDiff {
                    lost: vec![ShardIndex(1)],
                    gained: vec![ShardIndex(5)],
                    unchanged: vec![ShardIndex(2), ShardIndex(3), ShardIndex(4)],
                    removed: vec![],
                },
            ),
            gained_lost_and_removed: (
                ShardDiff::diff(
                    &[ShardIndex(3), ShardIndex(4), ShardIndex(5)],
                    &[ShardIndex(6), ShardIndex(7), ShardIndex(4)],
                    &[ShardIndex(1), ShardIndex(2)],
                ),
                ShardDiff {
                    lost: vec![ShardIndex(3), ShardIndex(5)],
                    gained: vec![ShardIndex(6), ShardIndex(7)],
                    unchanged: vec![ShardIndex(4)],
                    removed: vec![ShardIndex(1), ShardIndex(2)],
                },
            ),
        ]
    }
    fn test_shard_diff(computed_diff: ShardDiff, expected_result: ShardDiff) {
        assert_unordered_eq!(computed_diff.lost, expected_result.lost);
        assert_unordered_eq!(computed_diff.gained, expected_result.gained);
        assert_unordered_eq!(computed_diff.unchanged, expected_result.unchanged);
        assert_unordered_eq!(computed_diff.removed, expected_result.removed);
    }
```

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -79,6 +79,7 @@ mod tests {
         let (_sui_cluster, _cluster, client) =
             test_cluster::default_setup_with_epoch_duration_generic::<SimStorageNodeHandle>(
                 Duration::from_secs(60 * 60),
+                &[1u16, 2, 3, 3, 4],
```
mlegner: Question: (Why) Is the `u16` necessary here?

crates/walrus-sui/src/test_utils/system_setup.rs
```diff
@@ -196,15 +196,13 @@ pub async fn register_committee_and_stake(
         system_context.package_id,
         system_context.treasury_cap,
         &receiver_addrs,
-        *amounts_to_stake
-            .iter()
-            .max()
-            .ok_or_else(|| anyhow!("no staking amounts provided"))?,
+        1_000_000_000_000,
```
mlegner: Question/important: What is the motivation for this change? And how do we know this is sufficient to perform the staking?

crates/walrus-service/src/node.rs
```diff
@@ -1087,6 +1117,17 @@ impl StorageNodeInner {
         Ok(())
     }
 
+    async fn remove_storage_for_shards_in_background(
+        self: &Arc<Self>,
+        shards: Vec<ShardIndex>,
+    ) -> Result<(), anyhow::Error> {
+        let this = self.clone();
+        tokio::task::spawn_blocking(move || this.storage.remove_storage_for_shards(&shards))
```
mlegner: Question: (Why) Do we need `spawn_blocking` here?

crates/walrus-sui/src/test_utils/system_setup.rs
halfprice: > What is the motivation for this change

We trigger shard re-assignment by adding stakes. If we just mint `amounts_to_stake`, there is no extra we can stake.

> And how do we know this is sufficient to perform the staking?

We control the test and know how much to stake beforehand

crates/walrus-simtest/tests/simtest.rs
halfprice: Good catch. Not really needed here. Just copy-paste from other site.

crates/walrus-service/src/node.rs
halfprice: This follows the pattern of `create_storage_for_shards_in_background` above.

The reason for blocking is due to that we can't mark EpochChangeStart event as processed until we remove all the shards. Otherwise, crash during the process will cause shards remain in the node forever.

As for the reason for using spawn, I think this is due to that it gives the tokio scheduler ability to run other ACTORs while waiting for the deletion, as opposite to the entire thread is waiting for removing the shard.

crates/walrus-service/src/common/utils.rs
halfprice: Done.

crates/walrus-service/src/common/utils.rs
halfprice: Very nice! Updated.

crates/walrus-service/src/test_utils.rs
```diff
@@ -1642,13 +1666,19 @@ pub mod test_cluster {
         TestCluster,
         WithTempDir<client::Client<SuiContractClient>>,
     )> {
-        default_setup_with_epoch_duration_generic::<StorageNodeHandle>(epoch_duration).await
+        let node_weights = [1u16, 2, 3, 3, 4];
```
mlegner: Nit:
```suggestion
        let node_weights = [1, 2, 3, 3, 4];
```

crates/walrus-sui/src/test_utils/system_setup.rs
mlegner: Hmm... I think at the very least this would have to be specified somewhere. Currently, anyone can call this function with `amounts_to_stake` exceeding 1000 WAL. Maybe an option would be to take the maximum + 1000 WAL.

crates/walrus-service/src/node.rs
mlegner: But does this mean that the event processing is completely blocked when shards have to be removed? This is definitely not ideal as it means the node cannot store any new blobs for that time.

If this is the case, a better solution would be the approach we use for the "certified" events, which spawn a blob sync which then at the end marks the event as completed.

For the `create_storage_for_shards_in_background`, the delay is of course also not ideal, but we need to make sure that this completes before handling anything else.

crates/walrus-sui/src/test_utils/system_setup.rs
halfprice: I see what you are saying. I added a `wal_to_mint` parameter in the function, and the mint amount is set to max(wal_to_min, max(amounts_to_stake))I see what you are saying. I added a `wal_to_mint` parameter in the function, and the mint amount is set to max(wal_to_min, max(amounts_to_stake))

crates/walrus-service/src/node.rs
halfprice: Yeah, the event processing is blocked (but uploads/reads are still available).

Note that dropping column family is a RockDb metadata only operation, but I share your concern given that the last time when I tried to fix PTN, dropping 5 CFs takes about 18 seconds. In case there is large number of CFs we need to drop, this would block the node for minutes.

I added BackgroundShardRemover, and basically does the same thing, but in the background and does not block event processing. The pattern follows BlobSynchronizer and ShardSynchronizer, but much simpler.

crates/walrus-service/src/node.rs
```diff
@@ -1117,12 +1150,12 @@ impl StorageNodeInner {
         Ok(())
     }
 
-    async fn remove_storage_for_shards_in_background(
+    async fn remove_storage_for_shard_in_background(
         self: &Arc<Self>,
-        shards: Vec<ShardIndex>,
+        shards: ShardIndex,
```
mlegner: Nit: This should probably be called `shard` (singular).

crates/walrus-service/src/node/background_shard_remover.rs
```diff
@@ -0,0 +1,117 @@
...
+        }
+    }
+
+    /// Start a task to remove storage for the given shards.
+    /// Mark the event as completed after the removal is successful.
+    pub fn start_remove_storage_for_shards(
+        &self,
+        event_element: usize,
+        event: &EpochChangeStart,
+        shards: Vec<ShardIndex>,
```
mlegner, line 37: Question: Does this have to be a `Vec` or could this be a `&[]`? Same for `remove_storage_for_shards`.

crates/walrus-service/src/node.rs
```diff
@@ -1117,12 +1150,12 @@ impl StorageNodeInner {
...
     ) -> Result<(), anyhow::Error> {
         let this = self.clone();
-        tokio::task::spawn_blocking(move || this.storage.remove_storage_for_shards(&shards))
+        tokio::task::spawn_blocking(move || this.storage.remove_storage_for_shards(&[shards]))
             .in_current_span()
             .await??;
         Ok(())
```
mlegner: Question/suggestion: Is it important that this is a method on `StorageNodeInner`? If not, I'd suggest to either move it to `BackgroundShardRemover` or even inline it in `BackgroundShardRemover::remove_storage_for_shards`; it's a bit unintuitive jumping from `StorageNode::process_shard_changes_in_new_epoch` (in this file) to `BackgroundShardRemover::start_remove_storage_for_shards` in a submodule, and from there back here.

crates/walrus-service/src/node.rs
halfprice: Good point (and this function is no longer exist)

crates/walrus-service/src/node/background_shard_remover.rs
halfprice: The issue is that if passing a slice reference, we need to copy it to pass into a tokio::spawn. So I think it's less confusing to just pass in an owned vector.

Changed to &[ShardIndex] in `remove_storage_for_shards`

crates/walrus-service/src/node.rs
halfprice: Yeah, I think inline is better. Also, we already run the remover in a seperate thread, and we probably don't need to spin up another thread to remove individual shard.

PR #1060, opened 2024-10-14
feat(node)!: use checkpoint-based event processor by default

### LINKED ISSUE #1055: Add option to enable checkpoint-based event processor to `walrus-node setup`

mlegner:
- Minor breaking changes to the config file.
- Remove `--enable-checkpoint-event-processor` option for `walrus-deploy`, add new `--use-legacy-event-processor` option.

Closes #1055

### DIFF
--- a/scripts/local-testbed.sh
+++ b/scripts/local-testbed.sh
@@ -114,7 +114,7 @@ if ! $existing; then
     # Generate configs
     echo Generating configuration...
     cargo run --bin walrus-deploy -- generate-dry-run-configs \
-    --working-dir $working_dir
+    --working-dir $working_dir --use-legacy-event-provider
 fi
 
 i=0


crates/walrus-service/src/node/config.rs
```diff
@@ -108,7 +111,7 @@ impl Default for StorageNodeConfig {
             blob_recovery: Default::default(),
             tls: Default::default(),
             shard_sync_config: Default::default(),
-            event_processor_config: Default::default(),
+            event_provider_config: Default::default(),
```
giac-mysten, line 114: Remark: the simtests fail b/c this variable has not been updated at places, e.g.,  https://github.com/MystenLabs/walrus/pull/1060/files#diff-8a974540e8ac8c723e5d58ca0b859b04dde05bc69f49e7dc49ee0aa576f78f7bL353

crates/walrus-service/bin/deploy.rs
```diff
@@ -136,8 +136,11 @@ struct GenerateDryRunConfigsArgs {
...
     faucet_cooldown: Option<Duration>,
-    /// Enable checkpoint based event processor
-    /// [default: false]
+    /// Use the legacy event processor instead of the standard checkpoint-based event processor.
+    #[clap(long, action)]
+    use_legacy_event_provider: bool,
+    /// Deprecated; using the checkpoint-based event processor is now the default. Use
+    /// `--use-legacy-event-processor` to use the legacy event processor.
     #[clap(long, action)]
     enable_checkpoint_event_processor: bool,
```
giac-mysten: Question: why do we keep this? only a handful of external operators have even seen this flag.
Suggestion: remove?

crates/walrus-service/src/node.rs
```diff
@@ -299,19 +300,28 @@ impl StorageNodeBuilder {
...
-                        sui_config.event_polling_interval,
-                        &config.storage_path.join("events"),
-                        &metrics_registry,
+            let event_manager: Box<dyn EventManager> = match &config.event_provider_config {
+                EventProviderConfig::CheckpointBasedEventProcessor(event_processor_config) => {
+                    let event_processor_config =
+                        event_processor_config.clone().unwrap_or_else(|| {
+                            EventProcessorConfig::new_with_default_pruning_interval(
+                                sui_config.rpc.clone(),
+                            )
```
giac-mysten, line 310: Question: instead of `unwrap_or_else` in multiple places, could we just have that if there is no configuration, the default pruning interval is picked and we don't have the `Option` on the `EventProcerssorConfig`?

giac-mysten: ah I guess we may not have the rpc url yet. Please disregard the above.

crates/walrus-service/bin/deploy.rs
mlegner: We currently also have this in the `sui-operations` setup. But yes, I can remove this completely.

crates/walrus-service/src/node/config.rs
mlegner: Should be fixed by e1318ceff994675a8869b19c8f8b77af4e0dc2ae.

crates/walrus-service/bin/deploy.rs
mlegner: Removed in 0c274549f7723f60f217e67f6d5ae0c828b202f9.

PR #1061, opened 2024-10-14
feat(node): add `public_host` and `public_port` to config

### LINKED ISSUE #1002: Add `public_hostname` and `public_port` to node config file and use this also for the TLS server name
This can be used for the `register` command and also to update the node's data as part of #965.

mlegner:
Closes #1002

### DIFF
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -112,7 +112,7 @@ impl Default for StorageNodeConfig {
             db_config: Default::default(),
             protocol_key_pair: PathOrInPlace::InPlace(ProtocolKeyPair::generate()),
             network_key_pair: PathOrInPlace::InPlace(NetworkKeyPair::generate()),
-            public_host: Some("127.0.0.1".into()),
+            public_host: Some(defaults::rest_api_address().ip().to_string()),
             public_port: Some(defaults::rest_api_port()),
             metrics_address: defaults::metrics_address(),
             rest_api_address: defaults::rest_api_address(),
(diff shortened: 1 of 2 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -195,9 +189,12 @@ struct ConfigArgs {
     /// Initial storage capacity of this node in bytes.
     ///
     /// The value can either by unitless; have suffixes for powers of 1000, such as (B),
-    /// kilobytes (K), etc, or have suffixes for the IEC units such as kibibytes (Ki),
+    /// kilobytes (K), etc.; or have suffixes for the IEC units such as kibibytes (Ki),
     /// mebibytes (Mi), etc.
     node_capacity: ByteCount,
+    #[clap(long)]
+    /// The host name or public IP address of the node.
+    public_host: String,
```
halfprice, line 206: Question: what's the relationship between this in IP format and rest_api_address? Seem that they should match?

crates/walrus-service/bin/node.rs
```diff
@@ -483,37 +475,34 @@ mod commands {
...
+        let public_port = config.public_port.unwrap_or(REST_API_PORT);
         let public_address = if let Ok(ip_addr) = IpAddr::from_str(&public_host) {
-            NetworkAddress(SocketAddr::new(ip_addr, port).to_string())
+            NetworkAddress(SocketAddr::new(ip_addr, public_port).to_string())
         } else {
-            // Do a minor sanity check that the user has not included a port in the hostname
-            ensure!(
-                !public_host.contains(':'),
-                "DNS names must not contain ':', to specify a port, use the --port option."
-            );
```
halfprice, line 520: Nit: this check seems still useful

crates/walrus-service/bin/node.rs
mlegner: The `rest_api_address` can be something like `0.0.0.0` or `127.0.0.1`. The `public_host` should be the public IP address of the node or any reverse proxy in front of it.

crates/walrus-service/bin/node.rs
mlegner: This is now included in the `generate_config` function. But I can keep this here as well.

PR #1065, opened 2024-10-15
test: crash if any of the storage node processes fail with error

halfprice:
Right now, the tests silently dropped the error when any of the storage node components (rest API, node, and event processor) returned with an error, which mask failure.

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -658,7 +658,14 @@ impl StorageNodeHandleBuilder {
             let cloned_cancel_token = cancel_token.clone();
             let cloned_event_processor = event_processor.clone();
             tokio::spawn(
-                async move { cloned_event_processor.start(cloned_cancel_token).await }.instrument(
+                async move {
+                    let status = cloned_event_processor.start(cloned_cancel_token).await;
+                    if let Err(error) = status {
+                        tracing::error!("event processor stopped with error: {:?}", error);
+                        std::process::exit(1);
+                    }
+                }
+                .instrument(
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/test_utils.rs
```diff
@@ -658,7 +658,14 @@ impl StorageNodeHandleBuilder {
             let cloned_cancel_token = cancel_token.clone();
             let cloned_event_processor = event_processor.clone();
             tokio::spawn(
-                async move { cloned_event_processor.start(cloned_cancel_token).await }.instrument(
+                async move {
```
sadhansood, line 661: We maybe able to write a function to spawn task, log error and exit which helps with the readability a little bit:
```
fn spawn_process_crashable_task<F>(name: &str, f: F)
where
    F: Future<Output = Result<(), anyhow::Error>> + Send,
{
    tokio::spawn(async move {
        if let Err(e) = f.await {
            tracing::error!("Task {} failed with error: {:?}", name, e);
            std::process::exit(1);
        } 
    })
}```
spawn_process_crashable_task("event_processor", run_event_processor(config.clone(), cancel_token.clone()));
spawn_process_crashable_task("rest_api", run_rest_api(config.clone(), cancel_token.clone()));
spawn_process_crashable_task("node", run_node(config, cancel_token.clone()));

halfprice: That's a great idea!

I spent about an hour on this, and eventually decide to table this for the future. There are some challenges to make this generalization a bit hard. Mainly because that amount these processes, some are inside an Arc and some are not, which makes the spawn_process_crashable_task to take the ownership of the object a bit tricky.

PR #1066, opened 2024-10-16
fix: make process_epoch_change_start_event idempotent

### LINKED ISSUE #1045: Make sure nodes do not crash if they process an epoch change event multiple times

halfprice:
Same epoch change event may be processed multiple times. For example, an earlier certified blob event is taking long time to finish, and the node crashes after processing epoch change start event. Therefore, we need to make sure that we don't restart syncing shards that have been synced.

Close #1045

### DIFF
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -52,8 +52,8 @@ impl ShardSyncHandler {
             .contains_key(&shard_index)
         {
             tracing::info!(
-                "Shard index: {} is already being synced. Skipping starting new sync task.",
-                shard_index
+                shard_index=%shard_index,
+                "shard is already being synced; skipping starting new shard sync",
             );
             return Ok(());
         }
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/src/types.rs
```diff
@@ -290,4 +290,10 @@ impl Committee {
             .iter()
             .map(|member| (&member.network_address, &member.network_public_key))
     }
+
+    /// Helper function to returns the members of the committee as mutable in tests.
```
mlegner: Nit:
```suggestion
    /// Helper function that returns the members of the committee as mutable in tests.
```

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -85,17 +111,22 @@ impl ShardSyncHandler {
...
 
         // TODO(#705): implement rate limiting for shard syncs.
         let mut shard_sync_in_progress = self.shard_sync_in_progress.lock().await;
         let Entry::Vacant(entry) = shard_sync_in_progress.entry(shard_storage.id()) else {
-            tracing::info!(
+            // We have checked the shard_sync_in_progress map before starting the sync task. So,
+            // this is an unexpected state.
+            tracing::error!(
                 "Shard index: {} is already being synced. Skipping starting new sync task.",
                 shard_storage.id()
```
mlegner: Nit (not related to these changes): We generally try to record additional data as fields instead of including them in the string.
```suggestion
                shard_index=%shard_storage.id(),
                "shard is already being synced; skipping starting new sync task",
```

PR #1072, opened 2024-10-17
feat(staking): add simple early withdrawal mechanic

### LINKED ISSUE #1028: Allow request_withdrawal before activation epoch

damirka:
Allows direct stake withdrawal before `activation_epoch` before committee for the activation epoch is selected.

Closes #1028

### DIFF
--- a/contracts/walrus/sources/staking/pending_values.move
+++ b/contracts/walrus/sources/staking/pending_values.move
@@ -5,8 +5,9 @@ module walrus::pending_values;
 
 use sui::vec_map::{Self, VecMap};
 
-#[error]
-const EIncorrectValue: vector<u8> = b"Incorrect value, can't reduce non existing value";
+/// Attempt to reduce the value of a pending value for an epoch that does not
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/staking/pending_values.move
```diff
@@ -5,6 +5,9 @@ module walrus::pending_values;
 
 use sui::vec_map::{Self, VecMap};
 
+#[error]
+const EIncorrectValue: vector<u8> = b"Incorrect value, can't reduce non existing value";
+
```
karlwuest: Important: We should use an error code here and add a corresponding error to the rust code. I switched all errors back from clever errors to error codes, because clever errors don't work nicely with rust (neither do error codes, but they work better). See #1039 .

damirka: Nice! Yeah, smart errors are not that good in other ways too, like formatting changes require package upgrade due to line number change.

PR #1073, opened 2024-10-17
feat: add walrus-proxy for metrics

suiwombat:
## Description
create a walrus-proxy, similar to how we have a sui-proxy.  Some items are borrowed, but there is a considerable amount of customization here as well.

General flow of how this works;

* walrus-node spins up a reqwest client and connects to the walrus-proxy over anonymous TLS, accepting well known CA certs like we do today in other parts of our infra.
* the network private key is used to sign a uuidv7 and that signature is sent in an Authorization header along with a message. The signature is a recoverable secp256r1 signature which allows the pub key to be recovered service side as long as the message that was signed is also known.  We send the signature as base64 and also encode the uuidv7 as the message to be signed over as a string.
* the walrus-proxy will inspect this header and extract the pub key from the signature, decoding as needed. If it matches a known walrus-node, we accept the metric.  This authorization header can be swapped out later if we want mTLS, but there is a benefit in using the header mechanism because we can easily avoid tls termination on this service and instead rely on gcp's upstream load balancers to do this for us.
* walrus-proxy side, we have a process that polls for a staking object, we extract committee members and we grab their pub key. this is key to how we know who to accept metrics from.  This is done on a configurable interval and automatically adjusts to nodes joining or leaving the stacking object we depend on.  If the staking object id were to change, then we'd need to update the variable that holds this object id.  If this is a frequent occurrence, we could add an api endpoint to the service to reconfigure itself w/o a restart.
* in a follow-up PR, I will add an additional layer of replay protection using a sliding window approach.  The uuidv7 uses the most significant 48 bits to encode a unix epoch timestamp.  We can order these uuids based on time and apply a replay constraint to middleware to prevent the same message from being used over and over again.  Messages that fall outside of coordinated UTC windows will not be allowed, regardless of being seen prior.

## Test Plan
local tests

working as expected so far. I briefly had it on and captured some data. https://metrics.sui.io/goto/-R1hv-mHg?orgId=1

### DIFF
--- a/crates/walrus-service/src/common/utils.rs
+++ b/crates/walrus-service/src/common/utils.rs
@@ -353,7 +353,11 @@ impl MetricsAndLoggingRuntime {
     pub fn join(&mut self) -> Result<(), anyhow::Error> {
         if let Some(metric_push_handle) = self._metric_push_handle.take() {
             tracing::debug!("waiting for the metric runtime to shutdown...");
-            self.runtime.block_on(metric_push_handle)?
+            self.runtime.block_on(async {
+                // The first `?` handles the timeout result.
+                // The second `?` handles the handle result.
+                tokio::time::timeout(Duration::from_secs(5), metric_push_handle).await??
+            })
         } else {
             Ok(())
         }


Cargo.toml
```diff
@@ -67,6 +87,8 @@ sui-sdk = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.35.0" }
 sui-simulator = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.35.0" }
 sui-storage = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.35.0" }
 sui-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.35.0" }
+sui-proxy = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.35.0" }
```
pei-mysten: do we need this dep? didn't see anything refer to ti

crates/walrus-proxy/src/providers/walrus/provider.rs
```diff
@@ -0,0 +1,144 @@
...
+        ],
+    )
+    .unwrap()
+});
+
+/// AllowedPeers is a mapping of public key to AllowedPeer data
+pub type AllowedPeers = Arc<RwLock<HashMap<u64, NodeInfo>>>;
+
+/// WalrusNodeProvider queries the sui blockchain and keeps a record of known validators based on the response from
+/// sui_getValidators.  The node name, public key and other info is extracted from the chain and stored in this
```
pei-mysten: :)

crates/walrus-proxy/src/providers/walrus/provider.rs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            info!("loaded node:[{name}] network_address: [{network_address}]");
+        }
+        if committee.is_empty() {
+            error!("walrus committee is empty? refusing to attempt to update cache");
+            return;
+        }
+        let nodes: HashMap<u64, NodeInfo> = committee
+            .into_iter()
+            .map(|v| (xxh3_64(v.network_public_key.as_ref()), v))
```
pei-mysten: is this base64? one thing i start network_key is stored as base64 encoded in the key file, so it would be a lot easier to compare if both are using the same format

Cargo.toml
suiwombat: yes, there are a few crates i use, namely the protobuf decoding/encoding

crates/walrus-proxy/src/providers/walrus/provider.rs
suiwombat: at one point it is. At this particular point, it is bytes.

crates/walrus-service/src/node/push_metrics.rs
```diff
@@ -0,0 +1,177 @@
...
+    })?;
+
+    let uid = Uuid::now_v7();
+    let uids = uid.simple().to_string();
+    let signature = network_key_pair.sign_recoverable(uid.as_bytes());
+    let auth = serde_json::json!({"signature":signature.encode_base64(), "message":uids});
+
+    let response = client
+        .post(push_url)
+        .header(reqwest::header::AUTHORIZATION, auth.to_string())
```
rjs-mysten: You're already signing the bytes, which is good as there's no malleability regarding dashes. I'd suggest adding a Bearer scheme and encoding the auth header value itself as base64 to make sure no future proxy messes with anything 
```
diff --git a/crates/walrus-service/src/node/push_metrics.rs b/crates/walrus-service/src/node/push_metrics.rs
index 358e006..e78ad2e 100644
--- a/crates/walrus-service/src/node/push_metrics.rs
+++ b/crates/walrus-service/src/node/push_metrics.rs
@@ -10,8 +10,7 @@ use std::{
 
 use anyhow::{Context as _, Error, Result};
 use fastcrypto::{
-    secp256r1::Secp256r1KeyPair,
-    traits::{EncodeDecodeBase64, RecoverableSigner},
+    encoding::Base64, secp256r1::Secp256r1KeyPair, traits::{EncodeDecodeBase64, RecoverableSigner}
 };
 use prometheus::{Encoder, Registry};
 use serde_json;
@@ -145,13 +144,15 @@ async fn push_metrics(
     })?;
 
     let uid = Uuid::now_v7();
-    let uids = uid.simple().to_string();
     let signature = network_key_pair.sign_recoverable(uid.as_bytes());
-    let auth = serde_json::json!({"signature":signature.encode_base64(), "message":uids});
+    let message = Base64::from_bytes(uid.as_bytes());
+    let auth_value = serde_json::json!({"signature":signature.encode_base64(), "message": message.encoded()});
+    let auth_base64 = Base64::from_bytes(auth_value.to_string().as_bytes()).encoded();
+    let auth = format!("Bearer {auth_base64}");
 
     let response = client
         .post(push_url)
-        .header(reqwest::header::AUTHORIZATION, auth.to_string())
+        .header(reqwest::header::AUTHORIZATION, auth)
         .header(reqwest::header::CONTENT_ENCODING, "snappy")
         .header(reqwest::header::CONTENT_TYPE, prometheus::PROTOBUF_FORMAT)
         .body(compressed)
```

crates/walrus-proxy/src/middleware.rs
```diff
@@ -0,0 +1,213 @@
...
+        return Err((StatusCode::FORBIDDEN, "authorization header is required"));
+    };
+
+    let auth_info: AuthInfo = serde_json::from_str(auth_header.to_str().unwrap_or_default())
+        .map_err(|_| (StatusCode::FORBIDDEN, "authorization header is malformed"))?;
+
+    let Some(peer) = allower.get(&auth_info.recover().map_err(|e| {
+        error!("allower failed us; {}", e);
+        (StatusCode::FORBIDDEN, "authorization header is malformed")
+    })?) else {
```
rjs-mysten: if you make the changes I suggested on walrus-service/src/push_metrics.rs, then you also need to change this (split the header into scheme and value, test that the scheme is Bearer, base64_decode the header value, and parse the uuid from base64_decoded message.

crates/walrus-service/src/node/push_metrics.rs
suiwombat: done

Cargo.toml
```diff
@@ -37,17 +52,26 @@ mockall = "0.12.1"
...
 prometheus = "0.13.4"
+prost = "0.13"
+protobuf = { version = "2.28", features = ["with-bytes"] }
 quote = "1.0"
 rand = "0.8.5"
 raptorq = "2.0.0"
 rcgen = "0.13.1"
 regex = "1"
-reqwest = { version = "0.12.9", default-features = false, features = ["http2", "json", "rustls-tls"] }
+reqwest = { version = "0.12.8", default-features = false, features = [
```
pei-mysten: not sure if this is a typo, but maybe let's use 0.12.9? should have no difference whatsoever

crates/walrus-proxy/src/consumer.rs
```diff
@@ -0,0 +1,363 @@
...
+        &["operation"]
+    )
+    .expect("consumer_operations_duration_seconds"))
+});
+
+/// NodeMetric holds metadata and a metric payload from the calling node
+#[derive(Debug)]
+pub struct NodeMetric {
+    /// dns or ip found from walrus committee along with port;
+    /// walrus-testnet.blockscope.net:9185 or 57.129.49.124:9185
```
pei-mysten: given that walrus will be open source in the future, lets not keeping this in the code :)

crates/walrus-service/Cargo.toml
```diff
@@ -33,6 +33,8 @@ deploy = ["client", "node"]
 node = [
   "dep:async-trait",
   "dep:enum_dispatch",
+  "dep:humantime",
+  "dep:itertools",
```
mlegner: Nit: Instead of adding these to the `node` feature, we can just remove the `optional` for them.

crates/walrus-service/bin/node.rs
mlegner, line 1: Nit: Can you revert the reformatting of this file? We use a max line length of 100.

crates/walrus-service/src/node/config.rs
```diff
@@ -438,7 +452,7 @@ pub mod defaults {
         2000
     }
 
-    /// Returns true iff the value is the default.
+    /// Returns true if the value is the default.
```
mlegner: Nit: This is actually not a typo but an abbreviation for ["if and only if"](https://en.wikipedia.org/wiki/If_and_only_if).

crates/walrus-service/src/node/config.rs
```diff
@@ -103,6 +103,9 @@ pub struct StorageNodeConfig {
     /// Name of the storage node.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub name: Option<String>,
+    /// Metric push configuration.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub metrics: Option<MetricsConfig>,
```
mlegner: Nit: If this is only for the push config, can we call this `metrics_push` or something like that? If this is supposed to include all metrics config, then we should at least move the `metrics_address` here (although this would be a breaking change).

crates/walrus-service/bin/node.rs
```diff
@@ -458,6 +487,11 @@ mod commands {
         // Wait for the node runtime to complete, may take a moment as
         // the REST-API waits for open connections to close before exiting.
         node_runtime.join()?;
+
+        // see if we also should wait for metrics to shutdown
```
mlegner: Yes, we should also include the metrics runtime in these checks, I created an issue for this: #1152.

crates/walrus-service/bin/node.rs
```diff
@@ -379,6 +388,20 @@ mod commands {
         let cancel_token = CancellationToken::new();
         let (exit_notifier, exit_listener) = oneshot::channel::<()>();
 
+        let mut push_metrics_runtime: Option<push_metrics::MetricPushRuntime> = None;
+        if let Some(metric_config) = config.metrics.clone() {
+            if let Some(tagged_key_pair) = config.network_key_pair.clone().get() {
```
mlegner: This is guaranteed to be there as it is loaded above. But we could make this clearer by changing lines 368-371 to
```

        let network_key_pair = config.network_key_pair.load()?;
        tracing::info!(
            walrus.node.network_key = %network_key_pair.as_ref().public(),
            "walrus network key",
        );
```
and then use `network_key_pair` here directly.

mlegner: Also, I'm not sure all the clones here and in the following are necessary.

crates/walrus-service/src/node/push_metrics.rs
```diff
@@ -0,0 +1,173 @@
...
+        network_key_pair: Arc<Secp256r1KeyPair>,
+        config: MetricsConfig,
+        registry: Registry,
+    ) -> anyhow::Result<Self> {
+        let runtime = Builder::new_multi_thread()
+            .thread_name("event-manager-runtime")
+            .worker_threads(2)
+            .enable_all()
+            .build()
+            .context("event manager runtime creation failed")?;
```
mlegner: This was probably copied from the event-processor runtime and should be adjusted.
```suggestion
        let runtime = Builder::new_multi_thread()
            .thread_name("metrics-push-runtime")
            .worker_threads(1)
            .enable_all()
            .build()
            .context("metrics push runtime creation failed")?;
```

crates/walrus-service/src/node/push_metrics.rs
```diff
@@ -0,0 +1,173 @@
...
+
+use super::config::MetricsConfig;
+
+/// MetricPushRuntime to manage the metric push task
+#[allow(missing_debug_implementations)]
+pub struct MetricPushRuntime {
+    metric_push_handle: JoinHandle<anyhow::Result<()>>,
+    // INV: Runtime must be dropped last.
+    runtime: Runtime,
+}
```
mlegner: Question: Do we really need a separate runtime for pushing the metrics? Could we maybe include this in the other metrics runtime?

crates/walrus-service/src/node/config.rs
```diff
@@ -177,6 +181,16 @@ impl StorageNodeConfig {
     }
 }
 
+/// Configuration for metric push
+#[derive(Debug, Clone, Deserialize, Serialize)]
+pub struct MetricsConfig {
+    /// the interval of time we will allow to elapse before pushing metrics
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub push_interval_seconds: Option<u64>,
```
mlegner: Suggestion: Instead of having an option here and the default hidden in an `unwrap_or` in `crates/walrus-service/src/node/push_metrics.rs`, can we remove the `Option` here and have a default? Also, we generally use a `Duration` for time settings (the exception being the weird `rest_graceful_shutdown_period_secs`).

crates/walrus-service/src/node/push_metrics.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    /// join handle for the task
+    pub fn join(&mut self) -> Result<(), anyhow::Error> {
+        tracing::debug!("waiting for the metric push to shutdown...");
+        self.runtime.block_on(&mut self.metric_push_handle)?
+    }
+}
+
+/// create a request client builder that enforces some defaults
```
mlegner: Nit: Here and in other places: We normally try to write docstrings as follows:
- Indicative instead of imperative (except for commands/options of the binaries).
- Starting with a capital letter and ending in a `.`.

crates/walrus-proxy/src/providers/walrus/query.rs
mlegner, line 1: Question: Instead of implementing this from scratch, could we reuse the queries from the `walrus-sui` crate to get the current (and ideally also previous and next) committee?

crates/walrus-proxy/src/providers/mod.rs
mlegner, line 1: Nit: We generally use the "new" Rust naming convention where this would be `crates/walrus-proxy/src/providers.rs`.

crates/walrus-proxy/src/providers/mod.rs
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+/// walrus specific provider implementation for nodes in committee
+pub mod walrus;
```
mlegner: Question (here and in other places): Does this have to be `pub`? Especially if we have the re-export below.

crates/walrus-proxy/src/providers/walrus/provider.rs
```diff
@@ -0,0 +1,168 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// // Copyright (c) Mysten Labs, Inc.
+// // SPDX-License-Identifier: Apache-2.0
```
mlegner: Nit:
```suggestion
```

crates/walrus-proxy/src/metrics.rs
mlegner, line 1: Question: How is this related to all the metrics we already have?

Cargo.toml
suiwombat: fixed

crates/walrus-proxy/src/consumer.rs
suiwombat: renamed, but kept the doc string to explain what values could appear here

crates/walrus-service/src/node/config.rs
suiwombat: this is only for pushing metrics config, it isn't related to metrics_address. I can rename it to metrics_push to help disambiguate it further.

crates/walrus-service/bin/node.rs
suiwombat: sure, i can try. most of this formatting was automatic though from the linters in the contributing guide.  It's possible i messed up a rebase though.  I'll reapply my changes with formatting from main to see if it stays put this time.

crates/walrus-service/bin/node.rs
suiwombat: I'll see if i can get rid of some of the clones, but most were added bc of borrows/ownership issues.

crates/walrus-service/bin/node.rs
suiwombat: i'll redo this comment. i meant that since metrics push is optional, we are "going to see if we need to to wait for metrics to shut down", not should we have a graceful shutdown, that is implemented below in the if let block.

crates/walrus-service/src/node/config.rs
suiwombat: sure! no problem!  I meant to do this and forgot. good suggestion.

crates/walrus-service/src/node/push_metrics.rs
suiwombat: That i'm unsure of. maybe worth having a chat about it. i'll ping you on slack

crates/walrus-proxy/src/providers/walrus/query.rs
suiwombat: absolutely, i didn't know we had them.  i'll look.  is it ok to do that in another PR?  Rebasing this PR is growing difficult due to the age and size.

crates/walrus-service/src/node/push_metrics.rs
suiwombat: i'll try to clean these up

crates/walrus-service/Cargo.toml
suiwombat: i had to remove it from client as well, hope that was ok. it's in the pr now.

crates/walrus-service/src/node/config.rs
suiwombat: reverted

crates/walrus-proxy/src/providers/mod.rs
suiwombat: updated

crates/walrus-proxy/src/metrics.rs
suiwombat: it's for walrus-proxy, not related to any of the walrus-service code.  it provides k8s pod health, and the metrics related to the proxy itself.

crates/walrus-proxy/src/providers/walrus/query.rs
mlegner: Yes, we can do this in a follow-up. In this case, can you add a GitHub issue and a "TODO" comment in this file referencing that issue number, like 
> TODO: Replace this code by calls to functions implemented in `walrus-sui` (#1XXX).

crates/walrus-proxy/src/providers/mod.rs
mlegner: This is not really important, but I'm curious: Is there a reason we need to use the `mod.rs` naming here (and for `crates/walrus-proxy/src/providers/walrus/mod.rs`)?

crates/walrus-proxy/src/providers/walrus/query.rs
```diff
@@ -0,0 +1,309 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
```
mlegner, line 3:
```suggestion

// TODO: Reuse calls from `walrus-sui` here (#1170).

```

crates/walrus-service/src/node/config.rs
```diff
@@ -177,6 +181,23 @@ impl StorageNodeConfig {
...
+    #[serde(default = "push_interval_seconds_default")]
+    pub push_interval_seconds: Duration,
+    /// the url that we will push metrics to
+    pub push_url: String,
+}
+
+/// Configure the default push interval for metrics.
+fn push_interval_seconds_default() -> Duration {
+    Duration::from_secs(60)
+}
```
mlegner, line 201: Nit: We generally implement the default values in the `defaults` module further down in this file.

crates/walrus-service/src/common/utils.rs
```diff
@@ -463,13 +489,137 @@ impl MetricsAndLoggingRuntime {
             .with_json()
             .init();
 
+        let metric_push_handle = match mp_config {
+            Some(mp_config) => {
+                // associate a default tls provider for this runtime
+                let tls_provider = rustls::crypto::ring::default_provider();
+                tls_provider.install_default().expect(
+                    "unable to install default tls provider for rustls in MetricsAndLoggingRuntime",
+                );
```
mlegner, line 312: We try to use `expect` only in cases where we *know* that no error can occur. 

Suggestion: Either describe why no error can occur or return an error instead of panicking:
```suggestion
                tls_provider
                    .install_default()
                    .map_err(|_| anyhow!("unable to install default tls provider for rustls"))?;
```

crates/walrus-service/src/common/utils.rs
```diff
@@ -463,13 +489,137 @@ impl MetricsAndLoggingRuntime {
...
+}
+
+/// Responsible for sending data to walrus-proxy, used within the async scope of
+/// MetricPushRuntime::start.
+async fn push_metrics(
+    network_key_pair: Arc<Secp256r1KeyPair>,
+    client: &reqwest::Client,
+    push_url: &str,
+    registry: &Registry,
+) -> Result<(), anyhow::Error> {
```
mlegner, line 382: Nit (non-blocking): It seems that this could be represented well by a `PushMetricsClient` that stores the `reqwest::Client`, some parameters, and maybe other structs like the key pair or an encoder.

Also, headers that are used on all requests (like content encoding and type) could be set as `default_headers` on the client directly.

crates/walrus-service/bin/node.rs
```diff
@@ -443,12 +467,14 @@ mod commands {
...
     #[cfg(msim)]
     fn monitor_runtimes(
+        mut metrics_runtime: MetricsAndLoggingRuntime,
         mut node_runtime: StorageNodeRuntime,
         mut event_processor_runtime: EventProcessorRuntime,
         exit_listener: oneshot::Receiver<()>,
         cancel_token: CancellationToken,
     ) -> anyhow::Result<()> {
         let monitor_runtime = Runtime::new()?;
+        metrics_runtime.join()?;
```
mlegner: I think this should be moved down to the other `.join()` invocations. This may explain the current simtest failure. cc @halfprice

crates/walrus-service/src/common/utils.rs
```diff
@@ -463,13 +489,137 @@ impl MetricsAndLoggingRuntime {
...
+        }
+    }
+}
+
+/// Create a request client builder that is used to push metrics to mimir.
+fn create_push_client() -> reqwest::Client {
+    reqwest::Client::builder()
+        .timeout(Duration::from_secs(30))
+        .build()
+        .expect("unable to build client")
```
mlegner, line 372: As above, I'd prefer returning an error to panicking if the client cannot be built.

suiwombat: are you sure?  the decision to panic here is in the event that the node is configured to push metrics and it isn't able to build a client to do so.  This is a constructor with few options that shouldn't ever fail - it's more in the class of invariant violations where we shouldn't proceed if it can't be built.  It isn't a error type that is recoverable.  I can refactor this to return an error but there is no meaningful next step after that.  Metrics just won't get pushed out and i'd also speculate that no metrics would be getting scraped since the agent wouldn't have been configured either.

crates/walrus-service/bin/node.rs
suiwombat: yeah i can try that.

crates/walrus-service/src/common/utils.rs
suiwombat: some headers could be, but i'd prefer to save the optimizations for other PRs.  this isn't a hot code path, runs only every 60 seconds, once.  Also, some headers, cannot be pre-calculated, such as the auth header.  It didn't make sense to me to make them in two places, that was the thinking, readability.

The client isn't embedded because under certain error conditions, we recreate it.  It's handled in the async closure and doesn't make sense to me to need to deal with ownership of that client from an outside scope.

crates/walrus-service/src/common/utils.rs
suiwombat: We cannot function with metrics push without this working.  This should crash the node on startup if it cannot do this and is appropriate if the node is configured to push metrics.  There isn't a mode where the code that follows will work without this also first working (once).

crates/walrus-service/src/node/config.rs
suiwombat: I can put it there.  Will do so in another PR.  I almost did that but it felt like a one off to include it there, but i can do so.

crates/walrus-service/src/common/utils.rs
suiwombat: i think it's best to leave this as is.  there isn't anything here that will fail unless something catastrophic happens, which i cannot think of a single one.

crates/walrus-proxy/src/providers/walrus/query.rs
suiwombat: done.

crates/walrus-service/bin/node.rs
suiwombat: done. retrying.

crates/walrus-service/src/common/utils.rs
mlegner: I understand that point and I agree. However, as this is a public function in the library, I find it still inappropriate to panic unless this is explicitly documented. It is also inconsistent with the rest of the code which returns errors in such cases and lets the caller decide if/how to proceed. See for example the `.context("metrics runtime creation failed")?` a few lines above.

In practice, there is no change in behavior as the caller continues to propagate the error.

crates/walrus-proxy/src/providers/mod.rs
suiwombat: probably just a habit of mine. i can revisit the structure again but I think it was because I wanted providers to be isolated from the super space of the crate, which this is how I usually do that.  I understand it's older, but it makes it harder to get at the code imho and easier to decouple it. I could be wrong though.

crates/walrus-service/src/common/utils.rs
mlegner: According to the docs, the "method fails if a TLS backend cannot be initialized, or the resolver cannot load the system configuration."

I agree that this should always succeed; in this case I would describe this in the expect, like `expect("should be able to initialize TLS backend and load system configuration")`.

Issue #1068, opened 2024-10-16
Event processor seems to stop processing events after failing to read one


mlegner:
This issue was discovered when one of the initial storage nodes for Testnet stopped making progress.

When stopping the node some hours later, it reported an error and stacktrace.

The relevant log entries are:
```
Oct 15 07:35:21 walrus1 walrus-node[3271258]: {"timestamp":"2024-10-15T07:35:21.673139Z","level":"ERROR","fields":{"message":"failed to read next checkpoint","next_checkpoint":120348946,"error":"Rest Client Error: 'error sending request for url (https://mysten-rpc.testnet.sui.io/v2/checkpoints/120348946/full)'"},"target":"walrus_event","filename":"crates/walrus-event/src/lib.rs","line_number":266}
Oct 15 08:13:20 walrus1 walrus-node[3271258]: {"timestamp":"2024-10-15T08:13:20.234738Z","level":"WARN","fields":{"message":"very slow batch write","elapsed":"54.137130016","db_name":"\"storage\""},"target":"typed_store::rocks","filename":"/home/runner/.cargo/git/checkouts/sui-6ac459c53b1b685a/40d9ec7/crates/typed-store/src/rocks/mod.rs","line_number":1439,"span":{"name":"maybe_advance_event_cursor"},"spans":[{"messaging.client.id":"jZHaJ4GrAwb+CovvC3Plf9QXCve6DYqEceVTZNv3zXl1tjh127fg/G6LTWTTHuuW","messaging.destination.name":"blob_store","messaging.operation.type":"receive","messaging.system":"sui","otel.kind":"CONSUMER","walrus.blob_id":"None","walrus.event.checkpoint_seq":"120358136","walrus.event.index":237201,"walrus.event.kind":"end-of-checkpoint","walrus.event.tx_digest":"None","name":"blob_store receive"},{"name":"process_event"},{"name":"mark_element_at_index"},{"name":"mark_event_completed"},{"name":"maybe_advance_event_cursor"}]}
Oct 15 08:13:20 walrus1 walrus-node[3271258]: {"timestamp":"2024-10-15T08:13:20.234761Z","level":"WARN","fields":{"message":"very slow batch write","elapsed":"53.93858663","db_name":"\"events\""},"target":"typed_store::rocks","filename":"/home/runner/.cargo/git/checkouts/sui-6ac459c53b1b685a/40d9ec7/crates/typed-store/src/rocks/mod.rs","line_number":1439}
[...]

PR #1075, opened 2024-10-17
feat: Shutdown all runtimes if any one crashes

sadhansood:
The current setup waits to be terminated but runtimes could have crashed behind the scene. This PR tries to address this with putting the runtimes in a join set and any one runtime crashing should trigger the process to shutdown.

Contributes to #1068

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -313,6 +313,7 @@ mod commands {
     use std::time::Duration;
 
     use config::EventProviderConfig;
+    use tokio::task::JoinSet;
     use walrus_core::ensure;
     use walrus_service::utils;
     use walrus_sui::types::NetworkAddress;
@@ -374,7 +375,7 @@ mod commands {
         let cancel_token = CancellationToken::new();
         let (exit_notifier, exit_listener) = oneshot::channel::<()>();
 
-        let (event_manager, mut event_processor_runtime) = EventProcessorRuntime::start(
+        let (event_manager, event_processor_runtime) = EventProcessorRuntime::start(
             config
                 .sui
                 .clone()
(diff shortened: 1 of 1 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -393,16 +394,33 @@ mod commands {
             cancel_token.child_token(),
         )?;
 
-        wait_until_terminated(exit_listener);
+        let monitor_task = tokio::spawn(async move {
+            let mut set = JoinSet::new();
+            set.spawn_blocking(move || node_runtime.join());
+            set.spawn_blocking(move || event_processor_runtime.join());
```
halfprice: question: why event processor is separated from the other 2 processes (rest api, and node)?

crates/walrus-service/bin/node.rs
```diff
@@ -393,16 +394,33 @@ mod commands {
...
-        // Cancel the node runtime, if it is still executing.
-        cancel_token.cancel();
+            // Drain remaining runtimes
+            while set.join_next().await.is_some() {}
+            tracing::info!("All runtimes have shut down");
+        });
 
-        event_processor_runtime.join()?;
+        let monitor_runtime = Runtime::new()?;
+        monitor_runtime.block_on(async {
```
halfprice, line 415: Since this is `block_on`, do we need another runtime to wait for monitor_task to finish? Can we just call `monitor_task.await` instead of creating a new runtime?

sadhansood: Yes, because main() is not initialized with a default tokio runtime

crates/walrus-service/bin/node.rs
sadhansood: I think this was JP's callout to seperate event processor in its own runtime given in future we may have multiple nodes and event processor will run in a different node from actual storage nodes

PR #1087, opened 2024-10-21
perf: add metrics to track shard sync progress

### LINKED ISSUE #1019: Add metrics to track shard sync progress and stats

halfprice:
Hope to be included in the new patch as well.

Closes #1019


crates/walrus-service/src/node/metrics.rs
```diff
@@ -37,6 +37,15 @@ telemetry::define_metric_set! {
...
         (slivers_retrieved_total, "Total number of sliver instances returned", &["sliver_type"]),
+        (shard_sync_total, "The number of shard sync per status", &["status"]),
+        (sync_shard_sync_blob_total,
+            "Total number of blobs synced during shard sync", &["shard"]),
+        (sync_shard_recover_blob_total,
+            "Total number of blobs recovered during shard sync", &["shard"]),
+        (sync_shard_recover_blob_success_total,
+            "Total number of blobs recovered during shard sync", &["shard"]),
+        (sync_shard_recover_blob_error_total,
+            "Total number of blobs recovered during shard sync", &["shard"]),
```
mlegner: Nit: These descriptions need to be changed.

crates/walrus-service/src/node/metrics.rs
```diff
@@ -37,6 +37,15 @@ telemetry::define_metric_set! {
     IntCounterVec: [
         (slivers_stored_total, "The total number of slivers stored", &["sliver_type"]),
         (slivers_retrieved_total, "Total number of sliver instances returned", &["sliver_type"]),
+        (shard_sync_total, "The number of shard sync per status", &["status"]),
```
mlegner, line 40: Question: What is the "status" here?

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -146,6 +146,7 @@ impl ShardSyncHandler {
             );
 
             utils::retry(backoff, || async {
+                metrics::with_label!(node_clone.metrics.shard_sync_total, "start").inc();
```
mlegner, line 149: Remark/question: Given the "retry" wrapper, this means that a shard sync that is retried is counted multiple times. Is this the intention?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -607,6 +612,12 @@ impl ShardStorage {
             }
             batch.write()?;
 
+            metrics::with_label!(
+                node.metrics.sync_shard_sync_blob_total,
+                &self.id.to_string()
+            )
+            .inc_by(fetched_slivers.len() as u64);
```
mlegner, line 619: Important: This seems to be different by a factor of 2: There are twice as many slivers as blobs. So either the metric should be renamed to `sync_shard_sync_sliver_total` or we should only increase this for one sliver type.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -808,15 +819,33 @@ impl ShardStorage {
...
+        metrics::with_label!(
+            node.metrics.sync_shard_recover_blob_total,
+            &self.id.to_string()
+        )
+        .inc();
+
         let result = node
             .committee_service
             .recover_sliver(metadata.into(), sliver_id, sliver_type, epoch)
             .await;
```
mlegner, line 833: Not related to these changes, but possibly important: If the blob is no longer available for some reason (e.g., it was deletable and has been deleted), this may never complete. As a result, this may permanently occupy a semaphore permit in `recover_missing_blobs`, meaning that function may also not make any progress if there are too many such blobs.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -808,15 +819,33 @@ impl ShardStorage {
...
         match result {
-            Ok(sliver) => self.put_sliver(&blob_id, &sliver)?,
+            Ok(sliver) => {
+                metrics::with_label!(
+                    node.metrics.sync_shard_recover_blob_success_total,
+                    &self.id.to_string()
+                )
+                .inc();
+                self.put_sliver(&blob_id, &sliver)?;
+            }
```
mlegner: Important: As above, we seem to be counting _slivers_ here instead of _blobs_.

crates/walrus-service/src/node/metrics.rs
halfprice: The counter counts the number of shard sync in different status. The status can be `start`, `error`, or `complete`.

crates/walrus-service/src/node/shard_sync.rs
halfprice: Yes. The idea is that `shard_sync_total` represents the total number of sync job started. shard_sync_total['start'] - shard_sync_total['complete'] - shard_sync_total['error'] gives the total number of ongoing shard sync jobs.

From debugging purpose, it's more informative to show the job level stats rather than app level stats (the number of shards are being synced), since the former somewhat includes information of the latter.

crates/walrus-service/src/node/storage/shard.rs
halfprice: suggestion accepted.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Metrics renamed.

crates/walrus-service/src/node/storage/shard.rs
halfprice: I believe you are right. This was before blob deletion, but I think similarly, blob expiration would cause same issue.

Let me think about how to address syncing/recovering removed blob problem in general.

Created #1095 and referenced it here.

crates/walrus-service/src/node/metrics.rs
halfprice: This is not `Nit` at all... Thanks for catching it. Was working on this PR pretty late...

PR #1089, opened 2024-10-21
feat(staking): allow joining and splitting stake in withdrawing state

### LINKED ISSUE #906: Staking: enable `split` and `join` for StakedWal in `withdrawing` state
Currently, we store `pool_token_amount` in the `StakedWal`, hence `split` and `join` on an instance need to correctly recalculate the amount of pool tokens. If needed, this feature can be added in a safe upgrade without updating any type signatures or even the main application version.

damirka:
**blocked by #1154**

- closes #906
- allows joining stake in the `Withdrawing` state
- allows splitting stake in the `Withdrawing` state by performing pool_token recalculation

Note: in the _withdrawing_ state we no longer care about `activation_epoch` and instead check that the `withdrawing_epoch` matches. There's a pool test which demonstrates this behaviour.

### DIFF
--- a/contracts/walrus/sources/staking/staked_wal.move
+++ b/contracts/walrus/sources/staking/staked_wal.move
@@ -193,7 +193,7 @@ public fun join(sw: &mut StakedWal, other: StakedWal) {
 /// Aborts if the `amount` is greater than the `principal` of the staked WAL.
 /// Aborts if the `amount` is zero.
 public fun split(sw: &mut StakedWal, amount: u64, ctx: &mut TxContext): StakedWal {
-    assert!(sw.principal.value() >= amount, EInvalidAmount);
+    assert!(sw.principal.value() > amount, EInvalidAmount);
     assert!(amount > 0, EInvalidAmount);
 
     // Simple scenario - the staked WAL is not withdrawing.
--- a/contracts/walrus/tests/staking/staked_wal_tests.move
+++ b/contracts/walrus/tests/staking/staked_wal_tests.move
@@ -69,6 +69,16 @@ fun split_join_early_withdraw() {
 }
 
 #[test, expected_failure(abort_code = staked_wal::EInvalidAmount)]
+fun try_splitting_full_amount() {
+    let ctx = &mut tx_context::dummy();
+    let node_id = ctx.fresh_object_address().to_id();
+    let mut sw = staked_wal::mint(node_id, mint_balance(1000), 1, ctx);
+    let _v = sw.split(1000, ctx);
+
+    abort
+}
+
+#[test, expected_failure(abort_code = staked_wal::EInvalidAmount)]
 fun try_split_zero() {
     let ctx = &mut tx_context::dummy();
     let node_id = ctx.fresh_object_address().to_id();


contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -126,24 +119,42 @@ public fun pool_token_amount(sw: &StakedWal): u64 {
...
+        sw.principal.join(principal);
+        id.delete();
+        return
+    };
+
+    // Withdrawing scenario - we no longer check that the activation epoch is
+    // the same, as the staked WAL is in the process of withdrawing. Instead,
+    // we make sure that the withdraw epoch is the same.
+    assert!(sw.is_withdrawing() && other.is_withdrawing(), EMetadataMismatch);
+    assert!(sw.withdraw_epoch() == other.withdraw_epoch(), EMetadataMismatch);
```
karlwuest, line 169: Suggestion: For now, let's still keep the check that the activation epoch is the same. We can then potentially remove this, once the request for withdrawal works before the activation epoch, depending on how this is implemented. The reason for this is, that in this scenario we may not be able to rely on the pool token amount to be correct in the StakedWal, so I'd prefer if we first know how to handle that before we allow joining stakedWal that does not have the same activation epoch.

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -126,24 +119,42 @@ public fun pool_token_amount(sw: &StakedWal): u64 {
...
+    let StakedWal { id, principal, .. } = other;
+    sw.principal.join(principal);
     id.delete();
 
-    sw.principal.join(principal);
+    match (&mut sw.state) {
+        StakedWalState::Withdrawing { pool_token_amount: current_pool_token_amount, .. } => {
+            *current_pool_token_amount = *current_pool_token_amount + pool_token_amount;
+        },
+        _ => abort 264, // unreachable
```
karlwuest: Suggestion: Add this explicitly as a constant to make sure that we don't use this error code for anything else and to improve readability.

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -153,14 +164,38 @@ public fun join(sw: &mut StakedWal, other: StakedWal) {
...
+                state: StakedWalState::Withdrawing {
+                    withdraw_epoch: *withdraw_epoch,
+                    pool_token_amount: new_pool_token_amount,
+                },
+                node_id: sw.node_id,
+                principal: sw.principal.split(amount),
+                activation_epoch: sw.activation_epoch,
+            }
+        },
+        _ => abort 264, // unreachable
```
karlwuest: See above.

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -191,8 +191,10 @@ public fun join(sw: &mut StakedWal, other: StakedWal) {
 /// same for both the staked WALs.
 ///
 /// Aborts if the `amount` is greater than the `principal` of the staked WAL.
+/// Aborts if the `amount` is zero.
 public fun split(sw: &mut StakedWal, amount: u64, ctx: &mut TxContext): StakedWal {
     assert!(sw.principal.value() >= amount, EInvalidAmount);
```
karlwuest: Let's make this a `>`, otherwise we can end up with a zero value in the original StakedWal.

crates/walrus-sui/src/types/move_errors.rs
```diff
@@ -276,6 +276,8 @@ move_error_kind!(
     4,
     ECantSplitWithdrawing,
     5,
+    EAlreadyWithdrawing,
```
karlwuest, line 279: Just FYI: this will hopefully soon no longer be necessary, I have an open PR to automatically generate rust errors from the move code :)

PR #1091, opened 2024-10-21
fix(node): correct `BlobStatus` for expired blobs

### LINKED ISSUE #1076: Storage node should not return a `BlobStatus::Permanent` or `BlobStatus::Deletable` if those blobs are already expired
This causes unintuitive output on the client (#1037) and can cause issues for the publisher.

mlegner:
Closes #1076

### DIFF
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -291,7 +291,7 @@ pub(crate) trait BlobInfoApi {
     fn invalidation_event(&self) -> Option<EventID>;
 
     /// Converts the blob information to a `BlobStatus` object.
-    fn to_blob_status(&self) -> BlobStatus;
+    fn to_blob_status(&self, current_epoch: Epoch) -> BlobStatus;
 }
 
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -484,37 +484,64 @@ impl From<ValidBlobInfoV1> for BlobInfoV1 {
...
         let deletable_counts = DeletableCounts {
-            count_deletable_total: *count_deletable_total,
-            count_deletable_certified: *count_deletable_certified,
+            count_deletable_total: if self
+                .latest_seen_deletable_registered_epoch
+                .is_some_and(|e| e > current_epoch)
+            {
+                self.count_deletable_total
+            } else {
+                0
```
halfprice: Nit: since below, we check that if `deletable_counts != Default::default()`, we may want to use Default::default() here to make it clearer.

mlegner: Changed the whole expression to 
```rust
self.latest_seen_deletable_registered_epoch
    .is_some_and(|e| e > current_epoch)
    .then_some(self.count_deletable_total)
    .unwrap_or_default()
```
and the next one accordingly.

Issue #1095, opened 2024-10-22
Recovering blob needs to be cancelled when the recovered blob is deleted/expired/invalidated


halfprice:
This needs to be handled in both
- Blob sync during processing certified event
- Blob recovery during shard move

PR #1097, opened 2024-10-22
fix(node): don't start blob sync for already expired blobs

mlegner:
This does not directly solve #1095, but at least makes sure that no blob syncs are started for already expired blobs. This also offers a manual recovery path: Just restart the recovering node after the epoch change.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -713,6 +713,7 @@ impl StorageNode {
 
         if self.inner.storage.is_stored_at_all_shards(&event.blob_id)?
             || self.inner.storage.is_invalid(&event.blob_id)?
+            || self.inner.current_epoch() >= event.end_epoch
         {
             self.inner
                 .mark_event_completed(event_index, &event.event_id)?;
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -2303,6 +2303,56 @@ mod tests {
...
+        retry_until_success_or_timeout(TIMEOUT, || async {
+            node.storage_node
+                .inner
+                .storage
+                .get_blob_info(&BLOB_ID)?
+                .ok_or(anyhow!("blob info not updated"))
+        })
+        .await?;
+
+        assert_eq!(node.storage_node.blob_sync_handler.cancel_all().await?, 0);
```
giac-mysten, line 2352: question: so previously it would hang because the blob info request would trigger a sync that would never complete? interesting...

mlegner: The sync would be triggered by the event, not the info request. The request is just there to make sure that the test success is not just because the node never saw the event. I can add a comment about this.

giac-mysten: makes sense, thanks for the explanation.

mlegner: Done.

PR #1101, opened 2024-10-22
refactor: move events into walrus-service and create walrus-utils crate

sadhansood:
Today, `EventProcessor` is susceptible to Sui FullNode failures and would crash the binary on failure to download checkpoints and events after sometime. But, as we recently saw the full nodes can go down for short periods causing a Walrus outage. But, we shouldn't crash storage nodes because:
1. Walrus can still serve reads
2. Storage nodes which are lagging have potentially already downloaded events to their local db and can safely continue with event recovery even if newer events are not being fetched

This PR does a couple things:
1. Moves `event_processor.rs` into `walrus_service` crate so we can use `ExponentialBackoff`. `EventProcessor` needs to access `EventBlobs` for a future PR, so this cyclic dependency needs to be broken
2. Uses exponential backoff to retry checkpoint failures

### DIFF
--- a/crates/walrus-service/src/client/communication/node.rs
+++ b/crates/walrus-service/src/client/communication/node.rs
@@ -25,6 +25,7 @@ use walrus_sdk::{
     error::NodeError,
 };
 use walrus_sui::types::StorageNode;
+use walrus_utils::backoff::{self, ExponentialBackoff};
 
 use crate::{
     client::{
@@ -32,7 +33,7 @@ use crate::{
         error::{SliverStoreError, StoreError},
         utils::{string_prefix, WeightedResult},
     },
-    common::utils::{self, ExponentialBackoff, FutureHelpers},
+    common::utils::FutureHelpers,
 };
 
 /// Below this threshold, the `NodeCommunication` client will not check if the sliver is present on
@@ -237,7 +238,7 @@ impl<'a, W> NodeCommunication<'a, W> {
         blob_id: &BlobId,
         epoch: Epoch,
     ) -> Result<SignedStorageConfirmation, NodeError> {
-        let confirmation = utils::retry(self.backoff_strategy(), || {
+        let confirmation = backoff::retry(self.backoff_strategy(), || {
             self.client.get_confirmation(blob_id)
         })
         .await
@@ -484,7 +485,7 @@ impl<'a> NodeWriteCommunication<'a> {
         F: FnMut() -> Fut,
         Fut: Future<Output = Result<T, E>>,
     {
-        utils::retry(self.backoff_strategy(), f)
+        backoff::retry(self.backoff_strategy(), f)
             .batch_limit(self.node_write_limit.clone())
             .batch_limit(self.sliver_write_limit.clone())
             .await
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -52,7 +52,6 @@ use walrus_core::{
     SliverPairIndex,
     SliverType,
 };
-use walrus_event::{event_processor::EventProcessor, EventProcessorConfig, IndexedStreamElement};
 use walrus_sdk::api::{
     BlobStatus,
     EventProgress,
@@ -91,6 +90,7 @@ use self::{
 pub mod committee;
 pub mod config;
 pub mod contract_service;
+pub mod events;
 pub mod server;
 pub mod system_events;
 
@@ -116,10 +116,16 @@ use errors::{
     StoreSliverError,
     SyncShardServiceError,
 };
+use events::{
+    event_processor::EventProcessor,
+    EventProcessorConfig,
+    EventStreamCursor,
+    EventStreamElement,
+    IndexedStreamElement,
+};
 
 mod storage;
 pub use storage::{DatabaseConfig, NodeStatus, Storage};
-use walrus_event::{EventStreamCursor, EventStreamElement};
 
 use crate::{
     common::utils::ShardDiff,
@@ -1082,6 +1088,10 @@ impl StorageNode {
         );
 
         Ok(())
+    }
+
+    pub(crate) fn inner(&self) -> &Arc<StorageNodeInner> {
+        &self.inner
     }
 }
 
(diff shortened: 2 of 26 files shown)


scripts/local-testbed.sh
```diff
@@ -114,7 +114,7 @@ if ! $existing; then
     # Generate configs
     echo Generating configuration...
     cargo run --bin walrus-deploy -- generate-dry-run-configs \
-    --working-dir $working_dir --use-legacy-event-provider
+    --working-dir $working_dir
```
mlegner: Question: Does that work with the default fullnode? Or do we have to use the `mysten-rpc.*.sui.io` nodes?

crates/walrus-service/src/events/event_processor.rs
```diff
@@ -235,29 +246,24 @@ impl EventProcessor {
...
-                num_attempts += 1;
-                if start.elapsed() > MAX_TIMEOUT && num_attempts >= MIN_ATTEMPTS {
-                    bail!(
-                        "Failed to read checkpoint from full node: {}",
-                        result.err().unwrap()
-                    );
-                } else {
-                    handle_checkpoint_error(result.err(), next_checkpoint);
+                let sleep_delay = backoff.next_delay();
+                if sleep_delay.is_none() {
```
halfprice: Nit: how about

```
let Some(sleep_delay) = backoff.next_delay() else {
    backoff = create_backoff(&mut rng);
}
...
```

crates/walrus-service/src/events/event_processor.rs
```diff
@@ -235,29 +246,24 @@ impl EventProcessor {
...
+                    backoff = create_backoff(&mut rng);
                     continue;
                 }
+                tokio::time::sleep(sleep_delay.unwrap()).await;
+                handle_checkpoint_error(result.err(), next_checkpoint);
```
halfprice: Suggestion: probably make sense to handle error before sleeping.

scripts/local-testbed.sh
halfprice: +1

crates/walrus-utils/src/backoff.rs
mlegner, line 1: I assume this is just moved from `crates/walrus-service/src/common/utils.rs`, right?

crates/walrus-utils/Cargo.toml
mlegner, line 1: Question: Are all those dependencies still required?

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -9,17 +9,15 @@ use std::{
 use tokio::sync::Mutex;
 use walrus_core::ShardIndex;
 use walrus_sdk::error::ServiceError;
+use walrus_utils::backoff::{retry, ExponentialBackoff};
```
mlegner: Nit: It is generally recommended not to import functions directly (e.g., [here](https://doc.rust-lang.org/book/ch07-04-bringing-paths-into-scope-with-the-use-keyword.html#creating-idiomatic-use-paths)). So better use 
```suggestion
use walrus_utils::backoff::{self, ExponentialBackoff};
```
here and call the function as `backoff::retry`.

crates/walrus-service/src/client/communication/node.rs
```diff
@@ -25,14 +25,15 @@ use walrus_sdk::{
     error::NodeError,
 };
 use walrus_sui::types::StorageNode;
+use walrus_utils::backoff::{retry, ExponentialBackoff};
```
mlegner: Same comment about the function import as above. This may apply to other places as well.

crates/walrus-service/src/lib.rs
```diff
@@ -9,6 +9,8 @@ pub mod client;
 #[cfg(feature = "node")]
 pub mod node;
 
+pub mod events;
+
```
mlegner: Suggestion: Move the `events` module under `node`. Also, does this really have to be `pub`?

crates/walrus-service/src/events/event_blob_writer.rs
```diff
@@ -127,6 +132,7 @@ impl EventBlobWriter {
         })
     }
 
+    /// Create a new file for the next blob.
```
mlegner: Nit: Here and in other places: Docstrings should be in indicative, not imperative form.

crates/walrus-service/src/node.rs
```diff
@@ -1022,9 +1027,21 @@ impl StorageNode {
...
 
 impl StorageNodeInner {
+    pub(crate) fn encoding_config(&self) -> &EncodingConfig {
+        &self.encoding_config
+    }
+
+    pub(crate) fn storage(&self) -> &Storage {
+        &self.storage
+    }
+
```
mlegner: Remark: If the `events` module is a submodule of `node`, these should no longer be needed.

crates/walrus-utils/src/backoff.rs
sadhansood: yes, that is correct.

crates/walrus-service/src/node/shard_sync.rs
sadhansood: done

crates/walrus-service/src/client/communication/node.rs
sadhansood: done

crates/walrus-service/src/node.rs
```diff
@@ -1014,6 +1021,10 @@ impl StorageNode {
 
         Ok(())
     }
+
+    pub(crate) fn inner(&self) -> &Arc<StorageNodeInner> {
+        &self.inner
+    }
```
mlegner, line 1095: Question: Do we need this?

crates/walrus-service/src/events/event_blob_writer.rs
mlegner: There are still a few of them, but this is not blocking. We anyway have to clean up some stuff, can be done as part of #1153.

crates/walrus-service/src/node.rs
sadhansood: Ah no, I actually removed it in the stacked pr but i can remove it in this pr itself.

PR #1108, opened 2024-10-24
feat: pipelined checkpoint downloader to increase throughput

### LINKED ISSUE #1104: Increase performance of checkpoint processing
For storage nodes that are not close to a full node, checkpoint processing is very slow: There is a node in Brazil that can just process around 5 cp/s, where checkpoints are produced at a rate of ~4 cp/s. This means they take extremely long to catch up if they ever fall behind, thus delaying recovery.

sadhansood:
This is to address the issue where some nodes are having trouble keeping up with checkpoint generation rate because of higher network latency.

Closes #1104

### DIFF
--- a/crates/walrus-event/src/checkpoint_downloader.rs
+++ b/crates/walrus-event/src/checkpoint_downloader.rs
@@ -1,0 +1,636 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{
+    collections::{BTreeMap, BTreeSet},
+    fmt::Debug,
+    sync::Arc,
+    time::Duration,
+};
+
+use anyhow::{anyhow, Result};
+use prometheus::{register_int_gauge_with_registry, IntGauge, Registry};
+use serde::{Deserialize, Serialize};
+use sui_rest_api::{client::sdk, Client};
+use sui_types::{
+    full_checkpoint_content::CheckpointData,
+    messages_checkpoint::{CheckpointSequenceNumber, TrustedCheckpoint},
+};
+use tokio::{
+    sync::{mpsc, RwLock},
+    time::Instant,
+};
+use tokio_util::sync::CancellationToken;
+use typed_store::{rocks::DBMap, Map};
+
+/// Fetcher configuration options for the parallel checkpoint fetcher.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct ParallelDownloaderConfig {
+    /// Number of retries per checkpoint before giving up.
+    pub min_retries: u32,
+    /// Delay between retries.
+    pub retry_delay: Duration,
+}
+
+impl Default for ParallelDownloaderConfig {
+    fn default() -> Self {
+        // Checkpoint downloader configuration tuned for:
+        // - Checkpoint generation rate: 1 per 200ms (5 per second)
+        // - Network RTT: ~100-300ms
+        //
+        // Retry timing example with 100ms RTT:
+        // T+0ms:   Client requests checkpoint N
+        // T+100ms: Client learns checkpoint N unavailable
+        // T+200ms: Server generates checkpoint N
+        // T+250ms: Client retries after retry_delay
+        // T+350ms: Client gets response which should contain checkpoint N.
+
+        // Retry timing example with 300ms RTT:
+        // T+0ms:   Client requests checkpoint N
+        // T+200ms: Server generates checkpoint N
+        // T+300ms: Client learns checkpoint N unavailable
+        // T+450ms: Client retries after retry_delay
+        // T+750ms: Client gets response which should contain checkpoint N.
+        // The retry_delay of 150ms is chosen to balance in steady state with
+        // fully caught up client:
+        // - Not retrying too quickly (which could waste network resources)
+        // - Not waiting too long (which would cause us to fall behind)
+        // TODO: Use adaptive retry delay based on RTT and error type (issue-1123)
+        Self {
+            min_retries: 10,
+            retry_delay: Duration::from_millis(150),
+        }
+    }
+}
+
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct ChannelConfig {
+    /// Additional buffer factor for work queue
+    /// Helps keep the worker busy in case of delays
+    /// in draining the checkpoint queue.
+    pub work_queue_buffer_factor: usize,
+    /// Buffer factor for result queue
+    /// Helps handle delays in processing the results.
+    pub result_queue_buffer_factor: usize,
+}
+
+impl ChannelConfig {
+    pub fn default() -> Self {
+        Self {
+            work_queue_buffer_factor: 3,
+            result_queue_buffer_factor: 3,
+        }
+    }
+}
+
+/// Configuration for the adaptive worker pool.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct AdaptiveDownloaderConfig {
+    /// Minimum number of workers.
+    pub min_workers: usize,
+    /// Maximum number of workers.
+    pub max_workers: usize,
+    /// Initial number of workers.
+    pub initial_workers: usize,
+    /// Checkpoint lag threshold for scaling up.
+    pub scale_up_lag_threshold: u64,
+    /// Checkpoint lag threshold for scaling down.
+    pub scale_down_lag_threshold: u64,
+    /// Minimum time between scaling operations.
+    pub scale_cooldown: Duration,
+    /// Base configuration.
+    pub base_config: ParallelDownloaderConfig,
+    /// Channel configuration.
+    pub channel_config: ChannelConfig,
+}
+
+impl Default for AdaptiveDownloaderConfig {
+    fn default() -> Self {
+        Self {
+            min_workers: 1,
+            max_workers: 10,
+            initial_workers: 5,
+            scale_up_lag_threshold: 100,
+            scale_down_lag_threshold: 10,
+            scale_cooldown: Duration::from_secs(10),
+            base_config: ParallelDownloaderConfig::default(),
+            channel_config: ChannelConfig::default(),
+        }
+    }
+}
+
+impl AdaptiveDownloaderConfig {
+    fn message_queue_size(&self) -> usize {
+        self.max_workers * self.channel_config.work_queue_buffer_factor
+    }
+
+    fn checkpoint_queue_size(&self) -> usize {
+        self.max_workers * self.channel_config.work_queue_buffer_factor
+    }
+
+    fn result_queue_size(&self) -> usize {
+        self.max_workers
+            * self.channel_config.work_queue_buffer_factor
+            * self.channel_config.result_queue_buffer_factor
+    }
+}
+
+/// Metrics for the event processor.
+#[derive(Clone, Debug)]
+struct AdaptiveDownloaderMetrics {
+    /// The number of checkpoint downloader workers.
+    pub num_workers: IntGauge,
+    /// The current checkpoint lag between the local store and the full node.
+    pub checkpoint_lag: IntGauge,
+}
+
+impl AdaptiveDownloaderMetrics {
+    pub fn new(registry: &Registry) -> Self {
+        Self {
+            num_workers: register_int_gauge_with_registry!(
+                "checkpoint_downloader_num_workers",
+                "Number of workers active in the worker pool",
+                registry,
+            )
+            .expect("this is a valid metrics registration"),
+            checkpoint_lag: register_int_gauge_with_registry!(
+                "checkpoint_downloader_checkpoint_lag",
+                "Current checkpoint lag between local store and full node",
+                registry,
+            )
+            .expect("this is a valid metrics registration"),
+        }
+    }
+}
+
+/// Worker message types.
+enum WorkerMessage {
+    /// Download a checkpoint with the given sequence number.
+    Download(CheckpointSequenceNumber),
+    /// Shutdown the worker.
+    Shutdown,
+}
+
+/// Entry in the checkpoint fetcher queue.
+#[derive(Debug)]
+pub struct CheckpointEntry {
+    pub sequence_number: u64,
+    pub result: Result<CheckpointData>,
+}
+
+impl CheckpointEntry {
+    pub fn new(sequence_number: u64, result: Result<CheckpointData>) -> Self {
+        Self {
+            sequence_number,
+            result,
+        }
+    }
+}
+
+/// Configuration for the pool monitor.
+#[derive(Clone)]
+struct PoolMonitorConfig {
+    downloader_config: AdaptiveDownloaderConfig,
+    checkpoint_store: DBMap<(), TrustedCheckpoint>,
+    client: Client,
+    metrics: AdaptiveDownloaderMetrics,
+}
+
+/// Channels for the pool monitor.
+#[derive(Clone)]
+struct PoolMonitorChannels {
+    message_sender: async_channel::Sender<WorkerMessage>,
+    message_receiver: async_channel::Receiver<WorkerMessage>,
+    checkpoint_sender: mpsc::Sender<CheckpointEntry>,
+}
+
+/// Parallel checkpoint downloader that fetches checkpoints in parallel.
+#[derive(Clone)]
+pub struct ParallelCheckpointDownloader {
+    full_node_client: Client,
+    checkpoint_store: DBMap<(), TrustedCheckpoint>,
+    config: AdaptiveDownloaderConfig,
+    metrics: AdaptiveDownloaderMetrics,
+}
+
+impl ParallelCheckpointDownloader {
+    /// Creates a new parallel checkpoint downloader.
+    pub fn new(
+        full_node_client: Client,
+        checkpoint_store: DBMap<(), TrustedCheckpoint>,
+        config: AdaptiveDownloaderConfig,
+        registry: &Registry,
+    ) -> Result<Self> {
+        let metrics = AdaptiveDownloaderMetrics::new(registry);
+        Ok(Self {
+            full_node_client,
+            checkpoint_store,
+            config,
+            metrics,
+        })
+    }
+
+    /// Starts fetching checkpoints from the given sequence number.
+    pub fn start(
+        &self,
+        sequence_number: CheckpointSequenceNumber,
+        cancellation_token: CancellationToken,
+    ) -> mpsc::Receiver<CheckpointEntry> {
+        let inner = ParallelCheckpointDownloaderInner::new(
+            self.full_node_client.clone(),
+            self.checkpoint_store.clone(),
+            self.config.clone(),
+            cancellation_token,
+            self.metrics.clone(),
+        );
+        inner.start(sequence_number)
+    }
+}
+
+/// Parallel checkpoint downloader that fetches checkpoints in parallel.
+struct ParallelCheckpointDownloaderInner {
+    message_sender: async_channel::Sender<WorkerMessage>,
+    checkpoint_receiver: mpsc::Receiver<CheckpointEntry>,
+    config: AdaptiveDownloaderConfig,
+    worker_count: Arc<RwLock<usize>>,
+    cancellation_token: CancellationToken,
+}
+
+impl ParallelCheckpointDownloaderInner {
+    /// Creates a new parallel checkpoint downloader.
+    fn new(
+        full_node_client: Client,
+        checkpoint_store: DBMap<(), TrustedCheckpoint>,
+        config: AdaptiveDownloaderConfig,
+        cancellation_token: CancellationToken,
+        metrics: AdaptiveDownloaderMetrics,
+    ) -> Self {
+        let (message_sender, message_receiver) =
+            async_channel::bounded(config.message_queue_size());
+        let (checkpoint_sender, checkpoint_receiver) =
+            mpsc::channel(config.checkpoint_queue_size());
+
+        for worker_id in 0..config.initial_workers {
+            let full_node_client = full_node_client.clone();
+            let cloned_message_receiver = message_receiver.clone();
+            let config = config.clone();
+            let cloned_checkpoint_sender = checkpoint_sender.clone();
+            Self::spawn_new_worker(
+                worker_id,
+                full_node_client,
+                cloned_message_receiver,
+                cloned_checkpoint_sender,
+                config.base_config,
+            );
+        }
+
+        let worker_count = Arc::new(RwLock::new(config.initial_workers));
+        let cloned_checkpoint_store = checkpoint_store.clone();
+        let cloned_cancel_token = cancellation_token.clone();
+        let cloned_client = full_node_client.clone();
+        let cloned_message_sender = message_sender.clone();
+        let cloned_config = config.clone();
+        let cloned_worker_count = worker_count.clone();
+        tokio::spawn(async move {
+            let config = PoolMonitorConfig {
+                downloader_config: cloned_config.clone(),
+                checkpoint_store: cloned_checkpoint_store,
+                client: cloned_client,
+                metrics,
+            };
+            let channels = PoolMonitorChannels {
+                message_sender: cloned_message_sender,
+                message_receiver,
+                checkpoint_sender,
+            };
+            Self::pool_monitor(config, channels, cloned_worker_count, cloned_cancel_token).await?;
+            anyhow::Ok(())
+        });
+
+        Self {
+            message_sender,
+            checkpoint_receiver,
+            config,
+            worker_count,
+            cancellation_token,
+        }
+    }
+
+    fn spawn_new_worker(
+        worker_id: usize,
+        full_node_client: Client,
+        message_receiver: async_channel::Receiver<WorkerMessage>,
+        checkpoint_sender: mpsc::Sender<CheckpointEntry>,
+        config: ParallelDownloaderConfig,
+    ) {
+        tokio::spawn(async move {
+            Self::worker_loop(
+                worker_id,
+                full_node_client,
+                message_receiver,
+                checkpoint_sender,
+                config,
+            )
+            .await?;
+            anyhow::Ok(())
+        });
+    }
+
+    /// Returns the current checkpoint lag between the local store and the full node.
+    async fn current_checkpoint_lag(
+        checkpoint_store: &DBMap<(), TrustedCheckpoint>,
+        client: &Client,
+    ) -> Result<u64> {
+        let Ok(Some(current_checkpoint)) = checkpoint_store.get(&()) else {
+            return Err(anyhow!("Failed to fetch current checkpoint"));
+        };
+        let latest_checkpoint = client.get_latest_checkpoint().await?;
+        let current_lag =
+            latest_checkpoint.sequence_number - current_checkpoint.inner().sequence_number;
+        Ok(current_lag)
+    }
+
+    /// Worker loop that fetches checkpoints.
+    async fn worker_loop(
+        worker_id: usize,
+        client: Client,
+        message_receiver: async_channel::Receiver<WorkerMessage>,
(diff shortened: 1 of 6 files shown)


crates/walrus-event/src/event_processor.rs
```diff
@@ -238,39 +229,37 @@ impl EventProcessor {
...
+            ensure!(
+                *checkpoint.checkpoint_summary.sequence_number() == next_checkpoint,
+                "Received out of order checkpoint: expected {}, got {}",
+                next_checkpoint,
+                checkpoint.checkpoint_summary.sequence_number()
+            );
             self.metrics
                 .latest_downloaded_checkpoint
-                .set(next_checkpoint as i64);
+                .set(checkpoint.checkpoint_summary.sequence_number as i64);
```
mlegner: Nit: This change is not necessary AFAICT.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub max_timeout: Duration,
+}
+
+impl Default for PipelinedDownloaderConfig {
+    fn default() -> Self {
+        Self {
+            window_size: 5,
+            min_retries: 10,
+            retry_delay: Duration::from_millis(250),
+            max_timeout: Duration::from_secs(60),
```
mlegner: Question: How were these values chosen?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,224 @@
...
+use serde::{Deserialize, Serialize};
+use sui_rest_api::Client;
+use sui_types::{
+    full_checkpoint_content::CheckpointData,
+    messages_checkpoint::CheckpointSequenceNumber,
+};
+use tokio::{sync::mpsc, time::Instant};
+use tokio_util::sync::CancellationToken;
+
+use crate::handle_checkpoint_error;
```
mlegner: Nit: This function is only used in this module and could be moved here.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,224 @@
...
+        client: &Client,
+        sequence: CheckpointSequenceNumber,
+        config: &PipelinedDownloaderConfig,
+    ) -> CheckpointEntry {
+        let mut attempts = 0;
+        let start = Instant::now();
+        while attempts < config.min_retries || start.elapsed() < config.max_timeout {
+            let res = client.get_full_checkpoint(sequence).await;
+            let Ok(checkpoint) = res else {
+                handle_checkpoint_error(res.err(), sequence);
```
mlegner: This logs an `ERROR`, even though this is not necessarily problematic as we have retries. 

Suggestion: Change that to a `WARN` and only log an `ERROR` if this is unsuccessful even after retries.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,224 @@
...
+            }
+        });
+
+        result_rx
+    }
+
+    /// Downloads a checkpoint with retries.
+    async fn download_with_retry(
+        client: &Client,
+        sequence: CheckpointSequenceNumber,
```
mlegner: Nit: Can we call this `sequence_number`?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,224 @@
...
+        let (checkpoint_sender, checkpoint_receiver) = mpsc::channel(config.window_size);
+
+        for worker_id in 0..config.window_size {
+            let client = client.clone();
+            let job_receiver = job_receiver.clone();
+            let config = config.clone();
+            let cloned_sender = checkpoint_sender.clone();
+            tokio::spawn(async move {
+                Self::worker_loop(worker_id, client, job_receiver, cloned_sender, config).await?;
+                anyhow::Ok::<()>(())
```
mlegner: Nit: I think the generic type is not necessary here.
```suggestion
                anyhow::Ok(())
```

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,224 @@
...
+    checkpoint_seq_num_sender: mpsc::Sender<u64>,
+    checkpoint_receiver: mpsc::Receiver<CheckpointEntry>,
+    config: PipelinedDownloaderConfig,
+}
+
+impl PipelinedCheckpointDownloader {
+    /// Creates a new pipelined checkpoint downloader.
+    pub fn new(client: Client, config: PipelinedDownloaderConfig) -> Result<Self> {
+        let (checkpoint_seq_num_sender, checkpoint_seq_num_receiver) =
+            mpsc::channel(config.window_size);
```
mlegner: Important/question: The "single consumer" part of this channel is not really a good match for how it is used. I see that you work around this by adding a mutex, but it seems to me that something like [flume](https://github.com/zesterer/flume) or [async-channel](https://github.com/smol-rs/async-channel) might be better suited. Did you look into those?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,224 @@
...
+            sender.send(entry).await?;
+        }
+        tracing::info!("Checkpoint fetcher worker {} shutting down", worker_id);
+        Ok(())
+    }
+
+    /// Starts fetching checkpoints from the given sequence number.
+    pub fn start(
+        mut self,
+        start_sequence: CheckpointSequenceNumber,
```
mlegner: Nit: Can we use `sequence_number` instead of `sequence` everywhere?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,224 @@
...
+                        tracing::error!("Failed to send job to workers - channel closed");
+                        return;
+                    }
+                    in_flight.insert(sequence);
+                    sequence += 1;
+                }
+
+                if in_flight.is_empty() {
+                    break;
+                }
```
mlegner: Question: How can this happen? We just added new numbers above.

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: Not very thoroughly, but i've added some comments justifying the reasoning

crates/walrus-event/src/event_processor.rs
sadhansood: fixed

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: fixed

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: fixed

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: fixed

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: fixed

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: done

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: good call out, I wasn't aware of these crates and upon noticing tokio doesn't have an mpmc queue, I just used a mutex. But these crates look good, I eneded up using `async-channel`

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: yeah good point, it shouldn't happen unless user chooses bad config values i.e. `initial_workers: 0`

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,628 @@
...
+                checkpoint_store: cloned_checkpoint_store,
+                client: cloned_client,
+                metrics,
+            };
+            let channels = PoolMonitorChannels {
+                message_sender: cloned_message_sender,
+                message_receiver,
+                checkpoint_sender,
+            };
+            Self::pool_monitor(config, channels, cloned_worker_count, cloned_cancel_token).await?;
```
halfprice, line 306: Seems the error in the pool monitor is silently dropped? We may not want to fail the node, but I guess we want to retry restarting the pool?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,628 @@
...
+        config: PoolMonitorConfig,
+        channels: PoolMonitorChannels,
+        worker_count: Arc<RwLock<usize>>,
+        cancellation_token: CancellationToken,
+    ) -> Result<()> {
+        let downloader_config = config.downloader_config.clone();
+        let mut next_worker_id = downloader_config.initial_workers;
+        let mut last_scale = Instant::now();
+
+        while !cancellation_token.is_cancelled() {
```
halfprice: Question: cancellation doesn't happen immediately, is this part of the design?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,628 @@
...
+            }
+
+            let Ok(lag) =
+                Self::current_checkpoint_lag(&config.checkpoint_store, &config.client).await
+            else {
+                tracing::error!("Failed to download current checkpoint lag");
+                continue;
+            };
+            let num_workers = *worker_count.read().await;
+            config.metrics.num_workers.set(num_workers as i64);
```
halfprice: Nit: feel it's better to set the num_workers after it is changed. Otherwise, the metric has some lag between when the value is changed and when observed.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,628 @@
...
+    async fn download_with_retry(
+        client: &Client,
+        sequence_number: CheckpointSequenceNumber,
+        config: &PipelinedDownloaderConfig,
+    ) -> CheckpointEntry {
+        let mut attempts = 0;
+        let start = Instant::now();
+        let mut err: Option<_> = None;
+        while attempts < config.min_retries || start.elapsed() < config.max_timeout {
+            let res = client.get_full_checkpoint(sequence_number).await;
```
halfprice: Question: what if the checkpoint corresponding to the sequence_number doesn't exist yet? Rest API seems to return [CheckpointNotFoundError](https://github.com/MystenLabs/sui/blob/13c1ec7b2a7a18e304444d63277c20f8f52a1bc2/crates/sui-rest-api/src/checkpoints.rs#L443), will this cause the `start_tailing_checkpoints` loop to quit?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,628 @@
...
+            let Ok(checkpoint) = res else {
+                err = res.err();
+                handle_checkpoint_error(err.as_ref(), sequence_number);
+                attempts += 1;
+                tokio::time::sleep(config.retry_delay).await;
```
halfprice: Suggestion (can be done in a followup PR): I think we want to use exponential backoff here. Retrying with such short delay is a receipt for even higher load on the fullnode under stress condition.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,628 @@
...
+                }
+
+                if result_tx.send(entry).await.is_err() {
+                    tracing::debug!("Result receiver dropped, stopping checkpoint fetcher");
+                    break;
+                }
+
+                next_expected += 1;
+
+                while let Some(checkpoint) = pending.remove(&next_expected) {
```
halfprice, line 477: ~~Remark: previously, after we download the checkpoint, we process it immediately. But now, the checkpoint is copied twice: one from worker to the downloader, and one from downloader to the event processor. Given the size of the checkpoint may not be very small, I wonder if we should store it in a common place similar to the `pending` map, and only pass the sequence number around.~~


Sorry, I was in the wrong mindset when reading the code (was thinking about transaction_manager in sui where we need to only share reference for some struct). I think the checkpoints are not copied. So please ignore my comment.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,628 @@
...
+        rocks,
+        rocks::{errors::typed_store_err_from_rocks_err, MetricConf, ReadWriteOptions},
+    };
+
+    use super::*;
+    use crate::tests::global_test_lock;
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
+    async fn test_pipelined_fetcher() -> Result<()> {
+        let client = Client::new("http://localhost:9000");
```
halfprice, line 568: Question: how does this test work? I'm not sure if I understand how the checkpoint is fetched given that there isn't a SUI cluster.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,628 @@
...
+                in_flight.remove(&entry.sequence_number);
+
+                if entry.sequence_number != next_expected {
+                    // Store out-of-order response
+                    pending.insert(entry.sequence_number, entry);
+                    continue;
+                }
+
+                if result_tx.send(entry).await.is_err() {
+                    tracing::debug!("Result receiver dropped, stopping checkpoint fetcher");
```
halfprice: Nit: feel the logging level should be `info!` at least. Similar below

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: i think this task will not fail on its but only as part of larger shutdown i.e. cancellation token getting cancelled. In `pool_monitor()` there is only place task can return with an error which is:
```
channels
                    .message_sender
                    .await
                    .context("Failed to send shutdown message to worker")?;
```
which only happens on channel close which only happens on cancellation token getting cancelled

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: Can you elaborate, sorry I did not fully understand the comment.

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: Yes, for now you are right. But my hope is we can land this [PR](https://github.com/MystenLabs/walrus/pull/1101) where we never crash the node on failing to fetch a checkpoint but only retry with backoffs

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: Agree, this [PR](https://github.com/MystenLabs/walrus/pull/1101) adds the exponential backoff which I am hoping to land soon,

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: done

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: So even if we cannot fetch the checkpoint (because there is no live cluster), the failures still return `CheckpointEntry` which has checkpoint sequence number. We can test that checkpoint sequence numbers are being returned in order (even though through failures)

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: done

crates/walrus-event/src/checkpoint_downloader.rs
halfprice: What I meant was when cancellation_token.cancel() is called, the checkpoint downloader may inside the loop and not quit immediately. Is this expected?

crates/walrus-event/src/checkpoint_downloader.rs
halfprice: I see. Will the new pipelined downloader being exercised in simtest?

crates/walrus-event/src/checkpoint_downloader.rs
halfprice: Can you do the similar thing here? It seems that we just need to replace `bail!("Failed to download checkpoint: {}", entry.sequence_number);` with a `handle_checkpoint_error` and a delayed continue. Otherwise, I don't this PR cannot be merged before #1101

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: yes

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: ok, happy to put it inside a select! but i was just trying to avoid the extra layer of indentations that makes the code complex to navigate. Also, notice we still have few await points inside the select! where function could be stuck at when cancellation token is cancelled.

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: The only thing we can do is to make `download_with_retry` not return with failures to achieve the same thing (given this is a different design with introduction of channels, etc) but hopefully with the other pr landed soon, we can make it a retry with exponential backoff soon.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+        // T+0ms:   Client requests checkpoint N
+        // T+200ms: Server generates checkpoint N
+        // T+300ms: Client learns checkpoint N unavailable
+        // T+450ms: Client retries after retry_delay
+        // T+750ms: Client gets response which should contain checkpoint N.
+        // The retry_delay of 150ms is chosen to balance in steady state with
+        // fully caught up client:
+        // - Not retrying too quickly (which could waste network resources)
+        // - Not waiting too long (which would cause us to fall behind)
+        // TODO: Use adaptive retry delay based on RTT and error type.
```
mlegner: Nit: Can you create an issue for that and include the issue number here?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+    sync::{mpsc, RwLock},
+    time::Instant,
+};
+use tokio_util::sync::CancellationToken;
+use typed_store::{rocks::DBMap, Map};
+
+/// Fetcher configuration options for the pipelined checkpoint fetcher.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct PipelinedDownloaderConfig {
+    /// Number of retries per checkpoint before giving up
```
mlegner: Nit (here and in other places): We normally terminate docstrings with a `.`.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+use tokio::{
+    sync::{mpsc, RwLock},
+    time::Instant,
+};
+use tokio_util::sync::CancellationToken;
+use typed_store::{rocks::DBMap, Map};
+
+/// Fetcher configuration options for the pipelined checkpoint fetcher.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+pub struct PipelinedDownloaderConfig {
```
mlegner: Question: Does this and other structs have to be `pub`?

mlegner: Nit: Strictly speaking, "pipelined" is probably not the right term for how the checkpoints are downloaded; "parallel" would be more accurate.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+}
+
+/// Pipelined checkpoint downloader that fetches checkpoints in parallel.
+struct PipelinedCheckpointDownloaderInner {
+    message_sender: async_channel::Sender<WorkerMessage>,
+    checkpoint_receiver: mpsc::Receiver<CheckpointEntry>,
+    config: AdaptiveDownloaderConfig,
+    worker_count: Arc<RwLock<usize>>,
+    cancellation_token: CancellationToken,
+}
```
mlegner: Question: Why do we need both `PipelinedCheckpointDownloader` and `PipelinedCheckpointDownloaderInner`?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+    checkpoint_sender: mpsc::Sender<CheckpointEntry>,
+}
+
+/// Pipelined checkpoint downloader that fetches checkpoints in parallel.
+#[derive(Clone)]
+pub struct PipelinedCheckpointDownloader {
+    full_node_client: Client,
+    checkpoint_store: DBMap<(), TrustedCheckpoint>,
+    config: AdaptiveDownloaderConfig,
+    metrics: AdaptiveDownloaderMetrics,
```
mlegner, line 213: Suggestion: Replace these by a single field `inner: PipelinedCheckpointDownloaderInner` (if we actually need both structs).

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+                Self::worker_loop(
+                    worker_id,
+                    full_node_client,
+                    cloned_message_receiver,
+                    cloned_checkpoint_sender,
+                    config.base_config,
+                )
+                .await?;
+                anyhow::Ok(())
+            });
```
mlegner: This seems to be duplicated with the scaling up in `pool_monitor`.

Suggestion: Create a new method `spawn_new_worker`.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+        client: &Client,
+    ) -> Result<u64> {
+        let Ok(Some(current_checkpoint)) = checkpoint_store.get(&()) else {
+            return Err(anyhow!("Failed to fetch current checkpoint"));
+        };
+        let latest_checkpoint = client.get_latest_checkpoint().await?;
+        let current_lag =
+            latest_checkpoint.sequence_number - current_checkpoint.inner().sequence_number;
+        Ok(current_lag)
+    }
```
mlegner, line 351: Important: It feels a bit wasteful to get the latest checkpoint every second just to throw it away. This is something like a 20% overhead. Is there a way to just get the latest checkpoint *number*? Alternatively, can we store the latest checkpoint in the "pending" checkpoints?

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+                    }
+
+                    let lag = Self::current_checkpoint_lag(
+                        &config.checkpoint_store, &config.client).await?;
+                    config.metrics.checkpoint_lag.set(lag as i64);
+                    let num_workers = *worker_count.read().await;
+                    let mut new_workers = num_workers;
+                    if lag > downloader_config.scale_up_lag_threshold &&
+                        num_workers < downloader_config.max_workers {
+                        new_workers = (num_workers + 1).min(downloader_config.max_workers);
```
mlegner: 2 nits:
* The name `new_workers` could be misunderstood as meaning "newly created workers".
* The `.min(downloader_config.max_workers)` is not needed as we checked that `num_workers < downloader_config.max_workers`.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+                            Self::worker_loop(next_worker_id, cloned_client, cloned_receiver,
+                                cloned_checkpoint_sender, cloned_config).await?;
+                            anyhow::Ok(())
+                        });
+                        *worker_count.write().await = new_workers;
+                        last_scale = now;
+                        next_worker_id += 1;
+                    } else if lag < downloader_config.scale_down_lag_threshold &&
+                            num_workers > downloader_config.min_workers {
+                        new_workers = (num_workers - 1).max(downloader_config.min_workers);
```
mlegner: Nits: Same as above.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+            let mut pending = BTreeMap::new();
+            let mut next_expected = sequence_number;
+
+            while !self.cancellation_token.is_cancelled() {
+                let num_workers = *self.worker_count.read().await;
+                let max_in_flight_requests =
+                    num_workers * self.config.channel_config.work_queue_buffer_factor;
+                while in_flight.len() < max_in_flight_requests {
+                    if message_sender
+                        .send(WorkerMessage::Download(sequence_number))
```
mlegner, line 447: Suggestion: Could we somehow keep track of the expected latest checkpoint sequence number and only send `WorkerMessage::Download` messages for sequence numbers that are at most 1 higher than that? Could also be done in a follow-up.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+
+impl Default for AdaptiveDownloaderConfig {
+    fn default() -> Self {
+        Self {
+            min_workers: 1,
+            max_workers: 10,
+            initial_workers: 2,
+            scale_up_lag_threshold: 100,
+            scale_down_lag_threshold: 10,
+            scale_cooldown: Duration::from_secs(30),
```
mlegner: Suggestion: Reduce this (to 10s?) to make this more reactive. As far as I understand, there are two main situations for the checkpoints: (1) The steady state, where the node downloads checkpoints as fast as they are produced and (2) a node catching up, where it wants to downloads checkpoints as fast as possible. The transition between those is pretty abrupt, so ideally the down-/upscaling should also be relatively fast.

crates/walrus-event/src/checkpoint_downloader.rs
```diff
@@ -0,0 +1,620 @@
...
+    /// Channel configuration.
+    pub channel_config: ChannelConfig,
+}
+
+impl Default for AdaptiveDownloaderConfig {
+    fn default() -> Self {
+        Self {
+            min_workers: 1,
+            max_workers: 10,
+            initial_workers: 2,
```
mlegner: Suggestion: Increase this closer to the max: When a node starts, it probably has to download a larger number of checkpoints and thus benefits of a larger number of initial workers.

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: This is because we need to invoke `start` with a non-mutable `&self` reference as it is called from `start` in `event_processor` which itself is invoked with non-mutable `&self` reference. Hence, we store`PipelinedCheckpointDownloader` in `EventProcessor` struct but inside `start` create `PipelinedCheckpointDownloaderInner` which we can mutate.

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: Sure, nice idea. Will do in a follow up. Thanks!

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: Keeping it for now, as mentioned we need to create `PipelinedCheckpointDownloaderInner` inside event processor `start` and be able to mutate it.

crates/walrus-event/src/checkpoint_downloader.rs
sadhansood: So, here we are only downloading the checkpoint summary which is just the header of a checkpoint while the actual checkpoint is much much larger with all txns, effects, objects and events.

PR #1112, opened 2024-10-28
feat: add repair DB command in walrus-node binary

halfprice:
Test the command locally

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -143,6 +143,8 @@ enum Commands {
     },
 
     /// Repair a corrupted RocksDB database due to non-clean shutdowns.
+    /// Hidden command for emergency use only.
+    #[clap(hide = true)]
     RepairDb {
         #[clap(long)]
         /// Path to the RocksDB database directory.


crates/walrus-service/bin/node.rs
```diff
@@ -618,6 +630,15 @@ mod commands {
         Ok(())
     }
 
+    pub fn repair_db(db_path: PathBuf) -> anyhow::Result<()> {
+        let mut opts = Options::default();
+
+        // In case the integrity of the entire database is compromised.
+        opts.create_if_missing(true);
+        opts.set_max_open_files(512_000);
+        DB::repair(&opts, db_path).map_err(|err| err.into())
```
sadhansood, line 639: I may also suggest to try and open the db after repair done with enabled `set_paranoid_checks(true)` to check the db for any errors and fail if so.

sadhansood: Actually, I am not sure if this is effective during repair command. So, ignore my suggestion. I think the guarantees of this cli command is unclear (as to whether how much of the data is recovered, if the db is consistent after repair and if it safe to operate the node after repair). We can still however add this command as long as we are careful while suggesting its usage.

halfprice: I think when a DB in a corruption state, this is the only hope to bring it back to life other than wipe out. I haven't incorporate your suggestion of using different consistency mode, since the default one seems to give the best consistency guarantee. We've used this to fix one corruption in SUI so I'll keep an eye on whether other modes are needed.

PR #1118, opened 2024-10-29
fix(client): various publisher performance improvements

giac-mysten:
- ensures the publisher does not check resources in the wallet: The storage and registration reuse logic in the client may slow down the publisher significantly when 1000s of registrations are present, because checking them takes multiple round trips to the full node
- introduces a concurrency limit for the publisher: the concurrency limit ensures that there is a maximum (small) amount of blobs stored concurrently by the client. This measure helps with avoiding resource contention and alleviates the locking problem at epoch change.
- introduces a limit to the total number of blob requests enqueued for the publisher: Helps with providing feedback to the clients early that the publisher is under load (and to possibly switch/retry later).

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -387,7 +387,7 @@ impl<T: ContractClient> Client<T> {
             symbol_size=%symbol_size,
             primary_sliver_size=%pair.primary.symbols.len() * usize::from(symbol_size),
             secondary_sliver_size=%pair.secondary.symbols.len() * usize::from(symbol_size),
-            duration = %humantime::Duration::from(encode_duration),
+            duration = ?encode_duration,
             "encoded sliver pairs and metadata"
         );
 
@@ -411,7 +411,7 @@ impl<T: ContractClient> Client<T> {
             .get_blob_status_with_retries(&blob_id, &self.sui_client)
             .await?;
         tracing::info!(
-            duration = %humantime::Duration::from(status_start_timer.elapsed()),
+            duration = ?status_start_timer.elapsed(),
             "retrieved blob status"
         );
 
@@ -422,7 +422,7 @@ impl<T: ContractClient> Client<T> {
             .store_operation_for_blob(metadata, epochs_ahead, persistence, store_when, blob_status)
             .await?;
         tracing::info!(
-            duration = %humantime::Duration::from(store_op_timer.elapsed()),
+            duration = ?store_op_timer.elapsed(),
             "blob resource obtained"
         );
 
@@ -458,7 +458,7 @@ impl<T: ContractClient> Client<T> {
                     let certify_duration = certify_start_timer.elapsed();
                     let blob_size = blob_object.size;
                     tracing::info!(
-                        duration =  %humantime::Duration::from(certify_duration),
+                        duration =  ?certify_duration,
                         blob_size = blob_size,
                         "finished sending blob data and collected certificate"
                     );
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -547,6 +555,19 @@ mod default {
...
+
+    pub(crate) fn max_request_buffer_size() -> usize {
+        // 1x the number of concurrent requests by default means that we start rejecting requests
+        // rather soon to avoid overloading the publisher.
+        //
+        // In total there can be 1x processing and 1x in the buffer, for 2x number of concurrent
+        // requests allowed before we start rejecting.
+        max_concurrent_requests()
+    }
+
```
giac-mysten, line 575: Note: I chose these constants based on the experiments above, and on the fact with these, the last request that comes in before the rate limiting (i.e., the 32nd request) will return successfully after about 1 minute 30 seconds (public testnet, local publisher).

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -355,9 +355,17 @@ pub struct PublisherArgs {
     #[serde(flatten)]
     pub daemon_args: DaemonArgs,
     /// The maximum body size of PUT requests in KiB.
-    #[clap(short, long = "max-body-size", default_value_t = default::max_body_size_kib())]
+    #[clap(long = "max-body-size", default_value_t = default::max_body_size_kib())]
```
mlegner, line 358: Important: We need to be careful with changing our args as users may rely on them. I think this particular case is fine, but we should probably clean up all the arguments at some point and only keep the "short" version on very-frequently used ones.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -355,9 +355,17 @@ pub struct PublisherArgs {
...
     #[serde(default = "default::max_body_size_kib")]
     pub max_body_size_kib: usize,
+    /// The maximum number of requests that can be buffered before the server starts rejecting new
+    /// ones.
+    #[clap(long = "max-buffer-size", default_value_t = default::max_request_buffer_size())]
+    #[serde(default = "default::max_request_buffer_size")]
+    pub max_request_buffer_size: usize,
+    #[clap(long, default_value_t = default::max_concurrent_requests())]
+    #[serde(default = "default::max_concurrent_requests")]
+    pub max_concurrent_requests: usize,
```
mlegner, line 373: Important: Can you add a docstring here?

crates/walrus-service/src/client.rs
```diff
@@ -449,10 +470,15 @@ impl<T: ContractClient> Client<T> {
...
         };
 
+        let sui_cert_timer = Instant::now();
         self.sui_client
             .certify_blob(blob_object.clone(), &certificate)
             .await
             .map_err(|e| ClientError::from(ClientErrorKind::CertificationFailed(e)))?;
+        tracing::info!(
+            duration = %humantime::Duration::from(sui_cert_timer.elapsed()),
+            "certified blob on SUI"
```
```suggestion
            "certified blob on Sui"
```

crates/walrus-service/src/client.rs
```diff
@@ -372,12 +383,12 @@ impl<T: ContractClient> Client<T> {
 
         let pair = pairs.first().expect("the encoding produces sliver pairs");
         let symbol_size = pair.primary.symbols.symbol_size().get();
-        tracing::debug!(
+        tracing::info!(
             symbol_size=%symbol_size,
             primary_sliver_size=%pair.primary.symbols.len() * usize::from(symbol_size),
             secondary_sliver_size=%pair.secondary.symbols.len() * usize::from(symbol_size),
             duration = %humantime::Duration::from(encode_duration),
```
mlegner: Remark: While it was I who decided to use the `humantime` formatter here, I'm no longer sure this is the best choice. Maybe it would be better to just have a decimal formatting of the duration in seconds. But not necessarily in this PR.

giac-mysten: i'll change it.

PR #1125, opened 2024-10-31
fix(node): cancel blob syncs for expired blobs on epoch change

mlegner:
Contributes to #1095

### DIFF
--- a/crates/walrus-service/src/node/blob_sync.rs
+++ b/crates/walrus-service/src/node/blob_sync.rs
@@ -93,7 +93,7 @@ impl BlobSyncHandler {
     pub async fn cancel_sync_and_mark_event_complete(
         &self,
         blob_id: &BlobId,
-    ) -> anyhow::Result<bool> {
+    ) -> anyhow::Result<()> {
         let Some(handle) = self
             .blob_syncs_in_progress
             .lock()
@@ -104,15 +104,14 @@ impl BlobSyncHandler {
                 sync.cancel()
             })
         else {
-            return Ok(false);
+            return Ok(());
         };
 
         if let Some(sync_result) = handle.await?? {
             self.mark_event_completed(sync_result)?;
-            return Ok(true);
         }
 
-        Ok(false)
+        Ok(())
     }
 
     /// Cancels all existing blob syncs for blobs that are already expired in the `current_epoch`


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -80,13 +103,63 @@ impl BlobSyncHandler {
...
+    ) -> anyhow::Result<usize> {
+        tracing::debug!("cancelling all blob syncs for expired blobs");
+
+        let join_handles: Vec<_> = self
+            .blob_syncs_in_progress
+            .lock()
+            .expect("should be able to acquire lock")
+            .iter_mut()
+            .filter_map(|(_, sync)| {
+                (sync.latest_expiration_epoch <= current_epoch)
```
halfprice: Important: should we just use BlobInfo::is_certified() to check whether a blob needs sync or not?

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -225,19 +310,26 @@ impl BlobSyncHandler {
...
 }
 
-type SyncJoinHandle = JoinHandle<Result<Option<(usize, EventID)>, anyhow::Error>>;
+type SyncJoinHandle = JoinHandle<Result<Option<BlobSyncResult>, anyhow::Error>>;
 
 #[derive(Debug)]
 struct InProgressSyncHandle {
     cancel_token: CancellationToken,
     blob_sync_handle: Option<SyncJoinHandle>,
+    latest_expiration_epoch: Epoch,
```
halfprice: Suggestion: given that BlobSyncHandler already has a reference to `StorageNodeInner`, do we need explicitly track the blob expiration here? I wonder we could just expose BlobInfo via StorageNodeInner and use `is_certified` to check if a blob needs sync or not. In that way, blob expiration info only lives in the BlobInfo table, and we don't need to implement another one here.

crates/walrus-service/src/node/blob_sync.rs
mlegner: Yes, I think this should work. 👍

mlegner: Done in 00c251393652e056b67daffd93c43d98514d317b.

crates/walrus-service/src/node/blob_sync.rs
mlegner: We already have a function `is_blob_certified`, which I use in 00c251393652e056b67daffd93c43d98514d317b.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -68,8 +76,24 @@ impl BlobSyncHandler {
...
+    ///
+    /// To avoid interference with later events (e.g., cancelling a sync initiated by a later event
+    /// or immediately restarting the sync that is cancelled here), this function should be called
+    /// before any later events are being handled.
     #[tracing::instrument(skip_all)]
-    pub async fn cancel_sync(&self, blob_id: &BlobId) -> anyhow::Result<Option<(usize, EventID)>> {
+    pub async fn cancel_sync_and_mark_event_complete(
+        &self,
+        blob_id: &BlobId,
+    ) -> anyhow::Result<bool> {
```
halfprice: Question: is the boolean result going to be used somehow? Doesn't look like it is used right now.

crates/walrus-service/src/node.rs
```diff
@@ -822,6 +803,11 @@ impl StorageNode {
             return Ok(());
         }
 
+        // Cancel all blob syncs for blobs that are expired in the *current epoch*.
+        self.blob_sync_handler
+            .cancel_all_expired_syncs_and_mark_events_completed()
```
halfprice, line 808: Question: I see that you moved this call after `begin_committee_change` in the latest commit. Any particular reason for this? I think it should be before `begin_committee_change`, so that blob sync cancellation is done despite of epoch change status.

mlegner: I agree with your sentiment; however, this is probably not too relevant in practice as nodes that are far behind will now use your recovery process.

If it were before the `begin_committee_change`, the `current_epoch` would not yet be updated, meaning the cancellation would not work. If you feel strongly about this, I can add another function on `StorageNodeInner` that takes an epoch as an input instead of using the `current_epoch` to decide if a blob is expired. WDYT?

crates/walrus-service/src/node/blob_sync.rs
mlegner: No, I can remove it.

mlegner: Done in 91393510e4f9af2193453b038ed8c21c0f08cee3.

crates/walrus-service/src/node.rs
halfprice: Good point that `current_epoch` is not updated. I think we can merge this first. Yeah, I think for the recovering nodes, since we skip syncing, too many pending blob syncs may not be an issue. We can monitor this and see if what you suggested (taking epoch as input in `is_blob_certified`) is needed.

mlegner: (BTW: I also only discovered this myself through the test case you set up. 😁)

Issue #1088, opened 2024-10-21
Report additional metrics in storage nodes


mlegner:
- [ ] Information from `health` endpoint (including shards and their status)?
- [x] Number of shards in the current epoch -> #1127
- [ ] Number of shards in the next epoch (if already determined).
- [ ] Node ID?
- [ ] Required storage capacity in the current and next epoch.

PR #1127, opened 2024-10-31
feat(metrics): export number of currently owned shards on node

mlegner:
Contributes to #1088.

### DIFF
--- a/crates/walrus-service/src/node/committee/committee_service.rs
+++ b/crates/walrus-service/src/node/committee/committee_service.rs
@@ -448,13 +448,10 @@ where
         metrics.current_epoch_state.set_from_committees(committees);
 
         if let Some(local_identity) = self.local_identity.as_ref() {
-            metrics.current_shards_owned.set(
+            metrics.shards_owned.set(
                 committees
                     .current_committee()
-                    .shards_for_node_public_key(local_identity)
-                    .len()
-                    .try_into()
-                    .expect("number of shards must fit into a u64"),
+                    .n_shards_for_node_public_key(local_identity),
             );
         }
     }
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/common/telemetry.rs
```diff
@@ -416,21 +416,11 @@ pub(crate) struct CurrentEpochMetric(GenericGauge<AtomicU64>);
 
 impl CurrentEpochMetric {
     pub fn new() -> Self {
-        let opts = Opts::new("current_epoch", "The current Walrus epoch").namespace("walrus");
-        let metric = create_metric!(GenericGauge<AtomicU64>, opts, []);
-        Self(metric)
-    }
-
-    /// Registers the metric with the registry, and returns a new instance of it.
-    pub fn register(registry: &Registry) -> Self {
```
giac-mysten, line 425: question: how come this is not needed anymore?

crates/walrus-service/src/common/telemetry.rs
```diff
@@ -330,7 +330,7 @@ macro_rules! define_metric_set {
     (
         $(#[$outer:meta])*
         struct $name:ident {
-            $($new_type_field:ident: $new_type_field_type:ident),*
```
giac-mysten, line 333: question: this was a bug, correct? the previous version allowed an arbitrary number of commas at the end?

crates/walrus-service/src/common/telemetry.rs
mlegner: I don't know, this isn't called anywhere and the `CurrentEpochStateMetric` also doesn't have it. I expect that the macros are now set up in a way that they automatically call the `register` function, but I'm not sure...

crates/walrus-service/src/common/telemetry.rs
mlegner: Not really a bug. The previous version didn't allow *any* comma at the end. The new version *requires* a comma at the end, which makes more sense when having the two types of fields. The combination would be possible with a `$(,)?`, but this doesn't seem to work as we have it already below and having it twice confuses the parser somehow.

PR #1128, opened 2024-10-31
feat: Retry downloading checkpoints based on exponential backoff

sadhansood:
This PR updates the checkpoint download retry to start using exponential backoff and fixes an issue with pool monitor in checkpoint downloader.

### DIFF
--- a/crates/walrus-utils/src/checkpoint_downloader.rs
+++ b/crates/walrus-utils/src/checkpoint_downloader.rs
@@ -12,6 +12,7 @@ use std::{
 
 use anyhow::{anyhow, Result};
 use prometheus::{register_int_gauge_with_registry, IntGauge, Registry};
+use rand::{rngs::StdRng, SeedableRng};
 use serde::{Deserialize, Serialize};
 use sui_rest_api::{client::sdk, Client};
 use sui_types::{
@@ -30,8 +31,11 @@ use typed_store::{rocks::DBMap, Map};
 pub struct ParallelDownloaderConfig {
     /// Number of retries per checkpoint before giving up.
     pub min_retries: u32,
-    /// Delay between retries.
-    pub retry_delay: Duration,
+    /// Initial delay before the first retry.
+    pub initial_delay: Duration,
+    /// Maximum delay between retries. Once this delay is reached, the
+    /// fetcher will keep retrying with this duration.
+    pub max_delay: Duration,
 }
 
 impl Default for ParallelDownloaderConfig {
@@ -60,7 +64,8 @@ impl Default for ParallelDownloaderConfig {
         // TODO: Use adaptive retry delay based on RTT and error type (issue-1123)
         Self {
             min_retries: 10,
-            retry_delay: Duration::from_millis(150),
+            initial_delay: Duration::from_millis(150),
+            max_delay: Duration::from_secs(2),
         }
     }
 }
@@ -338,7 +343,9 @@ impl ParallelCheckpointDownloaderInner {
         });
     }
 
-    /// Returns the current checkpoint lag between the local store and the full node.
+    /// Returns the current checkpoint lag between the local store and the full node
+    /// in terms of sequence numbers. This works by downloading the latest checkpoint
+    /// summary from the full node and comparing it with the current checkpoint in the store.
     async fn current_checkpoint_lag(
         checkpoint_store: &DBMap<(), TrustedCheckpoint>,
         client: &Client,
@@ -361,8 +368,10 @@ impl ParallelCheckpointDownloaderInner {
         config: ParallelDownloaderConfig,
     ) -> Result<()> {
         tracing::info!("Starting checkpoint download worker {}", worker_id);
+        let mut rng = StdRng::from_entropy();
         while let Ok(WorkerMessage::Download(sequence_number)) = message_receiver.recv().await {
-            let entry = Self::download_with_retry(&client, sequence_number, &config).await;
+            let entry =
+                Self::download_with_retry(&client, sequence_number, &config, &mut rng).await;
             checkpoint_sender.send(entry).await?;
         }
         tracing::info!("Checkpoint download worker {} shutting down", worker_id);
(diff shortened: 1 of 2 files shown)


crates/walrus-utils/src/checkpoint_downloader.rs
```diff
@@ -392,8 +400,12 @@ impl ParallelCheckpointDownloaderInner {
                         continue;
                     }
 
-                    let lag = Self::current_checkpoint_lag(
-                        &config.checkpoint_store, &config.client).await?;
+                    let Ok(lag) = Self::current_checkpoint_lag(
+                        &config.checkpoint_store, &config.client).await else {
+                        tracing::error!("Failed to fetch current checkpoint lag");
```
mlegner: Suggestion: Can we make this a warning as this is expected to happen whenever the full node is not accessible?

crates/walrus-utils/src/checkpoint_downloader.rs
```diff
@@ -527,6 +553,21 @@ impl ParallelCheckpointDownloaderInner {
     }
 }
 
+/// Helper function to create backoff with consistent settings
+#[cfg(not(test))]
```
mlegner, line 555: Question: Why is the `cfg(not(test))` necessary?

crates/walrus-utils/src/checkpoint_downloader.rs
```diff
@@ -32,6 +33,8 @@ pub struct ParallelDownloaderConfig {
     pub min_retries: u32,
     /// Delay between retries.
     pub retry_delay: Duration,
+    /// Max timeout
+    pub timeout: Duration,
```
mlegner: AFAIU, these are now parameters for the exponential backoff. So the `retry_delay` is the initial delay, which is then increased up to the `timeout`. At that point, it just keeps creating the same delay infinitely.

If this understanding is correct, then these fields should be renamed and their docstrings adjusted.

crates/walrus-utils/src/checkpoint_downloader.rs
```diff
@@ -495,21 +508,34 @@ impl ParallelCheckpointDownloaderInner {
...
+        loop {
+            let result = client.get_full_checkpoint(sequence_number).await;
+            let Ok(checkpoint) = result else {
+                let err = result.err();
+                handle_checkpoint_error(err.as_ref(), sequence_number);
+
+                let Some(delay) = backoff.next_delay() else {
+                    backoff = create_backoff(rng, config);
+                    continue;
+                };
```
mlegner: Why would this happen? If no `max_retries` is defined, the backoff should create an infinite sequence of delays.

sadhansood: You are right, I absolutely did not see the implementation of this function and thought it somehow works like an iterator and terminates after max timeout with a `None`.

crates/walrus-utils/src/checkpoint_downloader.rs
sadhansood: done

crates/walrus-utils/src/checkpoint_downloader.rs
sadhansood: If we don't do this, `clippy test` does not see this function being used and returns with an error.

crates/walrus-utils/src/checkpoint_downloader.rs
sadhansood: done

crates/walrus-utils/src/backoff.rs
```diff
@@ -117,7 +117,7 @@ impl<R: Rng> ExponentialBackoff<R> {
         }
     }
 
-    fn next_delay(&mut self) -> Option<Duration> {
+    pub fn next_delay(&mut self) -> Option<Duration> {
```
mlegner: Question: Is this necessary? There is the `BackoffStrategy::next_delay` defined below.

crates/walrus-utils/src/checkpoint_downloader.rs
mlegner: Ah, because the `download_with_retry` function that calls this is also annotated with `cfg(not(test))`.

crates/walrus-utils/src/checkpoint_downloader.rs
```diff
@@ -30,8 +31,11 @@ use typed_store::{rocks::DBMap, Map};
 pub struct ParallelDownloaderConfig {
     /// Number of retries per checkpoint before giving up.
     pub min_retries: u32,
-    /// Delay between retries.
-    pub retry_delay: Duration,
+    /// Initial delay before the first retry.
+    pub initial_delay: Duration,
+    /// Maximum time to wait for a checkpoint to be fetched. Once this delay is reached, the
+    /// fetcher will keep retrying with this timeout.
+    pub timeout: Duration,
```
mlegner: I still feel that the name `timeout` and the docstring don't properly describe the behavior. This is not really a timeout, but a time between retries.

halfprice: Rather, I think this is currently used as the max delay in backoff.

crates/walrus-utils/src/checkpoint_downloader.rs
```diff
@@ -495,21 +509,31 @@ impl ParallelCheckpointDownloaderInner {
...
-            tokio::time::sleep(config.retry_delay).await;
-            checkpoint = client.get_full_checkpoint(sequence_number).await;
+        let mut backoff = create_backoff(rng, config);
+        loop {
+            let result = client.get_full_checkpoint(sequence_number).await;
+            let Ok(checkpoint) = result else {
+                let err = result.err();
+                handle_checkpoint_error(err.as_ref(), sequence_number);
+
+                let delay = backoff.next_delay().unwrap_or(Duration::from_secs(1));
```
halfprice: Remark: should we call `.expect()` instead of setting a default value? It seems that the backoff is not set with a max retries.

crates/walrus-utils/src/backoff.rs
sadhansood: fixed

crates/walrus-utils/src/checkpoint_downloader.rs
sadhansood: fixed

PR #1131, opened 2024-11-01
feat: storage node complete recovery

halfprice:
This PR implements a mechanism which brings a node that is lagging for arbitrarily long to the last epoch.

The main approach is to only process BlobInfo update without syncing blobs, and after reaching the latest epoch, the node scans the BlobInfo table and finds all the certified blobs to recover/sync.

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -1148,32 +1148,18 @@ where
     F: Fn() -> Fut,
     Fut: Future<Output = ClientResult<R>>,
 {
-    let mut retry_count = 0;
-    loop {
-        let current_epoch = client.committees.read().await.epoch();
-        let result = func().await;
-        retry_count += 1;
-        match result {
-            Err(error) if error.may_be_caused_by_epoch_change() => {
-                if let Some(max_retry) = client
-                    .config
-                    .communication_config
-                    .request_rate_config
-                    .max_retries
-                {
-                    if retry_count > max_retry {
-                        tracing::warn!("request max out retries");
-                        return Err(error);
-                    }
-                }
-                tracing::warn!(
-                    %error,
-                    "an error occurred during the current operation, \
-                    which may be caused by epoch change; refreshing the committees and retrying"
-                );
-                client.refresh_committees(current_epoch).await?;
-            }
-            result => return result,
+    let current_epoch = client.committees.read().await.epoch();
+    let result = func().await;
+    match result {
+        Err(error) if error.may_be_caused_by_epoch_change() => {
+            tracing::warn!(
+                %error,
+                "an error occurred during the current operation, \
+                which may be caused by epoch change; refreshing the committees and retrying"
+            );
+            client.refresh_committees(current_epoch).await?;
+            func().await
         }
+        result => result,
     }
 }
--- a/crates/walrus-service/src/node/committee/committee_service.rs
+++ b/crates/walrus-service/src/node/committee/committee_service.rs
@@ -357,6 +357,25 @@ where
 
         Ok(())
     }
+
+    async fn extend_services_from_committee(
+        &self,
+        committee: &Arc<Committee>,
+        service_factory: &mut Box<dyn NodeServiceFactory<Service = T>>,
+    ) -> Result<(), BeginCommitteeChangeError> {
+        let new_services =
+            create_services_from_committee(service_factory, committee, &self.inner.encoding_config)
+                .await
+                .map_err(BeginCommitteeChangeError::AllServicesFailed)?;
+
+        let mut services = self
+            .inner
+            .services
+            .lock()
+            .expect("thread did not panic with mutex");
+        services.extend(new_services);
+        Ok(())
+    }
 }
 
 pub(super) struct NodeCommitteeServiceInner<T> {
@@ -616,43 +635,22 @@ where
 
         let mut service_factory = self.inner.service_factory.lock().await;
 
+        self.inner
+            .services
+            .lock()
+            .expect("thread did not panic with mutex")
+            .clear();
+
         if let Some(previous_committee) = latest.previous_committee() {
-            let new_services = create_services_from_committee(
-                &mut service_factory,
-                previous_committee,
-                &self.inner.encoding_config,
-            )
-            .await
-            .map_err(BeginCommitteeChangeError::AllServicesFailed)?;
-
-            let mut services = self
-                .inner
-                .services
-                .lock()
-                .expect("thread did not panic with mutex");
-            services.extend(new_services);
+            self.extend_services_from_committee(previous_committee, &mut service_factory)
+                .await?;
         }
 
-        {
-            let new_services = create_services_from_committee(
-                &mut service_factory,
-                latest.current_committee(),
-                &self.inner.encoding_config,
-            )
-            .await
-            .map_err(BeginCommitteeChangeError::AllServicesFailed)?;
+        self.extend_services_from_committee(latest.current_committee(), &mut service_factory)
+            .await?;
 
-            let mut services = self
-                .inner
-                .services
-                .lock()
-                .expect("thread did not panic with mutex");
-            services.extend(new_services);
-        }
-
-        self.inner.committee_tracker.send_if_modified(|tracker| {
+        self.inner.committee_tracker.send_modify(|tracker| {
             tracker.update_active_committees(latest);
-            true
         });
 
         Ok(())
@@ -682,6 +680,7 @@ impl<T> std::fmt::Debug for NodeCommitteeService<T> {
     }
 }
 
+/// Create services for each member of the committee.
 async fn create_services_from_committee<T: NodeService>(
     service_factory: &mut Box<dyn NodeServiceFactory<Service = T>>,
     committee: &Committee,
@@ -692,6 +691,7 @@ async fn create_services_from_committee<T: NodeService>(
     Ok(services)
 }
 
+/// Add services for each member of the committee.
 #[tracing::instrument(skip_all, fields(walrus.epoch = committee.epoch))]
 async fn add_members_from_committee<T: NodeService>(
     services: &mut HashMap<PublicKey, T>,
--- a/crates/walrus-simtest/tests/simtest.rs
+++ b/crates/walrus-simtest/tests/simtest.rs
@@ -247,7 +247,7 @@ mod tests {
         // Starts a background workload that a client keeps writing and retrieving data.
         // All requests should succeed even if a node crashes.
         let workload_handle = tokio::spawn(async move {
-            let mut data_length = 31415;
+            let mut data_length = 64;
             loop {
                 tracing::info!("writing data with size {}", data_length);
 
@@ -258,7 +258,7 @@ mod tests {
 
                 tracing::info!("finish writing data with size {}", data_length);
 
-                data_length += 1024;
+                data_length += 64;
             }
         });
 


crates/walrus-sdk/src/client.rs
```diff
@@ -172,7 +172,9 @@ impl UrlEndpoints {
 
     fn server_health_info(&self) -> (Url, &'static str) {
         (
-            self.0.join("/v1/health").expect("this is a valid URL"),
+            self.0
+                .join("/v1/health?detailed=true")
+                .expect("this is a valid URL"),
             HEALTH_URL_TEMPLATE,
         )
```
mlegner: Suggestion: Add `detailed: bool` as a function argument.

crates/walrus-service/src/node/committee.rs
```diff
@@ -171,6 +171,8 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
     /// Otherwise, an error is returned.
     fn end_committee_change(&self, epoch: Epoch) -> Result<(), EndCommitteeChangeError>;
 
+    async fn update_to_latest_committee(&self) -> Result<(), BeginCommitteeChangeError>;
```
mlegner: It's a bit unclear how this is related to the `begin_committee_change`. If it's something different, it might also warrant its own error type.

crates/walrus-service/src/node.rs
```diff
@@ -568,13 +578,16 @@ impl StorageNode {
                     // Clear the starting epoch, so that we never make this check again.
                     maybe_epoch_at_start = None;
 
-                    // if event.event_epoch() < starting_epoch - 1
-                    if event.event_epoch() + 1 < epoch_at_start {
+                    if node_status == NodeStatus::Active && event.event_epoch() + 1 < epoch_at_start
+                    {
                         tracing::warn!(
-                            "the current epoch ({}) is too far ahead of the event epoch: {}",
+                            "the current epoch ({}) is too far ahead of the event epoch: {}; node entering recovery mode",
```
mlegner: Nit: Given that we have the recovery mode now, we could remove the "too" 😁.
```suggestion
                            "the current epoch ({}) is far ahead of the event epoch: {}; node entering recovery mode",
```

crates/walrus-service/src/node.rs
```diff
@@ -813,6 +837,53 @@ impl StorageNode {
...
+                tracing::info!(
+                    "processing event reaches the latest epoch, start recover entire node"
+                );
+
+                self.inner
+                    .storage
+                    .set_node_status(NodeStatus::RecoveryInProgress)?;
+
+                // Cancel any shard sync;
+                self.shard_sync_handler.cancel_all().await;
```
mlegner: Question: Why don't we want to do shard syncs? The node may have gained a shard right in this epoch, in which case it would be great to sync that shard instead of recovering it.

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,140 @@
...
+            task_handle: Arc::new(Mutex::new(None)),
+        }
+    }
+
+    pub fn start_node_recovery(&self, epoch: Epoch) -> Result<(), TypedStoreError> {
+        let start = tokio::time::Instant::now();
+        let mut locked_task_handle = self.task_handle.lock().unwrap();
+        assert!(locked_task_handle.is_none());
+
+        let node: Arc<StorageNodeInner> = self.node.clone();
```
mlegner: Nit: Is the type here necessary?
```suggestion
        let node = self.node.clone();
```

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,140 @@
...
+                    match blob_result {
+                        Ok((blob_id, blob_info)) => {
+                            if !blob_info.is_certified(epoch) {
+                                // Skip blobs that are not certified in the given epoch. This includes blobs that are
+                                // invalid or expired.
+                                tracing::info!(
+                                    blob_id = ?blob_id,
+                                    epoch = epoch,
+                                    "skip non-certified blob"
+                                );
```
mlegner: Nit: This should probably be `debug!` as there may be many.

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,140 @@
...
+                                continue;
+                            }
+
+                            if let Ok(stored_at_all_shards) =
+                                node.storage.is_stored_at_all_shards(&blob_id)
+                            {
+                                if stored_at_all_shards {
```
mlegner: Nit: This version is a bit shorter:
```suggestion
                            if let Ok(true) = node.storage.is_stored_at_all_shards(&blob_id) {
```

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,140 @@
...
+                let mut contain_recovery_blob = false;
+                for blob_result in node.storage.certified_blob_info_iter_before_epoch(epoch) {
+                    match blob_result {
+                        Ok((blob_id, blob_info)) => {
+                            if !blob_info.is_certified(epoch) {
+                                // Skip blobs that are not certified in the given epoch. This includes blobs that are
+                                // invalid or expired.
+                                tracing::info!(
+                                    blob_id = ?blob_id,
+                                    epoch = epoch,
```
mlegner: Suggestion: Use the `Display` formatter here. Also, the `=` is not necessary. Both also apply to other places.
```suggestion
                                    %blob_id,
                                    epoch,
```

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,140 @@
...
+                        Err(err) => {
+                            tracing::error!(error = ?err, "failed to read certified blob");
+                        }
+                    }
+                }
+
+                if !contain_recovery_blob {
+                    tracing::info!("no recovery blob found; stop recovery task");
+                    break;
+                }
```
mlegner: Question: Do we actually need this boolean flag when we have the `all_blob_syncs` vector?

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,140 @@
...
+                            match start_sync_result {
+                                Ok(notify) => {
+                                    all_blob_syncs.push(notify);
+                                }
+                                Err(err) => {
+                                    tracing::error!(
+                                        blob_id = ?blob_id,
+                                        error = ?err,
+                                        "failed to start recovery sync for blob",
+                                    );
```
mlegner: Question: What do we do with the blobs we fail to recover?

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -44,6 +48,12 @@ struct Permits {
     sliver: Arc<Semaphore>,
 }
 
+#[derive(Debug, Copy, Clone)]
+pub struct EventInfo {
+    pub event_id: EventID,
+    pub event_index: usize,
+}
```
mlegner: Just FYI, this has some overlap with my PR #1125. 😁

Should be simple to resolve though.

crates/walrus-service/src/node/blob_sync.rs
mlegner, line 1: Question: With these changes to the blob sync, could we also use it for the recovery if shard sync fails? This would then probably solve #1095 automatically. There would still remain the inefficiency of checking all shards instead of focusing on the one(s) that need to be recovered.

crates/walrus-service/src/node/storage.rs
```diff
@@ -52,13 +55,39 @@ pub(crate) use shard::{ShardStatus, ShardStorage};
 #[derive(Debug, Clone, Copy)]
 pub struct WouldBlockError;
 
+// TODO: add status to health endpoint
```
mlegner: Suggestion: Can we then also add a metric for it? (I'm currently adding some metrics in #1127.)

crates/walrus-sdk/src/client.rs
halfprice: Done

crates/walrus-service/src/node.rs
halfprice: I removed this: if there is any existing shard sync on going, we won't cancel it to save some work.

The main reason I wanted to cancel the existing shard sync is for simplicity. Theoretically, if the node gained the shard 2 epoch ago, the original source may not hold the shard anymore, and stop the sync immediately.

crates/walrus-service/src/node/node_recovery.rs
halfprice: Good point. Removed

crates/walrus-service/src/node/storage.rs
halfprice: Done.

crates/walrus-service/src/node/blob_sync.rs
halfprice: Yes, but probably in the next PR. I agree that one nice thing about reuse blob sync handler is that cancellation is handled.

crates/walrus-service/src/node/node_recovery.rs
halfprice: It's actually impossible in this code path. There only place where start_sync can return error is when marking event complete. In this codepath, there is no event to mark, and therefore cannot fail. I added more comments here.

crates/walrus-service/src/node/committee.rs
halfprice: The reason I reused the error type is mainly due to the duplicated code path between `update_to_latest_committee` and `begin_committee_change` where they fetch the new committee on chain. I wonder if it make sense to change `update_to_latest_committee` to `begin_committee_change_to_latest_committee` so that reusing the error makes more sense?

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -611,6 +604,57 @@ where
     fn end_committee_change(&self, epoch: Epoch) -> Result<(), EndCommitteeChangeError> {
         self.end_committee_change_to(epoch)
     }
+
+    async fn update_to_latest_committee(&self) -> Result<(), BeginCommitteeChangeError> {
```
halfprice: Hi @giac-mysten , can you please take a look at this function and see if it makes sense? I didn't touched the committee updating code path before, so I don't know if I may miss something. The idea in this function is to update the committee in the node to match the latest committee on chain. Different from `begin_committee_change` where it has to move from committee in epoch X to epoch X+1, this function has no restriction on the new committee's epoch: we just want to get the latest on chain.

Thanks!

crates/walrus-sdk/src/client.rs
```diff
@@ -170,11 +170,20 @@ impl UrlEndpoints {
...
+                    .join("/v1/health?detailed=true")
+                    .expect("this is a valid URL"),
+                HEALTH_URL_TEMPLATE,
+            )
+        } else {
+            (
+                self.0.join("/v1/health").expect("this is a valid URL"),
+                HEALTH_URL_TEMPLATE,
+            )
+        }
```
mlegner: Nit:
```suggestion
        let mut url = self.0.join("/v1/health").expect("this is a valid URL");
        url.set_query(detailed.then_some("detailed=true"));
        (url, HEALTH_URL_TEMPLATE)
```

crates/walrus-service/src/node.rs
```diff
@@ -568,13 +578,24 @@ impl StorageNode {
...
                             epoch_at_start,
                             event.event_epoch()
                         );
+                        self.inner
+                            .metrics
+                            .current_node_status
+                            .set(NodeStatus::RecoveryCatchUp.to_i64());
+                        self.inner
+                            .storage
+                            .set_node_status(NodeStatus::RecoveryCatchUp)?;
```
mlegner: Nit: Can we create a function `StorageNodeInner::set_node_status` that combines these two? That way we won't forget to update the metrics.

crates/walrus-sdk/src/client.rs
```diff
@@ -956,7 +968,7 @@ mod tests {
     #[test]
     fn test_url_health_info_endpoint() {
         let endpoints = UrlEndpoints(Url::parse("https://node.com").unwrap());
-        let (url, _) = endpoints.server_health_info();
+        let (url, _) = endpoints.server_health_info(false);
 
         assert_eq!(url.to_string(), "https://node.com/v1/health");
     }
```
mlegner: Here we could add a test for the detailed version:
```rust


    param_test! {
        test_url_health_info_endpoint: [
            default: (false, "https://node.com/v1/health"),
            detailed: (true, "https://node.com/v1/health?detailed=true"),
        ]
    }
    fn test_url_health_info_endpoint(detailed: bool, expected_url: &str) {
        let endpoints = UrlEndpoints(Url::parse("https://node.com").unwrap());
        let (url, _) = endpoints.server_health_info(detailed);

        assert_eq!(url.to_string(), expected_url);
    }
```

crates/walrus-service/src/node.rs
```diff
@@ -819,6 +877,62 @@ impl StorageNode {
...
+            .metrics
+            .current_node_status
+            .set(NodeStatus::RecoveryInProgress(event.epoch).to_i64());
+        self.inner
+            .storage
+            .set_node_status(NodeStatus::RecoveryInProgress(event.epoch))?;
+
+        // // Cancel any shard syncs that are currently in progress. Since the node is lagging,
+        // self.shard_sync_handler.cancel_all().await;
+
```
```suggestion
```

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -170,65 +173,81 @@ impl BlobSyncHandler {
...
                     "otel.kind" = "CONSUMER",
                     "otel.status_code" = field::Empty,
                     "otel.status_message" = field::Empty,
-                    "walrus.event.index" = event_index,
-                    "walrus.event.tx_digest" = ?event.event_id.tx_digest,
-                    "walrus.event.event_seq" = ?event.event_id.event_seq,
-                    "walrus.event.kind" = event.label(),
-                    "walrus.blob_id" = %event.blob_id,
+                    "walrus.event.index" = event_info.map(|info| info.event_index),
+                    "walrus.event.tx_digest" = ?event_info.map(|info| info.event_id),
```
mlegner:
```suggestion
                    "walrus.event.tx_digest" = ?event_info.map(|info| info.event_id.tx_digest),
```

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -275,8 +300,8 @@ impl BlobSyncHandler {
         self.remove_sync_handle(&synchronizer.blob_id).await;
 
         let label = match output {
-            Ok(Some(_)) => metrics::STATUS_SUCCESS,
-            Ok(None) => metrics::STATUS_CANCELLED,
+            Ok(None) => metrics::STATUS_SUCCESS,
```
mlegner: This could also happen when the sync is cancelled and `synchronizer.event_info.is_none()`, right?

crates/walrus-service/src/node/committee.rs
mlegner: That sounds good, yes.

crates/walrus-service/src/node/metrics.rs
```diff
@@ -90,7 +91,10 @@ telemetry::define_metric_set! {
             // Buckets from 2^18 (256 KiB) to 2^34 (16 GiB, inclusive), which covers the ~212 KiB to
             // 14 GiB unencoded blobs which are possible in a system with 1000 shards.
             buckets: (18..=34).map(|power| (1u64 << power) as f64).collect::<Vec<_>>()
-        }
+        },
+
+        #[help = "Indicates the current node status"]
+        current_node_status: IntGauge[],
```
mlegner, line 97: Remark: We use a different way of encoding the epoch state, see [here](https://github.com/MystenLabs/walrus/blob/9a31a5e3ab04b99513d8f1c2d5d1f54ef04bd7ae/crates/walrus-service/src/common/telemetry.rs#L448). I think JP tried to mimic the systemd metrics, see https://github.com/MystenLabs/walrus/pull/983. However, your version seems simpler and more efficient to me. 🤷

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -215,6 +215,14 @@ impl ShardSyncHandler {
...
     }
 
+    // /// Cancels all the shard sync tasks.
+    // pub async fn cancel_all(&self) {
+    //     let mut shard_sync_in_progress = self.shard_sync_in_progress.lock().await;
+    //     for (_, task) in shard_sync_in_progress.drain() {
+    //         task.abort();
+    //     }
+    // }
+
```
mlegner: Can this be removed?
```suggestion
```

crates/walrus-service/src/test_utils.rs
```diff
@@ -250,6 +251,8 @@ pub struct SimStorageNodeHandle {
     pub cancel_token: CancellationToken,
     /// The wrapped simulator node id.
     pub node_id: sui_simulator::task::NodeId,
+    /// The storage capability object of the node.
+    pub storage_capability: Option<StorageNodeCap>,
```
mlegner: Nit: Can we call this `storage_node_capability`?

crates/walrus-service/src/test_utils.rs
```diff
@@ -1649,6 +1680,9 @@ pub mod test_cluster {
         node::{committee::DefaultNodeServiceFactory, contract_service::SuiSystemContractService},
     };
 
+    /// The weight of each storage node in the test cluster.
+    pub const FROSTER_PER_NODE_WEIGHT: u64 = 1_000_000;
```
mlegner: Nit: Should this be `FROST`?
```suggestion
    pub const FROST_PER_NODE_WEIGHT: u64 = 1_000_000;
```

crates/walrus-service/src/node/storage.rs
```diff
@@ -52,13 +55,48 @@ pub(crate) use shard::{ShardStatus, ShardStorage};
...
+}
+
+impl Display for NodeStatus {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let display_string = match self {
+            NodeStatus::Active => "Active".to_string(),
+            NodeStatus::RecoveryCatchUp => "RecoveryCatchUp".to_string(),
+            NodeStatus::RecoveryInProgress(epoch) => format!("RecoveryInProgress ({epoch})"),
+        };
+        write!(f, "{}", display_string)
```
mlegner: Nit: This version seems simpler and clearer to me:
```suggestion
        match self {
            NodeStatus::Active => write!(f, "Active"),
            NodeStatus::RecoveryCatchUp => write!(f, "RecoveryCatchUp"),
            NodeStatus::RecoveryInProgress(epoch) => write!(f, "RecoveryInProgress ({epoch})"),
        }
```

crates/walrus-service/src/node/storage.rs
mlegner: Nit:
```suggestion
```

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,155 @@
...
+            if recovering_epoch == self.node.current_epoch() {
+                return self.start_node_recovery(self.node.current_epoch());
+            } else {
+                assert!(recovering_epoch < self.node.current_epoch());
+                tracing::warn!(
+                    recovering_epoch = recovering_epoch,
+                    current_epoch = self.node.current_epoch(),
+                    "recovery epoch mismatch; skip recovery restart; next epoch change start event
+                    will bring node to the latest state"
+                );
```
mlegner, line 138: Important: Don't we have to set the status to `RecoveryCatchUp` for this to be true?

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,155 @@
...
+            }
+        });
+        *locked_task_handle = Some(task_handle);
+
+        Ok(())
+    }
+
+    /// Restarts any in progress recovery.
+    pub fn restart_recovery(&self) -> Result<(), TypedStoreError> {
+        if let crate::node::NodeStatus::RecoveryInProgress(recovering_epoch) =
```
mlegner: Nit:
```suggestion
        if let NodeStatus::RecoveryInProgress(recovering_epoch) =
```

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -0,0 +1,155 @@
...
+                                        blob_id, err,
+                                    );
+                                }
+                            }
+                        }
+                        Err(err) => {
+                            tracing::error!(error = ?err, "failed to read certified blob");
+                        }
+                    }
+                }
```
mlegner: How about this version that reduces the indentation by two levels?
```suggestion
                for (blob_id, blob_info) in node
                    .storage
                    .certified_blob_info_iter_before_epoch(epoch)
                    .filter_map(|blob_result| {
                        blob_result.inspect_err(|error| {
                            tracing::error!(?error, "failed to read certified blob")
                        })
                        .ok()
                    })
                {
                    if !blob_info.is_certified(epoch) {
                        // Skip blobs that are not certified in the given epoch. This
                        // includes blobs that are invalid or expired.
                        tracing::debug!(%blob_id, %epoch, "skip non-certified blob");
                        continue;
                    }

                    if let Ok(stored_at_all_shards) = node.storage.is_stored_at_all_shards(&blob_id)
                    {
                        if stored_at_all_shards {
                            tracing::debug!(
                                %blob_id,%epoch,"blob is stored at all shards; skip recovery"
                            );
                            continue;
                        }
                    } else {
                        tracing::warn!(
                            %blob_id,
                            "failed to check if blob is stored at all shards; start blob sync"
                        );
                    }

                    tracing::debug!(%blob_id, "start recovery sync for blob");

                    // TODO: rate limit start sync to avoid OOM.
                    let start_sync_result = blob_sync_handler
                        .start_sync(
                            blob_id,
                            blob_info.initial_certified_epoch().expect(
                                "certified blob should have an initial certified epoch set",
                            ),
                            None,
                            start,
                        )
                        .await;
                    match start_sync_result {
                        Ok(notify) => {
                            all_blob_syncs.push(notify);
                        }
                        Err(err) => {
                            // The only place where start_sync can fail is when marking the
                            // event complete, which is not applicable here since the there
                            // is no event associated with the recovery task.
                            panic!(
                                "failed to start recovery sync for blob {}: {}",
                                blob_id, err,
                            );
                        }
                    }
```

crates/walrus-service/src/node/committee.rs
halfprice: Done

crates/walrus-sdk/src/client.rs
halfprice: Good point! I didn't read where the `server_health_info` is used carefully...

crates/walrus-service/src/node/blob_sync.rs
halfprice: Amazing catch! Add a field to be explicit about whether the sync is cancelled.

crates/walrus-service/src/node/metrics.rs
halfprice: Yeah, I'm aware of that and I suspect using field to indicate status also hard to observe in dashboard, since there will me multiple lines jumping between 0 and 1. So I guess we keep it as is? Let me know what do you prefer.

crates/walrus-service/src/test_utils.rs
halfprice: Sure. Done

crates/walrus-service/src/node/node_recovery.rs
halfprice: And of course you are right! I thought about this issue pretty late last night, and only added the branch without thinking through it. I think what needed is a test to do repeated crash-recovery. I'll create one tomorrow.

crates/walrus-service/src/node.rs

crates/walrus-service/src/node/metrics.rs
mlegner: Yes, let's keep it as it is. 👍

crates/walrus-service/src/client.rs
```diff
@@ -1148,18 +1148,32 @@ where
...
+                tracing::warn!(
+                    %error,
+                    "an error occurred during the current operation, \
+                    which may be caused by epoch change; refreshing the committees and retrying"
+                );
+                client.refresh_committees(current_epoch).await?;
+            }
+            result => return result,
         }
-        result => result,
```
mlegner: Question: What motivated these changes? AFAIU, the `request_rate_config.max_retries` is intended for individual requests to the storage nodes, not a complete operation. Also, if the error occurs due to an epoch change, retrying once should resolve it.

So my suggestion would be to revert these changes. cc @giac-mysten

giac-mysten: yes, this retry mechanism is in place only to refresh the committee if the client fails because of epoch change. 

The only reason why it may make sense to try multiple times is if the client tries to refresh, but fails for so long that another epoch change occurs -- i don't think we should consider this case though 😅

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -611,6 +604,59 @@ where
...
+            )
+            .await
+            .map_err(BeginCommitteeChangeError::AllServicesFailed)?;
+
+            let mut services = self
+                .inner
+                .services
+                .lock()
+                .expect("thread did not panic with mutex");
+            services.extend(new_services);
```
giac-mysten: Suggestion: this code is duplicated twice. separate into a function `self.extend_services_from_committee` or similar?

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -392,14 +392,7 @@ where
...
-            committees.current_committee(),
-            &encoding_config,
-        )
-        .await?;
-        add_members_from_committee(
-            &mut services,
+        let services = create_services_from_committee(
             &mut service_factory,
             committees.current_committee(),
             &encoding_config,
```
giac-mysten, line 417: Question: was this a bug where the same services were added twice?

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -611,6 +604,59 @@ where
...
+                .services
+                .lock()
+                .expect("thread did not panic with mutex");
+            services.extend(new_services);
+        }
+
+        self.inner.committee_tracker.send_if_modified(|tracker| {
+            tracker.update_active_committees(latest);
+            true
+        });
```
giac-mysten: Question: are we absolutely sure that the committe has been modified?

Suggestion: if we are sure, then
```suggestion
        self.inner.committee_tracker.send_modfiy(|tracker| {
            tracker.update_active_committees(latest);
        });
```

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -611,6 +604,59 @@ where
     fn end_committee_change(&self, epoch: Epoch) -> Result<(), EndCommitteeChangeError> {
         self.end_committee_change_to(epoch)
     }
+
+    async fn begin_committee_change_to_latest_committee(
```
giac-mysten, line 627: Question/imporant: since were are completely replacing the committee with the current one fetched from chain, should we also remove the old entries in `services`, in addition to extending with the new ones? similar to `end_commitee_change_to`?

crates/walrus-service/src/node/committee/committee_service.rs
giac-mysten: Thanks @halfprice ! left just a few nits and questions.

crates/walrus-service/src/node/committee/committee_service.rs
giac-mysten: maybe i'm wrong, but seems to me that the `end_committee_change` will only remove the services for the previous committee (newly updated), instead of all the ones that may be left from the original (very old) committee.

crates/walrus-service/src/node/committee/committee_service.rs
halfprice: Yeah, `add_members_from_committee` is already called inside `create_services_from_committee`: https://github.com/MystenLabs/walrus/blob/9e74f03a72791244ed79da22652e893e84e388e9/crates/walrus-service/src/node/committee/committee_service.rs#L691

crates/walrus-service/src/client.rs
halfprice: The reason for this is exactly what @giac-mysten mentioned: after retry, before finishing the retry, another epoch change is triggered. This is primarily due to that in the test, I set epoch duration to 30 seconds, and blob uploads also takes around 30 seconds (simulated time) to finish.

In the latest commit, I changed the workload to use very small blob size, which helps with this situation. So I reverted the repeated retry. Let's see if CI is ok.

crates/walrus-service/src/node/committee/committee_service.rs
halfprice: Done

crates/walrus-service/src/node/committee/committee_service.rs
halfprice: Good call out. I seems to recall that when @jpcsmith was talking about committee change, he mentioned that for any ongoing RPCs, removing the service would cause the RPC to hang. However, I just test it that before creating new services, I clear the existing services since we will just use the latest one. The test seems ok.

giac-mysten: I see, where do yuo remove the old services?

halfprice: Line 638 below, where I simply clear the services map.

halfprice: Hi @giac-mysten , I'm going to merge the PR for now, but feel free to let me know if you have other comments. Thanks for pointing out the issue regarding removing services.

PR #1140, opened 2024-11-04
chore: replace cargo-sort by taplo

mlegner:


.pre-commit-config.yaml
```diff
@@ -25,15 +25,15 @@ repos:
...
-- repo: https://github.com/DevinR528/cargo-sort
-  rev: 55ec89082466f6bb246d870a8d56d166a8e1f08b
-  hooks:
-  - id: cargo-sort
-    args: ["--workspace"]
 - repo: https://github.com/EmbarkStudios/cargo-deny
   rev: 0.16.1
   hooks:
   - id: cargo-deny
+- repo: https://github.com/ComPWA/taplo-pre-commit
```
karlwuest: Suggestion: Instead of taking a dependency on this repo for the pre-commit hooks, can we just call taplo directly? It's a bit hard to tell how trustworthy the repo is (and it doesn't seem affiliated with taplo) and running the pre-commit can in principle execute arbitrary code if the code in the repo changes.
If this is to cumbersome to do, just specifying the version using a commit hash should also be fine.

mlegner: Good point. I've added a local hook instead in d88e5ea2d4a02ff6cd11868c7dcec0799e503139.

PR #1143, opened 2024-11-04
feat(test): Use checkpoint based event processor in tests wherever possible

sadhansood:
We start using the checkpoint based event processor everywhere in tests with this PR. Requires running event processor in a seperate runtime in tokio runtime environment for tests to not timeout.

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -13,7 +13,8 @@ use std::{
     num::NonZeroU16,
     path::PathBuf,
     str::FromStr,
-    sync::{Arc, Mutex},
+    sync::{Arc, Mutex, MutexGuard, OnceLock},
+    thread,
 };
 
 use async_trait::async_trait;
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/test_utils.rs
```diff
@@ -875,6 +880,62 @@ fn committee_partner(node_config: &StorageNodeTestConfig) -> Option<StorageNodeT
...
+            if let Err(error) = status {
+                tracing::error!("event processor stopped with error: {:?}", error);
+                std::process::exit(1);
+            }
+        }
+        .instrument(tracing::info_span!("cluster-node", address = %rest_api_address)),
+    )
+}
+
+/// Returns a committee constructed from the provided members.
```
halfprice: Remark: the commet is irrelevent.

crates/walrus-service/src/test_utils.rs
```diff
@@ -1908,6 +1965,22 @@ pub fn storage_node_config() -> WithTempDir<StorageNodeConfig> {
...
 }
 
+async fn wait_for_event_processor_to_start(
+    event_processor: EventProcessor,
+    client: sui_rest_api::Client,
+) -> anyhow::Result<()> {
+    // Wait until event processor is actually running and downloaded a few checkpoints
+    tokio::time::sleep(Duration::from_secs(5)).await;
+    let checkpoint = client.get_latest_checkpoint().await?;
+    while let Some(event_processor_checkpoint) = event_processor.checkpoint_store.get(&())? {
```
halfprice, line 2033: Question: can `checkpoint_store` be empty, say in a brand new node? If so, this will return immediately without waiting for any checkpoint to be downloaded.

sadhansood: No, it can't be. When we bootstrap the node - we initialize the `checkpoint_store` with the first checkpoint where system contract was published.

crates/walrus-service/src/test_utils.rs
sadhansood: fixed

PR #1154, opened 2024-11-06
feat(staking): adds early withdraw functionality

damirka:
Approach:

- request for withdrawal before activation epoch means "scheduling" withdrawal in the activation epoch
- given that we don't know the `pool_token_amount` before the activation epoch, we:
  - set amount in `StakedWal` to `option::none` (impossible to get this value otherwise)
  - set `withdraw_epoch` in `StakedWal` as `activation_epoch + 1` (mind the `+1`) 
  - add pending **early** withdrawal to be calculated in `process_pending_stake`
- during epoch advancement, we calculate `pool_token_amount` for early withdrawals knowing that their `activation_epoch` is the previous one; and then deduct from the pool values

### DIFF
--- a/contracts/walrus/sources/staking/staking_pool.move
+++ b/contracts/walrus/sources/staking/staking_pool.move
@@ -40,6 +40,8 @@ const ENotWithdrawing: u64 = 10;
 const EWithdrawEpochNotReached: u64 = 11;
 /// Attempt to withdraw before `activation_epoch`.
 const EActivationEpochNotReached: u64 = 12;
+/// Requesting withdrawal for the stake that can be withdrawn directly.
+const EWithdrawDirectly: u64 = 13;
 
 /// Represents the state of the staking pool.
 ///
@@ -246,7 +248,7 @@ public(package) fun request_withdraw_stake(
     // reached, and the stake is already counted for the next committee selection
     if (staked_wal.activation_epoch() > wctx.epoch()) {
         // only allow requesting if the stake cannot be withdrawn directly
-        assert!(!staked_wal.can_withdraw_early(wctx));
+        assert!(!staked_wal.can_withdraw_early(wctx), EWithdrawDirectly);
 
         let withdraw_epoch = staked_wal.activation_epoch() + 1;
         // register principal in the early withdrawals, the value will get converted to
@@ -256,7 +258,7 @@ public(package) fun request_withdraw_stake(
         return
     };
 
-    assert!(staked_wal.activation_epoch() <= wctx.epoch());
+    assert!(staked_wal.activation_epoch() <= wctx.epoch(), EActivationEpochNotReached);
 
     // If the node is in the committee, the stake will be withdrawn in E+2,
     // otherwise in E+1.
--- a/contracts/walrus/tests/staking/staking_pool/pool_early_withdraw_tests.move
+++ b/contracts/walrus/tests/staking/staking_pool/pool_early_withdraw_tests.move
@@ -59,6 +59,7 @@ fun withdraw_processing_at_different_epochs() {
     assert_eq!(pool.wal_balance_at_epoch(E1), 1000);
     let balance = pool.withdraw_stake(alice, &wctx);
     assert_eq!(balance.destroy_for_testing(), 1000);
+    assert_eq!(pool.wal_balance_at_epoch(E1), 0);
 
     // Bob stakes before committee selection, stake applied E+1
     let mut bob = pool.stake(mint_balance(1000), &wctx, ctx);
@@ -117,7 +118,7 @@ fun withdraw_processing_at_different_epochs() {
 }
 
 
-#[test, expected_failure]
+#[test, expected_failure(abort_code = walrus::staking_pool::ENotWithdrawing)]
 // Scenario:
 // 1. Alice stakes in E0,
 // 2. Committee selected
@@ -138,7 +139,7 @@ fun request_withdraw_after_committee_selection() {
     destroy(pool);
 }
 
-#[test, expected_failure]
+#[test, expected_failure(abort_code = walrus::staking_pool::EWithdrawDirectly)]
 // Scenario (see `request_withdraw_can_withdraw_directly` for symmetrical successful case)
 // 1. Alice stakes in E0 before committee selection
 // 2. Alice requests withdrawal (even though it's not needed)


contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -134,6 +134,9 @@ public(package) fun new(
     let mut exchange_rates = table::new(ctx);
     exchange_rates.add(activation_epoch, pool_exchange_rate::empty());
 
+    // adds `PendingValues` for early withdrawals for each pool
+    df::add(&mut id, EarlyWithdrawalsKey(), pending_values::empty());
```
karlwuest: Suggestion: Add a TODO comment here + an issue to replace this before mainnet with a normal field.

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -321,14 +347,32 @@ public(package) fun advance_epoch(
 public(package) fun process_pending_stake(pool: &mut StakingPool, wctx: &WalrusContext) {
     let current_epoch = wctx.epoch();
 
+    // each value in pending withdrawals contains the principal which became
+    // active in the previous epoch. so unlike other pending values, we need to
+    // flush it one by one, recalculating the exchange rate and pool token amount
+    // for each early withdrawal epoch.
```
karlwuest, line 369: Question: This should always only be a single epoch (for active pools), right?

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -215,24 +218,39 @@ public(package) fun request_withdraw_stake(
...
 
-    // If the node is in the committee, the stake will be withdrawn in E+2,
-    // otherwise in E+1.
-    // TODO: add a check that the node is in the committee: `node_in_committee &&`
-    // let node_in_committee = wctx.committee().contains(pool.id.as_inner());
-    let withdraw_epoch = if (wctx.committee_selected()) {
-        wctx.epoch() + 2
+    // if stake hasn't been activated yet, we mark it as withdrawing starting
+    // its activation epoch before the committee selection.
+    let (withdraw_epoch, token_amount) = if (staked_wal.activation_epoch() > wctx.epoch()) {
```
karlwuest: Important: I think this check is not sufficient, since this succeeds with stake that has not affected the committee selection yet. This stake could afterwards be immediately withdrawn with the special case in the withdraw function. This will then cause the total stake in the pool to be wrong, because the withdrawal is counted twice (once, because we add it to the pending early withdrawals, and once, because the pending stake is reduced in the withdraw function).

We can either disallow this behaviour here, i.e. if stake has not affected the committee selection yet, it can only be withdrawn directly, or, and this is probably nicer, we need to add an additional check in the withdraw function for the withdrawal of non-active stake.

contracts/walrus/tests/staking/staking_pool/pool_early_withdraw_tests.move
```diff
@@ -34,48 +41,124 @@ fun withdraw_before_activation_before_committee_selection() {
...
+
+//     // Alice stakes in E0 after committee selection, withdraws in the next epoch
+//     let (wctx, ctx) = test.select_committee();
+//     let sw1 = pool.stake(mint_balance(1000), &wctx, ctx);
+//     assert_eq!(sw1.activation_epoch(), E2);
+//     assert_eq!(pool.wal_balance_at_epoch(E2), 1000);
+
+//     let (wctx, _ctx) = test.next_epoch();
+//     let amount = pool.withdraw_stake(sw1, &wctx).destroy_for_testing();
+
```
karlwuest: Question: what is the plan with these commented-out tests?

contracts/walrus/tests/staking/staking_pool/pool_early_withdraw_tests.move
```diff
@@ -1,6 +1,13 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
+// Early withdrawal mechanics (for E0, E0', E1, E1', E2, E2'):
+// ```
+// - stake(E0,  AE=E1) -> immediate withdrawal(E0)              // no rewards
+// - stake(E0,  AE=E1) -> request_withdraw(E0') -> withdraw(E2) // rewards for E1
+// - stake(E0', AE=E2) -> immediate withdrawal(E0', E1)         // no rewards
+// - stake(E0', AE=E2) -> request_withdraw(E1') -> withdraw(E3) // rewards for E2
+
```
karlwuest, line 10: Can you add a test for the behaviour mentioned in one of my comments above, where Alice stakes, then immediately requests withdrawal before withdrawing directly (all before the committee change), then do committee change + epoch change and check if the pool balance (WAL and tokens) is correct?

contracts/walrus/sources/staking/staking_pool.move
damirka: For active pool - yes.

contracts/walrus/tests/staking/staking_pool/pool_early_withdraw_tests.move
damirka: If we decide to go this route, I'll add tests for failure scenarios in early withdrawing.

contracts/walrus/sources/staking/staking_pool.move
karlwuest: Since we decided to use feature flags in the corresponding rust code, we can replace this with a normal field now already.

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -78,11 +80,46 @@ public(package) fun into_balance(sw: StakedWal): Balance<WAL> {
...
+    // if stake is to be applied in the next epoch but committee hasn't been selected
+    if (
+        !is_withdrawing &&
+        sw.activation_epoch == current_epoch + 1 &&
+        !wctx.committee_selected()
+    ) {
+        return true
+    };
+
+    false
```
karlwuest: I think we can simplify this a bit by adding the `is_withdrawing` to the early return:
```suggestion
    // early return if stake is already active or if withdrawal has been requested
    if (activation_epoch <= current_epoch || is_withdrawing) {
        return false
    };
    activation_epoch == current_epoch + 2 || 
    activation_epoch == current_epoch + 1 && !wctx.committee_selected()
```

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -11,6 +11,7 @@
 // - advance_epoch - initiates the epoch change
 // - initiate epoch change - bumped in `advance_epoch`
 // - get "epoch_sync_done" event
+#[allow(deprecated_usage)]
```
karlwuest: Question: What is this needed for? Was this for the `FixedPoint32`?

damirka: Yup. Exactly for it.

contracts/walrus/tests/staking/staking_pool/pool_early_withdraw_tests.move
```diff
@@ -34,48 +41,140 @@ fun withdraw_before_activation_before_committee_selection() {
...
     assert_eq!(pool.wal_balance_at_epoch(E1), 1000);
+    assert_eq!(sw1.activation_epoch(), E1);
 
     let (wctx, _ctx) = test.select_committee();
+    let _ = pool.withdraw_stake(sw1, &wctx).destroy_for_testing();
+
+    destroy(pool);
+}
+
+#[test, expected_failure]
```
karlwuest: Nit: Can you add the abort code to the expected_failure here?

contracts/walrus/tests/staking/staking_pool/pool_early_withdraw_tests.move
```diff
@@ -34,48 +41,140 @@ fun withdraw_before_activation_before_committee_selection() {
...
+    let mut pool = pool().build(&wctx, ctx);
+
+    // Alice stakes before committee selection, stake applied E+1
+    // And she performs the withdrawal right away
+    let alice = pool.stake(mint_balance(1000), &wctx, ctx);
+    assert_eq!(alice.activation_epoch(), E1);
+    assert_eq!(pool.wal_balance_at_epoch(E1), 1000);
+    let balance = pool.withdraw_stake(alice, &wctx);
+    assert_eq!(balance.destroy_for_testing(), 1000);
+
```
karlwuest, line 63: Suggestion: add another assert for the pool balance
```suggestion
    assert_eq!(pool.wal_balance_at_epoch(E1), 0);
    
```

contracts/walrus/tests/staking/staking_pool/pool_early_withdraw_tests.move
damirka: Yup, added here an in another place too.

PR #1155, opened 2024-11-08
fix: sending epoch_sync_done should not block event processing

halfprice:
This PR implements that sending `epoch_sync_done` does not block next event processing.

We renamed `BackgroundShardRemover` to `StartEpochChangeFinisher`, which in addition to removing shards, also sends `epoch_sync_done` message.

Note that it possible that `epoch_sync_done` is not executed successfully, but the next epoch arrives. In this case, `epoch_sync_done` would abort due to unexpected epoch, and the event should be able to complete.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -908,7 +908,7 @@ impl StorageNode {
         // Last but not least, we need to remove any shards that are no longer owned by the node.
         if !shard_diff.removed.is_empty() {
             self.start_epoch_change_finisher
-                .start_finish_epoch_change_start_tasks(
+                .start_finish_epoch_change_tasks(
                     element_index,
                     event,
                     shard_diff.removed.clone(),
@@ -997,8 +997,7 @@ impl StorageNode {
         }
 
         // Here we need to wait for the previous shard removal to finish so that for the case
-        // where
-        // same shard is moved in again, we don't have shard removal and move-in running
+        // where same shard is moved in again, we don't have shard removal and move-in running
         // concurrently.
         //
         // Note that we expect this call to finish quickly because removing RocksDb column
@@ -1025,7 +1024,7 @@ impl StorageNode {
         }
 
         self.start_epoch_change_finisher
-            .start_finish_epoch_change_start_tasks(
+            .start_finish_epoch_change_tasks(
                 element_index,
                 event,
                 shard_diff.removed.clone(),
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1007,21 +996,22 @@ impl StorageNode {
                 .context("failed to lock shard")?;
         }
 
-        // Here we need to wait for the previous shard removal to finish so that for the case where
+        // Here we need to wait for the previous shard removal to finish so that for the case
+        // where
```
karlwuest: Nit: merge with next line.

crates/walrus-service/src/node/start_epoch_change_finisher.rs
```diff
@@ -0,0 +1,165 @@
...
+            task_handle: Arc::new(Mutex::new(None)),
+        }
+    }
+
+    /// Starts background tasks to finish the epoch change.
+    /// This includes:
+    ///    - Sending epoch sync done if there is no newly scheduled shard syncs.
+    ///    - Removing no longer owned storage for shards.
+    ///    - Marking the event as completed.
+    pub fn start_finish_epoch_change_start_tasks(
```
karlwuest: Nit: This name sound a bit confusing, maybe remove the second "start"?
```suggestion
    pub fn start_finish_epoch_change_tasks(
```

crates/walrus-service/src/node/start_epoch_change_finisher.rs
```diff
@@ -0,0 +1,165 @@
...
+        shards: Vec<ShardIndex>,
+        committees: ActiveCommittees,
+        ongoing_shard_sync: bool,
+    ) {
+        let self_clone = self.clone();
+        let event_clone = event.clone();
+        let committees_clone = committees.clone();
+
+        let mut locked_task_handle = self.task_handle.lock().unwrap();
+        assert!(locked_task_handle.is_none());
```
karlwuest, line 54: Question: Do we assume that `start_finish_epoch_change_start_tasks` is only ever called on a newly constructed `StartEpochChangeFinisher`?
Suggestion: If that is the case, it may be better to remove the explicit constructor and construct + return it in the start function to prevent the function from being called on a `StartEpochChangeFinisher` that already has a handle set or otherwise we could return an error instead of panicking.

halfprice: StartEpochChangeFinisher is a long lived object in the node. There is only every going to be one instance, and we call its start_finish_epoch_change_start_tasks for every epoch change start event.

crates/walrus-service/src/node/start_epoch_change_finisher.rs
halfprice: Yeah, that sounds a lot better

PR #1157, opened 2024-11-08
feat(client): implement client mulitplexer and parallel-write publisher

giac-mysten:
Improves the publisher, adding multiple clients that can serve write requests in parallel. Since by default the number of clients is the same as the maximum number of concurrent requests allowed by the axum server, each client typically serves only one request at the time (note that this is not strictly enforced though).

The gas and wal refill infrastructure from the stress client is repurposed to ensure the clients have sufficient funds, and it has been moved to the service crate.

### DIFF
--- a/crates/walrus-service/src/client/multiplexer.rs
+++ b/crates/walrus-service/src/client/multiplexer.rs
@@ -149,7 +149,7 @@ impl WriteClientPool {
         refiller: &Refiller<G>,
     ) -> anyhow::Result<Self> {
         tracing::info!(%n_clients, "creating write client pool");
-        let pool = SubClientBuilder::new(config, sub_wallets_dir, sui_env, gas_budget, refiller)
+        let pool = SubClientLoader::new(config, sub_wallets_dir, sui_env, gas_budget, refiller)
             .create_or_load_sub_clients(n_clients)
             .await?;
 
@@ -182,7 +182,7 @@ impl WriteClientPool {
 }
 
 /// Helper struct to build or load sub clients for the client multiplexer.
-struct SubClientBuilder<'a, G> {
+struct SubClientLoader<'a, G> {
     config: &'a Config,
     sub_wallets_dir: &'a Path,
     sui_env: SuiEnv,
@@ -190,7 +190,7 @@ struct SubClientBuilder<'a, G> {
     refiller: &'a Refiller<G>,
 }
 
-impl<'a, G: CoinRefill> SubClientBuilder<'a, G> {
+impl<'a, G: CoinRefill> SubClientLoader<'a, G> {
     fn new(
         config: &'a Config,
         sub_wallets_dir: &'a Path,


crates/walrus-service/src/client/dedup_metrics.rs
karlwuest, line 1: Question: Can this file be deleted or did you forget to commit something?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -384,13 +385,21 @@ impl ClientCommandRunner {
...
-            self.wallet,
+        let config = &self.config?;
+        let sui_contract_client = config
+            .new_contract_client(self.wallet?, self.gas_budget)
+            .await?;
+
+        let client = ClientMultiplexer::new(
+            args.n_clients,
+            sui_contract_client,
+            config,
```
karlwuest: Question/Suggestion: Can we just construct the contract client in the `ClientMultiplexer` instead of passing it in?

crates/walrus-service/src/client/daemon.rs
```diff
@@ -27,33 +27,71 @@ use tower::{
...
 
+pub trait WalrusReadClient {
+    fn read_blob(
+        &self,
+        blob_id: &BlobId,
+    ) -> impl std::future::Future<Output = ClientResult<Vec<u8>>> + Send;
+    fn set_metric_registry(&mut self, registry: &Registry);
+}
+
+pub trait WalrusWriteClient: WalrusReadClient {
```
karlwuest, line 47: Question: Is it necessary that a WalrusWriteClient implements the WalrusReadClient trait?

crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -0,0 +1,251 @@
...
+        let cur_idx = *idx;
+
+        let client = self
+            .pool
+            .get(cur_idx)
+            .expect("the index is computed modulo the length and clients cannot be removed")
+            .clone();
+
+        let next = (cur_idx + 1) % self.pool.len();
+        *idx = next;
```
karlwuest: Nit: This could be reordered to release the lock earlier. I don't think we gain anything from locking across the `pool.get(cur_idx)`.  You could also potentially replace the mutex with an atomic counter to achieve the same thing.

Remark: This does not prevent the same client from being used multiple times. If that's intended, that's fine, just want to make sure you are aware of this.

crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -0,0 +1,251 @@
...
+        client
+    }
+}
+
+/// A temporary write client for client multiplexing.
+#[derive(Debug)]
+pub(crate) struct TempWriteClient(WithTempDir<Client<SuiContractClient>>);
+
+impl TempWriteClient {
```
karlwuest: Important: I don't think that we should use temporary clients for the multiplexer that will be used in production on publishers, since this will lead to lost funds if the publisher crashes. I would suggest to instead create a directory that will store all the sui configs and keys for the created clients. We could then also have an option to load existing clients from that directory when the Multiplexer is created.

crates/walrus-service/src/client/dedup_metrics.rs
giac-mysten: yes, forgt about adding it to the commit, sorry

crates/walrus-service/src/client/daemon.rs
giac-mysten: It is, because the `ClientDaemon<T>` serves both reads and writes from the `T`

karlwuest: But wouldn't it be sufficient then for each of the implementors to implement both traits separately?
It's fine to leave it as is, just to be clear, I was just wondering if there is a necessity for it.

crates/walrus-service/src/client/multiplexer.rs
giac-mysten: >Nit: This could be reordered to release the lock earlier. I don't think we gain anything from locking across the pool.get(cur_idx). You could also potentially replace the mutex with an atomic counter to achieve the same thing.

i like the idea of an atomic counter, changed to that 👍 

>Remark: This does not prevent the same client from being used multiple times. If that's intended, that's fine, just want to make sure you are aware of this.

correct, this is a form of load balancing for the moment. In principle, a client may be serving more than request at the time. (the pr description mentions that the 1-req-per-client is not enforced)

crates/walrus-service/src/client/multiplexer.rs
giac-mysten: good point! wanted to keep this as a follow up, but was quite quick to add.

crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -136,116 +145,128 @@ impl WriteClientPool {
...
-        *idx = next;
         client
     }
 }
 
-/// A temporary write client for client multiplexing.
-#[derive(Debug)]
-pub(crate) struct TempWriteClient(WithTempDir<Client<SuiContractClient>>);
+/// Helper struct to build or load sub clients for the client multiplexer.
+struct SubClientBuilder<'a, G> {
```
karlwuest: Nit: This isn't really using the builder pattern, maybe call it `SubClientLoader` instead?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -377,10 +377,13 @@ pub struct PublisherArgs {
     #[clap(long, default_value_t = default::n_publisher_clients())]
     #[serde(default = "default::n_publisher_clients")]
     pub n_clients: usize,
-    /// The interval of time between refilling the publisher's sub clients' wallets.
+    /// The interval of time between refilling the publisher's sub-clients' wallets.
     #[clap(long, value_parser = humantime::parse_duration, default_value = "1s" )]
     #[serde(default = "default::refill_interval")]
     pub refill_interval: Duration,
+    /// The directory where the publisher will store the sub-wallets used for client multiplexing.
+    #[clap(long)]
```
karlwuest, line 385: Suggestion: add a default path for the `sub_wallets_dir`.

giac-mysten: i went back and forth on this. It's important that the operator knows where the wallet configs are, otherwise funds may be lost. So my idea was to ensure this by forcing to specify the path. Also, what if the default path does not exist? should it be created? wdyt?

karlwuest: Hmmm... that's a good point. Let's keep it as is, then 👍

Issue #1156, opened 2024-11-08
Make event handling more robust against programming errors


mlegner:
Currently, our event handling is somewhat brittle and error-prone: It relies on several implicit conventions. In particular, it assumes that every event is marked as completed exactly once by calling the `StorageNodeInner::mark_event_completed` function. With a now relatively large number of event types and sometimes complex handling, it becomes quite likely that for some code path, this function is called multiple times or not at all, which causes the whole node to crash or stall.

It would be better if we relied on the compiler for ensuring this expectation. This can be achieved by adding a new `EventHandle` struct with a custom `Drop` implementation that marks the event complete. By creating this exactly once for each event when it is read, not having any `Clone` or `Copy` functionality, we thus achieve the intended goal.

PR #1158, opened 2024-11-10
fix: earlier EpochChangeStart is not marked as complete in nodes in RecoveryCatchUp state

halfprice:
Also add a simtest which triggers repeated node crash and recovery in a single node.

We check that in the test, the event cursor should continue make progress eventually. Without the fix, the test fails with stuck cursor.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -18,5 +18,5 @@ slow-timeout = { period = "2m", terminate-after = 5 } # Mark E2E tests as slow o
 threads-required = 4
 
 [profile.simtest]
-slow-timeout = { period = "5m", terminate-after = 3 } # Timeout tests after 15 minutes
+slow-timeout = { period = "5m", terminate-after = 6 } # Timeout tests after 30 minutes
 test-threads = "num-cpus"
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -549,6 +549,7 @@ impl Storage {
             .certified_blob_info_iter_before_epoch(epoch, std::ops::Bound::Unbounded)
     }
 
+    /// Returns the current event cursor.
     pub(crate) fn get_event_cursor_progress(&self) -> Result<EventProgress, TypedStoreError> {
         self.event_cursor.get_event_cursor_progress()
     }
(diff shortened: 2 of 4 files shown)


.config/nextest.toml
```diff
@@ -18,5 +18,5 @@ slow-timeout = { period = "2m", terminate-after = 5 } # Mark E2E tests as slow o
 threads-required = 4
 
 [profile.simtest]
-slow-timeout = { period = "5m", terminate-after = 3 } # Timeout tests after 15 minutes
+slow-timeout = { period = "5m", terminate-after = 6 } # Timeout tests after 30 minutes
```
halfprice, line 21: The test is getting bigger and it's hard to finish them in a timely manner in CI (local testing is a lot faster), even though I have tried to make it as lightweight as possible...

So I had to increase the timeout.

crates/walrus-service/src/node.rs
```diff
@@ -830,7 +835,7 @@ impl StorageNode {
                 .begin_committee_change_to_latest_committee()
                 .await?;
 
-            let mut need_to_mark_event_complete = false;
+            let mut need_to_mark_event_complete = true;
```
halfprice, line 838: This one line of code change is the fix. Sorry for not catching this in the original PR.

mlegner: I'm actually working on a setup that should make our event handling more robust against this type of errors (#1156).

crates/walrus-sdk/src/api.rs
```diff
@@ -291,3 +293,13 @@ pub enum ShardStatus {
...
+
+/// Represents the progress of the events.
+#[derive(Debug, Clone, Copy, Default, Deserialize, Serialize, utoipa::ToSchema, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct EventProgress {
+    /// The number of events that have been persisted.
+    pub persisted: u64,
+    /// The number of events that are pending in memory.
+    pub pending: u64,
+}
```
mlegner, line 305: Remark: Moving this here makes it part of our external API. In other such cases we sometimes had two types: The external type in the `walrus-sdk` crate and an internal type in `walrus-service`. I guess we could keep a single type in this case, need to be careful if we ever want to change it.

halfprice: Let me do it in a separate PR.

PR #1161, opened 2024-11-11
test: create a test that reproduce the issue that extending blob doesn't extend blob registration

halfprice:
Syncing extended certified blob currently is blocked due to that `is_blob_registered` returns false after the original end_epoch is passed, even though the blob is extended.

Related issue: #1163

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -3983,7 +3983,7 @@ mod tests {
             .inner
             .is_blob_certified(blob_details.blob_id())?);
 
-        // TODO: fix that blob registration is not extended when extending blob life time.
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -3941,4 +3941,54 @@ mod tests {
...
+        )?;
+
+        advance_cluster_to_epoch(&cluster, &[&events], 5).await?;
+
+        assert!(cluster.nodes[0]
+            .storage_node
+            .inner
+            .is_blob_certified(blob_details.blob_id())?);
+
+        // TODO: fix that blob registration is not extended when extending blob life time.
```
mlegner:
```suggestion
        // TODO: fix that blob registration is not extended when extending blob life time (#1163).
```

PR #1164, opened 2024-11-11
fix(node): correct status tracking for blob extensions

### LINKED ISSUE #1163: Internal blob tracking does not correctly handle blob extensions
The registration expiration is not correctly updated such that nodes conclude the blob is no longer registered. See #1161 for the resulting issue.

mlegner:
Contains a temporary fix for already existing `blob_info` entries that may be incorrect due to the bug.

The added tests failed before the change and pass with the change.

Closes #1163

### DIFF
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -456,12 +456,13 @@ pub(crate) enum BlobInfoV1 {
     Valid(ValidBlobInfoV1),
 }
 
-// INV: count_deletable_total > count_deletable_certified
+// INV: count_deletable_total >= count_deletable_certified
 // INV: permanent_total.is_none() => permanent_certified.is_none()
-// INV: permanent_total.count > permanent_certified.count
-// INV: permanent_total.end_epoch > permanent_certified.end_epoch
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -865,9 +905,20 @@ impl BlobInfoApi for BlobInfoV1 {
...
         let exists_registered_permanent_blob = permanent_total
             .as_ref()
-            .is_some_and(|p| p.end_epoch > current_epoch);
+            .is_some_and(|p| p.end_epoch > current_epoch)
+
+            // TODO(mlegner): This is a temporary workaround due to a previous bug (#1163) with the
+            // blob-status tracking. This is no longer needed after a full redeployment.
+            || permanent_certified
+                .as_ref()
+                .is_some_and(|p| p.end_epoch > current_epoch);
```
halfprice, line 914: I suspect that this is needed unless we run a one shot job to fix all the existing register. I think it's fine to just leave the additional check permanenetly.

mlegner: Yes, exactly. We can revisit this in a few months and decide if we want to keep it or remove it (will be slightly more performant).

PR #1165, opened 2024-11-11
refactor(node): make event handling more robust

### LINKED ISSUE #1156: Make event handling more robust against programming errors

mlegner:
This introduces a new `EventHandle` struct, which is used to track handling of events and to mark them as complete. As this struct does not implement `Clone`, it is impossible to mark the same event as complete multiple times. Furthermore, it automatically logs an error if it is dropped without being marked as complete.

No actual behavior is changed, but this should prevent programming errors in the future.

Closes #1156


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -402,13 +359,14 @@ impl BlobSynchronizer {
     }
 
     #[tracing::instrument(skip_all)]
-    async fn run(&self, sliver_permits: Arc<Semaphore>) -> anyhow::Result<()> {
+    async fn run(&self, sliver_permits: Arc<Semaphore>) {
```
halfprice, line 431: Question: why removed error propagation?

crates/walrus-service/src/node/system_events.rs
```diff
@@ -17,13 +18,142 @@ use walrus_event::{
...
+    fn mark_as_incomplete_during_shutdown(self);
+}
+
+impl CompletableHandle for EventHandle {
+    fn mark_as_complete(self) {
+        self.mark_as_complete();
+    }
+
+    fn mark_as_incomplete_during_shutdown(mut self) {
+        self.can_be_dropped = true;
```
halfprice: I suspect that it'll be challenging to mark all the events that currently exist in the system when shutting down. There could be other events in the middle of processing.

I think a more robust way is to check that when an event is dropped, whether the event stream is still open. If so, it indicates that the stream is still going and we shouldn't drop an event without marking it complete. If the stream is gone, it's ok to drop them without marking them as complete.

mlegner: Very good idea. I agree that this mechanism is also error-prone (exactly what I wanted to avoid 😅) but didn't think of this option. Let me try to implement this.

crates/walrus-service/src/node/blob_sync.rs
mlegner: We already [panic in case of DB issues with slivers](https://github.com/MystenLabs/walrus/pull/1165/files/1948881d4745f13351894b7c3d159eac2258bcde#diff-51d59d2838dbf3ec2d6b89da95cd7419432c6149658c2938360ba7495eeab257R420), this just makes the handling of DB errors in the metadata recovery consistent with this.

crates/walrus-service/src/node/system_events.rs
mlegner: I made an attempt in a76a1a964b205c413a6f89c1588261cd84bd7ba2. It's a bit ugly, and I'm not sure if that is the best option. I tried using the cancellation token, but it is a bit tricky to set on the `StorageNodeInner`. Let me know if you have a better idea.

crates/walrus-service/src/node.rs
```diff
@@ -1205,6 +1211,14 @@ impl StorageNodeInner {
         self.metrics.current_node_status.set(status.to_i64());
         self.storage.set_node_status(status)
     }
+
+    fn shut_down(&self) {
+        self.is_shutting_down.store(true, Ordering::Relaxed)
```
halfprice: probably should use Ordering::SeqCst so that it always reads the latest value.

crates/walrus-service/src/node/system_events.rs
halfprice: I just rebased the PR to the latest main and looks like right now there are decent amount of failures due to that EventHandle dropping but not mark as completed. Right now, we only call `shut_down()` when the node is explicitly cancelled, but I'm not sure if the node object may get destructed without invoking cancellation.

One quick and hacky way I can think of right now is to expose a test API in storage node, similar to `shut_down()`, that in test body, before finishing the test, we call `node.shut_down()` explicitly. But ideally, if we can define a more robust way to signal events shutting down, that'll be great.

Another thing we should make sure to test is that `systemctl stop` works properly without panicing.

mlegner: Unfortunately, I haven't found a robust way to check for a shutting-down node in simtests. A temporary solution is to just not panic in simtests (970f0cf52124b791fa0ad18ee94a920c69319970)... 🤷 

> Another thing we should make sure to test is that systemctl stop works properly without panicing.

As far as I understand the existing code [here](https://github.com/MystenLabs/walrus/blob/a52970fbfc03f552d543420f4716ebef1e1b24a0/crates/walrus-service/bin/node.rs#L425-L433), we explicitly cancel the `CancellationToken`, which is passed into the `StorageNode::run` function. So this *should* work, but I can try to test it.

crates/walrus-service/src/node.rs
```diff
@@ -884,17 +871,16 @@ impl StorageNode {
             .cancel_all_expired_syncs_and_mark_events_completed()
             .await?;
 
-        self.process_shard_changes_in_new_epoch(element_index, event)
+        self.process_shard_changes_in_new_epoch(event_handle, event)
             .await
     }
 
     /// Starts the node recovery process.
```
halfprice, line 875: Nit: maybe also document that this function must complete `event_handle`

crates/walrus-service/src/node.rs
halfprice: This one was missed?

crates/walrus-service/src/node/start_epoch_change_finisher.rs
```diff
@@ -66,25 +66,24 @@ impl StartEpochChangeFinisher {
...
-                self_clone
-                    .node
-                    .mark_event_completed(event_element, &event_clone.event_id)
+                anyhow::Ok(())
             })
             .await;
 
             if let Err(err) = result {
+                // Question(mlegner): What should we do with the event in this case?
                 tracing::error!(
```
halfprice, line 79: Question: will the node panic in this case, since the event is not handled?

crates/walrus-service/src/node/system_events.rs
```diff
@@ -1,29 +1,159 @@
...
+        match () {
+            _ if thread::panicking() => {
+                tracing::debug!("event handle dropped during panic",);
+            }
+            _ if self.node.is_shutting_down() => {
+                tracing::debug!("event handle dropped during shutdown",);
+            }
+            _ => {
+                tracing::error!("event handle dropped before being marked as complete",);
+                // Panic in tests (not in simtests) if an event handle is dropped.
```
halfprice, line 114: Remark: I suspect that we don't want to skip panicing in simtest, since I feel that is the place where missing mark event complete can be found more effectively. If you agree, maybe add a TODO here, so that it doesn't block this PR?

crates/walrus-service/src/node.rs
mlegner: Ah, indeed. I think I only changed it for reading as it doesn't matter for writing (because we only ever change `false` to `true`), but it also doesn't hurt.

crates/walrus-service/src/node/start_epoch_change_finisher.rs
mlegner: Currently not, the panic is the result of a `debug_assert`. It would just log an error. However, it probably makes sense to panic here as this would not be recoverable, and the event would never be marked as complete. WDYT?

mlegner: Alternatively, we could mark the event as complete as the only reason for an error is that the shard removal failed. This can be done later as well.

mlegner: Actually, this should anyway never happen as we retry indefinitely above. Added a comment about that and made sure that the event is always marked as complete.

crates/walrus-service/src/node/system_events.rs
mlegner: Created a follow-up issue: #1232

PR #1171, opened 2024-11-13
chore: modify workflow to feature flag contract modifications

karlwuest:


contracts/walrus/sources/system.move
```diff
@@ -23,12 +23,15 @@ const VERSION: u64 = 0;
 public struct System has key {
     id: UID,
     version: u64,
+
+    // TODO: remove, only for testing
+    foo: u64,
```
karlwuest: Will remove this (and all changes based on this) from the PR later, this is just here to have a breaking change behind the feature flag to make sure that the CI workflow is correct.

.pre-commit-config.yaml
mlegner, line 1: Nit: This seems to have been reformatted. No strong opinion on the formatting itself, but this somehow clutters the history. 😅

crates/walrus-sui/Cargo.toml
```diff
@@ -1,13 +1,14 @@
...
-version.workspace = true
 edition.workspace = true
 license.workspace = true
+name = "walrus-sui"
+publish = false
+version.workspace = true
 
 [features]
 default = []
+mainnet-contracts = []
```
mlegner, line 11: Question: Is the idea that we mark all code paths that are (only) relevant for the new contract version with a `#[cfg(feature = "mainnet-contracts")]`? Do we need a similar feature on `walrus-service` and/or other crates?

Suggestion: Document this somewhere.

.github/workflows/code.yml
```diff
@@ -119,10 +119,17 @@ jobs:
...
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
       - uses: Swatinem/rust-cache@v2.7.5
         with:
           save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
-      - name: Run tests
+      - name: Run tests with mainnet contracts
+        run: cargo nextest run --profile ci --run-ignored all --features mainnet-contracts
```
mlegner: AFAIU, this enables the feature `mainnet-contracts` on all crates that have it. Maybe it would be better to only enable this for the crate where we actually expect it:
```suggestion
        run: cargo nextest run --profile ci --run-ignored all --features walrus-sui/mainnet-contracts
```

crates/walrus-sui/Cargo.toml
karlwuest: Yes, that's the idea. I don't think that we need a similar feature yet on any of the other crates, but we may need it in the future.

What would be the best location to document this? In CONTRIBUTING.md ?

.pre-commit-config.yaml
karlwuest: Let me try to figure out why this got reformatted, I don't think I changed anything in my settings.

karlwuest: Actually, for some of the files for the github actions we use the same indentation as this reformatted version and for some we use the old one, so this is generally very inconsistent. I would be in favour of changing the formatting for all yaml files to be consistent 😅 but I don't really care which one we use.

mlegner: Haha, I can add (yet another) linter to our already impressive portfolio. 😆

crates/walrus-sui/Cargo.toml
mlegner: Yes, I think the CONTRIBUTING.md file would be a good place. 👍

.pre-commit-config.yaml
karlwuest: Having a linter for this would probably actually be good. It seems that block sequences should be indented as in the formatted version: https://yaml.org/spec/1.2.2/#821-block-sequences

mlegner: I will add https://github.com/adrienverge/yamllint.

crates/walrus-sui/Cargo.toml
karlwuest: Added some documentation there, ptal!

.github/workflows/code.yml
```diff
@@ -119,10 +119,34 @@ jobs:
         with:
           tool: cargo-nextest
       - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
```
karlwuest:
```suggestion
```

CONTRIBUTING.md
```diff
@@ -62,16 +62,42 @@ Also make sure you use the correct version of Rustfmt. See
...
 The majority of our code is covered by automatic unit and integration tests which you can run
 through `cargo test` or `cargo nextest run` (require [nextest](https://nexte.st/)).
 
 Integration and end-to-end tests are excluded by default when running `cargo nextest` as they depend on
-additional packages and take longer to run. You can run these test as follows:
+additional packages and take longer to run. These tests can either be run with the newest contract
+version (including breaking changes that are incompatible with testnet) or they can be run with the
+contracts as deployed on testnet.
+
+For the first option, you can run these test as follows:
```
mlegner: Typo:
```suggestion
For the first option, you can run these tests as follows:
```

CONTRIBUTING.md
```diff
@@ -62,16 +62,42 @@ Also make sure you use the correct version of Rustfmt. See
...
 
 ```sh
+cargo nextest run --run-ignored ignored-only --features walrus-sui/mainnet-contracts
+```
+
+For the second option, you first need to checkout the correct version of the contracts. To do this
+and cleanup afterwards, you can run the following:
+
+```sh
+GIT_STASH_TAG="${RANDOM}" git stash push -m "${GIT_STASH_TAG}" -a -- contracts
```
mlegner: Question: The `-a` flag includes also ignored files, do we want that? While it would be not to lose build files, I had some issues when trying this locally. Maybe `-u` would be better?

Also, I think the variable needs to be defined in a separate command. And, the random number could be included in some other stash message, so better to prefix it with something.

Overall suggestion:
```suggestion
GIT_STASH_TAG="TESTING${RANDOM}"
git stash push -m "${GIT_STASH_TAG}" -u -- contracts
```

crates/walrus-sui/Cargo.toml
```diff
@@ -1,13 +1,14 @@
 [package]
-name = "walrus-sui"
-publish = false
 authors.workspace = true
-version.workspace = true
 edition.workspace = true
 license.workspace = true
+name = "walrus-sui"
+publish = false
+version.workspace = true
```
mlegner: Question: Did this get reformatted automatically?

CONTRIBUTING.md
```diff
@@ -62,16 +62,42 @@ Also make sure you use the correct version of Rustfmt. See
...
-additional packages and take longer to run. You can run these test as follows:
+additional packages and take longer to run. These tests can either be run with the newest contract
+version (including breaking changes that are incompatible with testnet) or they can be run with the
+contracts as deployed on testnet.
+
+For the first option, you can run these test as follows:
 
 ```sh
+cargo nextest run --run-ignored ignored-only --features walrus-sui/mainnet-contracts
+```
```
mlegner, line 88: Looking at this again, I realize that running `cargo nextest run` or `cargo test` by themselves will no longer work as soon as we have breaking changes in the contracts. This is a bit unfortunate as many developers probably use them frequently. 

Unfortunately, I don't have a good solution for this. We could turn around the feature-flag setup and instead have a `testnet-contracts` feature; however, then we need to activate this feature for all our builds instead... 🤷

CONTRIBUTING.md
```diff
@@ -62,16 +62,42 @@ Also make sure you use the correct version of Rustfmt. See
...
 for example Clippy.
 
+## Contract Versions
+
+To allow breaking changes in the contracts, while keeping compatibility with the deployed testnet
+version in the Rust codebase, we gate any incompatible changes in the Rust bindings for the
+contracts behind the `mainnet-contracts` feature in the `walrus-sui` crate.
+
+If you make any changes to the contracts that necessitates corresponding changes in Rust, make sure
+to gate these changes behind the feature, e.g. by adding the `#[cfg(feature = "walrus-mainnet")]` attribute.
```
mlegner: This should probably be
```suggestion
to gate these changes behind the feature, e.g. by adding the `#[cfg(feature = "mainnet-contracts")]` attribute.
```

.github/workflows/code.yml
```diff
@@ -123,6 +123,27 @@ jobs:
...
+      - uses: taiki-e/install-action@v2.44.58
+        with:
+          tool: cargo-nextest
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          fetch-tags: true
+      - uses: Swatinem/rust-cache@v2.7.5
+        with:
+          save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
```
mlegner, line 144: Suggestion: Share the cache between the two test jobs (otherwise we may go over the 10GiB limit):
```suggestion
          save-if: ${{ github.ref == 'refs/heads/main' && 'true' || 'false' }}
          shared-key: v0-rust-test
```
This requires the same change to the other test job above.

CONTRIBUTING.md
karlwuest: Yeah, I had similar thoughts (I think I mentioned this on slack). I don't have strong opinions about this, if you think it's better to have a `testnet-contracts` feature and then specify it for our builds, we can do that as well. For now, it might be easier to just keep it as is, because we anyway don't have external developers accessing this repo yet.

CONTRIBUTING.md
karlwuest: Good catch, thanks!

crates/walrus-sui/Cargo.toml
karlwuest: Yes. I assumed that was from taplo. Should it not have been sorted?

CONTRIBUTING.md
karlwuest: How did you test it locally (so that I can try to reproduce it)? With my local tests it always worked, but using `-u` instead should probably be fine.

>Also, I think the variable needs to be defined in a separate command. And, the random number could be included in some other stash message, so better to prefix it with something.

Good suggestion, I will add a prefix.

crates/walrus-sui/Cargo.toml
mlegner: I think it _should not_ due to this configuration:
```
[[rule]]
include = ["**/Cargo.toml", "**/Move.toml"]
keys = ["bench", "package", "workspace.package"]

[rule.formatting]
align_comments = true
reorder_keys = false
```
Don't know why it does it... 🤷

CONTRIBUTING.md
mlegner: Essentially it just took very long as I had build and trace files there.

crates/walrus-sui/Cargo.toml
karlwuest: Hmmm... maybe it was VSCode, I'll change it back and then try to find out why it's reformatting this.

karlwuest: ~Apparently the "Even Better TOML" extension with taplo does not use the taplo config file. Do you use anything else with VSCode for TOML files?~ 
Nevermind, it just needs restarting to apply the settings...

PR #1184, opened 2024-11-14
feat: add walrus-proxy for metrics

### LINKED ISSUE #571: Testnet metrics & logs from walrus 3rd party operators
More details:

* early on in testnet we might want to ask storage nodes to push metrics to a mysten-run loki
* we'll want a sui-proxy like solution for the "active" storage nodes to push metrics to a mysten-run metrics server
* long term (mainnet) we wont ask storage nodes to push logs

suiwombat:
## Description
create a walrus-proxy, similar to how we have a sui-proxy.  Some items are borrowed, but there is a considerable amount of customization here as well.

General flow of how this works;

* walrus-node spins up a reqwest client and connects to the walrus-proxy over anonymous TLS, accepting well known CA certs like we do today in other parts of our infra.
* the network private key is used to sign a uuidv7 and that signature is sent in an Authorization header along with a message. The signature is a recoverable secp256r1 signature which allows the pub key to be recovered service side as long as the message that was signed is also known.  We send the signature as base64 and also encode the uuidv7 as the message to be signed over as a string.
* the walrus-proxy will inspect this header and extract the pub key from the signature, decoding as needed. If it matches a known walrus-node, we accept the metric.  This authorization header can be swapped out later if we want mTLS, but there is a benefit in using the header mechanism because we can easily avoid tls termination on this service and instead rely on gcp's upstream load balancers to do this for us.
* walrus-proxy side, we have a process that polls for a staking object, we extract committee members and we grab their pub key. this is key to how we know who to accept metrics from.  This is done on a configurable interval and automatically adjusts to nodes joining or leaving the stacking object we depend on.  If the staking object id were to change, then we'd need to update the variable that holds this object id.  If this is a frequent occurrence, we could add an api endpoint to the service to reconfigure itself w/o a restart.
* in a follow-up PR, I will add an additional layer of replay protection using a sliding window approach.  The uuidv7 uses the most significant 48 bits to encode a unix epoch timestamp.  We can order these uuids based on time and apply a replay constraint to middleware to prevent the same message from being used over and over again.  Messages that fall outside of coordinated UTC windows will not be allowed, regardless of being seen prior.

## Test Plan
local tests

working as expected so far. I briefly had it on and captured some data. https://metrics.sui.io/goto/-R1hv-mHg?orgId=1

Closes #571

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -365,8 +365,8 @@ mod commands {
             walrus.node.public_key = %config.protocol_key_pair.load()?.as_ref().public(),
             "walrus protocol public key",
         );
-        // // Load the network_key_pair so that it can be passed to the
-        // // MetricsAndLoggingRuntime.
+        // Load the network_key_pair so that it can be passed to the
+        // MetricPushRuntime.
         let network_key_pair = config.network_key_pair.load()?;
         tracing::info!(
             walrus.node.network_key = %network_key_pair.as_ref().public(),
--- a/crates/walrus-service/src/common/utils.rs
+++ b/crates/walrus-service/src/common/utils.rs
@@ -297,7 +297,7 @@ impl MetricsAndLoggingRuntime {
 }
 
 /// A config struct to initialize the push metrics. Some binaries that depend on
-/// MetricsAndLoggingRuntime do not need nor is it appropriate to have push metrics.
+/// MetricPushRuntime do not need nor is it appropriate to have push metrics.
 #[derive(Debug)]
 pub struct EnableMetricsPush {
     /// token that is used to gracefully shut down the metrics push process


crates/walrus-service/bin/node.rs
```diff
@@ -357,8 +365,11 @@ mod commands {
             walrus.node.public_key = %config.protocol_key_pair.load()?.as_ref().public(),
             "walrus protocol public key",
         );
+        // // Load the network_key_pair so that it can be passed to the
+        // // MetricsAndLoggingRuntime.
```
mlegner: If you really want to have the separate runtimes for this PR, there are some comments and docstrings here and in `crates/walrus-service/src/common/utils.rs` that still refer to the `MetricsAndLoggingRuntime` and need to be cleaned up.

suiwombat: ah fix this thanks, the comment.

PR #1191, opened 2024-11-15
feat(staking): removes `New` state of the pool

damirka:
Removes the `New` state which was introduced at early stages of development and was never used.

### DIFF
--- a/crates/walrus-sui/src/types/move_structs.rs
+++ b/crates/walrus-sui/src/types/move_structs.rs
@@ -156,9 +156,7 @@ impl Display for StorageNodeCap {
 
 #[derive(Debug, PartialEq, Eq, Clone, Deserialize)]
 enum PoolState {
-    // The pool is new and awaits the stake to be added.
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -156,9 +156,7 @@ impl Display for StorageNodeCap {
 
 #[derive(Debug, PartialEq, Eq, Clone, Deserialize)]
 enum PoolState {
-    // The pool is new and awaits the stake to be added.
-    New,
-    // The pool is active and can accept stakes.
+    // The pool is active and can accept stakes. This is the default state.
```
karlwuest, line 159: Important: add the feature flag instead of just removing it:
```suggestion
    #[cfg(not(feature = "mainnet-contracts"))]
    // The pool is new and awaits the stake to be added.
    New,
    // The pool is active and can accept stakes. This is the default state.
```

PR #1196, opened 2024-11-18
feat(staking): no more `New` state of the pool

damirka:
replace: #1191 to avoid rebase conflicts

### DIFF
--- a/contracts/walrus/sources/staking/staking_pool.move
+++ b/contracts/walrus/sources/staking/staking_pool.move
@@ -30,8 +30,7 @@ const EPoolAlreadyWithdrawing: u64 = 5;
 const EPoolIsNotActive: u64 = 6;
 /// Trying to stake zero amount.
 const EZeroStake: u64 = 7;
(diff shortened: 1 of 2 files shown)


contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -151,18 +148,13 @@ public(package) fun new(
         EInvalidProofOfPossession,
     );
 
-    let (activation_epoch, state) = if (wctx.committee_selected()) {
-        (wctx.epoch() + 1, PoolState::New)
-    } else {
-        (wctx.epoch(), PoolState::Active)
-    };
-
+    let activation_epoch = if (wctx.committee_selected()) wctx.epoch() + 1 else wctx.epoch();
```
karlwuest, line 151: Nit: I think the readability is better with braces.

damirka: Sure! I'll address it in the next PR.

PR #1203, opened 2024-11-19
chore: exclude autogenerated file from yamlfmt

mlegner:


crates/walrus-service/client_config_example.yaml
```diff
@@ -1,8 +1,8 @@
 system_object: 0xa2637d13d171b278eadfa8a3fbe8379b5e471e1f3739092e5243da17fc8090eb
 staking_object: 0xca7cf321e47a1fc9bfd032abc31b253f5063521fd5b4c431f2cdd3fee1b4ec00
 exchange_object:
-  - 0xa9b00f69d3b033e7b64acff2672b54fbb7c31361954251e235395dea8bd6dcac
-  - 0x26a8a417b553b18d13027c23e8016c3466b81e7083225436b55143c127f3c0cb
+- 0xa9b00f69d3b033e7b64acff2672b54fbb7c31361954251e235395dea8bd6dcac
```
karlwuest, line 4: Question: can we make sure that this file is generated with the standard yaml formatting?

mlegner: Unfortunately I didn't find a simple way in my quick search. The `serde_yaml` crate does not support this. See, for example, https://github.com/dtolnay/serde-yaml/issues/337.

Issue #1037, opened 2024-10-11
Some improvements to the output of the `walrus` binary


mlegner:
- [x] Improve info output #979
- [X] Describe optimizations done for `store` in the store helptext
- [x] Nicer error if trying to store with a too high number of epochs (max number of epochs exceeded)
- [X] Better output for blob-status if blob is expired -> #1076
- [x] Tone down warnings to infos or debug if nodes are down
- [X] Clearer output for some optimizations like “was already stored for a sufficient number of epochs”
- [x] Correct singular/plural in outputs
- [X] Add warning when `--epochs` is not set for `store`.

Issue #979, opened 2024-10-08
Add additional information to `info` output


mlegner:
- epoch duration
- maximum number of epochs
- clarify FROST/WAL relationship
- add write price
- node IDs
- node stake
- next committee (if already selected)

PR #1204, opened 2024-11-19
feat(client): batch of minor client UX updates

giac-mysten:
Better reviewed commit by commit:

* c2124200 feat(client): generate-sui-wallet tries to save in the default directory - Closes: #1071
* 9f232417 feat(client): add -y flag to skip confirmation prompt during blob deletion - Closes: #1062 
* 7ffb056c feat(client): report if the delete action fully removes the blob - Closes: #1018 

The following contribute to: #1037 

* 06e9fc6b docs(client): document store optimizations in CLI help 
* a32698ff docs(client): clearer output for blob already stored 
* 4bc8a962 feat(client): show warning if --epochs is not passed explicitly 
* 89a2142f feat(client): improve error output when storing for too many epochs 
* a0f23789 refactor(client): tone down warnings for nodes not replying
* 6ad68ca9 fix(client): correct singular/plural in outputs

The following contribute to: #979 

* 7a6a8696 feat(client): add additional details to the info command
* 8f66e2d0 feat(client): add stake distribution to info command

Closes: #1037
Closes: #979

### DIFF
--- a/crates/walrus-service/src/client/cli/cli_output.rs
+++ b/crates/walrus-service/src/client/cli/cli_output.rs
@@ -38,7 +38,6 @@ use crate::client::{
         StorageNodeInfo,
         WalletOutput,
     },
-    string_prefix,
     BlobStoreResult,
 };
 
@@ -385,12 +384,9 @@ fn print_storage_node_table(n_shards: &NonZeroU16, storage_nodes: &[StorageNodeI
     table.set_titles(row![
         b->"Idx",
         b->"Name",
-        b->"Node ID",
         b->"# Shards",
         b->"Stake",
-        b->"Pk prefix",
         b->"Address",
-        b->"Shards"
     ]);
     for (i, node) in storage_nodes.iter().enumerate() {
         let n_owned = node.n_shards;
@@ -398,15 +394,15 @@ fn print_storage_node_table(n_shards: &NonZeroU16, storage_nodes: &[StorageNodeI
         table.add_row(row![
             bFg->format!("{i}"),
             node.name,
-            node.node_id,
-            format!("{} ({:.2}%)", n_owned, n_owned_percent),
-            HumanReadableFrost::from(node.stake),
-            string_prefix(&node.public_key),
+            r->format!("{} ({:.2}%)", n_owned, n_owned_percent),
+            r->HumanReadableFrost::from(node.stake),
             node.network_address,
-            DisplayShardList(&node.shard_ids),
         ]);
     }
     table.printstd();
+    for (i, node) in storage_nodes.iter().enumerate() {
+        print_storage_node_info(node, i, n_shards);
+    }
 }
 
 struct DisplayShardList<'a>(&'a [ShardIndex]);
@@ -588,4 +584,31 @@ fn blob_and_file_str(blob_id: &BlobId, file: &Option<PathBuf>) -> String {
     } else {
         format!("{}", blob_id)
     }
+}
+
+/// Print the full information of the storage node to stdoud.
+fn print_storage_node_info(node: &StorageNodeInfo, node_idx: usize, n_shards: &NonZeroU16) {
+    let n_owned = node.n_shards;
+    let n_owned_percent = (n_owned as f64) / (n_shards.get() as f64) * 100.0;
+    printdoc!(
+        "
+
+        {heading}
+        Owned shards: {n_owned} ({n_owned_percent:.2} %)
+        Total stake: {stake}
+        Node address: {network_address}
+        Node ID: {node_id}
+        Public key: {public_key}
+        Network public key: {network_public_key}
+        Owned shards:
+        {shards}
+        ",
+        heading = format!("{}: {}", node_idx, node.name).bold().magenta(),
+        stake = HumanReadableFrost::from(node.stake),
+        network_address = node.network_address,
+        node_id = node.node_id,
+        public_key = node.public_key,
+        network_public_key = node.network_public_key,
+        shards = DisplayShardList(&node.shard_ids),
+    );
 }


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -260,6 +275,9 @@ pub enum CliCommands {
         #[clap(flatten)]
         #[serde(flatten)]
         target: FileOrBlobIdOrObjectId,
+        /// Proceed to delete the blob without confirmation.
+        #[clap(short, long)]
```
karlwuest:
```suggestion
        #[clap(short, long, action)]
```

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -212,16 +217,24 @@ impl CliOutput for BlobStatusOutput {
...
+                        Expiry epoch: {end_epoch}\n\
                         Related event: {}\
                         {initial_certified_str}",
                     format_event_id(&status_event)
                 );
-                if count_deletable_certified > 0 {
-                    println!(
-                        "There are also {count_deletable_certified} certified deletable Blob \
+                match count_deletable_certified {
+                    2..=u32::MAX => {
```
karlwuest: Nit:
```suggestion
                    2.. => {
```

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -432,10 +481,63 @@ impl CliOutput for DeleteOutput {
...
+            }
+        }
+    }
+}
+
+fn removed_instance_string(blob_status: &BlobStatus) -> String {
+    const STILL_AVAILABLE: &str =
+        "The above blob objects were removed, but the blob still exist on Walrus.";
+    match blob_status {
+        BlobStatus::Nonexistent => "The blob was completely removed from Walrus.".to_owned(),
```
karlwuest: Suggestion: This may convey some notion of data privacy, i.e., that it is impossible for the blob to still be stored. Maybe we could add a warning that the data may still be publicly available?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -432,10 +481,63 @@ impl CliOutput for DeleteOutput {
...
+                "{} There are still one or more {} permanent instances {}available.",
+                STILL_AVAILABLE,
+                if *is_certified {
+                    "certified"
+                } else {
+                    "registered"
+                },
+                if deletable_counts.count_deletable_total > 0 {
+                    format!(
+                        "and deletable instances ({})",
```
karlwuest: Nit: I think this needs a space at the end
```suggestion
                        "and deletable instances ({}) ",
```

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -432,10 +481,63 @@ impl CliOutput for DeleteOutput {
...
+            ..
+        } => {
+            format!(
+                "{} There are still one or more {} permanent instances {}available.",
+                STILL_AVAILABLE,
+                if *is_certified {
+                    "certified"
+                } else {
+                    "registered"
+                },
```
karlwuest: Remark/Suggestion: I'm not a huge fan of this large nested format invocation, but I'm also not sure how to make it more readable. Maybe something like the following?
```suggestion
                is_certified.then_some("certified")
                    .unwrap_or("registered"),
```
Feel free to ignore this if you don't like it.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -456,7 +504,11 @@ impl ClientCommandRunner {
         BlobIdConversionOutput::from(blob_id_decimal).print_output(self.json)
     }
 
-    pub(crate) async fn delete(self, target: FileOrBlobIdOrObjectId) -> Result<()> {
+    pub(crate) async fn delete(
+        self,
+        target: FileOrBlobIdOrObjectId,
```
karlwuest, line 510: Question: I know that this is not part of this PR, but why is `FileOrBlobOrObjectId` not an Enum?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -716,3 +739,31 @@ mod tests {
...
+#[derive(Debug, Clone, Default)]
+pub enum UserConfirmation {
+    // The user needs to confirm the action.
+    #[default]
+    Required,
+    // The action can proceed without confirmation from the user.
+    Automatic,
+}
+
+impl UserConfirmation {
```
karlwuest: Nit: The name sounds as if this is represents the confirmation, but it represents whether the confirmation is required, i.e. there is kind of an "inversion" of the meaning of the CLI flag. Maybe a better name would be `AutoConfirmation` with e.g. variants `Provided` and `ConfirmationRequired`.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -716,3 +739,31 @@ mod tests {
...
+    /// Converts a boolean flag to an `ActionConfirmation`.
+    ///
+    /// If `yes` is `true`, the action is automatic; otherwise, the action requires confirmation.
+    pub(crate) fn from_yes_flag(yes: bool) -> Self {
+        if yes {
+            UserConfirmation::Automatic
+        } else {
+            UserConfirmation::Required
+        }
+    }
```
karlwuest: Question/Suggestion: Is there any reason not to implement `From<bool>` (except maybe the "inversion of meaning" mentioned above)? This might also allow to specify the clap argument directly as the enum type and still have it act as a flag.

crates/walrus-service/src/client/responses.rs
```diff
@@ -260,6 +265,52 @@ pub(crate) struct StorageNodeInfo {
     pub(crate) network_public_key: NetworkPublicKey,
     pub(crate) n_shards: usize,
     pub(crate) shard_ids: Vec<ShardIndex>,
+    pub(crate) object_id: ObjectID,
+    pub(crate) stake: u64,
+}
+
+impl StorageNodeInfo {
+    fn from_nodes_and_stake(value: StorageNode, stake: u64) -> Self {
```
karlwuest: Nit:
```suggestion
    fn from_node_and_stake(value: StorageNode, stake: u64) -> Self {
```

crates/walrus-service/src/client/responses.rs
```diff
@@ -260,6 +265,52 @@ pub(crate) struct StorageNodeInfo {
     pub(crate) network_public_key: NetworkPublicKey,
     pub(crate) n_shards: usize,
     pub(crate) shard_ids: Vec<ShardIndex>,
+    pub(crate) object_id: ObjectID,
```
karlwuest: Question: Is there a reason to call it `object_id` instead of `node_id`?

```suggestion
    pub(crate) node_id: ObjectID,
```

crates/walrus-service/src/client/responses.rs
```diff
@@ -260,6 +265,52 @@ pub(crate) struct StorageNodeInfo {
...
+    committee
+        .members()
+        .iter()
+        .cloned()
+        .map(|node| {
+            let stake = stake_assignment
+                .iter()
+                .find(|(object_id, _)| object_id == &node.node_id)
+                .map(|(_, stake)| *stake)
+                .expect("a node in the committee must have stake");
```
karlwuest: Suggestion: This iterates through the `stake_assignment` vector once for every node. Instead, you could create a hashmap from `stake_assignment` first, and then just read from that.

crates/walrus-service/src/client/responses.rs
```diff
@@ -326,7 +354,12 @@ impl InfoOutput {
         dev: bool,
     ) -> anyhow::Result<Self> {
         let committee = sui_read_client.current_committee().await?;
-        let price_per_unit_size = sui_read_client.storage_price_per_unit_size().await?;
+        let (storage_price_per_unit_size, write_price_per_unit_size) = sui_read_client
+            .storage_and_write_price_per_unit_size()
+            .await?;
+        let fixed_params = sui_read_client.fixed_system_parameters().await?;
+        let next_committee = sui_read_client.next_committee().await?;
+        let stake_assignment = sui_read_client.stake_assignment().await?;
```
karlwuest, line 363: Remark: We are now doing a lot of calls that then do RPC requests, sometimes for the same objects (e.g. both `next_committee` and `stake_assignment` will fetch the staking object). For the CLI, this probably doesn't matter much, but if we want to enable http APIs to get the info on e.g. publishers, we should start caching values to not do this calls for every request.
One option even now would be to expose the `get_staking_object` function and then expose functions `next_committee()` and `stake_assignment()` on `StakingObject` (and similar for the SystemObject), s.t. we just have one call to retrieve them within this function. 
Just to be clear: this is not needed for this PR, but you could maybe add a TODO.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -142,15 +142,30 @@ pub enum Commands {
...
+    /// `--force` flag, the store operation stops. Otherwise, the operation proceeds as follows:
+    ///
+    /// - check if the blob is registered (but not certified) in the current wallet for a sufficient
+    ///   duration, and if so reuse the registration to store the blob;
+    /// - otherwise, check if there is an appropriate storage resource (with sufficient space
+    ///   and for a sufficient duration) that can be used to register the blob; or
+    /// - if the above fails, it purchase a new storage resource and register the blob.
+    ///
+    /// If the `--force` flag is used, this operation always creates a new certification for the
+    /// blob possibly reusing storage resources or uncertified but registered blobs).
```
```suggestion
    /// blob (possibly reusing storage resources or uncertified but registered blobs).
```

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -149,6 +154,11 @@ pub trait ReadClient: Send + Sync {
     fn fixed_system_parameters(
         &self,
     ) -> impl Future<Output = SuiClientResult<FixedSystemParameters>> + Send;
+
+    /// Returns the mapping between object IDs and stake in the staking object.
```
mlegner: Question: I guess the "object IDs" are the node IDs (=object IDs of the staking pools)?

Also, why not return a `HashMap` here?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -340,24 +366,47 @@ impl CliOutput for InfoOutput {
...
+                "\n(dev) Next committee: Storage node details and shard distribution\n"
+                    .bold()
+                    .yellow()
+            );
+            print_storage_node_table(n_shards, storage_nodes);
+        };
     }
 }
 
+fn print_storage_node_table(n_shards: &NonZeroU16, storage_nodes: &[StorageNodeInfo]) {
```
mlegner, line 381: Suggestion: Can we also include the node IDs here? This would also address https://github.com/MystenLabs/walrus-docs/issues/167.

crates/walrus-service/src/client/cli/runner.rs
mlegner: Because this is not yet supported by clap, see https://github.com/clap-rs/clap/issues/2621.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -507,11 +561,24 @@ impl ClientCommandRunner {
...
+            // Wait to ensure that the information on the deletion is propagated.
+            tokio::time::sleep(Duration::from_secs(1)).await;
+            Some(
+                client
+                    .get_blob_status_with_retries(&deleted_blob_id, client.sui_client())
+                    .await?,
+            )
+        } else {
+            None
+        };
```
mlegner: Suggestion: Given that this check takes quite some extra time, can we add a flag to disable it?

crates/walrus-service/src/client/cli/cli_output.rs
giac-mysten: good point!

crates/walrus-service/src/client/cli/args.rs
giac-mysten: awesome, I did not know this could work!

crates/walrus-service/src/client/responses.rs
giac-mysten: Makes sense, I opened #1222 , though this should be a quick one.

crates/walrus-service/src/client/cli/cli_output.rs
giac-mysten: i had left it out because the table is now becoming huge. Will add it. I'm wondering however if it would be best to keep the table narrow, and add full per-node info in another way.

mlegner: Yes, that's a good point. The list of shard IDs also takes up a lot of space and is probably not relevant for most users. Also, the public key could potentially be removed as it isn't needed for any user action.

giac-mysten: what about the following? concise table and full info later
image

mlegner: I like it. ❤️ 

A few other ideas here (but can be done separately):
- The shard and stake columns could be right-aligned.
- The stake is maybe a bit unnecessarily precise. Also, the "WAL" could maybe be added to the title to make the column narrower.
- We could either split up the `info` command into smaller ones or allow users to control what is included with some options.

giac-mysten: will keep this latter one, if there are no objections :)

PR #1205, opened 2024-11-19
chore: add testnet contracts, use them if `mainnet-contracts` not set

karlwuest:
The contracts in `testnet-contracts` are taken exactly from the `testnet-v1.0.1` tag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7355,8 +7355,7 @@ dependencies = [
  "pin-project-lite",
  "signal-hook-registry",
  "socket2 0.5.7",
- "tokio-macros",
- "tracing",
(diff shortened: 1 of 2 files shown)


crates/walrus-orchestrator/src/protocol/target.rs
mlegner, line 1: Important: If we check for the `mainnet-contracts` feature in `walrus-orchestrator` and `walrus-service`, we also need to add it to those crates.

karlwuest: Good catch 🤦‍♂️

karlwuest: Added in 5743cd2d58a7bcfaf3018f9531b7da97acbc9865

PR #1206, opened 2024-11-19
feat: docker compose for local testbed

pei-mysten:
Adding docker compose setup for local testbed. Please refer to the README file for detailed instruction
All you need to start a local testbed is `docker compose up`, or `./build-local-image.sh` to use the locally built image

### DIFF
--- a/docker/local-testbed/README.md
+++ b/docker/local-testbed/README.md
@@ -11,7 +11,7 @@ This setup launches:
 
 Make sure you're under `walrus/docker/local-testbed`
 
-1. ./build-local-image.sh
+1. `./build-local-image.sh`
    1. only if you want to run containers built from local repo
    2. takes about 10 min on M2 Pro Macbook
    3. be sure assign 16GB+ to docker process for image building process


docker/local-testbed/files/run-walrus.sh
```diff
@@ -0,0 +1,52 @@
...
+mkdir -p /root/.config/walrus
+mkdir -p /opt/walrus/outputs
+
+# copy from deploy outputs
+cp /opt/walrus/outputs/${HOSTNAME}-sui.yaml /root/.sui/sui_config/client.yaml
+cp /opt/walrus/outputs/${HOSTNAME}.aliases /root/.sui/sui_config/sui.aliases
+
+# extract object IDs from the deploy outputs
+SYSTEM_OBJECT=$(grep "system_object" /opt/walrus/outputs/deploy | awk '{print $2}')
+STACKING_OBJECT=$(grep "staking_object" /opt/walrus/outputs/deploy | awk '{print $2}')
```
mlegner: Nit: This should probably be `STAKING_OBJECT`. Also below and in `docker/local-testbed/files/walrus-client-config.yaml`.

docker/local-testbed/README.md
```diff
@@ -0,0 +1,61 @@
+# Walrus Network Docker Compose Setup
+
+Tested on M2 MacOS 15.1 with Docker v27.3.1.
+
+This setup launches:
+
+- A Sui Network with 4 validators, 1 fullnode, and 1 faucet.
```
mlegner, line 7: Question: Would 2 validators also work, or is that more effort to set up?

docker/local-testbed/README.md
```diff
@@ -0,0 +1,61 @@
...
+- 4 Walrus Storage Nodes.
+
+## How to Run
+
+Make sure you're under `walrus/docker/local-testbed`
+
+1. ./build-local-image.sh
```
mlegner: Nit:
```suggestion
1. `./build-local-image.sh`
```

docker/local-testbed/README.md
```diff
@@ -0,0 +1,61 @@
...
+   1. only if you want to run containers built from local repo
```
mlegner, line 15: Question: What if I don't want that?

docker/local-testbed/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+   - Run `docker exec -it dryrun-node-0/1/2/3 bash`.
+   - Each Walrus node's configuration file is located at `/opt/walrus/outputs/dryrun-node-[0|1|2|3].yaml`.
+   - Each storage node contains Sui wallet information for all network participants (storage nodes and admin).
+
+3. To obtain more WAL tokens for testing:
+
+   - Shell into a container and run:
+     - `sui client faucet --url http://sui-localnet:9123/gas` to get more Sui tokens.
+     - `walrus get-wal <amount_in_mist>` (e.g., `500000000000` for 500 WAL).
```
mlegner, line 37: Question: How many SUI does the faucet return by default? If it's the same as for Testnet, you have to call the faucet many times until you reach 500 SUI.

docker/local-testbed/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+4. To store data on the Walrus Network:
+
+   - Shell into any storage node container and run (all of them are pre-configured with a `walrus` client):
+     - `walrus store <file>` to upload data.
+     - Similarly, use `walrus` commands to retrieve stored data.
+
+5. Optionally, you can expose port 9185 from any storage node container and interact with the REST API from you local development environment.
```
mlegner, line 45: Question: If we just expose the port of one node, we can only communicate with that one, right? In this case, interacting with Walrus from the host system would not actually work.

Would it be possible to set this up in such a way that all storage nodes are reachable from the host system? Interacting with Walrus from the host system would be much more convenient and should be the default if possible.

docker/local-testbed/build-local-image.sh
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) Mysten Labs, Inc.
+# SPDX-License-Identifier: Apache-2.0
+#!/bin/bash
+
+PLATFORM=linux/arm64
```
mlegner, line 5: Question: Does this still work on an x86 hardware?

docker/local-testbed/docker-compose.yaml
```diff
@@ -0,0 +1,161 @@
...
+      - walrus-deploy-outputs:/opt/walrus/outputs
+      - ./files/walrus-client-config.yaml:/root/walrus-client-config.yaml
+      - ./files/run-walrus.sh:/root/run-walrus.sh
+    command: >
+      /bin/bash -c "/root/run-walrus.sh"
+    logging:
+      driver: "json-file"
+      options:
+        max-file: "10"
+        max-size: "1g"
```
mlegner: Given the amount of overlap between the different nodes, it might be worth deduplicating code with [fragments](https://docs.docker.com/reference/compose-file/fragments/). I'm not an expert on these things but AFAIU the following should work:

```yaml
  walrus-node-0: &node
    depends_on:
      sui-localnet:
        condition: service_healthy
      walrus-deploy:
        condition: service_completed_successfully
    networks:
      testbed-network:
        ipv4_address: 10.0.0.10
    image: ${WALRUS_IMAGE_NAME:-mysten/walrus-service:9eeee2f17198ed8683967af8c8ed7d7c985398ef}
    platform: ${WALRUS_PLATFORM:-linux/amd64}
    hostname: dryrun-node-0
    container_name: dryrun-node-0
    environment:
      - NODE_NAME=dryrun-node-0
    volumes:
      - sui-bin:/root/sui_bin
      - walrus-deploy-outputs:/opt/walrus/outputs
      - ./files/walrus-client-config.yaml:/root/walrus-client-config.yaml
      - ./files/run-walrus.sh:/root/run-walrus.sh
    command: >
      /bin/bash -c "/root/run-walrus.sh"
    logging:
      driver: "json-file"
      options:
        max-file: "10"
        max-size: "1g"

  walrus-node-1:
    <<: *node
    networks:
      testbed-network:
        ipv4_address: 10.0.0.11
    hostname: dryrun-node-1
    container_name: dryrun-node-1
    environment:
      - NODE_NAME=dryrun-node-1
```

docker/local-testbed/files/deploy-walrus.sh
```diff
@@ -0,0 +1,24 @@
...
+cp -r walrus-docs/contracts /opt/walrus
+
+cd /opt/walrus
+
+rm -rf /opt/walrus/outputs/*
+
+/opt/walrus/bin/walrus-deploy deploy-system-contract \
+  --working-dir /opt/walrus/outputs \
+  --sui-network 'http://sui-localnet:9000;http://sui-localnet:9123/gas' \
+  --n-shards 10 \
```
mlegner: Suggestion: We should probably increase this a bit to make this at least slightly more realistic. Maybe 100 would be better.

docker/local-testbed/files/deploy-walrus.sh
```diff
@@ -0,0 +1,24 @@
...
+  --host-addresses 10.0.0.10 10.0.0.11 10.0.0.12 10.0.0.13 \
+  --storage-price 5 \
+  --write-price 1 \
+  --epoch-duration 1h >/opt/walrus/outputs/deploy
+
+/opt/walrus/bin/walrus-deploy generate-dry-run-configs \
+  --working-dir /opt/walrus/outputs \
+  --use-legacy-event-provider
```
mlegner: Ideally we would use the new checkpoint-based event provider. I don't know if the Sui fullnode has the REST API enabled though. Maybe @sadhansood knows more about this.

docker/local-testbed/README.md
mlegner: ~~Also, the client wallet should have 1M WAL at the beginning (see [here](https://github.com/MystenLabs/walrus/blob/a52970fbfc03f552d543420f4716ebef1e1b24a0/crates/walrus-service/src/testbed.rs#L409-L417)), this should be sufficient.~~

Ah, I see, you are using the Sui wallet from the node, not the client wallet.

docker/local-testbed/files/deploy-walrus.sh
```diff
@@ -0,0 +1,24 @@
...
+rm -rf /opt/walrus/outputs/*
+
+/opt/walrus/bin/walrus-deploy deploy-system-contract \
+  --working-dir /opt/walrus/outputs \
+  --sui-network 'http://sui-localnet:9000;http://sui-localnet:9123/gas' \
+  --n-shards 10 \
+  --host-addresses 10.0.0.10 10.0.0.11 10.0.0.12 10.0.0.13 \
+  --storage-price 5 \
+  --write-price 1 \
+  --epoch-duration 1h >/opt/walrus/outputs/deploy
```
mlegner: Question: Would it be possible to make this configurable when starting the network?

docker/local-testbed/files/run-walrus.sh
```diff
@@ -0,0 +1,52 @@
...
+# copy binaries
+cp /root/sui_bin/sui /usr/local/bin/
+cp /opt/walrus/bin/walrus /usr/local/bin/
+
+# copy walrus client config
+cp /root/walrus-client-config.yaml /root/.config/walrus/client_config.yaml
+# replace <SYSTEM_OBJECT> and <STACKING_OBJECT> with the actual object IDs from the deploy outputs
+sed -i "s/<SYSTEM_OBJECT>/${SYSTEM_OBJECT}/g" /root/.config/walrus/client_config.yaml
+sed -i "s/<STACKING_OBJECT>/${STACKING_OBJECT}/g" /root/.config/walrus/client_config.yaml
+sed -i "s/<EXCHANGE_OBJECT>/${EXCHANGE_OBJECT}/g" /root/.config/walrus/client_config.yaml
```
mlegner: Nit: How about this simpler version?
```suggestion
# write walrus client config
echo "system_object: $SYSTEM_OBJECT
staking_object: $STAKING_OBJECT
exchange_object: $EXCHANGE_OBJECT" > /root/.config/walrus/client_config.yaml
```

docker/local-testbed/README.md
pei-mysten: 2 would be great but `walrus-deploy` requires 4 nodes to start in the `--host-addresses` arg

docker/local-testbed/files/deploy-walrus.sh
pei-mysten: sounds good, will change!

docker/local-testbed/README.md
mlegner: I was referring to the Sui nodes, does the cluster there also require 4 nodes?

docker/local-testbed/files/deploy-walrus.sh
pei-mysten: good point, let me double check the config generated

docker/local-testbed/files/deploy-walrus.sh
pei-mysten: yes we can put them as an env var to the .env file, i'll change the script to support that

docker/local-testbed/files/run-walrus.sh
pei-mysten: ha you're 100% right. this was an legacy from my previous attempts and forgot we ca have a way simpler method to tackle it

docker/local-testbed/README.md
pei-mysten: it will download and use the default image version: 
```
image: ${WALRUS_IMAGE_NAME:-mysten/walrus-service:9eeee2f17198ed8683967af8c8ed7d7c985398ef}
    platform: ${WALRUS_PLATFORM:-linux/amd64}
```

mlegner: Ah, I see. Would this work on MacOS?

docker/local-testbed/README.md
pei-mysten: oh yes same for Sui, 4 nodes are needed, good thing is that they're all packed in one container and from what i see the resource usage is pretty mild, should be good!

docker/local-testbed/README.md
pei-mysten: we'll have to expose several different ports for different containers, as the destination is our local machine, we cannot reuse the same port again and again.
for example:
node-0 9185 -> local 9185
node-1 9185 -> local 9186
etc.

docker/local-testbed/build-local-image.sh
pei-mysten: yes, i built some public images with x86. reason for using arm64 here is cross compiling on mac is super slow.. took me more than 1 hour to finish the build

docker/local-testbed/README.md
pei-mysten: it will! as long as rosetta is enabled

docker/local-testbed/files/deploy-walrus.sh
pei-mysten: yes REST API is enabled, let me remove this

docker/local-testbed/docker-compose.yaml
pei-mysten: thanks! refactored accordingly

docker/local-testbed/README.md
mlegner: Hmm... But this won't work with the client on the host as there is a mismatch between the host addresses on chain and the addresses through which they can be reached. Is there no way to connect at `10.0.0.10:9185` etc.?

PR #1208, opened 2024-11-19
fix: Make event processor handle package upgrades correctly

sadhansood:
This PR addresses the problem with checkpoint event processor not being able to handle events emitted by different versions of walrus package. The testing is a bit convoluted but is robust and convincing that this change works.

Testing steps:

1. Made a simple change in a different branch which added a contract function called `initiate_epoch_change_dummy` which only emits the epoch change event without doing anything meaningful. This can be viewed here https://github.com/MystenLabs/walrus/pull/1207/files
2. Started the 4 node cluster with ` scripts/local-testbed.sh  `
3. Upgraded the contract and got a new package id
4. Switched to this current branch (with event processor fixed), and killed one of the four nodes. Started that node again but now in this current branch
5. Now, 3/4 nodes are running old event processor (which can only read events from version 1 of the contract) and 1/4 node is running the updated code (which can read events from all versions of the contract)
6. Invoked the contract function `initiate_epoch_change_dummy` on the new package we published which emitted an epoch change start event
7. Observed that 3/4 node did not receive the event from new package (nothing printed in their logs), but the 1/4 node running new code printed this in the log: 
```
{"timestamp":"2024-11-19T23:47:23.285471Z","level":"INFO","fields":{"message":"EpochChangeStart event received: EpochChangeStart { epoch: 2, event_id: EventID { tx_digest: TransactionDigest(YXKWQYEH9xu2FzahTQB32paPMukN9EKSLEbuaSTrybw), event_seq: 0 } }. Index: 2667"},"target":"walrus_service::node","filename":"crates/walrus-service/src/node.rs","line_number":699,"span":{"name":"process_epoch_change_event"},"spans":[{"messaging.client.id":"olHhxDLuzcMgAmTwiRoaRY4gLdc6jt7ommIxtkBa7uJjdTC6HUakTgkdNM1c9XYB","messaging.destination.name":"blob_store","messaging.operation.type":"receive","messaging.system":"sui","otel.kind":"CONSUMER","walrus.blob_id":"None","walrus.event.checkpoint_seq":"531824","walrus.event.index":2667,"walrus.event.kind":"epoch-change-start","walrus.event.tx_digest":"Some(TransactionDigest(YXKWQYEH9xu2FzahTQB32paPMukN9EKSLEbuaSTrybw))","walrus.node_status":"Active","name":"blob_store receive"},{"name":"process_event"},{"name":"process_epoch_change_event"}]}
{"timestamp":"2024-11-19T23:47:23.793736Z","level":"ERROR","fields":{"message":"failed to initiate a transition to the new epoch","error":"LookupError(unexpected total number of shards in committee members)"},"target":"walrus_service::node","filename":"crates/walrus-service/src/node.rs","line_number":979,"span":{"name":"begin_committee_change"},"spans":[{"messaging.client.id":"olHhxDLuzcMgAmTwiRoaRY4gLdc6jt7ommIxtkBa7uJjdTC6HUakTgkdNM1c9XYB","messaging.destination.name":"blob_store","messaging.operation.type":"receive","messaging.system":"sui","otel.kind":"CONSUMER","walrus.blob_id":"None","walrus.event.checkpoint_seq":"531824","walrus.event.index":2667,"walrus.event.kind":"epoch-change-start","walrus.event.tx_digest":"Some(TransactionDigest(YXKWQYEH9xu2FzahTQB32paPMukN9EKSLEbuaSTrybw))","walrus.node_status":"Active","name":"blob_store receive"},{"name":"process_event"},{"name":"process_epoch_change_event"},{"name":"process_epoch_change_start_event"},{"name":"begin_committee_change"}]}
{"timestamp":"2024-11-19T23:47:23.793870Z","level":"ERROR","fields":{"message":"storage node exited with an error","error":"unexpected total number of shards in committee members"},"target":"walrus_node","filename":"crates/walrus-service/bin/node.rs","line_number":869}
```
This confirms that new code is successful in parsing events from different contract versions.

### DIFF
--- a/crates/walrus-service/src/node/events/event_processor.rs
+++ b/crates/walrus-service/src/node/events/event_processor.rs
@@ -3,10 +3,11 @@
 
 //! Event processor for processing events from the full node.
 
-use std::{fmt::Debug, path::Path, sync::Arc, time::Duration};
+use std::{collections::HashMap, fmt::Debug, path::Path, sync::Arc, time::Duration};
 
 use anyhow::{anyhow, bail, Result};
 use async_trait::async_trait;
+use futures_util::future::try_join_all;
 use move_core_types::{
     account_address::AccountAddress,
     annotated_value::{MoveDatatypeLayout, MoveTypeLayout},
@@ -36,10 +37,14 @@ use sui_types::{
     full_checkpoint_content::CheckpointData,
     message_envelope::Message,
     messages_checkpoint::{TrustedCheckpoint, VerifiedCheckpoint},
-    object::Object,
+    object::{Data, Object},
     SYSTEM_PACKAGE_ADDRESSES,
 };
-use tokio::{select, sync::Mutex, time::sleep};
+use tokio::{
+    select,
+    sync::{Mutex, RwLock},
+    time::sleep,
+};
 use tokio_util::sync::CancellationToken;
 use typed_store::{
     rocks,
@@ -84,6 +89,8 @@ pub struct LocalDBPackageStore {
     package_store_table: DBMap<ObjectID, Object>,
     /// The full node REST client.
     fallback_client: Client,
+    /// Cache for original package ids.
+    original_id_cache: Arc<RwLock<HashMap<AccountAddress, ObjectID>>>,
 }
 
 /// Metrics for the event processor.
@@ -122,6 +129,8 @@ pub struct EventProcessor {
     pub metrics: EventProcessorMetrics,
     /// Pipelined checkpoint downloader.
     pub checkpoint_downloader: ParallelCheckpointDownloader,
+    /// Local package store.
+    pub package_store: LocalDBPackageStore,
 }
 
 impl Debug for LocalDBPackageStore {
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -201,12 +204,13 @@ impl EventProcessor {
...
-                    &self.walrus_package_store,
-                    std::iter::once((object.id(), object)),
-                )?;
+            if !object.is_package() {
+                continue;
             }
+            write_batch.insert_batch(
+                &self.walrus_package_store,
+                std::iter::once((object.id(), object)),
+            )?;
```
karlwuest, line 219: Question: What's the reason for this change? This seems to be doing exactly the same, just with an inverted condition.

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -347,12 +351,26 @@ impl EventProcessor {
...
                 let tx_events = tx.events.unwrap_or_default();
+                let package_ids: Vec<_> = tx_events
+                    .data
+                    .iter()
+                    .map(|event| event.package_id)
+                    .collect();
+                let original_package_ids = try_join_all(
+                    package_ids
+                        .iter()
+                        .map(|&id| self.package_store.get_original_package_id(id.into())),
```
karlwuest: Question: How expensive is resolving all original package ids for all events in a checkpoint? 
An alternative could be to store a hashset with all package IDs for our package and then just check if `event.package_id` is contained in that set.

Remark: At the moment, it would even work to just check the package from the event type, but I really like your approach because it also works for events defined in future package versions!

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -347,12 +351,26 @@ impl EventProcessor {
...
+                )
+                .await?;
```
mlegner: Suggestion: Instead of collecting the package IDs in a vector and then immediately consuming it, how about this shorter version?
```suggestion
                let original_package_ids: Vec<_> =
                    try_join_all(tx_events.data.iter().map(|event| {
                        self.package_store
                            .get_original_package_id(event.package_id.into())
                    }))
                    .await?;
```

PR #1210, opened 2024-11-20
fix: panics in blob sync tasks should be propagated to the node runtime

halfprice:
Currently it is silently dropped due to that the sync_handle's results are never checked.

### DIFF
--- a/crates/walrus-simtest/tests/simtest.rs
+++ b/crates/walrus-simtest/tests/simtest.rs
@@ -546,15 +546,15 @@ mod tests {
     #[ignore = "ignore E2E tests by default"]
     #[walrus_simtest(config = "latency_config()")]
     async fn test_repeated_shard_move_with_workload() {
-        // We use a very short epoch duration of 10 seconds so that we can exercise more epoch
+        // We use a very short epoch duration of 60 seconds so that we can exercise more epoch
         // changes in the test.
         let (_sui_cluster, walrus_cluster, client) =
             test_cluster::default_setup_with_epoch_duration_generic::<SimStorageNodeHandle>(
-                Duration::from_secs(60),
+                Duration::from_secs(30),
                 &[1, 2, 3, 3, 4],
                 true,
                 ClientCommunicationConfig::default_for_test_with_reqwest_timeout(
-                    Duration::from_secs(5),
+                    Duration::from_secs(2),
                 ),
             )
             .await
@@ -589,7 +589,7 @@ mod tests {
                 .await
                 .expect("stake with node pool should not fail");
 
-            tokio::time::sleep(Duration::from_secs(120)).await;
+            tokio::time::sleep(Duration::from_secs(70)).await;
         }
 
         workload_handle.abort();


crates/walrus-service/src/node.rs
```diff
@@ -2617,6 +2628,44 @@ mod tests {
...
+            .delete_shard()
+            .unwrap();
+
+        // Start a sync to trigger the blob sync task.
+        cluster.nodes[0]
+            .storage_node
+            .blob_sync_handler
+            .start_sync(*blob.blob_id(), 1, None, Instant::now())
+            .await
+            .unwrap();
```
mlegner: I don't fully understand why this causes a panic: After deleting the shard, the blob sync should not even try to recover the slivers for the no-longer-existing shard. 🤔

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -79,6 +80,83 @@ impl BlobSyncHandler {
...
+                    // Collect all finished in-progress syncs;
+                    let mut syncs = blob_syncs.lock().expect("should be able to acquire lock");
+                    let mut completed = Vec::new();
+
+                    // Collect handles to be awaited
+                    for (blob_id, handle) in syncs.iter_mut() {
+                        if let Some(sync_handle) = &handle.blob_sync_handle {
+                            if sync_handle.is_finished() {
+                                tracing::info!(
+                                    blob_id = %blob_id,
```
mlegner: Nit: This could be simplified (also in other places):
```suggestion
                                    %blob_id,
```

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -79,6 +80,83 @@ impl BlobSyncHandler {
...
+                    for blob_id in completed {
+                        syncs.remove(&blob_id);
+                    }
+                }
+
+                // Now await the handles
+                for (blob_id, handle) in handles {
+                    match handle.await {
+                        Ok(Ok(_)) => {
+                            // Normal completion. This should not happen.
```
mlegner, line 123: Can you add some more info on why this should not happen? Maybe something like "Normal completion. This should not happen because if the handle is finished, the `BlobSyncHandler::sync` must have completed, in which case the sync should have been removed from the `blob_syncs_in_progress`."

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -79,6 +80,83 @@ impl BlobSyncHandler {
...
+                        }
+                        Err(e) => {
+                            if e.is_panic() {
+                                std::panic::resume_unwind(e.into_panic());
+                            }
+                            tracing::warn!(
+                                blob_id = %blob_id,
+                                error = %e,
+                                "Blob sync task panicked"
+                            );
```
mlegner: Question: Do we have to change the order here? Also, is there a scenario where this is not due to a panic?

Also, in this case we could use the Debug print for the error to get additional info.
```suggestion
                            if e.is_panic() {
                                tracing::error!(
                                    %blob_id,
                                    error = ?e,
                                    "Blob sync task panicked"
                                );
                                std::panic::resume_unwind(e.into_panic());
                            }
```

crates/walrus-service/src/node.rs
halfprice: This only deletes the RocksDB, but not motify the shards HashMap. So techinically, the node still thinks it has the shard. Just that any RocksDB op will panic.

Added more documentation about the test, and renamed delete_shard() to delete_shard_storage().

crates/walrus-service/src/node/blob_sync.rs
halfprice: Done. PTAL

crates/walrus-service/src/node/blob_sync.rs
halfprice: I moved the logging before the is_panic() check. I think cancellation may also return here.

Good idea for the debug print.

PR #1211, opened 2024-11-20
feat: implement transaction builder for Walrus PTBs

### LINKED ISSUE #1135: Make sui bindings more PTB friendly
Refactor the sui bindings to allow building ptbs with multiple operations. This will make it easier to merge coins before registering ablob and allows registering or certifying multiple blobs in a single transaction and more customizability in the future.

karlwuest:
closes #1135

### DIFF
--- a/crates/walrus-sui/src/client/transaction_builder.rs
+++ b/crates/walrus-sui/src/client/transaction_builder.rs
@@ -254,7 +254,7 @@ impl WalrusPtbBuilder {
         Ok(result_arg)
     }
 
-    /// Adds a call to `certify_blob` to the `pt_builder` and returns the result [`Argument`].
+    /// Adds a call to `certify_blob` to the `pt_builder`.
     pub async fn certify_blob(
         &mut self,
         blob_object: ArgumentOrOwnedObject,


crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -518,3 +518,30 @@ impl Display for StakedWal {
         writeln!(f, "    activation_epoch: {}", self.activation_epoch)
     }
 }
+
+/// Sui type for an exchange rate: `wal` WAL = `sui` SUI.
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct ExchangeRate {
+    wal: u64,
+    sui: u64,
+}
```
giac-mysten, line 527: Nit:
```suggestion
}

```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! A transaction builder for Walrus PTBs.
```
giac-mysten: Nit:
```suggestion
//! A transaction builder for PTBs containing Walrus-related contract calls.
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    }
+}
+
+impl From<&Argument> for ArgumentOrOwnedObject {
+    fn from(arg: &Argument) -> Self {
+        Self::Argument(*arg)
+    }
+}
+
+impl From<ObjectID> for ArgumentOrOwnedObject {
```
giac-mysten, line 71: Question: no implementation for the `&ObjectID` version?

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            used_wal_coins: BTreeSet::new(),
+            wal_coin_arg: None,
+            sender_address,
+            args_to_consume: HashSet::new(),
+        }
+    }
+
+    /// Fills up the WAL coin argument of the PTB to at least `min_balance` by merging additional
+    /// coins if necessary. This is a no-op if the current available balance that has already been
+    /// added to the PTB (and hasn't been consumed yet) is larger than `min_balance`.
```
giac-mysten: Nit
```suggestion
/// Fills up the WAL coin argument of the PTB to at least `min_balance`.
    ///
    /// This function merges additional coins if necessary, and it is a no-op if the current
    /// available balance---that has been added to the PTB and hasn't been consumed yet---is larger
    /// than `min_balance`.
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+        } else {
+            let coin = coins
+                .pop()
+                .ok_or_else(|| SuiClientError::NoCompatibleWalCoins)?;
+            added_balance += coin.balance;
+            let coin_arg = self.pt_builder.input(coin.object_ref().into())?;
+            self.wal_coin_arg = Some(coin_arg);
+            coin_arg
+        };
+        if !coins.is_empty() {
```
giac-mysten, line 160: Important/question: consider the following scenario

- `additional_balance > 0`
- `coins.is_empty() == true`
- `coin_arg.is_some() == true`

~~then, IIUC, the result will be that the function will return `Ok(())`, but `added_balance == 0`, so the `tx_wal_balance` won't change.~~

~~from the function description, it seems to me that it should instead return `NoCompatibleWalCoins`. Is this true?~~

giac-mysten: Scratch the above, this is returned by `get_coins_with_total_balance`, but the documentation is outdated.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -273,64 +269,34 @@ impl SuiReadClient {
...
-                cursor,
-                None,
-            )
-        })
-        .await
-    }
-
     /// Returns a vector of coins of provided `coin_type` whose total balance is at least `balance`.
     ///
     /// Returns `None` if no coins of sufficient total balance are found.
```
giac-mysten: Remark: the documentation is outdated.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            return Err(SuiClientError::Internal(anyhow::anyhow!(
+                "trying to reduce WAL balance below 0"
+            )));
+        }
+        self.tx_wal_balance -= amount;
+        Ok(())
+    }
+
+    /// Adds a move call to the PTB.
+    /// Always returns an [`Argument::Result`] if no error is returned.
```
giac-mysten:
```suggestion
    /// Adds a move call to the PTB.
    ///
    /// Always returns an [`Argument::Result`] if no error is returned.
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    pub fn move_call(
```
giac-mysten: Question: should this be `pub`? in principle, the transaction that is added may use WAL and break the accounting of `tx_wal_balance`?

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    /// Adds a call to `reserve_space` to the `pt_builder` and returns the result Argument.
+    pub async fn reserve_space(
+        &mut self,
+        encoded_size: u64,
+        epochs_ahead: EpochCount,
+    ) -> SuiClientResult<Argument> {
+        let price = self
+            .storage_price_for_encoded_length(encoded_size, epochs_ahead)
+            .await?;
+        self.fill_wal_balance(price).await?;
```
giac-mysten, line 211: Question/suggestion: in the worst case, `fill_wal_balance` could be added for every new funciton called on the walrus ptb.

would it make sense to keep a running amount of the `tx_wal_balance`, but only call the `fill_wal_balance` on finish? IIUC, this would require quite some reworking, as the coin merging needs to happen before all other transactions 😬 so unclear if this is feasible.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            self.wal_coin_arg()?,
+        ];
+        let result_arg = self.move_call(contracts::system::register_blob, register_arguments)?;
+        self.reduce_wal_balance(price)?;
+        self.mark_arg_as_consumed(&storage_resource_arg);
+        self.add_result_to_be_consumed(result_arg);
+        Ok(result_arg)
+    }
+
+    /// Adds a call to `certify_blob` to the `pt_builder` and returns the result Argument.
```
giac-mysten: Remark: this function does not return the result Argument. 
```suggestion
    /// Adds a call to `certify_blob` to the `pt_builder`.
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            self.pt_builder.pure(epochs_ahead)?,
+            self.wal_coin_arg()?,
+        ];
+        let result_arg = self.move_call(contracts::system::reserve_space, reserve_arguments)?;
+        self.reduce_wal_balance(price)?;
+        self.add_result_to_be_consumed(result_arg);
+        Ok(result_arg)
+    }
+
+    /// Adds a call to `register_blob` to the `pt_builder` and returns the result Argument.
```
giac-mysten:
```suggestion
    /// Adds a call to `register_blob` to the `pt_builder` and returns the result [`Argument`].
```

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -194,30 +193,29 @@ impl SuiReadClient {
...
-        ))
+        Ok(ObjectArg::SharedObject {
+            id: object_id,
+            initial_shared_version,
+            mutable,
+        })
     }
 
-    pub(crate) async fn call_arg_from_system_obj(&self, mutable: bool) -> SuiClientResult<CallArg> {
+    pub(crate) async fn object_arg_from_system_obj(
```
giac-mysten: Nit?
```suggestion
    pub(crate) async fn object_arg_for_system_obj(
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+    ) -> SuiClientResult<Argument> {
+        let price = self
+            .write_price_for_encoded_length(blob_metadata.encoded_size)
+            .await?;
+        self.fill_wal_balance(price).await?;
+
+        let storage_resource_arg = self.argument_from_arg_or_obj(storage_resource).await?;
+
+        let register_arguments = vec![
+            self.system_arg(true).await?,
```
giac-mysten: Nit: would be nice to strongly type this flag
```suggestion
            self.system_arg(Mutability::Mutable).await?,
```
or something similar

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            self.argument_from_arg_or_obj(blob_object).await?,
+            self.pt_builder.pure(certificate.signature.as_bytes())?,
+            self.pt_builder.pure(&signers)?,
+            self.pt_builder.pure(&certificate.serialized_message)?,
+        ];
+        self.move_call(contracts::system::certify_blob, certify_args)?;
+        Ok(())
+    }
+
+    /// Adds a call to `delete_blob` to the `pt_builder` and returns the result Argument.
```
giac-mysten:
```suggestion
    /// Adds a call to `delete_blob` to the `pt_builder` and returns the result [`Argument`].
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            args.push(self.argument_from_arg_or_obj(arg_or_obj).await?);
+        }
+        args.iter().for_each(|arg| self.mark_arg_as_consumed(arg));
+        self.pt_builder
+            .transfer_args(recipient.unwrap_or(self.sender_address), args);
+        Ok(())
+    }
+
+    /// Transfers all outputs that have not been consumed yet by another command in the PTB.
+    /// If the recipient is `None`, the sender address is used.
```
giac-mysten:
```suggestion
    /// Transfers all outputs that have not been consumed yet by another command in the PTB.
    ///
    /// If the recipient is `None`, the sender address is used.
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            self.staking_arg(true).await?,
+            self.argument_from_arg_or_obj(storage_node_cap).await?,
+            self.pt_builder.pure(epoch)?,
+            self.pt_builder.obj(CLOCK_OBJECT_ARG)?,
+        ];
+        self.move_call(contracts::staking::epoch_sync_done, args)?;
+        Ok(())
+    }
+
+    /// Adds call to initiate epoch change to the PTB.
```
giac-mysten:
```suggestion
    /// Adds a call to initiate epoch change to the PTB.
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+            .pt_builder
+            .command(Command::SplitCoins(wal_coin_arg, vec![amount_arg]));
+        self.transfer(Some(recipient), vec![split_coin.into()])
+            .await?;
+        self.reduce_wal_balance(amount)?;
+        Ok(())
+    }
+
+    /// Transfers all remaining outputs and returns the PTB and the additional balance needed in the
+    /// gas coin to cover the SUI cost of the transaction.
```
giac-mysten: Nit: maybe make explicit that this does not cover gas costs? (only the exchange cost rn iiuc)

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -228,18 +226,16 @@ impl SuiReadClient {
         Ok(*initial_shared_version)
     }
 
-    pub(crate) async fn call_arg_from_staking_obj(
+    pub(crate) async fn object_arg_from_staking_obj(
```
giac-mysten: Nit?
```suggestion
    pub(crate) async fn object_arg_for_staking_obj(
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -0,0 +1,543 @@
...
+
+    fn mark_arg_as_consumed(&mut self, arg: &Argument) {
+        self.args_to_consume.remove(arg);
+    }
+
+    fn add_result_to_be_consumed(&mut self, arg: Argument) {
+        self.args_to_consume.insert(arg);
+    }
+
+    // all the functions
```
giac-mysten: Nit?
```suggestion
```

crates/walrus-sui/src/client.rs
```diff
@@ -597,23 +481,14 @@ impl ContractClient for SuiContractClient {
...
-            self.wallet_address,
-            Argument::NestedResult(reserve_result_index, 0),
-        );
-
-        let res = self
-            .sign_and_send_ptb(&wallet, pt_builder.finish(), None)
-            .await?;
+        let mut pt_builder = WalrusPtbBuilder::new(self.read_client.clone(), self.wallet_address);
+        pt_builder.reserve_space(encoded_size, epochs_ahead).await?;
+        let (ptb, _sui_cost) = pt_builder.finish().await?;
```
giac-mysten, line 491: Nit: I think we tried to keep unused arguments consistently `_`, but up to your preference (here and in the following)
```suggestion
        let (ptb, _) = pt_builder.finish().await?;
```

crates/walrus-sui/src/client.rs
```diff
@@ -673,65 +531,30 @@ impl ContractClient for SuiContractClient {
...
-        for i in 0..contracts::system::register_blob.n_object_outputs {
-            pt_builder.transfer_arg(
-                self.wallet_address,
-                Argument::NestedResult(register_result_index, i),
-            );
-        }
-
-        let res = self
-            .sign_and_send_ptb(&wallet, pt_builder.finish(), None)
+        // Blob is transferred automatically in the call to `finish`
```
giac-mysten:
```suggestion
        // Blob is transferred automatically in the call to `finish`.
```

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: ObjectID is `copy`, so we usually don't use it as a reference, but I can add the implementation for this.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: Good catch, I'll update the documentation for `get_coins_with_total_balance`.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: Yeah, I thought about this. I think eventually we want to allow adding arbitrary commands, but for now I'll restrict it and we can open it up once we need it (perhaps with additional safeguards). Once we allow adding additional commands, we anyway don't have any guarantees that the transactions are well-formed.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: I looked into this when I wrote it and this would require either buffering all other calls to then apply them at the end, or to more fundamentally change/replace the `ProgrammableTransactionBuilder` instead of using it internally, because the commands there are applied in the order that they are added and it's not possible to edit or replace a previous one.
I decided that neither of these are worth the effort, because the common case will be that there are just a few coins in which case the overhead for the merge is not very significant.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: The `certify` function does not return a result because it only mutates the `Blob` object.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: That's what the "additional" is trying to convey 😅 I'll make it more explicit.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: Ah, that was a todo comment for me, must have missed it at the end, because I didn't actually add the word "TODO" 🤦‍♂️

crates/walrus-sui/src/client.rs
karlwuest: I think we use both, depending on the situation. Here I like having the name, because it makes it clear what we ignore.

crates/walrus-sui/src/client/read_client.rs
karlwuest: I realized that the function actually returns the wrong error (in some cases), since we use it for SUI and WAL, but always return `NoCompatibleWalCoins`. I've now updated this to take an enum as argument that restricts the coin type to SUI and WAL and returns the correct error.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: Ah, I completely misread your comment and in my head swapped around the diff of the suggestion 🤦‍♂️

crates/walrus-sui/src/client.rs
mlegner: Additional remark: It actually makes a difference in some cases as the objects are dropped directly if assigned to `_`. (Doesn't matter here though.)

crates/walrus-sui/src/client/transaction_builder.rs
giac-mysten: yes, this makes a lot of sense. Even though I imagined this was infeasible, I though to mention it in case you had other options in mind.

PR #1212, opened 2024-11-20
fix(logs): canonicalize log entries and reduce overall verbosity

### LINKED ISSUE #1086: Revisit log levels
We should make sure we have consistent log levels everywhere so that we can globally set log levels as follows:
- INFO: Normal operation. Sufficient output to be able to follow the storage-node or client process and get notified of issues.
- DEBUG: Additional help with low-level issues but not too verbose such that it can be run for ~days.
- TRACE: Any additional messages that would otherwise pollute the `DEBUG` log.

mlegner:
Among others, this includes the following changes:
- reduce verbosity for some log entries, notably for successful requests to the store sliver endpoint
- increase level of some epoch-related log entries to INFO
- consistently print Debug formatting for errors on ERROR level
- use consistent capitalization
- prefer recording metadata as fields as opposed to including them in the log string
- qualify tracing macros with `tracing::` prefix
- simplify syntax for including fields in log entries

Closes #1086

### DIFF
--- a/crates/walrus-service/src/client/cli/runner.rs
+++ b/crates/walrus-service/src/client/cli/runner.rs
@@ -13,10 +13,12 @@ use std::{
 use anyhow::{Context, Result};
 use prometheus::Registry;
 use rand::seq::SliceRandom;
+use sui_config::{sui_config_dir, SUI_CLIENT_CONFIG};
 use sui_sdk::wallet_context::WalletContext;
 use sui_types::base_types::ObjectID;
 use walrus_core::{
     encoding::{encoded_blob_length_for_n_shards, EncodingConfig, Primary},
+    ensure,
     BlobId,
     EpochCount,
 };
@@ -33,16 +35,19 @@ use super::args::{
     FileOrBlobIdOrObjectId,
     PublisherArgs,
     RpcArg,
+    UserConfirmation,
 };
 use crate::{
     client::{
         cli::{
+            self,
             get_contract_client,
             get_read_client,
             get_sui_read_client_from_rpc_node_or_wallet,
             load_configuration,
             read_blob_from_file,
             success,
+            warning,
             BlobIdDecimal,
             CliOutput,
             HumanReadableMist,
@@ -158,7 +163,11 @@ impl ClientCommandRunner {
 
             CliCommands::ListBlobs { include_expired } => self.list_blobs(include_expired).await,
 
-            CliCommands::Delete { target } => self.delete(target).await,
+            CliCommands::Delete {
+                target,
+                yes,
+                no_status_check,
+            } => self.delete(target, yes.into(), no_status_check).await,
 
             CliCommands::Stake { node_id, amount } => {
                 self.stake_with_node_pool(node_id, amount).await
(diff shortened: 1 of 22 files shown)


crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -268,7 +268,7 @@ impl ClientCommandRunner {
         let client = get_contract_client(self.config?, self.wallet, self.gas_budget, &None).await?;
 
         if dry_run {
-            tracing::info!("Performing dry-run store for file {}", file.display());
+            tracing::info!("performing dry-run store for file '{}'", file.display());
```
giac-mysten, line 320: General question: where do we use string formatting, and where use metadata fields instead? is this because it's just the client startup runner?

crates/walrus-service/src/client/refill.rs
```diff
@@ -113,7 +113,7 @@ impl WalletCoinRefill {
     }
 
     async fn send_gas(&self, address: SuiAddress) -> Result<()> {
-        tracing::debug!("Sending gas to address: {:?}", &address);
+        tracing::debug!("sending gas to {address}");
```
giac-mysten:
```suggestion
        tracing::debug!(address, "sending gas to client");
```

crates/walrus-service/src/client/refill.rs
```diff
@@ -131,7 +131,7 @@ impl WalletCoinRefill {
     }
 
     async fn send_wal(&self, address: SuiAddress) -> Result<()> {
-        tracing::debug!("Sending wal to address: {:?}", &address);
+        tracing::debug!("sending WAL to {address}");
```
giac-mysten: Nit:
```suggestion
        tracing::debug!(address, "sending WAL to client");
```

crates/walrus-service/src/node.rs
```diff
@@ -592,7 +592,7 @@ impl StorageNode {
                         && event.event_epoch() + 1 < epoch_at_start
                     {
                         tracing::warn!(
-                            "the current epoch ({}) is far ahead of the event epoch: {};
+                            "the current epoch ({}) is far ahead of the event epoch ({}); \
                             node entering recovery mode",
                             epoch_at_start,
                             event.event_epoch()
```
giac-mysten, line 607:
```suggestion
                            current_epoch = epoch_at_start,
                            event_epoch = event.event_epoch()
                            "the current epoch is far ahead of the event epoch; \
                            node entering recovery mode",

```

crates/walrus-service/src/client/cli/runner.rs
mlegner: I think I used/kept string formatting in ~~two~~ three cases:
1. When the logs are mainly intended for end users as in this case. These cases should actually be replaced by the mechanisms added in #1214.
2. When the formatted variables are (somewhat) necessary to create a useful message in the first place.
3. In tests where logs are mainly manually inspected and not filtered through tools.

I could rebase this PR on #1214 and improve the first category. WDYT?

crates/walrus-service/src/client/refill.rs
mlegner: Done in 15229c1989a12e7b8e4b84180e0a007b5505fde9.

crates/walrus-service/src/client/refill.rs
mlegner: Done in 15229c1989a12e7b8e4b84180e0a007b5505fde9.

crates/walrus-service/src/node.rs
mlegner: Given this is a one-time log entry at startup, I'd rather leave it as it is now.

crates/walrus-service/src/client/cli/runner.rs
mlegner: I tried to minimize (2) in 15229c1989a12e7b8e4b84180e0a007b5505fde9. Maybe (1) could be done in #1214 or in a follow-up?

giac-mysten: sounds good! thanks for the clarification

giac-mysten: ~~I can try to add (1) to #1214~~
tracked in an issue #1231

Issue #557, opened 2024-06-20
Add progress output to the CLI


mlegner:
When storing larger blobs with the CLI, there is a long period of no output during which the user doesn't know if anything is happening at all. It might be useful to provide some intermediate progress output. This could include the current stage (encoding, reserving space, storing data, certification) as well as the progress within a stage (especially the storing stage).

https://github.com/console-rs/indicatif looks like a great crate to use for this.

PR #1214, opened 2024-11-21
feat(client): add progress bars and spinners

giac-mysten:
Adds progress bars and spinners (with a cute blob animation) to the CLI store and read commands.
The `indicatif` library should take care of not showing the bars in the logs.

Also changes the accents for success and info heading commands to match the "Walrus brand" (cyan and magenta instead of green/yellow) 😬 

Closes: #557

### DIFF
--- a/crates/walrus-service/src/client/cli.rs
+++ b/crates/walrus-service/src/client/cli.rs
@@ -12,7 +12,7 @@ use std::{
 };
 
 use anyhow::{anyhow, Context, Result};
-use colored::{ColoredString, Colorize};
+use colored::{Color, ColoredString, Colorize};
 use indicatif::{ProgressBar, ProgressStyle};
 use num_bigint::BigUint;
 use serde::{Deserialize, Serialize};
@@ -154,7 +154,7 @@ pub async fn get_sui_read_client_from_rpc_node_or_wallet(
 
 /// Returns the string `Success:` colored in green for terminal output.
 pub fn success() -> ColoredString {
-    "Success:".bold().cyan()
+    "Success:".bold().walrus_teal()
 }
 
 /// Returns the string `Error:` colored in red for terminal output.
@@ -165,6 +165,58 @@ pub fn error() -> ColoredString {
 /// Returns the string `Warning:` colored in yellow for terminal output.
 pub fn warning() -> ColoredString {
     "Warning:".bold().yellow()
+}
+
+// Custom colors to match the Walrus brand identity.
+
+/// The Walrus teal color.
+///
+/// The original walrus teal is `#99efe4`. However, to remain compatible with 256-color
+/// terminals, we use the closest color available in the 256-color palette, which is `122
+/// (#87ffd7)`.
+pub const WALRUS_TEAL: Color = Color::TrueColor {
+    r: 135,
+    g: 255,
+    b: 215,
+};
+/// The Walrus purple color.
+///
+/// The original walrus purple is `#c484f6 `. However, to remain compatible with 256-color
+/// terminals, we use the closest color available in the 256-color palette, which is `177
+/// (#d787ff)`.
+pub const WALRUS_PURPLE: Color = Color::TrueColor {
+    r: 215,
+    g: 135,
+    b: 255,
+};
+
+/// Trait to color strings with the Walrus brand colors.
+pub trait WalrusColors {
+    /// Colors the string in the Walrus teal color.
+    fn walrus_teal(self) -> ColoredString;
+
+    /// Colors the string in the Walrus purple color.
+    fn walrus_purple(self) -> ColoredString;
+}
+
+impl WalrusColors for ColoredString {
+    fn walrus_teal(self) -> ColoredString {
+        self.color(WALRUS_TEAL)
+    }
+
+    fn walrus_purple(self) -> ColoredString {
+        self.color(WALRUS_PURPLE)
+    }
+}
+
+impl<'a> WalrusColors for &'a str {
+    fn walrus_teal(self) -> ColoredString {
+        self.color(WALRUS_TEAL)
+    }
+
+    fn walrus_purple(self) -> ColoredString {
+        self.color(WALRUS_PURPLE)
+    }
 }
 
 /// Type to help with formatting bytes as human-readable strings.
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/client/cli.rs
```diff
@@ -152,7 +154,7 @@ pub async fn get_sui_read_client_from_rpc_node_or_wallet(
 
 /// Returns the string `Success:` colored in green for terminal output.
 pub fn success() -> ColoredString {
-    "Success:".bold().green()
+    "Success:".bold().cyan()
```
karlwuest: Suggestion: Instead of cyan, we could use the exact color that marketing materials use. I'm sure we have the values specified somewhere, I assume the ones on walrus.xyz are canonical. This would be 99efe4 (as "teal" instead of cyan) and c484f6 (as "purple" instead of magenta). For the purples, different values are used, this would be the fully saturated shade.

crates/walrus-service/Cargo.toml
```diff
@@ -71,6 +71,7 @@ futures-util.workspace = true
 git-version.workspace = true
 home.workspace = true
 humantime.workspace = true
+indicatif = "0.17.9"
```
mlegner: Nit: Can you move the version to the workspace `Cargo.toml` file and replace this by
```suggestion
indicatif.workspace = true
```

crates/walrus-service/src/client.rs
```diff
@@ -423,22 +431,13 @@ impl<T: ContractClient> Client<T> {
...
 
             let certificate = match blob_status.initial_certified_epoch() {
                 Some(certified_epoch) if !committees.is_change_in_progress() => {
-                    // The blob is already certified on chain: the slivers are already available.
-                    // However, during epoch change we may need to store the slivers again, as the
-                    // current committee may not have synced them yet.
+                    // If the blob is already certified on chain and there is no committee change in
+                    // progress, all nodes already have the slivers.
                     self.get_certificate_standalone(&blob_id, certified_epoch)
                         .await?
```
mlegner: Important: There seems to have been some kind of merge/rebase issue with #1202.

crates/walrus-service/src/client.rs
```diff
@@ -599,13 +598,24 @@ impl<T> Client<T> {
             .communication_factory
             .node_write_communications(&committees, Arc::new(Semaphore::new(sliver_write_limit)))?;
 
+        let pb = styled_progress_bar(bft::min_n_correct(committees.n_shards()).get().into());
```
mlegner: Nit: Can we call this `progress_bar`? This is used 50 lines below and at that point it's unclear what this refers to.

crates/walrus-service/src/client.rs
```diff
@@ -599,13 +598,24 @@ impl<T> Client<T> {
...
                 metadata,
                 pairs_per_node
                     .remove(&n.node_index)
                     .expect("there are shards for each node"),
             )
+            .inspect({
+                let value = pb.clone();
+                move |result| {
+                    if result.is_ok() && !value.is_finished() {
+                        value.inc(n.n_owned_shards().get().into())
```
mlegner: Question: Why not use the weight of the result (`result.1`) for this?

Related remark: Maybe we should change `NodeResult` from a tuple struct to a "normal" struct to make it clear what the different fields contain. (In a follow-up.)

crates/walrus-service/src/client.rs
```diff
@@ -762,6 +788,11 @@ impl<T> Client<T> {
     {
         let committees = self.committees.read().await;
 
+        // Create a progress bar to track the progress of the sliver retrieval.
+        // TODO: this will be moved will need a flag to check if we are running in CLI mode.
```
mlegner: Remark: This TODO should be addressed or related to a follow-up issue.

Also: Does this apply to other functions here as well?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -415,9 +416,12 @@ impl ClientCommandRunner {
         };
 
         tracing::debug!(%n_shards, "encoding the blob");
+        let spinner = styled_spinner();
+        spinner.set_message("computing the blob ID");
         let metadata = EncodingConfig::new(n_shards)
             .get_blob_encoder(&read_blob_from_file(&file)?)?
             .compute_metadata();
+        spinner.finish_with_message("blob ID computed");
```
mlegner: Suggestion: How about we print the blob ID here?

crates/walrus-service/src/client.rs
```diff
@@ -382,6 +389,7 @@ impl<T: ContractClient> Client<T> {
             duration = ?encode_duration,
             "encoded sliver pairs and metadata"
         );
+        spinner.finish_with_message("blob encoded");
```
mlegner: Suggestion: Print the blob ID here.

crates/walrus-service/src/client/cli.rs
giac-mysten: unfortunately, I could not find a way to change the color of the spinners in the indicatif crate to a custom one. Therefore, it's probably better to keep a consistent cyan/magenta?

crates/walrus-service/src/client.rs
giac-mysten: good point! yes, opened an issue  #1224

crates/walrus-service/src/client.rs
giac-mysten: this was actually solved, and I forgot to remove it: the `indicatif` crate should automatically detect if the client is not running in a terminal and then not print the bars.

crates/walrus-service/src/client.rs
giac-mysten: i think this should've been solved by the last rebase

crates/walrus-service/src/client/cli.rs
karlwuest: It looks like [this method](https://docs.rs/console/latest/console/struct.Style.html#method.from_dotted_str) is used to convert the style, which means that custom colors should be supported, but only [256](https://docs.rs/console/latest/console/enum.Color.html). We could just use the closest available color to the "correct" ones and then also use the same one for the text. I'll try to figure out the correct values.

karlwuest: Ok, so the closest colors (assuming that the 256 colors are xterm256 colors) would be:

- "Teal": 122 -> closest rgb is 87FFD7
- "Purple": 177 -> closest rgb is D787FF

karlwuest: Alternatives that actually seem visually closer (especially the purple) would be:
- 123 -> 87FFFF
- ~178 -> D7AF0~ (the color picker I used didn't update after checking the original color, this is definitely not purple)

giac-mysten: cool! i'll use the two above

Issue #1218, opened 2024-11-25
Return error early if WeightedFuture collected `anti-quorum` weighted errors.


halfprice:
Currently, WeightedFutures returns only if a set of successful responses returned that exceeds the weight (2f+1), or no more future can be consumed. During events such as epoch change where client may get consistent EpochMismatch error, it may make sense to return error results if more than (f+1) weighted error is collect. Currently, the WeightedFutures wait for collecting all error responses.

This is discovered in simtest where client may have unnecessary long blob upload latency during node failure: even though the failed node contains # of shards < f+1, it may block the WeightedFuture to complete if all other requests also returned error (e.g., during epoch change)

Issue #1172, opened 2024-11-13
Investigate simtest failures and make them more robust


mlegner:
It seems that simtests have been failing somewhat frequently on various PRs even when changes should not affect the tests. We should investigate the failures and either fix bugs that are discovered or make the tests more robust.

Some of these issues have been solved by now. Some are still open:

- [ ] Make sure inconsistent sliver is not assigned to failed node in `test_inconsistency`
- [ ] Investigate RocksDB panic (e.g., [here](https://github.com/MystenLabs/walrus/actions/runs/11851977497/job/33029372368?pr=1180#step:5:1827)):
  ```
  storage creation must succeed: rocksdb error: IO error: No such file or directory: While renaming a file to /tmp/tmp.mUiLqNhTFy/.tmpmCdskL/CURRENT: /tmp/tmp.mUiLqNhTFy/.tmpmCdskL/000005.dbtmp: No such file or directory
  ```
- [ ] Intermittent failures of `simtest_walrus_with_single_node_crash_and_restart` and `simtest_test_lagging_node_recovery`. See for example [here](https://github.com/MystenLabs/walrus/actions/runs/11852431669/job/33030604216?pr=1179).
- [ ] Intermittent failures of `simtest_test_repeated_node_crash`. See for example [here](https://github.com/MystenLabs/walrus/actions/runs/12002779152/job/33455268296?pr=1210). Potentially related to this error:
  ```
  2021-12-27T15:31:09.182324Z ERROR node{id=10 name="r+2hKA9BwSgVsBs4kX7hSI1zl/hPARfGPSonFtnlx/HG6gO65XY/Qz8tlQ63S3XK"}: walrus_service::node::shard_sync: crates/walrus-service/src/node/shard_sync.rs:158: Failed to sync shard index: shard-2 to before epoch: 5. Error: RequestError(NodeError { kind: StatusWithMessage { inner: reqwest::Error { kind: Status(400), url: "https://10.10.0.11:1314/v1/migrate/sync_shard" }, message: "Invalid epoch. Client epoch: 5. Server epoch: 4", service_error: Some(InvalidEpoch { request_epoch: 5, server_epoch: 4 }) } })
  ```
- [ ] Failure of `simtest_test_repeated_shard_move_with_workload`. First failed after merging #1204, see [here](https://github.com/MystenLabs/walrus/actions/runs/12047162217/job/33589139229).

PR #1219, opened 2024-11-25
fix: deflake test_lagging_node_recovery test

halfprice:
The main reason for the test failure that happens pretty often is due to that after the node crashes, it takes more than 30 seconds (reqwest timeout) for the client to get an error from the crashed node. However, 30 seconds is also the epoch length in the simtest, so by the time it retries the request, another epoch passed, so we have two consecutive upload blob fails, which caused the test to fail.

Although the ultimate fix would be to address #1218 , this PR deflake the test by reducing reqwest timeout in simtest, so that it can return error early without waiting for 30 seconds.

100 seeds passed.

Contributes to #1172

### DIFF
--- a/crates/walrus-service/src/client/communication/node.rs
+++ b/crates/walrus-service/src/client/communication/node.rs
@@ -300,19 +300,22 @@ impl<'a> NodeWriteCommunication<'a> {
                 .store_metadata_with_retries(metadata)
                 .await
                 .map_err(StoreError::Metadata)?;
-            tracing::debug!(?metadata_status, "finished storing metadata on node");
+            tracing::debug!(node = %self.node.public_key, ?metadata_status,
+                "finished storing metadata on node");
 
             let n_stored_slivers = self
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/client/communication/node.rs
```diff
@@ -300,19 +300,22 @@ impl<'a> NodeWriteCommunication<'a> {
...
-            tracing::debug!(n_stored_slivers, "finished storing slivers on node");
+            tracing::debug!(node = %self.node.public_key, n_stored_slivers,
+                "finished storing slivers on node");
 
             self.get_confirmation_with_retries_inner(metadata.blob_id(), self.committee_epoch)
                 .await
                 .map_err(StoreError::Confirmation)
         }
         .await;
-        tracing::debug!("successfully stored metadata and sliver pairs");
```
halfprice, line 315: This is misleading. "storing metadata and sliver pairs" may not be successful.

PR #1226, opened 2024-11-27
fix: is_stored_at_all_shards needs to check with contract, not local shard storage

halfprice:
Since local shard map may not have the most up-to-date shard information.

This was discovered during the epoch 33 epoch change, where blob sync is started due to shards that are supposed to move out (but the node doesn't know, since epoch change 32 didn't run in the node).

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -903,12 +903,9 @@ impl StorageNode {
         // Given that the storage node is severely lagging, the node may contain shards in outdated
         // status. We need to set the status of all currently owned shards to `Active` despite
         // their current status.
-        for shard in committees
-            .current_committee()
-            .shards_for_node_public_key(public_key)
-        {
+        for shard in self.inner.owned_shards() {
             storage
-                .shard_storage(*shard)
+                .shard_storage(shard)
                 .expect("we just create all storage, it must exist")
                 .set_active_status()?;
         }
@@ -1108,7 +1105,7 @@ impl StorageNodeInner {
         self.committee_service
             .active_committees()
             .current_committee()
-            .shards_for_node_public_key(self.protocol_key_pair.public())
+            .shards_for_node_public_key(self.public_key())
             .to_vec()
     }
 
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -520,9 +521,9 @@ impl StorageNode {
         Ok(())
     }
 
-    /// Returns the shards currently owned by the storage node.
-    pub fn shards(&self) -> Vec<ShardIndex> {
-        self.inner.storage.shards()
+    /// Returns the existing shards currently owned by the storage node.
```
mlegner: Question: Does this mean the shards for which the node has storage? This only looks at the storage, not the status or whether the node owns them, correct?
```suggestion
    /// Returns the shards which the node currently manages in its storage.
    ///
    /// This neither considers the current shard assignment from the Walrus contracts nor the status
    /// of the local shard storage.
```

crates/walrus-service/src/node.rs
```diff
@@ -1110,6 +1114,28 @@ impl StorageNodeInner {
         &self.encoding_config
     }
 
+    pub(crate) fn owned_shards(&self) -> Vec<ShardIndex> {
+        self.committee_service
+            .active_committees()
+            .current_committee()
+            .shards_for_node_public_key(self.protocol_key_pair.public())
```
mlegner: Nit: This function could also be used in `StorageNode::start_node_recovery` and `StorageNodeInner::shard_health_status`.

Also: There is a `StorageNodeInner::public_key()` function to simplify this slightly.

crates/walrus-service/src/node.rs
```diff
@@ -1110,6 +1114,28 @@ impl StorageNodeInner {
...
+            .to_vec()
+    }
+
+    pub(crate) fn is_stored_at_all_shards(&self, blob_id: &BlobId) -> anyhow::Result<bool> {
+        for shard in self.owned_shards() {
+            match self.storage.is_stored_at_shard(blob_id, shard) {
+                Ok(false) => return Ok(false),
+                Ok(true) => continue,
+                Err(e) => {
+                    tracing::warn!(?e, "failed to check if blob is stored at shard");
```
mlegner: Nit: As the metadata field is recorded by name, can we rename this?
```suggestion
                Err(error) => {
                    tracing::warn!(?error, "failed to check if blob is stored at shard");
```

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -486,15 +486,6 @@ where
...
     }
 
-    fn committee(&self) -> Arc<Committee> {
-        self.inner
-            .committee_tracker
-            .borrow()
-            .committees()
-            .current_committee()
-            .clone()
-    }
```
mlegner, line 496: Question: Why was this function removed?

crates/walrus-service/src/node.rs
mlegner: Done in 258041ec51a0480e7fc0e6e9ed3b21a8260e2440.

Issue #1139, opened 2024-11-04
Fix rust test coverage report and re-enable weekly reports


mlegner:
The code-coverage tool has been broken for a while, see the [failed weekly reports](https://github.com/MystenLabs/walrus/actions/workflows/scheduled_reports.yml).

PR #1227, opened 2024-11-27
chore: some repository cleanup

mlegner:
## Description

This PR contains some repository cleanup, in particular:
- exclude some crates from default members so they are not built and tested by default
- remove some dead code
- improve tarpaulin setup
- use new module naming convention everywhere
- consolidate diff action in CI
- update and clean up README file

## Test plan

No behavioral changes. CI changes should be tested automatically.

### DIFF
--- a/crates/walrus-service/bin/deploy.rs
+++ b/crates/walrus-service/bin/deploy.rs
@@ -13,7 +13,6 @@ use std::{
 use anyhow::Context;
 use clap::{Parser, Subcommand};
 use humantime::Duration;
-use tokio::sync::oneshot;
 use walrus_core::EpochCount;
 use walrus_service::{
     node::config::{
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -4,7 +4,6 @@ repos:
     hooks:
       - id: check-merge-conflict
       - id: check-yaml
-      - id: trailing-whitespace
```
giac-mysten, line 7: Question: why remove this? covered by editorconfig?

.tarpaulin.toml
```diff
@@ -2,12 +2,13 @@
...
   "walrus-core",
-  "walrus-event",
+  "walrus-e2e-tests",
   "walrus-sdk",
   "walrus-service",
   "walrus-sui",
+  "walrus-utils",
 ]
-run-types = ["Doctests", "Lib", "Tests"]
+run-types = ["Lib", "Tests"]
```
giac-mysten, line 11: Questions: i guess we don't have doctests?

crates/walrus-service/bin/deploy.rs
```diff
@@ -312,38 +311,6 @@ mod commands {
     }
 }
 
-/// Wait for SIGINT and SIGTERM (unix only).
```
giac-mysten, line 315: question: i see this was a duplicate. Why did CI not complain?

.pre-commit-config.yaml
mlegner: Because we have the same a few lines below... 😅

.tarpaulin.toml
mlegner: We do, but somehow they don't seem to work in Tarpaulin atm... 🤷 

But I will have to do some additional work on this (#1139).

crates/walrus-service/bin/deploy.rs
mlegner: This *might* be due to the `#[tokio::main]` attribute, but I'm not sure. 🤷 

I believe the duplication happened when reorganizing the crate structure.

Issue #1040, opened 2024-10-11
Automatically keep move errors and their counterparts in rust in sync


karlwuest:
Or at least automatically test to make sure that they are in sync.

PR #1234, opened 2024-11-28
chore: automatically generate Rust error types from Move

karlwuest:
## Description

Automatically generates Rust error types from the Move contracts.

closes WAL-264 (#1040 )

## Test plan

- Manually compared the auto-generated error definitions with the previous ones

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/build.rs
+++ b/crates/walrus-sui/build.rs
@@ -34,31 +34,35 @@ impl ModuleErrorDefs {
 }
 
 fn main() {
-    let out_dir = env::var_os("OUT_DIR").unwrap();
-    let dest_path = Path::new(&out_dir).join("move_error_defs.rs");
+    let out_dir = env::var_os("OUT_DIR").expect("OUT_DIR should be set automatically by cargo");
 
+    for contracts_dir in ["../../contracts", "../../testnet-contracts"] {
+        generate_error_defs_for_contracts(Path::new(contracts_dir), Path::new(&out_dir));
+        // Tell cargo to rerun if relevant files change
+        println!("cargo:rerun-if-changed={}", contracts_dir);
+    }
+}
+
+fn generate_error_defs_for_contracts(contracts_dir: &Path, out_dir: &Path) {
+    let dest_path = out_dir.join(format!(
+        "{}_error_defs.rs",
+        contracts_dir
+            .file_name()
+            .expect("contracts dir should have a name")
+            .to_string_lossy()
+    ));
     // Collect error definitions from Move files
-    let error_defs = collect_error_definitions();
+    let error_defs = collect_error_definitions(contracts_dir);
 
     // Generate Rust code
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/build.rs
```diff
@@ -0,0 +1,201 @@
...
+        format!("{}Error", self.enum_variant_name())
+    }
+
+    fn enum_variant_name(&self) -> String {
+        self.module_name().to_pascal_case()
+    }
+}
+
+fn main() {
+    let out_dir = env::var_os("OUT_DIR").unwrap();
```
mlegner: Nit:
```suggestion
    let out_dir = env::var_os("OUT_DIR").expect("this is automatically set by cargo");
```

crates/walrus-sui/build.rs
```diff
@@ -0,0 +1,201 @@
...
+    let error_defs = collect_error_definitions();
+
+    // Generate Rust code
+    let error_defs_code = generate_all_error_defs(&error_defs);
+
+    // Write to file
+    fs::write(&dest_path, error_defs_code).expect("should be able to write to file");
+
+    // Tell cargo to rerun if relevant files change
+    println!("cargo:rerun-if-changed=build.rs");
```
mlegner: If I understand [the docs](https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed) correctly, this is unnecessary.
```suggestion
```

crates/walrus-sui/build.rs
```diff
@@ -0,0 +1,201 @@
...
+    println!("cargo:rerun-if-changed=../testnet-contracts");
+}
+
+/// Collect the error names and codes from the Move source code for each module.
+fn collect_error_definitions() -> Vec<ModuleErrorDefs> {
+    let contracts_dir = if cfg!(feature = "mainnet-contracts") {
+        Path::new("../../contracts")
+    } else {
+        Path::new("../../testnet-contracts")
+    };
```
mlegner: Question: How does that work when re-compiling with a different set of features? When I run `cargo build`, it seems to run the build script with and without the `mainnet-contracts` feature. But I don't fully understand how this works...

An alternative (maybe more robust?) would be to just generate the error definitions for both contract versions and select the correct one in `crates/walrus-sui/src/types/move_errors.rs` depending on the feature.

karlwuest: >An alternative (maybe more robust?) would be to just generate the error definitions for both contract versions and select the correct one in crates/walrus-sui/src/types/move_errors.rs depending on the feature.

That's a good suggestion! I was also slightly worried about this, because it's surprisingly hard to find information about how the build script handles features. When I tested it, it always generated the correct version, but your suggestion seems more robust.

karlwuest: Done!

PR #1237, opened 2024-11-28
fix: split coin in call to exchange

karlwuest:
## Description

Fixes the transaction to exchange SUI for WAL

## Test plan

- adds an integration test for the exchange
- also adds a test for WAL coin squashing

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: fixes a bug that prevented exchanging SUI for WAL

### DIFF
--- a/crates/walrus-sui/tests/test_walrus_sui.rs
+++ b/crates/walrus-sui/tests/test_walrus_sui.rs
@@ -377,59 +377,93 @@ async fn test_exchange_sui_for_wal() -> anyhow::Result<()> {
 #[ignore = "ignore integration tests by default"]
 async fn test_automatic_wal_coin_squashing() -> anyhow::Result<()> {
     _ = tracing_subscriber::fmt::try_init();
-    let (sui_cluster_handle, walrus_admin_client) = initialize_contract_and_wallet().await?;
+    let (sui_cluster_handle, client_1) = initialize_contract_and_wallet().await?;
 
-    let original_balance = walrus_admin_client.as_ref().balance(CoinType::Wal).await?;
+    let original_balance = client_1.as_ref().balance(CoinType::Wal).await?;
 
-    let walrus_client = new_contract_client_on_sui_test_cluster(
-        sui_cluster_handle.clone(),
-        walrus_admin_client.as_ref(),
-    )
-    .await?;
+    let client_2 =
+        new_contract_client_on_sui_test_cluster(sui_cluster_handle.clone(), client_1.as_ref())
+            .await?;
+
+    let client_1_address = client_1.as_ref().address();
+    let client_2_address = client_2.as_ref().address();
 
     let amount = 100_000;
 
     // Fund the wallet with two separate WAL coins.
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/tests/test_walrus_sui.rs
```diff
@@ -340,3 +348,88 @@ async fn test_register_candidate() -> anyhow::Result<()> {
...
+
+    let post_balance = walrus_client.as_ref().balance(CoinType::Wal).await?;
+    assert_eq!(post_balance, pre_balance + exchange_val);
+
+    Ok(())
+}
+
+#[tokio::test]
+#[ignore = "ignore integration tests by default"]
+async fn test_automatic_wal_coin_squashing() -> anyhow::Result<()> {
```
giac-mysten, line 378: question: how are we testing the squashing here? iiuc, the function never checks the number of coins?

Suggestion: use something like `select_coins` or `get_coins_with_total_balance` ?

crates/walrus-sui/tests/test_walrus_sui.rs
```diff
@@ -340,3 +348,88 @@ async fn test_register_candidate() -> anyhow::Result<()> {
...
+        .pay_wal(walrus_client.as_ref().address(), amount)
+        .await?;
+    walrus_admin_client
+        .as_ref()
+        .sign_and_send_ptb(&wallet, tx_builder.finish().await?.0, None)
+        .await?;
+    assert_eq!(walrus_client.as_ref().balance(CoinType::Wal).await?, amount);
+
+    // Second transaction.
+    let mut tx_builder = walrus_admin_client.as_ref().transaction_builder();
```
giac-mysten: Question: we need 2 separate tx because otherwise the wal would be squashed here, correct?

crates/walrus-sui/tests/test_walrus_sui.rs
karlwuest: Do you mean adding an assert for this after the two transactions that send the coins to the second client? That's a good idea!

crates/walrus-sui/tests/test_walrus_sui.rs
karlwuest: I don't think they would be squashed (since they are the outputs, not the inputs), but I just wanted to make it very clear that we have two coins. However, I'll add an explicit check that we need to coins to cover the last transaction and then I can do a single PTB here.

crates/walrus-sui/tests/test_walrus_sui.rs
giac-mysten: And also after they are sent back? to check that the first client receives the merged coin?

karlwuest: Sure, I can add a check there as well, although it's not as important because we are interested in the inputs being automatically squashed.

karlwuest: Done

PR #1238, opened 2024-11-28
fix(publisher): improve coin management

giac-mysten:
## Description

Several coin-management improvements to the publisher:

- if the full node responds with an error, the publisher does _not_ try to refill the wallet of the sub-client
- on startup, the publisher checks if the sub-client already has sufficient funds (as specified by the `--sub-wallets-initial-balance` flag), and does not send funds by default anymore.
- on startup, the publisher merges all WAL and SUI coins in the sub wallets, to help with coin management
- the number of default sub wallets has been reduced from 16 to 8

Closes WAL-338

This PR also simplifies the refiller code path, by removing the possiblity to refill from faucet.

Closes WAL-259

## Test plan

Manually tested several scenarios where the publisher:
- did not have sub wallets
- had sub wallets with sufficient balance
- had unmerged coins
- has sub wallets with insufficient balance

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [x] Publisher: Fixes issues with the coin management of sub-wallets: The main wallet now only refills sub-wallets on startup if they actually have insufficient balance and handles errors from full nodes better. The default number of sub-wallets is reduced from 16 to 8.
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/multiplexer.rs
+++ b/crates/walrus-service/src/client/multiplexer.rs
@@ -11,29 +11,23 @@ use std::{
     },
 };
 
-use futures::StreamExt as _;
 use prometheus::Registry;
 use sui_sdk::{
-    rpc_types::Coin,
     sui_client_config::SuiEnv,
     types::base_types::SuiAddress,
     wallet_context::WalletContext,
 };
-use sui_types::{
-    programmable_transaction_builder::ProgrammableTransactionBuilder,
-    transaction::Command,
-};
 use walrus_core::{BlobId, EpochCount};
 use walrus_sui::{
-    client::{get_system_package_id, BlobPersistence, SuiContractClient, SuiReadClient},
-    utils::{create_wallet, sign_and_send_ptb},
+    client::{BlobPersistence, SuiContractClient, SuiReadClient},
+    utils::create_wallet,
 };
 
 use super::{
     cli::PublisherArgs,
     daemon::{WalrusReadClient, WalrusWriteClient},
     metrics::ClientMetrics,
-    refill::{CoinRefill, RefillHandles, Refiller, WalletCoinRefill},
+    refill::{RefillHandles, Refiller},
     responses::BlobStoreResult,
     Client,
     ClientResult,
@@ -62,15 +56,10 @@ impl ClientMultiplexer {
         let sui_read_client = contract_client.read_client.clone();
         let read_client = Client::new_read_client(config.clone(), sui_read_client).await?;
 
-        let system_pkg_id = get_system_package_id(&sui_client, config.system_object).await?;
         let refiller = Refiller::new(
-            WalletCoinRefill::new(
-                contract_client,
-                args.gas_refill_amount,
-                args.wal_refill_amount,
-                gas_budget,
-            )?,
-            system_pkg_id,
+            contract_client,
+            args.gas_refill_amount,
+            args.wal_refill_amount,
             args.sub_wallets_min_balance,
         );
 
@@ -151,13 +140,13 @@ pub struct WriteClientPool {
 
 impl WriteClientPool {
     /// Creates a new client pool with `n_client`, based on the given `config` and `sui_env`.
-    pub async fn new<G: CoinRefill + 'static>(
+    pub async fn new(
         n_clients: usize,
         config: &Config,
         sui_env: SuiEnv,
         gas_budget: u64,
         sub_wallets_dir: &Path,
-        refiller: &Refiller<G>,
+        refiller: &Refiller,
         min_balance: u64,
     ) -> anyhow::Result<Self> {
         tracing::info!(%n_clients, "creating write client pool");
@@ -201,23 +190,23 @@ impl WriteClientPool {
 }
 
 /// Helper struct to build or load sub clients for the client multiplexer.
-struct SubClientLoader<'a, G> {
+struct SubClientLoader<'a> {
     config: &'a Config,
     sub_wallets_dir: &'a Path,
     sui_env: SuiEnv,
     gas_budget: u64,
-    refiller: &'a Refiller<G>,
+    refiller: &'a Refiller,
     /// The minimum balance the sub-wallets should have, below which they are refilled at startup.
     min_balance: u64,
 }
 
-impl<'a, G: CoinRefill + 'static> SubClientLoader<'a, G> {
+impl<'a> SubClientLoader<'a> {
     fn new(
         config: &'a Config,
         sub_wallets_dir: &'a Path,
         sui_env: SuiEnv,
         gas_budget: u64,
-        refiller: &'a Refiller<G>,
+        refiller: &'a Refiller,
         min_balance: u64,
     ) -> Self {
         Self {
@@ -251,12 +240,13 @@ impl<'a, G: CoinRefill + 'static> SubClientLoader<'a, G> {
         let mut wallet = self.create_or_load_sub_wallet(sub_wallet_idx)?;
         self.top_up_if_necessary(&mut wallet, self.min_balance)
             .await?;
-        self.merge_coins(&mut wallet).await?;
 
         let sui_client = self
             .config
             .new_contract_client(wallet, self.gas_budget)
             .await?;
+        // Merge existing coins to avoid fragmentation.
+        sui_client.merge_coins().await?;
 
         Ok(Client::new_contract_client(self.config.clone(), sui_client).await?)
     }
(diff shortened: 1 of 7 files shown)


crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -262,11 +281,93 @@ impl<'a, G: CoinRefill> SubClientLoader<'a, G> {
...
-    async fn refill_wallet(&self, wallet: &mut WalletContext) -> anyhow::Result<()> {
+    /// Ensures the wallet has at least 1 coin of at least`initial_balance` SUI and WAL.
+    async fn top_up_if_necessary(
+        &self,
+        wallet: &'a mut WalletContext,
+        initial_balance: u64,
+    ) -> anyhow::Result<()> {
+        let wal_coin_type = self.refiller.wal_coin_type();
+        //let pkg_id = self.refiller.system_pkg_id.clone();
+        //let wal_coin_type = format!("{}::wal::WAL", pkg_id);
```
karlwuest: Nit
```suggestion
```

crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -262,11 +281,93 @@ impl<'a, G: CoinRefill> SubClientLoader<'a, G> {
...
+            .collect()
+            .await;
+
+        let wal_coins: Vec<_> = sui_client
+            .coin_read_api()
+            .get_coins_stream(address, Some(wal_coin_type))
+            .collect()
+            .await;
+
+        if sui_coins.len() > 1 || sui_coins.len() > 1 {
```
karlwuest:
```suggestion
        if sui_coins.len() > 1 || wal_coins.len() > 1 {
```

crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -262,11 +281,93 @@ impl<'a, G: CoinRefill> SubClientLoader<'a, G> {
...
+            sign_and_send_ptb(address, wallet, ptb, sui_coins, self.gas_budget).await?;
+        } else {
+            tracing::debug!(%address, "no coins to merge in sub-wallet");
+        }
+
+        Ok(())
+    }
+
+    /// Downloads the list of all coins of a type, and adds the commands to merge them to the ptb.
+    async fn merge_coin_commands(
```
karlwuest: Suggestion: I don't think this needs to be `async`.
```suggestion
    fn merge_coin_commands(
```

crates/walrus-service/src/client/refill.rs
```diff
@@ -401,3 +398,122 @@ pub struct RefillHandles {
...
+    .inspect_err(|error| {
+        tracing::debug!(
+            ?error,
+            %address,
+            "failed checking the amount of coins owned by the address"
+        );
+    })
+    // If the returned value is an error, we assume that the RPC failed, the
+    // address has enough coins, and we do not try to refill.
+    .unwrap_or(true)
```
karlwuest: Question/Suggestion: Do we still need to check anywhere that we have at least a certain number of coins? 
As I mentioned on slack earlier this week, I think it should be sufficient to check the balance, because we now squash coins anytime we need them in the SuiContractClient. This massively simplifies things, because then the functions `has_enough_coins_of_value` and `get_coins_stream_with_errors` are no longer needed.

```suggestion
    min_balance: u64,
) -> bool {
    sui_client
        .coin_read_api()
        .get_balance(address, coin_type)
        .await
        .map(|balance| balance.total_balance < min_balance as u128)
        .inspect_err(|error| {
            tracing::debug!(
                ?error,
                %address,
                "failed checking the balance of the address"
            );
        })
        // If the returned value is an error, we assume that the RPC failed, the
        // address has enough coins, and we do not try to refill.
        .unwrap_or(true)
```

giac-mysten: are the gas coins also squashed? in that case, I think we can remove entirely the "num_coins" logic

karlwuest: Yes, gas coins are always squashed

giac-mysten: ok, then i'll fully remove the num_coins logic, that was only needed b/c we did not have the WAL coin and needed 2 coins to pay in SUI

crates/walrus-service/src/client/cli/args.rs
mlegner, line 1: Question/suggestion for the `max_concurrent_requests` and `max_request_buffer_size` parameters: AFAIU, these are not automatically adjusted based on the `n_clients` parameter, correct? That means if I want to change `n_clients` to some non-default value, I also need to adjust the other parameters.

If that understanding is correct, could we make the other parameters an `Option<usize>` and then do an `unwrap_or(n_clients)`?

Also, more general suggestion (for a follow-up): Given the large number of CLI parameters it might make sense to replace them by a publisher section in the client config.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -603,7 +619,7 @@ pub(crate) mod default {
     }
 
     pub(crate) fn max_concurrent_requests() -> usize {
-        16
+        8
```
mlegner, line 622: Question: This does not require any changes for deployments, correct? The wallets 8--15 are then just not used anymore?

crates/walrus-service/src/client/refill.rs
```diff
@@ -155,95 +110,34 @@ impl CoinRefill for WalletCoinRefill {
...
-        match self {
-            Self::Network(faucet) => faucet.send_wal_request(address).await,
-            Self::Wallet(wallet) => wallet.send_wal_request(address).await,
-        }
-    }
-}
-
 /// Refills gas and WAL for the clients.
 #[derive(Debug)]
 pub struct Refiller<G> {
```
karlwuest: Question: Does this still need to be generic now that the network refiller is removed? 

Suggestion: If `refill_inner` is always a `WalletCoinRefill`, make it non-generic, or even flatten it. This would allow simplifying it much further, e.g. we don't need a separate sui client anymore, because `WalletCoinRefill` already contains a `SuiContractClient`, and we would also not need the `system_pkg_id` anymore, because we get the coin type etc. from that as well.

crates/walrus-service/src/client/refill.rs
```diff
@@ -18,25 +15,9 @@ use sui_sdk::{
...
-const MIN_NUM_COINS: usize = 1;
-/// The amount in MIST that is transferred from the wallet refill account to the stress clients at
-/// each request.
-const WALLET_MIST_AMOUNT: u64 = 1_000_000_000;
-/// The amount in FROST that is transferred from the wallet refill account to the stress clients at
-/// each request.
-const WALLET_FROST_AMOUNT: u64 = 1_000_000_000;
 
 /// Trait to request gas and Wal coins for a client.
 pub trait CoinRefill: Send + Sync {
```
karlwuest: Question: Do we still need the trait after removing `NetworkCoinRefill`?

crates/walrus-service/src/client/refill.rs
```diff
@@ -330,7 +222,6 @@ impl<G: CoinRefill + 'static> Refiller<G> {
         sui_client: SuiClient,
         coin_type: Option<String>,
         min_coin_value: u64,
-        min_num_coins: usize,
```
karlwuest, line 333: Nit: I think the `#[allow(clippy::too_many_arguments)]` above can be removed now.

crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -262,11 +287,83 @@ impl<'a, G: CoinRefill> SubClientLoader<'a, G> {
...
+                ?wal_coins,
+                "merging SUI or WAL coins in sub-wallet"
+            );
+            Self::merge_coin_commands(wal_coins, &mut pt_builder)?;
+            let ptb = pt_builder.finish();
+            // Even if there are no WAL coins, we do a no-op to smash the SUI gas coins.
+            sign_and_send_ptb(address, wallet, ptb, sui_coins, self.gas_budget).await?;
+        } else {
+            tracing::debug!(%address, "no coins to merge in sub-wallet");
+        }
```
karlwuest: Suggestion: Instead of doing this here by passing a `WalletContext`, this could be done in `SuiContractClient` and then you could just call it on each of the clients. This should be quite simple with the `WalrusPtbBuilder`.

Alternatively, you could pass a `SuiContractClient` to this function instead of the `WalletContext` and then do roughly (please double check, only wrote this here) the following:

````suggestion
    /// Merges the WAL and SUI coins in the wallet.
    async fn merge_coins(&self, client: &SuiContractClient) -> anyhow::Result<()> {
        let wallet = client.wallet().await;
        let mut tx_builder = client.transaction_builder();
        let sui_balance = client
            .sui_client()
            .coin_read_api()
            .get_balance(client.address(), None)
            .await?;
        let wal_balance = client
            .sui_client()
            .coin_read_api()
            .get_balance(client.address(), Some(client.wal_coin_type())
            .await?;
        if wal_balance.coin_object_count > 1 {
            tx_builder.fill_wal_balance(balance.total_balance as u64).await?;
        }
        if sui_balance.coin_object_count > 1 || wal_balance.coin_object_count > 1 {
            client
            .sign_and_send_ptb(&wallet, tx_builder.finish().await?.0, sui_balance.total_balance as u64)
            .await?;
        }
````

crates/walrus-service/src/client/cli/args.rs
giac-mysten: yes, that should work.

crates/walrus-service/src/client/multiplexer.rs
giac-mysten: like this. moved to the sui contract client

crates/walrus-service/src/client/cli/args.rs
giac-mysten: My initial idea was that `max_concurrent_requests`, `max_request_buffer_size`, and `n_clients` are only loosely coupled: they are set in this was as defaults because it seemed sensible to me, and not because there is a formula that controls their relative size.

Therefore, it seems a bit odd to me, e.g., to opaquely increase the `max_request_buffer_size` if I increase the `n_clients` and `max_concurrent_requests`: maybe I just wanted to reduce the client latency by having a shallower buffer. 

Not sure about this though, I'm open to suggestions.

>Also, more general suggestion (for a follow-up): Given the large number of CLI parameters it might make sense to replace them by a publisher section in the client config.

good idea, I added an [issue](https://linear.app/mysten-labs/issue/WAL-408/consolidate-the-publisherargs-with-the-client-config) for this.

mlegner: > My initial idea was that `max_concurrent_requests`, `max_request_buffer_size`, and `n_clients` are only loosely coupled: they are set in this was as defaults because it seemed sensible to me, and not because there is a formula that controls their relative size.
> 
> Therefore, it seems a bit odd to me, e.g., to opaquely increase the `max_request_buffer_size` if I increase the `n_clients` and `max_concurrent_requests`: maybe I just wanted to reduce the client latency by having a shallower buffer.

Hmm... I see. But we could still say in the docstring that they are set to `n_clients` by default, then it's not really opaque. If anyone explicitly touches these parameters, we won't change them, we just define the default based on the `n_clients`.

mlegner: But we can also address this as part of the [config follow-up task](https://linear.app/mysten-labs/issue/WAL-408).

PR #1239, opened 2024-11-29
chore: several fixes and improvements to the `local-testbed.sh` script

mlegner:
## Description

- Use checkpoint-based event processor.
- Be more careful about killing tmux sessions.
- Allow setting custom Sui networks (in the form `<RPC_URL>;<FAUCET_URL>`.
- Use release instead of debug builds.
 
Note that the testbed works with Sui Devnet, but currently has problems when starting up due to the large number of initial checkpoint download workers (see WAL-303, WAL-406).

An immediate fix could be to either start the nodes in sequence to avoid the initial burst or adjust the config to have a lower number of initial workers.

## Test plan

Run the script manually and interact with the resulting system.

### DIFF
--- a/scripts/local-testbed.sh
+++ b/scripts/local-testbed.sh
@@ -7,7 +7,7 @@ set -euo pipefail
 trap ctrl_c INT
 
 function kill_tmux_sessions() {
-    tmux ls | { grep -o "dryrun-node-\d*" || true; } | xargs -n1 tmux kill-session -t
+    { tmux ls || true; } | { grep -o "dryrun-node-\d*" || true; } | xargs -n1 tmux kill-session -t
 }
 
 function ctrl_c() {
@@ -24,6 +24,7 @@ function usage() {
     echo "  -s <n_shards>         Number of shards (default: 10)"
     echo "  -n <network>          Sui network to generate configs for (default: devnet)"
     echo "  -d <duration>         Set the length of the epoch (in human readable format, e.g., '60s', default: 1h)"
+    echo "  -t                    Use testnet contracts"
     echo "  -e                    Use existing config"
     echo "  -h                    Print this usage message"
 }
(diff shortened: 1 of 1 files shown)


scripts/local-testbed.sh
```diff
@@ -78,15 +81,9 @@ if ! [ "$shards" -ge "$committee_size" ] 2>/dev/null; then
...
-    echo "Invalid argument: $network is not a valid network (${nets[@]})."
-    usage
-    exit 1
-fi
-
 
-echo Building walrus-node and walrus-deploy binaries...
-cargo build --bin walrus-node --bin walrus-deploy
+echo Building walrus, walrus-node, and walrus-deploy binaries...
+cargo build --release --bin walrus --bin walrus --bin walrus-node --bin walrus-deploy --features "mainnet-contracts,deploy"
```
karlwuest: Suggestion: Should we allow specifying whether `mainnet-contracts` should be used?

scripts/local-testbed.sh
```diff
@@ -107,14 +104,18 @@ if ! $existing; then
...
+      --host-addresses $ips \
+      --storage-price 5 \
+      --write-price 1 \
+      --epoch-duration $epoch_duration
 
     # Generate configs
     echo Generating configuration...
-    cargo run --bin walrus-deploy -- generate-dry-run-configs \
-    --working-dir $working_dir --use-legacy-event-provider
+    ./target/release/walrus-deploy generate-dry-run-configs --working-dir $working_dir
```
karlwuest, line 130: Question: Does this work with a default local sui network? If not, can you add the necessary cli arguments to the [readme](https://github.com/MystenLabs/walrus/blob/3450b46c65185966658926c3160d3b04a312806f/README.md?plain=1#L103)?

Nit: since you mentioned that the downloading of the checkpoints takes a while, we could maybe add an option to the script to use the legacy event provider.

mlegner: > Does this work with a default local sui network?

~~Yes. I tested it with Devnet at least (the default).~~

I haven't tested it yet. Let me take a look.

> we could maybe add an option to the script to use the legacy event provider

I'd actually prefer not to "allow" this anymore to have this as close to the deployed networks. Also, this will probably be removed in the near future anyway. In the meantime, if someone *really* wants to use the legacy provider, they can always edit the file. 😅

mlegner: I tested it with the default localnet and it works out-of-the-box.

scripts/local-testbed.sh
mlegner: Done in b824440ca8394530356867c45ec0211717b28785.

PR #1248, opened 2024-12-02
feat(client): burn blobs, and return blobs to the sender through publisher

giac-mysten:
## Description

This PR adds the ability to burn blob objects through the CLI.

It also changes the behavior of the publisher regarding newly-stored blobs:
- there is now a `send_object_to` query keyword to the `store` endpoint to allow users to specify to which sui address the resulting `Blob` object should be sent to 
- if no `send_object_to` is specified in the query:
   - if the publisher is run with the `--keep` flag, the publisher will send the created `Blob` object from the sub-wallet to the main wallet (for safekeeping)
   - if the `--keep` flag is not specified, the new `Blob` object is `burn`t.

Closes WAL-415
Closes WAL-312
Closes WAL-421

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [x] Publisher: Changes the behavior of the publisher regarding newly-stored blobs, allowing to `send_object_to`, `burn`, or `--keep`.
- [x] CLI: Adds the ability to burn blob objects through the CLI.

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -455,10 +455,12 @@ pub struct PublisherArgs {
     #[clap(long, default_value_t = default::sub_wallets_min_balance())]
     #[serde(default = "default::sub_wallets_min_balance")]
     pub sub_wallets_min_balance: u64,
-    /// If set, the publisher will keep the created Blob object ids in the wallet.
+    /// If set, the publisher will keep the created Blob objects in its _main_ wallet.
     ///
-    /// If unset, the publisher will immediately burn all created blob objects by default, and only
-    /// send back the ones that are created with the `send_object_to` query.
+    /// If unset, the publisher will immediately burn all created blob objects by default. However,
+    /// note that this flag _does not affect_ the use of the `send_object_to` query parameter:
+    /// Regardless of this flag's status, the publisher will send created objects to the address in
+    /// the `send_object_to` query parameter, if it is specified in the PUT request.
     #[clap(long, action)]
     #[serde(default)]
     pub keep: bool,
@@ -654,41 +656,27 @@ pub struct BurnSelection {
 }
 
 impl BurnSelection {
-    pub(crate) fn exactly_one_provided(&self) -> Result<()> {
-        match (self.object_ids.is_empty(), self.all, self.all_expired) {
-            (false, false, false) => Ok(()),
-            (true, true, false) => Ok(()),
-            (true, false, true) => Ok(()),
-            _ => Err(anyhow!(
-                "exactly one of `object-ids`, `all`, or `all-expired` must be specified"
-            )),
-        }
-    }
-
     pub(crate) async fn get_object_ids(
         &self,
         client: &SuiContractClient,
     ) -> anyhow::Result<Vec<ObjectID>> {
-        self.exactly_one_provided()?;
-
-        if !self.object_ids.is_empty() {
-            Ok(self.object_ids.clone())
-        } else if self.all {
-            Ok(client
+        match (self.object_ids.is_empty(), self.all, self.all_expired) {
+            (false, false, false) => Ok(self.object_ids.clone()),
+            (true, true, false) => Ok(client
                 .owned_blobs(None, ExpirySelectionPolicy::All)
                 .await?
                 .into_iter()
                 .map(|blob| blob.id)
-                .collect::<Vec<_>>())
-        } else if self.all_expired {
-            Ok(client
+                .collect::<Vec<_>>()),
+            (true, false, true) => Ok(client
                 .owned_blobs(None, ExpirySelectionPolicy::Expired)
                 .await?
                 .into_iter()
                 .map(|blob| blob.id)
-                .collect::<Vec<_>>())
-        } else {
-            unreachable!("we have already checked that exactly one is provided")
+                .collect::<Vec<_>>()),
+            _ => Err(anyhow!(
+                "exactly one of `objectIds`, `all`, or `allExpired` must be specified"
+            )),
         }
     }
 }
--- a/crates/walrus-service/src/client/cli/runner.rs
+++ b/crates/walrus-service/src/client/cli/runner.rs
@@ -441,7 +441,10 @@ impl ClientCommandRunner {
             .new_contract_client(self.wallet?, self.gas_budget)
             .await?;
         let blobs = contract_client
-            .owned_blobs(None, ExpirySelectionPolicy::from_flag(include_expired))
+            .owned_blobs(
+                None,
+                ExpirySelectionPolicy::from_include_expired_flag(include_expired),
+            )
             .await?;
         blobs.print_output(self.json)
     }
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -173,9 +173,9 @@ pub enum ExpirySelectionPolicy {
 
 impl ExpirySelectionPolicy {
     /// Returns the policy for a give `include_expired` flag.
-    pub fn from_flag(include_expired: bool) -> Self {
+    pub fn from_include_expired_flag(include_expired: bool) -> Self {
         if include_expired {
-            Self::Expired
+            Self::All
         } else {
             Self::Valid
         }


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -327,6 +327,27 @@ pub enum CliCommands {
...
+    /// Walrus and _without_ refunding the storage. Running this command will result in the loss of
+    /// control over the Blob objects and the data they represent. Importantly, after burning:
+    ///
+    /// - Permanent blobs cannot be extended;
+    /// - deletable blobs cannot be extended nor deleted.
+    BurnBlobs {
+        /// The object IDs of the Blob objects to burn.
+        #[clap(num_args=1.., required=true)]
+        #[serde_as(as = "Vec<DisplayFromStr>")]
+        object_ids: Vec<ObjectID>,
```
mlegner: Suggestion: Can we accept something like `--all` and `--all-expired` as alternatives?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -630,6 +634,39 @@ impl ClientCommandRunner {
...
+        self,
+        object_ids: Vec<ObjectID>,
+        confirmation: UserConfirmation,
+    ) -> Result<()> {
+        if confirmation.is_required() {
+            let object_list = object_ids
+                .iter()
+                .map(|id| id.to_string())
+                .collect::<Vec<_>>()
+                .join("\n");
```
mlegner: Nit: If you `use itertools::Itertools as _;`, you can directly join the iterator:
```suggestion
                .join("\n");
```

crates/walrus-sui/src/client.rs
```diff
@@ -760,6 +766,21 @@ impl SuiContractClient {
...
+    /// Burns the blob objects with the given object IDs.
+    pub async fn burn_blobs(&self, blob_object_ids: &[ObjectID]) -> SuiClientResult<()> {
+        tracing::debug!(?blob_object_ids, "burning blobs");
+        let mut pt_builder = self.transaction_builder();
+
+        // Lock the wallet here to ensure there are no race conditions with object references.
+        let wallet = self.wallet().await;
+        for object_id in blob_object_ids.iter() {
+            pt_builder.burn_blob(object_id.into()).await?;
+        }
```
mlegner: Important: I guess there is a limit on the number of calls, right? So if `blob_object_ids.len()` is above that, we need to use multiple transactions, correct?

crates/walrus-service/src/client/daemon.rs
```diff
@@ -71,9 +89,33 @@ impl WalrusWriteClient for Client<SuiContractClient> {
...
+                    sui_client.sign_and_send_ptb(&wallet, ptb, None).await?;
+                }
+                PostStoreAction::Burn => self
+                    .sui_client()
+                    .burn_blobs(&[blob_object.id])
+                    .await
+                    .map_err(ClientError::other)?,
+                PostStoreAction::Keep => (),
+            }
+        };
```
mlegner: Suggestion: Integrate these contract calls into the certification PTB.

crates/walrus-service/src/client/daemon.rs
```diff
@@ -31,11 +32,25 @@ use walrus_core::{encoding::Primary, BlobId, EpochCount};
...
 
+/// The action to be performed for newly-created blobs.
+#[derive(Debug, Clone)]
+pub enum PostStoreAction {
+    /// Burn the blob object.
+    Burn,
+    /// Transfer the blob object to the given address.
+    TransferTo(SuiAddress),
+    /// Keep the blob object in the wallet that created it.
+    Keep,
```
mlegner: Question: Is this actually used? I guess we could add a CLI option to the publisher to keep objects instead of burning them.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -196,6 +205,8 @@ pub(super) struct PublisherQuery {
     force: bool,
     #[serde(default)]
     deletable: bool,
+    #[serde(default)]
+    return_to: Option<SuiAddress>,
```
mlegner: Remark: I'm not entirely convinced by the naming of this option because it's not clear what is returned; how about something like `send_object_to` or `object_recipient`? 

Also, can you add a docstring to make the generated OpenAPI docs more useful? Maybe also for the other parameters where we neglected that so far. On `force` we could add a `#[deprecated]` attribute instead.

crates/walrus-sui/src/client.rs
karlwuest: The limit is 1024. That's actually a good point that also applies more generally (although it's less likely to occur for other commands). Maybe we could limit this in the transaction builder and return a `TooManyTransactions` error if the limit is about to be exceeded, s.t. we can automatically stop adding more.

crates/walrus-service/src/client/daemon.rs
giac-mysten: yes, now that the calls to burn/transfer are integrated in the certification ptb, the standard CLI `store` command uses `keep`.

>I guess we could add a CLI option to the publisher to keep objects instead of burning them.

the nicer way would probably be to have the sub-wallets send the blob objects to the main publisher wallet, wdyt?

mlegner: > the nicer way would probably be to have the sub-wallets send the blob objects to the main publisher wallet, wdyt?

Yes, that would be perfect!

crates/walrus-sui/src/client.rs
giac-mysten: good point, I now ensure that multiple ptbs in sequence are used in that case.

crates/walrus-sui/src/client.rs
```diff
@@ -160,6 +160,37 @@ pub enum BlobPersistence {
...
+    /// Select only valid (non-expired) objects.
+    Valid,
+}
+
+impl ExpirySelectionPolicy {
+    /// Returns the policy for a give `include_expired` flag.
+    pub fn from_flag(include_expired: bool) -> Self {
+        if include_expired {
+            Self::Expired
+        } else {
```
mlegner: Important: I think this should be `Self::All`. Also, the function name doesn't specify which flag this is for; this could be an `include_expired` flag or an `only_expired` one.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -450,6 +455,13 @@ pub struct PublisherArgs {
     #[clap(long, default_value_t = default::sub_wallets_min_balance())]
     #[serde(default = "default::sub_wallets_min_balance")]
     pub sub_wallets_min_balance: u64,
+    /// If set, the publisher will keep the created Blob object ids in the wallet.
```
```suggestion
    /// If set, the publisher will keep the created Blob objects in the wallet.
```
Also, two questions: Does it keep them in its main wallet? And does it still support the `send_object_to` parameter?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -621,6 +633,66 @@ impl FileOrBlobIdOrObjectId {
...
+}
+
+impl BurnSelection {
+    pub(crate) fn exactly_one_provided(&self) -> Result<()> {
+        match (self.object_ids.is_empty(), self.all, self.all_expired) {
+            (false, false, false) => Ok(()),
+            (true, true, false) => Ok(()),
+            (true, false, true) => Ok(()),
+            _ => Err(anyhow!(
+                "exactly one of `object-ids`, `all`, or `all-expired` must be specified"
```
mlegner: Remark: AFAIU, this function is only relevant for the JSON mode as this is automatically checked by the clap, right? In this case, using camelCase would be more appropriate.

mlegner: Also, this could potentially get integrated into the `get_object_ids` function because we essentially have the same case distinction there.

crates/walrus-service/src/client/cli/args.rs
giac-mysten: > Does it keep them in its main wallet? 

yes, the objects are sent to the main wallet with the certification ptb

> And does it still support the send_object_to parameter?

yes, the send_object_to parameter overrides burn/keep

Will specify better in the docstring.

As the pr description says:

> It also changes the behavior of the publisher regarding newly-stored blobs:
> - there is now a `send_object_to` query keyword to the `store` endpoint to allow users to specify to which sui address the resulting `Blob` object should be sent to 
> - if no `send_object_to` is specified in the query:
>    - if the publisher is run with the `--keep` flag, the publisher will send the created `Blob` object from the sub-wallet to the main wallet (for safekeeping)
>    - if the `--keep` flag is not specified, the new `Blob` object is `burn`t.

crates/walrus-sui/src/client.rs
giac-mysten: mmm, good point!

crates/walrus-service/src/client/cli/args.rs
giac-mysten: i wanted to have a standalone function to make this check, but i guess it's not needed anywhere else.

PR #1252, opened 2024-12-03
test(staking): rounding errors + hardening

damirka:
## Description

- Closes WAL-411
- Closes WAL-229

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/system/blob.move
+++ b/contracts/walrus/sources/system/blob.move
@@ -282,6 +282,7 @@ public fun add_metadata(self: &mut Blob, metadata: Metadata) {
 ///
 /// Aborts if the metadata does not exist.
 public fun take_metadata(self: &mut Blob): Metadata {
+    assert!(dynamic_field::exists_(&self.id, METADATA_DF), EMissingMetadata);
     dynamic_field::remove(&mut self.id, METADATA_DF)
 }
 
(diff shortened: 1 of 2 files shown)


contracts/walrus/tests/staking/staking_pool/pool_rounding_tests.move
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+module walrus::pool_rounding_tests;
+
+use walrus::test_utils::{mint_balance, pool, context_runner, assert_eq};
+
+#[test, expected_failure(abort_code = ::walrus::staking_pool::EPoolNotEmpty)]
+// Failure reason: rewards left in the pool after all stakes are withdrawn
```
karlwuest, line 9: Question: We cannot remove this dust by e.g. paying it out to the storage node, because we don't know if it's dust or rewards that still need to be paid out, right? The reason being that we account the pool token when withdrawal is requested and not when it is actually performed.

Remark: I think that's fine because there is no strict need to be able to delete a pool after it loses all its stake. It would be a nice-to-have, but it's definitely not needed to go to mainnet.

Could you add a comment to explain this and create an issue for "later/optional" to look into this after mainnet?

damirka: Created an issue!

PR #1254, opened 2024-12-03
feat: use type origin map to get correct types when upgrading package

karlwuest:
## Description

Allow specifying a Walrus package ID (e.g. for package upgrades) and use the package to get the correct fully qualified types for all move structs defined in the Walrus package.

This currently still assumes that all packages (`wal`, `wal_exchange`, and `walrus`) are published as a single package. I will separate this in a follow up (see WAL-425).

Contributes to WAL-346

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/contracts.rs
+++ b/crates/walrus-sui/src/contracts.rs
@@ -138,7 +138,7 @@ impl<'a> StructTag<'a> {
         })
     }
 
-    /// Converts a [StructTag] to a [MoveStructTag] using the matching package ID from the given
+    /// Converts a [`StructTag`] to a [`MoveStructTag`] using the matching package ID from the given
     /// type origin map.
     pub(crate) fn to_move_struct_tag_with_type_map(
         &self,


crates/walrus-service/client_config_example.yaml
```diff
@@ -1,8 +1,9 @@
 system_object: 0xa2637d13d171b278eadfa8a3fbe8379b5e471e1f3739092e5243da17fc8090eb
 staking_object: 0xca7cf321e47a1fc9bfd032abc31b253f5063521fd5b4c431f2cdd3fee1b4ec00
+walrus_package: 0xa9b00f69d3b033e7b64acff2672b54fbb7c31361954251e235395dea8bd6dcac
 exchange_object:
-- 0xa9b00f69d3b033e7b64acff2672b54fbb7c31361954251e235395dea8bd6dcac
 - 0x26a8a417b553b18d13027c23e8016c3466b81e7083225436b55143c127f3c0cb
+- 0xaf3819c82de2e0257c0cc2177dfce6432efa42ca6c04c0b774dbb3c5ca2573cd
```
giac-mysten, line 6: question: is this change necessary?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -376,6 +398,86 @@ impl SuiReadClient {
             })
     }
 
+    /// Get the [`StorageNodeCap`] object associated with the address.
```
giac-mysten, line 401: Question: any specfic reason to move these functions  to methods on the read client? just cleanup?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -706,6 +808,30 @@ pub async fn get_system_package_id(
     Ok(object_pkg_id)
 }
 
+/// Gets the type origin map for a given package.
+async fn type_origin_map_for_package(
```
giac-mysten, line 812: nice!

crates/walrus-sui/src/contracts.rs
```diff
@@ -115,8 +118,10 @@ pub struct StructTag<'a> {
 }
 
 impl<'a> StructTag<'a> {
-    /// Returns a Move StructTag for the identified struct, within the published contract module.
-    pub fn to_move_struct_tag(
+    /// Returns a [MoveStructTag] for the identified struct with the given package ID.
```
giac-mysten:
```suggestion
    /// Returns a [`MoveStructTag`] for the identified struct with the given package ID.
```

crates/walrus-sui/src/contracts.rs
```diff
@@ -132,6 +137,19 @@ impl<'a> StructTag<'a> {
             type_params: type_params.into(),
         })
     }
+
+    /// Converts a [StructTag] to a [MoveStructTag] using the matching package ID from the given
```
giac-mysten:
```suggestion
    /// Converts a [`StructTag`] to a [`MoveStructTag`] using the matching package ID from the given
```

crates/walrus-service/client_config_example.yaml
karlwuest: That's automatically generated and I think it changed because of the order in which the random ObjectIDs are generated.

crates/walrus-sui/src/client/read_client.rs
karlwuest: Yes, mainly cleanup. I think, we used to use (some of) them separately during the setup, before we actually constructed a read client. But I think this is the right place for them (it also simplifies their signatures a bit).

crates/walrus-sui/src/client/read_client.rs
karlwuest: thanks!

PR #1255, opened 2024-12-03
feat(client)!: accept multiple file for store

joyqvq:
## Description
https://github.com/MystenLabs/walrus-sites/issues/169
## Test plan

bench result 
number of 10kb blobs | before | after | delta
-- | -- | -- | --
1 | 11.35269483 | 14.79881596 | 0.3035509343
2 | 23.45166217 | 16.69676471 | -0.2880349124
5 | 50.6061375 | 22.46064713 | -0.5561675276
10 | 118.9244417 | 34.11470796 | -0.7131396416
  |   |   |  
number of 100mb blobs | before | after | delta
1 | 30.22572454 | 34.70246792 | 0.1481103743
2 | 198.2747552 | 119.4213113 | -0.3976978502
5 | 496.7119502 | 194.706034 | -0.6080101679
10 | 907.32686125 | 563.0411232 | -0.3794506178


tested with cli

error if total blob size is too large (only applies if > 1 blob is uploaded)
```
target/release/walrus store ~/Desktop/old_random_file_9.bin ~/Desktop/old_random_file_10.bin ~/Desktop/old_random_file_11.bin ~/Desktop/old_random_file_12.bin ~/Desktop/old_random_file_13.bin ~/Desktop/old_random_file_14.bin ~/Desktop/old_random_file_15.bin ~/Desktop/old_random_file_16.bin ~/Desktop/old_random_file_17.bin ~/Desktop/old_random_file_18.bin ~/Desktop/old_random_file_4.bin

2024-12-18T15:41:23.633340Z  INFO walrus: client version: 1.7.0-d63f9be124f4-dirty
2024-12-18T15:41:23.634368Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2024-12-18T15:41:23.635361Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/.sui/sui_config/client.yaml'
Warning: The number of epochs for which to store the blob was not specified. Using the default value of 1 epoch. Use `--epochs` to store for a longer period.
2024-12-18T15:41:25.393767Z  INFO walrus_service::client::cli::runner: storing 11 files as blobs on Walrus
Error: client internal error: total blob size 1153433600 exceeds the maximum limit of 1073741824
```

cli output with path printed
```
target/release/walrus store ~/Desktop/old_random_file_9.bin ~/Desktop/old_random_file_10.bin ~/Desktop/old_random_file_11.bin ~/Desktop/old_random_file_12.bin ~/Desktop/old_random_file_13.bin ~/Desktop/old_random_file_14.bin ~/Desktop/old_random_file_15.bin ~/Desktop/old_random_file_16.bin ~/Desktop/old_random_file_17.bin ~/Desktop/old_random_file_18.bin
2024-12-18T15:55:48.692623Z  INFO walrus: client version: 1.7.0-d63f9be124f4-dirty
2024-12-18T15:55:48.693734Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2024-12-18T15:55:48.695468Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/.sui/sui_config/client.yaml'
Warning: The number of epochs for which to store the blob was not specified. Using the default value of 1 epoch. Use `--epochs` to store for a longer period.
2024-12-18T15:55:50.394852Z  INFO walrus_service::client::cli::runner: storing 10 files as blobs on Walrus
 ○ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:55:52.319773Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.719286166s
 ◉ blob encoded; blob ID: JgBsmO7Rux9nzYVXBs8VqwPxuUDz5e6jnVCEI38_czY [00:00:01]
 ◎ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:55:53.950550Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.630525125s
 ◉ blob encoded; blob ID: TbwUrmC9xvQC5hLeUU7Vh29K9GLNw4EXXNNto8YGrlw [00:00:01]
 ◎ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:55:55.575388Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.624610667s
 ◉ blob encoded; blob ID: 5b0pWtVRrW6x7F3MM6lqmKpZhnGByOUGWaw7oTdmr8o [00:00:01]
 ◎ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:55:57.231215Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.65561525s
 ◉ blob encoded; blob ID: seY45E4VBTnLsAeCeK9dSPXVxmOXuFEFgzxbH3hTnRA [00:00:01]
 ◎ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:55:58.868500Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.637081959s
 ◉ blob encoded; blob ID: zmtuesYjC3eM7F5EKbP9uuiWaUeCxxSPgqx197b2Zhs [00:00:01]
 ○ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:56:00.580398Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.711719583s
 ◉ blob encoded; blob ID: DeTan4_GyWHnoDoXvQ6LqU2-MiXuKUPjDhfXK2Cr_v4 [00:00:01]
 ○ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:56:02.299908Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.7193195s
 ◉ blob encoded; blob ID: BwNKnYCbdDokA1X6DZ-RvvXkvdz2kyqENDzlZMmGzZQ [00:00:01]
 ◎ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:56:03.959659Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.659541042s
 ◉ blob encoded; blob ID: V6HPv2R_iV71VMT3WGXmmoMg63OCRXTalpNtnKobu8k [00:00:01]
 ○ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:56:05.677051Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.717180375s
 ◉ blob encoded; blob ID: bclHwGC4zdsJhQfXxfCmzCJBxQGAhaZwzxtOl_hRNow [00:00:01]
 ○ encoding the blob [00:00:01]                                                                                                                                                                                                                                                                                                                                                            2024-12-18T15:56:07.403046Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=482 primary_sliver_size=319084 secondary_sliver_size=158578 duration=1.725797375s
 ◉ blob encoded; blob ID: c35joezV0FaWmGf_ORFiaCoPRj_w05OHj6YtdCrKI-I [00:00:01]                                                                                                                                                                                                                                                                                                           2024-12-18T15:56:07.405274Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded files to metadata: V6HPv2R_iV71VMT3WGXmmoMg63OCRXTalpNtnKobu8k -> /Users/joy/Desktop/old_random_file_16.bin, seY45E4VBTnLsAeCeK9dSPXVxmOXuFEFgzxbH3hTnRA -> /Users/joy/Desktop/old_random_file_12.bin, zmtuesYjC3eM7F5EKbP9uuiWaUeCxxSPgqx197b2Zhs -> /Users/joy/Desktop/old_random_file_13.bin, 5b0pWtVRrW6x7F3MM6lqmKpZhnGByOUGWaw7oTdmr8o -> /Users/joy/Desktop/old_random_file_11.bin, TbwUrmC9xvQC5hLeUU7Vh29K9GLNw4EXXNNto8YGrlw -> /Users/joy/Desktop/old_random_file_10.bin, DeTan4_GyWHnoDoXvQ6LqU2-MiXuKUPjDhfXK2Cr_v4 -> /Users/joy/Desktop/old_random_file_14.bin, BwNKnYCbdDokA1X6DZ-RvvXkvdz2kyqENDzlZMmGzZQ -> /Users/joy/Desktop/old_random_file_15.bin, bclHwGC4zdsJhQfXxfCmzCJBxQGAhaZwzxtOl_hRNow -> /Users/joy/Desktop/old_random_file_17.bin, c35joezV0FaWmGf_ORFiaCoPRj_w05OHj6YtdCrKI-I -> /Users/joy/Desktop/old_random_file_18.bin, JgBsmO7Rux9nzYVXBs8VqwPxuUDz5e6jnVCEI38_czY -> /Users/joy/Desktop/old_random_file_9.bin
2024-12-18T15:56:07.405342Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: storing 10 sliver pairs with metadata
2024-12-18T15:56:07.946788Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: retrieved 10 blobs statuses duration=541.37875ms
2024-12-18T15:56:28.703480Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: 10 blobs resources obtained duration=20.756606709s
2024-12-18T15:56:28.906141Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=V6HPv2R_iV71VMT3WGXmmoMg63OCRXTalpNtnKobu8k
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.926729Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=seY45E4VBTnLsAeCeK9dSPXVxmOXuFEFgzxbH3hTnRA
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.933331Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=bclHwGC4zdsJhQfXxfCmzCJBxQGAhaZwzxtOl_hRNow
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.940556Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=DeTan4_GyWHnoDoXvQ6LqU2-MiXuKUPjDhfXK2Cr_v4
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.947998Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=5b0pWtVRrW6x7F3MM6lqmKpZhnGByOUGWaw7oTdmr8o
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.955657Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=JgBsmO7Rux9nzYVXBs8VqwPxuUDz5e6jnVCEI38_czY
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.962773Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=zmtuesYjC3eM7F5EKbP9uuiWaUeCxxSPgqx197b2Zhs
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.968711Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=BwNKnYCbdDokA1X6DZ-RvvXkvdz2kyqENDzlZMmGzZQ
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.974257Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=c35joezV0FaWmGf_ORFiaCoPRj_w05OHj6YtdCrKI-I
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-18T15:56:28.979225Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=TbwUrmC9xvQC5hLeUU7Vh29K9GLNw4EXXNNto8YGrlw
 ◉ slivers sent [00:02:04] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ slivers sent [00:02:04] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ slivers sent [00:02:30] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◎ sending slivers [00:03:07] [#####################################################################################################################################################################################################################################>--------------------------------------------------------------------------------------------------------] 459/667 (2y)2024-12-18T15:59:36.294306Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=BwNKnYCbdDokA1X6DZ-RvvXkvdz2kyqENDzlZMmGzZQ duration=187.325873416s blob_size=104857600
 ◉ additional slivers stored [00:01:02]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T15:59:36.685324Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=TbwUrmC9xvQC5hLeUU7Vh29K9GLNw4EXXNNto8YGrlw duration=187.706386709s blob_size=104857600
 ◉ additional slivers stored [00:01:15]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T16:00:15.464735Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=c35joezV0FaWmGf_ORFiaCoPRj_w05OHj6YtdCrKI-I duration=226.490821958s blob_size=104857600
 ◉ slivers sent [00:03:51] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ slivers sent [00:03:58] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ slivers sent [00:04:09] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ additional slivers stored [00:01:56]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T16:02:16.616321Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=seY45E4VBTnLsAeCeK9dSPXVxmOXuFEFgzxbH3hTnRA duration=347.690129084s blob_size=104857600
 ◉ additional slivers stored [00:01:59]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T16:02:27.258872Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=V6HPv2R_iV71VMT3WGXmmoMg63OCRXTalpNtnKobu8k duration=358.353327583s blob_size=104857600
 ◉ additional slivers stored [00:02:05]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T16:02:44.288904Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=DeTan4_GyWHnoDoXvQ6LqU2-MiXuKUPjDhfXK2Cr_v4 duration=375.348923167s blob_size=104857600
 ◉ slivers sent [00:06:27] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ slivers sent [00:06:57] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ slivers sent [00:07:08] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ slivers sent [00:07:12] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ additional slivers stored [00:03:14]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T16:06:10.242015Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=bclHwGC4zdsJhQfXxfCmzCJBxQGAhaZwzxtOl_hRNow duration=581.309573584s blob_size=104857600
 ◉ additional slivers stored [00:03:29]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T16:06:55.845942Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=5b0pWtVRrW6x7F3MM6lqmKpZhnGByOUGWaw7oTdmr8o duration=626.898906459s blob_size=104857600
 ◉ additional slivers stored [00:03:34]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T16:07:12.457984Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=JgBsmO7Rux9nzYVXBs8VqwPxuUDz5e6jnVCEI38_czY duration=643.503319958s blob_size=104857600
 ◉ additional slivers stored [00:03:36]                                                                                                                                                                                                                                                                                                                                                    2024-12-18T16:07:17.949498Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=zmtuesYjC3eM7F5EKbP9uuiWaUeCxxSPgqx197b2Zhs duration=648.987722125s blob_size=104857600
2024-12-18T16:07:17.949556Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: get 10 blobs certificates duration=649.246953333s
2024-12-18T16:07:20.268471Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: certified 10 blobs on Sui duration=2.318894083s
2024-12-18T16:07:20.709093Z  INFO walrus_service::client::cli::runner: 10 out of 10 blobs stored duration=690.315266625s
Path: /Users/joy/Desktop/old_random_file_16.bin
Blob ID: V6HPv2R_iV71VMT3WGXmmoMg63OCRXTalpNtnKobu8k
Sui object ID: 0x0faeb8ff6339fbafb9597c5d323ad24054ae3037f5ee59d1bead6f9250f4a914
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_12.bin
Blob ID: seY45E4VBTnLsAeCeK9dSPXVxmOXuFEFgzxbH3hTnRA
Sui object ID: 0x212344ae747fe5a48c7ee0c870cd306fc9df54cb3d3203f714976c83c8b43738
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_17.bin
Blob ID: bclHwGC4zdsJhQfXxfCmzCJBxQGAhaZwzxtOl_hRNow
Sui object ID: 0x23064719ce369a107fa851b9fd17520ff6ec43b8d03bda33fa32b12447562adc
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_14.bin
Blob ID: DeTan4_GyWHnoDoXvQ6LqU2-MiXuKUPjDhfXK2Cr_v4
Sui object ID: 0x376579aa3416b282b5bb5714bb870420a64c4a22862b51054b12ddf15aec1bbb
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_11.bin
Blob ID: 5b0pWtVRrW6x7F3MM6lqmKpZhnGByOUGWaw7oTdmr8o
Sui object ID: 0xa40d8ada54697d089387ff83bf9643ca8c97bc8bd3bd2a43331e94d5ce1847d0
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_9.bin
Blob ID: JgBsmO7Rux9nzYVXBs8VqwPxuUDz5e6jnVCEI38_czY
Sui object ID: 0xb1d48ef65866fcf57e3a5b28bad271e4d499416fe6d23b906bc86ce0ffeedc2e
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_13.bin
Blob ID: zmtuesYjC3eM7F5EKbP9uuiWaUeCxxSPgqx197b2Zhs
Sui object ID: 0xb9010c56b2bccda4248276c36b4432934163b8379eb404f322a254e394dea0d1
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_15.bin
Blob ID: BwNKnYCbdDokA1X6DZ-RvvXkvdz2kyqENDzlZMmGzZQ
Sui object ID: 0xbc77a2fbcdb3ab23bf25617349b62a67a2488d50c7d85538dc5620bdf0309475
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_18.bin
Blob ID: c35joezV0FaWmGf_ORFiaCoPRj_w05OHj6YtdCrKI-I
Sui object ID: 0xc4f22622cc9ac54e456d5936457b832a26f5130bbcf1318a0b570d3e2a653e26
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)

Path: /Users/joy/Desktop/old_random_file_10.bin
Blob ID: TbwUrmC9xvQC5hLeUU7Vh29K9GLNw4EXXNNto8YGrlw
Sui object ID: 0xdfa43d798474590fd7717c16a3fc6c2dd429f65be5f2a129a1de9de4542f3ea7
Unencoded size: 100 MiB
Encoded size (including replicated metadata): 517 MiB
Cost (excluding gas): 0.0011 WAL (storage was purchased, and a new blob object was registered)
```
example bulk certify blob ptb: https://suiscan.xyz/testnet/tx/J5ATdF2fFNNxUWY4Mm5w6cHfvP5UsmA5wpVVXZhFR7Ys
example bulk reserve + register ptb: https://suiscan.xyz/testnet/tx/EJb8BLWo8WdXc91PKcXXrDjqxQWEqNjfLRYcxPcuyzHp

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Allow multiple files to be uploaded with the store command

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -664,11 +664,9 @@ impl Client<SuiContractClient> {
         >,
     > {
         // Filters out all the failed blobs that cannot fetch status and continue.
-        let failed_blobs = Arc::new(RwLock::new(Vec::with_capacity(pairs_and_metadata.len())));
         let blob_id_to_metadata_with_status: HashMap<BlobId, (_, _, _)> =
-            futures::future::join_all(pairs_and_metadata.iter().map(|(pair, metadata)| {
-                let failed_blobs_write = failed_blobs.clone();
-                async move {
+            futures::future::join_all(pairs_and_metadata.iter().map(
+                |(pair, metadata)| async move {
                     let blob_id = *metadata.blob_id();
                     match async {
                         self.check_blob_id(&blob_id)?;
@@ -682,13 +680,11 @@ impl Client<SuiContractClient> {
                         Ok(result) => Some(result),
                         Err(e) => {
                             tracing::warn!("Failed to process blob {}: {}", blob_id, e);
-                            let mut guard = failed_blobs_write.write().await;
-                            guard.push((blob_id, e));
                             None
                         }
                     }
-                }
-            }))
+                },
+            ))
             .await
             .into_iter()
             .flatten()
--- a/crates/walrus-service/src/client/cli/runner.rs
+++ b/crates/walrus-service/src/client/cli/runner.rs
@@ -360,7 +360,7 @@ impl ClientCommandRunner {
                 .into_iter()
                 .map(|file| read_blob_from_file(&file).map(|blob| (file, blob)))
                 .collect::<Result<Vec<(PathBuf, Vec<u8>)>>>()?;
-            let result = client
+            let results = client
                 .reserve_and_store_blobs_retry_epoch_with_path(
                     &blobs,
                     epochs,
@@ -369,13 +369,26 @@ impl ClientCommandRunner {
                     PostStoreAction::Keep,
                 )
                 .await?;
+            let blobs_len = blobs.len();
+            if results.len() != blobs_len {
+                let original_paths: Vec<_> = blobs.into_iter().map(|(path, _)| path).collect();
+                let not_stored = results
+                    .iter()
+                    .filter(|blob| !original_paths.contains(&blob.path))
+                    .map(|blob| blob.blob_store_result.blob_id())
+                    .collect::<Vec<_>>();
+                tracing::warn!(
+                    "some blobs ({}) are not stored",
+                    not_stored.into_iter().join(", ")
+                );
+            }
             tracing::info!(
                 duration = ?start_timer.elapsed(),
                 "{} out of {} blobs stored",
-                result.len(),
-                blobs.len()
+                results.len(),
+                blobs_len
             );
-            result.print_output(self.json)
+            results.print_output(self.json)
         }
     }
 


crates/walrus-service/src/client.rs
```diff
@@ -398,43 +410,74 @@ impl Client<SuiContractClient> {
...
         persistence: BlobPersistence,
-    ) -> ClientResult<BlobStoreResult> {
-        let blob_id = *metadata.blob_id();
-        self.check_blob_id(&blob_id)?;
-        tracing::Span::current().record("blob_id", blob_id.to_string());
+    ) -> ClientResult<Vec<BlobStoreResult>> {
+        // todo: figure out what span for this trace now
+        // tracing::Span::current().record("blob_id", blob_id.to_string());
 
+        // fetch status for each blob, then build mapping: blob id -> (pair, metadata, status)
```
giac-mysten: Nit(here and in other places): our code comment convention requires capitalization and a final `.`
```suggestion
        // Fetch status for each blob, then build mapping: blob id -> (pair, metadata, status).
```

crates/walrus-service/src/client.rs
```diff
@@ -320,21 +320,29 @@ impl Client<SuiContractClient> {
             .await)
     }
 
-    /// Stores the blob to Walrus, retrying if it fails because of epoch change.
+    /// Stores the list of blobs to Walrus, retrying if it fails because of epoch change.
     #[tracing::instrument(skip_all, fields(blob_id))]
     pub async fn reserve_and_store_blob_retry_epoch(
```
giac-mysten: Nit: here and in other function names
```suggestion
    pub async fn reserve_and_store_blobs_retry_epoch(
```

crates/walrus-service/src/client.rs
```diff
@@ -320,21 +320,29 @@ impl Client<SuiContractClient> {
...
         epochs_ahead: EpochCount,
         store_when: StoreWhen,
         persistence: BlobPersistence,
-    ) -> ClientResult<BlobStoreResult> {
-        let (pairs, metadata) = self.encode_pairs_and_metadata(blob).await?;
+    ) -> ClientResult<Vec<BlobStoreResult>> {
+        if blobs.is_empty() {
+            return Ok(Vec::new());
+        }
+        // todo: should any max batch size limit be imposed in addition to blob size?
```
giac-mysten: Remark: if the list of blobs is long, the client risks running out of memory here (each blob requires 5-6x it's original size in memory).

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -318,36 +318,45 @@ impl ClientCommandRunner {
...
-                .reserve_and_store_blob_retry_epoch(
-                    &read_blob_from_file(&file)?,
-                    epochs,
-                    store_when,
-                    persistence,
-                )
+                .reserve_and_store_blob_retry_epoch(&blobs, epochs, store_when, persistence)
                 .await?;
-            result.print_output(self.json)
+            result[0].print_output(self.json)
```
giac-mysten: Question: why print only the first result?

crates/walrus-service/src/client/resource.rs
```diff
@@ -123,56 +123,62 @@ impl<'a> ResourceManager<'a> {
...
-        if !store_when.is_store_always() && !persistence.is_deletable() {
-            if let Some(result) =
-                self.blob_status_to_store_result(*metadata.blob_id(), epochs_ahead, blob_status)
-            {
-                return Ok(StoreOp::NoOp(result));
-            }
-        };
+    ) -> ClientResult<Vec<StoreOp>> {
+        let mut results = Vec::with_capacity(metadata_with_status.len());
+        for (metadata, blob_status) in metadata_with_status {
```
giac-mysten: Suggestion: separate the inner part of the for loop to a function

crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -102,10 +102,15 @@ impl ClientMultiplexer {
...
+            .reserve_and_store_blob_retry_epoch(
+                &[blob.to_vec()],
+                epochs_ahead,
+                store_when,
+                persistence,
+            )
             .await?;
 
-        Ok(result)
+        Ok(result.into_iter().next().unwrap())
```
giac-mysten: Nit: as a general rule, we never use `unwrap` in this codebase, and if needed use instead:
```suggestion
        Ok(result.into_iter().next().expect("there is exactly one element"))
```

Nit: elsewhere, you returned `result[0]`. Use either form consistently.

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -56,6 +56,20 @@ pub trait CliOutput: Serialize {
         }
         Ok(())
     }
+    /// Writes a list of outputs to stdout formatted depending on the output mode.
+    fn print_output_list(outputs: &[Self], json: bool) -> Result<()>
```
giac-mysten: Question: where is this used?

Remark: this will not work very well, because it will, for example, print "success" multiple times when storing several blobs. 
Suggestion: I think we should create change the `BlobStoreResult` to possibly collect multiple results and print them nicely.

crates/walrus-service/src/client.rs
```diff
@@ -469,28 +512,42 @@ impl Client<SuiContractClient> {
...
+                        certified_epoch: Some(write_committee_epoch),
+                        ..blob.clone()
+                    },
+                    resource_operation: resource_operation.clone(),
+                    cost,
+                }
+            })
+            .collect();
+
+        // todo: check if order matters here
```
giac-mysten: Remark: the `BlobStoreResult` has no information on which file generated it. 
Suggestion: either:
1. maintain the order, so that the caller can understand what file generated what blob ID.
2. (preferred) add info on the file that generated the result  to the `BlobStoreResult`.

crates/walrus-service/src/client/resource.rs
```diff
@@ -123,56 +123,62 @@ impl<'a> ResourceManager<'a> {
...
+                    results.push(StoreOp::NoOp(result));
+                    continue;
+                }
+            };
 
-        let (blob, op) = self
-            .get_existing_or_register(metadata, epochs_ahead, persistence, store_when)
-            .await?;
+            let (blob, op) = self
+                .get_existing_or_register(metadata, epochs_ahead, persistence, store_when)
```
giac-mysten: Question/Suggestion: It would be nice if all the registrations could use the same PTB

crates/walrus-service/src/client.rs
```diff
@@ -398,43 +410,74 @@ impl Client<SuiContractClient> {
...
-            let committees = self.committees.read().await;
+        // return early if all operations were noops
+        if new_blobs_and_ops.is_empty() {
+            return Ok(noop_results);
+        }
+
+        // get certificates for all new blobs
+        let committees = self.committees.read().await;
+        let mut blobs_with_certificates = Vec::with_capacity(new_blobs_and_ops.len());
+        for (blob_object, resource_operation) in new_blobs_and_ops.iter() {
```
giac-mysten: Remark/Question: the upload is still sequential, right? do I understand correctly that the only speedup this implementation would bring over calling the client repeatedly is that the certification call to sui happens concurrently, rather than sequentially?

crates/walrus-service/src/client.rs
```diff
@@ -398,43 +410,74 @@ impl Client<SuiContractClient> {
...
+        // todo: figure out what span for this trace now
+        // tracing::Span::current().record("blob_id", blob_id.to_string());
 
+        // fetch status for each blob, then build mapping: blob id -> (pair, metadata, status)
         let status_start_timer = Instant::now();
-        let blob_status = self
-            .get_blob_status_with_retries(&blob_id, &self.sui_client)
-            .await?;
+        let metadata_with_status: HashMap<BlobId, (_, _, _)> =
+            futures::future::try_join_all(pairs_and_metadata.iter().map(
```
gdanezis: Here if any of the requests fail for some reason, the whole thing will fail, and abort. Is this what we want? The more blobs the less reliable this may become?

crates/walrus-service/src/client.rs
```diff
@@ -398,43 +410,74 @@ impl Client<SuiContractClient> {
...
 
-        let (mut blob_object, resource_operation) = match store_operation {
-            StoreOp::NoOp(result) => return Ok(result),
-            StoreOp::RegisterNew { blob, operation } => (blob, operation),
-        };
+        // collect store ops for noops and new blobs
+        let mut noop_results: Vec<_> = Vec::new();
+        let mut new_blobs_and_ops: Vec<_> = Vec::new();
+        store_operations.iter().for_each(|store_op| match store_op {
+            StoreOp::NoOp(result) => noop_results.push(result.clone()),
```
gdanezis: Is the clone needed?

crates/walrus-service/src/client.rs
joyqvq: are ther cases that this may be desirable to be atomic, maybe it should be a strategy option 1) all or nothing 2) or keep retrying with some backed off batch size?

one way we can optimize, maybe at the retry layer, to batch them with a reasonable total batch size, with committee refresh check. (also addressing to this issue https://github.com/MystenLabs/walrus/pull/1255#issuecomment-2518074405 @giac-mysten )

joyqvq: revised this to collect and filter only the successful ones and continue, return early on fails

crates/walrus-service/src/client.rs
joyqvq: i set a total max blob size (len * size) as 2gb (somewhat arbitrarily but conservatively, lmk if this is a good limit)

crates/walrus-service/src/client/cli/runner.rs
joyqvq: revised to print the list!

crates/walrus-service/src/client/cli/cli_output.rs
joyqvq: removed in favor of a list print

crates/walrus-service/src/client/resource.rs
joyqvq: revised a bit here to first filter out the noop and continue - PTAL. imo its cleaner to not separate to keep the results more readable in this case

crates/walrus-service/src/client.rs
joyqvq: maintaining the order is a bit tricky here (since i built the results with filters on noop and collapse them into one list). attaching the raw outputs for the command currently, i think the output prints makes sense with blob id and info printed.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -118,39 +119,48 @@ async fn run_store_and_read_with_crash_failures(
...
+        read_blobs.len(),
+        "Number of blobs should match"
+    );
 
+    for original in original_blobs {
+        assert!(
+            read_blobs.iter().any(|read| read == original),
+            "Each original blob should have a matching read blob"
+        );
+    }
```
joyqvq, line 162: since the ordering is not preserved (noops and newly registered are preprocessed) we just compare the hashset equality

crates/walrus-service/src/client.rs
```diff
@@ -326,48 +339,84 @@ impl Client<SuiContractClient> {
...
     }
 
-    /// Stores the blob to Walrus, retrying if it fails because of epoch change.
+    /// Stores a list of blobs to Walrus, retrying if it fails because of epoch change.
     #[tracing::instrument(skip_all, fields(blob_id))]
-    pub async fn reserve_and_store_blob_retry_epoch(
+    pub async fn reserve_and_store_blobs_retry_epoch(
         &self,
-        blob: &[u8],
+        blobs: &[&[u8]],
```
joyqvq, line 344: since there are test files calling this without file path, we preserve this function and add `reserve_and_store_blobs_retry_epoch_with_path` below.

crates/walrus-service/src/client.rs
```diff
@@ -326,48 +339,84 @@ impl Client<SuiContractClient> {
...
         persistence: BlobPersistence,
         post_store: PostStoreAction,
-    ) -> ClientResult<BlobStoreResult> {
-        let (pairs, metadata) = self.encode_pairs_and_metadata(blob).await?;
+    ) -> ClientResult<Vec<BlobStoreResultWithPath>> {
+        let EncodedResult {
+            pairs_and_metadata,
+            id_to_path,
+        } = self
+            .encode_blobs_to_pairs_and_metadata_with_path(blobs_with_paths)
```
joyqvq, line 380: a design decision here is to save the mapping here blob_id -> file_path. then after we get vec<BlobStoreResult> we map it into Vec<BlobStoreResultWithPath>. imo its cleaner than propagating filepath down in the call graph.

crates/walrus-service/src/client.rs
```diff
@@ -376,6 +425,86 @@ impl Client<SuiContractClient> {
...
+    /// Returns the successful list of sliver and metadata and logs the indices of failed blobs.
+    pub async fn encode_blobs_to_pairs_and_metadata(
+        &self,
+        blobs: &[&[u8]],
+    ) -> ClientResult<Vec<(Vec<SliverPair>, VerifiedBlobMetadataWithId)>> {
+        if blobs.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        if blobs.len() > 1 {
```
joyqvq, line 467: we share the function here for store and multistore (its more concise and we don't need to add another command for single/multistore anymore)

crates/walrus-service/src/client/daemon.rs
```diff
@@ -80,16 +80,16 @@ impl WalrusWriteClient for Client<SuiContractClient> {
...
+                &[blob],
                 epochs_ahead,
                 store_when,
                 persistence,
                 post_store,
             )
             .await?;
 
-        Ok(result)
+        Ok(result.into_iter().next().expect("expected one blob"))
```
joyqvq: we only change the cli to accept multistore in this pr, not the daemon.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -278,48 +295,63 @@ async fn test_store_with_existing_blob_resource(
...
-        .encoding_config()
-        .get_blob_encoder(&blob)?
-        .encode_with_metadata();
-    let metadata = metadata.metadata().to_owned();
-    let blob_id = BlobId::from_sliver_pair_metadata(&metadata);
-    let metadata = VerifiedBlobMetadataWithId::new_verified_unchecked(blob_id, metadata);
-
-    // Register a new blob.
-    let (original_blob_object, _) = client
+    let blob_data = walrus_test_utils::random_data_list(31415, 10);
```
giac-mysten: Question: do we need 10? any insigth on how much this slows down the testst?

crates/walrus-service/src/client.rs
giac-mysten: Remark: i was leaning towards a separate command to avoid breaking changes to the store interface. but since the output interface is broken anyways, it's probably ok to break it here too.

crates/walrus-service/src/client/daemon.rs
giac-mysten:
```suggestion
        Ok(result.into_iter().next().expect("there is only one blob, as store was called with one blob"))
```

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -325,36 +325,56 @@ impl ClientCommandRunner {
...
                     store_when,
                     persistence,
                     PostStoreAction::Keep,
                 )
                 .await?;
+            tracing::info!(
+                duration = ?start_timer.elapsed(),
+                "{} out of {} blobs stored",
+                result.len(),
+                blobs.len()
```
giac-mysten: Question: is there ever a case where `result.len() != blobs.len()`?

crates/walrus-sui/src/utils.rs
```diff
@@ -176,6 +176,25 @@ where
     Ok(inner)
 }
 
+pub(crate) async fn get_sui_objects<U>(
```
giac-mysten: Remark: this will have to be kept in sync with [this pr](https://github.com/MystenLabs/walrus/pull/1281). (I can take care of this).

crates/walrus-sui/src/client.rs
```diff
@@ -334,104 +339,114 @@ impl SuiContractClient {
     ///
     /// `blob_size` is the size of the unencoded blob. The encoded size of the blob must be
     /// less than or equal to the size reserved in `storage`.
-    pub async fn register_blob(
+    pub async fn register_blobs(
         &self,
-        storage: &StorageResource,
-        blob_metadata: BlobObjectMetadata,
+        blob_metadata_and_storage: Vec<(BlobObjectMetadata, StorageResource)>,
```
giac-mysten, line 342: Question: Does `StorageResource` need to be passed by value?

crates/walrus-sui/src/client.rs
```diff
@@ -334,104 +339,114 @@ impl SuiContractClient {
...
+        let expected_num_blobs = blob_metadata_and_storage.len();
+        for (blob_metadata, storage) in blob_metadata_and_storage.into_iter() {
+            pt_builder
+                .register_blob(storage.id.into(), blob_metadata, persistence)
+                .await?;
+        }
         let (ptb, _sui_cost) = pt_builder.finish().await?;
         let res = self.sign_and_send_ptb(&wallet, ptb, None).await?;
-        let blob_obj_id = get_created_sui_object_ids_by_type(
+        let blob_obj_ids = get_created_sui_object_ids_by_type(
```
giac-mysten, line 358: Remark: If i remember correctly, the created objects in the effects are not in the order they were created, but just ordered lexicographically by object ID.

As far as i can see, no ordering assumption is used later on in the code, but just mentioning it to make sure :)

crates/walrus-service/src/client.rs
giac-mysten: Can we make it test-only then? (`#[cfg(test)]`)

crates/walrus-service/src/client.rs
giac-mysten: i like it!

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -57,25 +58,41 @@ pub trait CliOutput: Serialize {
...
+            BlobStoreResult::AlreadyCertified {
                 blob_id,
                 event_or_object,
                 end_epoch,
             } => {
                 println!(
                     "{} Blob was already available and certified within Walrus, \
                     for a sufficient number of epochs.\n\
-                    Blob ID: {}\n{event_or_object}\nExpiry epoch (exclusive): {}",
+                    Blob ID: {}\n{event_or_object}\nExpiry epoch (exclusive): {}\nPath: {}",
```
giac-mysten: Suggestion: add the path first, before the blob id.

crates/walrus-service/src/client.rs
```diff
@@ -3,7 +3,7 @@
 
 //! Client for the Walrus service.
 
-use std::{collections::HashMap, fmt::Display, sync::Arc, time::Instant};
+use std::{collections::HashMap, fmt::Display, path::PathBuf, sync::Arc, time::Instant};
 
 use anyhow::anyhow;
 use cli::{styled_progress_bar, styled_spinner};
```
giac-mysten, line 9: Remark: spinners and progress bars are currently broken for multiple stores. 
In a followup, maye we could:
- add the path names to the prints, so we know which blob is being stored,
- ensure that the progress bars are not called in parallel (maybe only use a single progress bar for the number of blobs stored).

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -97,7 +114,8 @@ impl CliOutput for BlobStoreResult {
                     Sui object ID: {}\n\
                     Unencoded size: {}\n\
                     Encoded size (including replicated metadata): {}\n\
-                    Cost (excluding gas): {} {}",
+                    Cost (excluding gas): {} {}\n\
+                    Path: {}",
```
giac-mysten: Suggestion: as above, move the path first.

crates/walrus-service/src/client/resource.rs
```diff
@@ -123,59 +123,78 @@ impl<'a> ResourceManager<'a> {
...
-            );
-            tracing::debug!(
-                "there is a deletable certified blob in the wallet, and we are not forcing a store"
-            );
-            StoreOp::NoOp(BlobStoreResult::AlreadyCertified {
-                blob_id: *metadata.blob_id(),
-                event_or_object: EventOrObjectId::Object(blob.id),
-                end_epoch: blob.certified_epoch.unwrap(),
+        // Filter for already certified/invalid blobs and add to result,
+        // otherwise add it to to_be_processed.
```
giac-mysten: Nit: we use line length 100 for code and comments. Is your editor picking up the `.editorconfig` correctly?

crates/walrus-service/src/client/resource.rs
```diff
@@ -123,59 +123,78 @@ impl<'a> ResourceManager<'a> {
...
+        let to_be_processed = metadata_with_status
+            .iter()
+            .filter(|(metadata, blob_status)| {
+                if !store_when.is_store_always() && !persistence.is_deletable() {
+                    if let Some(result) = self.blob_status_to_store_result(
```
giac-mysten, line 146: Suggestion(for follow up): add instrumentation for the blob ids to these sub functions, as it is not clear to which files/blob ids the logs are referring to.

crates/walrus-service/src/client/resource.rs
```diff
@@ -123,59 +123,78 @@ impl<'a> ResourceManager<'a> {
...
+                    operation: op,
+                }
+            };
+            results.push(store_op);
+        }
+        Ok(results)
     }
 
-    /// Returns a [`Blob`] registration object for the specified metadata and number of epochs.
+    /// Returns a list of [`Blob`] registration object for a list of specified metadata and number
```
giac-mysten:
```suggestion
    /// Returns a list of [`Blob`] registration objects for a list of specified metadata and number
```

crates/walrus-service/src/client/resource.rs
```diff
@@ -190,29 +209,40 @@ impl<'a> ResourceManager<'a> {
     #[tracing::instrument(skip_all, err(level = Level::DEBUG))]
     pub async fn get_existing_or_register(
         &self,
-        metadata: &VerifiedBlobMetadataWithId,
+        metadatum: &[&VerifiedBlobMetadataWithId],
```
giac-mysten: Nit:
`metadata` is already plural, `metadatum` is singular.
something like `metadata_list`?

crates/walrus-service/src/client/resource.rs
```diff
@@ -222,72 +252,145 @@ impl<'a> ResourceManager<'a> {
 
     async fn get_existing_or_register_with_resources(
         &self,
-        encoded_length: u64,
+        encoded_lengths: &[u64],
         epochs_ahead: EpochCount,
-        metadata: &VerifiedBlobMetadataWithId,
+        metadatum: &[&VerifiedBlobMetadataWithId],
```
giac-mysten: Nit: as above.

crates/walrus-service/src/client/resource.rs
```diff
@@ -222,72 +252,145 @@ impl<'a> ResourceManager<'a> {
...
+
+        let mut reused_metadata_with_storage = Vec::new();
+        let mut reused_encoded_lengths = Vec::new();
+
+        let mut new_list = Vec::new();
+        let mut new_encoded_lengths = Vec::new();
+
+        // This keeps tracks of selected storage objects and exclude them from selecting again.
+        // todo: might not be efficient if the list is long, consider multiselect in
+        // owned_storage_for_size_and_epoch.
```
giac-mysten: Question: is this todo going to be resolved? if not add an issue and mention the issue number here?
```suggestion
        // TODO(joy): Might not be efficient if the list is long, consider multiselect in
        // owned_storage_for_size_and_epoch.
```

crates/walrus-service/src/client/resource.rs
```diff
@@ -222,72 +252,145 @@ impl<'a> ResourceManager<'a> {
...
+        let mut excluded = Vec::new();
+
+        // For all the metadata, if the blob is registered in wallet, add it directly to results.
+        // Otherwise, check if there is existing storage resource selected for the encoded length,
+        // add it to reused_metadata_with_storage and its length to reused_encoded_lengths.
+        // Otherwise, add it to new_list and its length to new_list and new_encoded_lengths.
```
giac-mysten:
```suggestion
        // Otherwise, add it to new_list and its length to new_encoded_lengths.
```

crates/walrus-service/src/client/resource.rs
```diff
@@ -222,72 +252,145 @@ impl<'a> ResourceManager<'a> {
...
+        debug_assert!(
+            encoded_lengths.len() == metadatum.len(),
+            "inconsistent metadata and encoded lengths"
+        );
+        let mut results = Vec::new();
+
+        let mut reused_metadata_with_storage = Vec::new();
+        let mut reused_encoded_lengths = Vec::new();
+
+        let mut new_list = Vec::new();
```
giac-mysten: Nit: (here and in comment strings)
```suggestion
        let mut new_metadata_list = Vec::new();
```

crates/walrus-service/src/client.rs
```diff
@@ -88,6 +89,18 @@ mod multiplexer;
 
 type ClientResult<T> = Result<T, ClientError>;
 
+const MAX_TOTAL_BLOB_SIZE: usize = 1024 * 1024 * 1024; // 1GB
```
giac-mysten: Nit:
```suggestion
const MAX_TOTAL_BLOB_SIZE: usize = 1024 * 1024 * 1024; // 1GiB
```

crates/walrus-service/src/client.rs
```diff
@@ -376,6 +427,86 @@ impl Client<SuiContractClient> {
...
+        &self,
+        blobs: &[&[u8]],
+    ) -> ClientResult<Vec<(Vec<SliverPair>, VerifiedBlobMetadataWithId)>> {
+        if blobs.is_empty() {
+            return Ok(Vec::new());
+        }
+
+        if blobs.len() > 1 {
+            let total_blob_size = blobs.iter().map(|blob| blob.len()).sum::<usize>();
+            if total_blob_size > MAX_TOTAL_BLOB_SIZE {
```
giac-mysten: Remark: the maximum blob size that can be stored on walrus is ~14 GB. We cannot limit the upload to 1GB in all cases.
Question: Can we make this configurable instead?

crates/walrus-service/src/client.rs
```diff
@@ -406,102 +537,228 @@ impl Client<SuiContractClient> {
...
+    ) -> ClientResult<Vec<BlobStoreResult>> {
+        let metadata_len = pairs_and_metadata.len();
+        tracing::info!("storing {} sliver pairs with metadata", metadata_len);
         let status_start_timer = Instant::now();
-        let blob_status = self
-            .get_blob_status_with_retries(&blob_id, &self.sui_client)
-            .await?;
+
+        // Fetch status for each blob, then build mapping: blob id -> (pair, metadata, status).
+        // Filters out all the failed blobs that cannot fetch status and continue.
```
giac-mysten: Question: is this a silent failure? how is the user informed?

crates/walrus-service/src/client.rs
```diff
@@ -406,102 +537,228 @@ impl Client<SuiContractClient> {
...
+        let failed_blobs = Arc::new(RwLock::new(Vec::with_capacity(metadata_len)));
```
giac-mysten: Question: failed blobs are pushed into this, but they are never read after? or am i missing something?

crates/walrus-e2e-tests/tests/test_client.rs
giac-mysten: Question: can we add a test somewhere that checks that the `(blob_id, path)` mapping is respected? or is this too much work? can be done in a followup

crates/walrus-service/src/client.rs
```diff
@@ -406,102 +537,228 @@ impl Client<SuiContractClient> {
...
-                        tokio::time::sleep(self.config.communication_config.registration_delay)
-                            .await;
+        // Get certificates for all new blobs.
+        let get_cert_timer = Instant::now();
+        let committees = self.committees.read().await;
+        let metadata_ref = &metadata_with_status;
+
+        let mut failed_indices = Vec::with_capacity(new_blobs_and_ops.len());
+        let mut blobs_with_certificates = Vec::with_capacity(new_blobs_and_ops.len());
+        futures::future::join_all(new_blobs_and_ops.iter().map(
```
giac-mysten, line 720: Praise: now the store is also parallel, nice!
I wonder how this affects the performance and the number of open connections.

Question: in `ClientCommunicationConfig` we have configurations such as `max_concurrent_writes` and `max_data_in_flight`. I guess these are not respected anymore, right? and are a function of the number of files?

If this is the case, should we change the documentation? @mlegner thoughts here?

crates/walrus-e2e-tests/tests/test_client.rs
joyqvq: revised to 4, should be sufficient for testing

crates/walrus-service/src/client.rs
joyqvq: this function is also used by the multiplexer where the path is not specified so cant just mark it to test-only. 

alternatively i can add a default path and switch it to reserve_and_store_blobs_retry_epoch_with_path here, then make reserve_and_store_blobs_retry_epoch test only. any preferences?

crates/walrus-service/src/client.rs
joyqvq: the command itself its not broken iiuc since the variable name remains the same. the old command should just work: `walrus store <path1>`. is there any deps (e.g. walrus-site) / downstream services that i should double check?

crates/walrus-service/src/client.rs
joyqvq: i currently start observing timeouts / retries  e.g. `storing metadata and pairs failed node=29 error=the metadata could not be stored` when im uploading 10 100mb files so the time to upload takes a lot longer. it feels like the multistore will have performance degrades when more than 1gb. 

i agree that its a bit more strenuous to enforce here, should this instead live in the top level in this config file with default? https://github.com/MystenLabs/walrus/blob/b6196e7362cdbd4094a509ac770515450909b111/crates/walrus-service/src/client/config.rs#L46

crates/walrus-service/src/client.rs
joyqvq: its only logged with tracing::warning but not in the return type, since the return type vec<BlobStoreResult> cant really handle list of errors. note that i handled this in a similar fashion for 1) get encoding sizes 2) get blob status + check blacklist (here) 3) get blob cert. 

a future improvement is to change the return result type to return a list of success and failures. its a bit too much of a refactoring to ClientResult right now so i think its good to leave it out - wdyt?

crates/walrus-service/src/client.rs
joyqvq: the failed ones are logged in warning and will not continue to be tried with storage ops later. if all of the blobs failed at this step it will be be returned early as an error result. 

note that `get_blob_status_with_retries` itself is unchanged and still will be retried at the per blob level.

crates/walrus-service/src/client/cli/runner.rs
joyqvq: in the happy path should be the same, when the blobs fail at some previous steps with client errors: 1) encoding 2) get blob status / check blob id 3) get blob cert. they all each log the error, but do not return it in the result

crates/walrus-service/src/client/resource.rs
joyqvq: good point, added in this pr as much as i can.

crates/walrus-service/src/client/resource.rs
joyqvq: lol i remembered as the other around, fixed!

crates/walrus-service/src/client/resource.rs
joyqvq: im a bit unsure if this would be an overkill or more error prone to do this. the idea here is this function wants to do, is to select a list of storages based one a list of encoded lengths. it could be further done in one pass, instead of select one, exclude it, select the next one, etc. do you think its a meaningful improvement?

i can see it being useful if there is a long list of storages and a list of blobs to be stored, then we should sort and select them in one pass. 

it feels like it depends on how this is done `// TODO(giac): consider splitting the storage before reusing it (#811).` i added a todo for myself here and in linear for now and lmk what you think!

crates/walrus-sui/src/client.rs
joyqvq: yeah the order is no longer preserved in the walrus code as well!

crates/walrus-e2e-tests/tests/test_client.rs
joyqvq: added!

crates/walrus-service/src/client/resource.rs
joyqvq: i have it as a precommit, fixed with auto format instead

crates/walrus-sui/src/utils.rs
joyqvq: rebased and fixed based on your pr

crates/walrus-service/src/client.rs
joyqvq: iiuc, progress bar happens and still works as expected at the per blob level: 

https://github.com/MystenLabs/walrus/blob/5cc2fe05bff6d08c5d96859bbd81441e2527eb72/crates/walrus-service/src/client.rs#L883

you mean, we should introduce a global progress bar for all blobs to upload?

crates/walrus-sui/src/client.rs
joyqvq: the resource is by value when im creating a vector. its more efficient to just use the owned values since the vector is already owned upon creation, im also using the value once without cloning. 

https://github.com/MystenLabs/walrus/blob/5cc2fe05bff6d08c5d96859bbd81441e2527eb72/crates/walrus-service/src/client/resource.rs#L322

crates/walrus-service/src/client.rs
joyqvq: good call: skimmed through the limit in config, iiuc this is currently enforced on the per blob level. we can either replace it by the total blob encoding length and add the limit here, or alternatively, we add an additional config with Semaphore to impose a max_concurrent_get_cert limit (in addition to per blob limit) at top level here. 

which way do we prefer? (i prefer to set it to another pr for improvement)

crates/walrus-service/src/client.rs
```diff
@@ -406,102 +546,229 @@ impl Client<SuiContractClient> {
         Ok((pairs, metadata))
     }
 
-    async fn reserve_and_store_encoded_blob(
+    async fn reserve_and_store_encoded_blobs(
```
mlegner, line 549: Suggestion: As this function now has almost 200 LOC, it might be good to split it up a bit.

crates/walrus-service/src/client.rs
mlegner: @giac-mysten @joyqvq Sorry, I missed this question directed at me.

I think we do need to make some changes to this, but this can be in a follow-up. It might make sense to look at this a bit more holistically and/or together with other open issues like #685. 

We currently don't have these config parameters documented publicly in any way (except that they exist), so I wouldn't worry about this too much in the short term.

crates/walrus-service/src/client.rs
joyqvq: good point, i extracted `get_all_blob_certificates` and `get_blob_statuses`

crates/walrus-service/src/client.rs
joyqvq: thanks! i left a todo here to address in the future

crates/walrus-service/src/client.rs
joyqvq: moved it to client config file.

crates/walrus-service/src/client.rs
giac-mysten: right, my question was what is the `failed_blobs` variable needed for?

joyqvq: ah i see, yes i can remove it, since we don't want to care about the errors in the return type

crates/walrus-service/src/client.rs
giac-mysten: mmm i'm wondering here if it's better to fail out directly in these scenarios.
Consider the single-blob case, you'd store, the program would end, and nothing would have happened (without user feedback, as users don't have logs on). Can we do anything better?

crates/walrus-service/src/client/cli/runner.rs
giac-mysten: Can we use this to at least inform the user that some blobs were not stored?

crates/walrus-service/src/client.rs
giac-mysten: ah good point. I guess it's fine then!

crates/walrus-service/src/client.rs
giac-mysten: maybe, yes... but followup :)

crates/walrus-service/src/client/resource.rs
giac-mysten: i think it's fine as a followup! thanks for the explanation

crates/walrus-service/src/client.rs
giac-mysten: left a suggestion above :)

crates/walrus-service/src/client/cli/runner.rs
joyqvq: added logs if results len != blobs len, log all the blob ids that are not in the result / not stored properly (then client can look for warning logs for particular error, they are keyed by blob id so should be good enough to track)

PR #1256, opened 2024-12-04
fix: secondary sliver column family should be created last

halfprice:
## Description

The issue was found in simtest that when a node crashes in the middle of shard storage initialization, restarting only checks if secondary sliver column family exists or not. After adding status/shard_sync/recovery column families in the shard, this is no longer correct.

To make the change simple, I just move secondary shard creation to the last.

Also rename `ShardStorage::existing_shards` to `ShardStorage::existing_cf_shard_ids` to differentiate from `Storage::existing_shards`

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -113,7 +113,14 @@ impl Storage {
         db_opts.create_missing_column_families(true);
         db_opts.create_if_missing(true);
 
-        let existing_shards_ids = ShardStorage::existing_shards(path, &db_opts);
+        let existing_shards_ids = ShardStorage::existing_cf_shards_ids(path, &db_opts);
+        tracing::info!(
+            "open storage for existing shards IDs: {}",
+            existing_shards_ids
+                .iter()
+                .map(ToString::to_string)
+                .join(", ")
+        );
         let mut shard_column_families = existing_shards_ids
             .iter()
             .copied()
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/node/storage.rs
```diff
@@ -113,7 +113,11 @@ impl Storage {
         db_opts.create_missing_column_families(true);
         db_opts.create_if_missing(true);
 
-        let existing_shards_ids = ShardStorage::existing_shards(path, &db_opts);
+        let existing_shards_ids = ShardStorage::existing_cf_shards_ids(path, &db_opts);
+        tracing::info!(
+            "open storage existing shards ids: {:?}",
+            existing_shards_ids.iter().copied().collect::<Vec<_>>()
```
mlegner: Nit:
```suggestion
            "open storage for existing shards IDs: {}",
            existing_shards_ids
                .iter()
                .map(ToString::to_string)
                .join(", ")
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -192,6 +183,19 @@ impl ShardStorage {
             rw_options
         );
 
+        // Make sure that sliver column families are created last. They are used to identify
+        // whether the shard storage is initialized.
```
mlegner: Nit: Can we point to the corresponding function?
```suggestion
        // Make sure that sliver column families are created last. They are used to identify
        // whether the shard storage is initialized in `existing_cf_shards_ids`.
```

PR #1257, opened 2024-12-04
perf: optimze key aggregation

jonas-lj:
## Description

* Use `uncompressed_g1_sum` to aggregate public keys.
* Compute the aggregate public key from a total aggregate key and subtract non-signers.
* Pass indices of signers as a bitmap.

Fixes WAL-223.

## Test plan

Unit tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -3821,7 +3821,7 @@ mod tests {
 
             // Waits for the shard sync process to stop.
             wait_until_no_sync_tasks(&cluster.nodes[1].storage_node.shard_sync_handler).await?;
-            check_all_blobs_are_synced(&_blob_details, &_shard_storage_dst)?;
+            check_all_blobs_are_synced(&_blob_details, &_shard_storage_dst, &[])?;
 
             Ok(())
         }
@@ -3849,7 +3849,7 @@ mod tests {
             wait_until_no_sync_tasks(&cluster.nodes[1].storage_node.shard_sync_handler).await?;
 
             // All blobs should be recovered in the new dst node.
-            check_all_blobs_are_synced(&_blob_details, &_shard_storage_dst)?;
+            check_all_blobs_are_synced(&_blob_details, &_shard_storage_dst, &[])?;
 
             Ok(())
         }


contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -18,7 +18,7 @@ const ENotEnoughStake: u64 = 2;
 const EIncorrectCommittee: u64 = 3;
 
 public struct BlsCommitteeMember has copy, drop, store {
-    public_key: Element<G1>,
+    public_key: Element<UncompressedG1>,
```
jonas-lj, line 21: Note that the deserialization on the Rust side can handle this change because the. `from_bytes` function on BLS12381PublicKey also accepts uncompressed format.

crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -425,6 +425,9 @@ pub(crate) struct BlsCommittee {
     n_shards: u16,
     /// The current epoch
     epoch: Epoch,
+    #[cfg(feature = "mainnet-contracts")]
+    /// Aggregated key for all committee members
+    aggregated_keys: PublicKey,
```
karlwuest: I think this should fix the error:
```suggestion
    /// Aggregated key for all committee members
    #[serde(deserialize_with = "deserialize_public_key")]
    aggregated_keys: PublicKey,
```

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -61,7 +70,7 @@ public(package) fun new_bls_committee_member(
     node_id: ID,
 ): BlsCommitteeMember {
     BlsCommitteeMember {
-        public_key,
+        public_key: bls12381::g1_to_uncompressed_g1(&public_key),
```
karlwuest: Suggestion: I think it may be best to store the keys in uncompressed form in the `StorageNodeInfo` already when the node registers (or sets a new key). Otherwise, we will do the decompression for all nodes in the committee in the function that performs the epoch change, which is already one of the most expensive functions (and thus most likely to hit execution limits once we scale up to more nodes).

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -200,6 +219,21 @@ public(package) fun verify_certificate(
...
 }
 
+/// Returns the complement of the given list with respect to the range [0, n)
+/// assuming that the list is in strictly increasing order.
+fun complement(list: &vector<u16>, n: u16): vector<u16> {
+    assert!(list.length() <= n as u64);
+    let mut result: vector<u16> = vector::empty();
+    let mut offset = 0;
+    list.do_ref!(|index| {
+        assert!(*index >= offset);
```
karlwuest: Nit: emit error code in the assert (`ETotalMemberOrder`).

contracts/walrus/tests/system/bls_tests.move
```diff
@@ -58,21 +61,6 @@ public fun test_incorrect_signature_error() {
     );
 }
 
-#[test, expected_failure(abort_code = bls_aggregate::ETotalMemberOrder)]
```
jonas-lj, line 61: This test no longer makes sense because the order of the signers is given from the bitmap.

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -153,35 +162,58 @@ public(package) fun verify_quorum(self: &BlsCommittee, weight: u16): bool {
...
     self: &BlsCommittee,
     signature: &vector<u8>,
-    signers: &vector<u16>,
+    signers_bitmap: &vector<u8>,
     message: &vector<u8>,
 ): u16 {
     // Use the signers flags to construct the key and the weights.
 
     // Lower bound for the next `member_index` to ensure they are monotonically
     // increasing
```
karlwuest: Nit
```suggestion
```

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -153,35 +162,58 @@ public(package) fun verify_quorum(self: &BlsCommittee, weight: u16): bool {
...
+            };
+        });
+        offset = offset + 8;
     });
 
     // The expression below is the solution to the inequality:
     // n_shards = 3 f + 1
     // stake >= 2f + 1
-    assert!(self.verify_quorum(aggregate_weight), ENotEnoughStake);
+    let aggregate_weight = self.n_shards - non_signer_aggregate_weight;
```
karlwuest: I just noticed that we didn't delete the comment after separating this out into `verify_quorum`, I think we can delete the comment here.

contracts/walrus/tests/system/bls_tests.move
```diff
@@ -104,27 +92,31 @@ public fun test_cert_incorrect_epoch() {
...
     let message = messages::certified_message_bytes(epoch, 0xABC);
 
-    // Create the aggregate sig for keys 0, 1, 2, 3, 4, 5, 6
+    // Create the aggregate sig for keys 0, 1, 2, 3, 4, 7, 8
+    let signers = vector[0, 1, 2, 3, 4, 7, 8];
     let mut sigs = vector[];
-    7u64.do!(|i| sigs.push_back(bls_min_pk_sign(&message, &sks[i])));
+    signers.do!(|i| sigs.push_back(bls_min_pk_sign(&message, &sks[i as u64])));
+
+    let signers_bitmap = x"9f01";
```
karlwuest: Nit: Can you add a `signers_to_bitmap` helper function in `test_utils` that we can then use in tests?

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -278,6 +282,17 @@ impl WalrusPtbBuilder {
         Ok(())
     }
 
+    #[cfg(feature = "mainnet-contracts")]
+    fn signers_to_bitmap(signers: &[u16]) -> Vec<u8> {
+        let mut bitmap = vec![0; (signers.len() + 7) / 8];
```
karlwuest: Nit:
```suggestion
        let mut bitmap = vec![0; signers.len().div_ceil(8)];
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -267,6 +267,10 @@ impl WalrusPtbBuilder {
     ) -> SuiClientResult<()> {
         let mut signers = certificate.signers.clone();
         signers.sort_unstable();
```
karlwuest: Nit: since `signers_to_bitmap` doesn't need a sorted input, we can add `#[cfg(not(feature = "mainnet-contracts"))]` to the sorting.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -387,6 +402,10 @@ impl WalrusPtbBuilder {
...
```
karlwuest: Nit: same as above ( since signers_to_bitmap doesn't need a sorted input, we can add `#[cfg(not(feature = "mainnet-contracts"))]` to the sorting.)

contracts/walrus/tests/system/bls_tests.move
jonas-lj: Sure!

Issue #919, opened 2024-10-03
Improve sliver recovery limits.


jpcsmith:
Currently, sliver recovery is both limited in terms of number of simultaneous blobs and number of slivers being synced. 

With 1000 shards, sliver sizes range around 350 B to 45 MB on the size of the blob and the sliver type. Proofs for the sliver are ~320 B per symbol which amounts to around 100 kB to 200 kB per sliver. Recovering a sliver therefore requires around 100 kB to 45 MB of memory.

Ignoring additional overheads, we would expect being able to support between **22 and 10,000 slivers per GB of RAM** being recovered at once. Therefore, a fixed limit of concurrent sliver recoveries can either heavily over-utilise or under-utilise the available memory when recovering slivers.

A better approach may be to have limits in place on the basis of memory, using the blob's size from taken from the available metadata. This could be, for example, "1 semaphore permit = 100 kB of memory". Recovering a large sliver would then require 450 permits and a minimum sized sliver would require only 1 permit.

PR #1261, opened 2024-12-05
fix: shard recovery needs to cancel any blob recovery when the blob is expired/deleted/invalid

halfprice:
## Description

Instead of exposing a hook from shard sync to node for cancelling blob recover, I implemented a simpler approach where blob recover also monitors the blob status, and quit the recover task. This is less efficient than what blob_sync does (tracking active blob sync and cancel them upon receiving blob deleted/invalid/expired events), but the code is a lot simpler than if we want to use similar approach as blob_sync.

Opinions are welcome!

Close WAL-283
Contribute WAL-343

## Test plan

---

## Release notes

- [x] Storage node: Storage node fixes
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -19,9 +19,9 @@ use futures::{stream::FuturesUnordered, StreamExt};
 use regex::Regex;
 use rocksdb::{Options, DB};
 use serde::{Deserialize, Serialize};
-#[cfg(msim)]
-use sui_macros::fail_point_arg;
 use sui_macros::fail_point_if;
+#[cfg(msim)]
+use sui_macros::{fail_point, fail_point_arg};
 use tokio::sync::Semaphore;
 use typed_store::{
     rocks::{


crates/walrus-service/src/node.rs
```diff
@@ -2471,19 +2476,33 @@ mod tests {
...
+    /// The function also takes custom function to determine the end epoch of a blob, and whether
+    /// the blob should be deletable.
+    async fn cluster_with_partially_stored_blobs_in_shard_0<'a, F, G>(
         assignment: &[&[u16]],
         blobs: &[&'a [u8]],
         initial_epoch: Epoch,
         mut blob_index_store_at_shard_0: F,
-    ) -> TestResult<(TestCluster, Vec<EncodedBlob>)>
+        mut blob_index_to_end_epoch: G,
+        deletable_blob_indices: &[usize],
```
mlegner, line 2501: Nit: The way `deletable_blob_indices` is passed is inconsistent with the two functions that are used for `blob_index_store_at_shard_0` and `blob_index_to_end_epoch`. Could this be a function too?

crates/walrus-service/src/node.rs
```diff
@@ -2510,13 +2529,35 @@ mod tests {
...
-            all_other_node_events
-                .send(BlobRegistered::for_testing(*blob_details.blob_id()).into())?;
+            let blob_end_epoch = blob_index_to_end_epoch(i);
+            let deletable = deletable_blob_indices.contains(&i);
+            node_0_events.send(
+                BlobRegistered {
+                    deletable,
+                    end_epoch: blob_end_epoch,
+                    ..BlobRegistered::for_testing(*blob_details.blob_id())
+                }
```
mlegner, line 2539: Nit: This could be assigned to a variable to avoid defining it twice. Same for the certified message.

crates/walrus-service/src/node.rs
```diff
@@ -3307,41 +3361,61 @@ mod tests {
     }
 
     // Checks that all primary and secondary slivers match the original encoding of the blobs.
+    // Don't check if the blob is in the skip list.
```
mlegner, line 3364: Nit: AFAIU, this function explicitly checks that blobs in the skip list *are not synced*.

crates/walrus-service/src/node.rs
```diff
@@ -3501,19 +3585,115 @@ mod tests {
...
+        register_fail_point_if(
+            "shard_recovery_skip_initial_blob_certification_check",
+            move || skip_blob_certification_at_recovery_beginning,
+        );
+
+        let skip_stored_blob_index: [usize; 12] = [3, 4, 5, 9, 10, 11, 15, 18, 19, 20, 21, 22];
+
+        // Blob 3 expires at epoch 2, which is the current epoch when
+        // `setup_shard_recovery_test_cluster` returns.
+        let blob_end_epoch: HashMap<usize, Epoch> = HashMap::from([(3, 2)]);
```
mlegner, line 3622: Nit: I feel it would be more straight-forward to directly define the closure here:
```suggestion
        let blob_end_epoch = |blob_index| if blob_index == 3 { 2 } else { 42 };
```

crates/walrus-service/src/node/metrics.rs
```diff
@@ -61,6 +61,12 @@ telemetry::define_metric_set! {
         #[help = "Total number of slivers failed to recover during shard sync"]
         sync_shard_recover_sliver_error_total: IntCounterVec["shard"],
 
+        #[help = "Total number of slivers skipped during shard sync"]
+        sync_shard_recover_sliver_skip_total: IntCounterVec["shard"],
+
+        #[help = "Total number of slivers successfully recovered during shard sync"]
```
mlegner, line 67:
```suggestion
        #[help = "Total number of cancelled slivers recoveries during shard sync"]
```

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -273,8 +273,6 @@ pub(super) trait Mergeable: Sized {
 pub(crate) trait BlobInfoApi {
     /// Returns a boolean indicating whether the metadata of the blob is stored.
     fn is_metadata_stored(&self) -> bool;
-    /// Returns true iff the blob is invalid.
-    fn is_invalid(&self) -> bool;
```
mlegner, line 277: Question: Is this removed because it's currently not used?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -835,12 +894,47 @@ impl ShardStorage {
...
+                tokio::time::interval(Duration::from_secs(1))
+            };
+
+            loop {
+                check_interval.tick().await;
+                if !node_clone.is_blob_certified(&blob_id)? {
+                    return Ok(());
+                }
+            }
+        };
```
mlegner, line 913: Hmm... Originally I thought that this is a lot of work (including DB accesses), which is probably wasted for the vast majority of blobs. However, most recoveries should be done within a minute, in which case the check never actually happens, is that correct?

Irrespective of that: Can we make the time interval configurable?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -783,23 +784,77 @@ impl ShardStorage {
...
+                );
+                metrics::with_label!(
+                    node.metrics.sync_shard_recover_sliver_skip_total,
+                    &self.id.to_string()
+                )
+                .inc();
+                self.pending_recover_slivers
+                    .remove(&(sliver_type, blob_id))?;
+            }
         }
```
mlegner, line 821: Important (probably for a separate PR): IIUC, we create an unbounded number of futures here, which may exceed the available memory, see also #919. We may need to limit the number of futures that exist simultaneously here.

crates/walrus-service/src/node/storage/shard.rs
halfprice: > However, most recoveries should be done within a minute, in which case the check never actually happens, is that correct?

Exactly. Even under a low bandwidth setting (say 20 Mbps), transferring 1 sliver of maximum file size (35MB for 14G file) takes less than 15s. Added some additional time for computing the sliver on the source side. Yes, I can make this configurable since I don't have any idea about how long does it take to compute the sliver.

sadhansood: Suggestion: Would it then make sense to have the initial check delay be at 60s and then follow up at slightly lower delay i.e 10s? The existing suggestion of making this configurable is probably enough but the hybrid setting of different init+periodic delay might be an improvement.

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: Yes.

PR #1264, opened 2024-12-06
feat: when shard sync fails, we need to retry shard recovery directly

halfprice:
## Description

We've seen this before that if the source node is down, currently the returned error will terminate the shard sync task, but the shard is not synced. Instead, we want to retry shard sync and directly use the recovery mechanism.

Contributes to WAL-195

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -142,6 +142,7 @@ impl ShardSyncHandler {
                 shard_index.0 as u64, // Seed the backoff with the shard index.
             );
 
+            // TODO(WAL-444): see if we can remove the mutex.
             let directly_recover_shard = Arc::new(Mutex::new(false));
 
             backoff::retry(backoff, || async {


crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -142,50 +142,76 @@ impl ShardSyncHandler {
                 shard_index.0 as u64, // Seed the backoff with the shard index.
             );
 
+            let directly_recover_shard = Arc::new(Mutex::new(false));
```
mlegner, line 146: Question: Why do we need a mutex here?

crates/walrus-service/src/node/shard_sync.rs
mlegner, line 1: Remark: The `start_shard_sync_impl` function is getting really long and complex with now more than 200 LOC. Can we refactor this a bit sometime to make it more readable?

crates/walrus-service/src/node/shard_sync.rs
sadhansood: +1, This seems like it could've been a simple boolean flag.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -569,62 +574,64 @@ impl ShardStorage {
...
-            let fetched_slivers = node
-                .committee_service
-                .sync_shard_before_epoch(
-                    self.id(),
-                    next_starting_blob_id,
-                    sliver_type,
-                    config.sliver_count_per_sync_request,
+        if !directly_recover_shard {
+            while let Some((next_starting_blob_id, _)) = next_blob_info {
+                tracing::debug!(
```
sadhansood, line 579: Suggestion/Nit: I think this function could be slightly refactored, and use early return:
```
async fn sync_shard_before_epoch_internal(
    &self,
    epoch: Epoch,
    node: Arc<StorageNodeInner>,
    sliver_type: SliverType,
    last_synced_blob_id: Option<BlobId>,
    config: &ShardSyncConfig,
    directly_recover_shard: bool,
) -> Result<(), SyncShardClientError> {
    #[cfg(msim)]
    let mut scan_count: u64 = 0;

    let mut blob_info_iter = node
        .storage
        .blob_info
        .certified_blob_info_iter_before_epoch(
            epoch,
            last_synced_blob_id.map_or(Unbounded, Excluded),
        );

    let mut next_blob_info = blob_info_iter.next().transpose()?;
    
    // Validate epoch transition
    assert!(epoch != 0 && (epoch > 1 || next_blob_info.is_none()));
    
    if next_blob_info.is_none() {
        return Ok(());
    }

    if directly_recover_shard {
        return self.recover_remaining_blobs(sliver_type, &mut blob_info_iter, next_blob_info).await;
    }

    return self.sync_blobs_sequentially(
        epoch,
        node,
        sliver_type,
        config,
        &mut blob_info_iter,
        next_blob_info,
    ).await;
}
```

and

```
async fn recover_remaining_blobs(
    &self,
    sliver_type: SliverType,
    blob_info_iter: &mut BlobInfoIterator,
    mut next_blob_info: Option<(BlobId, BlobInfo)>,
) -> Result<(), SyncShardClientError> {
    let mut batch = self.pending_recover_slivers.batch();
    
    while let Some((blob_id, _)) = next_blob_info {
        batch.insert_batch(
            &self.pending_recover_slivers,
            [((sliver_type, blob_id), ())],
        )?;
        next_blob_info = blob_info_iter.next().transpose()?;
    }
    
    batch.write()?;
    Ok(())
}
```

```
async fn sync_blobs_sequentially(
    &self,
    epoch: Epoch,
    node: Arc<StorageNodeInner>,
    sliver_type: SliverType,
    config: &ShardSyncConfig,
    blob_info_iter: &mut BlobInfoIterator,
    mut next_blob_info: Option<(BlobId, BlobInfo)>,
) -> Result<(), SyncShardClientError> {
    while let Some((next_starting_blob_id, _)) = next_blob_info {

     }
     if next_blob_info.is_some() {
        self.recover_remaining_blobs(sliver_type, blob_info_iter, next_blob_info).await?;
    }

    Ok(())
}
```

crates/walrus-service/src/node/shard_sync.rs
halfprice: We want to update `directly_recover_shard` in backoff::retry(|| async {}), and it doesn't allow reference inside the async block to escape. I haven't found a nice work around, but I agree that it's weird to have a mutex here.

Created WAL-444 to see if we can remove it later.

crates/walrus-service/src/node/shard_sync.rs
halfprice: I will extract the retry function out. Will do it in a follow up PR.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Will also create a followup PR for this. Want to catch the 1.6.0 release, and don't have much time over the weekend to work on this.

PR #1266, opened 2024-12-06
chore: bump Rust version to 1.83 and use Debian Bookworm in Dockerfiles

mlegner:
## Description

Bumps the rust version in the toolchain and in all Dockerfiles. Also uses newer Debian Bookworm instead of Bullseye for Docker containers.

## Test plan

Existing CI workflows.

### DIFF
--- a/crates/walrus-sui/src/utils.rs
+++ b/crates/walrus-sui/src/utils.rs
@@ -49,13 +49,14 @@ use crate::{
     types::NodeRegistrationParams,
 };
 
-// Keep in sync with the same constant in `contracts/walrus/sources/system.move`.
+// Keep in sync with the same constant in `contracts/walrus/sources/system/system_state_inner.move`.
 // The storage unit is used in doc comments for CLI arguments in the files
 // `crates/walrus-service/bin/deploy.rs` and `crates/walrus-service/bin/node.rs`.
 // Change the unit there if it changes.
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/src/utils.rs
```diff
@@ -59,7 +59,7 @@ pub const BYTES_PER_UNIT_SIZE: u64 = 1_024 * 1_024; // 1 MiB
 /// Calculates the number of storage units required to store a blob with the
 /// given encoded size.
 pub fn storage_units_from_size(encoded_size: u64) -> u64 {
-    (encoded_size + BYTES_PER_UNIT_SIZE - 1) / BYTES_PER_UNIT_SIZE
+    encoded_size.div_ceil(BYTES_PER_UNIT_SIZE)
```
karlwuest, line 63: Nit: I know it does the same thing, but do we maybe want to keep it explicit here to have the same expression as in the contracts? Perhaps add a comment to that effect.

mlegner: How about I replace the code in Move by `size.divide_and_round_up(BYTES_PER_UNIT_SIZE)`?

mlegner: Done in bfdd17c65b4994d5a42b0f3c2602930e93c89372.

PR #1278, opened 2024-12-10
fix: Only check for at least one correct node on invalidate blob

jonas-lj:
## Description

Currently, to mark a blob id as invalid, we need a CertifiedMessage, which requires a quorum. However, for marking a blob as invalid, only one correct node needs to have verified the proof, which means that we only need f + 1 signatures.

## Test plan

Unit tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/system/bls_aggregate.move
+++ b/contracts/walrus/sources/system/bls_aggregate.move
@@ -134,8 +134,8 @@ public(package) fun to_vec_map(self: &BlsCommittee): VecMap<ID, u16> {
 
 /// Verifies that a message is signed by a quorum of the members of a committee.
 ///
-/// The signers are listed as indices into the `members` vector of the committee
-/// in increasing order and with no repetitions.
+/// The signers are given as a bitmap for the indices into the `members` vector of
+/// the committee.
 ///
 /// If the signers form a quorum and the signature is valid, the function returns
 /// a new `CertifiedMessage` with the message, the epoch, and the total stake of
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -127,43 +135,75 @@ public(package) fun to_vec_map(self: &BlsCommittee): VecMap<ID, u16> {
 /// Verifies that a message is signed by a quorum of the members of a committee.
 ///
 /// The signers are listed as indices into the `members` vector of the committee
-/// in increasing
-/// order and with no repetitions. The total weight of the signers (i.e. total
-/// number of shards)
-/// is returned, but if a quorum is not reached the function aborts with an
-/// error.
+/// in increasing order and with no repetitions.
```
karlwuest: Nit: change comment to reflect the change to the bitmap, e.g.:
```
    /// The signers are given as a bitmap for the indices into the `members` vector of the committee.
```

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -127,43 +135,75 @@ public(package) fun to_vec_map(self: &BlsCommittee): VecMap<ID, u16> {
...
 /// Returns true if the weight is more than the aggregate weight of quorum members of a committee.
-public(package) fun verify_quorum(self: &BlsCommittee, weight: u16): bool {
+public(package) fun is_quorum(self: &BlsCommittee, weight: u16): bool {
     3 * (weight as u64) >= 2 * (self.n_shards as u64) + 1
 }
 
+/// Verifies that a message is signed by at least one correct node of a committee.
+///
+/// The signers are listed as indices into the `members` vector of the committee
+/// in increasing order and with no repetitions.
```
karlwuest: Nit: see above
```suggestion
/// The signers are given as a bitmap for the indices into the `members` vector of the committee.
```

PR #1281, opened 2024-12-10
feat: wrap SuiClient to retry RPC errors

giac-mysten:
## Description

- Adds retries to calls to the `SuiClient`, in case there is an RPC network error (statuses 429, 500, 502).
- Adds a more informative error message in case the client returns an error that is due to RPC problems. 

Closes: WAL-416
Closes: WAL-313
Closes: WAL-417
Closes: WAL-339
Closes: WAL-290

## Test plan

Manual tests for cli/publisher/aggregator.

---

## Release notes

- [x] Storage node: Retries when the RPC fails with a network error.
- [x] Aggregator: Retries when the RPC fails with a network error.
- [x] Publisher: Retries when the RPC fails with a network error.
- [x] CLI: Improved error messages when issues are caused by the RPC.

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -59,6 +59,7 @@ use walrus_sui::{
     },
 };
 use walrus_test_utils::WithTempDir;
+use walrus_utils::backoff::ExponentialBackoffConfig;
 
 #[cfg(msim)]
 use crate::node::config::{self, SuiConfig};
@@ -1746,7 +1747,6 @@ pub mod test_cluster {
             DEFAULT_GAS_BUDGET,
         },
     };
-    use walrus_utils::backoff::ExponentialBackoffConfig;
 
     use super::*;
     use crate::{


crates/walrus-sui/src/client/retry_rpc.rs
```diff
@@ -0,0 +1,146 @@
...
+pub trait RpcRetriableClient {
+    /// Sets a backoff configuration for RPC retries.
+    ///
+    /// If the [`super::SuiReadClient`] encounters a [`RetriableRpcError`], it will retry the RPC
+    /// call using after an exponentially-increasing backoff time.
+    fn with_rpc_retries(&mut self, config: ExponentialBackoffConfig) -> &mut Self;
+
+    /// Gets the backoff strategy used for RPC retries.
+    fn get_rpc_backoff_strategy(&self) -> Option<ExponentialBackoff<StdRng>>;
+}
```
karlwuest: Suggestion: I don't think this needs to be a trait, these can just be methods on the SuiReadClient, and we could even pass the backoff strategy in the constructor.

crates/walrus-sui/src/client/retry_rpc.rs
```diff
@@ -0,0 +1,146 @@
...
+use walrus_utils::backoff::{BackoffStrategy, ExponentialBackoff};
+
+use super::SuiClientError;
+
+/// The list of HTTP status codes that are retriable.
+const RETRIABLE_RPC_ERRORS: &[&str] = &["429", "500", "502"];
+
+/// Wrapper for the configuration for the exponential backoff strategy.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct ExponentialBackoffConfig {
```
karlwuest: Suggestion: I think it would make sense to define this in `walrus_utils` where the strategy is also defined. This maybe useful for other walrus crates as well.

Question: How do we configure the strategy currently for other parts that use an exponential backoff?

crates/walrus-sui/src/client/retry_rpc.rs
```diff
@@ -0,0 +1,146 @@
...
+    pub fn new(min_backoff: Duration, max_backoff: Duration, max_retries: Option<u32>) -> Self {
+        ExponentialBackoffConfig {
+            min_backoff,
+            max_backoff,
+            max_retries,
+        }
+    }
+
+    /// Gets a new [`ExponentialBackoff`] strategy with the given seed from the configuration.
+    pub fn get_strategy(&self, seed: u64) -> ExponentialBackoff<StdRng> {
```
karlwuest: Remark: In most cases, we probably want to seed this with a random number from the ThreadRng. We only really need the ability to specify a seed for (deterministic) tests. I just noticed that we often actually seed the rng with some "fixed" numbers (node index, shard index etc.) in many places, which is probably not ideal.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -600,16 +619,25 @@ impl SuiReadClient {
...
     }
-}
 
-enum WhichCommittee {
-    Current,
-    Previous,
-    Next,
-}
+    /// Retries the given function while it returns retriable errors.
+    pub async fn retry_rpc_errors<F, T, E, Fut>(&self, mut func: F) -> Result<T, E>
```
karlwuest: Question: Does this need to be public?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -760,6 +788,44 @@ impl ReadClient for SuiReadClient {
...
+
+    // Version with args.
+    ($function:ident($($arg_name:ident: $arg_type:ty),*) -> $return:ty) => {
+        async fn $function(&self, $($arg_name: $arg_type),*) -> $return {
+            self.retry_rpc_errors(|| async { self.$function($($arg_name),*).await }).await
+        }
+    }
+}
+
+impl ReadClient for SuiReadClient {
```
karlwuest: Suggestion: I already mentioned this, but I think it would be preferable to wrap the `SuiClient` instead of the functions in the `SuiReadClient` and perform the retries on the functions there. When I mentioned this, this was more of an intuition, but I think there are good reasons for it.

One, I already mentioned, namely with the current approach we retry the whole function instead of just the failing rpc call. More importantly: We have several functions in `SuiReadClient` that we do not expose in the trait, but we use in the `SuiContractClient`. With the current approach, we do not get retries for those.

I quickly checked and there are not that many different functions that we call on the `SuiClient`, so exposing the ones that we need on a wrapped client should not take a lot of effort, and we can expose functions like `get_sui_object` directly on the wrapped type.

crates/walrus-sui/src/client/read_client.rs
giac-mysten: nope, good call.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -0,0 +1,372 @@
...
+
+use super::{SuiClientError, SuiClientResult};
+use crate::{
+    contracts::{AssociatedContractStruct, TypeOriginMap},
+    types::move_structs::{SuiDynamicField, SystemObjectForDeserialization},
+    utils::{get_package_id_from_object_response, get_sui_object_from_object_response},
+};
+
+/// The list of HTTP status codes that are retriable.
+const RETRIABLE_RPC_ERRORS: &[&str] = &["429", "500", "502"];
```
karlwuest, line 47: Question: are there other errors that we may want to retry?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -0,0 +1,372 @@
...
+
+/// Trait to test if an error is produced by a temporary RPC failure and can be retried.
+pub trait RetriableRpcError: Debug {
```
karlwuest, line 50: Nit: Here you use "retriable", but in other places (e.g. the client), you use "retryable". I don't think there is actually a "correct" spelling, since both seem to be used, but we should probably choose one. 
"Retriable" seems a bit more natural to me, because usually `y` endings (following a consonant) are changed to an `i` if a suffix is added, but I think either of the two is fine.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -0,0 +1,372 @@
...
+
+    /// Returns a [`SuiObjectResponse`] based on the provided [`ObjectID`].
+    ///
+    /// Calls [`sui_sdk::apis::ReadApi::get_object_with_options`] internally.
+    pub async fn get_object_with_options(
+        &self,
+        object_id: ObjectID,
+        options: SuiObjectDataOptions,
+    ) -> SuiRpcResult<SuiObjectResponse> {
+        retry_rpc_errors(self.get_strategy(), || async {
```
karlwuest, line 244: Suggestion (for all calls of `retry_rpc_errors`): It could be useful to create a new span with the function name and then instrument the future with that span (`retry_rpc_errors(...).instrument(span).await`) to provide better context for the log messages that are created in `retry_rpc_errors`.

crates/walrus-sui/src/client/retry_client.rs
giac-mysten: these are the only ones I have seen so far. If more come up we can add them

crates/walrus-sui/src/client/retry_client.rs
giac-mysten: good catch, i was also on the line on this an forgot to check consistency in the end 😅

crates/walrus-sui/src/client/retry_client.rs
giac-mysten: used instrumentation on the functions directly

PR #1284, opened 2024-12-10
refactor: small changes across the the contracts

benr-ml:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/staked_wal.move
+++ b/contracts/walrus/sources/staking/staked_wal.move
@@ -211,7 +211,6 @@ public fun split(sw: &mut StakedWal, amount: u64, ctx: &mut TxContext): StakedWa
         // If the staked WAL is withdrawing, we need to perform pool token amount
         // calculation based on the amount being split. Needn't worry about the
         // rounding errors as the value is always subtracted from the principal.
-        // TODO: disable split during withdraw.
         StakedWalState::Withdrawing { withdraw_epoch, pool_token_amount } => {
             // reclaculate the pool token amount if it is set, otherwise ignore
             let new_pool_token_amount = if (pool_token_amount.is_some()) {


contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -389,6 +389,9 @@ public(package) fun certify_event_blob(
         return
     };
 
+    // [ben] event_blob_certification_state is internally a vector, and a node
```
benr-ml: related to https://github.com/MystenLabs/walrus/pull/1282/files

contracts/walrus/sources/staking.move
```diff
@@ -52,6 +52,8 @@ public(package) fun create(
 // === Public API: Storage Node ===
 
 /// Creates a staking pool for the candidate, registers the candidate as a storage node.
+// [ben] anyone can register a candidate with any name and network address - are we worried 
+// about impersonation (i.e., use mysten-labs as a name even though there is another node with the same name)?
```
karlwuest: That's a good question. I'm actually not sure if there is a good way to prevent this. After all, the system is permissionless. I don't know if this is handled some way in sui, but it might be worth to check.

contracts/walrus/sources/staking/committee.move
```diff
@@ -39,6 +42,7 @@ public(package) fun initialize(assigned_number: VecMap<ID, u16>): Committee {
 ///
 /// This assumes that the number of shards in the new committee is equal to the
 /// number of shards in the current committee. Check for this is not performed.
+// [ben] why it's not checked?
```
karlwuest: This should be guaranteed by the caller. 
But I think adding a check should be doable without a big overhead, since we can just sum up the number of shards in the previous committee, and the number of shards in the new committee while iterating through them.

contracts/walrus/sources/staking/exchange_rate.move
```diff
@@ -29,22 +29,33 @@ public(package) fun new(wal_amount: u64, pool_token_amount: u64): PoolExchangeRa
...
 
+/// Assumptions:
+/// - token_amount is at most the amount of pool tokens in the pool
 public(package) fun get_wal_amount(exchange_rate: &PoolExchangeRate, token_amount: u64): u64 {
     // When either amount is 0, that means we have no stakes with this pool.
     // The other amount might be non-zero when there's dust left in the pool.
+
+    // [ben] when can pool_token_amount = 0 but not wal_amount?
+    // [ben] assuming we only care about wal_amount = 0, should we use an enum to be more explicit?
+    // [ben] if wal_amount=0, why should we return token_amount here?
```
karlwuest: I'm not sure if this can still happen in the exchange rates, because now we set the number of pool tokens based on the WAL amount (instead of subtracting pool tokens from withdrawals). But this used to be possible due to rounding (which is unavoidable).

More generally, the exchange rate should be 1:1 if either of the two values is 0, because that means that there is no active stake, so the exchange rate is essentially reset to the default (1:1). Afaik, this is handled the same way in Sui staking. cc @damirka

contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -211,6 +211,10 @@ public fun split(sw: &mut StakedWal, amount: u64, ctx: &mut TxContext): StakedWa
         // If the staked WAL is withdrawing, we need to perform pool token amount
         // calculation based on the amount being split. Needn't worry about the
         // rounding errors as the value is always subtracted from the principal.
+
+        // [ben] why is this permitted? as opposed to only support split for Staked?
+        // This rounding below does not guarantee "fair" distribution of pool tokens 
+        // (can be repeated multiple times to make it more significant than off by 1)
```
karlwuest: The goal for this was to enable more functionality for owners. In particular, once a `StakedWal` is in withdrawing state, it will (in mainnet) take a significant amount of time until it can be withdrawn. Because of this, it's desirable from an economics PoV to still allow splitting it to make it tradeable.

The logic behind the handling of the rounding error here is that for the total amount of money that the owner of a `StakedWal` in this state receives, only the pool token amount should matter, because that is what's used to calculate the withdrawal amount, so it doesn't matter whether the distribution is "fair". 

However, I am happy that you flagged this, because I think there are actually some edge cases where this creates an issue. In particular, if `pool_token_amount * amount < sw.principal.value()`, we will end up with a `new_pool_token_amount` of 0, but a non-zero `amount`. This will then allow to withdraw more than they should have been allowed to.

@damirka let's disable this again to prevent this from happening and create a (post-mainnet) issue to revisit this to see if we can reintroduce it.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -234,6 +233,7 @@ public(package) fun calculate_votes(self: &StakingInnerV1): EpochParams {
         write_prices.insert(pool.write_price(), weight);
         storage_prices.insert(pool.storage_price(), weight);
         // The vote for capacity is determined by the node capacity and number of assigned shards.
+        // [ben] i don't follow this comment. also, why divide here by the "fixed" n_shards?
```
karlwuest: We don't divide by n_shards here, we multiply by it, but it may be more readable (and reduce the rounding error) to rearrange the assignment to
```
let capacity_vote = pool.node_capacity() * (self.n_shards as u64) / weight;
```

One issue with voting for the capacity is that nodes don't know how many shards they will get assigned before voting, so instead we have them vote for how much capacity they can support, which then gets scaled by the number of shards that they hold. 

E.g. they vote for a capacity of 10TB and get 10 shards assigned, that then translates into a vote of 1TB per shard, or 1000 TB in total (which is what we calculate here).

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -359,8 +359,10 @@ public(package) fun advance_epoch(
 
     // split the commission from the rewards
     let total_rewards = rewards.value();
+    // [ben] why do we update commission_rate before this line? shouldn't it be for the next epoch?
```
karlwuest: We update it here to the commission rate for the current epoch. The epoch in which the commission rate is applied is set in the function that sets the new commission rate. This is done because there may be two pending commission rate updates (one for the next epoch and one for the epoch after that).

CC @damirka correct me if I'm wrong

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -510,6 +511,7 @@ public(package) fun destroy_empty(pool: StakingPool) {
 /// Returns the exchange rate for the given current or future epoch. If there
 /// isn't a value for the specified epoch, it will look for the most recent
 /// value down to the pool activation epoch.
+// [ben] i don't see why we fallback to the most recent rate, and it's only confusing this way
```
karlwuest: We fall back to the latest exchange rate before the activation epoch. This is needed because for nodes that are not part of the committee, the exchange rate may not be set for some epochs.

E.g. a user stakes with a node in epoch 1 (before committee selection), which means that the activation epoch on the StakedWal will be 2. Then the node is not selected for the committee and therefore they will not have an exchange rate set during the epoch change (because we cannot do this for all registered nodes). But then, the node is selected for the committee starting in epoch 3. Now, if the user wants to withdraw the stake, we are looking for the exchange rate of epoch 2, which doesn't exist. However, since the node didn't receive a reward for epoch 1, the exchange rate also didn't change, so we use the previous one.

CC @damirka again, correct me if I'm wrong

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -519,7 +521,7 @@ public(package) fun exchange_rate_at_epoch(pool: &StakingPool, mut epoch: u32):
         epoch = epoch - 1;
     };
 
-    pool_exchange_rate::empty()
+    pool_exchange_rate::empty() // [ben] why not assert? this should never happen, no?
```
karlwuest: This is used to get the initial exchange rate for a node that was not active before.

cc @damirka

contracts/walrus/sources/staking/storage_node.move
```diff
@@ -40,6 +40,7 @@ public(package) fun new(
     network_public_key: vector<u8>,
 ): StorageNodeInfo {
     assert!(network_public_key.length() == 33, EInvalidNetworkPublicKey);
+    // [ben] maybe let's check that network_address starts with https://?
```
karlwuest: We discussed doing some validation on the network address in the contracts but decided against it. 

We don't really gain anything from doing it, because no matter what we check, the address can still be invalid (e.g. no DNS, not controlled by the node), and we support both DNS names as well as a combination of IP & port.

contracts/walrus/sources/system/blob.move
```diff
@@ -148,6 +148,7 @@ public(package) fun new(
 
     // Cryptographically verify that the Blob ID authenticates
     // both the size and fe_type.
+    // [ben] why does the user need to provide the blobid? just a sanity check?
```
karlwuest: Yes, this is just a sanity check and, maybe more important, to make it obvious to the caller which blob they are registering.

contracts/walrus/sources/system/blob.move
```diff
@@ -252,6 +253,7 @@ public(package) fun extend_with_resource(blob: &mut Blob, extension: Storage, cu
 }
 
 /// Emits a `BlobCertified` event for the given blob.
+// [ben] what is an extension?
```
karlwuest: This is to extend the expiration time of a blob.

contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -17,6 +17,8 @@ const ESigVerification: u64 = 1;
 const ENotEnoughStake: u64 = 2;
 const EIncorrectCommittee: u64 = 3;
 
+// [ben] why copy is needed in the next 2 structs?
```
karlwuest: That's a good question. I think we might be able to remove it, afaict, we only use it in a few places where we can replace it with a reference, and in one test.

contracts/walrus/sources/system/events.move
```diff
@@ -18,6 +18,7 @@ public struct BlobRegistered has copy, drop {
     end_epoch: u32,
     deletable: bool,
     // The object id of the related `Blob` object
+    // [ben] why is this needed? if only for ref, document the reason
```
karlwuest: This is needed to correctly keep track of deletable blobs on the storage nodes. I will add some documentation.

contracts/walrus/sources/system/messages.move
```diff
@@ -83,8 +83,8 @@ public(package) fun verify_proof_of_possession(
 public struct CertifiedMessage has drop {
     intent_type: u8,
     intent_version: u8,
-    cert_epoch: u32,
-    stake_support: u16,
+    cert_epoch: u32, 
+    stake_support: u16, // [ben] it's confusing that it's not part of the cert, maybe metadata_stake_support and put this field last?
```
karlwuest: I'm not sure if I understand the comment correctly. 
The value states by how much of the stake the message was certified. I don't mind reordering the fields.

contracts/walrus/sources/staking.move
benr-ml: Sui checks for duplications - https://github.com/MystenLabs/sui/blob/a6a80846d7a48ce6c602991221bcc03581d8da33/crates/sui-framework/packages/sui-system/sources/validator.move#L624

karlwuest: Hmmm, I'm not sure how I feel about these duplicate checks. 

They only mitigate impersonating a node that has already registered, but at the same time they would allow name or address squatting and prevent a legitimate node from registering with their correct data by just registering first.

We can create an issue on the backlog for this, but I don't think it is a priority or blocking for mainnet.

contracts/walrus/sources/staking/staked_wal.move
benr-ml: While it's secure/fair with regards to the pool, what I meant by "fairness" here was referring to the actual object, that may be restaked/shared by others, and then might represent values that are not "fair" compared to what the user expects.

Will replace my comment with a TODO.

karlwuest: I think the core point regarding fairness here is that the value for a `StakedWal` in the withdrawing state is solely determined by the `pool_token_amount` and should be treated as such.

But, as I mentioned, I do think that this can cause issues, so we should disable it for now.

contracts/walrus/sources/system/events.move
benr-ml: added

karlwuest: thanks!

contracts/walrus/sources/staking/staking_pool.move
damirka: Exchange rates only exist for epochs where a node is in the committee, but the function can be called for any epoch. So we search the last exchange rate when a node was active.

contracts/walrus/sources/staking/staking_pool.move
damirka: That is correct. If a node is in the committee (hence, advance_epoch is called), the exchange rate will appear and won't be empty. If not, all pool tokens are 1:1 to wal.

For clarity we could return an Option here, but I'm not sure if it's worth it.

contracts/walrus/sources/staking/exchange_rate.move
damirka: Indeed, it is handled the same way in Sui System, if any of the values is `0`, the rate is `1:1`.

contracts/walrus/sources/staking.move
benr-ml: https://linear.app/mysten-labs/issue/WAL-464/prevent-impersonation-of-existing-nodes

contracts/walrus/sources/staking/exchange_rate.move
benr-ml: Ok, do you see a case where the current code would have pool_token_amount = 0 and wal_amount != 0?

benr-ml: If only wal_amount=0 is an option, then maybe we should use an explicit enum `{  Flat, Rate{wal_amount, pool_token_amount} }`

contracts/walrus/sources/staking/staking_pool.move
benr-ml: perhaps we should rename empty -> flat to be clearer here

contracts/walrus/sources/system/blob.move
```diff
@@ -147,7 +147,7 @@ public(package) fun new(
     assert!(encoded_size <= storage.storage_size(), EResourceSize);
 
     // Cryptographically verify that the Blob ID authenticates
-    // both the size and fe_type.
+    // both the size and fe_type (sanity check).
```
karlwuest: Not sure why we called it `fe_type` here, but we can change this as well for readability.

```suggestion
    // both the size and encoding_type (sanity check).
```

contracts/walrus/sources/system/messages.move
```diff
@@ -83,9 +83,9 @@ public(package) fun verify_proof_of_possession(
 public struct CertifiedMessage has drop {
     intent_type: u8,
     intent_version: u8,
-    cert_epoch: u32,
-    stake_support: u16,
+    cert_epoch: u32,     
     message: vector<u8>,
+    stake_support: u16, // Metadata, not part of the actual certified message.
```
karlwuest, line 88: Ah, now I understand your previous comment 😄 makes sense 👍

contracts/walrus/sources/staking/staked_wal.move
karlwuest: I added this [on linear](https://linear.app/mysten-labs/project/disable-split-of-stakedwal-in-withdrawing-state-6b79c8bbf775) to keep track.

contracts/walrus/sources/staking/exchange_rate.move
karlwuest: I think using an explicit enum makes sense. Do you want to add this to this PR @benr-ml ?

benr-ml: SG, will refactor that in a follow up PR

karlwuest: Ok, in that case I'll stamp this one!

PR #1285, opened 2024-12-11
fix: Fix security issues with event blobs

sadhansood:
## Description

Fixes issues related to being able to increment certified checkpoint incorrectly and causing OOM by repeatedly attesting invalid blob ids.
## Test plan

Added test cases written by Ben

### DIFF
--- a/contracts/walrus/sources/system/event_blob.move
+++ b/contracts/walrus/sources/system/event_blob.move
@@ -23,9 +23,7 @@ public struct EventBlob has copy, drop, store {
 }
 
 /// State of event blob stream.
-#[allow(unused_field)]
-public struct EventBlobCertificationState has key, store {
-    id: UID,
+public struct EventBlobCertificationState has store {
     /// Latest certified event blob.
     latest_certified_blob: Option<EventBlob>,
     /// Current event blob being attested.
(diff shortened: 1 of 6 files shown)


contracts/walrus/tests/system/event_blob_tests.move
```diff
@@ -224,6 +224,119 @@ public fun test_event_blob_certify_change_epoch() {
     destroy(system);
 }
 
+#[test]
+public fun test_certify_invalid_blob_id() {
```
benr-ml, line 220: let's fix format and clean the tests a bit

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -430,7 +430,9 @@ public(package) fun certify_event_blob(
             ending_checkpoint_sequence_num,
             blob_id,
         );
-    self.event_blob_certification_state.stop_tracking_blob(blob_id);
```
karlwuest, line 433: Suggestion: I think the `stop_tracking_blob` function is now no longer needed. Can you remove that function?

sadhansood: done

contracts/walrus/sources/system/event_blob.move
```diff
@@ -24,8 +24,7 @@ public struct EventBlob has copy, drop, store {
 
 /// State of event blob stream.
 #[allow(unused_field)]
```
karlwuest: I think the `#[allow(unused_field)]` can be removed now. You'll probably also need to change the structs in rust (with the `mainnet-contracts` feature), to make sure the deserialization works (this is probably why the rust tests are failing).

PR #1293, opened 2024-12-12
refactor: convert PoolExchangeRate to an enum

benr-ml:
## Description

Convert `PoolExchangeRate` to an enum to be more explicit about its state.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/exchange_rate.move
+++ b/contracts/walrus/sources/staking/exchange_rate.move
@@ -15,7 +15,8 @@ public enum PoolExchangeRate has copy, drop, store {
     Variable {
         /// Amount of staked WAL tokens + rewards.
         wal_amount: u128,
-        /// Amount of total tokens in the pool (<= wal_amount, as long as slashing is not implemented).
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/staking/exchange_rate.move
```diff
@@ -5,56 +5,62 @@
...
-    /// Amount of staked WAL tokens this epoch.
-    wal_amount: u128,
-    /// Amount of total tokens in the pool this epoch.
-    pool_token_amount: u128,
+public enum PoolExchangeRate has copy, drop, store {
+    Flat, // one to one exchange rate
+    Variable {
+        /// Amount of staked WAL tokens + rewards.
+        wal_amount: u128,
+        /// Amount of total tokens in the pool (<= wal_amount).
```
karlwuest: The invariant that `pool_token_amount <= wal_amount` only holds as long as we don't have slashing. Can you add a comment that mentions this here and below where the check is performed?

PR #1294, opened 2024-12-12
fix(node): don't call epoch-change functions if they already happened

mlegner:
## Description

Currently, storage nodes call `initiate_epoch_change` and `end_voting` even if they were successfully called by a different node already.

With this change, they first check if calling the function is even necessary. It can still happen that between this check and the call it is triggered by another node, but it is much less likely.

## Test plan

Automatic tests and deployment to PTN.

### DIFF
--- a/crates/walrus-service/src/node/epoch_change_driver.rs
+++ b/crates/walrus-service/src/node/epoch_change_driver.rs
@@ -492,7 +492,16 @@ impl EpochOperation for VotingEndOperation {
     }
 
     async fn invoke(&self, contract: &dyn SystemContractService) -> Result<(), anyhow::Error> {
-        tracing::debug!("attempting to end voting");
+        let epoch_under_vote = self.epoch_under_vote.get();
+        let (current_epoch, state) = contract.get_epoch_and_state().await?;
+        if current_epoch >= epoch_under_vote
+            || (current_epoch + 1 == epoch_under_vote
+                && matches!(state, EpochState::NextParamsSelected(_)))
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -492,7 +492,15 @@ impl EpochOperation for VotingEndOperation {
     }
 
     async fn invoke(&self, contract: &dyn SystemContractService) -> Result<(), anyhow::Error> {
-        tracing::debug!("attempting to end voting");
+        let epoch_under_vote = self.epoch_under_vote.get();
+        let (current_epoch, state) = contract.get_epoch_and_state().await?;
+        if current_epoch + 1 >= epoch_under_vote
+            && matches!(state, EpochState::NextParamsSelected(_))
```
karlwuest: Important: I think this should be the following (otherwise the condition won't hold if we're e.g. in the first half of the next epoch)
```suggestion
        if current_epoch > epoch_under_vote
            || current_epoch == epoch_under_vote
            && matches!(state, EpochState::NextParamsSelected(_))
```

mlegner: Good point about the `||`, but I think we still need the `+ 1`, don't we? AFAIU the `epoch_under_vote` is the next epoch.

karlwuest: Oh, right, yes absolutely

PR #1295, opened 2024-12-12
feat: storage node support of new nodes joining cluster

halfprice:
## Description

There are mainly two things:

- new nodes will be in RecoveryCatchup state initially, but need to skip recovery and jump into normal shard sync when it just becomes a committee member
- Shard sync also needs to sync metadata

## Test plan

---

## Release notes

- [x] Storage node: Storage nodes that become committee members for the first time now sync the metadata for all live blobs before attempting to sync the gained shards.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -834,7 +834,7 @@ impl StorageNode {
         event_handle: EventHandle,
         event: &EpochChangeStart,
     ) -> anyhow::Result<()> {
-        // TODO: need to check if the node is lagging or not.
+        // TODO(WAL-479): need to check if the node is lagging or not.
 
         // Irrespective of whether we are in this epoch, we can cancel any scheduled calls to change
         // to or end voting for the epoch identified by the event, as we're already in that epoch.
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -131,8 +131,9 @@ impl ShardSyncHandler {
             }
         }
 
-        // TODO: create a end point that can transfer multiple blob metadata at once.
-        // TODO: do this in parallel to speed up the sync.
+        // TODO(WAL-478):
+        //   - create a end point that can transfer multiple blob metadata at once.
+        //   - do this in parallel to speed up the sync.
         for blob_info in blob_infos {
             let (blob_id, blob_info) = blob_info?;
 
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -680,7 +680,7 @@ impl ShardStorage {
 
             #[cfg(any(test, feature = "test-utils"))]
             {
-                assert!(node.storage.has_metadata(blob_id)?);
+                debug_assert!(node.storage.has_metadata(blob_id)?);
             }
 
             match sliver {


crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -63,6 +63,9 @@ pub enum ShardStatus {
     /// The shard is active in this node serving reads and writes.
     Active,
 
+    /// The shard is waiting for metadata recovery.
+    RecoverMetadata,
+
```
mlegner: Question: Why is this a per-shard status? This should either affect all or none of the shards AFAIU.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -449,7 +457,7 @@ impl ShardStorage {
         config: &ShardSyncConfig,
         directly_recover_shard: bool,
     ) -> Result<(), SyncShardClientError> {
-        tracing::info!(walrus.epoch = epoch, %directly_recover_shard, "syncing shard");
+        tracing::info!(walrus.epoch = epoch, %self.id, %directly_recover_shard, "syncing shard");
```
mlegner: Nit: I don't think this is needed as this is already added as `walrus.shard_index` in the `#[tracing::instrument(...)]` attribute above.

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -32,10 +33,99 @@ impl ShardSyncHandler {
...
+        recover_metadata: bool,
+    ) -> Result<(), SyncShardClientError> {
+        // Check if the shards are created in the storage.
+        for shard in shards.iter() {
+            let Some(shard_storage) = self.node.storage.shard_storage(*shard) else {
+                tracing::info!(
+                    walrus.shard_index = %shard,
+                    "shard is not assigned to this node; skipping sync"
+                );
+                return Err(ShardNotAssigned(*shard, self.node.current_epoch()).into());
```
mlegner: Question: How can this actually happen? And is it intentional to skip the whole sync when a single shard is not assigned?

crates/walrus-service/src/node.rs
```diff
@@ -852,7 +852,22 @@ impl StorageNode {
                     epoch = %event.epoch,
                     "processing event reaches the latest epoch, start recover entire node"
                 );
-                self.start_node_recovery(event_handle, event).await?;
+                let active_committees = self.inner.committee_service.active_committees();
+                if active_committees.previous_committee().is_none()
+                    || !active_committees
+                        .previous_committee()
+                        .unwrap()
+                        .contains(self.inner.public_key())
```
mlegner: Nit, to avoid the `unwrap`:
```suggestion
                if !active_committees
                    .previous_committee()
                    .is_some_and(|c| c.contains(self.inner.public_key()))
```

crates/walrus-service/src/node.rs
```diff
@@ -852,7 +852,22 @@ impl StorageNode {
...
+                {
+                    tracing::info!(
+                        "node is not in the previous committee, set node status to active"
+                    );
+                    self.inner.storage.set_node_status(NodeStatus::Active)?;
```
mlegner: Suggestion: Add two new global statuses: `Inactive` (for nodes that aren't part of any committee) and `NewCommitteeMember` or similar. With this global status we could remove the per-shard `RecoverMetadata` status and maybe some `new_node_joining` function parameters.

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -32,10 +33,99 @@ impl ShardSyncHandler {
...
+                return Err(SyncShardClientError::InvalidShardStatusToSync(
+                    *shard,
+                    shard_status,
+                ));
+            }
+
+            if recover_metadata {
+                // TODO: Add test to test crash at this point.
+                shard_storage.set_recover_metadata_status()?;
+            }
```
mlegner: As suggested above, we may be able to use a global status instead of this per-shard status.

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -95,6 +186,7 @@ impl ShardSyncHandler {
     /// Restarts syncing shards that were previously syncing. This method is used when restarting
     /// the node.
     pub async fn restart_syncs(&self) -> Result<(), anyhow::Error> {
```
mlegner, line 212: I think this could also be simplified assuming either all or no shards are waiting for metadata.

crates/walrus-service/src/node/storage/shard.rs
halfprice: This was created as per-shard status but actually used as a global state. Main reason was because I thought this is easier to accommodate for restarts. But after thinking about your other comments, I move this to a per node state.

crates/walrus-service/src/node/shard_sync.rs
halfprice: Good catch. This is one of the caveats that using `RecoverMetadata` as a shard state: the shard needs to be created so that we can set it to `RecoverMetadata`, and I simply copy pasted the old check to here, which was incorrect.

Now I've moved RecoverMetadata to a node state, and this whole parts before starting the Metadata recovery is gone. So I think it's much cleaner now.

crates/walrus-service/src/node.rs
halfprice: Great suggestion! The old implementation was kinda of ad-hoc in a sense that I was trying to fill holes in the new node joining pipeline. Your suggestion made me to try to think through how exactly should the node state transition be, which made me realized that there are much more nuance than I expected (e.g., whether a node is lagging or not, wether a node is in a committee or not, all adds additional dimension for each node state).

After thinking about this quite a bit, I now create 5 state for a storage node: Standby, RecoverMetadata, RecoveryCatchUp, RecoveryInProgress, and Active.

I documented the state transition here: https://www.notion.so/mystenlabs/Walrus-Storage-Node-Epoch-Change-5019443967f2456087cceda8fba94c86?pvs=4#10b6d9dcb4e980fe8a82c5ed94c65a55 , which you can use as a reference when reviewing the code.

Overall, I think the new version is much cleaner than the previous version (and also, after this thinking process, the previous version was actually implemented incorrect). So thanks for the suggestion!

crates/walrus-service/src/node.rs
```diff
@@ -835,6 +834,8 @@ impl StorageNode {
         event_handle: EventHandle,
         event: &EpochChangeStart,
     ) -> anyhow::Result<()> {
+        // TODO: need to check if the node is lagging or not.
```
halfprice: I'll create issue for all the TODOs once the approach looks good.

crates/walrus-service/src/node.rs
mlegner: Sound great!

Just one question about the state diagram: What state does a node get that starts up with an empty DB when it is already part of the current committee?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -674,6 +677,12 @@ impl ShardStorage {
             //TODO(#705): verify sliver validity.
             //  - blob is certified
             //  - metadata is correct
+
+            #[cfg(any(test, feature = "test-utils"))]
+            {
+                assert!(node.storage.has_metadata(blob_id)?);
+            }
```
mlegner: Question: Why not use a `debug_assert` here?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -604,4 +608,123 @@ mod tests {
 
         workload_handle.abort();
     }
+
+    #[ignore = "ignore E2E tests by default"]
+    #[walrus_simtest]
+    async fn test_new_node_joining_cluster() {
```
mlegner, line 614: AFAIU this tests the following:
- Node 5 is not in the initial committee.
- After a few epochs, node 5 is started.
- A bit later, it becomes part of the committee.

This is what we expect for the nodes we are going to onboard this week.

However, the following scenarios may also be relevant in same cases, for example when a node loses it's drives and has to restart with an empty DB:
- The node becomes part of the committee before it is started.
- The node becomes part of the committee and then gets/loses some shards, before it is started.

Can we test these as well, maybe in a follow-up?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -904,17 +913,15 @@ impl ShardStorage {
...
         );
 
-        let Some(metadata) = node.storage.get_metadata(&blob_id)? else {
+        let metadata = if let Some(metadata) = node.storage.get_metadata(&blob_id)? {
+            metadata
+        } else {
             if !node.is_blob_certified(&blob_id)? {
                 self.skip_recover_blob(blob_id, sliver_type, &node)?;
                 return Ok(());
             }
```
mlegner, line 922: Question: Do we even call this function for non-certified blobs? If not, we could directly use `get_or_recover_blob_metadata` instead of this conditional block.

crates/walrus-service/src/node.rs
halfprice: In that case, it needs to process all the events from beginning again (given it's empty DB), so the node will start in `Standby` mode, and then enter `RecoveryCatchUp`. When it is caught up, it'll enter the node recovery mode.

crates/walrus-simtest/tests/simtest.rs
halfprice: Those scenarios sound more like doing node recovery. Let's discuss it tomorrow in our 1:1

crates/walrus-service/src/node/storage/shard.rs
halfprice: The blob may be certified when we fetching it, and then become non-certified here. If we don't have this condition, we may fail getting blob metadata.

crates/walrus-service/src/node.rs

crates/walrus-service/src/node/storage/shard.rs
halfprice: Updated.

PR #1297, opened 2024-12-13
fix: shard sync should ignore non-certified blobs at the new epoch

halfprice:
## Description

The assumption that expired blobs are not deleted are already not true given that if a node performed a full recovery, it won't recover any expired blobs. Therefore, the `certified_blob_iterator` needs to also take the current certification status into consideration. Otherwise, produced item may not exist on the src node during shard think, which can falsely terminate the happy path shard sync on the destination node.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -127,7 +127,6 @@ impl BlobInfoTable {
                     .safe_range_iter((starting_blob_id_bound, Unbounded)),
             ),
             before_epoch,
-            true,
         )
     }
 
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -3800,17 +3800,24 @@ mod tests {
...
+                // Tests that there is a discrepancy between the source and destination shards in
+                // terms of certified blobs. If the source doesn't return any blobs, the destination
+                // should finish the sync process.
+                return_empty: ("fail_point_sync_shard_return_empty"),
+                // Tests that when direct shard sync request fails, the shard sync process will be
+                // retried using shard recovery.
+                return_error: ("fail_point_sync_shard_return_error")
+            ]
+        }
+        async fn sync_shard_src_abnormal_return(fail_point: &'static str) -> TestResult {
```
halfprice, line 3814: This is just a cleanup to merge two existing tests since they looked too similar. It has nothing to do with the rest of the PR...

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -222,11 +222,17 @@ where
             let Ok((_, blob_info)) = &item else {
                 return Some(item);
             };
+
+            // The iterator should return blobs that are certified before `before_epoch` and
+            // are valid and remain certified at `before_epoch`.
+            //
+            // It is important to only return certified blobs certified before `before_epoch`
+            // because we don't want to fetch blobs that are just certified at `before_epoch`.
             if !self.only_certified
```
mlegner: Remark: The semantics are a bit unclear if `only_certified == false`; however, I think we don't currently use that at all. Maybe it would be best to remove that field until we use it and then define the semantics properly?

halfprice: Make sense. I don't remember why that option was there. It was indeed not used.

PR #1298, opened 2024-12-13
feat(node): add per-object blob-info DB table

mlegner:
## Description

This adds a database table to track individual `Blob` objects. This will subsequently be used to update the aggregate blob info on epoch change (WAL-473) and to be able to add the object ID to the certification message of deletable blobs (WAL-301).

~~It (ab)uses the `mainnet-contracts` feature to ensure this is not a breaking change for the current Testnet. Maybe we should change that feature name to something like `walrus-mainnet`?~~

It uses the `walrus-mainnet` feature (renamed in #1308) to ensure this is not a breaking change for the current Testnet.

Closes WAL-257

## Test plan

The general behavior is covered by existing tests. Additional unit tests will be added here and/or as part of WAL-258.

---

## Release notes

This does not become active immediately as the added features are gated behind the `mainnet-contracts` feature.

### DIFF
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -134,7 +134,10 @@ impl BlobInfoTable {
         tracing::debug!(?operation, "updating blob info");
 
         let mut batch = self.aggregate_blob_info.batch();
-        self.merge_blob_info_batch(&mut batch, &event.blob_id(), &operation)?;
+        batch.partial_merge_batch(
+            &self.aggregate_blob_info,
+            [(event.blob_id(), operation.to_bytes())],
+        )?;
         #[cfg(feature = "walrus-mainnet")]
         if let Some(object_id) = event.object_id() {
             let per_object_operation =
@@ -153,18 +156,19 @@ impl BlobInfoTable {
         latest_handled_index.map_or(false, |i| event_index <= i)
     }
 
-    pub fn merge_blob_info_batch(
+    pub fn set_metadata_stored<'a>(
         &self,
-        batch: &mut DBBatch,
+        batch: &'a mut DBBatch,
         blob_id: &BlobId,
-        operand: &BlobInfoMergeOperand,
-    ) -> Result<(), TypedStoreError> {
-        batch.partial_merge_batch(&self.aggregate_blob_info, [(blob_id, operand.to_bytes())])?;
-        Ok(())
-    }
-
-    pub fn delete(&self, batch: &mut DBBatch, blob_id: &BlobId) -> Result<(), TypedStoreError> {
-        batch.delete_batch(&self.aggregate_blob_info, [blob_id])
+        metadata_stored: bool,
+    ) -> Result<&'a mut DBBatch, TypedStoreError> {
+        batch.partial_merge_batch(
+            &self.aggregate_blob_info,
+            [(
+                blob_id,
+                &BlobInfoMergeOperand::MarkMetadataStored(metadata_stored).to_bytes(),
+            )],
+        )
     }
 
     /// Returns an iterator over all blobs that were certified before the specified epoch in the
@@ -203,7 +207,7 @@ impl BlobInfoTable {
         operand: &BlobInfoMergeOperand,
     ) -> Result<(), TypedStoreError> {
         let mut batch = self.batch();
-        self.merge_blob_info_batch(&mut batch, blob_id, operand)?;
+        batch.partial_merge_batch(&self.aggregate_blob_info, [(blob_id, operand.to_bytes())])?;
         batch.write()
     }
 
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -822,7 +822,7 @@ impl StorageNode {
         self.blob_sync_handler
             .cancel_sync_and_mark_event_complete(&event.blob_id)
             .await?;
-        self.inner.storage.delete_blob(&event.blob_id, false)?;
+        self.inner.storage.delete_blob_data(&event.blob_id)?;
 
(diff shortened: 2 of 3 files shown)


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -24,22 +26,38 @@ use walrus_core::{BlobId, Epoch};
...
+#[cfg(feature = "walrus-mainnet")]
+use self::per_object_blob_info::{PerObjectBlobInfo, PerObjectBlobInfoMergeOperand};
 use super::{database_config::DatabaseTableOptions, DatabaseConfig};
 
 #[derive(Debug, Clone)]
 pub(super) struct BlobInfoTable {
-    blob_info: DBMap<BlobId, BlobInfo>,
+    aggregate_blob_info: DBMap<BlobId, BlobInfo>,
+    #[cfg(feature = "walrus-mainnet")]
+    per_object_blob_info: DBMap<ObjectID, PerObjectBlobInfo>,
```
karlwuest, line 37: Question: What is the reason to gate this behind the feature? Do we get an inconsistent state if we would add this table now? Or is this because the change on `BlobStatusChangeInfo` will break the WAL of the db?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1141,28 +1215,192 @@ impl Mergeable for BlobInfo {
...
+                change_info,
+            }: PerObjectBlobInfoMergeOperand,
+        ) -> Self {
+            assert_eq!(self.blob_id, change_info.blob_id);
+            assert_eq!(self.deletable, change_info.deletable);
+            assert!(!self.deleted);
+            self.event = change_info.status_event;
+            match change_type {
+                BlobStatusChangeType::Register => {
+                    unreachable!("cannot register an already registered blob");
```
karlwuest: Nit: this should probably not be `unreachable!`, but just `panic!`.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: If we add this to the existing testnet, we would have to backfill the data based on old events. Otherwise, we could run into issues when nodes process a certified event without a previous registered event, etc. And given that we will have a Testnet redeployment soon, I didn't want to spend all the effort to do the backfilling properly.

karlwuest: Makes sense 👍

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -834,6 +908,7 @@ impl PermanentBlobInfoV1 {
             end_epoch: new_end_epoch,
             status_event: new_status_event,
             deletable,
+            ..
```
halfprice: Nit: should we be more explicit here using `walrus-mainnet` feature and blob_id? Not sure if this is important or not. We may overlook this place when removing `walrus-mainnet` feature.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1011,6 +1086,7 @@ impl Mergeable for BlobInfoV1 {
                     epoch: _,
                     end_epoch,
                     status_event,
+                    ..
```
halfprice: Same as above

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: May also worth adding to the message or add a comment here that the event idempotency has been addressed when reaching to this point.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -89,8 +133,18 @@ impl BlobInfoTable {
...
 
-        let mut batch = self.blob_info.batch();
+        let mut batch = self.aggregate_blob_info.batch();
         self.merge_blob_info_batch(&mut batch, &event.blob_id(), &operation)?;
+        #[cfg(feature = "walrus-mainnet")]
+        if let Some(object_id) = event.object_id() {
+            let per_object_operation =
+                PerObjectBlobInfoMergeOperand::from_blob_info_merge_operand(operation)
+                    .expect("we know this is a registered, certified, or deleted event");
+            batch.partial_merge_batch(
```
halfprice, line 146: Important: this op needs to be in the same DB batch as updating aggregate_blob_info. Otherwise, there will be a moment of inconsistency between per object table and aggregate_blob_info. Especially, if db batch fails to write, but we have updated the per object table, re-processing the same event may cause assertion failures.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Thanks for calling out my laziness here, you are absolutely right. 😅

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Good point, but AFAIU my code does that already. The batch is created in line 136, the aggregate blob info is added in line 137 (inside `merge_blob_info_batch`), then the detailed blob info is updated here. Let me know if I misunderstood something.

halfprice: Oh Yeah, you are absolutely right! Merging into BlobInfo uses a helper function `merge_blob_info_batch` and here we directly added to the batch. I missed this and thought only one of them is operating on the batch.

Maybe consider removing `merge_blob_info_batch` then 😊. Not important. Up to you.

mlegner: That function is used in multiple locations. But looking at it, it probably shouldn't be. 😬 
Let me refactor this a bit.

PR #1300, opened 2024-12-15
fix: handle_sync_shard_request should respect the sliver count in the request

halfprice:
## Description

The semantic of `SyncShardRequest` is that the des expects the src to return SyncShardRequest.sliver_count number of slivers starting at starting_blob_id. However, currently, the src storage fetches `sliver_count` number of BlobInfo first and then fetch slivers. If the slivers are not stored, less slivers are returned. The problematic situation is that if all slivers from the result of reading BlobInfo is not stored, the request returns empty slivers, which made the dst node to think that no more slivers can be returned and quit happy path shard sync. Even though there are still valid blobs after this batch.

This change fixed the above problem by always trying to return SyncShardRequest.sliver_count of slivers until no more slivers can be returned.

Contribute WAL-195

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -5,7 +5,7 @@ use core::fmt::{self, Display};
 use std::{
     collections::{hash_map::Entry, HashMap},
     fmt::Debug,
-    ops::Bound::Included,
+    ops::Bound::{Excluded, Included},
     path::Path,
     sync::{Arc, RwLock, TryLockError},
 };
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/storage.rs
```diff
@@ -530,20 +530,36 @@ impl Storage {
...
+
+            if blobs_to_fetch.is_empty() {
+                // No more blobs to fetch.
+                break;
+            }
+
+            // Update last fetched ID for next iteration
+            last_fetched_blob_id = blobs_to_fetch.last().cloned();
+
+            let mut slivers = shard.fetch_slivers(request.sliver_type(), &blobs_to_fetch)?;
```
halfprice, line 558: I thought about just changing the logic to `fetch blobinfo and then fetch blob` one after another. But I think we still want to leverage the MultiGet in fetch_slivers to improve DB read performance. So I ended up using this batched way of getting slivers.

PR #1301, opened 2024-12-15
feat: add initial custom upgrade policy

karlwuest:
## Description

contributes to WAL-357

Follow-up:
- Quorum-based upgrades

## Test plan

- unit test in move for custom upgrade policy
- existing tests
- will work on adding integration tests for upgrades with WAL-481

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/init.move
+++ b/contracts/walrus/sources/init.move
@@ -3,8 +3,14 @@
 
 module walrus::init;
 
-use sui::clock::Clock;
-use walrus::{staking, system};
+use std::type_name;
+use sui::{clock::Clock, package::UpgradeCap};
+use walrus::{events, staking::{Self, Staking}, system::{Self, System}, upgrade};
+
+// Error codes
+// Error types in `walrus-sui/types/move_errors.rs` are auto-generated from the Move error codes.
+const EInvalidMigration: u64 = 0;
+const EInvalidUpgradeCap: u64 = 1;
 
 /// Must only be created by `init`.
 public struct InitCap has key, store {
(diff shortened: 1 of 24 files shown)


contracts/walrus/sources/staking.move
```diff
@@ -259,6 +271,43 @@ public fun withdraw_stake(
...
+
+/// Returns the current epoch of the staking object.
+public fun epoch(staking: &Staking): u32 {
+    staking.inner().epoch()
+}
+
+// === Upgrade ===
+
+public(package) fun set_new_package_id(staking: &mut Staking, new_package_id: ID) {
+    staking.new_package_id = option::some(new_package_id);
```
damirka, line 305: Should it override or "fill"?

contracts/walrus/sources/staking.move
```diff
@@ -259,6 +271,43 @@ public fun withdraw_stake(
...
+public(package) fun migrate(staking: &mut Staking) {
+    assert!(staking.version < VERSION, EInvalidMigration);
+
+    // Move the old system state inner to the new version.
+    let staking_inner: StakingInnerV1 = df::remove(&mut staking.id, staking.version);
+    df::add(&mut staking.id, VERSION, staking_inner);
+    staking.version = VERSION;
+
+    // Set the new package id.
+    staking.package_id = staking.new_package_id.extract();
```
damirka, line 322: Missing explicit check for is_some

contracts/walrus/sources/staking.move
karlwuest: It should override. E.g. if we have an upgrade following another upgrade without calling migrate in between, then this should be set to the latest upgraded version.

contracts/walrus/sources/init.move
```diff
@@ -24,17 +28,40 @@ fun init(ctx: &mut TxContext) {
 /// This can only be called once, after which the `InitCap` is destroyed.
 /// TODO: decide what to add as system parameters instead of constants.
 public fun initialize_walrus(
```
manolisliolios, line 32: nit: even though this is done in a controlled environment, I'd recommend we check that the supplied UpgradeCap is the correct one

karlwuest: How can we check this?

contracts/walrus/sources/upgrade.move
manolisliolios, line 1: Do we expect more than 1 `EmergencyUpgradeCap` objects to ever be created from this package? (asking because I see the M:N checks)

contracts/walrus/sources/init.move
manolisliolios: Could be done by checking against the "address" of `InitCap` type: `type_name::get<InitCap>().get_address() == package.address()`

karlwuest: Oh, I didn't know this was possible. Nice!

contracts/walrus/sources/upgrade.move
karlwuest: Do you mean the check for the id of the upgrade manager?
No, this is just a sanity check. We should never have more than 1 `EmergencyUpgradeCap` or `UpgradeManager`.

manolisliolios: Yep that's why I was asking! Thanks!

contracts/walrus/sources/init.move
karlwuest: added!

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -501,13 +539,50 @@ impl SuiReadClient {
...
+        }
+        let inner = self
+            .sui_client
+            .get_dynamic_field_object::<u64, SystemStateInnerV1>(
+                self.system_object_id,
+                TypeTag::U64,
+                version,
+            )
+            .await?;
+        let system_object = SystemObject {
```
giac-mysten: Nit: this is defined to be immediately returned.

PR #1303, opened 2024-12-16
feat: Add ability to stop serving blobs in deny list

sadhansood:
## Description

This PR adds the ability for stoarge nodes to stop service blobs added to deny list. 

closes WAL-423

## Test plan

Added end to end test


crates/walrus-service/src/client/blocklist.rs
mlegner, line 1: If we use this in both client and node, this should be moved to the `common` module.

crates/walrus-service/src/client/blocklist.rs
```diff
@@ -32,37 +41,132 @@ impl Blocklist {
...
     }
 
     /// Checks if a blob ID is blocked.
     #[inline]
     pub fn is_blocked(&self, blob_id: &BlobId) -> bool {
-        self.blocked_blobs.contains(blob_id)
+        let guard = self
+            .blocked_blobs
+            .read()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
```
karlwuest: Suggestion: I think if the RWLock is poisoned, we probably want to panic here, because that means another thread that was writing to it panicked.

crates/walrus-service/src/client/blocklist.rs
```diff
@@ -32,37 +41,132 @@ impl Blocklist {
...
     ///
     /// Returns whether the ID was newly inserted.
     #[inline]
-    pub fn insert(&mut self, blob_id: BlobId) -> bool {
-        self.blocked_blobs.insert(blob_id)
+    pub fn insert(&mut self, blob_id: BlobId) -> Result<bool> {
+        let mut guard: std::sync::RwLockWriteGuard<'_, HashSet<BlobId>> = self
+            .blocked_blobs
+            .write()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
```
karlwuest: Important: This should definitely panic, because here we are writing to the locked data.

crates/walrus-service/src/client/blocklist.rs
```diff
@@ -32,37 +41,132 @@ impl Blocklist {
...
     ///
     /// Returns whether the ID was previously blocked.
     #[inline]
-    pub fn remove(&mut self, blob_id: &BlobId) -> bool {
-        self.blocked_blobs.remove(blob_id)
+    pub fn remove(&mut self, blob_id: &BlobId) -> Result<bool> {
+        let mut guard: std::sync::RwLockWriteGuard<'_, HashSet<BlobId>> = self
+            .blocked_blobs
+            .write()
+            .unwrap_or_else(|poisoned| poisoned.into_inner());
```
karlwuest: Important: Same as above, this should panic.

crates/walrus-service/src/client/blocklist.rs
```diff
@@ -32,37 +41,132 @@ impl Blocklist {
...
+
+        let mut guard = self.blocked_blobs.write().expect("mutex poisoned");
+        let old_blobs = guard.iter().cloned().collect::<HashSet<_>>();
+        let new_blobs = blocklist.0.iter().cloned().collect::<HashSet<_>>();
+        let added_blobs = new_blobs
+            .symmetric_difference(&old_blobs)
+            .collect::<Vec<_>>();
+        let removed_blobs = old_blobs
+            .symmetric_difference(&new_blobs)
+            .collect::<Vec<_>>();
```
karlwuest: Important: I think this doesn't do what it should and specifically `added_blobs == removed_blobs`. This should probably be: 
```suggestion
        let added_blobs = new_blobs
            .difference(&old_blobs)
            .collect::<Vec<_>>();
        let removed_blobs = old_blobs
            .difference(&new_blobs)
            .collect::<Vec<_>>();
```

crates/walrus-service/src/test_utils.rs
```diff
@@ -1839,13 +1863,25 @@ pub mod test_cluster {
...
             contract_clients.push(client);
+            blocklist_files.push(
+                blocklist_dir
+                    .clone()
+                    .unwrap_or(
+                        tempfile::tempdir()
+                            .expect("temporary directory creation must succeed")
+                            .path()
+                            .to_path_buf(),
+                    )
```
karlwuest: Question/Nit: Do we need one tempdir per node? If not, you can move the `unwrap_or` outside of the loop and pass the same directory for all to improve readability.

Nit 2: `unwrap_or_else` may be better here.

Important: Once the tempdir goes out of scope it is deleted, so we're only left with the path, but no directory with that path. I think in other places we use `Box::leak` to make sure the directory remains until the program exits. Alternatively, you could use `client.temp_dir` to also store the blocklist file. On some platforms it works as is, because writing to the file (with the `create` option) will also create the path, but afaik that is not guaranteed behaviour.

crates/walrus-service/src/client.rs
mlegner, line 1: Nit (maybe for a follow-up): We should try to deduplicate some code between `request_slivers_and_decode ` and `decode_sliver_by_sliver`.

crates/walrus-service/src/node/server/responses.rs
mlegner, line 1: Suggestion: We should consider also blocking the metadata for blocked blobs and/or add this to the blob status.

crates/walrus-service/src/test_utils.rs
sadhansood: Fixed

crates/walrus-service/src/client/blocklist.rs
sadhansood: done

crates/walrus-service/src/node/server/responses.rs
sadhansood: done

crates/walrus-service/src/client.rs
sadhansood: will do this in a follow-up

PR #1306, opened 2024-12-16
feat: Add function to contracts to allow adding rewards

georgedigkas:
## Description

This PR adds function to contracts to allow adding rewards

## Test plan

Move tests added

### DIFF
--- a/contracts/walrus/sources/staking.move
+++ b/contracts/walrus/sources/staking.move
@@ -259,13 +259,6 @@ public fun withdraw_stake(
     staking.inner_mut().withdraw_stake(staked_wal, ctx)
 }
 
-/// Adds rewards to the system for the specified number of epochs ahead.
-/// The rewards are split equally across the future accounting ring buffer up to the
-/// specified epoch.
-public fun add_subsidy(system: &mut System, subsidy: Coin<WAL>, epochs_ahead: u32) {
-    system.add_subsidy(subsidy, epochs_ahead);
-}
-
 // === Internals ===
 
 /// Get a mutable reference to `StakingInner` from the `Staking`.
--- a/contracts/walrus/sources/system.move
+++ b/contracts/walrus/sources/system.move
@@ -143,6 +143,9 @@ public fun extend_blob(
     self.inner_mut().extend_blob(blob, epochs_ahead, payment);
 }
 
+/// Adds rewards to the system for the specified number of epochs ahead.
+/// The rewards are split equally across the future accounting ring buffer up to the
+/// specified epoch.
 public fun add_subsidy(system: &mut System, subsidy: Coin<WAL>, epochs_ahead: u32) {
     system.inner_mut().add_subsidy(subsidy, epochs_ahead)
 }


contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -434,6 +435,40 @@ public(package) fun certify_event_blob(
...
+) {
+    // Check the period is within the allowed range.
+    assert!(epochs_ahead > 0, EInvalidEpochsAhead);
+    assert!(epochs_ahead <= self.future_accounting.max_epochs_ahead(), EInvalidEpochsAhead);
+
+    let mut subsidy_balance = subsidy.into_balance();
+    let reward_per_epoch = subsidy_balance.value() / (epochs_ahead as u64);
+    let leftover_rewards = subsidy_balance.value() % (epochs_ahead as u64);
+
+    0u32.range_do!(epochs_ahead, |i| {
```
karlwuest: Nit:
```suggestion
    epochs_ahead.do!(|i| {
```

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -434,6 +435,40 @@ public(package) fun certify_event_blob(
...
+    // Add leftover rewards to the first epoch's accounting.
+    if (leftover_rewards > 0) {
+        let accounts = self.future_accounting.ring_lookup_mut(0);
+        let rewards_balance = accounts.rewards_balance();
+        rewards_balance.join(subsidy_balance.split(leftover_rewards));
+    };
+
+    // TODO: is there a better way to destroy the subsidy balance without using assert?
+    assert!(subsidy_balance.value() == 0, ESubsidyBalanceNotZero);
+    subsidy_balance.destroy_zero();
```
karlwuest: Suggestion: We can just join the remainder directly, even if it's zero:
```suggestion
    // Add leftover rewards to the first epoch's accounting.
    self
        .future_accounting.ring_lookup_mut(0)
        .rewards_balance();
        .join(subsidy_balance);
```

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -434,6 +435,40 @@ public(package) fun certify_event_blob(
     blob.burn();
 }
 
+/// Adds rewards to the system for the specified number of epochs ahead.
+/// The rewards are split equally across the future accounting ring buffer up to the
+/// specified epoch.
+public(package) fun add_rewards(
```
karlwuest: Nit: maybe `add_subsidy` is a better name.

karlwuest: Important: You also need to add a public function on ~`staking`~ `system` to call this function. This is needed to allow the subsidy contract to live in a separate package.

Edit: sorry, `system` not `staking`

contracts/walrus/tests/system/system_state_inner_tests.move
```diff
@@ -0,0 +1,123 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+#[test_only]
+module walrus::system_state_inner_tests;
+
+use sui::{clock, test_utils::destroy};
+use walrus::{storage_accounting as sa, system_state_inner, test_utils::mint};
+
+#[test]
+fun test_add_rewards_zero_rewards() {
```
karlwuest: Suggestion: The first three tests are essentially the same code. Instead you could have one function that takes parameters for these tests (reward, epochs_ahead) and then another function that calls the parametrized function a few times with different parameters. It would then also be good to handle the case where the reward is not cleanly divisible by the number of epochs, then the 4th test can also be covered with the same function.

contracts/walrus/sources/system/system_state_inner.move
georgedigkas: indeed! Renamed to `add_subsidy`

contracts/walrus/tests/system/system_state_inner_tests.move
georgedigkas: Good idea, tests refactored!

contracts/walrus/sources/system/system_state_inner.move
georgedigkas: Thanks @karlwuest  `add_subsidy` added to `walrus::system` module

PR #1313, opened 2024-12-16
feat: adds deny list functionality to system + move rewards distribution to system

damirka:
## Description

- depends on #1348
- moves rewards distribution to _system_
- staking now takes a `VecMap<ID, Balance<WAL>>` instead of a `Balance`
- cleanups in tests
- minor naming patches

## Test plan

- adds an e2e test for rewards distribution
- adds a (de)serialization test for new message types

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/types/events.rs
+++ b/crates/walrus-sui/src/types/events.rs
@@ -4,7 +4,6 @@
 //! Walrus event type bindings. Replicates the move event types in Rust.
 
 use anyhow::anyhow;
-use move_core_types::u256::U256;
 use serde::{Deserialize, Serialize};
 use sui_sdk::rpc_types::SuiEvent;
 use sui_types::{base_types::ObjectID, event::EventID};
@@ -628,7 +627,7 @@ pub struct RegisterDenyListUpdateEvent {
     /// The epoch in which the deny list update was registered.
     pub epoch: Epoch,
     /// The root hash of the deny list update.
-    pub root: U256,
+    pub root: [u8; 32],
     /// The deny list update ID.
     pub sequence_number: u64,
     /// The ID of the event.
@@ -665,7 +664,7 @@ pub struct DenyListUpdateEvent {
     /// The epoch in which the deny list update was registered.
     pub epoch: Epoch,
     /// The root hash of the deny list update.
-    pub root: U256,
+    pub root: [u8; 32],
     /// The deny list update ID.
     pub sequence_number: u64,
     /// The ID of the event.
--- a/crates/walrus-sui/src/types/move_structs.rs
+++ b/crates/walrus-sui/src/types/move_structs.rs
@@ -6,7 +6,6 @@
 use std::{fmt::Display, num::NonZeroU16};
 
 use fastcrypto::traits::ToFromBytes;
-// use move_core_types::u256::U256;
 use serde::{
     de::{DeserializeOwned, Error},
     Deserialize,


contracts/walrus/sources/system/blob.move
damirka, line 1: Minor refactoring of messages to match the payload in the signed message.

contracts/walrus/sources/system/events.move
```diff
@@ -86,6 +86,26 @@ public struct ContractUpgraded has copy, drop {
     version: u64,
 }
 
+/// Signals that a Denylist update has started.
+public struct RegisterDenyListUpdate has copy, drop {
```
karlwuest, line 90: Important: Add the epoch here and in `DenyListUpdate` as well. Each Walrus event should contain the epoch in which it was emitted.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -55,6 +56,8 @@ public struct SystemStateInnerV1 has key, store {
     future_accounting: FutureAccountingRingBuffer,
     /// Event blob certification state
     event_blob_certification_state: EventBlobCertificationState,
+    deny_list_sizes: ExtendedField<VecMap<ID, u64>>,
+    leftover_rewards: Balance<WAL>,
```
karlwuest: Suggestion: Instead of storing the leftover rewards separately, just merge them into the FutureAccounting for the next epoch.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -111,7 +115,38 @@ public(package) fun advance_epoch(
...
-    accounts_old_epoch.unwrap_balance()
+
+    let mut total_rewards = accounts_old_epoch.unwrap_balance();
+    total_rewards.join(self.leftover_rewards.withdraw_all());
+
+    // === Rewards distribution ===
+
+    // for all nodes: stored[node_idx] = weight[node_idx] * (used_capacity - deny_list_size[node_idx])
+    // total_stored = sum(stored)
+    // for all nodes: reward_per_node[node_idx] = stored[node_idx]*total_reward_value / total_stored
```
karlwuest: Suggestion: Instead of this pseudocode, add a short explanation of what we're doing here. Ideally, also come up with better names, especially for `total_stored` because that is not literally what is stored. What the nodes store (and the total_stored) would actually be the current value divided by the total number of shards. However, to keep rounding errors as small as possible, we only do the division afterwards when computing rewards, because the ratios are what matters for the "sizes".

Also, add a comment somewhere to the "deny_list_size" that explains that this is the sum of the total encoded sizes of the blobs on the deny list.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -503,18 +529,95 @@ public(package) fun n_shards(self: &SystemStateInnerV1): u16 {
...
+
+    let epoch = certified_message.cert_epoch();
+    let message = certified_message.deny_list_update_message();
+    let node_id = message.storage_node_id();
+    let size = message.size();
+
+    assert!(epoch == self.epoch(), EInvalidIdEpoch);
+    assert!(node_id == cap.node_id());
+    assert!(cap.deny_list_sequence() < message.sequence_number());
+    assert!(cap.deny_list_root() != message.root());
```
karlwuest: Important: Remove this check, the root may be the same. It's possible that we have an update in which only the size is updated, e.g. because some blobs expired or were deleted.

contracts/walrus/sources/system/system_state_inner.move
damirka: Added in 1035cf3, can you please look at this commit to check if I got it right?

contracts/walrus/tests/e2e_tests.move
```diff
@@ -730,3 +783,16 @@ public fun test_emergency_upgrade() {
     test_scenario::return_to_address(admin, emergency_cap);
     runner.destroy();
 }
+
+// local alias for simplicity
+use fun quorum_sign as vector.quorum_sign;
+
+fun quorum_sign(nodes: &vector<TestStorageNode>, message: vector<u8>): (vector<u8>, vector<u8>) {
```
karlwuest: Nit: This gives a signature from all nodes instead of just a quorum, so maybe rename to a more fitting name?

damirka: Done. Now it's just `sign`.

crates/walrus-sui/src/types/events.rs
```diff
@@ -317,6 +361,10 @@ impl TryFrom<SuiEvent> for BlobEvent {
             contracts::events::BlobRegistered => Ok(BlobEvent::Registered(value.try_into()?)),
             contracts::events::BlobCertified => Ok(BlobEvent::Certified(value.try_into()?)),
             contracts::events::InvalidBlobID => Ok(BlobEvent::InvalidBlobID(value.try_into()?)),
+            contracts::events::BlobDeleted => Ok(BlobEvent::Deleted(value.try_into()?)),
```
karlwuest, line 363: Did we not have this in here until now? 😳

damirka: Yup, we did not have it until now.

crates/walrus-service/src/node/metrics.rs
```diff
@@ -181,6 +182,7 @@ impl TelemetryLabel for ContractEvent {
             ContractEvent::BlobEvent(event) => event.label(),
             ContractEvent::EpochChangeEvent(event) => event.label(),
             ContractEvent::PackageEvent(event) => event.label(),
+            ContractEvent::DenyListEvent(_event) => todo!("DenyListEvent"),
```
karlwuest: Suggestion: Implement `TelemetryLabel` for `DenyListEvent` and then just do:
```suggestion
            ContractEvent::DenyListEvent(event) => event.label(),
```

This should be relatively simple and will give us logging for these events and make sure we don't crash if we receive one.

crates/walrus-service/src/node.rs
```diff
@@ -857,6 +857,9 @@ impl StorageNode {
                 self.process_package_event(event_handle, package_event)
                     .await?;
             }
+            EventStreamElement::ContractEvent(ContractEvent::DenyListEvent(_event)) => {
+                todo!("DenyListEvent event handling");
```
karlwuest: Suggestion: I think, it's better here to just ignore the event (and mark it as complete) instead of panicking. The reason for this is that any registered node can trigger such an event, which would cause all nodes to crash. 

Then you can add a TODO as comment and refer to the issue on linear to implement the deny list.

Edit: also add the above explanation as a comment to why we should not have a panic here.

crates/walrus-service/src/node.rs
```diff
@@ -888,6 +891,9 @@ impl StorageNode {
             BlobEvent::InvalidBlobID(event) => {
                 self.process_blob_invalid_event(event_handle, event).await?;
             }
+            BlobEvent::DenyListBlobDeleted(_) => {
+                todo!("DenyListBlobDeleted event handling");
```
karlwuest, line 900: Suggestion: Can you add a comment referencing the linear issue? 
And also add a comment explaining that it's fine to panic here with a todo!, because in order to trigger this event, we need f+1 signatures and until the rust integration is implemented no such event should be emitted.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -518,6 +518,7 @@ impl From<&BlobEvent> for BlobInfoMergeOperand {
             BlobEvent::Certified(event) => event.into(),
             BlobEvent::Deleted(event) => event.into(),
             BlobEvent::InvalidBlobID(event) => event.into(),
+            BlobEvent::DenyListBlobDeleted(_) => todo!("DenyListBlobDeleted"),
```
karlwuest: See my comment above: Add comment to reference linear issue and explain why it's fine to panic here.

crates/walrus-service/src/node.rs
damirka: Done. Thank you!

crates/walrus-service/src/node.rs
```diff
@@ -858,7 +858,8 @@ impl StorageNode {
                     .await?;
             }
             EventStreamElement::ContractEvent(ContractEvent::DenyListEvent(_event)) => {
-                todo!("DenyListEvent event handling");
+                // TODO: Implement DenyListEvent handling
```
karlwuest: Nit: 
```suggestion
                // TODO: Implement DenyListEvent handling (WAL-424)
```

crates/walrus-service/src/node/metrics.rs
```diff
@@ -182,7 +182,7 @@ impl TelemetryLabel for ContractEvent {
             ContractEvent::BlobEvent(event) => event.label(),
             ContractEvent::EpochChangeEvent(event) => event.label(),
             ContractEvent::PackageEvent(event) => event.label(),
-            ContractEvent::DenyListEvent(_event) => todo!("DenyListEvent"),
+            ContractEvent::DenyListEvent(event) => event.label(),
```
karlwuest, line 201: Question: Does this compile? I think you'll need to add an impl for the `TelemetryLabel`, e.g.:

````
impl TelemetryLabel for DenyListEvent {
    fn label(&self) -> &'static str {
        match self {
            RegisterDenyListUpdate(_) => "register-deny-list-update",
            DenyListUpdate(_) => "deny-list-update",
        }
    }
}
````

Comment (damirka) [reply crates/walrus-service/src/node/metrics.rs:201]: My bad, was testing build a wrong crate. All fixed, and now builds.

File: crates/walrus-service/src/node/metrics.rs [hunk 13]
```diff
@@ -176,6 +184,15 @@ impl TelemetryLabel for PackageEvent {
     }
 }
 
+impl TelemetryLabel for DenyListEvent {
+    fn label(&self) -> &'static str {
+        match self {
+            DenyListEvent::DenyListUpdate(_) => "deny-list-updated",
+            DenyListEvent::RegisterDenyListUpdate(_) => "deny-list-deleted",
```
karlwuest:
````suggestion
            DenyListEvent::RegisterDenyListUpdate(_) => "register-deny-list-update",
````

PR #1318, opened 2024-12-17
feat(staking): adds node_metadata module and type

damirka:
## Description

- closes WAL-428
- adds new util `extended_field` which moves the value away from object but preserves the type
- adds `NodeMetadata` which is required to create a new staking pool
- moves ActiveSet into an `ExtendedField`, temporarily disables read API

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking.move
+++ b/contracts/walrus/sources/staking.move
@@ -153,7 +153,7 @@ public fun set_node_capacity_vote(self: &mut Staking, cap: &StorageNodeCap, node
 // === Get/ Update Node Parameters ===
 
 /// Get `NodeMetadata` for the given node.
-public fun node_metadata(self: &mut Staking, node_id: ID): NodeMetadata {
+public fun node_metadata(self: &Staking, node_id: ID): NodeMetadata {
     self.inner().node_metadata(node_id)
 }
 
--- a/contracts/walrus/tests/e2e_tests.move
+++ b/contracts/walrus/tests/e2e_tests.move
@@ -500,14 +500,14 @@ fun node_update_metadata() {
     runner.tx!(node.sui_address(), |staking, _, _| {
         let mut metadata = staking.node_metadata(node.node_id());
         metadata.set_description(b"Tusk Crew".to_string());
-        metadata.set_project_url(b"https://crew.walrus.sites/".to_string());
+        metadata.set_project_url(b"https://crew.walrus.site/".to_string());
         staking.set_node_metadata(node.cap(), metadata);
     });
 
     runner.tx!(node.sui_address(), |staking, _, _| {
         let metadata = staking.node_metadata(node.node_id());
         assert_eq!(metadata.description(), b"Tusk Crew".to_string());
-        assert_eq!(metadata.project_url(), b"https://crew.walrus.sites/".to_string());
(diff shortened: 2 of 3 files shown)


contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -80,6 +82,8 @@ public struct StakingPool has key, store {
     voting_params: VotingParams,
     /// The storage node info for the pool.
     node_info: StorageNodeInfo,
+    /// The metadata of the storage node.
+    node_metadata: ExtendedField<NodeMetadata>,
```
karlwuest: Question: What do you think about storing this in the `node_info` instead of the staking pool?

damirka: I have a counter-question. What do you think about storing them together in a dynamic field? Does it complicate things?

karlwuest: You mean one extended field for StorageNodeInfo + NodeMetadata?

The only concrete issue that comes to mind is that we need the bls keys during epoch change, so those should not be in the dynamic field. But I think we could move the name into the NodeMetadata.

contracts/walrus/sources/staking.move
```diff
@@ -150,7 +150,12 @@ public fun set_node_capacity_vote(self: &mut Staking, cap: &StorageNodeCap, node
     self.inner_mut().set_node_capacity_vote(cap, node_capacity);
 }
 
-// === Update Node Parameters ===
+// === Get/ Update Node Parameters ===
+
+/// Get `NodeMetadata` for the given node.
+public fun node_metadata(self: &mut Staking, node_id: ID): NodeMetadata {
```
karlwuest: I don't think this needs to be `mut`
```suggestion
public fun node_metadata(self:  &Staking, node_id: ID): NodeMetadata {
```

contracts/walrus/tests/e2e_tests.move
```diff
@@ -467,6 +467,53 @@ fun test_epoch_change_with_rewards_and_commission() {
...
+            NODE_CAPACITY,
+            ctx,
+        );
+        node.set_storage_node_cap(cap);
+    });
+
+    runner.tx!(node.sui_address(), |staking, _, _| {
+        let mut metadata = staking.node_metadata(node.node_id());
+        metadata.set_description(b"Tusk Crew".to_string());
+        metadata.set_project_url(b"https://crew.walrus.sites/".to_string());
```
karlwuest: Nit:
```suggestion
        metadata.set_project_url(b"https://crew.walrus.site/".to_string());
```
(also, I just grabbed the testnet suins name for this 😄)

crates/walrus-service/src/client/responses.rs
```diff
@@ -280,6 +280,8 @@ impl StorageNodeInfo {
             next_epoch_public_key,
             network_public_key,
             shard_ids,
+            #[cfg(feature = "walrus-mainnet")]
+                metadata: _,
```
karlwuest: I think you can just do the following, then we don't need the feature flag:
```suggestion
            ..
```

PR #1321, opened 2024-12-17
feat: Implement contract for subsidies in Move

georgedigkas:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/subsidies/sources/subsidies.move
+++ b/contracts/subsidies/sources/subsidies.move
@@ -204,6 +204,10 @@ public fun reserve_space(
 
 // === Accessors ===
 
+public fun admin_cap_subsidies_id(admin_cap: &AdminCap): ID {
+    admin_cap.subsidies_id
+}
+
 /// Returns the current value of the subsidy pool.
 public fun subsidy_pool_value(self: &Subsidies): u64 {
     self.subsidy_pool.value()
--- a/contracts/subsidies/tests/subsidies_tests.move
+++ b/contracts/subsidies/tests/subsidies_tests.move
@@ -4,8 +4,8 @@
 #[test_only]
 module subsidies::subsidies_tests;
 
-use subsidies::subsidies;
-use sui::coin::Coin;
+use subsidies::subsidies::{Self, Subsidies};
+use sui::{coin::Coin, test_scenario as test};
 use wal::wal::WAL;
 use walrus::{
     blob::{Self, Blob},
@@ -23,6 +23,57 @@ const SIZE: u64 = 5_000_000;
 const EPOCH: u32 = 0;
 
 const N_COINS: u64 = 1_000_000_000;
+
+#[test]
+fun test_new() {
+    let user = @0xa11ce;
+    let mut test = test::begin(user);
+
+    let admin_cap = subsidies::new(test.ctx());
+
+    test.next_tx(user);
+    let subsidies = test.take_shared<Subsidies>();
+
+    assert!(subsidies.buyer_subsidy_rate() == 0);
+    assert!(subsidies.system_subsidy_rate() == 0);
+    assert!(subsidies.subsidy_pool_value() == 0);
+    assert!(admin_cap.admin_cap_subsidies_id() == object::id(&subsidies));
+
+    admin_cap.destroy_admin_cap();
+    subsidies.destroy_subsidies();
+
+    test.end();
+}
+
+#[test]
+fun test_new_with_initial_rates_and_funds_public_fn() {
+    let user = @0xa11ce;
+    let mut test = test::begin(user);
+
+    let initial_buyer_subsidy_rate: u16 = 5_00; // 5%
+    let initial_storage_node_subsidy_rate: u16 = 10_00; // 10%
+    let initial_funds_value = 1_000_000;
+
+    let admin_cap = subsidies::new_with_initial_rates_and_funds(
+        initial_buyer_subsidy_rate,
+        initial_storage_node_subsidy_rate,
+        mint(initial_funds_value, test.ctx()),
+        test.ctx(),
+    );
+
+    test.next_tx(user);
+    let subsidies = test.take_shared<Subsidies>();
+
+    assert!(subsidies.buyer_subsidy_rate() == initial_buyer_subsidy_rate);
+    assert!(subsidies.system_subsidy_rate() == initial_storage_node_subsidy_rate);
+    assert!(subsidies.subsidy_pool_value() == initial_funds_value);
+    assert!(admin_cap.admin_cap_subsidies_id() == object::id(&subsidies));
+
+    admin_cap.destroy_admin_cap();
+    subsidies.destroy_subsidies();
+
+    test.end();
+}
 
 #[test]
 fun test_new_subsidy_object(): System {


contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,141 @@
...
+/// This module allows adding funds to a subsidy pool, setting the rate of subsidies, and reserving storage space
+// while providing a discounted cost based on the current subsidy rate.
+module walrus::subsidies;
+
+use sui::{balance::{Self, Balance}, coin::Coin};
+use wal::wal::WAL;
+use walrus::{blob::Blob, storage_resource::Storage, system::System};
+
+/// Subsidy rate is in basis points (1/100 of a percent).
+//TODO: Maybe it has to be u64 to avoid castings?
```
karlwuest: I think having a `u16` for this is best and all casts go from u16 to u64, so they will always succeed.

contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,141 @@
...
+
+/// Creates a new `Subsidies` object and an `AdminCap`.
+public fun create(ctx: &mut TxContext): (Subsidies, AdminCap) {
+    let subsidies = Subsidies {
+        id: object::new(ctx),
+        subsidy_rate: 0,
+        subsidy_pool: balance::zero(),
+    };
+    let admin_cap = AdminCap { id: object::new(ctx) };
+    (subsidies, admin_cap)
```
karlwuest: The subsidies object should be a shared object, so instead returning it here, it should be shared.

contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,141 @@
...
+}
+
+/// Add additional funds to the subsidy pool.
+public fun add_funds(self: &mut Subsidies, funds: Coin<WAL>) {
+    self.subsidy_pool.join(funds.into_balance());
+}
+
+/// Set the subsidy rate.
+public fun set_subsidy_rate(self: &mut Subsidies, _: &AdminCap, new_rate: u16) {
```
karlwuest: Currently, anyone can create a new `Subsidies` object, which means just checking for the presence of an `AdminCap` is not sufficient. The cap in this case should store the ID of the `Subsidies` object and we need to check if they match. Alternatively, we could just have one single `Subsidies` object that is created in the `init` function.

contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,141 @@
...
+    assert!(new_rate <= MAX_SUBSIDY_RATE, EInvalidSubsidyRate);
+    self.subsidy_rate = new_rate;
+}
+
+/// Extends a blob's lifetime and applies the subsidy.
+// TODO: The signature, of  public fun extend_blob(self: &mut System, blob: &mut Blob, epochs_ahead: u32, payment: &mut Coin<WAL>)
```
karlwuest: The signature for both functions (`extend_blob` and `reserve_space`) should be the same as in `walrus`.

contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,141 @@
...
+    let cost = initial_payment_value - payment.value();
+    let subsidy_amount = cost * (self.subsidy_rate as u64) / (MAX_SUBSIDY_RATE as u64);
+
+    // Ensure sufficient funds in subsidy pool.
+    assert!(self.subsidy_pool.value() >= subsidy_amount as u64, EInsufficientSubsidyFunds);
+
+    let subsidy_balance = self.subsidy_pool.split(subsidy_amount);
+    payment.join(subsidy_balance.into_coin(ctx));
+
+    system.add_subsidy(payment, epochs_ahead);
```
karlwuest: (also applies to `extend_blob`): When we subsidize a blob, we want to subsidize "both sides", so the subsidy amount should be added once to the contract and then also once refunded to the buyer.

contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,141 @@
...
+    ctx: &mut TxContext,
+): Storage {
+    let initial_payment_value = payment.value();
+    let storage = system.reserve_space(storage_amount, epochs_ahead, &mut payment, ctx);
+
+    let cost = initial_payment_value - payment.value();
+    let subsidy_amount = cost * (self.subsidy_rate as u64) / (MAX_SUBSIDY_RATE as u64);
+
+    // Ensure sufficient funds in subsidy pool.
+    assert!(self.subsidy_pool.value() >= subsidy_amount as u64, EInsufficientSubsidyFunds);
```
karlwuest: We don't want the operation to fail if there is no money left, in that case we just don't provide a subsidy, but the transaction should still succeed.

contracts/subsidies/sources/subsidies.move
georgedigkas: Thanks, I keep it as u16

contracts/subsidies/sources/subsidies.move
georgedigkas: Changed it to shared object

contracts/subsidies/sources/subsidies.move
georgedigkas: Changed to this one:
```
public struct AdminCap has key {
    id: UID,
    subsidies_id: ID,
}
public struct Subsidies has key, store {
    id: UID,
    buyer_subsidy_rate: u16,
    storage_node_subsidy_rate: u16,
    subsidy_pool: Balance<WAL>,
}
```
But personally I prefer the following:
```
public struct AdminCap has key {
    id: UID,
}
public struct Subsidies has key, store {
    id: UID,
    admin: ID,
    buyer_subsidy_rate: u16,
    storage_node_subsidy_rate: u16,
    subsidy_pool: Balance<WAL>,
}
```
What do you think?

contracts/subsidies/sources/subsidies.move
georgedigkas: Thanks for the feedback, Karl. I understand that we want the `extend_blob` and `reserve_space` signatures to match the `walrus::system::extend_blob signature`. The primary differences I am facing are:
```
fun extend_blob(
    self: &mut Subsidies,
    system: &mut System,
    blob: &mut Blob,
    epochs_ahead: u32,
    mut payment: Coin<WAL>,
    ctx: &mut TxContext,
)
```
signature of `walrus::system`:
```
public fun extend_blob(
    self: &mut System,
    blob: &mut Blob,
    epochs_ahead: u32,
    payment: &mut Coin<WAL>,
) {
```
`system: &mut System` and `self: &mut Subsidies` are required. Is there a recommended way for accessing `System` context inside the `extend_blob` function within the `Subsidies` module instead of passing it as a parameter?

contracts/subsidies/sources/subsidies.move
karlwuest: Between these two, I prefer the former, because if a client owns multiple `AdminCap`s, it is easier to match them to the correct `Subsidies` object.

contracts/subsidies/sources/subsidies.move
karlwuest: Sorry for being a imprecise there. I think we cannot make the signatures completely the same, because we do need the Subsidies object. Passing a TxContext is also fine. However, the other arguments should be the same, i.e. we should not pass an owned coin, but just pass a mutable reference.

georgedigkas: What about the `system: &mut System`?
I think it we should also pass it

karlwuest: Yes, we do need that

contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,247 @@
...
+/// Extends a blob's lifetime and applies the buyer and storage node subsidies.
+public fun extend_blob(
+    self: &mut Subsidies,
+    system: &mut System,
+    blob: &mut Blob,
+    epochs_ahead: u32,
+    payment: &mut Coin<WAL>,
+    ctx: &mut TxContext,
+) {
+    let payment_value = payment.value();
```
karlwuest: Nit:
```suggestion
    let initial_payment_value = payment.value();
```

contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,247 @@
...
+    subsidies_id: ID,
+}
+
+/// Subsidy rates are expressed in basis points (1/100 of a percent).
+/// A subsidy rate of 100 basis points means a 1% subsidy.
+/// The maximum subsidy rate is 10,000 basis points (100%).
+public struct Subsidies has key, store {
+    id: UID,
+    buyer_subsidy_rate: u16,
+    storage_node_subsidy_rate: u16,
```
karlwuest: Nit:
Let's call the subsidies for the system `system_subsidy` instead of `storage_node_subsidy`.

contracts/subsidies/sources/subsidies.move
```diff
@@ -0,0 +1,247 @@
...
+    };
+
+    // Correct cost accounting after subsidy
+    let cost_after_subsidy = cost - actual_subsidy;
+    let payment_value = payment.value();
+
+    // The amount to pay *after* subsidy
+    let amount_to_system = payment_value - (payment_value - cost_after_subsidy);
+
+    add_subsidy_to_system(payment, amount_to_system, system, epochs_ahead, ctx);
```
karlwuest: Important:
I don't think this does what we want. The `payment` coin should only be used here to add the buyer subsidy as a refund. Then a separate subsidy should be paid to the system. Something like the following:

```suggestion
    // Return early if the subsidy pool is empty.
    if (self.subsidy_pool.value() == 0) {
        return
    };
    
    let buyer_subsidy = cost * (self.buyer_subsidy_rate as u64) / (MAX_SUBSIDY_RATE as u64);
    let system_subsidy =
        cost * (self.storage_node_subsidy_rate as u64) / (MAX_SUBSIDY_RATE as u64);
    let total_subsidy = buyer_subsidy + system_subsidy;
    
    // Apply subsidy up to the available amount in the pool.
    let (buyer_subsidy, system_subsidy) = if (self.subsidy_pool.value() >= total_subsidy) {
        (buyer_subsidy, system_subsidy)
    } else {
        // If we don't have enough in the pool to pay the full subsidies, split the remainder proportionally
        // between the buyer and system subsidies.
        let pool_value = self.subsidy_pool.value();
        let total_subsidy_rate = self.buyer_subsidy_rate + self.storage_node_subsidy_rate;
        let buyer_subsidy = pool_value * self.buyer_subsidy_rate / total_subsidy_rate;
        let system_subsidy = pool_value * self.storage_node_subsidy_rate / total_subsidy_rate;
        (buyer_subsidy, system_subsidy)
    };

    let buyer_subsidy_coin = self.subsidy_pool.split(buyer_subsidy).into_coin(ctx);
    payment.join(buyer_subsidy_coin);
    
    let system_subsidy_coin = self.subsidy_pool.split(system_subsidy).into_coin(ctx);
    system.add_subsidy(system_subsidy_coin, epochs_ahead);
```

contracts/subsidies/tests/subsidies_tests.move
```diff
@@ -0,0 +1,437 @@
...
+//     let ctx = &mut tx_context::dummy();
+//     let system = system::new_for_testing();
+//     let (subsidies, admin_cap) = subsidies::new_with_initial_rates_and_funds_for_testing(
+//         initial_buyer_subsidy_rate,
+//         initial_storage_node_subsidy_rate,
+//         mint(initial_funds_value, ctx),
+//         ctx,
+//     );
+//     (system, subsidies, admin_cap)
+// }
```
karlwuest: Nit:
```suggestion
```

contracts/subsidies/tests/subsidies_tests.move
```diff
@@ -0,0 +1,437 @@
...
+    (system, subsidies, admin_cap)
+}
+
+// fun create_and_certify_blob(system: &mut System, storage: Storage): Blob {
+//     let mut blob = register_default_blob(system, storage, false);
+//     let certify_message = messages::certified_blob_message_for_testing(EPOCH, blob.blob_id());
+//     blob.certify_with_certified_msg(system.epoch(), certify_message);
+//     assert!(blob.certified_epoch().is_some());
+//     blob
+// }
```
karlwuest: Nit:
```suggestion
```

contracts/walrus/sources/system/blob.move
```diff
@@ -175,7 +175,7 @@ public(package) fun new(
 
 /// Certifies that a blob will be available in the storage system until the end epoch of the
 /// storage associated with it, given a [`CertifiedBlobMessage`].
-public(package) fun certify_with_certified_msg(
+public fun certify_with_certified_msg(
```
karlwuest: Important: This should not be public, but you can create a `test_only` function to certify a blob.

contracts/subsidies/tests/subsidies_tests.move
```diff
@@ -0,0 +1,437 @@
...
+    subsidies.add_funds(mint(initial_funds_value, ctx));
+    subsidies.set_buyer_subsidy_rate(&admin_cap, 10_00); // 10%
+    subsidies.set_storage_node_subsidy_rate(&admin_cap, 10_00); // 10%
+
+    let mut payment = mint(1000, ctx);
+
+    let storage = subsidies.reserve_space(&mut system, SIZE, 3, &mut payment, ctx);
+
+    // assert!(payment.value() == ?);
+    // assert!(subsidies.subsidy_pool_value() == 86);
```
karlwuest: I think we should have these checks in all tests.

contracts/walrus/sources/system/blob.move
```diff
@@ -312,3 +312,27 @@ public fun insert_or_update_metadata_pair(self: &mut Blob, key: String, value: S
...
+    // Check that the message is from the current epoch
+    assert!(message.certified_epoch() == current_epoch, EWrongEpoch);
+
+    // Check that the storage in the blob is still valid
+    assert!(message.certified_epoch() < blob.storage.end_epoch(), EResourceBounds);
+
+    // Mark the blob as certified
+    blob.certified_epoch.fill(message.certified_epoch());
+
+    blob.emit_certified(false);
```
karlwuest:
```suggestion
    certify_with_certified_msg(blob, current_epoch, message)
```

PR #1322, opened 2024-12-17
fix: wal_balance_at_epoch should take pre_active_withdrawals into account

benr-ml:
## Description

wal_balance_at_epoch should take pre_active_withdrawals into account

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/staking_pool.move
+++ b/contracts/walrus/sources/staking/staking_pool.move
@@ -551,7 +551,9 @@ public(package) fun wal_balance_at_epoch(pool: &StakingPool, epoch: u32): u64 {
     pre_active_withdrawals.keys().do_ref!(|old_epoch| if (*old_epoch <= epoch) {
         let wal_value = pre_active_withdrawals.get(old_epoch);
         // recall that pre_active_withdrawals contains stakes that were
-        // active for exactly 1 epoch.
+        // active for exactly 1 epoch. since the node might have been
+        // inactive, this list may contain more than one value
+        // (although exchange_rate_at_epoch will return the same value).
         let activation_epoch = *old_epoch - 1;
         let token_value_for_epoch = pool
             .exchange_rate_at_epoch(activation_epoch)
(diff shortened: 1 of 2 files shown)


contracts/walrus/tests/staking/staking_pool/staking_pool_tests.move
```diff
@@ -414,6 +414,68 @@ fun wal_balance_at_epoch() {
     pool.destroy_empty()
 }
 
+#[test]
+// Check that wal_balance_at_epoch correctly updates after a pre-active stake
+// withdrawal.
+fun wal_balance_after_pre_active_withdrawal() {
```
karlwuest, line 420: Can you add an additional test where we have two epoch changes after requesting a withdrawal without actually withdrawing before the second epoch change (edit: and, importantly, where rewards are paid in the epoch changes)?

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -533,14 +533,27 @@ public(package) fun exchange_rate_at_epoch(pool: &StakingPool, mut epoch: u32):
 /// the given epoch, due to the complexity of the pending stake and withdrawal
 /// requests, and lack of immediate updates.
 public(package) fun wal_balance_at_epoch(pool: &StakingPool, epoch: u32): u64 {
-    let mut expected = pool.wal_balance;
     let exchange_rate = pool_exchange_rate::new(pool.wal_balance, pool.pool_token_balance);
+
+    let mut pre_active_token_withdraw = 0;
+    let pre_active_withdrawals = pool.pre_active_withdrawals.unwrap();
+    pre_active_withdrawals.keys().do_ref!(|old_epoch| if (*old_epoch <= epoch) {
```
karlwuest, line 551: Remark: I think that `pre_active_withdrawals` can only be longer than 1 for a node that is/was inactive, in which case the exchange rates for all the values should be the same and thus the conversion later to WAL gives the correct value. Can you add a comment explaining this?
(Also see the comment below on adding a test that should catch if it is incorrect)

contracts/walrus/tests/staking/staking_pool/staking_pool_tests.move
karlwuest: I tried to convince myself that the calculations are correct and created a test case in #1325 . It seems that there are additional bugs with the rewards for the pre_active_withdrawals. Can you have a look at the test and investigate?

PR #1324, opened 2024-12-18
fix(publisher): improve response formats

mlegner:
## Description

Currently the publisher's responses contain fields that are not camelCased. Also, the response for an already existing blob is overly complicated.

Closes WAL-484.

## Test plan

Run a publisher and look at its OpenAPI specification.

---

## Release notes

- [x] Publisher: Minor modifications to the response format when storing blobs.


crates/walrus-service/src/client/responses.rs
```diff
@@ -85,6 +86,7 @@ pub enum BlobStoreResult {
         ///
         /// The object ID of the registered blob is used in place of the event ID when the blob is
         /// deletable, already certified, and owned by the client.
+        #[serde(flatten)]
```
giac-mysten, line 89: Question: Is this the fix for the "overly complicated response", i guess?

PR #1328, opened 2024-12-18
feat: support split again in Withdrawing state

benr-ml:
## Description

 support split again in Withdrawing state

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/staked_wal.move
+++ b/contracts/walrus/sources/staking/staked_wal.move
@@ -169,27 +169,12 @@ public fun split(sw: &mut StakedWal, amount: u64, ctx: &mut TxContext): StakedWa
     assert!(sw.principal.value() > amount, EInvalidAmount);
     assert!(amount > 0, EInvalidAmount);
 
-    match (sw.state) {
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/staking/staked_wal.move
```diff
@@ -181,6 +180,15 @@ public fun split(sw: &mut StakedWal, amount: u64, ctx: &mut TxContext): StakedWa
                 activation_epoch: sw.activation_epoch,
             }
         },
+        StakedWalState::Withdrawing { withdraw_epoch } => {
```
karlwuest: Suggestion: I think we can simplify this by just replacing the whole `match` statement with the following, because the two cases should now be the same:

```
    StakedWal {
        id: object::new(ctx),
        state: sw.state, // state is preserved
        node_id: sw.node_id,
        principal: sw.principal.split(amount),
        activation_epoch: sw.activation_epoch,
    }
```

benr-ml: oops

PR #1331, opened 2024-12-19
feat: implement quorum-based upgrade policy

karlwuest:
## Description

Implements custom quorum-based upgrades.
closes WAL-357
depends on #1301 

## Test plan

- added move tests for the upgrades
- existing tests
- will work on adding integration tests for upgrades with WAL-481

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


contracts/walrus/sources/upgrade.move
```diff
@@ -1,26 +1,77 @@
...
 }
 
+/// Create a new upgrade manager.
+///
+/// This is called from the `init::initialize_walrus` function and will
+/// create a unique `UpgradeManager` and `EmergencyUpgradeCap` object.
 public(package) fun new(cap: UpgradeCap, ctx: &mut TxContext): EmergencyUpgradeCap {
-    let upgrade_manager = UpgradeManager { id: object::new(ctx), cap };
+    let table = table::new(ctx);
+    let upgrade_manager = UpgradeManager { id: object::new(ctx), cap, upgrade_proposals: table };
```
manolisliolios: nit: can be inlined
```suggestion
    let upgrade_manager = UpgradeManager { id: object::new(ctx), cap, upgrade_proposals: table::new(ctx) };
```

contracts/walrus/sources/upgrade.move
```diff
@@ -29,7 +80,67 @@ public(package) fun new(cap: UpgradeCap, ctx: &mut TxContext): EmergencyUpgradeC
...
+// === Upgrade Manager Public API ===
+
+/// Vote for an upgrade given the digest of the package to upgrade to.
+///
+/// This will create a new upgrade proposal if none exists for the given digest.
+public fun vote_for_upgrade(
+    self: &mut UpgradeManager,
+    staking: &Staking,
+    auth: Authenticated,
+    node_id: ID,
```
manolisliolios, line 95: Wonder if it'd simplify things if we saved `node_id` in `Authenticated`, and the validity of a node_id is guaranteed on the auth layer, instead of needing to check for validity on each public API.

PS: I also think that'd need a naming reversal, as having `node_id` access is an authorization thing, not an authentication.

contracts/walrus/sources/upgrade.move
```diff
@@ -29,7 +80,67 @@ public(package) fun new(cap: UpgradeCap, ctx: &mut TxContext): EmergencyUpgradeC
...
+        if (proposal.epoch != epoch || proposal.version != self.cap.version() + 1) {
+            *proposal = fresh_proposal(epoch, digest, self.cap.version() + 1);
+        };
+        proposal
+    } else {
+        let proposal = fresh_proposal(epoch, digest, self.cap.version() + 1);
+        self.upgrade_proposals.add(digest, proposal);
+        self.upgrade_proposals.borrow_mut(digest)
+    };
+    proposal.add_vote(node_id, weight);
```
manolisliolios, line 117: Verifying my assumption: Since the vote works on epoch boundaries, that's why we're guaranteed to have the same "weight" for a node for the entire vote?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -275,6 +276,36 @@ fun take_threshold_value(vote_queue: &mut PriorityQueue<u64>, threshold_weight:
...
+): bool {
+    auth.matches(self.pools[node_id].governance_authorized())
+}
+
+/// Returns the current node weight for the given node id.
+public(package) fun get_current_node_weight(self: &StakingInnerV1, node_id: ID): u16 {
+    // Check if the node is in the committee.
+    assert!(self.committee.inner().contains(&node_id), ENotInCommittee);
+    let weight = self.committee.shards(&node_id).length();
+    assert!(weight <= std::u16::max_value!() as u64, EInvalidNodeWeight);
```
manolisliolios, line 309: Is this possible?

contracts/walrus/sources/upgrade.move
```diff
@@ -29,7 +80,67 @@ public(package) fun new(cap: UpgradeCap, ctx: &mut TxContext): EmergencyUpgradeC
...
+    staking: &Staking,
+    digest: vector<u8>,
+): UpgradeTicket {
+    let digest = package_digest(digest);
+
+    assert!(self.upgrade_proposals.contains(digest), ENoProposalForDigest);
+    let proposal = self.upgrade_proposals.remove(digest);
+
+    // Check that the proposal is for the current epoch and that the quorum is reached.
+    assert!(proposal.epoch == staking.epoch(), EWrongEpoch);
```
manolisliolios, line 136: Same-epoch execution is guaranteed here, I'd guess this is planned for, right?

contracts/walrus/sources/staking/staking_inner.move
karlwuest: No, this is just a sanity check.

contracts/walrus/sources/upgrade.move
karlwuest: >Wonder if it'd simplify things if we saved node_id in Authenticated, and the validity of a node_id is guaranteed on the auth layer, instead of needing to check for validity on each public API.

The `Authenticated` only authenticates the sender address / an owned object. To get this, there are no authorization checks. The authorization happens in `check_governance_authorization` where we look up if the authenticated entity is authorized for the given `node_id`.

>I also think that'd need a naming reversal, as having node_id access is an authorization thing, not an authentication.

See above 😉

contracts/walrus/sources/upgrade.move
karlwuest: Yes

contracts/walrus/sources/upgrade.move
karlwuest: Yes, if an upgrade is not committed in the same epoch that it was voted for, the vote needs to be restarted because the committee may have changed.

contracts/walrus/sources/staking/auth.move
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+module walrus::auth;
+
+/// Authentication for either a sender or an object.
+/// Unlike the `Authorized` type, it cannot be stored and must be used or ignored in the same
+/// transaction.
+public enum Authenticated has drop {
```
manolisliolios, line 9: Should we just hold the `address` here to simplify the flow? We cannot have an overlap of an address/objectID anyways

contracts/walrus/sources/upgrade.move
manolisliolios: I understand this! My recommendation was to abstract the authorization part too in order to blindly "trust" the witness, but it's fine to not do it if we do not plan on having many public APIs needing that.

contracts/walrus/sources/staking/auth.move
karlwuest: This is essentially just a refactor of the code that was previously in `commission.move`. When @damirka implemented that, we discussed the option of just storing an address, but decided against it. The main reason was that this is more readable and makes the distinction between the two cases clear. I also don't think that it would simplify the flow, because if would only change the inner workings of the functions in the `auth` module.

That being said, I think we may anyway want to change the mechanism to just use another capability object (as a follow up, not as a part of this PR). For context, see [this comment on slack](https://mysten-labs.slack.com/archives/C080QP8N1A8/p1735553261670409), it would be great to get your thoughts on this as well!

contracts/walrus/sources/upgrade.move
karlwuest: Ah, I see. So the idea would be to replace the `Authenticated` with a `AuthenticatedAndAuthorized` that also contains the node id and for which we first do the authorization check as well?

I think generally that would be a nice direction, but then we also need to store in that object for which operation we checked the authorization. The advantage of the current approach is that the authentication can happen completely separate from the authorization.

contracts/walrus/sources/staking/auth.move
damirka: +1 to what @karlwuest is saying here. Explicitness is an important property of this design choice.

contracts/walrus/sources/staking/auth.move
damirka, line 1: Love this one! Maybe something worth upstreaming somewhere.

contracts/walrus/sources/upgrade.move
```diff
@@ -1,26 +1,77 @@
...
+const ENotAuthorized: u64 = 1;
+const EDuplicateVote: u64 = 2;
+const EInvalidPackageDigest: u64 = 3;
+const ENoProposalForDigest: u64 = 4;
+const EWrongEpoch: u64 = 5;
+const ENotEnoughVotes: u64 = 6;
+const EWrongVersion: u64 = 7;
+
+/// Newtype for package digests, ensures that the digest is always 32 bytes long.
+public struct PackageDigest(vector<u8>) has copy, drop, store;
```
damirka, line 31: Should we then use the type that is constrained to 32 bytes? An `address` could be a good match due to it keeping the order of bytes.

contracts/walrus/sources/upgrade.move
```diff
@@ -63,11 +174,66 @@ public fun commit_upgrade(
...
+    proposal.voters.insert(node_id);
+    proposal.voting_weight = proposal.voting_weight + weight;
+}
+
+// === Package Digest ===
+
+/// Creates a new package digest given a byte vector.
+///
+/// Aborts if the digest is not 32 bytes long.
+fun package_digest(digest: vector<u8>): PackageDigest {
```
damirka: nit: consider making it a local macro to inline and avoid function call.

contracts/walrus/tests/upgrade_tests.move
```diff
@@ -0,0 +1,208 @@
...
+        });
+    });
+
+    // === advance clock and change epoch ===
+
+    runner.clock().increment_for_testing(e2e_runner::default_epoch_duration());
+    runner.tx!(nodes[0].sui_address(), |staking, system, _| {
+        staking.voting_end(runner.clock());
+        staking.initiate_epoch_change(system, runner.clock());
+        assert!(system.epoch() == 2);
```
damirka: nit: assert_eq would result in a meaningful error, macro error tracing is not perfect

contracts/walrus/tests/upgrade_tests.move
```diff
@@ -0,0 +1,208 @@
...
+
+    // === try to upgrade to digest 2 ===
+
+    runner.tx_with_upgrade_manager!(nodes[0].sui_address(), |staking, _, upgrade_manager, _| {
+        // Test should fail here.
+        let _upgrade_ticket = upgrade_manager.authorize_upgrade(staking, digest_2);
+    });
+
+    // Unreachable
+    abort EUnreachable
```
damirka: You can just leave `abort`

contracts/walrus/sources/upgrade.move
manolisliolios: I mean you could wrap it as an `AuthorizationProof(Authenticated, node_id)` (or something similar to have) `type-safe` access control [or not even sure you care about having Authenticated in there since you'd have a proof(node_id)]. Fwiw, I am not sure if its worth the trouble of changing things here.

contracts/walrus/sources/upgrade.move
karlwuest: All of the functions in `sui::package` just use vector, which is why having a vector internally is convenient. I only added the newtype to enforce the check.

contracts/walrus/sources/staking/auth.move
manolisliolios: > ... and makes the distinction between the two cases clear.

> Explicitness is an important property of this design choice.

Why is the distinction between the two authentication methods important here? (why do we care about being explicit on "how" we authorized?)


> The main reason was that this is more readable

Why do we consider it more readable? I actually find quite more bulky (more APIs, more complex checks e.g. on `matches` function).

**To be clear, I don't have a strong opinion on going either way.**

karlwuest: >Why is the distinction between the two authentication methods important here?

I think the explicit distinction and typing here can help to avoid errors. If we just have a single API to create an `Authorized`, we would always need to pass an address, even if we want to create it for an object. Having clear types and separate functions makes the intent clearer and makes it less likely to pass a wrong value (granted, the typing only applies if called from another contracts, since if we create a ptb, `ID` and `address` are effectively the same type).

>more APIs

It's only one more function in the API, namely when setting the `Authorized` value. 

>Why do we consider it more readable?

This is, of course, subjective, but I find it more readable because it doesn't conflate user addresses with object IDs and it makes it immediately obvious which of the two we're using. 
Yes, `matches` is a bit more complex than if we just used addresses, but that is a small internal function. The external interfaces are much clearer this way in my opinion (the above w.r.t. typing and naming applies here as well).

PR #1333, opened 2024-12-19
fix: include object ID in certification message for deletable blobs

karlwuest:
## Description

closes WAL-301

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/docs/msg_formats.txt
+++ b/contracts/walrus/docs/msg_formats.txt
@@ -43,7 +43,7 @@ BLOB_CERT message
 -----------------
 
 The body is the blob_id : u256 (32 bytes) of a blob followed by the bcs encoding of the blob
-persistence type defined below, i.e. 1 byte that is 0 for deletable blobs and 1 for permanent
+persistence type defined below, i.e. 1 byte that is 0 for permanent blobs and 1 for deletable
 blobs, followed by the object id of the deletable blob if it is a deletable blob.
 
 ```


contracts/walrus/docs/msg_formats.txt
mlegner, line 1: Nit: Having a `0` for the "default" permanent type would be more intuitive. But I also don't see a downside of the current version.

contracts/walrus/sources/system/messages.move
```diff
@@ -291,6 +306,36 @@ public(package) fun blob_id(self: &DenyListBlobDeleted): u256 {
...
+public(package) fun object_id(self: &BlobPersistenceType): ID {
+    match (self) {
+        BlobPersistenceType::Deletable { object_id } => *object_id,
+        BlobPersistenceType::Permanent => abort ENotDeletable,
+    }
+}
+
+// === BCS deserialization ===
+
+public(package) fun peel_blob_persistence_type(bcs: &mut BCS): BlobPersistenceType {
```
mlegner, line 327: Question: There isn't any generic BCS deserialization in Move, correct?

contracts/walrus/sources/system/messages.move
```diff
@@ -305,18 +350,42 @@ public fun certified_message_for_testing(
 }
 
 #[test_only]
-public fun certified_blob_message_for_testing(_epoch: u32, blob_id: u256): CertifiedBlobMessage {
```
mlegner, line 308: Question: Why did we have the epoch there before?

contracts/walrus/sources/system/messages.move
```diff
@@ -305,18 +350,42 @@ public fun certified_message_for_testing(
...
+public fun certified_deletable_message_bytes(epoch: u32, blob_id: u256, object_id: ID): vector<u8> {
     let mut message = vector<u8>[];
     message.push_back(BLOB_CERT_MSG_TYPE);
     message.push_back(INTENT_VERSION);
     message.push_back(APP_ID);
     message.append(bcs::to_bytes(&epoch));
     message.append(bcs::to_bytes(&blob_id));
+    message.append(bcs::to_bytes(&BlobPersistenceType::Deletable { object_id }));
     message
 }
```
mlegner: Nit: I feel these could be combined into a function that just takes an additional `BlobPersistenceType` argument.

contracts/walrus/tests/system/blob_tests.move
```diff
@@ -98,6 +101,61 @@ fun blob_certify_single_function() {
     system.destroy_for_testing();
 }
 
+#[test, expected_failure(abort_code = blob::EInvalidBlobPersistenceType)]
+fun blob_certify_deletable_msg_for_permanent_blob() {
```
mlegner, line 136: Suggestion: In addition to the two added negative tests, could you also add a positive test case for deletable blobs?

crates/walrus-core/src/messages/storage_confirmation.rs
```diff
@@ -16,28 +16,54 @@ pub enum StorageConfirmation {
...
+    /// The blob is deletable and was with the given object ID.
+    Deletable {
+        /// The object ID of the associated Sui object.
+        object_id: SuiObjectId,
+    },
+    /// The blob is permanent.
+    Permanent,
+}
+
+/// The message body for a `StorageConfirmationMsg`,
```
mlegner: Nits:
```suggestion
/// The message body for a [`Confirmation`].
```

crates/walrus-core/src/messages/storage_confirmation.rs
```diff
@@ -16,28 +16,54 @@ pub enum StorageConfirmation {
...
+    /// The blob id of the blob that is being confirmed.
+    pub blob_id: BlobId,
+    /// Whether the blob is permanent or deletable.
+    /// For deletable blobs, the object id of the blob is included.
+    pub blob_type: BlobPersistenceType,
+}
+
 /// A Confirmation that a storage node has stored all respective slivers
-/// of a blob in their shards.
+/// of a permanent blob in their shards.
```
mlegner: Suggestion: Revert this.

crates/walrus-core/src/messages/storage_confirmation.rs
```diff
@@ -16,28 +16,54 @@ pub enum StorageConfirmation {
     Signed(SignedStorageConfirmation),
 }
 
+/// Indicates the persistence of a blob.
+///
+/// For deletable blobs the object ID of the associated Sui object is included.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
```
mlegner: Suggestion: As this is very small, we can also derive `Copy` for this:
```suggestion
#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
```

crates/walrus-service/src/client.rs
```diff
@@ -1133,17 +1141,9 @@ impl<T> Client<T> {
...
-            bcs::to_bytes(&Confirmation::new(
-                committees.write_committee().epoch,
-                *blob_id,
-            ))
-            .expect("serialization should always succeed"),
-            aggregate,
-        );
+        let cert =
+            ConfirmationCertificate::from_signed_messages_and_indices(signed_messages, signers)
+                .map_err(ClientError::other)?;
```
mlegner, line 1146: Question: This refactoring is independent of the main changes of this PR, correct?

crates/walrus-sdk/src/client.rs
```diff
@@ -96,10 +103,41 @@ impl UrlEndpoints {
...
+        blob_id: &BlobId,
+        object_id: &ObjectID,
+    ) -> (Url, &'static str) {
         (
-            self.blob_resource(blob_id, "confirmation"),
-            STORAGE_CONFIRMATION_URL_TEMPLATE,
+            self.blob_resource(blob_id, &format!("confirmation/deletable/{object_id}")),
+            DELETABLE_BLOB_CONFIRMATION_URL_TEMPLATE,
         )
     }
```
mlegner: Important: AFAIU, this changes the behavior when the `walrus-mainnet` feature is not enabled. This means that clients would call the `confirmation/deletable` endpoint, which may not be provided by storage nodes yet. In fact, this fails when I run the new client against the Walrus Testnet:
```
2025-01-07T15:44:04.764344Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: storing metadata and pairs failed node=32 error=unexpected end of input
...
2025-01-07T15:44:04.765943Z  WARN reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: Failed to get blob certificate blob_object.blob_id=eUyCBl9h0he57BNP2VaWT8SdtCMOo_ZxhyIorC_KKf8 error=could not retrieve enough confirmations to certify the blob: 0 / 667 required;
Error: could not retrieve enough confirmations to certify the blob: 0 / 667 required;

File: contracts/walrus/docs/msg_formats.txt
Comment (karlwuest) [reply contracts/walrus/docs/msg_formats.txt:1]: I went back and forth between the two and decided to use `0` for deletable, because it is not persistent (i.e. persistence is "false"), but now that you mention it, I realize that we otherwise generally use "true" for deletable 🤦‍♂️🤦‍♂️. Will change this.

File: contracts/walrus/sources/system/messages.move (hunk 1 above)
Comment (karlwuest) [reply contracts/walrus/sources/system/messages.move:327]: Correct

File: contracts/walrus/sources/system/messages.move (hunk 2 above)
Comment (karlwuest) [reply contracts/walrus/sources/system/messages.move:308]: I think that was a leftover from the earlier contract version where the epoch was still in the specific message types (instead of them containing only the body) and for some reason it didn't get removed in a previous refactor.

File: contracts/walrus/sources/system/messages.move (hunk 3 above)
Comment (karlwuest) [reply contracts/walrus/sources/system/messages.move]: I can't replace the functions (because all enum variants in move are private to the module), but I can add another function that I call from these functions to keep it a bit tidier.

File: crates/walrus-service/src/client.rs (hunk 8 above)
Comment (karlwuest) [reply crates/walrus-service/src/client.rs:1146]: Yes, mostly. Without this change, the message body would need to be reconstructed, which would have required passing additional arguments.

File: crates/walrus-sdk/src/client.rs (hunk 9 above)
Comment (karlwuest) [reply crates/walrus-sdk/src/client.rs]: Ah, that's a good point. I did not consider newer clients connecting to older storage nodes. I'll fix that.

Comment (karlwuest) [reply crates/walrus-sdk/src/client.rs]: Fixed in e3b3b4fcc57f08bc22a8faa6259328fb06c2f9b0. The endpoint now returns the signed messages with the previous message format and I tested the store with the new client on the current Testnet.

File: contracts/walrus/docs/msg_formats.txt [hunk 10]
```diff
@@ -42,9 +42,19 @@ size array) of the signer.
 BLOB_CERT message
 -----------------
 
-The body is the blob_id : u256 (32 bytes) of a blob. The storage node certified that all
-shards that were assigned to it for the message epoch have been received, checked and
-persisted to store.
+The body is the blob_id : u256 (32 bytes) of a blob followed by the bcs encoding of the blob
+persistence type defined below, i.e. 1 byte that is 0 for deletable blobs and 1 for permanent
+blobs, followed by the object id of the deletable blob if it is a deletable blob.
```
mlegner: Important: This needs to be changed after swapping the `Permanent` and `Deletable` variants:
```suggestion
The body is the blob_id : u256 (32 bytes) of a blob followed by the bcs encoding of the blob
persistence type defined below, i.e. 1 byte that is 0 for permanent blobs and 1 for deletable
blobs, followed by the object id of the deletable blob if it is a deletable blob.
```

PR #1335, opened 2024-12-19
feat(client): add post store action for shared blob

joyqvq:
## Description

Contributes to WAL-263

## Test plan

simtest

also tested with cli

```
target/release/walrus store ~/Desktop/old_10kb_random_file_6.bin ~/Desktop/old_10kb_random_file_5.bin --shared-blob                                                                                                                                                                             ✔  10027  15:37:57
2024-12-19T20:38:56.876134Z  INFO walrus: client version: 1.8.0-7f22ea46bd31-dirty
2024-12-19T20:38:56.876300Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2024-12-19T20:38:56.876650Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/.sui/sui_config/client.yaml'
Warning: The number of epochs for which to store the blob was not specified. Using the default value of 1 epoch. Use `--epochs` to store for a longer period.
2024-12-19T20:38:58.495081Z  INFO walrus_service::client::cli::runner: storing 2 files as blobs on Walrus
 • encoding the blob [00:00:00]                                                                                                                                                                                                                                                                                                                                                            2024-12-19T20:38:59.202091Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=1 primary_sliver_size=662 secondary_sliver_size=329 duration=706.036458ms
 ◉ blob encoded; blob ID: JfIoR9RRMXSSWZg6Urnzh36QBMyhui3PnhXYeg6b-s8 [00:00:00]
 • encoding the blob [00:00:00]                                                                                                                                                                                                                                                                                                                                                            2024-12-19T20:38:59.880712Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded sliver pairs and metadata symbol_size=1 primary_sliver_size=662 secondary_sliver_size=329 duration=678.4085ms
 ◉ blob encoded; blob ID: czKYhytqStnEtCqegOrolY8ghqGbjpX-Cr6bwW29Kl0 [00:00:00]                                                                                                                                                                                                                                                                                                           2024-12-19T20:38:59.880901Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: encoded files to metadata: JfIoR9RRMXSSWZg6Urnzh36QBMyhui3PnhXYeg6b-s8 -> /Users/joy/Desktop/old_10kb_random_file_6.bin, czKYhytqStnEtCqegOrolY8ghqGbjpX-Cr6bwW29Kl0 -> /Users/joy/Desktop/old_10kb_random_file_5.bin
2024-12-19T20:38:59.880921Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: storing 2 sliver pairs with metadata
2024-12-19T20:39:00.387803Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: retrieved 2 blobs statuses duration=506.849417ms
2024-12-19T20:39:08.302515Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: 2 blobs resources obtained duration=7.914612958s
2024-12-19T20:39:08.504307Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=JfIoR9RRMXSSWZg6Urnzh36QBMyhui3PnhXYeg6b-s8
 ◉ sending slivers [00:00:00] [------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------] 0/667 (0s)2024-12-19T20:39:08.518129Z  INFO reserve_and_store_blobs_retry_epoch_with_path:send_blob_data_and_get_certificate: walrus_service::client: starting to send data to storage nodes blob_id=czKYhytqStnEtCqegOrolY8ghqGbjpX-Cr6bwW29Kl0
 ◉ slivers sent [00:00:03] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ slivers sent [00:00:03] [#################################################################################################################################################################################################################################################################################################################################################] 667/667 (0s)
 ◉ additional slivers stored [00:00:02]                                                                                                                                                                                                                                                                                                                                                    2024-12-19T20:39:13.650847Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=czKYhytqStnEtCqegOrolY8ghqGbjpX-Cr6bwW29Kl0 duration=5.132676041s blob_size=10240
 ◉ additional slivers stored [00:00:02]                                                                                                                                                                                                                                                                                                                                                    2024-12-19T20:39:13.704819Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: finished sending blob data and collecting certificate blob_id=JfIoR9RRMXSSWZg6Urnzh36QBMyhui3PnhXYeg6b-s8 duration=5.200490792s blob_size=10240
2024-12-19T20:39:13.704860Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: get 2 blobs certificates duration=5.402173792s
2024-12-19T20:39:15.576884Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_sui::client: shared blob object created object_id=0x49b5949e28ae82d2428cdae6e97c9332221d219aa49300d8ddb00c72053b2abd
2024-12-19T20:39:15.576976Z  INFO reserve_and_store_blobs_retry_epoch_with_path: walrus_service::client: certified 2 blobs on Sui duration=1.872093875s
2024-12-19T20:39:15.579741Z  INFO walrus_service::client::cli::runner: 2 out of 2 blobs stored duration=17.084410125s
Path: /Users/joy/Desktop/old_10kb_random_file_6.bin
Blob ID: JfIoR9RRMXSSWZg6Urnzh36QBMyhui3PnhXYeg6b-s8
Sui object ID: 0xc3a60cbb66a9a20251875d61258aa98364a85de7a43e1eae7c99b33deecf3836
Unencoded size: 10.0 KiB
Encoded size (including replicated metadata): 62.0 MiB
Cost (excluding gas): 0.0001 WAL (storage was purchased, and a new blob object was registered)
Shared blob object ID (if created): 0x49b5949e28ae82d2428cdae6e97c9332221d219aa49300d8ddb00c72053b2abd

Path: /Users/joy/Desktop/old_10kb_random_file_5.bin
Blob ID: czKYhytqStnEtCqegOrolY8ghqGbjpX-Cr6bwW29Kl0
Sui object ID: 0xefa09d15eac114aaaa2c77c66afdc3f9c2287212ef91a39af3f13d1ea3289d76
Unencoded size: 10.0 KiB
Encoded size (including replicated metadata): 62.0 MiB
Cost (excluding gas): 0.0001 WAL (storage was purchased, and a new blob object was registered)
Shared blob object ID (if created): 0x49b5949e28ae82d2428cdae6e97c9332221d219aa49300d8ddb00c72053b2abd
```

example txn: https://suiscan.xyz/testnet/tx/DnsFy5ktSwZb7e6H5gkkfmR2mPx43jiKoHbAFd3R1ysm

deletable blob cannot be shared 
```
target/release/walrus store ~/Desktop/old_10kb_random_file_1.bin --deletable --share                                                                                                                                                                                                                  ✔  10012  16:55:04
2025-01-06T21:56:01.999776Z  INFO walrus: client version: 1.8.0-d046caa53534-dirty
2025-01-06T21:56:01.999825Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2025-01-06T21:56:01.999978Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/.sui/sui_config/client.yaml'
Warning: The number of epochs for which to store the blob was not specified. Using the default value of 1 epoch. Use `--epochs` to store for a longer period.
Error: deletable blobs cannot be shared
```

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Add a PostStoreAction::Share that puts a blob into a SharedBlob.

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -487,7 +487,7 @@ impl SuiContractClient {
         )?;
         ensure!(
             object_ids.len() == blobs_with_certificates.len(),
-            "unexpected number of shared blob objects created: {} expected {}",
+            "unexpected number of shared blob objects created: {} (expected {})",
             object_ids.len(),
             blobs_with_certificates.len()
         );


crates/walrus-sui/src/client.rs
```diff
@@ -421,13 +436,15 @@ impl SuiContractClient {
...
+    ///
+    /// If the post store action is `share`, returns a mapping blob ID -> shared_blob_object_id.
     // NB: This intentionally takes an owned `Blob` object even though it is not required, as the
     // corresponding object on Sui will be changed in the process.
     pub async fn certify_blobs(
         &self,
         blobs_with_certificates: &[(&Blob, ConfirmationCertificate)],
         post_store: PostStoreAction,
-    ) -> SuiClientResult<()> {
+    ) -> SuiClientResult<HashMap<BlobId, ObjectID>> {
```
joyqvq, line 448: since the transaction effect only gives a list of created object ids - this call needs to fetch the full shared object to map the blob id to the shared blob object id.

so the caller can optionally add it to blob store result and return to in the output. 

alternatively, if we do not care about the mapping we can just return a list of object id without mapping to blob.. or do not return this info at all.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -189,6 +189,11 @@ pub enum CliCommands {
         #[clap(long, action)]
         #[serde(default)]
         deletable: bool,
+
+        /// Whether to put the blob into a shared blob object.
+        #[clap(long, action)]
+        #[serde(default)]
+        shared_blob: bool,
```
mlegner: Suggestion: Can we just call this `shared` or `share`?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -115,7 +116,8 @@ impl CliOutput for BlobStoreResultWithPath {
                     Sui object ID: {}\n\
                     Unencoded size: {}\n\
                     Encoded size (including replicated metadata): {}\n\
-                    Cost (excluding gas): {} {}\n",
+                    Cost (excluding gas): {} {}\n\
+                    Shared blob object ID (if created): {}\n",
```
mlegner: Suggestion: Only add this when there is a shared object.

crates/walrus-service/src/client/responses.rs
```diff
@@ -108,6 +108,8 @@ pub enum BlobStoreResult {
         resource_operation: RegisterBlobOp,
         /// The storage cost, excluding gas.
         cost: u64,
+        /// The shared blob object ID if created.
+        shared_blob_object: Option<ObjectID>,
```
mlegner: Suggestion:
```suggestion
        /// The shared blob object ID if created.
        #[serde_as(as = "Option<DisplayFromStr>")]
        #[serde(skip_serializing_if = "Option::is_none")]
        shared_blob_object: Option<ObjectID>,
```

crates/walrus-service/src/client/cli/args.rs
mlegner: Also, more important: It is a bit unclear how this interacts with other flags like `deletable` and `force`:
- What is the behavior if the blob is already stored?
- If `deletable` is `true`, will the shared blob be deletable as well? Not sure if there is a good reason for that.

crates/walrus-sui/src/client.rs
mlegner: I think the mapping makes sense. However, can we add a special case for when there is just a single blob to avoid having to do another RPC call, see suggestion below?

crates/walrus-sui/src/client.rs
```diff
@@ -444,15 +461,42 @@ impl SuiContractClient {
...
+                blobs_with_certificates.len()
+            );
+
+            // Fetch and fetch the SharedBlob object and parse the blob ID.
+            let mut shared_blob_objects = HashMap::new();
+            for object_id in object_ids {
+                let shared_blob: SharedBlob = self.sui_client().get_sui_object(object_id).await?;
+                shared_blob_objects.insert(shared_blob.blob.blob_id, object_id);
+            }
+            Ok(shared_blob_objects)
```
mlegner: Two suggestions here:
1. If there is just one object, we don't need to request anything.
2. If there are multiple objects, we should use `get_sui_objects` (which takes a vector of inputs) to avoid multiple sequential calls.

crates/walrus-sui/src/client.rs
```diff
@@ -444,15 +461,42 @@ impl SuiContractClient {
...
         let res = self.sign_and_send_ptb(&wallet, ptb, None).await?;
-        if res.errors.is_empty() {
-            Ok(())
+
+        if !res.errors.is_empty() {
+            return Err(anyhow!("could not certify blob: {:?}", res.errors).into());
+        }
+
+        // If the blobs are shared, create a mapping blob ID -> shared_blob_object_id.
+        if post_store == PostStoreAction::Share {
```
mlegner: Nit: To avoid the extra indentation, you could do 
```rust
        if post_store != PostStoreAction::Share {
            return Ok(HashMap::new());
        }
```

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -609,6 +609,21 @@ impl SuiReadClient {
...
+        let obj = self
+            .sui_client
+            .get_object_with_options(
+                object_id,
+                SuiObjectDataOptions::new().with_bcs().with_type(),
+            )
+            .await?;
+        Ok(get_sui_object_from_object_response(&obj)?)
+    }
 }
```
mlegner: There already exists a function doing pretty much this in `crates/walrus-sui/src/client/retry_client.rs`: `get_sui_object`. Can we use that one and remove the newly added one?

crates/walrus-sui/src/client.rs
joyqvq: +1 on 1)
re 2) we could indeed use get_sui_objects, but we will lose the ordering of object ids, and cannot map it back into per blob store result: https://github.com/MystenLabs/walrus/blob/7ab39cb11d0418858e2a671119838caf0a5daa41/crates/walrus-service/src/client.rs#L642

i did this bc we put this object id into each BlobStoreResult, but maybe we do not care the mapping much..?

crates/walrus-sui/src/client/read_client.rs
joyqvq: i see its under sui_client().sui_client() not read_client 😓 revised and make get_sui_object public.

crates/walrus-service/src/client/cli/args.rs
joyqvq: >What is the behavior if the blob is already stored?

good catch! currently only newly registered blob will look at [post store action](https://github.com/MystenLabs/walrus/blob/7ab39cb11d0418858e2a671119838caf0a5daa41/crates/walrus-sui/src/client.rs#L215) ie to be shared. i can add this to also check noop on already certified blobs, qq is, do we want this behavior for all of the post store actions (i.e burn and transfer can also be done on certified blobs)? 

a related q, post store action is currently only interpret Keep, should we allow others e.g. burn and transfer to? cc @giac-mysten 

> will the shared blob be deletable as well

i think we should disallow deletable blob to use this option bc otherwise anyone can delete the blob..?

mlegner: > i think we should disallow deletable blob to use this option bc otherwise anyone can delete the blob..?

Yes, I think that makes sense.

crates/walrus-sui/src/client.rs
mlegner: > re 2) we could indeed use get_sui_objects, but we will lose the ordering of object ids, and cannot map it back into per blob store result:

So the objects returned from there are not in the same order as the input? As far as I understand the [docs](https://docs.sui.io/sui-api-ref#sui_multigetobjects), the response should contain the object IDs, so it should be possible to do the matching.

joyqvq: ah i see what you mean, revised!

crates/walrus-sui/src/client.rs
```diff
@@ -445,15 +462,50 @@ impl SuiContractClient {
...
+
+        // If the blobs are shared, create a mapping blob ID -> shared_blob_object_id.
+        let object_ids = get_created_sui_object_ids_by_type(
+            &res,
+            &contracts::shared_blob::SharedBlob
+                .to_move_struct_tag_with_type_map(&self.read_client.type_origin_map(), &[])?,
+        )?;
+        ensure!(
+            object_ids.len() == blobs_with_certificates.len(),
+            "unexpected number of shared blob objects created: {} expected {}",
```
mlegner: Nit:
```suggestion
            "unexpected number of shared blob objects created: {} (expected {})",
```

crates/walrus-sui/src/client.rs
mlegner: Hmm... I still see individual calls to `get_sui_object`... 🤔

crates/walrus-service/src/client/cli/args.rs
joyqvq: discussed offline - we do not share objects if they are already stored (similar to other PostStoreAction). rather a new command will be added to store existing blob (similar to burn)

PR #1336, opened 2024-12-19
feat: add recovery metadata metrics; add sliver type for shard sync metrics

halfprice:
## Description

Says by the title.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/lib.rs
+++ b/crates/walrus-core/src/lib.rs
@@ -114,7 +114,11 @@ impl BlobId {
     ///
     /// The extracted can be used to monitor the progress of tasks that scans over blob IDs.
     pub fn first_two_bytes(&self) -> u16 {
-        u16::from_be_bytes(self.0[0..2].try_into().unwrap())
+        u16::from_be_bytes(
+            self.0[0..2]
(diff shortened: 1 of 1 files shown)


crates/walrus-core/src/lib.rs
```diff
@@ -109,6 +109,14 @@ impl BlobId {
         blob_id
     }
 
+    /// Extracts the first two bytes of the blob ID as a `u16`, with the left most bit being the
+    /// most significant.
+    ///
+    /// The extracted can be used to monitor the progress of tasks that scans over blob IDs.
+    pub fn first_two_bytes(&self) -> u16 {
+        u16::from_be_bytes(self.0[0..2].try_into().unwrap())
```
mlegner: Nit: Can you replace the `unwrap` by an `expect`?
```suggestion
        u16::from_be_bytes(self.0[0..2].try_into().expect("two bytes can be converted to a u16"))
```

halfprice: Done

PR #1345, opened 2024-12-24
feat(client): add jwt layer for publisher

### LINKED ISSUE #1316: client: add auth for publisher
### Introduction
The publisher can run as internal service, people will like to use it with auth layer.  Such that no additional reverse proxy needed for auth and the overall performance of publisher service will be great, and easier to use. And it will be flexible to fit in any kind of existing web service.

### Proposal
First, the proposal is to add an optional publisher config, namely `publisher_config.yaml`, or an optional field `publisher_config` in `client_config.yaml`.  Please noted, the config memory copy will be different from Daemon and Client, so no mater we put the publisher config in side the `client_config.yaml` or not we need to take/copy it into daemon.

For now, there will be a optional `jwt_decode_secret` field for a static decoding key under `publish_config`, user can easier to change the key by themself for any other service.  In the future, we can easier to add another optional fields for other key management service.

The another optional parameter will implement at the same time is `jwt_token_expire_sec`.  If the request is from weird in too long ago, we will drop the request.

yanganto:
## Description
Add optional JWT auth layer for publisher.
Fix #1316 

#### Mechanism
The objective for this layer is to provide an optional authentication layer
1. If the secret is provided with optional algorithm setting, the JWT token and the `exp` will be verified to authenticate the user
2. If the `jwt_expiring_sec` is provided, the layer will check the expiring time follow the agreement of token usage, this prevent abusing never expiring tokens or any secret leak from other party with a different expiring period.
3. If the `jwt_verify_upload` is set, the layer will use the extra `address` and `epoch` field from the token to make sure store in a valid epoch and the sui blob object will send to the valid owner
4. In some using scenario, an APIGateway check the JWT token before the request coming into the publisher, and we can disable JWT token verification and the `jwt_expiring_sec` and `jwt_verify_upload` still work, so publisher can only do the additional check and safe computing.

## Test plan
unit tests on axum Services
- basic `auth_layer` with HMAC (1.)
- `verify_upload` (3.)
- `verify_upload_skip_check_token` for the APIGateway scenario (4.)
- `verify_exp` period to prevent abusing (2.)
- `eddsa_auth` for one of asymmetric decryption mechanism

There is no test for RSA relating mechanism, because all the decryption is using `jsonwebtoken` crate, so we do not need redundant test on these.

---

## Release notes
- [ ] Storage node:
- [ ] Aggregator:
- [X] Publisher: Add optional JWT auth layer for publisher
- [ ] CLI:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -617,7 +617,7 @@ checksum = "de3415c9dbaf54397292da0bb81a907e2b989661ce068e4ccfebac33dc9e245e"
 dependencies = [
  "async-graphql",
  "async-trait",
- "axum",
+ "axum 0.7.9",
  "bytes",
  "futures-util",
  "serde_json",
@@ -666,15 +666,6 @@ dependencies = [
  "indexmap 2.7.0",
  "serde",
  "serde_json",
-]
-
-[[package]]
-name = "async-lock"
-version = "2.8.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "287272293e9d8c41773cec55e365490fe034813a2f172f502d6ddcf75b2f582b"
-dependencies = [
- "event-listener 2.5.3",
 ]
 
 [[package]]
@@ -1104,12 +1095,40 @@ dependencies = [
 
 [[package]]
 name = "axum"
+version = "0.6.20"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "3b829e4e32b91e643de6eafe82b1d90675f5874230191a4ffbc1b336dec4d6bf"
+dependencies = [
+ "async-trait",
+ "axum-core 0.3.4",
+ "bitflags 1.3.2",
+ "bytes",
+ "futures-util",
+ "http 0.2.12",
+ "http-body 0.4.6",
+ "hyper 0.14.32",
+ "itoa",
+ "matchit 0.7.3",
+ "memchr",
+ "mime",
+ "percent-encoding",
+ "pin-project-lite",
+ "rustversion",
+ "serde",
+ "sync_wrapper 0.1.2",
+ "tower 0.4.13",
+ "tower-layer",
+ "tower-service",
+]
+
+[[package]]
+name = "axum"
 version = "0.7.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "edca88bc138befd0323b20752846e6587272d3b03b0343c8ea28a6f819e6e71f"
 dependencies = [
  "async-trait",
- "axum-core",
+ "axum-core 0.4.5",
  "axum-macros",
  "base64 0.22.1",
  "bytes",
@@ -1142,6 +1161,23 @@ dependencies = [
 
 [[package]]
 name = "axum-core"
+version = "0.3.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "759fa577a247914fd3f7f76d62972792636412fbfd634cd452f6a385a74d2d2c"
+dependencies = [
+ "async-trait",
+ "bytes",
+ "futures-util",
+ "http 0.2.12",
+ "http-body 0.4.6",
+ "mime",
+ "rustversion",
+ "tower-layer",
+ "tower-service",
+]
+
+[[package]]
+name = "axum-core"
 version = "0.4.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "09f2bd6146b97ae3359fa0cc6d6b376d9539582c7b4220f041a33ec24c226199"
@@ -1167,8 +1203,8 @@ version = "0.9.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c794b30c904f0a1c2fb7740f7df7f7972dfaa14ef6f57cb6178dc63e5dca2f04"
 dependencies = [
- "axum",
- "axum-core",
+ "axum 0.7.9",
+ "axum-core 0.4.5",
  "bytes",
  "fastrand",
  "futures-util",
@@ -1371,15 +1407,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d86b93f97252c47b41663388e6d155714a9d0c398b99f1005cbc5f978b29f445"
 
 [[package]]
-name = "beef"
-version = "0.5.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3a8241f3ebb85c056b509d4327ad0358fbbba6ffb340bf388f26350aeda225b1"
-dependencies = [
- "serde",
-]
-
-[[package]]
 name = "bellpepper"
 version = "0.4.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1425,8 +1452,8 @@ dependencies = [
 
 [[package]]
 name = "bin-version"
-version = "1.40.1"
-source = "git+https://github.com/MystenLabs/sui?tag=testnet-v1.40.1#5ff708baf2b6591fcbaac7545f4c78953c76a521"
+version = "1.41.0"
+source = "git+https://github.com/MystenLabs/sui?tag=testnet-v1.41.0#4b4e533f03530cc8eb9af1ed004aecfe95756e28"
 dependencies = [
  "const-str",
  "git-version",
@@ -1710,16 +1737,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bstr"
-version = "1.11.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "531a9155a481e2ee699d4f98f43c0ca4ff8ee1bfd55c31e9e98fb29d2b176fe0"
-dependencies = [
- "memchr",
- "serde",
-]
-
-[[package]]
 name = "bumpalo"
 version = "3.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1899,6 +1916,12 @@ dependencies = [
  "libc",
  "shlex",
 ]
+
+[[package]]
+name = "cesu8"
+version = "1.1.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6d43a04d8753f35258c91f8ec639f792891f748a1edbd759cf1dcea3382ad83c"
 
 [[package]]
 name = "cexpr"
@@ -2180,7 +2203,7 @@ dependencies = [
 [[package]]
 name = "consensus-config"
 version = "0.1.0"
-source = "git+https://github.com/MystenLabs/sui?tag=testnet-v1.40.1#5ff708baf2b6591fcbaac7545f4c78953c76a521"
+source = "git+https://github.com/MystenLabs/sui?tag=testnet-v1.41.0#4b4e533f03530cc8eb9af1ed004aecfe95756e28"
 dependencies = [
  "fastcrypto 0.1.8 (git+https://github.com/MystenLabs/fastcrypto?rev=69d496c71fb37e3d22fe85e5bbfd4256d61422b9)",
  "mysten-network",
@@ -2192,7 +2215,7 @@ dependencies = [
 [[package]]
 name = "consensus-core"
 version = "0.1.0"
-source = "git+https://github.com/MystenLabs/sui?tag=testnet-v1.40.1#5ff708baf2b6591fcbaac7545f4c78953c76a521"
+source = "git+https://github.com/MystenLabs/sui?tag=testnet-v1.41.0#4b4e533f03530cc8eb9af1ed004aecfe95756e28"
 dependencies = [
  "anemo",
  "anemo-build",
@@ -2210,9 +2233,6 @@ dependencies = [
  "fastcrypto 0.1.8 (git+https://github.com/MystenLabs/fastcrypto?rev=69d496c71fb37e3d22fe85e5bbfd4256d61422b9)",
  "futures",
  "http 1.2.0",
- "hyper 1.5.2",
- "hyper-rustls 0.27.5",
- "hyper-util",
  "itertools 0.13.0",
  "mockall 0.11.4",
  "mysten-common",
@@ -2221,23 +2241,23 @@ dependencies = [
  "nom",
  "parking_lot 0.12.3",
  "prometheus",
- "prost",
+ "prost 0.13.4",
  "quinn-proto",
  "rand 0.8.5",
  "rustls 0.23.21",
  "serde",
  "shared-crypto",
  "strum_macros 0.24.3",
+ "sui-http",
  "sui-macros",
  "sui-protocol-config",
  "sui-tls",
  "tap",
  "thiserror 1.0.69",
  "tokio",
- "tokio-rustls 0.26.1",
  "tokio-stream",
  "tokio-util 0.7.13",
- "tonic",
+ "tonic 0.12.3",
  "tonic-build",
  "tonic-rustls",
  "tower 0.4.13",
@@ -2257,6 +2277,43 @@ dependencies = [
  "once_cell",
  "unicode-width 0.2.0",
  "windows-sys 0.59.0",
+]
+
+[[package]]
+name = "console-api"
+version = "0.6.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd326812b3fd01da5bb1af7d340d0d555fd3d4b641e7f1dfcf5962a902952787"
+dependencies = [
+ "futures-core",
+ "prost 0.12.6",
+ "prost-types 0.12.6",
+ "tonic 0.10.2",
+ "tracing-core",
+]
+
+[[package]]
+name = "console-subscriber"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7481d4c57092cd1c19dd541b92bdce883de840df30aa5d03fd48a3935c01842e"
+dependencies = [
+ "console-api",
+ "crossbeam-channel",
+ "crossbeam-utils",
+ "futures-task",
+ "hdrhistogram",
+ "humantime",
+ "prost-types 0.12.6",
+ "serde",
+ "serde_json",
+ "thread_local",
+ "tokio",
+ "tokio-stream",
+ "tonic 0.10.2",
+ "tracing",
+ "tracing-core",
+ "tracing-subscriber",
 ]
 
 [[package]]
@@ -3205,7 +3262,7 @@ dependencies = [
 [[package]]
 name = "enum-compat-util"
 version = "0.1.0"
-source = "git+https://github.com/MystenLabs/sui?tag=testnet-v1.40.1#5ff708baf2b6591fcbaac7545f4c78953c76a521"
+source = "git+https://github.com/MystenLabs/sui?tag=testnet-v1.41.0#4b4e533f03530cc8eb9af1ed004aecfe95756e28"
 dependencies = [
  "serde_yaml 0.8.26",
 ]
@@ -3584,12 +3641,6 @@ checksum = "b90ca2580b73ab6a1f724b76ca11ab632df820fd6040c336200d2c1df7b3c82c"
 
 [[package]]
 name = "event-listener"
-version = "2.5.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0206175f82b8d6bf6652ff7d71a1e27fd2e4efde587fd368662814d6ec1d9ce0"
-
-[[package]]
-name = "event-listener"
 version = "5.4.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3492acde4c3fc54c845eaab3eed8bd00c7a7d881f78bfc801e43a93dec1331ae"
@@ -3605,7 +3656,7 @@ version = "0.5.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3c3e4e0dd3673c1139bf041f3008816d9cf2946bbfac2945c09e523b8d7b05b2"
 dependencies = [
- "event-listener 5.4.0",
+ "event-listener",
  "pin-project-lite",
 ]
 
@@ -4250,19 +4301,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a8d1add55171497b4705a648c6b583acafb01d58050a51727785f0b2c8e0a2b2"
 
 [[package]]
-name = "globset"
-version = "0.4.15"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "15f1ce686646e7f1e19bf7d5533fe443a45dbfb990e00629110797578b42fb19"
-dependencies = [
- "aho-corasick",
- "bstr",
- "log",
- "regex-automata 0.4.9",
- "regex-syntax 0.8.5",
-]
-
-[[package]]
 name = "gloo-timers"
 version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4693,7 +4731,6 @@ dependencies = [
  "rustls-native-certs 0.6.3",
  "tokio",
  "tokio-rustls 0.23.4",
- "webpki-roots 0.22.6",
 ]
 
 [[package]]
@@ -4722,6 +4759,7 @@ dependencies = [
  "http 1.2.0",
  "hyper 1.5.2",
  "hyper-util",
+ "log",
  "rustls 0.23.21",
  "rustls-native-certs 0.8.1",
  "rustls-pki-types",
@@ -4729,6 +4767,18 @@ dependencies = [
  "tokio-rustls 0.26.1",
  "tower-service",
  "webpki-roots 0.26.7",
+]
+
+[[package]]
+name = "hyper-timeout"
+version = "0.4.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bbb958482e8c7be4bc3cf272a766a2b0bf1a6755e7a6ae777f017a31d11b13b1"
+dependencies = [
+ "hyper 0.14.32",
+ "pin-project-lite",
+ "tokio",
+ "tokio-io-timeout",
 ]
 
 [[package]]
@@ -5245,6 +5295,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "jni"
+version = "0.19.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c6df18c2e3db7e453d3c6ac5b3e9d5182664d28788126d39b91f2d1e22b017ec"
+dependencies = [
+ "cesu8",
+ "combine",
+ "jni-sys",
+ "log",
+ "thiserror 1.0.69",
+ "walkdir",
+]
+
+[[package]]
+name = "jni-sys"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8eaf4bc02d17cbdd7ff4c7438cafcdf7fb9a4613313ad11b4f8fefe7d3fa0130"
+
+[[package]]
 name = "jobserver"
 version = "0.1.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5280,8 +5350,9 @@ dependencies = [
 
 [[package]]
 name = "jsonrpsee"
-version = "0.16.2"
-source = "git+https://github.com/wlmyng/jsonrpsee.git?rev=b1b300784795f6a64d0fcdf8f03081a9bc38bde8#b1b300784795f6a64d0fcdf8f03081a9bc38bde8"
+version = "0.24.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c5c71d8c1a731cc4227c2f698d377e7848ca12c8a48866fc5e6951c43a4db843"
 dependencies = [
  "jsonrpsee-core",
  "jsonrpsee-http-client",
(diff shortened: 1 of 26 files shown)


crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -459,9 +463,26 @@ impl ClientCommandRunner {
...
+                .take()
+                .map(AuthConfig::new)
+                .unwrap_or(Ok(AuthConfig::default()))?;
+            c.expiring_sec = args.jwt_expiring_sec;
+            c.verify_upload = args.jwt_verify_upload;
+            c.algorithm = args.jwt_algorithm;
+            Some(c)
+        } else {
+            None
+        };
```
giac-mysten: Suggestion: maybe separate this to another function?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -459,9 +463,26 @@ impl ClientCommandRunner {
             &args,
         )
         .await?;
+        let auth_config = if args.jwt_decode_secret.is_some()
+            || args.jwt_expiring_sec > 0
+            || args.jwt_verify_upload
+        {
+            let mut c = args
```
giac-mysten: Nit: avoid 1-letter variable names

crates/walrus-service/src/client/error.rs
```diff
@@ -22,6 +22,13 @@ pub enum StoreError {
     Confirmation(NodeError),
 }
 
+/// Transforming arguments to config, failed.
+#[derive(Debug, thiserror::Error)]
+pub enum ConfigError {
```
giac-mysten: Question/suggestion: do we expect additional variants here? If not, convert to a struct.

giac-mysten: Suggestion: Maybe a more specific name?
```suggestion
pub enum PublisherAuthConfigError {
```

crates/walrus-service/src/client/error.rs
```diff
@@ -22,6 +22,13 @@ pub enum StoreError {
     Confirmation(NodeError),
 }
 
+/// Transforming arguments to config, failed.
```
giac-mysten: Nit:
```suggestion
/// The contents of the publisher's authentication config are invalid. 
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -465,6 +466,28 @@ pub struct PublisherArgs {
     #[clap(long, action)]
     #[serde(default)]
     pub keep: bool,
+    /// If set, the publisher will verify the JWT token.
+    /// If API Gateway already check the token, we can skip the signature validation
+    /// The secret can be hexyl string start with `0x`.
```
giac-mysten:
```suggestion
    /// If set, the publisher will verify the JWT token.
    ///
    /// If not specified, the verification is diabled. This is useful, e.g., in case the API Gateway has already checked the token.
    /// The secret can be hex string, starting with `0x`.
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -465,6 +466,28 @@ pub struct PublisherArgs {
...
+    #[clap(long)]
+    #[serde(default)]
+    pub jwt_decode_secret: Option<String>,
+    /// If unset, the jwt algorithm will be HMAC.
+    /// Following algorithms are supported "HS256", "HS384", "HS512", "ES256", "ES384", "RS256",
+    /// "RS384", "PS256", "PS384", "PS512", "RS512", "EdDSA"
```
giac-mysten:
```suggestion
    /// If unset, the JWT authentication algorithm will be HMAC.
    ///
    /// The following algorithms are supported: "HS256", "HS384", "HS512", "ES256", "ES384", "RS256",
    /// "RS384", "PS256", "PS384", "PS512", "RS512", "EdDSA".
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -465,6 +466,28 @@ pub struct PublisherArgs {
...
+    #[clap(long)]
+    #[serde(default)]
+    pub jwt_algorithm: Option<Algorithm>,
+    /// If set and greater than 0, `exp` will be check based on `iat`
```
giac-mysten:
```suggestion
    /// If set and greater than 0, the publisher will check if the JWT token is expired based on the "issued at" (`iat`) value.
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -465,6 +466,28 @@ pub struct PublisherArgs {
...
+    #[clap(long)]
+    #[serde(default)]
+    pub jwt_expiring_sec: u64,
+    /// - Verify `epochs` in query should be the same as `epoch` in JWT claim if set
+    /// - Verify `send_object_to` in query should be the same as `address` in JWT claim if set
+    // TODO: /// - Verify the size/hash of uploaded file
```
giac-mysten:
```suggestion
    /// If set, the publisher will verify that the requested upload matches the claims in the JWT.
    ///
    /// Specifically, the publisher will:
    /// - Verify that the number of `epochs` in query is the the same as `epoch` in the JWT, if
    ///   present;
    /// - Verify that the `send_object_to` field in the query is the same as the `address` in the
    ///   JWT, if present;
    // TODO: /// - Verify the size/hash of uploaded file
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -465,6 +466,28 @@ pub struct PublisherArgs {
...
+    /// "RS384", "PS256", "PS384", "PS512", "RS512", "EdDSA"
+    #[clap(long)]
+    #[serde(default)]
+    pub jwt_algorithm: Option<Algorithm>,
+    /// If set and greater than 0, `exp` will be check based on `iat`
+    #[clap(long)]
+    #[serde(default)]
+    pub jwt_expiring_sec: u64,
+    /// - Verify `epochs` in query should be the same as `epoch` in JWT claim if set
+    /// - Verify `send_object_to` in query should be the same as `address` in JWT claim if set
```
giac-mysten: Question: can we keep the names of `epochs` and `send_object_to` in the JWT as well, to avoid confusion?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -459,9 +463,26 @@ impl ClientCommandRunner {
             &args,
         )
         .await?;
+        let auth_config = if args.jwt_decode_secret.is_some()
+            || args.jwt_expiring_sec > 0
+            || args.jwt_verify_upload
```
giac-mysten: Question: does it make sense to have `jwt_decode_secret == None`, but `args.jwt_verify_upload == true` or `args.jwt_expiring_sec > 0 == true`? 

Is this the case in which the API gateway has made the check before? But then, should it not have checked expiry and the upload as well?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -494,8 +515,18 @@ impl ClientCommandRunner {
...
         args.print_debug_message("attempting to run the Walrus daemon");
+        let auth_config = if let Some(secret) = args.jwt_decode_secret.take() {
+            let mut c = AuthConfig::new(secret).map_err(anyhow::Error::msg)?;
+            c.expiring_sec = args.jwt_expiring_sec;
+            c.verify_upload = args.jwt_verify_upload;
+            c.algorithm = args.jwt_algorithm;
+            Some(c)
+        } else {
+            None
+        };
```
giac-mysten: Nit: maybe add debug logs with the created config? Also above, possibly

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -494,8 +515,18 @@ impl ClientCommandRunner {
         Ok(())
     }
 
-    pub(crate) async fn daemon(self, registry: &Registry, args: PublisherArgs) -> Result<()> {
+    pub(crate) async fn daemon(self, registry: &Registry, mut args: PublisherArgs) -> Result<()> {
         args.print_debug_message("attempting to run the Walrus daemon");
+        let auth_config = if let Some(secret) = args.jwt_decode_secret.take() {
+            let mut c = AuthConfig::new(secret).map_err(anyhow::Error::msg)?;
```
giac-mysten: Question: why are daemon and publisher creating the `auth_config` differently? should they not use the same function?

crates/walrus-service/src/client/config.rs
```diff
@@ -95,6 +99,45 @@ pub enum ExchangeObjectConfig {
     Multiple(Vec<ObjectID>),
 }
 
+/// Configure for publisher authentication
```
giac-mysten:
```suggestion
/// Confugration for the JWT authentication on the publisher.
```

crates/walrus-service/src/client/config.rs
```diff
@@ -95,6 +99,45 @@ pub enum ExchangeObjectConfig {
...
+#[derive(Default, Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+pub struct AuthConfig {
+    /// The scecret for JWT decode
```
giac-mysten:
```suggestion
    /// The secret with which to authenticate the JWT.
```

crates/walrus-service/src/client/config.rs
```diff
@@ -95,6 +99,45 @@ pub enum ExchangeObjectConfig {
...
+    pub(crate) secret: Option<Vec<u8>>,
+    /// The algorithm for JWT
```
giac-mysten:
```suggestion
    /// The authentication algorithm for the JWT.
```

crates/walrus-service/src/client/config.rs
```diff
@@ -95,6 +99,45 @@ pub enum ExchangeObjectConfig {
...
+    pub(crate) algorithm: Option<Algorithm>,
+    /// If not 0, publisher will verify the expiring seconds
```
giac-mysten:
```suggestion
    /// The duration, in seconds, after which the publisher will consider the JWT as expired.
```

crates/walrus-service/src/client/config.rs
```diff
@@ -95,6 +99,45 @@ pub enum ExchangeObjectConfig {
...
+    pub(crate) expiring_sec: u64,
```
giac-mysten, line 119: Question: any reason to use `u64` instead of eg `std::time::Duration`?

crates/walrus-service/src/client/config.rs
```diff
@@ -95,6 +99,45 @@ pub enum ExchangeObjectConfig {
...
+    /// verify upload file size and address for `send_object_to`
```
giac-mysten:
```suggestion
    /// If `true, the publisher verifies the upload file size, and the address to which the object is sent after publishing (`send_object_to`).
```
Question: Is the upload size not a todo? And what about `epochs`?

crates/walrus-service/src/client/config.rs
```diff
@@ -95,6 +99,45 @@ pub enum ExchangeObjectConfig {
...
+                    ..Default::default()
+                })
+            }
+        } else {
+            Ok(Self {
+                secret: Some(secret.into_bytes()),
+                ..Default::default()
+            })
+        }
+    }
```
giac-mysten: Suggestion: this could be simplified, I think, by using `serde_as`? similarly to what is done in `sui_serde::HexAccountAddress` in the sui repo for example.

crates/walrus-service/src/client/daemon.rs
```diff
@@ -213,19 +223,38 @@ impl<T: WalrusWriteClient + Send + Sync + 'static> ClientDaemon<T> {
...
+                    .layer(BufferLayer::new(max_request_buffer_size))
+                    .layer(ConcurrencyLimitLayer::new(max_concurrent_requests));
+                self.router = self.router.route(
+                    BLOB_PUT_ENDPOINT,
+                    put(routes::put_blob)
+                        .route_layer(publisher_layers)
+                        .options(routes::store_blob_options),
+                );
+            }
+        }
```
giac-mysten: Suggestion: may be a tiny bit cleaner, up to you:
```suggestion
        let base_layers = ServiceBuilder::new()
            .layer(DefaultBodyLimit::max(max_body_limit))
            .layer(HandleErrorLayer::new(handle_publisher_error))
            .layer(LoadShedLayer::new())
            .layer(BufferLayer::new(max_request_buffer_size))
            .layer(ConcurrencyLimitLayer::new(max_concurrent_requests));

        if let Some(auth_config) = auth_config {
            self.router = self.router.route(
                BLOB_PUT_ENDPOINT,
                put(routes::put_blob)
                    .route_layer(
                        ServiceBuilder::new()
                            .layer(JwtLayer::new(auth_config))
                            .layer(base_layers),
                    )
                    .options(routes::store_blob_options),
            );
        } else {
            self.router = self.router.route(
                BLOB_PUT_ENDPOINT,
                put(routes::put_blob)
                    .route_layer(base_layers)
                    .options(routes::store_blob_options),
            );
        }
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+use tracing::error;
+
+use crate::client::config::AuthConfig;
+
+/// Claim follow RFC7519 with extra storage parameters: address, epoch
+#[derive(Clone, Deserialize, Debug)]
+#[cfg_attr(test, derive(serde::Serialize))]
+struct Claim {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    /// Token is issued at (timestamp)
```
giac-mysten:
```suggestion
    /// Token is issued at (timestamp).
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+    pub iat: Option<u64>,
+
+    /// Token is expired at (timestamp)
```
giac-mysten:
```suggestion
    /// Token is expires at (timestamp).
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+    pub exp: u64,
+
+    /// Address is the sui blob object owner
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub address: Option<String>,
```
giac-mysten: Question: can we call this `send_object_to`?

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+
+    /// Epoch is the storage time
```
giac-mysten: Question: this is the number of epochs in the future, right?
```suggestion
    /// The number of epochs the blob should be stored for.
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub epoch: Option<u32>,
```
giac-mysten:
```suggestion
    pub epochs: Option<u32>,
```

crates/walrus-service/src/client/error.rs
yanganto: I can use a struct for now, once we need more, we do it enum.

yanganto: It is `DecodeError`, I think it is general error easily happened in config, it is changed into a error struct , so we will not have no naming issue in the enum.

crates/walrus-service/src/client/cli/args.rs
yanganto: Yes, I will change these.

crates/walrus-service/src/client/cli/runner.rs
yanganto: It make sense, APIGateway just share the loading on verify the request.
I think normally using scenario is enable all, and once we want to share the loading of publiser to get a better performance, we will remove the decode secret and put the APIGateway in front of it, and `jwt_verify_upload` and `jwt_expiring_sec` is still set.  

By the way, APIGateway is not a universal medicine to solve every conditions, we already have a discussion in the [thread](https://suifoundation.slack.com/archives/C05LR8KJP8U/p1734689383509879?thread_ts=1734572995.545039&cid=C05LR8KJP8U)

crates/walrus-service/src/client/config.rs
yanganto: We will check the `exp - iat` with this setting.
The `exp` and `iat` will need deserialize from the token into `Datetime`, and we do not force user to use `iat` in the JWT claim [here](https://github.com/MystenLabs/walrus/blob/272ab4bb57046ed98102d9d07afb81aba373627c/crates/walrus-service/src/client/daemon/auth.rs#L22), so it is option type with this macro `skip_serializing_if = "Option::is_none"`, Datetime is a generic struct with timezone, and also `i64` casting inside.  Overall, it is overkill for us.

Most important thing is we even can accept the extra code complexity, but it will be painful to make sure we are not reducing the performance in every dependency update because the token deserialization and verification will do in every request.

That is the reason I think `u64` is good and clear enough here.

crates/walrus-service/src/client/daemon.rs
yanganto: Great. It will be easier to find out the difference.

crates/walrus-service/src/client/config.rs
yanganto: Yes, it is `epochs`.

crates/walrus-service/src/client/config.rs
yanganto: No, I think it is different, the length of account is determined in runtime, so you can use a fixed length slice then deser it.

For the current case, the secret field is dynamic and can use in any kind algorithums.  I believe the current implementation is future prevent.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+    pub iat: Option<u64>,
+
+    /// Token expires at (timestamp).
+    pub exp: u64,
+
+    /// The owner address of the sui blob object.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub send_object_to: Option<String>,
+
+    /// The number of epochs the blob should be stored for.
```
gdanezis, line 31: I suspect this is an attempt to prevent "too expensive" publish operations. We seem to be limiting the duration but not the size? 

I wonder @giac-mysten : would it instead make sense to include here a WAL value? Namely say: I authorize this token for X amount of WAL over a time period? The challenge is how to ensure some accounting in the publisher to not exceed this.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+
+/// Middleware for validating a valid JWT token is present on "authorization: bearer <token>"
+#[derive(Clone)]
+pub struct Jwt<S> {
+    inner: S,
+    auth_config: Box<AuthConfig>,
+    _phantom: PhantomData<Claim>,
+}
+
+#[pin_project(project = JwtFutureProj)]
```
gdanezis: Writing Futures by hand is pretty hardcore, but hard to maitain. Are we sure we cannot compose one with off-the-shelf rust async primitives like notifiers and channels?

crates/walrus-service/src/client/daemon/auth.rs
yanganto: **Every parameter will able be verified.**  
Publisher only trust the token issuer not the user who uploading the file.
Issuer will have all condition and flexible to restrict what user can upload.

There will be a next PR to add  (`blob-id`, `shard_number`) and/or file size in the claim, so the publiser have more information, and verify the uploading from the user is valid as issuer grant. There is also a [TODO](https://github.com/MystenLabs/walrus/blob/aa8dfea8ba1bee1e6c6447361089a01c056f4deb/crates/walrus-service/src/client/cli/args.rs#L497) for this.  We will do it, because we have campaign need this, so it will not be a TODO but never done. :smile: The hash/size is separated because I believe we will discuss when is the good timing to reject the upload, and it will out of auth layer topic.

Overall, publisher turn the web3 back to web2, so the issuer is the point the see something in web3 and then grant a JWT session token in web2.  In other words, the publisher is customs, the issuer is the government, and the JWT token is the (passport + visa).

crates/walrus-service/src/client/daemon/auth.rs
yanganto: I had tried implement the mechanism in easy ways, and eventually a layer is needed, the key point is the auth config is dynamic, it is not determined in compiling time.  A function as a layer is not good enough for our case.  If you already have some idea on this, could you show me some code snippets, I am happy to dig in more and make it compile and work on rest of things. :pray:

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+    fn layer(&self, inner: S) -> Self::Service {
+        Jwt {
+            inner,
+            auth_config: Box::new(self.auth_config.clone()),
+            _phantom: self._phantom,
+        }
+    }
+}
+
+/// Middleware for validating a valid JWT token is present on "authorization: bearer <token>"
```
giac-mysten:
```suggestion
/// Middleware for validating that a valid JWT token is present in "authorization: bearer <token>".
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+#[pin_project(project = JwtFutureProj)]
+pub enum JwtFuture<
+    TService: Service<Request<ReqBody>, Response = Response<ResBody>>,
+    ReqBody,
+    ResBody,
+> {
+    ValidateError(StatusCode),
+    WaitForFuture {
+        #[pin]
+        future: TService::Future,
```
giac-mysten: Question: do I understand correctly that this is needed to ensure that we can either return early upon validation failure, or continue polling the service future if validation is fine?

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+                match Claim::from_token(bearer.token().trim(), &decode_key, &validation) {
+                    Ok(claim) => {
+                        let mut valid_upload = true;
+                        if self.auth_config.expiring_sec > 0
+                            && (claim.exp - claim.iat.unwrap_or_default())
+                                != self.auth_config.expiring_sec
+                        {
+                            error!("invalid expiring token: {}", bearer.token());
+                            valid_upload = false;
+                        }
```
giac-mysten, line 176: question: the problem here is that the expiry duration is different from the one that is set in the authentication.
suggestion: should we allow shorter durations, but not longer?

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+                        if self.auth_config.verify_upload {
+                            let query = req.uri().query();
+                            if let Some(epochs) = claim.epochs {
+                                if !check_query(query, "epochs", epochs.to_string()) {
+                                    error!("upload with invalid epochs: {}", epochs);
```
giac-mysten: Nit: use the full path, and prefer using [fields](https://docs.rs/tracing/latest/tracing/#recording-fields) instead of string formatting.
```suggestion
                                    tracing::error!(epochs, "upload with invalid epochs");
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+fn check_query(queries: Option<&str>, field: &str, value: impl Display) -> bool {
+    if let Some(queries) = queries {
+        for q in queries.split('&') {
+            if q.starts_with(field) && q != format!("{field:}={value:}") {
+                return false;
+            }
+        }
+    }
+    true
+}
```
giac-mysten: Question: this basically takes the query string, splits it and checks that there is a string with the expected parameters. It makes me a bit uncomfortable. Could we parse the query string instead (maybe serde) according to some expected template, and then use that to perform the checks?

This would have the added benefit that we could say, for example, "accept requests for up to 10 epochs" and not "exactly 10 epochs".

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+                            && (claim.exp - claim.iat.unwrap_or_default())
+                                != self.auth_config.expiring_sec
+                        {
+                            error!("invalid expiring token: {}", bearer.token());
+                            valid_upload = false;
+                        }
+                        if self.auth_config.verify_upload {
+                            let query = req.uri().query();
+                            if let Some(epochs) = claim.epochs {
+                                if !check_query(query, "epochs", epochs.to_string()) {
```
giac-mysten, line 180: Question: Can we also check that epochs <= max_epochs? instead of just epochs == max_epochs? see other comments on `check_query`.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: Maybe @jpcsmith has more experience with such tower issues?

crates/walrus-service/src/client/daemon/auth.rs
yanganto: The `expiring_sec` should be a part of agreement between issuer and publisher, even the token seems valid, but it issuer does not follow the agreement.  It is really possible the secret is leak, so that is the reason I will reject it, not just providing error log here.  

If we do not like this trick, we can set `expiring_sec` to 0(the default), and the `iat` is not required.  So the overhead will be less.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: Exactly, If there is jwt token, and it can not verified. The request will go [here](https://github.com/MystenLabs/walrus/blob/53771431c162700d44362b04a4d66e90a281fce3/crates/walrus-service/src/client/daemon/auth.rs#L231).
If there is an error token or absent, the request will go [here](https://github.com/MystenLabs/walrus/blob/53771431c162700d44362b04a4d66e90a281fce3/crates/walrus-service/src/client/daemon/auth.rs#L234) or [here](https://github.com/MystenLabs/walrus/blob/53771431c162700d44362b04a4d66e90a281fce3/crates/walrus-service/src/client/daemon/auth.rs#L237)

These are early return.  The only thing is we need to keep the auth layer the most outside place, when assemble router.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: Yeh, the epochs is not force field, and we can have a `max-epochs` for this purpose.  Will we let this in other PR? I can do it later, and I think this is different topic.  I can create an issue and working on this when this PR merged.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: We can do this in another feature, I think this is beyond current thread, and the exactly `epochs` is still needed no mater we have `max-epochs` or not.
https://github.com/MystenLabs/walrus/pull/1345#discussion_r1912472431

Cargo.lock
```diff
@@ -11623,6 +11629,7 @@ dependencies = [
  "pin-project",
  "prettytable",
  "prometheus",
+ "querystr",
```
giac-mysten, line 14156: Question: do i understand correctly that this your own fork of the (unmaintainde) `querystring` crate? since the library seems tiny, can we import the functions directly in the codebase so that we know they will be maintained?

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -239,10 +239,10 @@ where
     }
 }
 
-fn check_query(queries: Option<&str>, field: &str, value: impl Display) -> bool {
+fn check_query(queries: Option<&str>, field: &str, value: impl AsRef<str>) -> bool {
     if let Some(queries) = queries {
-        for q in queries.split('&') {
-            if q.starts_with(field) && q != format!("{field:}={value:}") {
+        for (key, var) in querystr::querify(queries) {
+            if key == field && var != value.as_ref() {
```
giac-mysten, line 219: Remark: we still need exact matches here, and we cannot do `<` or `>=` comparisons with the values. I guess this can be done in a followup.

Cargo.lock
yanganto: Yeh, exactly.  Parsing query string should be slim and simple to have a good performace, 
at first I use querystring it provides `Vec<(string, string)>` as an output,  it is better than `serde-querystring`.

And then I just thought we can do `&'str` to save more memory, and then I found a folk already did that.  I use the folk in Cargo.toml, but our CI rule denys the dependency on a github repo, because it is not a crate.  

It is really simple and good for our purpose, eventually we included it, and it is all about how. But it is MIT license, and the best way is to maintain it as crate and show the original author.  That is what I think to show the respect to the author.
https://github.com/yanganto/querystr/blob/532f72e330de4feee2fd9c805d1e221252897a68/Cargo.toml#L4

You will not worry about the maintenance, all my rust crate keep updated.  It is welcome that if mystenlabs want to take over it.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: For `max-epochs`, we can cast into `u32` as needed.  For now `&str` save memory here.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+    }
+}
+
+impl<S> Layer<S> for JwtLayer {
+    type Service = Jwt<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        Jwt {
+            inner,
+            auth_config: Box::new(self.auth_config.clone()),
```
jpcsmith: Question: Why clone + box as opposed to cloning an Arc?

crates/walrus-service/src/client/daemon/auth.rs
jpcsmith: Remark: The layer itself is good and probably provides better context than just having tower create the layer using [`tower::layer::layer_fn()`](https://docs.rs/tower/latest/tower/layer/fn.layer_fn.html).

The `JwtFuture` seems like it could be replaced with `https://docs.rs/futures/latest/futures/future/enum.Either.html`, where one variant is the inner future and the other branch is a `std::future::ready(response_body)`.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,638 @@
...
+    type Output = Result<TService::Response, TService::Error>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut std::task::Context<'_>) -> Poll<Self::Output> {
+        match self.as_mut().project() {
+            JwtFutureProj::ValidateError(code) => {
+                let response = Response::builder()
+                    .status(*code)
+                    .body(Default::default())
+                    .unwrap();
+                Poll::Ready(Ok(response))
```
jpcsmith: Important: Ensure that this conforms to the API error format (see `walrus_sdk::api::errors`). The easiest would be to define an error enum for variants that correspond the various failure cases, and `derive(RestApiError)` on it (see the `walrus-proc-macros`) crate or ping me on slack.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,646 @@
...
+                let response = Response::builder()
+                    .status(*code)
+                    .body(Default::default())
+                    .expect("Response is valid without any customized headers");
+                Poll::Ready(Ok(response))
+            }
+            JwtFutureProj::WaitForFuture { future } => future.poll(cx),
+        }
+    }
+}
```
jpcsmith:
```suggestion
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,646 @@
...
+impl<S, ReqBody, ResBody> Service<Request<ReqBody>> for Jwt<S>
+where
+    S: Service<Request<ReqBody>, Response = Response<ResBody>> + Send + Clone + 'static,
+    S::Future: Send + 'static,
+    ResBody: Default,
+    for<'de> Claim: Deserialize<'de> + Send + Sync + Clone + 'static,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    type Future = JwtFuture<S, ReqBody, ResBody>;
```
jpcsmith:
```suggestion
    type Future = Either<S::Future, Ready<Result<S::Response, S::Error>>>;
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,646 @@
...
+                            if let Some(send_object_to) = claim.send_object_to {
+                                if !check_query(query, "send_object_to", &send_object_to) {
+                                    error!("upload to an invalid address: {}", send_object_to);
+                                    valid_upload = false;
+                                }
+                            }
+                        }
+                        if valid_upload {
+                            let future = self.inner.call(req);
+                            Self::Future::WaitForFuture { future }
```
jpcsmith:
```suggestion
                            self.inner.call(req).left_future()
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,646 @@
...
+                        } else {
+                            Self::Future::ValidateError(StatusCode::PRECONDITION_FAILED)
```
jpcsmith:
```suggestion
                            validation_failed_future(StatusCode::PRECONDITION_FAILED).right_future()
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,646 @@
...
+                        }
+                    }
+                    Err(code) => Self::Future::ValidateError(code),
```
jpcsmith:
```suggestion
                    Err(code) => validation_failed_future(code).right_future(),
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,646 @@
...
+                }
+            }
+            None => Self::Future::ValidateError(StatusCode::UNAUTHORIZED),
```
jpcsmith:
```suggestion
            None => validation_failed_future(StatusCode::UNAUTHORIZED).right_future(),
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -0,0 +1,646 @@
...
+        }
+    }
+}
+
```
jpcsmith, line 205: Where the helper function would construct the body. Note, however, that this does **not** comply with our API responses.

```suggestion

fn validation_failed_future<B, E>(code: StatusCode) -> Ready<Result<Response<B>, E>>
where
    B: Default,
{
    std::future::ready(Ok(Response::builder()
        .status(code)
        .body(Default::default())
        .expect("Response is valid without any customized headers")))
}

```

PR #1346, opened 2024-12-24
feat: parallel fetching blob metadata during shard sync for new committee members

halfprice:
## Description

Currently we are fetching metadata 1 by 1, which is very inefficient. This PR changes metadata fetching to go in parallel. The parallelism is controlled by a config in node config.

The next step is to add an end point to fetch multiple metadata in a single request.

Contribute WAL-478

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -111,6 +111,11 @@ impl ShardSyncHandler {
     }
 
     /// Syncs the certified blob metadata before the current epoch.
+    ///
+    /// This function performs the following steps:
+    /// 1. Retrieves all certified blob info from storage before the current epoch
+    /// 2. Processes blobs concurrently up to max_concurrent_metadata_fetch limit
+    /// 3. For each blob, syncs its metadata using sync_single_blob_metadata
     async fn sync_certified_blob_metadata(&self) -> Result<(), SyncShardClientError> {
         tracing::info!("start syncing blob metadata");
         let blob_infos = self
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -119,42 +120,93 @@ impl ShardSyncHandler {
...
         }
 
-        // TODO(WAL-478):
-        //   - create a end point that can transfer multiple blob metadata at once.
-        //   - do this in parallel to speed up the sync.
+        let semaphore = Arc::new(tokio::sync::Semaphore::new(
+            self.config.max_concurrent_metadata_fetch,
+        ));
+        let mut futures = FuturesUnordered::new();
+        let mut active_count = 0;
```
mlegner, line 132: Question: Given we keep track of active syncs through this variable, why do we need the semaphore?

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -119,42 +120,93 @@ impl ShardSyncHandler {
...
-                        .expect("certified blob must have certified epoch set"),
-                )
-                .await?;
-
-            self.node.metrics.sync_blob_metadata_count.inc();
+            // TODO(WAL-478):
+            //   - create a end point that can transfer multiple blob metadata at once.
+            futures.push(tokio::spawn(async move {
+                Self::sync_single_blob_metadata(permit, node_clone, blob_id, blob_info).await
+            }));
```
mlegner: Question: Why do we need to spawn a new task here? I expect the vast majority of the delay to come from networking, not CPU, so it seems to me that using futures directly would be sufficient to enable parallelism here.

PR #1348, opened 2024-12-25
feat(staking): uses sorted committee to boost performance + optimise df access

damirka:
## Description

- adds sorting to the `Committee` type and its transitions;
- instead of iterating through dynamic fields in `advance_epoch`, next epoch public keys are collected at `voting_end`
- optimises `advance_epoch` function reducing the number of iterations and `VecMap` lookup performed there
- adds a `diff` function for the `Committee` to determine the diff faster than performing a lookup of each member

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/committee.move
+++ b/contracts/walrus/sources/staking/committee.move
@@ -155,11 +155,8 @@ public(package) fun to_inner(cmt: &Committee): VecMap<ID, vector<u16>> {
 /// sets of nodes, one set for nodes that are in the first committee but not in
 /// the second, and the other set is vice versa.
 ///
-/// Requires both committees to be sorted, so we can do a linear scan.
+/// Committee is always sorted by the node ID, so the diff algorithm is simple.
 public(package) fun diff(cmt_1: &Committee, cmt_2: &Committee): (vector<ID>, vector<ID>) {
-    assert!(cmt_1.is_sorted());
-    assert!(cmt_2.is_sorted());
-
     let mut i = 0;
     let mut j = 0;
     let mut diff_1 = vector[];


contracts/walrus/sources/staking/committee.move
```diff
@@ -145,3 +148,70 @@ public(package) fun inner(cmt: &Committee): &VecMap<ID, vector<u16>> {
 public(package) fun to_inner(cmt: &Committee): VecMap<ID, vector<u16>> {
     cmt.0
 }
+
+/// Finds the difference between two committees, returns the difference in two
+/// sets of nodes, one set for nodes that are in the first committee but not in
+/// the second, and the other set is vice versa.
```
benr-ml, line 156:
```suggestion
/// the second, and the other set is vice versa (i.e., cmt_1 \ cmt_2, cmt_2 \ cmt_1).
```

contracts/walrus/sources/staking/committee.move
```diff
@@ -145,3 +148,70 @@ public(package) fun inner(cmt: &Committee): &VecMap<ID, vector<u16>> {
...
+    };
+
+    (diff_1, diff_2)
+}
+
+/// Uses insertion sort to sort the committee by the node ID transformed into `u256`.
+///
+/// We sort by keys, but every swap operation also swaps the values, so the
+/// committee is sorted by node ID and the shards are kept in place.
+public(package) fun sort(cmt: Committee): Committee {
```
benr-ml: Committee should always be sorted, thus this function and the next one should be private or can even be removed

contracts/walrus/sources/staking/committee.move
```diff
@@ -122,7 +125,7 @@ public(package) fun transition(cmt: &Committee, mut new_assignments: VecMap<ID,
         new_cmt.insert(key, current_shards);
     });
 
-    Committee(new_cmt)
+    Committee(new_cmt).sort()
```
benr-ml: nit - 
```suggestion
    Committee(sort::sort_vec_map_by_node_id!(new_cmt))
```

contracts/walrus/sources/staking/committee.move
```diff
@@ -32,7 +33,9 @@ public(package) fun contains(cmt: &Committee, node_id: &ID): bool {
 /// - The values of assigned_number are <= 1000 (i.e., the limit of a vector)
 public(package) fun initialize(assigned_number: VecMap<ID, u16>): Committee {
     let mut shard_idx: u16 = 0;
-    let (keys, values) = assigned_number.into_keys_values();
+    let sorted_assignments = sort::sort_vec_map_by_node_id!(assigned_number);
```
benr-ml, line 38: let's document above `public struct Committee` that the vector is sorted by id

contracts/walrus/sources/staking/committee.move
```diff
@@ -145,3 +148,70 @@ public(package) fun inner(cmt: &Committee): &VecMap<ID, vector<u16>> {
...
     cmt.0
 }
+
+/// Finds the difference between two committees, returns the difference in two
+/// sets of nodes, one set for nodes that are in the first committee but not in
+/// the second, and the other set is vice versa.
+///
+/// Requires both committees to be sorted, so we can do a linear scan.
+public(package) fun diff(cmt_1: &Committee, cmt_2: &Committee): (vector<ID>, vector<ID>) {
+    assert!(cmt_1.is_sorted());
```
benr-ml: can remove this check (as we make sure it's an invariant of Committee above)

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -452,11 +460,23 @@ public(package) fun select_committee(self: &mut StakingInnerV1) {
...
     let distribution = vec_map::from_keys_values(active_ids, shards);
 
+    // sorts the distribution by node ID to simplify the transition.
+    let distribution = sort::sort_vec_map_by_node_id!(distribution);
+
     // if we're dealing with the first epoch, we need to assign the shards to the
     // nodes in a sequential manner. Assuming there's at least 1 node in the set.
     let committee = if (self.committee.size() == 0) committee::initialize(distribution)
-    else self.committee.transition(distribution);
+    else self.committee.transition(distribution).sort();
```
benr-ml: no need for .sort() (should be handled in transition)

contracts/walrus/sources/utils/sort.move
```diff
@@ -0,0 +1,50 @@
...
+
+    if (self.size() <= 1) return self;
+    let (mut keys, mut values) = self.into_keys_values();
+    let len = keys.length();
+    let mut i = 1;
+
+    while (i < len) {
+        let mut j = i;
+        while (j > 0 && keys[j - 1].to_address().to_u256() > keys[j].to_address().to_u256()) {
+            keys.swap(j - 1, j);
```
benr-ml, line 23: nit (not sure if significant) - you can create temp vector with (ID, u16) that holds an ID and it's index in self -> sort that vector using the current code -> update self accordingly.

contracts/walrus/sources/staking/committee.move
damirka: Done!

contracts/walrus/sources/utils/sort.move
damirka: It would've been great, but we don't support tuples.

contracts/walrus/sources/staking/committee.move
```diff
@@ -7,13 +7,16 @@
...
+use walrus::sort;
 
 // Error codes
 // Error types in `walrus-sui/types/move_errors.rs` are auto-generated from the Move error codes.
 const EInvalidShardAssignment: u64 = 0;
 
 /// Represents the current committee in the system. Each node in the committee
 /// has assigned shard IDs.
+///
+/// The `VecMap` inside the `Committee` is guaranteed to be sorted by the node ID.
```
karlwuest, line 19: Suggestion: I think if we use a sorted vector as underlying data structure, it makes sense to use it directly instead of a `VecMap`. If we need direct access to one committee member, we could then also (at some point) use binary search to access it instead of the linear scan that the VecMap performs.

karlwuest: This can be a follow up later. Since it doesn't change any public interfaces, we can also do this as an upgrade.

damirka: Let's discuss it during out call, and I'll add an issue to linear.

PR #1351, opened 2024-12-30
feat: add utility function to calculate rewards with dev_inspect

karlwuest:
## Description

Adds a utility function to calculate rewards with the intention of being used with `dev_inspect`. Also includes some minor refactoring and cleanup.

Closes WAL-418

## Test plan

- existing tests cover the reward calculation

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking.move
+++ b/contracts/walrus/sources/staking.move
@@ -302,7 +302,7 @@ public fun epoch(staking: &Staking): u32 {
 // === Utility functions ===
 
 /// Calculate the rewards for an amount with value `staked_principal`, staked in the pool with
-/// the given `node_id` between `activation_epoch` and `last_epoch`.
+/// the given `node_id` between `activation_epoch` and `withdraw_epoch`.
 ///
 /// This function can be used with `dev_inspect` to calculate the expected rewards for a `StakedWal`
 /// object or, more generally, the returns provided by a given node over a given period.
@@ -311,9 +311,9 @@ public fun calculate_rewards(
     node_id: ID,
     staked_principal: u64,
     activation_epoch: u32,
-    last_epoch: u32,
+    withdraw_epoch: u32,
 ): u64 {
-    staking.inner().calculate_rewards(node_id, staked_principal, activation_epoch, last_epoch)
+    staking.inner().calculate_rewards(node_id, staked_principal, activation_epoch, withdraw_epoch)
 }
 
 // === Upgrade ===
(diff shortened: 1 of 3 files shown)


contracts/walrus/sources/staking.move
```diff
@@ -299,6 +299,23 @@ public fun epoch(staking: &Staking): u32 {
...
 }
 
+// === Utility functions ===
+
+/// Calculate the rewards for an amount with value `staked_principal`, staked in the pool with
+/// the given `node_id` between `activation_epoch` and `last_epoch`.
+///
+/// This function can be used with `dev_inspect` to calculate the expected rewards for a `StakedWal`
+/// object or, more generally, the returns provided by a given node over a given period.
+public fun calculate_rewards(
```
damirka, line 309: Are you sure you want to expose this function as public? If the goal is to run dev inspect, then it's enough to add it as a private function (with an `#[allow(unused_function)]` attribute. It would make things tad easier.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -703,6 +698,25 @@ public(package) fun has_pool(self: &StakingInnerV1, node_id: ID): bool {
...
+public(package) fun calculate_rewards(
+    self: &StakingInnerV1,
+    node_id: ID,
+    staked_principal: u64,
+    activation_epoch: u32,
+    last_epoch: u32,
+): u64 {
+    if (!self.pools.contains(node_id)) {
+        abort EPoolNotFound
+    };
```
damirka:
```suggestion
    assert!(self.pools.contains(node_id), EPoolNotFound);
```

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -703,6 +698,25 @@ public(package) fun has_pool(self: &StakingInnerV1, node_id: ID): bool {
...
+
+    let pool = &self.pools[node_id];
+    pool.calculate_rewards(staked_principal, activation_epoch, last_epoch)
```
damirka:
```suggestion
    self.pools[node_id].calculate_rewards(staked_principal, activation_epoch, last_epoch)
```

contracts/walrus/sources/staking.move
karlwuest: Just commenting here for completeness since we discussed this on slack: I'll leave this public because it should be considered part of the external API (the functionality is needed by partners).

PR #1353, opened 2025-01-03
fix: use a specific msim version to work around rust 1.83 non-determinism in MacOs

halfprice:
## Description

We upgraded to rust 1.83 where MacOs calls a new native function to get random bytes that are no longer determinism. @mystenmark fixed it in the msim repo. Before Sui 1.41 is released, we need to use a custom version of msim to get determinism when running simtest in MacOs.

Linux and CI are not affected.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -125,3 +125,8 @@ inherits = "test"
 overflow-checks = true
 # opt-level 1 gives >5x speedup for simulator tests without slowing down build times very much.
 opt-level = 1
+
+# TODO: Remove this once Sui 1.41 is released.
+[patch."https://github.com/MystenLabs/mysten-sim.git"]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -125,3 +125,8 @@ inherits = "test"
 overflow-checks = true
 # opt-level 1 gives >5x speedup for simulator tests without slowing down build times very much.
 opt-level = 1
+
+# TODO: Remove this once Sui 1.41 is released.
+[patch."https://github.com/MystenLabs/mysten-sim.git"]
+msim = { git = "https://www.github.com/MystenLabs/mysten-sim.git", rev = "2a170f4cd81c5cd10f5e4a5e810068f3045f41b6" }
```
halfprice, line 131: the `www` in the URL is required because otherwise the patch does not allow pointing to the same repo.

PR #1356, opened 2025-01-05
chore: version structs used in metadata and event cursor table

halfprice:
## Description

Close WAL-496, WAL-497

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/encoding/blob_encoding.rs
+++ b/crates/walrus-core/src/encoding/blob_encoding.rs
@@ -632,7 +632,7 @@ mod tests {
 
     use super::*;
     #[cfg(feature = "walrus-mainnet")]
-    use crate::metadata::BlobMetadataApi;
+    use crate::metadata::BlobMetadataApi as _;
     use crate::metadata::UnverifiedBlobMetadataWithId;
 
     param_test! {
--- a/crates/walrus-core/src/metadata.rs
+++ b/crates/walrus-core/src/metadata.rs
@@ -195,7 +195,8 @@ pub struct BlobMetadata {
 
 #[cfg(not(feature = "walrus-mainnet"))]
 impl BlobMetadata {
-    /// Creates a new [`BlobMetadata`] with the given encoding type, unencoded length, and sliver.
+    /// Creates a new [`BlobMetadata`] with the given encoding type, unencoded length, and sliver
+    /// hashes.
     pub fn new(
         encoding_type: EncodingType,
         unencoded_length: u64,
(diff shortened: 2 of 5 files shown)


crates/walrus-core/src/metadata.rs
```diff
@@ -183,6 +181,7 @@ impl<const V: bool> AsRef<BlobMetadata> for BlobMetadataWithId<V> {
     }
 }
 
+#[cfg(not(feature = "walrus-mainnet"))]
 /// Metadata about a blob, without its corresponding [`BlobId`].
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct BlobMetadata {
```
mlegner, line 187: Question/suggestion: Instead of keeping this old definition, could we not have the `BlobMetadataV1` for both Mainnet and Testnet and have a type alias like the following?
```rust
#[cfg(feature = "walrus-mainnet")]
pub type BlobMetadata = BlobMetadataV1
```

This would reduce some of the duplicated code here.

crates/walrus-core/src/metadata.rs
```diff
@@ -239,6 +252,169 @@ impl BlobMetadata {
...
+#[enum_dispatch(BlobMetadataApi)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum BlobMetadata {
+    /// Version 1 of the blob metadata.
+    V1(BlobMetadataV1),
+}
+
+#[cfg(feature = "walrus-mainnet")]
+impl BlobMetadata {
+    /// Creates a new [`BlobMetadata`] with the given encoding type, unencoded length, and sliver
```
mlegner, line 325: Nit:
```suggestion
    /// Creates a new [`BlobMetadata`] with the given encoding type, unencoded length, and sliver hashes.
```

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -631,6 +631,8 @@ mod tests {
     use walrus_test_utils::{param_test, random_data, random_subset};
 
     use super::*;
+    #[cfg(feature = "walrus-mainnet")]
+    use crate::metadata::BlobMetadataApi;
```
mlegner: Nit (here and in other places): The `BlobMetadataApi` could be implemented on `BlobMetadata` independent of the feature. Then this could be replaced by
```suggestion
    use crate::metadata::BlobMetadataApi as _;
```

crates/walrus-core/src/metadata.rs
halfprice: Thanks a lot @mlegner for the review! I thought about this right after I send this PR for review, but was lazy to get to it... Since my other PR remove all of this, I'm going to save the hassle and just keep it as is.

PR #1361, opened 2025-01-07
feat: add compute next committee function

georgedigkas:
implement compute_next_committee fn

## Description

Add function to compute next committee before the next-epoch parameter selection.

## Test plan

Sui Move tests added

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/staking_inner.move
+++ b/contracts/walrus/sources/staking/staking_inner.move
@@ -493,13 +493,7 @@ public(package) fun compute_next_committee(self: &StakingInnerV1): Committee {
 public(package) fun select_committee(self: &mut StakingInnerV1) {
     assert!(self.next_committee.is_none(), ECommitteeSelected);
 
-    let (active_ids, shards) = self.apportionment();
-    let distribution = vec_map::from_keys_values(active_ids, shards);
-
-    // if we're dealing with the first epoch, we need to assign the shards to the
-    // nodes in a sequential manner. Assuming there's at least 1 node in the set.
-    let committee = if (self.committee.size() == 0) committee::initialize(distribution)
-    else self.committee.transition(distribution);
+    let committee = self.compute_next_committee();
 
     // inherently sorted by node ID
     let public_keys = vec_map::from_keys_values(


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -478,17 +478,28 @@ public(package) fun withdraw_stake(
...
+    // nodes in a sequential manner. Assuming there is at least 1 node in the set.
+    if (self.committee.size() == 0) committee::initialize(distribution)
+    else self.committee.transition(distribution)
+}
+
+/// Compute the next committee. This function can be called to preview the
+/// next committee without actually performing an epoch change.
+public fun compute_next_committee(self: &StakingInnerV1): Committee {
+    calculate_committee(self)
+}
```
karlwuest: Suggestion: I don't think we need a wrapper function here and can just change it to the following.

```suggestion
/// Computes the committee for the next epoch.
public(package) fun compute_next_committee(self: &StakingInnerV1): Committee {
    let (active_ids, shards) = self.apportionment();
    let distribution = vec_map::from_keys_values(active_ids, shards);

    // if we are dealing with the first epoch, we need to assign the shards to the
    // nodes in a sequential manner. Assuming there is at least 1 node in the set.
    if (self.committee.size() == 0) committee::initialize(distribution)
    else self.committee.transition(distribution)
}
```

Important: The function should not be `public` and instead a `public` function that calls this function should be added in `staking`. Since #1363 (just merged today), the inner object is no longer a dynamic object field, so to use the function with devinspect, it must be on `Staking`.

contracts/walrus/tests/staking/staking_inner_compute_tests.move
```diff
@@ -0,0 +1,138 @@
...
+    let clock = clock::create_for_testing(ctx);
+    let mut staking = staking_inner::new(0, EPOCH_DURATION, 6, &clock, ctx);
+
+    let pool_one = test::pool().register(&mut staking, ctx);
+    let pool_two = test::pool().register(&mut staking, ctx);
+    let pool_three = test::pool().register(&mut staking, ctx);
+
+    let wal_alice = staking.stake_with_pool(test::mint(1000, ctx), pool_one, ctx);
+    let wal_bob = staking.stake_with_pool(test::mint(1000, ctx), pool_two, ctx);
+    let wal_karl = staking.stake_with_pool(test::mint(1000, ctx), pool_three, ctx);
```
karlwuest: > wal_karl

sneaky 😄

contracts/walrus/sources/staking/staking_inner.move
georgedigkas: > Suggestion: I don't think we need a wrapper function here and can just change it to the following.

I have extracted the shared logic into the `calculate_committee` function. This avoids code duplication, as this same calculation is performed in both `compute_next_committee` and `select_committee`. This makes the code easier to maintain and ensures consistency across those functions.

karlwuest: But the current `compute_next_committee` just forwards the call and doesn't do anything else. 
I agree with you if the function would be public (as currently), but the public function should be moved to `staking` and then I think we can get rid of the wrapper function here.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -489,17 +489,17 @@ fun calculate_committee(self: &StakingInnerV1): Committee {
...
     assert!(self.next_committee.is_none(), ECommitteeSelected);
 
-    let committee = calculate_committee(self);
+    let (active_ids, shards) = self.apportionment();
+    let distribution = vec_map::from_keys_values(active_ids, shards);
+
+    // if we're dealing with the first epoch, we need to assign the shards to the
+    // nodes in a sequential manner. Assuming there's at least 1 node in the set.
+    let committee = if (self.committee.size() == 0) committee::initialize(distribution)
+    else self.committee.transition(distribution);
```
karlwuest:
```suggestion
    let committee = self.compute_next_committee();
```

PR #1362, opened 2025-01-07
chore: version the data struct used in slivers column families

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage/shard.rs
+++ b/crates/walrus-service/src/node/storage/shard.rs
@@ -783,7 +783,7 @@ impl ShardStorage {
                     #[cfg(feature = "walrus-mainnet")]
                     batch.insert_batch(
                         &self.primary_slivers,
-                        [(blob_id, &PrimarySliverData::V1(primary.clone()))],
+                        [(blob_id, &PrimarySliverData::from(primary.clone()))],
                     )?;
                 }
                 Sliver::Secondary(secondary) => {
@@ -793,7 +793,7 @@ impl ShardStorage {
                     #[cfg(feature = "walrus-mainnet")]
                     batch.insert_batch(
                         &self.secondary_slivers,
-                        [(blob_id, &SecondarySliverData::V1(secondary.clone()))],
+                        [(blob_id, &SecondarySliverData::from(secondary.clone()))],
                     )?;
                 }
             }


crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -220,8 +272,24 @@ impl ShardStorage {
...
         match sliver {
-            Sliver::Primary(primary) => self.primary_slivers.insert(blob_id, primary),
-            Sliver::Secondary(secondary) => self.secondary_slivers.insert(blob_id, secondary),
+            Sliver::Primary(primary) => {
+                #[cfg(not(feature = "walrus-mainnet"))]
+                let result = self.primary_slivers.insert(blob_id, primary);
+                #[cfg(feature = "walrus-mainnet")]
+                let result = self
+                    .primary_slivers
+                    .insert(blob_id, &PrimarySliverData::from(primary.clone()));
```
halfprice, line 281: The new version currently is only used in this file for simplicity. Once mainnet version is used, we can use the new struct from the source to avoid a copy here.

sadhansood, line 281: We don't need a clone here i think.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -220,8 +272,24 @@ impl ShardStorage {
...
+                result
+            }
+            Sliver::Secondary(secondary) => {
+                #[cfg(not(feature = "walrus-mainnet"))]
+                let result = self.secondary_slivers.insert(blob_id, secondary);
+                #[cfg(feature = "walrus-mainnet")]
+                let result = self
+                    .secondary_slivers
+                    .insert(blob_id, &SecondarySliverData::from(secondary.clone()));
```
sadhansood, line 290: same here

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -696,11 +778,23 @@ impl ShardStorage {
             match sliver {
                 Sliver::Primary(primary) => {
                     assert_eq!(sliver_type, SliverType::Primary);
+                    #[cfg(not(feature = "walrus-mainnet"))]
                     batch.insert_batch(&self.primary_slivers, [(blob_id, primary)])?;
+                    #[cfg(feature = "walrus-mainnet")]
+                    batch.insert_batch(
+                        &self.primary_slivers,
+                        [(blob_id, &PrimarySliverData::V1(primary.clone()))],
```
sadhansood: Should we standardize on `PrimarySliverData::from` everywhere instead of `PrimarySliverData::V1`

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -696,11 +778,23 @@ impl ShardStorage {
...
+                    )?;
                 }
                 Sliver::Secondary(secondary) => {
                     assert_eq!(sliver_type, SliverType::Secondary);
+                    #[cfg(not(feature = "walrus-mainnet"))]
                     batch.insert_batch(&self.secondary_slivers, [(blob_id, secondary)])?;
+                    #[cfg(feature = "walrus-mainnet")]
+                    batch.insert_batch(
+                        &self.secondary_slivers,
+                        [(blob_id, &SecondarySliverData::V1(secondary.clone()))],
```
sadhansood: same here `SecondarySliverData::from` instead of `SecondarySliverData::V1`

crates/walrus-service/src/node/storage/shard.rs
halfprice: Good catch. Done.

crates/walrus-service/src/node/storage/shard.rs

crates/walrus-service/src/node/storage/shard.rs
halfprice: `primary` is a reference, so either here or PrimarySliverData::from needs to clone the data. I'm not sure how to work around it.

crates/walrus-service/src/node/storage/shard.rs
halfprice: please see above

PR #1363, opened 2025-01-07
feat: use dynamice field for inner staking/system objects

karlwuest:
## Description

Change from dynamic object fields to dynamic fields for the inner objects.
Damir (currently on PTO) and I had discussed this a while ago, because there is no benefit to having dynamic object fields here. As a nice side effect, this saves one RPC request any time we request the system or staking object.

## Test plan

Covered by existing tests

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/read_client.rs
+++ b/crates/walrus-sui/src/client/read_client.rs
@@ -572,7 +572,7 @@ impl SuiReadClient {
         }
         let inner = self
             .sui_client
-            .get_dynamic_field_object::<u64, SystemStateInnerV1>(
+            .get_dynamic_field::<u64, SystemStateInnerV1>(
                 self.system_object_id,
                 TypeTag::U64,
                 version,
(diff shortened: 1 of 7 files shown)


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -330,6 +330,7 @@ impl RetriableSuiClient {
         .await
     }
 
+    #[allow(unused)]
```
giac-mysten, line 333: Question: will we need this in the future?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -622,11 +622,7 @@ impl SuiReadClient {
         }
         let inner = self
             .sui_client
-            .get_dynamic_field_object::<u64, StakingInnerV1>(
-                self.staking_object_id,
-                TypeTag::U64,
-                version,
-            )
+            .get_dynamic_field::<u64, StakingInnerV1>(self.staking_object_id, TypeTag::U64, version)
```
giac-mysten, line 625: Nice that this uses `derive_dynamic_field_id` internally!

crates/walrus-sui/src/client/retry_client.rs
karlwuest: I'm not sure. I just kept it because figuring out how to do this correctly was a bit of a pain (and now it's there if we need it) 😅 But I can also remove it.

PR #1370, opened 2025-01-08
feat: Limit the number of shards per node

benr-ml:
## Description

Limit the number of shards per node when >=20 are used.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/staking_inner.move
+++ b/contracts/walrus/sources/staking/staking_inner.move
@@ -41,7 +41,7 @@ const TEMP_ACTIVE_SET_SIZE_LIMIT: u16 = 100;
 const MIN_NODES_FOR_SHARDS_LIMIT: u8 = 20;
 
 /// The maximum number of shards per node as a denominator of the total number of shards.
-/// When the number of nodes is smaller than MIN_NODES_FOR_SHARDS_LIMIT, the shards limit 
+/// When the number of nodes is smaller than MIN_NODES_FOR_SHARDS_LIMIT, the shards limit
 /// is multiplied by MIN_NODES_FOR_SHARDS_LIMIT / number of nodes.
 const SHARDS_LIMIT_DENOMINATOR: u8 = 10; // 10%
 


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -37,6 +37,12 @@ const MIN_STAKE: u64 = 0;
 /// TODO: Remove once solutions are in place to prevent hitting move execution limits (#935).
 const TEMP_ACTIVE_SET_SIZE_LIMIT: u16 = 100;
 
+/// The minimum number of nodes for the shards limit to be applied.
+const MIN_NODES_FOR_SHARDS_LIMIT: u8 = 20;
+
+/// The maximum number of shards per node as a denominator of the total number of shards.
```
karlwuest, line 43: Suggestion: Can you add a comment here that this must be less than `MIN_NODES_FOR_SHARDS_LIMIT` to ensure that all shards can be assigned?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -549,8 +561,10 @@ fun dhondt(
         // quotient.
         let (_quotient, tie_breaker, index) = quotients.pop_max();
```
karlwuest, line 564: Suggestion: Add a comment here, why it's safe to call `pop_max` without causing an abort. Previously, this was the case because the queue always had the same length. Now this should still hold because `SHARDS_LIMIT_DENOMINATOR <= MIN_NODES_FOR_SHARDS_LIMIT`.

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -526,19 +533,25 @@ fun dhondt(
     let n_shards = n_shards as u64;
     assert!(total_stake > 0, ENoStake);
 
+    // Limit the number of shards per node if there are enough nodes.
+    let max_shards = if (n_nodes >= (MIN_NODES_FOR_SHARDS_LIMIT as u64))
+        n_shards / (SHARDS_LIMIT_DENOMINATOR as u64) else n_shards;
+
```
karlwuest: We could do something like the following to prevent the issue that @mlegner mentioned:

```suggestion
    let max_shards = if (n_nodes >= (MIN_NODES_FOR_SHARDS_LIMIT as u64)) {
        n_shards / (SHARDS_LIMIT_DENOMINATOR as u64) 
    } else {
        n_shards * MIN_NODES_FOR_SHARDS_LIMIT / (n_nodes * SHARDS_LIMIT_DENOMINATOR)
    };

```

benr-ml: This doesn't work for equally distributed stakes like vector[25000, 25000, 25000, 25000] (breaks the invariant of `quotients`).

karlwuest: That's true, I guess we would need to round up? Since this hurts readability even more, moving this into a separate `max_shards` function may be better. But, as mentioned, while I do share the preference of @mlegner to have a smooth behaviour, I don't feel very strongly about this. I'm also fine with making the assumption that we will always have >= 20 nodes in practice (unless in a test setting), but then we should document this as well.

Side note: The invariant may actually also break in the first case if `n_shards` is not cleanly divisible by `SHARDS_LIMIT_DENOMINATOR` and `SHARDS_LIMIT_DENOMINATOR = MIN_NODES_FOR_SHARDS_LIMIT`, so we should reflect that in the invariant.

mlegner: I'd still prefer a smoother behavior at the `MIN_NODES_FOR_SHARDS_LIMIT` boundary, having the logic in a separate function sounds good to me. I'm not dead-set on this though as I agree that it will most likely not be relevant in practice.

benr-ml: @mlegner please let me know the concrete calculation you prefer here and I'll add it.

mlegner: @benr-ml I'm probably missing something, but I don't understand yet why the version suggested by @karlwuest doesn't work (at least for the given constants). However, rounding up should make this work in all cases. So how about a function like the following?

```
fun max_shards_per_node(n_nodes: u64, n_shards: u64): u64 {
    if (n_nodes >= (MIN_NODES_FOR_SHARDS_LIMIT as u64)) {
        n_shards / (SHARDS_LIMIT_DENOMINATOR as u64)
    } else {
        n_shards * (MIN_NODES_FOR_SHARDS_LIMIT as u64).divide_and_round_up(
            n_nodes * (SHARDS_LIMIT_DENOMINATOR as u64)
        )
    }
}
```

If we have 4 nodes, the given constants should result in a maximum of 20000/40 = 500 (with or without rounding up).

benr-ml: Just making sure I'm using the approach you prefer before I fix effected tests. Note that even the simple code above has a bug :)

mlegner: Ah, whoops, didn't set the parentheses correctly... 🤦

PR #1371, opened 2025-01-08
feat: use separate package IDs in the SuiReadClient

karlwuest:
## Description

- Automatically gets the correct `wal::WAL` type from the walrus package
- Uses the exchange package ID instead of the walrus package ID for calls to the exchange
- Minor refactoring to group all package/object ID arguments for the constructors of the clients

closes WAL-425

## Test plan

Existing tests

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -268,6 +271,23 @@ impl RetriableSuiClient {
...
+    /// Calls [`sui_sdk::apis::ReadApi::get_normalized_move_modules_by_package`] internally.
+    #[tracing::instrument(level = Level::DEBUG, skip_all)]
+    pub async fn get_normalized_move_modules_by_package(
+        &self,
+        package_id: ObjectID,
+    ) -> SuiRpcResult<BTreeMap<String, SuiMoveNormalizedModule>> {
+        retry_rpc_errors(self.get_strategy(), || async {
+            self.sui_client
+                .read_api()
+                .get_normalized_move_modules_by_package(package_id)
```
giac-mysten, line 285: Question: What is a "normalized" move module? Just the additional metadata about struct, functions, etc.?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -427,6 +467,56 @@ impl RetriableSuiClient {
             .collect())
     }
 
+    /// Retrieves the WAL type from the walrus package by getting the type tag of the `Balance`
+    /// in the `StakedWal` Move struct.
+    #[tracing::instrument(err, skip(self))]
+    pub(crate) async fn wal_type_from_package(
```
giac-mysten, line 460: very cool! did not know this kind of introspection was possible.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -234,50 +240,50 @@ impl SuiReadClient {
...
-        // TODO(WAL-474): remove this when removing the `walrus-mainnet` feature flag.
-        package_id: Option<ObjectID>,
+        contract_config: &ContractConfig,
     ) -> SuiClientResult<Self> {
         #[cfg(not(feature = "walrus-mainnet"))]
-        let walrus_package_id = package_id.unwrap_or(
+        let walrus_package_id = contract_config.walrus_package.unwrap_or(
             // We use `unwrap_or` here because we want to call the function even if the package ID
             // is provided to check if the system and staking objects exist.
             sui_client
```
giac-mysten: question: remind me why we are still using `contract_config.walrus_package`, even if we are also checking `get_system_package_id_from_system_object` for both `mainnet` and `!mainnet`?

karlwuest: `get_system_package_id_from_system_object` for `!mainnet` just gets the package ID from the type definition of `System`, which will always be the initial package version, so we need this to be compatible with upgrades in the current testnet.
(However, we're anyway now removing the feature flag, so that is redundant soon)

crates/walrus-sui/src/client/retry_client.rs
karlwuest: I'm not sure what exactly it does, but my understanding is that it structures the module data, e.g. by including (human readable) type definitions. It may do more than that, but that's the part that we need.

PR #1372, opened 2025-01-08
feat(cli): add command to share blob, fund/extend shared blob

joyqvq:
## Description

Added three new commands 1) share a blob 2) fund a shared blob 3) extend a shared blob 

## Test plan

- added simtest 
- tested on testnet
new shared blob: https://suiscan.xyz/testnet/tx/7noAcfNTwQioBekTi9DScL8RtkX5Hdsp7ES4Hmsp41mf 
new funded shared blob (in devnet): https://suiscan.xyz/devnet/tx/C4pFuT3xW5PPj6d88ijyUPUrBY1LSWyWSCVaMA6Vphb4
fund: https://suiscan.xyz/testnet/tx/4SCm8x3ep6ac2DoqB2KoneVs9GPWAc12nY39E7Fnarzs
extend: https://testnet.suivision.xyz/txblock/D7KmHuQtVLCumQ3yGa6VWvpu2ExgTjte4r7QdxUu9UDv

cli output:
```
target/release/walrus share --blob-obj-id 0x4f9fb9792b7a4dd19a7cd124bfc741e7a8c446378a20e80c94bd8adccd347c1a                                                                                                                                                                                 

2025-01-08T20:05:42.428841Z  INFO walrus: client version: 1.9.0-388169f6343a
2025-01-08T20:05:42.429002Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2025-01-08T20:05:42.429594Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/.sui/sui_config/client.yaml'
 ◉ done [00:00:09]                                                                                                                                                                                                                                                                                                                                                                         Success: The blob has been shared with id: 0xc98c65209441d1cdc46638063f165265360d364869bf1ed31fa95e9b02b27ddc

target/release/walrus fund --shared-blob-obj-id 0xc98c65209441d1cdc46638063f165265360d364869bf1ed31fa95e9b02b27ddc --amount 100

2025-01-08T20:06:54.468541Z  INFO walrus: client version: 1.9.0-388169f6343a
2025-01-08T20:06:54.468600Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2025-01-08T20:06:54.468743Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/.sui/sui_config/client.yaml'
 ◉ done [00:00:01]                                                                                                                                                                                                                                                                                                                                                                         Success: The blob has been funded with 100 FROST

target/release/walrus extend --shared-blob-obj-id 0xc98c65209441d1cdc46638063f165265360d364869bf1ed31fa95e9b02b27ddc --epochs-ahead 1

2025-01-08T20:13:42.337980Z  INFO walrus: client version: 1.9.0-388169f6343a
2025-01-08T20:13:42.338055Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2025-01-08T20:13:42.338216Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/.sui/sui_config/client.yaml'
 ◉ done [00:00:03]                                                                                                                                                                                                                                                                                                                                                                         Success: The blob has been extended by 1 epochs

# using walrus testbed + sui devnet 
target/release/walrus share --blob-obj-id 0x8cc3d56d7977a0d24f6f4b5e62ad4b1d7eb142eadd702c83f636c5586205f85a --amount 1                                                                                                                                                                       ✔  10067  10:42:22
2025-01-10T15:42:35.111804Z  INFO walrus: client version: 1.9.0-742345ef3cfe-dirty
2025-01-10T15:42:35.111951Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2025-01-10T15:42:35.112506Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/mysten/walrus/working_dir/sui_client.yaml'
 ◉ done [00:00:02]                                                                                                                                                                                                                                                                                                                                                                         Success: The blob has been shared, object id: 0x5da9547c2a2eefcb35afa1ca51d2af04969d8d79110f8ee782733f475333713b
```
---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Added share blob, fund shared blob, extend shared blob commands.

### DIFF
--- a/contracts/subsidies/Move.lock
+++ b/contracts/subsidies/Move.lock
@@ -50,6 +50,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.39.2"
+compiler-version = "1.40.1"
 edition = "2024.beta"
 flavor = "sui"
--- a/contracts/wal/Move.lock
+++ b/contracts/wal/Move.lock
@@ -21,6 +21,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.39.2"
+compiler-version = "1.40.1"
 edition = "2024.beta"
 flavor = "sui"
--- a/contracts/wal_exchange/Move.lock
+++ b/contracts/wal_exchange/Move.lock
@@ -30,6 +30,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.39.2"
+compiler-version = "1.40.1"
 edition = "2024.beta"
 flavor = "sui"
--- a/contracts/walrus/Move.lock
+++ b/contracts/walrus/Move.lock
@@ -40,6 +40,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.39.2"
+compiler-version = "1.40.1"
 edition = "2024.beta"
 flavor = "sui"


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -360,6 +360,31 @@ pub enum CliCommands {
...
+    },
+    /// Extend a shared blob.
+    Extend {
+        /// The object ID of the shared blob to extend.
+        #[clap(long)]
+        shared_blob_obj_id: ObjectID,
+        /// The number of epochs to extend the shared blob for.
+        #[clap(long)]
+        epochs_ahead: EpochCount,
+    },
```
mlegner, line 390: Remark: We will also want to offer an `extend` command for owned blobs in the future, which will have slightly different semantics (requires WAL). I think we can use the same `extend` command for both and just have both `shared_blob_obj_id` and `blob_obj_id` options (exactly one of which must be set, similar to the `FileOrBlobIdOrObjectId` construct).

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -360,6 +360,31 @@ pub enum CliCommands {
         #[serde(default)]
         yes: bool,
     },
+
+    /// Fund a shared blob.
+    Fund {
```
mlegner: Suggestion: Call this something like `FundSharedBlob`, otherwise it's a bit unclear what is funded.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -360,6 +360,31 @@ pub enum CliCommands {
...
+        /// The number of epochs to extend the shared blob for.
+        #[clap(long)]
+        epochs_ahead: EpochCount,
+    },
+    /// Share a blob.
+    Share {
+        /// The object ID of the (owned) blob to share.
+        #[clap(long)]
+        blob_obj_id: ObjectID,
+    },
```
mlegner: Suggestion: Add an option to directly fund the shared blob in the same transaction.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -216,6 +216,78 @@ impl ClientCommandRunner {
...
+                    .fund_shared_blob(shared_blob_obj_id, amount)
+                    .await
+                    .map_err(|e| anyhow::anyhow!(e))?;
+
+                spinner.finish_with_message("done");
+                println!(
+                    "{} The blob has been funded with {} FROST",
+                    success(),
+                    amount
+                );
```
mlegner: Two suggestions:
1. Only print this `if !self.json` (also applies to similar cases below).
2. Instead of `{} FROST`, you could use the `HumanReadableFrost` (some examples in `crates/walrus-service/src/client/cli/cli_output.rs`).

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -360,6 +360,31 @@ pub enum CliCommands {
...
+        amount: u64,
+    },
+    /// Extend a shared blob.
+    Extend {
+        /// The object ID of the shared blob to extend.
+        #[clap(long)]
+        shared_blob_obj_id: ObjectID,
+        /// The number of epochs to extend the shared blob for.
+        #[clap(long)]
+        epochs_ahead: EpochCount,
```
mlegner, line 389: Nit: Something like `additional_epochs` would seem clearer to me. For `epochs_ahead` it's unclear if they are ahead of the current epoch or the expiration epoch.

Edit: I just realized that I used that parameter myself in the contracts and we have the same in the `extend_blob` contract... 🤦🤡 I guess we can leave it as `epochs_ahead`, unless we want to change it everywhere (cc @karlwuest).

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -216,6 +216,78 @@ impl ClientCommandRunner {
...
+                    .config?
+                    .new_contract_client(self.wallet?, self.gas_budget)
+                    .await?;
+                let spinner = styled_spinner();
+                spinner.set_message("funding blob...");
+
+                sui_client
+                    .fund_shared_blob(shared_blob_obj_id, amount)
+                    .await
+                    .map_err(|e| anyhow::anyhow!(e))?;
```
mlegner: Nit: I think this is not actually required (also below):
```suggestion
                    .await?;
```

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -216,6 +216,78 @@ impl ClientCommandRunner {
...
+                    .share_blob(blob_obj_id)
+                    .await
+                    .map_err(|e| anyhow::anyhow!(e))?;
+
+                spinner.finish_with_message("done");
+                println!(
+                    "{} The blob has been shared with id: {}",
+                    success(),
+                    shared_blob_obj_id
+                );
```
mlegner: Suggestion: Here we should have a response that includes the new object ID and also works in JSON mode. You can do something similar as we do for the `StakeOutput` defined/implemented in `crates/walrus-service/src/client/responses.rs` and `crates/walrus-service/src/client/cli/cli_output.rs`.

crates/walrus-service/src/client/cli/args.rs
joyqvq: good idea to add it as an option, but iiuc it cannot be the same transaction since it creates a shared object!

crates/walrus-service/src/client/cli/args.rs
joyqvq: makes sense, noted for the next pr for extend owned blob

crates/walrus-service/src/client/cli/args.rs
mlegner: Ah, good point, you're right of course. Maybe it would make sense to add a `new_funded` function to the contract similar to [what we have in the exchange](https://github.com/MystenLabs/walrus/blob/f4265dd3e79d0e349784740d9b63201fbe11b75d/contracts/wal_exchange/sources/wal_exchange.move#L73)?

joyqvq: yeah that would work! didnt think changing contract is an option, i guess its not too late to do?

also, do we still need `new` without funding it?

mlegner: No, I think this should be fine (cc @karlwuest). ~~You may need to use some `#[cfg(feature = "walrus-mainnet")]` annotations temporarily.~~

karlwuest: Yes, it's still fine to change the contracts (especially if it's just adding a function but not changing a struct).

crates/walrus-service/src/client/cli/args.rs
karlwuest: I think the name came from the initial blob creation and we just used the same one for the extensions. I don't mind changing it for the extension, but for the reservation, we should not rename it to `additional_epochs`.

crates/walrus-service/src/client/cli/args.rs
mlegner: Just FYI: We've removed the `walrus-mainnet` feature in #1380, so you don't have to worry about that.

crates/walrus-service/src/client/cli/args.rs
mlegner: > but for the reservation, we should not rename it to additional_epochs

Yes, fully agree, `additional_epochs` doesn't actually make sense there.

contracts/subsidies/Move.lock
mlegner, line 1: Nit: Can you revert the changes to the `Move.lock` files?

crates/walrus-sui/src/client.rs
```diff
@@ -1068,6 +1068,75 @@ impl SuiContractClient {
...
+    ) -> SuiClientResult<ObjectID> {
+        let blob: Blob = self
+            .read_client
+            .sui_client()
+            .get_sui_object(blob_obj_id)
+            .await?;
+        let wallet = self.wallet().await;
+        let mut pt_builder = self.transaction_builder();
+
+        if let Some(amount) = amount {
```
mlegner, line 1118: Nit: You could add a check that the amount is nonzero.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -365,6 +365,75 @@ impl WalrusPtbBuilder {
         self.mark_arg_as_consumed(&blob_arg);
         Ok(())
     }
+    /// Adds a call to create a new shared blob and fund it.
```
mlegner, line 377: Nit:
```suggestion

    /// Adds a call to create a new shared blob and fund it.
```

crates/walrus-service/src/client/responses.rs
```diff
@@ -493,3 +493,27 @@ pub struct ExchangeOutput {
     /// The amount of SUI exchanged (in MIST).
     pub amount_sui: u64,
 }
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+/// The output of the `walrus share` command.
+pub struct ShareBlobOutput {
```
mlegner, line 500: Suggestion: Optionally add also the funding amount if specified.

PR #1376, opened 2025-01-09
chore: Fix dir paths in event blob writer

sadhansood:
## Description

Prefix the paths in event blob writer as they are currently confusing.

### DIFF
--- a/crates/walrus-service/src/node/events/event_blob_writer.rs
+++ b/crates/walrus-service/src/node/events/event_blob_writer.rs
@@ -183,6 +183,16 @@ pub struct EventBlobWriterFactory {
 }
 
 impl EventBlobWriterFactory {
+    /// Returns the path to the database directory.
+    pub fn db_path(root_dir_path: &Path) -> PathBuf {
+        root_dir_path.join("event_blob_writer").join("db")
+    }
+
+    /// Returns the path to the blobs directory.
+    pub fn blobs_path(root_dir_path: &Path) -> PathBuf {
+        root_dir_path.join("event_blob_writer").join("blobs")
+    }
+
     /// Create a new event blob writer factory.
     pub fn new(
         root_dir_path: &Path,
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -183,23 +183,40 @@ pub struct EventBlobWriterFactory {
...
+            root_dir_path.join("db")
+        } else {
+            root_dir_path.join("event_blob_writer").join("db")
+        }
+    }
+
+    /// Returns the path to the blobs directory.
+    pub fn blobs_path(root_dir_path: &Path) -> PathBuf {
+        if cfg!(not(feature = "walrus-mainnet")) {
+            root_dir_path.join("blob")
```
halfprice: I think it was `blobs` before?

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -183,23 +183,40 @@ pub struct EventBlobWriterFactory {
...
         node: Arc<StorageNodeInner>,
         registry: &Registry,
         num_checkpoints_per_blob: Option<u32>,
     ) -> Result<EventBlobWriterFactory> {
-        let dirs = ["db", "blobs"];
-        for dir in dirs.iter() {
-            fs::create_dir_all(root_dir_path.join(dir))?;
-        }
+        let db_path = Self::db_path(root_dir_path);
+        fs::create_dir_all(db_path.as_path())?;
```
halfprice, line 204: I'm not sure if we need to create db_path, but we probably need to create the blob dir

sadhansood: I think we need to but not 100% certain if rocksdb will create the dir. Since this is how things were before, we can keep the existing behavior.

PR #1378, opened 2025-01-09
perf: Rate limit shard sync

liquid-helium:
## Description

Limit the number of concurrent shard syncs.

Close WAL-522

## Test plan

unit test added

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -105,6 +105,9 @@ jobs:
     needs: diff
     if: ${{ needs.diff.outputs.isRelevantForRustTests == 'true' }}
     runs-on: ubuntu-ghcloud
+    env:
+      RUST_LOG: warn
+      RUST_MIN_STACK: 67108864
     steps:
       - uses: taiki-e/install-action@v2.47.2
         with:


crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -300,7 +302,16 @@ impl ShardSyncHandler {
...
-        // TODO(#705): implement rate limiting for shard syncs.
+        // Panics if error occurs.
+        // The rate limit is enforced by the semaphore, without considering
+        // the priority of the syncs.
+        let permit = self
+            .shard_sync_semaphore
+            .clone()
+            .acquire_owned()
+            .await
+            .expect("Failed to acquire shard sync semaphore");
```
halfprice: suggestion: put this inside let shard_sync_task = tokio::spawn(async move { <here> )) so that you don't need to explicitly drop the permit, which is more robust.

For testing, maybe you can keep track of a counter about active shard sync. This is also a good metric to expose for monitoring.

liquid-helium: That's a good suggestion to expose the number of active sync task. 

Here is the reason I placed the semaphore like this, please help check the logic, since I am still ramping up my Rust.
With the semaphore placed outside of `shard-sync_task` and acquired before we insert a `sync_task` into `shard_sync_in_progress` (`entry.insert(shard_sync_task);` ), we won't need to spawn a task unless we are going to run it, instead of potentially blocking on the semaphore inside `shard_sync_task`. Therefore the size of `shard_sync_in_progress` is the number of active syncs. Per your suggestion, we can expose this metric.

halfprice: Yes, the code works as you expect. The intention of the rate limiting here is more about control shard sync traffic as opposite to tokio thread (but it is still good if we can limit it).

So the main thing I was suggesting is that we probably don't want to call `drop(permit)` explicitly, which may cause confusing if later we want to exit the function in the middle.

Another cleaner way to do this is probably:

```
let permit = self
            .shard_sync_semaphore
            .clone()
            .acquire_owned()
            .await
            .expect("Failed to acquire shard sync semaphore");
...
tokio::spawn(async move {
  let _permit = permit;  // here, the ownership of the permit is moved into this function. And thread exit automatically drop and release the permit;

  <rest code>
})

```

In this case, when the async function returns, no matter at which point it returns, the permit is guaranteed to be released.

liquid-helium: I see your point now, that's a nice catch, thanks for the suggestion!

After giving a second thought, I feel moving the semaphore inside the `sync_task` is more beneficial, in terms of the backoff retry, i.e., we don't want to hold it when waiting for the next retry.

crates/walrus-service/src/node.rs
```diff
@@ -2797,6 +2803,7 @@ mod tests {
             TestCluster::<StorageNodeHandle>::builder()
                 .with_shard_assignment(assignment)
                 .with_system_event_providers(events.clone())
+                .with_shard_sync_config(shard_sync_config.unwrap_or_default())
```
halfprice: Suggestion: typically, we don't pass in default, but let the builder to control what should be default config.

So you can just do something like

builder = TestCluster::<StorageNodeHandle>::builder()
                .with_shard_assignment(assignment)
                .with_system_event_providers(events.clone());

if let Some(shard_sync_config) = shard_sync_config {
    builder = builder.with_shard_sync_config(shard_sync_config);
}

builder.build().await?;

crates/walrus-service/src/node.rs
```diff
@@ -3823,6 +3854,29 @@ mod tests {
         Ok(())
     }
 
+    async fn wait_for_shards_in_active_state(shard_storage_set: &ShardStorageSet) -> TestResult {
+        // Waits for the shard to be synced.
+        tokio::time::timeout(Duration::from_secs(15), async {
+            loop {
+                let mut all_active = true;
+                for shard_storage in &shard_storage_set.shard_storage {
+                    let status = shard_storage.status().unwrap();
```
halfprice: Nit: the current convention in the code base is to use .expect("<reason>") instead of .unwrap(). (although there may be existing instances of unwrap() in the codebase still)

crates/walrus-service/src/node.rs
```diff
@@ -3833,9 +3887,26 @@ mod tests {
     async fn sync_shard_complete_transfer(
         wipe_metadata_before_transfer_in_dst: bool,
     ) -> TestResult {
-        let (cluster, blob_details, storage_dst, shard_storage_dst) =
-            setup_cluster_for_shard_sync_tests().await?;
+        let assignment: Option<&[&[u16]]> = Some(&[&[0, 1, 2], &[3]]);
+        let shard_sync_config: ShardSyncConfig = ShardSyncConfig {
+            shard_sync_concurrency: if let Some(assignment) = assignment {
```
halfprice: looks like the assignment is always set? So you may not need the `if let Some()` (here and below).

crates/walrus-service/src/test_utils.rs
```diff
@@ -561,6 +562,12 @@ impl StorageNodeHandleBuilder {
         self
     }
 
+    /// Sets the shard sync config for the node.
+    pub fn with_shard_sync_config(mut self, shard_sync_config: ShardSyncConfig) -> Self {
```
halfprice, line 566: Similar to all the other options, let's pass in `Option<ShardSyncConfig>`

crates/walrus-service/src/test_utils.rs
```diff
@@ -1464,6 +1474,12 @@ impl TestClusterBuilder {
         &self.storage_node_configs
     }
 
+    /// Sets the shard sync config for the cluster.
+    pub fn with_shard_sync_config(mut self, shard_sync_config: ShardSyncConfig) -> Self {
```
halfprice, line 1478: ditto

crates/walrus-service/src/node/config.rs
```diff
@@ -388,6 +390,7 @@ impl Default for ShardSyncConfig {
             max_concurrent_blob_recovery_during_shard_recovery: 5,
             blob_certified_check_interval: Duration::from_secs(60),
             max_concurrent_metadata_fetch: 10,
+            shard_sync_concurrency: 100,
```
halfprice: 100 maybe too large. maybe use 5.

crates/walrus-service/src/node.rs
liquid-helium: Thanks for the suggestion, change applied.

crates/walrus-service/src/node.rs
liquid-helium: Done.

crates/walrus-service/src/node.rs
liquid-helium: Done.

crates/walrus-service/src/test_utils.rs
liquid-helium: No strong preference, but since when this is called, the caller is explicitly setting the shard_sync_config. keep as for now unless you have a strong opinion

crates/walrus-service/src/node/config.rs
liquid-helium: The local tests become much slower with 5, do we want to keep it high until we know a small concurrency help in prod?

PR #1379, opened 2025-01-10
fix: storage lag check should only be performed on being processed events

halfprice:
## Description

When blob writer is catching up, the node may writing events that are already processed. These events may be old from the processed event and should not be used to check event lagging.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/node.rs
```diff
@@ -768,33 +789,40 @@ impl StorageNode {
                 "error.type" = field::Empty,
             );
 
-            if let Some(epoch_at_start) = maybe_epoch_at_start {
```
halfprice, line 771: The fix is basically move this code block inside should process

PR #1382, opened 2025-01-10
feat(node)!: add command to export parameters for third-party registration

mlegner:
## Description

This PR modifies the `walrus-node setup` command to optionally export registration parameters to be used by a third party.

In addition, this includes the following changes:

- Don't use the faucet by default.
- Don't overwrite files by default when generating configuration files with the `walrus-node` binary.
- Make parameters `public_host`, `public_port`, and `name` mandatory in the configuration file.
- Add parameters to the `setup` command to set the node metadata.
- Minor refactoring.

Contributes to WAL-153
Contributes to WAL-274
Contributes to WAL-506

## Test plan

Automatic tests and manual testing of the `walrus-node setup` command. For example, run the following and check the files generated in `/tmp/config`:

```
cargo run --bin walrus-node -- setup \
  --config-directory /tmp/config \
  --storage-path /tmp/storage \
  --node-capacity 1G \
  --public-host foobar.com \
  --name "My Node" \
  --description "Some description of the node..." \
  --registering-third-party 0x7cbf9f40ef49df41c9b98946bdd1e5e439759671256ca1d43d97db83b2f17932
```

---

## Release notes

- [x] Storage node: Make parameters `public_host`, `public_port`, and `name` mandatory in the configuration file. Don't request SUI tokens by default when generating a wallet. Modify the `walrus-node setup` command to optionally export registration parameters to be used by a third party.
- [x] CLI: Don't request SUI tokens by default when generating a wallet.

### DIFF
--- a/crates/walrus-sui/tests/test_walrus_sui.rs
+++ b/crates/walrus-sui/tests/test_walrus_sui.rs
@@ -392,7 +392,6 @@ async fn test_set_authorized() -> anyhow::Result<()> {
     let proof_of_possession = utils::generate_proof_of_possession(
         &protocol_key_pair,
         &walrus_client.inner,
-        &registration_params,
         walrus_client.inner.current_epoch().await?,
     );
 


crates/walrus-sui/src/utils.rs
```diff
@@ -407,13 +406,21 @@ pub async fn request_sui_from_faucet(
...
+    bls_sk: &ProtocolKeyPair,
+    sui_address: SuiAddress,
+    current_epoch: Epoch,
+) -> SignedMessage<ProofOfPossessionMsg> {
     bls_sk.sign_message(&ProofOfPossessionMsg::new(
         current_epoch,
-        sui_address,
-        registration_params.public_key.clone(),
+        sui_address.to_inner(),
+        bls_sk.public().clone(),
```
mlegner, line 424: Question @karlwuest: Why did we need/use the registration params instead of the key directly? If this was intentional, I'm happy to revert this.

karlwuest: I don't think we needed it. Probably, it's just that we had the `registration_params` before adding the PoP and then didn't change it. I think the change is fine.

crates/walrus-service/bin/node.rs
```diff
@@ -211,6 +234,9 @@ struct ConfigArgs {
     #[clap(long)]
     /// The host name or public IP address of the node.
     public_host: String,
+    /// The name of the storage node used in the registration.
+    #[clap(long)]
+    name: String,
```
karlwuest, line 239: Question: Do we also want to allow the nodes to specify the [node metadata](https://github.com/MystenLabs/walrus/blob/596a09373751d9f40e0f3315cbe5a7117f9acd87/crates/walrus-sui/src/types.rs#L94) or are we fine with just setting empty values for all nodes initially?

mlegner: Ah, good point. I can add this explicitly.

mlegner: Added in 1753e4d84de52b1cac0d13e097c1d8774e1c59c3.

karlwuest: Looks good, thanks!

PR #1383, opened 2025-01-10
feat(test): Add event blob check to simtests

sadhansood:
## Description

This PR adds a check that event blobs are being created. And enabling it in couple sim tests for now. Intentionally not enabling it for every simtest as the tests slow down a lot (due to single threaded runtime) and fail abruptly due to timeouts and delays in different places. So, this might require some work on individual tests before enabling it every where.

## Test plan

Running locally works.

### DIFF
--- a/crates/walrus-simtest/tests/simtest.rs
+++ b/crates/walrus-simtest/tests/simtest.rs
@@ -23,7 +23,7 @@ mod tests {
         client::{responses::BlobStoreResult, Client, ClientCommunicationConfig, StoreWhen},
         test_utils::{test_cluster, SimStorageNodeHandle},
     };
-    use walrus_sui::client::{BlobPersistence, PostStoreAction, SuiContractClient};
+    use walrus_sui::client::{BlobPersistence, PostStoreAction, ReadClient, SuiContractClient};
     use walrus_test_utils::WithTempDir;
 
     const FAILURE_TRIGGER_PROBABILITY: f64 = 0.01;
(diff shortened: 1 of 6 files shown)


crates/walrus-simtest/tests/simtest.rs
```diff
@@ -78,22 +78,39 @@ mod tests {
...
-        let read_blob = client
+        // Read the blob using primary slivers. Retry because nodes may not have received the blob
+        // certify event yet and until then the slivers will not be readable.
+        let mut read_blob_result = client
             .as_ref()
             .read_blob::<Primary>(&blob_confirmation.blob_id)
-            .await
-            .context("should be able to read blob we just stored")?;
+            .await;
+        while read_blob_result.is_err() {
```
halfprice, line 87: nit: can we add a timeout here?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -136,16 +153,31 @@ mod tests {
...
+                Some(10),
             )
             .await
             .unwrap();
 
         write_read_and_check_random_blob(&client, 31415, false)
             .await
             .expect("workload should not fail");
+
+        loop {
```
halfprice, line 168: This test may not be very interesting. Can we move the check to `test_repeated_shard_move_with_workload` and `test_new_node_joining_cluster`?

sadhansood: done - added to both tests

PR #1385, opened 2025-01-11
fix: fix walrus info command

halfprice:
## Description

It was reporting object not found when getting `active_set`.

(absolutely not sure if this is the right fix...)

Contribute to WAL-513

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/read_client.rs
+++ b/crates/walrus-sui/src/client/read_client.rs
@@ -54,7 +54,6 @@ use crate::{
         move_structs::{
             EpochState,
             EventBlob,
-            Key,
             StakingInnerV1,
             StakingObjectForDeserialization,
             StakingPool,
@@ -862,15 +861,11 @@ impl ReadClient for SuiReadClient {
 
         let staking_object = self.get_staking_object().await?;
         let active_set_id = staking_object.inner.active_set;
-        let key_tag = contracts::extended_field::Key
-            .to_move_struct_tag_with_type_map(&self.type_origin_map(), &[])?;
+        let type_map = self.type_origin_map().clone();
+
         let active_set = self
             .sui_client
-            .get_dynamic_field::<Key, ActiveSet>(
-                active_set_id,
-                key_tag.into(),
-                Key { dummy_field: false },
-            )
+            .get_extended_field::<ActiveSet>(active_set_id, &type_map)
             .await?;
         Ok(active_set.nodes.into_iter().collect())
     }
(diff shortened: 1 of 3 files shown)


crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -638,3 +638,14 @@ pub struct SharedBlob {
 impl AssociatedContractStruct for SharedBlob {
     const CONTRACT_STRUCT: StructTag<'static> = contracts::shared_blob::SharedBlob;
 }
+
+/// Sui type for the key of an extended field.
```
karlwuest, line 642: Can you add a comment here and create an issue to revisit? It really feels unclean that an empty Move struct is not actually empty, maybe we can get some insights from the Move team.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -863,9 +864,13 @@ impl ReadClient for SuiReadClient {
...
             .to_move_struct_tag_with_type_map(&self.type_origin_map(), &[])?;
-        let active_set: ActiveSet = self
+        let active_set = self
             .sui_client
-            .get_dynamic_field(active_set_id, key_tag.into(), ())
+            .get_dynamic_field::<Key, ActiveSet>(
+                active_set_id,
+                key_tag.into(),
+                Key { dummy_field: false },
+            )
```
karlwuest: Suggestion: We could put this into a function `get_extended_field<T>` on the RetryClient to abstract the inner workings.

crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -638,3 +638,14 @@ pub struct SharedBlob {
...
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct Key {
```
karlwuest: Suggestion: Change the visibility to `pub(crate)`.

crates/walrus-sui/src/types/move_structs.rs
halfprice: Sure. Created WAL-513

crates/walrus-sui/src/client/read_client.rs
halfprice: Done

crates/walrus-sui/src/types/move_structs.rs
halfprice: Done

PR #1386, opened 2025-01-11
fix: ensure that the signers bitmap is correct

karlwuest:
## Description

- fixes an issue where the signers bitmap is too short

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/system/bls_aggregate.move
+++ b/contracts/walrus/sources/system/bls_aggregate.move
@@ -209,8 +209,12 @@ fun verify_certificate_and_weight(
     let mut offset: u64 = 0;
 
     // The signers bitmap must be exactly long enough to hold all members.
-    let excess_bits = signers_bitmap.length() * 8 - self.members.length();
-    assert!(excess_bits >= 0 && excess_bits < 8, EInvalidBitmap);
+    let signers_bit_length = signers_bitmap.length() * 8;
+    assert!(
+        signers_bit_length >= self.members.length()
+            && signers_bit_length <= self.members.length() + 8,
+        EInvalidBitmap,
+    );
 
     signers_bitmap.do_ref!(|byte| {
         (8u8).do!(|i| {
--- a/crates/walrus-sui/src/client/transaction_builder.rs
+++ b/crates/walrus-sui/src/client/transaction_builder.rs
@@ -19,6 +19,7 @@ use sui_types::{
     SUI_CLOCK_OBJECT_ID,
     SUI_CLOCK_OBJECT_SHARED_VERSION,
 };
+use tokio::sync::OnceCell;
 use tracing::instrument;
 use walrus_core::{
     messages::{ConfirmationCertificate, InvalidBlobCertificate, ProofOfPossession},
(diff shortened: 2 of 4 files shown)


contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -209,8 +209,12 @@ fun verify_certificate_and_weight(
...
 
     // The signers bitmap must be exactly long enough to hold all members.
-    let excess_bits = signers_bitmap.length() * 8 - self.members.length();
-    assert!(excess_bits >= 0 && excess_bits < 8, EInvalidBitmap);
+    let signers_bit_length = signers_bitmap.length() * 8;
+    assert!(
+        signers_bit_length >= self.members.length()
+            && signers_bit_length <= self.members.length() + 8,
+        EInvalidBitmap,
+    );
```
karlwuest, line 217: This check does the same, but since Move doesn't have signed integers, the previous version would have failed with an underflow error instead of a meaningful one, in the case where the bitmap is too short.

@mlegner, @jonas-lj : We could also remove the minimum length requirement here. Then we would not need to fetch the committee size from chain to construct the bitmap (to simplify the code a bit again), and could just use the maximum signer index (+ 1) to determine the bitmap size.

mlegner: > We could also remove the minimum length requirement here. Then we would not need to fetch the committee size from chain to construct the bitmap (to simplify the code a bit again), and could just use the maximum signer index (+ 1) to determine the bitmap size.

Sounds like a good idea to me. 👍

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -301,14 +309,15 @@ impl WalrusPtbBuilder {
         Ok(())
     }
 
-    fn signers_to_bitmap(signers: &[u16]) -> Vec<u8> {
-        let mut bitmap = vec![0; signers.len().div_ceil(8)];
+    async fn signers_to_bitmap(&self, signers: &[u16]) -> SuiClientResult<Vec<u8>> {
+        let committee_size = self.system_object().await?.committee_size() as usize;
+        let mut bitmap = vec![0; committee_size.div_ceil(8)];
         for signer in signers {
```
mlegner, line 315: Question, just to make sure: The signers are guaranteed to be in the range `(0..committee_size)`?

contracts/walrus/sources/system/bls_aggregate.move
mlegner: Then we would use something like `signers.iter().max() + 1` instead of `committee_size` to determine the vector length, correct?

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: Yes, those are the indices into the members vector.

contracts/walrus/sources/system/bls_aggregate.move
karlwuest: Exactly. IIUC, we anyway want to redeploy PTN again on Monday? In that case I'll update the PR and then it's ready for Monday.

karlwuest: Or do we want to merge this one to get the current PTN running correctly and then have another one to make the additional change?

mlegner: Let's merge it as it is now and optionally make the contract change on Monday.

PR #1387, opened 2025-01-12
chore: refactor EventProcessorRuntime into its own crate

wbbradley:
## Description

Code organization: pull EventProcessorRuntime into its own crate.

## Test plan

Tested with local tests and CI pipeline.

No release impact.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -25,6 +25,7 @@ use rand::{rngs::StdRng, thread_rng, Rng, SeedableRng};
 use serde::Serialize;
 use start_epoch_change_finisher::StartEpochChangeFinisher;
 use storage::blob_info::PerObjectBlobInfoApi;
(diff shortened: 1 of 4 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -95,11 +96,35 @@ use self::{
     committee::{CommitteeService, NodeCommitteeService},
     config::{StorageNodeConfig, SuiConfig},
     contract_service::{SuiSystemContractService, SystemContractService},
-    errors::IndexOutOfRange,
+    errors::{
```
wbbradley, line 98: This was just me consolidating the `use` statements into one unit, rather than have them be spread out. I also noticed there was some inconsistency with respect to how symbols were imported from relative (`self`) crates.

Rationale here is that in my experience, having `use` statements together, and having `mod` statements together leads to simpler reading comprehension.

Happy to discuss.

PR #1388, opened 2025-01-12
chore: split walrus info command

liquid-helium:
## Description

complete WAL-340

Added sub-commands for `info`

`info epoch`
`info storage`
`info size`
`info price`
`info dev`

The new syntax is not fully compatible with the previous one, i.e., the output from `info dev` is different, it contains only the InfoDevOutput from the previous version.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: changes to the syntax and output of the `info` command


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -399,6 +398,27 @@ pub enum CliCommands {
     },
 }
 
+/// Subcommands for the Info command
```
giac-mysten: Nit (here and elsewhere): Our comments convention requires a `.` at the end.
```suggestion
/// Subcommands for the Info command.
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -399,6 +398,27 @@ pub enum CliCommands {
...
+    /// Print epoch information
+    Epoch,
+    /// Print storage information
+    Storage,
+    /// Print size information
+    Size,
+    /// Print price information
+    Price,
+    /// Print BFT information
+    Bft,
```
giac-mysten: Remark: `Bft` may be a bit opaque. In this case, it's probably not a problem since only people who know what bft means will reach out for this command. Maybe expand a bit in the docstring?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -277,40 +282,116 @@ impl CliOutput for BlobIdConversionOutput {
...
         } = self;
 
         // NOTE: keep text in sync with changes in the contracts.
         printdoc!(
             "
 
             {top_heading}
+            ",
+            top_heading = "Walrus system information".bold(),
+        );
```
giac-mysten: Remark: I guess since this is just a heading, we don't need `printdoc`?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -493,9 +504,32 @@ impl ClientCommandRunner {
...
         )
         .await?;
-        InfoOutput::get_system_info(&sui_read_client, dev)
-            .await?
-            .print_output(self.json)
+
+        match command {
+            None | Some(InfoCommands::All) => InfoOutput::get_system_info(&sui_read_client)
+                .await?
+                .print_output(self.json),
```
giac-mysten: Question: Why do `None` and `All` coincide with printing everything? Looks a bit counterintuitive to me. Is it for backwards compatibility?
Suggestion: Maybe we can have `None` print a minimal information summary (e.g., what `info` without `--dev` does now?)

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: Replaced with a println!

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Done.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Thanks for the catch, change applied here and other places.

crates/walrus-service/src/client/cli/runner.rs
liquid-helium: Thanks for the suggestion, change applied: info print basic information, while info all prints everything.

PR #1390, opened 2025-01-13
feat: allow short bitmaps in move code for certs

karlwuest:
## Description

Makes the bitmap for certificates more permissive. This will allow us to later construct certificates without querying the committee size (WAL-514). WAL-514 is not included in this PR to not make this a breaking change for the current PTN.

## Test plan

- added unit tests

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/system/bls_aggregate.move
+++ b/contracts/walrus/sources/system/bls_aggregate.move
@@ -212,7 +212,7 @@ fun verify_certificate_and_weight(
     let mut non_signer_aggregate_weight = 0;
     let mut non_signer_public_keys: vector<Element<UncompressedG1>> = vector::empty();
     let mut offset: u64 = 0;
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/system/bls_aggregate.move
```diff
@@ -207,23 +212,29 @@ fun verify_certificate_and_weight(
     let mut non_signer_aggregate_weight = 0;
     let mut non_signer_public_keys: vector<Element<UncompressedG1>> = vector::empty();
     let mut offset: u64 = 0;
+    let n_members = self.members.length();
```
jonas-lj: nit: You can use the new `n_members` functions here, or make it `test_only`.

PR #1391, opened 2025-01-13
fix!: revise all configuration files

mlegner:
## Description

This PR revises all configuration files to be more consistent and flexible. It also adds additional checks to ensure future changes are detected and handled appropriately. This includes the following:

- For optional parameter groups represented by structs, moved from a pattern of including an option of the struct in the parent config to a pattern of including the struct itself, but implementing `Default` for the struct and deriving `#[serde(default)]` on it.
- Added configuration parameters for the default database options.
- Consistently use `_secs` and `_millis` for durations.
- Simplified the event-processor configuration.
- Removed some redundant configuration parameters.
- Renamed `exchange_object` to `exchange_objects` and only accept a list of objects.
- Extended tests for the deserialization of partial configuration files.
- Improved tests to check and update the example configuration files.
- Added a check to the CI workflow to warn about changes to the example configuration files.

Closes WAL-153
Closes WAL-226
Closes WAL-243
Closes WAL-302

## Test plan

Existing automatic tests and improved/extended/added tests. New CI check.

Additional tests:

- [ ] Tests with a local testbed.
- [ ] Deploy to PTN.

---

## Release notes

- [x] Storage node: Breaking changes to the configuration file.
- [x] CLI: Breaking changes to the configuration file.

### DIFF
--- a/crates/walrus-service/src/node/storage/database_config.rs
+++ b/crates/walrus-service/src/node/storage/database_config.rs
@@ -5,7 +5,7 @@ use rocksdb::{DBCompressionType, Options};
 use serde::{Deserialize, Serialize};
 
 /// Options for configuring a column family.
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
+#[derive(Debug, Clone, Default, Deserialize, Serialize, PartialEq)]
 #[serde(default)]
 pub struct DatabaseTableOptions {
     /// Set it to true to enable key-value separation.
@@ -39,8 +39,8 @@ pub struct DatabaseTableOptions {
     max_bytes_for_level_base: Option<u64>,
 }
 
-impl Default for DatabaseTableOptions {
-    fn default() -> Self {
+impl DatabaseTableOptions {
+    fn standard() -> Self {
         Self {
             enable_blob_files: Some(false),
             min_blob_size: Some(0),
@@ -55,9 +55,7 @@ impl Default for DatabaseTableOptions {
             max_bytes_for_level_base: Some(512 << 20),
         }
     }
-}
 
-impl DatabaseTableOptions {
     fn optimized_for_blobs() -> Self {
         Self {
             enable_blob_files: Some(true),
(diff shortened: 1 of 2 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -877,43 +867,34 @@ impl EventProcessorRuntime {
...
-                        )),
-                        tokio::spawn(async { std::future::pending().await }),
+            if use_legacy_event_provider {
+                let read_client = runtime.block_on(async { sui_config.new_read_client().await })?;
+                (
+                    Box::new(SuiSystemEventProvider::new(
+                        read_client,
+                        sui_config.event_polling_interval,
+                    )),
+                    tokio::spawn(async { std::future::pending().await }),
```
jpcsmith, line 877: Question: Why is a forever-pending future that does nothing being spawned here?

crates/walrus-service/src/node/storage/database_config.rs
jpcsmith, line 1: Question: Do we need the default and optimised for blobs fields?

crates/walrus-service/src/node/config.rs
```diff
@@ -496,9 +484,14 @@ pub mod defaults {
         2000
     }
 
-    /// Returns true iff the value is the default.
+    /// Returns true iff the value is the default and we don't run in test mode.
     pub fn is_default<T: PartialEq + Default>(t: &T) -> bool {
-        t == &T::default()
+        !cfg!(test) && t == &T::default()
+    }
```
jpcsmith: Nit: Could you please add a (non-doc-string) comment as to why we dont want this run in test mode?

crates/walrus-service/bin/node.rs
mlegner: Since #1075 we are checking for crashed runtimes and crash the node in that case because we had issues with runtimes silently crashing before. I believe this was just a way to align the behavior of the two event providers (cc @sadhansood).

Happy to revisit this, but I think we will get rid of the legacy event provider soon anyway.

crates/walrus-service/src/node/storage/database_config.rs
mlegner: We use one of two default settings for all database tables. So far, it was not possible to make changes that applied to multiple tables, they would have to add configurations for all tables separately. This now allows to change the defaults and thus multiple tables simultaneously.

mlegner: I've just made some changes in ba3cf11d8f79deeee42c5d09fb72da552c6427f9 to how the DB config is handled and added some docs for that as well. In the future, we _could_ make this properly hierarchical, but that's beyond this PR.

jpcsmith: Okay, thank you for the clarification.

Nit: The fact that we have the two "classes" is a bit of an oversimplification of the settings, with time I would expect that we would have somewhat somewhat unique defaults for each of the tables in our code. So, I am not sure we should also give the user the ability to change the behaviour of multiple tables in one go. If you think it may be useful though, keep it in and we can always revise it later as needed.

mlegner: Hmm... I see your point. I still think the PR's setup is an improvement over what we had before, but I will create a follow-up issue to revise the DB configuration.

PR #1397, opened 2025-01-13
feat: boilerplate for the backup node

wbbradley:
## Description

Create most of the boilerplate for running a backup node.

## Test plan

Minimal testing at this point.

---

## Release notes

- [x] Backup node: an entrypoint called walrus-backup will now be included in the build process. It has minimal (tailing Walrus-related Sui events) functionality at this point.

### DIFF
--- a/crates/walrus-service/bin/backup.rs
+++ b/crates/walrus-service/bin/backup.rs
@@ -1,0 +1,161 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Walrus Backup Node entry point.
+
+use std::path::PathBuf;
+
+use clap::{Parser, Subcommand};
+use tokio::{runtime::Runtime, sync::oneshot};
+use tokio_util::sync::CancellationToken;
+use walrus_service::{
+    backup::{BackupNodeConfig, BackupNodeRuntime},
+    node::events::event_processor_runtime::EventProcessorRuntime,
+    utils::{self, version, wait_until_terminated, LoadConfig as _, MetricsAndLoggingRuntime},
+};
+
+const VERSION: &str = version!();
+
+/// Manage and run a Walrus backup node.
+#[derive(Parser)]
+#[clap(rename_all = "kebab-case")]
+#[clap(name = env!("CARGO_BIN_NAME"))]
+#[clap(version = VERSION)]
+#[derive(Debug)]
+struct Args {
+    #[command(subcommand)]
+    command: BackupCommands,
+}
+
+#[derive(Subcommand, Debug, Clone)]
+#[clap(rename_all = "kebab-case")]
+enum BackupCommands {
+    /// Run a backup node with the provided configuration.
+    Run { config_path: PathBuf },
+}
+
+fn main() -> anyhow::Result<()> {
+    let args = Args::parse();
+
+    if !matches!(args.command, BackupCommands::Run { .. }) {
+        utils::init_tracing_subscriber()?;
+    }
+
+    match args.command {
+        BackupCommands::Run { config_path } => {
+            let config = BackupNodeConfig::load(config_path)?;
+            commands::run_backup_node(config)?
+        }
+    };
+    Ok(())
+}
+
+mod commands {
+    #[cfg(not(msim))]
+    use tokio::task::JoinSet;
+    use walrus_service::utils;
+
+    use super::*;
+
+    pub(super) fn run_backup_node(config: BackupNodeConfig) -> anyhow::Result<()> {
+        let metrics_runtime = MetricsAndLoggingRuntime::start(config.metrics_address)?;
+        let registry_clone = metrics_runtime.registry.clone();
+        metrics_runtime.runtime.spawn(async move {
+            registry_clone
+                .register(mysten_metrics::uptime_metric(
+                    "walrus_backup_node",
+                    VERSION,
+                    "walrus",
+                ))
+                .unwrap();
+        });
+
+        tracing::info!(version = VERSION, "Walrus backup binary version");
+        tracing::info!(
+            metrics_address = %config.metrics_address, "started Prometheus HTTP endpoint",
+        );
+
+        utils::export_build_info(&metrics_runtime.registry, VERSION);
+
+        let cancel_token = CancellationToken::new();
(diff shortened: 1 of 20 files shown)


crates/walrus-service/src/backup.rs
```diff
@@ -0,0 +1,307 @@
...
+impl BackupNodeRuntime {
+    /// Starts a new backup node runtime.
+    pub fn start(
+        metrics_registry: Registry,
+        exit_notifier: oneshot::Sender<()>,
+        event_manager: Box<dyn EventManager>,
+        cancel_token: CancellationToken,
+    ) -> anyhow::Result<Self> {
+        let runtime = runtime::Builder::new_multi_thread()
+            .thread_name("walrus-node-runtime")
```
halfprice: Nit: maybe walrus-backup-runtime?

crates/walrus-service/src/backup.rs
```diff
@@ -0,0 +1,307 @@
...
+                Err(err) => return Err(err),
+            },
+            _ = cancel_token.cancelled() => {
+            },
+        }
+        Ok(())
+    }
+
+    fn get_storage_node_cursor(&self) -> anyhow::Result<EventStreamCursor> {
+        // TODO: restart from a saved cursor.
```
halfprice: Suggestion: the convention in Walrus codebase is that we always (at least in the recent months) create an issue and link it here to track it.

crates/walrus-service/src/backup.rs
```diff
@@ -0,0 +1,307 @@
...
+        // Important: Events must be handled consecutively and in order to prevent (intermittent)
+        // invariant violations and interference between different events.
+        while let Some((element_index, stream_element)) = indexed_element_stream.next().await {
+            let element_index = u64::try_from(element_index).expect("should fit into a u64");
+            if element_index < cursor_element_index {
+                return Err(anyhow!(
+                    "event stream out of sync [element_index={element_index}, \
+                        cursor_index={cursor_element_index}]"
+                ));
+            }
```
halfprice: Question: what is this check about? IIUC, the cursor_element_index is the initial index of the cursor, so this check seems never be true.

crates/walrus-service/src/backup.rs
```diff
@@ -0,0 +1,307 @@
...
+            .inner
+            .event_manager
+            .events(event_cursor.clone())
+            .await?;
+        let event_index = event_cursor
+            .clone()
+            .element_index
+            .try_into()
+            .expect("64-bit architecture");
+
```
halfprice: Remark: I don't think you need to deal with init_state here yet. Just return `Return Ok((Pin::from(event_stream), event_index));`.

`init_state` is only needed when we restore event cursor from event blobs. We may need it later on, but since we are not reading from the event blob yet, let's don't add it here yet.

crates/walrus-service/src/common/utils.rs
```diff
@@ -791,6 +791,33 @@ pub async fn collect_event_blobs_for_catchup(
...
+/// When a node starts up for the first time, older event blobs might have expired due to the
+/// MAX_EPOCHS_AHEAD limit. Let's say the earliest available event starts at index N ( where N >
+/// 0).
+///
+/// In this case, the event stream can only provide events starting from index N. But the node,
+/// being brand new, wants to start processing from index 0. In this scenario, we actually want
+/// to reposition the node's cursor to index N. This is safe because those events are no longer
+/// available in the system anyway (they've expired), and marking them as completed prevents the
+/// event tracking system from flagging this natural gap as an error.
+pub fn should_reposition_cursor(event_index: usize, actual_event_index: usize) -> bool {
```
halfprice: Although we don't want to deal with reposition yet, feel free to leave this here and don't need to move back.

crates/walrus-service/src/backup.rs
wbbradley: Done. Good idea.

crates/walrus-service/src/backup.rs
wbbradley: This is copy pasta from the storage node's `process_events`.

wbbradley: I agree though that is is probably unnecessary. I thought about converting it to a `panic!`, but didn't want to pull the trigger on that.

wbbradley: In the future the cursor_element_index will be some non-zero value, loaded from the DB, I presume...

crates/walrus-service/src/backup.rs
wbbradley: Good catch.

crates/walrus-service/src/backup.rs
wbbradley: Gotcha. Thanks!

crates/walrus-service/bin/backup.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    fn join(&mut self) -> Result<(), anyhow::Error> {
+        tracing::debug!("waiting for the event processor to shutdown...");
+        self.runtime.block_on(&mut self.event_processor_handle)?
+    }
+}
+/// Wait for SIGINT and SIGTERM (unix only).
+#[tracing::instrument(skip_all)]
+async fn wait_until_terminated(mut exit_listener: oneshot::Receiver<()>) {
```
halfprice: Nit: feel this can be extracted out into a library too since the `wait_until_terminated` in the node.rs is almost identical.

crates/walrus-service/src/backup.rs
```diff
@@ -0,0 +1,239 @@
...
+        bail!("event stream for blob events stopped")
+    }
+}
+
+/// The internal state of a Walrus backup node.
+#[derive(Debug)]
+struct BackupNodeInner {
+    event_manager: Box<dyn EventManager>,
+    // current_epoch: watch::Sender<Epoch>,
+    // is_shutting_down: AtomicBool,
```
halfprice: nit: can we remove this? In general we don't have comment out code in the codebase. We prefer `TODO`.

crates/walrus-service/bin/backup.rs
wbbradley: Yeah, pulling this out...

crates/walrus-service/src/backup.rs
wbbradley: Good catch.

PR #1399, opened 2025-01-14
feat(metrics): automatically add name and host labels

suiwombat:
## Description
enforce labels for host and name within metric data.

## Test Plan
local testing

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -249,7 +249,7 @@ impl MetricsPushConfig {
     fn set_host(&mut self, host: String) {
         self.labels
             .get_or_insert_with(HashMap::new)
-            .insert("host".into(), host.into());
+            .insert("host".into(), host);
     }
 }
 


crates/walrus-service/src/node/config.rs
```diff
@@ -227,6 +227,32 @@ pub struct MetricsPushConfig {
...
+    pub fn set_name(&mut self, name: &str) {
+        self.labels
+            .get_or_insert_with(HashMap::new)
+            .insert("name".into(), name.into());
+
+        // also set the host label and try to get the hostname to do it.  if we cannot
+        // use name as a fallback.
+        let host =
+            hostname::get().map_or_else(|_| name.into(), |v| v.to_string_lossy().to_string());
+        self.set_host(host);
```
mlegner, line 245: Suggestion: Don't overwrite any `host` set in the metrics push configuration.

pei-mysten: @mlegner this one is majorly for @IngaAgibalova since it would be a lot easier for them to match the host name in contract and in grafana

suiwombat: Hi @mlegner i'm not sure i understand the comment.  Pei and I decided last night to use the name from the config to set the metric label for name.  We wanted to set the host label from the hostname of the underlying os.  We default to the name in the config if that operation fails.

If a user sets host or name in the label config for metrics, this will overwrite it, but it's a tradeoff we made to ensure something is always there.  I COULD set these values only if they are not already present - if that's what you mean.  That's not a bad idea, but i purposefully didn't do that to make this predictable for us.  Otherwise people can set whatever they want for name and host labels.  thoughts?

mlegner: OK, that makes sense. I was just confused because in the deployment instructions we ask users to set the `host` explicitly. Then I guess we can remove that, right?

pei-mysten: yup @mlegner with that one we don't need our users to set `host` explicitly.

crates/walrus-service/src/node/config.rs
```diff
@@ -227,6 +227,32 @@ pub struct MetricsPushConfig {
...
+    }
+
+    /// set metric label host = bar
+    fn set_host(&mut self, host: String) {
+        self.labels
+            .get_or_insert_with(HashMap::new)
+            .insert("host".into(), host.into());
```
mlegner: Nit (Clippy finding):
```suggestion
            .insert("host".into(), host);
```

PR #1403, opened 2025-01-14
feat: publish packages separately

karlwuest:
## Description

Publishes the `wal`, `wal_exchange`, and `walrus` packages separately.

## Test plan

Existing tests + manually tested with local test cluster.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/system_setup.rs
+++ b/crates/walrus-sui/src/system_setup.rs
@@ -113,16 +113,16 @@ pub(crate) struct PublishSystemPackageResult {
 
 /// Copy files from the `source` directory to the `destination` directory recursively.
 #[tracing::instrument(err, skip(source, destination))]
-async fn copy_recursively(source: impl AsRef<Path>, destination: impl AsRef<Path>) -> Result<()> {
-    tokio::fs::create_dir_all(destination.as_ref()).await?;
+fn copy_recursively(source: impl AsRef<Path>, destination: impl AsRef<Path>) -> Result<()> {
+    std::fs::create_dir_all(destination.as_ref())?;
     for entry in WalkDir::new(source.as_ref()) {
         let entry = entry?;
         let filetype = entry.file_type();
         let dest_path = entry.path().strip_prefix(source.as_ref())?;
         if filetype.is_dir() {
-            tokio::fs::create_dir_all(destination.as_ref().join(dest_path)).await?;
+            std::fs::create_dir_all(destination.as_ref().join(dest_path))?;
         } else {
-            tokio::fs::copy(entry.path(), destination.as_ref().join(dest_path)).await?;
+            std::fs::copy(entry.path(), destination.as_ref().join(dest_path))?;
         }
     }
     Ok(())
@@ -142,7 +142,7 @@ pub async fn publish_coin_and_system_package(
     deploy_directory: Option<PathBuf>,
 ) -> Result<PublishSystemPackageResult> {
     let walrus_contract_directory = if let Some(deploy_directory) = deploy_directory {
-        copy_recursively(&walrus_contract_directory, &deploy_directory).await?;
+        copy_recursively(&walrus_contract_directory, &deploy_directory)?;
         deploy_directory
     } else {
         walrus_contract_directory


Cargo.toml
```diff
@@ -73,6 +73,7 @@ serde_test = "1.0.177"
 serde_with = { version = "3.12", features = ["base64"] }
 serde_yaml = "0.9"
 snap = "1.1.0"
+sui = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.40.1" }
```
mlegner: Suggestion: Add a TODO to get rid of this with its 2k dependencies again. 😅

crates/walrus-service/bin/deploy.rs
```diff
@@ -57,8 +57,8 @@ struct DeploySystemContractArgs {
     #[clap(long, default_value = "testnet")]
     sui_network: SuiNetwork,
     /// The directory in which the contracts are located.
-    #[clap(long, default_value = "./contracts/walrus")]
-    contract_path: PathBuf,
+    #[clap(long, default_value = "./contracts")]
+    contract_dir: PathBuf,
```
mlegner, line 85: Important: In the `sui-operations` repo we use the `--contract-path` option ([here](https://github.com/MystenLabs/sui-operations/blob/c63cdc20271b7247b0869fc1d2266f0d7c579e06/.github/workflows/walrus-deploy-ptn.yaml#L184)); that would need to be adjusted.

crates/walrus-service/bin/deploy.rs
```diff
@@ -98,6 +98,9 @@ struct DeploySystemContractArgs {
     /// The maximum number of epochs ahead for which storage can be obtained.
     #[arg(long, default_value_t = 104)]
     max_epochs_ahead: EpochCount,
+    /// Copy contracts to `working_dir` and publish from there to keep the `Move.toml` unchanged.
+    #[arg(long, action)]
+    deploy_contracts_from_working_dir: bool,
```
mlegner: Suggestion: This could be made the default with an opposite flag `do_not_copy_contracts`.

crates/walrus-service/src/testbed.rs
```diff
@@ -351,11 +360,7 @@ pub async fn deploy_walrus_contract(
     .await?;
 
     let exchange_object = contract_client
-        .create_and_fund_exchange(
-            // TODO(WAL-230): Use exchange package ID once published separately.
-            contract_client.read_client().get_system_package_id(),
-            WAL_AMOUNT_EXCHANGE,
-        )
+        .create_and_fund_exchange(system_ctx.wal_exchange_pkg_id, WAL_AMOUNT_EXCHANGE)
```
mlegner, line 365: Suggestion (for a follow-up): Create multiple exchange objects here.

crates/walrus-sui/src/system_setup.rs
```diff
@@ -136,36 +104,55 @@ pub(crate) async fn publish_package(
...
+    pub wal_pkg_id: ObjectID,
+    pub wal_exchange_pkg_id: ObjectID,
     pub init_cap_id: ObjectID,
     pub upgrade_cap_id: ObjectID,
     pub treasury_cap_id: ObjectID,
 }
 
+/// Copy files from the `source` directory to the `destination` directory recursively.
+#[tracing::instrument(err, skip(source, destination))]
+async fn copy_recursively(source: impl AsRef<Path>, destination: impl AsRef<Path>) -> Result<()> {
```
mlegner: Question: Is this taken from Stackoverflow? (I was checking if there is a standard way of doing this... 😅)

karlwuest: No, it's not. I was also looking for a standard way, but only found a solution that was actually implemented recursively (which I didn't like 😅).

crates/walrus-service/bin/deploy.rs
karlwuest: https://github.com/MystenLabs/sui-operations/pull/4785

crates/walrus-service/src/testbed.rs
karlwuest: Created an issue on linear -> https://linear.app/mysten-labs/issue/WAL-520/create-additional-exchange-objects-when-deploying-testbed

PR #1404, opened 2025-01-14
feat: third-party registration and staking with multiple nodes

mlegner:
## Description

- Modifies the `walrus stake` command to allow staking with multiple nodes in a single PTB.
- Adds a new `walrus-deploy register-nodes` command to allow registering multiple third-party nodes based on the registration parameters exported through the `walrus-node setup` command.

Closes WAL-506.

## Test plan

Ran the commands in a local testbed, see transactions for [registration](https://suiscan.xyz/devnet/tx/Bog4ERtczKMjw3ETtuPHaR6PwUQMHQn3fYxSibZy5BoH) and [staking](https://suiscan.xyz/devnet/tx/E7UvSgrVaPp2tbs31CH4L5wcfzwZfmP2wSUBEyXR6TPT).

---

## Release notes

- [x] CLI: Allows staking with multiple nodes simultaneously.

### DIFF
--- a/crates/walrus-service/bin/deploy.rs
+++ b/crates/walrus-service/bin/deploy.rs
@@ -38,8 +38,8 @@ struct Args {
 #[derive(Subcommand, Debug, Clone)]
 #[clap(rename_all = "kebab-case")]
 enum Commands {
-    /// Register nodes based on parameters exported by the `walrus-node setup` command and
-    /// optionally stake with them.
+    /// Register nodes based on parameters exported by the `walrus-node setup` command, send the
+    /// storage-node capability to the respective node's wallet, and optionally stake with them.
     RegisterNodes(RegisterNodesArgs),
     /// Deploy the Walrus system contract on the Sui network.
     DeploySystemContract(DeploySystemContractArgs),
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -348,7 +348,7 @@ impl SuiContractClient {
         // Lock the wallet here to ensure there are no race conditions with object references.
         let wallet = self.wallet().await;
 
-        let mut pt_builder = WalrusPtbBuilder::new(self.read_client.clone(), self.wallet_address);
+        let mut pt_builder = self.transaction_builder();
         pt_builder.reserve_space(encoded_size, epochs_ahead).await?;
         let (ptb, _sui_cost) = pt_builder.finish().await?;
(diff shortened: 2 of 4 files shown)


crates/walrus-service/src/client/config.rs
```diff
@@ -67,6 +68,24 @@ impl Config {
...
+    /// configuration.
+    pub async fn new_contract_client_with_wallet_in_config(
+        &self,
+        gas_budget: u64,
+    ) -> anyhow::Result<SuiContractClient> {
+        let Some(wallet_config) = self.wallet_config.as_ref() else {
+            bail!(
+                "path to Sui wallet must be specified in client config through the 'wallet_config' \
+                field"
+            );
```
karlwuest, line 83: Question/Suggestion: Is it intentional that we don't resolve to the default sui wallet config location if no config is specified? 
If not, the nicest option would be to set the default sui wallet config path as default in `client::config::Config`.

crates/walrus-service/bin/deploy.rs
```diff
@@ -38,13 +38,37 @@ struct Args {
 #[derive(Subcommand, Debug, Clone)]
 #[clap(rename_all = "kebab-case")]
 enum Commands {
+    /// Register nodes based on parameters exported by the `walrus-node setup` command and
+    /// optionally stake with them.
```
karlwuest: Nit: add that the capabilities are send to the node addresses.

crates/walrus-sui/src/client.rs
```diff
@@ -634,17 +634,60 @@ impl SuiContractClient {
...
-    ) -> SuiClientResult<StakedWal> {
+        registration_params_with_stake_amounts: Vec<(
+            NodeRegistrationParams,
+            ProofOfPossession,
+            SuiAddress,
+        )>,
+    ) -> SuiClientResult<Vec<StorageNodeCap>> {
+        let count = registration_params_with_stake_amounts.len();
+
+        let mut pt_builder = WalrusPtbBuilder::new(self.read_client.clone(), self.wallet_address);
```
```suggestion
        let mut pt_builder = self.transaction_builder();
```

crates/walrus-service/src/client/config.rs
mlegner: This is actually intentional in this case as I assumed that it would most likely not be used with the "normal" wallet but with some specific admin wallet.

crates/walrus-sui/src/client.rs
mlegner: I've made the same changes for other instances of this throughout the file.

PR #1406, opened 2025-01-14
feat(cli): Add option to extend owned blob

joyqvq:
## Description

Closes WAL-178.

## Test plan

- added simtest
- tested in testnet, example output
```
target/release/walrus extend --blob-obj-id 0x288b4877726744be57417b1759ddb4d23a398e33fd075f05d34c3cd8b96eb450 --amount 78 --epochs-ahead 1
2025-01-14T21:06:14.651218Z  INFO walrus: client version: 1.10.0-e7df7304e5ab-dirty
2025-01-14T21:06:14.651524Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/joy/.config/walrus/client_config.yaml'
2025-01-14T21:06:14.652138Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/joy/.sui/sui_config/client.yaml'
 • extending blob... [00:00:00]                                                                                                                 efdk pochs_ahead 1, amount 78
epochs_ahead 1, amount 78
 ◉ done [00:00:01]                                                                                                                              Success: The blob has been extended by 1 epochs
```
https://testnet.suivision.xyz/txblock/3m62UankFxqpAAnov7DJA5UDXNg91VQbXcoZ8Q5o1rvN

invalid cli will not succeed
```
target/release/walrus extend --shared-blob-obj-id 0x288b4877726744be57417b1759ddb4d23a398e33fd075f05d34c3cd8b96eb450 --amount 1
error: the argument '--shared-blob-obj-id <SHARED_BLOB_OBJ_ID>' cannot be used with '--amount <AMOUNT>'

Usage: walrus extend --epochs-ahead <EPOCHS_AHEAD> --shared-blob-obj-id <SHARED_BLOB_OBJ_ID>

target/release/walrus extend --blob-obj-id 0x288b4877726744be57417b1759ddb4d23a398e33fd075f05d34c3cd8b96eb450 --epochs-ahead 1
error: the following required arguments were not provided:
  --amount <AMOUNT>

Usage: walrus extend --epochs-ahead <EPOCHS_AHEAD> --blob-obj-id <BLOB_OBJ_ID> --amount <AMOUNT>

For more information, try '--help'.
```
---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Add option to extend owned blobs with WAL

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -403,7 +403,7 @@ pub enum CliCommands {
         #[clap(long)]
         amount: u64,
     },
-    /// Extend a shared blob.
+    /// Extend an owned or shared blob.
     Extend {
         /// The object ID of the blob to extend.
         #[clap(long)]
@@ -413,7 +413,7 @@ pub enum CliCommands {
         #[clap(long)]
         shared: bool,
 
-        /// The number of epochs to extend the shared blob for.
+        /// The number of epochs to extend the blob for.
         ///
         /// If set to `max`, the blob is stored for the maximum number of epochs allowed by the
         /// system object on chain. Otherwise, the blob is stored for the specified number of
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -1239,7 +1239,7 @@ impl SuiContractClient {
         .await
     }
 
-    /// Extends the owned blob object with given epochs_ahead and WAL amount.
+    /// Extends the owned blob object with given epochs_ahead.
     pub async fn extend_blob(
         &self,
         blob_obj_id: ObjectID,


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -386,9 +386,18 @@ pub enum CliCommands {
...
+        shared_blob_obj_id: Option<ObjectID>,
+
+        /// The object ID of the owned blob to extend.
+        #[clap(long, requires = "amount")]
+        blob_obj_id: Option<ObjectID>,
+
+        /// The WAL amount for extending the owned blob.
+        #[clap(long, requires = "blob_obj_id")]
+        amount: Option<u64>,
+
```
mlegner: Important: I think this should not be needed, it should be possible to compute the amount based on the encoded size of the blob.

What we *could* do here and for other commands like `store` is to set a maximum budget (similar to Sui's gas budget). But I'd rather do that in a separate PR.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -386,9 +386,18 @@ pub enum CliCommands {
...
         /// The object ID of the shared blob to extend.
-        #[clap(long)]
-        shared_blob_obj_id: ObjectID,
-        /// The number of epochs to extend the shared blob for.
+        #[clap(long, conflicts_with_all = ["blob_obj_id", "amount"])]
+        shared_blob_obj_id: Option<ObjectID>,
+
+        /// The object ID of the owned blob to extend.
+        #[clap(long, requires = "amount")]
+        blob_obj_id: Option<ObjectID>,
```
mlegner: Suggestion: Use an [`ArgGroup`](https://docs.rs/clap/latest/clap/struct.ArgGroup.html) similar to `FileOrBlobId` or `FileOrBlobIdOrObjectId` defined in this file. This enforces that *exactly one* of the arguments is required.

joyqvq: it's actually quite awkward to use it here since there is actually more information than just "needing exactly one arugment". it also tells me if it is shared blob or not to call the right function. i break it down into --shared flag and an objectId arg instead.

crates/walrus-service/src/client/cli/args.rs
joyqvq: iiuc, the blob.storage.size is a good proxy for the encoded_size since the blob object onchain only shows the unencoded size? 

modified to remove this from arg.

crates/walrus-service/src/client/cli/args.rs
mlegner: Suggestion: We could just have the object ID (removing the new `--shared` flag), then request that object from the full node, and then call one of the two functions depending on whether it's a `Blob` or `SharedBlob`. WDYT?

joyqvq: `get_sui_object` requires a associated contract struct when calling so the code will look a bit ugly to try with both types.

i don't have a strong opinion here since the user is expected to know what type they are using when calling.

mlegner: Hmm... Maybe it's not worth the extra complexity. If we leave it like this, I'd suggest to explicitly add to the `Extend` and/or `blob_obj_id` docstring that `--shared` must be used to extend a shared blob.

joyqvq: revised!

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -403,9 +403,14 @@ pub enum CliCommands {
     },
     /// Extend a shared blob.
```
```suggestion
    /// Extend a shared or owned blob.
```

crates/walrus-sui/src/client.rs
```diff
@@ -1188,6 +1188,27 @@ impl SuiContractClient {
         );
         Ok(shared_blob_obj_id[0])
     }
+
+    /// Extends the owned blob object with given epochs_ahead and WAL amount.
```
mlegner: Nit:
```suggestion
    /// Extends the owned blob object with given `epochs_ahead`.
```

Issue #997, opened 2024-10-09
Revise all API (node, aggregator, publisher)


mlegner:
Includes the following (not necessarily exhaustive):

- [X] ~~Change the recovery-symbol path to something like ~~`~~/v1/recovery/:blob_id/:sliver_type/:sliver_pair_index/:target_pair_index~~`~~ (see #542, #952)~~
- [ ] Decide which node APIs are "internal" and add documentation that authentication may be required for them in the future
- [ ] Potentially combine some API endpoints to reduce overhead
- [X] Make response formats (especially errors) on aggregator/publisher consistent with those on the storage nodes

Issue #193, opened 2024-04-02
`WeightedFutures` refactoring in the client (tracking issue)


giac-mysten:
An issue to track the possible reasons to refactor the `WeightedFutures` implementation in the client.

- It is too generic, and implementing their functionality case-by-case where needed may be better for recovery and performance ([comment](https://github.com/MystenLabs/walrus/pull/141/files#r1542786816)).
- It may be used in ways that are not safe, since the responsibility of all the checks (epochs, non-overlapping stake weights, etc.) is in the underlying futures, and not in the `WeightedFutures` ([comment](https://github.com/MystenLabs/walrus/pull/141/files#r1542764915)). 
- The `FuturesUnorderered`, on which this implementation is based, is executed on a single task. This should not be a problem, given that the CPU-intensive in the client is encoding and decoding the blob, _which is not performed within `WeightedFutures`_. However, it may be necessary at some point to split all the network requests into separate tasks. ([comment](https://github.com/MystenLabs/walrus/pull/141/files#r1542782670))
- The synchrony assumptions may need to be more fine-grained. These are anyways handled by the futures themselves, but the `WeightedFutures` could be too inflexible ([comment](https://github.com/MystenLabs/walrus/pull/141/files#r1543282703)).

Issue #542, opened 2024-06-17
Change URL path for recovery symbols


mlegner:
This is currently `/v1/blobs/:blob_id/slivers/:sliver_pair_index/:sliver_type/:target_pair_index` which is a bit confusing as it looks like a sliver endpoint even though it actually returns a recovery symbol. When looking at metrics and logs, having "recovery" somewhere in the path would be much clearer. For example, `/v1/blobs/:blob_id/slivers/:sliver_pair_index/:sliver_type/recovery/:target_pair_index` .

This would be a breaking change, but not client-facing.

PR #1407, opened 2025-01-14
feat!: refactor apis on storage node and daemon

jpcsmith:
## Description

This PR accomplishes the following:
- **Finalises the error response format.** The errors returned by the storage node and daemon APIs have been refactored to follow the [AIP #193](https://google.aip.dev/193) guidelines. This allows identifying errors in a machine-readable way without needing to compare error messages; attach metadata such as fields to error responses; and allows later adding more details to errors via the `details` array.
- **Adds a derive-macro for API errors.** The derive macro added in `walrus-proc-macros` adding the information needed to craft the error responses to error enums and structs, directly in the definition. Most notably, it allows using the docstring from the errors as the descriptions in the OpenAPI schema; which would otherwise be quite cumbersome with `macro_rules!`.
- **Makes REST API endpoints names consistent across daemon and storage node.** As per [AIP 122](https://google.aip.dev/122), more endpoints now follow the `/v1/{resource_name}/{resource_id}` convention; use nouns instead of verbs for their collection and resource names; and are in camelCase. 
- **Return consistent error types across daemon and storage node.** The daemon now also uses the same error format for its error responses as the storage node.
- **Fixes OpenAPI schemas.** This fixes the broken API schema for the storage node and also writes the schemas and rendered HTML for the storage node, daemon, client, and aggregator to the repo. The HTML files can be used to quickly view the endpoints and the saved schema can be used for checking for changes that render it invalid (TODO).

Contributes to #997.
Closes WAL-160.
Contributes to WAL-161.
Contributes to WAL-253.
Closes WAL-511.
Closes WAL-515.

**Remark:** These changes are heavily influenced by Google's API Improvement Proposals (AIPs). The AIPs condense a lot of general best practices regarding REST API designs in one location and is therefore a good reference for questions around implementing REST APIs, such as how to do versioning, long-running operations, batch jobs, etc.

~**Open question**: On the topic of versioning, we may benefit from already beginning our versioning practice with this testnet release by changing the `/v1` endpoints to `/v1beta`. This would make it a bit clearer that the APIs are still subject to change before mainnet, while at the same time allowing ourselves to refine and release a `/v1` set of endpoints at mainnet (which could just be the exact same endpoints as `/v1beta` if there are no further changes. We would support the `/v1beta` endpoints post-mainnet for some time in parallel with the `/v1` endpoints. See [AIP 185](https://google.aip.dev/185) for more on versioning.~

## Test plan

These changes are primarily refactoring the APIs, and are tested by the current unit and integration tests.

---

## Release notes

- [x] Storage node: Update error responses returned by storage nodes to be machine-readable and contain more details
- [x] Aggregator: Add error statuses to the responses of failed requests; update endpoint URLs; and add more detailed OpenAPI schemas.
- [x] Publisher: Add error statuses to the responses of failed requests; update endpoint URLs; and add more detailed OpenAPI schemas.
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/api/errors.rs
+++ b/crates/walrus-sdk/src/api/errors.rs
@@ -137,6 +137,11 @@ impl Status {
         self.inner.details.push(details);
         None
     }
+
+    /// Returns true if the error's reason and domain match those specified.
+    pub fn is_for_reason(&self, reason: &str, domain: &str) -> bool {
+        self.reason() == Some(reason) && self.domain() == Some(domain)
+    }
 }
 
 impl Display for Status {
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -593,10 +593,14 @@ async fn test_blocklist() -> TestResult {
         // sleep for a bit to allow the nodes to sync
         tokio::time::sleep(Duration::from_secs(30)).await;
     }
-    assert!(matches!(
-        blob_read_result.unwrap_err().kind(),
-        ClientErrorKind::BlobIdBlocked(_blob_id)
-    ));
+
+    let error = blob_read_result.expect_err("result must be an error");
+
+    assert!(
+        matches!(error.kind(), ClientErrorKind::BlobIdBlocked(_)),
+        "unexpected error {:?}",
+        error
+    );
 
     // Remove the blob from the blocklist
     for blocklist in blocklists.iter_mut() {
--- a/crates/walrus-sdk/src/error.rs
+++ b/crates/walrus-sdk/src/error.rs
@@ -43,16 +43,20 @@ impl NodeError {
         Some(StatusCode::NOT_FOUND) == self.http_status_code()
     }
 
-    /// Returns true if the HTTP error status code associated with the error is
-    /// [`StatusCode::FORBIDDEN`].
-    pub fn is_status_forbidden(&self) -> bool {
-        Some(StatusCode::FORBIDDEN) == self.http_status_code()
+    /// Returns true if the error is due to the blob being blocked.
+    pub fn is_blob_blocked(&self) -> bool {
+        // TODO(jsmith): use a constant shared between client and server.
+        self.status()
+            .map(|status| status.is_for_reason("FORBIDDEN_BLOB", STORAGE_NODE_ERROR_DOMAIN))
+            .unwrap_or(false)
     }
 
-    /// Returns true if the HTTP error status code associated with the error is
-    /// [`StatusCode::MISDIRECTED_REQUEST`].
+    /// Returns true if the error is due to the shard not being assigned to the storage node.
     pub fn is_shard_not_assigned(&self) -> bool {
-        self.http_status_code() == Some(StatusCode::MISDIRECTED_REQUEST)
+        // TODO(jsmith): use a constant shared between client and server.
+        self.status()
+            .map(|status| status.is_for_reason("SHARD_NOT_ASSIGNED", STORAGE_NODE_ERROR_DOMAIN))
+            .unwrap_or(false)
     }
 
     /// Wrap a standard error as a Node error.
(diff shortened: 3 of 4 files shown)


crates/walrus-proc-macros/src/derive_api_errors.rs
```diff
@@ -0,0 +1,498 @@
...
+
+                    match serde_json::to_value(&self) {
+                        Ok(serde_json::value::Value::Object(serialized)) => {
+                            error_info.metadata_mut().extend(serialized);
+                        },
+                        Ok(_) => {
+                            tracing::error!(?self, "errors must serialize to json objects");
+                        },
+                        Err(error) => {
+                            tracing::error!(?self, "serialization to json failed");
```
halfprice: Suggestion: also log the error here

crates/walrus-e2e-tests/Cargo.toml
```diff
@@ -22,7 +22,7 @@ tokio-stream.workspace = true
 tracing.workspace = true
 tracing-subscriber.workspace = true
 walrus-core.workspace = true
-walrus-proc-macros.workspace = true
+walrus-proc-macros = { workspace = true, features = ["walrus-simtest"] }
```
halfprice, line 25: Question: do we need to add `walrus-simtest` feature in walrus-proc-macros in `crates/walrus-e2e-tests/Cargo.toml` as well?

crates/walrus-sdk/src/api/errors.rs
```diff
@@ -0,0 +1,445 @@
...
+    message: String,
+
+    /// Machine readable details of the error.
+    ///
+    /// Always contains an [`ErrorInfo`], which provides a machine-readable
+    /// representation of the of the `message` field.
+    // The array of error details can be expanded to add further details such as localized messages,
+    // or details about where exactly the error occurred.
+    #[schema(value_type = Vec<Object>)]
+    details: Vec<ErrorDetails>,
```
halfprice, line 185: Question: can you explain a bit more on why this is a Vector? Are localized messages being inserted while populating the error up?

crates/walrus-service/src/node/errors.rs
```diff
@@ -122,34 +264,64 @@ impl From<RetrieveMetadataError> for InconsistencyProofError {
...
     #[error(transparent)]
+    #[rest_api_error(delegate)]
     Internal(#[from] InternalError),
 }
 
 /// Error returned when the epoch in a request is invalid.
-#[derive(Debug, thiserror::Error, Serialize, Clone)]
-#[error("Invalid epoch. Client epoch: {request_epoch}. Server epoch: {server_epoch}")]
+///
+/// This is a multiline
```
halfprice: Question: is this doc incomplete?

crates/walrus-sdk/src/api/errors.rs
jpcsmith: Expanded this into the following comment which should answer the question:
```
    // The details vector is where we can add additional, standardized extensions to the error
    // response. For example, for internal errors we add `ErrorDetails::DebugInfo` with the ID of
    // the current trace. Other possible "extensions" include region-localized error messages,
    // the duration after which to retry failed requests, details as to which fields in the request
    // are invalid and why, links to documentation, etc. The following link gives further examples
    // https://github.com/googleapis/googleapis/blob/master/google/rpc/error_details.proto
```

crates/walrus-service/src/node/errors.rs
jpcsmith: Thanks for catching this, it was an informal test for a longer comment. Removed it now.

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -68,7 +68,7 @@ pub const DELETABLE_BLOB_CONFIRMATION_ENDPOINT: &str =
 pub const RECOVERY_ENDPOINT: &str =
     "/v1/blobs/:blob_id/slivers/:sliver_pair_index/:sliver_type/:target_pair_index";
```
mlegner, line 69: Question: Is it intentional that this is unchanged? (#542)

jpcsmith: Yes. We'll be providing an endpoint to batch the symbol requests, I will use the new endpoint naming for that and deprecate this endpoint.

There was also a tropic of flattening the 2D symbol recovery index into 1D index (or at least a single field), which I want to flesh out first.

mlegner: OK, then I'm closing the related issue as not planned.

PR #1428, opened 2025-01-16
feat: add earliest_expiry to store, add expiry ms to blob-info

joyqvq:
## Description

Closes WAL-246.
Closes WAL-279.

## Test plan

tested with local testbed
```
target/release/walrus store ~/Desktop/old_10kb_random_file_2.bin --end-epoch 6
...
Path: /Users/joy/Desktop/old_10kb_random_file_2.bin
Blob ID: idwm21qlo2tCSedDf6bLeYs0Db0y8wDTj2p5AqmywPA
Sui object ID: 0x674dca9304d9d95f4544fc0d532ef5c6e78474e9e2b7f4e452b20d678db90bfe
Unencoded size: 10.0 KiB
Encoded size (including replicated metadata): 45.9 KiB
Cost (excluding gas): 26 FROST (storage was purchased, and a new blob object was registered)
Expiry epoch (exclusive): 6

target/release/walrus blob-status --blob-id idwm21qlo2tCSedDf6bLeYs0Db0y8wDTj2p5AqmywPA
...
There is a certified permanent Blob object for blob ID idwm21qlo2tCSedDf6bLeYs0Db0y8wDTj2p5AqmywPA.
Expiry epoch: 6
Estimated expiry timestamp: 2025-01-22T08:18:40.221+00:00
Related event: (tx: BHLDCCgScd55pREAkMtW7CQV9eCKb17v4y4BmteALFU, seq: 0)
Initially certified in epoch: 1

target/release/walrus store ~/Desktop/old_10kb_random_file_1.bin --earliest-expiry-time "2025-01-22 03:30:00"
...
Path: /Users/joy/Desktop/old_10kb_random_file_1.bin
Blob ID: kv9j_XA9CPEi3TLcZS-zge8RfpM18sFpKhs7P2pJhZ0
Sui object ID: 0xe2f50c42a4ee6c2ff7ee3073b762bacfd38dceb36ac4c6e0f94b15a70b1c5a77
Unencoded size: 10.0 KiB
Encoded size (including replicated metadata): 45.9 KiB
Cost (excluding gas): 6 FROST (storage was purchased, and a new blob object was registered)
Expiry epoch (exclusive): 2

target/release/walrus store ~/Desktop/old_10kb_random_file_3.bin --earliest-expiry-time "2025-01-20 03:30:00"
Error: earliest_expiry_time must be greater than the current epoch start time

target/release/walrus store ~/Desktop/old_10kb_random_file_4.bin --earliest-expiry-time "2026-01-20 03:30:00"
Error: blobs can only be stored for up to 104 epochs ahead; 8713 epochs were requested

target/release/walrus store ~/Desktop/old_10kb_random_file_2.bin --end-epoch 106
Error: blobs can only be stored for up to 104 epochs ahead; 105 epochs were requested
```
---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Add `--earliest-expiry-time` to `store` command so a unix timestamp in ms can be specified. Show human readable time in GMT for estimated expiry time in `blob-status` command.

### DIFF
--- a/crates/walrus-service/src/client/cli/runner.rs
+++ b/crates/walrus-service/src/client/cli/runner.rs
@@ -412,35 +412,47 @@ impl ClientCommandRunner {
         let system_object = client.sui_client().read_client.get_system_object().await?;
         let max_epochs_ahead = system_object.max_epochs_ahead();
 
-        let epochs_ahead = if let Some(epochs) = epoch_arg.epochs {
-            epochs.try_into_epoch_count(max_epochs_ahead)?
-        } else if let Some(earliest_expiry_time) = epoch_arg.earliest_expiry_time {
-            let staking_object = client.sui_client().read_client.get_staking_object().await?;
-            let epoch_state = staking_object.epoch_state();
-            let estimated_start_of_current_epoch = match epoch_state {
-                EpochState::EpochChangeDone(epoch_start)
-                | EpochState::NextParamsSelected(epoch_start) => *epoch_start,
-                EpochState::EpochChangeSync(_) => Utc::now(),
-            };
-            let earliest_expiry_ts = DateTime::from(earliest_expiry_time);
-            ensure!(
-                earliest_expiry_ts > estimated_start_of_current_epoch
-                    && earliest_expiry_ts > Utc::now(),
-                "earliest_expiry_time must be greater than the current epoch start time
+        let epochs_ahead = match epoch_arg {
+            EpochArg {
+                epochs: Some(epochs),
+                ..
+            } => epochs.try_into_epoch_count(max_epochs_ahead)?,
+            EpochArg {
+                earliest_expiry_time: Some(earliest_expiry_time),
+                ..
+            } => {
+                let staking_object = client.sui_client().read_client.get_staking_object().await?;
+                let epoch_state = staking_object.epoch_state();
+                let estimated_start_of_current_epoch = match epoch_state {
+                    EpochState::EpochChangeDone(epoch_start)
+                    | EpochState::NextParamsSelected(epoch_start) => *epoch_start,
+                    EpochState::EpochChangeSync(_) => Utc::now(),
+                };
+                let earliest_expiry_ts = DateTime::from(earliest_expiry_time);
+                ensure!(
+                    earliest_expiry_ts > estimated_start_of_current_epoch
+                        && earliest_expiry_ts > Utc::now(),
+                    "earliest_expiry_time must be greater than the current epoch start time
                 and the current time"
-            );
-            let delta =
-                (earliest_expiry_ts - estimated_start_of_current_epoch).num_milliseconds() as u64;
-            (delta / staking_object.epoch_duration() + 1) as u32
-        } else if let Some(end_epoch) = epoch_arg.end_epoch {
-            let current_epoch = client.sui_client().current_epoch().await?;
-            ensure!(
-                end_epoch > current_epoch,
-                "end_epoch must be greater than the current epoch"
-            );
-            (end_epoch - current_epoch) as u32
-        } else {
-            anyhow::bail!("either epochs or earliest_expiry_time must be provided");
+                );
+                let delta = (earliest_expiry_ts - estimated_start_of_current_epoch)
+                    .num_milliseconds() as u64;
+                (delta / staking_object.epoch_duration() + 1) as u32
+            }
+            EpochArg {
+                end_epoch: Some(end_epoch),
+                ..
+            } => {
+                let current_epoch = client.sui_client().current_epoch().await?;
+                ensure!(
+                    end_epoch > current_epoch,
+                    "end_epoch must be greater than the current epoch"
+                );
+                end_epoch - current_epoch
+            }
+            _ => {
+                anyhow::bail!("either epochs or earliest_expiry_time or end_epoch must be provided")
+            }
         };
 
         // Check that the number of epochs is lower than the number of epochs the blob can be stored


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -168,8 +168,11 @@ pub enum CliCommands {
         #[serde(deserialize_with = "crate::utils::resolve_home_dir_vec")]
         files: Vec<PathBuf>,
         /// The number of epochs ahead for which to store the blob.
-        #[clap(long)]
-        epochs: EpochCount,
+        #[clap(long, conflicts_with = "earliest_expiry_time")]
+        epochs: Option<EpochCount>,
+        /// The earliest time (in milliseconds since UNIX epoch) when the blob can expire.
+        #[clap(long, conflicts_with = "epochs")]
+        earliest_expiry_time: Option<u64>,
```
mlegner: Suggestion: As in #1406, use an `ArgGroup` here.

mlegner: Suggestion: Use [humantime](https://docs.rs/humantime/latest/humantime/) to parse timestamps in various user-friendly formats.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -507,10 +538,36 @@ impl ClientCommandRunner {
         let status = client
             .get_verified_blob_status(&blob_id, &sui_read_client, timeout)
             .await?;
+
+        // Compute estimated blob expiry in DateTime if it is a permanent blob.
+        let estimated_expiry_timestamp = match status {
+            BlobStatus::Permanent { end_epoch, .. } => {
```
mlegner: Nit: Given we only care about the `Permanent` status, the `match status` could also become an `if let BlobStatus::Permanent { end_epoch, ..}`.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -507,10 +538,36 @@ impl ClientCommandRunner {
...
+                let params = sui_read_client.fixed_system_parameters().await?;
+                let current_epoch = sui_read_client.current_epoch().await?;
+                let epoch_state = sui_read_client.epoch_state().await?;
```
mlegner: Important: This currently causes multiple RPC requests to the fullnodes. It might be good to solve this in a separate PR within the `SuiReadClient` though.

@liquid-helium is currently working on reducing the number of RPC calls for the `walrus info` command. Maybe we could combine these efforts?

cc @karlwuest @giac-mysten

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -507,10 +538,36 @@ impl ClientCommandRunner {
...
+
+                let estimated_start_of_current_epoch = match epoch_state {
+                    EpochState::EpochChangeDone(epoch) | EpochState::NextParamsSelected(epoch) => {
```
mlegner: Nit: The variable naming is a bit misleading here. Can we replace `epoch` by `epoch_start` or `timestamp`?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -277,6 +279,22 @@ impl SuiReadClient {
...
+            .read_api()
+            .get_object_with_options(SUI_CLOCK_OBJECT_ID, SuiObjectDataOptions::full_content())
+            .await?;
+        let bytes = response
+            .move_object_bcs()
+            .expect("Clock must be a Move object");
+        let clock: Clock =
+            bcs::from_bytes(bytes).expect("Clock object deserialization cannot fail");
+        Ok(clock.timestamp_ms)
+    }
```
mlegner: Question: Why do we need this? Can't we just get the current time with `DateTime`?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -507,10 +538,36 @@ impl ClientCommandRunner {
...
+                    }
+                    EpochState::EpochChangeSync(_) => {
+                        let timestamp = sui_read_client.current_clock_timestamp().await?;
+                        DateTime::from_timestamp_millis(timestamp as i64)
+                            .expect("timestamp must be valid")
+                    }
+                };
+                Some(
+                    estimated_start_of_current_epoch
+                        + params.epoch_duration * (end_epoch - current_epoch),
```
mlegner: Important: We should probably check somewhere that `end_epoch > current_epoch`. This *should* be the case as long as nodes behave correctly, but better to return an error than panicking.

crates/walrus-service/src/client/cli/args.rs
mlegner: Also, while you're at it, could you also add an `end_epoch` as an option?

crates/walrus-service/src/client/cli/runner.rs
joyqvq: do you think it's worth adding to the read client api just for this one use case?

should we just expose `get_staking_object` and `get_system_object` from pub(crate) to pub already? 

alternatively, if there is a pr to do this im happy to rebase

crates/walrus-service/src/client/cli/runner.rs
joyqvq: sure, i still need to handle the else branch to evaluate to None, but fine by me to use if else vs match. revised.

crates/walrus-sui/src/client/read_client.rs
joyqvq: probably safe just to use system clock here. revised.

crates/walrus-service/src/client/cli/runner.rs
mlegner: I think we can leave it for now or make the `get_staking_object` and `get_system_object` public. 👍

joyqvq: revised - had to add a few getters for inner value, but this should save some rpc calls.

crates/walrus-service/src/client/cli/args.rs
joyqvq: revised! now its using parse_rfc3339 from humantime

joyqvq: added!

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -172,13 +172,12 @@ pub enum CliCommands {
         #[clap(required = true, value_name = "FILES")]
         #[serde(deserialize_with = "crate::utils::resolve_home_dir_vec")]
         files: Vec<PathBuf>,
-        /// The number of epochs ahead for which to store the blob.
+        /// The epoch argument to specify either the number of epochs to store the blob, or the
+        /// end epoch, or the earliest expiry in UNIX timestamp ms.
         ///
```
mlegner: Nit: This no longer uses a UNIX timestamp.
```suggestion
        /// The epoch argument to specify either the number of epochs to store the blob, or the
        /// end epoch, or an earliest expiry time.
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -823,6 +822,44 @@ impl EpochCountOrMax {
     }
 }
 
+/// The number of epochs to store the blob for.
+#[serde_as]
+#[derive(Debug, Clone, Args, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct EpochArg {
```
mlegner, line 926: Suggestion to have clap check that exactly one of the options is set.
```suggestion
#[group(required = true, multiple = false)]
pub struct EpochArg {
```

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -126,7 +126,8 @@ impl CliOutput for BlobStoreResultWithPath {
                     Sui object ID: {}\n\
                     Unencoded size: {}\n\
                     Encoded size (including replicated metadata): {}\n\
-                    Cost (excluding gas): {} {}{}",
+                    Cost (excluding gas): {} {} \n\
+                    Expiry epoch (exclusive): {}{}\n",
```
mlegner, line 132: Nit: Similar to the blob status, could we also report the earliest expiry time here?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -378,29 +382,67 @@ impl ClientCommandRunner {
...
         post_store: PostStoreAction,
     ) -> Result<()> {
+        epoch_arg.exactly_one_is_some()?;
+
         let client = get_contract_client(self.config?, self.wallet, self.gas_budget, &None).await?;
 
+        let system_object = client.sui_client().read_client.get_system_object().await?;
+        let max_epochs_ahead = system_object.max_epochs_ahead();
+
+        let epochs_ahead = if let Some(epochs) = epoch_arg.epochs {
```
mlegner: Nit: Given we have three cases here now, a `match` statement would be more readable.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -378,29 +382,67 @@ impl ClientCommandRunner {
...
+            (delta / staking_object.epoch_duration() + 1) as u32
+        } else if let Some(end_epoch) = epoch_arg.end_epoch {
+            let current_epoch = client.sui_client().current_epoch().await?;
+            ensure!(
+                end_epoch > current_epoch,
+                "end_epoch must be greater than the current epoch"
+            );
+            (end_epoch - current_epoch) as u32
+        } else {
+            anyhow::bail!("either epochs or earliest_expiry_time must be provided");
```
mlegner: Nit: I think this is actually `unreachable` because we have the `exactly_one_is_some` check above.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -378,29 +382,67 @@ impl ClientCommandRunner {
...
+            let delta =
+                (earliest_expiry_ts - estimated_start_of_current_epoch).num_milliseconds() as u64;
+            (delta / staking_object.epoch_duration() + 1) as u32
+        } else if let Some(end_epoch) = epoch_arg.end_epoch {
+            let current_epoch = client.sui_client().current_epoch().await?;
+            ensure!(
+                end_epoch > current_epoch,
+                "end_epoch must be greater than the current epoch"
+            );
+            (end_epoch - current_epoch) as u32
```
mlegner: Nit: The conversion is not necessary here.
```suggestion
            end_epoch - current_epoch
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -823,6 +822,44 @@ impl EpochCountOrMax {
...
+    /// The number of epochs the blob is stored for.
+    ///
+    /// If set to `max`, the blob is stored for the maximum number of epochs allowed by the
+    /// system object on chain. Otherwise, the blob is stored for the specified number of
+    /// epochs. The number of epochs must be greater than 0.
+    #[clap(short, long, value_parser = EpochCountOrMax::parse_epoch_count)]
```
mlegner: Remark: I thought I had removed the "short" version here at some point to avoid confusion. Not sure where this was added again. 🤔 

Suggestion: Remove this as all the options here could conceivably be shortened to `-e`.
```suggestion
    #[clap(long, value_parser = EpochCountOrMax::parse_epoch_count)]
```

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -378,29 +382,67 @@ impl ClientCommandRunner {
...
+            let estimated_start_of_current_epoch = match epoch_state {
+                EpochState::EpochChangeDone(epoch_start)
+                | EpochState::NextParamsSelected(epoch_start) => *epoch_start,
+                EpochState::EpochChangeSync(_) => Utc::now(),
+            };
+            let earliest_expiry_ts = DateTime::from(earliest_expiry_time);
+            ensure!(
+                earliest_expiry_ts > estimated_start_of_current_epoch,
+                "earliest_expiry_time must be greater than the current epoch start time"
+            );
```
mlegner: Suggestion: Also check that this is in the future and return an error otherwise.

crates/walrus-service/src/client/cli/cli_output.rs
joyqvq: this was intentional that the earliest expiry time is not shown here - since 1) the blob struct here doest have everything i need 2) computing this requires a few rpc calls (also needs to instantiate a read client that is not available here). 

i could though, rewrite the BlobStoreResult a bit to make the `BlobStatus` available there and add fetch blob status, epoch state etc. do we want to do this / worth the effort? or the user should just use a second command for blob-status to opt in for the additional info?

crates/walrus-service/src/client/cli/runner.rs
joyqvq: you mean, changing `struct EpochArg` into `enum EpochArg`...? or you mean doing something like

```
match epoch_args.epochs {
    Some(epoch) => do something,
    None => match epoch_args.earliest_expiry_time {
                      Some(earliest_expiry_time) => do something, 
                      None => match epoch_args.end_epoch {
                                          Some(end_epoch) => do something, 
                                          None => anyhow::bail}  
                        }
}
```

i think the nested match branch looks more ugly, no?

crates/walrus-service/src/client/cli/runner.rs
joyqvq: since `EpochCount` from the first branch is a u32, these should be casted down after doing the computation in u64.

crates/walrus-service/src/client/cli/runner.rs
mlegner: I meant something like the following:
```rust
        match epoch_arg {
            EpochArg{epochs: Some(epochs), ..} => {}
            EpochArg{earliest_expiry_time: Some(earliest_expiry_time), ..} => {}
            EpochArg{end_epoch: Some(end_epoch), ..} => {}
        }
```

crates/walrus-service/src/client/cli/runner.rs
mlegner: I think you need the cast for the second branch but not this one as both `end_epoch` and `current_epoch` are already `u32`s.

crates/walrus-service/src/client/cli/cli_output.rs
mlegner: You're right, this is probably not worth the extra effort. Let's keep it as it is. 👍

PR #1433, opened 2025-01-16
chore: avoid usize where u64 is adequate and improves legibility

wbbradley:
## Description

This is a code hygiene PR. I believe it is better to rely on the type system rather than utilizing machine-specific pointer/integer size checks at runtime. I did not see a clear reason that usize should be preferred in any of the affected locations. If we move to u32 or u128 in the future, this change will be compatible with such a refactoring by way of simple type checking.

## Test plan

CI Pipeline.

No release impact.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -610,23 +610,20 @@ impl StorageNode {
         event_blob_writer: &mut Option<EventBlobWriter>,
     ) -> anyhow::Result<(
         Pin<Box<dyn Stream<Item = PositionedStreamEvent> + Send + Sync + '_>>,
-        usize,
+        u64,
     )> {
         let event_cursor = std::cmp::min(storage_node_cursor, event_blob_writer_cursor);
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -4730,15 +4726,15 @@ mod tests {
...
+                .update_blob_info(index as u64, event)?;
         }
         let intermediate_blob_info = node.storage_node.inner.storage.get_blob_info(&BLOB_ID)?;
 
         for (index, event) in repeated_events.iter().enumerate() {
             node.storage_node
                 .inner
                 .storage
-                .update_blob_info(index + count_setup_events, event)?;
+                .update_blob_info(index as u64 + count_setup_events, event)?;
```
halfprice, line 4737: `update_blob_info` probably can also just take u64 as well.

wbbradley: Yep, it does. This test is using an index via enumerate, so it needs to be cast to u64.

PR #1435, opened 2025-01-17
feat: [jh/use walrus-sui instead of raw json queries]

suiwombat:
## Description
* convert to walrus-sui client and stop using the raw json queries

## Test Plan
```
2025-01-17T18:33:22.600323Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[silent] network_address: [storage.walrus.silentvalidator.com:9185]
2025-01-17T18:33:22.600330Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[latitude-storage-testnet] network_address: [walrus-storage-testnet.latitude-sui.com:9185]
2025-01-17T18:33:22.600337Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Nodeinfra] network_address: [walrus-testnet-storage.nodeinfra.com:9185]
2025-01-17T18:33:22.600343Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Cosmostation] network_address: [walrus-testnet.cosmostation.io:9185]
2025-01-17T18:33:22.600350Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Astro-Stakers] network_address: [walrus.astrostakers.com:9185]
2025-01-17T18:33:22.600357Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Studio Mirai] network_address: [storage.testnet.walrus.mirainfra.com:9185]
2025-01-17T18:33:22.600363Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Staking4All] network_address: [walrus-testnet.staking4all.org:9185]
2025-01-17T18:33:22.600370Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[ZKV] network_address: [walrus-testnet-storage-node-1.zkv.xyz:9185]
2025-01-17T18:33:22.600376Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[natsai] network_address: [walrus-testnet-node.natsai.xyz:9185]
2025-01-17T18:33:22.600383Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[zeroservices.eu - EYL1] network_address: [eyl1.walrus.zeroservices.eu:9185]
2025-01-17T18:33:22.600390Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[LakeStake] network_address: [walrus.lakestake.io:9185]
2025-01-17T18:33:22.600397Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Global Stake] network_address: [walrus.globalstake.io:9185]
2025-01-17T18:33:22.600403Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Piconbello] network_address: [37.27.141.109:9185]
2025-01-17T18:33:22.600410Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Staking Defense League] network_address: [walrus-testnet.stakingdefenseleague.com:9185]
2025-01-17T18:33:22.600416Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Imperator.co] network_address: [walrus-testnet.imperator.co:9185]
2025-01-17T18:33:22.600423Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Krates] network_address: [walrus.krates.ai:9185]
2025-01-17T18:33:22.600430Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Everstake] network_address: [walrus-testnet.everstake.one:9185]
2025-01-17T18:33:22.600437Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[SenseiNode] network_address: [walrus-testnet.senseinode.com:9185]
2025-01-17T18:33:22.600443Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[kubenode] network_address: [walrus01.kubenode.xyz:9185]
2025-01-17T18:33:22.600450Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[overclock-testnet-storage] network_address: [walrus-storage-testnet.overclock.run:9185]
2025-01-17T18:33:22.600457Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[MozComputing] network_address: [walrus-node-01.mozcomputing.dev:9185]
2025-01-17T18:33:22.600464Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[ChainodeTech] network_address: [walrus-testnet.chainode.tech:9185]
2025-01-17T18:33:22.600471Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[pinata-bot] network_address: [16.16.120.87:9185]
2025-01-17T18:33:22.600477Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Blockscope] network_address: [walrus-testnet.blockscope.net:9185]
2025-01-17T18:33:22.600484Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Mysten Labs Baremetal] network_address: [mia-tnt-sto-00.walrus-testnet.walrus.space:9185]
2025-01-17T18:33:22.600491Z  INFO walrus_proxy::providers::walrus::provider: loaded node:[Luganodes] network_address: [walrus-testnet.lgns.xyz:9185]
2025-01-17T18:33:22.604811Z  INFO walrus_proxy::providers::walrus::provider: 58 walrus nodes managed to make it on the allow list
```

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-proxy/src/providers/walrus/query.rs
+++ b/crates/walrus-proxy/src/providers/walrus/query.rs
@@ -1,11 +1,7 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{
-    collections::HashSet,
-    hash::Hash,
-    vec,
-};
+use std::{collections::HashSet, hash::Hash, vec};
 
 use anyhow::Error;
 use fastcrypto::traits::ToFromBytes;


crates/walrus-proxy/src/providers/walrus/query.rs
```diff
@@ -227,86 +32,70 @@ pub struct NodeInfo {
...
-        .header("Content-Type", "application/json")
-        .json(&body)
-        .send()
-        .await?;
-    if !response.status().is_success() {
-        let error_message = response.text().await?;
-        return Err(anyhow::anyhow!("query call failed; {error_message}"));
+impl Hash for NodeInfo {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.network_address.hash(state);
```
karlwuest: Question: Why do we only hash the network address here?

crates/walrus-proxy/src/providers/walrus/query.rs
```diff
@@ -227,86 +32,70 @@ pub struct NodeInfo {
...
+}
 
-    let network_public_key: Vec<u8> = network_public_key
-        .iter()
-        // somewhat difficult to extract u8
-        .filter_map(|v| v.as_u64().map(|num| num as u8))
-        .collect();
+impl PartialEq for NodeInfo {
+    fn eq(&self, other: &Self) -> bool {
+        self.network_address == other.network_address
```
karlwuest: Same question as above: why do we only care about the network address?

crates/walrus-proxy/src/providers/walrus/query.rs
```diff
@@ -227,86 +32,70 @@ pub struct NodeInfo {
...
 
-    Ok(NodeInfo {
-        name: name.to_string(),
-        network_address: network_address.to_string(),
-        network_public_key: network_public_key.to_vec(),
-    })
+impl Eq for NodeInfo {}
+
+/// merge the Committee types into a hashset of NodeInfo and then return a vec of it.
+/// we use previous, current and next epoch members and return a unique vec of them
```
karlwuest: Nit:
```suggestion
/// Merges the Committee types into a hashset of NodeInfo and then returns a vec of it.
/// We use previous, current and next epoch members and return a unique vec of them.
```

crates/walrus-proxy/src/config.rs
```diff
@@ -55,6 +55,8 @@ pub struct DynamicPeerValidationConfig {
     /// the interval we will update our peer cache
     #[serde_as(as = "DurationSeconds<u64>")]
     pub interval: Duration,
+    /// system object id to query for current nodes
+    pub system_object_id: String,
     /// staking object id to query for current nodes
     pub staking_object_id: String,
```
karlwuest, line 61: Nit: You can use the contract config directly here, then you don't need to manually convert the strings from the config into `ObjectId`s and manually construct the `ContractConfig`.

```suggestion
    #[serde(flatten)]
    /// Walrus contract config used to query for current nodes.
    pub contract_config: ContractConfig,
```

crates/walrus-proxy/src/providers/walrus/query.rs
```diff
@@ -4,220 +4,25 @@
 // TODO: Reuse calls from `walrus-sui` here (#1170).
 // TODO: Include the previous and next committees (#1174).
```
karlwuest: Suggestion: I think these TODOs can be removed with this PR?

crates/walrus-proxy/src/providers/walrus/query.rs
```diff
@@ -227,86 +32,70 @@ pub struct NodeInfo {
...
+        })
+        .collect::<HashSet<_>>()
+        .into_iter()
+        .collect()
 }
 
-/// get_walrus_committee is the entry point for pub callers to get the set of
-/// committee members
+/// public entry point to get walrus committee data for the proxy
 pub async fn get_walrus_committee(
```
karlwuest: Nit: Since this doesn't just return the current committee, perhaps `get_walrus_nodes` may be a better name.

crates/walrus-proxy/src/providers/walrus/query.rs
```diff
@@ -227,86 +32,70 @@ pub struct NodeInfo {
...
+            network_public_key: storage_node.network_public_key.as_bytes().to_vec(),
+        })
+        .collect::<HashSet<_>>()
+        .into_iter()
+        .collect()
 }
 
-/// get_walrus_committee is the entry point for pub callers to get the set of
-/// committee members
+/// public entry point to get walrus committee data for the proxy
```
karlwuest: Nit:
```suggestion
/// Returns the [`NodeInfo`] for all nodes that are part of the last, current,
/// or next Walrus committee.
```

crates/walrus-proxy/src/providers/walrus/query.rs
```diff
@@ -227,86 +32,70 @@ pub struct NodeInfo {
...
-            "unable to get any committee objects; \
-the node cache won't be updated until this call succeeds"
-        ));
-    }
+    let contract_config = ContractConfig::new(
+        ObjectID::from_hex_literal(system_object_id)?,
+        ObjectID::from_hex_literal(staking_object_id)?,
+    );
+    let backoff_config = ExponentialBackoffConfig::default();
+    let c = walrus_sui::client::SuiReadClient::new_for_rpc(
```
karlwuest, line 65: Nit: Instead of constructing the `SuiReadClient` here, you could construct it in `WalrusNodeProvider::new` and store it there as a field (instead of storing the config values), then this function could just be a method on `WalrusNodeProvider` and doesn't need to construct a new `SuiReadClient` every time you poll. This would save a few RPC requests.

crates/walrus-proxy/src/providers/walrus/query.rs
suiwombat: i didn't see a need for a compund key. the other one we could hash is the pub key, which may make more sense, but i can see a case for either.  would you prefer one over the other?  I'm happy to use another hash strategy.

crates/walrus-proxy/src/providers/walrus/query.rs
suiwombat: I've found that long running client connections eventually fail because of legit regions that are difficult to debug.  If you want me to switch to a long running one i can, but just be aware that if the underlying connection misbehaves and the client implementation isn't robust enough, we can get dead client states that suck to debug and require binary restarts to correct until patched.  spinning a client up each time guarantees we have a good client each time, assuming everything else is working.

crates/walrus-proxy/src/providers/walrus/query.rs
karlwuest: I was mainly just curious, I don't think it matters too much for the use case.

With the network address, we don't necessarily get uniqueness, because a (malicious) node could set the same network address as another node. The cleanest way would probably be to use the node ID, because that is persistent, cannot be changed by a node, and is guaranteed to be unique.

crates/walrus-proxy/src/providers/walrus/query.rs
karlwuest: I don't think that's really an issue, because the client will just reconnect, but it's also fine to leave it as is. I don't know what the polling frequency is, but I'm guessing sending a few more RPC requests doesn't really have a significant impact.

crates/walrus-proxy/src/config.rs
suiwombat: I could yes, but then i create a code dep that doesn't seem necessary.  If ContractConfig is ever refactored, it's going to potentially cause a problem here.  I would prefer to leave these as they are, even if it could be done more compactly.  changing this would also cascasde over into the pulumi config, which i don't want to change haha

crates/walrus-proxy/src/providers/walrus/query.rs
suiwombat: on second thought, i'll make this more robust. it's fair to question it.

PR #1438, opened 2025-01-19
feat:  retry the exact same Sui transaction when encounter retriable error

halfprice:
## Description

When a contract is experiencing retryable error, it's unclear whether this transaction has locked object on chain or not. To reduce the chance of equivocation and locked object, we should retry the exact same transaction instead of building a new one.

The main motivation is that we have seen the epoch_sync_done transaction equivocated several times, which blocks the epoch change progress. Although this is less of a concern on mainnet since the epoch on mainnet is much longer than sui epoch, and locked objects can be unlocked, we still don't want epoch_sync_done transaction to take a day to commit.

This PR only targets at transactions that perform retry currently.

Close WAL-494

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -594,7 +594,9 @@ fn maybe_return_injected_error_in_stake_pool_transaction(
 
         // Simulate a retriable RPC error (502 Bad Gateway)
         Err(sui_sdk::error::Error::RpcError(
-            anyhow::anyhow!("Server returned an error status code: 502").into(),
+            jsonrpsee::core::ClientError::Custom(
+                "Server returned an error status code: 502".into(),
+            ),
         ))?;
     }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3564,7 +3564,7 @@ dependencies = [
  "hashers",
  "http 0.2.12",
  "instant",
- "jsonwebtoken",
+ "jsonwebtoken 8.3.0",
  "once_cell",
  "pin-project",
  "reqwest 0.11.27",
@@ -5505,13 +5505,27 @@ dependencies = [
 
 [[package]]
 name = "jsonwebtoken"
+version = "8.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6971da4d9c3aa03c3d8f3ff0f4155b534aad021292003895a469716b2a230378"
+dependencies = [
+ "base64 0.21.7",
+ "pem 1.1.1",
+ "ring 0.16.20",
+ "serde",
+ "serde_json",
+ "simple_asn1",
+]
+
+[[package]]
+name = "jsonwebtoken"
 version = "9.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b9ae10193d25051e74945f1ea2d0b42e03cc3b890f7e4cc5faa44997d808193f"
 dependencies = [
  "base64 0.21.7",
  "js-sys",
- "pem",
+ "pem 3.0.4",
  "ring 0.17.8",
  "serde",
  "serde_json",
@@ -14136,14 +14150,14 @@ dependencies = [
  "git-version",
  "hex",
  "home",
- "http-body-util",
  "hostname",
+ "http-body-util",
  "humantime",
  "indicatif",
  "indoc",
  "integer-encoding 4.0.2",
  "itertools 0.13.0",
- "jsonwebtoken",
+ "jsonwebtoken 9.3.0",
  "mime",
  "mockall 0.12.1",
  "move-core-types",
@@ -14252,6 +14266,7 @@ dependencies = [
  "fastcrypto 0.1.8 (git+https://github.com/MystenLabs/fastcrypto?rev=2f502fd8570fe4e9cff36eea5bbd6fef22002898)",
  "futures",
  "inflections",
+ "jsonrpsee",
  "move-core-types",
  "rand 0.8.5",
  "regex",
--- a/crates/walrus-sui/Cargo.toml
+++ b/crates/walrus-sui/Cargo.toml
@@ -60,4 +60,5 @@ walkdir = "2.5.0"
 workspace = true
 
 [target.'cfg(msim)'.dependencies]
+jsonrpsee = { version = "0.24" }
 sui-simulator.workspace = true


.github/workflows/code.yml
```diff
@@ -105,6 +105,9 @@ jobs:
     needs: diff
     if: ${{ needs.diff.outputs.isRelevantForRustTests == 'true' }}
     runs-on: ubuntu-ghcloud
+    env:
+      RUST_LOG: warn
+      RUST_MIN_STACK: 67108864
```
halfprice: Hi @karlwuest , please advise if this is a good idea. This PR makes all e2e test to have stack overflow in CI during contract publishing steps, however it does not touch any move related code.

crates/walrus-sui/src/client.rs
```diff
@@ -345,25 +368,25 @@ impl SuiContractClient {
...
-                .to_move_struct_tag_with_type_map(&self.read_client.type_origin_map(), &[])?,
-        )?;
-
-        ensure!(
-            storage_id.len() == 1,
-            "unexpected number of storage resources created: {}",
-            storage_id.len()
-        );
-        self.sui_client().get_sui_object(storage_id[0]).await
+        self.with_wallet_critical_section(|wallet| async move {
```
halfprice, line 365: All of these changes with `with_wallet_critical_section` was originated by moving the wallet into `RetryPtbExecutor`. But later, I moved the wallet back to the `SuiContractClient`. I left `with_wallet_critical_section` in the PR because I feel it makes the critical section clearer, but I don't mind to revert the code to the original struct.

.github/workflows/code.yml
karlwuest: I have no clue why there is a stack overflow without this. Locally, I cannot reproduce it, i.e. `cargo nextest run --features test-utils --run-ignored all` runs without issue on the latest main. Do you have a link to a failing run (ideally with debug output)?

I don't think increasing the stack size should be an issue as a temporary fix. I plan to both refactor the deployment code and remove `sui` as a dependency ([WAL-518](https://linear.app/mysten-labs/issue/WAL-518/refactor-contract-deployment-code)), but possibly only after mainnet since it's not high priority. Hopefully, that will make this unnecessary again.

crates/walrus-service/src/common/config.rs
```diff
@@ -39,7 +39,7 @@ pub struct SuiConfig {
     /// Location of the wallet config.
     #[serde(deserialize_with = "utils::resolve_home_dir")]
     pub wallet_config: PathBuf,
-    /// The configuration for the backoff strategy used for retries.
+    /// The configuration for the backoff strategy used for logic transaction retries.
```
karlwuest: Nit: revert this change. We still use the backoff config for read retries as well.

crates/walrus-sui/src/client.rs
karlwuest: I really like making this more explicit!

I had discussed this a couple of times with @giac-mysten, and I think it would be even nicer to do the following:
Instead of storing a `Mutex<WalletContext>`, we could store a `Mutex<SuiContractClientInner>`, where `SuiContractClientInner` stores the wallet and has all methods that send transactions, but takes `&mut self` for those. This way, we can enforce that critical sections always need to lock, but still allow `SuiContractClient` to operate on `&self`. What do you think about this?

This could then also take care of the transaction signing and leave the RPC retries to the `RetriableSuiClient` (see comment on `RetryPtbExecutor` below).

Maybe for this PR, we can keep the `with_wallet_critical_section` but still make the changes that I suggested on the `RetryPtbExecutor` and then separate out the `inner` in a follow up if it requires too much refactoring.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -92,7 +92,7 @@ impl RetriableRpcError for SuiClientError {
 }
 
 /// Retries the given function while it returns retriable errors.[
-async fn retry_rpc_errors<S, F, T, E, Fut>(mut strategy: S, mut func: F) -> Result<T, E>
+pub async fn retry_rpc_errors<S, F, T, E, Fut>(mut strategy: S, mut func: F) -> Result<T, E>
```
karlwuest: Nit: 
```suggestion
pub(crate) async fn retry_rpc_errors<S, F, T, E, Fut>(mut strategy: S, mut func: F) -> Result<T, E>
```

crates/walrus-sui/src/client/retry_ptb_executor.rs
```diff
@@ -0,0 +1,115 @@
...
+    base_types::{ObjectRef, SuiAddress},
+    transaction::{ProgrammableTransaction, TransactionData},
+};
+use walrus_utils::backoff::{ExponentialBackoff, ExponentialBackoffConfig};
+
+use super::retry_client::retry_rpc_errors;
+
+/// PTB executor that executes and retries upon receiving a retriable error.
+#[derive(Debug)]
+pub struct RetryPtbExecutor {
```
karlwuest: Suggestion:

Quoting from your comment on the main thread for context:

>We could add sign_and_sent_ptb function with in the RetriableSuiClient and requires the function to have input wallet. However, since ultimately what we want to reuse is the retry_rpc_errors function which is outside of the RetriableSuiClient, I feel doing it this way will only hurt the functionality and readability of RetriableSuiClient.

I agree with this and I understand why you added this functionality in a new struct, in particular because we currently need the wallet to send the transaction. 

However, I think it may be best to keep the signing in the `SuiContractClient` (or `SuiContractClientInner` once added) and then do the actual RPC retries in the `RetriableSuiClient`. We can do this without passing the wallet, if we use the [QuorumDriverApi](https://mystenlabs.github.io/sui/sui_sdk/apis/struct.QuorumDriverApi.html) directly instead of `wallet.execute_transaction_may_fail` and then pass an already signed transaction. I think this would keep the interfaces cleaner, because then all RPC functionality for Sui is contained in the `RetriableSuiClient`.

crates/walrus-service/src/common/config.rs
halfprice: Yeah, sorry about that. I added another backoff config initially and reverted that change.

crates/walrus-sui/src/client.rs
```diff
@@ -269,6 +269,8 @@ pub struct SuiContractClient {
     wallet: Mutex<WalletContext>,
     /// Client to read Walrus on-chain state.
     pub read_client: SuiReadClient,
```
halfprice, line 271: @karlwuest , given the new change in this PR which we add txn execution in the RetriableSuiClient, it probably makes more sense to rename this to something like `sui_client`, since it also does transaction execution now? If so,  I can do it in a follow up PR.

crates/walrus-sui/src/client.rs
halfprice: > Instead of storing a Mutex<WalletContext>, we could store a Mutex<SuiContractClientInner>, where SuiContractClientInner stores the wallet and has all methods that send transactions, but takes &mut self for those. This way, we can enforce that critical sections always need to lock, but still allow SuiContractClient to operate on &self. What do you think about this?

Yeah, this sounds better! I didn't have too much time today to refactor this and I'm not sure how much change this involves (could be quite a few). Let me send a follow up PR about this.

.github/workflows/code.yml
halfprice: This is weird... I believe @liquid-helium also ran into the similar issue in his PR which also doesn't touch contract at all. I asked @liquid-helium to increased the min stack size in his PR so that he can merge it. We may want to look into this if this becomes a problem.

crates/walrus-sui/src/client/retry_client.rs
halfprice: Reverted the change since now this is only used in this file.

crates/walrus-sui/src/client/retry_ptb_executor.rs
halfprice: This SGTM, and I can see why keeping tx signing in SuiContractClient is a good idea (since it's about executing a tx). This brings a little bit inconvenience as now the `read_client` also does tx execution, and we may want to rename it (as I mentioned in my other replies). Other than this, I've incorporate this comment, moved signing into `SuiContractClient::sign_and_send_ptb`, and create `execute_transaction` in RetriableSuiClient. PTAL!

crates/walrus-sui/src/client.rs
karlwuest: I don't think this needs to be renamed, because we're not executing in the `SuiReadClient`, but in the `RetriableSuiClient` that is returned by the read client. If we rename it to `sui_client`, it is more likely to be confused with a `RetriableSuiClient`.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -520,8 +528,75 @@ impl RetriableSuiClient {
...
+#[cfg(msim)]
+fn maybe_return_injected_error_in_stake_pool_transaction(
+    transaction: &Transaction,
+) -> anyhow::Result<()> {
+    // Check if this transaction contains a stake_with_pool operation
+    let is_stake_pool_tx = transaction
+        .transaction_data()
+        .move_calls()
+        .iter()
+        .any(|(_, _, function_name)| *function_name == "stake_with_pool");
```
karlwuest: Nit:
```suggestion
        .any(|(_, _, function_name)| function_name == contracts::staking::stake_with_pool.name);
```

halfprice: Done

PR #1446, opened 2025-01-20
refactor: make sliver requests parallel

jpcsmith:
## Description

This uses a separate tokio task for the recovery of each sliver pair, therefore allowing them to be distributed across several threads. Existing semaphore limits still apply.

Depends on #1443

## Test plan

This is a refactor, existing integration and unit tests apply.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/blob_sync.rs
+++ b/crates/walrus-service/src/node/blob_sync.rs
@@ -301,6 +301,7 @@ impl BlobSyncHandler {
     ) -> Option<EventHandle> {
         let queued_gauge =
             metrics::with_label!(self.node.metrics.recover_blob_backlog, STATUS_QUEUED);
+        let start = tokio::time::Instant::now();
         let _permit = permits
             .blob
             .acquire_owned()
@@ -310,7 +311,6 @@ impl BlobSyncHandler {
 
         let in_progress_gauge =
             metrics::with_label!(self.node.metrics.recover_blob_backlog, STATUS_IN_PROGRESS);
-        let start = tokio::time::Instant::now();
 
         let _decrement_guard = GaugeGuard::acquire(&in_progress_gauge);
         let blob_id = synchronizer.blob_id;
@@ -501,12 +501,15 @@ impl BlobSynchronizer {
                                 // abort the task, therefore should never happen.
                                 //
                                 // Returning from this function as if the task was cancelled
-                                // would leave a sliver unrecovered and blog progress, since the
+                                // would leave a sliver unrecovered and block progress, since the
                                 // recovery was not cancelled.
                                 //
                                 // Treating it as success would be worse as we would progress
                                 // but not have the sliver stored. We therefore panic.
-                                tracing::error!("a sliver recovery task cancelled unexpectedly");
+                                tracing::error!(
+                                    error = ?join_err,
+                                    "a sliver recovery task cancelled unexpectedly"
+                                );
                                 panic!("a sliver recovery task cancelled unexpectedly");
                             }
                         }


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -312,22 +310,25 @@ impl BlobSyncHandler {
 
         let in_progress_gauge =
             metrics::with_label!(self.node.metrics.recover_blob_backlog, STATUS_IN_PROGRESS);
+        let start = tokio::time::Instant::now();
```
halfprice: Question: why moving the start time? I feel the old start time is also pretty useful in a sense that it including the time acquiring the lock, which gives us some idea about blob sync contention. For this new time duration, I think it is the same as the `run` span in `tracing_span_latency` (still feel amazing that we have this metric!). I would say keep the old time, and if you necessary, we can add another time here to explicitly capture the `synchronizer.run` time (although we can already get it from the `tracing_span_latency`.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -465,46 +460,96 @@ impl BlobSynchronizer {
...
+
+        future::try_join(
+            self.clone()
+                .recover_sliver::<Primary>(shard, metadata.clone())
+                .observe(histograms.clone(), labels_from_sliver_result::<Primary>),
+            self.clone()
+                .recover_sliver::<Secondary>(shard, metadata.clone())
+                .observe(histograms.clone(), labels_from_sliver_result::<Secondary>),
+        )
+        .await?;
```
halfprice, line 546: question: looks like we are recovering both slivers in the same shard in the same thread concurrently. Should we future put them in separate threads?

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -465,46 +460,96 @@ impl BlobSynchronizer {
...
+                                std::panic::resume_unwind(reason);
+                            }
+                            Err(join_err) => {
+                                assert!(join_err.is_cancelled());
+                                // We do not poll after cancelling the tasks, and as we have the
+                                // join handle in our JoinSet, no one else should be able to
+                                // abort the task, therefore should never happen.
+                                //
+                                // Returning from this function as if the task was cancelled
+                                // would leave a sliver unrecovered and blog progress, since the
```
halfprice: s/blog/block ?

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -465,46 +460,96 @@ impl BlobSynchronizer {
...
+                                // recovery was not cancelled.
+                                //
+                                // Treating it as success would be worse as we would progress
+                                // but not have the sliver stored. We therefore panic.
+                                tracing::error!("a sliver recovery task cancelled unexpectedly");
```
halfprice: Nit: log the error

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -465,46 +460,96 @@ impl BlobSynchronizer {
...
         let mut futures_with_permits = std::pin::pin!(futures_with_permits);
-        let mut pending_futures = FuturesUnordered::new();
+        // When stored in a JoinSet, tasks are aborted on drop.
+        let mut pending_tasks = JoinSet::new();
 
         loop {
             tokio::select! {
                 biased;
 
                 Some(future) = futures_with_permits.next() => {
```
halfprice, line 470: Remark: it seems we are acquiring 1 permit to sync two slivers now.

crates/walrus-service/src/node/blob_sync.rs
jpcsmith: This should be a total of 1k  (~667 + 334) requests for symbols (before the change upcoming batch change), where most operations are network-IO bound. This will go down to a few hundred requests with the batched requests. This should be fine for a single thread until we see otherwise.

crates/walrus-service/src/node/blob_sync.rs
jpcsmith: Yes, that is the case. It changes the semantics of the the config value.

crates/walrus-service/src/node/blob_sync.rs
jpcsmith: `tracing_span_latency` is nice when you haven't instrumented the code you're interested in with metrics, but it's not sufficient, for e.g., every span with the name run is added to the same metric and you cannot use labels.

I had thought that we also had metrics around the time in queue, but that seems not to be the case, so I now moved the start time to before we acquire the permit as suggested, but kept it in the same function to avoid the extra argument. Let me know if that works.

PR #1448, opened 2025-01-20
feat: add configurable db connection to walrus-backup

wbbradley:
## Description

Add a DB connection to the backup service.

## Test plan

Local manual testing.

No release impact.

### DIFF
--- a/crates/walrus-service/Cargo.toml
+++ b/crates/walrus-service/Cargo.toml
@@ -30,7 +30,7 @@ required-features = ["backup"]
 backup = [
   "dep:async-channel",
   "dep:diesel",
-  "dep:diesel-async"
+  "dep:diesel-async",
 ]
 client = [
   "dep:colored",
@@ -59,7 +59,7 @@ test-utils = [
 
 [dependencies]
 anyhow.workspace = true
-async-channel = {workspace = true, optional = true}
+async-channel = { workspace = true, optional = true }
 async-trait = { workspace = true, optional = true }
 axum = { workspace = true, features = [
   "http1",
--- a/crates/walrus-service/migrations/2025-01-20-173100_create_stream_events/up.sql
+++ b/crates/walrus-service/migrations/2025-01-20-173100_create_stream_events/up.sql
@@ -34,7 +34,7 @@ CREATE TABLE blob_state (
     PRIMARY KEY (blob_id),
     CONSTRAINT valid_blob_state
     CHECK (
-           (state = 'archived'
+        (state = 'archived'
             AND initiate_fetch_after IS NULL
             AND backup_url IS NOT NULL
             AND fetch_attempts IS NULL)
--- a/crates/walrus-service/src/backup.rs
+++ b/crates/walrus-service/src/backup.rs
@@ -425,6 +425,8 @@ async fn backup_fetcher_core(
 pub mod defaults {
     use std::net::{Ipv4Addr, SocketAddr};
 
+    use super::WORKER_COUNT;
+
     /// Default backup node metrics port.
     pub const METRICS_PORT: u16 = 10184;
 
@@ -441,7 +443,7 @@ pub mod defaults {
     }
     /// Default backup work item queue size.
     pub fn message_queue_size() -> u32 {
-        WORKER_COUNT
+        WORKER_COUNT.try_into().expect("WORKER_COUNT overflow")
     }
     /// Default backup max_fetch_attempts_per_blob.
     pub fn max_fetch_attempts_per_blob() -> u32 {
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,7 +36,7 @@ chrono = "0.4"
 clap = { version = "4.5.27", features = ["derive"] }
 colored = "2.2.0"
 criterion = "0.5.1"
-diesel = { version = "2.2", features = ["postgres", "uuid", "chrono"] }
+diesel = { version = "2.2", features = ["chrono", "postgres", "uuid"] }
 diesel-async = { version = "0.5", features = ["postgres"] }
 enum_dispatch = "0.3"
 fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "2f502fd8570fe4e9cff36eea5bbd6fef22002898" }
--- a/crates/walrus-service/src/backup/schema.rs
+++ b/crates/walrus-service/src/backup/schema.rs
@@ -1,3 +1,6 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
 // @generated automatically by Diesel CLI.
 
 diesel::table! {
@@ -30,8 +33,4 @@ diesel::table! {
     }
 }
 
-diesel::allow_tables_to_appear_in_same_query!(
-    blob_state,
-    epoch_change_start_event,
-    stream_event,
-);
+diesel::allow_tables_to_appear_in_same_query!(blob_state, epoch_change_start_event, stream_event,);


crates/walrus-service/Cargo.toml
```diff
@@ -79,6 +79,8 @@ byteorder.workspace = true
 chrono.workspace = true
 clap.workspace = true
 colored = { workspace = true, optional = true }
+diesel.workspace = true
+diesel-async.workspace = true
```
halfprice: Suggestion: make them optional and only use in backup feature

crates/walrus-service/migrations/00000000000000_diesel_initial_setup/up.sql
```diff
@@ -0,0 +1,36 @@
...
+-- in the modified columns)
+--
+-- # Example
+--
+-- ```sql
+-- CREATE TABLE users (id SERIAL PRIMARY KEY, updated_at TIMESTAMP NOT NULL DEFAULT NOW());
+--
+-- SELECT diesel_manage_updated_at('users');
+-- ```
+CREATE OR REPLACE FUNCTION diesel_manage_updated_at(_tbl regclass) RETURNS VOID AS $$
```
halfprice, line 19: I'm not very familiar with postgres and by reading the comment, this trigger just updates the `updated_at` field of a table. Where is this trigger being used?

I probably would ask @sadhansood to take a look at postgres related code since he may have experience.

crates/walrus-service/migrations/2025-01-20-173100_create_stream_events/up.sql
```diff
@@ -0,0 +1,7 @@
+CREATE TABLE stream_event (
+    checkpoint_sequence_number BIGINT NOT NULL,
+    counter BIGINT NOT NULL,
+    element_index BIGINT NOT NULL,
+    element JSONB NOT NULL,
+    PRIMARY KEY (checkpoint_sequence_number, counter)
```
halfprice: Question: since we already store checkpoints and events in local RocksDb, how do you envision this table being used in the backup logic? Depending on how this is used, it may also be useful to key this table by element_index.

crates/walrus-service/Cargo.toml
wbbradley: Good call!

crates/walrus-service/migrations/00000000000000_diesel_initial_setup/up.sql
wbbradley: Removed prior comment. I misread. This appears to be part of diesel's design in the case that folks want to use tables with automatically updating updated_at fields.

crates/walrus-service/migrations/2025-01-20-173100_create_stream_events/up.sql
wbbradley: My goal is to remove rocksdb as per our earlier conversation.

crates/walrus-service/migrations/2025-01-20-173100_create_stream_events/up.sql
```diff
@@ -0,0 +1,25 @@
+-- Initialize the stream_event and blob_state tables.
+
+CREATE TABLE stream_event (
```
halfprice, line 3: Question: will this table eventually replacing the event_store in EventProcessor? Currently it stores all the events but only used during restarts to find the correct cursor to continue event stream.

wbbradley: I currently have no plans to mix data storage solutions between the EventProcessor engine (encapsulated state machine) and the backup engine. I think it makes sense to keep them separate. If we want to migrate the EventProcessor to alternate backing stores (like Postgres) we should probably consider that within the context of EventProcessor and its dependencies (including storage node, and I think eventually a public SDK version of it.)

Regarding storing all the events, it's only storing Walrus-related events, not all Sui events. And, my impression from our prior conversations was that this was helpful for debugging reasons (as a more accessible/queryable form of Walrus event data than on-chain Sui checkpoints...). If perf is an issue, I could make the cardinality of this table 1, and only store the most recently processed latest checkpoint sequence number, etc... I think for now though, this is fine, and we can tweak it once it's in testnet.

crates/walrus-service/src/common/utils.rs
```diff
@@ -311,7 +312,30 @@ impl MetricsAndLoggingRuntime {
...
         Ok(Self {
-            runtime,
+            runtime: Some(runtime),
+            registry: walrus_registry,
+            _telemetry_guards: telemetry_guards,
+            _tracing_handle: tracing_handle,
+        })
+    }
+    /// Avoid creation of a new runtime.
+    pub fn start_free_running(mut metrics_address: SocketAddr) -> anyhow::Result<Self> {
```
halfprice: What does free running mean?

crates/walrus-service/src/testbed.rs
```diff
@@ -471,6 +470,33 @@ pub async fn create_client_config(
...
+pub async fn create_backup_config(
+    system_ctx: &SystemContext,
+    working_dir: &Path,
+    database_url: &str,
+    rpc: String,
+) -> anyhow::Result<BackupNodeConfig> {
+    use crate::{
+        backup::defaults::{max_fetch_attempts_per_blob, message_queue_size, metrics_address},
+        common::config::{defaults::polling_interval, SuiReaderConfig},
+    };
```
halfprice: Nit: I don't see we do scoped `use` in our codebase, except in `mod test`. Can we move this to the top of the file?

crates/walrus-service/bin/deploy.rs
```diff
@@ -378,13 +384,27 @@ mod commands {
...
+        if let Some(database_url) = backup_database_url {
+            let backup_config = create_backup_config(
+                &testbed_config.system_ctx,
+                working_dir.as_path(),
+                database_url.as_str(),
+                testbed_config.sui_network.env().rpc,
+            )
+            .await?;
+            let serialized_backup_config = serde_yaml::to_string(&backup_config)
+                .context("Failed to serialize backup config")?;
```
halfprice, line 402:
```suggestion
                .context("failed to serialize backup config")?;
```

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
 }
 
-impl LoadConfig for BackupNodeConfig {}
+async fn stream_events(
+    event_processor: Arc<EventProcessor>,
+    _metrics_registry: Registry,
+    mut pg_connection: AsyncPgConnection,
+) -> Result<()> {
+    let event_cursor = get_backup_node_cursor(&mut pg_connection).await?;
+    tracing::debug!(?event_cursor, "[stream_events] starting");
```
halfprice: Nit: probably can be in `info` level since this is only printed once at start

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
+        element_index,
+        PositionedStreamEvent {
+            element,
+            checkpoint_event_position,
+        },
+    )) = indexed_element_stream.next().await
+    {
+        match &element {
+            EventStreamElement::ContractEvent(ref contract_event) => {
+                println!("{}", serde_json::to_string(contract_event).unwrap());
```
halfprice: Nit: use tracing for logging?

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
+                 DO UPDATE SET
+                     end_epoch = GREATEST(EXCLUDED.end_epoch, blob_state.end_epoch),
+                     state = CASE WHEN
+                         blob_state.state = 'archived' THEN
+                             blob_state.state
+                         ELSE
+                             'waiting'
+                         END,
+                     fetch_attempts = CASE WHEN
+                         blob_state.state = 'waiting' THEN
```
halfprice: Question: is this for when the new state is `waiting` or the existing (before the UPDATE) state is `waiting`?

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
-    pub fn new(event_manager: Box<dyn EventManager>, metrics_registry: Registry) -> Self {
-        Self {
-            metrics_registry,
-            event_manager,
+async fn dispatch_contract_event(
+    contract_event: &ContractEvent,
+    conn: &mut AsyncPgConnection,
+) -> Result<()> {
+    match contract_event {
+        ContractEvent::BlobEvent(BlobEvent::Registered(_)) => {}
```
halfprice: Nit: I think you can probably eliminate all these ()=>{} by doing something like:

```
match contract_event {
        ContractEvent::BlobEvent(blob_event) => match blob_event {
            BlobEvent::Certified(blob_certified) => {
               ...
            }
            _ => {}
        },
        ContractEvent::EpochChangeEvent(EpochChangeEvent::EpochChangeStart(EpochChangeStart { epoch, .. })) => {
           ...
        }
        _ => {}
    }
```

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
+    blob_id: BlobId,
+}
+
+/// Starts a new backup node runtime.
+pub async fn start_backup_node(
+    metrics_registry: Registry,
+    event_processor: Arc<EventProcessor>,
+    config: BackupNodeConfig,
+) -> Result<()> {
+    // Connect to the database, if we are configured to do so.
```
halfprice:
```suggestion
    // Connect to the database.
```

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
-            if let Err(ref error) = result {
-                tracing::error!(?error, "backup node exited with an error");
-            }
+async fn backup_fetcher_core(
+    worker_id: usize,
+    backup_config: BackupNodeConfig,
+    rx: async_channel::Receiver<BackupWorkItem>,
+) -> Result<()> {
+    tracing::info!("[backup_fetcher:{worker_id}] starting worker");
+    let mut conn = AsyncPgConnection::establish(&backup_config.database_url)
```
halfprice: Nit: use `establish_connection`?

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
+/// push them into the work queue. It should be a singleton.
+async fn backup_delegator(
+    backup_config: BackupNodeConfig,
+    tx: async_channel::Sender<BackupWorkItem>,
+) -> Result<()> {
+    use self::models::BlobIdRow;
 
-        bail!("event stream for blob events stopped")
+    tracing::info!("[backup_delegator] Starting worker with config: {backup_config:?}");
+    let mut conn = AsyncPgConnection::establish(&backup_config.database_url).await?;
```
halfprice: Nit: use `establish_connection`?

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
+                            LIMIT $2
+                        ),
+                        _updated_count AS (
+                            UPDATE blob_state
+                            SET
+                                initiate_fetch_after = NOW() + INTERVAL '45 minute',
+                                fetch_attempts = fetch_attempts + 1
+                            WHERE blob_id IN (SELECT blob_id FROM ready_blob_states)
+                        )
+                        SELECT blob_id FROM ready_blob_states",
```
halfprice, line 302: Remark: This query may perform poorly if it has already stored large number of blobs. We have this `blob_state_delegate_after` index, why not use that to fetch all the blobs waiting to be fetched? If that can be used, we can replace the implementation in a followup PR.

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
+                        "gs://backup/{blob_id}",
+                        blob_id = item.blob_id
+                    ))
+                    .bind::<diesel::sql_types::Bytea, _>(item.blob_id.as_ref().to_vec())
+                    .execute(conn)
+                    .await?)
+                }
+                .scope_boxed()
+            })
+            .await?;
```
halfprice: Suggestion: since we haven't store the blob yet, can we implement updating the row to `archived` when actually storing the blob?

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
+        // Pull a work item from the queue.
+        let item = rx.recv().await.expect("channel closed!");
+        tracing::info!(
+            "[backup_fetcher:{worker_id}] processing blob {blob_id:?}",
+            blob_id = item.blob_id
+        );
+        // Fetch the blob from Walrus network.
+        let _blob: Vec<u8> = read_client
+            .read_blob::<Primary>(&item.blob_id)
+            .await
```
halfprice, line 380: Question: any error in this loop terminates the worker thread. Should we just log the error here and let the thread continue running?

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
 
-        bail!("event stream for blob events stopped")
+    tracing::info!("[backup_delegator] Starting worker with config: {backup_config:?}");
+    let mut conn = AsyncPgConnection::establish(&backup_config.database_url).await?;
+    let max_fetch_attempts_per_blob = i32::try_from(backup_config.max_fetch_attempts_per_blob)
+        .expect("max_fetch_attempts_per_blob config overflow");
+    let message_queue_size = i32::try_from(backup_config.message_queue_size)
+        .expect("message_queue_size config overflow");
+    loop {
+        // Poll the db for new work items. Use the message_queue_size as an upper bound.
```
halfprice, line 271: Suggestion: should we use `WORKER_COUNT` as the bound? Not sure what we get by using `message_queue_size` as the upper bound except it'll fill the channel quickly.

crates/walrus-service/src/backup.rs
```diff
@@ -48,122 +63,370 @@ pub struct BackupNodeConfig {
...
+
+        tracing::info!(
+            "[backup_delegator] found {count} blobs in waiting state",
+            count = blob_id_rows.len()
+        );
+        for BlobIdRow { blob_id } in blob_id_rows {
+            tx.send(BackupWorkItem {
+                blob_id: BlobId::try_from(blob_id.as_slice())?,
+            })
+            .await?;
```
halfprice, line 323: ~~Important: when the channel is full, here it will terminate the backup_delegator. We should either add a retry sending items, or use an unbounded channel.~~

Ignore this. My mistake that I thought this was doing a try_send().

crates/walrus-service/src/common/utils.rs
wbbradley: It's an allusion to free-running threads, for lack of a better name at the time. I meant to basically say that this is just using normal async/await machinery, not trying to scuplt its runtime in a special fashion using runtime::block_on, etc... I'll try to think of a better name. I really want to say "normal async/await" here...

wbbradley: OK, I've cleaned this up.

crates/walrus-service/src/node/events/event_processor_runtime.rs
```diff
@@ -118,6 +118,34 @@ impl EventProcessorRuntime {
...
 
+    /// Starts the event processor in a free-running fashion.
+    pub async fn start_free_running(
+        sui_config: SuiReaderConfig,
+        event_processor_config: EventProcessorConfig,
+        db_path: &Path,
+        metrics_registry: &Registry,
+        cancel_token: CancellationToken,
+    ) -> anyhow::Result<Arc<EventProcessor>> {
+        tracing::error!("[start_free_running] 111111");
```
halfprice: Remove or update the debug logging to something more meaningful?

crates/walrus-service/src/backup.rs
wbbradley: Good idea.

crates/walrus-service/src/backup.rs
wbbradley: `EXCLUDED` is the incoming row data that conflicted, so `blob_state` is the previously existing (before the UPDATE) row that `EXCLUDED` conflicted with.

crates/walrus-service/src/backup.rs
wbbradley: This used to be optional. Good catch.

crates/walrus-service/src/backup.rs
wbbradley: Good catch.

crates/walrus-service/src/backup.rs
wbbradley: I think some sort of ideal number would be the count of idle (or soon-to-be idle) workers. However, in reality it's a lot easier to avoid synchronization of this datapoint. This is especially true because I want to move the backup_fetcher logic out to a separate process/container eventually.

I agree, let's reduce the size of the query for now.

wbbradley: For now, I'll hardcode the default message_queue_size to WORKER_COUNT. Another thing I should do is check the size of the queue and subtract that from the local var before issuing the query. I'll get to that later. Logged WAL-560 to track that

crates/walrus-service/src/backup.rs
wbbradley: Good catch. I've reworked this to handle that in the next PR. I'll leave it like this for now.

crates/walrus-service/src/backup.rs
wbbradley: This code will never run in production, and I've already got the GCS bit written in the next PR. Would like to just land this, then get the storage part written to get closer to end-to-end if that's OK...

crates/walrus-service/src/backup.rs
halfprice: My bad. Thanks for educating me on using modern DB!

crates/walrus-service/src/node/events/event_processor_runtime.rs
wbbradley: Done.

PR #1455, opened 2025-01-21
feat(node): configure parameters during bootstrap

liquid-helium:
## Description

Exposed necessary APIs in SuiContractClient to update the following StorageNode parameters:
* name
* network_address
* network_public_key
* next_public_key
* storage_price
* write_price
* node_capacity

Added a `sync_node_params` step when constructing a StorageNode, which will compare the local configs with the on-chain values and update them if necessary.

TODO: next_public_key update requires more changes to make sure the current key-pair will be around until the next epoch.

Resolves WAL-245

## Test plan

---

## Release notes

- [x] Storage node:
  -- Added the feature to change storage node config parameter as following, by changing the local config file and restart node.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/node.rs
```diff
@@ -447,6 +447,76 @@ pub struct NodeParameters {
...
+        .get_address_capability_object(address)
+        .await?
+        .ok_or(SuiClientError::StorageNodeCapabilityObjectNotSet)?;
+
+    let pool = contract_client
+        .read_client
+        .get_staking_pool(node_cap.node_id)
+        .await?;
+
+    let node_info: &walrus_sui::types::StorageNode = &pool.node_info;
```
halfprice: Nit: probably don't need `&walrus_sui::types::StorageNode`

crates/walrus-service/src/node.rs
```diff
@@ -447,6 +447,76 @@ pub struct NodeParameters {
...
+        &node_info.network_public_key,
+        &pool.voting_params,
+    );
+
+    if let Some(update_params) = update_params {
+        tracing::info!(
+            node_name = config.name,
+            node_id = ?node_info.node_id,
+            update_params = ?update_params,
+            "Update Node Params"
```
halfprice: nit: s/Update Node Params/update node params

Our convention is that we don't capital letters in logs.

crates/walrus-service/src/node.rs
```diff
@@ -447,6 +447,76 @@ pub struct NodeParameters {
...
+        );
+        // TODO: support updating public key.
```
halfprice: nit: attach associated issue here in the TODO.

crates/walrus-sui/src/client.rs
```diff
@@ -1188,6 +1189,41 @@ impl SuiContractClient {
...
+            .ok_or(SuiClientError::StorageNodeCapabilityObjectNotSet)?;
+
+        tracing::debug!(
+            wallet_address = ?self.wallet_address,
+            network_address = ?node_parameters.network_address,
+            "updating node parameters"
+        );
+
+        // Lock the wallet here to ensure there are no race conditions with object references.
+        let wallet = self.wallet().await;
```
halfprice: Note: this has slightly changed after #1438

crates/walrus-service/src/node.rs
```diff
@@ -447,6 +447,76 @@ pub struct NodeParameters {
...
+        if update_params.next_public_key.is_some() {
+            return Err(anyhow::anyhow!("not supported"));
+        }
+        let proof_of_possession = if update_params.next_public_key.is_some() {
```
halfprice: Question: can `proof_of_possession` be put inside `update_params`?

crates/walrus-service/src/node.rs
```diff
@@ -458,6 +528,7 @@ impl StorageNode {
         node_params: NodeParameters,
     ) -> Result<Self, anyhow::Error> {
         let start_time = Instant::now();
+        sync_node_params(config).await?;
```
halfprice: Suggestion: I wonder if we should add an option here to skip sync. Currently this is a blocking operation, and in the case where the transaction fails (can due to sui fullnode unavailable), the storage node won't be able to start. By allowing bypass this update, the storage node can at least serve reads.

@mlegner @karlwuest what do you guys think?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -860,4 +863,158 @@ mod tests {
...
+                    .storage_node_capability
+                    .as_ref()
+                    .unwrap()
+                    .node_id,
+                test_cluster::FROST_PER_NODE_WEIGHT * 3,
+            )
+            .await
+            .expect("stake with node pool should not fail");
+
+        tokio::time::sleep(Duration::from_secs(150)).await;
```
halfprice: The sleep here can be much shorter (30s maybe, the epoch duration) I think. Given that the simtest runs sui + walrus cluster, which is quite expensive, the test runtime is proportional to the expected wall time.

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -860,4 +863,158 @@ mod tests {
             tokio::time::sleep(Duration::from_secs(1)).await;
         }
     }
+
+    #[walrus_simtest]
+    #[ignore = "ignore simtests by default"]
+    async fn test_registered_node_update_params() {
```
halfprice, line 869: Love the test!!!

crates/walrus-service/src/node.rs
karlwuest: Remark (for the follow up): We will probably need to store two keys in the config, because we still need to use the old one until the epoch change happens.

crates/walrus-service/src/node/config.rs
```diff
@@ -206,6 +215,75 @@ impl StorageNodeConfig {
...
+    /// update params if there are any changes, so that the source of the passed-in parameters
+    /// can be updated to the node parameters.
+    pub fn generate_update_params(
+        &self,
+        name: &str,
+        network_address: &str,
+        public_key: &PublicKey,
+        network_public_key: &NetworkPublicKey,
+        voting_params: &VotingParams,
+    ) -> Option<NodeUpdateParams> {
```
karlwuest: Suggestion: This currently contains a lot of chaining of options (e.g. returns an `Option<NodeUpdateParams>` which contains an `Option<NodeVotingParams>` which contains an Option for each vote.

I think the readability (here and when using it in the transaction builder) could be improved by: 
1. returning a `NodeUpdateParams` instead of an option and providing a function on `NodeUpdateParams` that checks if an update is needed.
2. Flattening `NodeUpdateParams` to just contain an option for each parameter (including individual votes) and creating a default `NodeUpdateParams` at the start. Then you can just set the relevant values.

crates/walrus-service/src/node.rs
karlwuest: I agree, I think this should be grouped with the public key in `update_params`.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -807,6 +808,96 @@ impl WalrusPtbBuilder {
...
+                self.staking_arg(Mutability::Mutable).await?,
+                self.argument_from_arg_or_obj(storage_node_cap).await?,
+                self.pt_builder.pure(name)?,
+            ];
+            self.walrus_move_call(contracts::staking::set_name, args)?;
+        }
+
+        // Update next public key if provided
+        if let Some(next_public_key) = params.next_public_key {
+            let Some(proof_of_possession) = proof_of_possession else {
```
karlwuest: Remark: If the proof of possession and the public key are grouped in the `NodeUpdateParams` (as mentioned above), this error case can no longer happen.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -807,6 +808,96 @@ impl WalrusPtbBuilder {
...
+                );
+            };
+            let args = vec![
+                self.staking_arg(Mutability::Mutable).await?,
+                self.argument_from_arg_or_obj(storage_node_cap).await?,
+                self.pt_builder.pure(next_public_key.as_bytes())?,
+                self.pt_builder
+                    .pure(proof_of_possession.signature.as_bytes())?,
+            ];
+            self.walrus_move_call(contracts::staking::set_next_public_key, args)?;
```
karlwuest: Suggestion: Add separate methods for all these calls on the transaction builder and then use those here instead of constructing all contract calls directly.

crates/walrus-service/src/node.rs
karlwuest: Would you add this as a cli flag? I think that would make sense, although the node functionality will anyway be quite restricted if the transaction fails in a non-retriable way (or if e.g. the network address or keys change).

crates/walrus-service/src/node.rs
liquid-helium: Agreed!

I am drafting a [RP](https://github.com/MystenLabs/walrus/pull/1461) here to make the local config a mapping of the remote config, i.e., adding a local next_protocol_key_pair.

That will require some control flow changes to make sure the storage node react to the event changes with a proper config change.

crates/walrus-simtest/tests/simtest.rs
liquid-helium: Thanks for the tip, change applied.

crates/walrus-service/src/node.rs
liquid-helium: a flag is added, default to false.

This is an interesting point of discussion. As both of you noted. On one hand, a read-only node can offer some value. On the other hand, if there's a potential configuration mismatch, such as an incorrect network address, other nodes wouldn't be able to locate it anyway (assuming the storage node's address is made public through the SUI network).

I've added this check to the StorageNode **constructor** to enforce this check. Since if this operation fails, the node might progressively reach a more problematic state. It could be more effective to fail immediately, prompting operators to notice and address the issue.

crates/walrus-service/src/node.rs
liquid-helium: I like this suggestion very much! Thanks folks.

crates/walrus-sui/src/client.rs
liquid-helium: Thanks for the update! I really like the changes made in #1438—great job!

crates/walrus-service/src/node.rs
liquid-helium: Change applied, thanks for the tip.

crates/walrus-service/src/node.rs
liquid-helium: I removed the TODO since it will be resolved in https://github.com/MystenLabs/walrus/pull/1461 soon

crates/walrus-service/src/node.rs
liquid-helium: Done.

crates/walrus-sui/src/client/transaction_builder.rs
liquid-helium: Thanks for the suggestion, change applied.

crates/walrus-sui/src/client/transaction_builder.rs
liquid-helium: Done, thats a good idea.

crates/walrus-service/src/node/config.rs
liquid-helium: Thanks, change applied, looks much cleaner.

PR #1456, opened 2025-01-22
feat: add new recovery symbol endpoint using symbol ID

jpcsmith:
## Description

Add a new recovery symbol endpoint that uses an ID calculated from the primary and secondary sliver indices.  Additionally, deprecate the old recovery symbol endpoint in favour of the new one.

## Test plan

Majority of the change was a refactor of the `retrieve_recovery_symbol` method. As such, it is tested with existing tests. No code currently uses the new endpoint.

---

## Release notes

- [x] Storage node: Deprecate the recovery symbol endpoint in favour of a new endpoint. Integrations using recovery symbols should migrate to the new endpoint.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/storage_openapi.yaml
```diff
@@ -219,6 +219,55 @@ paths:
...
+      responses:
+        '200':
+          description: BCS-encoded primary/secondary recovery symbol
+          content:
+            application/octet-stream:
+              schema:
+                type: string
+                format: binary
+        '400':
+          description: May be returned when (1)  The index identifying the resource is out-of-range for the system. (2)  The recovery symbol ID is invalid for the system. (3)  The index identifying the resource is out-of-range for the system. (4)  The shard associated with the operation is not assigned to this storage node.
```
giac-mysten: Nit: ` (1)  The index identifying the resource is out-of-range for the system. ` is repeated twice

crates/walrus-service/storage_openapi.yaml
```diff
@@ -219,6 +219,55 @@ paths:
...
+              schema:
+                $ref: '#/components/schemas/Status'
+        '451':
+          description: ' The sliver cannot be returned, as the associated blob has been blocked on this storage node.'
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
+        '500':
+          description: May be returned when (1) An internal server error has occurred. Please report this error. (2) An internal server error has occurred. Please report this error.
```
giac-mysten: Nit: some descriptions have a leading `' `, others do not

giac-mysten: Also the description is duplicated?

crates/walrus-service/storage_openapi.yaml
```diff
@@ -368,7 +417,7 @@ paths:
                 type: string
                 format: binary
         '400':
-          description: May be returned when (1)  The index identifying the resource is out-of-range for the system. (2)  The index identifying the resource is out-of-range for the system. (3)  The shard associated with the operation is not assigned to this storage node.
+          description: May be returned when (1)  The index identifying the resource is out-of-range for the system. (2)  The recovery symbol ID is invalid for the system. (3)  The index identifying the resource is out-of-range for the system. (4)  The shard associated with the operation is not assigned to this storage node.
```
giac-mysten: Nit: (1) (3) duplicated again? is this on purpose?

crates/walrus-core/src/lib.rs
```diff
@@ -587,6 +599,68 @@ impl Display for SliverType {
...
+
+    /// Returns the pair `(primary_sliver_index, secondary_sliver_index)` that corresponds
+    /// to the decoding symbol.
+    ///
+    /// Returns None if the `DecodingSymbolId` is greater than or equal to `n_shards^2`.
+    pub fn to_sliver_indices(&self, n_shards: NonZeroU16) -> Option<(SliverIndex, SliverIndex)> {
+        let n_shards_u64 = u64::from(n_shards.get());
+        if n_shards_u64 * n_shards_u64 <= self.0 {
+            return None;
+        }
```
giac-mysten: Question: why the soft fail?

crates/walrus-core/src/lib.rs
```diff
@@ -807,3 +881,56 @@ macro_rules! ensure {
...
+
+    param_test! {
+        decoding_symbol_id_conversion: [
+            zero: (7, DecodingSymbolId(0), 0, 0),
+            eighteen: (7, DecodingSymbolId(18), 2, 4),
+            forty_one: (7, DecodingSymbolId(41), 5, 6),
+            last: (7, DecodingSymbolId(48), 6, 6),
+        ]
+    }
+    fn decoding_symbol_id_conversion(
```
giac-mysten: Nit: the two tests could be merged in one?

crates/walrus-service/src/common/api.rs
```diff
@@ -288,9 +290,9 @@ pub(crate) fn into_responses(
 
 /// Convert the path with variables of the form `:id` to the form `{id}`.
 pub(crate) fn rewrite_route(path: &str) -> String {
-    regex::Regex::new(r":(?<param>\w+)")
+    regex::Regex::new(r"/:(?<param>\w+)")
```
giac-mysten: question (for me): why is `/` needed?

crates/walrus-service/src/node.rs
```diff
@@ -1925,36 +1945,65 @@ impl ServiceState for StorageNodeInner {
...
-        sliver_pair_index: SliverPairIndex,
-        sliver_type: SliverType,
-        target_pair_index: SliverPairIndex,
+        symbol_id: DecodingSymbolId,
+        sliver_type: Option<SliverType>,
     ) -> Result<RecoverySymbol<MerkleProof>, RetrieveSymbolError> {
-        // Before touching the database, verify that the target_pair_index is possibly valid, and
-        // not out of range. Checking the sliver_pair_index is done by retrieve_sliver.
-        self.check_index(target_pair_index)?;
+        tracing::error!("at the start of the method call");
```
giac-mysten: Nit: leftover from debugging?

crates/walrus-service/src/node.rs
```diff
@@ -1616,6 +1613,34 @@ impl StorageNodeInner {
...
+    ) -> Result<RecoverySymbol<MerkleProof>, RetrieveSymbolError> {
+        let sliver = self.retrieve_sliver(blob_id, sliver_pair_index, sliver_type.orthogonal())?;
+
+        let symbol_result = match sliver {
+            Sliver::Primary(inner) => inner
+                .recovery_symbol_for_sliver(target_pair_index, &self.encoding_config)
+                .map(RecoverySymbol::Secondary),
+            Sliver::Secondary(inner) => inner
+                .recovery_symbol_for_sliver(target_pair_index, &self.encoding_config)
+                .map(RecoverySymbol::Primary),
```
giac-mysten: Remark: does it still make sense to have primary and secondary recovery symbols now?

crates/walrus-service/src/common/api.rs
jpcsmith: Because it's possible to have a colon in the URL that doesn't start a variable. All our variables follow a `/` so this distinguishes the two. This function is now removed on main, however, as paths now use `{}` for variables with the latest axum version.

crates/walrus-core/src/lib.rs
jpcsmith: To remind me that I need to check it beforehand in the caller.

crates/walrus-core/Cargo.toml
```diff
@@ -13,6 +13,7 @@ test-utils = ["walrus-test-utils"]
 [dependencies]
 base64.workspace = true
 bcs.workspace = true
+bytes.workspace = true
```
jpcsmith:
```suggestion
```

Cargo.toml
jpcsmith, line 1: Remark: The packages here were perhaps incorrectly removed during my rebase. I will re-add them.

crates/walrus-service/src/node.rs
```diff
@@ -1925,36 +1972,65 @@ impl ServiceState for StorageNodeInner {
...
-                .recovery_symbol_for_sliver(target_pair_index, &self.encoding_config)
-                .map(RecoverySymbol::Secondary),
-            Sliver::Secondary(inner) => inner
-                .recovery_symbol_for_sliver(target_pair_index, &self.encoding_config)
-                .map(RecoverySymbol::Primary),
-        };
+        symbol_id: SymbolId,
+        sliver_type: Option<SliverType>,
+    ) -> Result<GeneralRecoverySymbol, RetrieveSymbolError> {
+        tracing::error!("at the start of the method call");
```
jpcsmith:
```suggestion
```

crates/walrus-service/src/node.rs
jpcsmith: They have been unified in the latest version.

crates/walrus-core/src/lib.rs
jpcsmith: These tests have been replaced since no conversion is necessary now to the 1d index.

crates/walrus-service/storage_openapi.yaml
jpcsmith: Added `.trim()` call to remove the whitespace that comes from the docstrings.

crates/walrus-service/storage_openapi.yaml
jpcsmith: This is fixed in #1472.

Cargo.toml
giac-mysten: were they readded? look still removed for me?

crates/walrus-core/src/encoding/errors.rs
```diff
@@ -140,4 +140,7 @@ pub enum SymbolVerificationError {
     /// The provided metadata is itself invalid.
     #[error("the metadata is invalid")]
     InvalidMetadata,
+    /// The recovery symbol cannot be used to recover the identified sliver.
+    #[error("the symbol cannot be used to recovery the identified sliver")]
+    SymbolNotUsable,
```
giac-mysten: Nit: can we be more specific here? The errors above are also reasons why the "symbol is not usable".

crates/walrus-core/src/encoding/symbols.rs
```diff
@@ -341,6 +344,211 @@ impl<T: EncodingAxis> Display for DecodingSymbol<T> {
...
+
+        if self.proof.verify_proof(
+            expected_root,
+            self.symbol.data(),
+            self.target_index.as_usize(),
+        ) {
+            Ok(())
+        } else {
+            Err(SymbolVerificationError::InvalidProof)
+        }
```
giac-mysten: Nit: maybe?
```suggestion
        self.proof
            .verify_proof(
                expected_root,
                self.symbol.data(),
                self.target_index.as_usize(),
            )
            .then_some(())
            .ok_or(SymbolVerificationError::InvalidProof)
```

crates/walrus-core/src/encoding/symbols.rs
```diff
@@ -341,6 +344,211 @@ impl<T: EncodingAxis> Display for DecodingSymbol<T> {
...
+/// A recovery symbol taken from a blob.
+///
+/// As the symbol is taken from the intersection of two slivers, a primary and a secondary,
+/// it can be used in the recovery of either sliver, irrespective of its source.
+///
+/// Can be converted into a [`DecodingSymbol<Primary>`] or [`DecodingSymbol<Secondary>`].
+#[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
+pub struct GeneralRecoverySymbol<U = MerkleProof> {
+    symbol: EitherDecodingSymbol,
+    target_index: SliverIndex,
```
giac-mysten, line 419: Remark: for the "general" recovery symbol we could use the `SymbolId`? duplicates a `u16` of information, but may be clearer/simpler? just a note

crates/walrus-core/src/lib.rs
```diff
@@ -815,3 +920,30 @@ macro_rules! ensure {
...
+#[cfg(test)]
+mod tests {
+    use serde_test::{Configure as _, Token};
+
+    use super::*;
+
+    #[test]
+    fn symbol_id_serde_compact() {
+        let symbol_id = SymbolId::new(17.into(), 21.into());
+        serde_test::assert_tokens(
```
giac-mysten, line 933: Cool! didn't know about this

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -70,6 +72,8 @@ pub const DELETABLE_BLOB_CONFIRMATION_ENDPOINT: &str =
 /// The path to get recovery symbols.
 pub const RECOVERY_ENDPOINT: &str =
     "/v1/blobs/{blob_id}/slivers/{sliver_pair_index}/{sliver_type}/{target_pair_index}";
+/// The path to get recovery symbols.
+pub const RECOVERY_SYMBOL_ENDPOINT: &str = "/v1/blobs/{blob_id}/recoverySymbols/{symbol_id}";
```
giac-mysten, line 76: QQ: the `symbol_id` would be `12-41` now, right?

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -396,19 +401,74 @@ pub async fn get_recovery_symbol<S: SyncServiceState>(
...
+    };
+    let symbol_id = SymbolId::new(primary_index, secondary_index);
+    let symbol = state
+        .retrieve_recovery_symbol(&blob_id, symbol_id, Some(sliver_type))?
+        .into();
 
     match symbol {
         RecoverySymbol::Primary(inner) => Ok(Bcs(inner).into_response()),
         RecoverySymbol::Secondary(inner) => Ok(Bcs(inner).into_response()),
     }
```
giac-mysten, line 434: Remark: going back and forth with indices is annoying here, but well, it's deprecated...

Cargo.toml
giac-mysten: sorry, saw this changed again.

crates/walrus-core/src/encoding/symbols.rs
jpcsmith: How about the following? Not so fond of the boolean chaining as it blurs the fact that a boolean is being returned.
```suggestion
        if !self.proof.verify_proof(
            expected_root,
            self.symbol.data(),
            self.target_index.as_usize(),
        ) {
            return Err(SymbolVerificationError::InvalidProof);
        }

        Ok(())
```

crates/walrus-service/src/node/server/routes.rs
jpcsmith: Yes, that's correct.

crates/walrus-service/src/node/server/routes.rs
jpcsmith: Yea, it's just until we remove this endpoint completely after sufficient operators have upgraded.

crates/walrus-core/src/encoding/symbols.rs
jpcsmith: Added an `id()` method that returns the `SymbolId`. Since the data is already there, I would avoid the duplication of information.

PR #1459, opened 2025-01-22
fix: Add retries to rpc client calls

sadhansood:
## Description

Addresses WAL-406 by adding a new `RetryRPClient` to handle retries on common server failures for rpc client calls i.e. `get_object()`. Also starts using `RetriableSuiClient` in places where it was not already being used in event handling code.

Closes WAL-406

## Test plan

Existing tests


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -564,6 +608,76 @@ impl RetriableSuiClient {
     }
 }
 
+/// A `sui_rpc_api::Client` that retries RPC calls with backoff in case of network errors.
```
mlegner: Suggestion: Could you describe somewhere the difference between the clients (`RetryRpcClient` and `RetriableSuiClient`)? 

Also, nit: Can we use the same naming unless there is a good reason not to ("retry" vs. "retriable")?

Also, nit:
```suggestion
/// A [`sui_rpc_api::Client`] that retries RPC calls with backoff in case of network errors.
```

sadhansood: done

PR #1460, opened 2025-01-23
fix(deps): update axum and utoipa versions

jpcsmith:
## Description

Bumps the versions of axum, and as a result, several other dependencies like utoipa, axum_extra, etc.

As these were major version bumps, there were several breaking changes in our code's usage of these crates that needed to be addressed. This change should not have any external-facing API impact.

## Test plan

Tested by existing test cases.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/lib.rs
+++ b/crates/walrus-sui/src/lib.rs
@@ -50,5 +50,6 @@ pub struct ObjectIdSchema(());
     value_type = String,
     title = "Sui address",
     description = "Sui address encoded as a hexadecimal string",
+    examples("0x02a212de6a9dfa3a69e22387acfbafbb1a9e591bd9d636e7895dcfc8de0"),
 )]
 pub struct SuiAddressSchema(());


crates/walrus-sui/src/lib.rs
```diff
@@ -13,3 +13,42 @@ pub mod system_setup;
...
+)]
+pub struct ObjectIdSchema(());
+
+/// Schema for Sui addresses.
+#[derive(Debug, utoipa::ToSchema)]
+#[schema(
+    as = SuiAddress,
+    value_type = String,
+    title = "Sui address",
+    description = "Sui address encoded as a hexadecimal string",
```
halfprice, line 52: Nit: do we want an example here as well?

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -49,28 +52,29 @@ use crate::{
 };
 
 /// OpenAPI documentation endpoint.
-pub const API_DOCS: &str = "/v1/api";
+pub const API_DOCS_ENDPOINT: &str = "/v1/api";
 /// The path to get and store blob metadata.
-pub const METADATA_ENDPOINT: &str = "/v1/blobs/:blob_id/metadata";
+pub const METADATA_ENDPOINT: &str = "/v1/blobs/{blob_id}/metadata";
```
halfprice, line 57: question: this won't change the existing endpoint, correct? (user can use the same URL to access the endpoint)

jpcsmith: Correct, this was just a change in axum as to how variables are represented in the configured routes.

crates/walrus-sui/src/lib.rs
jpcsmith, line 52:
```suggestion
    description = "Sui address encoded as a hexadecimal string",
    examples("0x02a212de6a9dfa3a69e22387acfbafbb1a9e591bd9d636e7895dcfc8de0"),
```

PR #1461, opened 2025-01-23
feat: Update protocol keypair

liquid-helium:
## Description

1. Allow operators to update their protocol key pairs via the local `next_protocol_key_pair` config field
2. Automatically reload the new protocol key pair when a new epoch starts
3. Refresh the committee info in the background to keep it up-to-date, in case some nodes changes their parameters

Resolves WAL-245
## Test plan

---

## Release notes

- [x] Storage node: Storage nodes automatically update and rotate the protocol key pair. This only requires setting the `next_protocol_key_pair` in the config file.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/contract_service.rs
+++ b/crates/walrus-service/src/node/contract_service.rs
@@ -159,11 +159,14 @@ impl SystemContractService for SuiSystemContractService {
         config: &StorageNodeConfig,
         node_capability_object_id: ObjectID,
     ) -> Result<(), SyncNodeConfigError> {
-        let contract_client = self.contract_client.lock().await;
-
+        let node_capability = self
+            .get_node_capability_object(Some(node_capability_object_id))
+            .await?;
+        let contract_client: tokio::sync::MutexGuard<'_, SuiContractClient> =
+            self.contract_client.lock().await;
         let pool = contract_client
             .read_client
-            .get_staking_pool(node_capability_object_id)
+            .get_staking_pool(node_capability.node_id)
             .await?;
 
         let node_info = &pool.node_info;
@@ -174,7 +177,6 @@ impl SystemContractService for SuiSystemContractService {
             &node_info.network_public_key,
             &pool.voting_params,
         );
-
         let action = calculate_protocol_key_action(
             config.protocol_key_pair().public().clone(),
             config


crates/walrus-service/src/node/config.rs
```diff
@@ -123,6 +127,12 @@ pub struct StorageNodeConfig {
     /// Metric push configuration.
     #[serde(default, skip_serializing_if = "defaults::is_none")]
     pub metrics_push: Option<MetricsPushConfig>,
+    /// Interval between config monitoring checks
+    #[serde(default = "defaults::config_synchronizer_interval")]
+    pub config_synchronizer_interval: Duration,
+    /// Enable the config monitor
+    #[serde(default = "defaults::config_synchronizer_enabled")]
+    pub enable_config_synchronizer: bool,
```
mlegner: Suggestion: Could you use a similar construct as with the `event_processor_config`?

* Create a separate `ConfigSynchronizerConfig` with a `#[serde(default, skip_serializing_if = "defaults::is_default")]` attribute and a custom `Default` implementation.
* Parse the duration with `#[serde_as(as = "DurationSeconds")]`, `#[serde(rename = "interval_secs")]`.

crates/walrus-service/src/node/config.rs
```diff
@@ -155,14 +166,80 @@ impl Default for StorageNodeConfig {
...
+                "failed to read config file at '{}': {e}",
+                path.as_ref().display()
+            )
+        })?;
+        let mut config: StorageNodeConfig = serde_yaml::from_str(&config_content).map_err(|e| {
+            anyhow::anyhow!(
+                "failed to parse config file at '{}': {e}",
+                path.as_ref().display()
+            )
+        })?;
```
mlegner: Suggestion: Use the `crate::common::utils::load_from_yaml` function here.

crates/walrus-service/src/node/config.rs
```diff
@@ -155,14 +166,80 @@ impl Default for StorageNodeConfig {
...
+        // Remove old file if it exists
+        if path.as_ref().exists() {
+            std::fs::remove_file(path.as_ref())?;
+        }
+
+        // Rename temporary file to actual config file
+        // If crash happens before rename happens, the old config will be lost,but the new config
+        // has already been written to disk.
+        // This is a trade-off in case rename() is not atomic on some platforms.
+        std::fs::rename(&temp_path, path.as_ref())?;
```
mlegner: Question: What is the benefit of this approach over simply overwriting the existing file? Is this is case the node crashes exactly during the operation?

Also, there is a subtlety when serializing the configuration: We skip parameters that are the default values. It could be that the configuration file contained some of those before, to make sure they are fixed even if the defaults change; those would be lost after writing the file.

contracts/walrus/sources/staking.move
mlegner, line 1: Question: Why do we need these resets? Can't we just set the next public key to the current one to get the same behavior? cc @karlwuest due to changes to the contracts.

crates/walrus-service/src/testbed.rs
```diff
@@ -640,7 +640,7 @@ pub async fn create_storage_node_configs(
                 node_capacity: node.node_capacity,
             },
             metrics_push: None,
-            metadata: Default::default(),
+            ..Default::default()
```
mlegner: Suggestion: List all fields explicitly to make sure we properly set any parameters we add later.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -115,6 +127,96 @@ impl SuiSystemContractService {
 
 #[async_trait]
 impl SystemContractService for SuiSystemContractService {
+    /// Syncs the node parameters with the on-chain values.
+    /// If the node parameters are not in sync, it updates the node parameters on-chain.
+    /// Note this could return error if the node needs reboot, e.g., when protocol key pair
+    /// rotation is required.
+    async fn sync_node_params(&self, config: &StorageNodeConfig) -> Result<(), anyhow::Error> {
```
mlegner: Suggestion: Define an error enum with several variants (`ActionRequired`, `SuiClientError`, ?) that can be matched properly without having to use the `downcast_ref` approach.

crates/walrus-service/src/lib.rs
```diff
@@ -8,6 +8,7 @@ pub mod client;
 
 #[cfg(feature = "node")]
 pub mod node;
+pub use node::errors::StorageNodeError;
```
mlegner: Important: This has to be annotated with `#[cfg(feature = "node")]` as well.

crates/walrus-service/bin/node.rs
```diff
@@ -320,11 +321,40 @@ fn main() -> anyhow::Result<()> {
...
+                    }
+                    Err(e)
+                        if matches!(
+                            e.downcast_ref::<StorageNodeError>(),
+                            Some(StorageNodeError::NodeNeedsReboot)
+                        ) =>
+                    {
+                        tracing::info!("Node needs reboot, restarting...");
+                        continue;
+                    }
```
mlegner: Important: Is it really necessary to restart the node completely when some keys need to be updated? This causes all runtimes to be restarted, cancels in-process blob and shard syncs, etc. It would be much nicer if we would only restart the components that are affected.

liquid-helium: I see your point about how this could impact node performance. 

I've been opting for full restarts mainly because:

Right now, we have a clear understanding of which parameters affect specific runtimes. However, as more components and parameters get added in the future, these relationships might change. A full reboot seems like a safe approach in that context. Additionally, given that parameter updates aren't usually a daily occurrence, the overall impact should remain minimal.

What do you think about this approach?

crates/walrus-service/src/lib.rs
liquid-helium: Thanks for the catch, fixed.

crates/walrus-service/src/node/contract_service.rs
liquid-helium: thanks for the suggestion, change applied.

crates/walrus-service/src/testbed.rs
liquid-helium: Fix applied.

contracts/walrus/sources/staking.move
liquid-helium: The aim here is to be clear about the intended purpose.

Regarding the storage node, you can either set the next_protocol_key to None or keep it the same as the current key—both options will function properly.

crates/walrus-service/src/node/config.rs
liquid-helium: Done.

crates/walrus-service/src/node/config.rs
liquid-helium: Thanks for the suggestion, change applied.

crates/walrus-service/src/node/config.rs
liquid-helium: Yes, this is to avoid partial writes in case of crash.
On unix system, we can simplify it to just use the rename(), since it is already atomic, but not Windows.

That's a great catch! Thanks! I added a serde_yaml::Value as a indirection to avoid remove any fields.

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -0,0 +1,125 @@
...
+            tracing::warn!("Config monitor is disabled, skipping background run");
+            // If disabled, wait forever instead of returning
+            std::future::pending::<()>().await;
+            unreachable!();
+        }
+
+        loop {
+            tokio::time::sleep(self.check_interval).await;
+
+            if let Err(e) = self.sync_node_params().await {
```
karlwuest, line 50: Question: Are we reloading the config anywhere? If not, why do we need to sync anything here?

contracts/walrus/sources/staking.move
```diff
@@ -213,6 +213,11 @@ public fun set_next_public_key(
     self.inner_mut().set_next_public_key(cap, public_key, proof_of_possession, ctx);
 }
 
+/// Resets the next public key of the storage node.
+public fun reset_next_public_key(self: &mut Staking, cap: &StorageNodeCap) {
+    self.inner_mut().reset_next_public_key(cap);
+}
+
```
karlwuest: Important: I don't think we need a separate function to reset this. If the node changes it's key back again, we can just set the next public key. At this point, we should not add public functions to the contracts unless absolutely needed.

contracts/walrus/sources/staking.move
karlwuest: I agree with @mlegner here. I don't think we need these resets.

contracts/walrus/sources/staking.move
liquid-helium: I see.

removed reset_next_public_key

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: No, that will be addressed in a subsequent PR, as the current one is already quite messy.

The existing background loop is essentially inactive and serves only as a placeholder. 
Node params are presently synced through manual calls during startup and at the beginning of epoch changes.

contracts/walrus/sources/staking.move
liquid-helium: Removed. Thanks @mlegner @karlwuest for the suggestion!

crates/walrus-service/src/node/errors.rs
```diff
@@ -345,3 +347,23 @@ pub enum SyncShardClientError {
...
+    ProtocolKeyPairRotationRequired,
+    /// The node configuration has changed.
+    #[error("Node needs reboot")]
+    NodeNeedsReboot,
+    /// A SuiClientError occurred.
+    #[error(transparent)]
+    SuiClientError(#[from] SuiClientError),
+    /// The node configuration is inconsistent with the on-chain configuration.
+    #[error("Node config inconsistent")]
+    NodeConfigInconsistent(String),
```
mlegner: Suggestion:
```suggestion
    #[error("Node config is inconsistent: {0}")]
    NodeConfigInconsistent(String),
```

crates/walrus-service/src/testbed.rs
```diff
@@ -640,7 +641,8 @@ pub async fn create_storage_node_configs(
                 node_capacity: node.node_capacity,
             },
             metrics_push: None,
-            metadata: Default::default(),
+            metadata: NodeMetadata::default(),
```
mlegner: Nit: I think this change is not relevant.
```suggestion
            metadata: Default::default(),
```

crates/walrus-sui/src/types.rs
```diff
@@ -132,6 +132,13 @@ pub struct UpdatePublicKeyParams {
     pub proof_of_possession: ProofOfPossession,
 }
 
+/// Action to take on the next public key of the storage node.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum UpdateNextPublicKeyAction {
+    /// Update the next public key of the storage node.
+    Update(UpdatePublicKeyParams),
+}
+
```
mlegner: Suggestion: Given we don't have the reset anymore, I think we also no longer need this enum.

crates/walrus-service/src/node/contract_service.rs
mlegner: Could we apply this also to the `run` function so we can simplify the matching in the `bin/node.rs` file?

crates/walrus-service/bin/node.rs
mlegner: Hmm... I agree that this should be a pretty rare occurrence. Can you create a follow-up issue to investigate how to optimize this and add a TODO comment in the code?

crates/walrus-service/src/node/config.rs
```diff
@@ -60,7 +81,13 @@ pub struct StorageNodeConfig {
     pub db_config: DatabaseConfig,
     /// Key pair used in Walrus protocol messages.
     #[serde_as(as = "PathOrInPlace<Base64>")]
+    #[serde(rename = "protocol_key_pair")]
```
mlegner: Question: Why do we need the `rename` here and below?

crates/walrus-service/src/node/config.rs
```diff
@@ -155,14 +186,79 @@ impl Default for StorageNodeConfig {
...
+    /// This happens when the on-chain protocol key pair has been rotated.
+    /// The protocol_key_pair is set to the new key pair, and the
+    /// next_protocol_key_pair is cleared.
+    pub fn rotate_protocol_key_pair_persist(path: impl AsRef<Path>) -> anyhow::Result<()> {
+        // Load config from path, preserving the raw Value to maintain all fields
+        let config_str = std::fs::read_to_string(path.as_ref())?;
+        let mut original_value: serde_yaml::Value = serde_yaml::from_str(&config_str)?;
+        let mut config: StorageNodeConfig = serde_yaml::from_str(&config_str)?;
+
+        assert!(config.protocol_key_pair.is_path());
```
mlegner: Important: Why do we have this `assert` here?

crates/walrus-service/src/node.rs
```diff
@@ -155,6 +156,10 @@ mod start_epoch_change_finisher;
 pub(crate) mod errors;
 mod storage;
 
+// Add new module
```
```suggestion
```

crates/walrus-service/src/node.rs
```diff
@@ -533,14 +485,45 @@ impl StorageNode {
...
-                .then(|| {
-                    tracing::warn!("Failed to sync node params: {}", e);
-                })
-                .ok_or(e)
-        })?;
+        tracing::info!(
+            "Creating StorageNode with config monitor enabled: {}, interval: {:?}",
+            config.config_synchronizer.enabled,
+            config.config_synchronizer.interval
+        );
```
mlegner: Remark: While I like the added visibility, I think we should either print all relevant configuration parameters or none.

crates/walrus-service/src/node.rs
```diff
@@ -533,14 +485,45 @@ impl StorageNode {
...
+            .await
+            .or_else(|e| match e {
+                SyncNodeConfigError::ProtocolKeyPairRotationRequired => Err(e),
+                SyncNodeConfigError::NodeNeedsReboot => {
+                    tracing::info!("Ignore the error since we are booting");
+                    Ok(())
+                }
+                _ => {
+                    if node_params.ignore_sync_failures {
+                        tracing::warn!("Failed to sync node params: {}", e);
```
mlegner: Nit: We generally start log entries with lowercase letters and try to record any relevant data as fields instead of adding them to the string directly:
```suggestion
                        tracing::warn!(error = ?e, "failed to sync node params");
```

This also applies to other added log entries.

crates/walrus-service/src/node.rs
```diff
@@ -533,14 +485,45 @@ impl StorageNode {
...
+                config.config_synchronizer.interval,
+            ))
+        } else {
+            Arc::new(ConfigSynchronizer::disabled(
+                config.clone(),
+                contract_service.clone(),
+                committee_service.clone(),
+                config.config_synchronizer.interval,
+            ))
+        };
```
mlegner: Question: Couldn't we just store an `Option<Arc<ConfigSynchronizer>>` instead of creating a disabled synchronizer?

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -0,0 +1,125 @@
...
+};
+
+/// Monitors and syncs node configuration with on-chain parameters.
+/// Syncs committee member information with on-chain committee information
+pub struct ConfigSynchronizer {
+    config: StorageNodeConfig,
+    contract_service: Arc<dyn SystemContractService>,
+    committee_service: Arc<dyn CommitteeService>,
+    check_interval: Duration,
+    enabled: AtomicBool,
```
mlegner: Related to a previous comment: Why do we need disabled config synchronizers?

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -25,20 +25,33 @@ use walrus_sui::{
...
-use super::committee::CommitteeService;
+use super::{committee::CommitteeService, config::StorageNodeConfig, errors::SyncNodeConfigError};
 use crate::common::config::SuiConfig;
 
 const MIN_BACKOFF: Duration = Duration::from_secs(1);
 const MAX_BACKOFF: Duration = Duration::from_secs(3600);
 
+enum ProtocolKeyAction {
+    UpdateRemoteNextPublicKey(PublicKey),
+    ResetRemoteNextPublicKey,
```
mlegner: Question: Do we actually need this option here?

crates/walrus-service/src/node/errors.rs
liquid-helium: Fix applied.

crates/walrus-service/bin/node.rs
liquid-helium: Create WAL-573!

crates/walrus-service/src/testbed.rs
liquid-helium: opps! Reverted!

crates/walrus-service/src/node/contract_service.rs
liquid-helium: Oh, nice suggestion, removed.

crates/walrus-service/src/node/contract_service.rs
liquid-helium: Let it return a StorageNodeError now!

crates/walrus-service/src/node.rs
liquid-helium: Print all node config now.

crates/walrus-service/src/node.rs
liquid-helium: Thanks for the tip, change applied.

crates/walrus-service/src/node.rs
liquid-helium: I like this, change applied.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: it requires working SystemContractService to work, and will fail with MockSystemContractService, etc.

Change it to an Option<> in StorageNode, so it won't need this bool now.

crates/walrus-sui/src/types.rs
liquid-helium: Change applied.

crates/walrus-service/src/node/config.rs
liquid-helium: Just being cautious—if anyone changes the field names (which may never happen), it could affect the rotateKey function, which relies on these names for matching.

crates/walrus-service/src/node/config.rs
liquid-helium: This is an interesting point—thanks for bringing it up! I'm not entirely sure if I'm approaching this correctly. 
As we consider writing out the keys, it appears that in our default setup (from generate_config()), we have separate files for the keys. When we switch to InPlace, will we be writing the keys directly to the config file instead?

crates/walrus-service/src/node.rs
liquid-helium: opps, removed.

crates/walrus-service/src/node/contract_service.rs
liquid-helium: I’ve reverted the above change because the error propagation is intertwined with other runtimes in monitor_runtimes, and those runtimes also read data from the config file. While we could modify all runtimes to return this new type, I feel that using anyhow::Error provides us with greater flexibility. What are your thoughts on this?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -197,11 +197,15 @@ mod tests {
...
-                &[1, 2, 3, 3, 4],
-                false,
-                ClientCommunicationConfig::default_for_test(),
-                None,
+                TestNodesConfig {
+                    node_weights: vec![1, 2, 3, 3, 4],
+                    use_legacy_event_processor: false,
+                    disable_event_blob_writer: false,
+                    blocklist_dir: None,
+                    enable_node_config_synchronizer: false,
```
halfprice: Nit: since these setup are duplicated many times, consider creating a TestNodesConfig::default

crates/walrus-service/src/test_utils.rs
```diff
@@ -316,33 +335,72 @@ impl SimStorageNodeHandle {
...
+                                    protocol_key = ?config_guard.protocol_key_pair().public(),
+                                    next_protocol_key = ?config_guard
+                                        .next_protocol_key_pair()
+                                        .map(|kp| kp.public()),
+                                    "rotating protocol key pair"
+                                );
+                                config_guard.rotate_protocol_key_pair();
+                                sui_simulator::task::kill_current_node(Some(Duration::from_secs(
+                                    10,
+                                )));
```
halfprice, line 387: Remark: the behavior here is a bit different from node.rs. Here it is a clean restart whereas in node.rs, we don't restart the process. Can we also use `loop {rotate; continue}` patten?

Ideally, if we can extract this as a common function so that both places can call, that will be great, but I can see that since both places update the config in different way, so maybe it's not so easy to do so.

crates/walrus-service/src/node/committee.rs
```diff
@@ -173,6 +173,10 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
         &self,
     ) -> Result<(), BeginCommitteeChangeError>;
 
+    /// Update information about committee members from the latest committee on chain.
+    /// This does not change the committee, but updates the configurations of the nodes.
```
halfprice: Nit: I think it maybe worth documenting that this sync is to primarily deal with other storage nodes' address change.

crates/walrus-service/src/node/contract_service.rs
mlegner: Ah, I see. Yes, that works for me.

crates/walrus-service/src/node/config.rs
mlegner: I agree with the goal but I'm not sure this actually achieves it: We already have various checks to avoid breaking changes for config files. How about instead adding a comment that if these fields are renamed, the `rotate_key` function needs to be updated as well?

crates/walrus-service/src/node/config.rs
mlegner: For the keys, we have two options:
1. *In place*: The key is added directly in the config file as a base64-encoded string.
2. *Path*: The config file contains a path to the key file.

I don't see a reason why we couldn't support the in-place option here as well, it should work exactly the same.

crates/walrus-service/src/node/config.rs
```diff
@@ -155,14 +186,79 @@ impl Default for StorageNodeConfig {
...
+            self.protocol_key_pair = next_key_pair;
+            self.next_protocol_key_pair = None;
+        }
+    }
+
+    /// Rotates the protocol key pair and persists the config to disk.
+    /// This happens when the on-chain protocol key pair has been rotated.
+    /// The protocol_key_pair is set to the new key pair, and the
+    /// next_protocol_key_pair is cleared.
+    pub fn rotate_protocol_key_pair_persist(path: impl AsRef<Path>) -> anyhow::Result<()> {
```
mlegner, line 212: Nit/suggestion: Define constants for the key strings `protocol_key_pair` and `next_ protocol_key_pair` to make them more visible.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -312,3 +412,201 @@ impl SystemContractService for SuiSystemContractService {
...
+
+            // If local has next key and it differs from remote's next key (or remote has none),
+            // update remote's next key
+            (Some(local_next), remote_next) => {
+                if Some(local_next.clone()) != remote_next {
+                    Ok(ProtocolKeyAction::UpdateRemoteNextPublicKey(local_next))
+                } else {
+                    Ok(ProtocolKeyAction::DoNothing)
+                }
+            }
```
mlegner: Nit: The `if` condition can be included in the pattern:
```suggestion
            (Some(local_next), remote_next) if Some(&local_next) != remote_next.as_ref() => {
                Ok(ProtocolKeyAction::UpdateRemoteNextPublicKey(local_next))
            }
```

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -312,3 +412,201 @@ impl SystemContractService for SuiSystemContractService {
...
+            local_public_key = ?local_public_key,
+            remote_public_key = ?remote_public_key,
+            local_next_public_key = ?local_next_public_key,
+            remote_next_public_key = ?remote_next_public_key,
+            error = error_msg,
+            "protocol key mismatch"
+        );
+
+        Err(SyncNodeConfigError::NodeConfigInconsistent(
+            error_msg.to_owned(),
```
mlegner: Nit: I think the `to_owned()` is not necessary here.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -312,3 +412,201 @@ impl SystemContractService for SuiSystemContractService {
...
+        // Check if local next key matches remote current key
+        if let Some(local_next) = local_next_public_key.clone() {
+            if local_next == remote_public_key {
+                // Warn if remote has a next key set
+                if remote_next_public_key.is_some() {
+                    tracing::warn!(
+                        local_public_key = ?local_public_key.clone(),
+                        remote_public_key = ?remote_public_key.clone(),
+                        local_next_public_key = ?local_next_public_key.clone(),
+                        remote_next_public_key = ?remote_next_public_key.clone(),
```
mlegner: Suggestion: Instead of recording all these fields for each log entry, you could add a `#[tracing::instrument]` to the function.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -312,3 +412,201 @@ impl SystemContractService for SuiSystemContractService {
...
+                }
+                return Ok(ProtocolKeyAction::RotateLocalKeyPair);
+            } else {
+                // Update remote's next key only if it differs from local next or is not set
+                tracing::error!(
+                    local_public_key = ?local_public_key.clone(),
+                    remote_public_key = ?remote_public_key.clone(),
+                    local_next_public_key = ?local_next_public_key.clone(),
+                    remote_next_public_key = ?remote_next_public_key.clone(),
+                    error = error_msg,
```
mlegner: Question: Why do we need this additional error message?

mlegner: Also, we may have two error logs in this case.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -312,3 +412,201 @@ impl SystemContractService for SuiSystemContractService {
...
+                    Ok(ProtocolKeyAction::DoNothing)
+                }
+            }
+
+            // Keys match or both have no next key - do nothing
+            _ => Ok(ProtocolKeyAction::DoNothing),
+        }
+    }
+    // Case 2: Local public key doesn't match remote public key
+    else {
```
mlegner, line 499: Remark: This layered `if` clause is pretty difficult to follow. I *think* I understand it and it is correct, but I'm not 100% sure... Is there any way to rewrite it to make it easier to understand?

crates/walrus-service/src/node.rs
```diff
@@ -533,14 +484,38 @@ impl StorageNode {
...
         let start_time = Instant::now();
-        sync_node_params(config).await.or_else(|e| {
-            node_params
-                .ignore_sync_failures
-                .then(|| {
-                    tracing::warn!("Failed to sync node params: {}", e);
-                })
-                .ok_or(e)
-        })?;
+        tracing::info!("creating StorageNode with config:\n{:?}", config);
```
mlegner: Important question: Does this print the private keys?

crates/walrus-simtest/tests/simtest.rs
liquid-helium: Done.

crates/walrus-service/src/test_utils.rs
liquid-helium: I share your concerns about the different code paths for the StorageNode bootstrap in production, testing, and simtest, this has been troubling me through out this PR.

The clean restart is also necessary for this reason; we need to ensure proper cleanup, or else the previous runtime might still hold onto resources.

Instead of making changes specifically for this issue, I believe we should prioritize unifying the bootstrap process, as it will become increasingly challenging to address this over time.

I will give it a try soon.

crates/walrus-service/src/node/config.rs
liquid-helium: Comments added!

crates/walrus-service/src/node/config.rs
liquid-helium: Contants defined.

crates/walrus-service/src/node/config.rs
liquid-helium: Thanks for the explaination and offline discussion, change applied.

crates/walrus-service/src/node/contract_service.rs
liquid-helium: Done!

crates/walrus-service/src/node/contract_service.rs
liquid-helium: Removed.

crates/walrus-service/src/node/contract_service.rs
liquid-helium: Thanks for the tip. #[tracing::instrument] added.

crates/walrus-service/src/node/contract_service.rs
liquid-helium: I revised it a little bit.

crates/walrus-service/src/node.rs
liquid-helium: opps! Nice catch, removed!

crates/walrus-service/src/node/committee.rs
liquid-helium: Done.

crates/walrus-service/src/test_utils.rs
halfprice: SG! It is always a pain point to not have a common `run` function between node in bin and node in simtest, and it has high priority on my todo list, but feel free to take it and appreciate the help!

crates/walrus-service/src/node/config.rs
```diff
@@ -80,13 +80,13 @@ pub struct StorageNodeConfig {
     #[serde(default, skip_serializing_if = "defaults::is_default")]
     pub db_config: DatabaseConfig,
     /// Key pair used in Walrus protocol messages.
+    /// Important: this name should be in-sync with the name used in `rotate_protocol_key_pair()`
```
mlegner: Nit: Can you make this a normal comment instead of a docstring? This is not something to communicate to the users of the code but to anyone modifying it.
```suggestion
    // Important: this name should be in-sync with the name used in `rotate_protocol_key_pair()`
```

crates/walrus-service/src/node/config.rs
```diff
@@ -80,13 +80,13 @@ pub struct StorageNodeConfig {
...
     #[serde_as(as = "PathOrInPlace<Base64>")]
-    #[serde(rename = "protocol_key_pair")]
     pub protocol_key_pair: PathOrInPlace<ProtocolKeyPair>,
     /// The next protocol key pair to use for the storage node.
+    /// Important: this name should be in-sync with the name used in `rotate_protocol_key_pair()`
```
mlegner: Nit: As above:
```suggestion
    // Important: this name should be in-sync with the name used in `rotate_protocol_key_pair()`
```

crates/walrus-service/src/node/committee.rs
```diff
@@ -175,6 +175,8 @@ pub trait CommitteeService: std::fmt::Debug + Send + Sync {
 
     /// Update information about committee members from the latest committee on chain.
```
mlegner, line 176: Nit: Add an empty line to get a nicer display on the Rust docs generated with `cargo doc` (separate heading from details).
```suggestion
    /// Update information about committee members from the latest committee on chain.
    ///
```

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -457,50 +454,39 @@ fn calculate_protocol_key_action(
...
+
+        if local_next_public_key.is_none() {
+            return Err(SyncNodeConfigError::NodeConfigInconsistent(error_msg));
+        }
+
+        // Check if local next key matches remote current key, this indicates that
+        // the on-chain protocol public key is updated, so we need to rotate the local key pair.
+        let local_next = local_next_public_key
+            .as_ref()
+            .expect("local next public key is not set");
```
mlegner: Nit: This could be simplified a bit (also removing the need for the `expect`):
```suggestion
        let Some(local_next) = local_next_public_key else {
            return Err(SyncNodeConfigError::NodeConfigInconsistent(error_msg));
        };

        // Check if local next key matches remote current key, this indicates that
        // the on-chain protocol public key is updated, so we need to rotate the local key pair.
```

crates/walrus-service/src/node/contract_service.rs
mlegner: Thanks, this helps a lot! 🙏

crates/walrus-service/src/node/config.rs
mlegner: Important: I still see the asserts in the code. Can you remove them?

crates/walrus-service/src/node/contract_service.rs
liquid-helium: nice!

PR #1462, opened 2025-01-23
feat: add `db-tool` command in walrus-node to access storage node DB

halfprice:
## Description

During production debugging, we often in a situation where a storage node is stuck in processing some events and we want to know what is that event, and possibly the blob associated with the event if any.

This PR added a `db-tool` command in `walrus-node` binary to allow direct read of the event/BlobInfo/PerObjectBlobInfo table.

The command is hidden by default and only meant for debugging purpose.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -35,7 +35,7 @@ use walrus_service::{
     common::config::SuiConfig,
     node::{
         config::{self, defaults::REST_API_PORT, StorageNodeConfig},
-        dbtool_commands::DbToolCommands,
+        dbtool::DbToolCommands,
         events::event_processor_runtime::EventProcessorRuntime,
         server::{RestApiConfig, RestApiServer},
         system_events::EventManager,
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -139,7 +139,7 @@ use crate::{
 pub mod committee;
 pub mod config;
 pub mod contract_service;
-pub mod dbtool_commands;
+pub mod dbtool;
 pub mod events;
 pub mod server;
 pub mod system_events;
--- a/crates/walrus-service/src/node/dbtool.rs
+++ b/crates/walrus-service/src/node/dbtool.rs
@@ -3,12 +3,14 @@
 
 //! Tools for inspecting and maintaining the RocksDB database.
 
-use std::{path::PathBuf, str::FromStr};
+use std::path::PathBuf;
 
 use anyhow::Result;
 use bincode::Options;
 use clap::Subcommand;
 use rocksdb::{Options as RocksdbOptions, DB};
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
 use sui_types::base_types::ObjectID;
 use typed_store::rocks::be_fix_int_ser;
 use walrus_core::BlobId;
@@ -17,7 +19,8 @@ use super::events::PositionedStreamEvent;
 use crate::node::{
     events::event_processor::EVENT_STORE,
     storage::blob_info::{
-        merge_mergeable,
+        blob_info_cf_options,
+        per_object_blob_info_cf_options,
         BlobInfo,
         PerObjectBlobInfo,
         AGGREGATE_BLOB_INFO_COLUMN_FAMILY_NAME,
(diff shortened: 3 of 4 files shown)


crates/walrus-service/src/node/dbtool_commands.rs
mlegner, line 1: Nit: Maybe just call this `dbtool.rs`?

crates/walrus-service/src/node/dbtool_commands.rs
```diff
@@ -0,0 +1,200 @@
...
+#[clap(rename_all = "kebab-case")]
+pub enum DbToolCommands {
+    /// Repair a corrupted RocksDB database due to non-clean shutdowns.
+    RepairDb {
+        /// Path to the RocksDB database directory.
+        #[clap(long)]
+        db_path: PathBuf,
+    },
+
+    /// Scan events from the event_store table in RocksDB.
```
mlegner: Question: What does "scan" mean here? Print?

crates/walrus-service/src/node/dbtool_commands.rs
```diff
@@ -0,0 +1,200 @@
...
+        count: u64,
+    },
+
+    /// Read blob info from the RocksDB database.
+    ReadBlobInfo {
+        /// Path to the RocksDB database directory.
+        #[clap(long)]
+        db_path: PathBuf,
+        /// Blob ID to read.
+        blob_id: String,
```
mlegner: Suggestion: Directly use `BlobId` here with a `serde_as(as = "DisplayFromStr)` similar to the `walrus` CLI (see [here](https://github.com/MystenLabs/walrus/blob/99dd4879a4e5bb17931bd86b6411bc12a98a7027/crates/walrus-service/src/client/cli/args.rs#L214)).

crates/walrus-service/src/node/dbtool_commands.rs
```diff
@@ -0,0 +1,200 @@
...
+    },
+
+    /// Read object blob info from the RocksDB database.
+    ReadObjectBlobInfo {
+        /// Path to the RocksDB database directory.
+        #[clap(long)]
+        db_path: PathBuf,
+        /// Object ID to read.
+        object_id: String,
```
mlegner: Suggestion: Directly use `ObjectID` here.

crates/walrus-service/src/node/dbtool_commands.rs
```diff
@@ -0,0 +1,200 @@
...
+    let db = DB::open_cf_with_opts_for_read_only(
+        &RocksdbOptions::default(),
+        db_path,
+        [(AGGREGATE_BLOB_INFO_COLUMN_FAMILY_NAME, blob_info_options)],
+        false,
+    )?;
+
+    let cf = db
+        .cf_handle(AGGREGATE_BLOB_INFO_COLUMN_FAMILY_NAME)
+        .expect("Aggregate blob info column family should exist");
```
mlegner: Question/suggestion: Could this be implemented in `crates/walrus-service/src/node/storage/blob_info.rs` instead? Same below for the per-object info.

crates/walrus-service/src/node/dbtool_commands.rs
```diff
@@ -0,0 +1,200 @@
...
+    let blob_id = BlobId::from_str(&blob_id)?;
+    println!("Reading blob info for Blob ID: {:?}", blob_id);
```
mlegner: Nit: I wouldn't use the Debug print here.
```suggestion
    println!("Reading blob info for Blob ID: {}", blob_id);
```

crates/walrus-service/src/node/dbtool_commands.rs
halfprice: `Scan` means the command is supposed to start from a index, and continue reading `--count` number of events.

crates/walrus-service/src/node/dbtool_commands.rs
halfprice: SG. Done.

crates/walrus-service/src/node/dbtool_commands.rs
halfprice: Done

crates/walrus-service/src/node/dbtool_commands.rs
halfprice: The open DB part cannot be reused. I instead create a helper to generate the cf option, and use them in both places. PTAL.

crates/walrus-service/src/node/dbtool_commands.rs
halfprice: Done

PR #1463, opened 2025-01-23
feat(node): allow setting metrics push config in setup command

mlegner:
## Description

Allow setting the metrics push configuration directly in the `walrus-node` setup command.

This includes some refactoring and cleanup.

## Test plan

Run the `walrus-node setup` command manually and inspect the resulting configuration.

---

## Release notes

- [x] Storage node: Allow setting the metrics push configuration directly in the `walrus-node` setup command.

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -56,11 +56,8 @@ use walrus_sui::{client::SuiContractClient, types::move_structs::VotingParams, u
 const VERSION: &str = version!();
 
 /// Manage and run a Walrus storage node.
-#[derive(Parser)]
-#[clap(rename_all = "kebab-case")]
-#[clap(name = env!("CARGO_BIN_NAME"))]
-#[clap(version = VERSION)]
-#[derive(Debug)]
+#[derive(Debug, Parser)]
+#[clap(rename_all = "kebab-case", name = env!("CARGO_BIN_NAME"), version = VERSION)]
 struct Args {
     #[command(subcommand)]
     command: Commands,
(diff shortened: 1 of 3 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -209,62 +207,66 @@ struct SetupArgs {
...
     /// processing.
     ///
     /// If not provided, the RPC node from the wallet's active environment will be used.
+    #[clap(long)]
     sui_rpc: Option<String>,
-    #[clap(long, action)]
     /// Use the legacy event provider instead of the standard checkpoint-based event processor.
-    use_legacy_event_provider: bool,
     #[clap(long, action)]
+    use_legacy_event_provider: bool,
```
halfprice, line 246: not for this PR, but just curious: do we still need the legacy event provider? Seems the checkpoint based event processor is stable enough.

crates/walrus-service/src/node/config.rs
```diff
@@ -456,6 +455,16 @@ pub mod defaults {
         2000
     }
 
+    /// Configure the default push interval for metrics.
+    pub fn push_interval() -> Duration {
+        Duration::from_secs(60)
+    }
+
+    /// Configure the default push interval for metrics.
```
halfprice: Nit: the comment does not seem to match the function?

crates/walrus-service/bin/node.rs
mlegner: Indeed, we should remove that soon ([WAL-405: Remove legacy event provider](https://linear.app/mysten-labs/issue/WAL-405/remove-legacy-event-provider)).

crates/walrus-service/src/node/config.rs
mlegner: Whoops, copy-paste error. 😬

PR #1464, opened 2025-01-23
feat: add highest finished event index in health endpoint

halfprice:
## Description

So that we know which event the node stucks on.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/node/storage/event_cursor_table.rs
```diff
@@ -175,14 +180,23 @@ impl EventCursorTable {
...
 
-        Ok(EventProgress { persisted, pending })
+        // Update highest_finished_event_index if the current event_index is higher
+        let prev_index = self
+            .highest_finished_event_index
+            .fetch_max(event_index, Ordering::SeqCst);
+        Ok(EventProgress {
+            persisted,
+            pending,
+            highest_finished_event_index: std::cmp::max(prev_index, event_index),
```
liquid-helium, line 190: nit: use AtomicU64::max to get the max directly, since we are not using prev_index.

liquid-helium: my bad, ignore this!

PR #1465, opened 2025-01-23
chore: remove sui as dependency for publishing contracts

karlwuest:
## Description

contributes to WAL-518

hopefully fixes the stack overflow during publishing

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -105,6 +105,8 @@ jobs:
     needs: diff
     if: ${{ needs.diff.outputs.isRelevantForRustTests == 'true' }}
     runs-on: ubuntu-ghcloud
+    env:
+      RUST_LOG: warn
     steps:
       - uses: taiki-e/install-action@v2.47.18
         with:
--- a/crates/walrus-sui/src/system_setup.rs
+++ b/crates/walrus-sui/src/system_setup.rs
@@ -81,9 +81,8 @@ fn get_pkg_id_from_tx_response(tx_response: &SuiTransactionBlockResponse) -> Res
 pub(crate) async fn publish_package_with_default_build_config(
     wallet: &mut WalletContext,
     package_path: PathBuf,
-    gas_budget: Option<u64>,
 ) -> Result<SuiTransactionBlockResponse> {
-    publish_package(wallet, package_path, Default::default(), gas_budget).await
+    publish_package(wallet, package_path, Default::default(), None).await
 }
 
 #[tracing::instrument(err, skip(wallet))]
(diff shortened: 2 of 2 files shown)


.github/workflows/code.yml
```diff
@@ -105,9 +105,6 @@ jobs:
     needs: diff
     if: ${{ needs.diff.outputs.isRelevantForRustTests == 'true' }}
     runs-on: ubuntu-ghcloud
-    env:
-      RUST_LOG: warn
```
halfprice, line 109: suggestion: keep the `warn`? With `info`, any log e2e test failure is not readable in CI, and take too long to load.

crates/walrus-sui/src/system_setup.rs
```diff
@@ -67,45 +78,153 @@ fn get_pkg_id_from_tx_response(tx_response: &SuiTransactionBlockResponse) -> Res
         .ok_or_else(|| anyhow!("no immutable object was created"))
 }
 
+pub(crate) async fn publish_package_with_default_build_config(
+    wallet: &mut WalletContext,
+    package_path: PathBuf,
+    gas_budget: Option<u64>,
```
halfprice: Nit: all the call sites use None as gas_budget. Consider removing it if we are not using yet?

crates/walrus-sui/src/utils.rs
```diff
@@ -450,3 +464,63 @@ pub fn generate_proof_of_possession_for_address(
...
+        .await
+        .context("dry run failed")?
+        .effects;
+    let gas_cost_summary = effects.gas_cost_summary();
+
+    let safe_overhead = GAS_SAFE_OVERHEAD * reference_gas_price;
+    let computation_cost_with_overhead = gas_cost_summary.computation_cost + safe_overhead;
+    let gas_usage_with_overhead = gas_cost_summary.net_gas_usage() + safe_overhead as i64;
+    Ok(computation_cost_with_overhead.max(gas_usage_with_overhead.max(0) as u64))
+}
```
halfprice, line 500: Suggestion: I feel this is probably not necessary... Any downside if we always use `MAX_GAS_BUDGET` in publishing tx? The CLI gas estimation is a heuristic, and I feel duplicate it here will only add confusion.

karlwuest: Initially, I had it set up to either use a gas budget passed in as command line argument or twice the default gas budget (as default gas budget, we currently use 0.5 SUI, which is not sufficient for publishing). Always using the `MAX_GAS_BUDGET` is a bit overkill, since that's 50 SUI.

One of the reasons why I added this here is that we probably also want to use this for other operations in the future, so it's not just for the publishing (WAL-516). I think our current approach with a rigid gas budget is not ideal.

PR #1467, opened 2025-01-24
feat: Add metrics to track network throughput for different http endpoints

sadhansood:
## Description

This lets us track response throughput of client requests vs recovery and would be useful in debugging issues in the future hopefully.

## Test plan

Deployed on private testnet and verified the results.

[
Screenshot 2025-01-23 at 11 12 18 PM
](url)

### DIFF
--- a/crates/walrus-service/src/common/telemetry.rs
+++ b/crates/walrus-service/src/common/telemetry.rs
@@ -14,6 +14,7 @@ use std::{
 };
 
 use axum::{
+    body::Body,
     extract::{ConnectInfo, MatchedPath, State},
     http::{
         self,
@@ -22,6 +23,7 @@ use axum::{
     },
     middleware,
 };
+use futures::StreamExt;
 use opentelemetry::propagation::Extractor;
 use prometheus::{
     core::{AtomicU64, Collector, GenericGauge},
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/common/telemetry.rs
```diff
@@ -259,30 +285,88 @@ pub(crate) async fn metrics_middleware(
...
+    let request = Request::from_parts(parts, counted_body);
 
     let response = next.run(request).await;
 
-    let histogram =
-        metrics.with_label_values(&[method.as_str(), &route, response.status().as_str()]);
-    histogram.observe(start.elapsed().as_secs_f64());
+    // Record response metrics
+    let method_clone = method.clone();
+    let route_clone = route.clone();
```
halfprice, line 304: Nit: it seems that we don't need `clone` here?

crates/walrus-service/src/common/telemetry.rs
```diff
@@ -42,6 +46,28 @@ use super::active_committees::ActiveCommittees;
 /// Route string used in metrics for invalid routes.
 pub(crate) const UNMATCHED_ROUTE: &str = "invalid-route";
 
+/// HTTP metrics for tracking request/response statistics
+#[derive(Debug, Clone)]
+pub(crate) struct HttpMetrics {
+    pub(crate) duration: HistogramVec,
+    pub(crate) request_size: CounterVec,
+    pub(crate) response_size: CounterVec,
```
jpcsmith: Suggestion: Could you also add histograms for the size of request/response bodies, so that we can also get insight into the individual requests?

crates/walrus-service/src/common/telemetry.rs
```diff
@@ -259,30 +285,88 @@ pub(crate) async fn metrics_middleware(
...
+        "request_size_bytes_total",
+        "Total size of HTTP requests in bytes.",
+    )
+    .namespace("http");
+
+    let response_size_opts = prometheus::Opts::new(
+        "response_size_bytes_total",
+        "Total size of HTTP responses in bytes.",
+    )
+    .namespace("http");
```
jpcsmith, line 343: Suggestion: Rename to `request_bytes_total` and `response_bytes_total`.

crates/walrus-service/src/common/telemetry.rs
sadhansood: done

PR #1468, opened 2025-01-24
chore: refactor SuiContractClient

### LINKED ISSUE #1023: Client: improve locking of the wallet
Currently we ensure that the client does not submit conflicting transactions by locking the wallet and passing it around while it's needed.

Another option would be to require a mutable reference to the client itself, and lock that.

We would need to investigate the changes required for this, and the performance implications.

[Originating comment.](https://github.com/MystenLabs/walrus/pull/1014#pullrequestreview-2357658806)

halfprice:
## Description

Create SuiContractClientInner and put all PTB related functionality in it. SuiContractClient holds a Mutex<SuiContractClientinner> so that sending transactions do not overlap.

Close #1023 
Close WAL-261

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/subsidies/Move.lock
+++ b/contracts/subsidies/Move.lock
@@ -50,6 +50,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.39.1"
+compiler-version = "1.41.0"
 edition = "2024.beta"
 flavor = "sui"
--- a/contracts/wal/Move.lock
+++ b/contracts/wal/Move.lock
@@ -21,6 +21,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.39.1"
+compiler-version = "1.41.0"
 edition = "2024.beta"
 flavor = "sui"
--- a/contracts/wal_exchange/Move.lock
+++ b/contracts/wal_exchange/Move.lock
@@ -30,6 +30,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.39.1"
+compiler-version = "1.41.0"
 edition = "2024.beta"
 flavor = "sui"
--- a/contracts/walrus/Move.lock
+++ b/contracts/walrus/Move.lock
@@ -40,6 +40,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.39.1"
+compiler-version = "1.41.0"
 edition = "2024.beta"
 flavor = "sui"


crates/walrus-sui/src/client.rs
```diff
@@ -269,12 +269,13 @@ pub type SuiClientResult<T> = Result<T, SuiClientError>;
 
 /// Client implementation for interacting with the Walrus smart contracts.
 pub struct SuiContractClient {
-    wallet: Mutex<WalletContext>,
+    inner: Mutex<SuiContractClientInner>,
     /// Client to read Walrus on-chain state.
     pub read_client: SuiReadClient,
```
halfprice: We leave SuiReadClient here since SuiContractClient needs to implement ReadClient trait and I don't think we want to always acquire lock to access it. SuiContractClientInner also needs to use SuiReadClient. I'm currently just cloning it. But can change to Arc<> if needed.

crates/walrus-sui/src/client.rs
```diff
@@ -868,7 +1183,7 @@ impl SuiContractClient {
         match authorized {
             Authorized::Address(receiver) => {
                 ensure!(
-                    receiver == self.wallet_address,
+                    receiver == self.wallet.active_address()?,
```
halfprice, line 1235: I guess the main goal of making all these function a `&self mut` is that we can directly get active address from the wallet, which also requires a `&mut`, is that correct?

crates/walrus-sui/src/client.rs
karlwuest: I think using `Arc` may be better because then we're sure that the read clients are in sync after a contract upgrade (i.e. when calling `refresh_package_id`).

crates/walrus-sui/src/client.rs
```diff
@@ -396,32 +360,444 @@ impl SuiContractClient {
...
+}
+
+struct SuiContractClientInner {
+    wallet: WalletContext,
+    read_client: SuiReadClient,
+    gas_budget: u64,
+}
+
+impl SuiContractClientInner {
+    /// Constructor for [`SuiContractClient`] with an existing [`SuiReadClient`].
```
```suggestion
    /// Constructor for [`SuiContractClientInner`] with an existing [`SuiReadClient`].
```

crates/walrus-sui/src/client.rs
karlwuest: No, the main goal is that we have exclusive access to make sure that we don't have race conditions between reads and creating the transactions. Essentially the same that we had before with the `critical_section` / locking the wallet before reads, but cleaner.

For this function, just having a immutable reference (if we would still store the address in the client) should actually be fine. We just need to have `&mut self` for all functions that create a transaction. Having to call `self.wallet.active_address()` may be a good way to enforce this, though.

crates/walrus-sui/src/client.rs
halfprice: SG. Done.

crates/walrus-sui/src/client.rs
halfprice: Make sense.

crates/walrus-sui/src/client.rs
halfprice: Done

contracts/subsidies/Move.lock
```diff
@@ -50,6 +50,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.41.0"
+compiler-version = "1.39.1"
```
karlwuest: Nit: Can you roll these back?

crates/walrus-service/aggregator_openapi.html
```diff
@@ -22,7 +22,7 @@
...
...schema":{"$ref":"#/components/schemas/Status"}}}}}}}},"components":{"schemas":{"BlobId":{"type":"string","format":"byte","description":"The ID of a blob.","examples":["E7_nNXvFU_3qZVu3OH1yycRG7LZlyn1-UxEDCDDqGGU"]},"Status":{"type":"object","description":"A message returned from a failed API call.\n\nContains both human-readable and machine-readable details of the error,\nto assist in resolving the error.","required":["error"],"properties":{"error":{"allOf":[{"oneOf":[{"type":"object","required":["status","code"],"properties":{"code":{"type":"integer","format":"int32","description":"HTTP status code associated with the error.","minimum":0},"status":{"type":"string","description":"General type of error, given as an UPPER_SNAKE_CASE string."}}}],"description":"The status code corresponding to the error."},{"type":"object","required":["message","details"],"properties":{"details":{"type":"array","items":{"type":"object"},"description":"Machine readable details of the error.\n\nAlways contains an [`ErrorInfo`], which provides a machine-readable\nrepresentation of the of the `message` field."},"message":{"type":"string","description":"A message describing the error in detail."}}}]}}}}}},
```
karlwuest: Question: Did you figure out where these changes come from?

crates/walrus-sui/src/client.rs
```diff
@@ -399,32 +367,460 @@ impl SuiContractClient {
...
+            .reserve_and_register_blobs(epochs_ahead, blob_metadata_list, persistence)
+            .await
+    }
+
+    /// Certifies the specified blob on Sui, given a certificate that confirms its storage and
+    /// returns the certified blob.
+    ///
+    /// If the post store action is `share`, returns a mapping blob ID -> shared_blob_object_id.
+    // NB: This intentionally takes an owned `Blob` object even though it is not required, as the
+    // corresponding object on Sui will be changed in the process.
```
karlwuest: Nit: This is unrelated to your PR, but this comment doesn't seem to hold anymore, maybe you can just delete it.

crates/walrus-sui/src/client.rs
```diff
@@ -1041,52 +1309,40 @@ impl SuiContractClient {
...
-        )
-        .await
+        self.sign_and_send_ptb_inner(programmable_transaction, 0, minimum_gas_coin_balance)
+            .await
     }
 
     // TODO(giac): Currently we pass the wallet as an argument to ensure that the caller can lock
     // before taking the object references. This ensures that no race conditions occur. We could
     // consider a more ergonomic approach, where this function takes `&mut self`, and the whole
     // client needs to be locked. (#1023).
```
karlwuest: Suggestion: This comment can now be deleted.

crates/walrus-service/aggregator_openapi.html
mlegner: This is automatically updated because the version changed. I'm currently preparing a PR to set this to a dummy value.

mlegner: --> https://github.com/MystenLabs/walrus/pull/1481

Should be fixed after rebasing.

crates/walrus-sui/src/client.rs
halfprice: Done.

crates/walrus-sui/src/client.rs
halfprice: Yeah, missed this. Removed. Thanks!

contracts/subsidies/Move.lock
halfprice: Sorry for not catching this. I was using an outdated `sui` version locally...

PR #1469, opened 2025-01-24
refactor(publisher): simplify auth layer

giac-mysten:
## Description

- simplifies the authentication code by removing the custom layer implementation, and using `axum::middleware::from_fn_with_state`
- parses the query parameters using the `Query` extractor, such that they are correctly typed
- parses the secret and creates the decoding key only at startup.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/config.rs
+++ b/crates/walrus-service/src/client/config.rs
@@ -114,7 +114,9 @@ pub struct AuthConfig {
     pub(crate) algorithm: Option<Algorithm>,
     /// The duration, in seconds, after which the publisher will consider the JWT as expired.
     ///
-    /// If set to `0`, the JWT will never be considered expired.
+    /// If set to `0`, the publisher will not check that the expiration is correctly set based in
+    /// the issued-at time (iat) and expiration time (exp) in the JWT. I.e., if `expiring_sec > 0`,
+    /// the publisher will check that `exp - iat == expiring_sec`.
     pub(crate) expiring_sec: u64,
     /// verify upload epochs and address for `send_object_to`
     pub(crate) verify_upload: bool,


crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -73,208 +74,142 @@ impl Claim {
...
+        Ok(claim) => {
+            let mut valid_upload = true;
+            if auth_config.expiring_sec > 0
+                && (claim.exp - claim.iat.unwrap_or_default()) != auth_config.expiring_sec
+            {
+                error!(toker = bearer.token(), "token with invalid expiration");
+                valid_upload = false;
+            }
+            // TODO(giac): We never actually check that the expiration date is in the future.
+            // i.e., check that claim.exp > SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)
```
giac-mysten: @yanganto just checking, is this correct?

giac-mysten: ah now i see that the library automatically checks the `exp` claim

crates/walrus-service/src/client/config.rs
```diff
@@ -109,40 +110,57 @@ pub enum ExchangeObjectConfig {
 }
 
 /// Configuration for the JWT authentication on the publisher.
-#[derive(Default, Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+#[allow(missing_debug_implementations)]
```
jpcsmith: Hint: You could consider adding a manual implementation of the Debug trait using [`debug_struct`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct]).

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -1,19 +1,20 @@
...
 use serde::Deserialize;
-use tower::{Layer, Service};
+use sui_types::base_types::SuiAddress;
 use tracing::error;
 
+use super::routes::PublisherQuery;
 use crate::client::config::AuthConfig;
 
-/// Claim follow RFC7519 with extra storage parameters: address, epoch
+/// Claim follow RFC7519 with extra storage parameters: send_object_to, epochs.
```
jpcsmith: Nit:
```suggestion
/// Claim follows RFC7519 with extra storage parameters: send_object_to, epochs.
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -73,208 +74,142 @@ impl Claim {
...
+    } else {
+        Validation::default()
+    };
+
+    let default_key = DecodingKey::from_secret(&[]);
+    let decode_key = auth_config.decoding_key.as_ref().unwrap_or_else(|| {
+        // No decoding key is provided in the configuration, so we disable signature validation.
+        validation.insecure_disable_signature_validation();
+        &default_key
+    });
```
jpcsmith, line 103: Question: Why disable signature verification but not the entire verification? 

Remark: This may give the impression that authentication is turned on, since some JWT tokens are rejected, whereas the reality is that signatures are not being checked so anyone can get around the auth by attaching a passing, unsigned token.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -73,208 +74,142 @@ impl Claim {
...
+            // i.e., check that claim.exp > SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)
+
+            if auth_config.verify_upload {
+                if let Some(epochs) = claim.epochs {
+                    if query.epochs != epochs {
+                        tracing::error!(
+                            expected = claim.epochs,
+                            actual = query.epochs,
+                            "upload with invalid epochs"
+                        );
```
jpcsmith: Suggestion: Perhaps tracing::debug/trace here as this is not an error with the system, but a bad request from a user.

The same can be said about the errors below.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -73,208 +74,142 @@ impl Claim {
...
 
-fn validation_failed_future<B, E>(code: StatusCode) -> Ready<Result<Response<B>, E>>
+fn validation_failed_response<B>(code: StatusCode) -> Result<(), Response<B>>
 where
     B: Default,
 {
-    std::future::ready(Ok(Response::builder()
+    Err(Response::builder()
         .status(code)
         .body(Default::default())
```
jpcsmith: Suggestion: Make the response a detailed JSON error response. See the storage node errors for examples (i.e., define an Error enum, and then `derive(RestApiError)` and annotate each variant with a status code and reason.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: This is to remain compatible with @yanganto 's original implementation. I agree that this is a bit confusing.

@yanganto maybe you can add a bit of context to this choice?

yanganto: Yes, there are two things we will check with the token.
1. the token is valid, which need to check the signature
2. the upload payload is exactly the same as description in these extra fields in the token

We can enable all these verification on the publisher with `jwt_decode_secret`(1) and `jwt_verify_upload`(2). 

People may use publisher with different network infra.  For example, an API Gateway can set in front of publisher, which can easily to check the (1) but not (2), so this is the using case to disable (1) and keep (2) enabled if the signature of token is already checked by gateway.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: Yes, it is checked by the lib.  We can remove the TODO.

crates/walrus-service/src/client/config.rs
```diff
@@ -109,40 +110,57 @@ pub enum ExchangeObjectConfig {
...
+#[derive(Default, Clone)]
 pub struct AuthConfig {
     /// The secret with which to authenticate the JWT.
-    pub(crate) secret: Option<Vec<u8>>,
+    pub(crate) decoding_key: Option<DecodingKey>,
     /// The authentication algorithm for the JWT.
     pub(crate) algorithm: Option<Algorithm>,
     /// The duration, in seconds, after which the publisher will consider the JWT as expired.
+    ///
+    /// If set to `0`, the JWT will never be considered expired.
```
yanganto: If set to `0`, the `exp` of JWT still is considered and verified in lib based on current time, so it is still expired if the request comes too late.. 

If set to `0`, the publisher will not check the `exp` is correctly based on `iat`.

giac-mysten: Good catch! updated the docstring.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: Here is the context in previous PR help you to know the context.
https://github.com/MystenLabs/walrus/pull/1345#discussion_r1906265194

PR #1471, opened 2025-01-27
fix: Only retry on unexpected transaction execution errors while certifying event blobs

sadhansood:
## Description

This PR addresses the issue discovered with event blob writer not retrying aborted transaction due to congestion. 

## Test plan
Existing tests

### DIFF
--- a/crates/walrus-service/src/node/contract_service.rs
+++ b/crates/walrus-service/src/node/contract_service.rs
@@ -22,7 +22,10 @@ use walrus_sui::{
         SuiClientError,
         SuiContractClient,
     },
-    types::move_structs::EpochState,
+    types::{
+        move_errors::{MoveExecutionError, SystemStateInnerError},
+        move_structs::EpochState,
+    },
 };
 use walrus_utils::backoff::{self, ExponentialBackoff};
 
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/node/contract_service.rs
```diff
@@ -236,7 +239,15 @@ impl SystemContractService for SuiSystemContractService {
                             "Storage node capability object not set");
                         Some(())
                     }
-                    Err(SuiClientError::TransactionExecutionError(e)) => {
+                    Err(
+                        e @ SuiClientError::TransactionExecutionError(
+                            MoveExecutionError::SystemStateInner(
+                                SystemStateInnerError::EInvalidIdEpoch(_)
+                                | SystemStateInnerError::EIncorrectAttestation(_)
+                                | SystemStateInnerError::ERepeatedAttestation(_),
```
karlwuest: I think it might be better to explicitly match on the errors that we want to retry instead of just matching the ones that we know we don't want to retry. 

Otherwise, it's easy to miss an error that will just cause all retried transactions to fail again. E.g. we don't want to retry the `SystemStateInnerError::ENotCommitteeMember`. On the other hand, we also may not want the errors that we're not retrying to just be treated as success. Perhaps we should log them at least as a warning?

Perhaps, we could have three groups of errors:
1. Expected errors that can be treated as success.
2. Unexpected errors that we probably don't want to retry (e.g. all other errors that are `MoveAbort`s. Those, we definitely need to log at least as `WARN`, but retrying is likely useless.
3. Rare errors that we know may happen, such as the congestion, which we do want to retry (and log). For now, we could just match on `MoveExecutionError::NotParsable(e)`, which covers all transaction execution errors that are not caused by a move abort. However, we should probably add explicit error types for some of those, since we may not want to retry all of them.

sadhansood: Thanks Karl. I've categorized the errors as: expected (no retry needed), unexpected non-retriable, expected retriable and unexpected retriable. 
Expected errors are: `EInvalidEpoch` for instance as we know this is possible during epoch changes.
Unexpected non-retriable: `EIncorrectAttestation`, `ERepeatedAttestation`, etc as we know retrying them is useless. Move execution errors (which get parsed to a known error type) also fall under this category. They get printed as warn!
Expected retriable: ShardObjectCongestion falls under this category. It make sense to retry them. They get printed as warn!
Unexpected retriable: Non-parsed move execution errors, and all other sui client errors. They get retried as we don't know what went wrong really and retrying them is likely going to fix it. They get printed as error!. I am not a big fan of this category myself. A better way is to let `certify_event_blob` fail and still let event processing continue. I am working on a follow-up change to do this decoupling as well.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -236,18 +239,59 @@ impl SystemContractService for SuiSystemContractService {
...
+                                SystemStateInnerError::EInvalidIdEpoch(_)
+                            ),
+                        ),
+                    ) => {
                         tracing::debug!(
                             walrus.epoch = epoch,
                             error = ?e,
                             blob_id = ?blob_id,
-                            "Transaction execution error while attesting event blob"
+                            "Non-retryable event blob certification error while attesting event blob"
```
karlwuest: Nit (for consistency):
```suggestion
                            "Non-retriable event blob certification error while attesting event blob"
```

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -236,18 +239,59 @@ impl SystemContractService for SuiSystemContractService {
...
+                                | SystemStateInnerError::ERepeatedAttestation(_)
+                                | SystemStateInnerError::ENotCommitteeMember(_),
+                            ),
+                        ),
+                    ) => {
+                        tracing::warn!(
+                            walrus.epoch = epoch,
+                            error = ?e,
+                            blob_id = ?blob_id,
+                            "Unexpected non-retryable event blob certification error while attesting event blob"
```
karlwuest:
```suggestion
                            "Unexpected non-retriable event blob certification error while attesting event blob"
```

crates/walrus-sui/src/client.rs
```diff
@@ -136,6 +136,29 @@ pub enum SuiClientError {
...
+            .map(|objects_match| {
+                objects_match
+                    .as_str()
+                    .split(", ")
+                    .filter_map(|id| ObjectID::from_hex_literal(id).ok())
+                    .collect::<Vec<_>>()
+            })
+            .map(Self::SharedObjectCongestion)
+            .ok_or_else(|| anyhow::anyhow!("not a congestion error: {}", error))
+    }
```
karlwuest: Suggestion: I think having an explicit function instead of `try_from` would be preferable here. From a `try_from(&str)` I would expect that other error variants could also result instead of just the `SharedObjectCongestion`.

PR #1479, opened 2025-01-28
fix: Use the right rest api address during startup

sadhansood:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -890,7 +890,7 @@ impl StorageNodeRuntime {
         rest_api_address.set_ip(IpAddr::V4(Ipv4Addr::UNSPECIFIED));
         let rest_api_handle = tokio::spawn(async move {
             let result = rest_api
-                .run()
+                .run(rest_api_address)
                 .await
                 .inspect_err(|error| tracing::error!(?error, "REST API exited with an error"));
 
@@ -901,7 +901,7 @@ impl StorageNodeRuntime {
 
             result
         });
-        tracing::info!("started REST API on {}", node_config.rest_api_address);
+        tracing::info!("started REST API on {}", rest_api_address);
 
         Ok(Self {
             runtime,
--- a/crates/walrus-service/src/node/server.rs
+++ b/crates/walrus-service/src/node/server.rs
@@ -160,7 +160,7 @@ where
     }
 
     /// Runs the server, may only be called once for a given instance.
-    pub async fn run(&self) -> Result<(), std::io::Error> {
+    pub async fn run(&self, address: SocketAddr) -> Result<(), std::io::Error> {
         {
             let handle = self.handle.lock().await;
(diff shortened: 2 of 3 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -890,7 +890,7 @@ impl StorageNodeRuntime {
         rest_api_address.set_ip(IpAddr::V4(Ipv4Addr::UNSPECIFIED));
```
halfprice, line 890: Hi @sadhansood . Can you describe what is the problem of the old code? To me, it seems that we just need to remove line 889 and 890, and always use the address specified in the config in the rest server.

sadhansood: Hey @halfprice - I notice this in adhoc cluster where we specify the ip in the configs but upon node startup trying to listen on that node's ip, the server fails to bind to the network endpoint sometimes. For instance if the ip is 10.0.0.2 and we spin up storage node trying to listen on 10.0.0.2:9185, it randomly fails to bind to the endpoint and fails to startup. I don't really know if there is some timing issue (like network interface might not be fully up when we are starting the storage node), or another security/firewall issue tbh. But, listening on all interfaces with 0.0.0.0:9185 works a lot more reliably. 
I see the code is trying to do the same thing i.e. switch to Ipv4Addr::UNSPECIFIED but because of this bug, that is not happening. Also, this pattern of using  Ipv4Addr::UNSPECIFIED is used for metrics server in this codebase.

halfprice: I see. Thanks @sadhansood for the explanation! This changes from listen to specific IP to 0.0.0.0 in the default case too. Should we just add an option or use 0.0.0.0 in the walrus-deploy and use it only in the testing environment? I'm not sure if we want to allow walrus-node to be accessed from any network interface.

mlegner: I believe we should indeed remove the two lines 889 and 890. I think always using `Ipv4Addr::UNSPECIFIED` for the metrics address ([here](https://github.com/MystenLabs/walrus/blob/9a86ee28e275a575985a81ac36706d0107e47a51/crates/walrus-service/src/common/utils.rs#L299)) is also a bug.

The defaults (if not specified in the config file) are `127.0.0.0` for the [metrics address](https://github.com/MystenLabs/walrus/blob/5b21595fe56e801faa5ca239ae65ff9edaab1ec7/crates/walrus-service/src/node/config.rs#L464) and `0.0.0.0` for the [REST API address](https://github.com/MystenLabs/walrus/blob/5b21595fe56e801faa5ca239ae65ff9edaab1ec7/crates/walrus-service/src/node/config.rs#L469).

mlegner: I created a PR (#1630) for this.

PR #1484, opened 2025-01-28
fix(audit): early withdraw does not update `active_set`

damirka:
## Description

Cases:

- [ ] stake + early withdraw -> node is not expected in the committee

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/active_set.move
+++ b/contracts/walrus/sources/staking/active_set.move
@@ -59,6 +59,7 @@ public(package) fun new(max_size: u16, threshold_stake: u64): ActiveSet {
 /// Inserts the node if it is not already in the active set, otherwise updates its stake.
 /// Returns true if the node is in the set after the operation, false otherwise.
 public(package) fun insert_or_update(set: &mut ActiveSet, node_id: ID, staked_amount: u64): bool {
+    // Currently, the `threshold_stake` is set to `0`, so we need to account for that.
     if (staked_amount == 0 || staked_amount < set.threshold_stake) {
         set.remove(node_id);
         return false


contracts/walrus/sources/staking/active_set.move
```diff
@@ -59,10 +59,13 @@ public(package) fun new(max_size: u16, threshold_stake: u64): ActiveSet {
 /// Inserts the node if it is not already in the active set, otherwise updates its stake.
 /// Returns true if the node is in the set after the operation, false otherwise.
 public(package) fun insert_or_update(set: &mut ActiveSet, node_id: ID, staked_amount: u64): bool {
-    if (set.update(node_id, staked_amount)) {
-        return true
+    if (staked_amount == 0 || staked_amount < set.threshold_stake) {
```
karlwuest, line 63:
```suggestion
    if (staked_amount < set.threshold_stake) {
```

karlwuest: Actually, why don't we keep this as it was before and remove the entry in `update` instead?

contracts/walrus/sources/staking/active_set.move
```diff
@@ -87,9 +90,6 @@ public(package) fun update(set: &mut ActiveSet, node_id: ID, staked_amount: u64)
 public(package) fun insert(set: &mut ActiveSet, node_id: ID, staked_amount: u64): bool {
     assert!(set.nodes.find_index!(|entry| entry.node_id == node_id).is_none(), EDuplicateInsertion);
 
-    // Check if the staked amount is enough to be included in the active set.
-    if (staked_amount < set.threshold_stake) return false;
-
```
karlwuest, line 92: Question: Why do we not want to keep this in?

contracts/walrus/sources/staking/active_set.move
damirka: One main reason: `insert` is never called separately and I'm voting to make it a private function (update is called however)

damirka: As to your correction - in epoch 0 threshold stake is `0`, so we cannot remove this condition.

PR #1486, opened 2025-01-28
feat(client): support multiple node IDs and URLs for health command

mlegner:
## Description

Adds the ability to provide a list of node IDs or URLs for the `walrus health` command. This is a breaking change for the JSON mode because `nodeId` and `nodeUrl` are changed to `nodeIds` and `nodeUrls`, respectively. Through the alias this is backwards-compatible for the "normal" CLI.

Includes some simplifications and refactoring within `walrus_service::client` and `walrus_sui::client`.

Currently, if a node ID is not found, the error message is not very useful, as it doesn't contain the concrete ID that was not found.

Closes WAL-562.

## Test plan

Manually running some variants of the `walrus health` command:

```sh
cargo run --bin walrus -- health --node-ids 0xb763f0defd9d23eb6f6bd98371c4d6559749c264d0ddbcad0b1ba9bc917119f9 0x81dae13f34dfe76170901f9f24a920c3b361acee7136bbc2ef6a0a15bfa085c8
cargo run --bin walrus -- health --node-urls mia-tnt-sto-00.walrus-testnet.walrus.space:9185 uc1-tnt-sto-00.walrus-testnet.walrus.space:9185
cargo run --bin walrus -- health --committee
```

---

## Release notes

- [x] CLI: Adds the ability to provide a list of node IDs or URLs for the `walrus health` command. This is a breaking change for the JSON mode because `nodeId` and `nodeUrl` are changed to `nodeIds` and `nodeUrls`, respectively. The "normal" CLI is backwards-compatible because `--node-id` and `--node-url` are set as aliases for the new options.

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -135,9 +135,6 @@ pub enum SuiClientError {
     /// The sender is not authorized to perform the action on the pool.
     #[error("the sender is not authorized to perform the action on the pool with node ID {0}")]
     NotAuthorizedForPool(ObjectID),
-    /// The storage node is not found.
-    #[error("the storage node {0} is not found")]
-    StorageNodeNotFound(ObjectID),
     /// Transaction execution was cancelled due to shared object congestion
     #[error("execution cancelled due to shared object congestion on objects {0:?}")]
     SharedObjectCongestion(Vec<ObjectID>),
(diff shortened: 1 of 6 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -1675,16 +1672,16 @@ impl ReadClient for SuiContractClient {
         self.read_client.next_committee().await
     }
 
-    async fn get_storage_nodes_from_active_set(&self) -> SuiClientResult<Vec<StorageNode>> {
+    async fn get_storage_nodes_from_active_set(&self) -> Result<Vec<StorageNode>> {
```
giac-mysten, line 1675: Question: why change to anyhow?

mlegner: Because this calls the `get_sui_objects`, which returns an `anyhow::Result`.

PR #1487, opened 2025-01-28
feat(publisher): add checks for max size and max epochs

giac-mysten:
## Description

- adds the possibility to specify `max_epochs` in the claim
- adds the possibliity to specify the `max_size` in the claim

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon.rs
+++ b/crates/walrus-service/src/client/daemon.rs
@@ -41,7 +41,7 @@ use crate::{
     common::telemetry::{metrics_middleware, register_http_metrics, MakeHttpSpan},
 };
 
-mod auth;
+pub mod auth;
 mod openapi;
 mod routes;
 
--- a/crates/walrus-service/src/client/daemon/auth.rs
+++ b/crates/walrus-service/src/client/daemon/auth.rs
@@ -9,7 +9,7 @@ use jsonwebtoken::{
     DecodingKey,
     Validation,
 };
-use serde::Deserialize;
+use serde::{Deserialize, Serialize};
 use sui_types::base_types::SuiAddress;
 use tracing::error;
 use walrus_core::EpochCount;
@@ -21,8 +21,7 @@ use crate::{client::config::AuthConfig, common::api::RestApiError};
 pub const PUBLISHER_AUTH_DOMAIN: &str = "auth.publisher.walrus.space";
 
 /// Claim follow RFC7519 with extra storage parameters: send_object_to, epochs.
-#[derive(Clone, Deserialize, Debug, Default)]
-#[cfg_attr(test, derive(serde::Serialize))]
+#[derive(Clone, Debug, Default, Serialize, Deserialize)]
 pub struct Claim {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     /// Token is issued at (timestamp).


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -210,6 +232,38 @@ pub(super) async fn put_blob<T: WalrusWriteClient>(
     response
 }
 
+/// Checks if the JWT claim has a maximum size and if the blob exceeds it.
+///
+/// IMPORTANT: This function does _not_ check the validity of the claim (i.e., does not
+/// authenticate the signature). The assumption is that a previous middleware has already done
+/// so.
+///
+/// The function just decodes the token and checks that the size in the claim is not exceeded.
```
giac-mysten, line 241: Remark: here is my main concern aboit this PR. The function does not re-verify the signature because it assumes that, if there is a JWT in the header, it has already been verified by a previous layer (or it does not metter if it is verified). Does this sound correct? 

Also, if the claim cannot be decoded without verification, I return an "internal server error". This also sounds ok to me, but flagging to get another opinion.

crates/walrus-service/src/client/config.rs
```diff
@@ -118,7 +118,9 @@ pub struct AuthConfig {
     /// the issued-at time (iat) and expiration time (exp) in the JWT. I.e., if `expiring_sec > 0`,
     /// the publisher will check that `exp - iat == expiring_sec`.
     pub(crate) expiring_sec: u64,
-    /// verify upload epochs and address for `send_object_to`
+    /// Verify the upload epochs and address for `send_object_to` in the request.
+    ///
+    /// The token expiration is still checked, even if `verify_upload == true`.
```
yanganto, line 123: Do we need to mention any about token expiration in the `verify upload` field? I thought it is a separated feature.

crates/walrus-service/src/client/daemon.rs
```diff
@@ -270,7 +271,14 @@ pub(crate) async fn auth_layer(
     request: Request,
     next: Next,
 ) -> Response {
-    if let Err(resp) = verify_jwt_claim(query, bearer_header, &auth_config) {
+    // Get a hint on the body size if possible.
+    // Note: this is a preliminary check. It is fine if the body is actually larger than this,
+    // because we will check again the size when storing to Walrus. Hence, we can use `0` as the
+    // body size safely, if the hint is not available.
```
yanganto: Could we straightforward say
"Try to get a body hint to reject a oversize payload as fast as possible"

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -79,12 +93,140 @@ impl Claim {
...
+    }
+
+    /// Checks if the claim has expired yet.
+    fn is_expired(&self, auth_config: &AuthConfig) -> bool {
+        if auth_config.expiring_sec == 0 {
+            return false;
+        }
+
+        // Check that the difference between `exp` and `iat` is equal to the configured
+        // TODO(giac): should we allow <= instead of == ?
```
yanganto: No! Just use `==` here.

As I mentioned before, please do not back and forth at this point
https://github.com/MystenLabs/walrus/pull/1345#discussion_r1909869604

There is more literally meaning to check `exp - iat`, because `iat` is not required, and the validity of JWT is from `exp` not `iat`. 

The `exp - iat` is a simple second factor for the security.
- The first factor is the issuer **has** the secret key
- The second factor is the issuer **know** what is the expiring second
It is 2 factors for security, if you enable `expiring_sec`.

It is better not to loose the security feature.  We can remove all the `iat` things by disabling `expiring_sec`. This check is for security, if not, we just wast computing power to check nothing really meaningful.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -35,8 +36,21 @@ struct Claim {
...
     /// The number of epochs the blob should be stored for.
+    ///
+    /// This is an exact number of epochs the blob should be stored for, no more, no less.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub epochs: Option<u32>,
+
+    /// The maximum number of epochs the blob can be stored for (inclusive).
+    ///
+    /// If both `epochs` and `max_epochs` are present, `max_epochs` must be greater than or equal
+    /// to `epochs`. In this case, `max_epochs` is disregarded, and `epochs` is used.
```
yanganto: Will this better to let `epochs` and `max_epochs` conflict and reject earlier?  I do not see any meaning to handle a both settings from a uncertainty issuer.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -79,12 +93,140 @@ impl Claim {
...
+            return Err(error);
+        }
+
+        self.check_send_object_to(query.send_object_to)?;
+
+        Ok(())
+    }
+
+    /// Checks if the claim has expired yet.
+    fn is_expired(&self, auth_config: &AuthConfig) -> bool {
```
yanganto: It will be less confuse to use `check_expiring_sec` not `is_expired`.  
A expired token is the `exp` less than current time, it is nothing about the `iat`.  Please also make sure the doc of `fn` will be correct.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -79,12 +93,140 @@ impl Claim {
...
+                }
+            }
+            (Some(epochs), Some(max_epochs)) => {
+                if epochs > max_epochs {
+                    return Err(PublisherAuthError::InvalidMaxEpochs);
+                }
+                if query_epochs == epochs {
+                    Ok(())
+                } else {
+                    Err(PublisherAuthError::InvalidEpochs)
```
yanganto: It maybe good to return error than wasting computing on checking for a uncertainty issuer.  The issuer wastes the payload overhead on the internet and the computing on the publisher.

crates/walrus-service/src/client/daemon/routes.rs
yanganto: The web is an onion structure.  Each layer focus on different things, a layer can be a software layer included the program or a infra from other machine to share the work loading.

There are two things can be done in the auth layer, and it is possible to let other infra share the loading to verify the signature, mentioned in https://github.com/MystenLabs/walrus/pull/1469#discussion_r1935503649

The `check_blob_size` is nothing about the signature, so it is good not to mix thing up.  
However, the `check_blob_size` is derived from `verify_upload` by original idea.
https://github.com/MystenLabs/walrus/blob/1017e31a466c03caaa4c9b488c4e54d67944d179/crates/walrus-service/src/client/cli/args.rs#L612-L615  I think this is good interface to let user decide to verify the upload or not.  We can remove `IMPORTANT` of doc stirngs, the `fn check_blob_size` never say it will check signature, and it should not do it in my view.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -12,17 +12,18 @@ use jsonwebtoken::{
...
 
 pub const PUBLISHER_AUTH_DOMAIN: &str = "auth.publisher.walrus.space";
 
-/// Claim follows RFC7519 with extra storage parameters: send_object_to, epochs.
-#[derive(Clone, Deserialize, Debug)]
+/// Claim follow RFC7519 with extra storage parameters: send_object_to, epochs.
+#[derive(Clone, Deserialize, Debug, Default)]
 #[cfg_attr(test, derive(serde::Serialize))]
-struct Claim {
+pub(crate) struct Claim {
```
yanganto: It is good to keep this public.  Once the walrus open source, developer can easy to reuse the structure to build other web service with publisher.

crates/walrus-service/src/client/daemon/routes.rs
giac-mysten: cool, glad you agree.
> We can remove IMPORTANT of doc stirngs, the fn check_blob_size never say it will check signature, and it should not do it in my view.
i would keep the docstring, to make sure that it is clear that, even if we take the `Authorization<Bearer>` header, we don't check the signature.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: I remember the conversation. 

However, let me stress that this is _not_ a security feature. At best it's a sanity check. 
The security is fully dependent on the signature: If an adversary can forge the signature, it can also very quickly discover what the `exp - iat` is (for example, by trial and error, or even more by just observing a valid token) and then forge the token.

I'll keep this as is for now, but I still think the extra flexibility would be nice, with no loss of security.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: Sure, can do that.

crates/walrus-service/src/client/config.rs
giac-mysten: I think so, to resolve confusion about what `verify_upload` means.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: When talking about multi factors, it is good to open the textbook again.  It is much more than OTP code.

**CISSP Domain 5: identity and access management(IAM)**
 Authentication section
 
- Authentication by knowledge (Something you **know**)
- Authentication by ownership (Something you **have**)
- How you behave or your physiology (something you are)

That means.
- issuer **know** the `expiring_sec` and use correct `exp - iat` to prove
- issuer **has** the `secret` and use `signature` to prove

In my point of view, 
- The `exp - iat` is made by a setting of digital machine and not analogical detector, why we need to give a flexibility?
- If an error log (not Rust Error) raised in auth layer, the secret should change.  
- If the system not in developing anymore, the error is essentially an critical level.

So there should no space even once for trial and error by attacker, because this error only happen when we still in developing integration systems.

crates/walrus-service/src/client/daemon/routes.rs
yanganto: The exactly size and hash check is already in my action items for a while.
https://github.com/sui-foundation/walrus-storage-nft-drop/tree/main/publisher#action-items

If you are doing the exactly size not the max size, you save my time.  

The exactly `size` is much more important than a `max-size`.  

Because the token is a session thing should not be long term for security, and our using case is the issuer already know what will upload, so we will use `size` not `max-size`.  

Besides, I can not imagine why an issuer will give a token with a long session token with `max-size`?  The system will not be determined anymore, and let unknown files comes in, however I will not against to have `max-size` or `max-epochs`, and will avoid using it.

crates/walrus-service/src/client/config.rs
yanganto: `verify_upload` verify the uploads to the system.  It should be clear from doc of the [jwt_verify_upload](https://github.com/MystenLabs/walrus/blob/403c5f0287422d0e139bfd8f02bea999cd21fe26/crates/walrus-service/src/client/cli/args.rs#L605-L615) of cli args.

For a storage system, the object/file upload comes in and take time and space, so following things are included.

- time - the epoch user want to store is valid or not
- space - the size user want to store is valid or not
- integrity - the file is exactly the shard_id (file size is the simple version of integrity)
- ownership - the blob will send to

So the user enable it, then publisher will check `epoch`, `size`, `max-epoch`, etc.  If a non public personal uploading, he can upload without JWT token.  

In other words, the `max-size`, `max-epoch` take actions only when`verify_upload` is true.  
Note: [jwt_verify_upload](https://github.com/MystenLabs/walrus/blob/403c5f0287422d0e139bfd8f02bea999cd21fe26/crates/walrus-service/src/client/cli/args.rs#L605-L615) is not [jwt_decode_secret](https://github.com/MystenLabs/walrus/blob/403c5f0287422d0e139bfd8f02bea999cd21fe26/crates/walrus-service/src/client/cli/args.rs#L592) setting, so it is nothing about the signature, but still rely on JWT token.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: If you are willing to pick this, we can save time.
https://github.com/MystenLabs/walrus/pull/1536

It is still good to keep this in separate due to different issue.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: I still disagree: The `expiring_sec` is _public knowledge_, because anyone with a JWT can see it.
Imagine in the case of the storage airdrop:

- the adversary gets an airdrop nft
- the adversary claims the nft and gets a JWT
- the adversary looks at the JWT and now knows `expiring_sec`

Therefore, the only secret -- and the only factor -- that the issuer holds is the secret key.

Anyways, let's move forward.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: yes, I added this here.

PR #1489, opened 2025-01-28
feat: add batch endpoint for recovery symbols

jpcsmith:
## Description

This adds the list endpoint for recovery symbols. It will be disabled until all nodes have deployed this version of the storage node.

## Test plan

This is primarily a refactor, so existing tests should pass. Several minor tests were also added.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -335,7 +335,7 @@ impl RecoverySymbolsFilter {
     }
 }
 
-fn serialize_ids_in_query<S>(symbols: &Vec<SymbolId>, serializer: S) -> Result<S::Ok, S::Error>
+fn serialize_ids_in_query<S>(symbols: &[SymbolId], serializer: S) -> Result<S::Ok, S::Error>
 where
     S: Serializer,
 {
@@ -807,7 +807,7 @@ impl Client {
         struct EmptyResponse;
 
         let mut symbols = self
-            .list_recovery_symbols(metadata.blob_id(), &filter)
+            .list_recovery_symbols(metadata.blob_id(), filter)
             .await?;
         tracing::trace!(
             n_symbols = symbols.len(),
@@ -822,7 +822,7 @@ impl Client {
             )
             .entered();
 
-            if !filter.accepts(&symbol) {
+            if !filter.accepts(symbol) {
                 tracing::warn!("server returned a symbol with an unrequested proof axis");
                 return false;
             }


crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -474,6 +479,541 @@ where
...
+
+    fn decode_sliver_by_axis<A: EncodingAxis, I>(
+        &self,
+        recovery_symbols: I,
+    ) -> Option<Result<Sliver, InconsistencyProofEnum>>
+    where
+        I: IntoIterator<Item = RecoverySymbolData<A, MerkleProof>>,
+        SliverData<A>: Into<Sliver>,
+        InconsistencyProof<A, MerkleProof>: Into<InconsistencyProofEnum>,
+    {
```
jpcsmith, line 635: Remark: These functions are largely unchanged from that of `LegacyRecoverSliver`.

crates/walrus-sdk/src/client.rs
```diff
@@ -64,8 +74,10 @@ const SLIVER_STATUS_TEMPLATE: &str =
 const PERMANENT_BLOB_CONFIRMATION_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/confirmation/permanent";
 const DELETABLE_BLOB_CONFIRMATION_URL_TEMPLATE: &str =
     "/v1/blobs/:blob_id/confirmation/deletable/:object_id";
-const RECOVERY_URL_TEMPLATE: &str =
+const LEGACY_RECOVERY_URL_TEMPLATE: &str =
     "/v1/blobs/:blob_id/slivers/:sliver_pair_index/:sliver_type/:target_pair_index";
+const RECOVERY_SYMBOL_URL_TEMPLATE: &str = "/v1/blobs/:blob_id/recoverySymbols/:symbol_id";
```
halfprice, line 79: Question: you may have mentioned this before. Can you remind again why you need this endpoint? It seems that `LIST_RECOVERY_SYMBOLS_URL_TEMPLATE` can also query specific symbol?

crates/walrus-sdk/src/client.rs
```diff
@@ -214,6 +240,80 @@ impl UrlEndpoints {
...
+        /// The ID of the target sliver being recovered.
+        target: SliverIndex,
+        /// The type of the sliver being recovered.
+        target_type: SliverType,
+    },
+}
+
+impl RecoverySymbolsFilter {
+    /// Returns a new filter for the identified list of symbols.
+    pub fn ids(ids: Vec<SymbolId>, target_type: SliverType) -> Self {
```
halfprice: Nit: the ids probably shouldn't be empty.

crates/walrus-sdk/src/client.rs
```diff
@@ -608,19 +708,113 @@ impl Client {
...
+            .list_recovery_symbols(metadata.blob_id(), filter)
+            .await?;
+        tracing::trace!(
+            n_symbols = symbols.len(),
+            "the server returned recovery symbols"
+        );
+        let mut final_error = NodeError::other(EmptyResponse);
+
+        symbols.retain(|symbol| {
+            if symbol.proof_axis() == filter_target_type {
```
halfprice: Question: is this even possible? If the remote node generated a proof from the same type as the target type, is that indicates that the remote sliver is the same as the one being recovered? Just wonder if this should be an assertion instead of dropping it.

crates/walrus-sdk/src/client.rs
jpcsmith: We don't at the moment, but there's no harm keeping it in in the event we find use for it shortly after main-net. (E.g., point lookup in blobs).

crates/walrus-sdk/src/client.rs
jpcsmith: Ideally we would not have this check at all (i.e., not limit the symbols returned to coming from any given axis), because it does not matter: the proof can be verified irrespective of the axis and the symbol can be used to reconstruct the intended sliver. This check was as the `InconsistencyProof` has not been updated to use the new `GeneralRecoverySymbol`s, and so, this is needed to ensure we have only 1 type of recovery symbol for constructing the proof.

> If the remote node generated a proof from the same type as the target type, is that indicates that the remote sliver is the same as the one being recovered?

That's correct, but for the recovery-use-case, there's nothing in the protocol preventing a remote storage node from tracking more shards than assigned to it, and serving responses based on the additional shards stored. Other use-cases may simply not care about the proof axis, and for those use-cases this check would be removed.

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -469,6 +478,79 @@ pub async fn get_recovery_symbol_by_id<S: SyncServiceState>(
...
+
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, utoipa::ToSchema)]
+#[serde(rename_all = "camelCase", untagged)]
+pub(crate) enum ListRecoverySymbolsFilter {
+    /// Limit the results to the specified symbols.
+    #[serde(rename_all = "camelCase")]
+    Id {
+        /// One or more symbol IDs to be returned.
+        #[serde_as(as = "OneOrMany<_>")]
```
giac-mysten, line 502: Nice, did not know that `OneOrMany` existed. Maybe we could use it to simplify the syntax in the JSON client? 🤔

crates/walrus-sdk/src/client.rs
giac-mysten: just to try out my understanding 🙂 : this endpoint does not require query parameters, which may a bit cumbersome if requesting a single symbol. The other endpoint is more generic, but also requires more boilerplate to make the query... maybe?

crates/walrus-sdk/src/client.rs
giac-mysten: Well, it indicates the remote node is byzantine. And with an assertion it could crash the local node by just sending an incorrect response 😅 I think it's correct to skip it, but maybe we should log it with a warning?

crates/walrus-sdk/src/client.rs
```diff
@@ -608,19 +708,113 @@ impl Client {
...
+                return false;
+            }
+
+            if let Err(error) = symbol.verify(
+                metadata.metadata(),
+                encoding_config,
+                target_index,
+                target_type,
+            ) {
+                tracing::debug!(?error, "recovery symbol verification failed");
```
giac-mysten: Suggestion: maybe also increase the log level here?

crates/walrus-sdk/src/client.rs
jpcsmith: Yea, the resource collection is `/v1/blobs/:blob_id/recoverySymbols`. A `GET` to `/v1/blobs/:blob_id/recoverySymbols/:symbol_id` retrieves a single resource by its name (id), whereas a `GET` to the collection endpoint `/v1/blobs/:blob_id/recoverySymbols` lists all the symbols in the collection. Since we don't want to return them all, the query parameters are required.

crates/walrus-sdk/src/client.rs
```diff
@@ -608,19 +708,113 @@ impl Client {
...
+    #[tracing::instrument(
+        skip_all, fields(walrus.blob_id = %metadata.blob_id(),), err(level = Level::DEBUG)
+    )]
+    pub async fn list_and_verify_recovery_symbols(
+        &self,
+        filter: RecoverySymbolsFilter,
+        metadata: &VerifiedBlobMetadataWithId,
+        encoding_config: &EncodingConfig,
+        target_index: SliverIndex,
+        target_type: SliverType,
```
halfprice, line 803: Nit: `target_index` and `target_type` seem redundant and can be derived from `filter`?

crates/walrus-sdk/src/client.rs
halfprice: Maybe also log the symbol id? (Since we only return the last error after the `retain`)

crates/walrus-service/src/node.rs
```diff
@@ -2106,6 +2129,69 @@ impl ServiceState for StorageNodeInner {
...
+                            SymbolId::new(pair_stored.to_sliver_index::<Primary>(n_shards), target)
+                        }
+                    }
+                });
+                output.extend(symbol_ids.filter_map(|id| try_for_symbol(id, target_type)));
+            }
+        }
+
+        if output.is_empty() {
+            Err(last_error)
```
halfprice, line 2181: Suggestion: for lagging node, it possible that it doesn't have any slivers stored yet, and another node tries to recover a blob. I feel `NoSymbolsSpecified` maybe confusing. Can we create a case for `NoSymbolsAvailable`, or simply return EmptyResponse in this case?

crates/walrus-service/src/node/server.rs
```diff
@@ -1198,4 +1214,33 @@ mod tests {
...
+    async fn list_recovery_symbols(filter: RecoverySymbolsFilter) {
+        let (config, _handle) = start_rest_api_with_test_config().await;
+        let client = storage_node_client(config.as_ref());
+        let blob_id = walrus_core::test_utils::random_blob_id();
+
+        let symbols = client
+            .list_recovery_symbols(&blob_id, filter)
+            .await
+            .expect("request should succeed");
+        assert!(symbols.len() >= 2);
```
halfprice, line 1253: Question: just to verify my understanding of this test. The `retrieve_multiple_recovery_symbols` in `MockServiceState` always return 2 symbols. This test just tests that the request/response works with rest api server, is that correct?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -474,6 +479,567 @@ where
...
+                    match result {
+                        Ok(n_symbols) => {
+                            tracing::trace!(
+                                %n_symbols,
+                                "successfully collected the desired number of recovery symbols"
+                            );
+                            return self.decode_sliver(symbol_tracker);
+                        },
+                        Err(n_symbols_remaining) => {
+                            tracing::trace!(
```
halfprice, line 584: Suggestion: since this error is on sliver level, consider change this to info!.

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -474,6 +479,567 @@ where
...
+/// Since we request multiple symbols from each storage node, track the remaining set of storage
+/// nodes from the committee, as well as their shards from which we have not yet requested symbols.
+///
+/// Storage nodes are returned in a random order, with the chance of a node appearing earlier being
+/// proportional to the number of shards that it has.
+#[derive(Debug, Clone, Default)]
+struct RemainingShards {
+    // Store the node indices as u16 instead of usize to save on space.
+    upcoming_nodes: VecDeque<u16>,
+    next_shard_index: usize,
```
halfprice: Suggestion: can you add some doc for `next_shard_index`? It may be easily confused with ShardIndex. (Rather this is the index of the shard in the list of shards that the node owns). Took me a while to understand this.

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -474,6 +479,567 @@ where
...
+        shared: &'a NodeCommitteeServiceInner<T>,
+    ) -> Self {
+        // Clear counts of in-progress collections.
+        tracker.clear_in_progress();
+
+        let upcoming_nodes = if let Some(committee) = committee.upgrade() {
+            let mut rng_guard = shared.rng.lock().expect("mutex not poisoned");
+            RemainingShards::new(&committee, &mut rng_guard)
+        } else {
+            RemainingShards::default()
```
halfprice, line 705: Question: is this ok to use a default here because the `CollectRecoverySymbols` should be cancelled soon?

crates/walrus-sdk/src/client.rs
halfprice: Asking just because this endpoint does not seem to be used currently.

crates/walrus-sdk/src/client.rs
halfprice: Good point about byzantine that I didn't think about. I was more thinking about catching bugs and don't silently drop a potential issue. How about a debug_assert + warning?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -474,6 +479,567 @@ where
...
+    fn refill_pending_requests(&mut self) {
+        let mut new_request_count = 0;
+        let Some(committee) = self.committee.upgrade() else {
+            tracing::trace!("committee has been dropped, skipping refill");
+            return;
+        };
+
+        while let Some((node_index, shard_ids)) = self
+            .upcoming_nodes
+            .take_at_most(self.tracker.number_of_symbols_to_request(), &committee)
```
halfprice, line 752: Note (something to consider but may not need to change in this PR): when number_of_symbols_to_request() is smaller than next node's shard count, here we only request exact number_of_symbols_to_request() from the next node. I think this is good when all the remote nodes can return all the shards to not sending extra symbols back, but potentially require a few extra RPCs if some of the earlier requests are missing symbols.

I feel if we just always request all symbols from the node, (1) take_at_most() implementation can be simpler, and (2) sliver recovery may have lower latency (one of the issue we saw earlier was that we only request extra symbol when seeing a failure from earlier request, which adds up the latency of sliver recovery).

crates/walrus-sdk/src/client.rs
jpcsmith: Yes. I raised this point in the meeting on the 20th of Jan, and the response was that we or someone else may find a use for this endpoint in testnet/mainnet, so leave it for the moment.

crates/walrus-sdk/src/client.rs
jpcsmith: I am conflicted about increasing the verbosity level for log messages originating from the verification of responses coming from untrusted sources.

On one hand, it would help spotting an issue if one of the nodes we control return an invalid or unexpected response during development. On the other hand, since in production the message is coming from a possibly byzantine source, it enables a log-based attack vectors as warning messages are almost always enabled in production systems. 

I've changed both this and the level at `if symbol.proof_axis() == filter_target_type {` to `tracing::warn!` for now, but it's something we should re-evaluate at the system matures.

crates/walrus-sdk/src/client.rs
jpcsmith: Changed this to returning an Option if the list is empty.

crates/walrus-sdk/src/client.rs
jpcsmith: Refactored this such that there is now a `proof_axis` which is optional and identifies the desired axis from which proofs should originate. The 2 variants now are just `Vec<SymbolId>` and `(index, slivertype)`.

crates/walrus-service/src/node.rs
jpcsmith: It should not be possible for `NoSymbolsSpecified` to be returned if the storage node has any shards. For each shard it has, it attempts to retrieve the sliver/symbol which either results in a success or an error which overwrites the last error.

crates/walrus-service/src/node/server.rs
jpcsmith: Yes, it's just checking the parsing of the headers and the ability for the client-server to communicate.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: Given the frequency and quantity, I am a bit reluctant to place info messages in this loop, how about debug?

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: Added a doc and changed the variable name to `shard_id_range_start`.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: Yes, added a comment to that effect.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: Yea, that's a good point. My concern was that currently every additional symbol requires a database read and a full expansion of a sliver. The sliver expansions are not cheap and the extra data will just be discarded, so I erred on the side of not requiring them from the storage nodes, but we can re-evaluate later.

crates/walrus-service/src/node.rs
halfprice: Oh I see. Maybe `NoSymbolsAvailable` is closer to what the error is about?

crates/walrus-service/src/node/committee/request_futures.rs
halfprice: Sounds good.

crates/walrus-service/src/node.rs
jpcsmith: The original code could have been clearer. I refactored it now to create an iterator that generates the symbol IDs that the user has specified (either explicitly or implicitly). Then, we iterate over those symbols, if the iterator is non-empty (i.e., a symbol was specified), then our result will either be a list of at least 1 success or `ListSymbolsError::Last(_)`. If the iterator is empty, then the error is `ListSymbolsError:NoSymbolsSpecified`.

```rust
        let mut output = vec![];
        let mut last_error = ListSymbolsError::NoSymbolsSpecified;

        // If a specific proof axis is requested, then specify the target-type to the retrieve
        // function, otherwise, specify only the symbol IDs.
        let target_type_from_proof = filter.proof_axis().map(|axis| axis.orthogonal());

        for symbol_id in symbol_id_iter {
            match self.retrieve_recovery_symbol(blob_id, symbol_id, target_type_from_proof) {
                Ok(symbol) => output.push(symbol),

                // Callers may request symbols that are not stored with this shard, or
                // completely invalid symbols. These are ignored unless there are no successes.
                Err(error) => {
                    tracing::debug!(%error, %symbol_id, "failed to get requested symbol");
                    last_error = error.into();
                }
            }
        }
```

PR #1495, opened 2025-01-29
chore: fix dockerfile contracts location

pei-mysten:
## Description

kaniko is a little bit particular about relative folder location, so we use a absolute path for that

## Test plan

pulled the image and inspected, the `contracts/` folder is in the right place


docker/walrus-service/Dockerfile
```diff
@@ -11,7 +11,8 @@ RUN apt-get update && apt-get install -y cmake clang
 
 COPY Cargo.toml Cargo.lock ./
 COPY crates crates
-COPY contracts contracts
+COPY contracts ./contracts
```
mlegner, line 14: Question: Why does it work for `crates` but not `contracts`?

PR #1478, opened 2025-01-27
feat: connect walrus-backup to Google Cloud Storage

wbbradley:
## Description

Connect walrus-backup to Google Cloud Storage.

## Test plan

Manual testing via local-testbed.sh

No release impact.

PR #1496, opened 2025-01-30
chore: backup service should use a different builder image

pei-mysten:
## Description

`walrus-backup` binary added dependencies of `libpg5.so`
in this case, binaries that does not need `libpg5.so` should use a different folder

## Test plan

built an image based on this branch: `us-central1-docker.pkg.dev/cryptic-bolt-398315/walrus/walrus-service:b49b90ac8a8f73ad3ee9183fe5baf61759b1ab68`

pulled and tried to run `walrus-node` and `walrus-deploy`, both are working fine without complaining missing `libpg.so.5`
image

### DIFF
--- a/docker/walrus-service/Dockerfile
+++ b/docker/walrus-service/Dockerfile
@@ -14,18 +14,9 @@ COPY crates crates
 COPY contracts ./contracts
 COPY contracts /contracts
 
-FROM base AS backup-builder
-RUN cargo build --features walrus-service/backup \
-    --profile $PROFILE \
-    --bin walrus \
-    --bin walrus-backup \
-    --bin walrus-node \
-    --bin walrus-deploy
-
 FROM base AS builder
 RUN cargo build --profile $PROFILE \
     --bin walrus \
-    --bin walrus-backup \
     --bin walrus-node \
     --bin walrus-deploy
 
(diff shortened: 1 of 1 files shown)


docker/walrus-service/Dockerfile
```diff
@@ -78,13 +76,21 @@ ARG GIT_REVISION
 LABEL build-date=$BUILD_DATE
 LABEL git-revision=$GIT_REVISION
 
+FROM base AS backup-builder
+RUN cargo build --features walrus-service/backup \
+    --profile $PROFILE \
+    --bin walrus \
+    --bin walrus-backup \
+    --bin walrus-node \
+    --bin walrus-deploy
```
mlegner, line 85: Question: Why do we build the other binaries (`walrus`, `walrus-node`, and `walrus-deploy`) here as well?

wbbradley: I have a change in my current PR to remove these.

wbbradley: https://github.com/MystenLabs/walrus/pull/1478/files#diff-4cdff7fdfeffb77244cffed1fac8d91c6b812236dc9c32b91ff80a6877491879L82

PR #1498, opened 2025-01-30
fix(staking performance): unite iterations into one in `voting_end`

damirka:
## Description

While benchmarking dynamic fields access we discovered that `voting_end` does two iterations over pools, and this PR helps avoid it.

## Test plan

- all existing tests must pass
- compatibility is kept, functions are `public(package)`

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/staking_inner.move
+++ b/contracts/walrus/sources/staking/staking_inner.move
@@ -237,7 +237,7 @@ public(package) fun select_committee_and_calculate_votes(self: &mut StakingInner
     // prepare next epoch public_keys collection
     let committee = self.compute_next_committee();
     let mut public_keys = vector[];
-    let (node_ids, shard_assignment) = (*committee.inner()).into_keys_values();
+    let (node_ids, shard_assignments) = (*committee.inner()).into_keys_values();
 
     // prepare voting parameters
     let mut write_prices = priority_queue::new(vector[]);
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -231,34 +230,51 @@ public(package) fun voting_end(self: &mut StakingInnerV1, clock: &Clock) {
...
+public(package) fun select_committee_and_calculate_votes(self: &mut StakingInnerV1) {
+    assert!(self.next_committee.is_none(), ECommitteeSelected);
 
-    let committee = self.next_committee.borrow();
-    let size = committee.size();
-    let inner = committee.inner();
+    // prepare next epoch public_keys collection
+    let committee = self.compute_next_committee();
+    let mut public_keys = vector[];
+    let (keys, weights) = (*committee.inner()).into_keys_values();
```
karlwuest: Nit: `keys` here are the node ids, right? Also `weights` are not actually weights.
```suggestion
    let (node_ids, shard_assignment) = (*committee.inner()).into_keys_values();
```

damirka: Correct! My bad, these are shard assignments.

PR #1499, opened 2025-01-30
feat: allow multiple capability objects per storage node

### LINKED ISSUE #928: Revisit whether we want to restrict that one address can only register as walrus node once
This makes easier to find the capability object when interacting with the contract

halfprice:
## Description

This PR looses the constraint that each wallet can only register once and have one capability object, so that sending capability object to a node wallet won't bring the node down.

When a wallet contains multiple capability objects, it must specify the capability object ID in the node config file.

However, this change is still backward compatible in a sense that if the walrus-config file does not contain a capability object, contract call will try to find a capability object from the wallet. If there is only one, the call will use that object. If there are more than one, the call returns an error.

The `walrus-node register` now updates the config file with the newly created capability object in the file.

Close #928 
Close WAL-233

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/contract_service.rs
+++ b/crates/walrus-service/src/node/contract_service.rs
@@ -56,7 +56,7 @@ pub trait SystemContractService: std::fmt::Debug + Sync + Send {
     /// Submits a notification to the contract that this storage node epoch sync is done.
     ///
     /// If `node_capability` is provided, it will be used to set the node capability object ID.
-    async fn epoch_sync_done(&self, epoch: Epoch, node_capability: ObjectID);
+    async fn epoch_sync_done(&self, epoch: Epoch, node_capability_object_id: ObjectID);
 
     /// Ends voting for the parameters of the next epoch.
     async fn end_voting(&self) -> Result<(), anyhow::Error>;


crates/walrus-sui/src/client.rs
```diff
@@ -978,23 +996,29 @@ impl SuiContractClientInner {
...
+        let node_capability = if let Some(node_cap) = storage_node_cap {
+            node_cap
+        } else {
+            let node_capability = self
+                .read_client
+                .get_address_capability_object(self.wallet.active_address()?)
+                .await?
+                .ok_or(SuiClientError::StorageNodeCapabilityObjectNotSet)?;
+            node_capability.id
+        };
```
karlwuest: Suggestion: This check is done in multiple calls. Can you factor this out into its own function and then just call that?

crates/walrus-service/bin/node.rs
```diff
@@ -590,6 +600,11 @@ mod commands {
         println!("Successfully registered storage node:",);
         println!("      Capability object ID: {}", node_capability.id);
         println!("      Node ID: {}", node_capability.node_id);
+
+        // Update the config in `config_path` with the new storage node capability object ID.
+        config.storage_node_cap = Some(node_capability.id);
+        write_config_to_file(&config, &config_path, true)?;
```
mlegner, line 611: Remark: This is slightly dangerous: If the user defined any configuration parameters that coincide with defaults, those will not be included in the newly written config file. This is probably not too important in practice as the `register` command is most likely called immediately after the `setup` command. Still, might make sense to either try to avoid this behavior or at least document it.

crates/walrus-service/src/node.rs
```diff
@@ -483,15 +484,22 @@ async fn sync_node_params(config: &StorageNodeConfig) -> anyhow::Result<()> {
     let contract_client = node_wallet_config.new_contract_client().await?;
     let address = contract_client.address();
 
-    let node_cap = contract_client
-        .read_client
-        .get_address_capability_object(address)
-        .await?
-        .ok_or(SuiClientError::StorageNodeCapabilityObjectNotSet)?;
+    let node_capability = if let Some(node_cap) = config.storage_node_cap {
```
mlegner: Suggestion: Store this `node_capability` somewhere in the `StorageNodeInner` or the `SuiContractClient` so that it only has to be fetched a single time during startup.

crates/walrus-sui/src/client.rs
halfprice: Sounds good. Created `get_storage_node_capability_object_id`.

crates/walrus-service/bin/node.rs
halfprice: Not sure if I understand the `dangerous` part. If the user uses default, and it is omitted in the config rewrite, the node will still use the default value? Is there a particular scenario that this is unexpected?

I changed this to just append the `storage_node_cap` to the end of the config file.

crates/walrus-service/src/node.rs
halfprice: Here it is before the `StorageNodeInner` is created, but I preserved `node_capability` in the StorageNodeInner, so everywhere else this is needed is just read from `StorageNodeInner`.

crates/walrus-service/bin/node.rs
mlegner: > If the user uses default, and it is omitted in the config rewrite, the node will still use the default value? Is there a particular scenario that this is unexpected?

Indeed, the change doesn't have an immediate effect. However, if the defaults change in the future, the modified config would use the new defaults instead of the old hardcoded values. This behavior may not be expected.

mlegner: Important: Appending the parameter to the file will cause issues if the field was already set and the `--force` option is used. Maybe we can keep the previous version and just document the behavior with default values in the docstring?

crates/walrus-service/src/node.rs
```diff
@@ -451,6 +451,7 @@ pub struct StorageNodeInner {
     current_epoch: watch::Sender<Epoch>,
     is_shutting_down: AtomicBool,
     blocklist: Arc<Blocklist>,
+    node_capability: Option<ObjectID>,
```
mlegner: Question: Can we remove the option and just once get the capability object when the node is created?

crates/walrus-service/src/node.rs
mlegner: AFAICT, you're only storing the ID if it is set in the configuration. Could we instead store it in all cases?

crates/walrus-service/bin/node.rs
halfprice: Yes, make sense. I missed the `force` option during the revision. Good catch.

crates/walrus-service/src/node.rs
halfprice: Yeah, good suggestion. I now understand what do you mean from your original comments.

It's a good idea. I changed the code so that during node start up (when creating the StorageNode object), it tries to populate the `node_capability` either by filing from the config or by reading from the wallet. If multiple found in wallet, it returns failure.

Because of this change, I find it very challenging to test the specific node creation behavior (the idea is similar to the old `test_register_candidate` test). It requires a lot of plumbing in the test setup code, which requires decent amount of refactoring. This PR is still tested in simtest that we probabilistically add a second node cap object in the storage node to make sure that the node can work properly. Let me know if you feel strong about preserving `test_register_candidate` test, and I can try to refactor the test setup code to make it work (but probably in a follow up PR).

crates/walrus-service/src/node.rs
halfprice: Done, and please see the other comment.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -55,7 +56,7 @@ pub trait SystemContractService: std::fmt::Debug + Sync + Send {
     /// Submits a notification to the contract that this storage node epoch sync is done.
     ///
     /// If `node_capability` is provided, it will be used to set the node capability object ID.
-    async fn epoch_sync_done(&self, epoch: Epoch, node_capability: Option<ObjectID>);
+    async fn epoch_sync_done(&self, epoch: Epoch, node_capability: ObjectID);
```
mlegner: Nit, to be consistent:
```suggestion
    async fn epoch_sync_done(&self, epoch: Epoch, node_capability_object_id: ObjectID);
```

crates/walrus-service/bin/node.rs
mlegner: @halfprice Not required, but @liquid-helium has implemented a nice approach for a similar situation [here](https://github.com/MystenLabs/walrus/pull/1461/files#diff-08e649f77d08d50a6320719ca97aa3ceb6c8244b6c3ad7c110dba81d98c3e651R207).

PR #1501, opened 2025-01-30
fix(system): change FutureAccounting to store `used_capacity` for future epochs

damirka:
## Description

- Reuse the field `storage_to_reclaim` in FutureAccounting as `used_capacity`.
- [Here](https://github.com/MystenLabs/walrus/blob/5fbe3330f48fb8aeb94a1de148365a102bc7349c/contracts/walrus/sources/system/system_state_inner.move#L202-L205), iterate through all values in the ring buffer up to `epochs_ahead - 1` and increase the `used_capacity` by `storage_amount` . This requires looping through all epochs, but for mainnet, this would be at most 52 iterations, which should not increase the gas cost significantly.
- Similarly [here](https://github.com/MystenLabs/walrus/blob/5fbe3330f48fb8aeb94a1de148365a102bc7349c/contracts/walrus/sources/system/system_state_inner.move#L346-L355), increase the `used_capacity` for all values between `start_offset` and `end_offset - 1`. Here we should also add a check that the new value for any of these is below `total_capacity` and abort otherwise.
- Update `used_capacity` [here](https://github.com/MystenLabs/walrus/blob/5fbe3330f48fb8aeb94a1de148365a102bc7349c/contracts/walrus/sources/system/system_state_inner.move#L124)
- Preserve the check for `total_capacity` in `reserve_space` (and do it for future epochs) but do not perform it for event blobs.

## Other changes

- renames `Storage.storage_size` to `Storage.size`
- adds `#[test_only]` on unused functions in `System`
- features tests for used capacity

## Test plan

- added a test for the original issue
- all previous tests pass without any changes to the setup

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/tests/system/storage_resource_tests.move
+++ b/contracts/walrus/tests/system/storage_resource_tests.move
@@ -24,11 +24,7 @@ public fun test_split_epoch() {
         new_storage.end_epoch() == 10,
         0,
     );
-    assert!(
-        storage.size() == storage_amount &&
-        new_storage.size() == storage_amount,
-        0,
-    );
+    assert!(storage.size() == storage_amount && new_storage.size() == storage_amount, 0);
     storage.destroy();
     new_storage.destroy();
 }


contracts/walrus/sources/system/storage_accounting.move
```diff
@@ -32,16 +34,30 @@ public(package) fun epoch(accounting: &FutureAccounting): u32 {
     *&accounting.epoch
 }
 
+/// Accessor for used_capacity, read-only.
+public fun used_capacity(accounting: &FutureAccounting): u64 {
+    accounting.storage_to_reclaim
+}
+
+#[deprecated(note = b"Use `used_capacity` instead.")]
```
karlwuest: Suggestion: I think we can directly delete this, since it's not a public interface.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -48,6 +48,9 @@ public struct SystemStateInnerV1 has store {
     committee: BlsCommittee,
     // Some accounting
     total_capacity_size: u64,
+    /// [DEPRECATED]
+    /// This field is deprecated in favor of the accounting ring buffer which
+    /// stores the used capacity for each future epoch.
```
karlwuest: Nit: We're still using it below, so it's not really deprecated.

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -340,19 +344,16 @@ public(package) fun extend_blob(
...
-    // Account the space to reclaim in the future.
-
-    // First account for the space not being freed in the original end epoch.
-    self
-        .future_accounting
-        .ring_lookup_mut(start_offset - 1)
-        .decrease_storage_to_reclaim(storage_size);
+    // Account the used space: increase the used capacity for each epoch in the future
+    // Iterates: [start, end)
+    (start_offset - 1).range_do!(end_offset, |i| {
```
karlwuest: I think this should just be:
```suggestion
    start_offset.range_do!(end_offset, |i| {
```

contracts/walrus/sources/system/storage_accounting.move
```diff
@@ -14,6 +14,8 @@ const ETooFarInFuture: u64 = 0;
 /// storage needs to be reclaimed and the rewards to be distributed.
 public struct FutureAccounting has store {
     epoch: u32,
+    /// This field stores `used_capacity` for the epoch.
+    /// Currently, impossible to rename due to package upgrade limitations.
```
karlwuest, line 18: Let's rename this now. Otherwise we'll forget before publishing to mainnet and then we're locked in.

contracts/walrus/sources/system/storage_accounting.move
```diff
@@ -32,19 +34,15 @@ public(package) fun epoch(accounting: &FutureAccounting): u32 {
     *&accounting.epoch
 }
 
-/// Accessor for storage_to_reclaim, mutable.
-public(package) fun storage_to_reclaim(accounting: &FutureAccounting): u64 {
+/// Accessor for used_capacity, read-only.
+public fun used_capacity(accounting: &FutureAccounting): u64 {
```
karlwuest: Question: does this need to be public?

contracts/walrus/sources/system/system_state_inner.move
```diff
@@ -443,6 +449,7 @@ public(package) fun certify_event_blob(
     let storage = self.reserve_space_without_payment(
         encoded_blob_length(size, encoding_type, num_shards),
         epochs_ahead,
+        false, // do not check total capacity
```
karlwuest: Suggestion: Add more context in the comment
```suggestion
        // Do not check total capacity, event blobs are certified already at this point.
        false, 
```

contracts/walrus/sources/system/storage_accounting.move
damirka: Nice catch! My bad!

contracts/walrus/sources/system/storage_accounting.move
damirka: Renamed!

contracts/walrus/sources/system/system_state_inner.move
damirka: Done!

PR #1503, opened 2025-01-30
fix: Change how we check committee.

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/types.rs
+++ b/crates/walrus-sui/src/types.rs
@@ -396,7 +396,7 @@ impl Committee {
         // Compare epoch
         if self.epoch != other.epoch {
             error_msgs.push(format!(
-                "Epoch mismatch: left {}, right {}",
+                "Epoch mismatch: existing {}, new {}",
                 self.epoch, other.epoch
             ));
         }
@@ -404,7 +404,7 @@ impl Committee {
         // Compare n_shards
         if self.n_shards != other.n_shards {
             error_msgs.push(format!(
-                "Number of shards mismatch: left {}, right {}",
+                "Number of shards mismatch: existing {}, new {}",
                 self.n_shards, other.n_shards
             ));
         }
@@ -425,7 +425,7 @@ impl Committee {
         // Check each node in left committee exists in right with matching shards
         for (node_id, left_shards) in self_nodes.iter() {
             if let Some(right_shards) = other_nodes.get(node_id) {
-                if *left_shards != *right_shards {
+                if left_shards != right_shards {
                     error_msgs.push(format!(
                         "Shard assignment mismatch for node {:?}: left {:?}, right {:?}",
                         node_id, left_shards, right_shards


crates/walrus-sui/src/types.rs
```diff
@@ -387,4 +388,73 @@ impl Committee {
     pub fn members_mut(&mut self) -> &mut Vec<StorageNode> {
         &mut self.members
     }
+
+    /// Compares two committees based only on epoch, n_shards, and storage node ids/shards
+    pub fn compare_essential(&self, other: &Committee) -> Result<(), String> {
```
halfprice: Suggestion: return `anyhow::Result<()>` so that the returned error is a proper error type.

crates/walrus-sui/src/types.rs
```diff
@@ -387,4 +388,73 @@ impl Committee {
...
+        let mut error_msgs = Vec::new();
+
+        // Compare epoch
+        if self.epoch != other.epoch {
+            error_msgs.push(format!(
```
halfprice, line 398: anyhow::bail!("epoch mismatch: left {}, right {}", self.epoch, other.epoch);

crates/walrus-sui/src/types.rs
```diff
@@ -387,4 +388,73 @@ impl Committee {
...
+                        node_id
+                    ));
+                }
+                Some(right_shards) if *left_shards != *right_shards => {
+                    error_msgs.push(format!(
+                        "Shard assignment mismatch for node {:?}: left {:?}, right {:?}",
+                        node_id, left_shards, right_shards
+                    ));
+                }
+                _ => {}
```
halfprice: A common pattern is

             if let Some(other_shards) = other_nodes.get(node_id) {
                if shards != other_shards {
                    error_msgs.push(format!("Shard assignment mismatch for node {:?}: left {:?}, right {:?}", 
                        node_id, shards, other_shards));
                }
            } else {
                error_msgs.push(format!("Node {:?} exists in left but not in right", node_id));
            }

liquid-helium: Thanks for the tip, change applied.

crates/walrus-sui/src/types.rs
liquid-helium: Thanks, change applied.

crates/walrus-sui/src/types.rs
liquid-helium: I try to collect as much information as possible, to let it continue and check which nodes are missing.

crates/walrus-sui/src/types.rs
```diff
@@ -387,4 +388,71 @@ impl Committee {
...
+            .iter()
+            .map(|node| (&node.node_id, &node.shard_ids))
+            .collect();
+
+        // Check each node in left committee exists in right with matching shards
+        for (node_id, left_shards) in self_nodes.iter() {
+            if let Some(right_shards) = other_nodes.get(node_id) {
+                if *left_shards != *right_shards {
+                    error_msgs.push(format!(
+                        "Shard assignment mismatch for node {:?}: left {:?}, right {:?}",
```
halfprice, line 430: Instead of `left`, `right`, how about `existing` and `new`? This way, we don't need to read the code and figure out what is left and what is right.

crates/walrus-sui/src/types.rs
```diff
@@ -387,4 +388,71 @@ impl Committee {
...
+        let other_nodes: HashMap<_, _> = other
+            .members
+            .iter()
+            .map(|node| (&node.node_id, &node.shard_ids))
+            .collect();
+
+        // Check each node in left committee exists in right with matching shards
+        for (node_id, left_shards) in self_nodes.iter() {
+            if let Some(right_shards) = other_nodes.get(node_id) {
+                if *left_shards != *right_shards {
```
halfprice: Nit: I think you don't need the * in both sides of the comparison.

liquid-helium: Done.

crates/walrus-sui/src/types.rs
liquid-helium: done.

PR #1509, opened 2025-01-31
feat: support PKCS#8 PEM network keys

jpcsmith:
## Description

This PR adds support PEM network keys on the walrus storage node. It 

- allows parsing the `network_key_pair` in the config from a path to a PEM file;
- removes the `tls.pem_files` config argument, using instead the `network_key_pair` and `tls.certficiate_path` config arguments;
- adds CLI options to the `key-gen` subcommand for generating network keys in PEM format, as well as converting an existing key to PEM.

## Test plan

Several new tests were added for the aforementioned features.

---

## Release notes

- [x] Storage node: The config block `tls.pem_files` has been removed. Provide a TLS certificate path by the `tls.certificate_path` config argument, and the PKCS#8 PEM-encoded TLS private using `network_key_pair.path`.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -14,6 +14,7 @@ use std::{
 
 use anyhow::{bail, Context};
 use clap::{Parser, Subcommand, ValueEnum as _};
+use commands::generate_or_convert_key;
 use config::PathOrInPlace;
 use fs::File;
 use humantime::Duration;
@@ -344,26 +345,14 @@ fn main() -> anyhow::Result<()> {
             force,
             format,
             convert,
-        } => {
-            if let Some(input_key_path) = convert {
-                commands::convert(
-                    &input_key_path,
-                    out.as_deref()
-                        .unwrap_or_else(|| Path::new(key_type.default_filename())),
-                    key_type,
-                    force,
-                    format,
-                )?
-            } else {
-                commands::keygen(
-                    out.as_deref()
-                        .unwrap_or_else(|| Path::new(key_type.default_filename())),
-                    key_type,
-                    force,
-                    format,
-                )?
-            }
-        }
+        } => generate_or_convert_key(
+            out.as_deref()
+                .unwrap_or_else(|| Path::new(key_type.default_filename())),
+            key_type,
+            force,
+            format,
+            convert.as_deref(),
+        )?,
 
         Commands::GenerateConfig {
             path_args,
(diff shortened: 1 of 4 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -541,21 +571,79 @@ mod commands {
...
+    pub(super) fn keygen(
+        path: &Path,
+        key_type: KeyType,
+        force: bool,
+        format: KeyFormat,
+    ) -> anyhow::Result<()> {
+        walrus_core::ensure!(
+            format != KeyFormat::Pkcs8 || key_type == KeyType::Network,
+            "`--format=pkcs8` is only supported with `--key-type=network`"
+        );
```
mlegner: Nit: Can we move this up to the callsite of this function as this is shared with the `convert` function?

crates/walrus-service/bin/node.rs
```diff
@@ -541,21 +571,79 @@ mod commands {
...
+        force: bool,
+        output_format: KeyFormat,
+    ) -> anyhow::Result<()> {
+        walrus_core::ensure!(
+            output_format != KeyFormat::Pkcs8 || key_type == KeyType::Network,
+            "`--format=pkcs8` is only supported with `--key-type=network`"
+        );
+
+        println!(
+            "Converting {key_type} key pair form '{}' and writing it to '{}'",
```
mlegner: Typo:
```suggestion
            "Converting {key_type} key pair from '{}' and writing it to '{}'",
```

crates/walrus-service/src/node/config.rs
```diff
@@ -311,8 +313,14 @@ pub struct TlsConfig {
...
-    pub key_path: PathBuf,
+    /// Path to the PEM-encoded PKCS#8 certificate private key.
+    ///
+    /// Prefer using the `network_key_pair` option instead, as it also accepts a path to a
+    /// PKCS#8 PEM-encoded private key and is required.
+    ///
+    /// If this option is set, then `network_key_pair` must be set to the same path.
+    #[serde(default)]
+    pub key_path: Option<PathBuf>,
 }
```
mlegner: Suggestion: Just remove the `key_path`. I don't think we need to worry about having a breaking change here, nobody is actually using this yet AFAIK. The current solution is still a breaking change as we now have the additional requirement of having the same path as the network key pair.

This would also allow us to remove the extra hierarchy level.

jpcsmith: Removed, and I now directly use the path for the `network_key_pair` if available, so that `axum_server` can read the full PEM file.

crates/walrus-service/bin/node.rs
jpcsmith: I now merged the two functions to remove duplication.

PR #1512, opened 2025-02-01
feat: metadata for Blobs

liquid-helium:
## Description

Exposed methods in Client to operate on Blob Metadata.

Related to WAL-331
Resolves WAL-598

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -1044,6 +1051,353 @@ async fn test_post_store_action(
...
+        ("author".to_string(), "walrus_test".to_string()),
+        ("timestamp".to_string(), "2024-01-01".to_string()),
+        ("addr".to_string(), "Mars".to_string()),
+    ]);
+    let blobs_to_create = 2;
+    let (_sui_cluster_handle, _cluster, mut client) = test_cluster::default_setup().await?;
+    let blob = walrus_test_utils::random_data(314);
+    let blobs = vec![blob.as_slice()];
+
+    // Store multiple copies of the same blob with different end times
```
giac-mysten: Nit: here and elsewhere, comments and docstring should end with `.` :)
```suggestion
    // Store multiple copies of the same blob with different end times.
```

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -1044,6 +1051,353 @@ async fn test_post_store_action(
     Ok(())
 }
 
+#[ignore = "ignore E2E tests by default"]
+#[walrus_simtest]
+/// Tests blob metadata operations.
```
giac-mysten: Remark: this test is nice and comprehensive, though a bit massive 😅 Could not come up with ways to split it up, anyways.

crates/walrus-sui/src/client.rs
```diff
@@ -737,6 +744,76 @@ impl SuiContractClient {
...
     }
 
+    /// Adds metadata to a blob object.
+    ///
+    /// If metadata does not exist, it is created with the given key-value pairs.
+    /// If metadata already exists, an error is returned unless `force` is true.
+    /// If `force` is true, the metadata is updated with the given key-value pairs.
+    pub async fn add_blob_metadata(
+        &mut self,
+        blob_id: ObjectID,
```
giac-mysten: Remark: `blob_id` is usually the `BlobId` of the blob. maybe `blob_object`? here and following.

crates/walrus-sui/src/client.rs
```diff
@@ -737,6 +744,76 @@ impl SuiContractClient {
...
+            ))) if force => {
+                inner
+                    .insert_or_update_blob_metadata_pairs(blob_id, &metadata)
+                    .await
+            }
+            Err(e) => Err(e),
+        }
+    }
+
+    /// Removes the metadata from a blob object.
```
giac-mysten:
```suggestion
    /// Removes the metadata dynamic field from a blob object.
```

crates/walrus-sui/src/client.rs
```diff
@@ -737,6 +744,76 @@ impl SuiContractClient {
...
+        let mut inner = self.inner.lock().await;
+        match inner.add_blob_metadata(blob_id, &metadata).await {
+            Ok(()) => Ok(()),
+            Err(SuiClientError::TransactionExecutionError(MoveExecutionError::Blob(
+                BlobError::EDuplicateMetadata(_),
+            ))) if force => {
+                inner
+                    .insert_or_update_blob_metadata_pairs(blob_id, &metadata)
+                    .await
+            }
```
giac-mysten: Question: I like the structure of the command with `force`. However, if `force == true` and the metadata is duplicate, we have to wait for an extra command execution. Why not call directly `insert_or_update..` if `force==true`, and save the time and gas of the first failure?

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -827,6 +842,25 @@ impl ReadClient for SuiReadClient {
             .collect())
     }
 
+    async fn get_blob_metadata(&self, blob_id: ObjectID) -> SuiClientResult<Option<Metadata>> {
+        self.sui_client
+            .get_dynamic_field::<Vec<u8>, Metadata>(
```
giac-mysten: So nice that we have our own api to `get_dynamic_field`, that does the efficient thing of deriving the df id internally :)

crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -106,6 +107,77 @@ impl AssociatedContractStruct for Blob {
     const CONTRACT_STRUCT: StructTag<'static> = contracts::blob::Blob;
 }
 
+/// The metadata struct for Blob objects.
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct Metadata {
```
giac-mysten: Suggestion: This is easy to confuse with the `BlobMetadata` in core/redstuff encoding. Can we have a more descriptive name? `OnChainMetadata`? `OwnedMetadata`? don't have many good ideas :(

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -1044,6 +1051,353 @@ async fn test_post_store_action(
...
+
+    // Test adding initial metadata
+    let mut metadata = Metadata::default();
+    let mut count = 0;
+    for (key, value) in key_value_pairs.iter() {
+        if random::<bool>() {
+            metadata.insert(key.clone(), value.clone());
+            count += 1;
+        }
+    }
```
giac-mysten: Suggestion: implement `From<HashMap> for Metadata` ?

giac-mysten: and `len(&self)` for metadata?

crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -106,6 +107,77 @@ impl AssociatedContractStruct for Blob {
...
+            self.metadata.contents.push(Entry { key, value });
+        }
+    }
+
+    /// Returns an iterator over the key-value pairs in the metadata.
+    pub fn iter(&self) -> MetadataIter {
+        MetadataIter {
+            inner: self.metadata.contents.iter(),
+        }
+    }
```
giac-mysten, line 169: Question: `iter` here clones the inner values. This is a bit different from the usual meaning of `iter`, that does not clone (while `into_iter` consumes). Correct (i'm not too sure of this myself 😬 )? Maye rename?

crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -106,6 +107,77 @@ impl AssociatedContractStruct for Blob {
...
+}
+
+impl AssociatedContractStruct for Metadata {
+    const CONTRACT_STRUCT: StructTag<'static> = contracts::metadata::Metadata;
+}
+
+/// A blob with its metadata.
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
+pub struct BlobWithMetadata {
```
giac-mysten: 😬 ?
```suggestion
pub struct BlobWith(Object|Sui|Chain)Metadata {
```

crates/walrus-sui/src/client.rs
mlegner: We use `blob_obj_id` in a few places.

crates/walrus-sui/src/types/move_structs.rs
liquid-helium: Renamed to BlobWithObjectMetadata, thanks for the suggestion!

crates/walrus-sui/src/types/move_structs.rs
liquid-helium: Good point.
The return type is now <&String, &String>

crates/walrus-e2e-tests/tests/test_client.rs
liquid-helium: good point. change applied.

crates/walrus-e2e-tests/tests/test_client.rs
liquid-helium: Thanks for the tip, fixed.

crates/walrus-sui/src/types/move_structs.rs
liquid-helium: I renamed it as BlobAttribute, open to any suggestions!

crates/walrus-sui/src/client.rs
liquid-helium: Change applied.

crates/walrus-sui/src/client.rs
liquid-helium: Good point!

The only caveat is when the metadata dynamic field has not been created, the `insert_or_update` will also fail with the same error :(, without changing the contract implementation.

crates/walrus-sui/src/client.rs
liquid-helium: Renamed to `blob_obj_id`!

crates/walrus-e2e-tests/tests/test_client.rs
liquid-helium: Sorry for the mess.
I made some changes, move most of the bulky and messy part to a helper struct, hopefully it is easier to read now...

PR #1514, opened 2025-02-02
test(used capacity): adds tests for used_capacity mechanics

damirka:
## Description

- adds system tests for used capacity at different epochs
- cleans up storage tests
- marks certain package functions as `#[test_only]` to indicate they're not used in production
- adds `.size()` alias to `storage`, since we cannot change function names in upgrades

## Test plan

I am test.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/system/storage_resource.move
+++ b/contracts/walrus/sources/system/storage_resource.move
@@ -28,10 +28,7 @@ public fun end_epoch(self: &Storage): u32 {
     self.end_epoch
 }
 
-// alias for easier access
-public use fun storage_size as Storage.size;
-
-public fun storage_size(self: &Storage): u64 {
+public fun size(self: &Storage): u64 {
     self.storage_size
 }
 
--- a/contracts/walrus/sources/system/blob.move
+++ b/contracts/walrus/sources/system/blob.move
@@ -150,7 +150,7 @@ public(package) fun new(
         encoding_type,
         n_shards,
     );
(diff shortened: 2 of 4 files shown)


contracts/walrus/sources/system/storage_resource.move
```diff
@@ -28,6 +28,9 @@ public fun end_epoch(self: &Storage): u32 {
     self.end_epoch
 }
 
+// alias for easier access
+public use fun storage_size as Storage.size;
+
 public fun storage_size(self: &Storage): u64 {
     self.storage_size
```
damirka, line 32: If there's a room for last minute breaking changes, I'd change this function to just `size`, but it's super minor.

contracts/walrus/sources/system/storage_resource.move
damirka, line 1: Just to double check: `storage.destroy()` is public - is that intentional?

contracts/walrus/sources/system/storage_resource.move
karlwuest: Sure, I think that's fine to change.

contracts/walrus/sources/system/storage_resource.move
karlwuest: Yes, I think users should generally have the option to destroy/burn owned objects, otherwise people will end up just sending it to the zero address (that's what I do with unwanted objects 😅).

PR #1530, opened 2025-02-04
feat: CI correctness check

liquid-helium:
## Description

A boilerplate with limited functions.

A binary to check the storage nodes health, to be used in CI test.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-node-prober/src/health_checker.rs
+++ b/crates/walrus-node-prober/src/health_checker.rs
@@ -29,6 +29,9 @@ impl HealthLogEntry {
     }
 }
 
+/// HealthChecker is a struct that checks the health of the nodes in the network.
+/// It periodically checks the health of the nodes and logs the results to a file.
+/// TODO(WAL-524): Add a validator to generate a health report for the nodes.
 pub struct HealthChecker {
     log_dir: PathBuf,
     node_probers: Arc<Mutex<HashMap<ObjectID, Arc<Mutex<NodeProber>>>>>,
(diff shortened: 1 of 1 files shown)


docker/walrus-node-prober/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+FROM rust:1.84-bookworm AS builder
+ARG PROFILE=release
+ARG GIT_REVISION
+ENV GIT_REVISION=$GIT_REVISION
+WORKDIR "$WORKDIR/walrus"
+RUN apt-get update && apt-get install -y cmake clang
+
+COPY Cargo.toml Cargo.lock ./
+COPY crates crates
+COPY contracts contracts
```
pei-mysten, line 14: do we need the move contracts in the image? from the code I don't think it's quite needed but just to confirm

crates/walrus-node-prober/src/health_checker.rs
```diff
@@ -0,0 +1,164 @@
...
+                let probers = node_probers.lock().await;
+                let prober_refs: Vec<Arc<Mutex<NodeProber>>> = probers.values().cloned().collect();
+                drop(probers);
+
+                for prober in prober_refs {
+                    let mut prober = prober.lock().await;
+                    let log_entry = prober.check_health().await;
+                    // TODO(WAL-524): consider batching log entries
+                    if let Ok(bytes) = log_entry.as_bytes() {
+                        let _ = log_file.write_all(&bytes).await;
```
pei-mysten, line 99: given that this node-prober will likely run in our CI pipeline, it might make more sense to output the logs to stdout, so that journalctl can pick it up and then our logging agent will be able to read from syslog.

PR #1531, opened 2025-02-04
fix: do not block the runtime when retrieving multiple symbols

jpcsmith:
## Description

Retrieving symbols in a loop is currently a long-running synchronous operation. This can 'block' a tokio thread for several hundreds of milliseconds. Spawn a background threads to handle the request instead.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/encoding/slivers.rs
+++ b/crates/walrus-core/src/encoding/slivers.rs
@@ -313,12 +313,15 @@ impl<T: EncodingAxis> SliverData<T> {
     /// [`InconsistencyProof`] is generated and returned. If recovery fails or another error occurs,
     /// a [`SliverRecoveryError`] is returned.
     ///
-    /// Symbols that fail verification are logged and dropped.
+    /// If `verify_symbols` is set to true, symbols that fail verification are logged and
+    /// dropped. Otherwise, it is the caller's responsibility to ensure that the recovery symbols
+    /// have been verified to be useable to recover the identified symbol.
     pub fn recover_sliver_or_generate_inconsistency_proof<I, U>(
         recovery_symbols: I,
         target_index: SliverIndex,
         metadata: &BlobMetadata,
         encoding_config: &EncodingConfig,
+        verify_symbols: bool,
     ) -> Result<SliverOrInconsistencyProof<T, U>, SliverRecoveryOrVerificationError>
     where
         I: IntoIterator,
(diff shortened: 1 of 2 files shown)


crates/walrus-sdk/src/client.rs
```diff
@@ -796,56 +796,58 @@ impl Client {
...
         );
-        let mut final_error = NodeError::other(EmptyResponse);
 
-        symbols.retain(|symbol| {
-            let _guard = tracing::info_span!(
-                "list_and_verify_recovery_symbols__retain",
-                walrus.symbol.id = %symbol.id()
-            )
-            .entered();
+        tokio::task::spawn_blocking(move || {
```
halfprice, line 813: Just to confirm: all the changes in this function just to wrap the `symbols.retain` into a background thread, correct?

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -313,26 +313,33 @@ impl<T: EncodingAxis> SliverData<T> {
...
-    /// Symbols that fail verification are logged and dropped.
+    /// If `verify_symbols` is set to true, symbols that fail verification are logged and
+    /// dropped. Otherwise, it is the caller's responsibility to ensure that the recovery symbols
+    /// have been verified to be useable to recover the identified symbol.
     pub fn recover_sliver_or_generate_inconsistency_proof<I, U>(
         recovery_symbols: I,
         target_index: SliverIndex,
         metadata: &BlobMetadata,
         encoding_config: &EncodingConfig,
+        verify_symbols: bool,
```
halfprice, line 324: Question: trying to understand why we can afford to not verify symbols. Say we use the returned symbols to successfully reconstruct the sliver without verifying individual symbols, and the recovered sliver can be verified using the metadata, that is enough to show the authenticity of the sliver, is that correct?

halfprice: Follow up question: if `verify_symbols` is false, we may not need to fetch proof of returned symbol, which also save some work on the source side. Is this something you will be considering?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -613,33 +616,44 @@ where
...
+        let index = self.target_index;
+        let metadata = self.metadata.clone();
+        let encoding_config = self.shared.encoding_config.clone();
+        let result = tokio::task::spawn_blocking(move || {
+            SliverData::<A>::recover_sliver_or_generate_inconsistency_proof(
+                recovery_symbols,
+                index,
+                metadata.metadata(),
+                &encoding_config,
+                false,
```
halfprice, line 652: Similar to what I asked above, should we check that the recovered sliver matches the merkle hash in the metadata?

crates/walrus-sdk/src/client.rs
jpcsmith: Correct, it pushes the work onto one of tokio's blocking background threads, and the `await` yields to the tokio runtime until the result is complete.

crates/walrus-core/src/encoding/slivers.rs
jpcsmith: We verified that they were valid for the recovery of the sliver in `list_and_verify_recovery_symbols` in `client.rs`. I had placed the verification there, so that both network failures and invalid symbols being returned result in just fetching new symbols from a different storage node.

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: The check against the metadata is performed inside the function call, irrespective of the value `true/false` here.

PR #1532, opened 2025-02-04
feat(client): support for multiple environments (Testnet/Mainnet) in client config

wbbradley:
## Description

Add support for multiple contexts within Walrus client config.

## Test plan

Added new unit tests around the new scenarios. Manual testing.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: support has been introduced for multiple contexts, and a default_context for walrus client invocations. Context can be overriden with the new --context <context> argument.

Here is an example (with some fake addresses) multi-context config file.

```
---
contexts:
  testnet:
    system_object: 0x98ebc47370603fe81d9e15491b2f1443d619d1dab720d586e429ed233e1255c1
    staking_object: 0x20266a17b4f1a216727f3eef5772f8d486a9e3b5e319af80a5b75809c035561d
    exchange_objects:
      - 0x59ab926eb0d94d0d6d6139f11094ea7861914ad2ecffc7411529c60019133997
      - 0x89127f53890840ab6c52fca96b4a5cf853d7de52318d236807ad733f976eef7b
      - 0x9f9b4f113862e8b1a3591d7955fadd7c52ecc07cf24be9e3492ce56eb8087805
      - 0xb60118f86ecb38ec79e74586f1bb184939640911ee1d63a84138d080632ee28a
    wallet_config:
      path: ~/.sui/sui_config/client.yaml
      active_env: testnet
  devnet:
    system_object: 0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
    staking_object: 0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb
    exchange_objects:
      - 0xcccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccccc
      - 0xdddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddddd
      - 0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee
      - 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
    wallet_config:
      path: ~/.sui/sui_config/client.yaml
      active_env: devnet

default_context: testnet

```

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -52,8 +52,15 @@ pub struct App {
     /// 4. In `~/.walrus/`.
     // NB: Keep this in sync with `crate::cli`.
     #[clap(long, verbatim_doc_comment, global = true)]
-    #[serde(default, deserialize_with = "crate::utils::resolve_home_dir_option")]
+    #[serde(
+        default,
+        deserialize_with = "walrus_utils::config::resolve_home_dir_option"
+    )]
     pub config: Option<PathBuf>,
+    /// The configuration context to use for the client, if omitted the default_context is used.
+    #[clap(long, global = true)]
+    #[serde(default)]
+    pub context: Option<String>,
     /// The path to the Sui wallet configuration file.
     ///
     /// The wallet configuration is taken from the following locations:
@@ -67,7 +74,10 @@ pub struct App {
     /// is returned.
     // NB: Keep this in sync with `crate::cli`.
     #[clap(long, verbatim_doc_comment, global = true)]
-    #[serde(default, deserialize_with = "crate::utils::resolve_home_dir_option")]
+    #[serde(
+        default,
+        deserialize_with = "walrus_utils::config::resolve_home_dir_option"
+    )]
     pub wallet: Option<PathBuf>,
     /// The gas budget for transactions.
     ///
@@ -186,7 +196,7 @@ pub enum CliCommands {
     Store {
         /// The files containing the blob to be published to Walrus.
         #[clap(required = true, value_name = "FILES")]
-        #[serde(deserialize_with = "crate::utils::resolve_home_dir_vec")]
+        #[serde(deserialize_with = "walrus_utils::config::resolve_home_dir_vec")]
         files: Vec<PathBuf>,
         /// The epoch argument to specify either the number of epochs to store the blob, or the
         /// end epoch, or the earliest expiry time in rfc3339 format.
@@ -239,7 +249,10 @@ pub enum CliCommands {
         ///
         /// If unset, prints the blob to stdout.
         #[clap(long)]
-        #[serde(default, deserialize_with = "crate::utils::resolve_home_dir_option")]
+        #[serde(
+            default,
+            deserialize_with = "walrus_utils::config::resolve_home_dir_option"
+        )]
         out: Option<PathBuf>,
         /// The URL of the Sui RPC node to use.
         #[clap(flatten)]
@@ -318,7 +331,7 @@ pub enum CliCommands {
     /// Encode the specified file to obtain its blob ID.
     BlobId {
         /// The file containing the blob for which to compute the blob ID.
-        #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
+        #[serde(deserialize_with = "walrus_utils::config::resolve_home_dir")]
         file: PathBuf,
         /// The number of shards for which to compute the blob ID.
         ///
@@ -737,7 +750,7 @@ pub struct PublisherArgs {
     pub refill_interval: Duration,
     /// The directory where the publisher will store the sub-wallets used for client multiplexing.
     #[clap(long)]
-    #[serde(deserialize_with = "crate::utils::resolve_home_dir")]
+    #[serde(deserialize_with = "walrus_utils::config::resolve_home_dir")]
     pub sub_wallets_dir: PathBuf,
     /// The amount of MIST transferred at every refill.
     #[clap(long, default_value_t = default::gas_refill_amount())]
@@ -889,7 +902,10 @@ pub struct DaemonArgs {
     pub(crate) metrics_address: SocketAddr,
     /// Path to a blocklist file containing a list (in YAML syntax) of blocked blob IDs.
     #[clap(long)]
-    #[serde(default, deserialize_with = "crate::utils::resolve_home_dir_option")]
+    #[serde(
+        default,
+        deserialize_with = "walrus_utils::config::resolve_home_dir_option"
+    )]
     pub(crate) blocklist: Option<PathBuf>,
 }
 
@@ -1457,6 +1473,7 @@ mod tests {
     fn test_json_string_extraction(json: &str, command: Commands) -> TestResult {
         let mut app = App {
             config: None,
+            context: None,
             wallet: None,
             gas_budget: None,
             json: false,
--- a/crates/walrus-service/src/client/config.rs
+++ b/crates/walrus-service/src/client/config.rs
@@ -2,14 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
+    collections::HashMap,
     env,
     fmt,
     num::{NonZeroU16, NonZeroUsize},
-    path::PathBuf,
+    path::{Path, PathBuf},
     time::Duration,
 };
 
-use anyhow::bail;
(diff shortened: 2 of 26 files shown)


crates/walrus-sui/src/config.rs
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Loads the wallet context from the given optional wallet config (optional path and optional
+    /// Sui env).
+    ///
+    /// If no path is provided, tries to load the configuration first from the local folder, and
+    /// then from the standard Sui configuration directory.
+    // NB: When making changes to the logic, make sure to update the argument docs in
+    // `crates/walrus-service/bin/client.rs`.
+    pub fn load_wallet_context(wallet_config: Option<&WalletConfig>) -> Result<WalletContext> {
+        let mut default_paths = vec!["./sui_config.yaml".into()];
```
halfprice, line 89: Nit: can we define sui_config.yaml into a const string? And similar for the client.yaml below?

crates/walrus-service/src/client/config.rs
```diff
@@ -34,6 +38,55 @@ use crate::{
...
+#[allow(clippy::large_enum_variant)]
+pub enum MultiConfig {
+    SingletonConfig(Config),
+    MultiConfig {
+        contexts: HashMap<String, Config>,
+        default_context: String,
+    },
+}
+
+pub fn load_config_from_multi_config(
```
halfprice: Suggestion: given this is a public function, can you add some documentation to it, especially specify what are the return values?

crates/walrus-service/src/client/config.rs
```diff
@@ -34,6 +38,55 @@ use crate::{
...
+    path: impl AsRef<Path>,
+    context: Option<&str>,
+) -> anyhow::Result<(Config, Option<String>)> {
+    let path = path.as_ref();
+    match utils::load_from_yaml(path)? {
+        MultiConfig::SingletonConfig(config) => {
+            if let Some(context) = context {
+                bail!(
+                    "context '{}' not found in single-config file '{}'",
```
halfprice: Nit: maybe `cannot specify context in single-config file` ?

Also, I feel we can just ignore context in this case despite whether use sets it or not.

crates/walrus-service/src/client/config.rs
```diff
@@ -34,6 +38,55 @@ use crate::{
...
+            }
+            Ok((config, None))
+        }
+        MultiConfig::MultiConfig {
+            contexts,
+            default_context,
+        } => {
+            let target_context = context.unwrap_or(&default_context);
+            Ok((
+                contexts.get(target_context).cloned().ok_or_else(|| {
```
halfprice: Should we check if `contexts` is not empty?

crates/walrus-sui/src/config.rs
```diff
@@ -0,0 +1,117 @@
...
+/// override for the active environment.
+pub enum WalletConfig {
+    /// The path to the wallet configuration file. The wallet context will use the active
+    /// environment from the given configuration file.
+    Path(#[serde(default, deserialize_with = "resolve_home_dir")] PathBuf),
+    /// The path to the wallet configuration file with an optional overridden `active_env`. The
+    /// wallet context will use the specified `active_env` if it is non-null. Loading of the
+    /// `WalletContext` will fail if an invalid `active_env` is used. Specifying a null `active_env`
+    /// will use the `active_env` from the configuration file.
+    PathWithOverride {
```
halfprice, line 29: Suggestion: can you add an example, maybe in the PR description how this option is used?

crates/walrus-sui/src/config.rs
```diff
@@ -0,0 +1,117 @@
...
+/// The wallet configuration is a mechanism for specifying the sui wallet config, with a potential
+/// override for the active environment.
+pub enum WalletConfig {
+    /// The path to the wallet configuration file. The wallet context will use the active
+    /// environment from the given configuration file.
+    Path(#[serde(default, deserialize_with = "resolve_home_dir")] PathBuf),
+    /// The path to the wallet configuration file with an optional overridden `active_env`. The
+    /// wallet context will use the specified `active_env` if it is non-null. Loading of the
+    /// `WalletContext` will fail if an invalid `active_env` is used. Specifying a null `active_env`
+    /// will use the `active_env` from the configuration file.
```
mlegner: Nit: Can you also mention the address override in the docstrings?

crates/walrus-sui/src/config.rs
```diff
@@ -0,0 +1,117 @@
...
+) -> Result<WalletContext> {
+    WalletConfig::load_wallet_context(wallet_path.map(WalletConfig::from_path).as_ref())
+}
+
+impl WalletConfig {
+    /// Construct a simple pass-through wallet configuration from a path.
+    pub fn from_path(path: impl AsRef<Path>) -> Self {
+        WalletConfig::Path(path.as_ref().to_path_buf())
+    }
+    /// Retrieves the path to the wallet configuration file.
```
mlegner, line 57: Nit: Could you add empty lines between the function definitions (also below)?

crates/walrus-service/src/client/config.rs
```diff
@@ -34,6 +38,55 @@ use crate::{
     common::utils,
 };
 
+/// Multi config for the client.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(untagged)]
+#[allow(clippy::large_enum_variant)]
```
mlegner, line 47: Question: Is this actually needed? The variants don't look that large to me...

crates/walrus-service/src/client/config.rs
mlegner: Question/nit: Could this be a function on `Config`?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -45,8 +45,15 @@ pub struct App {
...
     // NB: Keep this in sync with `crate::cli`.
     #[clap(long, verbatim_doc_comment, global = true)]
-    #[serde(default, deserialize_with = "crate::utils::resolve_home_dir_option")]
+    #[serde(
+        default,
+        deserialize_with = "walrus_utils::config::resolve_home_dir_option"
+    )]
     pub config: Option<PathBuf>,
+    /// The configuration context to use for the client, if omitted the default_context is used.
+    #[clap(long, verbatim_doc_comment, global = true)]
```
mlegner: Nit: I believe the `verbatim_doc_comment` is not needed here.

crates/walrus-sui/src/config.rs
wbbradley: This code was copied almost verbatim from the prior implementation of `load_wallet_context`. I'm not sure it makes sense to pull `./sui_config.yaml` out into a constant when in fact I don't want any other pieces of code to reference that location. In other words, this is intended to stay DRY. Also, there is a second fallback that is more complex `~/.sui/sui_config/client.yaml`, and putting that into a global const is a bit more involved due the dynamic nature of constructing paths...

crates/walrus-service/src/client/config.rs
wbbradley: I personally loathe "ignore" behaviors because it is equivalent to a silent failure. The user said they wanted a particular context, if we can't find it, I think we should fail loudly. I will make it fail if a context is provided to a single-config file with your suggested error message. I agree that that does seem unnecessary and potentially problematic.

crates/walrus-service/src/client/config.rs
wbbradley: I'm not sure that's necessary. Or rather, I'm not sure how I'd change the flow of the code if there are no contexts. Note that in order to select this codepath, the config would have to have been parsed as a MultiConfig, so it almost certainly has at least one context.

crates/walrus-service/src/client/config.rs
wbbradley: Yeah, unfortunately, Config is pretty large, but HashMap + String is small. So, the discrepancy gets called out by clippy.

crates/walrus-sui/src/config.rs
wbbradley: I'll leave it for now, but let me know if you feel strongly about it and I can pull it out into a const.

PR #1533, opened 2025-02-05
fix: Make event blob writer non blocking

sadhansood:
## Description

Currently, when attestation fails due to errors, the event blob writer continues to retry indefinitely for some type of errors, blocking progress for the storage node. This change de-couples event blob writing from event processing and:

1. Adds state management to pause attestations until the next epoch when invalid epoch errors occur
2. Adds logging for attestation state changes
3. Adds finite number of retries for the `certify_event_blob` contract call

This ensures event processing continues even when attestations fail, and attestations automatically resume in the next epoch. The change prevents unnecessary load from failed attestation attempts and ensures smoother epoch transitions.

### DIFF
--- a/crates/walrus-service/src/node/events/event_blob_writer.rs
+++ b/crates/walrus-service/src/node/events/event_blob_writer.rs
@@ -5,10 +5,8 @@
 //! Event blob writer.
 
 use std::{
-    fs,
-    fs::{File, OpenOptions},
-    io,
-    io::{BufWriter, Seek, SeekFrom, Write},
+    fs::{self, File, OpenOptions},
+    io::{self, BufWriter, Seek, SeekFrom, Write},
     path::{Path, PathBuf},
     sync::Arc,
 };
@@ -17,6 +15,7 @@ use anyhow::{Context, Result};
 use byteorder::{BigEndian, WriteBytesExt};
 use futures_util::future::try_join_all;
 use prometheus::{register_int_gauge_with_registry, IntGauge, Registry};
+use rand::{thread_rng, Rng};
 use rocksdb::Options;
 use serde::{Deserialize, Serialize};
 use sui_types::{event::EventID, messages_checkpoint::CheckpointSequenceNumber};
@@ -26,7 +25,15 @@ use typed_store::{
     Map,
 };
 use walrus_core::{ensure, metadata::VerifiedBlobMetadataWithId, BlobId, Epoch, Sliver};
-use walrus_sui::types::{BlobEvent, ContractEvent, EpochChangeEvent};
+use walrus_sui::{
+    client::SuiClientError,
+    types::{
+        move_errors::{MoveExecutionError, SystemStateInnerError},
+        BlobEvent,
+        ContractEvent,
+        EpochChangeEvent,
+    },
+};
 
 use crate::node::{
     errors::StoreSliverError,
@@ -44,6 +51,7 @@ use crate::node::{
 const CERTIFIED: &str = "certified_blob_store";
 const ATTESTED: &str = "attested_blob_store";
 const PENDING: &str = "pending_blob_store";
+const FAILED_TO_ATTEST: &str = "failed_to_attest_blob_store";
 const MAX_BLOB_SIZE: usize = 100 * 1024 * 1024;
 const NUM_CHECKPOINTS_PER_BLOB: u32 = 18_000;
 
@@ -64,6 +72,9 @@ pub struct EventBlobMetadata<T, U> {
 
 /// Metadata for a blob that is waiting for attestation.
 type PendingEventBlobMetadata = EventBlobMetadata<CheckpointSequenceNumber, ()>;
+
+/// Metadata for a blob that failed to attest.
+type FailedToAttestEventBlobMetadata = EventBlobMetadata<CheckpointSequenceNumber, BlobId>;
 
 /// Metadata for a blob that is last attested.
 type AttestedEventBlobMetadata = EventBlobMetadata<CheckpointSequenceNumber, BlobId>;
@@ -94,6 +105,16 @@ impl PendingEventBlobMetadata {
             event_cursor: self.event_cursor,
             epoch: self.epoch,
             blob_id: *blob_metadata.blob_id(),
+        }
+    }
+
+    fn to_failed_to_attest(&self, blob_id: BlobId) -> FailedToAttestEventBlobMetadata {
+        FailedToAttestEventBlobMetadata {
+            start: self.start,
+            end: self.end,
+            event_cursor: self.event_cursor,
+            epoch: self.epoch,
+            blob_id,
         }
     }
 }
@@ -194,6 +215,8 @@ pub struct EventBlobWriterFactory {
     attested: DBMap<(), AttestedEventBlobMetadata>,
     /// Pending blobs metadata.
     pending: DBMap<u64, PendingEventBlobMetadata>,
+    /// Failed to attest blobs metadata.
+    failed_to_attest: DBMap<(), FailedToAttestEventBlobMetadata>,
     /// Number of checkpoints per blob.
     num_checkpoints_per_blob: Option<u32>,
 }
@@ -231,6 +254,7 @@ impl EventBlobWriterFactory {
                 (PENDING, Options::default()),
                 (ATTESTED, Options::default()),
                 (CERTIFIED, Options::default()),
+                (FAILED_TO_ATTEST, Options::default()),
             ],
         )?;
         if database.cf_handle(CERTIFIED).is_none() {
@@ -246,6 +270,11 @@ impl EventBlobWriterFactory {
         if database.cf_handle(PENDING).is_none() {
             database
                 .create_cf(PENDING, &Options::default())
+                .map_err(typed_store_err_from_rocks_err)?;
+        }
+        if database.cf_handle(FAILED_TO_ATTEST).is_none() {
+            database
+                .create_cf(FAILED_TO_ATTEST, &Options::default())
                 .map_err(typed_store_err_from_rocks_err)?;
         }
         let certified: DBMap<(), CertifiedEventBlobMetadata> = DBMap::reopen(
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -96,6 +107,28 @@ impl PendingEventBlobMetadata {
...
+}
+
+impl FailedToAttestEventBlobMetadata {
+    fn to_attested(&self, blob_metadata: VerifiedBlobMetadataWithId) -> AttestedEventBlobMetadata {
+        AttestedEventBlobMetadata {
+            start: self.start,
+            end: self.end,
+            event_cursor: self.event_cursor,
+            epoch: self.epoch,
+            blob_id: *blob_metadata.blob_id(),
```
halfprice: question: is it possible that FailedToAttestEventBlobMetadata::blob_id is different from the blob_id in blob_metadata? Do we need the `blob_metadata` passed in?

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -366,23 +429,31 @@ impl EventBlobWriterFactory {
...
+///  |  EventBlobWriter  |
+///  +-------------------+
+///           |
+///           v
+///   +------------+    +-------------+    +-------------+    +---------------+
+///   |Current Blob|--->| Pending Blob|--->|Attested Blob|--->|System Contract|
+///   +------------+    +-------------+    +-------------+    +---------------+
+///         |                 |                  ^                   |
+///         |                 |                  |                   |
+///         |                 v                  |                   |
```
halfprice, line 430: I think the arrow should be attested blob -> failed to attest -> pending blob ?

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -739,16 +882,24 @@ impl EventBlobWriter {
...
-        &self,
+        &mut self,
         metadata: &VerifiedBlobMetadataWithId,
         checkpoint_sequence_number: CheckpointSequenceNumber,
     ) -> Result<()> {
+        let blob_id = *metadata.blob_id();
+
+        if self.attestations_paused() {
+            tracing::debug!("attestations are paused, skipping blob: {}", blob_id);
+            return Ok(());
```
halfprice, line 884: Remark: is return `Ok(())` the correct behavior? Attestation will be paused when certify_event_blob returns `EInvalidIdEpoch`, which could just mean that the current node may be slightly lagging. We shouldn't just skip attest this blob, should we?

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -489,6 +571,63 @@ pub struct EventBlobWriterConfig {
...
+        if self.attempts < self.max_attempts {
+            self.attempts += 1;
+            self.current_wait *= 2
+        }
+        // Add random jitter between -10% and +10%
+        let jitter = thread_rng().gen_range(-0.1..0.1);
+        self.current_wait = (self.current_wait as f64 * (1.0 + jitter)) as u64;
+    }
+
+    /// Records an event and checks if enough events have passed to retry.
```
mlegner: Nit:
```suggestion
    /// Records an event.
```

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -489,6 +571,63 @@ pub struct EventBlobWriterConfig {
...
+#[derive(Debug)]
+struct EventBackoff {
+    /// Base number of events to wait before retry
+    base_events: u64,
+    /// Current number of events to wait
+    current_wait: u64,
+    /// Number of events processed since last attempt
+    events_since_attempt: u64,
+    /// Maximum number of retry attempts
+    max_attempts: u32,
```
mlegner: Important: This doesn't seem to do what the name and docstring suggests. Instead, it defines the number of attempts after which the `current_wait` is no longer increased.

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -760,33 +911,96 @@ impl EventBlobWriter {
...
-            }
-            Err(e) => {
-                tracing::error!(
-                    error = ?e,
-                    blob_id = ?metadata.blob_id(),
-                    "failed to attest event blob"
-                );
+            Ok(()) => Ok(()),
+            Err(err) => {
+                let result = match err.downcast_ref::<SuiClientError>() {
```
mlegner: Suggestion: Have `certify_event_blob` directly return a `SuiClientError` instead of having to do the `downcast_ref`.

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -760,33 +911,96 @@ impl EventBlobWriter {
...
+                    Some(SuiClientError::StorageNodeCapabilityObjectNotSet) => {
+                        tracing::debug!(blob_id = ?blob_id,
+                                "Storage node capability object not set");
+                        Ok(())
+                    }
```
mlegner: Nit: This should be unreachable given that we do pass in the capability object.

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -760,33 +911,96 @@ impl EventBlobWriter {
...
+                    Some(
+                        e @ SuiClientError::TransactionExecutionError(
+                            MoveExecutionError::SystemStateInner(
+                                SystemStateInnerError::EIncorrectAttestation(_)
+                                | SystemStateInnerError::ERepeatedAttestation(_),
+                            ),
+                        ),
+                    ) => {
+                        tracing::warn!(
```
mlegner: Question: Why is this only a warning? It seems that something is seriously wrong in this case.

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -760,33 +911,96 @@ impl EventBlobWriter {
...
 
+    #[cfg(test)]
+    async fn attest_blob(
+        &mut self,
+        _metadata: &VerifiedBlobMetadataWithId,
+        _checkpoint_sequence_number: CheckpointSequenceNumber,
+    ) -> Result<()> {
+        tracing::info!("attesting blob in test");
+        Ok(())
+    }
```
mlegner: Question: Why don't we do the attestation in tests? Seems like an important component.

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -760,33 +911,96 @@ impl EventBlobWriter {
...
+
     /// Attests the next pending blob.
     ///
     /// This method processes the next pending blob by storing its slivers,
     /// attesting it, and updating the database state.
```
mlegner: Can you also describe what this function returns?

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -934,22 +1225,85 @@ impl EventBlobWriter {
...
+            return Ok(None);
+        };
+
+        if metadata.blob_id != blob_id {
+            return Ok(None);
+        }
+
+        batch.delete_batch(&self.failed_to_attest, std::iter::once(()))?;
+        Ok(Some(metadata.to_certified()))
+    }
```
mlegner: Remark: These two functions do exactly the same, just on different DB tables. Can you unify them?

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: good point, made it error!

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: fixed, thanks!

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: This seems correct to me. 1> failure path: pending -> failed_to_attest -> attested and 2> happy path: pending -> attested.

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: No, it is fine because we re-attest already attested blob on epoch change. So, this blob will be in attested queue and will be retried attestation upon epoch change

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: done

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: The stub contract used to throw error by default. I've fixed it now, and hence this is not needed anymore.

crates/walrus-service/src/node/events/event_blob_writer.rs
halfprice: Yes, you are right. I didn't understand the process before making that comment... My bad.

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -489,6 +560,64 @@ pub struct EventBlobWriterConfig {
...
+            max_attempts,
+            attempts: 0,
+        }
+    }
+
+    /// Returns the next wait count. Once max attempts is reached,
+    /// keeps returning the current wait count without doubling.
+    fn update_next_attempt(&mut self) {
+        if self.attempts < self.max_attempts {
+            self.attempts += 1;
```
mlegner: Suggestion: Increment this even after `max_attempts` in case we ever want to use this for something else.

crates/walrus-service/src/node/events/event_blob_writer.rs
mlegner: Nit: Can we also change the name to something like `increase_backoff_for_attempts` or similar?

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: done

PR #1535, opened 2025-02-05
feat: Store: extend existing blobs if possible

liquid-helium:
## Description

Resolve WAL-551

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: 
The walrus store command now checks for matching owned Blob objects and extends their lifetime if necessary instead of registering the blob again.

### DIFF
--- a/crates/walrus-service/src/client/resource.rs
+++ b/crates/walrus-service/src/client/resource.rs
@@ -58,10 +58,7 @@ impl PriceComputation {
             RegisterBlobOp::ReuseAndExtendNonCertified {
                 encoded_length,
                 epochs_extended,
-            } => {
-                self.storage_fee_for_encoded_length(*encoded_length, *epochs_extended)
-                    + self.write_fee_for_encoded_length(*encoded_length)
-            }
+            } => self.storage_fee_for_encoded_length(*encoded_length, *epochs_extended),
             _ => 0, // No cost for reusing registration or no-op.
         }
     }
@@ -194,7 +191,7 @@ impl<'a> ResourceManager<'a> {
         let to_be_processed = metadata_with_status
             .iter()
             .filter(|(metadata, blob_status)| {
-                if !matches!(store_when, StoreWhen::Always) && !persistence.is_deletable() {
+                if !store_when.is_ignore_status() && !persistence.is_deletable() {
                     if let Some(result) = self.blob_status_to_store_result(
                         *metadata.blob_id(),
                         epochs_ahead,


crates/walrus-service/src/client/responses.rs
```diff
@@ -127,6 +127,15 @@ pub enum BlobStoreResult {
         #[schema(value_type = Option<ObjectIdSchema>)]
         shared_blob_object: Option<ObjectID>,
     },
+    /// The blob was extended in lifetime.
+    Extended {
```
mlegner: Suggestion: Just use the `NewlyCreated` variant with `RegisterBlobOp::ReuseAndExtend`.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -382,6 +382,58 @@ async fn test_store_with_existing_blob_resource(
     Ok(())
 }
 
+/// Tests that the client can store multiple blobs with different epochs-ahead.
```
mlegner: Nit: This docstring doesn't seem to match the actual test.

crates/walrus-service/src/client/resource.rs
```diff
@@ -111,6 +123,12 @@ pub enum StoreOp {
         blob: Blob,
         operation: RegisterBlobOp,
     },
+    // Extend the lifetime of a registered blob.
+    // We only need to make sure it has enough resource, but not the exact amount.
```
mlegner: Question: I don't understand the meaning of "We only need to make sure it has enough resource, but not the exact amount.".

Also, can you make this into a docstring instead of a normal comment?

crates/walrus-service/src/client/resource.rs
```diff
@@ -426,10 +465,17 @@ impl<'a> ResourceManager<'a> {
...
                     && blob.storage.end_epoch >= self.write_committee_epoch + epochs_ahead
                     && blob.deletable == persistence.is_deletable()
-                    && (include_certified || blob.certified_epoch.is_none())
+                    && (include_certified || blob.certified_epoch.is_none()))
+                // Case 2: Certified blob that needs lifetime extension
+                || (blob.blob_id == *blob_id
+                    && include_certified
+                    && blob.deletable == persistence.is_deletable()
+                    && blob.certified_epoch.is_some()
+                    && blob.storage.end_epoch < self.write_committee_epoch + epochs_ahead)
```
mlegner: Suggestion: With these changes, I would remove the `epochs_ahead` function parameter altogether and make some simplifications:

```rust
    /// Finds a blob object with the given `blob_id` owned by the active wallet.
    ///
    /// Only includes non-expired blobs with the provided `persistence`.
    ///
    /// If `include_certified` is `true`, the function includes already certified blobs owned by the
    /// wallet.
    async fn find_blob_owned_by_wallet(
        &self,
        blob_id: &BlobId,
        persistence: BlobPersistence,
        include_certified: bool,
    ) -> ClientResult<Option<Blob>> {
        Ok(self
            .sui_client
            .owned_blobs(None, ExpirySelectionPolicy::Valid)
            .await?
            .into_iter()
            .find(|blob| {
                (blob.blob_id == *blob_id
                    && blob.storage.end_epoch > self.write_committee_epoch
                    && blob.deletable == persistence.is_deletable()
                    && (include_certified || blob.certified_epoch.is_none()))
            }))
    }
```

crates/walrus-service/src/client/resource.rs
```diff
@@ -149,6 +167,7 @@ impl<'a> ResourceManager<'a> {
             .iter()
             .filter(|(metadata, blob_status)| {
                 if !store_when.is_store_always() && !persistence.is_deletable() {
+                    // TODO: also collect known certified blobs and extend their lifetime if needed.
```
mlegner: Question: Are you referring to shared blobs here?

crates/walrus-service/src/client/resource.rs
```diff
@@ -175,15 +194,20 @@ impl<'a> ResourceManager<'a> {
 
         for (blob, op) in blobs_with_ops {
             // If the blob is deletable and already certified, add it results as noop.
```
mlegner: Nit: This comment doesn't seem to match any more.

crates/walrus-service/src/client.rs
```diff
@@ -626,18 +626,39 @@ impl Client<SuiContractClient> {
...
+                StoreOp::Extend { blob, end_epoch } => {
+                    extended_blobs_and_ops.push((blob, end_epoch))
+                }
+            });
+
+        let mut extended_results = Vec::with_capacity(extended_blobs_and_ops.len());
+        for (blob, end_epoch) in extended_blobs_and_ops {
+            if blob.certified_epoch.is_none() {
+                return Err(ClientError::from(ClientErrorKind::Other(
+                    "attempting to extend lifetime of an uncertified blob".into(),
```
mlegner: Important: I believe we currently don't check this in `get_existing_or_register_with_resources`. We *could* actually also accept non-certified blobs, but would then have to certify them before being able to extend them.

liquid-helium: Yes, that’s a valuable observation.

When this happens, does it imply that there is a concurrent operation trying to create a new blob object for the same blob, or could it indicate that a previous operation failed, resulting in the blob object being left in a non-certified state? Is that a possibility?

If there are no concerns about conflicts, similar to what we aimed to avoid with the StorageNodeCap, I think it could be a good idea to either wait for the certification or take the initiative to generate an operation that certifies it first with a shorter lifetime and then extends it.

What are your thoughts on this?

crates/walrus-service/src/client/resource.rs
liquid-helium: deleted!

crates/walrus-service/src/client/resource.rs
liquid-helium: Oops, that was intended as a temporary note for me to review later.

I have a question to help me better understand this code:
Does `metadata_with_status` only include shared objects, which appear to be gathered from the storage nodes?

crates/walrus-service/src/client/resource.rs
liquid-helium: deleted the second line, meant for impl notes :)

crates/walrus-e2e-tests/tests/test_client.rs
liquid-helium: Update the comments!

crates/walrus-service/src/client/responses.rs
liquid-helium: Nice suggestion, change applied.

crates/walrus-service/src/client/resource.rs
liquid-helium: Change applied, thanks!

crates/walrus-service/src/client.rs
liquid-helium: Change applied, thanks for the suggestion!

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -143,6 +143,9 @@ impl CliOutput for BlobStoreResultWithPath {
                         "(an existing registration was reused)"
                     }
                     RegisterBlobOp::ReuseAndExtend { .. } => "(the blob was extended in lifetime)",
+                    RegisterBlobOp::ReuseAndExtendNonCertified { .. } => {
+                        "(the blob was extended in lifetime)"
```
mlegner: Nit:
```suggestion
                        "(an existing registration was reused and extended)"
```

crates/walrus-service/src/client/resource.rs
```diff
@@ -51,6 +51,20 @@ impl PriceComputation {
             RegisterBlobOp::ReuseStorage { encoded_length } => {
                 self.write_fee_for_encoded_length(*encoded_length)
             }
+            RegisterBlobOp::ReuseAndExtend {
+                encoded_length,
+                epochs_extended,
+            } => {
+                self.storage_fee_for_encoded_length(*encoded_length, *epochs_extended)
+                    + self.write_fee_for_encoded_length(*encoded_length)
```
mlegner: Important (somewhat): There shouldn't be a write fee for a pure extension.

crates/walrus-service/src/client/resource.rs
```diff
@@ -96,21 +119,32 @@ impl RegisterBlobOp {
...
+            | RegisterBlobOp::ReuseAndExtendNonCertified { encoded_length, .. } => *encoded_length,
         }
     }
 
     /// Returns if the operation involved issuing a new registration.
     pub fn is_registration(&self) -> bool {
         matches!(self, RegisterBlobOp::RegisterFromScratch { .. })
-            // Reusing storage also requires a new registration.
-            || matches!(self, RegisterBlobOp::ReuseStorage { .. })
+    }
```
mlegner, line 124: Important: This change changes the semantics of this function. Before it returned whether or not a new blob object was registered, new it returns whether it creates a new registration from scratch. Can you make sure that this doesn't break any uses of this? Also, can you modify the docstring?

crates/walrus-service/src/client/resource.rs
```diff
@@ -96,21 +119,32 @@ impl RegisterBlobOp {
...
+
+    pub fn is_reuse_registration(&self) -> bool {
+        matches!(self, RegisterBlobOp::ReuseRegistration { .. })
     }
 
     /// Returns if the operation involved extending the lifetime of a registered blob.
     pub fn is_extend(&self) -> bool {
         matches!(self, RegisterBlobOp::ReuseAndExtend { .. })
     }
```
mlegner: Here as well the docstring is no longer correct.

crates/walrus-service/src/client/resource.rs
```diff
@@ -123,11 +157,13 @@ pub enum StoreOp {
...
         operation: RegisterBlobOp,
     },
-    // Extend the lifetime of a registered blob.
-    // We only need to make sure it has enough resource, but not the exact amount.
+    // Extend the lifetime of a registered blob instead of creating a new one.
     Extend {
+        // The original blob object.
         blob: Blob,
-        end_epoch: Epoch,
+        // The new end epoch.
```
mlegner: Nit: This docstring doesn't make sense. 😅 
```suggestion
```

crates/walrus-service/src/client/resource.rs
mlegner, line 1: Suggestion: Refactor the `get_existing_or_register_with_resources` function a bit. This has become very long (~140 LOC) and increasingly complex with deeply nested if/else constructs.

crates/walrus-service/src/client/resource.rs
mlegner: No, the `metadata_with_status` includes *all* blobs owned or shared. The status is fetched from storage nodes that have to keep track of all blobs.

crates/walrus-service/src/client/resource.rs
```diff
@@ -123,11 +157,13 @@ pub enum StoreOp {
...
+        encoded_length: u64,
+        epochs_extended: EpochCount,
     },
```
mlegner: Question: Do we actually need this separate variant given that we have the `RegisterBlobOp::ReuseAndExtend`?

crates/walrus-service/src/client/resource.rs
```diff
@@ -327,13 +366,29 @@ impl<'a> ResourceManager<'a> {
...
+                            RegisterBlobOp::ReuseAndExtend {
+                                encoded_length: *encoded_length,
+                                epochs_extended: epoch_delta,
+                            },
+                        ));
+                    } else {
+                        tracing::debug!(
+                            blob_id=%blob.blob_id,
+                            "blob is already registered but its lifetime is too short; extending it"
+                        );
```
mlegner: Nit: This should probably be moved up one level as it also applies to the other branch of the if statement.

crates/walrus-service/src/client/resource.rs
```diff
@@ -303,6 +342,7 @@ impl<'a> ResourceManager<'a> {
         let mut new_encoded_lengths = Vec::with_capacity(max_len);
 
         let mut extended_blobs = Vec::with_capacity(max_len);
+        let mut extended_blobs_noncertified = Vec::with_capacity(max_len);
```
mlegner: Question: (Why) do we need these separate lists? Could we not directly add the corresponding entries to `results`?

crates/walrus-service/src/client/resource.rs
```diff
@@ -303,6 +342,7 @@ impl<'a> ResourceManager<'a> {
         let mut new_encoded_lengths = Vec::with_capacity(max_len);
```
mlegner, line 319: General remark (not related to your PR): Not sure if it makes sense to specify the capacity here as this is an upper bound only and may be pretty wasteful. 🤷

crates/walrus-service/src/client.rs
mlegner, line 1: Suggestion: Also the `reserve_and_store_encoded_blobs` function would benefit from some refactoring, it has almost 200 LOC by now...

crates/walrus-service/src/client/resource.rs
mlegner, line 1: General remark (only partially related to your PR): I think the `store_operation_for_blobs` function could be refactored a bit: It currently puts all results in a single vector, which is then separated again at the caller (`reserve_and_store_encoded_blobs`). Instead, it could for example directly create and return the relevant vectors `noop_results`, `new_blobs_and_ops`, `extended_blobs_and_ops`, `certify_and_extend_blobs_and_ops`. Then we could also remove the `StoreOp` altogether.

crates/walrus-service/src/client.rs
```diff
@@ -626,34 +626,95 @@ impl Client<SuiContractClient> {
...
+            extended_results.push(BlobStoreResult::NewlyCreated {
+                blob_object: blob,
+                resource_operation: RegisterBlobOp::ReuseAndExtend {
+                    encoded_length,
+                    epochs_extended,
+                },
+                cost,
+                shared_blob_object: None,
+            });
+        }
```
mlegner: AFAIU, this extends each of the blobs in a separate transaction, is this correct? Could all these extensions be added to the confirmation PTB we already create?

crates/walrus-service/src/client.rs
```diff
@@ -626,34 +626,95 @@ impl Client<SuiContractClient> {
...
             )
             .await?;
+        let blobs_with_cert_and_extend: Vec<(&Blob, ConfirmationCertificate, Option<EpochCount>)> =
+            blobs_with_certificates
+                .into_iter()
+                .map(|(blob, cert)| {
+                    let epochs_ahead = certify_and_extend_blobs_and_ops
+                        .iter()
+                        .find(|(b, _)| b.blob_id == blob.blob_id)
+                        .map(|(_, _)| epochs_ahead);
```
mlegner: Suggestion: Instead of searching through these blobs, can't we just check if the `end_epoch` of the `blob` is sufficient?

crates/walrus-sui/src/client.rs
```diff
@@ -769,6 +769,20 @@ impl SuiContractClient {
             .multiple_pay_wal(address, amount, n)
             .await
     }
+
+    /// Certifies and extends the specified blob on Sui in a single transaction.
+    /// Returns the shared blob object ID if the post store action is Share.
+    pub async fn certify_and_extend_blobs(
```
mlegner: Suggestion: Can we clarify the semantics of this function a bit more? Also, instead of an "anonymous" tuple, it might make sense to use a custom struct to clarify the meaning of the different inputs (especially the `Option<EpochCount>`).

This would also allow adding the blobs that *only* need to be extended, by making both the certificate and the epochs optional.

crates/walrus-sui/src/client.rs
```diff
@@ -1617,6 +1610,93 @@ impl SuiContractClientInner {
...
+                        .await?;
+                }
+                PostStoreAction::Burn => {
+                    pt_builder.burn_blob(blob.id.into()).await?;
+                }
+                PostStoreAction::Keep => (),
+                PostStoreAction::Share => {
+                    pt_builder.new_shared_blob(blob.id.into()).await?;
+                }
+            }
```
mlegner: Nit: This seems to be the same as the code in the `certify_blobs` function. Can this be factored out into a separate function?

liquid-helium: Nice observation, change applied.

liquid-helium: Nice suggestion, doen.

crates/walrus-sui/src/client.rs
liquid-helium: Done.

crates/walrus-service/src/client.rs
liquid-helium: I switched it to a HashMap<> to improve the efficiency.

I believe your suggestion will work well. The reason I opted for this indirect method is to eliminate implicit information, similar to the way certify_and_extend() operates.

crates/walrus-service/src/client.rs
liquid-helium: Nice catch, change applied.

crates/walrus-service/src/client.rs
liquid-helium: I'll create an issue to keep track of this and handle the cleanup in a separate PR.

crates/walrus-service/src/client/resource.rs
liquid-helium: Yes, I checked that.

In fact, there are only two call sites.

crates/walrus-service/src/client/resource.rs
liquid-helium: Make the fee 0.

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: Change applied.

crates/walrus-service/src/client/resource.rs
liquid-helium: updated doc strings for all of them.

crates/walrus-service/src/client/resource.rs
liquid-helium: applied.

crates/walrus-service/src/client/resource.rs
liquid-helium: Will do in a separate PR :)

crates/walrus-service/src/client/resource.rs
liquid-helium: I agree with you.

Fun fact: This variant just doesn't seem to go away! I've removed it twice, yet it always manages to come back due to some revert operations on my part.

crates/walrus-service/src/client/resource.rs
liquid-helium: Done.

crates/walrus-service/src/client/resource.rs
liquid-helium: I agree with you, this code needs some cleanup, let me also include this in the cleanup issue.

crates/walrus-service/src/client/resource.rs
liquid-helium: We can put them into a single one, currently this is used for bookkeeping.  let me put all these into the new cleanup issue.

crates/walrus-service/src/client/resource.rs
```diff
@@ -202,8 +194,7 @@ impl<'a> ResourceManager<'a> {
         let to_be_processed = metadata_with_status
             .iter()
             .filter(|(metadata, blob_status)| {
-                if !store_when.is_store_always() && !persistence.is_deletable() {
-                    // TODO: also collect known certified blobs and extend their lifetime if needed.
+                if !matches!(store_when, StoreWhen::Always) && !persistence.is_deletable() {
```
mlegner: I think we should use `store_when.is_ignore_status()` here.

crates/walrus-service/src/client/resource.rs
```diff
@@ -51,6 +51,17 @@ impl PriceComputation {
...
+            RegisterBlobOp::ReuseAndExtend {
+                encoded_length,
+                epochs_extended,
+            } => self.storage_fee_for_encoded_length(*encoded_length, *epochs_extended),
+            RegisterBlobOp::ReuseAndExtendNonCertified {
+                encoded_length,
+                epochs_extended,
+            } => {
+                self.storage_fee_for_encoded_length(*encoded_length, *epochs_extended)
+                    + self.write_fee_for_encoded_length(*encoded_length)
```
mlegner: Important: I just realized that we also don't have to pay the write fee here: This is only due for the registration, so the `ReuseAndExtend` and `ReuseAndExtendNonCertified` cases are actually the same.

liquid-helium: Oh, thanks for the catch!
Now I see, the write fee is included already when first register it :)
Fix applied.

crates/walrus-service/src/client/resource.rs
liquid-helium: Oh, you are right, change applied.

PR #1539, opened 2025-02-05
feat(publisher): add actor to sync committee state

giac-mysten:
## Description

Introduces a `CommitteeRefresher` actor, the job of which is to track and sync the changes in the committee for all the clients that may be running (most importantly in a publisher or aggregator). The refresher notifies the running store and read operations that a committee change has occurred, prompting a restart of the operations with the new committee.

The `CommitteeRefresher` provides an interface (through a channel) with which clients can request the latest cached committee. If the refresher realizes that the cached committee is too old, it refreshes it from chain before answering.

Moreover, the refresher periodically refreshes the committee from chain. The frequency of these refreshes _greatly increase shortly before (5m) the expected time of epoch change_. This is to ensure that even i very long running tasks (big store operations) are running on the daemon during epoch change, the change is detected promptly.

Finally, if the epoch change is not detected through a refresh, but through a _failure_, the clients that fail because of epoch change can _force_ a refresh of the cached commitee information.

## Test plan

Manual testing storing with the publisher with very short epochs (30s).

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/refresh.rs
+++ b/crates/walrus-service/src/client/refresh.rs
@@ -260,18 +260,15 @@ impl<T: ReadClient> CommitteesRefresher<T> {
             Self::get_latest(&self.sui_client).await?;
 
         // First update, then notify if needed.
-        let comparison =
-            compare_current_previous_committees(&committees, self.last_committees.as_ref());
+        let are_different =
+            are_current_previous_different(&committees, self.last_committees.as_ref());
         self.last_committees = Arc::new(committees);
         self.last_price_computation = price_computation;
         self.epoch_state = epoch_state;
 
         // If the committee has changed, send a notification to the clients.
-        if let Err(error) = comparison {
-            tracing::info!(
-                differences=%error,
-                "the active committees have changed, notifying the clients"
-            );
+        if are_different {
+            tracing::debug!("the active committees have changed, notifying the clients");
             self.notify.notify_waiters();
         } else {
             tracing::trace!("the active committee has not changed");
@@ -373,17 +370,18 @@ impl CommitteesRefresherHandle {
 }
 
 /// Checks if two committes are different enough to require a notification to the clients.
-fn compare_current_previous_committees(
-    first: &ActiveCommittees,
-    second: &ActiveCommittees,
-) -> anyhow::Result<()> {
+fn are_current_previous_different(first: &ActiveCommittees, second: &ActiveCommittees) -> bool {
     // Compare the current committees.
-    first
+    if let Err(error) = first
         .current_committee()
-        .compare_functional_equivalence(second.current_committee())?;
+        .compare_functional_equivalence(second.current_committee())
+    {
+        tracing::debug!(differences = %error, "current committees are different");
+        return true;
+    }
 
     // Compare the previous committees, if present.
-    match (first.previous_committee(), second.previous_committee()) {
+    let previous_comparison = match (first.previous_committee(), second.previous_committee()) {
         (Some(first_previous), Some(second_previous)) => {
             first_previous.compare_functional_equivalence(second_previous)
         }
@@ -391,7 +389,11 @@ fn compare_current_previous_committees(
         _ => Err(anyhow::anyhow!(
             "one of the two sets has a previous committee, the other does not"
         )),
-    }?;
+    };
+    if let Err(error) = previous_comparison {
+        tracing::debug!(differences = %error, "previous committees are different");
+        return true;
+    }
 
-    Ok(())
+    false
 }


crates/walrus-service/src/client/metrics.rs
```diff
@@ -27,6 +27,7 @@ pub const WRITE_WORKLOAD: &str = "write";
 /// The name of the read workload.
 pub const READ_WORKLOAD: &str = "read";
 
+// TODO(giac): I realized this was not used except at startup?
```
giac-mysten: Question: the client metric set was initialized and written only once. Should we keep it? and if so, what is the goal?

crates/walrus-service/src/client.rs
```diff
@@ -81,6 +83,10 @@ mod refill;
 pub use refill::{RefillHandles, Refiller};
 mod multiplexer;
 
+/// The maximum number of retries for an operation that is stopped because of a committee change.
+// TODO: make this configurable.
```
karlwuest, line 87: Nit: link an issue?

crates/walrus-service/src/client.rs
```diff
@@ -1677,26 +1680,50 @@ async fn verify_blob_status_event(
     Ok(())
 }
 
-/// Retries the given function if the error may be caused by an epoch change, after refreshing the
-/// set of active committees.
-async fn retry_if_epoch_change<T, F, R, Fut>(client: &Client<T>, func: F) -> ClientResult<R>
+/// Retries the given function if the client gets notified that the committees have changed.
+async fn retry_if_committees_change<T, F, R, Fut>(client: &Client<T>, func: F) -> ClientResult<R>
```
karlwuest: Question: Is there a reason why this is not a method on `Client`?

crates/walrus-service/src/client/metrics.rs
karlwuest: I guess this is a leftover from the code that was originally in the stress client?

crates/walrus-service/src/client/cli.rs
```diff
@@ -56,6 +71,29 @@ pub fn load_configuration(path: &Option<PathBuf>) -> Result<Config> {
...
+) -> Result<(mpsc::Sender<RefreshKind>, Arc<Notify>)> {
+    let notify = Arc::new(Notify::new());
+    let (req_tx, req_rx) = mpsc::channel(REFRESHER_CHANNEL_SIZE);
+    let mut refresher = CommitteeRefresher::new(
+        DEFAULT_CACHE_VALIDITY,
+        sui_read_client,
+        req_rx,
+        notify.clone(),
+    )
+    .await?;
```
karlwuest: Question/Suggestion: Would this makes sense as a constructor on `CommitteeRefresher`?

crates/walrus-service/src/client/cli.rs
```diff
@@ -56,6 +71,29 @@ pub fn load_configuration(path: &Option<PathBuf>) -> Result<Config> {
...
+
+    tokio::spawn(async move {
+        if let Err(e) = refresher.run().await {
+            tracing::error!("failed to run the committee refresher: {:?}", e);
+        }
+    });
+
+    Ok((req_tx, notify))
```
karlwuest: Suggestion: 
Instead of returning a tuple here, these could be moved into a struct `CommitteeRefresherHandle` (perhaps also add a cancel token). This handle could be returned by a call on `CommitteeRefresherHandle` and would then allow creating clients by just passing the handle instead of both values separately.

crates/walrus-service/src/client.rs
```diff
@@ -215,8 +221,13 @@ impl Client<()> {
 
 impl<T: ReadClient> Client<T> {
     /// Creates a new read client starting from a config file.
-    pub async fn new_read_client(config: Config, sui_read_client: T) -> ClientResult<Self> {
-        Ok(Client::new(config, &sui_read_client)
+    pub async fn new_read_client(
```
karlwuest, line 211: Suggestion (also applies to `new_contract_client`): Often, only a single client is created. Maybe it would make sense to have another constructor that creates and starts the `CommitteeRefresher` for these scenarios?

crates/walrus-service/src/client/refresh.rs
```diff
@@ -0,0 +1,265 @@
...
+                "the active committee is fresh, skipping refresh"
+            );
+        }
+        Ok(())
+    }
+
+    /// Refreshes the data in the cache and _notifies_ the clients if the committee has changed.
+    ///
+    /// This function does _not_ update the last refresh time.
+    pub async fn refresh(&mut self) -> Result<()> {
```
karlwuest: Question/Suggestion: Should this be private instead of `pub`?

crates/walrus-service/src/client/refresh.rs
```diff
@@ -0,0 +1,265 @@
...
+        tracing::debug!("getting the latest active committee and price computation from chain");
+        let (committees, price_computation, epoch_state) =
+            Self::get_latest(&self.sui_client).await?;
+
+        // If the committee has changed, send a notification to the clients.
+        if are_committees_different(&committees, self.last_committees.as_ref()) {
+            tracing::info!("the active committee has changed, notifying the clients");
+            self.notify.notify_waiters();
```
karlwuest, line 272: Nit: There shouldn't be any race conditions (because of the cache validity, assuming `refresh` is only called as it is now), but it may still be better to update `last_committees` before notifying.

crates/walrus-service/src/client/refresh.rs
```diff
@@ -0,0 +1,265 @@
...
+        return false;
+    }
+
+    if first.current_committee() == second.current_committee()
+        && first.previous_committee() == second.previous_committee()
+    {
+        // The relevant committees for storing and reading are the same.
+        return false;
+    }
+
```
karlwuest: Suggestion: We may want to have a slightly more lenient comparison, probably as a special comparison function on `Committee`, because in many cases we don't care if the `Committee`s are not equal. In particular, we don't care if the name of a node changes or if the `next_epoch_public_key` changes.

Additionally, it's not necessary to first check if the `ActiveCommittees` are identical, you can just directly do:
```suggestion
    first.current_committee() != second.current_committee()
        || first.previous_committee() != second.previous_committee()
```

crates/walrus-service/src/client/refresh.rs
```diff
@@ -0,0 +1,265 @@
...
+            self.last_refresh = Instant::now();
+        } else if is_hard && self.last_hard_refresh.elapsed() > self.cache_validity {
+            tracing::debug!(
+                elapsed = ?self.last_hard_refresh.elapsed(),
+                "the active committee is forced to refresh, refreshing"
+            );
+            self.refresh().await?;
+            self.last_hard_refresh = Instant::now();
+            self.last_refresh = Instant::now();
+        } else {
```
karlwuest: Question: I kind of get the idea here, but I can't quite wrap my head around the distinction between hard and soft refresh, because they are both using the same `cache_validity`. I guess the only difference is that `last_refresh` is updated more frequently and if it is only used when it is really needed, then it should actually force a refresh?

crates/walrus-service/src/client/cli.rs
giac-mysten: very good point! I can also use the handle itself to manage the oneshot channels, which saves a bit of code here and there.

>perhaps also add a cancel token

would you add this to stop the refresher, i guess?

karlwuest: Sorry, I added the comment about the cancel token before going through everything and didn't change it afterwards. I don't think it's needed, because the refresher stops anyway once the mpsc channel is closed.

crates/walrus-service/src/client/cli.rs
giac-mysten: good point!

crates/walrus-service/src/client.rs
giac-mysten: makes sense!

crates/walrus-service/src/client/refresh.rs
giac-mysten: ah good point. No problems here i think, but yes, better change the order.

crates/walrus-service/src/client/refresh.rs
giac-mysten: Let me try to explain the idea behind this :) 
They are both a way to update the `self.last_committees`. The soft refresh is the regular way, where the client just wants to get the committes. The hard refresh is the way for the client to communicate that _it knows_ that whatever previous value it had did not work, and so a refresh from chain is needed immediately.

Imagine, for example, if the client fails because it cannot certify and therefore knows that the committee may have changed or the epoch advanced.

However, we could also have the event in which many store processes fail b/c of epoch change at the same time, and therefore trigger many hard refreshes in a short time. In that case, refreshing multiple times is not useful and could lead to rate limiting. Therefore, the hard refresh also has a "validity" period, where if one occurred soon before, no refresh is triggered. This period is set as the same `cache_validity` (but could be different).

Let me know if this makes sense!

giac-mysten: In general there is a tension here which i'm not too sure about. It came out of the discussion with @mlegner (who may have further input).

There are now two ways to update the committees value:
- with a refresh request from the client, if the last refresh was further away than `cache_validity`; or
- if the refresh timer fires -- this timer is used to make sure that the refresh occurs in proximity of epoch change, even if no requests come from clients.

now, we could eliminate the soft refresh by setting `DEFAULT_CACHE_VALIDITY == MAX_AUTO_REFRESH_INTERVAL`, and keep only the hard refresh. This way, the oldest value that a client can get (without asking for a hard refresh) is at most `DEFAULT_CACHE_VALIDITY` away.

wdyt?

karlwuest: Yes, I get all of that 😅 
Just to confirm: this just works essentially just because it's used less frequently, e.g. the normal refresh happens about every 30 seconds, but the hard refresh would be called after epoch change and therefore will actually cause a refresh?

karlwuest: >now, we could eliminate the soft refresh by setting DEFAULT_CACHE_VALIDITY == MAX_AUTO_REFRESH_INTERVAL, and keep only the hard refresh. This way, the oldest value that a client can get (without asking for a hard refresh) is at most DEFAULT_CACHE_VALIDITY away.
>wdyt?

I think this sounds reasonable. I guess in practice, the client (in an aggregator or publisher) would anyway end up refreshing every `DEFAULT_CACHE_VALIDITY` ?

karlwuest: Just to clarify: I also think it's fine to keep as is. 

It was just a bit confusing that the internals for soft vs hard refresh do almost exactly the same (as in: the behaviour is identical if you would only use hard refreshs vs only using soft refreshs, and the difference only matters if the two are mixed).

crates/walrus-service/src/client.rs
giac-mysten: i think there was a reason at some point 😬 but it does not anymore. Moved.

crates/walrus-service/src/client/refresh.rs
giac-mysten: I think `compare_essential` on committee would work for this, right?

karlwuest: TIL that this function exists 😄
I think we probably also want equality for the network addresses and network keys here?

giac-mysten: yeah, and also public_key?

karlwuest: Yes! (although that should be guaranteed not to change within the same epoch, so this would mainly be a sanity check)

crates/walrus-service/src/client/refresh.rs
giac-mysten: > this just works essentially just because it's used less frequently, e.g. the normal refresh happens about every 30 seconds, but the hard refresh would be called after epoch change and therefore will actually cause a refresh?

> the difference only matters if the two are mixed).

yes, exactly. The two are the same, and are only a way to express the urgency or priority of a refresh.

karlwuest: Thanks!

giac-mysten: > > now, we could eliminate the soft refresh by setting DEFAULT_CACHE_VALIDITY == MAX_AUTO_REFRESH_INTERVAL, and keep only the hard refresh. This way, the oldest value that a client can get (without asking for a hard refresh) is at most DEFAULT_CACHE_VALIDITY away.
> > wdyt?
> 
> I think this sounds reasonable. I guess in practice, the client (in an aggregator or publisher) would anyway end up refreshing every `DEFAULT_CACHE_VALIDITY` ?

yep, that is the main reason why i think removing the soft refresh may make sense here, and instead of `RefreshKind` have `RequestKind { GetCommittee, RefreshCommittee }`, where `GetCommittee` does not issue a refresh. And the periodic refresh is handled by the timer

karlwuest: That sounds good

crates/walrus-service/src/client/metrics.rs
giac-mysten: Maybe -- will remove for now then!

crates/walrus-service/src/client.rs
giac-mysten: actually already implemented now

crates/walrus-service/src/client/refresh.rs
```diff
@@ -239,27 +324,74 @@ impl<T: ReadClient> CommitteeRefresher<T> {
...
-        // The relevant committees for storing and reading are the same.
-        return false;
+    /// Awaits for a notification from the refresher that the active committee has changed.
+    pub async fn change_notified(&self) {
+        self.notify.notified().await
     }
+}
+
+/// Checks if two committes are different enough to require a notification to the clients.
+fn compare_current_previous_committees(
```
karlwuest: Nit: I think it would be a bit nicer to return a bool here instead of an error.

I assume this is to keep it consistent with the `compare_essential` that also returns an error? 

I think `compare_essential` is only used so far to do a sanity check that the committee after epoch change is the same as the `next_committee` before epoch change, which is why it returns an error, but here it's a normal occurence to have different committees.

crates/walrus-service/src/client/refresh.rs
```diff
@@ -176,22 +254,29 @@ impl<T: ReadClient> CommitteeRefresher<T> {
...
+        self.last_committees = Arc::new(committees);
+        self.last_price_computation = price_computation;
+        self.epoch_state = epoch_state;
+
         // If the committee has changed, send a notification to the clients.
-        if are_committees_different(&committees, self.last_committees.as_ref()) {
-            tracing::info!("the active committee has changed, notifying the clients");
+        if let Err(error) = comparison {
+            tracing::info!(
+                differences=%error,
```
karlwuest: Nit: The differences should probably not be part of the logging at the info level.

crates/walrus-service/src/client/refresh.rs
giac-mysten: that makes sense, i will print here the differences though, because it's nice to see them in the logs.

PR #1540, opened 2025-02-05
feat: finalize WAL token contract

karlwuest:
## Description

Finalizes the WAL token contract to be ready for TGE.
Closes WAL-202

## Test plan

Existing tests + manually ran integration tests with existing `testnet-contracts` to ensure compatibility

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/system_setup.rs
+++ b/crates/walrus-sui/src/system_setup.rs
@@ -24,7 +24,14 @@ use sui_move_build::{
 };
 use sui_sdk::{
     apis::ReadApi,
-    rpc_types::{SuiExecutionStatus, SuiTransactionBlockEffectsAPI, SuiTransactionBlockResponse},
+    rpc_types::{
+        SuiExecutionStatus,
+        SuiObjectDataFilter,
+        SuiObjectDataOptions,
+        SuiObjectResponseQuery,
+        SuiTransactionBlockEffectsAPI,
+        SuiTransactionBlockResponse,
+    },
     types::{
         base_types::ObjectID,
         programmable_transaction_builder::ProgrammableTransactionBuilder,
@@ -39,6 +46,7 @@ use sui_types::{
     SUI_CLOCK_OBJECT_ID,
     SUI_CLOCK_OBJECT_SHARED_VERSION,
     SUI_FRAMEWORK_ADDRESS,
+    SUI_FRAMEWORK_PACKAGE_ID,
 };
 use walkdir::WalkDir;
 use walrus_core::{ensure, EpochCount};
@@ -46,7 +54,6 @@ use walrus_core::{ensure, EpochCount};
 use crate::{
     client::retry_client::RetriableSuiClient,
     contracts::{self, StructTag},
-    test_utils::system_setup::mint_wal_to_addresses,
     utils::{get_created_sui_object_ids_by_type, resolve_lock_file_path},
 };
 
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/src/system_setup.rs
```diff
@@ -252,8 +254,7 @@ fn copy_recursively(source: impl AsRef<Path>, destination: impl AsRef<Path>) ->
 
 /// Publishes the `wal`, `wal_exchange`, and `walrus` packages.
 ///
-/// Returns the IDs of the walrus package and the `InitCap` as well as the `TreasuryCap`
-/// of the `WAL` coin.
+/// Returns the IDs of the packages, the `InitCap`, and.
```
mlegner: Nit: "and ..."? 😁

crates/walrus-sui/src/test_utils/system_setup.rs
```diff
@@ -282,14 +260,18 @@ pub async fn register_committee_and_stake(
...
-        // stake with storage nodes
-        if *amount_to_stake > 0 {
-            let _staked_wal = contract_client
-                .stake_with_pools(&[(node_cap.node_id, *amount_to_stake)])
-                .await?;
-        }
         node_capabilities.push(node_cap);
     }
+    // Use admin wallet to stake with storage nodes
+    let amounts_to_stake: Vec<_> = node_capabilities
```
mlegner: Nit: The variable name here is misleading, it's rather like `node_ids_with_stake_amounts`.

crates/walrus-sui/src/system_setup.rs
```diff
@@ -285,11 +286,31 @@ pub async fn publish_coin_and_system_package(
...
+    if let Some(treasury_cap_id) =
+        get_created_sui_object_ids_by_type(&transaction_response, &treasury_cap_struct_tag)?.first()
+    {
+        let sender = wallet.active_address()?;
+        if wallet
+            .get_object_owner(treasury_cap_id)
+            .await
+            .is_ok_and(|owner| owner == sender)
+        {
+            // Mint WAL to the publisher.
```
mlegner: Nit: I was first confused by the term "publisher" here; can you clarify that this is the wallet that published the contract?

crates/walrus-sui/src/system_setup.rs
```diff
@@ -285,11 +286,31 @@ pub async fn publish_coin_and_system_package(
...
+            mint_wal_to_addresses(
```
mlegner: Remark: This function is now only used here, is unnecessarily complicated as we only mint WAL to a single address, and has some slightly confusing comments like "Mint WAL to stake with storage nodes." Maybe we can move that function into this module and simplify it.

crates/walrus-sui/src/system_setup.rs
```diff
@@ -46,9 +46,12 @@ use walrus_core::{ensure, EpochCount};
 use crate::{
     client::retry_client::RetriableSuiClient,
     contracts::{self, StructTag},
+    test_utils::system_setup::mint_wal_to_addresses,
     utils::{get_created_sui_object_ids_by_type, resolve_lock_file_path},
 };
 
+const WAL_MINT_AMOUNT: u64 = 5_000_000_000 * 1_000_000_000; // 5 billion WAL
```
damirka, line 60: Does it make sense to write it the same way in Move? I find it a little confusing that total supply and decimals in move are separate.

karlwuest: In move, I used the decimals, because that's one of the metadata parameters, so we anyway need to specify it, and I think it helps readability to specify the amount in WAL and then use that to calculate the amount of FROST. 
Here it's not very important, because this will never apply to mainnet.

crates/walrus-sui/src/system_setup.rs
karlwuest: Done in 37e06a9a60024f4c3f02b8058a3bde0d075aa307

PR #1545, opened 2025-02-06
chore: run diesel migrations on walrus-backup boot or bust

wbbradley:
## Description

Ensure that walrus-backup database migrations are run upon boot of any walrus-backup process.

## Test plan

Manual testing locally, manual testing in testnet.

No release impact.

### DIFF
--- a/crates/walrus-service/build.rs
+++ b/crates/walrus-service/build.rs
@@ -1,0 +1,8 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Build script for the walrus-service crate.
+fn main() {
+    #[cfg(feature = "backup")]
+    println!("cargo:rerun-if-changed=migrations");
+}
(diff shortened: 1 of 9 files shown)


crates/walrus-service/build.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Build script for the walrus-service crate.
+fn main() {
```
halfprice, line 5: What is this file for?

wbbradley: In general, build.rs files are run to aid in specifying certain build parameters/conditions. In this instance, I'm doing this to ensure that if I change the .sql files (the migrations) then running `cargo build` will cause a rebuild of the `walrus-backup` binary. It's an affordance for iterating on schema/data migrations locally.

wbbradley: For reference: https://doc.rust-lang.org/cargo/reference/build-scripts.html

PR #1546, opened 2025-02-06
feat: Cli for blob metadata

liquid-helium:
## Description

Related to WAL-331

Closes WAL-597

Sample output:

```
walrus get-blob-attribute 0xb56e6c2611a140f498286278540cd73ccdbea2debfeb3b16da6712246312b487`
2025-02-12T21:13:59.564873Z  INFO walrus: client version: 1.13.0-4e78c6c9378f-dirty
2025-02-12T21:13:59.565441Z  INFO walrus_service::client::cli: using Walrus configuration from '/Users/heliu/.config/walrus/client_config.yaml'
2025-02-12T21:13:59.566445Z  INFO walrus_service::common::utils: using Sui wallet configuration from '/Users/heliu/.sui/sui_config/client.yaml'
2025-02-12T21:13:59.570429Z  INFO walrus_service::client::cli: using RPC URL set in wallet configuration

Attribute
alias: sea-elephant
date: date
today: today
  =+:   =+
+-(: +-(
```

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI:
Added the following new commands for blob attribute:
- walrus set-blob-attribute <BLOB_OBJ_ID>
- walrus get-blob-attribute [OPTIONS] <BLOB_OBJ_ID>
-  walrus remove-blob-attribute-fields [OPTIONS] <BLOB_OBJ_ID>
-  walrus remove-blob-attribute [OPTIONS] <BLOB_OBJ_ID>

### DIFF
--- a/crates/walrus-sui/src/types/move_structs.rs
+++ b/crates/walrus-sui/src/types/move_structs.rs
@@ -110,7 +110,7 @@ impl AssociatedContractStruct for Blob {
 /// The attribute struct for Blob objects.
 #[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
 pub struct BlobAttribute {
-    /// The key-value pairs.
+    /// The metadata key-value pairs.
     pub metadata: VecMap<String, String>,
 }
 
@@ -122,14 +122,13 @@ impl Default for BlobAttribute {
     }
 }
 
-impl BlobAttribute {
-    /// Creates a new BlobAttribute from any collection of key-value pairs.
-    pub fn from<I, K, V>(iter: I) -> Self
-    where
-        I: IntoIterator<Item = (K, V)>,
-        K: Into<String>,
-        V: Into<String>,
-    {
+impl<I, K, V> From<I> for BlobAttribute
+where
+    I: IntoIterator<Item = (K, V)>,
+    K: Into<String>,
+    V: Into<String>,
+{
+    fn from(iter: I) -> Self {
         let contents = iter
             .into_iter()
             .map(|(key, value)| Entry {
@@ -142,7 +141,9 @@ impl BlobAttribute {
             metadata: VecMap { contents },
         }
     }
+}
 
+impl BlobAttribute {
     /// Insert a key-value pair into the metadata.
     pub fn insert(&mut self, key: String, value: String) {
         if let Some(idx) = self.metadata.contents.iter().position(|e| e.key == key) {


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -449,6 +449,50 @@ pub enum CliCommands {
...
+    /// Get the attribute of a blob.
+    GetBlobAttribute {
+        /// The object ID of the blob to get the attribute of.
+        #[clap(index = 1)]
+        blob_obj_id: ObjectID,
+        /// If unset, prints the blob to stdout.
+        #[clap(long)]
+        #[serde(default, deserialize_with = "crate::utils::resolve_home_dir_option")]
+        out: Option<PathBuf>,
+    },
```
mlegner: Question: It's unclear *which* attribute is requested here. Also, why is the blob written anywhere?

crates/walrus-service/src/client/responses.rs
```diff
@@ -633,6 +640,14 @@ pub struct ShareBlobOutput {
     pub amount: Option<u64>,
 }
 
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+/// The output of the `walrus get-blob-attribute` command.
+pub struct GetBlobAttributeOutput {
+    /// The attribute of the blob.
```
mlegner, line 646: Nit: This seems to contain both the blob *and* the attribute. It's a bit unintuitive to get the blob object as well, why not just use `Option<BlobAttribute>` here?

crates/walrus-sui/src/client.rs
```diff
@@ -801,21 +801,36 @@ impl SuiContractClient {
...
             .await
+        {
+            Ok(()) => Ok(()),
+            Err(SuiClientError::TransactionExecutionError(MoveExecutionError::Blob(
+                BlobError::EMissingMetadata(_),
+            ))) if force => {
+                inner
+                    .add_blob_attribute(blob_obj_id, &BlobAttribute::from(pairs_clone))
+                    .await
+            }
```
mlegner: Suggestion: Check first if the attribute exists and create it if needed instead of potentially sending an invalid transaction first.

crates/walrus-sui/src/types/move_structs.rs
mlegner, line 1: Important: Doesn't that have to match the field names of the contracts? (cc @karlwuest)

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1235,3 +1279,14 @@ impl UserConfirmation {
         matches!(self, UserConfirmation::Required)
     }
 }
+
+fn parse_attr_pair(s: &str) -> Result<(String, String)> {
+    let parts: Vec<&str> = s.trim().split('=').collect();
```
mlegner: Important: Does this mean we cannot have an `=` in any of the attributes?

liquid-helium: Ah, that's a great catch! Thank you!

It seems we can address this by prompting the user to be more explicit, such as using a format like --key "xxxx" --value "yyyy" --key "zzzzz" --value "aaaaa," or even utilizing JSON formats.

Do you have any suggestions off the top of your head?

crates/walrus-service/src/client/cli/args.rs
liquid-helium: This command retrieves the complete BlobAttribute, including all key-value pairs.

It will fetch both the BlobAttribute and the Blob Object, but it does not retrieve the actual blob data.

I've included this information in the comments.

If you think we should add an API for fetching a specific key, let me know. Just keep in mind that we always fetch the entire Metadata dynamic field from Sui under the hood.

crates/walrus-sui/src/types/move_structs.rs
liquid-helium: I made this change because the 'metadata' field name in the JSON output can be somewhat confusing. 

I tested it on the testnet, and it works for the CLI commands. However, if we need to keep it consistent, I can revert it back to 'metadata.'

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Oh, my mistake, we already have a --json mode for the CLI, path removed!

crates/walrus-sui/src/client.rs
liquid-helium: Thank you for the suggestion! I agree that this approach would help prevent potentially failed write attempts.

Does this suggest that reading from Sui is way cheaper than writing (or write attempts)? I conducted 'blind writes' in both the insert_pair and add_attribute cases, based on the (incorrect) assumption that read and write operations have similar costs.
For the insert_or_update case, if we're checking each time and the attribute is usually already present (since it only needs to be created once), it leads to unnecessary reads for every operation.

This brings to mind that if we assume insert or update operations will occur much more frequently than the initial create, perhaps I should revise the logic for add_attribute as well to perform an insert operation when force is true.

I will keep them as for now, since these two APIs actually covers the same functionalities essentially, maybe we can keep only one of them?

crates/walrus-service/src/client/responses.rs
liquid-helium: Good idea, change applied.

crates/walrus-sui/src/types/move_structs.rs
mlegner: I'm a bit surprised that this works. Probably because we only look at the BCS encoding. @karlwuest should know more.

crates/walrus-sui/src/client.rs
mlegner: AFAIK, reads have much lower complexity than writes: They don't require validators to be involved, don't require signing, etc. Also, writes come with actual (gas) costs, even if the transaction fails.

crates/walrus-service/src/client/cli/args.rs
mlegner: Being able to provide the key-value pairs as JSON would definitely be great. But it would be good to have a simpler variant in addition. Both the existing variant and the `--key`/`--value` options sound good to me.

crates/walrus-sui/src/types/move_structs.rs
karlwuest: Yes, this works because we only look at the BCS, which means that the names of the fields of the structs don't matter (only their types matter).

crates/walrus-sui/src/client.rs
liquid-helium: Change applied!

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Switched to --attr "k1" "v1" --attr "k2" "v2" format, to allow multiple attribute fields.

For the json format, as we discussed, since we already have the --json mode, it might be not necessary to add it here.

crates/walrus-sui/src/types/move_structs.rs
liquid-helium: Thanks @mlegner and @karlwuest for the tips!
I changed it back to metadata, as discussed with @mlegner offline, to avoid any potential confusion.

mlegner, line 1: Remark (related to a previous PR): Instead of the `BlobAttribute::from` function, could you implement the `From` trait? This would be the more canonical way of doing it.

liquid-helium: Done, thanks for the suggestion!

PR #1548, opened 2025-02-07
feat: Sync storage node config

liquid-helium:
## Description

Periodically update on-chain storage node config, from local storage node config file.

Related to WAL-245

## Test plan

---

## Release notes

- [x] Storage node:
    Automatically reload the StorageNodeConfig from disk and update the on-chain configuration when changes are detected.
    Some changes may result in a automatic restart of the storage nodes.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -361,7 +361,7 @@ fn main() -> anyhow::Result<()> {
             let result = commands::run(
                 load_from_yaml(&config_path)?,
                 cleanup_storage,
-                Some(Arc::new(StorageNodeConfigLoader::new(config_path.clone()))),
+                Arc::new(StorageNodeConfigLoader::new(config_path.clone())),
                 ignore_sync_failures,
             );
 
@@ -465,7 +465,7 @@ mod commands {
     pub(super) fn run(
         mut config: StorageNodeConfig,
         cleanup_storage: bool,
-        config_loader: Option<Arc<dyn ConfigLoader>>,
+        config_loader: Arc<dyn ConfigLoader>,
         ignore_sync_failures: bool,
     ) -> anyhow::Result<()> {
         if cleanup_storage {
@@ -565,7 +565,7 @@ mod commands {
             exit_notifier,
             event_manager,
             cancel_token.child_token(),
-            config_loader,
+            Some(config_loader),
             ignore_sync_failures,
         )?;
 
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1127,7 +1127,7 @@ impl StorageNode {
         event_handle: EventHandle,
         event: &EpochChangeStart,
     ) -> anyhow::Result<()> {
-        #[cfg(not(test))]
+        // #[cfg(not(test))]
         if let Some(c) = self.config_synchronizer.as_ref() {
             c.sync_node_params().await?;
         }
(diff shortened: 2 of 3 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -462,6 +465,7 @@ mod commands {
     pub(super) fn run(
         mut config: StorageNodeConfig,
         cleanup_storage: bool,
+        config_loader: Option<Arc<dyn ConfigLoader>>,
```
halfprice: Nit: will there be an API where run() takes `None`? If not, should we take Arc<dyn ConfigLoader> here, and call
`StorageNode::builder().with_config_loader(Some(config_loader))`

crates/walrus-service/src/test_utils.rs
```diff
@@ -423,6 +449,7 @@ impl SimStorageNodeHandle {
     async fn build_and_run_node(
         config: Arc<RwLock<StorageNodeConfig>>,
         num_checkpoints_per_blob: Option<u32>,
+        config_loader: Option<Arc<dyn ConfigLoader>>,
```
halfprice: Nit: similar question here.

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -64,20 +93,24 @@ impl ConfigSynchronizer {
...
-        self.contract_service
-            .sync_node_params(&self.config, self.node_capability_object_id)
-            .await
+        if let Some(config_loader) = &self.config_loader {
+            let config = config_loader.load_storage_node_config().await?;
+            self.contract_service
+                .sync_node_params(&config, self.node_capability_object_id)
+                .await
+        } else {
+            tracing::warn!("config loader is not set");
```
halfprice, line 104: Question: it seems that the old behavior does periodic sync using the `self.config`. Even if the config loader is not set in the node, is it no longer needed to call `sync_node_params` on self.config?

crates/walrus-service/src/node.rs
```diff
@@ -1118,6 +1127,7 @@ impl StorageNode {
         event_handle: EventHandle,
         event: &EpochChangeStart,
     ) -> anyhow::Result<()> {
+        #[cfg(not(test))]
```
halfprice: Can you add some docs why this is excluded from tests?

crates/walrus-service/bin/node.rs
liquid-helium: Thats a good point, thanks!

Change applied. I stopped at StorageNodeRuntime::start() with the hope that we will use this as a common entry point for prod and tests in the future. Let me know if you have strong preference!

crates/walrus-service/src/test_utils.rs
liquid-helium: Done.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Yes, this setup is intentional. Without the config-loader, the config_synchronizer is limited to performing only a few basic checks, which are more like cleanup operations.

crates/walrus-service/src/node.rs
liquid-helium: Nice catch! This is not necessary any more, removed.

PR #1551, opened 2025-02-07
feat: ignore resources also from the CLI

giac-mysten:
## Description

- Introduces the `--ignore-resources` flag in the CLI, to skip checks of owned resources/blobs when storing.
- Caches the owned blobs during multi-blob stores.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: introduces the `--ignore-resources` flag.

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -117,12 +117,15 @@ pub enum StoreWhen {
 }
 
 impl StoreWhen {
-    /// Returns `true` if the operation is [`Self::Always`].
+    /// Returns `true` if the operation ignore the blob status.
+    ///
+    /// If `true`, the client should store the blob, even if the blob is already stored on Walrus
+    /// for a sufficient number of epochs.
     pub fn is_ignore_status(&self) -> bool {
         matches!(self, Self::Always | Self::AlwaysIgnoreResources)
     }
 
-    /// Returns `true` if the operation is [`Self::NotStoredIgnoreResources`].
+    /// Returns `true` if the operation should ignore the resources in the wallet.
     pub fn is_ignore_resources(&self) -> bool {
         matches!(
             self,


crates/walrus-service/src/client.rs
```diff
@@ -125,12 +125,15 @@ impl StoreWhen {
...
-        if force {
-            Self::Always
-        } else {
-            Self::NotStored
+    /// Returns [`Self`] based on the value of the `force` and `ignore-resources` flags.
+    pub fn from_flags(force: bool, ignore_resources: bool) -> anyhow::Result<Self> {
+        match (force, ignore_resources) {
+            (true, true) => Err(anyhow!(
+                "both force and ignore-resources flags are set; please only specify one of them"
+            )),
```
mlegner: I think the two flags are somewhat orthogonal: `force` ignores the current status while `ignore-resources` ignores owned resources. IMO it *does* make sense to specify both of them simultaneously.

mlegner: Maybe we could change the `StoreWhen` and either add additional variants or make it a struct with boolean fields like `ignore_resources`, `ignore_status`, and possibly others in the future.

giac-mysten: yea, i was thinking about this too. My idea was that forcing when you don't even know what resources you have is not great, but it may make sense sometimes I guess.
Added a `AlwaysIgnoreResources` to the enum.
If we add other flags, it may make sens to change to a struct.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -170,6 +170,13 @@ pub enum CliCommands {
     ///
     /// If the `--force` flag is used, this operation always creates a new certification for the
     /// blob (possibly reusing storage resources or uncertified but registered blobs).
+    ///
+    /// The `--ignore-resources` flag can be used to ignore the storage resources owned by the
+    /// wallet and always purchase new storage resources, bypassing the need to check the owned
+    /// resources on chain. Using this flag could speed up the store operation in case there are
+    /// thousands of resources or blobs owned by the wallet.
+    ///
+    /// At this moment, only one between `--force` and `--ignore-resources` can be used at once.
```
```suggestion
```

crates/walrus-service/src/client.rs
```diff
@@ -108,10 +108,12 @@ pub enum StoreWhen {
...
-    Always,
     /// Check the status of the blob before storing it, and store it only if it is not already.
     NotStored,
+    /// Store the blob always, without checking the status.
+    Always,
+    /// Store the blob always, without checking the status, and ignore the resources in the wallet.
+    AlwaysIgnoreResources,
 }
 
 impl StoreWhen {
```
mlegner, line 119: Important: The methods need to be adjusted:
```rust
    /// Returns `true` if the operation indicates to ignore the status of the blob.
    pub fn is_store_always(&self) -> bool {
        matches!(self, Self::Always | Self::AlwaysIgnoreResources)
    }

    /// Returns `true` if the operation indicates to ignore the resources in the current wallet.
    pub fn is_ignore_resources(&self) -> bool {
        matches!(
            self,
            Self::NotStoredIgnoreResources | Self::AlwaysIgnoreResources
        )
    }
```

We could also rename the `is_store_always` to something like `is_ignore_status`.

giac-mysten: :facepalm:

crates/walrus-service/src/client.rs
```diff
@@ -118,13 +118,16 @@ pub enum StoreWhen {
...
     /// Returns `true` if the operation is [`Self::Always`].
-    pub fn is_store_always(&self) -> bool {
-        matches!(self, Self::Always)
+    pub fn is_ignore_status(&self) -> bool {
+        matches!(self, Self::Always | Self::AlwaysIgnoreResources)
     }
 
     /// Returns `true` if the operation is [`Self::NotStoredIgnoreResources`].
     pub fn is_ignore_resources(&self) -> bool {
```
mlegner: Nit: Can you also update the docstrings?

PR #1553, opened 2025-02-07
chore: ensure process exits when a panic! is encountered

wbbradley:
## Description

Force Walrus binaries to crash when a panic! is encountered to avoid running with no progress or undefined behavior.

## Test plan

CI, manual testing.

---

No functional release impact.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -126,10 +126,17 @@ missing_debug_implementations = "warn"
 missing_docs = "warn"
 unexpected_cfgs = { level = "allow", check-cfg = ['cfg(msim)'] }
 
+[profile.release]
+panic = 'abort'
+
+[profile.dev]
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -126,10 +126,17 @@ missing_debug_implementations = "warn"
 missing_docs = "warn"
 unexpected_cfgs = { level = "allow", check-cfg = ['cfg(msim)'] }
 
+[profile.release]
+panic = 'abort'
+
+[profile.dev]
+panic = 'abort'
```
bmwill, line 133: Just note that in tests things are forced to unwind since thats required of the test-harness machinery

wbbradley: Yep, that's fine/expected. Thanks!

PR #1554, opened 2025-02-07
chore: use with_label to record metrics

halfprice:
## Description

Just to be consistent in the way we record metrics.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/common/telemetry.rs
+++ b/crates/walrus-service/src/common/telemetry.rs
@@ -38,7 +38,6 @@ use tower_http::trace::{MakeSpan, OnResponse};
 use tracing::{field, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 use walrus_core::Epoch;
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/common/telemetry.rs
```diff
@@ -38,6 +38,7 @@ use tower_http::trace::{MakeSpan, OnResponse};
 use tracing::{field, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 use walrus_core::Epoch;
+pub(crate) use walrus_utils::with_label;
```
mlegner: Question: Why do we need this?

halfprice: Good catch. I think this line was somewhere in the middle of this file and after changing the place of with_label, it got moved automatically.

I've removed and use `walrus_utils::with_label` everywhere.

PR #1557, opened 2025-02-07
chore: add a metrics_address parameter to walrus-backup

wbbradley:
## Description

Allow metrics_address to be changed for walrus-backup via the command line. This is to ensure that the orchestrator and fetcher can be run on the same machine easily for local testing purposes.

## Test plan

Manual testing.

No release impact.

### DIFF
--- a/crates/walrus-service/bin/backup.rs
+++ b/crates/walrus-service/bin/backup.rs
@@ -27,6 +27,12 @@ use walrus_service::{
 struct Args {
     #[clap(long, short, help = "Specify the config file path to use")]
     config: PathBuf,
+    #[clap(
+        long,
+        short,
+        help = "Override the metrics address to use (ie: 127.0.0.1:9184)"
+    )]
+    metrics_address: Option<std::net::SocketAddr>,
     #[command(subcommand)]
     command: BackupCommands,
 }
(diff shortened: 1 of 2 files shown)


crates/walrus-service/bin/backup.rs
```diff
@@ -20,6 +20,12 @@ use walrus_service::{
 struct Args {
     #[clap(long, short, help = "Specify the config file path to use")]
     config: PathBuf,
+    #[clap(
+        long,
+        short,
+        help = "Specify the metrics address to use (ie: 127.0.0.1:9184)"
```
halfprice: Nit: use `defaults::metrics_address` here so that this won't need to be an Option<>?

wbbradley: It usually comes in the config file, so forcing it to be specified on the command line seems like a break from the rest of the configuration that is otherwise unneeded? This affordance is just to enable locally running it... I can see it both ways, hard to say which is "correct?"

halfprice: I mean doing something like this: https://github.com/MystenLabs/walrus/blob/03922c7a1528542f2e0d44f7944e2165617a07d5/crates/walrus-service/bin/node.rs#L292-L293

So that if the the command line doesn't provide one, a default one will be used and metrics_address can then be std::net::SocketAddr instead of Option<std::net::SocketAddr>

wbbradley: Yep, that would break the configuration file behavior, because it would always read from the defaults and never read from the config. Note that there are now 3 sources of this datapoint: command-line, config-file, and defaults. My preferred order is command-line, config, default. With the change suggested here, it would be command-line, default.

halfprice: Make sense!

PR #1560, opened 2025-02-10
feat(publisher): add cache for JWT duplicate suppression

giac-mysten:
## Description

Adds a cache for seen JWTs, to ensure that the same JWT cannot be used multiple times.

Entries in the cache all have an expiration, which corresponds to `exp` in the JWT; after this expiration, the entries are removed from the cache. A background job periodically removes expired entries in the cache. It is the responsibility of the issuer of the JWTs  to ensure that each has a unique `jti` field (JWT ID, see the [RFC](https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7))

The cache has a maximum size; under load, if the max cache size is reached, the publisher will stop accepting requests. The cache size is a tunable parameter.

## Test plan

Added tests to all new components.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon/auth.rs
+++ b/crates/walrus-service/src/client/daemon/auth.rs
@@ -26,10 +26,10 @@ pub const PUBLISHER_AUTH_DOMAIN: &str = "auth.publisher.walrus.space";
 pub struct Claim {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     /// Token is issued at (timestamp).
-    pub iat: Option<u64>,
+    pub iat: Option<i64>,
 
     /// Token expires at (timestamp).
-    pub exp: u64,
+    pub exp: i64,
 
     /// A unique identifier for the token, the JTI (JWT ID).
     ///
@@ -250,12 +250,7 @@ pub async fn verify_jwt_claim(
         Ok(claim) => {
             // To avoid race conditions between store requests, we insert the token into the cache
             // now, and later remove it if the JWT verification fails.
-            let Some(expiration) = claim
-                .exp
-                .try_into()
-                .ok()
-                .and_then(|exp| DateTime::from_timestamp(exp, 0))
-            else {
+            let Some(expiration) = DateTime::from_timestamp(claim.exp, 0) else {
                 return Err(PublisherAuthError::InvalidTimestamp.to_response());
             };
 
@@ -381,12 +376,12 @@ mod tests {
     const OTHER_ADDRESS: &str =
         "0x1111111111111111111111111111111111111111111111111111111111111111";
 
-    const FAR_EXP: u64 = 33292598400; // 3025-01-01 00:00:00 UTC
+    const FAR_EXP: i64 = 33292598400; // 3025-01-01 00:00:00 UTC
 
     fn auth_config_for_tests(
         secret: Option<&str>,
         algorithm: Option<Algorithm>,
-        expiring_sec: u64,
+        expiring_sec: i64,
         verify_upload: bool,
     ) -> AuthConfig {
         let mut config = AuthConfig {
@@ -405,7 +400,7 @@ mod tests {
     }
 
     fn setup_router_and_token(
-        expiring_sec: u64,
+        expiring_sec: i64,
         verify_upload: bool,
         use_secret: bool,
         claim: Claim,
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -616,7 +616,7 @@ pub struct PublisherArgs {
     /// the "issued at" (`iat`) value.
     #[clap(long, default_value_t = 0)]
     #[serde(default)]
-    pub jwt_expiring_sec: u64,
+    pub jwt_expiring_sec: i64,
     /// If set, the publisher will verify that the requested upload matches the claims in the JWT.
     ///
     /// Specifically, the publisher will:
--- a/crates/walrus-service/src/client/config.rs
+++ b/crates/walrus-service/src/client/config.rs
@@ -124,7 +124,7 @@ pub struct AuthConfig {
     /// If set to `0`, the publisher will not check that the expiration is correctly set based in
     /// the issued-at time (iat) and expiration time (exp) in the JWT. I.e., if `expiring_sec > 0`,
     /// the publisher will check that `exp - iat == expiring_sec`.
-    pub(crate) expiring_sec: u64,
+    pub(crate) expiring_sec: i64,
     /// Verify the upload epochs and address for `send_object_to` in the request.
     ///
     /// The token expiration is still checked, even if `verify_upload == true`.


crates/walrus-service/src/client/daemon.rs
```diff
@@ -191,8 +196,10 @@ impl<T: WalrusWriteClient + Send + Sync + 'static> ClientDaemon<T> {
     }
 
     /// Constructs a new [`ClientDaemon`] with combined aggregator and publisher functionality.
+    #[allow(clippy::too_many_arguments)]
```
yanganto: If the option are `jwt-cache-size` and `jwt-cache-refresh-interval` and flatten in the existing auth config, and we do not need to pass additional argument, and `fn` will be easy to read.


In my personal opinion,
Although options is a long list, using `sturct`s do not help to reduce the complexity.  Let developer jumping in a lot of small files not always clean than a big options list.

If you do not want to change it is still good from my end.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -30,6 +31,13 @@ pub struct Claim {
     /// Token expires at (timestamp).
     pub exp: u64,
 
+    /// A unique identifier for the token, the JTI (JWT ID).
+    ///
+    /// See [RFC 7519][rfc7519s4.1.7].
+    ///
+    /// [rfc7519s4.1.7]: https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.7
+    pub jti: String,
```
yanganto, line 39: Will the cache force enabled?

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -320,14 +370,19 @@ mod tests {
...
+        daemon::{auth_layer, cache::CacheConfig},
+    };
 
     // Fixtures and helpers for tests.
 
     const ADDRESS: [u8; SUI_ADDRESS_LENGTH] = [42; SUI_ADDRESS_LENGTH];
     const OTHER_ADDRESS: &str =
         "0x1111111111111111111111111111111111111111111111111111111111111111";
 
+    const FAR_EXP: u64 = 33292598400; // 3025-01-01 00:00:00 UTC
```
yanganto: Just wonder is there anything bad to use `MAX` for a never expiring value in test cases.

crates/walrus-service/src/client/daemon.rs
```diff
@@ -239,7 +251,7 @@ impl<T: WalrusWriteClient + Send + Sync + 'static> ClientDaemon<T> {
                     .route_layer(
                         ServiceBuilder::new()
                             .layer(axum::middleware::from_fn_with_state(
-                                Arc::new(auth_config),
+                                (Arc::new(auth_config), Arc::new(token_cache)),
```
yanganto: Will a `JWTConfig` better here?

Cache config is misleading, people easy to think we cache some response, but filtering the tokens.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: The problem is that the conversion from number to datetime ([here](https://github.com/MystenLabs/walrus/blob/1caa4a65ec29da4eb46a0738b8e5e020cf899e38/crates/walrus-service/src/client/daemon/auth.rs#L257C1-L257C66)) requires a `i64`, and `u64::MAX` is not a valid `i64`.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: Yes, let me reply here also to the above on why this is necessary.

> The feature is force enabled now. Will we let this optional?
> If the token include the pre-calculated blob-id from client side and the token issuer only give token for exactly blob-id, the publisher will not be abused, user only can upload the exactly file with the exactly token. Multiple file uploading with an ambiguous token will not work any more. However, for the normal case, we will have no limitation from the cache, so if the other feature is ready, I will disable the cache.

there are a few problems with this:

1. In general, we cannot know the pre-calculated blob ID, because this requires encoding the blob, which is the job of the publisher. If the client can encode the blob, then it can also probably store the slivers itself (and maybe the certify transaction is done by the developer's wallet through an API call).
2. Even if we do have a pre-calculated blob ID (or file hash), the duplicate suppression is necessary. This is because just checking if a blob is already stored requires a lot of work (getting the blob ID, checking the status on the storage nodes, and possibly checking the event on chain). If we don't have duplicate suppression, someone with one valid token can completely stall the publisher for a long time.

In any case, if you feel this should be optional for any specific use case, we can do so in a follow up pr.

yanganto: Thanks for your explain.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: Actually, the JWT did not say anything about the format should be `i64`.  Is it good to use `u64` to implement cache and avoid to unnecessary casting when request coming.  
That means we do not need to be restricted by datetime far from some OS issues, and less casting in our program.

Besides, other application easy to run into issues by using valid unsign number bigger than `i64` in developing time, and we will raise error to them, and the error possibly ambiguous.

Also, we have a previous discussion about this, just in case you can not find it.
https://github.com/MystenLabs/walrus/pull/1345#discussion_r1906313052

crates/walrus-service/src/client/daemon.rs
```diff
@@ -233,13 +235,15 @@ impl<T: WalrusWriteClient + Send + Sync + 'static> ClientDaemon<T> {
...
+            // Create and run the cache to track the used JWT tokens.
+            let replay_suppression_cache = auth_config.replay_suppression_config.build_and_run();
             self.router = self.router.route(
                 BLOB_PUT_ENDPOINT,
                 put(routes::put_blob)
                     .route_layer(
                         ServiceBuilder::new()
                             .layer(axum::middleware::from_fn_with_state(
-                                Arc::new(auth_config),
+                                (Arc::new(auth_config), Arc::new(replay_suppression_cache)),
```
yanganto, line 246: Is it possible to put `replay_suppression_cache` in side the `auth_config`?  We can rename the `auth_config`, if you feel the variable name is not fit.  Then the state is only one `Arc`, and also the `fn auth_layer` will good to read after this.

giac-mysten: I'll avoid that, because the cache handle is semantically not part of the config. I used a single arc though, to improve readability.

crates/walrus-service/src/client/daemon/auth.rs
giac-mysten: I am not convinced by this argument, for the following reasons:

1. UNIX timestamps are supposed to be signed integers. So maybe we should transition to `i64` everywhere. 
2. there is almost no overhead in casting, as internally Datetime uses `i32` and `u32`s to represent the date. This also means that the performance of comparing datetime values is very close to comparing `i64`s (i've seen a 1ns difference at most).
3. The cache implementation may be used for other tasks, where precision below the second is required. At that point using integers does not work anymore.
4. There is a readability and clarity advantage for the developer, who does not need to remember that the integers are actually timestamps.

For these reasons, I would actually be in favor of having relevant fields on `Claim` be changed to `DateTime`, so that serde takes care of the casting and conversions, and we are always sure to be working with correct timestamps. Thoughts?

yanganto: Yeh, If 2 is acceptable, we change the relevant fields in on `Claim` into `Datetime`, it will be easier to handle it in the cache.  About 4, it is about making a `type Timestamp = u64`, and not a reasonable reason in my view.

Overall, I will not against to use `Datetime`(`i64`) in the future also including the fields of JWT token, if @mlegner also think the `Datetime` will be better here.

PR #1562, opened 2025-02-11
test: nightly simtest setup

halfprice:
## Description

Setup "nightly" walrus simtest.
- Running each simtest with 30 different seed.
- Schedule to run everyday 1pm PST to not conflict with Sui simtest, which is scheduled 1am PST.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/.github/workflows/simtest-nightly.yml
+++ b/.github/workflows/simtest-nightly.yml
@@ -36,7 +36,7 @@ jobs:
 
     steps:
       - name: Install Teleport
-        uses: teleport-actions/setup@v1 # pin@v1.0.6
+        uses: teleport-actions/setup@176c25dfcd19cd31a252f275d579822b243e7b9c # pin@v1.0.6
         with:
           version: 12.1.0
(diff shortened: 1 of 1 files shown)


.github/workflows/simtest-nightly.yml
```diff
@@ -34,9 +36,9 @@ jobs:
 
     steps:
       - name: Install Teleport
-        uses: teleport-actions/setup@176c25dfcd19cd31a252f275d579822b243e7b9c # pin@v1.0.6
+        uses: teleport-actions/setup@v1 # pin@v1.0.6
```
pei-mysten: nit: we've seen some incidents where the updates on the remote side breaking our workflow, so it still be the best to stick with a certain pin

halfprice: Make sense. The change is reverted.

PR #1564, opened 2025-02-11
feat: add dbtool to count total number of certified blobs

halfprice:
## Description

Reading local db and count the number of certified blobs.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/dbtool.rs
+++ b/crates/walrus-service/src/node/dbtool.rs
@@ -13,7 +13,7 @@ use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use sui_types::base_types::ObjectID;
 use typed_store::rocks::be_fix_int_ser;
-use walrus_core::BlobId;
+use walrus_core::{BlobId, Epoch};
 
 use super::events::PositionedStreamEvent;
 use crate::node::{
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/dbtool.rs
```diff
@@ -184,3 +196,44 @@ fn read_object_blob_info(db_path: PathBuf, object_id: ObjectID) -> Result<()> {
...
+    // Scan all the blob info and count the certified ones
+    let iter = db.iterator_cf(&cf, rocksdb::IteratorMode::Start);
+
+    let mut certified_count = 0;
+    let mut scan_count = 0;
+    for blob_info_raw in iter {
+        let (_key, value) = blob_info_raw?;
+        let blob_info: BlobInfo = bcs::from_bytes(&value)?;
+        if blob_info.is_certified(epoch) {
+            certified_count += 1;
```
wbbradley, line 222: Does the blob_info keep track of expired blobs? In other words, should we be filtering out expired or deleted blobs here?

halfprice: Yes, `is_certified` accounts for blob expiration and deletion. These are the true certified blobs.

PR #1566, opened 2025-02-11
chore(walrus-backup): retry writes that fail due to serializable transaction constraints

wbbradley:
## Description

When multiple writers attempt to write simultaneously, it is possible (and expected) for transactions in isolation level SERIALIZABLE to fail. This change ensures that those transactions are retried.

## Test plan

CI, manual testing.

No release impact.

### DIFF
--- a/crates/walrus-service/src/backup/service.rs
+++ b/crates/walrus-service/src/backup/service.rs
@@ -3,10 +3,14 @@
 
 //! Backup service implementation.
 
-use std::{pin::Pin, sync::Arc, time::Duration};
+use std::{panic::Location, pin::Pin, sync::Arc, time::Duration};
 
 use anyhow::{bail, Context, Result};
-use diesel::Connection as _;
+use diesel::{
+    result::{DatabaseErrorKind, Error},
+    sql_types::{Bytea, Int4, Text},
+    Connection as _,
+};
 use diesel_async::{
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/backup/service.rs
```diff
@@ -569,3 +592,43 @@ async fn upload_blob_to_storage(
     );
     Ok(blob_url)
 }
+
+async fn retry_serializable_query<'a, 'b, T, F>(
```
halfprice, line 600: Nice generalization of the retry logic!

crates/walrus-service/src/backup/service.rs
```diff
@@ -338,52 +353,57 @@ async fn backup_take_task(
...
+                                AND blob_state.retry_count < $1
+                            ORDER BY blob_state.initiate_fetch_after ASC
+                            LIMIT 1
+                        ),
+                        _updated_count AS (
+                            UPDATE blob_state
+                            SET
+                                initiate_fetch_after =
+                                    NOW()
+                                    + LEAST(86400, ($2 / 1.5) * POW(1.5, retry_count))
```
halfprice, line 391: Can you add a comment explain the time interval here so that it's easier to read?

wbbradley: Yep, will do!

PR #1572, opened 2025-02-13
feat: implement emergency upgrade integration

karlwuest:
## Description

Integrates the emergency update in the rust code (and functions for the quorum-based upgrade in the transaction builder).

Follow ups:
- integrate quorum-based upgrades
- integration tests for upgrades

## Test plan

- Tested manually in local test cluster.
- Will add integration tests as follow up

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -587,21 +587,6 @@ impl SuiContractClient {
             .await
     }
 
-    /// Executes an upgrade that has been voted for by a quorum of the committee.
-    ///
-    /// Returns the new package ID.
-    pub async fn execute_upgrade(
-        &self,
-        upgrade_manager: ObjectID,
-        package_path: PathBuf,
-    ) -> SuiClientResult<ObjectID> {
-        self.inner
-            .lock()
-            .await
-            .execute_upgrade(upgrade_manager, package_path)
-            .await
-    }
-
     /// Performs an emergency upgrade.
     ///
     /// Returns the new package ID.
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/system_setup.rs
```diff
@@ -96,17 +95,11 @@ pub(crate) async fn publish_package_with_default_build_config(
     publish_package(wallet, package_path, Default::default(), gas_budget).await
 }
 
-#[tracing::instrument(err, skip(wallet, build_config))]
-pub(crate) async fn publish_package(
-    wallet: &mut WalletContext,
+pub(crate) async fn compile_package_and_update_lock_file(
```
mlegner: Question: Where does that update the lock file? The corresponding call still seems to be [below](https://github.com/MystenLabs/walrus/pull/1572/files#diff-d1c133c032d51cbb6145a2e837e4c8c1c7551400d65d9b3d8a6212a5792294ddR195) in the `publish_package` function.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1091,6 +1096,118 @@ impl WalrusPtbBuilder {
         Ok(())
     }
 
+    /// Votes for an upgrade of the system contract from the Node with `node_id`.
+    pub async fn vote_for_upgrade(
+        &mut self,
+        upgrade_manager: ObjectID,
+        node_id: ObjectID,
+        digest: &[u8],
```
mlegner, line 1107: Question: Does the `digest` identify the upgrade?

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1091,6 +1099,118 @@ impl WalrusPtbBuilder {
...
+            )?,
+            self.argument_from_arg_or_obj(emergency_upgrade_cap).await?,
+            self.pt_builder.pure(digest)?,
+        ];
+        self.walrus_move_call(contracts::upgrade::authorize_emergency_upgrade, args)
+    }
+
+    /// Performs a contract upgrade.
+    ///
+    /// Returns the `UpgradeTicket` as result argument.
```
mlegner: Question: Isn't the ticket an input to this function? Do you mean `UpgradeReceipt`?

crates/walrus-sui/src/client.rs
```diff
@@ -1329,26 +1375,133 @@ impl SuiContractClientInner {
         Ok(())
     }
 
+    /// Executes an upgrade that has been voted for by a quorum of the committee.
+    ///
+    /// Returns the new package ID.
+    pub async fn execute_upgrade(
```
mlegner: Question: This assumes that the package has already been compiled, the digest calculated, and nodes have voted for it, correct? Should we check somewhere that the digest computed within this function corresponds to one voted on?

crates/walrus-sui/src/client.rs
```diff
@@ -1329,26 +1375,133 @@ impl SuiContractClientInner {
...
+    }
+
+    /// Performs an emergency upgrade.
+    ///
+    /// Returns the new package ID.
+    pub async fn emergency_upgrade(
+        &mut self,
+        upgrade_manager: ObjectID,
+        package_path: PathBuf,
+    ) -> SuiClientResult<ObjectID> {
```
mlegner, line 1370: Suggestion: AFAICT, this is almost the same as the `execute_upgrade`, except for the way it is authorized. Could those be consolidated?

crates/walrus-sui/src/system_setup.rs
karlwuest: I couldn't come up with a better name 😅. The function does update the lock file, but only temporarily.
The function updates the lock file before compiling the package and then resets it after compiling the package, because for the package graph resolution to work correctly, the package address needs to be set to zero (I don't really know the internals here).

I could just integrate this in the `compile_package` function (since we only need that with the graph resolution anyway). Wdyt?

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: This is the package digest of the upgraded package. In the contracts this identifies the upgrade.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: Yes, good catch!

crates/walrus-sui/src/client.rs
karlwuest: That's a good idea!

crates/walrus-sui/src/system_setup.rs
mlegner: > I could just integrate this in the compile_package function (since we only need that with the graph resolution anyway). Wdyt?

Sounds good. 👍

crates/walrus-sui/src/client.rs
karlwuest: I removed the `execute_upgrade` for now, but will consolidate them in the follow up adding the quorum based upgrades.

crates/walrus-sui/src/client.rs
karlwuest: I will add such a check in the follow-up for the quorum based upgrade.

PR #1573, opened 2025-02-13
feat: update active set size in testnet contracts

karlwuest:
## Description

Updates the testnet contracts.
TODO:

- [x] Merge after contracts are updated on chain with new `Move.lock`

## Test plan

- Tested in local cluster
- Will test on PTN before testnet

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


contracts/subsidies/Move.lock
mlegner, line 1: Question: Is it intentional that the lock files in the `contracts` are updated (as opposed to the `testnet-contracts`)?

karlwuest: No, that was the pre-commit script and I didn't catch that this did not apply to the `testnet-contracts`. Reverted for `contracts` but updated the version for `testnet-contracts/walrus`.

PR #1574, opened 2025-02-13
feat(cli): wal-587 upon dry run include sui gas costs - WIP

Tzal3x:
## Description

Currently the CLI when used with the `--dry-run` flag is used, it provides only FROST cost estimates.

In this PR we include Sui PTB execution gas costs. Example CLI output:
```
2025-02-18T10:02:37.443837Z  INFO reserve_and_store_blobs_retry_committees_with_path: walrus_service::client: storing 1 sliver pairs with metadata
2025-02-18T10:02:38.081764Z  INFO reserve_and_store_blobs_retry_committees_with_path: walrus_service::client: retrieved 1 blob statuses duration=637.876875ms
Dry run: a Sui PTB will be executed that will cost 1988000 MIST
Would you like to continue? (Y/n): 
```

However, there are steps in the process that depend on the PTB execution outputs. 
For example, the PTBs of _reserve_ and _register_ blobs need to happen before the certification PTB. 

## Changelog
`--dry-run` no longer interrupts the execution of the program, instead the store command continues, by prompting the user to either abort or continue after executing the command. 

## Open questions:
1. Do we need user prompts to continue/abort? 
2. Would we prefer calculating the total gas cost for _reserve_ and _register_ and just explain blob certification is not calculated yet?
3. Calculating the sui gas costs upon PTB construction doesn't fit well with the current `--dry-run` implementation because it will abort before reaching the store command:
```rust
if dry_run {
     return Self::store_dry_run(client, files, epochs_ahead, self.json).await;
 }
``` 
4. We should somehow export the estimated gas cost for register and store in order for the site-builder dry run to consume it.
5. Other alternatives?

## Test plan

1. Generate a test file of 6 MB to upload on walrus: `dd if=/dev/zero of=mediumsizefile.bin bs=1M count=6`
5. Upload a file: `cargo run --package walrus-service --bin walrus -- store --dry-run mediumsizefile.bin --epochs 1`
6. For now, the user will be prompted to either continue the file upload or abort before each PTB dry run.

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [X] CLI: TODO

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -24,3 +24,14 @@ retries = 0 # Do not retry these tests as they always succeed on the second run
 [profile.simtest]
 slow-timeout = { period = "5m", terminate-after = 6 } # Timeout tests after 30 minutes
 test-threads = "num-cpus"
+
+[profile.simtestnightly]
+# Print out output for failing tests as soon as they fail, and also at the end
+# of the run (for easy scrollability).
+failure-output = "immediate"
+# Show skipped tests in the CI output.
+status-level = "fail"
+# Do not cancel the test run on the first failure.
+fail-fast = false
+# Mark tests as slow after 30m, terminate after 1.5h
+slow-timeout = { period = "30m", terminate-after = 3 }
--- a/.editorconfig
+++ b/.editorconfig
@@ -25,7 +25,7 @@ max_line_length = 150
 indent_size = unset
 
 # Ignore paths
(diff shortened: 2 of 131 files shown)


mlegner:
@Tzal3x @giac-mysten I'm closing this for now as we haven't found a good way to deal with the issue of having multiple transactions. Let's revisit this in the context of the whole cost-calculator topic.

PR #1575, opened 2025-02-13
fix(publisher): retry if refresh fails

giac-mysten:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/refresh.rs
+++ b/crates/walrus-service/src/client/refresh.rs
@@ -199,11 +199,11 @@ impl<T: ReadClient> CommitteesRefresher<T> {
                         ?timer_interval,
                         "auto-refreshing the active committee"
                     );
-                    self.refresh().await.unwrap_or_else(|error| {
+                    let _ = self.refresh().await.inspect_err(|error| {
                         tracing::error!(
-                            "failed to refresh the active committee: {:?}; \
+                            %error,
+                            "failed to refresh the active committee; \
                             retrying again at the next interval",
-                            error
                         )
                     });
                 }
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client/refresh.rs
```diff
@@ -201,36 +199,47 @@ impl<T: ReadClient> CommitteesRefresher<T> {
                         ?timer_interval,
                         "auto-refreshing the active committee"
                     );
-                    self.refresh().await?;
+                    self.refresh().await.unwrap_or_else(|error| {
```
karlwuest: Nit: 
```suggestion
                    self.refresh().await.inspect_err(|error| {
```

crates/walrus-service/src/client/refresh.rs
```diff
@@ -201,36 +199,47 @@ impl<T: ReadClient> CommitteesRefresher<T> {
...
-                        .inspect_err(|_| {
-                            // This may happen because the client was notified of a committee
-                            // change, and therefore the receiver end of the channel was dropped.
-                            tracing::info!("failed to send the committee and price")
-                        });
+                        tracing::trace!(
+                            "received a request"
+                        );
+                        if request.is_refresh() {
+                            self.refresh_if_stale().await.unwrap_or_else(|error| {
```
karlwuest: Nit:
```suggestion
                            self.refresh_if_stale().await.inspect_err(|error| {
```

crates/walrus-service/src/client/refresh.rs
giac-mysten: The problem with this is that `inspect_err` does not consume the error. 
Therefore, it's either `unwrap_or_else` or `let _ = self.refres().await.inspect_err(|...`.
preferences?

karlwuest: Ah right. I have a slight preference for the latter, but no strong feelings.

PR #1579, opened 2025-02-14
fix: Ensure consistency between BlobInfo::to_blob_status() and is_registered()

liquid-helium:
## Description

Ensure consistency between `BlobInfo::to_blob_status()` and `is_registered()`.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1979,9 +1979,6 @@ impl ServiceState for StorageNodeInner {
         sliver_pair_index: SliverPairIndex,
         sliver_type: SliverType,
     ) -> Result<Sliver, RetrieveSliverError> {
-        if let Ok(BlobStatus::Nonexistent) = self.blob_status(blob_id) {
-            return Err(RetrieveSliverError::Unavailable);
-        }
         self.check_index(sliver_pair_index)?;
 
         ensure!(!self.is_blocked(blob_id), RetrieveSliverError::Forbidden);
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -960,6 +960,7 @@ impl BlobInfoApi for BlobInfoV1 {
     fn is_registered(&self, current_epoch: Epoch) -> bool {
         let Self::Valid(ValidBlobInfoV1 {
             count_deletable_total,
+            count_deletable_certified,
             permanent_total,
             latest_seen_deletable_registered_epoch,
             permanent_certified,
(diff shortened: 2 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1979,6 +1979,9 @@ impl ServiceState for StorageNodeInner {
         sliver_pair_index: SliverPairIndex,
         sliver_type: SliverType,
     ) -> Result<Sliver, RetrieveSliverError> {
+        if let Ok(BlobStatus::Nonexistent) = self.blob_status(blob_id) {
```
halfprice: Why `self.is_blob_registered(blob_id)?` check 6 lines below not covering this case? What's the test symptom?

mlegner: This should indeed be covered already... 🤔

liquid-helium: Here is one example of the failed test: https://github.com/MystenLabs/walrus/actions/runs/13323343454/job/37211768730?pr=1548

I didn't spend a lot of time on it, but it appears that the blob was deleted, and we can retrieve the Nonexistent status from the blob_status() call. However, the client still succeeds in reading the sliver, which suggests there may be a race condition or some other glitches occurring.

halfprice: Thanks @liquid-helium for the example! I think if `is_blob_registered` returns true, here the blob status also won't be `nonexistent`.

I couldn't find a local seed that reproduces the error, but I think what could happen was that N node processed the blob deletion event, where f+1 <= N <= 2f. This is [enough for the blob status to return non-existance](https://github.com/MystenLabs/walrus/blob/59a6ab04bb476d003ec21ddd8cdfb98607390424/crates/walrus-service/src/client.rs#L1583-L1593). However, when doing the blob read, we use primary sliver to read, which only requires reading f+1, and then decode. We may still have f+1 nodes that hasn't processed blob deletion yet, so the read succeeded.

I think the fix of the test can simply be using the `secondary` sliver to read the blob [here](https://github.com/MystenLabs/walrus/blob/a649efbf1ee43ee99f9af78aa79f66632b2d5e2c/crates/walrus-e2e-tests/tests/test_client.rs#L593).

liquid-helium: I think I find a subtle difference between is_blob_registered and how BlobInfo calculate Nonexistent state. There are comments about temporary workaround, we need to do a cleanup, and make the info cnsistent.

mlegner: Thanks a lot, @liquid-helium! The cleanup you mention has been done in parallel in #1581. 🙂

liquid-helium: That's even better! Thanks @mlegner. I will close this PR.

PR #1580, opened 2025-02-14
feat: Integrate subsidies into walrus system

sadhansood:
## Description

Closes WAL-438.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


contracts/subsidies/Move.lock
```diff
@@ -5,51 +5,51 @@ version = 3
...
-  { id = "WAL", name = "WAL" },
-  { id = "WAL_exchange", name = "WAL_exchange" },
+  { name = "Sui" },
+  { name = "WAL" },
+  { name = "WAL_exchange" },
 ]
 
 [move.toolchain-version]
-compiler-version = "1.42.1"
+compiler-version = "1.33.0"
```
karlwuest: Nit: You should update your local sui client 😄

contracts/subsidies/sources/subsidies.move
```diff
@@ -45,6 +49,8 @@ public struct Subsidies has key, store {
     system_subsidy_rate: u16,
     /// The balance of funds available in the subsidy pool.
     subsidy_pool: Balance<WAL>,
+    /// The version of the subsidies contract.
+    version: u64,
```
karlwuest, line 55: Important: More important than having the version check would be storing the package address in the Subsidies object to allow users to automatically fetch it from the object. This allows users to just configure the object ID of the Subsidies object and any contract updates (which need to be done every time the walrus contracts are upgraded) will then be handled automatically. Otherwise, users may still have an old version of the contract configured which will then fail for all calls.

contracts/wal/Move.lock
karlwuest, line 1: Important: Replace the lock files with files generated by a new compiler version.

crates/walrus-orchestrator/src/protocol/target.rs
```diff
@@ -169,6 +169,7 @@ impl ProtocolCommands for TargetProtocol {
             admin_wallet_path: None,
             do_not_copy_contracts: false,
             with_wal_exchange: true,
+            with_subsidies: true,
```
karlwuest: Nit: I think in the orchestrator we don't need the subsidies contracts.

crates/walrus-service/node_config_example.yaml
```diff
@@ -56,6 +56,7 @@ sui:
   rpc: https://fullnode.testnet.sui.io:443
   system_object: 0xa2637d13d171b278eadfa8a3fbe8379b5e471e1f3739092e5243da17fc8090eb
   staking_object: 0xca7cf321e47a1fc9bfd032abc31b253f5063521fd5b4c431f2cdd3fee1b4ec00
+  subsidies_object: 0xa9b00f69d3b033e7b64acff2672b54fbb7c31361954251e235395dea8bd6dcac
```
karlwuest: Nit: The nodes should never need the subsidies object, so we probably don't want to add it in the example config.

crates/walrus-service/src/client/config.rs
```diff
@@ -618,11 +620,28 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn parses_no_subsidies_object_config_file() -> TestResult {
+        let yaml = indoc! {"
+            system_object: 0xa2637d13d171b278eadfa8a3fbe8379b5e471e1f3739092e5243da17fc8090eb
+            staking_object: 0xca7cf321e47a1fc9bfd032abc31b253f5063521fd5b4c431f2cdd3fee1b4ec00
+            subsidies_object: []
```
karlwuest:
```suggestion
```

crates/walrus-service/src/node/config.rs
```diff
@@ -888,6 +888,7 @@ mod tests {
         let contract_config = ContractConfig {
             system_object: ObjectID::random_from_rng(&mut rng),
             staking_object: ObjectID::random_from_rng(&mut rng),
+            subsidies_object: Some(ObjectID::random_from_rng(&mut rng)),
```
karlwuest:
```suggestion
            subsidies_object: None,
```

crates/walrus-sui/src/client.rs
```diff
@@ -641,6 +644,27 @@ impl SuiContractClient {
             .await
     }
 
+    /// Creates a new [`contracts::wal_exchange::Exchange`] with a 1:1 exchange rate, funds it with
+    /// `amount` FROST, and returns its object ID.
```
karlwuest: Suggestion: Fix comment

crates/walrus-sui/src/client.rs
```diff
@@ -1473,6 +1497,46 @@ impl SuiContractClientInner {
         Ok(exchange_id[0])
     }
 
+    /// Creates a new [`contracts::wal_subsidies::Subsidies`],
+    /// funds it with `amount` FROST, and returns its object ID.
```
karlwuest:
```suggestion
    /// Creates a new [`contracts::wal_subsidies::Subsidies`] object, funds it with
    /// `amount` FROST, and returns its object ID, as well as the object ID of its admin cap.
```

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -254,12 +259,15 @@ impl From<Mutability> for bool {
 #[derive(Clone)]
 pub struct SuiReadClient {
     walrus_package_id: Arc<RwLock<ObjectID>>,
+    subsidies_package_id: Arc<RwLock<Option<ObjectID>>>,
```
karlwuest: Suggestion: Package all subsidy related fields into a separate struct, then we can have a single `Option` field here.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -959,6 +1025,18 @@ impl ReadClient for SuiReadClient {
...
     }
+
+    async fn refresh_subsidies_package_id(&self) -> SuiClientResult<()> {
+        let Some(subsidies_object_id) = self.subsidies_object_id else {
+            return Ok(());
+        };
+        let subsidies_package_id = self
+            .sui_client
+            .get_package_id_from_object(subsidies_object_id)
+            .await?;
```
karlwuest: Important: This will always return the same package id for the same object. Namely, this returns the ID of the package at the version in which the struct of the object is defined.

crates/walrus-sui/src/client.rs
```diff
@@ -1941,6 +2005,10 @@ impl ReadClient for SuiContractClient {
     async fn refresh_package_id(&self) -> SuiClientResult<()> {
         self.read_client.refresh_package_id().await
     }
+
+    async fn refresh_subsidies_package_id(&self) -> SuiClientResult<()> {
```
karlwuest, line 2288: Question: Is this called anywhere?

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -275,6 +291,37 @@ impl WalrusPtbBuilder {
...
+            self.pt_builder.pure(epochs_ahead)?,
+            self.wal_coin_arg()?,
+        ];
+        let result_arg = self.move_call(
+            subsidies_package_id,
+            contracts::subsidies::reserve_space,
+            reserve_arguments,
+        )?;
+        // TODO(WAL-512): fix this once we have a way to get the subsidy amount
+        self.reduce_wal_balance(price)?;
```
karlwuest: Remark: reducing the balance by the full amount here is correct. The subsidy contract requires the full payment and then just "refunds" the subsidy. In the case where the subsidy contract runs out of money, the transaction will still go through but without subsidies.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -251,11 +251,27 @@ impl WalrusPtbBuilder {
...
+            self.reserve_space_with_subsidies(encoded_size, epochs_ahead)
+                .await
+        } else {
+            self.reserve_space_without_subsidies(encoded_size, epochs_ahead)
+                .await
+        }
+    }
+
+    /// Adds a call to `reserve_space` to the `pt_builder` and returns the result [`Argument`].
+    pub async fn reserve_space_without_subsidies(
```
karlwuest: Suggestion: I think we should not publicly expose these functions and only have `reserve_space` and `extend_blob` as public interfaces. 

Alternatively, we expose the functions as `reserve_space`/`reserve_space_with_subsidies` (and equivalent for extend), such that `reserve_space` just calls the `reserve_space` in walrus without any subsidies and the case distinction for which function to call is made in the contract client. This option may be slightly cleaner, because it preserves the mapping of a function call here to a function call in the contracts better, but I don't have a strong preference between the two options.

crates/walrus-sui/src/client.rs
sadhansood: Yes, now wpon wrong version error

contracts/subsidies/sources/subsidies.move
```diff
@@ -257,16 +292,20 @@ public fun new_with_initial_rates_and_funds_for_testing(
...
     assert!(initial_buyer_subsidy_rate <= MAX_SUBSIDY_RATE, EInvalidSubsidyRate);
     assert!(initial_system_subsidy_rate <= MAX_SUBSIDY_RATE, EInvalidSubsidyRate);
+    let package_id = object::new(ctx);
     let subsidies = Subsidies {
         id: object::new(ctx),
         buyer_subsidy_rate: initial_buyer_subsidy_rate,
         system_subsidy_rate: initial_system_subsidy_rate,
         subsidy_pool: initial_funds.into_balance(),
+        version: VERSION,
+        package_id: object::uid_to_inner(&package_id),
```
karlwuest: Nit:
```suggestion
        package_id: package_id.to_inner(),
```

contracts/subsidies/sources/subsidies.move
```diff
@@ -235,16 +266,20 @@ public fun get_subsidy_pool(self: &Subsidies): &Balance<WAL> {
 
 #[test_only]
 public fun new_for_testing(ctx: &mut TxContext): (Subsidies, AdminCap) {
+    let package_id = object::new(ctx);
     let subsidies = Subsidies {
         id: object::new(ctx),
         buyer_subsidy_rate: 0,
         system_subsidy_rate: 0,
         subsidy_pool: balance::zero(),
+        package_id: object::uid_to_inner(&package_id),
```
karlwuest: Nit:
```suggestion
        package_id: package_id.to_inner(),
```

contracts/subsidies/tests/subsidies_tests.move
```diff
@@ -40,24 +43,27 @@ fun test_new() {
...
     let mut test = test::begin(user);
+    let ctx = test.ctx();
 
     let initial_buyer_subsidy_rate: u16 = 5_00; // 5%
     let initial_storage_node_subsidy_rate: u16 = 10_00; // 10%
     let initial_funds_value = 1_000_000;
+    let package_id = object::new(ctx);
 
     let admin_cap = subsidies::new_with_initial_rates_and_funds(
+        object::uid_to_inner(&package_id),
```
karlwuest: Nit:
```suggestion
        package_id: package_id.to_inner(),
```

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -854,6 +854,68 @@ async fn test_blocklist() -> TestResult {
...
+            PostStoreAction::Keep,
+        )
+        .await?;
+
+    let blob_object = match &store_result[0] {
+        BlobStoreResult::NewlyCreated { blob_object, .. } => blob_object.clone(),
+        _ => panic!("Expected newly created blob"),
+    };
+
+    // Verify initial storage cost
```
karlwuest: Nit: This comment doesn't match the code. There is no storage cost check.

crates/walrus-service/src/test_utils.rs
```diff
@@ -2310,6 +2316,26 @@ pub mod test_cluster {
...
+                    )
+                    .await?;
+                Some(ClientSubsidies::new(subsidies_pkg_id, subsidies_id))
+            }
+            None => None,
+        };
+
+        if let Some(subsidies) = subsidies {
+            *admin_contract_client.inner.read_client().subsidies_mut() = Some(subsidies);
+        }
```
karlwuest: Nit: This could be part of the first case in the match statement above.

crates/walrus-sui/src/client.rs
```diff
@@ -616,23 +661,29 @@ impl SuiContractClient {
...
     ///
     /// This must be called in the new package after an upgrade is committed in a separate
     /// transaction.
     pub async fn migrate_contracts(&self, new_package_id: ObjectID) -> SuiClientResult<()> {
-        self.inner
-            .lock()
-            .await
-            .migrate_contracts(new_package_id)
-            .await
+        self.retry_on_wrong_version(|| async {
```
karlwuest: Important: When we migrate the contracts, there is no point retrying, because this always calls the `migrate` function on `new_package_id`.

crates/walrus-sui/src/client.rs
```diff
@@ -642,11 +693,14 @@ impl SuiContractClient {
         exchange_package: ObjectID,
         amount: u64,
     ) -> SuiClientResult<ObjectID> {
-        self.inner
-            .lock()
-            .await
-            .create_and_fund_exchange(exchange_package, amount)
-            .await
+        self.retry_on_wrong_version(|| async {
```
karlwuest: Important: Similar to above, there is no point retrying here, since we always use `exchange_package` to call the functions.

crates/walrus-sui/src/client.rs
```diff
@@ -655,11 +709,39 @@ impl SuiContractClient {
         exchange_id: ObjectID,
         amount: u64,
     ) -> SuiClientResult<()> {
-        self.inner
-            .lock()
-            .await
-            .exchange_sui_for_wal(exchange_id, amount)
-            .await
+        self.retry_on_wrong_version(|| async {
```
karlwuest: Important: No point retrying, we should never get a wrong version error from the exchange, because the exchange contract is not versioned.

crates/walrus-sui/src/client.rs
```diff
@@ -964,11 +1086,48 @@ impl SuiContractClient {
...
+                .certify_and_extend_blobs(blobs_with_certificates, post_store)
+                .await
+        })
+        .await
+    }
+}
+
+/// Trait for retrying operations that may fail due to wrong package versions
+#[async_trait]
+trait RetryOnWrongVersion {
```
karlwuest: Question: Why do we need a trait for this?

Nit: `async_trait` is not necessary anymore for `async` functions in traits, see e.g. the `ReadClient` trait and its implementations.

crates/walrus-sui/src/client.rs
```diff
@@ -964,11 +1086,48 @@ impl SuiContractClient {
...
+    {
+        match f().await {
+            Err(SuiClientError::TransactionExecutionError(
+                MoveExecutionError::Subsidies(SubsidiesError::EWrongVersion(_))
+                | MoveExecutionError::Staking(StakingError::EWrongVersion(_))
+                | MoveExecutionError::System(SystemError::EWrongVersion(_)),
+            )) => {
+                self.read_client.refresh_package_id().await?;
+                self.read_client.refresh_subsidies_package_id().await?;
+                f().await
```
karlwuest: Suggestion: Check if the version actually changed after a refresh before retrying.

crates/walrus-sui/src/client.rs
```diff
@@ -1145,12 +1312,19 @@ impl SuiContractClientInner {
             "starting to reserve and register blobs"
         );
 
+        // Check if subsidies are available
```
karlwuest: Nit: The comment doesn't match the code, there is no check here.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -256,6 +256,25 @@ impl WalrusPtbBuilder {
         &mut self,
         encoded_size: u64,
         epochs_ahead: EpochCount,
+        subsidies_package_id: Option<ObjectID>,
```
karlwuest: Question: Why do you pass the `subsidies_package_id` here instead of getting it from `self.read_client` ?

(Also applies to extend)

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -275,6 +294,35 @@ impl WalrusPtbBuilder {
...
+            self.pt_builder.pure(encoded_size)?,
+            self.pt_builder.pure(epochs_ahead)?,
+            self.wal_coin_arg()?,
+        ];
+        let result_arg = self.move_call(
+            subsidies_package_id,
+            contracts::subsidies::reserve_space,
+            reserve_arguments,
+        )?;
+        // TODO(WAL-512): fix this once we have a way to get the subsidy amount
```
karlwuest:
```suggestion
```

karlwuest: See comment from last week below.

crates/walrus-sui/src/test_utils/system_setup.rs
```diff
@@ -153,6 +161,7 @@ pub async fn create_and_init_system_for_test(
         deploy_directory,
         true,
         false,
+        true,
```
karlwuest: Important: This makes all tests use subsidies, i.e. we don't have any e2e tests covering the case where no subsidy object is configured. Can you make sure that we have e2e tests with and without subsidies?

crates/walrus-sui/src/client/transaction_builder.rs
sadhansood: good call, no need to do it

contracts/subsidies/sources/subsidies.move
```diff
@@ -202,6 +227,12 @@ public fun reserve_space(
     storage
 }
 
+entry fun migrate(subsidies: &mut Subsidies, admin_cap: &AdminCap) {
+    check_admin(subsidies, admin_cap);
+    check_version_upgrade(subsidies);
+    subsidies.version = VERSION;
+}
```
karlwuest: Suggestion: Set the new package ID here, then it doesn't require a separate call.

crates/walrus-sui/src/client/contract_config.rs
```diff
@@ -13,15 +13,29 @@ pub struct ContractConfig {
     pub system_object: ObjectID,
     /// Object ID of the Walrus staking object.
     pub staking_object: ObjectID,
-    // TODO(WAL-438): Add package and object for subsidies
+    /// Object ID of the Walrus subsidies object.
+    pub subsidies_object: Option<ObjectID>,
```
karlwuest, line 18:
```suggestion
    #[serde(default, skip_serializing_if = "Option::is_none")]
    pub subsidies_object: Option<ObjectID>,
```

crates/walrus-service/node_config_example.yaml
```diff
@@ -56,6 +56,7 @@ sui:
   rpc: https://fullnode.testnet.sui.io:443
   system_object: 0xa2637d13d171b278eadfa8a3fbe8379b5e471e1f3739092e5243da17fc8090eb
   staking_object: 0xca7cf321e47a1fc9bfd032abc31b253f5063521fd5b4c431f2cdd3fee1b4ec00
+  subsidies_object: null
```
karlwuest: Suggestion: We should skip serialization here, see suggestion on the struct definition of the contract config.

crates/walrus-sui/src/client.rs
```diff
@@ -662,6 +709,31 @@ impl SuiContractClient {
...
+    /// funds it with the specified amount,
+    /// and returns the object ID and the admin cap ID.
+    pub async fn create_and_fund_subsidies(
+        &self,
+        subsidies_package: ObjectID,
+        initial_buyer_subsidy_rate: u16,
+        initial_system_subsidy_rate: u16,
+        amount: u64,
+    ) -> SuiClientResult<(ObjectID, ObjectID)> {
+        self.retry_on_wrong_version(|| async {
```
karlwuest: Suggestion: We don't need the retry here, because the package ID is passed as an argument and therefore cannot be refreshed and we don't have a subsidy object yet to check the version of.

crates/walrus-sui/src/client.rs
```diff
@@ -724,29 +796,42 @@ impl SuiContractClient {
...
+            self.inner.lock().await.delete_blob(blob_object_id).await
+        })
+        .await
     }
 
     /// Merges the WAL and SUI coins owned by the wallet of the contract client.
     pub async fn merge_coins(&self) -> SuiClientResult<()> {
-        self.inner.lock().await.merge_coins().await
+        self.retry_on_wrong_version(|| async { self.inner.lock().await.merge_coins().await })
```
karlwuest: Suggestion: This doesn't go through our contracts, no point retrying.

crates/walrus-sui/src/client.rs
```diff
@@ -724,29 +796,42 @@ impl SuiContractClient {
...
+            .await
     }
 
     /// Sends the `amount` gas to the provided `address`.
     pub async fn send_sui(&self, amount: u64, address: SuiAddress) -> SuiClientResult<()> {
-        self.inner.lock().await.send_sui(amount, address).await
+        self.retry_on_wrong_version(|| async {
```
karlwuest: Suggestion: Remove retry, this doesn't use our contracts.

crates/walrus-sui/src/client.rs
```diff
@@ -724,29 +796,42 @@ impl SuiContractClient {
...
+            self.inner.lock().await.send_sui(amount, address).await
+        })
+        .await
     }
 
     /// Sends the `amount` WAL to the provided `address`.
     pub async fn send_wal(&self, amount: u64, address: SuiAddress) -> SuiClientResult<()> {
-        self.inner.lock().await.send_wal(amount, address).await
+        self.retry_on_wrong_version(|| async {
```
karlwuest: Suggestion: Remove retry, this doesn't use our contracts.

PR #1582, opened 2025-02-14
feat: cancel metadata sync if blob becomes expired/deleted/invalidated

halfprice:
## Description

This PR tackles blob metadata sync cancellation using a more general approach, where we create a node wide
blob retirement notifier, and use it to communicate blob retirement.

Any tasks whose execution are affected by blob retirement can register a notification, and when blob is expired,
deleted, or invalidated, the notification will get notified, and the tasks can progress as blob is retired.

This is currently used only in shard recovery during epoch change shard sync. We will update how events cancel
blob sync using the same notifier in a follow up PR.

## Test plan

Unit test for BlobRetirementNotifier
Integration test to test blob expiration/deletion/invalidation.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -4116,7 +4116,7 @@ mod tests {
         G: FnMut(usize) -> Epoch,
         H: FnMut(usize) -> bool,
     {
-        let blobs: Vec<[u8; 32]> = (1..(blob_count + 1)).map(|i| [i; 32]).collect();
+        let blobs: Vec<[u8; 32]> = (1..=blob_count).map(|i| [i; 32]).collect();
         let blobs: Vec<_> = blobs.iter().map(|b| &b[..]).collect();
         let (cluster, blob_details, event_senders) =
             cluster_with_partially_stored_blobs_in_shard_0(
--- a/crates/walrus-service/src/node/blob_retirement_notifier.rs
+++ b/crates/walrus-service/src/node/blob_retirement_notifier.rs
@@ -138,7 +138,8 @@ impl Drop for BlobRetirementNotify {
             *ref_count
         };
 
-        // If ref count is 0, remove from BlobRetirementNotifier.
+        // If ref count is 1, remove from BlobRetirementNotifier. The only ref is the one stored in
+        // the HashMap.
         // Note that the blob may have already being dropped after notify_waiters is called.
         if new_ref_count == 1 {
             self.node_wide_blob_retirement_notifier


crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -997,6 +997,7 @@ impl ShardStorage {
         .inc();
 
         // Create a periodic check future which checks if the blob is still certified.
+        // TODO(zhewu): use blob_expiration_notifier to check if the blob is expired.
```
mlegner: Question: Are you planning to do that in this PR?

crates/walrus-service/src/node/blob_sync.rs
mlegner, line 1: Question: How do we feel about the `cancel_all_expired_syncs_and_mark_events_completed` function in general? Should we create a follow-up issue to change how we cancel syncs for expired blobs?

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -195,7 +195,12 @@ impl BlobSyncHandler {
                 self.node
                     .is_blob_certified(blob_id)
                     .is_ok_and(Not::not)
-                    .then(|| sync.cancel())
+                    .then(|| {
+                        self.node
+                            .blob_expiration_notifier
+                            .notify_blob_expiration(blob_id);
+                        sync.cancel()
+                    })
```
mlegner: Important: AFAIU, this only creates notifications for blobs that have ongoing syncs. This doesn't help with all other blobs that expire.

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -175,18 +175,33 @@ impl ShardSyncHandler {
...
+
+        // Check blob is certified must be after acquiring the notify handle.
+        if !node.is_blob_certified(&blob_id)? {
+            node.metrics.sync_blob_metadata_skipped.inc();
+            return Ok(());
+        }
+
+        tokio::select! {
+            _ = notified => {
+                tracing::info!("blob expired; skipping sync");
```
mlegner: Nit: Strictly speaking, it could also be marked as invalid or deleted. Maybe "blob is no longer live"?

crates/walrus-service/src/node/blob_sync.rs
mlegner: How about we do it in a different way: We also acquire `BlobExpirationNotify` objects in the blob sync, and replace this function by one that iterates over all entries in the `BlobExpirationNotifier`?

crates/walrus-service/src/node/storage/shard.rs
halfprice: Yes. It is done now.

This function now becomes very long, and I'll put some time to clean it up (put different steps in a sub function). I added a TODO, and may also be able to get it done in this PR (but prefer you can review this PR sooner)

crates/walrus-service/src/node/blob_sync.rs
halfprice: Great question!

The main reason I created BlobRetirementNotifier is actually want to use it in `cancel_all_expired_syncs_and_mark_events_completed` case. If you remember, `cancel_all_expired_syncs_and_mark_events_completed` may get stuck when there are a large number of blob sync, which further blocks event processing. By using the notifier, the notification is async so the `notify()` does not wait for the `notified().await` to start running, which will make `cancel_all_expired_syncs_and_mark_events_completed` non-blocking, which is much more desirable.

I didn't get this done in this PR because there appears to be quite a few tests need to be updated when switching to the BlobRetirementNotifier. So I think it's safer to do it in a separate PR.

crates/walrus-service/src/node/blob_sync.rs
halfprice: > Important: AFAIU, this only creates notifications for blobs that have ongoing syncs. This doesn't help with all other blobs that expire.

Yes, you are absolutely right! I was rushing sending out the PR so was just thinking that we need to put a notify expiration in all the places where we do blob sync cancellation. Here, we need to call notify in the place where the cancellation is invoked, not in side the cancellation...

This bug has been fixed now.

>How about we do it in a different way: We also acquire BlobExpirationNotify objects in the blob sync, and replace this function by one that iterates over all entries in the BlobExpirationNotifier?

This is related to your other question about `cancel_all_expired_syncs_and_mark_events_completed`, and is what I plan to do. But prefer to do it in a separate PR.

crates/walrus-service/src/node/shard_sync.rs
halfprice: Agreed. I now use `retire` throughout the PR to indicate `expire/delete/invalidate`. WDYT?

crates/walrus-service/src/node.rs
```diff
@@ -4903,6 +4855,217 @@ mod tests {
...
+        // Although both tests can run under `cargo nextest`, `check_certification_during_recovery`
+        // only works when running in simtest, since it uses failpoints to skip initial blob
+        // certification check.
+        simtest_param_test! {
+            shard_recovery_blob_not_recover_expired_invalid_deleted_blobs -> TestResult: [
+                check_certification_at_beginning: (false),
+                check_certification_during_recovery: (true),
+            ]
+        }
+        async fn shard_recovery_blob_not_recover_expired_invalid_deleted_blobs(
```
halfprice, line 4820: Note: you can skip review this test. I simply copied it from its old location without changing anything. I want to put cancellation related tests together, and after all the happy path tests.

crates/walrus-service/src/node/storage.rs
```diff
@@ -621,9 +621,17 @@ impl Storage {
     /// Clears the metadata in the storage for testing purposes.
     #[cfg(test)]
     pub fn clear_metadata_in_test(&self) -> Result<(), TypedStoreError> {
-        self.metadata.schedule_delete_all()?;
-        self.metadata
-            .compact_range(&BlobId([0; 32]), &BlobId([255; 32]))?;
```
halfprice, line 626: I found that these is not enough to make sure the blob id is removed from the cf. Need to investigate why, but switch to point delete for now. cc @sadhansood

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -977,12 +977,27 @@ impl ShardStorage {
...
+
+            tokio::select! {
+                _ = notified => {
+                    self.skip_recover_blob(blob_id, sliver_type, &node)?;
+                    return Ok(());
+                }
+                result = node.get_or_recover_blob_metadata(&blob_id, epoch) => {
+                    result?
+                }
+            }
```
mlegner, line 1000: Remark: I see this pattern in several different places. Any chance to factor this out into a separate function?

crates/walrus-service/src/node.rs
```diff
@@ -1204,6 +1222,12 @@ impl StorageNode {
...
 
+        // For blobs that are expired in the new epoch, sends a notification to all the tasks
+        // that may be affected by the blob expiration.
+        self.inner
+            .blob_retirement_notifier
+            .epoch_change_notify_all_pending_blob_retirement(self.inner.clone())?;
+
         // Cancel all blob syncs for blobs that are expired in the *current epoch*.
         self.blob_sync_handler
             .cancel_all_expired_syncs_and_mark_events_completed()
```
mlegner, line 1233: Question: This won't be needed when you modify the blob sync to also use the retirement notifier, correct?

crates/walrus-service/src/node.rs
```diff
@@ -4133,7 +4170,7 @@ mod tests {
         G: FnMut(usize) -> Epoch,
         H: FnMut(usize) -> bool,
     {
-        let blobs: Vec<[u8; 32]> = (1..24).map(|i| [i; 32]).collect();
+        let blobs: Vec<[u8; 32]> = (1..(blob_count + 1)).map(|i| [i; 32]).collect();
```
mlegner: Nit: Instead of the `+1`, you can also use the `..=` operator.
```suggestion
        let blobs: Vec<[u8; 32]> = (1..=blob_count).map(|i| [i; 32]).collect();
```

crates/walrus-service/src/node/blob_retirement_notifier.rs
```diff
@@ -0,0 +1,277 @@
...
+                self.blob_id,
+                *ref_count
+            );
+            *ref_count = ref_count.checked_sub(1).unwrap_or(0);
+            *ref_count
+        };
+
+        // If ref count is 0, remove from BlobRetirementNotifier.
+        // Note that the blob may have already being dropped after notify_waiters is called.
+        if new_ref_count == 1 {
```
mlegner: Nit: Maybe adjust the comment as the 0 vs. 1 could be confusing. (AFAIU, the check is `== 1` because there is always `self`.)

crates/walrus-service/src/node/blob_retirement_notifier.rs
```diff
@@ -0,0 +1,277 @@
...
+        let mut registered_blobs = self.registered_blobs.lock().unwrap();
+        for (blob_id, notify) in registered_blobs.iter_mut() {
+            if !node.is_blob_certified(blob_id)? {
+                tracing::debug!(%blob_id, "epoch change notify blob retirement");
+                notify.notify_waiters();
+            }
+        }
+
+        Ok(())
+    }
```
mlegner, line 70: Question: This does something similar as the `cancel_all_expired_syncs_and_mark_events_completed` on the blob-sync handler. Is the idea that the notification is so fast that it doesn't suffer from the same starvation issues?

crates/walrus-service/src/node/shard_sync.rs
mlegner: Sounds great! 💯

crates/walrus-service/src/node.rs
halfprice: Yes, that's right.

crates/walrus-service/src/node.rs

crates/walrus-service/src/node/blob_retirement_notifier.rs
halfprice: It may still suffer from the same starvation issue, but it is not blocking `EpochChangeStart` processing anymore.

Right now, we join on cancellation in cancel_all_expired_syncs_and_mark_events_completed(), which waits for the tasks to actually get cancelled. When switching to notification, the notify() returns immediate, even though the notifier may still take a long time to be notified due the tokio runtime saturation. But at least in this case, the `EpochChangeStart` event won't be blocked.

crates/walrus-service/src/node/blob_retirement_notifier.rs
halfprice: Done

crates/walrus-service/src/node/storage/shard.rs
halfprice: Don't have time to go through this. Will create a followup PR.

PR #1583, opened 2025-02-14
fix(client): reimplement rust SDK functions with retries

giac-mysten:
## Description

Unretriable failures occurred in the client when calling `select_coins`, because of the internal implementation in the rust SDK. This PR fixes the problem by reimplementing the function with retries.

## Test plan

manual tests. The failures have disappeared under load.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -285,7 +285,9 @@ impl RetriableSuiClient {
                             .await
                     })
                     .await
-                    .inspect_err(|error| tracing::warn!(%error, "failed to get coins after retries"))
+                    .inspect_err(
+                        |error| tracing::warn!(%error, "failed to get coins after retries"),
+                    )
                     .ok()?;
 
                     let mut data = page.data;


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -220,14 +222,92 @@ impl RetriableSuiClient {
...
+                            .await
+                    })
+                    .await
+                    {
+                        Ok(page) => page,
+                        Err(error) => {
+                            tracing::warn!(%error, "failed to get coins after retries");
+                            return None;
+                        }
+                    };
```
karlwuest: Question/Nit: Could this be shortened to the following?
```suggestion
                    let page = retry_rpc_errors(self.get_strategy(), || async {
                        self.sui_client
                            .coin_read_api()
                            .get_coins(owner, coin_type.clone(), cursor, Some(100))
                            .await
                    })
                    .await
                    .inspect_err(|error| tracing::warn!(%error, "failed to get coins after retries"))
                    .ok()?;
```

giac-mysten: think so, nice one!

PR #1584, opened 2025-02-14
fix: adress audit results

karlwuest:
## Description

- uses the new epochs `used_capacity_size` in `advance_epoch` when setting the `total_capacity` if `total_capacity` is exceeded
- introduces length limit for node names and network addresses
- checks that commission rate is within bounds
- adds an explicit sanity check that `weight != 0` before dividing by `weight` in `select_committee_and_calculate_votes` and ensures that the result of the apportionment has no assignment with weight 0
- refactor of `quorum_above` and `quorum_below` to improve readability and clarify intentions

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/staking_pool.move
+++ b/contracts/walrus/sources/staking/staking_pool.move
@@ -419,7 +419,9 @@ public(package) fun advance_epoch(
 
     // split the commission from the rewards
     let total_rewards = rewards.value();
-    let commission = rewards.split(total_rewards * (pool.commission_rate as u64) / (N_BASIS_POINTS as u64));
+    let commission = rewards.split(
+        total_rewards * (pool.commission_rate as u64) / (N_BASIS_POINTS as u64),
+    );
     pool.commission.join(commission);
 
     // add rewards to the pool and update the `wal_balance`


contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -18,6 +18,12 @@ use walrus::{
     walrus_context::WalrusContext
 };
 
+// Limit name length to 100 characters.
+const MAX_NAME_LENGTH: u64 = 100;
```
mlegner: Suggestion: Can we add a check in the Rust code to provide a more useful error in case this is exceeded?

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -185,6 +195,15 @@ public(package) fun new(
...
     );
 
+    // Verify name length.
+    assert!(name.length() <= MAX_NAME_LENGTH, EInvalidNameLength);
+
+    // Verify network address length.
+    assert!(network_address.length() <= MAX_NETWORK_ADDRESS_LENGTH, EInvalidNetworkAddressLength);
+
+    // Verify commission rate.
+    assert!(commission_rate <= 100_00, EIncorrectCommissionRate);
```
mlegner: Nit: Can you define a constant `N_BASIS_POINTS` or something like that? This "magic value" is used in 3 different places.

contracts/walrus/sources/staking/staking_pool.move
karlwuest: I added a check to the node setup command and to the deserialization of the config.

PR #1586, opened 2025-02-14
fix: Handle retries on grpc errors correctly

sadhansood:
## Description

The retry on rpc client is still doing it based on http status code. This PR fixes the retries based on grpc error codes.

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -64,6 +64,13 @@ use crate::{
 
 /// The list of HTTP status codes that are retriable.
 const RETRIABLE_RPC_ERRORS: &[&str] = &["429", "500", "502"];
+/// The list of gRPC status codes that are retriable.
+const RETRIABLE_GRPC_ERRORS: &[tonic::Code] = &[
+    tonic::Code::ResourceExhausted,
+    tonic::Code::Internal,
+    tonic::Code::Unavailable,
+    tonic::Code::DeadlineExceeded,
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -64,6 +64,13 @@ use crate::{
 
 /// The list of HTTP status codes that are retriable.
 const RETRIABLE_RPC_ERRORS: &[&str] = &["429", "500", "502"];
+/// The list of gRPC status codes that are retriable.
+const RETRIABLE_GRPC_ERRORS: &[tonic::Code] = &[
+    tonic::Code::ResourceExhausted,
+    tonic::Code::Internal,
```
halfprice, line 70: Is `Internal` also supposed to be retryable?

wbbradley: I think the intent of retrying here is basically "the server failed for unknown reasons, but we have no reason to think that that is by design, so let's retry."

wbbradley: (Similar to 500 above.)

halfprice: I see. Was just wondering if you guys observed some cases like this.

sadhansood: Yes, we actually did. So, with the load balancer the 429 is wrapped inside the Internal (in the message string), and retrying there is really useful.

PR #1589, opened 2025-02-15
chore(walrus-backup): add configuration for blob upload timeout

wbbradley:
## Description

Allow longer upload times for blobs. Also ensure expired blobs are not fetched, and allow retrying DB connections when encountering unexpected errors.

## Test plan

CI, manual testing

No release impact.

### DIFF
--- a/crates/walrus-service/src/backup/metrics.rs
+++ b/crates/walrus-service/src/backup/metrics.rs
@@ -41,6 +41,9 @@ telemetry::define_metric_set! {
 
         #[help = "The count of database reconnects"]
         db_reconnects: IntCounter[],
+
+        #[help = "Idle"]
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/backup/config.rs
```diff
@@ -147,4 +152,8 @@ pub mod defaults {
     pub fn db_serializability_retry_time() -> Duration {
         Duration::from_millis(100)
     }
+    /// Default time to allow blob uploads to take before timing out.
+    pub fn blob_upload_timeout() -> Duration {
+        Duration::from_secs(60)
```
halfprice, line 188: For large blobs (13G), this may be a bit short.

PR #1591, opened 2025-02-16
feat: Expose blob attribute endpoint in Aggregator

liquid-helium:
## Description

Resolves WAL-331
Resolves WAL-599

Exposed the aggregator endpoint to allow fetching blobs with associated attributes, if the attribute fields are in the allow-list, they will be included in the response headers.

The current allow-list is:
* "content-disposition"
* "content-encoding"
* "content-language"
* "content-location"
* "content-type"
* "link"

One example:
```
curl -I flag "127.0.0.1:31415/v1/blobs/0x937a36a938b953477f3797cafe8ffe1f4ae979f2b112ddede58969a9447733bb/with-attribute"                                                         
curl: (6) Could not resolve host: flag
HTTP/1.1 200 OK
... ...
link: walrus.io
date: Sun, 16 Feb 2025 00:17:32 GMT
```

## Test plan

---

## Release notes

- [ ] Storage node:
- [x] Aggregator: Exposed the aggregator endpoint to allow fetching blobs with associated attributes; if the attribute fields are in the allow-list, they will be included in the response headers.
  The current allow-list is:
  * "content-disposition"
  * "content-encoding"
  * "content-language"
  * "content-location"
  * "content-type"
  * "link"
- [ ] Publisher:
- [x] CLI: Added the allowed-headers option for daemon commands, allowing operators to specify custom allowed headers to filter key-value pairs from the blob attribute.


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -56,6 +65,8 @@ pub const STATUS_ENDPOINT: &str = "/status";
 pub const API_DOCS: &str = "/v1/api";
 /// The path to get the blob with the given blob ID.
 pub const BLOB_GET_ENDPOINT: &str = "/v1/blobs/{blob_id}";
+/// The path to get the metadata with the given blob ID.
+pub const BLOB_WITH_ATTRIBUTE_GET_ENDPOINT: &str = "/v1/blobs/{blob_object_id}/with-attribute";
```
giac-mysten: Remark: i defer a check here to @mlegner or @jpcsmith on the API-ness of this enpoint.

Question: could we just have `"/v1/blobs/{blob_object_id}` (so same as the other endpoint) and differentiate later based on whether the string is a blob id or an object id? And if it is an object id, return the attributes?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -126,6 +137,103 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+                // Add metadata headers if available
+                if let Some(attribute) = attribute {
+                    let allowed_headers = [
+                        (String::from("content-disposition"), CONTENT_DISPOSITION),
+                        (String::from("content-encoding"), CONTENT_ENCODING),
+                        (String::from("content-language"), CONTENT_LANGUAGE),
+                        (String::from("content-location"), CONTENT_LOCATION),
+                        (String::from("content-type"), CONTENT_TYPE),
+                        (String::from("link"), LINK),
+                    ];
```
giac-mysten: Remark (possibly for a future PR): we should bmake these `allowed_headers` configurable, such that different providers can decide to serve different headers.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -126,6 +137,103 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+
+                    for (key, header) in &allowed_headers {
+                        if let Some(value) = &attribute.get(key) {
+                            if let Ok(header_value) = HeaderValue::from_str(value) {
+                                headers.insert(header, header_value);
+                            }
+                        }
+                    }
+                }
+            }
```
giac-mysten: Suggestion (minor optimization): the attribute is implemented as a `VecMap`, which uses a vector to store kv pairs. Therefore, calling `get` for each of the `allowed_headers` may end up scanning the array (which could be pretty large). 

I would either put the `allowed_headers` in a hashmap, and then iterate over the attribute map, or directly parse the attribute map into a hashmap (small preference for the first i guess).

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -126,6 +137,103 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+
+            match &error {
+                GetBlobError::BlobNotFound => {
+                    tracing::debug!(
+                        ?blob_object_id,
+                        "the requested blob object ID does not exist"
+                    )
+                }
+                GetBlobError::Internal(error) => {
+                    tracing::error!(?error, "error retrieving blob metadata")
```
giac-mysten, line 209:
```suggestion
                    tracing::error!(%error, "error retrieving blob metadata")
```

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -126,6 +137,103 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+
+            response
+        }
+        Err(error) => {
+            let error = GetBlobError::from(error);
+
+            match &error {
+                GetBlobError::BlobNotFound => {
+                    tracing::debug!(
+                        ?blob_object_id,
```
giac-mysten, line 204:
```suggestion
                        %blob_object_id,
```

crates/walrus-service/src/client/daemon/routes.rs
mlegner: So we generally try to follow [Google's AIPs](https://google.aip.dev/general). Unfortunately, I didn't find a matching guideline for this scenario.

But I think I would prefer one of the following two versions:
1. `/v1/blobs/by-object-id/{blob_object_id}`
2. `/v1/blob-objects/{blob_object_id}`

In the current choice, we use different types of IDs in the same place, which is not great IMO.

I would also name the constant something like `BLOB_OBJECT_GET_ENDPOINT` or similar and modify the docstring.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -126,6 +137,103 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+                ("content-type" = String,
+                    description = "Content type header if provided in attribute"),
+                ("link" = String,
+                    description = "Link header if provided in attribute")
+            )
+        ),
+        GetBlobError,
+    ),
+)]
+pub(super) async fn get_blob_with_attribute<T: WalrusReadClient>(
```
mlegner: Important: This path needs to be added to the aggregator and daemon API specs in `crates/walrus-service/src/client/daemon/openapi.rs`.

crates/walrus-service/src/client/daemon/routes.rs
mlegner: I would actually keep the debug print in this case to get some additional insights into the error.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -126,6 +137,103 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+/// Retrieve a Walrus blob with its associated attribute.
+///
+/// Reconstructs the blob identified by the provided blob object ID from Sui and then reconstructs
+/// the blob from Walrus by the blob_id in the object and returns its binary data along with
+/// attribute information in the response headers.
+#[tracing::instrument(level = Level::ERROR, skip_all, fields(%blob_object_id))]
+#[utoipa::path(
+    get,
+    path = BLOB_WITH_ATTRIBUTE_GET_ENDPOINT,
+    params(("blob_object_id" = BlobId,)),
```
mlegner: Important: This needs to be an `ObjectID`.

liquid-helium: Oops! It seems to be working by pure coincidence when I tested it locally. :) I put `ObjectIdSchema` here; please let me know if that’s not the correct solution.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Thanks for pointing that out! I added this path to the aggregator and daemon API specs, PTAL.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Kept it as. Since it is only logged for debug! level.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: ditto

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Thank you both for the suggestions, it makes a lot of sense.
I changed it to` /v1/blobs/by-object-id/{blob_object_id}` since it suggests the fact that we are still getting a blob, but with a different identifier.

Also updated to `BLOB_OBJECT_GET_ENDPOINT`.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Good idea, I put it into a HashSet :)

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: That's an excellent suggestion; thank you!

I've added an allowed-headers option to the CLI command, which allows operators to define their own allow list.

A default set of allowed headers will be provided in case the operator does not specify one.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -744,6 +744,15 @@ pub struct DaemonArgs {
...
     pub(crate) blocklist: Option<PathBuf>,
+    /// Allowed headers for the daemon.
+    ///
+    /// This defines the allow-list of headers. It is currently used for the
+    /// /v1/blobs/by-object-id/{blob_object_id} aggregator endpoint. The response will include the
+    /// allowed headers if the specified header names are present in the BlobAttribute associated
+    /// with the requested blob.
+    #[clap(long)]
+    #[serde(default)]
+    pub(crate) allowed_headers: Option<Vec<String>>,
```
mlegner: Important (somewhat): If this is not used for the publisher, it should only be added as an argument for the `aggregator` and `daemon` commands.

crates/walrus-service/src/client/daemon.rs
```diff
@@ -50,6 +60,11 @@ pub trait WalrusReadClient {
         &self,
         blob_id: &BlobId,
     ) -> impl std::future::Future<Output = ClientResult<Vec<u8>>> + Send;
+
+    fn get_blob_with_attribute(
```
mlegner: Nit: Can we call this and related functions `get_blob_by_object_id`?

crates/walrus-service/src/client/cli/args.rs
mlegner: Also, suggestion: Remove the `Option` and instead add a default function to the `default` module below (essentially moving the `default_allowed_headers` function from `crates/walrus-service/src/client/daemon.rs`).

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -126,6 +133,82 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+                    for (key, value) in attribute.iter() {
+                        if allowed_headers.contains(key) {
+                            if let (Ok(header_name), Ok(header_value)) =
+                                (HeaderName::from_str(key), HeaderValue::from_str(value))
+                            {
+                                headers.insert(header_name, header_value);
+                            }
+                        }
+                    }
+                }
```
mlegner: Suggestion: Given the amount of nesting, could we move this header management to a separate function?

liquid-helium: Nice suggestion, a separate function is added.

crates/walrus-service/src/client/daemon.rs
liquid-helium: Done.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Nice suggestion! I created a AggregatorArgs, just in case more will be added in the future. And set the default via the clap default.

PR #1592, opened 2025-02-16
feat: Support update of NodeMetadata

liquid-helium:
## Description

Related to WAL-245

* Feature Addition: Implemented support for updating NodeMetadata.

* Refactor: Introduced SyncedNodeConfigSet to encapsulate storage node configurations that require synchronization. This change aims to simplify the process of extending the list of supported configurations in the future.

## Test plan

---

## Release notes

- [x] Storage node:
    Enabling updating storage node metadata via local storage node config file.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-sui/src/contracts.rs
```diff
@@ -246,6 +246,7 @@ pub mod staking {
     contract_ident!(fn staking::set_storage_price_vote);
     contract_ident!(fn staking::set_write_price_vote);
     contract_ident!(fn staking::set_node_capacity_vote);
+    contract_ident!(fn staking::set_next_commission);
```
mlegner: Question: This is not related to the main goal of this PR, correct? BTW: What is the plan to set the commission? (cc @karlwuest)

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -47,6 +47,7 @@ impl ConfigSynchronizer {
         loop {
             tokio::time::sleep(self.check_interval).await;
 
+            tracing::info!("config_synchronizer: syncing node params");
```
mlegner: Suggestion: Only add INFO logs if something was actually changed, and use DEBUG otherwise. This also applies to the `node parameters are in sync with on-chain values` message.

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -154,32 +190,23 @@ impl SystemContractService for SuiSystemContractService {
...
         let action = calculate_protocol_key_action(
             config.protocol_key_pair().public().clone(),
             config
                 .next_protocol_key_pair()
                 .map(|kp| kp.public().clone()),
-            node_info.public_key.clone(),
-            node_info.next_epoch_public_key.clone(),
+            synced_config.public_key.clone(),
+            synced_config.next_public_key.clone(),
         )?;
```
mlegner, line 207: Nit (unrelated to this PR): Could this function take references instead of owned parameters? Then we wouldn't have to clone everything.

crates/walrus-sui/src/contracts.rs
liquid-helium: I was planning to include the commission rate in this PR but hesitated due to some uncertainty regarding the correct way to set it. Will the following approach meet our needs? I can add it in a follow PR if it is correct.

During synchronization, we retrieve the on-chain commission rate data, which includes the current rate ((rate-c)) and a map of (epoch : rate) for up to 2 epochs ahead of the current epoch.
We then compare this data with the local commission rate and update it only if the local rate differs from the projected rate within the next 2 epochs—specifically, the last value in the list ([rate-c, rate-c+1, rate-c+2]), where (rate-c+1) and (rate-c+2) may not always be present?
@mlegner @karlwuest

karlwuest: For the commission rate we probably don't want automatic sync, because the node may not be authenticated by the StorageNodeCap for this.

Instead, I was planning to add a set of related commands to the walrus CLI with a `node-admin` subcommand, that include setting the authorized entities for commission/governance, voting for contract upgrades, withdrawing the commission, and setting the commission rate, since these are all commands that can potentially be called from a different address. 

I started working on this yesterday the same time as voting for upgrades, but have not made much progress yet due to other issues popping up 😅 
I'll see if I can continue on this today, otherwise we can perhaps quickly sync at the start of your day to see if you can take over.

crates/walrus-service/src/node/contract_service.rs
liquid-helium: I will change it in a following pr.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Thanks for the tip, this log is removed.

crates/walrus-sui/src/contracts.rs
liquid-helium: Sounds good!
I will remove it now, thanks @karlwuest and @mlegner !

PR #1594, opened 2025-02-17
feat: Added --sort-by --desc options for health and info.

liquid-helium:
## Description

related to WAL-552

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Added `--sort-by` and `--desc` options for `walrus health` and `walrus info` commands, to sort the storage nodes by node-id, node-name, node-status, or node-url.

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -1375,11 +1375,11 @@ pub enum HealthSortBy {
     #[default]
     Status,
     /// Sort by node ID
-    NodeId,
+    Id,
     /// Sort by node name
-    NodeName,
+    Name,
     /// Sort by node URL
-    NodeUrl,
+    Url,
 }
 
 /// Generic sort configuration that can be used with any ValueEnum type
--- a/crates/walrus-service/src/client/responses.rs
+++ b/crates/walrus-service/src/client/responses.rs
@@ -476,8 +476,8 @@ impl InfoCommitteeOutput {
                 .0
                 .to_lowercase()
                 .cmp(&b.network_address.0.to_lowercase()),
-            // Default to sorting by name (when None or NodeSortBy::Name)
-            _ => a.name.to_lowercase().cmp(&b.name.to_lowercase()),
+            Some(NodeSortBy::Name) => a.name.to_lowercase().cmp(&b.name.to_lowercase()),
+            None => a.name.to_lowercase().cmp(&b.name.to_lowercase()),
         };
 
         storage_nodes.sort_by(|a, b| if sort.desc { cmp(b, a) } else { cmp(a, b) });
@@ -747,9 +747,9 @@ impl ServiceHealthInfoOutput {
 
         // Sort health_info directly based on sort_by
         health_info.sort_by(|a, b| match sort.sort_by {
-            Some(HealthSortBy::NodeName) => a.cmp_by_name(b),
-            Some(HealthSortBy::NodeId) => a.cmp_by_id(b),
-            Some(HealthSortBy::NodeUrl) => a.cmp_by_url(b),
+            Some(HealthSortBy::Name) => a.cmp_by_name(b),
+            Some(HealthSortBy::Id) => a.cmp_by_id(b),
+            Some(HealthSortBy::Url) => a.cmp_by_url(b),
             Some(HealthSortBy::Status) => a.cmp_by_status(b),
             None => a.cmp_by_status(b),
         });
@@ -761,30 +761,6 @@ impl ServiceHealthInfoOutput {
         Ok(Self { health_info })
     }
 }
-
-impl Ord for NodeHealthOutput {
-    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
-        // Default to Status comparison
-        self.cmp_by_status(other)
-    }
-}
-
-// Required for Ord
-impl PartialOrd for NodeHealthOutput {
-    fn partial_cmp(&self, other: &Self) -> Option<std::cmp::Ordering> {
-        Some(self.cmp(other))
-    }
-}
-
-// Required for Ord/PartialOrd
-impl PartialEq for NodeHealthOutput {
-    fn eq(&self, other: &Self) -> bool {
-        self.cmp(other) == std::cmp::Ordering::Equal
-    }
-}
-
-// Required for PartialEq
-impl Eq for NodeHealthOutput {}
 
 impl NodeHealthOutput {
     pub fn cmp_by_name(&self, other: &Self) -> std::cmp::Ordering {


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -295,6 +295,16 @@ pub enum CliCommands {
         #[clap(long, action)]
         #[serde(default)]
         detail: bool,
+        /// Sort the output by the specified field.
+        ///
+        /// By default, the output is sorted by node status.
+        #[clap(long, value_enum)]
+        #[serde(default)]
+        sort_by: Option<HealthSortBy>,
```
mlegner: Nit: You could use the `HealthSortBy` here directly (instead of an `Option`) and set its `Default` implementation to `Default`. Same for the `NodeSortBy` below.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1290,3 +1322,60 @@ impl UserConfirmation {
...
+        match s.to_lowercase().as_str() {
+            "id" => Ok(Self::Id),
+            "name" => Ok(Self::Name),
+            "url" => Ok(Self::Url),
+            _ => Err(anyhow!(
+                "invalid sort option; valid options are: id, name, url"
+            )),
+        }
+    }
+}
```
mlegner: Question: Is this actually needed? This should be provided by the `ValueEnum` implementation. See, for example the `KeyType` and `KeyFormat` in `crates/walrus-service/bin/node.rs`.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1290,3 +1322,60 @@ impl UserConfirmation {
...
+
+/// Sort options for health information display
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, clap::ValueEnum)]
+#[serde(rename_all = "kebab-case")]
+pub enum HealthSortBy {
+    /// Sort by node status
+    Status,
```
mlegner: Suggestion: If we sort by status, it would be great to have a secondary sort, because most nodes usually have the same (`Active`) status.

crates/walrus-service/src/client/cli/args.rs
mlegner: Then you would also not need the sentence "By default, the output is sorted by node status."

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -523,7 +544,18 @@ pub enum InfoCommands {
...
+        /// Sort the committee information by the specified field.
+        ///
+        /// By default, the output is sorted by name.
+        #[clap(long, value_enum)]
+        #[serde(default)]
+        sort_by: Option<NodeSortBy>,
+        /// Sort in descending order.
+        #[clap(long, action)]
+        #[serde(default)]
+        desc: bool,
```
mlegner: Suggestion: Define a `struct SortBy<T: ValueEnum>` that combines these two fields. This would simplify specifying the sorting and passing it around.

crates/walrus-service/src/client/cli/cli_output.rs
mlegner, line 1: Question: Why is the sorting implemented in the `CliOutput` implementation for `ServiceHealthInfoOutput` while it is implemented in the "constructor" for `InfoCommitteeOutput`?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -861,26 +862,58 @@ impl CliOutput for ServiceHealthInfoOutput {
...
+                        (Ok(info_a), Ok(info_b)) => info_a
+                            .node_status
+                            .to_lowercase()
+                            .cmp(&info_b.node_status.to_lowercase()),
+                        (Err(err_a), Err(err_b)) => err_a.cmp(err_b),
+                        (Err(_), Ok(_)) => std::cmp::Ordering::Greater,
+                        (Ok(_), Err(_)) => std::cmp::Ordering::Less,
+                    }
+                }
+            }
```
mlegner: Suggestion: Define a compare function on the `NodeHealthOutput`.

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -861,26 +862,58 @@ impl CliOutput for ServiceHealthInfoOutput {
...
         let mut read_only_shards = 0;
         let mut node_statuses = std::collections::HashMap::new();
-        let mut error_nodes = Vec::new();
-
         let mut table = create_node_health_table();
-        // Collect summary information while building the table
-        for (node_idx, node) in self.health_info.iter().enumerate() {
+
+        // Create sorted indices based on self.sort_by
+        let mut indices: Vec<usize> = (0..self.health_info.len()).collect();
```
mlegner: Question: Why do we make the detour with the indices? So it is more efficient than sorting a vector of `NodeHealthOutput`?

liquid-helium: Nice observation, I used the indices since &self for this function.

I moved the sorting logic to the constructor instead, and added some comparators functions for it.

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: That's a good idea, change applied.

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: Good point, moved to the ctors!

crates/walrus-service/src/client/cli/args.rs
liquid-helium: That is a fantastic idea, I like it! Change applied, thanks!

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Great observation, thanks! Now sort by name in case of a tie.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: You are right, this is already covered by clap::ValueEnum, removed!

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Thanks for the tip!

Now it is wrapped by a generic SortBy, and it has this 'default' behavior as you suggested here.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1290,3 +1304,55 @@ impl UserConfirmation {
...
+}
+
+/// Generic sort configuration that can be used with any ValueEnum type
+#[derive(Debug, Clone, Deserialize, PartialEq, Eq, clap::Args)]
+#[serde(rename_all = "camelCase")]
+pub struct SortBy<T: clap::ValueEnum + Send + Sync + Default + 'static> {
+    /// Field to sort by
+    #[clap(long, value_enum)]
+    #[serde(default)]
+    pub sort_by: Option<T>,
```
mlegner, line 1392: Suggestion: I don't think you need the `Option` here. You should be able to derive a `Default` on the whole struct.

crates/walrus-service/src/client/responses.rs
```diff
@@ -706,13 +728,82 @@ impl ServiceHealthInfoOutput {
...
+
+// Required for Ord/PartialOrd
+impl PartialEq for NodeHealthOutput {
+    fn eq(&self, other: &Self) -> bool {
+        self.cmp(other) == std::cmp::Ordering::Equal
+    }
+}
+
+// Required for PartialEq
+impl Eq for NodeHealthOutput {}
```
mlegner: Question: (Why) do we need these?

liquid-helium: You are right, we don't need these, added due to previous attempts to implement the comparators.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: I had difficulty achieving that and was unable to succeed for some reason.

The default value for the generic type T will always resolve to one of the specific Enum types (NodeSortBy). 

I've reverted to using the Option<> approach, which allows us to correctly set different defaults for NodeSortBy and HealthSortBy.

We can revisit this if needed.

PR #1595, opened 2025-02-17
feat: add function to allow nodes to proactively join active set

karlwuest:
## Description

Adds a function to let nodes rejoin the active set (without staking) if they have a higher stake than the current node with the minimum stake in the active set.

## Test plan

- added unit test

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking.move
+++ b/contracts/walrus/sources/staking.move
@@ -315,6 +315,11 @@ public fun withdraw_stake(
     staking.inner_mut().withdraw_stake(staked_wal, ctx)
 }
 
+/// Allows a node to join the active set if it has sufficient stake.
+/// This can be useful if another node in the active had its stake
+/// reduced to be lower than that of the current node.
+/// In that case, the current node will be added to the active set either
+/// the next time stake is added or by calling this function.
 public fun try_join_active_set(staking: &mut Staking, cap: &StorageNodeCap) {
     staking.inner_mut().try_join_active_set(cap)
 }
--- a/contracts/walrus/sources/staking/active_set.move
+++ b/contracts/walrus/sources/staking/active_set.move
@@ -129,6 +129,7 @@ public(package) fun remove(set: &mut ActiveSet, node_id: ID) {
 }
 
 /// Sets the maximum size of the active set.
+#[test_only]
 public(package) fun set_max_size(set: &mut ActiveSet, max_size: u16) {
     set.max_size = max_size;
 }


contracts/walrus/sources/staking/active_set.move
```diff
@@ -128,6 +128,11 @@ public(package) fun remove(set: &mut ActiveSet, node_id: ID) {
     });
 }
 
+/// Sets the maximum size of the active set.
+public(package) fun set_max_size(set: &mut ActiveSet, max_size: u16) {
+    set.max_size = max_size;
+}
```
giac-mysten: Question: why should this be changed?

contracts/walrus/sources/staking.move
```diff
@@ -316,6 +315,10 @@ public fun withdraw_stake(
     staking.inner_mut().withdraw_stake(staked_wal, ctx)
 }
 
+public fun try_join_active_set(staking: &mut Staking, cap: &StorageNodeCap) {
```
giac-mysten, line 323: Remark: I noticed that sometimes the logical "self" parameter is called `self`, and other times it is called with the name of the struct (as in this case).

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -514,6 +514,20 @@ public(package) fun withdraw_stake(
...
+    let wctx = &self.new_walrus_context();
+    let pool = &self.pools[node_id];
+
+    // Active set only tracks the stake for the next vote, which either happens for the committee
+    // in wctx.epoch() + 1, or in wctx.epoch() + 2, depending on whether the vote already happened.
+    let balance = match (self.epoch_state) {
+        EpochState::NextParamsSelected(_) => pool.wal_balance_at_epoch(wctx.epoch() + 2),
+        _ => pool.wal_balance_at_epoch(wctx.epoch() + 1),
+    };
+    self.active_set.borrow_mut().insert_or_update(node_id, balance);
```
giac-mysten, line 528: Question: do i understand correctly that `insert_or_update` could result in the removal of a node? 🤔

contracts/walrus/sources/staking.move
```diff
@@ -316,6 +315,10 @@ public fun withdraw_stake(
     staking.inner_mut().withdraw_stake(staked_wal, ctx)
 }
```
giac-mysten, line 317: Question: docs?

contracts/walrus/sources/staking/active_set.move
giac-mysten: should this be test only?

contracts/walrus/sources/staking.move
damirka: It's a style choice, and there are two main approaches to this:

- from Move core to not use `self` because it's not really the same thing as self in Rust
- from Rust devs to use `self` because it's similar to Rust

damirka, line 323: Do I understand it correctly, the expectation is that nodes will call this function directly under some condition?

contracts/walrus/sources/staking/active_set.move
damirka: +1 to @giac-mysten's comments, if we don't use this function anywhere but in tests yet, I'd make it `#[test_only]`

contracts/walrus/sources/staking.move
giac-mysten: right, what i'm saying is that it would be nice to pick one :)

contracts/walrus/sources/staking/active_set.move
karlwuest: I'll make it `test_only` for now. 
We will need it for upgrades in the future since we likely want to expand the maximum number of storage nodes at some point (as we did for testnet).

contracts/walrus/sources/staking.move
karlwuest: Yeah, we're not really using it consistenly. I think we started using `self` in some cases after the method syntax was enabled in move, but have a bit of a mix.

I'll try to go through the contracts later today or tomorrow and make it consistent. I'm a bit unsure which style to use, though. Given that there are no `impl` blocks and no actual `self` keyword, maybe not using `self` is cleaner?

karlwuest: Yes. I'm not sure if or how soon we will actually integrate this in our tooling, but this at least provides the option to the storage nodes. This was suggested in the audit report since storage nodes can be kicked out of the active set, but later qualify again (if another node in the active set loses sufficient stake).

contracts/walrus/sources/staking/staking_inner.move
karlwuest: Yes

karlwuest: Actually in this call this cannot happen, because a node that has stake below the threshold would have already been removed.

giac-mysten: not in this call, I meant in general. 
Just found the name a bit confusing, because an insert can result in a remove.

karlwuest: Yes, in general it can lead to a removal. I would consider that part of the `update` in the name.

PR #1601, opened 2025-02-19
feat: enable contract deployment with existing token

karlwuest:
## Description

- adds a flag to the deployment that publishes the contracts without publishing `WAL`

## Test plan

- tested locally in testbed

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/system_setup.rs
+++ b/crates/walrus-sui/src/system_setup.rs
@@ -233,7 +233,6 @@ pub(crate) async fn publish_package(
 
 pub(crate) struct PublishSystemPackageResult {
     pub walrus_pkg_id: ObjectID,
-    pub wal_pkg_id: ObjectID,
     pub wal_exchange_pkg_id: Option<ObjectID>,
     pub init_cap_id: ObjectID,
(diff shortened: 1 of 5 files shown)


crates/walrus-sui/src/system_setup.rs
```diff
@@ -233,7 +233,6 @@ pub(crate) async fn publish_package(
 
 pub(crate) struct PublishSystemPackageResult {
     pub walrus_pkg_id: ObjectID,
-    pub wal_pkg_id: ObjectID,
```
mlegner, line 236: Question: Why don't we return the ID of the WAL package? Is it tricky to read it from the lock file?

karlwuest: I mainly removed it because we don't really need it (other than for minting just after publishing) and weren't using it anywhere. But reading it from the lock file should be feasible as well if we need it somewhere.

mlegner: If we don't need it, I think it's fine to remove it. 👍

PR #1602, opened 2025-02-19
feat: add RS2 encoding type based on Reed-Solomon erasure code

mlegner:
## Description

This PR adds support for a second encoding type `RS2` to `walrus-core`. This uses the already implemented RedStuff construction, but swaps RaptorQ for Reed-Solomon as the underlying erasure code.

The `EncodingConfig` now contains parameters for both encoding types, which can be used to create appropriate encoders/decoders for the used encoding type.

Outside of `walrus-core` the encoding type is generally fixed with constants to the old `RedStuff` encoding, but mechanisms are already in place to support both encoding types.

Closes WAL-605.

## Test plan

Existing test suite + added unit tests for the new encoding.

---

## Release notes

No release impact yet.

### DIFF
--- a/crates/walrus-core/src/encoding/config.rs
+++ b/crates/walrus-core/src/encoding/config.rs
@@ -616,7 +616,7 @@ pub struct ReedSolomonEncodingConfig {
     /// the Byzantine parameter.
     pub(crate) source_symbols_primary: NonZeroU16,
     /// The number of source symbols for the secondary encoding, which is, simultaneously, the
-    /// number of symbols per primary sliver.It must be strictly less than `n_shards - f`, where `f`
+    /// number of symbols per primary sliver. It must be at most `n_shards - f`, where `f`
     /// is the Byzantine parameter.
     pub(crate) source_symbols_secondary: NonZeroU16,
     /// The number of shards.


crates/walrus-core/src/encoding/basic_encoding/raptorq.rs
mlegner, line 1: This is essentially moved from `crates/walrus-core/src/encoding/basic_encoding.rs`.

crates/walrus-core/src/encoding/basic_encoding/reed_solomon.rs
mlegner, line 1: This is a slightly adapted copy of `crates/walrus-core/src/encoding/basic_encoding/raptorq.rs`. We should deduplicate some code in a follow-up (WAL-621).

crates/walrus-core/src/lib.rs
```diff
@@ -804,9 +805,11 @@ pub struct InvalidEncodingType;
 #[repr(u8)]
 #[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 pub enum EncodingType {
-    /// Default RaptorQ encoding.
-    #[default]
+    /// Original RedStuff encoding using the RaptorQ erasure code.
     RedStuff = 0,
```
mlegner: Question: Should we rename this to `RedStuffRaptorQ`? This would be a small breaking change for some APIs, but may avoid confusion in the future.

karlwuest: Which APIs does this break? I would generally be in favour of renaming it.

mlegner: I believe just our client's JSON API and the publisher.

karlwuest: In that case, I would probably rename it, but @giac-mysten probably knows best how much this would break.

mlegner: Renamed in 165f17cc1b4745de441210b3343054607a8fb66c.

crates/walrus-core/src/encoding/config.rs
```diff
@@ -137,211 +431,443 @@ impl EncodingConfig {
...
     }
+}
+/// Configuration of the Reed-Solomon encoding.
+///
+/// This consists of the number of source symbols for the two encodings and the total number of
+/// shards.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ReedSolomonEncodingConfig {
+    /// The number of source symbols for the primary encoding, which is, simultaneously, the number
+    /// of symbols per secondary sliver. It must be strictly less than `n_shards - 2f`, where `f` is
```
karlwuest: Nit: 
```suggestion
    /// of symbols per secondary sliver. It must be at most `n_shards - 2f`, where `f` is
```

crates/walrus-core/src/encoding/config.rs
```diff
@@ -137,211 +431,443 @@ impl EncodingConfig {
...
+    /// the Byzantine parameter.
+    pub(crate) source_symbols_primary: NonZeroU16,
+    /// The number of source symbols for the secondary encoding, which is, simultaneously, the
+    /// number of symbols per primary sliver.It must be strictly less than `n_shards - f`, where `f`
```
karlwuest: Nit:
```suggestion
    /// number of symbols per primary sliver. It must be at most `n_shards - f`, where `f`
```

crates/walrus-core/src/lib.rs
giac-mysten: i don't know of any component that relies on that name :)

crates/walrus-core/src/encoding/config.rs
```diff
@@ -137,211 +431,443 @@ impl EncodingConfig {
...
+    pub fn get_blob_decoder<E: EncodingAxis>(
         &self,
         blob_size: u64,
-    ) -> Result<BlobDecoder<T>, DataTooLargeError> {
+    ) -> Result<BlobDecoder<ReedSolomonDecoder, E>, DataTooLargeError> {
         BlobDecoder::new(self, blob_size)
     }
 }
 
+impl EncodingConfigTrait for &ReedSolomonEncodingConfig {
```
karlwuest, line 742: Question: Why the `impl` for `&ReedSolomonEncodingConfig` (in addition to `ReedSolomonEncodingConfig`) instead of just for `ReedSolomonEncodingConfig`, given that all functions take an `&self`?

crates/walrus-core/src/encoding/config.rs
```diff
@@ -527,34 +1081,71 @@ mod tests {
...
-            fifty: (51, 16, 32),
-            one_hundred_and_one: (101, 30, 63),
+            // RaptorQ
+            one_raptor: (1, 1, 1, EncodingType::RedStuff),
+            three_raptor: (3, 3, 3, EncodingType::RedStuff),
+            four_raptor: (4, 2, 3, EncodingType::RedStuff),
+            nine_raptor: (9, 5, 7, EncodingType::RedStuff),
+            ten_raptor: (10, 4, 7, EncodingType::RedStuff),
+            fifty_raptor: (51, 16, 32, EncodingType::RedStuff),
+            one_hundred_and_one_raptor: (101, 30, 63, EncodingType::RedStuff),
```
giac-mysten: Remark: I think these should be renamed to`RedStuffRaptorQ`.

crates/walrus-core/src/encoding/basic_encoding/reed_solomon.rs
karlwuest: >We should deduplicate some code in a follow-up

I'm not sure if it's worth putting too much effort into this, since we likely want to go to mainnet with RS2 only, so maybe we'll just remove RaptorQ? Of course, if there's is deduplication that is also useful for other independent encoding schemes it might be worth it, but it's probably less important than some of the other tasks.

crates/walrus-core/src/encoding/config.rs
mlegner: This is because otherwise the `enum_dispatch` doesn't work. I may be able to clean this up in a follow-up though.

PR #1604, opened 2025-02-20
feat: retry shard sync upon PRC error

halfprice:
## Description

Retry shard sync when the sync encounters networking error. This is to cover the case where shard sync is interrupted
by the source node restarting/temporarily unavailable.

The added simtest also discovered a scenario where when a node just becomes a committee member, the source node may not see the new committee yet, and reject shard sync due to unauthorized error. In this case, the node should
still retry shard sync instead of entering recovery directly.

Resolve WAL-444, WAL-603

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -333,7 +333,6 @@ impl ShardSyncHandler {
             return;
         };
 
-        let node_clone = self.node.clone();
         let shard_sync_handler_clone = self.clone();
         let shard_sync_task = tokio::spawn(async move {
             let shard_index = shard_storage.id();
@@ -351,13 +350,7 @@ impl ShardSyncHandler {
             let mut shard_sync_success = false;
             loop {
                 match shard_sync_handler_clone
-                    .sync_shard_impl(
-                        node_clone.clone(),
-                        &shard_sync_handler_clone,
-                        shard_storage.clone(),
-                        current_epoch,
-                        directly_recover_shard,
-                    )
+                    .sync_shard_impl(shard_storage.clone(), current_epoch, directly_recover_shard)
                     .await
                 {
                     SyncShardResult::Success => {
@@ -373,16 +366,15 @@ impl ShardSyncHandler {
                     }
                     SyncShardResult::RetryAfterBackoff(force_recovery) => {
                         let backoff_duration = backoff.next_delay();
-                        if backoff_duration.is_none() {
+                        let Some(backoff_duration) = backoff_duration else {
                             tracing::warn!(
                                 shard_index=%shard_index,
                                 "maximum number of retries reached; stop shard sync; \
                                 restart node to retry shard sync"
                             );
                             break;
-                        }
-                        tokio::time::sleep(backoff_duration.expect("backoff duration is None"))
-                            .await;
+                        };
+                        tokio::time::sleep(backoff_duration).await;
                         if start_time.elapsed()
                             > shard_sync_handler_clone
                                 .config
@@ -396,13 +388,14 @@ impl ShardSyncHandler {
             }
 
             // Remove the task from the shard_sync_in_progress map upon completion.
-            let mut epoch_sync_done = false;
-            if shard_sync_success {
+            let epoch_sync_done = if shard_sync_success {
                 let mut shard_sync_map =
                     shard_sync_handler_clone.shard_sync_in_progress.lock().await;
                 shard_sync_map.remove(&shard_index);
-                epoch_sync_done = shard_sync_map.is_empty();
-            }
+                shard_sync_map.is_empty()
+            } else {
+                false
+            };
 
             if epoch_sync_done {
                 shard_sync_handler_clone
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -24,6 +27,17 @@ use super::{
...
 
+/// The result of syncing a shard.
+enum SyncShardResult {
+    /// The shard sync finished successfully.
+    Success,
+    /// The shard sync is not finished and should be retried after a backoff.
+    /// The bool indicates whether to directly recover the shard instead of using shard sync.
+    RetryAfterBackoff(bool),
+    /// The shard sync contains errors and should be stopped.
+    StopRetry,
```
liquid-helium: nit: Failure or Error?

crates/walrus-service/src/node/config.rs
```diff
@@ -533,6 +537,7 @@ impl Default for ShardSyncConfig {
             blob_certified_check_interval: Duration::from_secs(60),
             max_concurrent_metadata_fetch: 10,
             shard_sync_concurrency: 10,
+            shard_sync_retry_switch_to_recovery_interval: Duration::from_secs(1800),
```
mlegner: Suggestion: I would probably go for a higher default (2 hours?), especially for Mainnet.  The cost for the recovery is really high, so waiting/retrying a few hours is pretty cheap in comparison.

crates/walrus-service/src/node/shard_sync.rs
halfprice: Changed to Failed.

crates/walrus-service/src/node/config.rs
halfprice: Make sense. Changed to 2 hours.

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -323,103 +337,67 @@ impl ShardSyncHandler {
...
+            // Whether to directly recover the shard instead of using shard sync.
+            let mut directly_recover_shard = false;
+            let mut shard_sync_success = false;
+            loop {
+                match shard_sync_handler_clone
+                    .sync_shard_impl(
                         node_clone.clone(),
-                        &shard_sync_handler_clone.config,
-                        recover_shard,
+                        &shard_sync_handler_clone,
```
wbbradley: Why are we passing another duplicate receiver in to this method?

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -323,103 +337,67 @@ impl ShardSyncHandler {
         let shard_sync_handler_clone = self.clone();
```
wbbradley, line 336: Is it necessary to clone `self.node` since we're already cloning `self`? It looks like the only usage of `node_clone` is itself cloned again below.

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -323,103 +337,67 @@ impl ShardSyncHandler {
...
-                        if !recover_shard {
-                            *directly_recover_shard.lock().await = true;
-                            tracing::info!(
-                                walrus.shard_index = %shard_index,
-                                "shard sync failed; directly recovering shard next time"
+                        break;
+                    }
+                    SyncShardResult::RetryAfterBackoff(force_recovery) => {
+                        let backoff_duration = backoff.next_delay();
+                        if backoff_duration.is_none() {
```
wbbradley:
```suggestion
                        let Some(backoff_duration) = backoff_duration else {
```

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -323,103 +337,67 @@ impl ShardSyncHandler {
...
+                                restart node to retry shard sync"
                             );
-                            false // Retry with direct recovery
-                        } else {
-                            // TODO(#705): also do retries for other retriable errors. E.g. RPC
-                            // error.
-                            true // Exit retry loop if direct recovery also failed
+                            break;
+                        }
+                        tokio::time::sleep(backoff_duration.expect("backoff duration is None"))
```
wbbradley: Then can remove this expect.

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -323,103 +337,67 @@ impl ShardSyncHandler {
...
                     }
                 }
-            })
-            .await;
+            }
 
             // Remove the task from the shard_sync_in_progress map upon completion.
-            let epoch_sync_done;
-            {
+            let mut epoch_sync_done = false;
```
wbbradley: Can we eliminate this extra mutable variable, and nest the logic for calling `epoch_sync_done` inside this next `if` statement?

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -440,6 +418,119 @@ impl ShardSyncHandler {
         entry.insert(shard_sync_task);
     }
 
+    /// Syncs a shard using shard sync. If `directly_recover_shard` is true, the shard will be
+    /// directly recovered instead of using shard sync.
+    async fn sync_shard_impl(
+        &self,
+        node: Arc<StorageNodeInner>,
+        shard_sync_handler: &ShardSyncHandler,
```
wbbradley: As mentioned above, why is there an extra ShardSyncHandler being passed into this method? `self` is already one...

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -440,6 +418,119 @@ impl ShardSyncHandler {
...
+                    return SyncShardResult::Success;
+                }
+
+                Self::handle_sync_error(&error, shard_index, directly_recover_shard)
+            }
+        }
+    }
+
+    /// Handles sync shard errors and determines whether/how to retry
+    fn handle_sync_error(
```
wbbradley, line 467: This method seems like it doesn't need to be inside the ShardSyncHandler, as it relies on no ShardSyncHandler instance or type variable. Could be pulled out into the module as an accessory function to clarify its purpose?

crates/walrus-service/src/node/shard_sync.rs
halfprice: Good catch! I think both node and shard_sync_handler are indeed not needed. I think every time I evolve the code, I just updating what's already there.

PTAL

crates/walrus-service/src/node/shard_sync.rs
halfprice: Good suggestion. Applied.

crates/walrus-service/src/node/shard_sync.rs
halfprice: Done

crates/walrus-service/src/node/shard_sync.rs
halfprice: Both `node` and `shard_sync_handler` are not needed. PTAL.

crates/walrus-service/src/node/shard_sync.rs
halfprice: I put it here because I want to only use it in the `ShardSyncHandler` namespace, because this is very specific to the shard sync logic, and probably not useful anywhere else.

PR #1606, opened 2025-02-20
fix: clarify extend semantics and fix client behavior

mlegner:
## Description

Using `epochs_ahead` for extensions in the contracts is misleading and a cause for confusion: The word indicates something relative to the current epoch (similar to the register) but the actual semantics is different.

This PR renames this parameter to `epochs_extended`, which is aligned with the semantics, and fixes two bugs in the client that resulted from this confusion. (Example of the bug is [here](https://github.com/MystenLabs/walrus-docs/actions/runs/13430256831/job/37520475061).)

## Test plan

Extended existing E2E test + manual test.

---

## Release notes

- [x] CLI: Fix `walrus store` behavior when blobs can be extended.

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -1078,16 +1078,10 @@ async fn test_extend_owned_blobs() -> TestResult {
             PostStoreAction::Keep,
         )
         .await?;
-    let (end_epoch, blob_object_id) = {
-        let BlobStoreResult::NewlyCreated { blob_object, .. } = result
-            .into_iter()
-            .next()
-            .expect("expect one blob store result")
-        else {
-            panic!("expect newly stored blob")
-        };
-        (blob_object.storage.end_epoch, blob_object.id)
+    let BlobStoreResult::NewlyCreated { blob_object, .. } = result[0].clone() else {
+        panic!("expect newly stored blob")
     };
+    let (end_epoch, blob_object_id) = (blob_object.storage.end_epoch, blob_object.id);
     assert_eq!(end_epoch, current_epoch + 1);
 
     // Extend it by 5 epochs.
@@ -1111,20 +1105,24 @@ async fn test_extend_owned_blobs() -> TestResult {
         .reserve_and_store_blobs(
             &[blob.as_slice()],
             20,
-            StoreWhen::Always,
+            StoreWhen::NotStored,
             BlobPersistence::Permanent,
             PostStoreAction::Keep,
         )
         .await?;
     let BlobStoreResult::NewlyCreated {
-        blob_object,
+        blob_object: second_store_blob_object,
         resource_operation,
         ..
     } = result[0].clone()
     else {
         panic!("unexpected result")
     };
-    assert_eq!(blob_object.storage.end_epoch, current_epoch + 20);
+    assert_eq!(second_store_blob_object.id, blob_object_id);
+    assert_eq!(
+        second_store_blob_object.storage.end_epoch,
+        current_epoch + 20
+    );
     assert!(resource_operation.is_extend());
 
     Ok(())


crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -1095,13 +1098,31 @@ async fn test_extend_owned_blobs() -> TestResult {
...
+            StoreWhen::Always,
+            BlobPersistence::Permanent,
+            PostStoreAction::Keep,
+        )
+        .await?;
+    assert!(matches!(
+        &result[0],
+        BlobStoreResult::NewlyCreated {
+            blob_object: Blob{storage:StorageResource{end_epoch, ..}, ..}, .. }
+    if *end_epoch == current_epoch + 20));
```
karlwuest: Nit: This `matches` statement with the `if` clause is not very readable. This is a matter of taste, of course, so feel free to ignore this if you disagree, but I would prefer something like the following:
```suggestion
    let BlobStoreResult::NewlyCreated { blob_object, .. } = result[0] else { bail!(...)};
    assert!(blob_object.storage.end_epoch == current_epoch + 20);
```

Question: Why does this return a `BlobStoreResult::NewlyCreated`? Maybe it would be good to have a `BlobStoreResult::Extended`?

crates/walrus-service/src/client/responses.rs
```diff
@@ -691,7 +691,7 @@ pub struct FundSharedBlobOutput {
 /// The output of the `walrus extend` command.
 pub struct ExtendBlobOutput {
     /// The number of epochs extended by.
-    pub epochs_ahead: u32,
+    pub epochs_extended: u32,
```
karlwuest: Nit:
```suggestion
    pub epochs_extended: EpochCount,
```

crates/walrus-sui/src/client.rs
```diff
@@ -762,16 +762,16 @@ impl SuiContractClient {
             .await
     }
 
-    /// Extends the shared object epoch.
+    /// Extends the shared blob's lifetime by `epochs_extended` epochs.
     pub async fn extend_shared_blob(
         &self,
         shared_blob_obj_id: ObjectID,
-        epochs_ahead: u32,
+        epochs_extended: u32,
```
```suggestion
        epochs_extended: EpochCount,
```

crates/walrus-sui/src/client.rs
```diff
@@ -1753,15 +1753,15 @@ impl SuiContractClientInner {
         Ok(())
     }
 
-    /// Extends the shared object epoch.
+    /// Extends the shared blob's lifetime by `epochs_extended` epochs.
     pub async fn extend_shared_blob(
         &mut self,
         shared_blob_obj_id: ObjectID,
-        epochs_ahead: u32,
+        epochs_extended: u32,
```
```suggestion
        epochs_extended: EpochCount,
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -550,7 +550,7 @@ impl WalrusPtbBuilder {
     pub async fn extend_shared_blob(
         &mut self,
         shared_blob_object_id: ObjectID,
-        epochs_ahead: Epoch,
+        epochs_extended: Epoch,
```
karlwuest: Nit:
```suggestion
        epochs_extended: EpochCount,
```

crates/walrus-e2e-tests/tests/test_client.rs
mlegner: > Why does this return a `BlobStoreResult::NewlyCreated`? Maybe it would be good to have a `BlobStoreResult::Extended`?

The semantics of the `NewlyCreated` have expanded over time, and already included the `ReuseRegistration` case, which is also not strictly a creation. But I agree that we could maybe rename this (although that would be a breaking change).

karlwuest: Ok, something for later then 😄

PR #1608, opened 2025-02-21
feat: implement support for RS2 in the client and node

giac-mysten:
## Description

Closes WAL-607

## Test plan

Existing test suite + planned manual tests on PTN and Testnet (see [WAL-625: Run manual tests on PTN](https://linear.app/mysten-labs/issue/WAL-625/run-manual-tests-on-ptn)).

---

## Release notes

- [x] Storage node: Storage node now supports multiple encoding types. Depending on features selected at compile time, it accepts a subset of them.
- [ ] Aggregator:
- [x] Publisher: Added `encoding_type` query parameter. Added additional variants for the encoding type.
- [x] CLI: The CLI supports multiple encoding types for reads and writes. Depending on features selected at compile time, it uses a different default type and accepts a subset of them for writes.

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -889,6 +889,7 @@ async fn test_blob_operations_with_subsidies() -> TestResult {
     let store_result = client
         .reserve_and_store_blobs(
             &blobs,
+            DEFAULT_ENCODING,
             1,
             StoreWhen::Always,
             BlobPersistence::Permanent,
--- a/crates/walrus-simtest/tests/simtest.rs
+++ b/crates/walrus-simtest/tests/simtest.rs
@@ -18,7 +18,6 @@ mod tests {
     use tokio::{sync::RwLock, task::JoinHandle, time::Instant};
     use walrus_core::{
         encoding::{Primary, Secondary},
-        EncodingType,
         DEFAULT_ENCODING,
     };
     use walrus_proc_macros::walrus_simtest;


crates/walrus-core/src/lib.rs
```diff
@@ -76,6 +76,40 @@ pub type Epoch = u32;
 /// The number of epochs.
 pub type EpochCount = u32;
 
+const ENCODING_CONFIG: (&[EncodingType], EncodingType) = {
+    #[cfg(feature = "raptorq-and-rs2")]
```
karlwuest: Nit: I think it's better to specify both features separately, because when building, the features can both be specified without the `raptorq-and-rs2` feature set.
```suggestion
    #[cfg(all(feature = "rs2", feature = "raptorq"))]
```

crates/walrus-core/Cargo.toml
```diff
@@ -7,6 +7,9 @@ edition.workspace = true
 license.workspace = true
 
 [features]
+raptorq = []
+raptorq-and-rs2 = ["raptorq", "rs2"]
```
giac-mysten, line 11: Nit(follow-up): not sure how useful the joint flag is, especially if we will have more than 2 encodings.

crates/walrus-core/src/lib.rs
```diff
@@ -76,6 +76,40 @@ pub type Epoch = u32;
 /// The number of epochs.
 pub type EpochCount = u32;
 
+const ENCODING_CONFIG: (&[EncodingType], EncodingType) = {
```
giac-mysten: Important(follow-up): this can go in a future PR, but for mainnet we possibly need to support only RS on mainnet, and RS and RQ on testnet. Therefore, we cannot have this set at compile time but based on a network-specific configuration (as in #1532 )

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -185,7 +189,11 @@ impl ClientCommandRunner {
                 file_or_blob_id,
                 timeout,
                 rpc_arg: RpcArg { rpc_url },
-            } => self.blob_status(file_or_blob_id, timeout, rpc_url).await,
+                encoding_type,
+            } => {
+                self.blob_status(file_or_blob_id, timeout, rpc_url, encoding_type)
```
karlwuest, line 194: Remark: In many commands, the `encoding_type` is passed even though it is only needed if a file is specified. We may want to later change the grouping in the CLI arguments a bit, s.t. we only pass it if it is a file, but don't need it if an object ID or blob ID is specified.

crates/walrus-e2e-tests/tests/test_event_blobs.rs
```diff
@@ -22,7 +22,7 @@ async fn test_event_blobs() -> anyhow::Result<()> {
         test_cluster::default_setup_with_num_checkpoints_generic::<StorageNodeHandle>(
             Duration::from_secs(60 * 60),
             TestNodesConfig {
-                node_weights: vec![1, 1],
+                node_weights: vec![2, 2],
```
giac-mysten, line 25: Question: why the changes in weights? here and elsewhere.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -226,7 +223,7 @@ async fn test_inconsistency(failed_nodes: &[usize]) -> TestResult {
     let (pairs, metadata) = client
         .as_ref()
         .encoding_config()
-        .get_for_type(ENCODING_TYPE)
+        .get_for_type(DEFAULT_ENCODING)
```
giac-mysten, line 232: Nit(follow-up): At some point would be good to test all supported encodings in these tests, not only the default.

crates/walrus-service/src/client/responses.rs
```diff
@@ -403,33 +414,67 @@ impl InfoPriceOutput {
             storage_units_from_size(metadata_storage_size) * storage_price_per_unit_size;
 
         let marginal_price = storage_units_from_size(
-            encoded_slivers_length_for_n_shards(n_shards, marginal_size, ENCODING_TYPE)
+            encoded_slivers_length_for_n_shards(n_shards, marginal_size, DEFAULT_ENCODING)
```
karlwuest:
```suggestion
            encoded_slivers_length_for_n_shards(n_shards, marginal_size, encoding_type)
```

crates/walrus-service/src/client/responses.rs
```diff
@@ -459,13 +504,13 @@ impl InfoCommitteeOutput {
...
+            source_symbols_for_n_shards(n_shards, DEFAULT_ENCODING);
 
         let max_sliver_size =
-            max_sliver_size_for_n_secondary(n_secondary_source_symbols, ENCODING_TYPE);
-        let max_blob_size = max_blob_size_for_n_shards(n_shards, ENCODING_TYPE);
+            max_sliver_size_for_n_secondary(n_secondary_source_symbols, DEFAULT_ENCODING);
+        let max_blob_size = max_blob_size_for_n_shards(n_shards, DEFAULT_ENCODING);
         let max_encoded_blob_size =
-            encoded_blob_length_for_n_shards(n_shards, max_blob_size, ENCODING_TYPE)
+            encoded_blob_length_for_n_shards(n_shards, max_blob_size, DEFAULT_ENCODING)
```
karlwuest, line 513: Remark: We should probably reconsider including these outputs as part of the committee output, especially if we add additional encoding types in the future (e.g. full replication). Can be ignored for this PR, though.

giac-mysten: Good point, i think we should create an `walrus info encoding` command -> WAL-630

crates/walrus-service/src/client/cli/runner.rs
giac-mysten: agreed, WAL-631

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -70,6 +74,7 @@ mod tests {
             .as_ref()
             .reserve_and_store_blobs_retry_committees(
                 &[blob.as_slice()],
+                ENCODING_TYPE,
```
mlegner: Should this be `DEFAULT_ENCODING`?

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -450,6 +450,31 @@ impl<'a> ExpandedMessageMatrix<'a> {
...
 }
 
+/// A wrapper around the blob decoder for different encoding types.
+#[derive(Debug)]
+pub enum BlobDecoderEnum<'a, E: EncodingAxis> {
+    /// The RaptorQ decoder.
+    RaptorQ(BlobDecoder<'a, RaptorQDecoder, E>),
+    /// The Reed-Solomon decoder.
+    ReedSolomon(BlobDecoder<'a, ReedSolomonDecoder, E>),
+}
```
mlegner, line 460: Question: Do we need the enum? Is this because of the `decode_sliver_by_sliver` construction? If so, we may be able to avoid this enum and just keep the generics, because we only need that function in the case of RaptorQ anyway.

Can be done as part of [WAL-621: Refactor encoding code](https://linear.app/mysten-labs/issue/WAL-621/refactor-encoding-code) and/or [WAL-611: Optimizations for new encoding type](https://linear.app/mysten-labs/issue/WAL-611/optimizations-for-new-encoding-type).

crates/walrus-core/src/encoding/config.rs
```diff
@@ -311,6 +312,23 @@ pub enum EncodingConfigEnum<'a> {
     ReedSolomon(&'a ReedSolomonEncodingConfig),
 }
 
+impl<'a> EncodingConfigEnum<'a> {
+    /// Returns a decoder for the given blob size and encoding axis.
+    pub fn get_blob_decoder<E: EncodingAxis>(
```
mlegner, line 317: Nit: I believe this could be added to the `EncodingConfigTrait`.

crates/walrus-service/src/node.rs
```diff
@@ -2064,6 +2069,19 @@ impl ServiceState for StorageNodeInner {
...
+            .context("database error when storing sliver")?
+            .ok_or(StoreSliverError::MissingMetadata)?;
+
+        // Check if encoding type is supported
+        let encoding_type = metadata.metadata().encoding_type();
+        if !encoding_type.is_supported() {
+            return Err(StoreSliverError::UnsupportedEncodingType(encoding_type));
+        }
+
         self.store_sliver_unchecked(blob_id, sliver_pair_index, sliver)
```
mlegner: Suggestion: This function also retrieves the metadata from the DB. Can we either move the check above to the inner function or pass in the metadata?

giac-mysten: Good point. Passing in the `VerifiedBlobMetadataWithId`, instead of the `blob_id` solves this.

crates/walrus-core/src/encoding/config.rs
giac-mysten: If we remove the `BlobDecoderEnum`, what is the return type of this? I think once the refactoring you propose above is done, we can probably remove this enum altogether?

crates/walrus-core/src/encoding/blob_encoding.rs
giac-mysten: > Is this because of the decode_sliver_by_sliver construction?

i think so

crates/walrus-core/src/lib.rs
mlegner: AFAIU, we just disable the _write path_ for some encoding at compile time and keep supporting the _read path_ for all. So we can compile this with just the `rs2` feature and should get the behavior we want.

crates/walrus-core/src/encoding/config.rs
mlegner: Indeed, then we cannot do it. Let's take a closer look in the follow-up work.

PR #1609, opened 2025-02-21
fix: return error codes for all asserts in the contracts

karlwuest:
## Description

- Makes sure that we return error codes for all asserts
- Adds comments to all error codes
- Adds a check that obtained storage has a non-zero size
- Sets the VERSION to 1 (to be consistent with the versioning in the Move.lock which starts with 1)
- Includes some small additional cleanup

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/subsidies/sources/subsidies.move
+++ b/contracts/subsidies/sources/subsidies.move
@@ -26,6 +26,7 @@ const MAX_SUBSIDY_RATE: u16 = 10_000; // 100%
 const EInvalidSubsidyRate: u64 = 0;
 /// The admin cap is not authorized for the `Subsidies` object.
 const EUnauthorizedAdminCap: u64 = 1;
+/// The package version is not compatible with the `Subsidies` object.
 const EWrongVersion: u64 = 2;
 
 // === Structs ===
--- a/contracts/walrus/sources/staking/exchange_rate.move
+++ b/contracts/walrus/sources/staking/exchange_rate.move
@@ -2,12 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /// A utility module which implements an `ExchangeRate` struct and its methods.
-/// It stores a fixed point exchange rate between the Wal token and pool shares.
+/// It stores a fixed point exchange rate between the WAL token and pool shares.
 module walrus::pool_exchange_rate;
 
 // Error codes
 // Error types in `walrus-sui/types/move_errors.rs` are auto-generated from the Move error codes.
-/// The exchange rate between the shares and the Wal token is invalid.
+/// The exchange rate between the shares and the WAL token is invalid.
 const EInvalidRate: u64 = 0;
 
 /// Represents the exchange rate for the staking pool.


contracts/subsidies/tests/subsidies_tests.move
```diff
@@ -400,24 +400,6 @@ fun test_reserve_space_funds_with_subsidies(): (System, Coin<WAL>, Storage) {
     (system, payment, storage)
 }
 
-#[test]
-fun test_reserve_space_no_funds_no_subsidies_zero_storage(): (System, Coin<WAL>, Storage) {
```
karlwuest, line 404: This test is removed because purchasing zero sized storage is no longer possible. A test for that is added in `contracts/walrus/tests/system/storage_resource_tests.move`.

contracts/subsidies/sources/subsidies.move
```diff
@@ -16,13 +16,15 @@ use wal::wal::WAL;
...
 
 /// Subsidy rate is in basis points (1/100 of a percent).
 const MAX_SUBSIDY_RATE: u16 = 10_000; // 100%
 
 // === Errors ===
+/// The provided subsidy rate is invalid.
 const EInvalidSubsidyRate: u64 = 0;
+/// The admin cap is not authorized for the `Subsidies` object.
 const EUnauthorizedAdminCap: u64 = 1;
 const EWrongVersion: u64 = 2;
```
mlegner, line 30: Question: I guess this one is obvious enough to not need a docstring?

contracts/walrus/sources/staking.move
```diff
@@ -259,21 +261,6 @@ public fun initiate_epoch_change(staking: &mut Staking, system: &mut System, clo
...
-///     shard recovery
-public fun shard_transfer_failed(
-    staking: &mut Staking,
-    cap: &StorageNodeCap,
-    other_node_id: ID,
-    shard_ids: vector<u16>,
-) {
-    staking.inner_mut().shard_transfer_failed(cap, other_node_id, shard_ids);
-}
-
```
mlegner, line 276: Question: This is removed because we don't currently use it, correct?

contracts/walrus/sources/staking/exchange_rate.move
```diff
@@ -2,11 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /// A utility module which implements an `ExchangeRate` struct and its methods.
-/// It stores a fixed point exchange rate between the Wal token and pool token.
+/// It stores a fixed point exchange rate between the Wal token and pool shares.
```
mlegner: Nit (tiny): The token should probably be written as `WAL` instead of `Wal`.

contracts/subsidies/sources/subsidies.move
karlwuest: Ah, I missed that after rebasing (Sadhan added this in his PR). I'll add a comment for this as well.

contracts/walrus/sources/staking.move
karlwuest: Yes. I had actually thought that we already removed the public function.

PR #1612, opened 2025-02-22
feat: add allowlist to walrus proxy

halfprice:
## Description

So that we can get metrics in epoch 0. Nodes in the allowlist are always allowed. Reading from the allowlist follows the
same cadence as reading onchian information, so updating the allowlist does not require restarting the proxy.

Resolve WAL-561

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-proxy/src/providers/walrus/provider.rs
+++ b/crates/walrus-proxy/src/providers/walrus/provider.rs
@@ -12,7 +12,6 @@ use std::{
 use fastcrypto::traits::EncodeDecodeBase64;
 use once_cell::sync::Lazy;
 use prometheus::{CounterVec, HistogramOpts, HistogramVec, Opts};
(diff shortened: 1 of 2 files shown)


crates/walrus-proxy/src/fixtures/allowlist.yaml
mlegner, line 1: Question: Would it be possible to accept node IDs instead or in addition? We get those directly from the node operators. We can of course also extract this info in an automatic way, but having a file that just consists of node IDs would be easier.

halfprice: Yeah, I think so. But it requires a bigger change since the proxy currently only takes node info to authenticate. Let me see if I can plume in something quick.

PR #1613, opened 2025-02-22
feat:  add exact size for JWT Token

yanganto:
## Description

Add size of JWT Token to restrict payload with size.

The CI flow do not use the correct version of Rust, and it causes red light of CI here and not really related with this PR.
So the CI issue is reported in https://github.com/MystenLabs/walrus/issues/1663 
Please ignore the this CI error.
## Test plan

A test case `verify_body_size_with_exact_size` is added.

---

## Release notes
- [ ] Storage node:
- [ ] Aggregator:
- [X] Publisher: Add `size` field of JWTToken
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon/auth.rs
+++ b/crates/walrus-service/src/client/daemon/auth.rs
@@ -122,7 +122,7 @@ impl Claim {
         }
 
         if self.max_size.is_some() && self.size.is_some() {
-            return Err(PublisherAuthError::SizeIncorrect);
+            return Err(PublisherAuthError::InvalidSize);
         }
 
         // If verify_upload is disabled, skip the rest of the checks.
@@ -331,6 +331,11 @@ pub enum PublisherAuthError {
     #[rest_api_error(reason = "INVALID_EPOCHS", status = ApiStatusCode::FailedPrecondition)]
     InvalidEpochs,
 
+    /// Misuse size, max-size field of the token to restrict upload
+    #[error("misuse size, max-size field of the token to restrict upload")]
+    #[rest_api_error(reason = "INVALID_SIZE", status = ApiStatusCode::FailedPrecondition)]
+    InvalidSize,
+
     /// Epochs is above the maximum allowed.
     #[error("the epochs field in the query is above the maximum allowed")]
     #[rest_api_error(reason = "EPOCHS_ABOVE_MAX", status = ApiStatusCode::FailedPrecondition)]
@@ -346,8 +351,8 @@ pub enum PublisherAuthError {
     #[rest_api_error(reason = "MISSING_SEND_OBJECT_TO", status = ApiStatusCode::FailedPrecondition)]
     MissingSendObjectTo,
 
-    /// The size of the body is incorrect, or misuse size conditions to restrict upload
-    #[error("the size of the body is incorrect, or misuse size conditions to restrict upload")]
+    /// The size of the body is incorrect
+    #[error("the size of the body is incorrect")]
     #[rest_api_error(reason = "SIZE_INCORRECT", status = ApiStatusCode::FailedPrecondition)]
     SizeIncorrect,
 


crates/walrus-core/src/metadata.rs
```diff
@@ -59,6 +59,16 @@ pub type VerifiedBlobMetadataWithId = BlobMetadataWithId<true>;
...
 
+impl VerifiedBlobMetadataWithId {
+    /// Show the original size from blob metadata.
+    pub fn size(&self) -> u64 {
+        let BlobMetadata::V1(BlobMetadataV1 {
+            unencoded_length, ..
+        }) = self.metadata;
+        unencoded_length
+    }
+}
```
mlegner: Question: Is this really needed? Can't you just call `verified_metadata.metadata.unencoded_length()`?

crates/walrus-sui/src/client.rs
```diff
@@ -316,17 +316,17 @@ impl BlobPersistence {
...
     /// Transfer the blob object to the given address.
-    TransferTo(SuiAddress),
+    TransferTo(SuiAddress, Option<u64>),
     /// Keep the blob object in the wallet that created it.
-    Keep,
+    Keep(Option<u64>),
     /// Put the blob into a shared blob object.
-    Share,
+    Share(Option<u64>),
 }
```
mlegner: Important: These changes seem very invasive and the semantics of the added value is unrelated to the purpose of the enum. I would suggest to pass this limit in a different way.

crates/walrus-service/src/client.rs
```diff
@@ -497,6 +497,15 @@ impl Client<SuiContractClient> {
...
             .await?;
+        let mut size = 0;
+        for pair_and_metadata in pairs_and_metadata.iter() {
+            size += pair_and_metadata.1.size();
+        }
+        if !post_store.check_size(size) {
+            return Err(ClientError::from(
+                ClientErrorKind::PayloadVerificationFailed,
+            ));
+        }
```
mlegner: Important: Instead of passing the maximum size in the `PostStoreAction`, which is pretty unrelated, why not just add another function parameter? Or, even better, check the size outside of this function by applying the [`encoded_blob_length_for_n_shards`](https://github.com/MystenLabs/walrus/blob/3e04b2986f74255132d05a18be64b402a827e254/crates/walrus-core/src/encoding/config.rs#L1018) function to the blobs.

crates/walrus-core/src/metadata.rs
yanganto: Agreed

crates/walrus-sui/src/client.rs
yanganto: I will try to avoid to use postaction

crates/walrus-service/src/client.rs
yanganto: agreed, let me do this.

crates/walrus-service/src/client.rs
```diff
@@ -493,10 +494,22 @@ impl Client<SuiContractClient> {
...
+        exact_size: Option<u64>,
     ) -> ClientResult<Vec<BlobStoreResult>> {
         let pairs_and_metadata = self
             .encode_blobs_to_pairs_and_metadata(blobs, encoding_type)
             .await?;
+        if let Some(exact_size) = exact_size {
+            let mut size = 0;
+            for pair_and_metadata in pairs_and_metadata.iter() {
+                size += pair_and_metadata.1.metadata().unencoded_length();
+            }
```
mlegner: Nit: Probably a matter of taste, but how about the following?
```suggestion
            let size = pairs_and_metadata.iter().fold(0, |sum, (_, metadata)| {
                sum + metadata.metadata().unencoded_length()
            });
```

crates/walrus-service/src/client/error.rs
```diff
@@ -170,6 +170,9 @@ pub enum ClientErrorKind {
         FROST for staking"
     )]
     StakeBelowThreshold(u64),
+    /// The payload client provided do not pass the verification.
```
mlegner: Nit:
```suggestion
    /// The payload provided by the client does not pass the verification.
```

crates/walrus-service/src/client.rs
```diff
@@ -493,10 +494,22 @@ impl Client<SuiContractClient> {
         store_when: StoreWhen,
         persistence: BlobPersistence,
         post_store: PostStoreAction,
+        exact_size: Option<u64>,
     ) -> ClientResult<Vec<BlobStoreResult>> {
         let pairs_and_metadata = self
             .encode_blobs_to_pairs_and_metadata(blobs, encoding_type)
             .await?;
+        if let Some(exact_size) = exact_size {
```
mlegner: Suggestion: Move this check to the [`ClientMultiplexer::write_blob`](https://github.com/MystenLabs/walrus/blob/9d214deb9c6da07aca74c4f5e8be842a1ef73210/crates/walrus-service/src/client/multiplexer.rs#L174) function. This is only used in the publisher, so it doesn't have to be in the inner client.

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -120,14 +124,37 @@ impl Claim {
...
         }
 
-        if let Some(max_size) = self.max_size {
-            if body_size_hint > max_size {
+        if let Some(body_size_upper_hint) = body_size_hint.upper() {
+            if let Some(max_size) = self.max_size {
+                if body_size_upper_hint > max_size {
+                    tracing::debug!(
+                        max_size = max_size,
+                        body_size_upper_hint = body_size_upper_hint,
```
mlegner: Nit: This is equivalent:
```suggestion
                        max_size,
                        body_size_upper_hint,
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -60,6 +60,10 @@ pub struct Claim {
     /// The maximum size of the blob that can be stored, in bytes.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub max_size: Option<u64>,
+
+    /// The exact size of the blob that can be stored, in bytes.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub size: Option<u64>,
```
mlegner: Question: Why do we need both here? And what is the benefit of enforcing an exact size? And can both be set at the same time? What happens if they are inconsistent?

crates/walrus-sui/src/client.rs
mlegner, line 1: I think these changes can be reverted.

yanganto: Yes, the `max_size` can be replaced by `size`, if it is good.  And it is thankful to figure out how to use `SizeHint` in previous works.  I will revert it in another commit, in case we want to keep both, we can easier to drop it.

Some previous discussion here.
https://github.com/MystenLabs/walrus/pull/1487#discussion_r1944022043

Overall, it is about the security, size and hash(`blob_id`) can identify a file.  
We can not use `blob_id` due to the computation overhead, but the `size` is a trade-off and cheap enough to check the identity of file is the same as user want to upload.

crates/walrus-service/src/client.rs
yanganto: I feel we are not in the same direction of this topic.  

Eventually, we will let user upload multiple blobs of a supper big file at the same time, and the `size` is still by a file not blob.  The uploads of blobs can be async.  So no mater which upload request is the last one, one of the requests will verified the overall about the size is correct or not.  That is why I put this in put PostAction at first.

Overall exact size is good enough for my current task and make sure it is secure, I will not hold any position to keep the check size in the end of process.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: The `mas_size` is removed.  Do we remove the `max_epoch` in this PR?  The `epoch` is implemented when the JWT auth introduced.
The `max_epoch` is not used from my end.  There are some discuss before and I am not sure this is needed also.
https://github.com/MystenLabs/walrus/pull/1345#discussion_r1912479008

The `max_epoch` and `max_size` are different features user can use one of the them.  We can raise error if `max_size` and `size` are set at the same time.  If this is good from you.  We can keep `max_size` and raise error when user set both.

crates/walrus-service/src/client.rs
mlegner: > I feel we are not in the same direction of this topic.

Agreed, I still don't fully understand what the overall plan is. Do you have a design document or some other description of that somewhere?

> Eventually, we will let user upload multiple blobs of a supper big file at the same time, and the size is still by a file not blob. The uploads of blobs can be async. So no mater which upload request is the last one, one of the requests will verified the overall about the size is correct or not. That is why I put this in put PostAction at first.

So there is a single token to authenticate all the requests? Or one token per request?

crates/walrus-service/src/client/daemon/auth.rs
mlegner: > The max_epoch and max_size are different features user can use one of the them. We can raise error if max_size and size are set at the same time. If this is good from you. We can keep max_size and raise error when user set both.

I still don't understand why we need the (exact) size and epoch. If we want to protect the publisher, having max values should be sufficient and provide more flexibility.

crates/walrus-sui/src/client.rs
mlegner: @yanganto I think I'm missing some context or background here:

> Overall, it is about the security, size and hash(blob_id) can identify a file.

Whom are we protecting? The publisher or the user? And from what type of attack?

Could you point me to some design document describing what you want to achieve?

crates/walrus-service/src/client.rs
yanganto: There is no async upload with multiple blobs or any design at this moment.  There is not a single token to authenticate all the requests.  And the size field is from the natural of file, so the number is a fact and never change.  No matter we breakdown a file into blobs to uploading in the future or keep in a single blob at this moment.

What I thought is the size of file should keep in the end, no matter we breakdown and async multiple uploading blobs or use different way in the future, the file size will in the final process (almost the same time with post action) and sum up all the unencoded size from the metadatas and check it in the end.  So we do not need to consider the exact size feature again in the future.  In that time(and the feature will be much stable even we do async uploading), in that time we just consider which requests(one or everyone) we need to pass the size field in to let the check happen.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: Yes, if only consider publisher, it is.  But it protects the all system. 

1. If a hacker hijack a token with exact size, he still can not use it to upload the files in different size as he want.  And the token is short session one, he need time to modify the file to fit the size against with the time.

2. If he can hijack a bigger size token with max-size, he can easier upload things small than it.

So with the exact size, even anyone find a way to hijack the token in the system, the vulnerability almost non-profitable.  And it is profitable with max-size field.  
In other word, if we only keep one, why not exact `size` but a loose `max-size`?

crates/walrus-sui/src/client.rs
yanganto: We do the campaign to pay WAL for user to free upload file via publisher.

So we want the exact user(send_object_to) can upload the exactly file(size) with the exactly storage time(epoch).  
Nothing else or vulnerability can do.

The `send_object_to` and `epoch` are introduced in the auth layer in #1345 and now I want to put file size in this PR.  

The size was considered as pre-calculated `blob-id` in frontend, but it is not reasonable so at least we can do check size.

You can see that project, and you are the only person make star for it. :smile: 
https://github.com/sui-foundation/walrus-storage-nft-drop

crates/walrus-service/src/client/daemon/auth.rs
mlegner: > If a hacker hijack a token with exact size, he still can not use it to upload the files in different size as he want. And the token is short session one, he need time to modify the file to fit the size against with the time.

I understand your point but I don't think it is very relevant. At best, it provides minimal extra protection as an attacker can easily pad a file they want to store to the exact size.

> In other word, if we only keep one, why not exact size but a loose max-size?

Because it makes the system less flexible. With the max size, an honest user could obtain a token and later decide what exact file they want to store (as long as it fits into the max size). 

I don't have a super strong opinion on this topic, but I don't think there are any real benefits in replacing the max size with an exact size.

crates/walrus-sui/src/client.rs
mlegner: > So we want the exact user(send_object_to) can upload the exactly file(size) in the exactly time(epoch).
Nothing else or vulnerability can do.

I fully agree with the first point, but I still don't see why we need an *exact* match for the size and epoch. IMO, it would be much more user-friendly to simply set an upper limit for those.

As mentioned [above](https://github.com/MystenLabs/walrus/pull/1613#discussion_r1989317634), I don't believe that this extra restriction provides any meaningful extra security for anyone.

crates/walrus-service/src/client.rs
mlegner: > What I thought is the size of file should keep in the end, no matter we breakdown and async multiple uploading blobs or use different way in the future, the file size will in the final process (almost the same time with post action) and sum up all the unencoded size from the metadatas and check it in the end.

I'm not sure I understand this approach correctly. If we allow users to upload multiple blobs asynchronously, we should check at any point if the user exceeds their allowance. 

AFAIK, there is no plan to limit the size in the CLI or any other client component, so the check should be performed at the earliest possible time, i.e., when the publisher gets the request (before actually performing any work).

crates/walrus-service/src/client/daemon/auth.rs
yanganto: Yes, we know the padding is also possible to work around it, and it still raise the cost of attacker to handle this and easy to make noise in the log.

The JWT token should be a session token, and it is good to keep it short.
Does a user change his mind to upload a smaller file in the short session time to get the real benefits from the `max-size`?  

Note: the token should use only once(related #1560), else the valid user can use the same token to upload a lot of files in the session time.

mlegner: > The JWT token should be a session token, and it is good to keep it short.
Does a user change his mind to upload a smaller file in the short session time to get the real benefits from the max-size?

Hmm... Probably not too much. I think overall neither has a very large benefit. I'd still prefer the max size but I'm fine with the exact size.

yanganto: Okay, let's keep both, just like `epoch` and `max-epoch`.  Developer can use one of them.

mlegner: > Okay, let's keep both, just like `epoch` and `max-epoch`. Developer can use one of them.

I'm fine with that as well, but then we need to make sure we define the semantics if both are set. I'd suggest to just check both separately and make sure on the issuer that they are consistent or only one of them is used.

mlegner: Suggestion: Can you add a statement like for the epochs above?
```rust
    /// If both `size` and `max_size` are present, this is considered a configuration mistake
    /// and the claim is rejected.
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -108,26 +112,53 @@ impl Claim {
...
+                    );
+                    return Err(PublisherAuthError::MaxSizeExceeded);
+                }
+            }
+            if let Some(size) = self.size {
+                if body_size_upper_hint < size {
+                    tracing::debug!(
+                        size = size,
+                        body_size_upper_hint = body_size_upper_hint,
+                        "upload with body size do not the same as size (upper hint unmacth)"
```
mlegner: Nit:
```suggestion
                        size,
                        body_size_upper_hint,
                        "body size does not match the size specified in the JWT (upper hint mismatch)"
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -108,26 +112,53 @@ impl Claim {
...
+        let body_size_lower_hint = body_size_hint.lower();
+        if let Some(size) = self.size {
+            if body_size_lower_hint > 0 && body_size_lower_hint > size {
                 tracing::debug!(
-                    max_size = max_size,
-                    body_size_hint = body_size_hint,
-                    "upload with body size greater than max_size"
+                    size = size,
+                    body_size_lower_hint = body_size_lower_hint,
+                    "upload with body size do not the same as size (lower hint unmatch)"
```
mlegner: Nit:
```suggestion
                        size,
                        body_size_lower_hint,
                        "body size does not match the size specified in the JWT (lower hint mismatch)"
```

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -313,11 +344,21 @@ pub enum PublisherAuthError {
...
     /// The size of the body is above the maximum allowed.
     #[error("the size of the body is above the maximum allowed.")]
     #[rest_api_error(reason = "MAX_SIZE_EXCEEDED", status = ApiStatusCode::FailedPrecondition)]
     MaxSizeExceeded,
 
+    /// The size of the body is incorrect.
+    #[error("the size of the body is incorrect")]
+    #[rest_api_error(reason = "SIZE_INCORRECT", status = ApiStatusCode::FailedPrecondition)]
+    SizeIncorrect,
+
```
mlegner: Remark: I think this separate error is not needed, we can use the same error for mismatch with the max size and exact size in the claim.

Also: For epochs, an `InvalidEpochs` error is returned in case there is a mismatch with the exact epochs in the claim. This should be changed to the same error as for a mismatch with the max epochs.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -342,17 +342,26 @@ fn check_blob_size(
...
+            if let Some(max_size) = claim.max_size {
+                if blob_size as u64 > max_size {
+                    return Err(PublisherAuthError::MaxSizeExceeded);
+                }
+            }
+            if let Some(size) = claim.size {
+                // Note: The assumption here is upload one file in one blob
+                // If the file is breakdown into more blobs and upload asynchronously in future,
+                // the size check should move to the post action and sum up the unencoded size from
+                // all blobs.
```
mlegner: I do not agree with this statement. If we allow uploading a file as multiple blobs in the future, we should just issue a separate token for each of them.

This is particularly relevant as each blob has an additional overhead. So if the token would only limit the total size, a user could split a file into blobs of 1 byte each and upload them separately, causing a lot of extra work for the system.

Also, it would be very tricky to do bookkeeping inside the publisher.

crates/walrus-service/src/client/daemon/auth.rs
yanganto: No really, it is also for `epoch` and `max-epoch` conflict case.
https://github.com/MystenLabs/walrus/blob/c07ce492f641471d0f9e0a326015e158c939c5e9/crates/walrus-service/src/client/daemon/auth.rs#L190

yanganto: I will use the same error for this case.

mlegner: There should be two errors for both cases:
1. There is a mismatch between the restrictions in the claim. Currently, this is `InvalidSizes` for the sizes.
2. There is a mismatch of the actual query with the claim. Currently, you use two different errors `MaxSizeExceeded` and `SizeIncorrect`. These could be merged IMO.

For the epochs, you use the `InvalidEpochs` for case 1 and (part of) 2, which does not make sense to me.

yanganto: Agree with 2.  I will do it.  

The `InvalidEpochs` and the `max-epoch` are not useful for me, exactly things will be secure in the system.

Overall, I think one error for size is enough, just make the message clear.

crates/walrus-service/src/client/daemon/routes.rs
yanganto: Okay, I see, I will remove this, the size means for the blob specific, for now it is the same for the file.  

If the file need to breakdown uploading async in future, the calculating size of file can be done in the client side.

crates/walrus-sui/src/client.rs
yanganto: Sure.  Well managing the token is much more important.  Let the hijacked token hard to use is just is silver not gold.

crates/walrus-service/src/client/daemon/auth.rs
mlegner: I still think we should distinguish between errors caused by the user (e.g., request size doesn't match the JWT claim) and those caused by the issuer (invalid claim, e.g., because both max and exact size are set). 

If you prefer to just use as single error type, that's fine for me. However, then we should use the same approach for the `send_object_to` and `epochs` parameters, where we currently have two different errors each:

```rust
    /// The epochs field in the query does not match the token.
    #[error("the epochs field in the query does not match the token")]
    #[rest_api_error(reason = "INVALID_EPOCHS", status = ApiStatusCode::FailedPrecondition)]
    InvalidEpochs,

    /// Epochs is above the maximum allowed.
    #[error("the epochs field in the query is above the maximum allowed")]
    #[rest_api_error(reason = "EPOCHS_ABOVE_MAX", status = ApiStatusCode::FailedPrecondition)]
    EpochsAboveMax,

    /// The send_object_to field in the query does not match the token, or is missing.
    #[error("the send_object_to field in the query does not match the token, or is missing")]
    #[rest_api_error(reason = "INVALID_SEND_OBJECT_TO", status = ApiStatusCode::FailedPrecondition)]
    InvalidSendObjectTo,

    /// The send_object_to field is missing from the query, but it is required.
    #[error("the send_object_to field is missing from the query, but it is required")]
    #[rest_api_error(reason = "MISSING_SEND_OBJECT_TO", status = ApiStatusCode::FailedPrecondition)]
    MissingSendObjectTo,
```

Can we just use `Invalid{Epochs,SendObjectTo,Size}` for those three?

PR #1617, opened 2025-02-24
fix: correct Subsidies struct, prevent empty transactions

karlwuest:
## Description

- fixes the Rust version of the `Subsidies` move struct (`version` and `package_id` were swapped)
- returns early for contract operations that take a vector as input and do nothing for empty vectors

## Test plan

- existing tests
- manual tests in local testbed & PTN

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/types/move_structs.rs
+++ b/crates/walrus-sui/src/types/move_structs.rs
@@ -800,10 +800,10 @@ pub struct Subsidies {
     pub system_subsidy_rate: u16,
     /// The total amount of WAL available for subsidies
     pub subsidy_pool: u64,
-    /// The version of the subsidies object
-    pub version: u64,
     /// The package ID of the subsidies contract
     pub package_id: ObjectID,
+    /// The version of the subsidies object
+    pub version: u64,
 }
 
 impl AssociatedContractStruct for Subsidies {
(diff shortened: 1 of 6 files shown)


crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -800,10 +800,10 @@ pub struct Subsidies {
     pub system_subsidy_rate: u16,
     /// The total amount of WAL available for subsidies
     pub subsidy_pool: u64,
-    /// The version of the subsidies object
-    pub version: u64,
     /// The package ID of the subsidies contract
     pub package_id: ObjectID,
+    /// The version of the subsidies object
+    pub version: u64,
```
mlegner, line 806: Question: Is there any automatic way to detect these kinds of bugs. Maybe just integration tests?

karlwuest: There is an integration test for this that for some reason still succeeded and I don't know why. I'm still trying to figure out why the test didn't catch this (and I've also asked @sadhansood if he might have an idea, since he implemented the integration). I must be missing something, because from what I can tell, the test should have failed previously.

In the debug output of the test, the `package_id` always looked normal, whereas it was shifted and "padded" with the version when running it against the local testbed and PTN.

karlwuest: If/when I figure out what we need to change to properly test this, I'll create a follow up.

PR #1619, opened 2025-02-24
fix(client): do not interrupt wallet operations

giac-mysten:
## Description

Ensures that, during a store operation, the sections in which a wallet is used to sign a transaction are never preempted. 

This change solves the issue whereby the transaction would be sent, but not awaited for, possibly creating conflicts in object versions.

The PR modifies the `reserve_and_store_encoded_blobs` function such that:
- only the parts where the blob status is fetched, and where the slivers are sent to the nodes, can be interrupted by the committee change notification
- after all blobs are registered, a check is introduced to ensure that the committee was not changed during this operation.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -351,6 +351,10 @@ impl<T: ReadClient> Client<T> {
     }
 
     /// Retries the given function if the client gets notified that the committees have changed.
+    ///
+    /// This function should not be used to retry function `func` that cannot be interrupted at
+    /// arbitrary await points. Most importantly, functions that use the wallet to sign and execute
+    /// transactions on Sui.
     async fn retry_if_notified_epoch_change<F, R, Fut>(&self, func: F) -> ClientResult<R>
     where
         F: Fn() -> Fut,
@@ -427,7 +431,7 @@ impl<T: ReadClient> Client<T> {
     ///
     /// Returns a [`ClientErrorKind::CommitteeChangeNotified`] error if the client is notified that
     /// the committee has changed.
-    async fn await_while_checking_notification<Fut, R>(&self, func: Fut) -> ClientResult<R>
+    async fn await_while_checking_notification<Fut, R>(&self, future: Fut) -> ClientResult<R>
     where
         Fut: Future<Output = ClientResult<R>>,
     {
@@ -435,7 +439,7 @@ impl<T: ReadClient> Client<T> {
             _ = self.committees_handle.change_notified() => {
                 Err(ClientError::from(ClientErrorKind::CommitteeChangeNotified))
             },
-            result = func => result,
+            result = future => result,
         }
     }
 }


crates/walrus-service/src/client.rs
```diff
@@ -356,55 +351,59 @@ impl<T: ReadClient> Client<T> {
     }
 
     /// Retries the given function if the client gets notified that the committees have changed.
-    async fn retry_if_committees_change<F, R, Fut>(&self, func: F) -> ClientResult<R>
+    async fn retry_if_notified_epoch_change<F, R, Fut>(&self, func: F) -> ClientResult<R>
```
karlwuest, line 358: Suggestion: Expand the comment here to explain that this should not be used on functions that cannot be interrupted at arbitrary await points.

crates/walrus-service/src/client.rs
```diff
@@ -423,6 +422,22 @@ impl<T: ReadClient> Client<T> {
             .await
             .map_err(ClientError::other)
     }
+
+    /// Executes the function while also awaiiting on the change notification.
+    ///
+    /// Returns a [`ClientErrorKind::CommitteeChangeNotified`] error if the client is notified that
+    /// the committee has changed.
+    async fn await_while_checking_notification<Fut, R>(&self, func: Fut) -> ClientResult<R>
```
karlwuest: Nit:
```suggestion
    async fn await_while_checking_notification<Fut, R>(&self, future: Fut) -> ClientResult<R>
```

crates/walrus-service/src/client.rs
```diff
@@ -760,15 +784,30 @@ impl Client<SuiContractClient> {
...
             .collect::<Vec<_>>();
+
+        // Check if the committee has changed while registering the blobs.
+        if are_current_previous_different(
+            committees.as_ref(),
+            self.get_committees().await?.as_ref(),
+        ) {
+            tracing::warn!("committees have changed while registering blobs");
+            return Err(ClientError::from(ClientErrorKind::CommitteeChangeNotified));
+        }
```
halfprice, line 804: Nit: I feel this check can be put into `await_while_checking_notification`.

giac-mysten: ah but the point here is that we don't want to `await_while_checking_notification`, because the operations just above required the wallet.

Further, because of the way `Notify` works, we cannot check if a notification was triggered while we were not awaiting on the `notify` channel. 

Therefore, the only way to check if a notification was triggered in the meantime (while we were not awaiting to ensure the wallet operation is not interrupted) is to get the committee again and check if it is different from what we had before.

giac-mysten: hope this makes sense :)

PR #1621, opened 2025-02-24
test: Select random encoding types in e2e tests.

liquid-helium:
## Description

Related to WAL-608

Contributes to WAL-641

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/test_utils.rs
+++ b/crates/walrus-core/src/test_utils.rs
@@ -6,7 +6,7 @@ use alloc::{vec, vec::Vec};
 use core::num::NonZeroU16;
 
 use fastcrypto::traits::{KeyPair, Signer as _};
-use rand::{rngs::StdRng, RngCore, SeedableRng};
+use rand::{rngs::StdRng, seq::SliceRandom, Rng, RngCore, SeedableRng};
 
 use crate::{
     encoding::{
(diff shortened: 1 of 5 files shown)


crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -92,6 +93,90 @@ async fn test_store_and_read_blob_without_failures(blob_size: usize) {
...
+{
+    // Generate random blobs.
+    let blob_data = walrus_test_utils::random_data_list(data_length, num_blobs);
+    let mut path_to_data: HashMap<PathBuf, Vec<u8>> = HashMap::new();
+    let mut blobs_by_encoding: HashMap<EncodingType, Vec<(PathBuf, Vec<u8>)>> = HashMap::new();
+    let mut path_to_blob_id: HashMap<PathBuf, BlobId> = HashMap::new();
+
+    // For each blob, create multiple encodings and paths.
+    for (i, data) in blob_data.iter().enumerate() {
+        let encodings = random_encoding_types();
```
mlegner, line 119: Question/suggestion: Could we not simply use the exhaustive list of encoding types instead (i.e., `SUPPORTED_ENCODING_TYPES`)?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -179,7 +179,7 @@ pub(super) async fn get_blob_by_object_id<T: WalrusReadClient>(
     match client.get_blob_by_object_id(&blob_object_id).await {
         Ok(BlobWithAttribute { blob, attribute }) => {
             // Get the blob data using the existing get_blob function
-            let mut response = get_blob(
+            let mut response: axum::http::Response<axum::body::Body> = get_blob(
```
mlegner, line 182: Nit: I think this could be reverted:
```suggestion
            let mut response = get_blob(
```

crates/walrus-core/src/test_utils.rs
```diff
@@ -27,13 +27,31 @@ use crate::{
...
+
+    SUPPORTED_ENCODING_TYPES
+        .choose_multiple(&mut rand::thread_rng(), count)
+        .copied()
+        .collect()
+}
+
+/// Randomly selects an encoding type from the supported encoding types.
+pub fn random_encoding_type() -> EncodingType {
+    random_encoding_types()[0]
```
mlegner, line 52: Nit: It feels weird to generate multiple values and then just using one of them. Can't we just generate a single one instead?

liquid-helium: Agreed, change applied.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Oh, thanks for the catch, reverted.

crates/walrus-e2e-tests/tests/test_client.rs
liquid-helium: Nice suggestion! Change applied.

PR #1622, opened 2025-02-24
fix: address new node joining simtest failure

halfprice:
## Description

Adding delay between retries in store operation to account for time discrepancy in processing EpochChangeStart
in different storage nodes.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client.rs
+++ b/crates/walrus-service/src/client.rs
@@ -386,7 +386,7 @@ impl<T: ReadClient> Client<T> {
                         Ok(result) => return Ok(result),
                         // Operation failed but may be retryable due to committee change
                         Err(error) if error.may_be_caused_by_epoch_change() => {
-                            tracing::info!(?attempts,
+                            tracing::debug!(?attempts,
                                 "operation failed; epoch change detected; \
                                 forcing committee refresh");
                             self.force_refresh_committees().await?;


crates/walrus-service/src/client.rs
```diff
@@ -94,7 +95,9 @@ mod multiplexer;
 
 /// The maximum number of retries for an operation that is stopped because of a committee change.
 // TODO: make this configurable.
-const MAX_COMMITTEE_CHANGE_RETRIES: u32 = 3;
+const MAX_COMMITTEE_CHANGE_RETRIES: u32 = 5;
+const COMMITTEE_CHANGE_RETRY_MIN_DELAY: Duration = Duration::from_secs(1);
+const COMMITTEE_CHANGE_RETRY_MAX_DELAY: Duration = Duration::from_secs(5);
```
liquid-helium, line 100: Suggestion: What would occur if we set the backoff duration even longer in `Prod`? This backoff should be designed for specifically during an ongoing epoch change, making it reasonable to retry multiple times or extend the duration.

crates/walrus-service/src/client.rs
```diff
@@ -360,43 +363,54 @@ impl<T: ReadClient> Client<T> {
...
                 }
                 result = func() => {
                     match result {
                         Ok(result) => return Ok(result),
+                        // Operation failed but may be retryable due to committee change
+                        Err(error) if error.may_be_caused_by_epoch_change() => {
+                            tracing::info!(?attempts,
+                                "operation failed; epoch change detected; \
+                                forcing committee refresh");
+                            self.force_refresh_committees().await?;
```
liquid-helium: nit: debug!, since in the worst case in prod, we will see the log for last retry?

crates/walrus-service/src/client.rs
halfprice: Agreed. As the above comment suggested, I think these should be in a config as opposite to fixed values. I can imagine different clients in different part of the world may have different settings.

crates/walrus-service/src/client.rs
halfprice: Done.

crates/walrus-service/src/client.rs
```diff
@@ -94,7 +95,9 @@ mod multiplexer;
 
 /// The maximum number of retries for an operation that is stopped because of a committee change.
 // TODO: make this configurable.
```
mlegner, line 97: I created a follow-up issue to make this configurable: [WAL-637: Make `MAX_COMMITTEE_CHANGE_RETRIES` configurable](https://linear.app/mysten-labs/issue/WAL-637/make-max-committee-change-retries-configurable)

PR #1625, opened 2025-02-25
feat(walrus-backup): implement basic garbage collection

wbbradley:
## Description

Allow the walrus-backup service to clean up expired blobs.

## Test plan

Manual testing locally and in Testnet.
---

No release impact.


crates/walrus-service/src/backup.rs
```diff
@@ -6,6 +6,9 @@
 mod config;
 pub use config::{BackupConfig, BACKUP_BLOB_ARCHIVE_SUBDIR};
 
+#[cfg(feature = "backup")]
+mod garbage;
```
halfprice: Wonder if the mod name may be a bit misleading :)

crates/walrus-service/src/backup/config.rs
```diff
@@ -84,14 +88,20 @@ pub struct BackupConfig {
...
-    /// successful fetches in order to ensure a rapid recovery if there is a queue building up.
+    /// How long to sleep after realizing the fetcher has no work to do.
     #[serde_as(as = "DurationMilliSeconds<u64>")]
     #[serde(
         rename = "idle_fetcher_sleep_time_milliseconds",
         default = "defaults::idle_fetcher_sleep_time"
     )]
     pub idle_fetcher_sleep_time: Duration,
+    /// How long to sleep after realizing the garbage collector has no work to do.
+    #[serde_as(as = "DurationMilliSeconds<u64>")]
```
halfprice: Nit: feel this is in second granularity. Maybe <DurationSeconds<u64>>?

crates/walrus-service/migrations/2025-02-24-213538_add-gc-field/up.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE blob_state
+    ADD COLUMN initiate_gc_after TIMESTAMP WITH TIME ZONE NULL;
+ALTER TABLE blob_state
+    ADD COLUMN size BIGINT NULL;
```
halfprice, line 4: question: are existing rows filling NULL in them?

crates/walrus-service/src/backup/service.rs
```diff
@@ -665,13 +717,19 @@ async fn backup_fetch_inner_core(
                                     initiate_fetch_after = NULL,
                                     retry_count = NULL,
                                     last_error = NULL,
-                                    fetcher_version = $2
+                                    size = $2,
+                                    md5 = $3,
+                                    sha256 = $4,
```
halfprice, line 719: question: why do we need these? Just for the purpose of storing some additional information about the blob?

crates/walrus-service/src/backup/service.rs
```diff
@@ -761,7 +819,7 @@ async fn upload_blob_to_storage(
     Ok(blob_url)
 }
 
-async fn retry_serializable_query<'a, 'b, T, F>(
+pub async fn retry_serializable_query<'a, 'b, T, F>(
```
halfprice: nit: pub(crate)?

crates/walrus-service/src/backup/service.rs
```diff
@@ -352,41 +373,70 @@ fn start_db_metrics_loop(metrics_runtime: &MetricsAndLoggingRuntime, config: &Ba
                 .expect("failed to connect to postgres for db metrics polling");
 
         loop {
-            let stats: Vec<BlobStateStatistic> = diesel::sql_query(
+            let stats: Vec<DbStatistic> = diesel::sql_query(
```
halfprice, line 373: Nit: wonder if you can document the stats listed here to help the reader to understand the query.

crates/walrus-service/src/backup/service.rs
```diff
@@ -352,41 +373,70 @@ fn start_db_metrics_loop(metrics_runtime: &MetricsAndLoggingRuntime, config: &Ba
...
+                    WHERE
+                        state = 'deleted' OR
+                        (state = 'waiting' AND
+                            COALESCE(end_epoch > (SELECT MAX(epoch) FROM epoch_change_start_event),
+                                    TRUE))
+
+                    GROUP BY 1
+                    UNION
+                    SELECT
+                        'garbage' AS name,
```
halfprice, line 398: question: is there a place where we list all the state? I see `waiting`, `archived`, `deleted`, `expired`, and `garbage`. Not sure what's the difference between the last two.

crates/walrus-service/src/backup/garbage.rs
```diff
@@ -0,0 +1,220 @@
...
+    #[diesel(sql_type = diesel::sql_types::Bytea)]
+    blob_id: Vec<u8>,
+}
+
+async fn collect_garbage(
+    config: &BackupConfig,
+    metric_set: &BackupGarbageCollectorMetricSet,
+) -> Result<()> {
+    // Start an infinite loop polling the database for blob state statistics and updating the
+    // metrics.
```
halfprice: Nit: looks like an outdated comments

crates/walrus-service/src/backup/garbage.rs
```diff
@@ -0,0 +1,220 @@
...
+                    }
+                }
+                .scope_boxed()
+            },
+        )
+        .await
+        .unwrap_or_default();
+
+        if garbage_blobs.is_empty() {
+            metric_set.idle_state.set(1.0);
```
halfprice: Suggestion: looks like you try to monitor the entrance and exit of this function, there is a macro, `mysten_metrics::monitored_scope`, which does this and also add some other metrics such as the number of times this block is entered and accumulated time spent in this block.

crates/walrus-service/src/backup/garbage.rs
```diff
@@ -0,0 +1,220 @@
...
+            tracing::info!("no garbage blobs found, sleeping");
+            tokio::time::sleep(config.idle_garbage_collector_sleep_time).await;
+            metric_set.idle_state.set(0.0);
+        } else {
+            for GarbageBlob { blob_id } in garbage_blobs.into_iter() {
+                let blob_id = BlobId::try_from(blob_id.as_slice()).unwrap();
```
halfprice: nit: I think the convention in walrus code repo is to use `expect(reason)` to extract from Result

crates/walrus-service/src/backup/config.rs
wbbradley: Yeah, that's a good idea. Will do.

crates/walrus-service/migrations/2025-02-24-213538_add-gc-field/up.sql
wbbradley: Yes, the implicit default for nullable columns is NULL.

crates/walrus-service/src/backup/service.rs
wbbradley: Yeah, just because it's interesting.

crates/walrus-service/src/backup/service.rs
wbbradley: Good idea, will do.

crates/walrus-service/src/backup/garbage.rs
wbbradley: That's what this function does, so it's not exactly out of date. But, perhaps I should move it down to the loop statement.

crates/walrus-service/src/backup.rs
wbbradley: Renamed to garbage_collector.

crates/walrus-service/src/backup/service.rs
wbbradley: Done.

crates/walrus-service/src/backup.rs
wbbradley: Fixed.

crates/walrus-service/src/backup/garbage.rs
wbbradley: Good call. Will use that in the future as needed.

crates/walrus-service/src/backup/garbage.rs
wbbradley: Fixed.

crates/walrus-service/src/backup/service.rs
wbbradley: The states in the DB are defined in the initial migration of the DB. I will list them out in a comment.

crates/walrus-service/src/backup.rs
```diff
@@ -2,10 +2,45 @@
...
+//!     once the blob has been successfully deleted.
+//!
+//! The `blob_state` table is the primary table used by the backup service. The available states are
+//!
+//!   - `'waiting'`  - the blob is ready to be fetched from the Walrus network and then archived in
+//!     backup storage. Note that there is a race condition between archiving and expiration. This
+//!     is acceptable as the garbage collector will clean up any expired blobs, and once a blob is
+//!     expired, it need not be archived anyway.
+//!   - `'archived'` - the blob has been archived in the backup storage.
+//!   - `'deleted'`  - the blob has been deleted from the backup storage.
```
halfprice, line 31: seems there was an `expired` state. Is it no longer needed? Not sure if it is already used in the existing backup system.

crates/walrus-service/src/backup.rs
```diff
@@ -2,10 +2,45 @@
 // SPDX-License-Identifier: Apache-2.0
 
 //! Walrus blob backup service.
-
+//!
+//! The backup service runs alongside the Walrus network. It should not be relied upon during
```
halfprice, line 6: Appreciate the documentation!

crates/walrus-service/src/backup.rs
wbbradley: There is a metric for "expired" items, but it is not reified in the actual database. It is a computed state.

PR #1628, opened 2025-02-25
chore(publisher): default to storing blob objects

giac-mysten:
## Description

Changes the default post-store behavior of the publisher to keeping the created `Blob` objects in the main publisher wallet, instead of burning them immediately.
Replaces the `--keep` flag with a `--burn-after-store` flag.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [x] Publisher: Changes the default post-store behavior of the publisher to keeping the created `Blob` objects in the main publisher wallet, instead of burning them immediately. Replaces the `--keep` flag with a `--burn-after-store` flag.
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -685,6 +685,17 @@ pub struct PublisherArgs {
     #[clap(long, default_value_t = default::sub_wallets_min_balance())]
     #[serde(default = "default::sub_wallets_min_balance")]
     pub sub_wallets_min_balance: u64,
+    /// Deprecated flag for backwards compatibility.
+    ///
+    /// By default, the publisher already keeps created Blob objects in its main wallet. This flag
+    /// is only available for backwards-compatibility reasons. Therefore _specifying this flag has
+    /// no effect_.
+    ///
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -685,15 +685,15 @@ pub struct PublisherArgs {
...
-    /// note that this flag _does not affect_ the use of the `send_object_to` query parameter:
-    /// Regardless of this flag's status, the publisher will send created objects to the address in
-    /// the `send_object_to` query parameter, if it is specified in the PUT request.
+    /// If unset, the publisher will keep all created blob objects in its _main wallet_ by default.
+    /// However, note that this flag _does not affect_ the use of the `send_object_to` query
+    /// parameter: Regardless of this flag's status, the publisher will send created objects to the
+    /// address in the `send_object_to` query parameter, if it is specified in the PUT request.
     #[clap(long, action)]
     #[serde(default)]
-    pub keep: bool,
```
mlegner, line 696: Suggestion: Can we keep the `--keep` flag (as a no-op) for backwards compatibility?

PR #1629, opened 2025-02-25
chore: update testnet contracts after publishing subsidies

karlwuest:
## Description

- Adds the `subsidies` contract to `testnet-contracts` after publishing
- Cleanup and dependency version bump in `Move.toml` files.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


testnet-contracts/wal/Move.toml
mlegner, line 1: Question: Do the deployed contracts now use the newer version of the `sui-framework`?

karlwuest: They all use the currently deployed version of the `sui-framework` (since that's a special package, there are no updates for the dependencies needed). 
The version bump is necessary because otherwise the local compilation fails if the dependency versions are not the same.

PR #1631, opened 2025-02-25
feat(walrus-sui): Add fallback to direct calls when subsidies contract is outdated

sadhansood:
## Description

When the main Walrus contracts are upgraded but the subsidies contract hasn't been updated yet, calls through the subsidies contract will fail with EWrongVersion errors. This change adds fallback logic to bypass the subsidies 
contract and call the main contracts directly in such cases.

The fallback is triggered specifically when we get an EWrongVersion error from the Walrus package (which the subsidies package depends on), indicating that the Walrus package was upgraded but the subsidies package hasn't been updated 
yet.

However, if the calls to walrus package also fail with EWrongVersion, the packages would be refreshed and contract call to packages retried (first to subsidies and then walrus if subsidies is not updated but walrus is)

closes WAL-639

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -15,6 +15,7 @@ use std::{
 
 use anyhow::{anyhow, Context, Result};
 use contract_config::ContractConfig;
+use futures::future::BoxFuture;
 use move_package::BuildConfig as MoveBuildConfig;
 use retry_client::RetriableSuiClient;
 use sui_package_management::LockCommand;
@@ -33,7 +34,7 @@ use sui_types::{
     base_types::SuiAddress,
     event::EventID,
     programmable_transaction_builder::ProgrammableTransactionBuilder,
-    transaction::{ProgrammableTransaction, TransactionData, TransactionKind},
+    transaction::{Argument, ProgrammableTransaction, TransactionData, TransactionKind},
     TypeTag,
 };
 use tokio::sync::Mutex;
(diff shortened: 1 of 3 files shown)


crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -260,8 +261,24 @@ impl WalrusPtbBuilder {
         let subsidies_package_id = self.read_client.get_subsidies_package_id();
         match subsidies_package_id {
             Some(pkg_id) => {
-                self.reserve_space_with_subsidies(encoded_size, epochs_ahead, pkg_id)
+                match self
```
karlwuest: Important: This cannot be handled in the transaction builder because when the transaction is constructed, we will not see any execution error, so the second branch will never be taken.

This must instead be handled in the client when the transaction is sent and executed (i.e. at the point where you added the refresh for the package IDs in the previous PR if an `EWrongVersion` is returned).

sadhansood: fixed, and thank you for catching this important bug!

crates/walrus-sui/src/client.rs
```diff
@@ -1210,15 +1210,77 @@ impl SuiContractClientInner {
...
+        let subsidies_package_id = self.read_client.get_subsidies_package_id();
+        match subsidies_package_id {
+            Some(pkg_id) => {
+                match self
+                    .reserve_space_with_subsidies(encoded_size, epochs_ahead, pkg_id)
+                    .await
+                {
+                    Ok(arg) => Ok(arg),
+                    Err(SuiClientError::TransactionExecutionError(
+                        MoveExecutionError::Staking(StakingError::EWrongVersion(_))
```
karlwuest: Nit: Here we should not get an error from staking, so this could be removed.

crates/walrus-sui/src/client.rs
```diff
@@ -1287,6 +1349,55 @@ impl SuiContractClientInner {
...
+                        epochs_ahead,
+                        blob_metadata_list.clone(),
+                        persistence,
+                        pkg_id,
+                    )
+                    .await
+                {
+                    Ok(result) => Ok(result),
+                    Err(SuiClientError::TransactionExecutionError(
+                        MoveExecutionError::Staking(StakingError::EWrongVersion(_))
```
karlwuest: Nit: Same as above, this error is unexpected here.

crates/walrus-sui/src/client.rs
```diff
@@ -2021,20 +2176,80 @@ impl SuiContractClientInner {
...
+        let subsidies_package_id = self.read_client.get_subsidies_package_id();
+        match subsidies_package_id {
+            Some(pkg_id) => {
+                match self
+                    .extend_blob_with_subsidies(blob_obj_id, epochs_extended, pkg_id)
+                    .await
+                {
+                    Ok(arg) => Ok(arg),
+                    Err(SuiClientError::TransactionExecutionError(
+                        MoveExecutionError::Staking(StakingError::EWrongVersion(_))
```
karlwuest: Nit: Same as above

crates/walrus-sui/src/client.rs
```diff
@@ -2080,6 +2295,47 @@ impl SuiContractClientInner {
...
+                }
+            }
+            None => {
+                self.certify_and_extend_blobs_without_subsidies(blobs_with_certificates, post_store)
+                    .await
+            }
+        }
+    }
+
+    async fn certify_and_extend_blobs_with_subsidies(
```
karlwuest, line 2481: Suggestion: There is a lot of duplication between this function and the `*_without_subsidies` function. Can some of this be factored out to remove duplication?

crates/walrus-sui/src/client.rs
sadhansood: done

crates/walrus-sui/src/client.rs
sadhansood: done

crates/walrus-sui/src/client.rs
sadhansood: fixed

crates/walrus-sui/src/client.rs
sadhansood: done

PR #1634, opened 2025-02-26
fix: expose JWT Claim

yanganto:
## Description

This PR was opened https://github.com/MystenLabs/walrus/pull/1536 and agreed at https://github.com/MystenLabs/walrus/pull/1536#issuecomment-2639180814.  However this feature was not really merge in, because the related PR did not perform a well cherry pick and implementation was not correct as the original PR.  

Following are the points missing, will check double the JWT layer can be integrated and reuse again.
- The `daemon` is a private mod, the pub Claim in side it does not really reusable.
- The `PublisherQuery` is also related and good to public to let user reuse it.

## Test plan

No additional test needed

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon/routes.rs
+++ b/crates/walrus-service/src/client/daemon/routes.rs
@@ -409,7 +409,7 @@ pub(super) async fn status() -> Response {
     "OK".into_response()
 }
 
-/// The parameters of query to publisher
+/// The query parameters for a publisher.
 #[derive(Debug, Deserialize, IntoParams)]
 pub struct PublisherQuery {
     /// The encoding type to use for the blob.


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -409,8 +409,9 @@ pub(super) async fn status() -> Response {
     "OK".into_response()
 }
 
+/// The parameters of query to publisher
 #[derive(Debug, Deserialize, IntoParams)]
-pub(crate) struct PublisherQuery {
+pub struct PublisherQuery {
```
mlegner, line 414: Question: (Why) does this have to be public?

crates/walrus-service/src/client/daemon/auth.rs
```diff
@@ -63,6 +63,7 @@ pub struct Claim {
 }
 
 impl Claim {
+    /// Build Claim from JWT token
```
mlegner: Nit:
```suggestion
    /// Builds the Claim from a JWT token.
```

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -409,8 +409,9 @@ pub(super) async fn status() -> Response {
     "OK".into_response()
 }
 
+/// The parameters of query to publisher
```
mlegner: Nit:
```suggestion
/// The query parameters for a publisher.
```

crates/walrus-service/src/client/daemon/routes.rs
yanganto: Because this function is public
https://github.com/MystenLabs/walrus/blob/e6f3bc2aeea69991457e60ce93cb9419664e2e09/crates/walrus-service/src/client/daemon/auth.rs#L107-L112

It is possible to let developer to have a dry run to make sure the claim is good before issuing, so this function is public.  Actually, I do not dry run before issuing a JWTToken, and it is good for me to make these private.

PR #1635, opened 2025-02-26
feat: automatically select encoding type based on system version

karlwuest:
## Description

- Automatically selects the default encoding type based on the version of the system object

closes WAL-642

## Test plan

- Manually tested, correctly uses RS2 on PTN and RaptorQ on testnet

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/cli/runner.rs
+++ b/crates/walrus-service/src/client/cli/runner.rs
@@ -773,7 +773,6 @@ impl ClientCommandRunner {
                 (n_shards, encoding_type)
             } else {
                 let config = self.config?;
-                tracing::debug!("reading `n_shards` from chain");
                 let sui_read_client = get_sui_read_client_from_rpc_node_or_wallet(
                     &config,
                     rpc_url,
@@ -784,6 +783,7 @@ impl ClientCommandRunner {
                 let n_shards = if let Some(n_shards) = n_shards {
                     n_shards
                 } else {
+                    tracing::debug!("reading `n_shards` from chain");
                     sui_read_client.current_committee().await?.n_shards()
                 };
                 let encoding_type = encoding_type_or_default_for_version(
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -510,11 +510,13 @@ impl ClientCommandRunner {
...
         post_store: PostStoreAction,
-        encoding_type: EncodingType,
+        encoding_type: Option<EncodingType>,
     ) -> Result<()> {
         epoch_arg.exactly_one_is_some()?;
-        if !encoding_type.is_supported() {
-            anyhow::bail!(ClientErrorKind::UnsupportedEncodingType(encoding_type));
+        if encoding_type.is_some_and(|encoding| !encoding.is_supported()) {
+            anyhow::bail!(ClientErrorKind::UnsupportedEncodingType(
+                encoding_type.expect("just checked that option is Some")
```
mlegner, line 518: ~~Nit: Why not use `encoding` here?~~
```suggestion
                encoding
```

Edit: Ah, that doesn't work, this is not inside the closure... 🤦

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -756,22 +766,32 @@ impl ClientCommandRunner {
...
-            )
-            .await?;
-            sui_read_client.current_committee().await?.n_shards()
-        };
+        let (n_shards, encoding_type) =
+            if let (Some(n_shards), Some(encoding_type)) = (n_shards, encoding_type) {
+                (n_shards, encoding_type)
+            } else {
+                let config = self.config?;
+                tracing::debug!("reading `n_shards` from chain");
```
mlegner: Nit: This should be moved into the `else` block below where the `n_shards` is actually read.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -1140,3 +1164,17 @@ pub fn ask_for_confirmation() -> Result<bool> {
     std::io::stdin().read_line(&mut input)?;
     Ok(input.trim().to_lowercase().starts_with('y'))
 }
+
+// TODO(WAL-621): Remove in refactor for mainnet
```
mlegner: Suggestion: Can we have a separate issue for this? WAL-621 is otherwise optional.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -1140,3 +1164,17 @@ pub fn ask_for_confirmation() -> Result<bool> {
...
+fn encoding_type_or_default_for_version(
```
mlegner, line 1169: Suggestion: Can we add a DEBUG log here?

crates/walrus-service/src/client/cli/runner.rs
karlwuest: Sure

karlwuest: -> https://linear.app/mysten-labs/issue/WAL-647/use-rs2-only-for-mainnet-release

PR #1636, opened 2025-02-26
fix(contracts): create-or-update flow for metadata

giac-mysten:
## Description

Simplifies the metadata update flow by adding funcitons that do not abort if the metatada is already present.

## Test plan

Added new tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -629,6 +629,7 @@ impl Drop for SimStorageNodeHandle {
 /// See function level documentation for details on the various configuration settings.
 #[derive(Debug)]
 pub struct StorageNodeHandleBuilder {
+    name: Option<String>,
     storage: Option<WithTempDir<Storage>>,
     blocklist_path: Option<PathBuf>,
     event_provider: Box<dyn SystemEventProvider>,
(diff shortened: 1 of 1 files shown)


contracts/walrus/sources/system/blob.move
```diff
@@ -309,6 +309,19 @@ public fun add_metadata(self: &mut Blob, metadata: Metadata) {
     dynamic_field::add(&mut self.id, METADATA_DF, metadata)
 }
 
+/// Adds the metadata dynamic field to the Blob, replacing the existing metadata if present.
+///
+/// Optionally returns the replaced metadata.
```
karlwuest: Nit:
```suggestion
/// Returns the replaced metadata if present.
```

contracts/walrus/sources/system/blob.move
```diff
@@ -339,6 +363,17 @@ public fun remove_metadata_pair(self: &mut Blob, key: &String): (String, String)
     self.metadata().remove(key)
 }
 
+/// Removes the metadata associated with the given key, if it exists.
+///
+/// Optionally returns the previous value associated with the key.
```
karlwuest: Nit:
```suggestion
/// Removes and returns the metadata associated with the given key, if it exists.
```

PR #1637, opened 2025-02-26
fix: address audit results

karlwuest:
## Description

Below are the issues/suggestions mentioned by ottersec and how they are addressed in this PR.
closes WAL-645

**Best reviewed by commit. The first commit covers everything except point 2., which is covered by the second commit (and quite noisy due to necessary changes in many tests).**

 -> fixed by performing calculation with u128 and then taking the minimum with u64::MAX

> 2. the share_amount in request_withdraw_stake can be zero, meaning that the pool's ratio can be easily manipulated by withdrawing small principal.

-> added a check to ensure that the share amount is non-zero, and added a minimum stake per StakedWal to otherwise reduce rounding errors (in af583a96aab8a181d530573d671f3d03df83c42d)

> 3. convert_to_wal_amount will abort if share_amount == 0. therefore, it is possible to cause Denial of Service in few places, including epoch advancement and stake withdrawal.

 -> fixed by also returning a flat exchange rate if share_amount == 0

> 4. staking_pool::advance_epoch uses new commission_rate when calculating the commission for the epoch. we believe that the old commission rate should be used instead

 -> fixed by setting the new commission rate after splitting off the commission

### Suggestions

> 1. the current implementation of staking::withdraw_node will always abort, as both functions call set_withdrawing

 -> was not used, removed

> 2. we suggest checking if old_epoch_used_capacity > deny_list_size in advance_epoch in order to prevent underflow

-> fixed by setting deny_list_size to deny_list_size.min(old_epoch_used_capacity)

> 3. active_set::insert_or_update can implicitly remove an entry. we suggest to handle the return value

-> added a comment to the docstring to explain that a node can be removed, the return value is not relevant for any of the current call sites

> 4. it seems that staking_pool::destroy_empty checks pending_stakes twice, once in pool.is_empty and second time at the end of the function

-> removed

> 5. the pending_values::reduce function could prevent an underflow by validating the value

-> added check and explicit error code

> 6. if EpochState == NextParamsSelected deposits and withdrawals will be activated on current_epoch+2. therefore we believe that the set_next_commission should be activated in current_epoch+3 to allow all users to react.

-> Not an issue, users can react by directly withdrawing the stake until the committee for epoch `current_epoch+2` has been selected, so there is always sufficient time.

-> done

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/staking/staked_wal.move
+++ b/contracts/walrus/sources/staking/staked_wal.move
@@ -13,6 +13,8 @@ use sui::balance::Balance;
 use wal::wal::WAL;
 use walrus::walrus_context::WalrusContext;
 
+// Keep in sync with corresponding value in
+// `crates/walrus-sui/src/client.rs`
 /// StakedWal objects must have a principal with at least this amount.
 const MIN_STAKING_THRESHOLD: u64 = 1_000_000_000; // 1 WAL
 
--- a/crates/walrus-service/src/client/error.rs
+++ b/crates/walrus-service/src/client/error.rs
@@ -5,7 +5,7 @@
 
 use walrus_core::{BlobId, EncodingType, Epoch, SliverPairIndex, SliverType};
 use walrus_sdk::error::{ClientBuildError, NodeError};
-use walrus_sui::client::SuiClientError;
+use walrus_sui::client::{SuiClientError, MIN_STAKING_THRESHOLD};
 
 /// Storing the metadata and the set of sliver pairs onto the storage node, and retrieving the
 /// storage confirmation, failed.
@@ -93,6 +93,9 @@ impl From<SuiClientError> for ClientError {
         let kind = match value {
             SuiClientError::NoCompatibleWalCoins => ClientErrorKind::NoCompatiblePaymentCoin,
             SuiClientError::NoCompatibleGasCoins => ClientErrorKind::NoCompatibleGasCoins,
+            SuiClientError::StakeBelowThreshold(amount) => {
+                ClientErrorKind::StakeBelowThreshold(amount)
+            }
             error => ClientErrorKind::Other(error.into()),
         };
         Self { kind }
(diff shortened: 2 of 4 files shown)


contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -480,7 +482,7 @@ public(package) fun process_pending_stake(pool: &mut StakingPool, wctx: &WalrusC
     // don't forget to flush the early withdrawals since we worked on a copy
     let _ = pool.pre_active_withdrawals.flush(current_epoch);
 
-    let shares_withdraw = pool.pending_shares_withdraw.flush(wctx.epoch());
+    let shares_withdraw = pool.pending_shares_withdraw.flush(current_epoch);
```
damirka, line 485: It is the same value, right? `let current_epoch = wctx.epoch();` ?

damirka: I am clarifying because we have off-by-one in Sui System in a similar function, and there `current_epoch` and `ctx.epoch()` are different, and it led to a lot of confusion.

karlwuest: Yes, it's the same value.

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -587,9 +588,6 @@ public(package) fun destroy_empty(pool: StakingPool) {
     commission.destroy_zero();
     rewards_pool.destroy_zero();
     extra_fields.destroy_empty();
-
-    let (_epochs, pending_stakes) = pending_stake.unwrap().into_keys_values();
```
damirka, line 591: Took me a while to figure this out, but just to confirm, we already check that the pool is empty, and this is why we can ignore these asserts?

contracts/walrus/sources/staking/staking_pool.move
```diff
@@ -338,6 +338,8 @@ public(package) fun request_withdraw_stake(
         .exchange_rate_at_epoch(staked_wal.activation_epoch())
         .convert_to_share_amount(principal_amount);
 
+    assert!(share_amount != 0, EZeroShares);
```
damirka, line 341: What are the conditions for it to be 0?

contracts/walrus/sources/staking/staking_pool.move
karlwuest: Yes (see suggestion 4 in the PR description).

contracts/walrus/sources/staking/staking_pool.move
karlwuest: This will be zero only if the principal is very small. 
With the minimum stake of 1 WAL, this should not happen for the foreseeable future (assuming a very high APY of 10%, for at least ~200 years, at that point the share price might cause high enough rounding errors again).

PR #1639, opened 2025-02-26
feat(staking): adds Sui Object Display

damirka:
## Description

Adds Sui Object Display implementation to the Walrus package.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/display.move
+++ b/contracts/walrus/sources/display.move
@@ -50,7 +50,7 @@ fun init_blob_display(p: &Publisher, ctx: &mut TxContext): Display<Blob> {
         // editorconfig-checker-disable-next-line
         b"Registered: {registered_epoch}; certified: {certified_epoch}; deletable: {deletable}".to_string(),
     );
-    d.add(b"image_url".to_string(), b"https://www.walrus.xyz/walrus-blob.svg".to_string());
+    d.add(b"image_url".to_string(), b"https://www.walrus.xyz/walrus-blob".to_string());
     d.add(b"project_url".to_string(), b"https://walrus.xyz/".to_string());
     d.add(b"link".to_string(), b"".to_string());
     d.update_version();
@@ -63,7 +63,7 @@ fun init_storage_display(p: &Publisher, ctx: &mut TxContext): Display<Storage> {
 
     d.add(b"name".to_string(), b"Walrus Storage Resource ({storage_size}b)".to_string());
     d.add(b"description".to_string(), b"Start: {start_epoch}; end: {end_epoch}".to_string());
-    d.add(b"image_url".to_string(), b"https://www.walrus.xyz/walrus-storage.svg".to_string());
+    d.add(b"image_url".to_string(), b"https://www.walrus.xyz/walrus-storage".to_string());
     d.add(b"project_url".to_string(), b"https://walrus.xyz/".to_string());
     d.add(b"link".to_string(), b"".to_string());
     d.update_version();
@@ -79,7 +79,7 @@ fun init_staked_wal_display(p: &Publisher, ctx: &mut TxContext): Display<StakedW
         b"description".to_string(),
         b"Staked for node: {node_id}, activates at: {activation_epoch}".to_string(),
     );
-    d.add(b"image_url".to_string(), b"https://www.walrus.xyz/walrus-stakedwal.svg".to_string());
+    d.add(b"image_url".to_string(), b"https://www.walrus.xyz/walrus-stakedwal".to_string());
     d.add(b"project_url".to_string(), b"https://walrus.xyz/".to_string());
     d.add(b"link".to_string(), b"".to_string());
     d.update_version();


contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -122,10 +125,13 @@ public struct StakingInnerV1 has store {
     /// the order of the nodes in the `next_committee`. The value is set in the `select_committee`
     /// function and consumed in the `next_bls_committee` function.
     next_epoch_public_keys: ExtendedField<VecMap<ID, Element<UncompressedG1>>>,
+    /// Stores `Publisher` and `Display` objects.
+    object_display: ObjectDisplay,
```
karlwuest: Let's maybe add this as a dynamic field. Otherwise, we lose compatibility with the struct definitions on testnet and we will have to add compile time flags to the rust code again (+ I'm not sure if partners may already rely on the struct format).

karlwuest: For the init cap it's fine to break compatibility, because we never deserialize it in rust and it's only used during deployment.

damirka: The only issue is that StakingInnerV1 doesn't have a UID to attach to, so there isn't a way for us to achieve that. What do you think about attaching it to the `Staking` object instead?

damirka: Eventually we will introduce Display Registry, and these definitions will migrate there. So I like the idea of having it as a dynamic field which doesn't leave trace in type signatures.

karlwuest: Actually, that might be a bit risky at this point in case partners just fetch all dynamic fields to get the staking_inner and expect there to be exactly one dynamic field. Since we probably cannot test this on testnet, that might not be the best solution. I'm worried in particular about breaking staking for BitGo.

Are there other possibilities that seem reasonable? We could also attach it to the `UpgradeManager` or the `EmergencyUpgradeCap` or just turn it into a shared object?

damirka: Let's do a separate object with potential to deprecate this module almost entirely and wipe the shared object when DisplayRegistry is created later this year. I like that. We'd only need to track the object ID, which is extra off-chain overhead we should be able to tolerate.

damirka: @karlwuest updated the PR to reflect it.

contracts/walrus/sources/init.move
```diff
@@ -14,17 +14,22 @@ const EInvalidMigration: u64 = 0;
...
 
 /// Init function, creates an init cap and transfers it to the sender.
 /// This allows the sender to call the function to actually initialize the system
 /// with the corresponding parameters. Once that function is called, the cap is destroyed.
-fun init(ctx: &mut TxContext) {
+fun init(otw: INIT, ctx: &mut TxContext) {
     let id = object::new(ctx);
-    let init_cap = InitCap { id };
+    let publisher = package::claim(otw, ctx);
+    let init_cap = InitCap { id, publisher };
```
karlwuest, line 32: Suggestion: Now that the ObjectDisplay gets shared, we could create it directly here and don't need to wrap the publisher in the InitCap anymore.

contracts/walrus/sources/display.move
```diff
@@ -0,0 +1,89 @@
...
+fun init_blob_display(p: &Publisher, ctx: &mut TxContext): Display<Blob> {
+    let mut d = display::new(p, ctx);
+
+    d.add(b"name".to_string(), b"Walrus Blob ({size}b)".to_string());
+    d.add(
+        b"description".to_string(),
+        b"Registered: {registered_epoch}; certified: {certified_epoch}; deletable: {deletable}".to_string(),
+    );
+    d.add(b"image_url".to_string(), b"".to_string());
+    d.add(b"link".to_string(), b"".to_string());
```
karlwuest: Question: should we add a thumbnail url as well? 
Also, we can probably skip the `link` until we have a `walrus://` protocol (with e.g. a browser extension).

damirka: The idea behind `link` field is being able to link a resource / webpage where this object can be viewed. Say, if you have a SuiFren NFT, a link could point to `suifrens.com/fren/{id}`. So if there was an official "blob viewer" or anything, it would map perfectly.

contracts/walrus/sources/init.move
damirka: Perhaps, I kinda liked the idea of creating all shared objects in one place though: staking / system / display. If you insist, let's remove the field and create it directly.

karlwuest: Ok, let's keep it as is, then.

contracts/walrus/sources/display.move
karlwuest: I'm a bit hesitant of hardcoding an aggregator url for this (because we also don't want to direct everyone to our aggregator). I would probably just skip it for now and eventually we can have a `walrus:://` link.

PR #1642, opened 2025-02-26
fix: ensure consensus on event blob checkpoint sequence numbers

sadhansood:
## Description

Previously, event blob certification only tracked consensus based on blob IDs, allowing the last attesting node to set an arbitrary checkpoint sequence number that would become the official sequence number for the certified blob. This
could lead to inconsistencies in the event blob timeline.

The fix modifies the event blob certification process to require consensus on both the blob ID and checkpoint sequence number by:
- Tracking attestations using a composite key of (blob_id, checkpoint_seq_num)
- Aggregating weights specific to each blob+checkpoint combination
- Only certifying when quorum is reached for a specific checkpoint sequence number and blob id

This ensures that nodes must agree on both the blob content (via blob ID) and its position in the checkpoint sequence before certification can occur.

## Test plan

Existing tests

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/walrus/sources/system/event_blob.move
+++ b/contracts/walrus/sources/system/event_blob.move
@@ -32,7 +32,7 @@ public struct EventBlobCertificationState has store {
     /// Latest certified event blob.
     latest_certified_blob: Option<EventBlob>,
     /// Current event blob being attested.
-    aggregate_weight_per_blob: VecMap<u256, u16>,
+    aggregate_weight_per_blob: VecMap<EventBlob, u16>,
 }
 
 // === Accessors related to event blob attestation ===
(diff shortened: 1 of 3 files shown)


contracts/walrus/Move.lock
```diff
@@ -30,6 +30,6 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.43.1"
+compiler-version = "1.43.0"
```
karlwuest: Nit: Keep the higher version here.

PR #1643, opened 2025-02-26
chore: add a counter to the download loop for event blobs

wbbradley:
## Description

Add the counter event_processor_catchup_event_blobs to track progress during the catchup phase of the EventProcessor. Also note that this change moves the starting of the MetricPushRuntime to be before the start of the EventProcessorRuntime to enable metrics events to be emitted during catchup.

## Test plan

Manual testing in Testnet on the backup service.

No release impact.

### DIFF
--- a/crates/walrus-service/src/backup/metrics.rs
+++ b/crates/walrus-service/src/backup/metrics.rs
@@ -7,7 +7,7 @@ use crate::common::telemetry;
 
 telemetry::define_metric_set! {
     /// Metrics exported by the backup fetcher node.
-    struct BackupFetcherMetricSet {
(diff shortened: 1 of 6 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -560,6 +547,19 @@ mod commands {
             None => None,
         };
 
+        let (event_manager, event_processor_runtime) = EventProcessorRuntime::start(
```
wbbradley, line 550: I moved the EventProcessorRuntime to _after_ initializing the MetricPushRuntime so that metrics during `start` are reported.

PR #1646, opened 2025-02-27
fix: improve the store cli output.

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/cli/cli_output.rs
+++ b/crates/walrus-service/src/client/cli/cli_output.rs
@@ -94,11 +94,24 @@ impl CliOutput for Vec<BlobStoreResultWithPath> {
             }
         }
 
-        println!(
-            "Summary for Modified or Created Blobs ({} newly certified, {} extended)",
-            total_count - reuse_and_extend_count,
-            reuse_and_extend_count
-        );
+        let mut parts = Vec::new();
+        if total_count - reuse_and_extend_count > 0 {
+            parts.push(format!(
+                "{} newly certified",
+                total_count - reuse_and_extend_count
+            ));
+        }
+        if reuse_and_extend_count > 0 {
+            parts.push(format!("{} extended", reuse_and_extend_count));
+        }
+
+        let summary = if !parts.is_empty() {
+            format!("({})", parts.join(", "))
+        } else {
+            String::new()
+        };
+
+        println!("Summary for Modified or Created Blobs {}", summary);
         println!(
             "Total encoded size: {}",
             HumanReadableBytes(total_encoded_size)


crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -93,7 +93,7 @@ impl CliOutput for Vec<BlobStoreResultWithPath> {
                 _ => 0,
             })
             .sum::<u64>();
-        println!("Summary for Newly Created Blobs");
+        println!("Summary for Modified or Created Blobs");
```
yanganto: Will this better to have a number to know how many blob involved and how may blob are newly created?

example:
`Summary for Modified or Created Blobs (5/5 Newly Created)`

yanganto: By the way, it will be nice to show `5/5 Modified` than `0/5  Newly Created`, if there is nothing newly created in my view, because user may wonder what is it when he get `0/5`, and just for a better UX, if you agree with this.  And `0/5  Newly Created` still fine if you insist to use the same format which are good for a parser.

mlegner: > Will this better to have a number to know how many blob involved and how may blob are newly created?

Is there a concrete use case for this? It's not 100% clear where we draw the line between "modified" and "created", is this about registration or certification? So, if we do make this more fine-grained, I'd list *all* the the different operations separately, but this may overload the output a bit. Maybe just having the total number would be a reasonable compromise?

> which are good for a parser

The CLI output is not intended to be parsed automatically and is also not stable. If you need to parse output automatically, use the JSON API or at least output in JSON format with the `--json` option.

liquid-helium: Agreed it is not so clear how to define "NewlyCreated". If we really go with more fine-grained, maybe the certification is an option to divide them, all previously certified are displayed as "modified", all others "created"?

mlegner: > If we really go with more fine-grained, maybe the certification is an option to divide them, all previously certified are displayed as "modified", all others "created"?

Works for me, but then I would probably call them "newly certified" and "extended". The "newly certified" would include all variants except the pure extension.

liquid-helium: Sounds great! @yanganto I will go with this for now, let know your thoughts.

yanganto: Yes, the terms are not really mater for user, and blob numbers will be.  

Users will like to know how much they pay for the storage, and check it to make sure it is reasonable.  The term is not really matter in the real case, and how many blob involved is.

As an engineer I like to use the same term to avoid mapping things if I need to do parser, but this is my personal experience on this.
So "modified"/"extended"/"newly certified" are already good enough for the user.

PR #1651, opened 2025-02-27
refactor: allow cancelling the wait for a permit during recovery and parallelize cancellation

jpcsmith:
## Description

Fixes a bug where it was not possible to cancel a blob recovery that is awaiting a permit to perform the recovery. This meant that waiting for the cancellation of expired blobs to complete would require also waiting for each blob to first acquire a semaphore.

Additionally, this uses the `rayon` crate to parallelise the checks of whether a blob is certified or not. While it's currently unclear whether the former bug, or the latter serial implementation was responsible for the slow handling of epoch change events, there is at least some evidence that the slowdown was happening even when no recoveries were cancelled (which would invalidate the bug from being the issue). 

This also adds a bit of info-logging around the cancellation during epoch-change, which should highlight any long delays when during cancellation on epoch-change.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/blob_sync.rs
+++ b/crates/walrus-service/src/node/blob_sync.rs
@@ -16,6 +16,7 @@ use futures::{
     TryFutureExt,
 };
 use mysten_metrics::{GaugeGuard, GaugeGuardFutureExt};
+use rayon::prelude::*;
 use tokio::{
     sync::{Notify, Semaphore},
     task::{JoinHandle, JoinSet},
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -16,6 +16,7 @@ use futures::{
     TryFutureExt,
 };
 use mysten_metrics::{GaugeGuard, GaugeGuardFutureExt};
+use rayon::prelude::*;
```
mlegner, line 19: Question: AFAIU, this just imports the relevant traits, correct?

jpcsmith: Yes, correct.

PR #1653, opened 2025-02-28
fix: shard move out and back in immediately

halfprice:
## Description

This PR fixes a bug exposed by simtest that when a shard moves out from node in epoch X, and moves back in
epoch X+1, the shard sync for that shard currently does not start properly and stay in `locked` mode, which
caused unavailability of data in that shard.

More specifically, before this PR, when calculating the [ShardDiff](https://github.com/MystenLabs/walrus/blob/c270549ffea21bd8f0bc673b861d6cfeb57a6f04/crates/walrus-service/src/node.rs#L1391), it also takes existing shard into consideration mainly for the purpose of finding out
shards that need to be removed, but in this case, the moved out shard still exists, but in locked mode,
so it does not consider that shard is moved in. So the shard remains in the locked mode and won't be unavailable
to serve data. In the test, this caused store to fail since writing to this shard just fail constantly.

There are two key fixes in this PR:
- Revisit the ShardDiff logic to make it more clear about it's usage when during normal epoch change vs. during full node recovery. In the former, shard diff between previous and current epoch assignments matter, and in the latter, what is currently assigned matter.
- For moved in shard, no matter it's current status, we need to start shard sync for it.

## Test plan

Unit test to reproduce the issue
Unit test to test ShardDiffCalculator
Integration test `test_repeated_shard_move_with_workload` which originally exposed the issue.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -1907,7 +1907,7 @@ async fn test_blob_attribute_fields_operations() -> TestResult {
 async fn test_shard_move_out_and_back_in_immediately() -> TestResult {
     telemetry_subscribers::init_for_testing();
     let (_sui_cluster_handle, walrus_cluster, client) =
-        test_cluster::default_setup_with_epoch_duration_generic::<StorageNodeHandle>(
+        test_cluster::default_setup_with_num_checkpoints_generic::<StorageNodeHandle>(
             Duration::from_secs(20),
             TestNodesConfig {
                 node_weights: vec![1, 1],
--- a/crates/walrus-sui/src/test_utils.rs
+++ b/crates/walrus-sui/src/test_utils.rs
@@ -34,13 +34,6 @@ use tokio::sync::mpsc;
 use tokio::sync::Mutex;
 use walrus_core::{
     keys::{NetworkKeyPair, ProtocolKeyPair},
-    messages::{
-        BlobPersistenceType,
-        Confirmation,
-        ConfirmationCertificate,
-        InvalidBlobCertificate,
-        InvalidBlobIdMsg,
-    },
     BlobId,
     DEFAULT_ENCODING,
 };


crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -1784,6 +1791,109 @@ async fn test_blob_attribute_fields_operations() -> TestResult {
     Ok(())
 }
 
+#[ignore = "ignore E2E tests by default"]
+#[walrus_simtest]
+async fn test_shard_move_out_and_back_in_immediately() -> TestResult {
```
halfprice, line 1907: This test can manifest the bug 100% without the fix.

crates/walrus-service/src/common/utils.rs
```diff
@@ -507,55 +507,71 @@ async fn push_metrics(
...
-    /// identified by the provided public key.
-    /// `exist` is the list of shards that the node currently holds, which is used to find out
-    /// the shards that are no longer needed in the node and can be removed.
-    pub fn diff_previous(
-        committees: &ActiveCommittees,
-        exist: &[ShardIndex],
-        id: &PublicKey,
-    ) -> ShardDiff {
-        let from: &[ShardIndex] = committees
+pub(crate) struct ShardDiffCalculator {
```
halfprice, line 514: The main motivation for this change is that, the old `ShardDiff` was used in 2 places: (1) calculating shard gain/lose in normal epoch change, and (2) calculating shards gain/lose in full recovery. However, the semantic for these two cases are different. So this change just makes it more explicit what we are calculating.

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -259,17 +259,9 @@ impl ShardSyncHandler {
             return Ok(());
         }
 
-        // Validate shard status
-        if shard_status != ShardStatus::None {
-            return Err(SyncShardClientError::InvalidShardStatusToSync(
-                shard_index,
-                shard_status,
-            ));
-        }
```
halfprice, line 268: This code removal is the core of the bug fix: we need to start the shard sync not just for moved in shards, but also for already existing shards in locked mode, which is also move in.

crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -280,6 +281,22 @@ pub struct StorageNode {
...
+            node_id: ObjectID::random(),
+            network_address: NetworkAddress("127.0.0.1:8080".to_string()),
+            public_key: ProtocolKeyPair::generate().public().clone(),
+            next_epoch_public_key: None,
+            network_public_key: NetworkKeyPair::generate().public().clone(),
+            metadata: ObjectID::random(),
+            shard_ids: vec![],
+        }
+    }
+}
```
jpcsmith: Question: Don't we already have a function for this in one of the test-utils modules?

Important: This should not be unconditionally public/compiled into the crate.

crates/walrus-service/src/node/storage.rs
```diff
@@ -628,6 +628,22 @@ impl Storage {
...
+        self.shards
+            .read()
+            .expect("Should acquire the lock successfully")
+            .values()
+            .filter(|&shard_storage| {
+                shard_storage
+                    .status()
+                    .is_ok_and(|status| status.is_owned_by_node())
+            })
+            .map(|shard_storage| shard_storage.id())
```
jpcsmith: Nit: You can use filter map here.
```suggestion
            .filter_map(|&shard_storage| {
                shard_storage
                    .status()
                    .is_ok_and(|status| status.is_owned_by_node())
                    .then_some(shard_storage.id())
            })
```

crates/walrus-service/src/common/utils.rs
```diff
@@ -507,55 +507,71 @@ async fn push_metrics(
...
+    shards_assigned_prev_epoch: Vec<ShardIndex>,
+    /// The list of shards that are assigned to the node in current epoch.
+    shards_assigned_current_epoch: Vec<ShardIndex>,
+    /// The list of shards that the node currently holds. In includes shards in all status.
+    shards_exist: Vec<ShardIndex>,
+}
+
+impl ShardDiffCalculator {
+    /// Create a new `ShardDiffCalculator` for a storage node.
+    pub fn new(committees: &ActiveCommittees, id: &PublicKey, shards_exist: &[ShardIndex]) -> Self {
```
jpcsmith, line 528: Suggestion: Since you're already refactoring this, please change this to using the node's ID instead of the public key.

crates/walrus-service/src/common/utils.rs
```diff
@@ -507,55 +507,71 @@ async fn push_metrics(
...
+        let shards_assigned_prev_epoch: Vec<ShardIndex> = committees
             .previous_committee()
-            .map_or(&[], |committee| committee.shards_for_node_public_key(id));
-        let to = committees
+            .map_or_else(Vec::new, |committee| {
+                committee.shards_for_node_public_key(id).to_vec()
+            });
```
jpcsmith: Nit: `.map(..).unwrap_or_default()` may be clearer.

crates/walrus-service/src/common/utils.rs
```diff
@@ -507,55 +507,71 @@ async fn push_metrics(
...
+    /// The list of shards that are assigned to the node in current epoch.
+    shards_assigned_current_epoch: Vec<ShardIndex>,
+    /// The list of shards that the node currently holds. In includes shards in all status.
+    shards_exist: Vec<ShardIndex>,
+}
+
+impl ShardDiffCalculator {
+    /// Create a new `ShardDiffCalculator` for a storage node.
+    pub fn new(committees: &ActiveCommittees, id: &PublicKey, shards_exist: &[ShardIndex]) -> Self {
+        let shards_assigned_prev_epoch: Vec<ShardIndex> = committees
```
jpcsmith:
```suggestion
        let shards_assigned_prev_epoch = committees
```

crates/walrus-service/src/common/utils.rs
```diff
@@ -507,55 +507,71 @@ async fn push_metrics(
...
+            shards_assigned_prev_epoch,
+            shards_assigned_current_epoch,
+            shards_exist: shards_exist.to_vec(),
         }
     }
+
+    /// Returns the list of shards that are assigned to the node in current epoch serving data.
+    pub fn all_owned_shards(&self) -> Vec<ShardIndex> {
+        self.shards_assigned_current_epoch.clone()
+    }
```
jpcsmith: Nit: allow the caller to determine if they need a copy.

```suggestion
    pub fn all_owned_shards(&self) -> &[ShardIndex] {
        &self.shards_assigned_current_epoch
    }
```

crates/walrus-service/src/common/utils.rs
```diff
@@ -507,55 +507,71 @@ async fn push_metrics(
...
+
+    /// These are the shards that are just moved out, and therefore should be locked. They
+    /// should not be removed yet given that they need to be served as the source of shard
+    /// sync for the new shard owner.
+    pub fn shards_to_lock(&self) -> Vec<ShardIndex> {
+        let prev: HashSet<_> = self.shards_assigned_prev_epoch.iter().copied().collect();
+        let current: HashSet<_> = self.shards_assigned_current_epoch.iter().copied().collect();
+        prev.difference(&current).copied().collect()
+    }
```
jpcsmith: Suggestion: Calculate these lists once and store them in the struct as they share they share similar requirements (such as prev and next) and are currently re-calculated for every call.

crates/walrus-service/src/common/utils.rs
```diff
@@ -507,55 +507,71 @@ async fn push_metrics(
...
+
+    /// These are the shards that are gained from the previous epoch, and can be synced using
+    /// shard sync.
+    pub fn gained_shards_from_prev_epoch(&self) -> Vec<ShardIndex> {
+        let current: HashSet<_> = self.shards_assigned_current_epoch.iter().copied().collect();
+        let prev: HashSet<_> = self.shards_assigned_prev_epoch.iter().copied().collect();
+        current.difference(&prev).copied().collect()
+    }
+
+    /// These are the shards that are no longer needed and can be removed.
```
jpcsmith: Suggestion: Expand this comment to make it clear that this set excludes those that we hold on to, to allow the receiving node to sync from us.

crates/walrus-service/src/common/utils.rs
halfprice: We currently don't store node-id in the node yet. We only store the capability object. The refactor may be a bit more. I added a TODO for it.

crates/walrus-service/src/node/storage.rs
halfprice: Done

crates/walrus-sui/src/types/move_structs.rs
halfprice: This one is for the StorateNode representation of the move object, which I don't think we have a test utility yet. I created new_move_storage_node_for_testing in the walrus-sui/src/test_utils.rs

crates/walrus-service/src/common/utils.rs
halfprice: Done

crates/walrus-service/src/common/utils.rs
halfprice: Done

crates/walrus-service/src/common/utils.rs
halfprice: Done.

crates/walrus-service/src/common/utils.rs
halfprice: Done.

PR #1664, opened 2025-03-03
fix: use event type package address instead of event address

karlwuest:
## Description

Fixes an issue that the event processor was filtering events based on the package that emitted the event, not based on the package that defines the event. This caused nodes to miss events emitted when a blob is extended using the subsidies contract (since the package of the subsidies contract is set as event package id).

The fix includes special casing for the public testnet to make sure that nodes start filtering the events at the same time (in sequential order of all events), s.t. future event blobs are written correctly.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/events/event_processor.rs
+++ b/crates/walrus-service/src/node/events/event_processor.rs
@@ -425,6 +425,8 @@ impl EventProcessor {
             == ObjectID::from_str(
                 "0x795ddbc26b8cfff2551f45e198b87fc19473f2df50f995376b924ac80e56f88b",
             )?;
+        let march_7_7_pm_utc =
+            chrono::DateTime::parse_from_rfc3339("2025-03-07T19:00:00+00:00")?.to_utc();
 
         while let Some(entry) = rx.recv().await {
             let Ok(checkpoint) = entry.result else {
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -460,14 +472,26 @@ impl EventProcessor {
                 self.verify_checkpoint(&checkpoint, prev_verified_checkpoint)?;
             let mut write_batch = self.stores.event_store.batch();
             let mut counter = 0;
+            // TODO(WAL-667): remove special casing
+            let march_7_7_pm_utc =
+                chrono::DateTime::parse_from_rfc3339("2025-03-07T19:00:00+00:00")?.to_utc();
```
mlegner: Nit: I guess this will be optimized by the compiler, but can we move it outside of the loop?

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -460,14 +472,26 @@ impl EventProcessor {
...
                     .map_err(|e| anyhow!("Failed to update walrus package store: {}", e))?;
                 let tx_events = tx.events.unwrap_or_default();
                 let original_package_ids: Vec<ObjectID> =
                     try_join_all(tx_events.data.iter().map(|event| {
-                        self.package_store
-                            .get_original_package_id(event.package_id.into())
+                        // TODO(WAL-667): remove special case
+                        if on_public_testnet && before_march_7_7pm_utc {
+                            self.package_store
+                                .get_original_package_id(event.package_id.into())
```
mlegner: Nit/question: Can we do something like `let pkg = if ...` and then `self.package_store.get_original_package_id(pkg)`?

PR #1671, opened 2025-03-04
chore: remove raptorq from contracts

karlwuest:
## Description

Removes RaptorQ from contracts, removes testnet specific special case for choosing the default encoding, and removes raptorq from e2e tests to maintain compatibility with contracts.

Closes WAL-647.

## Test plan

Existing tests

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/subsidies/tests/subsidies_tests.move
+++ b/contracts/subsidies/tests/subsidies_tests.move
@@ -9,17 +9,17 @@ use sui::{coin::Coin, test_scenario as test};
 use wal::wal::WAL;
 use walrus::{
     blob::{Self, Blob},
-    encoding,
     messages,
     storage_resource::Storage,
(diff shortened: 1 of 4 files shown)


contracts/walrus/sources/system/redstuff.move
mlegner, line 1: Remark: I would probably have kept the mechanisms to support multiple encoding types for future use. However, I also see benefits of removing it (cleaner and more efficient), so I'm fine with this version.

karlwuest: I thought about that, but opted for removing it to have cleaner code, since it's not clear if/how we will have multiple RedStuff versions, and we can add it again when needed.

PR #1673, opened 2025-03-04
feat: added estimated start time of the current epoch

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/responses.rs
+++ b/crates/walrus-service/src/client/responses.rs
@@ -345,9 +345,10 @@ impl InfoEpochOutput {
             | EpochState::NextParamsSelected(epoch_start) => {
                 EpochTimeOrMessage::DateTime(epoch_start)
             }
-            EpochState::EpochChangeSync(_) => {
-                EpochTimeOrMessage::Message("Epoch change is currently in progress...".to_string())
-            }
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client/responses.rs
```diff
@@ -333,9 +339,20 @@ impl InfoEpochOutput {
...
         let epoch_duration = fixed_params.epoch_duration;
         let max_epochs_ahead = fixed_params.max_epochs_ahead;
+        let epoch_state = sui_read_client.epoch_state().await?;
+        let start_of_current_epoch = match epoch_state {
+            EpochState::EpochChangeDone(epoch_start)
+            | EpochState::NextParamsSelected(epoch_start) => {
+                EpochTimeOrMessage::DateTime(epoch_start)
+            }
+            EpochState::EpochChangeSync(_) => {
+                EpochTimeOrMessage::Message("Epoch change is currently in progress...".to_string())
```
mlegner: Nit/suggestion: In this case we can still output an earliest end time of the current epoch.

liquid-helium: Change applied.

PR #1676, opened 2025-03-04
fix(deps): bump Sui to testnet-v1.44.3

mlegner:
## Description

Bumps all Sui dependencies to `testnet-v1.44.3` and modifies some code following changes to `typed-store`.

## Test plan

CI.

### DIFF
--- a/crates/walrus-service/src/node/events/event_processor.rs
+++ b/crates/walrus-service/src/node/events/event_processor.rs
@@ -15,6 +15,7 @@ use std::{
 
 use anyhow::{anyhow, bail, Context, Result};
 use async_trait::async_trait;
+use bincode::Options as _;
 use chrono::Utc;
 use futures_util::future::try_join_all;
 use move_core_types::{
@@ -252,9 +253,10 @@ impl EventProcessor {
 
     /// Polls the event store for new events starting from the given sequence number.
     pub fn poll(&self, from: u64) -> Result<Vec<PositionedStreamEvent>> {
-        let iter = self.stores.event_store.unbounded_iter();
-        Ok(iter
-            .skip_to(&from)?
+        Ok(self
+            .stores
+            .event_store
+            .iter_with_bounds(Some(from), None)
             .take(MAX_EVENTS_PER_POLL)
             .map(|(_, event)| event)
             .collect())
@@ -263,8 +265,7 @@ impl EventProcessor {
     /// Polls the event store for the next event starting from the given sequence number,
     /// and returns the event along with any InitState that exists at that index.
     pub fn poll_next(&self, from: u64) -> Result<Option<StreamEventWithInitState>> {
-        let mut iter = self.stores.event_store.unbounded_iter();
-        iter = iter.skip_to(&from)?;
+        let mut iter = self.stores.event_store.iter_with_bounds(Some(from), None);
         let Some((index, event)) = iter.next() else {
             return Ok(None);
         };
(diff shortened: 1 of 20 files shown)


crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -325,7 +325,11 @@ impl EventProcessor {
                     write_batch.write()?;
                     // This will prune the event store by deleting all the sst files relevant to the
                     // events before the commit index
-                    self.stores.event_store.delete_file_in_range(&0, &commit_index)?;
+                    self.stores.event_store.rocksdb.delete_file_in_range(
```
sadhansood, line 338: This is unfortunate that we have to invoke this method on the `rocksdb` instance vs the `DBMap` instance as it did not get restored after the removal. But, the key encoding in typed store is bincode with big endian. I think the right way would be to store it:
```
use bincode::Options;
let start = bincode::DefaultOptions::new()
                        .with_big_endian()
                        .with_fixint_encoding()
                        .serialize(&0)?;
                    let end = bincode::DefaultOptions::new()
                        .with_big_endian()
                        .with_fixint_encoding()
                        .serialize(&commit_index)?;
                    self.stores.event_store.rocksdb.delete_file_in_range(
                        &self.stores.event_store.cf(),
                        &start,
                        &end,
                    )?;
```

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -217,8 +217,12 @@ impl BlobInfoTable {
...
 
-    pub fn keys(&self) -> Result<Vec<BlobId>, TypedStoreError> {
-        self.aggregate_blob_info.keys().collect()
+    #[cfg(test)]
+    pub fn keys(&self) -> Vec<BlobId> {
+        self.aggregate_blob_info
+            .safe_iter()
+            .filter_map(|e| e.ok().map(|(k, _)| k))
+            .collect()
     }
```
sadhansood: I think this should return a Result with safe_iter:
```
pub fn keys(&self) -> Result<Vec<BlobId>, TypedStoreError> {
        self.aggregate_blob_info
            .safe_iter()
            .map(|result| result.map(|(k, _)| k))
            .collect()
    }
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -1132,8 +1132,8 @@ impl ShardStorage {
     #[cfg(test)]
     pub(crate) fn sliver_count(&self, sliver_type: SliverType) -> usize {
         match sliver_type {
-            SliverType::Primary => self.primary_slivers.keys().count(),
-            SliverType::Secondary => self.secondary_slivers.keys().count(),
+            SliverType::Primary => self.primary_slivers.safe_iter().count(),
+            SliverType::Secondary => self.secondary_slivers.safe_iter().count(),
         }
     }
```
sadhansood: I think this should return a Result too:
```
pub(crate) fn sliver_count(&self, sliver_type: SliverType) -> Result<usize, TypedStoreError> {
        match sliver_type {
            SliverType::Primary => self.primary_slivers.safe_iter().try_fold(0, |count, item| {
                item.map(|_| count + 1)
            }),
            SliverType::Secondary => self.secondary_slivers.safe_iter().try_fold(0, |count, item| {
                item.map(|_| count + 1)
            }),
        }
    }
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -1163,10 +1163,11 @@ impl ShardStorage {
...
-    pub(crate) fn all_pending_recover_slivers(
-        &self,
-    ) -> Result<Vec<(SliverType, BlobId)>, TypedStoreError> {
-        self.pending_recover_slivers.keys().collect()
+    pub(crate) fn all_pending_recover_slivers(&self) -> Vec<(SliverType, BlobId)> {
+        self.pending_recover_slivers
+            .safe_iter()
+            .filter_map(|e| e.ok().map(|(k, _)| k))
+            .collect()
     }
```
sadhansood: I think this should also return a Result still.

PR #1678, opened 2025-03-04
feat: add a generic type representing values and axes

jpcsmith:
## Description

Use a generic type to represent `Primary(_)` and `Secondary(_)` variants.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-core/src/by_axis.rs
```diff
@@ -0,0 +1,193 @@
...
+        $crate::by_axis::derive_from_trait!(ByAxis<$t, $u>, ());
+    };
+}
+
+/// Error returned when trying to extract the wrong variant (primary or secondary).
+#[derive(Debug, Error, PartialEq, Eq, Clone)]
+#[error("wrong axis variant")]
+pub struct WrongAxisError;
+
+macro_rules! derive_try_from_trait {
```
mlegner, line 162: Question: Do you expect that to be used independently from the `derive_from_trait`? So far I just see two otherwise identical invocations of the two.

jpcsmith: Maybe, just figured to keep it separate since they're separate traits.

PR #1681, opened 2025-03-05
chore: walrus-proxy rust build needs contracts folder now

pei-mysten:
## Description

walrus-proxy rust build needs contracts folder now, adding that to dockerfile

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/docker/walrus-proxy/Dockerfile
+++ b/docker/walrus-proxy/Dockerfile
@@ -8,6 +8,7 @@ RUN apt-get update && apt-get install -y cmake clang
 COPY .git/ .git/
 COPY Cargo.toml Cargo.lock ./
 COPY crates crates
+COPY contracts contracts
 
 RUN cargo build --profile ${PROFILE} --bin walrus-proxy
 


docker/walrus-proxy/Dockerfile
```diff
@@ -8,6 +8,7 @@ RUN apt-get update && apt-get install -y cmake clang
 COPY .git/ .git/
 COPY Cargo.toml Cargo.lock ./
 COPY crates crates
+COPY contracts contracts
```
mlegner, line 11: Question: We have a slightly more complicate setup in other Dockerfiles:
```
COPY contracts ./contracts
COPY contracts /contracts
```

(Why) don't we need it here, but only in other places?

pei-mysten: `./contracts` will be used for build time
`/contracts` will be copied over to the `walrus-deploy` docker image only. `walrus-deploy` needs the copy the contracts to perform. We could either download from walrus-docs repo, or just copy over, and the latter one provide better consistency

mlegner: Ah, I see. Thanks for the explanation!

PR #1683, opened 2025-03-05
feat: improve performance of the the recoverySymbols HTTP endpoint

jpcsmith:
## Description

This adds several improvements to the performance of the storage node's recoverySymbol endpoint. These include making servicing recoverySymbol requests concurrent, avoiding buffer bloat with large amounts of recoverySymbol requests, and caching recoverySymbol proofs.

## Test plan

Added tests for the cache/thread-pool. Otherwise this is a refactor.

## Benchmark results
image

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/recovery_symbol_service.rs
+++ b/crates/walrus-service/src/node/recovery_symbol_service.rs
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use std::{
-    mem,
     sync::Arc,
     task::{self, Context, Poll},
 };
@@ -24,6 +23,7 @@ use walrus_core::{
 };
 
 use super::{thread_pool, BoundedRayonThreadPool};
+use crate::utils;
 
 /// The key into the cache.
 ///
@@ -163,8 +163,7 @@ impl Service<RecoverySymbolRequest> for RecoverySymbolService {
     }
 
     fn call(&mut self, req: RecoverySymbolRequest) -> Self::Future {
-        let unready_clone = self.clone();
-        let mut this = mem::replace(self, unready_clone);
+        let mut this = utils::clone_ready_service(self);
 
         async move { this.handle_request_and_cache(req).await }.boxed()
     }
--- a/crates/walrus-service/src/common/utils.rs
+++ b/crates/walrus-service/src/common/utils.rs
@@ -8,6 +8,7 @@ use std::{
     env,
     fmt::Debug,
     future::Future,
+    mem,
     net::{IpAddr, Ipv4Addr, SocketAddr},
     path::Path,
     pin::Pin,
@@ -841,6 +842,28 @@ pub async fn wait_until_terminated(mut exit_listener: oneshot::Receiver<()>) {
             Ok(_) => tracing::info!("exit notification received"),
         }
     }
+}
+
+/// Clone a service and return the original instance.
+///
+/// When cloning a service the clone may not yet be ready. For example, consider cloning a service
+/// that holds a permit for a semaphore. The clone is (typically) returned without the permit
+/// which is left on the original service. This means that the clone is likely not ready ready to
+/// be `call`-ed.
+///
+/// Therefore, the correct way to clone a ready service, where you would like to the clone to have
+/// the readiness, is to `clone()` then call [`std::mem::replace`] to swap the ready, original
+/// instance with the clone.
+///
+/// This function is a convenience function for that pattern.
+///
+/// For more information see the [`tower::Service`] documentation.
+pub(crate) fn clone_ready_service<T, R>(svc: &mut T) -> T
+where
+    T: tower::Service<R> + Clone,
+{
+    let unready_clone = svc.clone();
+    mem::replace(svc, unready_clone)
 }
 
 #[cfg(test)]


crates/walrus-core/src/merkle.rs
```diff
@@ -144,6 +144,14 @@ where
     }
 }
 
+impl PartialEq for MerkleProof {
+    fn eq(&self, other: &Self) -> bool {
+        self.path.eq(&other.path)
+    }
```
mlegner, line 150: Question: Is the difference to the previous derived version that the `_hash_type` is not compared?

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -787,7 +822,7 @@ mod tests {
             one_byte_symbol_raptorq: (EncodingType::RedStuffRaptorQ, &[1,2,3,4], 4, 1),
             two_byte_symbol_raptorq: (EncodingType::RedStuffRaptorQ, &[1,2,3,4], 2, 2),
             two_byte_symbol_reed_solomon: (EncodingType::RS2, &[1,2,3,4], 2, 2),
-            four_byte_symbol_reed_solomon: (EncodingType::RS2, &[1,2,3,4,5,6,7,8], 2, 2),
+            four_byte_symbol_reed_solomon: (EncodingType::RS2, &[1,2,3,4,5,6,7,8], 2, 4),
```
mlegner, line 825: Thank you for the fix! 🙏

crates/walrus-core/benches/basic_encoding.rs
mlegner, line 1: Question: Could you share a high-level summary of the results of these benchmarks?

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -157,14 +158,20 @@ impl<T: EncodingAxis> SliverData<T> {
...
-            config
-                .encode_all_symbols::<T::OrthogonalAxis>(self.symbols.data())?
-                .into_iter()
-                .flatten()
-                .collect(),
-            self.symbols.symbol_size(), // The symbol size remains unvaried when re-encoding.
-        ))
+        let symbol_list = config.encode_all_symbols::<T::OrthogonalAxis>(self.symbols.data())?;
+        assert!(!symbol_list.is_empty(), "must be at least 1 symbol");
+        assert!(!symbol_list[0].is_empty(), "symbols must have data");
```
mlegner, line 163: Question: Do we need these asserts in release builds?

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -198,6 +207,32 @@ impl<T: EncodingAxis> SliverData<T> {
...
+    /// Returns a [`RecoverySymbolError::EncodeError`] if the sliver cannot be encoded. Returns a
+    /// [`RecoverySymbolError::IndexTooLarge`] error if `target_pair_index >= n_shards`.
+    pub fn decoding_symbol_for_sliver(
+        &self,
+        target_pair_index: SliverPairIndex,
+        config: &EncodingConfigEnum,
+    ) -> Result<DecodingSymbol<T::OrthogonalAxis>, RecoverySymbolError> {
+        Self::check_index(target_pair_index.into(), config.n_shards())?;
+
+        // TODO(jsmith): Avoid expanding all the symbols to get a single symbol (WAL-611).
```
mlegner, line 226: Remark: This is probably an important piece to improve the performance. Are you going to work on this or should I?

crates/walrus-core/src/merkle.rs
```diff
@@ -144,6 +144,14 @@ where
     }
 }
 
+impl PartialEq for MerkleProof {
```
jpcsmith:
```suggestion
impl<T> PartialEq for MerkleProof<T> {
```

crates/walrus-core/src/merkle.rs
jpcsmith: The hash-type is compared, since the function is only defined if `self` and `other` are of the same type have the same `T`.

The difference is that the derive only will only implement `PartialEq` if `T: PartialEq`, but in our case, `T` is the algorithm and the algorithm being comparable is irrelevant for the proof being comparable.

crates/walrus-core/benches/basic_encoding.rs
jpcsmith: Something formal or simple? The throughput graph for the two methods is shown below

image

crates/walrus-core/src/encoding/slivers.rs
jpcsmith: It's not on my current stack, so feel free to tackle it if you find the cycles.

crates/walrus-core/src/encoding/slivers.rs
jpcsmith: Asserts in release builds are not bad, why would we want to remove them?

crates/walrus-core/src/merkle.rs
mlegner: Ah, right. 👍

crates/walrus-core/benches/basic_encoding.rs
mlegner: Simple is good. 😄

crates/walrus-core/src/encoding/slivers.rs
mlegner: They just cause additional code, right?
I guess it doesn't matter much whether or not we have two additional comparisons, so fine to leave them in. 🙂

crates/walrus-service/src/node.rs
```diff
@@ -1764,16 +1780,21 @@ impl StorageNodeInner {
...
         &self,
         blob_id: &BlobId,
         sliver_pair_index: SliverPairIndex,
         target_sliver_type: SliverType,
         target_pair_index: SliverPairIndex,
     ) -> Result<GeneralRecoverySymbol, RetrieveSymbolError> {
+        // Claim a worker for performing the expansion necessary to get the symbol.
+        let mut worker = match self.symbol_service.clone().ready_oneshot().now_or_never() {
+            Some(result) => result.expect("polling the symbol service is infallible"),
+            None => return Err(Unavailable.into()),
```
halfprice, line 1800: Question: does client perform retry upon this error explicitly? Or it relies on the retry of failing the entire `CollectRecoverySymbols` (https://github.com/MystenLabs/walrus/blob/8085ba8e26463dc7f36e51c570da30063d8190c8/crates/walrus-service/src/node/committee/request_futures.rs#L577)

crates/walrus-service/src/node.rs
```diff
@@ -1789,34 +1810,15 @@ impl StorageNodeInner {
...
-        }
+
+        let request = RecoverySymbolRequest {
+            blob_id: *metadata.blob_id(),
+            source_sliver: sliver,
+            target_pair_index,
+            encoding_type: metadata.metadata().encoding_type(),
+        };
+
+        worker.call(request).map_err(convert_error).await
```
halfprice, line 1829: Question: I assume the work will get released if the function returned before the `call()`?

crates/walrus-service/src/node.rs
```diff
@@ -2257,8 +2264,16 @@ impl ServiceState for StorageNodeInner {
...
+        // We use FuturesOrdered to keep the results in the same order as the requests.
+        let mut symbols: FuturesOrdered<_> = symbol_id_iter
+            .map(|symbol_id| {
+                self.retrieve_recovery_symbol(blob_id, symbol_id, target_type_from_proof)
+                    .map(move |result| (symbol_id, result))
+            })
+            .collect();
+
+        while let Some((symbol_id, result)) = symbols.next().await {
+            match result {
```
halfprice, line 2286: Remark: I think now there is higher chance that not all symbols are returned due to service unavailability. IIUC, symbols are still returned but with less symbols that are assigned to this node.

Does the client need to check if they have get all the symbols they think this node should return? And if not, retry fetching the missing one if necessary?

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -0,0 +1,388 @@
...
+        Ok(recovery_symbol)
+    }
+}
+
+impl Service<RecoverySymbolRequest> for RecoverySymbolService {
+    type Response = GeneralRecoverySymbol;
+    type Error = RecoverySymbolError;
+    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Self::Error>> {
```
halfprice, line 154: question: there seems to be a race between multiple poll_ready with single available slot (single thread available, multiple concurrent poll ready all return Ready). Is that possible? Does this matter or it's ok to temporarily congest the service with more requests than threads. Caching may help but it's a bit hard to reason about what's the actual effects.

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -0,0 +1,388 @@
...
+        ));
+
+        thread_pool::unwrap_or_resume_panic(result);
+
+        Poll::Ready(Ok(()))
+    }
+
+    fn call(&mut self, req: RecoverySymbolRequest) -> Self::Future {
+        let unready_clone = self.clone();
+        let mut this = mem::replace(self, unready_clone);
```
halfprice: Nit: can you document why we need to replace self? IIUC, this is due to how the BoundedRayonThreadPool is cloned?

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -0,0 +1,388 @@
...
+
+        async move { this.handle_request_and_cache(req).await }.boxed()
```
halfprice, line 168: Question: do you think we should run all of these in the thread pool?

crates/walrus-service/src/node.rs
halfprice: I added this comment before I looked at how the client performs retries. If this also relies on the retry of entire CollectRecoverySymbols, maybe it's fine for now and we can revisit of we want a more efficient retry mechanism.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -0,0 +1,124 @@
...
+use tower::{limit::ConcurrencyLimit, Service};
+use tracing::span::Span;
+
+pub(crate) type ThreadPanicError = Box<dyn Any + Send + 'static>;
+
+/// Limit of the number of inflight requests on a [`BoundedRayonThreadPool`] to
+/// `DEFAULT_IN_FLIGHT_RATIO * pool.current_num_threads`.
+pub(crate) const DEFAULT_IN_FLIGHT_RATIO: f64 = 1.5;
+/// Minimum value for the number of inflight requests on a [`BoundedRayonThreadPool`].
+pub(crate) const MIN_IN_FLIGHT_REQUESTS: usize = 3;
```
halfprice, line 25: Suggestion: IIUC, the thread pool is 1 per node handling all the recovery symbol requests. Is 3 a bit low given that typical nodes have 10+ cores?

crates/walrus-service/src/node.rs
jpcsmith: The recovery future in the `CommitteeService` which uses this endpoint in the other storage nodes retires regardless of the cause of the error, so requests to this node will be retried after a backoff.

crates/walrus-service/src/node.rs
jpcsmith: Do you mean the `worker` will get released? Yes, it's a clone so when it's dropped any resources that are held will be freed. In this case, any acquired semaphore.

crates/walrus-service/src/node.rs
jpcsmith: In the event that a storage node is overloaded, then yes. It will not even try to return all the symbols that it has since it's already under load.

If the requesting storage node is unable to get sufficient symbols from the other storage nodes in the committee, then it will retry this node after a (exponential) backoff, which is the desired client behaviour for a server under load.

crates/walrus-service/src/node/recovery_symbol_service.rs
jpcsmith: There shouldn't be. Since `poll_ready` requires an exclusive reference (`&mut self`) to the instance, the exclusivity of the object is guaranteed by rust's borrow checker. For a single instance of the object, only 1 caller can be calling `poll_ready` at a time.

Given the above, sharing the service is then done by cloning it, multiple callers can each be calling `poll_ready` on their cloned instance. If there's only 1 underlying resource available (e.g., a semaphore permit as is used around the rayon thread pool) then only one of those methods will be able to acquire the permit (guaranteed by the semaphore) and return ready.

halfprice: Yeah, I think the clone + `poll_ready` is what I'm confused about. It seems that the [ready_oneshot()](https://github.com/MystenLabs/walrus/blob/485d5ffaee29acb58791f033e2f6e6b692baa971/crates/walrus-service/src/node.rs#L1791) always works on a cloned copy so the `&mut self` doesn't apply.

> then only one of those methods will be able to acquire the permit (guaranteed by the semaphore) and return ready.

This I understand, and just want to make sure that my understanding is correct that even one resource available, multiple workers may still try to run the request instead of [returning `Unavailable`](https://github.com/MystenLabs/walrus/blob/485d5ffaee29acb58791f033e2f6e6b692baa971/crates/walrus-service/src/node.rs#L1793).

crates/walrus-service/src/node/recovery_symbol_service.rs
jpcsmith: The rational is here: `https://docs.rs/tower/latest/tower/trait.Service.html#be-careful-when-cloning-inner-services`.

I've moved the code to a function under `crate::utils` and added a docstring there to describe the reason and link to that article.

crates/walrus-service/src/node/thread_pool.rs
jpcsmith: The default is `max(MIN_IN_FLIGHT_REQUESTS, DEFAULT_IN_FLIGHT_RATIO * num_cores)`. So for a 10 core machine it'll be 15. The minimum was added because when we're running in tests in single-thread mode, `1.5 * 1 =~ 1` which was too little.

crates/walrus-service/src/node/recovery_symbol_service.rs
jpcsmith: Maybe. Beside the merkle tree construction, the creation of the decoding symbol may benefit from being moved into the thread-pool if it remains CPU intensive after the change to compute only the specific symbol (instead of all of them) lands. At that point we can re-evaluate.

PR #1684, opened 2025-03-06
fix: address the race condition in creating shards and shard access in the new epoch

halfprice:
## Description

Exposed in simtest that during epoch change, after new committee is refreshed, and until new shards are created if
there is any new shard, there is a window that the storage node does not have all the shards that assigned to it,
and shard access to go through all the owned shards, such as blob sync, fails. This PR adds a critical section between
the new committee created and when the new shards are created.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -27,7 +27,7 @@ use prometheus::Registry;
 use rand::{rngs::StdRng, thread_rng, Rng, SeedableRng};
 use serde::Serialize;
 use start_epoch_change_finisher::StartEpochChangeFinisher;
-use storage::blob_info::PerObjectBlobInfoApi;
+use storage::{blob_info::PerObjectBlobInfoApi, StorageShardLock};
 pub use storage::{DatabaseConfig, NodeStatus, Storage};
 #[cfg(msim)]
 use sui_macros::fail_point_if;
@@ -139,7 +139,11 @@ use self::{
 };
 use crate::{
     client::Blocklist,
-    common::{config::SuiConfig, utils::should_reposition_cursor},
+    common::{
+        active_committees::ActiveCommittees,
+        config::SuiConfig,
+        utils::should_reposition_cursor,
+    },
     utils::ShardDiffCalculator,
 };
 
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1284,14 +1319,17 @@ impl StorageNode {
...
         let committees = self.inner.committee_service.active_committees();
         let shard_diff_calculator =
-            ShardDiffCalculator::new(&committees, public_key, &storage.existing_shards().await);
+            ShardDiffCalculator::new(&committees, public_key, shard_map_lock.existing_shards());
 
         // Since the node is doing a full recovery, its local shards may be out of sync with the
         // contract for multiple epochs. Here we need to make sure that all the shards that is
         // assigned to the node in the latest epoch are created.
+        //
+        // Note that the shard_map_lock will be unlocked after this function returns.
```
jpcsmith, line 1328: Question: Is there a reason not to hold the lock until the function completes? This would ensure that all changes are made before reads/writes can resume?

crates/walrus-service/src/node.rs
```diff
@@ -1396,14 +1434,24 @@ impl StorageNode {
...
         let committees = self.inner.committee_service.active_committees();
         assert!(event.epoch <= committees.epoch());
 
         let shard_diff_calculator =
-            ShardDiffCalculator::new(&committees, public_key, &storage.existing_shards().await);
+            ShardDiffCalculator::new(&committees, public_key, shard_map_lock.existing_shards());
+
+        let shards_gained = shard_diff_calculator.gained_shards_from_prev_epoch();
+        self.create_new_shards_and_start_sync(
+            shard_map_lock,
```
jpcsmith, line 1449: Question: Same as above, is there a reason not to hold the lock until the function completes? This would ensure that all changes are made before reads/writes can resume?

crates/walrus-service/src/node.rs
halfprice: Thanks for the review @jpcsmith , and sorry for the late reply... So the main reason for not holding it until the end of epoch change processing is due to other shard operation such as locking the shard and removing the shard.

Given that `StorageShardLock` does not expose the underlying ShardStorage, which means that all the shards are not mutable until the lock is released.

I think we can create a shard locking in `Storage` similar to shard creation that passing in a locked `StorageShardLock`, which just requires some code refactoring.

For shard removal, since we want to run it in the background and not blocking event processing, it is not too relevant.

Let me know what do you think. I'm going to merge this for now since another fix is depending on this fix.

crates/walrus-service/src/node.rs
halfprice: Please see above.

PR #1689, opened 2025-03-09
feat: Update commission rate

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -329,6 +329,10 @@ impl StorageNodeConfig {
         }
     }
 
+    /// Calculates the next commission rate for the storage node.
+    ///
+    /// This function compares the local commission rate with the on-chain projected commission
+    /// rate one epoch in the future and returns the local commission rate if it is different.
     fn calculate_next_commission_rate(
         &self,
         commission_rate_data: &CommissionRateData,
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/config.rs
```diff
@@ -328,6 +329,25 @@ impl StorageNodeConfig {
         }
     }
 
+    /// Calculates the next commission rate for the storage node.
+    ///
+    /// This function compares the local commission rate with the on-chain projected commission
+    /// rate one epoch in the future and returns the local commission rate if it is different.
```
halfprice, line 335: I think this comment is a bit inaccurate. IIUC, here the node checks if the latest commission rate is set to the local value, which is not necessarily one epoch in the future, but could be 2 (commission rate is always [set 2 epoch ahead](https://github.com/MystenLabs/walrus/blob/83c8e65738bc53786daf2ffcf5f0f3cbc67dcb9f/contracts/walrus/sources/staking/staking_pool.move#L507)).

I was a little worried that if we only look at one epoch in the future, we may repeatedly set the commission rate even though it is already updated, but the implementation seems correct.

liquid-helium: will update the doc in a following pr, it was meant to be two epcochs :)

PR #1690, opened 2025-03-09
fix: only update metadata is stored for event blob

halfprice:
## Description

It was updating for all blobs. We only need to do this for the event blob.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/events/event_blob_writer.rs
+++ b/crates/walrus-service/src/node/events/event_blob_writer.rs
@@ -1170,11 +1170,6 @@ impl EventBlobWriter {
             return Ok(());
         };
 
-        self.node
-            .storage()
-            .update_blob_info_with_metadata(&blob_id)
-            .context("unable to update metadata")?;
-
         let attested = self.attested.clone();
         let failed_to_attest = self.failed_to_attest.clone();
         let metadata = self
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -1189,6 +1184,11 @@ impl EventBlobWriter {
             return Ok(());
         };
 
+        self.node
+            .storage()
+            .update_blob_info_with_metadata(&blob_id)
+            .context("unable to update metadata")?;
```
mlegner, line 1190: Question @halfprice @sadhansood: Are we sure the blob is already tracked at this point?

sadhansood: I think so @mlegner. Before we get blob certify event for an event blob, we also get the blob registered event for it which ensures we start tracking the blob in blob info table before we invoke this function `update_blob_info_with_metadata`.

mlegner: OK, makes sense: In the [`process_events`](https://github.com/MystenLabs/walrus/blob/e6f3bc2aeea69991457e60ce93cb9419664e2e09/crates/walrus-service/src/node.rs#L799) function, we first call `process_event`, which updates the blob info, and only afterwards call `writer.write`.

PR #1693, opened 2025-03-10
feat: Change event blob size

sadhansood:
## Description

Add a check for public or private testnet, but otherwise increase the size of blobs to have 216_000 checkpoints per blob.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -7,6 +7,7 @@ use std::{
     future::Future,
     num::{NonZero, NonZeroU16},
     pin::Pin,
+    str::FromStr,
     sync::{
         atomic::{AtomicBool, Ordering},
         Arc,
@@ -18,7 +19,7 @@ use blob_retirement_notifier::BlobRetirementNotifier;
 use committee::{BeginCommitteeChangeError, EndCommitteeChangeError};
 use epoch_change_driver::EpochChangeDriver;
 use errors::{ListSymbolsError, Unavailable};
-use events::event_blob_writer::EventBlobWriter;
+use events::event_blob_writer::{EventBlobWriter, NUM_CHECKPOINTS_PER_BLOB};
 use fastcrypto::traits::KeyPair;
 use futures::{
     stream::{self, FuturesOrdered},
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -606,12 +609,37 @@ impl StorageNode {
...
+
+        // TODO(WAL-667): remove special case
+        let system_package_id = read_client.get_system_package_id();
+        let on_public_testnet = system_package_id
+            == ObjectID::from_str(
+                "0x795ddbc26b8cfff2551f45e198b87fc19473f2df50f995376b924ac80e56f88b",
+            )?;
+        let on_private_testnet = system_package_id
+            == ObjectID::from_str(
+                "0x11f5d87dab9494ce459299c7874e959ff121649fd2d4529965f6dea85c153d2d",
```
halfprice: Question: IIUC, redeploying PTN always requires updating this id?

[not in this PR] I wonder if we should introduce a network identifier (probably implemented in the contract, set it once at the beginning of contract publishing) such as `mainnet`, `testnet`, `private-testnet` etc. so that the node can automatically figuring out which network it is on.

crates/walrus-service/src/node.rs
```diff
@@ -606,12 +609,37 @@ impl StorageNode {
...
 
+        let read_client = config
+            .sui
+            .as_ref()
+            .ok_or_else(|| anyhow!("sui config is required for read client"))?
+            .new_read_client()
+            .await?;
+
+        // TODO(WAL-667): remove special case
+        let system_package_id = read_client.get_system_package_id();
```
halfprice: Nit: I think if you move this block to StorageNodeBuilder::build, there the system package is already resolved, and you don't need to create a new read client and fetch the system package again.

It depends on when this code block will be removed. If very soon, no big deal then.

crates/walrus-service/src/node.rs
sadhansood: No, if we redeploy PTN we will fallback to default path which is `NUM_CHECKPOINTS_PER_BLOB` and we don't need to worry about updating this at all. This is only temporary to support the current running deployments.
I agree, a network identifier is a good idea. Arguably, the system package id is sort of the on-chain network identifier which we are using here to identify the correct node config.

crates/walrus-service/src/node.rs
sadhansood: ah ok, yeah this could be removed once we redeploy public testnet, so hopefully soon.

crates/walrus-service/src/node.rs
halfprice: Ah, I see. This is to not use the new value in the existing network. This makes sense.

PR #1700, opened 2025-03-11
feat: Add workflow to attach binaries to the releases

ebmifa:
## Description
Add workflow to attach binaries to the releases

## Test plan
I will be testing this once it lands

### DIFF
--- a/binary-build-list.json
+++ b/binary-build-list.json
@@ -1,22 +1,3 @@
 {
-    "release_binaries": [
-        "sui",
-        "sui-node",
-        "sui-tool",
-        "sui-faucet",
-        "sui-data-ingestion",
-        "sui-bridge",
-        "sui-bridge-cli",
-        "sui-graphql-rpc",
-        "sui-test-validator",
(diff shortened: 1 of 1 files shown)


.github/workflows/attach-binaries-to-release.yml
```diff
@@ -0,0 +1,193 @@
...
+        shell: bash
+        run: |
+          sudo apt update && sudo apt install libpq-dev
+
+      - name: Remove unused apps (MacOS arm64)
+        if: ${{ matrix.os == 'macos-latest-xlarge' && env.walrus_tarball_exist == '' }}
+        continue-on-error: true
+        shell: bash
+        run: |
+          # MacOS arm64 runner only has 14GB avaialble, which is too small for our builds, so removing unused softwared.
```
pei-mysten, line 155: our own mac studio 🙈

ebmifa: No, this is a GH runner. Not ours

PR #1703, opened 2025-03-12
feat: monitor TLS certificate file

liquid-helium:
## Description

Resolve WAL-245
Close WAL-197

## Test plan

---

## Release notes

- [x] Storage node:
Storage node monitors TLS certificate, and reboots storage node upon changes.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/config_synchronizer.rs
+++ b/crates/walrus-service/src/node/config_synchronizer.rs
@@ -139,15 +139,12 @@ impl ConfigSynchronizer {
         }
     }
 
-    /// Internal implementation to load and hash the TLS certificate.
+    /// Loads and hashes the TLS certificate.
     ///
     /// Returns Ok(None) if TLS is disabled or certificate path is not configured.
     /// Returns Ok(Some(hash)) if certificate is successfully loaded and hashed.
     /// Returns Err for actual errors like file not found or relative paths.
-    async fn load_tls_cert_hash_internal(
-        &self,
-        tls_config: &TlsConfig,
-    ) -> anyhow::Result<Option<Vec<u8>>> {
+    async fn load_tls_cert_hash(&self, tls_config: &TlsConfig) -> anyhow::Result<Option<Vec<u8>>> {
         // Skip if TLS is disabled or not configured.
         if tls_config.disable_tls {
             return Ok(None);
@@ -178,21 +175,6 @@ impl ConfigSynchronizer {
         let current_hash = hasher.finalize().to_vec();
 
         Ok(Some(current_hash))
-    }
-
-    #[cfg(test)]
-    /// Loads the TLS certificate hash in tests.
-    pub async fn load_tls_cert_hash(
-        &self,
-        tls_config: &TlsConfig,
-    ) -> anyhow::Result<Option<Vec<u8>>> {
-        self.load_tls_cert_hash_internal(tls_config).await
-    }
-
-    #[cfg(not(test))]
-    /// Loads the TLS certificate hash.
-    async fn load_tls_cert_hash(&self, tls_config: &TlsConfig) -> anyhow::Result<Option<Vec<u8>>> {
-        self.load_tls_cert_hash_internal(tls_config).await
     }
 }
 


crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -108,6 +142,62 @@ impl ConfigSynchronizer {
...
+        let Some(cert_path) = &tls_config.certificate_path else {
+            return Ok(None);
+        };
+
+        // Return error if the path is relative.
+        let cert_path = PathBuf::from(cert_path);
+        if cert_path.is_relative() {
+            return Err(anyhow!(
+                "TLS certificate path must be absolute, got relative path: {:?}",
+                cert_path
```
mlegner: Nit:
```suggestion
                "TLS certificate path must be absolute, got relative path: {}",
                cert_path.display()
```

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -108,6 +142,62 @@ impl ConfigSynchronizer {
...
+            ));
+        }
+
+        // Read the certificate file.
+        let cert_data = fs::read(&cert_path).await.context(format!(
+            "Failed to read certificate file at {:?}",
+            cert_path
```
mlegner: Nit:
```suggestion
            "failed to read certificate file at {}",
            cert_path.display()
```

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -108,6 +142,62 @@ impl ConfigSynchronizer {
...
+        tls_config: &TlsConfig,
+    ) -> anyhow::Result<Option<Vec<u8>>> {
+        self.load_tls_cert_hash_internal(tls_config).await
+    }
+
+    #[cfg(not(test))]
+    /// Loads the TLS certificate hash.
+    async fn load_tls_cert_hash(&self, tls_config: &TlsConfig) -> anyhow::Result<Option<Vec<u8>>> {
+        self.load_tls_cert_hash_internal(tls_config).await
+    }
```
mlegner: Question: Why do we have two separate functions here? Is it just for the `pub` visibility?

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -75,11 +78,45 @@ impl ConfigSynchronizer {
...
+                tracing::warn!("config loader is not set");
+                continue;
+            };
+
+            let config = config_loader.load_storage_node_config().await?;
+
+            let new_cert_hash = self.load_tls_cert_hash(&config.tls).await?;
+            if let (Some(old_hash), Some(new_hash)) = (&cert_hash, &new_cert_hash) {
+                if old_hash != new_hash {
+                    tracing::error!("TLS certificate has changed");
```
mlegner: Important: I think this should not be an error as this is a normal behavior; `info` might be more appropriate.

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -75,11 +78,45 @@ impl ConfigSynchronizer {
...
+                    return Err(SyncNodeConfigError::NodeNeedsReboot);
+                }
+            } else {
+                cert_hash = new_cert_hash;
+            }
```
mlegner: Important: I think we should also reload the config if we previously had `None` and now `Some(_)`.

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -75,11 +78,45 @@ impl ConfigSynchronizer {
...
+        } else {
+            tracing::warn!("config loader is not set");
+        }
+
         loop {
             tokio::time::sleep(self.check_interval).await;
 
+            if let Err(e) = self.committee_service.sync_committee_members().await {
+                tracing::error!("failed to sync committee: {}", e);
+            }
```
jpcsmith:
```suggestion
            if let Err(error) = self.committee_service.sync_committee_members().await {
                tracing::error!(%error, "failed to sync committee");
            }
```

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -75,11 +78,45 @@ impl ConfigSynchronizer {
     /// Runs the config synchronization loop
     /// Errors are ignored except for NodeNeedsReboot and RotationRequired
     pub async fn run(&self) -> Result<(), SyncNodeConfigError> {
+        // Initialize TLS certificate hash.
+        let mut cert_hash = None;
+        if let Some(config_loader) = &self.config_loader {
+            let config = config_loader.load_storage_node_config().await?;
+            cert_hash = self.load_tls_cert_hash(&config.tls).await?;
+        } else {
+            tracing::warn!("config loader is not set");
```
jpcsmith: Question: What does this mean if I see it in my logs? When would this not be set? How should I resolve it?

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -75,11 +78,45 @@ impl ConfigSynchronizer {
...
+
+            let Some(config_loader) = &self.config_loader else {
+                tracing::warn!("config loader is not set");
+                continue;
+            };
+
+            let config = config_loader.load_storage_node_config().await?;
+
+            let new_cert_hash = self.load_tls_cert_hash(&config.tls).await?;
+            if let (Some(old_hash), Some(new_hash)) = (&cert_hash, &new_cert_hash) {
```
jpcsmith: Question: Doesn't the node also need a reboot if `new_cert_hash` is None?

If the certificate path has become unset, they would now be using self-signed certificates -> reboot to enable self-signed certificates.

If `disable_tls` is true resulting in new_cert_hash now being None, then a middlebox is terminating the TLS connection. In this case, they need to reboot to turn off TLS on their node.

crates/walrus-service/src/node/config_synchronizer.rs
jpcsmith: Question: why  does this need pub in tests?

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: That's a good point! Thanks for the catch.
I will reboot the node as long as the new_cert is different from the previous one, no matter if is is None vs Some or hash diff.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Right, there is no AI, it is supposed to be there in prod, and could be absent in tests, in any case, no action is required. Removed the warning log.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Thanks for the tip, change applied.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Absolutely, good observation! I adjusted it to reload when they differ, which certainly highlighted my lack of practical knowledge about TLS.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Change applied.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: The function is called only within ConfigSynchronizer, so it makes sense to keep it private, but in that case I won't be able to test it in the test session below. Open to other suggestions.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Change applied.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Change applied.

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -75,23 +78,51 @@ impl ConfigSynchronizer {
...
                     SyncNodeConfigError::NodeNeedsReboot
                         | SyncNodeConfigError::ProtocolKeyPairRotationRequired
                 ) {
-                    tracing::warn!("going to reboot node due to {}", e);
-                    return Err(e);
+                    tracing::warn!(%error, "going to reboot node");
+                    return Err(error);
                 }
-                tracing::error!("failed to sync node params: {}", e);
+                tracing::error!(%error, "failed to sync node params");
```
jpcsmith, line 115: Important: this should be in else otherwise it prints for reboots

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -75,23 +78,51 @@ impl ConfigSynchronizer {
...
-            if let Err(e) = self.sync_node_params().await {
+            if let Err(error) = self
+                .contract_service
+                .sync_node_params(&config, self.node_capability_object_id)
+                .await
+            {
                 if matches!(
-                    e,
+                    error,
                     SyncNodeConfigError::NodeNeedsReboot
```
jpcsmith, line 109: Question: Doesn't the node also need a reboot if the **network** key has changed? Or is that already captured here?

crates/walrus-service/src/node/config_synchronizer.rs
```diff
@@ -104,10 +135,65 @@ impl ConfigSynchronizer {
...
+        tls_config: &TlsConfig,
+    ) -> anyhow::Result<Option<Vec<u8>>> {
+        self.load_tls_cert_hash_internal(tls_config).await
+    }
+
+    #[cfg(not(test))]
+    /// Loads the TLS certificate hash.
+    async fn load_tls_cert_hash(&self, tls_config: &TlsConfig) -> anyhow::Result<Option<Vec<u8>>> {
+        self.load_tls_cert_hash_internal(tls_config).await
+    }
```
jpcsmith: Suggestion: remove the pub and the duplicate method and merge with the _internal method.

Tests can access private methods of structs in their parent modules, so this is all unnecessary.

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: I didn't get it, if it is an error for reboot, the function should have returned from the above if matches!?

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: Yes, it is already captured in sync_node_params().

crates/walrus-service/src/node/config_synchronizer.rs
liquid-helium: change applied. Thanks for the tip.

crates/walrus-service/src/node/config_synchronizer.rs
jpcsmith: Yes, sorry i missed the return

PR #1704, opened 2025-03-12
fix(docker): clean up Dockerfiles and bump Rust version

mlegner:
## Description

- Bump Rust version in Dockerfiles to 1.85 (aligned with the toolchain we use everywhere else).
- Don't create a separate `base` image if it is not actually reused.
- Use name `builder` consistently.

## Test plan

Would be good to have some basic lints/tests for our Docker setup... 🤔

### DIFF
--- a/docker/walrus-service/Dockerfile
+++ b/docker/walrus-service/Dockerfile
@@ -2,7 +2,7 @@
 #
 # Copy in all crates, Cargo.toml and Cargo.lock unmodified,
 # and build the application.
-FROM rust:1.84-bookworm AS base
+FROM rust:1.85-bookworm AS builder
 ARG PROFILE=release
 ARG GIT_REVISION
 ENV GIT_REVISION=$GIT_REVISION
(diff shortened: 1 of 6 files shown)


docker/walrus-service/Dockerfile
```diff
@@ -2,7 +2,7 @@
 #
 # Copy in all crates, Cargo.toml and Cargo.lock unmodified,
 # and build the application.
-FROM rust:1.84-bookworm AS base
+FROM rust:1.85-bookworm AS builder
```
mlegner, line 5: Question @pei-mysten: Is there any benefit of having a separate `base` image if it is only used a single time to create the `builder`?

pei-mysten: in general yes, but in our case here, no.
Sometimes the prod image is based on base, and without separating base & builder, you might bring in some unnecessary deps into the prod image. But in our case here, the prod image is based on debian-slim, so it really doesn't matter!

mlegner: OK, that was also my understanding. Thanks for confirming! 🙏

PR #1711, opened 2025-03-13
fix(client): improve native TLS certificate handling

mlegner:
## Description

* load native TLS certificates from the system trust store by default
* only load native certificates once and keep them in the `NodeCommunicationFactory`
* respect `disable_native_certs` configuration parameter for `walrus health` command

On MacOS, running against the ~100 node Testnet, caching the native certificates substantially reduces the time for various operations:

* `walrus read <blob-id>`: before ~16s, after ~2.5s
* `walrus blob-status --blob-id <blob-id>`: before ~15s, after ~2s
* `walrus health --committee`: before ~19s, after ~11s

Closes WAL-275

## Test plan

Existing tests + manual testing.

### DIFF
--- a/crates/walrus-service/src/client/communication/factory.rs
+++ b/crates/walrus-service/src/client/communication/factory.rs
@@ -11,6 +11,8 @@ use std::{
 use anyhow::anyhow;
 use rand::{seq::SliceRandom, thread_rng};
 use reqwest::Client as ReqwestClient;
+use rustls::pki_types::CertificateDer;
+use rustls_native_certs::CertificateResult;
 use tokio::sync::Semaphore;
 use walrus_core::{encoding::EncodingConfig, Epoch, NetworkPublicKey};
 use walrus_sdk::{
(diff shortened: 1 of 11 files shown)


crates/walrus-service/src/client/communication/factory.rs
```diff
@@ -30,19 +32,39 @@ pub(crate) struct NodeCommunicationFactory {
...
-    ) -> Self {
-        Self {
+    ) -> ClientResult<Self> {
+        let native_certs = if !config.disable_native_certs {
+            let CertificateResult { certs, errors, .. } = rustls_native_certs::load_native_certs();
+            if certs.is_empty() {
+                return Err(ClientError::from(ClientErrorKind::FailedToLoadCerts(
+                    errors,
+                )));
+            };
```
liquid-helium, line 50: Is it possible to have a fall-back path, when native_cert errors out, we treat it as disabled?

mlegner: Hmm... Good question. I believe this would cause issues when storage nodes actually have proper TLS certificates, because the client would fail to connect. I think it's safer to error out in this case.

PR #1713, opened 2025-03-13
test: create a simtest to test event blob fork and recovery

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-simtest/tests/simtest.rs
+++ b/crates/walrus-simtest/tests/simtest.rs
@@ -27,6 +27,7 @@ mod tests {
     use walrus_core::{
         encoding::{Primary, Secondary},
         keys::{NetworkKeyPair, ProtocolKeyPair},
+        test_utils::random_blob_id,
         DEFAULT_ENCODING,
     };
     use walrus_proc_macros::walrus_simtest;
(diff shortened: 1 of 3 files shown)


crates/walrus-simtest/tests/simtest.rs
```diff
@@ -1866,4 +1871,152 @@ mod tests {
...
+                tracing::info!(
+                    "waiting for event blob fork, seeing a new event blob: {:?}",
+                    last_certified_event_blob_after_folk
+                );
+                last_certified_event_blob_after_folk = current_certified_event_blob;
+                last_new_certified_event_blob = Instant::now();
+            }
+
+            if last_new_certified_event_blob.elapsed() > Duration::from_secs(20) {
+                tracing::info!("event blob certification has been stuck for 20 seconds, breaking");
```
sadhansood, line 1976: Is it possible that we timeout not because the event blob writer is stuck but because there is delay in forming blobs?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -1866,4 +1871,152 @@ mod tests {
...
+                break;
```
sadhansood, line 1977: One thing we could possibly do here check and ensure the last attested blob across all nodes is actually different or the quorum is really not achieved.

PR #1714, opened 2025-03-14
fix: Reset local state of event blob writer

sadhansood:
## Description

During startup, reset all pending, attested and failed to attest blobs once we realize we have accumulated quite a few of them since the last event blob was certified. This lets storage nodes generate new blobs after a code change and re-attest again upon epoch change.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -632,6 +632,7 @@ impl StorageNode {
             num_checkpoints_per_blob
         );
 
+        let last_certified_event_blob = contract_service.last_certified_event_blob().await?;
         let event_blob_writer_factory = if !config.disable_event_blob_writer {
             Some(EventBlobWriterFactory::new(
                 &config.storage_path,
@@ -640,6 +641,8 @@ impl StorageNode {
                 node_params
                     .num_checkpoints_per_blob
                     .or(num_checkpoints_per_blob),
+                last_certified_event_blob,
+                config.num_uncertified_blob_threshold,
             )?)
         } else {
             None
@@ -811,7 +814,10 @@ impl StorageNode {
             return Ok(None);
         };
         let read_client = config.new_read_client().await?;
-        let system_package_id = read_client.get_system_package_id();
+        let retriable_client = read_client.sui_client();
+        let system_package_id = retriable_client
+            .get_package_id_from_object(read_client.get_system_object_id())
+            .await?;
         let on_public_testnet = system_package_id
             == ObjectID::from_str(
                 "0x795ddbc26b8cfff2551f45e198b87fc19473f2df50f995376b924ac80e56f88b",
@@ -5383,6 +5389,9 @@ mod tests {
         contract_service
             .expect_get_epoch_and_state()
             .returning(move || Ok((0, EpochState::EpochChangeDone(Utc::now()))));
+        contract_service
+            .expect_last_certified_event_blob()
+            .returning(|| Ok(None));
         let node = StorageNodeHandle::builder()
             .with_system_event_provider(vec![ContractEvent::EpochChangeEvent(
                 EpochChangeEvent::EpochChangeStart(EpochChangeStart {
--- a/crates/walrus-service/src/node/events/event_blob_writer.rs
+++ b/crates/walrus-service/src/node/events/event_blob_writer.rs
@@ -36,6 +36,7 @@ use walrus_sui::{
     client::SuiClientError,
     types::{
         move_errors::{MoveExecutionError, SystemStateInnerError},
+        move_structs::EventBlob as SuiEventBlob,
         BlobEvent,
         ContractEvent,
         EpochChangeEvent,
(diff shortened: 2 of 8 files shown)


crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -420,6 +432,68 @@ impl EventBlobWriterFactory {
...
+    /// have a checkpoint sequence number greater than the last certified blob.
+    fn reset_unattested_blobs(
+        pending: &DBMap<u64, PendingEventBlobMetadata>,
+        attested: &DBMap<(), AttestedEventBlobMetadata>,
+        failed_to_attest: &DBMap<(), FailedToAttestEventBlobMetadata>,
+        num_unattested_blob_threshold: u32,
+        last_certified_event_blob: Option<SuiEventBlob>,
+    ) -> Result<()> {
+        // Early return if no certified blob exists
+        let Some(last_certified_event_blob) = last_certified_event_blob else {
```
halfprice, line 466: question: what if the first event blob is corrupted?

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -420,6 +432,68 @@ impl EventBlobWriterFactory {
...
+            return Ok(());
+        };
+        let pending_blobs = pending
+            .unbounded_iter()
+            .filter(|(_, metadata)| {
+                metadata.end > last_certified_event_blob.ending_checkpoint_sequence_number
```
halfprice, line 474: question: is it possible that we have pending blob whose metadat.end <= last_certified_event_blob.ending_checkpoint_sequence_number?

crates/walrus-service/src/node.rs
```diff
@@ -632,6 +632,7 @@ impl StorageNode {
             num_checkpoints_per_blob
         );
 
+        let last_certified_event_blob = contract_service.last_certified_event_blob().await?;
```
halfprice, line 635: Hmm, I'm not sure if we should use the last_certified_event_blob on chain.

Here is my understanding: if the local node whose last certified event blob id < last_certified_event_blob_on_chain, the reset logic will only reset from last_certified_event_blob_on_chain onward. But for blobs between local_last_certified_blob and last_certified_event_blob_on_chain, they may still be corrupted, which leave the node not recovered. Is this possible?

crates/walrus-service/src/node.rs
```diff
@@ -811,7 +814,10 @@ impl StorageNode {
             return Ok(None);
         };
         let read_client = config.new_read_client().await?;
-        let system_package_id = read_client.get_system_package_id();
+        let retriable_client = read_client.sui_client();
+        let system_package_id = retriable_client
+            .get_package_id_from_object(read_client.get_system_object_id())
```
halfprice, line 819: remark: as @mlegner pointed out, here it should use `get_original_package_id()`?

crates/walrus-service/src/node.rs
sadhansood: Yes, this is possible. However, my understanding is that we cannot safely remove the blobs before global `last_certified_event_blob` in the current setup. I have an example which I used to help build this intuition myself:
1> Imagine a node X is joining the network and catching up on past events
2> In the past there was an incident which led to event blob certification outage for several days
3>  X forms some blobs locally for the outage period and get restarted
4> X will move the cursor back, reset the local state and form those blobs again. If it keeps getting restarted, it keeps reseting the local state and can never actually make progress.
5> To fix this, we should only ever reset the local state before `last_certified_event_blob` if we know that setting `NUM_CHECKPOINTS_PER_BLOB` actually changed after restart. So, we need to track previous and current setting which is a little bit tricky.
6> I think a better way to handle historical events in event blob writer is to skip past those completely without making blobs locally. But that is a more involved change which is why I avoided tackling the issue of blob corruption/inconsistency for blobs before `last_certified_event_blob`.

crates/walrus-service/src/node.rs
sadhansood: `get_original_package_id` is only available on package store which we don't have/need to import here since we can just also get the original package id from object struct tag

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: Yes, we will not recover from it (due to the same reason as noted above that we cannot always safely remove local blobs). A more general solution for recovery from local state being inconsistent should be done as a follow up.

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: Yes, we could have several such blobs possible. Bur, we will not delete those. in fact, i think if there is any single attested/failed_attest/pending blob whose `metadata.end <= last_certified_event_blob.ending_checkpoint_sequence_number` then we shouldn't reset the state locally at all. This is because in that case the blobs after  `last_certified_event_blob.ending_checkpoint_sequence_number` are all blocked by ones which are yet to be certified locally (but have been globally certified already) and not because event blob certification is overall stuck. I'll make this change.

crates/walrus-service/src/node.rs
halfprice: Thanks a lot for the explanation and chat!

crates/walrus-service/src/node.rs
halfprice: I'm not too familiar with this... Maybe verify this with @karlwuest or @mlegner ?

karlwuest: Yes, this works. The `get_package_id_from_object` takes the package ID from the object type, which is the original package ID in the case of the system object.

PR #1715, opened 2025-03-14
feat(node): improve setup command

mlegner:
## Description

- Add optional `--network-key-path` and `--certificate-path` options to the `setup` command to use externally created credentials.
- Remove options `--use-legacy-event-provider` and `--disable-event-blob-writer`, which should not be used anyway.

## Test plan

Manual tests.

---

## Release notes

- [x] Storage node: Added optional `--network-key-path` and `--certificate-path` options to the `setup` command, removed `--use-legacy-event-provider` and `--disable-event-blob-writer`. Adjusted defaults for storage prices and commission rate.

### DIFF
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -105,10 +105,10 @@ event_processor_config:
   event_stream_catchup_min_checkpoint_lag: 20000
 use_legacy_event_provider: false
 disable_event_blob_writer: false
-commission_rate: 0
+commission_rate: 6000
 voting_params:
-  storage_price: 100
-  write_price: 2000
+  storage_price: 100000
+  write_price: 20000
   node_capacity: 250000000000
 metadata:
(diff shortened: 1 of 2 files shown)


crates/walrus-service/bin/node.rs
```diff
@@ -307,7 +310,7 @@ struct ConfigArgs {
     /// Initial vote for the write price in FROST per MiB.
     #[clap(long, default_value_t = config::defaults::write_price())]
```
karlwuest, line 311: Question: Do we want to update the defaults for the write and storage prices to the values that we suggest for mainnet?

mlegner: Ah, good point.

mlegner: Done in 31154968d93a6fc4908f8d8c1046c6b319ab2290. Also set the default commission rate to 60%.

PR #1723, opened 2025-03-14
test: various improvements to speed up nodes & simtests

halfprice:
## Description

- SuiContractClient does not need to lock itself during the request if the request is reading from Sui.
- Creating storage node in parallel in simtest
- Resolving RocksDb `list_cf` call nondeterminism.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/contract_service.rs
+++ b/crates/walrus-service/src/node/contract_service.rs
@@ -120,7 +120,10 @@ pub trait SystemContractService: std::fmt::Debug + Sync + Send {
 /// A [`SystemContractService`] that uses a [`SuiContractClient`] for chain interactions.
 #[derive(Debug, Clone)]
 pub struct SuiSystemContractService {
-    contract_client: Arc<TokioMutex<SuiContractClient>>,
+    // A client for sending transactions to the contract.
+    contract_tx_client: Arc<TokioMutex<SuiContractClient>>,
+    // A client for reading from the contract. Note that reading from the contract can be done
+    // concurrently.
     read_client: Arc<SuiReadClient>,
     committee_service: Arc<dyn CommitteeService>,
     rng: Arc<StdMutex<StdRng>>,
@@ -142,7 +145,7 @@ impl SuiSystemContractService {
     ) -> Self {
         Self {
             read_client: contract_client.read_client.clone(),
-            contract_client: Arc::new(TokioMutex::new(contract_client)),
+            contract_tx_client: Arc::new(TokioMutex::new(contract_client)),
             committee_service,
             rng: Arc::new(StdMutex::new(StdRng::seed_from_u64(seed))),
         }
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/test_utils.rs
```diff
@@ -1910,21 +1910,29 @@ impl TestClusterBuilder {
...
-                    self.system_context.clone(),
-                    nondeterministic!(
-                        tempfile::tempdir().expect("temporary directory creation must succeed")
-                    ),
-                    start_node_from_beginning,
-                    disable_event_blob_writer,
-                )
-                .await?,
-            );
+            // Build and run the storage nodes in parallel.
```
mlegner, line 1913: Remark: AFAIU, this makes the futures concurrent but not parallel, as they are still executed in the same Tokio task. Is that understanding correct?

crates/walrus-service/src/node/contract_service.rs
```diff
@@ -120,6 +121,7 @@ pub trait SystemContractService: std::fmt::Debug + Sync + Send {
 #[derive(Debug, Clone)]
 pub struct SuiSystemContractService {
     contract_client: Arc<TokioMutex<SuiContractClient>>,
```
sadhansood: nit: I wonder if we should rename `contract_client` to `rw_client` or `write_client` to be more consistent

crates/walrus-service/src/test_utils.rs
```diff
@@ -1910,21 +1910,29 @@ impl TestClusterBuilder {
...
+            node_futures.push(T::build_and_run(
+                builder,
+                self.sui_cluster_handle.clone(),
+                self.system_context.clone(),
+                nondeterministic!(
+                    tempfile::tempdir().expect("temporary directory creation must succeed")
+                ),
+                start_node_from_beginning,
+                disable_event_blob_writer,
+            ));
```
sadhansood, line 1923: I'm curious if this part make the tests run faster? It seems like the nodes already run on different tokio tasks, the handle of which are stored and returned from `StorageNodeHandle`? Or it makes the initialization of the node(s) concurrent which makes the tests run faster?

crates/walrus-service/src/test_utils.rs
halfprice: Yes. That's right. Not parallel, but concurrent.

crates/walrus-service/src/test_utils.rs
halfprice: In the case were we drop 40min to 20min runtime, this is the main reason for the time reduction. It makes the storage node initialization concurrent. Without this, nodes initialization took 5~10 seconds, which means that when the last node starts to initialize, we are already in epoch 2+.

However, I still don't know what caused 20min to 40min. What I can tell is after adding this [call](https://github.com/MystenLabs/walrus/blob/e680eb753c377f4c041c8c8631f15f1e4ab76779/crates/walrus-service/src/node.rs#L640), the time increased. However, it is not about the mutex locking. Something else happens weird in CI.

crates/walrus-service/src/node/contract_service.rs
halfprice: I changed it to `contract_tx_client` indicating this is for executing transactions. I wanted to keep contract in the name since the struct is called `SuiContractClient`.

crates/walrus-service/src/test_utils.rs
sadhansood: Ah ok, yeah i suspected that call (although not sure why) since i saw the time to jump when i added it.

PR #1728, opened 2025-03-17
fix: fix nightly simtest

halfprice:
## Description

Nightly simtest were not running due to two issues
- The seed was set incorrectly
- the test filter was empty (probably a new requirement in rust 1.85)

I'll also try to add oncall to the nightly result soon.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/scripts/simtest/simtest-run.sh
+++ b/scripts/simtest/simtest-run.sh
@@ -20,7 +20,9 @@ if [ -z "$NUM_CPUS" ]; then
 fi
 
 DATE=$(date '+%Y%m%d_%H%M%S')
-SEED="$DATE"
+
+# Using a random seed derived from the current time for the simulator tests.
+SEED=$(date +%s)
 
 # create logs directory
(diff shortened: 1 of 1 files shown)


scripts/simtest/simtest-run.sh
```diff
@@ -20,7 +20,9 @@ if [ -z "$NUM_CPUS" ]; then
 fi
 
 DATE=$(date '+%Y%m%d_%H%M%S')
-SEED="$DATE"
+
+# Using a random seed derived from the current time for the simulator tests.
+SEED=$(date +%s)
```
mlegner, line 25: Question: Why did the previous version not work?

halfprice: The previous version has an underscore in it, like this `20250311_210225`, which throws `ParseIntError`.

PR #1735, opened 2025-03-18
feat: Add fallback client for checkpoints

sadhansood:
## Description

This PR adds code to read checkpoints from a fallback checkpoint bucket. One caveat is that `checkpoint_downlaoder.rs` now needs to use `RetriableRpcClient` which causes it (`walrus_util`) to take dependency on `walrus_sui` crate which already depends on `walrus_util`. So, in order to break this circular dependency, I've moved it to `walrus_service` which seems not a bad place for it (but as always open to suggestions).

## Test plan

Tested locally by running catchup on mainnet, and it downloads the checkpoints using fallback just fine using command: 
```
RUST_LOG=info,walrus_sui::client::retry_client=debug ./target/debug/walrus-node catchup --db-path /tmp/catchup --system-object-id 0x2134d52768ea07e8c43570ef975eb3e4c27a39fa6396bef985b5abc58d03ddd2 --staking-object-id 0x10b9d30c28448939ce6c4d6c6e0ffce4a7f8a4ada8248bdad09ef8b70e4a3904 --sui-rpc-url "https://mysten-rpc.mainnet.sui.io:443" --runtime-duration 1m --event-stream-catchup-min-checkpoint-lag 100000 --checkpoint-bucket https://checkpoints.mainnet.sui.io 2>&1 | tee /tmp/logs
```
```
2025-03-20T06:33:52.567676Z DEBUG walrus_sui::client::retry_client: Downloading checkpoint from: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("checkpoints.mainnet.sui.io")), port: None, path: "/122521999.chk", query: None, fragment: None }
2025-03-20T06:33:52.568676Z DEBUG walrus_sui::client::retry_client: Checkpoint download successful for sequence: 122521989
2025-03-20T06:33:52.570394Z DEBUG walrus_sui::client::retry_client: Checkpoint download successful for sequence: 122521995
2025-03-20T06:33:52.584979Z DEBUG walrus_sui::client::retry_client: Primary client error while fetching checkpoint: Status { code: NotFound, message: "data from requested checkpoint 122522000 has been pruned", details: b"\x08\x05\x128data from requested checkpoint 122522000 has been pruned", metadata: MetadataMap { headers: {"content-type": "application/grpc", "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000", "via": "1.1 google", "content-length": "0", "x-sui-chain-id": "35834a8a", "x-sui-chain": "mainnet", "x-sui-epoch": "707", "x-sui-checkpoint-height": "124635129", "x-sui-timestamp-ms": "1742452431897", "x-sui-timestamp": "2025-03-20T06:33:51.897Z", "x-sui-lowest-available-checkpoint": "123334496", "x-sui-lowest-available-checkpoint-objects": "123727549", "vary": "origin, access-control-request-method, access-control-request-headers", "access-control-allow-origin": "*", "server-timing": "finish_request;dur=0", "date": "Thu, 20 Mar 2025 06:33:52 GMT"} }, source: None }
2025-03-20T06:33:52.585017Z DEBUG walrus_sui::client::retry_client: Using fallback client to fetch checkpoint as error: Status { code: NotFound, message: "data from requested checkpoint 122522000 has been pruned", details: b"\x08\x05\x128data from requested checkpoint 122522000 has been pruned", metadata: MetadataMap { headers: {"content-type": "application/grpc", "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000", "via": "1.1 google", "content-length": "0", "x-sui-chain-id": "35834a8a", "x-sui-chain": "mainnet", "x-sui-epoch": "707", "x-sui-checkpoint-height": "124635129", "x-sui-timestamp-ms": "1742452431897", "x-sui-timestamp": "2025-03-20T06:33:51.897Z", "x-sui-lowest-available-checkpoint": "123334496", "x-sui-lowest-available-checkpoint-objects": "123727549", "vary": "origin, access-control-request-method, access-control-request-headers", "access-control-allow-origin": "*", "server-timing": "finish_request;dur=0", "date": "Thu, 20 Mar 2025 06:33:52 GMT"} }, source: None }
2025-03-20T06:33:52.585057Z DEBUG walrus_sui::client::retry_client: Downloading checkpoint from: Url { scheme: "https", cannot_be_a_base: false, username: "", password: None, host: Some(Domain("checkpoints.mainnet.sui.io")), port: None, path: "/122522000.chk", query: None, fragment: None }
2025-03-20T06:33:52.589472Z DEBUG walrus_sui::client::retry_client: Checkpoint download successful for sequence: 122521997
2025-03-20T06:33:52.590492Z DEBUG walrus_sui::client::retry_client: Checkpoint download successful for sequence: 122521998
2025-03-20T06:33:52.590693Z DEBUG walrus_sui::client::retry_client: Checkpoint download successful for sequence: 122521999
```

---

## Release notes

- [x] Storage node: Support a fallback bucket from which to get checkpoints if they are unavailable from the main RPC node. Can be activated by adding the following to the node configuration file (example for Mainnet):
  ```
  sui:
    ...
    rpc_fallback_config:
      checkpoint_bucket: https://checkpoints.mainnet.sui.io
  ```

### DIFF
--- a/crates/checkpoint-downloader/src/downloader.rs
+++ b/crates/checkpoint-downloader/src/downloader.rs
@@ -1,232 +1,39 @@
 // Copyright (c) Mysten Labs, Inc.
 // SPDX-License-Identifier: Apache-2.0
 
-//! Checkpoint downloader util.
-
+//! Checkpoint downloader.
 use std::{
     collections::{BTreeMap, BTreeSet},
-    fmt::Debug,
     sync::Arc,
     time::Duration,
 };
 
 use anyhow::{anyhow, Result};
-use prometheus::{register_int_gauge_with_registry, IntGauge, Registry};
+use prometheus::Registry;
 use rand::{rngs::StdRng, SeedableRng};
-use serde::{Deserialize, Serialize};
-use serde_with::{serde_as, DurationMilliSeconds, DurationSeconds};
-use sui_rpc_api::{client::ResponseExt, Client};
-use sui_types::{
-    full_checkpoint_content::CheckpointData,
-    messages_checkpoint::{CheckpointSequenceNumber, TrustedCheckpoint},
-};
+use sui_rpc_api::client::ResponseExt;
+use sui_types::messages_checkpoint::{CheckpointSequenceNumber, TrustedCheckpoint};
 use tokio::{
     sync::{mpsc, RwLock},
     time::Instant,
 };
 use tokio_util::sync::CancellationToken;
-use tonic::Status;
 use typed_store::{rocks::DBMap, Map};
+use walrus_sui::client::retry_client::{RetriableClientError, RetriableRpcClient};
+#[cfg(not(test))]
+use walrus_utils::backoff::ExponentialBackoff;
 
-/// Fetcher configuration options for the parallel checkpoint fetcher.
-#[serde_as]
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
-#[serde(default)]
-pub struct ParallelDownloaderConfig {
-    /// Number of retries per checkpoint before giving up.
-    pub min_retries: u32,
-    /// Initial delay before the first retry.
-    #[serde_as(as = "DurationMilliSeconds")]
-    #[serde(rename = "initial_delay_millis")]
-    pub initial_delay: Duration,
-    /// Maximum delay between retries. Once this delay is reached, the
-    /// fetcher will keep retrying with this duration.
-    #[serde_as(as = "DurationMilliSeconds")]
-    #[serde(rename = "max_delay_millis")]
-    pub max_delay: Duration,
-}
-
-impl Default for ParallelDownloaderConfig {
-    fn default() -> Self {
-        // Checkpoint downloader configuration tuned for:
-        // - Checkpoint generation rate: 1 per 200ms (5 per second)
-        // - Network RTT: ~100-300ms
-        //
-        // Retry timing example with 100ms RTT:
-        // T+0ms:   Client requests checkpoint N
-        // T+100ms: Client learns checkpoint N unavailable
-        // T+200ms: Server generates checkpoint N
-        // T+250ms: Client retries after retry_delay
-        // T+350ms: Client gets response which should contain checkpoint N.
-
-        // Retry timing example with 300ms RTT:
-        // T+0ms:   Client requests checkpoint N
-        // T+200ms: Server generates checkpoint N
-        // T+300ms: Client learns checkpoint N unavailable
-        // T+450ms: Client retries after retry_delay
-        // T+750ms: Client gets response which should contain checkpoint N.
-        // The retry_delay of 150ms is chosen to balance in steady state with
-        // fully caught up client:
-        // - Not retrying too quickly (which could waste network resources)
-        // - Not waiting too long (which would cause us to fall behind)
-        // TODO: Use adaptive retry delay based on RTT and error type (#1123)
-        Self {
-            min_retries: 10,
-            initial_delay: Duration::from_millis(150),
-            max_delay: Duration::from_secs(2),
-        }
-    }
-}
-
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
-pub struct ChannelConfig {
-    /// Additional buffer factor for work queue
-    /// Helps keep the worker busy in case of delays
-    /// in draining the checkpoint queue.
-    pub work_queue_buffer_factor: usize,
-    /// Buffer factor for result queue
-    /// Helps handle delays in processing the results.
-    pub result_queue_buffer_factor: usize,
-}
-
-impl Default for ChannelConfig {
-    fn default() -> Self {
-        Self {
-            work_queue_buffer_factor: 3,
-            result_queue_buffer_factor: 3,
-        }
-    }
-}
-
-/// Configuration for the adaptive worker pool.
-#[serde_as]
-#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
-#[serde(default)]
-pub struct AdaptiveDownloaderConfig {
-    /// Minimum number of workers.
-    pub min_workers: usize,
-    /// Maximum number of workers.
-    pub max_workers: usize,
-    /// Initial number of workers.
-    pub initial_workers: usize,
-    /// Checkpoint lag threshold for scaling up.
-    pub scale_up_lag_threshold: u64,
-    /// Checkpoint lag threshold for scaling down.
-    pub scale_down_lag_threshold: u64,
-    /// Minimum time between scaling operations.
-    #[serde(rename = "scale_cooldown_secs")]
-    #[serde_as(as = "DurationSeconds")]
-    pub scale_cooldown: Duration,
-    /// Base configuration.
-    pub base_config: ParallelDownloaderConfig,
-    /// Channel configuration.
-    pub channel_config: ChannelConfig,
-}
-
-impl Default for AdaptiveDownloaderConfig {
-    fn default() -> Self {
-        Self {
-            min_workers: 1,
-            max_workers: 10,
-            initial_workers: 5,
-            scale_up_lag_threshold: 100,
-            scale_down_lag_threshold: 10,
-            scale_cooldown: Duration::from_secs(10),
-            base_config: ParallelDownloaderConfig::default(),
-            channel_config: ChannelConfig::default(),
-        }
-    }
-}
-
-impl AdaptiveDownloaderConfig {
-    fn message_queue_size(&self) -> usize {
-        self.max_workers * self.channel_config.work_queue_buffer_factor
-    }
-
-    fn checkpoint_queue_size(&self) -> usize {
-        self.max_workers * self.channel_config.work_queue_buffer_factor
-    }
-
-    fn result_queue_size(&self) -> usize {
-        self.max_workers
-            * self.channel_config.work_queue_buffer_factor
-            * self.channel_config.result_queue_buffer_factor
-    }
-}
-
-/// Metrics for the event processor.
-#[derive(Clone, Debug)]
-struct AdaptiveDownloaderMetrics {
-    /// The number of checkpoint downloader workers.
-    pub num_workers: IntGauge,
-    /// The current checkpoint lag between the local store and the full node.
-    pub checkpoint_lag: IntGauge,
-}
-
-impl AdaptiveDownloaderMetrics {
-    pub fn new(registry: &Registry) -> Self {
-        Self {
-            num_workers: register_int_gauge_with_registry!(
-                "checkpoint_downloader_num_workers",
-                "Number of workers active in the worker pool",
-                registry,
-            )
-            .expect("this is a valid metrics registration"),
-            checkpoint_lag: register_int_gauge_with_registry!(
-                "checkpoint_downloader_checkpoint_lag",
-                "Current checkpoint lag between local store and full node",
-                registry,
-            )
-            .expect("this is a valid metrics registration"),
-        }
-    }
-}
-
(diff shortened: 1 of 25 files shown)


crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -221,6 +222,8 @@ pub struct EventProcessorRuntimeConfig {
     pub event_polling_interval: Duration,
     /// The path to the database.
     pub db_path: PathBuf,
+    /// The path to the checkpoint download client.
```
mlegner:
```suggestion
    /// The URL for the fallback source of Sui checkpoints.
```

deny.toml
```diff
@@ -112,6 +112,7 @@ unknown-git = "deny"
 allow-git = [
   "https://github.com/asonnino/prometheus-parser",
   "https://github.com/bmwill/axum-server",
+  "https://github.com/mystenlabs/anemo.git",
```
mlegner: Question: Why is this needed? We already allow all of `MystenLabs` below.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -128,6 +137,17 @@ impl RetriableRpcError for tonic::Status {
...
 }
 
+impl RetriableRpcError for CheckpointDownloadError {
+    fn is_retriable_rpc_error(&self) -> bool {
+        match self {
+            CheckpointDownloadError::RequestFailed(error) => error
+                .status()
+                .map(|status| status.is_server_error())
+                .unwrap_or(false),
+            _ => false,
```
mlegner: Question: How do we deal with 429 errors?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -780,13 +800,65 @@ impl RetriableSuiClient {
...
+    pub async fn download(
+        &self,
+        sequence_number: u64,
+    ) -> Result<CheckpointData, CheckpointDownloadError> {
+        let url = self.base_url.join(&format!("{}.chk", sequence_number))?;
+        let response = self.client.get(url).send().await?;
+        let bytes = response.bytes().await?;
+        let checkpoint_data = Blob::from_bytes::<CheckpointData>(&bytes)
+            .map_err(|e| CheckpointDownloadError::DeserializationError(e.to_string()))?;
+        Ok(checkpoint_data)
```
mlegner: Nit: Why not the following?
```suggestion
        Blob::from_bytes::<CheckpointData>(&bytes)
            .map_err(|e| CheckpointDownloadError::DeserializationError(e.to_string()))?
```

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -818,10 +897,50 @@ impl RetriableRpcClient {
         &self,
         sequence: u64,
     ) -> Result<CheckpointData, tonic::Status> {
```
mlegner: Remark: It might be better to use some custom error type here instead of converting the fallback error to a `tonic::Status`.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -818,10 +897,50 @@ impl RetriableRpcClient {
...
+                fallback.download(sequence).await.map_err(Into::into)
+            })
+            .await;
+        }
+
+        let short_backoff = ExponentialBackoffConfig {
+            min_backoff: std::time::Duration::from_millis(100),
+            max_backoff: std::time::Duration::from_millis(300),
+            max_retries: Some(3),
+        };
```
mlegner: Suggestion: Can we make this configurable?

crates/walrus-sui/src/client/retry_client.rs
sadhansood: Good point, but so far I've not seen cloud buckets fail requests with 429. But I will add it to be safe

crates/walrus-sui/src/client/retry_client.rs
sadhansood: ok sure, will do

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -818,10 +897,50 @@ impl RetriableRpcClient {
...
     ) -> Result<CheckpointData, tonic::Status> {
-        retry_rpc_errors(self.get_strategy(), || async {
+        let primary_result = self.client.get_full_checkpoint(sequence).await;
+
+        let Err(err) = &primary_result else {
+            return primary_result;
+        };
+
+        let Some(ref fallback) = self.fallback_client else {
+            return retry_rpc_errors(self.get_strategy(), || async {
```
halfprice, line 975: Question: when fallback client is not set, do we always retry 1 more time as before? If so, can we document this behavior? It may not be an issue if retries are set >= 2, but when it is set to 0 or 1, it may be an surprising behavior.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -818,10 +897,50 @@ impl RetriableRpcClient {
...
             self.client.get_full_checkpoint(sequence).await
         })
-        .await
+        .await;
+
+        if short_retry_result.is_ok() {
+            return short_retry_result;
+        }
+
+        return retry_rpc_errors(self.get_strategy(), || async {
```
halfprice: Question: say the fullnode is down, it seems that we retry the fullnode for every checkpoint, and upon failure, then fallback to the archival. Is this the current behavior? If so, I wonder if we can directly use the archival for some period until retry with the fullnode. This may also affect how many checkpoint downloader we need. It's a combination of factors of # of checkpoint per seconds (4), timeout period, and latency of downloading checkpoints from the bucket.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: I agree, this is fixed now

crates/walrus-sui/src/client/retry_client.rs
sadhansood: sure, done

crates/walrus-sui/src/client/retry_client.rs
sadhansood: Yes, good callout - updated documentation on the function to reflect this.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: Yes I agree, this would be ideal. We could fallback to archives, and in the background poll the FN to see if it is back up. We should consider implementing it as a follow up.

deny.toml
sadhansood: ah, removed

crates/walrus-utils/Cargo.toml
```diff
@@ -7,23 +7,6 @@ license.workspace = true
...
-  "dep:rand",
-  "dep:rocksdb",
-  "dep:serde",
-  "dep:serde_with",
-  "dep:sui-rpc-api",
-  "dep:sui-types",
-  "dep:tokio",
-  "dep:tokio-util",
-  "dep:tonic",
-  "dep:typed-store",
```
mlegner, line 24: Remark: I think several of these dependencies could now be removed.

crates/walrus-utils/src/config.rs
```diff
@@ -1,10 +1,75 @@
...
+                    .unwrap_or(Duration::from_millis(300)),
+                max_retries: self.max_retries,
+            };
+            RpcFallbackConfig {
+                checkpoint_bucket: url.clone(),
+                quick_retry_config: backoff,
+            }
+        })
+    }
+}
```
mlegner: Suggestion: Keep this in the same place as the downloader code.

crates/walrus-utils/src/config.rs
```diff
@@ -1,10 +1,75 @@
...
+
+/// Configuration for the RPC endpoint fallback.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
+pub struct RpcFallbackConfig {
+    /// The endpoint of the checkpoint bucket that will be used to download the checkpoint if the
+    /// RPC endpoint is not available.
+    pub checkpoint_bucket: reqwest::Url,
+    /// The configuration for the backoff strategy for the RPC endpoint before falling back to the
+    /// checkpoint bucket.
+    #[serde(default)]
```
mlegner: Question: I think you used different values before making them configurable, correct? In this case, maybe you can define a function defining these defaults instead of just taking the `ExponentialBackoffConfig::default()`.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -128,6 +140,30 @@ impl RetriableRpcError for tonic::Status {
...
+}
+
+impl RetriableRpcError for RetriableClientError {
+    fn is_retriable_rpc_error(&self) -> bool {
+        match self {
+            RetriableClientError::RpcError(status) => status.is_retriable_rpc_error(),
+            RetriableClientError::FallbackError(fallback_error) => {
+                fallback_error.is_retriable_rpc_error()
+            }
+            RetriableClientError::Other(_) => true,
```
halfprice: Question: should here return false (generic error not retriable)?

Also, https://github.com/MystenLabs/walrus/blob/4947c27b4e568746ae4ee7fa64569f3e4023a187/crates/walrus-sui/src/client/retry_client.rs#L92, can just call it here as well.

crates/walrus-utils/src/config.rs
sadhansood: ok good catch, thanks @mlegner !

crates/walrus-sui/src/client/retry_client.rs
sadhansood: good catch @halfprice !

crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -576,48 +383,54 @@ impl ParallelCheckpointDownloaderInner {
...
+    match err {
+        Some(RetriableClientError::RpcError(status)) => {
+            if let Some(checkpoint_height) = status.checkpoint_height() {
+                if next_checkpoint > checkpoint_height {
+                    tracing::trace!(
+                        next_checkpoint,
+                        checkpoint_height,
+                        message = status.message(),
+                        "failed to read next checkpoint, probably not produced yet",
+                    );
```
halfprice: suggestions: I think we are missing some possibly important log here: we don't log when the RpcError does not return checkpoint height or the height is greater. Can we do things similar as before? Just return here, and move `tracing::warn!(next_checkpoint, ?err, "failed to read next checkpoint");` out of the match, so that it is logged every other case.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -780,46 +813,217 @@ impl RetriableSuiClient {
...
+// Implement automatic conversion from CheckpointDownloadError to tonic::Status
+impl From<CheckpointDownloadError> for tonic::Status {
+    fn from(error: CheckpointDownloadError) -> Self {
+        tonic::Status::internal(format!("Checkpoint download failed: {}", error))
+    }
+}
+
+/// A client for downloading checkpoint data from a remote server.
+#[derive(Clone, Debug)]
+pub struct CheckpointBucketClient {
```
halfprice, line 841: Suggestion: feel `CheckpointBucketClient` should be in the new `checkpoint_downloader` crate (but no need to do it in this PR).

crates/walrus-utils/src/config.rs
sadhansood: done

crates/checkpoint-downloader/src/downloader.rs
sadhansood: fixed

PR #1753, opened 2025-03-19
feat: db-tool

liquid-helium:
## Description

Closes WAL-703

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -59,7 +59,7 @@ pub(super) use event_cursor_table::EventProgress;
 mod event_sequencer;
 mod shard;
 
-pub use shard::{
+pub(crate) use shard::{
     pending_recover_slivers_column_family_options,
     primary_slivers_column_family_options,
     secondary_slivers_column_family_options,


crates/walrus-service/src/node/dbtool.rs
```diff
@@ -237,3 +356,168 @@ fn count_certified_blobs(db_path: PathBuf, epoch: Epoch) -> Result<()> {
...
+
+    Ok(())
+}
+
+fn list_column_families(db_path: PathBuf) -> Result<()> {
+    let result = rocksdb::DB::list_cf(&RocksdbOptions::default(), db_path);
+    if let Ok(column_families) = result {
+        println!("Column families: {:?}", column_families);
+    } else {
+        println!("Failed to get column families: {:?}", result);
```
halfprice, line 460: nit: return the error here?

crates/walrus-service/src/node/dbtool.rs
```diff
@@ -237,3 +356,168 @@ fn count_certified_blobs(db_path: PathBuf, epoch: Epoch) -> Result<()> {
...
+    for result in iter.take(count as usize) {
+        match result {
+            Ok((key, value)) => {
+                let blob_id: BlobId = bcs::from_bytes(&key)?;
+                let metadata: BlobMetadata = bcs::from_bytes(&value)?;
+                println!("Blob ID: {}, Metadata: {:?}", blob_id, metadata);
+            }
+            Err(e) => {
+                println!("Error: {:?}", e);
+                break;
```
halfprice: Nit: also return error here

crates/walrus-service/src/node/dbtool.rs
```diff
@@ -71,7 +83,10 @@ pub enum DbToolCommands {
         #[clap(long)]
         db_path: PathBuf,
         /// Object ID to read.
-        object_id: ObjectID,
+        object_id: Option<ObjectID>,
+        /// Count of objects to read.
+        #[clap(long, default_value = "1")]
+        count: u64,
```
halfprice, line 107: Suggestion: since we added a scan for reading object, we may as well add a count for reading blob info (ReadBlobInfo command).

crates/walrus-service/src/node/storage/constants.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/// Returns the name of the pending recover slivers column family for the given shard.
+pub fn pending_recover_slivers_column_family_name(id: ShardIndex) -> String {
+    format!(
+        "{}/{}",
+        base_column_family_name(id),
+        PENDING_RECOVER_SLIVERS_COLUMN_FAMILY_NAME
+    )
+}
```
halfprice: Suggestion: TBH, I'm a little afraid of changes like this because if the cf name change, we lost state from the previous cf for the node to behave correctly, and also the new cf (if it is wrong) may stay forever. In this PR, can we still use the old way of getting CF names? There are also some benefit of old structure, for example, the STATUS_COLUMN_FAMILY_NAME was in the scope of a shard, which indicates that the status of the shard.

Later if desired, we can again make this change, but I think we need more careful testing to make sure after upgrades, nothing breaks. To test this, a simple upgrade in PTN may not be enough because not all column families are constantly used (e.g. those are used only during shard moves).

liquid-helium: Thanks for sharing your concern!
I really think this is a good time to make this change since waiting might make things trickier down the line. The good news is we can easily check it now; I’ve added a test to compare the names of the `before` and `after` column families.

Looking ahead, I still believe that unifying the names will be way more beneficial.

crates/walrus-service/src/node/dbtool.rs
liquid-helium: added count and start_key for read_blob_info.

read_object_blob_info is already a scan, I renamed the input to start_object_id to make it clear.

crates/walrus-service/src/node/dbtool.rs
liquid-helium: Thanks for the tip, change applied.

crates/walrus-service/src/node/dbtool.rs
liquid-helium: Change applied.

crates/walrus-service/src/node/storage/constants.rs
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+use walrus_core::ShardIndex;
+
+/// Column family names used in RocksDB.
+const EVENT_STORE: &str = "event_store";
```
halfprice: I think this is double defined? The original definition is here: https://github.com/MystenLabs/walrus/blob/e9049f6df50b050188289f9b46e49bd4acc227b8/crates/walrus-service/src/node/events/event_processor.rs#L93, which doesn't seem to be removed.

Maybe leave the EVENT_STORE as before this PR, since it probably should stick together with other EventProcessor column families. EVENT_STORE is part of the `events` DB, not the node DB.

crates/walrus-service/src/node/storage/constants.rs
```diff
@@ -0,0 +1,133 @@
...
+const PER_OBJECT_BLOB_INFO_COLUMN_FAMILY_NAME: &str = "per_object_blob_info";
+const NODE_STATUS_COLUMN_FAMILY_NAME: &str = "node_status";
+const METADATA_COLUMN_FAMILY_NAME: &str = "metadata";
+const EVENT_INDEX_COLUMN_FAMILY_NAME: &str = "latest_handled_event_index";
+const BASE_COLUMN_FAMILY_NAME: &str = "shard";
+const PRIMARY_SLIVERS_COLUMN_FAMILY_NAME: &str = "primary-slivers";
+const SECONDARY_SLIVERS_COLUMN_FAMILY_NAME: &str = "secondary-slivers";
+const STATUS_COLUMN_FAMILY_NAME: &str = "status";
+const SYNC_PROGRESS_COLUMN_FAMILY_NAME: &str = "sync-progress";
+const PENDING_RECOVER_SLIVERS_COLUMN_FAMILY_NAME: &str = "pending-recover-slivers";
```
halfprice: Suggestion: can we add a `SHARD_` prefix for these 5 names? These are per shard, and I think particularly the STATUS_COLUMN_FAMILY_NAME may be ambiguous.

crates/walrus-service/src/node/storage/constants.rs
```diff
@@ -0,0 +1,133 @@
...
+        PENDING_RECOVER_SLIVERS_COLUMN_FAMILY_NAME
+    )
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_column_family_names() {
```
halfprice, line 112: nice!

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -55,10 +55,14 @@ use crate::node::{
     StorageNodeInner,
 };
 
-const CERTIFIED: &str = "certified_blob_store";
-const ATTESTED: &str = "attested_blob_store";
-const PENDING: &str = "pending_blob_store";
-const FAILED_TO_ATTEST: &str = "failed_to_attest_blob_store";
+/// The column family name for certified event blobs.
+pub const CERTIFIED: &str = "certified_blob_store";
```
halfprice: nit: pub(crate)

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -78,16 +82,16 @@ pub struct EventBlobMetadata<T, U> {
 }
 
 /// Metadata for a blob that is waiting for attestation.
-type PendingEventBlobMetadata = EventBlobMetadata<CheckpointSequenceNumber, ()>;
+pub type PendingEventBlobMetadata = EventBlobMetadata<CheckpointSequenceNumber, ()>;
```
halfprice: nit: pub(crate)

crates/walrus-service/src/node/storage.rs
```diff
@@ -541,17 +578,7 @@ impl Storage {
...
-            db_config.node_status().to_options(),
-        )
-    }
-
-    fn metadata_options(db_config: &DatabaseConfig) -> (&'static str, Options) {
-        (
-            Self::METADATA_COLUMN_FAMILY_NAME,
-            db_config.metadata().to_options(),
-        )
+        (node_status_cf_name(), db_config.node_status().to_options())
```
halfprice: Suggestion: `node_status` and `metadata` cfs are the two top level cfs and I think it's better to keep their options setting together. We can either move the node_status_options to the new location, or move the metadata_options back here, and use the default options in the DB tool when reading the metadata cf.

crates/walrus-service/src/node/storage.rs
```diff
@@ -149,17 +170,33 @@ impl Storage {
...
                 ]
             })
             .collect::<Vec<_>>();
 
         let (node_status_cf_name, node_status_options) = Self::node_status_options(&db_config);
-        let (metadata_cf_name, metadata_options) = Self::metadata_options(&db_config);
+        let metadata_options = metadata_options(&db_config);
+        let metadata_cf_name = metadata_cf_name();
         let blob_info_column_families = BlobInfoTable::options(&db_config);
         let (event_cursor_cf_name, event_cursor_options) = EventCursorTable::options(&db_config);
```
halfprice, line 206: Suggestion: also move the event cursor column family name to constants.rs. Right now, we move all the cfs name in the node storage DB to constants.rs except the event cursor.

crates/walrus-service/src/node/storage/constants.rs
halfprice: Sounds good, and thanks for adding the test! It looks good to me!

crates/walrus-service/src/node/storage/constants.rs
liquid-helium: That's a good idea! Thanks, change applied.

crates/walrus-service/src/node/storage/constants.rs
liquid-helium: That makes sense, change applied, thanks for the tip.

crates/walrus-service/src/node/events/event_blob_writer.rs
liquid-helium: Done.

crates/walrus-service/src/node/storage.rs
liquid-helium: Nice catch, thanks! I moved it to the new location, and make it consistent with the metadata_options.

crates/walrus-service/src/node/storage.rs
liquid-helium: Good idea, done.

PR #1754, opened 2025-03-19
feat: Quilt encoding part I

liquid-helium:
## Description

Related to WAL-685.
Contributes to WAL-858

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-core/Cargo.toml
```diff
@@ -16,8 +16,10 @@ test-utils = ["walrus-test-utils"]
 [dependencies]
 base64.workspace = true
 bcs.workspace = true
+crc = "3.2.1"
```
mlegner: Nit: Can we also set this in the workspace Cargo.toml file?

crates/walrus-core/src/encoding/errors.rs
```diff
@@ -145,3 +151,91 @@ pub enum SymbolVerificationError {
...
+    }
+    /// Index is out of bounds.
+    pub fn index_out_of_bounds(index: usize, max_index: usize) -> Self {
+        Self::IndexOutOfBounds(index, max_index)
+    }
+    /// Other error.
+    pub fn other(reason: String) -> Self {
+        Self::Other(reason)
+    }
+}
```
mlegner: Question: Why are these functions needed? One can simply instantiate the corresponding variants directly.

crates/walrus-core/src/metadata.rs
```diff
@@ -45,6 +46,163 @@ pub enum VerificationError {
...
+
+    /// Returns the CRC8 checksum of the block.
+    pub fn crc8(&self) -> u8 {
+        self.crc8
+    }
+
+    /// Sets the CRC8 checksum of the block.
+    pub fn set_crc8(&mut self, crc8: u8) {
+        self.crc8 = crc8;
+    }
```
mlegner: Question: Why not just make the fields `pub` if they can anyway be set and read individually?

crates/walrus-core/src/metadata.rs
```diff
@@ -45,6 +46,163 @@ pub enum VerificationError {
...
+pub enum QuiltVersion {
+    /// Version 1 of the quilt index format.
+    #[default]
+    V1 = 0,
+}
+
+/// A index over the blobs in a quilt.
+///
+/// Each [QuiltBlock] represents a blob stored in the quilt. And each blob is
+/// mapped to a continuous index range.
```
mlegner: Nit: "continuous" -> "contiguous"?
```suggestion
/// mapped to a contiguous index range.
```

crates/walrus-core/src/metadata.rs
```diff
@@ -45,6 +46,163 @@ pub enum VerificationError {
...
+    unencoded_length: u64,
+    /// The start sliver index of the block.
+    #[serde(skip)]
+    start_index: u16,
+    /// The end sliver index of the block.
+    end_index: u16,
+    /// The identifier of the block, it can be used to locate the blob in the quilt.
+    identifier: String,
+    /// A CRC8 checksum of the block.
+    crc8: u8,
```
mlegner: Question: What is the purpose of this?

crates/walrus-core/src/metadata.rs
```diff
@@ -45,6 +46,163 @@ pub enum VerificationError {
...
     UnencodedLengthTooLarge,
 }
+/// Represents a blob quilted into a single quilt blob.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct QuiltBlock {
+    /// The unencoded length of the blob.
+    unencoded_length: u64,
+    /// The start sliver index of the block.
+    #[serde(skip)]
+    start_index: u16,
```
mlegner: Question: Why is the `start_index` skipped? Because it can be inferred from the `unencoded_length` and `end_index`? If so, it would feel more intuitive to me to skip the `end_index` instead.

crates/walrus-core/src/metadata.rs
```diff
@@ -45,6 +46,163 @@ pub enum VerificationError {
...
+    /// Returns the [`BlobMetadata`] of the quilt.
+    pub fn metadata(&self) -> &BlobMetadata {
+        &self.metadata
+    }
+
+    /// Returns the [`BlobId`] of the quilt.
+    pub fn blob_id(&self) -> &BlobId {
+        &self.quilt_id
+    }
+}
```
mlegner: Question: Why are those functions needed?

crates/walrus-core/src/metadata.rs
```diff
@@ -45,6 +46,163 @@ pub enum VerificationError {
...
+        &self,
+        identifier: &str,
+    ) -> Result<&QuiltBlock, QuiltError> {
+        self.quilt_blocks
+            .iter()
+            .find(|block| block.identifier == identifier)
+            .ok_or(QuiltError::blob_not_found_in_quilt(&identifier))
+    }
+
+    /// Returns an iterator over (blob_id, blob_identifier) pairs in the quilt.
```
mlegner: The docstring doesn't seem to match the function signature.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1367 @@
...
+    ///
+    /// `quilt_blob` is a quilt constructed from a set of blobs.
+    /// This function loads the quilt blob to access its internal structures without having
+    /// to re-encode the blobs.
+    fn new_from_quilt_blob(
+        &self,
+        quilt_blob: Vec<u8>,
+        metadata: &QuiltMetadata,
+        n_shards: NonZeroU16,
+    ) -> Result<V::Quilt, QuiltError>;
```
mlegner: Question/remark: IIUC, this is something like a "parse" or "deserialize" function.

crates/walrus-core/src/metadata.rs
```diff
@@ -45,6 +46,163 @@ pub enum VerificationError {
...
+        for i in 1..self.quilt_blocks.len() {
+            let prev_end_index = self.quilt_blocks[i - 1].end_index;
+            self.quilt_blocks[i].start_index = prev_end_index;
+        }
+    }
+}
+
+/// Metadata associated with a quilt.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct QuiltMetadata {
```
mlegner: Suggestion: Call this `QuiltMetadataV1`.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1367 @@
...
+        // assert_eq!(self, &QuiltConfig::V1);
+        let encoding_config = EncodingConfig::new(n_shards);
+        let config = encoding_config.get_for_type(metadata.metadata().encoding_type());
+
+        let n_primary_source_symbols = config.n_primary_source_symbols().get();
+        let n_secondary_source_symbols = config.n_secondary_source_symbols().get();
+        let n_source_symbols = n_primary_source_symbols * n_secondary_source_symbols;
+
+        // Verify data alignment.
+        if quilt_blob.len() % n_source_symbols as usize != 0 {
```
mlegner: General remark: Try to replace ` as ` conversions by something like `from()`/`into()` or `try_from()/try_into()` to make it clear that/why the conversion is correct.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1367 @@
...
+            quilt_index,
+            symbol_size,
+        })
+    }
+}
+
+/// A quilt is a collection of encoded blobs stored together in a unified structure.
+///
+/// The data is organized as a 2D matrix where:
+/// - Each blob occupies a continuous range of columns (secondary slivers).
```
mlegner: Question: To use primary slivers instead, would we define a different quilt version?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1367 @@
...
+
+    /// Constructs a quilt from a quilt blob.
+    ///
+    /// `quilt_blob` is a quilt constructed from a set of blobs.
+    /// This function loads the quilt blob to access its internal structures without having
+    /// to re-encode the blobs.
+    fn new_from_quilt_blob(
+        &self,
+        quilt_blob: Vec<u8>,
+        metadata: &QuiltMetadata,
```
mlegner: Question: Why do we have to pass in the metadata and where would we get it from?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1367 @@
...
+            symbol_size,
+        })
+    }
+}
+
+/// A quilt is a collection of encoded blobs stored together in a unified structure.
+///
+/// The data is organized as a 2D matrix where:
+/// - Each blob occupies a continuous range of columns (secondary slivers).
+/// - The first column's initial `QUILT_INDEX_SIZE_PREFIX_SIZE` bytes contain the unencoded
```
mlegner, line 288: Suggestion: Explicitly include the quilt version at the very beginning of the raw data. That way we wouldn't need any additional information to parse a quilt.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1367 @@
...
+///   length of the [`QuiltIndexV1`]. It is guaranteed the column size is more than
+///   `QUILT_INDEX_SIZE_PREFIX_SIZE`.
+/// - The [`QuiltIndexV1`] is stored in the first one or multiple columns.
+/// - The blob layout is defined by the [`QuiltIndexV1`].
+#[derive(Default, Serialize, Deserialize, PartialEq, Eq)]
+pub struct QuiltV1 {
```
mlegner, line 300: Suggestion: Explicitly specify any invariants that hold for this. For example

```
// INV: `data.len()` is an integer multiple of `row_size * symbol_size`.
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1367 @@
...
+            span: tracing::span!(Level::ERROR, "QuiltEncoderV1"),
+        }
+    }
+}
+
+/// Finds the minimum symbol size needed to store blobs in a fixed number of columns.
+/// Each blob must be stored in consecutive columns exclusively.
+///
+/// The first column is used to store the quilt index, so the symbol size has to be
+/// large enough to store the quilt index in a single column.
```
mlegner: Important: Above you state that the index can occupy one or multiple columns. Can you clarify which is correct?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1367 @@
...
+/// * `blobs_sizes` - Slice of blob lengths.
+/// * `nc` - Number of columns available.
+/// * `nr` - Number of rows available.
+///
+/// # Returns
+/// * `Result<usize, QuiltError>` - The minimum symbol size needed, or an error if impossible.
+fn compute_symbol_size(
+    blobs_sizes: &[usize],
+    nc: usize,
+    nr: usize,
```
mlegner: Suggestion: Use clearer names like `n_columns` and `n_rows`.

mlegner (commented):
Thanks a lot for this PR, @liquid-helium. I made a pass (without looking at all the small details) and like the general approach. 💯 

My main recommendation would be to make the raw format of the quilt self-explanatory by including the version number and anything else that is needed for parsing at the beginning. Also, we should only include checksums or hashes if there is some use for them.

crates/walrus-core/src/metadata.rs
liquid-helium: this is something I am also discussing with @karlwuest 

The goal here is to read a blob/file from the quilt without reading all the slivers, and to catch any potential bugs/corruptions.

Suppose there is a bug or memory corruption in the quilt construction code, after receiving the raw blobs from the client, and before encoding the quilt blob.
The corruption won’t be detected, and the quilt blob will be consistent.

When we read the slivers with the corrupted file, we won’t notice it.

The possibility of either bug or memory corruption could be really low, but being paranoid, it is nice to have a hash from the un-processed data, and compare it with the decode data.

We can also do the verification right after encoding the quilt, and compare the hashes store in memory without storing it, but that means more computation in the write path.

I chose a 1-byte crc, so the storage overhead should be minimal.

mlegner: Hmm... If we take the possibility of memory corruption seriously, we probably need to change much more in our whole codebase. I don't think we should add this data just for this purpose, even if the size is pretty small.

That said, I'm not generally opposed to having additional per-subblob metadata, if it serves an important purpose.

crates/walrus-core/src/metadata.rs
liquid-helium: >Question: Why is the start_index skipped? Because it can be inferred from the unencoded_length and end_index?

Currently, it relies on all QuiltBlock has an end index, so after sorting them, and knowing the first start index from the QuiltIndex, we know their start indices as well. We don't need extra information outside of QuiltIndex to decode it.

>If so, it would feel more intuitive to me to skip the end_index instead.

The issue is when the last columns are empty, we won't know where the last blob ends.

crates/walrus-core/src/metadata.rs
liquid-helium: Another usage is to detect bug in the quilt constructing code, we can run many tests, but it would be good we can test it in any arbitrary tests, without requiring the tests to explicitly validate the correctness of the blobs by remembering blobs it has written. For example, with the stress-client.
We can keep this field optional, and populate it when needed.

crates/walrus-core/src/encoding/quilt_encoding.rs

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Thanks for the catch, this is a leftover some changes from a previous version.
Now `max_num_columns_for_quilt_index` is used to specify the max number of columns that could be used to store the index.
The current value is set as `MAX_NUM_COLUMNS_FOR_QUILT_INDEX = 1`.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good idea.
A 8-bit type is added to the raw data.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: This API is used when we failed to get the necessary slivers from the corresponding nodes.

Ideally we can recover the slivers from the symbols, but as stated in a previous discussion, this could incur heavy proof calculation load on the storage nodes.

Before we have a concrete solution, this is a simple solution to recover the slivers from the entire quilt blob, by reading the quilt blob, and then pass the blob to the Quilt.

the Metadata is already there due to the previous operations, so just pass it in.

crates/walrus-core/src/metadata.rs
liquid-helium: Removed.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: It is.
Renamed as `parse_from_quilt_blob`

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: I updated the 'as usize/u16' with try_from()/from() except some obvious cases. Thanks for the tip.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: We could reuse some code, but the data layout will be different, so another quilt version might be a clean solution.

Is there a particular motivation to use the primary slivers?

crates/walrus-core/src/metadata.rs
liquid-helium: Done.

crates/walrus-core/src/metadata.rs
liquid-helium: Thanks for the catch, fixed.

crates/walrus-core/src/metadata.rs
liquid-helium: Done.

crates/walrus-core/src/metadata.rs
liquid-helium: A habit from C++ to hide everything and provide setter/getter. :) 
I made all of the public and removed the functions, except for identifier, which is not supposed to be changed once constructed, since it will be used to calculate the size of the encoded QuiltIndex, while other fields all have a fixed size.

crates/walrus-core/Cargo.toml
liquid-helium: Removed, since we won't use it.

crates/walrus-core/src/encoding/errors.rs
liquid-helium: Change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Done.

crates/walrus-core/src/encoding/errors.rs
```diff
@@ -145,3 +148,41 @@ pub enum SymbolVerificationError {
     #[error("the symbol is not one of the symbols along the identified sliver's axis")]
     SymbolNotUsable,
 }
+
+/// Errors that may be encountered while interacting with quilt.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum QuiltError {
+    /// The quilt version is not supported.
+    // #[error("invalid quilt version, expected {0:?}, got {1:?}")]
+    // InvalidQuiltVersion(QuiltVersion, QuiltVersion),
```
mlegner: Nit: If this is not used, we should just remove it.

crates/walrus-core/src/metadata.rs
```diff
@@ -46,6 +50,137 @@ pub enum VerificationError {
...
+pub struct QuiltPatchV1 {
+    /// The unencoded length of the blob.
+    unencoded_length: u64,
+    /// The start sliver index of the blob.
+    #[serde(skip)]
+    pub start_index: u16,
+    /// The end sliver index of the blob.
+    pub end_index: u16,
+    /// The identifier of the blob, it can be used to locate the blob in the quilt.
+    identifier: String,
```
mlegner: Question: What is the reason to keep the `identifier` and `unencoded_length` private while the indices are public?

crates/walrus-core/src/metadata.rs
```diff
@@ -46,6 +50,137 @@ pub enum VerificationError {
...
+    /// QuiltIndexV1.
+    V1(QuiltIndexV1),
+}
+
+/// An index over the patches(blobs) in a quilt.
+///
+/// Each quilt patch represents a blob stored in the quilt. And each patch is
+/// mapped to a contiguous index range.
+///
+/// INV: The patches are sorted by their end indices, as well as their blob_ids.
```
mlegner: Nits:
```suggestion
/// An index over the [patches][QuiltPatchV1] (blobs) in a quilt.
///
/// Each quilt patch represents a blob stored in the quilt. And each patch is
/// mapped to a contiguous index range.
// INV: The patches are sorted by their end indices, as well as their blob_ids.
```

crates/walrus-core/src/metadata.rs
```diff
@@ -46,6 +50,137 @@ pub enum VerificationError {
...
+        identifier: &str,
+    ) -> Result<&QuiltPatchV1, QuiltError> {
+        self.quilt_patches
+            .iter()
+            .find(|patch| patch.identifier() == identifier)
+            .ok_or(QuiltError::BlobNotFoundInQuilt(identifier.to_string()))
+    }
+
+    /// Returns an iterator over the identifiers of the blobs in the quilt.
+    pub fn iter(&self) -> impl Iterator<Item = &str> {
```
mlegner: Nit: Can we call this method `identifiers`?

crates/walrus-core/src/metadata.rs
```diff
@@ -46,6 +50,137 @@ pub enum VerificationError {
...
+    /// Returns the quilt patch with the given blob identifier.
+    pub fn get_quilt_patch_by_identifier(
+        &self,
+        identifier: &str,
+    ) -> Result<&QuiltPatchV1, QuiltError> {
+        self.quilt_patches
+            .iter()
+            .find(|patch| patch.identifier() == identifier)
+            .ok_or(QuiltError::BlobNotFoundInQuilt(identifier.to_string()))
+    }
```
mlegner: Question: What access patterns do we expect here? This is fine as long as we access only a small number of patches by their identifiers, but is pretty inefficient if we go through many, in which case a map would be much more performant.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+#[allow(dead_code)] // TODO: remove this once follow up PRs are merged.
+pub trait QuiltDecoderApi<'a, V: QuiltVersion> {
+    /// Decodes the quilt index from received slivers.
+    ///
+    /// The decoded quilt index is stored in the decoder and can be retrieved
+    /// using the `get_quilt_index` method after this method returns.
+    fn decode_quilt_index(&mut self) -> Result<&V::QuiltIndex, QuiltError>;
+
+    /// Returns the decoded quilt index, if available.
+    fn get_quilt_index(&self) -> Option<&V::QuiltIndex>;
```
mlegner: Question/suggestion: What is the purpose of having separate functions here? Could we replace these by a combined function `get_or_decode_quilt_index`?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+const QUILT_INDEX_SIZE_PREFIX_SIZE: usize = 8;
+
+/// The number of bytes used to store the type of the quilt.
+const QUILT_TYPE_SIZE: usize = 1;
+
+/// The maximum number of columns a quilt index can have.
+const MAX_NUM_COLUMNS_FOR_QUILT_INDEX: usize = 1;
+
+/// Gets the quilt version enum from the data.
+#[allow(dead_code)] // TODO: remove this once follow up PRs are merged.
```
mlegner, line 47: Nit: You could also just add `#![allow(dead_code)]` at the top of the file, then you don't need individual attributes for each piece of code.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+    /// Returns the quilt index.
+    fn quilt_index(&self) -> &V::QuiltIndex;
+
+    /// Returns the data of the quilt.
+    fn data(&self) -> &[u8];
+
+    /// Returns the symbol size of the quilt.
+    fn symbol_size(&self) -> usize;
+}
+/// The configuration of the quilt.
```
mlegner, line 94: Nit: Can you add blank lines between blocks?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+    /// QuiltVersionV1.
+    V1,
+    /// Invalid version.
+    Invalid,
+}
+
+impl QuiltVersionEnum {
+    /// Creates a new `QuiltVersionEnum` from its serialized bytes.
+    #[allow(dead_code)] // TODO: remove this once follow up PRs are merged.
+    pub fn new_from_bytes(type_bytes: &[u8]) -> QuiltVersionEnum {
```
mlegner: Suggestion/nit: As the type field is one byte, we could set `type_bytes: u8` instead.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+}
+
+/// The version of the quilt.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum QuiltVersionEnum {
+    /// QuiltVersionV1.
+    V1,
+    /// Invalid version.
+    Invalid,
+}
```
mlegner: Question: Do we need the explicit `Invalid` type? If not, we could use a similar implementation as for the [`EncodingType`](https://github.com/MystenLabs/walrus/blob/9dc6842e0496f5637ba4f17bf9d496539f1f7f2e/crates/walrus-core/src/lib.rs#L678), in particular with `#[repr(u8)]`.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+    }
+}
+
+/// A wrapper around an owned blob and its identifier.
+#[derive(Debug, Clone)]
+#[allow(dead_code)] // TODO: remove this once follow up PRs are merged.
+pub struct BlobWithIdentifierOwned {
+    blob: Vec<u8>,
+    identifier: String,
+}
```
mlegner: Remark/question: I guess it will become clear in future PRs why we need separate types `BlobWithIdentifier` and `BlobWithIdentifierOwned`?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+            QuiltEnum::V1(quilt_v1) => Ok(QuiltIndex::V1(quilt_v1.quilt_index.clone())),
+        }
+    }
+}
+
+/// A wrapper around a blob and its identifier.
+#[derive(Debug, Clone)]
+pub struct BlobWithIdentifier<'a> {
+    blob: &'a [u8],
+    identifier: String,
```
mlegner, line 207: Question: Why can't we make the fields public?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+    type QuiltDecoder<'a>: QuiltDecoderApi<'a, Self>;
+    /// The type of the quilt.
+    type Quilt: QuiltApi<Self>;
+    /// The type of the quilt index.
+    type QuiltIndex: Clone;
+    /// The type of the quilt metadata.
+    type QuiltMetadata;
+
+    /// The serialized bytes of the quilt type.
+    fn quilt_type_bytes() -> &'static [u8];
```
mlegner: Suggestion: Given we use a single byte for the type, why not return a `u8` here?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+///   length of the [`QuiltIndexV1`]. It is guaranteed the column size is more than
+///   `QUILT_INDEX_SIZE_PREFIX_SIZE`.
+/// - The [`QuiltIndexV1`] is stored in the first one or multiple columns, up to
+///   `MAX_NUM_COLUMNS_FOR_QUILT_INDEX`.
+/// - The blob layout is defined by the [`QuiltIndexV1`].
+///
+/// INV:
+///  - `data.len()` is an integer multiple of `row_size`.
+///  - `row_size` is an integer multiple of `symbol_size`.
+///  - QuiltPatches are sorted by their identifiers.
```
mlegner: Nit: I would keep the invariants as comments as opposed to docstrings.

Also: It's a bit unclear how the last invariant is related to the invariant of `QuiltIndexV1`.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+            usize::from(encoding_config.n_source_symbols::<Primary>().get());
+        let n_secondary_source_symbols =
+            usize::from(encoding_config.n_source_symbols::<Secondary>().get());
+        let n_source_symbols: usize = n_primary_source_symbols * n_secondary_source_symbols;
+        assert!(
+            quilt_blob.len() % n_source_symbols == 0,
+            "quilt_blob length {} is not a multiple of n_source_symbols {}",
+            quilt_blob.len(),
+            n_source_symbols
+        );
```
mlegner: Important: I think it would be better to return an error in this case. Alternatively, describe this panic in the function's docstring.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+        let row_size = quilt_blob.len() / n_primary_source_symbols;
+        let symbol_size = quilt_blob.len() / n_source_symbols;
```
mlegner: Nit: Maybe this is optimized by the compiler, but it would otherwise be better to replace a division by a multiplication:
```suggestion
        let symbol_size = quilt_blob.len() / n_source_symbols;
        let row_size = symbol_size * n_secondary_source_symbols;
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+    }
+
+    fn symbol_size(&self) -> usize {
+        self.symbol_size
+    }
+}
+
+impl QuiltV1 {
+    /// Gets the blob represented by the given quilt patch.
+    fn get_blob(&self, quilt_patch: &QuiltPatchV1) -> Result<Vec<u8>, QuiltError> {
```
mlegner, line 363: Suggestion/question: This is again about access patterns: If we expect to get many blobs from a quilt, it would probably be more efficient to transpose the quilt once in the beginning instead of having to do the somewhat inefficient transpose during each access.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+            ));
+        }
+        list.finish()?;
+        writeln!(f)
+    }
+}
+
+/// Configuration for the quilt version 1.
+#[derive(Serialize, Deserialize, PartialEq, Eq, Debug)]
+pub struct QuiltConfigV1 {}
```
mlegner: Nit: Why not use a unit struct here?
```suggestion
pub struct QuiltConfigV1;
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+
+impl QuiltEncoderApi<QuiltVersionV1> for QuiltEncoderV1<'_> {
+    /// Constructs a [`QuiltV1`].
+    fn construct_quilt(&self) -> Result<QuiltV1, QuiltError> {
+        let _guard = self.span.enter();
+
+        let n_rows = self.config.n_source_symbols::<Primary>().get().into();
+        let n_columns = self.config.n_source_symbols::<Secondary>().get().into();
+        tracing::event!(
+            Level::DEBUG,
```
mlegner: Nit/question: Why not the shorthand version? (Also in other places.)
```suggestion
        tracing::debug!(
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+            "Constructing quilt with n_columns: {}, n_rows: {}",
+            n_columns,
+            n_rows
+        );
+
+        let mut blob_pairs = Vec::new();
+        for blob in self.blobs.iter() {
+            blob_pairs.push(blob);
+        }
```
mlegner: Nit:
```suggestion
        let mut blob_pairs = self.blobs.iter().collect::<Vec<_>>();
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+            blobs,
+            config,
+            span: tracing::span!(Level::ERROR, "QuiltEncoderV1"),
+        }
+    }
+}
+
+impl QuiltEncoderApi<QuiltVersionV1> for QuiltEncoderV1<'_> {
+    /// Constructs a [`QuiltV1`].
+    fn construct_quilt(&self) -> Result<QuiltV1, QuiltError> {
```
mlegner, line 530: Nit: Could this function be split into smaller ones?

mlegner: Remark (and note to self): I need to look at this function again in full detail.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+    fn encode_with_metadata(&self) -> Result<(Vec<SliverPair>, QuiltMetadata), QuiltError> {
+        let _guard = self.span.enter();
+        tracing::event!(Level::DEBUG, "starting to encode quilt with metadata");
+
+        let quilt = self.construct_quilt()?;
+        tracing::event!(
+            Level::DEBUG,
+            "construct quilt success {}",
+            quilt.data().len()
+        );
```
mlegner: Suggestion: Move the debug statements to the `construct_quilt` function.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+
+        let first_sliver = self
+            .slivers
+            .get(&index)
+            .ok_or(QuiltError::MissingSlivers(vec![index]))?;
+
+        assert_eq!(
+            QuiltVersionV1::quilt_type_bytes(),
+            utils::get_quilt_version_bytes(first_sliver.symbols.data())
+        );
```
mlegner: Suggestion: Can you document this potential panic or return an error here?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+#[derive(Debug)]
+pub struct QuiltDecoderV1<'a> {
+    slivers: HashMap<SliverIndex, &'a SliverData<Secondary>>,
+    quilt_index: Option<QuiltIndexV1>,
+}
+
+impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
+    fn decode_quilt_index(&mut self) -> Result<&QuiltIndexV1, QuiltError> {
+        let index = SliverIndex(0);
+
```
mlegner: Nit: Can we call this `first_sliver_index` and remove the blank line? (Or just remove the variable.)

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+        // Collect data from subsequent slivers if needed.
+        for i in 1..num_slivers_needed {
+            let next_index = SliverIndex(i as u16);
+            let next_sliver = self
+                .slivers
+                .get(&next_index)
+                .ok_or(QuiltError::Other(format!(
+                    "Unexpected state, sliver {} not found",
+                    next_index
+                )))?;
```
mlegner: Question: Didn't we just check above that the slivers exist? In this case it could be an `expect` here.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1442 @@
...
+            }
+        }
+        if !missing_slivers.is_empty() {
+            return Err(QuiltError::MissingSlivers(missing_slivers));
+        }
+        Ok(())
+    }
+}
+
+mod utils {
```
mlegner, line 844: Remark (note to self): Still need to review utils and tests in detail.

mlegner (commented):
Thank you so much for the amazing work, @liquid-helium! In addition to the format discussions offline, I have a few questions about the implementation. Most comments are pretty minor and/or can be addressed in a follow-up. Also, there are a few pieces of the code that I want to look at again tomorrow.

crates/walrus-core/src/encoding/errors.rs
liquid-helium: Removed.

crates/walrus-core/src/metadata.rs
liquid-helium: Good question, they were set as private to prevent accidental modification when constructing a quilt. I changed them to public since it don't provide too much protection and loses the consistency over all fields.

crates/walrus-core/src/metadata.rs
liquid-helium: Thanks for the suggestion, applied, removed the sort by id part for now and will add it back in a following PR when the ids are actually used.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice catch!
The only reason was the `&mut self` vs `&self `, I think in the previous implementation, some other functions of QuiltDecoder has to take &mut self for this reason, that's why I added both versions.

I just checked the latest implementation on the sdk client side, it seems we no-longer need it, removed.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Added.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: good idea, applied, thanks.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good idea, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Yes, `BlobWithIdentifier` will be used usually as the input, with the &[u8] provided by the caller, this aligns with most of current Apis.

`BlobWithIdentifierOwned` will be used as return value in a function `QuiltClient::read_blobs`, since I still want to return a mapping to let the caller know which blob maps to which identifier they provided, we could promise to return as the input order, but I find that a bit unsafe (personal opinion).

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: For this one, the intent is to avoid it being modified accidentally, it is set once in the constructor, and allows access only via the getter.

I will keep it as, but let me know if you have strong preference.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good idea, thanks, Done.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Changed to comments.
Updated the last one.

crates/walrus-core/src/encoding/quilt_encoding.rs

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Oh, nice trick, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice idea, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice catch, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Added `QuiltVersionMismatch` and `check_quilt_version`

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Renamed it.

crates/walrus-core/src/encoding/quilt_encoding.rs

crates/walrus-core/src/metadata.rs
liquid-helium: Agreed a map is better when it comes to many blobs per quilt read pattern.
Maybe we can revisit this once we finalize how we want to use these identifiers, I will start a offline thread to discussion this.

crates/walrus-core/src/metadata.rs
liquid-helium: change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Oh, I didn't know that, Thanks for the tip :)

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Agreed, I am thinking adding per quilt cache in the Client for optimization.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: @mlegner That's a good suggestion, I am currently doing it in a following PR, together with the addition of unencoded size at the beginning of each quilt patch, so if you don't mind, we can leave this review to next PR (which will be much smaller)

crates/walrus-core/src/metadata.rs
```diff
@@ -54,35 +54,37 @@ pub enum VerificationError {
...
+        if !identifier
+            .chars()
+            .all(|c| c.is_alphanumeric() || c == '_' || c == '-' || c == '.')
+        {
+            return Err(QuiltError::Other(
+                "Invalid identifier: must contain only alphanumeric, underscore, hyphen, or \
+                period characters"
+                    .to_string(),
+            ));
+        }
```
mlegner, line 91: Nit: Can we put this into a separate function to make it easier to find?

crates/walrus-core/src/metadata.rs
mlegner: Can you create a Linear issue to track this?

crates/walrus-core/src/metadata.rs
mlegner: I can't see this change. Did this get committed/pushed? I think a few other ones were also lost... 🤔

mlegner (commented):
Thanks for the changes, @liquid-helium. It seems to me that some of the changes got lost somewhere though... Could you check and re-request a review?

crates/walrus-core/src/metadata.rs
liquid-helium: That is weird... I can see the changes, maybe Gibhub has some issues pulling the newest data? Let me know if you are still seeing the old code, we can jump into a call if necessary.


```
/// An index over the [patches][QuiltPatchV1] (blobs) in a quilt.
///
/// Each quilt patch represents a blob stored in the quilt. And each patch is
/// mapped to a contiguous index range.
// INV: The patches are sorted by their end indices.
```

crates/walrus-core/src/metadata.rs
liquid-helium: Added WAL-829 to track it as a sub-issue.

crates/walrus-core/src/metadata.rs
mlegner: OK, I see it now along with other changes. Very weird, may have been some caching issue in the browser.

crates/walrus-core/src/metadata.rs
```diff
@@ -107,19 +108,22 @@ pub struct QuiltIndexV1 {
 
 impl QuiltIndexV1 {
     /// Returns the quilt patch with the given blob identifier.
+    /// TODO(WAL-829): Consider storing the quilt patch in a hashmap for O(1) lookup.
```
mlegner: Nit:
```suggestion
    // TODO(WAL-829): Consider storing the quilt patch in a hashmap for O(1) lookup.
```

crates/walrus-core/src/encoding/errors.rs
```diff
@@ -152,15 +152,15 @@ pub enum SymbolVerificationError {
 /// Errors that may be encountered while interacting with quilt.
 #[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
 pub enum QuiltError {
```
mlegner, line 154: Suggestion/nit: Given that 4 uses of the `Other` error are for empty inputs, it might make sense to define a separate variant for that.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -698,10 +713,7 @@ impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
             let next_sliver = self
                 .slivers
                 .get(&next_index)
-                .ok_or(QuiltError::Other(format!(
-                    "Unexpected state, sliver {} not found",
-                    next_index
-                )))?;
+                .expect("sliver should be present");
```
mlegner: Nit: Can we be a bit more explicit? For example:
```suggestion
                .expect("we know this exists because we ran check_missing_slivers above");
```

crates/walrus-core/src/encoding/quilt_encoding.rs
mlegner: OK, then I'll look at this again in the next PR. 👍

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+                "failed to compute symbol size: blobs are empty".to_string(),
+            ));
+        }
+
+        let mut min_val = cmp::max(
+            blobs_sizes
+                .iter()
+                .sum::<usize>()
+                .div_ceil(n_columns)
+                .div_ceil(n_rows),
```
mlegner: Nit: Probably more efficient:
```suggestion
                .div_ceil(n_columns * n_rows),
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+        while min_val < max_val {
+            let mid = (min_val + max_val) / 2;
+            if can_blobs_fit_into_matrix(blobs_sizes, n_columns, mid * n_rows) {
+                max_val = mid;
+            } else {
+                min_val = mid + 1;
+            }
+        }
+
+        let symbol_size = min_val.div_ceil(required_alignment) * required_alignment;
```
mlegner: Nit: I think we can use `next_multiple_of` here:
```suggestion
        let symbol_size = min_val.next_multiple_of(required_alignment);
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+}
+
+mod utils {
+    use super::*;
+
+    /// Finds the minimum symbol size needed to store blobs in a fixed number of columns.
+    /// Each blob must be stored in consecutive columns exclusively.
+    ///
+    /// # Arguments
+    /// * `blobs_sizes` - Slice of blob lengths.
```
mlegner: Question: This includes the index, correct? Can you clarify this in the docstring? Also, can you mention that it's length should be between 1 and `n_columns`?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+    pub fn compute_symbol_size(
+        blobs_sizes: &[usize],
+        n_columns: usize,
+        n_rows: usize,
+        max_num_columns_for_quilt_index: usize,
+        required_alignment: usize,
+    ) -> Result<usize, QuiltError> {
+        if blobs_sizes.len() > n_columns {
+            // The first column is not user data.
+            return Err(QuiltError::TooManyBlobs(blobs_sizes.len(), n_columns - 1));
```
mlegner: Remark: This is a bit confusing because the first `blob` isn't actually a blob but the index...

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+            blobs_sizes
+                .first()
+                .expect("blobs_sizes is not empty")
+                .div_ceil(n_rows * max_num_columns_for_quilt_index),
+        );
+        min_val = cmp::max(
+            min_val,
+            (QUILT_INDEX_SIZE_PREFIX_SIZE + QUILT_TYPE_SIZE).div_ceil(n_rows),
+        );
+        let mut max_val = blobs_sizes
```
mlegner, line 896: Suggestion: AFAICT, this computes the symbol size such that the largest blob fits into a single column. We can get a still correct but tighter bound by assuming the largest blob fitting into `n_columns / n_blobs` columns.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+    /// * `blobs_sizes` - Slice of blob lengths.
+    /// * `n_columns` - Number of columns available.
+    /// * `n_rows` - Number of rows available.
+    /// * `max_num_columns_for_quilt_index` - The maximum number of columns that can be used to
+    ///   store the quilt index.
+    /// * `required_alignment` - The alignment of the symbol size.
+    ///
+    /// # Returns
+    /// * `Result<usize, QuiltError>` - The minimum symbol size needed, or an error if impossible.
+    pub fn compute_symbol_size(
```
mlegner, line 866: Suggestion: Can you add a few comments explaining the algorithm? I think it is correct, but it took me a bit to understand it.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+        );
+        min_val = cmp::max(
+            min_val,
+            (QUILT_INDEX_SIZE_PREFIX_SIZE + QUILT_TYPE_SIZE).div_ceil(n_rows),
+        );
+        let mut max_val = blobs_sizes
+            .iter()
+            .max()
+            .copied()
+            .unwrap_or(0)
```
mlegner: Nit: This could actually be an `expect`, because we made sure the `blobs_sizes` is non-empty.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+            let mid = (min_val + max_val) / 2;
+            if can_blobs_fit_into_matrix(blobs_sizes, n_columns, mid * n_rows) {
+                max_val = mid;
+            } else {
+                min_val = mid + 1;
+            }
+        }
+
+        let symbol_size = min_val.div_ceil(required_alignment) * required_alignment;
+        if symbol_size > u16::MAX as usize {
```
mlegner: Nit: Can we use the [`MAX_SYMBOL_SIZE` constant](https://github.com/MystenLabs/walrus/blob/aa95d189627da61d9e4c44b75cb92b59db7ab2c5/crates/walrus-core/src/encoding/common.rs#L9) here?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+        let mut used_cols = 0;
+        for &blob in blobs_sizes {
+            let cur = blob.div_ceil(column_size);
+            if used_cols + cur > n_columns {
+                return false;
+            }
+            used_cols += cur;
+        }
+
+        true
```
mlegner: Nit: Slightly simpler:
```suggestion
        let required_columns = blobs_sizes
            .iter()
            .map(|blob_size| blob_size.div_ceil(column_size))
            .sum::<usize>();
        n_columns >= required_columns
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+        if combined_data.len() < QUILT_INDEX_SIZE_PREFIX_SIZE + QUILT_TYPE_SIZE {
+            return Err(QuiltError::FailedToExtractQuiltIndexSize);
+        }
+
+        let data_size = u64::from_le_bytes(
+            combined_data[QUILT_TYPE_SIZE..QUILT_INDEX_SIZE_PREFIX_SIZE + QUILT_TYPE_SIZE]
+                .try_into()
+                .map_err(|_| QuiltError::FailedToExtractQuiltIndexSize)?,
+        );
+        let data_size = usize::try_from(data_size).expect("data_size should fit in usize");
```
mlegner, line 955: Question: Is there any case where `data_size > u32::MAX`? If yes, this could panic on 32-bit architectures. If not, why not use a `u32` here?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+        Ok(data_size)
+    }
+
+    pub fn get_quilt_version_byte(data: &[u8]) -> Result<u8, QuiltError> {
+        if data.is_empty() {
+            return Err(QuiltError::Other(
+                "failed to get quilt version byte: data is empty".to_string(),
+            ));
+        }
+        Ok(data[0])
```
mlegner: Nit: How about this version?
```suggestion
        data.get(0).copied().ok_or(QuiltError::Other(
            "failed to get quilt version byte: data is empty".to_string(),
        ))
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+        let mut column = Vec::with_capacity(n_rows * symbol_size);
+
+        for row in 0..n_rows {
+            let start_idx = row * row_size + i * symbol_size;
+            let end_idx = start_idx + symbol_size;
+
+            // Check if we have enough data for this chunk.
+            if end_idx > data.len() {
+                break;
+            }
```
mlegner: Question: Can this actually happen? AFAICT, the maximum value of `end_idx` is `(n_rows-1) * row_size + (row_size-symbol_size) + symbol_size = n_rows*row_size = data.len()`.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -0,0 +1,1468 @@
...
+            .get(&first_sliver_index)
+            .ok_or(QuiltError::MissingSlivers(vec![first_sliver_index]))?;
+
+        utils::check_quilt_version::<QuiltVersionV1>(first_sliver.symbols.data())?;
+
+        let data_size = utils::get_quilt_index_data_size(first_sliver.symbols.data())?;
+
+        // Calculate how many slivers we need based on the data size.
+        let num_slivers_needed = data_size.div_ceil(first_sliver.symbols.data().len());
+        let prefix_size = QUILT_INDEX_SIZE_PREFIX_SIZE + QUILT_TYPE_SIZE;
```
mlegner: Nit: This is used in 5 places; how about defining a separate constant `PREFIX_SIZE`?

crates/walrus-core/src/encoding/quilt_encoding.rs
mlegner: I've reviewed the utils now in detail, but only did a quick pass over the tests.

mlegner (approved):
Thanks so much for your work and patience, @liquid-helium! This LGTM! 💯 

I'm approving the PR as I only have a few remaining questions and nits for your consideration, but nothing major.

crates/walrus-core/src/metadata.rs
liquid-helium: Fixed.

crates/walrus-core/src/encoding/errors.rs
liquid-helium: Nice observation, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Thanks for the tip! Change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice catch, I added a -1 here

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good idea! Improvement applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: algorithm detail added.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice! I like it!

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good idea, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: You are right, it shouldn't. Removed!

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Switched to u32.

crates/walrus-core/src/metadata.rs
liquid-helium: Done.

PR #1755, opened 2025-03-19
test: make simtest deterministic

halfprice:
## Description

By using tokio blocking threads in RecoverySymbolService instead of Rayon thread.

This PR creates a generic BoundedThreadPool implementation that can be either a tokio thread pool or Rayon thread pool.

Verified simtest determinism with test_event_blob_fork_recovery and #[walrus_simtest(check_determinism)]. 100 tests
passed. Before this change, the test fails within 5 attempts.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/thread_pool.rs
+++ b/crates/walrus-service/src/node/thread_pool.rs
@@ -40,8 +40,8 @@ pub(crate) const ASSUMED_REQUEST_LATENCY: Duration = Duration::from_millis(10);
 /// [`RayonThreadPool::bounded`].
 pub(crate) const DEFAULT_MAX_WAIT_TIME: Duration = Duration::from_secs(1);
 
-/// The default number of concurrent jobs for the tokio blocking pool.
-pub(crate) const DEFAULT_TOKIO_BLOCKING_POOL_CONCURRENT_JOBS: usize = 100;
+/// The default number of workers for the tokio blocking pool.
+pub(crate) const DEFAULT_TOKIO_BLOCKING_POOL_NUM_WORKERS: usize = 4;
 
 /// A [`BlockingThreadPool`] with a limit to the number of concurrent jobs.
 pub(crate) type BoundedThreadPool = ConcurrencyLimit<BlockingThreadPool>;
@@ -205,7 +205,10 @@ impl BlockingThreadPool {
         let n_workers = match self {
             Self::Rayon(ref pool) => pool.inner.current_num_threads() as f64,
             Self::Tokio(ref _pool) => std::thread::available_parallelism()
-                .unwrap_or(NonZero::new(DEFAULT_TOKIO_BLOCKING_POOL_CONCURRENT_JOBS).unwrap())
+                .unwrap_or(
+                    NonZero::new(DEFAULT_TOKIO_BLOCKING_POOL_NUM_WORKERS)
+                        .expect("default tokio blocking pool concurrent jobs must be non-zero"),
+                )
                 .get() as f64,
         };
 
@@ -238,8 +241,8 @@ where
 
     fn call(&mut self, req: Req) -> Self::Future {
         match self {
-            Self::Rayon(pool) => pool.call(Box::new(req)),
-            Self::Tokio(pool) => pool.call(Box::new(req)),
+            Self::Rayon(pool) => pool.call(req),
+            Self::Tokio(pool) => pool.call(req),
         }
     }
 }


crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -126,6 +155,164 @@ where
...
+pub(crate) struct BlockingThreadPool {
+    inner: ThreadPoolImpl,
+}
+
+/// The inner implementation of the thread pool.
+#[derive(Debug, Clone)]
+pub(crate) enum ThreadPoolImpl {
+    Rayon(RayonThreadPool),
+    Tokio(TokioBlockingPool),
+}
```
halfprice: This is probably the 5th version of this PR, that I feel I finally create one that do not pollute the the main code with this testing only addition.

I tried to create a generic trait that can be implemented by both Rayon and Tokio, which all ended up with a lot of generic parameters added everywhere.

Willing to change if there is better way to do this.

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -258,13 +271,33 @@ mod tests {
 
     #[tokio::test]
     #[should_panic]
```
halfprice, line 275: #[should_panic] does not work with async_param_test. But I didn't dig deeper why...

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -38,18 +39,32 @@ pub(crate) const ASSUMED_REQUEST_LATENCY: Duration = Duration::from_millis(10);
 /// [`RayonThreadPool::bounded`].
 pub(crate) const DEFAULT_MAX_WAIT_TIME: Duration = Duration::from_secs(1);
 
-/// A [`RayonThreadPool`] with a limit to the number of concurrent jobs.
-pub(crate) type BoundedRayonThreadPool = ConcurrencyLimit<RayonThreadPool>;
+/// The default number of concurrent jobs for the tokio blocking pool.
+pub(crate) const DEFAULT_TOKIO_BLOCKING_POOL_CONCURRENT_JOBS: usize = 100;
```
jpcsmith: Question: Is there a reason not to use the same number of in-flight jobs for both rayon and tokio?

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -38,18 +39,32 @@ pub(crate) const ASSUMED_REQUEST_LATENCY: Duration = Duration::from_millis(10);
...
-            .build()
-            .expect("thread pool construction must succeed")
-            .into(),
-    )
-    .bounded()
+/// A [`BlockingThreadPool`] with a limit to the number of concurrent jobs.
+pub(crate) type BoundedThreadPool = ConcurrencyLimit<BlockingThreadPool>;
+
+/// Returns a default constructed `BoundedThreadPool`.
+pub(crate) fn default_bounded() -> ConcurrencyLimit<BlockingThreadPool> {
```
jpcsmith:
```suggestion
pub(crate) fn default_bounded() -> BoundedThreadPool {
```

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -38,18 +39,32 @@ pub(crate) const ASSUMED_REQUEST_LATENCY: Duration = Duration::from_millis(10);
...
+        .bounded()
+    }
+
+    // Note that in simtest, we cannot use Rayon thread pool due to that it causes simtest to be
+    // non-deterministic. This is because Rayon threads are actually run in different threads than
+    // the tokio main threads, and not controlled by the tokio runtime.
+    #[cfg(msim)]
+    {
+        BlockingThreadPool::new_tokio(TokioBlockingPool::new()).bounded()
+    }
```
jpcsmith, line 68: Nit: use `if cfg!(msim) {`. I'm generally against having conditional compilations scattered throughout the  code base as it hinders editing and linting the code.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -77,24 +104,26 @@ pub(crate) struct RayonThreadPool {
 
 impl RayonThreadPool {
     /// Returns a new `RayonThreadPool` service.
+    #[cfg(any(test, not(msim)))]
```
jpcsmith: Suggestion: Use `#[allow(unused)]` instead of conditional compilation.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -77,24 +104,26 @@ pub(crate) struct RayonThreadPool {
...
         let n_workers = self.inner.current_num_threads() as f64;
         let time_in_queue = DEFAULT_MAX_WAIT_TIME.as_secs_f64();
         let task_latency = ASSUMED_REQUEST_LATENCY.as_secs_f64();
         let limit = (n_workers * time_in_queue / task_latency) as usize;
 
         tracing::debug!(%limit, "calculated limit for `BoundedRayonThreadPool`");
 
-        ConcurrencyLimit::new(self, limit)
+        ConcurrencyLimit::new(BlockingThreadPool::new_rayon(self), limit)
     }
```
jpcsmith: Suggestion: Move the `bounded` method to the `BlockingThreadPool` enum.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -126,6 +155,164 @@ where
...
+/// returns the result.
+///
+/// The error type of the service is returned if the function panics and
+/// is the result of [`std::panic::catch_unwind`].
+#[derive(Debug, Clone, Default)]
+pub(crate) struct TokioBlockingPool;
+
+impl TokioBlockingPool {
+    /// Returns a new `TokioBlockingPool` service.
+    #[cfg(any(test, msim))]
```
jpcsmith: Suggestion: Use `#[allow(unused)]` instead of conditional compilation.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -126,6 +155,164 @@ where
...
+    }
+
+    /// Converts this pool into a [`BoundedThreadPool`], whose inner pool is a TokioBlockingPool
+    /// from `self`.
+    pub fn bounded(self) -> ConcurrencyLimit<BlockingThreadPool> {
+        ConcurrencyLimit::new(
+            BlockingThreadPool::new_tokio(self),
+            DEFAULT_TOKIO_BLOCKING_POOL_CONCURRENT_JOBS,
+        )
+    }
```
jpcsmith: Suggestion: Move to `BlockingThreadPool` and in place of rayon's `current_num_threads`  in the calculation, use [`std::thread::available_parallelism`](https://doc.rust-lang.org/std/thread/fn.available_parallelism.html).

crates/walrus-service/src/node/thread_pool.rs
jpcsmith: Suggestion: Make `BlockingThreadPool` an enum instead of having the `ThreadPoolImpl` enum, as I am not seeing the reason to hide that detail from the crate.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -63,6 +78,18 @@ pub(crate) fn unwrap_or_resume_panic<T>(result: std::thread::Result<T>) -> T {
...
+trait ThreadPoolService: Send {
+    /// Calls a function on the thread pool and returns a future that resolves to the result.
+    fn service_call<T: Send + 'static>(
+        &mut self,
+        req: Box<dyn FnOnce() -> T + Send>,
+    ) -> ThreadPoolFuture<T>;
+
+    /// Polls the readiness of the thread pool.
+    fn service_poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), ThreadPanicError>>;
+}
```
jpcsmith: Suggestion: Remove this trait as it seems unnecessary. The code in service_call and service_poll_ready for the respective enum variants can be placed directly into the Service implementation.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -126,6 +155,164 @@ where
...
+        &mut self,
+        req: Box<dyn FnOnce() -> T + Send>,
+    ) -> ThreadPoolFuture<T> {
+        self.call(req)
+    }
+
+    fn service_poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), ThreadPanicError>> {
+        <RayonThreadPool as Service<fn()>>::poll_ready(self, cx)
+    }
+}
```
jpcsmith: Suggestion: Remove (see comment above).

crates/walrus-service/src/node/recovery_symbol_service.rs
jpcsmith: Having a look...

crates/walrus-service/src/node/thread_pool.rs
halfprice: Make sense and much cleaner now!

All of these indirection were all from the evolution of them previous attempts...

crates/walrus-service/src/node/thread_pool.rs
halfprice: Done

crates/walrus-service/src/node/thread_pool.rs
halfprice: I thought about this, and I currently strongly prefer to do the conditional compilation in this case is to prevent rayon thread being used in simtest at all. It was a very painful process to figure out where the non-determinisim is, so I really want to avoid this at all cost... If using `if cfg!(msim)`, now Rayon threadpool now visible in simtest and it is easy for someone to use it in simtest.

crates/walrus-service/src/node/thread_pool.rs
halfprice: done

crates/walrus-service/src/node/thread_pool.rs
halfprice: done. much cleaner now!

crates/walrus-service/src/node/thread_pool.rs
halfprice: done. still preserved DEFAULT_TOKIO_BLOCKING_POOL_CONCURRENT_JOBS in casse available_parallelism returned error.

crates/walrus-service/src/node/thread_pool.rs
halfprice: make sense. done

crates/walrus-service/src/node/thread_pool.rs
halfprice: done.

crates/walrus-service/src/node/thread_pool.rs
halfprice: changed. done.

crates/walrus-service/src/node/recovery_symbol_service.rs
jpcsmith: It's something weird with `#[tokio::test]` placing the `#[test]` attribute in a weird place... nothing to do but to work around it for now.

crates/walrus-service/src/node/thread_pool.rs
jpcsmith: That's fair. Okay.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -126,6 +127,123 @@ where
...
+    // Converts this pool into a [`BoundedThreadPool`], whose inner pool is from `self`.
+    //
+    // For `N` workers in the thread pool, the limit is set to
+    // `N * DEFAULT_MAX_WAIT_TIME / ASSUMED_REQUEST_LATENCY` which aims to limit the amount
+    // of time spent in the queue to [`DEFAULT_MAX_WAIT_TIME`].
+    pub fn bounded(self) -> ConcurrencyLimit<BlockingThreadPool> {
+        let n_workers = match self {
+            Self::Rayon(ref pool) => pool.inner.current_num_threads() as f64,
+            Self::Tokio(ref _pool) => std::thread::available_parallelism()
+                .unwrap_or(NonZero::new(DEFAULT_TOKIO_BLOCKING_POOL_CONCURRENT_JOBS).unwrap())
```
jpcsmith: Suggestion: Use expect, not unwrap just to make it clear why this should never panic.

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -38,18 +40,32 @@ pub(crate) const ASSUMED_REQUEST_LATENCY: Duration = Duration::from_millis(10);
...
-pub(crate) fn default_bounded() -> BoundedRayonThreadPool {
-    RayonThreadPool::new(
-        ThreadPoolBuilder::new()
-            .build()
-            .expect("thread pool construction must succeed")
-            .into(),
-    )
-    .bounded()
+/// The default number of concurrent jobs for the tokio blocking pool.
+pub(crate) const DEFAULT_TOKIO_BLOCKING_POOL_CONCURRENT_JOBS: usize = 100;
```
jpcsmith: Important: Since this is now used as the number of workers, from which the number of concurrent jobs is calculated, rename it and reduce to a conservative number (such as 4 or 6; the number of computed concurrent jobs will be much higher than 100, even with 4).

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -126,6 +127,123 @@ where
...
+        match self {
+            Self::Rayon(pool) => <RayonThreadPool as Service<fn()>>::poll_ready(pool, cx),
+            Self::Tokio(pool) => <TokioBlockingPool as Service<fn()>>::poll_ready(pool, cx),
+        }
+    }
+
+    fn call(&mut self, req: Req) -> Self::Future {
+        match self {
+            Self::Rayon(pool) => pool.call(Box::new(req)),
+            Self::Tokio(pool) => pool.call(Box::new(req)),
```
jpcsmith:
```suggestion
            Self::Rayon(pool) => pool.call(req),
            Self::Tokio(pool) => pool.call(req),
```

crates/walrus-service/src/node/thread_pool.rs
jpcsmith: Yea, I suspected.

crates/walrus-service/src/node/thread_pool.rs
halfprice: Good catch!

PR #1763, opened 2025-03-20
feat: expose event sequence consistency

halfprice:
## Description

This PR adds a metric to expose event sequence consistency, and can be used to check if a node missed any events
comparing to other nodes. We periodically record the event source (checkpoint + count) in a metrics, and record
the last 10 recordings using 10 buckets in a round robin manner. To make the metrics comparable across nodes,
we only record metrics if the node is relatively up-up-date meaning that it is in Active or RecoveryInProgress state.

Node that Event Blob creation already detects event inconsistency across nodes. However, stuck event blob
creation may be caused by other reason for example different number of events included. Therefore, this metrics
provide more information on event processing consistency.

If useful, we can also expose the recordings in the health command with a bit more information. But I want to see how
and if this metric is useful in prod.

Contributes to WAL-701

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -191,6 +191,7 @@ const NUM_CHECKPOINTS_PER_BLOB_ON_TESTNET: u32 = 18_000;
 // contain Walrus events. 20K events per recording is roughly 1 recording per 1.5 hours.
 const NUM_EVENTS_PER_DIGEST_RECORDING: u64 = 20_000;
 const NUM_DIGEST_BUCKETS: u64 = 10;
+const CHECKPOINT_EVENT_POSITION_SCALE: u64 = 100;
 
 /// Trait for all functionality offered by a storage node.
 pub trait ServiceState {
@@ -946,10 +947,12 @@ impl StorageNode {
 
                 // Ignore the error here since this is a best effort operation, and we don't want
                 // any error from it to stop the node.
-                let _ = self.maybe_record_event_source(
+                if let Err(error) = self.maybe_record_event_source(
                     element_index,
                     &stream_element.checkpoint_event_position,
-                );
+                ) {
+                    tracing::warn!(?error, "Failed to record event source");
+                }
 
                 let event_handle = EventHandle::new(
                     element_index,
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1615,6 +1630,57 @@ impl StorageNode {
...
+        // `NUM_EVENTS_PER_DIGEST_RECORDING` is chosen in a way that a node produces a recording
+        // every few hours.
+        if event_index % NUM_EVENTS_PER_DIGEST_RECORDING != 0 {
+            return Ok(());
+        }
+
+        // Only record digests for active or recovering nodes
+        match self.inner.storage.node_status()? {
+            NodeStatus::Active | NodeStatus::RecoveryInProgress(_) => {
+                let bucket = event_index / NUM_EVENTS_PER_DIGEST_RECORDING % NUM_DIGEST_BUCKETS;
```
jpcsmith: Nit: Could you add parentheses here to make it clear the order of operations. My operator precedence knowledge for `%` is somewhat rusty 🦀 .

crates/walrus-service/src/node.rs
```diff
@@ -1615,6 +1630,57 @@ impl StorageNode {
...
+                debug_assert!(bucket < NUM_DIGEST_BUCKETS);
+
+                // The event source is the combination of checkpoint sequence number and counter.
+                let event_source = event_source
+                    .checkpoint_sequence_number
+                    .checked_mul(100)
```
jpcsmith: Question: What is the source of the 100? Is it the max number of events before a new checkpoint or something similar? If the value is defined somewhere else in the crate/event-processor, could we use that variable, or define a global variable here?

crates/walrus-service/src/node.rs
```diff
@@ -1615,6 +1630,57 @@ impl StorageNode {
...
+    ) -> Result<(), TypedStoreError> {
+        // Only record every Nth event.
+        // `NUM_EVENTS_PER_DIGEST_RECORDING` is chosen in a way that a node produces a recording
+        // every few hours.
+        if event_index % NUM_EVENTS_PER_DIGEST_RECORDING != 0 {
+            return Ok(());
+        }
+
+        // Only record digests for active or recovering nodes
+        match self.inner.storage.node_status()? {
```
jpcsmith: Nit: Use 
```rust
if !matches!(self.inner.storage.node_status()?, NodeStatus::Active | NodeStatus::RecoveryInProgress(_)) { 
  return Ok(());
}

// ... Rest of function
```
which avoids having the bulk of the function body creeping rightward.

crates/walrus-service/src/node.rs
```diff
@@ -936,6 +944,13 @@ impl StorageNode {
                     }
                 }
 
+                // Ignore the error here since this is a best effort operation, and we don't want
+                // any error from it to stop the node.
```
jpcsmith, line 949: Suggestion: Log the error so that it is not wholly ignored.

crates/walrus-service/src/node.rs
halfprice: done

crates/walrus-service/src/node.rs
halfprice: It's basically a random number I picked. I was envisioning that there won't be more than 100 walrus events per checkpoints, which can be totally not true if someday Walrus becomes very popular. If we have more than 100 events, that is still ok because what we care about is that all the nodes produce the exact same value per bucket, but not the value itself.

crates/walrus-service/src/node.rs
halfprice: I didn't want to log it because I feel if something is wrong, 1 log per event may be too much. However, currently only reading the node status may return error, so if that's the case, some other part of the node will fail anyway, so I added a log here. We can re-evaluate if it produce too many logs.

PR #1777, opened 2025-03-21
feat: Added move call metrics

liquid-helium:
## Description

Added a new metric set: SuiClientMetrics for metrics related to sui rpc calls

In this PR, the metrics are collected via the RetriableSuiClient.

Event related rpc call metrics are not collected yet.
Related to WAL-711.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/node/contract_service.rs
```diff
@@ -195,7 +196,16 @@ impl SuiSystemContractServiceBuilder {
         config: &SuiConfig,
         committee_service: Arc<dyn CommitteeService>,
     ) -> Result<SuiSystemContractService, anyhow::Error> {
-        Ok(self.build(config.new_contract_client().await?, committee_service))
+        Ok(self.build(
+            config
+                .new_contract_client_with_metrics(Arc::new(SuiClientMetricSet::new(
+                    self.metrics_registry
+                        .as_ref()
+                        .unwrap_or_else(|| prometheus::default_registry()),
```
halfprice: Important: if the metrics_registry is not set, but the node contains another registry, by calling `default_registry` will crash the node. Each binary can only run 1 registry at a time. So it's better to use the `new_contract_client()` version to create the client here.

crates/walrus-sui/src/client/metrics.rs
```diff
@@ -0,0 +1,36 @@
...
+fn default_buckets_for_slow_operations() -> Vec<f64> {
+    prometheus::exponential_buckets(0.001, 2.0, 14).expect("count, start, and factor are valid")
+}
+
+walrus_utils::metrics::define_metric_set! {
+    #[namespace = "walrus"]
+    /// Metrics for the Sui client operations.
+    pub struct SuiClientMetricSet {
+        #[help = "Total number of Sui RPC calls made"]
+        sui_rpc_calls_count: IntCounterVec["method", "status"],
```
halfprice, line 18: nit: I think sui_rpc_calls may be a bit confusing since rpc call indicates a specific calls to read data from sui. Maybe sui_calls_count will be sufficient, with the helper string indicating that this includes both the rpc call and the contract call.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -162,16 +177,33 @@ impl RetriableRpcError for RetriableClientError {
...
+    method: &str,
+) -> Result<T, E>
 where
     S: BackoffStrategy,
     F: FnMut() -> Fut,
     Fut: Future<Output = Result<T, E>>,
-    E: RetriableRpcError,
+    E: RetriableRpcError + ToErrorType,
 {
     loop {
```
halfprice, line 234: Suggestion: can we also record the number of retries?

crates/walrus-sui/src/client/metrics.rs
liquid-helium: Nice suggestion, change applied.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Good idea, added.

crates/walrus-service/src/node/contract_service.rs
liquid-helium: Thanks for the tip, fixed!

crates/walrus-sui/src/client/metrics.rs
halfprice: Seems the code is the same as before?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -166,17 +175,81 @@ impl RetriableRpcError for RetriableClientError {
...
+    count: u64,
+    metrics: Arc<SuiClientMetricSet>,
+    status: String,
+}
+
+impl RetryCountGuard {
+    fn new(metrics: Arc<SuiClientMetricSet>, method: &str) -> Self {
+        Self {
+            method: method.to_string(),
+            count: 1,
```
halfprice: Question: if initial count is 1, and every call increment by 1, does this mean that the retry count will be recorded for every call? I feel the initial count should be 0.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -166,17 +175,81 @@ impl RetriableRpcError for RetriableClientError {
...
+            metrics,
+            status: "success".to_string(),
+        }
+    }
+
+    pub fn record_result<T, E>(&mut self, value: Result<&T, &E>)
+    where
+        E: RetriableRpcError + ToErrorType,
```
halfprice: Suggestion: I think RetriableRpcError is not necessary.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1133,3 +1332,67 @@ fn maybe_return_injected_error_in_stake_pool_transaction(
...
+            Self::InvalidSignature => "invalid_sig".to_string(),
+        }
+    }
+}
+
+impl ToErrorType for SuiClientError {
+    fn to_error_type(&self) -> String {
+        match self {
+            Self::SuiSdkError(error) => error.to_error_type(),
+            _ => "other".to_string(),
```
halfprice: nit: sui_client_error_other. And similar for other "other" to add a prefix to identify the error source.

liquid-helium: Good point, thanks.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: nice catch!

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Thanks for the catch, fixed.

PR #1778, opened 2025-03-21
test: turn on event blob recovery simtest, also fix a race condition in event blob certification and reset

halfprice:
## Description

This PR enables the event blob recovery simtest, cleans up the test a bit, and fixed a race condition discovered by the test.

The race condition is found by the simtest where the following situation could cause the event blob progress to stuck
forever in the system, assuming 4 nodes cluster:

- node 1 certify event blob A
- node 2 certify event blob A
- node1 node2 crash
- node 3 certify event blob 1, blob 1 is now certified on chain
- node 1 and node 2 restart, their local last certified is different from the last certified onchain, which makes them stuck forever.
- Since node 1 and node 2 are all stuck, which cause the quorum cannot be formed, the event blob progress stuck.

The fix is to check that the local last certified is either the same as the global certified, or 1 blob before the global certified.

After the change, 100 tests passed.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/events/event_blob_writer.rs
+++ b/crates/walrus-service/src/node/events/event_blob_writer.rs
@@ -533,20 +533,34 @@ impl EventBlobWriterFactory {
         let uncertified_blobs_after_last_certified =
             pending.len() + failed_to_attest.len() + attested.len();
 
-        let should_reset = total_uncertified_blobs == uncertified_blobs_after_last_certified
-            && uncertified_blobs_after_last_certified as u32 >= num_uncertified_blob_threshold;
+        let consecutive_uncertified = uncertified_blobs_after_last_certified as u32;
+
+        // We reset the node pending blobs if the local node has already certified the last
+        // certified blob on chain, and have pending blob greater than
+        // num_uncertified_blob_threshold to recover the possible local node event blob corruption.
+        // Node that we need to check either the local last certified blob is the same as the chain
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -500,20 +500,34 @@ impl EventBlobWriterFactory {
...
+        // We reset the node pending blobs if the local node has already certified the last
+        // certified blob on chain, and have pending blob greater than
+        // num_uncertified_blob_threshold to recover the possible local node event blob corruption.
+        // Node that we need to check either the local last certified blob is the same as the chain
+        // last certified blob or the local last certified blob is one checkpoint behind the chain
+        // last certified blob, to account for the situation where node is crashed after sending
+        // out the last certified blob and before receiving the certification event.
+        let should_reset = consecutive_uncertified >= num_uncertified_blob_threshold
+            && (total_uncertified_blobs == consecutive_uncertified as usize
+                || total_uncertified_blobs == consecutive_uncertified as usize + 1);
```
halfprice, line 547: Hi @sadhansood, here is the main fix, namely not only we need to check the last certified locally is the same as the global certified, but also just 1 event blob behind the global certified.

PR #1782, opened 2025-03-21
feat: mdbook-templates for dynamic operators list preproccessing

vlegakis:
## Description

This PR introduced a templating pattern in the docs subproject with mdbook-templates preprocessor which inject data from the json with the operators list in docs/book/assets/operators.json . The intention of this PR is to maintain a common source of truth for Walrus operators maintainable by the community in operators.json which can also be used as the source of truth for other projects (e.g walrus nodes monitoring). 

In addition to that, anyone adding or updating-removing an operator can add metadata for each eg like :
`"https://walrus-testnet-aggregator.whatever.com": { "cache": true, other-metadata-here },...`

In sort the list of operators inside the document is replaced by: {{ #testnet.aggregators }} and {{ #testnet.publishers }} respectively pointing to the relevant paths inside the json. In the future we can point also to the mainnet part of the json with:  {{ #mainnet.aggregators }} and {{ #mainnet.publishers }} 

## Test plan

I tested this by serving the book locally after following the installation instructions in README.md

### DIFF
--- a/docs/book/usage/web-api.md
+++ b/docs/book/usage/web-api.md
@@ -117,6 +117,7 @@ requiring some authentication and compensation for the funds used.
 The following is a list of know public aggregators; they are checked periodically, but each of them
 may still be temporarily unavailable:
 
+<!-- markdownlint-disable proper-names -->
 {{ #testnet.aggregators }}
 {{ /testnet.aggregators }}
 
@@ -124,6 +125,7 @@ may still be temporarily unavailable:
 
 {{ #testnet.publishers }}
 {{ /testnet.publishers }}
+<!-- markdownlint-enable proper-names -->
 
 ## HTTP API Usage
 


.github/actions/build-mdbook/action.yaml
```diff
@@ -37,6 +37,10 @@ runs:
         url="https://github.com/lzanini/mdbook-katex/releases/download/0.9.0-binaries/mdbook-katex-v0.9.0-x86_64-unknown-linux-gnu.tar.gz"
         curl -sSL "$url" | tar -xz --directory=bin
       shell: bash
+    - name: Install mdbook-templates from crates.io
+      uses: baptiste0928/cargo-install@v3
+      with:
+        crate: mdbook-templates
```
mlegner: I see that the source repo for this (https://github.com/vlegakis/mdbook-templates) is owned by you. Was this forked from somewhere? Also, can this be made public?

mlegner: We use `taiki-e/install-action` in other places. Also, can you pin the action to a specific commit?
```suggestion
      uses: taiki-e/install-action@970d55e3ce02a46d60ffae7b4fab3dedace6e769 # pin@v2.49.13
      with:
        tool: mdbook-templates@0.2.0
```

docs/book/usage/web-api.md
```diff
@@ -117,133 +117,13 @@ requiring some authentication and compensation for the funds used.
...
-- `http://walrus-testnet.equinoxdao.xyz:9000`
-- `https://walrus-testnet-aggregator.stakely.io`
-- `https://walrus-testnet-aggregator.talentum.id`
-- `http://37.27.230.228:9000`
-- `https://walrus-testnet-aggregator.luckyresearch.org/`
-- `http://walrus-testnet.suicore.com:9000`
-- `https://walrus-testnet.validators.services.kyve.network/aggregate`
-- `https://walrus-aggregator-testnet.suisec.tech`
-- `https://walrus-testnet-aggregator.dzdaic.com`
+{{ #testnet.aggregators }}
```
mlegner, line 121: Suggestion: Disable the `proper-names` lint here:
```md
<!-- markdownlint-disable proper-names -->
{{ #testnet.aggregators }}
{{ /testnet.aggregators }}

### Public publishers

{{ #testnet.publishers }}
{{ /testnet.publishers }}
<!-- markdownlint-enable proper-names -->
```

.github/actions/build-mdbook/action.yaml
vlegakis: Fot this preprocessing I tried many mdbook preprocessors from which none worked as expected (mdbook-variables, mdbook-terra etc)
Thus I created this small and handy plugin which does exactly what we want. The problem is that it does exactly what we want and is not as generic as I wanted it to be (e.g hardcoded sorting for https>http>bare IPs etc) 
I can sure make it public but wanted to work a bit more on it first since its on a very quick and dirty level yet :(

mlegner: OK, that sounds good to me. 👍

PR #1783, opened 2025-03-21
feat: add metrics to the thread pool and configuration variable for max concurrent tasks

jpcsmith:
## Description

This adds metrics for latency and queue length to the thread pool, as well as a config variable to configure the maximum number of concurrent requests, which overrides the default calculation (if non-None).

## Test plan

Local testbed + existing tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/thread_pool.rs
+++ b/crates/walrus-service/src/node/thread_pool.rs
@@ -107,7 +107,7 @@ impl ThreadPoolBuilder {
     /// Creates a [`BlockingThreadPool`] with the specified configuration.
     pub fn build(&mut self) -> BlockingThreadPool {
         let inner = self.thread_pool.take().unwrap_or_else(|| {
-            // The default is Rayon when unspecified, unless in msim.
+            // The default is Rayon when unspecified, unless in simtest.
             if cfg!(msim) {
                 BlockingThreadPoolInner::Tokio(TokioBlockingPool::default())
             } else {


crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -43,31 +50,94 @@ pub(crate) const DEFAULT_MAX_WAIT_TIME: Duration = Duration::from_secs(1);
...
+    #[cfg(test)]
+    pub(crate) fn tokio(&mut self, pool: TokioBlockingPool) -> &mut Self {
+        self.thread_pool = Some(BlockingThreadPoolInner::Tokio(pool));
+        self
+    }
+
+    /// Creates a [`BlockingThreadPool`] with the specified configuration.
+    pub fn build(&mut self) -> BlockingThreadPool {
+        let inner = self.thread_pool.take().unwrap_or_else(|| {
+            // The default is Rayon when unspecified, unless in msim.
```
halfprice:
```suggestion
            // The default is Rayon when unspecified, unless in simtest.
```

crates/walrus-service/src/node/thread_pool.rs
```diff
@@ -233,16 +339,39 @@ where
...
+        let queued_latency = walrus_utils::with_label!(self.metrics.latency_seconds, STATUS_QUEUED);
+        let processing_latency =
+            walrus_utils::with_label!(self.metrics.latency_seconds, STATUS_IN_PROGRESS);
+
+        let request = move || {
+            // Drop the guard as soon as the task begins being processed.
+            drop(in_queue_guard);
+
+            let processing_start = Instant::now();
+            queued_latency.observe(processing_start.duration_since(queued_start).as_secs_f64());
```
halfprice, line 364: nit: feel `queued_start.elapsed()` is easier to read.

crates/walrus-service/src/node/config.rs
```diff
@@ -168,6 +168,9 @@ pub struct StorageNodeConfig {
     /// Configuration for background SUI balance checks and alerting.
     #[serde(default, skip_serializing_if = "defaults::is_default")]
     pub balance_check: BalanceCheckConfig,
+    /// Configuration for the blocking thread pool.
+    #[serde(default, skip_serializing_if = "defaults::is_default")]
+    pub thread_pool: ThreadPoolConfig,
```
halfprice, line 173: Question: should this be part of the BlobRecoveryConfig? In case there will be another needs to create another thread pool to do some other work.

crates/walrus-service/src/node/thread_pool.rs
halfprice, line 364: IIUC, the latency/queue length measured here does not include the queuing delay enforced by the ConcurrencyLimit, is that right? If so, is this intended?

jpcsmith: Yea, it is, but this guarantees that the measured interval is continuous, i.e., `[queued_start, processing_start]` and `[processing_start, end]`.

crates/walrus-service/src/node/config.rs
jpcsmith: It is not part of the `BlobRecoveryConfig` because the same thread-pool will be used by other services within the node.

crates/walrus-service/src/node/thread_pool.rs
jpcsmith: Correct, this measures the queueing delay in Rayon's internal work queues. There is currently no queueing delay enforced by the `ConcurrencyLimit`, because we never block and wait on a permit to become available (since there is already queueing going on in Rayon).

PR #1784, opened 2025-03-21
chore: ignore initial params sync failures.

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -298,7 +298,6 @@ pub struct StorageNodeBuilder {
     committee_service: Option<Arc<dyn CommitteeService>>,
     contract_service: Option<Arc<dyn SystemContractService>>,
     num_checkpoints_per_blob: Option<u32>,
-    ignore_sync_failures: bool,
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -529,6 +529,7 @@ pub struct NodeParameters {
     // If not provided, the default value will be used.
     num_checkpoints_per_blob: Option<u32>,
     // Whether to ignore sync failures during node initialization
+    #[allow(dead_code)]
     ignore_sync_failures: bool,
```
mlegner: Question: If this is no longer used, can we just remove it?

liquid-helium: I was thinking some users my be affected by the Run command (it is mostly unlikely thought).

mlegner: I guess we could still keep it there and mark it as `#[deprecated]`. Maybe something like `#[deprecated(since = "1.19.0", note = "This is now the default behavior")]`.

PR #1786, opened 2025-03-21
chore: write a test driver for restaking WAL during antithesis tests

wbbradley:
## Description

- Automate running the antithesis pipeline locally.
- Write a test driver for restaking WAL during antithesis tests.
 
## Test plan

CI, Manual testing.

No release impact.

### DIFF
--- a/crates/walrus-stress/src/main.rs
+++ b/crates/walrus-stress/src/main.rs
@@ -4,6 +4,7 @@
 //! Load generators for stress testing the Walrus nodes.
 
 use std::{
+    collections::{HashMap, HashSet},
     net::{IpAddr, Ipv4Addr, SocketAddr},
     num::{NonZeroU64, NonZeroUsize},
     path::PathBuf,
@@ -12,12 +13,19 @@ use std::{
 };
 
 use anyhow::Context;
-use clap::Parser;
+use clap::{Parser, Subcommand};
+use rand::{seq::SliceRandom, RngCore};
+use sui_types::base_types::ObjectID;
 use walrus_service::{
     client::{metrics::ClientMetrics, Config, Refiller},
     utils::load_from_yaml,
 };
-use walrus_sui::{config::load_wallet_context_from_path, utils::SuiNetwork};
+use walrus_sui::{
+    client::{CoinType, ReadClient, SuiContractClient, MIN_STAKING_THRESHOLD},
+    config::WalletConfig,
+    types::StorageNode,
+    utils::SuiNetwork,
+};
 
 use crate::generator::LoadGenerator;
 
@@ -28,10 +36,48 @@ const COIN_REFILL_AMOUNT: u64 = 500_000_000;
 /// The minimum balance to keep in the wallet.
 const MIN_BALANCE: u64 = 1_000_000_000;
 
+#[derive(Parser)]
+#[clap(rename_all = "kebab-case")]
+#[clap(name = env!("CARGO_BIN_NAME"))]
+#[derive(Debug)]
+struct Args {
+    /// The path to the client configuration file containing the system object address.
+    #[clap(long, default_value = "./working_dir/client_config.yaml")]
+    config_path: PathBuf,
+
+    /// Sui network for which the config is generated.
+    #[clap(long, default_value = "testnet")]
+    sui_network: SuiNetwork,
+
+    /// The port on which metrics are exposed.
+    #[clap(long, default_value = "9584")]
+    metrics_port: u16,
+
+    /// The path to the Sui Wallet to be used for funding the gas.
+    ///
+    /// If specified, the funds to run the stress client will be taken from this wallet. Otherwise,
+    /// the stress client will try to use the faucet.
+    #[clap(long)]
+    wallet_path: Option<PathBuf>,
+
+    #[command(subcommand)]
+    command: Commands,
+}
+
+#[derive(Subcommand, Debug, Clone)]
+#[clap(rename_all = "kebab-case")]
+enum Commands {
+    /// Register nodes based on parameters exported by the `walrus-node setup` command, send the
+    /// storage-node capability to the respective node's wallet, and optionally stake with them.
+    Stress(StressArgs),
+    /// Deploy the Walrus system contract on the Sui network.
+    Staking,
+}
+
 #[derive(Parser, Debug, Clone)]
 #[clap(rename_all = "kebab-case")]
-#[command(author, version, about = "Walrus load generator", long_about = None)]
-struct Args {
+#[command(author, version, about = "Walrus stress load generator", long_about = None)]
+struct StressArgs {
     /// The target write load to submit to the system (writes/minute).
     /// The actual load may be limited by the number of clients.
     /// If the write load is 0, a single write will be performed to enable reads.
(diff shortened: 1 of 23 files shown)


docker/walrus-antithesis/build-test-config-image/pipeline.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+export DOCKER_DEFAULT_PLATFORM=linux/amd64
```
halfprice: Question: this is for running this cluster suite + clients locally?

docker/walrus-antithesis/build-test-config-image/files/run-staking.sh
```diff
@@ -0,0 +1,46 @@
...
+
+HOSTNAME=$(hostname)
+
+# create directories
+mkdir -p /root/.sui/sui_config
+mkdir -p /root/.config/walrus
+mkdir -p /opt/walrus/outputs
+
+# copy from deploy outputs
+# TODO(zhewu): here we are using the admin wallet to run the staking client, which is not ideal.
```
halfprice: question: is this setup actually work? Given that the read-write client and staking client all use the same wallet, it may encounter locked object issue on Sui. However, given that the stake client runs transaction very infrequently, there can be a run which does not encounter this issue.

Ultimately, we need to address this TODO in order to run multiple test client reliably. I can help with that.

crates/walrus-stress/src/main.rs
```diff
@@ -124,3 +180,118 @@ async fn main() -> anyhow::Result<()> {
...
+    args: StakingArgs,
+) -> anyhow::Result<()> {
+    let _current_epoch = 0;
+    // Start the re-staking machine.
+    let restaking_period = Duration::from_secs(args.restaking_period_seconds.get());
+    let contract_client: SuiContractClient = config.new_contract_client(wallet, None).await?;
+    // The amount of WAL staked for StakedWal.
+    let mut wal_staked: BTreeMap<ObjectID, u64> = Default::default();
+
+    // REVIEW: is it OK to stake prior to epoch 1?
```
halfprice: Yes.

crates/walrus-stress/src/main.rs
```diff
@@ -124,3 +180,118 @@ async fn main() -> anyhow::Result<()> {
...
+                if epoch <= current_epoch {
+                    let mut nodes: Vec<StorageNode> = committee.members().to_vec();
+
+                    // Shuffle the nodes to determine which ones get preferential staking
+                    // treatment.
+                    nodes.shuffle(&mut rand::thread_rng());
+
+                    // Allocate half the WAL to various nodes. This is a linear walk over all of the
+                    // WAL which we're going to stake, just to simplify the algorithm. Each "stake"
+                    // below is one unit of MIN_STAKING_THRESHOLD.
```
halfprice, line 228: A more interesting scenario would be to stake very different amount to different WAL, to exercise the shard re-assignment.

wbbradley: I'll take a pass at making the distribution more extreme.

docker/walrus-antithesis/build-test-config-image/files/run-staking.sh
wbbradley: Makes sense, yeah, I'll figure this out next.

docker/walrus-antithesis/build-test-config-image/pipeline.sh
wbbradley: This script is for running locally. That variable is no longer needed.

docker/walrus-antithesis/build-test-config-image/pipeline.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
```
halfprice, line 1: I think it's helpful to merge this script, so that in the future, we can directly use it to run a local cluster. Can we move it to walrus/scripts dir, since this is not needed for antithesis test?

docker/walrus-antithesis/build-walrus-image-for-antithesis/build.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
halfprice, line 1: Similarly, can we move this to walrus/scripts?

halfprice (approved):
Thank you so much @wbbradley ! This looks great and I look forward to it being super useful in antithesis test! Let's merge this and see what testing results it can produce.

docker/walrus-antithesis/build-walrus-image-for-antithesis/build.sh
wbbradley: This script is quite particular for the antithesis docker image build. I'm not sure if it should live in scripts..? I thought about moving it to `docker/walrus-antithesis` but that directory has multiple Dockerfiles, which sort of goes against the pattern in the other `docker/*` directories. So, if anything, I'd like to make the pattern of file layouts in the `docker/*` subdirs all identical, with their corresponding `build.sh` in that `docker/*/build.sh` position, until we unify all of the build.sh scripts into something that's in `scripts/`. Does that make sense? I see it as a couple hops before we get all the way cleaned up.

wbbradley: I take it back. This script is generic. Yes, we should move it and the others like it in `docker/*/build.sh` into scripts, or into `docker/`. The trouble is that there is already a pattern of using copies of this `build.sh` script over in the `sui` codebase that follows the simplified pattern discussed here. It would be good to avoid deviating from that pattern unless we plan to clean up all instances of the pattern across all codebases, imho

wbbradley: @pei-mysten I think you may be the owner/originator of this pattern. I could unify all of these with a little bit of work, but I'd like to understand your thoughts on this layout/duplication of the build.sh script, etc before doing anything beyond copy/pasting.

docker/walrus-antithesis/build-test-config-image/pipeline.sh
wbbradley: Currently this script is only used to handle the local antithesis-like build/run pipeline. I could move it elsewhere, but I'm not yet convinced that it would improve the legibility/usability of it significantly, while at least having it next to the docker-compose.yaml it relates to is somewhat helpful. I agree that we should iterate on file layout in the repo though. Let's hold off on that for this go round, if that's OK?

halfprice:
Hi @wbbradley , the simtest should be fixed. Hopefully rebasing the PR should make the CI passing

docker/walrus-antithesis/build-test-config-image/pipeline.sh
```diff
@@ -0,0 +1,77 @@
...
+    git checkout "$sui_version" || die "Failed to checkout SUI version '$sui_version'"
+    cd docker/sui-tools
+    ./build.sh -t "$local_sui_image_name" || die "Failed to build SUI image"
+    # Get SUI image and push to local registry.
+  ) || die "failed to build SUI image"
+
+  local_walrus_image="walrus-antithesis:$sui_version"
+
+  export WALRUS_IMAGE_NAME="$local_walrus_image"
+  export SUI_IMAGE_NAME="mysten/sui-tools:$sui_version"
```
shuowang12, line 48: Hi Will @wbbradley. I was wondering why `SUI_IMAGE_NAME` is not set to `local_sui_image_name` like `WALRUS_IMAGE_NAME`. I got errors of `no such image` when running this `pipeline.sh` locally.
``` 
Container build-test-config-image-sui-localnet-1 Error response from daemon: No such image: mysten/sui-tools:testnet-v1.67.1
Error response from daemon: No such image: mysten/sui-tools:testnet-v1.67.1
```
`mysten/sui-tools:testnet-v1.67.1` does exist remotely in the docker hub if I pull it but it is amd64 only and not for macbook's arm64.

wbbradley: Good question. I don't recall. It very well could have been a typo.

shuowang12: Let me change it then. 
I also find it takes more than 10 minutes to build sui and walrus every time I run this script as the cache is dropped. I managed to find a way to mount and store the cache and will submit a PR along with it.

PR #1788, opened 2025-03-23
fix: improvements to checkpoint downloader

mlegner:
## Description

- Add a configurable timeout to all gRPC requests.
- Correctly handle HTTP errors from fallback node.
- Do not directly use fallback for not yet produced checkpoints.
- Return error from main RPC node when doing "quick retries".
- Minor cleanup.

Contributes to WAL-716

## Test plan

* Automatic tests.
* When running catchup on Mainnet (`RUST_LOG=info,walrus_sui::client::retry_client=debug ./target/release/walrus-node catchup --db-path /tmp/catchup --system-object-id 0x2134d52768ea07e8c43570ef975eb3e4c27a39fa6396bef985b5abc58d03ddd2 --staking-object-id 0x10b9d30c28448939ce6c4d6c6e0ffce4a7f8a4ada8248bdad09ef8b70e4a3904 --sui-rpc-url "https://mysten-rpc.mainnet.sui.io:443" --event-stream-catchup-min-checkpoint-lag 100000 --checkpoint-bucket https://checkpoints.mainnet.sui.io 2>&1 | tee /tmp/logs`, successfully downloads old checkpoints that were pruned from the RPC node:
  ```
  2025-03-23T15:39:55.991755Z DEBUG get_full_checkpoint{sequence=122523195}: walrus_sui::client::retry_client: primary client error while fetching checkpoint: RpcError(Status { code: NotFound, message: "data from requested checkpoint 122523195 has been pruned", details: b"\x08\x05\x128data from requested checkpoint 122523195 has been pruned", metadata: MetadataMap { headers: {"content-type": "application/grpc", "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000", "via": "1.1 google", "content-length": "0", "x-sui-chain-id": "35834a8a", "x-sui-chain": "mainnet", "x-sui-epoch": "710", "x-sui-checkpoint-height": "125857499", "x-sui-timestamp-ms": "1742744395471", "x-sui-timestamp": "2025-03-23T15:39:55.471Z", "x-sui-lowest-available-checkpoint": "124566728", "x-sui-lowest-available-checkpoint-objects": "124923727", "vary": "origin, access-control-request-method, access-control-request-headers", "access-control-allow-origin": "*", "server-timing": "finish_request;dur=0", "date": "Sun, 23 Mar 2025 15:39:55 GMT"} }, source: None })
  2025-03-23T15:39:55.991763Z DEBUG get_full_checkpoint{sequence=122523195}: walrus_sui::client::retry_client: using fallback client to fetch checkpoint as error: RpcError(Status { code: NotFound, message: "data from requested checkpoint 122523195 has been pruned", details: b"\x08\x05\x128data from requested checkpoint 122523195 has been pruned", metadata: MetadataMap { headers: {"content-type": "application/grpc", "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000", "via": "1.1 google", "content-length": "0", "x-sui-chain-id": "35834a8a", "x-sui-chain": "mainnet", "x-sui-epoch": "710", "x-sui-checkpoint-height": "125857499", "x-sui-timestamp-ms": "1742744395471", "x-sui-timestamp": "2025-03-23T15:39:55.471Z", "x-sui-lowest-available-checkpoint": "124566728", "x-sui-lowest-available-checkpoint-objects": "124923727", "vary": "origin, access-control-request-method, access-control-request-headers", "access-control-allow-origin": "*", "server-timing": "finish_request;dur=0", "date": "Sun, 23 Mar 2025 15:39:55 GMT"} }, source: None })
  2025-03-23T15:39:55.991770Z DEBUG get_full_checkpoint{sequence=122523195}: walrus_sui::client::retry_client: downloading checkpoint from fallback bucket url=https://checkpoints.mainnet.sui.io/122523195.chk
  [...]
  2025-03-23T15:39:56.004833Z DEBUG get_full_checkpoint{sequence=122523195}: walrus_sui::client::retry_client: checkpoint download successful sequence_number=122523195
  ```
* Ran it on a local testbed.
* Deployed 3b1a2d1190ad730312b7301dcbf0b3ab7f73425c to one PTN node (`cdg-ptn-sto-00`). Before this change, the logs are full with the following warnings:
  ```
  Mar 23 15:41:39 cdg-ptn-sto-00 walrus-node[171180]: {"timestamp":"2025-03-23T15:41:39.515796Z","level":"WARN","fields":{"message":"failed to read next checkpoint","next_checkpoint":176468001,"err":"Some(FallbackError(DeserializationError(\"No discriminant in enum `BlobEncoding` matches the value `60`\")))"},"target":"checkpoint_downloader::downloader","filename":"crates/checkpoint-downloader/src/downloader.rs","line_number":407}
  ```
  After, checkpoints are still downloaded continuously with a single worker in the steady state, and logs look smooth again.

---

## Release notes

- [x] Storage node: Adds a configurable timeout to checkpoint requests.

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -894,9 +894,10 @@ impl RetriableClientError {
         match self {
             Self::TimeoutError(_) => true,
             Self::RpcError(status) => {
-                status
-                    .checkpoint_height()
-                    .is_some_and(|height| next_checkpoint < height)
+                (status.code() == tonic::Code::NotFound
+                    && status
+                        .checkpoint_height()
+                        .is_some_and(|height| next_checkpoint < height))
                     || (status.code() == tonic::Code::Internal
                         && status.message().contains("missing event"))
             }


crates/walrus-service/src/node/events.rs
```diff
@@ -35,6 +35,10 @@ pub struct EventProcessorConfig {
     #[serde_as(as = "DurationSeconds")]
     #[serde(rename = "pruning_interval_secs")]
     pub pruning_interval: Duration,
+    /// The timeout for the RPC client.
+    #[serde_as(as = "DurationSeconds")]
+    #[serde(rename = "timeout_secs")]
+    pub timeout: Duration,
```
halfprice: Nit: can we name this to be a bit specific like sui_rpc_timeout?

crates/walrus-service/src/node/events.rs
```diff
@@ -50,6 +54,7 @@ impl Default for EventProcessorConfig {
     fn default() -> Self {
         Self {
             pruning_interval: Duration::from_secs(3600),
+            timeout: Duration::from_secs(10),
```
halfprice: Remark: make default to be at least 1 minutes or even longer. This timeout is meant for catching abnormal RPC behavior, as opposite to improve performance. From overload protection perspective, shorter timeout may increase the load on the server side if the server handles requests with > 10 seconds latency.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -959,36 +992,32 @@ impl RetriableRpcClient {
...
-        if err.code() == tonic::Code::NotFound
-            || (err.code() == tonic::Code::Internal && err.message().contains("missing event"))
-        {
+        if err.should_use_fallback_directly(sequence) {
             tracing::debug!(
-                "Using fallback client to fetch checkpoint as error: {:?}",
+                "using fallback client to fetch checkpoint as error: {:?}",
                 err
             );
             return retry_rpc_errors(self.get_strategy(), || async {
```
halfprice, line 1027: Question: should we add a timeout for get_full_checkpoint from fallback as well? It's a Reqwest client so we can directly set it when creating it.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -856,41 +854,61 @@ impl CheckpointBucketClient {
...
 }
 
+impl RetriableClientError {
+    /// Returns `true` if the error should be handled by the fallback client immediately instead of
+    /// retrying the primary client.
+    fn should_use_fallback_directly(&self, next_checkpoint: u64) -> bool {
+        match self {
+            Self::TimeoutError(_) => true,
+            Self::RpcError(status) => {
+                status
```
sadhansood: I think we should check for status to be `NotFound` along with the checkpoint height. The original intention for quick fallback was for missing events checkpoints and pruned checkpoints (which is determined with NotFound and checkpoint height). All other errors should get retried on primary endpoint first imo.

crates/walrus-service/src/node/events.rs
mlegner: Good point, will make the change. 👍

sadhansood: I think the dilemma is that a timeout that's too small creates unnecessary load on the server (and even 10s seems too much when we're producing 5 checkpoints per second and not cancelling the underlying request). But if the timeout is too high, then every checkpoint request waits too long before falling back.
Maybe we should just accept that FN will get unresponsive sometimes. We could directly send requests to the fallback endpoint instead of always trying the primary first. Maybe use something like a leaky bucket that refills with timeouts every hour?
Like we start with 300s total timeout in the bucket. Each request timeout is 60s, and after 5 requests timeout, we just stop sending requests to primary completely. Then the bucket refills every hour or so.

crates/walrus-service/src/node/events.rs
```diff
@@ -50,6 +54,7 @@ impl Default for EventProcessorConfig {
     fn default() -> Self {
         Self {
             pruning_interval: Duration::from_secs(3600),
+            checkpoint_request_timeout: Duration::from_secs(60),
```
sadhansood, line 57: I think the dilemma is that a timeout that's too small creates unnecessary load on the server (and even 5s seems too much when we're producing 5 checkpoints per second). But if the timeout is too high, then every checkpoint request waits too long before falling back.
Maybe we should just accept that FN will get unresponsive sometimes. We could directly send requests to the fallback endpoint instead of always trying the primary first. Maybe use something like a leaky bucket that refills with timeouts every hour?
Like we start with 300s total timeout in the bucket. Each request timeout is 60s, and after 5 requests timeout, we just stop sending requests to primary completely. Then the bucket refills every hour or so.

mlegner: Hmm... I agree that the current solution doesn't help *during* a time where the RPC node is unresponsive. It only helps with recovery afterwards. In practice, when we now use the `mysten-rpc` nodes, which are redundant and have a load balancer, an outage like last week is much less likely.

I agree that a solution like you describe could fix this. However, I'm a bit hesitant to include such a relatively complex solution in a cherry-pick that will be deployed to Mainnet today.

So my suggestion would be to keep the immediate fix as it is now and work on the complete fallback story (including secondary RPC nodes) in a follow-up. WDYT?

crates/walrus-sui/src/client/retry_client.rs
mlegner: Done in 4036950793a17882718b7769125c2ca7d2328143.

PR #1790, opened 2025-03-24
feat: blob info consistency check

halfprice:
## Description

This PR adds a background task that produce a digest of the list of certified blob at the beginning of an epoch.

Given that all the nodes should process all the events in the same way updating the blob info table, all the node
should produce the exact same digest.

The digest is exposed via metrics.

This PR also adds consistency check in simtest to check that all storage nodes should have the same digest for
each epoch.

Contributes to WAL-701

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/consistency_check.rs
+++ b/crates/walrus-service/src/node/consistency_check.rs
@@ -1,4 +1,4 @@
-// Copyright (c) Mysten Labs, Inc.
+// Copyright (c) Walrus Foundation
 // SPDX-License-Identifier: Apache-2.0
 
 //! Background consistency check for the storage node.


crates/walrus-simtest/tests/simtest.rs
```diff
@@ -1015,6 +1026,24 @@ mod tests {
             }
             tokio::time::sleep(Duration::from_secs(1)).await;
         }
+
+        // Ensure that for all the epoch, all the nodes have the same certified blob digest.
+        let digest_map = certified_blob_digest_map.lock().unwrap();
+        for (epoch, node_digest_map) in digest_map.iter() {
```
halfprice: I'll extend the check to more test tomorrow. Here is the basic idea to check across node blob info consistency.

crates/walrus-service/src/node.rs
```diff
@@ -1983,6 +2001,86 @@ impl StorageNodeInner {
...
+
+    /// Schedule a background task to compute the hash of the list of certified blobs at the
+    /// beginning of the epoch. This is used to detect inconsistencies in the blob info table
+    /// between the nodes.
+    async fn schedule_background_blob_info_consistency_check(
+        self: &Arc<StorageNodeInner>,
+        epoch: Epoch,
+    ) -> anyhow::Result<()> {
+        let this = self.clone();
+        let (tx, rx) = tokio::sync::oneshot::channel();
```
jpcsmith: Suggestion: Move this task into its own module, the file is already over 2k lines long.

crates/walrus-service/Cargo.toml
jpcsmith, line 1: Question: Why `twox-hash` versus any of the other rust hash crates?

crates/walrus-service/src/node.rs
```diff
@@ -1983,6 +2001,86 @@ impl StorageNodeInner {
...
+            let blob_info_iterator = this.storage.certified_blob_info_iter_before_epoch(epoch);
+            let _ = tx.send(());
+
+            Self::compose_certified_blob_list_digest(this.clone(), blob_info_iterator, epoch);
+        });
+
+        // We need to make sure that the function returns only after the blob info iterator is
+        // created, so that it can operate on a consistent snapshot of the blob info table.
+        // Otherwise, processing future events may cause inconsistency in different nodes.
+        rx.await.map_err(|err| {
```
jpcsmith: Suggestion: construct the iterator outside of the blocking call, then move it into the blocking context, so that you can avoid the need for the channel. Also `tokio::sync::Barrier` achieves a similar effect and may be more efficient since it does not need to be able to send a value.

crates/walrus-service/src/node.rs
```diff
@@ -1983,6 +2001,86 @@ impl StorageNodeInner {
...
+            )
+        })?;
+        Ok(())
+    }
+
+    fn compose_certified_blob_list_digest(
+        node: Arc<StorageNodeInner>,
+        blob_info_iter: BlobInfoIterator,
+        epoch: Epoch,
+    ) {
```
jpcsmith: Question: Do we already have a plan on how this check will evolve to when a larger number of blobs is stored?

crates/walrus-service/Cargo.toml
halfprice: During my sui time, we had similar needs for a hash function that needs to be fast, non-cryptographic-required and more importantly deterministic across platform, which xxhash seems to be the best option. I couldn't find the thread to discuss this since slack already GCed it. But I remember we compared the popurlar ones like xxhash (fast, and deterministic across platform), ahash (fast, not portable across platform), and blake3 (slower, since this one is cryptographic). So xxhash was the best.

I also just came across this blog post: https://blog.goose.love/posts/rosetta-hashing/ which also shows the xxhash is among the fastest. But again, the portability of the xxhash across different platform makes it seem to be the only reasonable choice.

crates/walrus-service/src/node.rs
halfprice: done

crates/walrus-service/src/node.rs
halfprice: Hmm, Barrier seems odd. That is for when multiple tasks all need to reach to a certain point before moving forward, whereas in this case, is more like producer-consumer relationship between the tasks (main thread need to wait for the blocking thread to reach to a point, whereas the blocking thread doesn't need to wait for the main thread for anything). I think maybe `Notify` is another option, but the benefit of using a one-shot channel is that the task can also propagate any error back, if any.

crates/walrus-service/src/node.rs
halfprice: Yes, that's will be the second followup PR after this one (the next one will be the blob object table consistency). The idea is that we will maintain a iterator for each column family and while going through the blob info table, we also go through the slivers table. We also need to account for ongoing blob sync, so that will be a more complex PR.

The output will be the percentage of blobs whose slivers are missing. But let me know if there are other metrics that are interesting to show.

contracts/walrus/sources/staking/active_set.move
```diff
@@ -22,10 +22,11 @@ public struct ActiveSetEntry has copy, drop, store {
 /// to a 1000 nodes.
 /// Additionally, the active set tracks the total amount of staked WAL to make
 /// the calculation of the rewards and voting power distribution easier.
-// TODO: implement a reserve to track N + K nodes, where N is the active set
-// size and K is the number of nodes that are in the process of being added to
-// the active set. This will allow us to handle removals from the active set
-// without refetching the nodes from the storage.
+///
+/// TODO: implement a reserve to track N + K nodes, where N is the active set
```
halfprice: This is to remove lint warning.

crates/walrus-simtest/tests/simtest.rs
halfprice: This is done now.

contracts/walrus/sources/staking/active_set.move
jpcsmith: Question: ~What was the lint warning (all the GH actions for this branch seem to have completed successfully)~, and is there another way to resolve it that doesn't put the TODO in public contract documentation?

contracts/walrus/sources/staking/staking_inner.move
```diff
@@ -34,7 +34,7 @@ use walrus::{
 const MIN_STAKE: u64 = 0;
 
 /// Temporary upper limit for the number of storage nodes.
-// TODO: Remove once solutions are in place to prevent hitting move execution limits (#935).
+/// TODO: Remove once solutions are in place to prevent hitting move execution limits (#935).
```
jpcsmith: Same as above.

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    // We need to make sure that the function returns only after the blob info iterator is
+    // created, so that it can operate on a consistent snapshot of the blob info table.
+    // Otherwise, processing future events may cause inconsistency in different nodes.
+    rx.await.map_err(|err| {
+        anyhow::anyhow!(
+            "background task failed to create iterator. Error: {:?}",
+            err
+        )
+    })?;
```
jpcsmith: Question: `err` here is a [`RecvError`](https://docs.rs/tokio/latest/tokio/sync/oneshot/error/struct.RecvError.html) which simply means the channel was closed before sending, it gives no information of the reason for the error. Is there any point in printing it?

Hint: There's [`anyhow::Context`](https://docs.rs/anyhow/latest/anyhow/trait.Context.html) that does this.

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -0,0 +1,96 @@
...
+
+    let value = hasher.finish();
+
+    tracing::info!(?epoch, certified_blob_hash = ?value,
+            "background blob info consistency check finished");
+    walrus_utils::with_label!(
+        node.metrics.blob_info_consistency_check,
+        (epoch % 100).to_string()
+    )
+    .set(value as i64);
```
jpcsmith: Remark: It would be good to know how many elements were processed along with the duration (which I assume the monitored_scope above gives) to have an idea of how this scales with number of certified blobs in the database.

contracts/walrus/sources/staking/active_set.move
jpcsmith: Suggestion: Put the TODO comment before docstrings, but keep as an internal comment.

halfprice: The error complains about a floating docstrings. I see quite a few `/// TODO` in the move code so I assume this is normal. Let me move them up.

contracts/walrus/sources/staking/staking_inner.move
halfprice: done.

crates/walrus-service/src/node/consistency_check.rs
halfprice: Done. Added context around it.

crates/walrus-service/src/node/consistency_check.rs
halfprice: Good point. Added counter to track the number of certified blobs scanned.

PR #1794, opened 2025-03-24
feat: Added extra Sui fullnode endpoints, so storage node can switch the endpoint upon failures

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-simtest/tests/simtest_event_blob.rs
+++ b/crates/walrus-simtest/tests/simtest_event_blob.rs
@@ -210,7 +210,8 @@ mod tests {
         Ok(())
     }
 
-    /// This test verifies that the node can correctly recover from a forked event blob.
+    /// This test verifies that the node can correctly download checkpoints from
+    /// additional fullnodes.
     #[ignore = "ignore integration simtests by default"]
     #[walrus_simtest]
     async fn test_checkpoint_downloader_with_additional_fullnodes() {
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -352,7 +352,7 @@ impl<T> FailoverWrapper<T> {
             .load(std::sync::atomic::Ordering::Relaxed);
         let mut current_index = start_index;
 
-        for _ in 0..self.max_retries.min(self.client_count()) {
+        for _ in 0..self.max_retries {
             let instance = self.get_client(current_index).await;
 
             let result = {
@@ -1353,26 +1353,25 @@ impl RetriableRpcClient {
         &self,
         sequence_number: u64,
     ) -> Result<CertifiedCheckpointSummary, RetriableClientError> {
-        self.client
-            .with_failover(|client| {
-                Box::pin(async move {
-                    retry_rpc_errors(
-                        self.get_strategy(),
-                        || async {
-                            self.handle_checkpoint_timeout(
-                                self.request_timeout,
-                                client.get_checkpoint_summary(sequence_number),
-                                sequence_number,
-                            )
-                            .await
-                        },
-                        self.metrics.clone(),
-                        "get_checkpoint_summary",
-                    )
-                    .await
-                })
+        let request = |client: Arc<RpcClient>| {
+            Box::pin(async move {
+                retry_rpc_errors(
+                    self.get_strategy(),
+                    || {
+                        self.handle_checkpoint_timeout(
+                            self.request_timeout,
+                            client.get_checkpoint_summary(sequence_number),
+                            sequence_number,
+                        )
+                    },
+                    self.metrics.clone(),
+                    "get_checkpoint_summary",
+                )
+                .await
             })
-            .await
+        };
+
+        self.client.with_failover(request).await
     }
 
     /// Gets the full checkpoint data for the given sequence number from the primary client.
@@ -1380,26 +1379,25 @@ impl RetriableRpcClient {
         &self,
         sequence_number: u64,
     ) -> Result<CheckpointData, RetriableClientError> {
-        self.client
-            .with_failover(|client| {
-                Box::pin(async move {
-                    retry_rpc_errors(
-                        self.get_strategy(),
-                        || async {
-                            self.handle_checkpoint_timeout(
-                                self.request_timeout,
-                                client.get_full_checkpoint(sequence_number),
-                                sequence_number,
-                            )
-                            .await
-                        },
-                        self.metrics.clone(),
-                        "get_full_checkpoint",
-                    )
-                    .await
-                })
+        let request = |client: Arc<RpcClient>| {
+            Box::pin(async move {
+                retry_rpc_errors(
+                    self.get_strategy(),
+                    || {
+                        self.handle_checkpoint_timeout(
+                            self.request_timeout,
+                            client.get_full_checkpoint(sequence_number),
+                            sequence_number,
+                        )
+                    },
+                    self.metrics.clone(),
+                    "get_full_checkpoint",
+                )
+                .await
             })
-            .await
+        };
+
+        self.client.with_failover(request).await
     }
 
     /// Gets the full checkpoint data for the given sequence number.
@@ -1482,45 +1480,41 @@ impl RetriableRpcClient {
     pub async fn get_latest_checkpoint_summary(
         &self,
     ) -> Result<CertifiedCheckpointSummary, RetriableClientError> {
-        self.client
-            .with_failover(|client| {
-                Box::pin(async move {
-                    retry_rpc_errors(
-                        self.get_extended_strategy(),
-                        || async {
-                            Self::handle_rpc_timeout(
-                                self.request_timeout,
-                                client.get_latest_checkpoint(),
-                            )
-                            .await
-                        },
-                        self.metrics.clone(),
-                        "get_latest_checkpoint_summary",
-                    )
-                    .await
-                })
+        let request = |client: Arc<RpcClient>| {
+            Box::pin(async move {
+                retry_rpc_errors(
+                    self.get_extended_strategy(),
+                    || {
+                        Self::handle_rpc_timeout(
+                            self.request_timeout,
+                            client.get_latest_checkpoint(),
+                        )
+                    },
+                    self.metrics.clone(),
+                    "get_latest_checkpoint_summary",
+                )
+                .await
             })
-            .await
+        };
+
+        self.client.with_failover(request).await
     }
 
     /// Gets the object with the given ID.
     pub async fn get_object(&self, id: ObjectID) -> Result<Object, RetriableClientError> {
-        self.client
-            .with_failover(|client| {
-                Box::pin(async move {
-                    retry_rpc_errors(
-                        self.get_extended_strategy(),
-                        || async {
-                            Self::handle_rpc_timeout(self.request_timeout, client.get_object(id))
-                                .await
-                        },
-                        self.metrics.clone(),
-                        "get_object",
-                    )
-                    .await
-                })
+        let request = |client: Arc<RpcClient>| {
+            Box::pin(async move {
+                retry_rpc_errors(
+                    self.get_extended_strategy(),
+                    || Self::handle_rpc_timeout(self.request_timeout, client.get_object(id)),
+                    self.metrics.clone(),
+                    "get_object",
+                )
+                .await
             })
-            .await
+        };
+
+        self.client.with_failover(request).await
     }
 }
 


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -202,6 +203,127 @@ where
...
+#[derive(Clone, Debug)]
+pub struct FailoverWrapper<T> {
+    inner: Arc<Vec<(Arc<T>, String)>>,
+    current_index: Arc<AtomicUsize>,
+    client_count: usize,
+}
+
+impl<T> FailoverWrapper<T> {
+    /// Creates a new failover wrapper.
+    pub fn new(instances: Vec<(T, String)>) -> Self {
```
sadhansood: Maybe return `Result<Self>` and return an error on an empty vec? It'll also simplify the empty vec error handling everywhere

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -202,6 +203,127 @@ where
...
+                    }
+                };
+
+                match result {
+                    Ok(result) => {
+                        self.current_index
+                            .store(current_index, std::sync::atomic::Ordering::Relaxed);
+                        return Ok(result);
+                    }
+                    Err(e) => {
```
sadhansood: Should we consider endpoint failover for certain sub-class of errors? For instance - connection/network errors, server errors, overload, etc. But i think some errors like client errors i.e. 400s, etc should just be returned to the caller. When checking tonic::Status i think `InvalidArgument`, `Unimplemented` , etc are some errors which we don't need to retry ever.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: We should also pass an optional instance for triggering direct fallback for certain class of errors: `pub fn new (instances: Vec<(T, String>), fallback: Option<T>)`. This is because for `missing events` error we likely want to fallback to the bucket directly instead of retrying on other endpoints (as they all would likely have the same issue)

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1001,19 +1112,43 @@ impl RetriableRpcClient {
...
+                    Box::pin(async move {
+                        client
+                            .get_full_checkpoint(sequence)
+                            .await
+                            .map_err(RetriableClientError::from)
+                    })
+                },
+                self.request_timeout,
+            )
+            .await;
```
sadhansood: This will quickly switch between endpoints upon any failure. I think we will probably want to retry first on the current endpoint to account for transient failures (network hiccups, overload, etc) before falling over. This also reduces unnecessary endpoint switching and gives more predictable behavior.  And for certain errors like "missing event" we most likely want to quickly fallback to the bucket instead of switching endpoints

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -872,6 +872,8 @@ impl CliOutput for NodeHealthOutput {
                     {event_heading}
                     Events persisted: {persisted}
                     Events pending: {pending}{highest_finished_event_index_output}
+                    Latest checkpoint sequence number: {latest_checkpoint_sequence_number_output}
```
sadhansood: Is this change related to endpoint switching? Maybe we can tackle it in a follow up PR

liquid-helium: You are totally right, here is the [PR](https://github.com/MystenLabs/walrus/pull/1805), this PR was originally re-based to the other one, but after several merges, I lost the track of it :/

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Nice catch, now the `retry_rpc_errors` is used to wrap the inner call.

That's a great point of quick fall-back on certain errors, I can add a list of quick-fail errors.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: That's a great idea!
I am a little concerning about the complexity of adding more retry/fall-back logic inside the FailoverWapper, since current we already have 3 levels of nested calls.

Do you think we can pass in a list of errors, and we these errors occurs, we return immediately from FailoverWrapper, and let the caller to do the fall-back?

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Sure!
Let's pass in a list of errors. As also discussed in other comments.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Nice suggestion, done.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Actually, now I am choosing a simple solution, if the error is not `is_retriable_rpc_error`, it will return immediately. I added log here to see what happens in prod, and we can adjust in the future when we have more insights.

let me know if there works.

sadhansood:
Thanks @liquid-helium! Could you rebase this PR over the other PR, so the changes here are just for client fallback. Thank you so much, it'll really help with doing review.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: I think we need something different from `is_retriable_rpc_error` here. For instance, consider the grpc error `tonic::Code::NotFound`. It is non retriable on the same endpoint because server has pruned a checkpoint but is retriable across endpoints because there is a chance another server has this checkpoint. I think we probably need to extend the trait `RetriableRpcError` to include multi endpoint errors: 
```rust
pub trait RetriableRpcError: Debug {
    /// Returns `true` if the error is a retriable network error.
    fn is_retriable_rpc_error(&self) -> bool;
    /// Returns `true` is the error is retriable across endoints.
    fn is_mutli_endpoint_retriable_rpc_error(&self) -> bool;
}
```

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,149 @@ where
...
+    current_index: Arc<AtomicUsize>,
+}
+
+impl<T> FailoverWrapper<T> {
+    /// Creates a new failover wrapper.
+    pub fn new(instances: Vec<(T, String)>) -> anyhow::Result<Self> {
+        let count = instances.len();
+        if count == 0 {
+            return Err(anyhow::anyhow!("No clients available"));
+        }
```
sadhansood: This could be `instance.is_empty()`

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,149 @@ where
...
+
+    fn client_count(&self) -> usize {
+        self.inner.len()
+    }
+
+    /// Gets a client at the specified index (wrapped around if needed).
+    async fn get_client(&self, index: usize) -> Option<Arc<T>> {
+        if self.client_count() == 0 {
+            return None;
+        }
```
sadhansood: This check seems unnecessary now given we dont allow empty vector. Also, we can make the return type `Arc<T>`

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,149 @@ where
...
+
+        let wrapped_index = index % self.client_count();
+        Some(self.inner[wrapped_index].0.clone())
+    }
+
+    /// Gets the name of the client at the specified index.
+    fn get_name(&self, index: usize) -> Option<&str> {
+        if self.client_count() == 0 {
+            return None;
+        }
```
sadhansood: Same here

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,149 @@ where
...
+    {
+        let mut last_error = None;
+        let start_index = self
+            .current_index
+            .load(std::sync::atomic::Ordering::Relaxed);
+        let mut current_index = start_index;
+
+        if self.client_count() == 0 {
+            panic!("No clients available in FailoverWrapper");
+        }
```
sadhansood: same here, no need to check this anymore

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,149 @@ where
...
+    //     F: for<'a> Fn(Arc<T>) -> Fut,
+    //     Fut: Future<Output = Result<R, RetriableClientError>>,
+    //     {
+    //     match tokio::time::timeout(timeout, self.failover_inner(operation)).await {
+    //         Ok(result) => result,
+    //         Err(_timeout) => Err(RetriableClientError::NonRetryableTimeoutError(_timeout)),
+    //     }
+    // }
+}
+
```
sadhansood: We should remove these commented lines

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1066,6 +1209,16 @@ pub enum RetriableClientError {
     Other(#[from] anyhow::Error),
 }
 
+impl RetriableClientError {
+    /// Returns `true` if the error is a retriable timeout error.
```
sadhansood: The comment doesn't match the code.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,149 @@ where
...
+            };
+
+            match result {
+                Ok(result) => {
+                    self.current_index
+                        .store(current_index, std::sync::atomic::Ordering::Relaxed);
+                    return Ok(result);
+                }
+                Err(e) => {
+                    if !self.should_retry(&e) {
```
sadhansood: I think we don't need to check whether an error type is retryable or not here. My argument is that full nodes are running independently and error encountered from one FN endpoint is generally localized to that node. My recommendation is to retry next FN endpoint without checking the error returned from the previous FN endpoint.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,149 @@ where
...
+    }
+
+    /// Returns `true` if we should retry the operation on the next client.
+    fn should_retry(&self, error: &RetriableClientError) -> bool {
+        error.is_retriable_rpc_error() || error.is_timeout_error()
+    }
+
+    /// Executes an operation on the current inner instance, falling back to the next one
+    /// if it fails.
+    async fn with_failover<F, Fut, R>(&self, operation: F) -> Result<R, RetriableClientError>
```
sadhansood, line 344: I think we should limit the number of endpoints to try to a max of 2 or 3.

sadhansood (commented):
Thanks @liquid-helium. The PR looks pretty good, I just have a small number of comments.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,149 @@ where
     }
 }
 
+/// A wrapper that provides failover functionality for any inner type.
+/// When an operation fails on the current inner instance, it will try the next one.
+#[derive(Clone, Debug)]
+pub struct FailoverWrapper<T> {
```
sadhansood, line 300: Would it be possible to write some unit tests for this data structure?

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Thanks for the catch, fix applied.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: oh, my bad, removed.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Good idea, change applied.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Yes! Done!

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: The method is removed.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Done.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,110 @@ where
...
+        F: for<'a> Fn(Arc<T>) -> Fut,
+        Fut: Future<Output = Result<R, RetriableClientError>>,
+    {
+        let mut last_error = None;
+        let start_index = self
+            .current_index
+            .load(std::sync::atomic::Ordering::Relaxed);
+        let mut current_index = start_index;
+
+        for _ in 0..self.max_retries.min(self.client_count()) {
```
sadhansood: Do we need to compute the `min` here? Don't we already do it in the `new` function above?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -287,6 +287,110 @@ where
...
+
+            match result {
+                Ok(result) => {
+                    self.current_index
+                        .store(current_index, std::sync::atomic::Ordering::Relaxed);
+                    return Ok(result);
+                }
+                Err(e) => {
+                    last_error = Some(e);
+                    tracing::warn!(
```
sadhansood, line 387: This should probably be a debug level, as today we do not log at warn level when retrying the same endpoint (to keep the behavior consistent)

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1228,31 +1330,69 @@ impl RetriableRpcClient {
...
+                            .await
+                        },
+                        self.metrics.clone(),
+                        "get_checkpoint_summary",
+                    )
+                    .await
+                })
+            })
+            .await
+    }
```
sadhansood, line 1375: @liquid-helium this function has gotten many levels deep, could we clean up and reduce indentation a little bit to make it readable. I took a stab but feel free to do anything but this feels a little bit easier to read: 
```rust
async fn get_checkpoint_summary_from_primary(
    &self,
    sequence_number: u64,
) -> Result<CertifiedCheckpointSummary, RetriableClientError> {
    let request = |client| {
        Box::pin(async move {
            let checkpoint_request = client.get_checkpoint_summary(sequence_number);
            let timeout_handler = self.handle_checkpoint_timeout(
                self.request_timeout,
                checkpoint_request,
                sequence_number,
            );
            
            retry_rpc_errors(
                self.get_strategy(),
                || async { timeout_handler.await },
                self.metrics.clone(),
                "get_checkpoint_summary",
            )
            .await
        })
    };
    
    self.client.with_failover(request).await
}
```

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1228,31 +1330,69 @@ impl RetriableRpcClient {
...
+                            )
+                            .await
+                        },
+                        self.metrics.clone(),
+                        "get_full_checkpoint",
+                    )
+                    .await
+                })
+            })
+            .await
```
sadhansood: similarly for this one

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1349,34 +1475,45 @@ impl RetriableRpcClient {
...
+                            )
+                            .await
+                        },
+                        self.metrics.clone(),
+                        "get_latest_checkpoint_summary",
+                    )
+                    .await
+                })
+            })
+            .await
```
sadhansood: same here

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1349,34 +1475,45 @@ impl RetriableRpcClient {
...
+                            Self::handle_rpc_timeout(self.request_timeout, client.get_object(id))
+                                .await
+                        },
+                        self.metrics.clone(),
+                        "get_object",
+                    )
+                    .await
+                })
+            })
+            .await
```
sadhansood: and here

crates/walrus-simtest/tests/simtest_event_blob.rs
```diff
@@ -162,4 +165,113 @@ mod tests {
...
+
+            nodes_to_check = lagging_nodes;
+
+            tokio::time::sleep(poll_interval).await;
+        }
+
+        Ok(())
+    }
+
+    /// This test verifies that the node can correctly recover from a forked event blob.
```
sadhansood: This comment seem out of place

sadhansood (commented):
Thanks @liquid-helium for addressing the comments. I have some basic stylistic comments at this point, and then this PR looks good to land.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: That's a nice suggestion.
I have to move the `checkpoint_request` and `timeout_handler` into retry_rpc_errors due to the move of future violation, but still we can reduce one layer of indentation.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: removed.

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: Can we keep it as warn for a while before we can add a metric to track how often this switch endpoint happen?

crates/walrus-simtest/tests/simtest_event_blob.rs
liquid-helium: updated!

sadhansood (approved):
Thanks @liquid-helium for your work on this, looks good to me to land. This is going to be very useful for storage node reliability!

PR #1802, opened 2025-03-24
feat: add tokio metrics to major storage node tasks

jpcsmith:
## Description

This enables the use of the `tokio-metrics` crate with the `prometheus::Registry` and instruments HTTP routes, event processing, and blob sync tasks `tokio_metrics::TaskMonitor` to record metrics for when tasks are poll slowly or are scheduled but delayed to be run due to other tasks polling slowly.

## Test plan

Local testbed and a unit tests

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/common/telemetry.rs
+++ b/crates/walrus-service/src/common/telemetry.rs
@@ -36,13 +36,19 @@ use prometheus::{
     IntGauge,
     IntGaugeVec,
     Opts,
+    Registry,
 };
+use reqwest::Method;
 use tokio::time::Instant;
+use tokio_metrics::TaskMonitor;
 use tower_http::trace::{MakeSpan, OnResponse};
 use tracing::{field, Span};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 use walrus_core::Epoch;
-use walrus_utils::http::{http_body::Frame, BodyVisitor, VisitBody};
+use walrus_utils::{
+    http::{http_body::Frame, BodyVisitor, VisitBody},
+    metrics::TaskMonitorFamily,
+};
 
 use super::active_committees::ActiveCommittees;
 
@@ -321,12 +327,49 @@ fn http_version(request: &axum::extract::Request) -> &'static str {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(crate) struct MetricsMiddlewareState {
+    inner: Arc<MetricsMiddlewareStateInner>,
+}
+
+impl MetricsMiddlewareState {
+    pub fn new(registry: &Registry) -> Self {
+        Self {
+            inner: Arc::new(MetricsMiddlewareStateInner {
+                metrics: HttpServerMetrics::new(registry),
+                task_monitors: TaskMonitorFamily::new(registry.clone()),
+            }),
+        }
+    }
+
+    /// Returns the task monitor for the route and request, where `http_route` should be a
+    /// known route or `UNMATCHED_ROUTE`.
+    fn task_monitor(&self, method: Method, http_route: &str) -> TaskMonitor {
+        self.inner
+            .task_monitors
+            .get_or_insert_with_task_name(&(method.clone(), http_route.to_owned()), || {
+                format!("{} {}", method, http_route)
+            })
+    }
+
+    fn metrics(&self) -> &HttpServerMetrics {
+        &self.inner.metrics
+    }
+}
+
+#[derive(Debug)]
+struct MetricsMiddlewareStateInner {
+    metrics: HttpServerMetrics,
+    task_monitors: TaskMonitorFamily<(Method, String)>,
+}
+
 /// Middleware that records the elapsed time, HTTP method, and status of requests.
 pub(crate) async fn metrics_middleware(
-    State(metrics): State<HttpServerMetrics>,
+    State(state): State<MetricsMiddlewareState>,
     request: axum::extract::Request,
     next: middleware::Next,
 ) -> axum::response::Response {
+    let metrics = state.metrics();
     // Manually record the time in seconds, since we do not yet know the status code which is
     // required to get the concrete histogram.
     let start = Instant::now();
@@ -365,7 +408,9 @@ pub(crate) async fn metrics_middleware(
         ))
     });
 
-    let response = next.run(request).await;
+    let monitor = state.task_monitor(http_request_method.clone(), &http_route);
+    let response = monitor.instrument(next.run(request)).await;
+
     let response_available_at = Instant::now();
     let http_response_status_code = response.status();
     let error_type = if http_response_status_code.is_client_error()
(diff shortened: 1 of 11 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -875,93 +879,106 @@ impl StorageNode {
         let mut maybe_epoch_at_start = Some(self.inner.committee_service.get_epoch());
 
         let mut indexed_element_stream = index_stream.zip(event_stream);
+        let task_monitors = TaskMonitorFamily::<&'static str>::new(self.inner.registry.clone());
```
jpcsmith, line 889: The changes in this method are adding this line, then moving the loop body into an async block which is then instrumented.

There are no changes to the business logic in this method.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -277,14 +281,16 @@ impl BlobSyncHandler {
                 let blob_sync_handler_clone = self.clone();
                 let permits_clone = self.permits.clone();
 
-                let sync_handle = tokio::spawn(async move {
+                let monitor = self.task_monitors.get_or_insert(&"blob_recovery");
```
halfprice, line 284: question: sync all events' processing is monitored, do we need to explicitly add blob_recovery as well?

crates/walrus-service/src/common/telemetry.rs
```diff
@@ -321,12 +327,49 @@ fn http_version(request: &axum::extract::Request) -> &'static str {
...
+            }),
+        }
+    }
+
+    /// Returns the task monitor for the route and request, where `http_route` should be a
+    /// known route or `UNMATCHED_ROUTE`.
+    fn task_monitor(&self, method: Method, http_route: &str) -> TaskMonitor {
+        self.inner
+            .task_monitors
+            .get_or_insert_with_task_name(&(method.clone(), http_route.to_owned()), || {
```
halfprice, line 350: Question: what's the overhead of a monitor? Looks like we are creating one monitor per route, and the number of metrics tracked seems a lot.

crates/walrus-service/src/common/telemetry.rs
```diff
@@ -321,12 +327,49 @@ fn http_version(request: &axum::extract::Request) -> &'static str {
     }
 }
 
+#[derive(Clone, Debug)]
+pub(crate) struct MetricsMiddlewareState {
+    inner: Arc<MetricsMiddlewareStateInner>,
```
halfprice, line 332: Suggestion/question: do we need the `MetricsMiddlewareStateInner`? Seems we can just put `metrics` and `task_monitors` here, as all `MetricsMiddlewareStateInner` does is to hold these two struct, IIUC.

crates/walrus-utils/src/metrics/tokio.rs
```diff
@@ -0,0 +1,439 @@
...
+///
+/// Stores a thread-safe map of keys `K` to [`TaskMonitor`]s. Existing or new `TaskMonitor`s can
+/// be retrieved with [`Self::get_or_insert_with_task_name`].
+#[derive(Debug, Clone)]
+pub struct TaskMonitorFamily<K> {
+    inner: Arc<TaskMonitorFamilyInner<K>>,
+}
+
+#[derive(Debug)]
+struct TaskMonitorFamilyInner<K> {
```
halfprice: same question for the necessity of the inner struct here.

crates/walrus-utils/src/metrics/tokio.rs
```diff
@@ -0,0 +1,439 @@
...
+    (@as_f64 (($name:ident in seconds) -> $_:tt) $tokio_metrics:ident) => {
+        ($tokio_metrics.$name).as_secs_f64()
+    };
+    (@as_f64 $name:ident $tokio_metrics:ident) => {
+        ($tokio_metrics.$name) as f64
+    };
+}
+
+convert_metrics! {
+    TaskMonitorCollector: [
```
halfprice, line 266: Question: (this is due to my unfamiliarity of tokio monitor) by reading the docs of these metrics, these are all seem to be metrics about `one` tokio runtime. Are we still able to interpret the result if we create multiple monitors per runtime?

halfprice (commented):
Thanks @jpcsmith for adding the monitor and sorry for the delayed review. Overall, I think it is going to be very powerful! I have some clarification questions and mostly about the overhead of a monitor and the scope of a monitor.

crates/walrus-service/src/node/blob_sync.rs
jpcsmith: Could you rephrase please? I do not understand.

crates/walrus-service/src/common/telemetry.rs
jpcsmith: I do not know yet, but this was the suggested usage in the tokio-metrics docs. We'll see in PTN.

crates/walrus-service/src/common/telemetry.rs
jpcsmith: This was due to the fact that the struct needs to be cloned for every request. With the `Arc<inner>` approach, a clone is 8 bytes. Without, it scales with the number of metrics in the inner struct (8+ bytes per metric). I therefore added it since I added quite a number of metrics in this PR and assumed it may grow again in the future.

For the TaskMonitorFamily, I can probably remove it since it seems that the registry already uses this pattern and the total size is 16-bytes. HttpServerMetrics is already ~40 bytes and likely to grow, and together with the TaskMonitorFamily would be 48 bytes at this point.

crates/walrus-utils/src/metrics/tokio.rs
jpcsmith: Answered above.

crates/walrus-utils/src/metrics/tokio.rs
jpcsmith: These aren't the metrics about the runtime, but instead about the tasks that they were attached to. The overall metrics about the runtime are in the tokio crate itself and are currently unstable so I did not add them. For these, each collector has a label "task" attached that is exported and identifies it unique among all other collectors using the same registry (added in `TaskMonitorCollector::new()`).

crates/walrus-service/src/node/blob_sync.rs
halfprice: My question is more about (excuse my insufficient knowledge of tokio minitor), the `blob recovery` is a task spinned off from the event loop, and we already add a monitor to the event loop, why is this additional monitor needed?

crates/walrus-utils/src/metrics/tokio.rs
halfprice: Ok, look forward to reading the result!

halfprice (approved):
🚀🚀🚀🚀

crates/walrus-service/src/node/blob_sync.rs
jpcsmith: Because we need a monitor on the tokio poll-path for the task. In the case of the event loop, we attach the monitor within the task on the future for a single event, and so the monitor will be polled as follows:
`ELoopTask -> process_events -> monitor -> process_single_event`

Since we spawn the task for blob recovery, tokio's polling of that task looks like:
`BlobRecoveryTask -> run_blob_recovery`
which doesn't encounter the monitor that we added in the event loop task. Hence the need to add a monitor there.

PR #1805, opened 2025-03-24
feat: added latest_checkpoint_sequence_number.

liquid-helium:
## Description

Added latest_checkpoint_sequence_number in health endpoint

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI:
   Added `Latest checkpoint sequence number` and `Estimated checkpoint lag` in health output.

### DIFF
--- a/crates/walrus-service/src/node/events/event_processor.rs
+++ b/crates/walrus-service/src/node/events/event_processor.rs
@@ -1106,14 +1106,12 @@ impl EventProcessor {
 
     /// Updates the cached checkpoint sequence number.
     fn update_checkpoint_cache(&self, sequence_number: u64) {
-        // Use fetch_max to atomically update the value if sequence_number is greater
         self.latest_checkpoint_seq_cache
             .fetch_max(sequence_number, Ordering::SeqCst);
     }
 
     /// Gets the latest checkpoint sequence number, preferring the cache.
     pub fn get_latest_checkpoint_sequence_number(&self) -> Option<u64> {
-        // Try to get from cache first.
         Some(self.latest_checkpoint_seq_cache.load(Ordering::SeqCst))
     }
 }


crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -1084,6 +1099,36 @@ impl EventProcessor {
...
+
+        // Update cache if a value was found.
+        if let Some(num) = seq_num {
+            self.latest_checkpoint_seq_cache
+                .store(num, Ordering::SeqCst);
+        }
+
+        seq_num
+    }
 }
```
sadhansood: I didn't follow why we need to update the cache in this getter function as it seems like we update checkpoint while processing events as well. Maybe add some documentation to help readers like me?

sadhansood (commented):
Thanks @liquid-helium! I just have one question in the PR, after which it looks good to merge.

crates/walrus-service/src/node/events/event_processor.rs
liquid-helium: Nice observation! Comments added.
I think in the happy path, this should never happen, we should always be able to get a sequence number form the cache.

I was thinking when some bug exists and we didn't update the cache for some reason then we should back populate the cache.

I will keep it as, but let me know if you feel this is redundant, I will remove it.

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -706,6 +718,9 @@ impl EventProcessor {
                 .stores
                 .checkpoint_store
                 .insert(&(), verified_checkpoint.serializable_ref())?;
+
+            // Also update the cache with the bootstrap checkpoint sequence number.
+            event_processor.update_checkpoint_cache(*verified_checkpoint.sequence_number());
```
sadhansood, line 722: We also need to update the checkpoint in the `catchup_using_event_blobs` path where we are updating events using event blobs.

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -477,6 +482,10 @@ impl EventProcessor {
                 .inc();
             let verified_checkpoint =
                 self.verify_checkpoint(&checkpoint, prev_verified_checkpoint)?;
+
+            // Update the checkpoint cache immediately after verification.
+            self.update_checkpoint_cache(*verified_checkpoint.sequence_number());
+
```
sadhansood: I think this should be updated after the events are written into the db

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -1086,6 +1101,36 @@ impl EventProcessor {
...
+        let seq_num = match self.stores.checkpoint_store.get(&()) {
+            Ok(Some(checkpoint)) => Some(*checkpoint.inner().sequence_number()),
+            _ => None,
+        };
+
+        // Update the cache if we had to read from the store.
+        if let Some(num) = seq_num {
+            self.latest_checkpoint_seq_cache
+                .store(num, Ordering::SeqCst);
+        }
```
sadhansood: I am not sure why we are doing the get and update of the cache here. It seems like cache is always kept up to date in the download loop. Do we also need to update `latest_checkpoint_seq_cache` here?

crates/walrus-service/storage_openapi.yaml
```diff
@@ -1063,6 +1070,13 @@ components:
                 description: The number of events that have been persisted.
                 minimum: 0
           description: The event progress of the storage node.
+        latestCheckpointSequenceNumber:
+          type:
+          - integer
+          - 'null'
+          format: int64
+          description: The latest checkpoint sequence number processed by the node.
```
sadhansood: This should be "Sequence number of the latest checkpoint downloaded by the node."

crates/walrus-sdk/src/api.rs
```diff
@@ -238,6 +238,9 @@ pub struct ServiceHealthInfo {
     /// The status of the shards for which the node is responsible.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub shard_detail: Option<ShardStatusDetail>,
+    /// The latest checkpoint sequence number processed by the node.
```
sadhansood: The latest checkpoint sequence number downloaded by the node.

crates/walrus-service/storage_openapi.yaml
liquid-helium: Done, thanks for the catch!

crates/walrus-sdk/src/api.rs
liquid-helium: Fixed.

crates/walrus-service/src/node/events/event_processor.rs
liquid-helium: Thanks for catching that, fixed.

crates/walrus-service/src/node/events/event_processor.rs
liquid-helium: Removed, return directly from the cache now.

crates/walrus-service/src/node/events/event_processor.rs
liquid-helium: Done.

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -1086,6 +1103,19 @@ impl EventProcessor {
...
+    /// Updates the cached checkpoint sequence number.
+    fn update_checkpoint_cache(&self, sequence_number: u64) {
+        // Use fetch_max to atomically update the value if sequence_number is greater
+        self.latest_checkpoint_seq_cache
+            .fetch_max(sequence_number, Ordering::SeqCst);
+    }
+
+    /// Gets the latest checkpoint sequence number, preferring the cache.
+    pub fn get_latest_checkpoint_sequence_number(&self) -> Option<u64> {
+        // Try to get from cache first.
```
sadhansood: I think the comment is out of date now or it could just be removed

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -1086,6 +1103,19 @@ impl EventProcessor {
...
     }
+
+    /// Updates the cached checkpoint sequence number.
+    fn update_checkpoint_cache(&self, sequence_number: u64) {
+        // Use fetch_max to atomically update the value if sequence_number is greater
+        self.latest_checkpoint_seq_cache
+            .fetch_max(sequence_number, Ordering::SeqCst);
+    }
+
+    /// Gets the latest checkpoint sequence number, preferring the cache.
```
sadhansood, line 1113: This comment seems out of date or maybe just say:  `Gets the latest checkpoint sequence number from the cache`

sadhansood (approved):
Thanks @liquid-helium! Just a couple minor comments, but lgtm!

crates/walrus-service/src/node/events/event_processor.rs
liquid-helium: Fixed!

crates/walrus-service/src/node/events/event_processor.rs
liquid-helium: Comments removed!

PR #1811, opened 2025-03-25
feat: Modify orchestrator to deploy clients

asonnino:
## Description

Modifies the orchestrator to deploy stress clients as opposed to create storage nodes.
This PR also introduce 3 timers to get a finer grain analysis of write latency.

## Test plan

Nothing

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


Cargo.toml
```diff
@@ -3,6 +3,7 @@ default-members = [
   "crates/checkpoint-downloader",
   "crates/walrus-core",
   "crates/walrus-e2e-tests",
+  "crates/walrus-orchestrator",
```
halfprice: Do we need this to be the default member? By removing this, simtests should work again.

asonnino: I don't think we do indeed. Happy to remove it. Why does it make fail the simtests though (out of curiosity)?

halfprice: I replied some details here: https://mysten-labs.slack.com/archives/C079UUZQSM6/p1743487502346859?thread_ts=1742920188.842889&cid=C079UUZQSM6

TLDR is that the aws crate used in the `walrus-orchestrator` added a feature that requires explicit crypto provider to be set.

asonnino: mmh it seems it didn't work

halfprice: The current code doesn't seem to compile. The error in CI is a different error. Or are you still seeing the same `no process-level CryptoProvider available` locally?

asonnino: very strange, checking in a bit

asonnino: It is indeed complaining about something deep inside sui-core; it is a mystery how such a simple PR can trigger so many problems

.gitignore
```diff
@@ -36,3 +36,11 @@ working_dir/
...
 venv/
 **/tmp
+
+# Deployment assets
+crates/walrus-orchestrator/assets/sui-wallet*
+crates/walrus-orchestrator/assets/client_config.yaml
+crates/walrus-orchestrator/assets/client_parameters.yaml
+crates/walrus-orchestrator/assets/settings.yaml
+crates/walrus-orchestrator/assets/*.keystore
+crates/walrus-orchestrator/assets/*.aliases
```
mlegner: Nit: The following should also do the trick:
```suggestion
crates/walrus-orchestrator/assets/
!crates/walrus-orchestrator/assets/dashboard-template.json
```

mlegner (commented):
Thanks for the cleanup+improvements, @asonnino! This looks good to me, although I haven't looked at the orchestrator changes in much detail.

@giac-mysten Could you take a brief look at the changes to the client metrics as you set them up initially? 🙏

.gitignore
asonnino: Good idea, I guess we also need [settings-template.yaml](https://github.com/MystenLabs/walrus/blob/main/crates/walrus-orchestrator/assets/settings-template.yaml)

asonnino:
```suggestion
crates/walrus-orchestrator/assets/
!crates/walrus-orchestrator/assets/dashboard-template.json
!crates/walrus-orchestrator/assets/settings-template.yaml
```

crates/walrus-service/src/client/metrics.rs
```diff
@@ -95,6 +107,41 @@ impl ClientMetrics {
                 registry,
             )
             .expect("this is a valid metrics registration"),
+            encoding_latency_s: register_histogram_with_registry!(
+                "encoding_latency_s",
+                "Time to encode a blob",
+                LATENCY_SEC_BUCKETS.to_vec(),
```
giac-mysten: Important: These buckets are probably too coarse to record the time to encode a blob. the smallest bucket is `1s`. Maybe count `ms`? or change the buckets?

crates/walrus-service/src/client/metrics.rs
```diff
@@ -95,6 +107,41 @@ impl ClientMetrics {
...
+                registry,
+            )
+            .expect("this is a valid metrics registration"),
+            checking_blob_status_latency_s: register_histogram_with_registry!(
+                "checking_blob_status_latency_s",
+                "Time to check the status of a blob",
+                LATENCY_SEC_BUCKETS.to_vec(),
```
giac-mysten: Remark: may be a bit too coarse for this as well? as in most cases this will take between 0 and 2 secs in my experience

crates/walrus-service/src/client.rs
```diff
@@ -737,6 +747,7 @@ impl Client<SuiContractClient> {
         store_when: StoreWhen,
         persistence: BlobPersistence,
         post_store: PostStoreAction,
+        metrics: Option<&Arc<ClientMetrics>>,
```
giac-mysten, line 749: Question: would it make sense to store this in the `Client` itself, instead of passing it to every call?

crates/walrus-stress/src/generator/write_client.rs
```diff
@@ -47,10 +48,12 @@ use super::blob::{BlobData, WriteBlobConfig};
 pub(crate) struct WriteClient {
```
giac-mysten, line 47: Nit: (two lines above) double docstring.

crates/walrus-service/src/client/daemon.rs
```diff
@@ -118,6 +118,7 @@ impl WalrusWriteClient for Client<SuiContractClient> {
                 store_when,
                 persistence,
                 post_store,
+                None,
```
giac-mysten, line 121: Remark: the `ClientMultiplexer` already has an instance of the metrics itself. Now or in a followup, it would be good to ensure that this is passed down. 
(or alternatively that the metrics pointer is stored in the client)

giac-mysten (approved):
Thanks @asonnino ! A couple of comments on the PR, which can be fixed now or in followups.

crates/walrus-service/src/client/metrics.rs
asonnino: ha that is a very good point!

crates/walrus-service/src/client.rs
asonnino: That was my first though but then realised it would create a fairly large PR, and given that this one is already quite big though it be be better done later

crates/walrus-service/src/client/daemon.rs
asonnino: same this was a trick to try to keep this PR as small as possible

PR #1814, opened 2025-03-25
feat: add some initial database metrics

jpcsmith:
## Description

Adds metrics for some common read/write paths in shard and metadata storage.

## Test plan

Local testbed to observe metrics. Will be vetted on PTN with dashboards.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-utils/src/metrics.rs
+++ b/crates/walrus-utils/src/metrics.rs
@@ -67,6 +67,19 @@ macro_rules! define_metric_set {
             /// The namespace in which the metrics reside.
             pub const NAMESPACE: &'static str = $namespace;
 
+            /// Create a unique ID for this instance of the defined metric set, which will be used
+            /// to separate multiple instantiations of the metric, so that they can be registered
+            /// together.
+            fn metric_set_instance_id() -> String {
+                static TYPE_LOCAL_ID: std::sync::atomic::AtomicUsize =
+                    std::sync::atomic::AtomicUsize::new(0);
+                format!(
+                    "{:?}::{}",
+                    std::any::TypeId::of::<Self>(),
+                    TYPE_LOCAL_ID.fetch_add(1, std::sync::atomic::Ordering::Relaxed)
+                )
+            }
+
             /// Creates a new instance of the metric set.
             ///
             /// The instance is created with a const-label of `metric_set_instance_id` set to a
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -319,9 +399,23 @@ impl ShardStorage {
         &self,
         blob_id: &BlobId,
     ) -> Result<Option<PrimarySliver>, TypedStoreError> {
-        self.primary_slivers
+        let start = Instant::now();
+        let labels = Labels {
+            collection_name: &self.cf_names.primary_slivers,
+            operation_name: OperationType::Get,
+            query_summary: "GET sliver BY blob_id",
```
halfprice: question: do we want to differentiate primary and secondary here?

crates/walrus-service/src/node/storage/metrics.rs
```diff
@@ -0,0 +1,171 @@
...
+
+/// Labels that are commonly attached to metrics for database operations.
+#[derive(Debug, Default, Copy, Clone)]
+pub(super) struct Labels<'a> {
+    /// The name of the collection (e.g., RocksDB column family).
+    pub collection_name: &'a str,
+    /// Type of the operation.
+    pub operation_name: OperationType,
+    /// A low-cardinality summary of the query (e.g., "GET sliver BY blob_id").
+    pub query_summary: &'a str,
```
halfprice: suggestion: after reading all the metrics, I feel collection name + operation name currently already uniquely represents the query summary. Therefore, I don't feel we need it at the moment. I'm worried that if someone put some random string here in the future, the cardinality is going to get huge.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -368,18 +464,43 @@ impl ShardStorage {
         blob_id: &BlobId,
         type_: SliverType,
     ) -> Result<bool, TypedStoreError> {
-        match type_ {
+        let start = Instant::now();
+        let labels = Labels {
+            collection_name: &self.cf_names.secondary_slivers,
```
halfprice: remark: collection name matches the sliver type?

crates/walrus-service/src/node/storage.rs
```diff
@@ -432,11 +462,23 @@ impl Storage {
         blob_id: &BlobId,
         metadata: &BlobMetadata,
     ) -> Result<(), TypedStoreError> {
+        let start = Instant::now();
+        let labels = Labels {
+            collection_name: metadata_cf_name(),
+            operation_name: OperationType::Insert,
```
halfprice, line 468: note: you may be already aware that for basic operation like insert/put/read a column family, DBMap already records a ton: https://github.com/MystenLabs/sui/blob/2c0cea6aecae6ba3210287ab77b7d8d05fc2f817/crates/typed-store/src/metrics.rs#L963

So I think some of the metrics added in this PR are duplicated, at least for the success case. But I do appreciate the (relatively) detailed errors being recorded for metrics added in this PR.

crates/walrus-service/src/node/storage/shard.rs
jpcsmith: I went back-and-forth on this, but am now switching it back to `GET primary_sliver BY blob_id` as it identifies the operations irrespective of how we organise the column families.

crates/walrus-service/src/node/storage/metrics.rs
jpcsmith: I changed the `&'a str` to `&'static str` to force the use of a string defined at compile time, which should discourage someone using `format!()` to construct the summary.

The query summary is easier to group the metrics by as it's a lower cardinality than the collection name, which may contain things like the shard ID.

crates/walrus-service/src/node/storage/shard.rs
jpcsmith: Yes, thank you for catching this. I have fixed it.

crates/walrus-service/src/node/storage.rs
jpcsmith: Yea, I knew but forgot, in which case I probably wouldn't have created this PR yet, as this aims to provides relatively database-agnostic metrics. We'll see if it survives the later generalising of the DB layer.

PR #1818, opened 2025-03-26
ci: enable automatic updates for {bin,docs}.wal.app

mlegner:
## Description

This updates and enables our CI jobs to automatically update the `docs.wal.app` and `bin.wal.app` sites. The relevant [variables](https://github.com/MystenLabs/walrus/settings/variables/actions) were added.

Closes WAL-698.

## Test plan

Run the workflows.

### DIFF
--- a/.github/workflows/update-ws-binaries.yaml
+++ b/.github/workflows/update-ws-binaries.yaml
@@ -24,6 +24,9 @@ jobs:
     env:
       # Colors don't seem to work properly with the multiline commands.
       NO_COLOR: 1
+      RUST_LOG: info
+      EPOCHS: 10
+      BUILD_DIR: site
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
       - uses: ./.github/actions/set-up-walrus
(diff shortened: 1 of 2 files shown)


.github/workflows/update-ws-binaries.yaml
```diff
@@ -34,20 +37,25 @@ jobs:
...
           name=walrus-$env-latest-$arch
-          curl https://storage.googleapis.com/mysten-walrus-binaries/$name -o site/$name
+          curl https://storage.googleapis.com/mysten-walrus-binaries/$name -o ${{ env.BUILD_DIR }}/$name
           done
           done
 
+      # TODO: Remove this step once the site-builder supports supports storing multiple blobs
+      # simultaneously.
+      - name: Store blobs on Walrus
+        run: walrus store ${{ env.BUILD_DIR }}/* --epochs ${{ env.EPOCHS }}
```
vlegakis, line 53: Could a `site-builder publish` be used here instead? I suppose it is very similar though so it should be good.

mlegner: The benefit of this separate command is that it uses the "multistore" feature of the Walrus CLI, which is not yet used by the site-builder. This just makes it a bit more efficient.

PR #1828, opened 2025-03-27
refactor: Refactored blob store operation

liquid-helium:
## Description

Related to  WAL-739
Related to WAL-589

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/metrics.rs
+++ b/crates/walrus-service/src/client/metrics.rs
@@ -22,7 +22,7 @@ const LATENCY_SEC_BUCKETS: &[f64] = &[
 ];
 
 const LATENCY_SEC_SMALL_BUCKETS: &[f64] = &[
-    0.01, 0.03, 0.05, 0.07, 1., 1.3, 1.5, 1.7, 2., 2.3, 2.5, 2.7, 3.,
+    0.05, 0.01, 0.03, 0.05, 0.07, 1., 1.3, 1.5, 1.7, 2., 2.3, 2.5, 2.7, 3.,
 ];
 
 // Workload types for the client.


crates/walrus-service/src/client.rs
```diff
@@ -804,232 +1781,208 @@ impl Client<SuiContractClient> {
...
-                    }
-                },
-            ))
-            .await
-            .into_iter()
-            .flatten()
-            .collect();
+        encoded_blobs: Vec<WalrusStoreBlob<'a, T>>,
+    ) -> ClientResult<Vec<WalrusStoreBlob<'a, T>>> {
+        encoded_blobs.iter().for_each(|blob| {
```
wbbradley, line 989:
```suggestion
            #[cfg(debug_assertions)]
            encoded_blobs.iter().for_each(|blob| {
```

liquid-helium: Wow! Nice tip! Thanks.

crates/walrus-service/src/client/communication/node.rs
```diff
@@ -301,6 +301,11 @@ impl NodeWriteCommunication<'_> {
         pairs: impl IntoIterator<Item = &SliverPair>,
         blob_persistence_type: &BlobPersistenceType,
     ) -> NodeResult<SignedStorageConfirmation, StoreError> {
+        tracing::info!(
+            "debug-store: storing metadata and sliver pairs {}, {}",
+            metadata.blob_id(),
+            self.node.node_id
+        );
```
giac-mysten: Nit: use the [recordiing fields](https://docs.rs/tracing/latest/tracing/#recording-fields) syntax

also, what does "debug-store" mean? Is this a leftorver from debugging?

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+///
+/// The typical life cycle of a blob is as follows (symbolically):
+///
+/// - unencoded_blobs = WalrusStoreBlob::from_bytes( .. )
+/// - encoded_blobs = client.encode_blobs( unencoded_blobs )
+/// - blobs_with_statuses = client.get_statuses( encoded_blobs )
+/// - registered_blobs = client.register_blobs( blobs_with_statuses )
+/// - blobs_with_certificates = client.get_certificates( registered_blobs )
+/// - cert_results = client.certify_blobs( blobs_with_certificates )
+/// - final_result = blobs_with_certificates.complete( cert_results )
```
giac-mysten: Praise: this is awesome! 🤩

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+///
+#[derive(Clone)]
+pub enum WalrusStoreBlob<'a, T: Display + Send + Sync> {
```
giac-mysten: Suggestion: just a thought -- since `blob, identifier, span` are always necessary (and should not change?), would it make sense to have this as:

```rust
pub enum WalrusStoreBlob<'a, T: Display + Send + Sync> {
        blob: &'a [u8],
        identifier: T,
        span: Span,
        status: WalrusStoreStatus,
}


pub enum WalrusStoreStatus {
    Unencoded ,
    Encoded {
        pairs: Arc<Vec<SliverPair>>,
        metadata: Arc<VerifiedBlobMetadataWithId>,
    },
    WithStatus {
        pairs: Arc<Vec<SliverPair>>,
        metadata: Arc<VerifiedBlobMetadataWithId>,
        status: BlobStatus,
        span: Span,
    },
    ...
```

not sure this is worth changing :) just curious what your thoughts are.

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+    /// state and operation type.
+    pub fn needs_certify(&self) -> bool {
+        if let WalrusStoreBlob::Registered {
+            operation: StoreOp::RegisterNew { operation, blob },
+            ..
+        } = self
+        {
+            if blob.certified_epoch.is_some() {
+                return false;
+            }
```
giac-mysten: Question: what if it is certified, but for too few epochs? Is this accounted for somewhere else?

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+                | RegisterBlobOp::ReuseAndExtendNonCertified { .. } => true,
+            }
+        } else {
+            false
+        }
+    }
+
+    /// Returns true if the blob needs to be extended based on its current state and operation type.
+    ///
+    /// Note: it returns false if the blob needs to be certified and extended.
```
giac-mysten: Question: why this special case? maybe call the function `is_certified_and_needs_extend`?

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+        {
+            let _enter = self.get_span().enter();
+            tracing::event!(
+                BLOB_SPAN_LEVEL,
+                operation = "with_encode_result",
+                result = ?result
+            );
+        }
+
+        let new_state = match self {
```
giac-mysten: Nit: use `if let` for a match with only one arm? just a preference.

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+                    span.record("blob_id", blob_id.to_string());
+                    WalrusStoreBlob::Encoded {
+                        blob,
+                        identifier,
+                        pairs: Arc::new(pairs),
+                        metadata: Arc::new(metadata),
+                        span: span.clone(),
+                    }
+                }
+                Err(e) => WalrusStoreBlob::Error {
```
giac-mysten: Nit
```suggestion
                Err(error) => WalrusStoreBlob::Error {
```

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+        {
+            let _enter = self.get_span().enter();
+            tracing::event!(
+                BLOB_SPAN_LEVEL,
+                operation = "with_status",
+                status = ?status
+            );
+        }
+
+        let new_state = match self {
```
giac-mysten: Nit: same as above (and below)

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+                Ok(status) => WalrusStoreBlob::WithStatus {
+                    blob,
+                    identifier,
+                    pairs,
+                    metadata,
+                    status,
+                    span: span.clone(),
+                },
+                Err(e) => {
+                    let error_msg = Self::get_error_msg_or_fail_early(e)?;
```
giac-mysten: Remark: cool way to fail early, but we use the error and only have the error message? Could this be an issue when the consumer of the client is not a CLI but some API that needs to progammatically check the error type?

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+            WalrusStoreBlob::WithStatus { span, .. } => span,
+            WalrusStoreBlob::Registered { span, .. } => span,
+            WalrusStoreBlob::WithCertificate { span, .. } => span,
+            WalrusStoreBlob::Completed { span, .. } => span,
+            WalrusStoreBlob::Error { span, .. } => span,
+        }
+    }
+
+    /// Returns true if the given ClientResult should be returned as an error.
+    /// This is used to determine if we should propagate an error or continue processing.
```
giac-mysten:
```suggestion
    /// Returns true if the given ClientResult should be returned as an error.
    ///
    /// This is used to determine if we should propagate an error or continue processing.
```

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+            let _enter = self.get_span().enter();
+            tracing::event!(
+                BLOB_SPAN_LEVEL,
+                operation = "with_error",
+                error = ?error
+            );
+        }
+        let error_msg = Self::get_error_msg_or_fail_early(error)?;
+        let blob_id = self.get_blob_id();
+        let (blob, identifier, span) = match self {
```
giac-mysten: Remark: this could be simplified with the suggestion above.

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
 
-/// The result of encoding as a list of sliver pairs and metadata and a
-/// mapping from blob id to file path.
-#[derive(Debug)]
-pub struct EncodedResult {
-    /// The sliver pairs and metadata.
-    pairs_and_metadata: Vec<(Vec<SliverPair>, VerifiedBlobMetadataWithId)>,
-    /// The mapping from blob ID to path.
-    id_to_path: HashMap<BlobId, PathBuf>,
+/// The log level for all WalrusStoreBlob spans
```
giac-mysten: Suggestion: the following `~1000` lines could be moved to a separate module

crates/walrus-service/src/client.rs
```diff
@@ -518,18 +1477,36 @@ impl Client<SuiContractClient> {
...
+                    post_store,
+                )
+            })
+            .await?;
+
+        #[cfg(debug_assertions)]
+        debug_assert_eq!(results.len(), blobs.len());
+
+        // A trick to make sure the output order is the same as the input order.
+        results.sort_by_key(|blob| blob.get_identifier().to_string());
```
giac-mysten, line 544: Question: where did the input blobs get sorted?

crates/walrus-service/src/client.rs
```diff
@@ -101,14 +105,969 @@ mod multiplexer;
...
+/// - blobs_with_certificates = client.get_certificates( registered_blobs )
+/// - cert_results = client.certify_blobs( blobs_with_certificates )
+/// - final_result = blobs_with_certificates.complete( cert_results )
+///
+#[derive(Clone)]
+pub enum WalrusStoreBlob<'a, T: Display + Send + Sync> {
+    /// Initial phase with raw blob data and identifier.
+    Unencoded {
+        /// The raw blob data to be stored.
+        blob: &'a [u8],
```
giac-mysten: Suggestion: One thing i was thinking to remove the lifetimes and make some things simpler downstream is to move to [Bytes](https://docs.rs/bytes/latest/bytes/) -- may not be needed after this refactor though.

crates/walrus-service/src/client.rs
```diff
@@ -545,17 +1522,19 @@ impl Client<SuiContractClient> {
         persistence: BlobPersistence,
         post_store: PostStoreAction,
     ) -> ClientResult<Vec<BlobStoreResultWithPath>> {
-        let EncodedResult {
-            pairs_and_metadata,
-            id_to_path,
-        } = self
-            .encode_blobs_to_pairs_and_metadata_with_path(blobs_with_paths, encoding_type)
-            .await?;
+        // Not using Path as identifier because it's not unique.
```
giac-mysten, line 565: Remark (for followups): I think we can now deduplicate the `fn` and `fn_with_path` that are lying around.

crates/walrus-service/src/client.rs
```diff
@@ -587,61 +1578,98 @@ impl Client<SuiContractClient> {
...
+            .collect())
     }
 
-    async fn encode_blobs_to_pairs_and_metadata_with_path(
+    /// Encodes multiple blobs into sliver pairs and metadata.
+    ///
+    /// Returns a list of sliver pairs and metadata for each blob.
+    ///
+    /// Failed blobs are filtered out.
+    pub fn encode_blobs_to_pairs_and_metadata(
```
giac-mysten, line 654: Question: what is this function for? why are failures filtered? Seems that it its for tests only? maybe add `#[cfg(test)]`?

crates/walrus-service/src/client/resource.rs
```diff
@@ -190,67 +222,78 @@ impl<'a> ResourceManager<'a> {
...
-            })
-            .map(|(m, _)| *m)
-            .collect::<Vec<_>>();
+                (WalrusStoreBlob::WithStatus { .. }, ..) => {
+                    to_be_processed.push(blob);
+                }
+                _ => {
+                    debug_assert!(blob.is_completed());
+                    results.push(blob);
+                }
```
giac-mysten: Question: shouldn't this be impossible, since we `continue`d if `blob.is_completed()` at line  `240`?

crates/walrus-service/src/client/resource.rs
```diff
@@ -310,6 +353,104 @@ impl<'a> ResourceManager<'a> {
...
+        persistence: BlobPersistence,
+        store_when: StoreWhen,
+    ) -> ClientResult<Vec<WalrusStoreBlob<'b, T>>> {
+        blobs.iter().for_each(|b| {
+            debug_assert!(b.is_with_status());
+        });
+
+        let encoded_lengths: Result<Vec<_>, _> = blobs
+            .iter()
+            .map(|b| {
```
giac-mysten: Nit: 
```suggestion
            .map(|blob| {
```

crates/walrus-service/src/client/resource.rs
```diff
@@ -310,6 +353,104 @@ impl<'a> ResourceManager<'a> {
...
+                            ClientError::other(ClientErrorKind::Other(
+                                anyhow!(
+                                    "the provided metadata is invalid: could not compute the \
+                                    encoded size"
+                                )
+                                .into(),
+                            ))
+                        })
+                    })
+            })
```
giac-mysten: Suggestion: Move this as a function on `WalrusStoreBlob`?

giac-mysten (commented):
Awesome stuff! Made a first pass, and it's looking great!

crates/walrus-service/src/client/resource.rs
liquid-helium: Nice suggestion! Change applied.

crates/walrus-service/src/client/resource.rs
liquid-helium: Change applied, thanks.

crates/walrus-service/src/client/resource.rs
liquid-helium: Nice catch!
Removed.

crates/walrus-service/src/client.rs
liquid-helium: That's a good observation!
Yes, it is test-only, I removed it, and modified the test to use the new API.

There are other similar functions, and I will change them in following PRs, and focus this PR on the core changes.

crates/walrus-service/src/client.rs
liquid-helium: Agreed.

crates/walrus-service/src/client.rs
liquid-helium: Noted.
We can come back and revisit it if it turns out to be necessary.

crates/walrus-service/src/client.rs
liquid-helium: They were assigned an ordered identifier based on their position in the input vector at the start of this function.

crates/walrus-service/src/client.rs
liquid-helium: That's a good idea!
Move it to client_types.rs.

crates/walrus-service/src/client/communication/node.rs
liquid-helium: Thanks for the tip, sorry this is a temporary debugging info, removed.

crates/walrus-service/src/client.rs
liquid-helium: Thank you for your suggestion! I appreciate it. 

Initially, I experimented with both styles: the current 'state machine transition' style and a similar approach as you proposed, which uses a base type with an internal state that updates when the blob is processed. 

I did start with the second approach but ultimately transitioned to the current one because it captures the semantics of the transitions between different phases as a blob entity.

I am open to more discussion, but I will keep it as for now.

crates/walrus-service/src/client.rs
liquid-helium: That's a good observation!
It works here because for a certified blob with a shorter lifetime, it will be classified as `needs_extend`, and will be extended in a later step.

crates/walrus-service/src/client.rs
liquid-helium: Thanks for the catch!
The comment is outdated, removed the misleading part.

crates/walrus-service/src/client.rs
liquid-helium: Change applied, it is cleaner, thanks for the tip.

crates/walrus-service/src/client.rs
liquid-helium: Change applied.

crates/walrus-service/src/client.rs
liquid-helium: Change applied!

crates/walrus-service/src/client.rs
liquid-helium: Change applied, thanks for the suggestion!

crates/walrus-service/src/client.rs
liquid-helium: I like this idea a lot.

changed to propagate ClientError.

Keeping error string in StoreBlobResult for now, returning Error from it would require more refactor due to the various trait bound on that type. 

We can do it in a follow-up if it turns out to be necessary.

crates/walrus-service/src/client.rs
giac-mysten: sounds good to me!

crates/walrus-stress/src/generator/write_client.rs
mlegner, line 1: Question: Why are these changes needed?

crates/walrus-service/src/client/responses.rs
```diff
@@ -158,6 +166,7 @@ impl BlobStoreResult {
                 blob_object: Blob { blob_id, .. },
                 ..
             } => blob_id,
+            Self::Error { blob_id, .. } => blob_id.as_ref().unwrap_or(&BlobId::ZERO),
```
mlegner: Suggestion: Change the function to return an `Option` instead of returning a dummy value.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -132,7 +134,9 @@ where
             BlobPersistence::Permanent,
             PostStoreAction::Keep,
         )
-        .await?;
+        .await;
+
+    let store_result = store_result?;
```
mlegner: Question: Why this change?

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -323,12 +327,13 @@ async fn test_inconsistency(failed_nodes: &[usize]) -> TestResult {
...
         .expect("should register exactly one blob");
 
+    let pair_ref = pairs.iter().collect::<Vec<_>>();
     // Certify blob.
     let certificate = client
         .as_ref()
         .send_blob_data_and_get_certificate(
             &metadata,
-            &pairs,
+            &pair_ref,
```
mlegner: Question: Why this change?

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -548,6 +553,63 @@ async fn store_blob(
         .to_owned())
 }
 
+/// Tests that blobs can be extended when possible.
```
mlegner: Nit: The docstring doesn't seem to fit the test.

crates/walrus-sui/src/client.rs
```diff
@@ -347,6 +367,32 @@ impl PostStoreAction {
...
+#[derive(Debug, Clone)]
+pub enum PostStoreActionResult {
+    /// The blob was burned.
+    Burned,
+    /// The blob was transferred to the given address.
+    TransferredTo(SuiAddress),
+    /// The blob was kept in the wallet that created it.
+    Kept,
+    /// The blob was put into a shared blob object.
+    Shared(Option<ObjectID>),
```
mlegner: Question: Why is this an `Option`?

crates/walrus-sui/src/client.rs
```diff
@@ -1156,7 +1202,7 @@ impl SuiContractClient {
         &self,
         blobs_with_certificates: &[CertifyAndExtendBlobParams<'_>],
         post_store: PostStoreAction,
-    ) -> SuiClientResult<HashMap<BlobId, ObjectID>> {
+    ) -> SuiClientResult<Vec<CertifyAndExtendBlobResult>> {
```
mlegner, line 1241: Question: Why do we no longer need the `BlobId` in the result?

crates/walrus-sui/src/client.rs
```diff
@@ -2645,20 +2693,72 @@ impl SuiContractClientInner {
...
+            )?;
+            ensure!(
+                object_ids.len() == cert_and_extend_params.len(),
+                "unexpected number of shared blob objects created: {} (expected {})",
+                object_ids.len(),
+                cert_and_extend_params.len()
+            );
+
+            // Fetch all SharedBlob objects and collect them as a mapping blob id
+            // to shared blob object id
```
mlegner: Nit: AFAIU the mapping is from blob **object ID**, not blob ID.
```suggestion
            // Fetch all SharedBlob objects and collect them as a mapping from blob object ID
            // to shared blob object ID.
```

crates/walrus-service/src/client.rs
mlegner: Just want to mention that I had the same thought as Giacomo and would have a preference for reducing the duplicated fields. Not blocking IMO though.

crates/walrus-service/src/client.rs
```diff
@@ -545,17 +554,19 @@ impl Client<SuiContractClient> {
         persistence: BlobPersistence,
         post_store: PostStoreAction,
     ) -> ClientResult<Vec<BlobStoreResultWithPath>> {
-        let EncodedResult {
-            pairs_and_metadata,
-            id_to_path,
-        } = self
-            .encode_blobs_to_pairs_and_metadata_with_path(blobs_with_paths, encoding_type)
-            .await?;
+        // Not using Path as identifier because it's not unique.
```
mlegner, line 565: Remark: I just wonder if it ever makes sense to pass the same path multiple times... 🤔

crates/walrus-service/src/client.rs
```diff
@@ -518,18 +509,36 @@ impl Client<SuiContractClient> {
         persistence: BlobPersistence,
         post_store: PostStoreAction,
     ) -> ClientResult<Vec<BlobStoreResult>> {
-        let pairs_and_metadata = self.encode_blobs_to_pairs_and_metadata(blobs, encoding_type)?;
+        let blobs_with_identifiers = blobs
+            .iter()
+            .enumerate()
+            .map(|(i, blob)| WalrusStoreBlob::new_unencoded(blob, format!("blob_{:06}", i)))
+            .collect::<Vec<_>>();
```
mlegner: Suggestion: Factor this out into a function as this is used 4 times in this file.

crates/walrus-service/src/client.rs
```diff
@@ -564,14 +575,26 @@ impl Client<SuiContractClient> {
...
+            results.push(BlobStoreResultWithPath {
+                path: path.clone(),
+                blob_store_result: store_result.clone(),
+            });
+        }
+
+        Ok(results)
     }
 
     /// Encodes the blob, reserves & registers the space on chain, and stores the slivers to the
```
mlegner, line 610: Remark (not related to this PR): AFAICT, the `reserve_and_store_blobs` function is only used for tests. Could we mark it accordingly or replace its use by `reserve_and_store_blobs_retry_committees`?

crates/walrus-service/src/client.rs
```diff
@@ -775,45 +821,49 @@ impl Client<SuiContractClient> {
...
-            store_operations.iter().map(|op| format!("{:?}", op)).collect::<Vec<_>>().join("\n")
+            registered_blobs.len(),
+            registered_blobs
+                .iter()
+                .map(|blob| format!("{:?}", blob.get_operation()))
+                .collect::<Vec<_>>()
+                .join("\n")
+        );
+        let num_registered_blobs = registered_blobs.len();
+        #[cfg(debug_assertions)]
```
mlegner: Question: Why do we need this attribute? Isn't this implied by the `debug_assert_eq` itself? (Also in other places.)

crates/walrus-service/src/client/client_types.rs
```diff
@@ -0,0 +1,1019 @@
...
+/// The typical life cycle of a blob is as follows (symbolically):
+///
+/// - unencoded_blobs = WalrusStoreBlob::from_bytes( .. )
+/// - encoded_blobs = client.encode_blobs( unencoded_blobs )
+/// - blobs_with_statuses = client.get_statuses( encoded_blobs )
+/// - registered_blobs = client.register_blobs( blobs_with_statuses )
+/// - blobs_with_certificates = client.get_certificates( registered_blobs )
+/// - cert_results = client.certify_blobs( blobs_with_certificates )
+/// - final_result = blobs_with_certificates.complete( cert_results )
+pub enum WalrusStoreBlob<'a, T: Display + Clone + Send + Sync> {
```
mlegner: Praise: I really like the approach of encoding the blob lifecycle in some form of state machine.

I see one issue with the current approach though: We don't really leverage the type system for that, because the blob always has the same type. A result of this are some of the `panic`s that are needed throughout the code. 

We could instead use a form of [type-state pattern](https://cliffle.com/blog/rust-typestate/) here. Essentially, this would mean defining separate structs `UnencodedBlob`, `EncodedBlob`, ..., and then have

```
enum WalrusStoreBlob {
    Unencoded(UnencodedBlob),
    Encoded(EncodedBlob),
    [...]
}
```

Then we could pass around the `WalrusStoreBlob` if we don't know or care about the actual state, but use the structs directly to emphasize that a blob has some particular state.

Additionally, we could define a trait that is implemented for all the variants and use `enum_dispatch` to automatically implement it for the enum.

mlegner (commented):
Thank you so much for this much-needed refactoring! 🙏 ❤️ 

I have a few smaller questions and comments, and one larger one about the way the blob state is encoded. Maybe we could discuss that topic offline?

crates/walrus-service/src/client.rs
liquid-helium: Nice catch! Fixed.

crates/walrus-service/src/client.rs
liquid-helium: Sure!
I will do it in following PRs to limit the changes in this one, and likely will also try to unify as many code as possible in the near future.

crates/walrus-service/src/client.rs
liquid-helium: Good idea, done.

crates/walrus-service/src/client.rs
liquid-helium: From the perspective of our API coverage, we are capable of handling this scenario since we never stated it was disallowed.

I am open to changing the behavior, but I am inclined to maintain our ability to handle these situations, and implementation-wise, there is no difference.

crates/walrus-sui/src/client.rs
liquid-helium: The reason I put an option there is a compromise, I wanted to put an Error type there but it won't satisfy the trait bound.

The reason is this CertifyAndExtendResult is constructed in a separate read *after* we have already completed the certify_and_extend transaction.
If the read fails, and we forward that error directly back to the client, it would be confusing to a user's application whether the blobs are stored successfully or not.

The intent is to deliver an error saying that `the blobs are stored successfully, but the results are not retrieved`.

Ideally, if the same transaction can return the corresponding ObjectID, then we don't need to worry about this.

crates/walrus-service/src/client.rs
liquid-helium: When refactoring the code, it achieved similar 'reducing', I put the common fields in UnencodedBlob, and use it as a field for other variants.

crates/walrus-e2e-tests/tests/test_client.rs
liquid-helium: used for test log, reverted.

crates/walrus-e2e-tests/tests/test_client.rs
liquid-helium: Reverted, thanks for the catch.

crates/walrus-e2e-tests/tests/test_client.rs
liquid-helium: opps, I forgot to change that after copying the test body, fixed.

crates/walrus-sui/src/client.rs
liquid-helium: The object_id obtained after registering/reusing each blob is unique, and the caller has both the object_id and blob_id, so the object_id will be sufficient for the caller to map all the blobs with their results.

crates/walrus-sui/src/client.rs
liquid-helium: Thanks for the correction, fixed.

crates/walrus-service/src/client/responses.rs
liquid-helium: Done.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -578,7 +578,10 @@ impl ClientCommandRunner {
                 .collect::<Vec<_>>();
             tracing::warn!(
                 "some blobs ({}) are not stored",
-                not_stored.into_iter().join(", ")
+                not_stored
+                    .into_iter()
+                    .map(|blob_id| blob_id.map_or("N/A".to_string(), |id| id.to_string()))
```
mlegner: Nit: Can we include the path or something else to identify these (especially relevant for the `N/A` cases).

crates/walrus-service/src/client/resource.rs
```diff
@@ -310,6 +335,89 @@ impl<'a> ResourceManager<'a> {
...
+                    // If vec is now empty, remove the entry from the map
+                    if entries.is_empty() {
+                        blob_id_map.remove(&blob_id);
+                    }
+
+                    blob.with_register_result(Ok(StoreOp::new(operation, blob_obj)))
+                        .expect("should succeed on a Ok result")
+                } else {
+                    panic!("missing blob ID: {}", blob_id);
+                }
```
mlegner, line 417: Suggestion: Can you add TODOs and a follow-up issue to get rid of these (and other) panics as part of a re-refactor of the state machine (as discussed offline)?

crates/walrus-service/src/client/client_types.rs
```diff
@@ -0,0 +1,1808 @@
...
+    WithStatus(BlobWithStatus<'a, T>),
+    /// After registration, includes the store operation.
+    Registered(RegisteredBlob<'a, T>),
+    /// After certificate, includes the certificate.
+    WithCertificate(BlobWithCertificate<'a, T>),
+    /// Final phase with the complete result.
+    Completed(CompletedBlob<'a, T>),
+    /// Error occurred during the blob lifecycle.
+    Error(FailedBlob<'a, T>),
+}
```
mlegner: As discussed offline, I would suggest to create an enum that can either be one of the concrete states or completed or an error:
```rust
pub enum WalrusStoreBlobState<'a, S: WalrusBlobStateApi<'a, T>, T: Debug + Clone + Send + Sync> {
    /// Blob is in a concrete state.
    BlobInState(S),
    /// Final phase with the complete result.
    Completed(CompletedBlob<'a, T>),
    /// Error occurred during the blob lifecycle.
    Error(FailedBlob<'a, T>),
}
```

Then we can implement transition functions on the appropriate concrete states and the corresponding `WalrusStoreBlobState` enum (where they are just no-ops on the `Completed` and `Error` variants.

As a result, we can be sure that blobs are in the correct state to be able to call transition functions. This also allows us to remove all the functions returning options from the `WalrusStoreBlobApi`.

Can you add a TODO comment and follow-up issue for this?

mlegner (approved):
Thanks for this refactoring, @liquid-helium. Given the amount of code changes, I'm not 100% sure I covered everything. Did you run some manual tests on PTN or Testnet or a local testbed to confirm that the behavior is unchanged (or improved)? If not, could you do that before merging this? Also, the CLI testing you are working on will be quite handy for PRs like this. 😅

crates/walrus-service/src/client/client_types.rs
liquid-helium: Thanks @mlegner for the neat solution.
TODO added.

crates/walrus-service/src/client/resource.rs
liquid-helium: Added WAL-754.

crates/walrus-service/src/client/cli/runner.rs
liquid-helium: Nice catch, fix applied.

crates/walrus-stress/src/generator/write_client.rs
liquid-helium: reverted.

PR #1829, opened 2025-03-27
feat: Allow updating logging verbosity dynamically at runtime

sadhansood:
## Description

This PR adds code for updating log directives to control logging verbosity dynamically at runtime through unix socket endpoint.

## Test plan

sudo /opt/walrus/bin/walrus-node local-admin --socket-path /opt/walrus/admin.socket log-level --level "info"
Log level updated successfully to: info

---

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -38,7 +38,7 @@ use walrus_core::{
 use walrus_service::{
     DbCheckpointManager,
     SyncNodeConfigError,
-    common::config::SuiConfig,
+    common::{config::SuiConfig, telemetry::WalrusTracingHandle},
     node::{
         ConfigLoader,
         StorageNode,
@@ -84,6 +84,8 @@ struct AdminArgs {
     checkpoint_manager: Option<Arc<DbCheckpointManager>>,
     /// Admin socket path.
     admin_socket_path: Option<PathBuf>,
+    /// Tracing handle.
+    tracing_handle: WalrusTracingHandle,
 }
 
 #[derive(Subcommand, Debug, Clone)]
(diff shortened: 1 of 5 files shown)


mlegner (commented):
@sadhansood Sorry again for my original unqualified review. As already mentioned at #2020, we should ideally have a unified interface through which to perform run-time operations, either through a UNIX socket or through a separate port that is only available locally.

cc @liquid-helium

crates/walrus-service/bin/node.rs
```diff
@@ -1433,6 +1442,20 @@ impl StorageNodeRuntime {
     }
 }
 
+/// Handle log level commands from admin socket.
+async fn handle_log_level_command(level: String, args: &AdminArgs) -> AdminCommandResponse {
+    match args.tracing_handle.0.update_log(level.as_str()) {
```
liquid-helium: nit: can we use `*args.tracing_handle`, since it implements deref already?

crates/walrus-service/bin/node.rs
```diff
@@ -189,6 +191,12 @@ enum AdminCommands {
         #[command(subcommand)]
         command: CheckpointCommands,
     },
+    /// Log level management.
```
liquid-helium, line 194: Does this support more complex rules like, if so, can we be explicit about it?
update_log("warn,walrus_service::client::daemon=trace")

Cargo.lock
mlegner, line 1: Question: Are the changes to the lock file related to this PR? Can they be reverted?

crates/walrus-service/bin/node.rs
mlegner: Good idea. Maybe add a few examples or reference to what is possible with `RUST_LOG`?

mlegner: Also, can you add this to the release notes?

sadhansood: Yes, @liquid-helium it supports complex directives like the ones you mentioned

sadhansood:
> Thanks a lot for adding this, @sadhansood. This will be very helpful.

Thank you @mlegner !

PR #1831, opened 2025-03-27
docs: update list of aggregators and publishers

mlegner:
## Description

- Add missing instances.
- Update some URLs.
- Sort lists.
- Remove instances that are not working.

## Test plan

Inspect preview.


mlegner:
@VLegakis Thanks for the insights. 

> Also, if possible it would be good to add some metadata for them, especially the cache: true/false flag which enables them to be monitored for caching by our Nodes Monitoring !

Could we do it the other way round? We check all for a cache setup and mark those that have a cache in the list?

VLegakis:
> @VLegakis Thanks for the insights.
> 
> > Also, if possible it would be good to add some metadata for them, especially the cache: true/false flag which enables them to be monitored for caching by our Nodes Monitoring !
> 
> Could we do it the other way round? We check all for a cache setup and mark those that have a cache in the list?

@mlegner currently our monitoring tool gets these flags from the list and tests them for caching (or not (if false or undefined). So, we cannot currently specify if an operator supports caching or not, other than testing their response times and observing whether they are very fast (so we assume caching is supported) or slow.
@nikos-terzo what do you think?

nikos-terzo:
I think the only ways to programmatically infer whether a response was cached or not, would be to define a response time threshold, or a response time speed-up threshold, and assume that all responses under that threshold were cached. Correct? If so, I think, we would need to have the geolocation of the node, in order for this threshold to somewhat apply. So the question would change from whether a node has a cache setup to where the node lives. Anyway, the "assuming" nature of this path, makes me want to avoid such inferences, and have these metadata set explicitly. No?

mlegner:
@nikos-terzo I think we could request the same blob twice within a relatively short time period. If the second request is significantly faster, there is almost certainly a cache. Given the latency without cache is something like 2s and with cache likely below 500ms, this should be very easy to detect.

PR #1832, opened 2025-03-27
test: split simtest into multiple build target

halfprice:
## Description

The nightly simtest is getting longer and longer. One way to make it faster without digging into optimizing the test,
is to create multiple test binary and run the tests in parallel. In the past two days, for some tests, 30 minutes only
ran 12 seeds. Hope this will make tests run faster and provide better coverage.

The PR itself is a no-op. A follow up PR will address some of the inefficiency in the simtest.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


sadhansood:
Thanks @halfprice , after splitting them into different binaries and running in parallel do we observe any perf improvement? Also, I didn't see any build/cargo related changes for different test binaries. Are you planning to do that in a follow up PR?

halfprice:
> Thanks @halfprice , after splitting them into different binaries and running in parallel do we observe any perf improvement? Also, I didn't see any build/cargo related changes for different test binaries. Are you planning to do that in a follow up PR?

About performance: about to find out with 1pm. I plan to also start multiple seed execution currently, and that will be in a follow up PR.

Regarding cargo/build: each .rs file under tests/ folder will be compiled into a different test binary, so there is no need to change anything in the cargo.toml file. For example, with this RP, these will be the test binary built:

`ls target/simulator/deps/simtest* | grep -v rcgu | grep -v "\.d"`

target/simulator/deps/simtest-72df307a77c477f5
target/simulator/deps/simtest-ab6370ddf0c73d65
target/simulator/deps/simtest-cc3a96224971f0a0
target/simulator/deps/simtest_event_blob-a39ffe27538db7e1
target/simulator/deps/simtest_failure-1370c964071bc054
target/simulator/deps/simtest_param_sync-457133f5e050b132

sadhansood (approved):
Thanks for helping my understanding, I hope the performance improves with this change going forward!

PR #1836, opened 2025-03-27
docs: minor improvements to setup

mlegner:
## Description

- Allow specifying network for install script.
- Tell users to download script and configs from Walrus instead of GitHub.
- Add installation via Cargo.
- Fix links on "getting started" page.
- Enable linkcheck for `*.wal.app`.

## Test plan

- Tested commands manually. Note that some instructions will only work after this is merged and deployed.
- Inspect preview.


wbbradley (approved):
💯

PR #1841, opened 2025-03-28
fix: Adjust number of download workers in case of extended failures

sadhansood:
## Description

This PR enhances the checkpoint lag monitoring system with better error handling, improved logging, and a better recovery mechanism in case of extended failures by:
1. Improving the error handling for checkpoint lag monitoring failures. The system now provides more detailed logging that includes the specific error details, failure counts, and thresholds.
2. Adding a safety mechanism for handling persistent monitoring failures. When the system fails to monitor checkpoint lag for an extended period (10 consecutive failures), it now scales the worker pool to the average of min/max workers. This prevents the system from operating in an unknown state during extended failures and provides a more predictable fallback behavior.

## Test plan

Existing tests
---

### DIFF
--- a/crates/checkpoint-downloader/src/downloader.rs
+++ b/crates/checkpoint-downloader/src/downloader.rs
@@ -197,6 +197,58 @@ impl ParallelCheckpointDownloaderInner {
         Ok(())
     }
 
+    async fn adjust_workers(
+        next_worker_id: &mut usize,
+        current_workers: usize,
+        target_workers: usize,
+        channels: &PoolMonitorChannels,
+        config: &PoolMonitorConfig,
+        worker_count: &Arc<RwLock<usize>>,
+    ) -> Result<()> {
+        match current_workers.cmp(&target_workers) {
+            std::cmp::Ordering::Greater => {
+                let to_remove = current_workers - target_workers;
+                for _ in 0..to_remove {
+                    channels
+                        .message_sender
+                        .send(WorkerMessage::Shutdown)
+                        .await?;
+                    let new_count = {
+                        let mut count = worker_count.write().await;
+                        *count -= 1;
+                        *count
+                    };
+                    config.metrics.num_workers.set(new_count as i64);
+                }
+            }
+            std::cmp::Ordering::Less => {
+                let to_add = target_workers - current_workers;
+                for _ in 0..to_add {
+                    let cloned_client = config.client.clone();
+                    let cloned_receiver = channels.message_receiver.clone();
+                    let cloned_checkpoint_sender = channels.checkpoint_sender.clone();
+                    let cloned_config = config.downloader_config.base_config.clone();
+                    Self::spawn_new_worker(
+                        *next_worker_id,
+                        cloned_client,
+                        cloned_receiver,
+                        cloned_checkpoint_sender,
+                        cloned_config,
+                    );
+                    *next_worker_id += 1;
+                    let new_count = {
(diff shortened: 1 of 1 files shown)


crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -362,18 +436,25 @@ impl ParallelCheckpointDownloaderInner {
...
-            handle_checkpoint_error(err.as_ref(), sequence_number);
-            let err = err
-                .map(|e| e.into())
-                .unwrap_or_else(|| anyhow!("Failed to download checkpoint"));
-            return CheckpointEntry::new(sequence_number, Err(err));
-        };
-        CheckpointEntry {
-            sequence_number,
-            result: Ok(checkpoint),
+        let res = tokio::time::timeout(
```
halfprice, line 448: suggeestion: remove this timeout. I think all the RPCs inside RetriableRpcClient::get_full_checkpoint already contains tokio::time::timoout.

Also, given this also includes fallback path, 1 second timeout feels too short.

crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -197,6 +197,49 @@ impl ParallelCheckpointDownloaderInner {
...
+        worker_count: &Arc<RwLock<usize>>,
+    ) -> Result<()> {
+        match current_workers.cmp(&target_workers) {
+            std::cmp::Ordering::Greater => {
+                let to_remove = current_workers - target_workers;
+                for _ in 0..to_remove {
+                    channels
+                        .message_sender
+                        .send(WorkerMessage::Shutdown)
+                        .await?;
```
halfprice, line 215: question: what if in the middle of removal, there is a failure, and returned here? Seems that worker_count won't be updated in this case.

halfprice (commented):
Thanks @sadhansood for improving the lag monitor! I just some small comments.

crates/checkpoint-downloader/src/downloader.rs
liquid-helium: That's a great point. Taking it a step further, we could make the timeout parameter specific to each function call and pass it in from the appropriate caller, ensuring that we always have a single source of truth.

mlegner:
@sadhansood Given He and Zhe have already reviewed this, I'll remove myself from the reviewers. Feel free to re-request a review though.

sadhansood:
> 💯 Thanks, @sadhansood, for making it more reliable!
> 
> It looks fantastic!
> 
> Just one point for discussion: If the current_checkpoint_lag fails, it's likely due to the RPC endpoints. Does that mean other workers are likely to encounter the same error?
> 
> If the failures are correlated, perhaps we should consider reducing the number of workers in the event of consecutive failures?

If the lag computation fails, it does not always mean that workers will also fail given now we have fallback. With fallback, workers could keep making progress but pool monitor would just stop working. In that scenario, we can not do anything like today but we risk having too few or too many workers. So, I think average number of workers probably is probably a good compromise.

crates/checkpoint-downloader/src/downloader.rs
sadhansood: This timeout is only meant for test actually, to make the tests run quicker (since we know these calls will fail but we want them to fail quick in tests)

crates/checkpoint-downloader/src/downloader.rs
sadhansood: sure, fixed

PR #1842, opened 2025-03-28
fix the little bug in  wal.move

Zhujunnan:
1. Fixed method calls in total_supply and burn to use direct function calls instead of object method syntax
2. Added overflow verification in the init function
3. Corrected all spelling mistakes in the test_burn function
4. Fixed the method calls in the test functions to use the correct syntax
5. Properly called coin::split instead of using object method syntax

## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/wal/sources/wal.move
+++ b/contracts/wal/sources/wal.move
@@ -45,6 +45,8 @@ fun init(otw: WAL, ctx: &mut TxContext) {
     // Mint the total supply of WAL.
     let frost_per_wal = 10u64.pow(DECIMALS);
     let total_supply_to_mint = TOTAL_WAL_SUPPLY_TO_MINT * frost_per_wal;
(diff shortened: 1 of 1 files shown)


damirka:
Thank you for spending time on this PR! However, we have a policy to only accept substantial PRs from external contributors, see [here](https://github.com/MystenLabs/walrus/blob/main/CONTRIBUTING.md#important-note). 

As to the contents of the PR, in case you find this helpful:
- receiver style syntax is the standard way of calling functions, so this change is not adding anything and sets code back to Move pre 2024 edition
- overflow check you suggested is meaningless, if there was overflow, it would've happened during the package publish
- lastly, the WAL package is immutable, so any changes to it can never be applied in real world

PR #1843, opened 2025-03-28
fix: Delete uncertified blobs from filesystem when resetting

sadhansood:
## Description

When we reset uncertified blobs from db, we should also delete the corresponding blob from filesystem which we don't do today. While this is not a problem, it leaves garbage blobs behind which makes things hard to debug later on. This PR also cleans up the file system during reset.

## Test plan

Existing blob tests

### DIFF
--- a/crates/walrus-service/src/node/events/event_blob_writer.rs
+++ b/crates/walrus-service/src/node/events/event_blob_writer.rs
@@ -4,6 +4,7 @@
 //! Event blob writer.
 
 use std::{
+    ffi::OsStr,
     fs::{self, File, OpenOptions},
     io::{self, BufWriter, Seek, SeekFrom, Write},
     path::{Path, PathBuf},
@@ -64,8 +65,13 @@ const ATTESTED: &str = "attested_blob_store";
 const PENDING: &str = "pending_blob_store";
 /// The column family name for failed to attest event blobs.
 const FAILED_TO_ATTEST: &str = "failed_to_attest_blob_store";
+/// The file name for the current blob.
+const CURRENT_BLOB: &str = "current_blob";
+/// The maximum size of a blob file.
 const MAX_BLOB_SIZE: usize = 100 * 1024 * 1024;
+/// The number of checkpoints per blob.
 pub(crate) const NUM_CHECKPOINTS_PER_BLOB: u32 = 216_000;
+/// The default number of unattested blobs threshold.
 const DEFAULT_NUM_UNATTESTED_BLOBS_THRESHOLD: u32 = 3;
 
 pub(crate) fn certified_cf_name() -> &'static str {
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -563,17 +566,38 @@ impl EventBlobWriterFactory {
...
+        for blob_index in blobs_to_delete {
+            let blob_path = blobs_path.join(blob_index.to_string());
+            if blob_path.exists() {
+                if let Err(e) = fs::remove_file(&blob_path) {
+                    tracing::warn!(?blob_path, ?e, "Failed to remove blob file");
+                } else {
+                    tracing::info!(?blob_path, "Removed uncertified blob file");
+                }
+            }
+        }
```
halfprice, line 676: question: what if the node stops during the removal? That may leave the blob file permanently on disk? Should we move the `wb.write()` after the removal, or do something like `ls` then `rm`?

halfprice (commented):
Thanks @sadhansood for the improvement! I just one comment

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: @halfprice I think it is dangerous to remove the blobs first before the db write. If node crashes after blob removal from filesystem but before db write, upon next restart the latest certified event blob could have progressed further to potentially include some of the blobs we are removing from file system (but they exist in db). I think one thing we can do to address your concern is, upon startup just check all blobs in db and filesystem, and remove ones from fs which are not in db.

halfprice: > upon next restart the latest certified event blob could have progressed further to potentially include some of the blobs we are removing from file system

Are you talking about the case where after restart, the node no longer needs to reset the event blob? And the removed event blobs can be potentially useful again?

> upon startup just check all blobs in db and filesystem, and remove ones from fs which are not in db.

Yeah, this sounds good to me too.

sadhansood: > Are you talking about the case where after restart, the node no longer needs to reset the event blob? And the removed event blobs can be potentially useful again?

Yes

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -270,6 +271,66 @@ impl EventBlobWriterFactory {
...
+        failed_to_attest: &DBMap<(), FailedToAttestEventBlobMetadata>,
+        certified: &DBMap<(), CertifiedEventBlobMetadata>,
+    ) -> Result<()> {
+        let blobs = fs::read_dir(blobs_path)?;
+        for blob in blobs {
+            let blob_path = blob?.path();
+            if !blob_path.is_file() {
+                continue;
+            }
+            if blob_path.file_name() == Some(OsStr::new("current_blob")) {
```
halfprice: nit: maybe define all the const blob names (just 1 at the moment) as a `const &str` so that we won't overlook them.

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -270,6 +271,66 @@ impl EventBlobWriterFactory {
...
+                continue;
+            }
+            let event_index = blob_path
+                .file_name()
+                .and_then(|name| name.to_str())
+                .and_then(|name| name.parse::<u64>().ok());
+            if let Some(event_index) = event_index {
+                // Check if blob exists in any database
+                let exists_in_db = pending.safe_iter().any(|result| {
```
halfprice, line 305: Nit: it seems that for each blob, we scan all 4 tables. Maybe more efficient to build the exist index once and then iterate through blobs.

halfprice (approved):
Thanks @sadhansood for the improvements! LGTM with 2 minor nits.

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: good call, i can handle this in a follow up.

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: This might be something not very serious tbh, `pending` table is really small number of blobs and attested/failed_to_attest/certified all have just one entry.

PR #1855, opened 2025-03-28
ci: Switch release type to published

ebmifa:
## Description
Switch release type to published

## Test plan
👀


ebmifa:
Don't need this. There was a problem with the `ruleset` for the `main` branch, since it was renamed from `master` and `testnet`. After I fixed it this action starts automatically.

Issue #1153, opened 2024-11-06
Describe conventions for code, docstrings, logging messages, etc.


mlegner:
We should also clean up existing text to comply with these conventions.

- [ ] Docstrings: Full sentences, start with capital letter, split into heading and details if applicable.
- [ ] Errors/panics/unwraps/expects.
  - Don't use `unwrap` except in tests.
  - `expect` should be used if we know that it should succeed; in this case the string should describe why we expect this; e.g., `lock.lock().expect("mutex should not be poisoned")`.
  - Functions that can panic *must* describe this in the docstring.
- [ ] Logging messages: start with lower-case letter. Other related things see #1212.
- [ ] Variable naming.
- [ ] File names for modules (https://doc.rust-lang.org/reference/items/modules.html#module-source-filenames).

Himess:
Hi @mlegner ,

I've started working on the docstring cleanups mentioned in this issue. Currently, I'm fixing all instances where docstrings begin with a lowercase letter to ensure they follow the convention of starting with a capital letter.

If it's helpful, I can also review and complete any sentence fragments in docstrings that are not full sentences.

Additionally, if you’d like, I can apply comment wrapping using a rustfmt.toml configuration and reformat the relevant code to comply with that style as well.

Please let me know if you'd like me to proceed with all of these updates in a single PR.

Thanks!

mlegner:
Hi @Himess!

Thanks a lot for the offer to work on this, this is highly appreciated. 🙏  

However, please hold off on investing too much time, we're still in the process of getting our process for community contributions organized, and many issues including this one haven't been discussed fully within the team and/or don't contain all the necessary details. I don't want you to invest time in creating PRs that may then require lots of changes or cannot be landed.

I'll take a look at the PRs you already prepared later this week or early next week.

Himess:
Hi @mlegner , 
just a gentle follow-up to check if you had a chance to review the PRs I submitted. No rush at all — just wanted to make sure they didn't get lost in the shuffle. Thanks again for your time and guidance!

PR #1859, opened 2025-03-29
fix: replace unwrap with expect and improve panic documentation

Himess:
## Description

As discussed in issue #1153 , this PR refactors all `unwrap()` calls in non-test production code, replacing them with `expect()` and providing meaningful context in each message. This helps to improve code robustness and maintainability by avoiding silent panics.

Additionally, functions that may panic now include appropriate `# Panics` sections in their docstrings, explaining under which conditions they might fail, as required by the project guidelines.

## Test plan

- Verified compilation with `cargo check`
- Confirmed style and lint checks with `cargo clippy --all-targets --all-features -- -D warnings`

## Test Result
`cargo check` 

image

`cargo clippy --all-targets --all-features -- -D warnings`

image

All tests passed.

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


Himess:
Dear @mlegner , I haven’t received a response on this PR for a while and several conflicts had accumulated. I’ve resolved them now. Could you please take a look when you have time?

contracts/walrus/sources/staking/staking_pool.move
mlegner, line 1: These should be reverted, these `unwrap`s are of a different kind defined [here](https://github.com/MystenLabs/walrus/blob/aa95d189627da61d9e4c44b75cb92b59db7ab2c5/contracts/walrus/sources/staking/pending_values.move#L81).

mlegner (changes requested):
Thanks for the PR, @Himess, and sorry for my *very long* delay in reviewing. 🙏 

While I agree with the general direction, I see a few issues with the current version:
* The Move changes should be reverted, see the detailed comment.
* In general, neither `unwrap` nor `expect` should be used for error handling, but only in places where the operation is known to succeed. As a consequence, the messages should be something like "should succeed because X" or "we checked above that this is Some", etc. Also, this means that we don't need to document a possible panic in the function's docstring because we know that it cannot actually panic.
* In cases where we actually want to panic if something is `None` or `Err`, I'd prefer an explicit panic with `let Some(x) = y else {panic!("...");};`.
* IMO, `unwrap`s are fine in benchmark code, but I'm also happy to replace them with `expect`s following the conventions above.

If you are willing to revise this PR following the comments above, I'm happy to review it again, but I expect this to require some additional work.

PR #1865, opened 2025-03-31
chore: remove testnet special case for event filtering

karlwuest:
## Description

closes WAL-667

only merge after testnet redeployment

## Test plan

- existing tests

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/node.rs
```diff
@@ -816,33 +805,6 @@ impl StorageNode {
         Ok(EventStreamCursor::new(from_event_id, next_event_index))
     }
 
-    async fn get_num_checkpoints_per_blob(
```
karlwuest, line 819: Note: Removing this will break Testnet V2 -> only merge after we are no longer pushing node binary updates to testnet V2.

mlegner (approved):
Thanks for the cleanup! 🙏

sadhansood (approved):
LGTM!

PR #1869, opened 2025-03-31
feat: per object blob info consistency check

halfprice:
## Description

Similar to #1790 , this PR implements consistency check for per object blob info table. The bulk of the implementation
is to implement an iterator over the per object blob info table.

Also turned on per object blob info consistency check in all simtest.

Contributes to WAL-701

## Test plan

Unit test, simtest

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/consistency_check.rs
+++ b/crates/walrus-service/src/node/consistency_check.rs
@@ -115,19 +115,19 @@ fn compose_certified_blob_list_digest(
         "EpochChange::background_certified_blob_info_consistency_check",
     );
     let epoch_bucket = get_epoch_bucket(epoch);
-    let result = compose_blob_list_digest(
+
+    let value = match compose_blob_list_digest(
         blob_info_iter,
         epoch,
         &node.metrics.blob_info_consistency_check_certified_scanned,
-    );
-
-    if let Err(error) = result {
-        tracing::warn!(?error, "error when processing blob info");
-        node.metrics.blob_info_consistency_check_error.inc();
-        return;
-    }
-
-    let value = result.expect("just checked for error");
+    ) {
+        Ok(value) => value,
+        Err(error) => {
+            tracing::warn!(?error, "error when processing blob info");
+            node.metrics.blob_info_consistency_check_error.inc();
+            return;
+        }
+    };
 
     tracing::info!(?epoch, certified_blob_hash = ?value,
             "background blob info consistency check finished");
@@ -157,23 +157,23 @@ fn compose_certified_object_blob_list_digest(
         "EpochChange::background_certified_blob_object_info_consistency_check",
     );
     let epoch_bucket = get_epoch_bucket(epoch);
-    let result = compose_blob_list_digest(
+
+    let value = match compose_blob_list_digest(
         per_object_blob_info_iter,
         epoch,
         &node
             .metrics
             .per_object_blob_info_consistency_check_certified_scanned,
-    );
-
-    if let Err(error) = result {
-        tracing::warn!(?error, "error when processing per object blob info");
-        node.metrics
-            .per_object_blob_info_consistency_check_error
-            .inc();
-        return;
-    }
-
-    let value = result.expect("just checked for error");
+    ) {
+        Ok(value) => value,
+        Err(error) => {
+            tracing::warn!(?error, "error when processing per object blob info");
+            node.metrics
+                .per_object_blob_info_consistency_check_error
+                .inc();
+            return;
+        }
+    };
 
     tracing::info!(?epoch, certified_blob_hash = ?value,
             "background per-object blob info consistency check finished");
--- a/crates/walrus-service/src/node/consistency_check.rs
+++ b/crates/walrus-service/src/node/consistency_check.rs
@@ -8,12 +8,22 @@ use std::{collections::HashMap, sync::Mutex};
 use std::{hash::Hasher, sync::Arc};
 
 use anyhow::{Context, Result};
+use prometheus::IntCounterVec;
 #[cfg(msim)]
 use sui_types::base_types::ObjectID;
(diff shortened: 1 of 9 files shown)


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -327,15 +362,9 @@ pub(super) trait Mergeable: ToBytes + Debug + DeserializeOwned + Serialize + Siz
...
-// NB: Before adding functions to this trait, think twice if you really need it as it needs to be
-// implementable by future internal representations of the blob status as well.
+/// Trait defining methods for retrieving information about a certified blob.
 #[enum_dispatch]
-pub(crate) trait BlobInfoApi {
-    /// Returns a boolean indicating whether the metadata of the blob is stored.
-    fn is_metadata_stored(&self) -> bool;
-    /// Returns true iff there exists at least one non-expired deletable or permanent `Blob` object.
-    fn is_registered(&self, current_epoch: Epoch) -> bool;
+pub(crate) trait CertifiedBlobInfoApi {
```
jpcsmith, line 367: Question: Why the need for the new API?

jpcsmith (commented):
Looks okay, I will leave markus to make the final pass approval it touches the blob info tables.

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: The BlobInfoIter only returns certified blob info. In this PR, it is changed to iterate through both BlobInfo table and PerObjectBlobInfo table with different value type (one is BlobInfo, and one is PerObjectBlobInfo). Extracting checking whether a blob info represents a certified blob into an API so that it can be used in BlobInfoIter to support both tables.

crates/walrus-service/src/node/metrics.rs
```diff
@@ -143,6 +143,18 @@ walrus_utils::metrics::define_metric_set! {
         #[help = "The number of certified blobs scanned during the blob info consistency check."]
         blob_info_consistency_check_certified_scanned: IntCounterVec["epoch"],
 
+        #[help = "The number of errors occurred when checking the consistency of the per-object \
+        blob info table."]
+        per_object_blob_info_consistency_check: IntGaugeVec["epoch"],
```
mlegner: Question: Is this actually the "number of errors"? If so, the metric name should probably reflect that.

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -32,9 +42,22 @@ pub(super) async fn schedule_background_consistency_check(
...
         let _ = tx.send(());
 
+        // Right now, the computing the two digests are sequential, given that scanning blob info
+        // table is quick. We may consider parallelizing them in the future.
         compose_certified_blob_list_digest(node.clone(), blob_info_iterator, epoch);
+        compose_certified_object_blob_list_digest(
+            node.clone(),
+            per_object_blob_info_iterator,
+            epoch,
+        )
```
mlegner, line 60: Question (for my understanding): Essentially these expose the digests as metrics, correct?

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -45,43 +68,73 @@ pub(super) async fn schedule_background_consistency_check(
...
+        &node.metrics.blob_info_consistency_check_certified_scanned,
+    );
+
+    if let Err(error) = result {
+        tracing::warn!(?error, "error when processing blob info");
+        node.metrics.blob_info_consistency_check_error.inc();
+        return;
+    }
+
+    let value = result.expect("just checked for error");
```
mlegner: Nit: I find the following variant a bit nicer as it doesn't need the `expect` (same below):
```suggestion
    let value = match compose_blob_list_digest(
        blob_info_iter,
        epoch,
        &node.metrics.blob_info_consistency_check_certified_scanned,
    ) {
        Ok(value) => value,
        Err(error) => {
            tracing::warn!(?error, "error when processing blob info");
            node.metrics.blob_info_consistency_check_error.inc();
            return;
        }
    };
```

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -93,3 +146,53 @@ fn compose_certified_blob_list_digest(
             .insert(node.node_capability, value);
     });
 }
+
+/// Compose the digest of the certified object blob list.
+fn compose_certified_object_blob_list_digest(
```
mlegner, line 151: Nit: This function seems to have quite a lot of overlap with the `compose_certified_blob_list_digest`. Is there some chance for deduplication? If not, it's also fine as it is now.

mlegner (approved):
Thanks a lot for this, @halfprice! 💯 

Just a few minor comments.

crates/walrus-service/src/node/consistency_check.rs
halfprice: Correct. It's a u64 value.

crates/walrus-service/src/node/metrics.rs
halfprice: Good catch. Probably messed this part when resolving conflict.

crates/walrus-service/src/node/consistency_check.rs
halfprice: Great suggestion

crates/walrus-service/src/node/consistency_check.rs
halfprice: I started by dedup `compose_certified_blob_list_digest` and `compose_certified_object_blob_list_digest` completely, and metrics and failpoints needed to be passed made the code hard to read. After integrating your suggestion above, these functions are smaller and easier to read.

PR #1873, opened 2025-04-01
feat: Refactor rpc client for code simplification

sadhansood:
## Description

This PR refactors the code in `get_full_checkpoint` and simplify it. Today, the workflow is:
1. try primary FN
2. if failed, quick retry fallback in some cases
3. try primary again
4. retry fallback again

The new code path simplifies this to:
1. try primary FN
2. if failed, retry with fallback


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -961,96 +1008,123 @@ impl RetriableRpcClient {
...
+            },
+            Err(timeout_error) => {
+                if self.fallback_client.is_some() {
+                    Err(RetriableClientError::NonRetryableTimeoutError(
+                        timeout_error,
+                    ))
+                } else {
+                    Err(RetriableClientError::RetryableTimeoutError(timeout_error))
+                }
+            }
```
liquid-helium, line 1254: I believe you've already considered the trade-offs of this quick fallback.

I’d appreciate your insights on a couple of things I’m not familiar with: 

how often is the timeout error intermittent, or is a false positive when the client is overloaded?
Do we have any metrics comparing the performance between retrieving from the archive versus a full node?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -872,36 +884,71 @@ impl CheckpointBucketClient {
...
+
+impl From<tonic::Status> for RetriableClientError {
+    fn from(status: tonic::Status) -> Self {
+        Self::RpcError(RpcError::from(status))
+    }
+}
+
+/// Error type for RPC operations
+#[derive(Error, Debug)]
+pub struct RpcError {
```
liquid-helium: nit: maybe rename as CheckpointRpcError or other name to indicate this is related to checkpoint, since it has the checkpoint specific field checkpoint_seq_num

crates/walrus-sui/src/client/retry_client.rs
sadhansood: The timeout were added to fix the issue of grpc client requests getting stuck for some unknown reason in the tonic crate. We don't really expect to hit this timeout often (I think it is set to over a minute iirc) but it helps the node make progress if it does hit the tonic issue (which we have only observed once in the past). We don't actually have any metrics in the client code, but we should add those as i think they will be helpful for our understanding of client performance.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: sure

PR #1877, opened 2025-04-02
feat: treat jsonrpsee request timeout error as retriable error

halfprice:
## Description

The fuzzer in antithesis test sometimes generates few seconds to close to 1 minutes network congestion, which exposed
that all of our PRCs are currently subject to request timeout. In a few places, such timeout may cause the node to
restart. Instead of adding a retry logic in individual places, I think in general we should treat request timeout as retriable
error, since this will be expected anyway in WAN.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


sadhansood (approved):
LGTM!

PR #1884, opened 2025-04-03
feat: add an option that when resuming shard sync upon node restart always try shard transfer first

halfprice:
## Description

Since at the current stage, shard transfer is significantly faster than shard recovery.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -606,60 +606,48 @@ impl Default for CommitteeServiceConfig {
 #[serde(default)]
 pub struct ShardSyncConfig {
     /// The number of slivers to fetch in a single sync shard request.
-    #[serde(default = "defaults::sliver_count_per_sync_request")]
     pub sliver_count_per_sync_request: u64,
     /// The minimum backoff time for shard sync retries.
     #[serde_as(as = "DurationSeconds<u64>")]
     #[serde(rename = "shard_sync_retry_min_backoff_secs")]
-    #[serde(default = "defaults::shard_sync_retry_min_backoff")]
     pub shard_sync_retry_min_backoff: Duration,
     /// The maximum backoff time for shard sync retries.
     #[serde_as(as = "DurationSeconds<u64>")]
     #[serde(rename = "shard_sync_retry_max_backoff_secs")]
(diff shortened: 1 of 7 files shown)


halfprice:
@jpcsmith @mlegner sent out early draft. Will take another look tomorrow.

halfprice:
@jpcsmith @mlegner this PR is ready for review now.

crates/walrus-service/src/node/config.rs
```diff
@@ -606,42 +606,60 @@ impl Default for CommitteeServiceConfig {
...
     #[serde(rename = "shard_sync_retry_min_backoff_secs")]
+    #[serde(default = "defaults::shard_sync_retry_min_backoff")]
     pub shard_sync_retry_min_backoff: Duration,
     /// The maximum backoff time for shard sync retries.
     #[serde_as(as = "DurationSeconds<u64>")]
     #[serde(rename = "shard_sync_retry_max_backoff_secs")]
+    #[serde(default = "defaults::shard_sync_retry_max_backoff")]
     pub shard_sync_retry_max_backoff: Duration,
     /// The maximum number of concurrent blob recoveries during shard recovery.
+    #[serde(default = "defaults::max_concurrent_blob_recovery_during_shard_recovery")]
```
jpcsmith: Question: Why add the `#[serde(default = ...)]` attribute to each field? `#[serde(default)]` on the struct already deserialises missing fields with the default taken from the struct's default. See, for example, `BlobRecoveryConfig` and [this test case](https://github.com/MystenLabs/walrus/blob/ae35f928a5e214df7f22796e9375d575a8c0ca09/crates/walrus-service/src/node/config.rs#L1215).

jpcsmith (approved):

crates/walrus-service/src/node/config.rs
halfprice: Thanks for the comment @jpcsmith! I didn't know this and a past experience about adding one field causing a binary crash led to these changes. I reverted it and it's much nicer now!

mlegner:
@halfprice Given JP has already reviewed this, I'll remove myself from the reviewers. Feel free to re-request a review though.

PR #1890, opened 2025-04-04
feat: Add metrics to track primary and fallback performance

sadhansood:
## Description

Track latency and success rate from metrics for primary and fallback endpoints for rpc calls.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -20,6 +20,7 @@ use futures::{
     Stream,
     StreamExt,
 };
+use mysten_metrics::monitored_scope;
 use rand::{
     rngs::{StdRng, ThreadRng},
     Rng as _,
@@ -1409,32 +1410,6 @@ impl RetriableRpcClient {
         self.client.with_failover(request).await
     }
 
-    /// Records RPC latency metrics
-    fn record_rpc_latency(&self, method: &str, status: &str, duration: Duration) {
-        if let Some(metrics) = &self.metrics {
-            metrics
-                .rpc_latency
-                .with_label_values(&[method, self.client.get_current_client_name(), status])
-                .observe(duration.as_millis() as f64);
-        }
-    }
-
-    /// Records fallback metrics
-    fn record_fallback_metrics(
-        &self,
-        method: &str,
-        result: &Result<impl Sized, impl Sized>,
-        duration: Duration,
-    ) {
-        if let Some(metrics) = &self.metrics {
-            let status = if result.is_ok() { "success" } else { "failure" };
-            metrics
-                .rpc_latency
-                .with_label_values(&[method, "fallback", status])
-                .observe(duration.as_millis() as f64);
-        }
-    }
-
     /// Gets the full checkpoint data for the given sequence number.
     ///
     /// This function will first try to fetch the checkpoint from the primary client with retries.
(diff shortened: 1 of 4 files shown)


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1400,6 +1409,32 @@ impl RetriableRpcClient {
...
+    /// Records RPC latency metrics
+    fn record_rpc_latency(&self, method: &str, status: &str, duration: Duration) {
+        if let Some(metrics) = &self.metrics {
+            metrics
+                .rpc_latency
+                .with_label_values(&[method, self.client.get_current_client_name(), status])
+                .observe(duration.as_millis() as f64);
+        }
+    }
+
```
liquid-helium: Is it possible to make this a function of `SuiClientMetricSet `, to be consistent with other metrics in it.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1409,18 +1444,33 @@ impl RetriableRpcClient {
         &self,
         sequence_number: u64,
     ) -> Result<CheckpointData, RetriableClientError> {
+        let start_time = Instant::now();
         let error = match self.get_full_checkpoint_from_primary(sequence_number).await {
-            Ok(checkpoint) => return Ok(checkpoint),
+            Ok(checkpoint) => {
+                self.record_rpc_latency("get_full_checkpoint", "success", start_time.elapsed());
+                return Ok(checkpoint);
+            }
```
liquid-helium: This should have been recorded within `retry_rpc_errors` on line 1403, all prc calls wrapped in `retry_rpc_errors` are recorded.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1400,6 +1409,32 @@ impl RetriableRpcClient {
         self.client.with_failover(request).await
     }
 
+    /// Records RPC latency metrics
+    fn record_rpc_latency(&self, method: &str, status: &str, duration: Duration) {
+        if let Some(metrics) = &self.metrics {
+            metrics
+                .rpc_latency
+                .with_label_values(&[method, self.client.get_current_client_name(), status])
```
halfprice: nit: in Walrus repo, we use `walrus_utils::with_label!` macro in most of places, can we keep it consistent here?

crates/walrus-sui/src/client/retry_client.rs
halfprice: Indeed. I think this latency includes the latency of retries, maybe still worth adding.

One thing that maybe interesting to record in `retry_rpc_errors` is the number of retries.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1409,18 +1444,33 @@ impl RetriableRpcClient {
...
+        let result = self
+            .get_full_checkpoint_from_fallback_with_retries(fallback, sequence_number)
+            .await;
+
+        self.record_fallback_metrics(
+            "get_full_checkpoint",
+            &result,
+            fallback_start_time.elapsed(),
+        );
+
```
halfprice, line 1464: Suggestion: may also be interesting to record the whole function latency from `start_time` until it returns. Maybe use `monitored_scope`, which records the time spent in this whole function until return?

halfprice (approved):

crates/walrus-sui/src/client/retry_client.rs
sadhansood: @halfprice I checked, the number of tries is already being recorded in `retry_rpc_errors`, lmk if that doesn't feel enough.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1409,18 +1419,50 @@ impl RetriableRpcClient {
...
+        let _scope = monitored_scope("RetriableRpcClient::get_full_checkpoint");
+        let start_time = Instant::now();
         let error = match self.get_full_checkpoint_from_primary(sequence_number).await {
-            Ok(checkpoint) => return Ok(checkpoint),
+            Ok(checkpoint) => {
+                self.metrics.as_ref().inspect(|metrics| {
+                    metrics.record_rpc_latency(
+                        "get_full_checkpoint",
+                        self.client.get_current_client_name(),
+                        "success",
```
liquid-helium, line 1430: nit: This is [measured](https://github.com/MystenLabs/walrus/blob/c87d44a2184c8de4ec9285ed646c4a82ab0cab39/crates/walrus-sui/src/client/retry_client.rs#L1386-L1399) in `get_full_checkpoint_from_primary` as "`get_full_checkpoint`" already, if we want to measure the overall latency, can we change the name to reflect that? although they are in different metrics, I am afraid it would cause some confusion.

sadhansood: Ah, the underlying metric name is quite different, so I think we shouldn't technically get confused. But if it gets really hard to manage, we can rename it in the future.

PR #1891, opened 2025-04-04
feat: Increase default block cache size

sadhansood:
## Description

Increase default block cache size of column families, and allow for users to change certain subset of the db config easily.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage/database_config.rs
+++ b/crates/walrus-service/src/node/storage/database_config.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Walrus Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use rocksdb::{DBCompressionType, Options};
+use rocksdb::{BlockBasedOptions, Cache, DBCompressionType, Options};
 use serde::{Deserialize, Serialize};
 
 /// Options for configuring a column family.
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -69,6 +75,59 @@ impl DatabaseTableOptions {
...
+    }
+
+    fn blob_info() -> Self {
+        Self {
+            block_cache_size: Some(1 << 29),
+            ..Default::default()
+        }
+    }
+
+    pub fn with_default_override(
```
jpcsmith: Nit: Add a docstring.

Nit: Rename to `inherit_from` or similar as this confused me a bit as it does not use the `Default` implementation for anything and doesn't override a value set by the user, but instead fills in values that were not set.

crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -69,6 +75,59 @@ impl DatabaseTableOptions {
             write_buffer_size: Some(256 << 20),
             target_file_size_base: Some(64 << 20),
             max_bytes_for_level_base: Some(512 << 20),
+            block_cache_size: Some(1 << 28),
```
jpcsmith: Nit: In general these defaults would be more readable as something like `block_cache_size: mebibyte(256)`. Otherwise, make this `block_cache_size: 256 << 20`.

jpcsmith (approved):

mlegner:
@sadhansood Given JP has already reviewed this, I'll remove myself from the reviewers. Feel free to re-request a review though.

PR #1894, opened 2025-04-04
fix: reduce blocking on tokio core threads

jpcsmith:
## Description

This moves two verifications to run on background threads and changes the methods that store slivers such that they use `tokio::task::spawn_blocking`.

Depends on #1893 

## Test plan

Refactoring, existing tests apply.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1753,7 +1753,7 @@ impl StorageNodeInner {
             .get_and_verify_metadata(*blob_id, certified_epoch)
             .await;
 
-        self.storage.put_verified_metadata(&metadata)?;
+        self.storage.put_verified_metadata(&metadata).await?;
         tracing::debug!(%blob_id, "metadata successfully synced");
         Ok(metadata)
     }
@@ -2218,6 +2218,7 @@ impl ServiceState for StorageNodeInner {
 
         self.storage
             .put_verified_metadata(&verified_metadata_with_id)
+            .await
             .context("unable to store metadata")?;
 
         self.metrics
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -204,7 +204,7 @@ pub trait ServiceState {
     fn store_metadata(
         &self,
         metadata: UnverifiedBlobMetadataWithId,
-    ) -> Result<bool, StoreMetadataError>;
+    ) -> impl Future<Output = Result<bool, StoreMetadataError>> + Send;
 
     /// Returns whether the metadata is stored in the shard.
     fn metadata_status(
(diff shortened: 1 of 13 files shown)


crates/walrus-service/src/node/config.rs
```diff
@@ -1002,14 +1002,25 @@ impl Default for BalanceCheckConfig {
...
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(default)]
 pub struct ThreadPoolConfig {
     /// Specify the maximum number of concurrent tasks that will be pending on the thread pool.
     ///
     /// Defaults to an amount calculated from the number of cores.
     #[serde(skip_serializing_if = "defaults::is_none")]
     pub max_concurrent_tasks: Option<usize>,
+    /// Specify the maximum number of blocking threads to use for I/O.
+    pub max_blocking_io_threads: usize,
```
halfprice, line 1014: Nit: I think these two config currently have different scope (max_concurrent_tasks is only used in BoundedThreadPool whereas the max_blocking_io_thread is used in runtime setup), and I wonder if it make sense to be a bit more specific.

crates/walrus-service/src/node.rs
```diff
@@ -581,6 +582,10 @@ impl StorageNode {
         };
         tracing::info!("successfully opened the node database");
 
+        let thread_pool = ThreadPoolBuilder::default()
```
halfprice, line 585: Question: IIUC, right now, we are using 2 thread pools in storage nodes:
- this thread pool to do some verification blocking work in serving requests
- the tokio blocking thread pool for RocksDB IOs.

Is this fundamental, or we can unify them so that we only need to manage 1 thread pool per node?

halfprice (approved):
Thanks @jpcsmith for this important improvement! This looks great! I assume putting reading the sliver into blocking threads is on your radar too?

crates/walrus-service/src/node.rs
jpcsmith: I do not currently see a clear path to unifying them. Threads pools for CPU-bound work should only have as many threads as there are cores, whereas thread-pools for I/O-bound work benefit from a significantly higher limit as the threads there spend a "lot" of time sleeping. If we were to have a single thread pool, then it would need to have a large amount of threads to handle both cases. We would then need to limit the amount of CPU-bound work that has access to the pool. However, let's assume that we have 1000 threads there, and we have a batch of 1000 I/O bound requests coming in, then all the threads are doing I/O-bound work which means they may be sleeping for some time and CPU utilisation is low, but we cannot schedule any CPU-bound work since all threads are currently occupied.

crates/walrus-service/src/node/config.rs
jpcsmith: They do, and I considered adding an extra config section; but then thought that how we use these in the code is an implementation detail that the user may not care about. This way, the user just considers that the setting is related to "thread pools", whether it's one or multiple it's irrelevant.

PR #1903, opened 2025-04-07
Infer whether aggregators support cache

nikos-terzo:
## Description

Script to fill the cache-data for the operators.json used in mdbook.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/scripts/cache-inference/package.json
+++ b/scripts/cache-inference/package.json
@@ -1,0 +1,12 @@
+{
+  "name": "cache-inference",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC"
+}
--- a/scripts/cache-inference/tsconfig.json
+++ b/scripts/cache-inference/tsconfig.json
@@ -1,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
+    // For importing json
+    "resolveJsonModule": true,
+    "esModuleInterop": true
+  }
+}
--- a/scripts/cache-inference/types.ts
+++ b/scripts/cache-inference/types.ts
@@ -1,0 +1,24 @@
+
+export type Network = "mainnet" | "testnet";
+
+export type AggregatorData = {
+    cache?: {
+        hasCache: boolean;
+        headers: {
+            key: string;
+            value: string | null;
+        }[];
+        // TODO
+        speedup?: any
+    }
+}
+
+export type PublisherData = Object;
+
+export type NetworkData = {
+    aggregators?: Record<string, AggregatorData>;
+    publishers?: Record<string, PublisherData>;
+};
+
+export type Operators = Record<Network, NetworkData>
+
--- a/docs/book/assets/operators.json
+++ b/docs/book/assets/operators.json
@@ -1,125 +1,1553 @@
 {
   "mainnet": {
     "aggregators": {
-      "https://aggregator.mainnet.walrus.mirai.cloud": {},
-      "https://aggregator.suicore.com": {},
-      "https://aggregator.walrus-mainnet.tududes.com": {},
-      "https://aggregator.walrus-mainnet.walrus.space": {},
-      "https://aggregator.walrus.atalma.io": {},
-      "https://aggregator.walrus.mainnet.mozcomputing.dev": {},
-      "https://aggregator.walrus.silentvalidator.com": {},
-      "https://mainnet-aggregator.hoh.zone": {},
-      "https://mainnet-aggregator.walrus.graphyte.dev": {},
-      "https://mainnet-walrus-aggregator.kiliglab.io": {},
-      "https://sm1-walrus-mainnet-aggregator.stakesquid.com": {},
-      "https://sui-walrus-mainnet-aggregator.bwarelabs.com": {},
-      "https://suiftly.mhax.io": {},
-      "https://wal-aggregator-mainnet.staketab.org": {},
-      "https://walmain.agg.chainflow.io": {},
-      "https://walrus-agg-main.bucketprotocol.io": {},
-      "https://walrus-agg.mainnet.obelisk.sh": {},
-      "https://walrus-aggregator-mainnet.chainode.tech:9002": {},
-      "https://walrus-aggregator.chainbase.online": {},
-      "https://walrus-aggregator.n1stake.com": {},
-      "https://walrus-aggregator.natsai.xyz": {},
-      "https://walrus-aggregator.rockfin.io": {},
-      "https://walrus-aggregator.rubynodes.io": {},
-      "https://walrus-aggregator.stakely.io": {},
-      "https://walrus-aggregator.stakin-nodes.com": {},
-      "https://walrus-aggregator.staking4all.org": {},
-      "https://walrus-aggregator.starduststaking.com": {},
-      "https://walrus-aggregator.talentum.id": {},
-      "https://walrus-aggregator.thcloud.ai": {},
-      "https://walrus-aggregator.thepassivetrust.com": {},
-      "https://walrus-cache-mainnet.latitude.sh": {},
-      "https://walrus-cache-mainnet.overclock.run": {},
-      "https://walrus-main-aggregator.4everland.org": {},
-      "https://walrus-mainnet-aggregator-1.zkv.xyz": {},
-      "https://walrus-mainnet-aggregator.crouton.digital": {},
-      "https://walrus-mainnet-aggregator.dzdaic.com": {},
-      "https://walrus-mainnet-aggregator.everstake.one": {},
-      "https://walrus-mainnet-aggregator.imperator.co": {},
-      "https://walrus-mainnet-aggregator.luckyresearch.org": {},
-      "https://walrus-mainnet-aggregator.nami.cloud": {},
-      "https://walrus-mainnet-aggregator.nodeinfra.com": {},
-      "https://walrus-mainnet-aggregator.redundex.com": {},
-      "https://walrus-mainnet-aggregator.rpc101.org": {},
-      "https://walrus-mainnet-aggregator.stakecraft.com": {},
-      "https://walrus-mainnet-aggregator.stakeengine.co.uk": {},
-      "https://walrus-mainnet-aggregator.stakingdefenseleague.com": {},
-      "https://walrus-mainnet-aggregator.trusted-point.com": {},
-      "https://walrus.aggregator.stakepool.dev.br": {},
-      "https://walrus.blockscope.net": {},
-      "https://walrus.globalstake.io": {},
-      "https://walrus.lakestake.io": {},
-      "https://walrus.prostaking.com:9443": {},
-      "https://walrus.veera.com": {},
-      "https://walrusagg.pops.one": {},
-      "http://cvchfgpvfth0e3ss0dkg.bdnodes.net:9000": {},
-      "http://walrus-aggregator.brightlystake.com:9000": {},
-      "http://walrus-aggregator.winsnip.site:9000": {},
-      "http://walrus.equinoxdao.xyz:9000": {},
-      "http://67.220.194.10:9000": {}
+      "https://aggregator.mainnet.walrus.mirai.cloud": {
+        "cache": {
+          "hasCache": true,
+          "headers": [
+            {
+              "key": "cache-control",
+              "value": "public, max-age=86400, stale-while-revalidate=3600"
+            },
+            {
+              "key": "cf-cache-status",
+              "value": "DYNAMIC"
+            }
+          ]
+        }
+      },
+      "https://aggregator.suicore.com": {
+        "cache": {
+          "hasCache": true,
+          "headers": [
+            {
+              "key": "cache-control",
+              "value": "public, max-age=86400, stale-while-revalidate=3600"
+            },
+            {
+              "key": "cf-cache-status",
+              "value": "DYNAMIC"
+            },
+            {
+              "key": "x-cache-status",
+              "value": "HIT"
+            }
+          ]
+        }
+      },
+      "https://aggregator.walrus-mainnet.tududes.com": {
+        "cache": {
+          "hasCache": true,
+          "headers": [
+            {
+              "key": "cache-control",
+              "value": "public, max-age=86400, stale-while-revalidate=3600"
+            }
+          ]
+        }
+      },
+      "https://aggregator.walrus-mainnet.walrus.space": {
+        "cache": {
+          "hasCache": true,
+          "headers": [
+            {
+              "key": "cache-control",
+              "value": "public, max-age=86400, stale-while-revalidate=3600"
+            },
+            {
+              "key": "cf-cache-status",
+              "value": "HIT"
+            },
+            {
+              "key": "x-cache-status",
+              "value": "MISS"
+            }
+          ]
+        }
+      },
+      "https://aggregator.walrus.atalma.io": {
+        "cache": {
+          "hasCache": true,
+          "headers": [
+            {
+              "key": "cache-control",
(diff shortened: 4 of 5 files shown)


nikos-terzo:
@mlegner, during for checking for cache relevant headers, I observed that different aggregators return different patterns of headers, so I thought we can include it in the script response. Do you think this is okay or TMI?
eg:
```json
    "https://aggregator.walrus-testnet.walrus.space": {
        "cache": {
          "hasCache": true,
          "headers": [
            {
              "key": "cache-control",
              "value": "public, max-age=86400, stale-while-revalidate=3600"
            },
            {
              "key": "cf-cache-status",
              "value": "HIT"
            },
            {
              "key": "x-cache-status",
              "value": "MISS"
            }
          ]
        }
      },
 ```

**Comment by nikos-terzo** [comment]
> I currently don't see any benefit of adding the caching headers in the JSON file. Do you have a concrete use case in mind? If not, I'd suggest not to include them as they add a lot of noise to the file.

Thanks for the feedback @mlegner !
I was seeing that the header-patterns are grouped into 3-4 clusters, and was wondering whether these clusters would correlate with the speedup measurements to be implemented, but I agree that this might not be the right place for this data; will remove.

**Comment by nikos-terzo** [comment]
@mlegner, can you take a look? The GH action is currently a manual one and it has the inputs for the blob-ids to check for mainnet and testnet. It automatically creates [a new PR](https://github.com/MystenLabs/walrus/pull/1988).
In order to make the GH action run automatically once every 1 week, one needs to set the blob-ids [here](https://github.com/MystenLabs/walrus/blob/nt/aggregator-cache-inference/.github/workflows/update-operators-cache.yml#L18-L19) , to non-expired blob-ids and uncomment [these lines](https://github.com/MystenLabs/walrus/blob/nt/aggregator-cache-inference/.github/workflows/update-operators-cache.yml#L14-L15).
I have the cron-trigger still disabled until a solution to the "indefinitely available blob-ids" is found.

File: .github/workflows/update-operators-cache.yml [hunk 1]
```diff
@@ -0,0 +1,54 @@
+name: Update Operators Cache
```
mlegner: Nit:
```suggestion
name: Update Aggregator Cache Info
```

.github/workflows/update-operators-cache.yml
```diff
@@ -0,0 +1,54 @@
...
+      testnet-blob-id:
+        description: 'Testnet blob ID'
+        required: true
+        type: string
+  # schedule:
+  #   - cron: '0 11 * * 1'  # Every Monday at 11:00 UTC
+
+env:
+  MAINNET_BLOB_ID: ${{ github.event.inputs.mainnet-blob-id || '6SCk-C85lApQY_3bGwJMpVve5ASWiIhxWplg22fKHvo' }}
+  TESTNET_BLOB_ID: ${{ github.event.inputs.testnet-blob-id || 'AYTUffH740r9eHKRgICusBEw-M6DucKqIBAHctZDCC8' }}
```
mlegner: > In order to make the GH action run automatically once every 1 week, one needs to set the blob-ids [here](https://github.com/MystenLabs/walrus/blob/nt/aggregator-cache-inference/.github/workflows/update-operators-cache.yml#L18-L19) , to non-expired blob-ids and uncomment [these lines](https://github.com/MystenLabs/walrus/blob/nt/aggregator-cache-inference/.github/workflows/update-operators-cache.yml#L14-L15). I have the cron-trigger still disabled until a solution to the "indefinitely available blob-ids" is found.

Ideally we would simply store a new random blob within the workflow before running the measurements. That would also ensure that we have a cache miss on the first request and a hit on the second. 

I can set this up in a follow-up. For now, let's just store some blobs for 53 epochs and use them, we can also use them as defaults in the `workflow_dispatch` inputs.

.github/workflows/update-operators-cache.yml
```diff
@@ -0,0 +1,54 @@
...
+
+jobs:
+  update-cache:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
mlegner: Suggestion: Can you pin all the actions to specific commits like we do in other workflow files?

.github/workflows/update-operators-cache.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: |
+          npx ts-node scripts/cache-inference/cache-inference.ts $MAINNET_BLOB_ID $TESTNET_BLOB_ID > docs/book/assets/new-operators.json
+          mv docs/book/assets/new-operators.json docs/book/assets/operators.json
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "Update operators.json with new cache measurements"
+          title: "chore: Update operators.json with new cache measurements"
```
mlegner: Nit:
```suggestion
          commit-message: "chore: update aggregator cache info"
          title: "chore: update aggregator cache info"
```

scripts/cache-inference/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "cache-inference",
+  "version": "1.0.0",
+  "description": "",
```
mlegner: Nit: Can you add a description here?

scripts/cache-inference/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
```
mlegner: Question: Why not `ES2024`?

scripts/cache-inference/cache-inference.ts
mlegner, line 1: Can you document how this is expected to be installed and run manually? For example, when I run `npm install` (as in the GH workflow), a separate `package-lock.json` file is created. If this is expected, should we check this in as well?

Or should this be installed with `pnpm install`?

scripts/cache-inference/cache-inference.ts
```diff
@@ -0,0 +1,143 @@
...
+            headers2 = resp2.headers;
+        } catch (e) {
+            console.error(`Error during measuring ${blobUrl}:`);
+            console.error(e);
+            continue;
+        }
+        const speedupMs = fetch1 - fetch2;
+
+        let output1 = headerKeyContainsCache(headers1);
+        let output2 = headerKeyContainsCache(headers2);
```
mlegner: Nit: These could be `const` as well:
```suggestion
        const output1 = headerKeyContainsCache(headers1);
        const output2 = headerKeyContainsCache(headers2);
```
Also, the variable names are not very informative.

scripts/cache-inference/types.ts
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+export type Network = "mainnet" | "testnet";
+
+export type AggregatorData = {
+    cache?: {
+        hasCache: boolean;
+        speedupMs?: number | [number, [number, number]]
```
mlegner: Question: What is the reason to add this to the JSON file? I don't see a concrete use case for this.

scripts/cache-inference/cache-inference.ts
```diff
@@ -0,0 +1,143 @@
...
+                hasCache,
+                headers: merged,
+                speedupMs: [speedupMs, [fetch1, fetch2]]
+            }
+        };
+    }
+    // let results = {
+    //     aggregators: aggregatorsVerbose,
+    // }
+    // console.log(JSON.stringify(results, null, 2));
```
mlegner: Nit: Can we either uncomment or remove this?

.github/workflows/update-operators-cache.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+
+      - name: Install dependencies
+        run: cd scripts/cache-inference && npm install
```
mlegner: Remark: This currently results in the creation of a `package-lock.json` file, which is subsequently included in the PR, see [here](https://github.com/MystenLabs/walrus/pull/1988/files).

scripts/cache-inference/cache-inference.ts
```diff
@@ -0,0 +1,143 @@
...
+}
+
+const THRESHOLD: number = 1000;
```
mlegner: Nit: Can you move that to the top of the file and add a comment explaining the unit (I guess ms)?

scripts/cache-inference/cache-inference.ts
```diff
@@ -0,0 +1,143 @@
...
+function headersHaveCacheHit(matches: HeaderMatch[]): boolean {
+    return matches.some(({ key: _, value }) => {
+        return value?.toLowerCase().includes("hit");
+    });
+}
+
+async function updateAggregatorCacheInfo(
+    aggregators: Record<string, AggregatorData>,
+    blobId: string,
+    threshold: number,
```
mlegner: Nit: If this is always called with `THRESHOLD`, we could just remove the function parameter.

scripts/cache-inference/cache-inference.ts
```diff
@@ -0,0 +1,143 @@
...
+            });
+        }
+
+        aggregatorsVerbose[url] = {
+            cache: {
+                hasCache,
+                headers: merged,
+                speedupMs: [speedupMs, [fetch1, fetch2]]
+            }
+        };
```
mlegner: Question: (Where) do we actually use all this?

mlegner (commented):
Thanks a lot for adding this, @nikos-terzo! I am not very familiar with TS, so I have a bunch of questions. 🙏

scripts/cache-inference/cache-inference.ts
nikos-terzo: Yes it is expected. We can commit the package-lock.json as well.

Done.

scripts/cache-inference/cache-inference.ts
nikos-terzo: So I was using it in case something does not run as expected as better debug information by printing them with the commented below lines. This is how I actually found all cache-relevant headers.
Now that I rethink of this, maybe we should use an argument like `--debug` or `--verbose` to print this verbose information only when running it locally?
EDIT: Actually the `KnownCacheKeys` was used before I was searching for any header containing the "cache" substring. With this simple substring search, we can remove this. I still think that the verbose print might prove useful though.

scripts/cache-inference/cache-inference.ts
nikos-terzo: Similar to above. Will add optional argument.

scripts/cache-inference/tsconfig.json
nikos-terzo: No reason, will update.

scripts/cache-inference/types.ts
nikos-terzo: Ok, will reset to `cache?: boolean`.

.github/workflows/update-operators-cache.yml
nikos-terzo: Stored file with contents: `testblob for measuring cache` for `--max` epochs in testnet and 53 epochs in mainnet (`bfH-ZqYV5faXTiq2-FVBT2cLQD61VYRuDje0cIkMWMo`).
Also uncommented cron job.

scripts/cache-inference/tsconfig.json
```diff
@@ -0,0 +1,8 @@
+{
+  "compilerOptions": {
+    "target": "ES2022",
```
nikos-terzo, line 3: This is the most recent supported by the latest lts node version.

scripts/cache-inference/cache-inference.ts
mlegner: Could you add a (short) README.md to `scripts/cache-inference/` just describing the purpose and usage of this script?

.github/workflows/update-operators-cache.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          fetch-depth: 0
+
+      - name: Setup Node.js
+        uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020
+        with:
+          node-version: '20'
+
+      - name: Install dependencies
+        run: cd scripts/cache-inference && npm install
```
mlegner: Nit: The following is a bit cleaner:
```suggestion
        working-directory: ./scripts/cache-inference
        run: npm install
```

scripts/cache-inference/cache-inference.ts
```diff
@@ -0,0 +1,148 @@
...
+        return value?.toLowerCase().includes("hit");
+    });
+}
+
+async function updateAggregatorCacheInfo(
+    aggregators: Record<string, AggregatorData>,
+    blobId: string,
+    verbose: boolean,
+) {
+    const aggregatorsVerbose: Record<string, AggregatorDataVerbose> = {};
```
mlegner: Question: How can this be `const`?

scripts/cache-inference/cache-inference.ts
```diff
@@ -0,0 +1,148 @@
...
+                headersWithCacheKey2.map(({ key, value }) => [key, value])
+            );
+            const merged = headersWithCacheKey1.reduce<Record<string, [HeaderValue, HeaderValue]>>(
+                (acc, { key, value }) => {
+                    acc[key] = [value, map2[key] ?? undefined];
+                    return acc;
+                }, {}
+            );
+
+            aggregatorsVerbose[url].cacheHeaders = merged;
```
mlegner: Important: When running `npx ts-node scripts/cache-inference/cache-inference.ts -v bfH-ZqYV5faXTiq2-FVBT2cLQD61VYRuDje0cIkMWMo`, I get the following error:
```
/Users/markuslegner/github/walrus/scripts/cache-inference/cache-inference.ts:90
            aggregatorsVerbose[url].cacheHeaders = merged;
                                                ^
TypeError: Cannot set properties of undefined (setting 'cacheHeaders')
    at updateAggregatorCacheInfo (/Users/markuslegner/github/walrus/scripts/cache-inference/cache-inference.ts:90:49)
    at processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async run (/Users/markuslegner/github/walrus/scripts/cache-inference/cache-inference.ts:142:5)
```

mlegner (commented):
Thanks for the changes, @nikos-terzo! A few small remaining comments.

scripts/cache-inference/cache-inference.ts
nikos-terzo: So, `const` in javascript is about assignment, not mutation. In other words, `const` can be thought of meaning that this variable will point to the same place in memory and not be reassigned. It doesn't mean that the contents of the variable are immutable.
(I know, javascript 😬)

scripts/cache-inference/cache-inference.ts
nikos-terzo: Sorry about that, will fix.

scripts/cache-inference/README.md
```diff
@@ -0,0 +1,35 @@
...
+## Overview
+
+The script makes two sequential requests to each aggregator and determines caching status through:
+1. **Headers**: Detects cache-related headers with "hit" values
+2. **Performance**: Identifies caching through response time improvements (>1000ms)
+
+## Usage
+
+```bash
+npx ts-node ./cache-inference.ts <mainnetBlobId> [testnetBlobId] [--verbose]
```
mlegner, line 15: Nit: Can you add how to install dependencies with `npm`?

mlegner (approved):
Thanks so much for your work and patience, @nikos-terzo! 🙏

scripts/cache-inference/README.md
mlegner:
```suggestion
npm install --prefix scripts/cache-inference
npx ts-node ./cache-inference.ts <mainnetBlobId> [testnetBlobId] [--verbose]
```

PR #1904, opened 2025-04-07
fix: save and reuse metrics instead of adding a unique ID

jpcsmith:
## Description

Adds a wrapper around `prometheus::Registry` that stores metrics as they are registered and provides the method `get_or_register` which attempts to register a metric or returns the stored metric if its already registered.

## Test plan

Added unit tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12208,6 +12208,7 @@ dependencies = [
  "serde_json",
  "serde_with",
  "tempfile",
+ "thiserror 2.0.12",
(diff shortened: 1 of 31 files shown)


mlegner (approved):
Thank you very much for the fix, this LGTM, although I don't fully understand all the details.

PR #1905, opened 2025-04-07
docs: add docs on the authenticated publisher

giac-mysten:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/docs/book/operator-guide/auth-publisher.md
+++ b/docs/book/operator-guide/auth-publisher.md
@@ -1,7 +1,7 @@
 # The authenticated publisher
 
 We now describe the authenticated publisher, which requires the HTTP request to store a blob to be
-authenticated. Such authenticated publisher can be used as a building block for services that
+authenticated. Such an authenticated publisher can be used as a building block for services that
 require storing over HTTP on Walrus `mainnet`, where an "open" publisher is undesirable (because of
 the `SUI` and `WAL` cost of publishing to Walrus).
 
@@ -32,6 +32,39 @@ The authenticated publishing flow occurs, at a high level, as follows:
 1. (Optional) **Asset return:**: If so specified, the publisher returns the newly-created `Blob`
    object to the Sui Address set in the request.
 
+### Request authentication flow
+
+We now provide an example of how the authenticated publisher can be used in a webapp that allows
+users to upload files to Walrus through a web frontend. This is especially useful if the users are
+*not* required to have a wallet, and therefore cannot store blobs to Walrus on their own.
+
+- The user connects to the webapp, authenticates (e.g., through name and password, as they have no
+  wallet).
+- The user uploads the file it wants to store, and selects the number of epochs for which it would
+  like to store it.
+- The webapp frontend sends information on the size of the file and the number of epochs to the
+  backend.
+- With this information, the backend computes if the user is authorized to store the given amount of
+  data for the selected number of epochs, and possibly reduces the user quota. This accounting can
+  be done locally, but also directly on Sui. Importantly, at this point the backend can also check
+  the cost of the upload (in terms of SUI and WAL), and check if it is within the user's budget.
+- If the user is authorized to store, the backend returns a JWT token with the size and epoch limits
+  to the frontend.
+- The frontend sends the file to the publisher via a `PUT` request, setting the JWT in the
+  Authorization header as a Bearer token.
+- Upon receipt, the publisher verifies:
+  - The token signature using the configured secret;
+  - Token expiration;
+  - Token uniqueness (prevents replay attacks);
+  - Upload parameters against claims (if enabled).
+- If all checks succeed, the publisher stores the file on Walrus.
+- If set, the publisher finally returns the created `Blob` object to the address specified by the
+  user.
+
+One important note here is that the JWT token itself is a "single use" token, and should not be used
+for accounting purposes. I.e., one token can be used to store only one file, and after that the
+token is rejected by the replay suppression system.
+
 ## Publisher setup
 
 The publisher is configured at startup using the following command line arguments:
@@ -40,7 +73,7 @@ The publisher is configured at startup using the following command line argument
   only store blobs with valid JWTs.
 - `--jwt-algorithm`: The algorithm used for JWT verification (defaults to HMAC).
 - `--jwt-expiring-sec`: Duration in seconds after which the JWT is considered expired.
-- `--jwt-verify-upload`: Enable verification of upload parameters against JWT claims
+- `--jwt-verify-upload`: Enable verification of upload parameters against JWT claims.
 
 Additional details follow.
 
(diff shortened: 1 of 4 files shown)


docs/book/operator-guide/auth-publisher.md
```diff
@@ -0,0 +1,126 @@
...
+The authenticated publishing flow occurs, at a high level, as follows:
+
+1. **Publisher setup**: The publisher operator:
+    - Funds the publisher's wallet with sufficient `SUI` and `WAL`;
+    - Configures the publisher to only accept authenticated requests. This entails setting the
+      algorithm to authenticate JWTs, the expiration time for JWTs, and the JWT authentication
+      secret.
+1. **Authentication channel setup**: The publisher operator sets up a channel through which users
+   can obtain the JWT tokens. This step can be performed in any way that produces a valid JWT, and
+   is not provided in this implementation.
```
gdanezis, line 23: Can we provide at least some example software that may be used to set this up?

docs/book/operator-guide/auth-publisher.md
```diff
@@ -0,0 +1,126 @@
+# The authenticated publisher
+
+We now describe the authenticated publisher, which requires the HTTP request to store a blob to be
+authenticated. Such authenticated publisher can be used as a building block for services that
```
gdanezis: "Such an" ...

docs/book/operator-guide/auth-publisher.md
```diff
@@ -0,0 +1,126 @@
...
+disables the checks that ensure the contents of the JWT claim match the requested blob upload.
+
+Specifically, the publisher will:
+
+- Verify that the number of `epochs` in query is the the same as `epochs` in the JWT, if present;
+- Verify that the `send_object_to` field in the query is the same as the `send_object_to` in the
+  JWT, if present;
+- Verify the size of uploaded file;
+- Verify the uniqueness of the `jti` claim.
+
```
gdanezis, line 114: Tell me here under what conditions I should disable verification? What is this intended to do?

docs/book/operator-guide/auth-publisher.md
```diff
@@ -0,0 +1,126 @@
...
+The replay suppression supports the following configuration parameters:
+
+- `--jwt-cache-size`: The maximum size of the publisher's JWT cache, where the `jti` JWT IDs of the
+  "used" JWTs are kept until their expiration. This is a hard upperbound on the number of entries in
+  the cache, after which additional requests to store are rejected. This hard bound is introduced to
+  avoid DoS attacks on the publisher through the cache.
+- `--jwt-cache-refresh-interval`: The interval (in seconds) after which the cache is refreshed, and
+  expired JWTs are removed (possibly creating space for additional JWTs to be inserted).
+
```
gdanezis, line 132: Explain here: Am I meant to provide a fresh jwt token per upload - or if not how am I meant to use replay supression?

docs/book/operator-guide/auth-publisher.md
```diff
@@ -0,0 +1,126 @@
...
+## JWT Fields
+
+The current authenticated publisher implementation does not provide a way to generate the JWTs and
+distribute them to the clients. These can be generated with any tool, as long as they respect the
+following constraints.
+
```
gdanezis, line 140: Here a couple of major example services would really help

docs/book/operator-guide/auth-publisher.md
```diff
@@ -0,0 +1,126 @@
...
+`max_size`. Using both in either case will result in token rejection.
+
+## Request authentication flow
+
+- The token is sent in the Authorization header as a Bearer token
+- The publisher verifies:
+  - Token signature using the configured secret
+  - Token expiration
+  - Token uniqueness (prevents replay attacks)
+  - Upload parameters against claims (if enabled)
```
gdanezis: This doc is a good start but needs to describe the common case: how do I ensure as a publisher that an authenticated user can use up to X WAL/SUI for uploads. What combination of the above config should I use?

docs/book/operator-guide/auth-publisher.md
giac-mysten: The original idea, I think, was to separate the enabiling/disabling cryptographic checks and "semantic" checks, based on the specifics of the publisher. 

I'm not sure, though, where this would be needed. @yanganto do you have a usecase in mind?

docs/book/operator-guide/auth-publisher.md
giac-mysten: Yes, the idea is the following: 
- the user uploads a file. The app (frontend, or local) checks the size and communicates with the backend
- the backend does some accounting to decide if the user can store that amount (for the number of epochs required)
- the backend returns the token for this specific file

Importantly, the token is not to be used for accounting purposes.

I added a "end-to-end" example in the beginning.

docs/book/operator-guide/auth-publisher.md
giac-mysten: I added a Rust-based example. @yanganto do you know what the standard is for `js` implementations?

docs/book/operator-guide/auth-publisher.md
giac-mysten: Added a bit of information. 
Currently the publisher does not support metering the use of WAL, if not indirectly (by limiting size and epochs).

Again, the JWT is not an accounting tool, but an authorization mechanism. 
So the backend should decide if the upload is too costly before issuing the JWT.

docs/book/operator-guide/auth-publisher.md
yanganto: It is common and also possible for a user to have a local publisher for himself and use browser as a UI, so every upload is from the wallet owner and we do not need extra validation here.

It is really common to have a local server and a local browser app in different domain.

For example, 
[ungit](https://www.youtube.com/watch?v=hkBVAi3oKvo) for git.
[ceph-web](https://github.com/tobegit3hub/ceph-web) for distributing storage system.

I do not show a strong application now, but it is really common, and it is good to keep this flexible and community can easier to leverage this to build a web UI for local usage.

docs/book/operator-guide/auth-publisher.md
yanganto: I also have working backend service with Rust, and have no example with `js`.  
It is good to use the same struct with Rust, and also good in performance.  

Besides, it always comes out a brand new package with different syntax or idea but working in the same thing in `js` domain, so I think we do not need to provide something in js, it always changes.  If something need, it maybe only Vanilla JS, and only this is standard.

yanganto (approved):

docs/book/operator-guide/auth-publisher.md
giac-mysten: thank you! added further clarification.

PR #1906, opened 2025-04-07
fix: simplify wording for gas errors

wbbradley:
## Description

Clarify the insufficient Sui error messages to say SUI, instead of "gas."

## Test plan

CI Pipeline.


karlwuest (approved):
lgtm!
Suggestion: Let's make the analogous change for NoCompatiblePaymentCoin and WAL as well

PR #1909, opened 2025-04-07
fix: Limit fallback for NotFound checkpoint errors

sadhansood:
## Description

Limit fallback for `NotFound` errors to only pruned checkpoints. Once we have a way to estimate whether a FN is slow or lagging, we will fall back for recent checkpoints as well.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -1202,6 +1202,19 @@ impl From<tonic::Status> for RetriableClientError {
     }
 }
 
+impl RetriableClientError {
+    fn is_eligible_for_fallback(&self, next_checkpoint: u64) -> bool {
+        match self {
+            Self::RpcError(rpc_error) if rpc_error.status.code() == tonic::Code::NotFound => {
+                rpc_error
+                    .checkpoint_seq_num
+                    .is_some_and(|height| next_checkpoint < height)
+            }
+            _ => true,
+        }
+    }
+}
+
 /// Error type for RPC operations
 #[derive(Error, Debug)]
 pub struct CheckpointRpcError {
@@ -1451,6 +1464,21 @@ impl RetriableRpcClient {
             });
             return Err(error);
         };
+
+        if !error.is_eligible_for_fallback(sequence_number) {
+            tracing::debug!(
+                "primary client error while fetching checkpoint is not eligible for fallback"
+            );
+            self.metrics.as_ref().inspect(|metrics| {
+                metrics.record_rpc_latency(
+                    "get_full_checkpoint",
+                    self.client.get_current_client_name(),
+                    "failure",
+                    start_time.elapsed(),
+                )
+            });
+            return Err(error);
+        }
 
         let fallback_start_time = Instant::now();
         let result = self


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1452,6 +1465,21 @@ impl RetriableRpcClient {
...
+            );
+            self.metrics.as_ref().inspect(|metrics| {
+                metrics.record_rpc_latency(
+                    "get_full_checkpoint",
+                    self.client.get_current_client_name(),
+                    "failure",
+                    start_time.elapsed(),
+                )
+            });
+            return Err(error);
```
liquid-helium, line 1480: Question: will this error fail the download worker, and eventually we will start more workers to retry?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1202,6 +1202,19 @@ impl From<tonic::Status> for RetriableClientError {
...
 
+impl RetriableClientError {
+    fn is_eligible_for_fallback(&self, next_checkpoint: u64) -> bool {
+        match self {
+            Self::RpcError(rpc_error) if rpc_error.status.code() == tonic::Code::NotFound => {
+                rpc_error
+                    .checkpoint_seq_num
+                    .is_some_and(|height| next_checkpoint < height)
+            }
+            _ => true,
```
halfprice, line 1213: ~~Question: should default be `false`?~~

Reading it closely, this change basically means `we will always do fallback except NotFound with checkpoint being caught up`, yes? I try to see where does `missing events` use fallback, and it is here.

Is this intended? Asking because the original idea of the PR is to `fallback only for two scenarios for now 1> missing events error 2> pruned checkpoints`.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1202,6 +1202,19 @@ impl From<tonic::Status> for RetriableClientError {
     }
 }
 
+impl RetriableClientError {
+    fn is_eligible_for_fallback(&self, next_checkpoint: u64) -> bool {
+        match self {
+            Self::RpcError(rpc_error) if rpc_error.status.code() == tonic::Code::NotFound => {
+                rpc_error
+                    .checkpoint_seq_num
```
halfprice, line 1210: question: checkpoint_seq_num is the same as the next_checkpoint, is it? Should we use `rpc_error.status.checkpoint_height()` here?

crates/walrus-sui/src/client/retry_client.rs
liquid-helium: in the case of missing events, the symptom will be the same? the error is `NotFound`, and the full-node has seen events with higher seq?

sadhansood: In case of  missing events error, the error code is `InternalError`, so it will go to fallback

crates/walrus-sui/src/client/retry_client.rs
sadhansood: `checkpoint_height()` is a function on status? I am not sure, my rust analyzer is so slow - I can't even tell :(. In any case, i think the end result is the same - so will keep it

crates/walrus-sui/src/client/retry_client.rs
sadhansood: @halfprice yes, we want to fallback for all errors except for most recent checkpoints (since we are not sure if they have been produced or not yet)

crates/walrus-sui/src/client/retry_client.rs
sadhansood: Yes, it will fail the worker and worker will just retry this call with some backoff again

crates/walrus-sui/src/client/retry_client.rs
halfprice: My main question is, `rpc_error. checkpoint_seq_num` is actually the checkpoint height returned from the fullnode?

It seems that [the rpc_error.checkpoint_seq_num is always set to be the checkpoint number that this request tries to fetch](https://github.com/MystenLabs/walrus/blob/f3b8641456b0a56e644fa82c90bc42f6c46d8330/crates/walrus-sui/src/client/retry_client.rs#L1352), which makes this check here always a `false`. Am I reading the code correctly?

PR #1911, opened 2025-04-07
fix: Fix the checkpoint height computation

sadhansood:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -29,7 +29,7 @@ use reqwest::Url;
 use serde::{de::DeserializeOwned, Serialize};
 #[cfg(msim)]
 use sui_macros::fail_point_if;
-use sui_rpc_api::Client as RpcClient;
+use sui_rpc_api::{client::ResponseExt, Client as RpcClient};
 use sui_sdk::{
     apis::{EventApi, GovernanceApi},
     error::SuiRpcResult,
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1207,7 +1207,8 @@ impl RetriableClientError {
         match self {
             Self::RpcError(rpc_error) if rpc_error.status.code() == tonic::Code::NotFound => {
                 rpc_error
-                    .checkpoint_seq_num
+                    .status
```
liquid-helium, line 1210: nit: is it possible to set checkpoint_seq_num [here](https://github.com/MystenLabs/walrus/blob/9834cef47989a1aac0948394c8bfa55dfa64469b/crates/walrus-sui/src/client/retry_client.rs#L1238-L1245)

halfprice (approved):
LGTM! Thanks for the quick fix @sadhansood . The DB config change included in this PR is also intended? It looks ok to me, just want to make sure they are not something WIP.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: yeah sure i can add it to the `CheckpointRpcError` struct. Good call

sadhansood:
> LGTM! Thanks for the quick fix @sadhansood . The DB config change included in this PR is also intended? It looks ok to me, just want to make sure they are not something WIP.
Thanks @halfprice! Sorry, i will rebase - the other files were for a different PR.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: Actually, i think it probably is not super useful to add it here. Since, it can just be derived from `status` itself which we are storing here

PR #1914, opened 2025-04-07
refactor: move most of walrus_service::client into walrus-sdk

wbbradley:
## Description

This is the first big step in extracting CLI logic from the walrus_service crate, down into the walrus_sdk crate.

## Test plan

Pipeline, and manual local-testbed testing.


crates/walrus-sdk/src/config.rs
```diff
@@ -0,0 +1,175 @@
...
+        "using Walrus configuration from '{}' with {} context",
+        path.display(),
+        context.map_or("default".to_string(), |c| format!("'{}'", c))
+    );
+    Ok(config)
+}
+
+/// Config for the client.
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+pub struct Config {
```
halfprice: Suggestion: I think we may want to rename this to something like `ClientConfig`. `Config` is too general and caused some confusion before. Maybe add a TODO here?

crates/walrus-sdk/src/config/client_config.rs
```diff
@@ -0,0 +1,470 @@
...
+    }
+
+    /// This corresponds to 100Mb, i.e., 1 second on a 100 Mbps connection.
+    pub fn max_data_in_flight() -> usize {
+        12_500_000
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
halfprice: suggestion: can we move these tests to `src/config.rs`? I think these are testing the `[Client]Config`

crates/walrus-sdk/src/config/client_config.rs
```diff
@@ -0,0 +1,470 @@
...
+    pub max_concurrent_writes: usize,
+    /// The maximum number of concurrent sliver reads from the storage nodes.
+    pub max_concurrent_sliver_reads: usize,
+    /// The maximum number of concurrent metadata reads from the storage nodes.
+    pub max_concurrent_metadata_reads: usize,
+    /// The maximum number of concurrent status reads from the storage nodes.
+    pub max_concurrent_status_reads: usize,
+    /// The maximum amount of data in flight (in bytes) for the client.
+    pub max_data_in_flight: usize,
+}
```
halfprice: I have a few suggestions:
- (No strong opinion on this) I feel `MultiConfig` can be placed near [Client]Config in src/config.rs, since it's small and also more of an implementation detail of [Client]Config. We can also remove `pub` after moving there.
- (Not caused by your PR, but since I already see it), ExchangeObjectConfig is no longer used and can be removed.
- (if the first two are adopted) If CommunicationLimits is the only struct left in this file, maybe consider renaming this file to something like src/config/communication_limits.rs?

halfprice (commented):
Thanks @wbbradley and this is a great way to initiating the sdk work! Appreciate the effort to make the refactoring "minimum" for easier review. I just have some nits and should be quick to address.

crates/walrus-sdk/src/config.rs
wbbradley: I'll just go ahead and make this change.

crates/walrus-sdk/src/config/client_config.rs
wbbradley: Sure.

crates/walrus-sdk/src/config/client_config.rs
wbbradley: I lumped the CommunicationLimits into communication_config.rs, seems to make sense there. I also renamed MultiConfig to MultiClientConfig, since that's more inline with the rename to ClientConfig. Deleted ExchangeObjectConfig.

crates/walrus-sdk/client_config_example.yaml
mlegner, line 1: I haven't really looked at the code changes, but could you update https://github.com/MystenLabs/walrus/blob/main/.github/actions/diffs/action.yml#L66 in accordance with this move?

wbbradley: Nice catch. Thanks.

crates/walrus-sdk/src/utils.rs
```diff
@@ -3,16 +3,35 @@
...
-use futures::{stream::FuturesUnordered, Future, StreamExt};
+use anyhow::{Context as _, Result};
+use futures::{stream::FuturesUnordered, StreamExt};
+use indicatif::{ProgressBar, ProgressStyle};
+use serde::de::DeserializeOwned;
 use tokio::time;
 use tracing::Level;
 
+/// Load the config from a YAML file located at the provided path.
+pub fn load_from_yaml<P: AsRef<Path>, T: DeserializeOwned>(path: P) -> anyhow::Result<T> {
```
halfprice, line 22: suggestion: add a TODO that this should probably go to walrus-utils crate

crates/walrus-sdk/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+pub mod error;
+pub mod store_when;
+/// Utilities for the Walrus SDK.
+pub mod utils;
+
+pub use sui_types::event::EventID;
+pub use walrus_sui as sui;
+
+/// Format the event ID as the transaction digest and the sequence number.
+pub fn format_event_id(event_id: &EventID) -> String {
```
halfprice, line 19: nit: this inside lib.rs feels weird. Can this go to walrus-utils? If so, just add a TODO here and address it later.

crates/walrus-sdk/src/active_committees.rs
liquid-helium, line 1: Does it make sense to move these 'common' structs into a separate crate, and let client and server import these common structs, (could in another PR).

crates/walrus-sdk/src/client/communication.rs
liquid-helium, line 1: Same for this one.

halfprice (approved):
@wbbradley Thank you very much for preparing the code change so quickly! LGTM 💯

While reading the PR, one TODO I think we should do is to create a test/mini example app to use the SDK and make sure that there is no dependency to walrus-service crate. Can you create an issue to track this?

crates/walrus-sdk/src/utils.rs
wbbradley: Added a TODO. https://linear.app/mysten-labs/issue/WAL-764/move-load-from-yaml-into-walrus-utils

crates/walrus-sdk/src/lib.rs
wbbradley: walrus-utils does not import EventID (sui-types). Having this here is a bit strange, I agree. I'll leave here for now, and we can easily move it later if this placement becomes problematic.

wbbradley:
>  there are a bunch of newly added files, are they moved from the original file? It is hard to tell the changes.
Sorry they are not 1:1. I agree it's frustrating that it's not so direct. Some are almost exact copies of old files, some were simply moving structs/enums out, etc.. Hard to categorize everything, but bear in mind that we can change this around later, as needed. This is just to get the code out of the "service" layer for now.

wbbradley:
> @wbbradley Thank you very much for preparing the code change so quickly! LGTM 💯
> 
> While reading the PR, one TODO I think we should do is to create a test/mini example app to use the SDK and make sure that there is no dependency to walrus-service crate. Can you create an issue to track this?

Once the overall SDK work is complete, the Walrus Client CLI will be an example user, as well as various other places in the codebase. I'll add a task to track documentation of the SDK, which would include examples. In fact, we can use the `cargo --example` magic to make it nice.

crates/walrus-sdk/src/client/communication.rs
wbbradley: I've created a Linear task to track this. https://linear.app/mysten-labs/issue/WAL-766/consider-moving-activecommittees-and-nodecommunicationfactory-into

crates/walrus-sdk/src/active_committees.rs
wbbradley: Tracked in https://linear.app/mysten-labs/issue/WAL-766/consider-moving-activecommittees-and-nodecommunicationfactory-into

PR #1919, opened 2025-04-08
fix: shard transfer resets timer for switching to recovery after it makes progress

halfprice:
## Description

Shard transfer is a long running process and even though it may encounter error (e.g. the source server restart),
it should continue using shard transfer as long as it can continue make progress. The shard transfer should only
switch to shard recovery if it encounters error for extended period of time without any progress.

## Test plan

Integration tests with specific scenario created.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -34,7 +34,10 @@ enum SyncShardResult {
     /// The shard sync is not finished and should be retried after a backoff.
     /// The first bool indicates whether to directly recover the shard instead of using shard sync.
     /// The second bool indicates whether the shard sync made progress.
-    RetryAfterBackoff(bool, bool),
+    RetryAfterBackoff {
+        force_recovery: bool,
+        shard_sync_made_progress: bool,
+    },
     /// The shard sync contains errors and should be stopped.
     Failed,
 }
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -32,8 +32,9 @@ enum SyncShardResult {
     /// The shard sync finished successfully.
     Success,
     /// The shard sync is not finished and should be retried after a backoff.
-    /// The bool indicates whether to directly recover the shard instead of using shard sync.
-    RetryAfterBackoff(bool),
+    /// The first bool indicates whether to directly recover the shard instead of using shard sync.
+    /// The second bool indicates whether the shard sync made progress.
+    RetryAfterBackoff(bool, bool),
```
jpcsmith: Suggestion: Since there are two values here, perhaps use a named enum variant.

```suggestion
    /// The first bool indicates whether to directly recover the shard instead of using shard sync.
    /// The second bool indicates whether the shard sync made progress.
    RetryAfterBackoff{force_recovery: bool, shard_sync_made_progress: bool},
````

jpcsmith (approved):

crates/walrus-service/src/node/shard_sync.rs
halfprice: Looks better! Thanks!

PR #1921, opened 2025-04-08
feat: update docs tools for walrus (ref. awesome-walrus)

nidz-the-fact:
## Description

Add more tools on docs for Walrus. 
ref. https://github.com/MystenLabs/awesome-walrus

## Test plan

Yes, I tested it.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI:


mlegner:
Thanks for the contribution, but I don't think we should duplicate that information, as it will be tedious to keep this in sync.

PR #1932, opened 2025-04-09
fix: certified blob recovery should only recover the shard that is assigned to the node in the epoch that the blob is certified

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


halfprice:
Hi @jpcsmith @mlegner , hope this PR can catch the 1.21 release which will be cut this Thursday.

The build failure is the CI issue, which should be addressed by #1930

crates/walrus-service/src/node.rs
```diff
@@ -1714,16 +1717,42 @@ impl StorageNodeInner {
         self.node_capability
     }
 
-    pub(crate) fn owned_shards(&self) -> Vec<ShardIndex> {
-        self.committee_service
-            .active_committees()
-            .current_committee()
-            .shards_for_node_public_key(self.public_key())
-            .to_vec()
+    pub(crate) fn owned_shards(&self, epoch: Epoch) -> Vec<ShardIndex> {
```
mlegner: Suggestion/nit: In many places we still use the previous behavior, namely looking for the currently owned shards. Could we keep the existing function under the name `currently_owned_shards` or similar? Same for the `is_stored_at_all_shards` function.

crates/walrus-service/src/node.rs
```diff
@@ -1714,16 +1717,42 @@ impl StorageNodeInner {
...
+                .previous_committee()
+                .expect("previous committee should be set")
+                .shards_for_node_public_key(self.public_key())
+                .to_vec()
+        } else {
+            // Otherwise, return the shard assignment based on the current committee.
+            //
+            // Here we don't validate the epoch given that there are already multiple places
+            // in the node logic where storage node cannot be more than 2 epoch lagging. If so,
+            // it'll enter recovery mode.
```
mlegner: Important: I understand the reasoning here, but this could be a bit dangerous when someone just looks at the function name and signature. Could we instead return an error or panic in cases where the event input is not the current or previous epoch? Also, can we describe this behavior in the docstring of the function?

mlegner (commented):
Thanks a lot for addressing this important issue, @halfprice. This generally LGTM, just two comments.

crates/walrus-service/src/node.rs
halfprice: Yeah, I added `owned_shards_at_<latest_>epoch` and `is_stored_at_all_shards_at_<latest_>epoch`. I feel `current` is a bit relative since it can be the current latest epoch or the current epoch the event is from.

crates/walrus-service/src/node.rs
halfprice: I share your concern, and I was particularly wanted to avoid adding a panic. But I do agree that the old function semantics has implicit assumption. What I changed to is to return error if the epoch is neither the current or the previous epoch. The only callsite that will fail is when trying to recover assigned slivers. Currently the node panic and restarting will make the node into recovery mode.

halfprice:
Thanks a lot @mlegner for the review! I agree with all your comments and addressed them accordingly. PTAL!

mlegner (approved):
Thanks a lot, @halfprice! 🙏

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -462,10 +462,33 @@ impl BlobSynchronizer {
...
+        //   - Since the blob is certified in an earlier epoch, it is this node's responsibility to
+        //     hold, recover, and transfer the shard to the new owner.
+        //
+        // If the node is severally lagging behind and the certified_epoch is 2 epochs older,
+        // this function panics since no shard assignment info is found. Upon restarting the node,
+        // the node will enter recovery mode until catching up with all the events and start
+        // recovering all the missing blobs.
+        let futures_iter = this
+            .node
+            .owned_shards_at_epoch(this.node.current_event_epoch())
```
halfprice, line 480: Hi @mlegner , simtest revealed an issue that the previous version of using this.certified_epoch (the event's epoch) here is actually incorrect. This is because there can be a high lag between the certified event is processed, and the start of blob recovery. So it is not guaranteed that when the blob sync task is scheduled to run, the node is still at certified_event.epoch.

The correct fix here is to use the epoch the latest event is in. And in order to do that, we need to track it. I added latest_event_epoch in StorageNodeInner, and updated it when encounter epoch change start event.

Let me know if you see any problem of current solution.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -462,10 +462,33 @@ impl BlobSynchronizer {
             .await
             .expect("database operations should not fail");
 
-        let futures_iter = this.node.owned_shards().into_iter().map(|shard| {
-            this.clone()
-                .recover_slivers_for_shard(shared_metadata.clone(), shard)
-        });
+        // Note that we only need to recover the slivers in the shards that are assigned to the
+        // node at the time of the blob certification. Due to slow event processing, the contract
```
mlegner: Nit: This comment is no longer fully in sync with the code.

crates/walrus-service/src/node/blob_sync.rs
mlegner: Thanks a lot for catching this! 🙏 

Just wondering about a few corner cases:
* Is it a problem if the `current_event_epoch` is updated before we create new shard storage?
* What happens if we later lose one of the shards for which we now try to recover slivers?

I believe those are ok, but just want to make sure.

mlegner (approved):

crates/walrus-service/src/node/blob_sync.rs
halfprice: Done.

crates/walrus-service/src/node/blob_sync.rs
halfprice: Thanks for the review @mlegner !

> Is it a problem if the current_event_epoch is updated before we create new shard storage?

Yes, because after updating the current_event_epoch, the blob sync will use the new config, but we haven't created the storage yet.

> What happens if we later lose one of the shards for which we now try to recover slivers?

Recovering shards always use the "current" view of the walrus system. So shards cannot change when blob sync starts. However, are you saying if a blob sync takes so long that during it's syncing, a shard moves out, and then
deleted? Yeah, I think this is an issue, but only if a blob sync takes longer than 2 epoch duration, so which may be
rare.

PR #1933, opened 2025-04-09
Nix flake

poelzi:
## Description

Add nix flake for deterministic builds of walrus packages.
Added docker image generation
When https://github.com/NixOS/nixpkgs/pull/387337 lands on master, I will change nixpkgs to unstable again.

Nix is save against supply chain attacks and allows builds to be reproducible, even years later.
Nix flakes allows users to properly compose development environments with composition and overrides, making the dev and deployment experience much smoother and faster, including rollbacks.

Next step is to add NixOS modules for config and service generation

Requirements:
* [nix](https://nixos.org/download/) package manager (works on all linux systems)
* optional: direnv for best developer experience

## Test plan

Did local testing, I would like add workflows as well

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -50,6 +50,7 @@ pub struct App {
     /// 2. If the environment variable `XDG_CONFIG_HOME` is set, in `$XDG_CONFIG_HOME/walrus/`.
     /// 3. In `~/.config/walrus/`.
     /// 4. In `~/.walrus/`.
+    /// 5. The environment variable `WALRUS_CONFIG` as complete path
     // NB: Keep this in sync with `crate::cli`.
     #[arg(long, verbatim_doc_comment, global = true)]
     #[serde(
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,361 @@
+{
+  inputs = {
+    fenix = {
+      url = "github:nix-community/fenix";
+      inputs.nixpkgs.follows = "nixpkgs";
+    };
+
+    #nixpkgs.url = "nixpkgs/nixos-unstable";
+    # we need to wait for https://github.com/NixOS/nixpkgs/pull/387337
+    # nixpkgs.url = "github:TomaSajt/nixpkgs?ref=fetch-cargo-vendor-dup";
+    # rebased version on master
+    nixpkgs.url = "github:poelzi/nixpkgs?ref=fetch-cargo-vendor-dup";
+
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs =
+    {
+      self,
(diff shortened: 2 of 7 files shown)


bmwill:
Love seeing more folks interested in deterministic builds. 

We've been working on getting deterministic build infrastructure setup for the suite of applications that Mysten builds and maintains and we are close to integrating fully deterministic builds into our release pipelines. I'm hoping we'll be able to do so in the coming weeks. We do not use Nix and likely don't plan on adding support for it, but instead are leveraging [StageX](https://codeberg.org/stagex/stagex) for this. Here you can find a [comparison](https://codeberg.org/stagex/stagex#comparison) between StageX and Nix.

poelzi:
@mlegner @bmwill I can understand the reasoning and think stagex images are better for production images.
The provided flake is more for development. It is very hard to provide fully functional, easy to use development environments that are correctly enforced on each machine. You easily clash with local tools in different versions, tools are missing or tools need patching, ...
Nix flakes fixes this and allows downstream compos-ability. Everybody who ever used nix flakes will sign this.
SUI wants to be developers driven, without a nix flake, it is super hard to use walrus on NixOS systems, or projects to use walrus.
The advantage of the produced images is, that they reuse the nix store and the cache server provided. So, you can rebuild them in a few seconds. Long term, I want nix flakes to be stored on sui/walrus, making it super easy and fast to provide composable environments.
For the production image, the full bootstrapped stagex image is of course better.

Please reopen

flake.nix
mlegner, line 1: Could we move the nix-related files to a subdirectory and add a README file that contains a disclaimer that this is a community contribution and not maintained by the Walrus core team?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -50,6 +50,7 @@ pub struct App {
     /// 2. If the environment variable `XDG_CONFIG_HOME` is set, in `$XDG_CONFIG_HOME/walrus/`.
     /// 3. In `~/.config/walrus/`.
     /// 4. In `~/.walrus/`.
+    /// 5. The environment variable `WALRUS_CONFIG` as complete path
```
mlegner, line 53: Suggestion: I think this should have higher priority than the default paths: It should be checked immediately after the explicitly specified path. (In particular as it's supposed to be the full path (including the filename) AFAIU as opposed to the default directories.)

mlegner (commented):
Thanks for the feedback, @poelzi.

I wonder if this might be better suited for a separate repository and listed on https://github.com/MystenLabs/awesome-walrus. WDYT?

If we include it in this repository, I would like to make it clear that this is not maintained by the Walrus core team.

PR #1934, opened 2025-04-09
fix(node): adjust default params for shard sync

mlegner:
## Description

This increases several default parameters for shard transfer and recovery to make it more peformant.

## Test plan

Most of these parameters were already tested in the recent epoch change on Mainnet.

---

## Release notes

- [x] Storage node: Optimized default parameters for shard transfer and recovery.


halfprice (approved):
Thanks @mlegner !

PR #1937, opened 2025-04-09
feat: Add db tool command to print init state

sadhansood:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/dbtool.rs
+++ b/crates/walrus-service/src/node/dbtool.rs
@@ -20,6 +20,7 @@ use walrus_core::{
     ShardIndex,
 };
 
+use super::events::event_processor::INIT_STATE;
 use crate::node::{
     events::{
         event_blob_writer::{
@@ -32,7 +33,8 @@ use crate::node::{
             FailedToAttestEventBlobMetadata,
             PendingEventBlobMetadata,
         },
-        event_processor::event_store_cf_name,
+        event_processor::EVENT_STORE,
+        InitState,
         PositionedStreamEvent,
     },
     storage::{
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -86,17 +86,17 @@ use crate::{
 };
 
 /// The name of the checkpoint store.
-const CHECKPOINT_STORE: &str = "checkpoint_store";
+pub const CHECKPOINT_STORE: &str = "checkpoint_store";
```
halfprice: nit: event store is exposed using `pub(crate) fn event_store_cf_name() -> &'static str {}` below. Maybe do the similar things for other CFs for consistency?

crates/walrus-service/src/node/dbtool.rs
```diff
@@ -629,6 +653,28 @@ fn read_secondary_slivers(
...
+        return Ok(());
+    };
+
+    let iter = db.iterator_cf(&cf, rocksdb::IteratorMode::Start);
+    for result in iter {
+        let (key, value) = result?;
+        let init_state: InitState = bcs::from_bytes(&value)?;
+        let config = bincode::DefaultOptions::new()
+            .with_big_endian()
+            .with_fixint_encoding();
```
halfprice: nit: I think we can move `config` out of the for loop to only initialize it once?

halfprice (approved):
LGTM! Just a few nits

crates/walrus-service/src/node/events/event_processor.rs
sadhansood: That function seems a bit unnecessary tbh. I am not totally certain why adding another layer of encapsulation is better over just making it a pub const. If there isn't a strong reason, we can probably remove that other function.

halfprice: yeah. removing it is ok too. Let's just keep them consistent. (and maybe also change here to pub(crate)).

liquid-helium: There won't be any functional difference, I added them just to make it look nicer (personal preference) 😄 
when making that change, the major goal is to put all constants in a single place, so that it would be easier to find.

sadhansood: I see, thanks for adding context @liquid-helium. Do you like [this](https://github.com/MystenLabs/walrus/pull/1940) approach of putting constants together in a mod names constants and then importing them as `use event_processor::constants::{self as event_processor_constants},` which makes it clear where these are being imported from.

PR #1946, opened 2025-04-10
docs: update all aggregators and publishers

mlegner:
## Description

Update all aggregator and publisher URLs on Mainnet and Testnet. Also include the operator name.

## Test plan

Downloaded/uploaded an example blob with all instances. Result can be inspected in the preview.


nikos-terzo (approved):
Great! I wonder whether we should use name as key instead of the url, as it is more readable, and I guess it should also be unique (no 2 operators with the same name should exist).
We can investigate this in a different PR if it breaks the listing of the HTTP endpoints in documentation.

PR #1948, opened 2025-04-10
fix(client): prevent panic in genesis epoch

mlegner:
## Description

Without this change, the panic a few lines below can be triggered in the genesis epoch (which results in a cryptic error output):

```
thread 'main' panicked at crates/walrus-service/src/client/communication/factory.rs:302:13:
internal error: entered unreachable code: `all()` guarantees at least 1 result and all results are errors
```

## Test plan

None.


giac-mysten:
@mlegner i rebased to resolve conflicts, hope you don't mind!

PR #1953, opened 2025-04-11
fix: add CORS headers to server endponints

giac-mysten:
## Description

Adds CORS headers to the node endpoints, and improves the CORS header on the daemon endpoints.
Depends on #1929 , because it resolves a RUSTSEC advisory in tokio that would have otherwise been added to the `deny.toml` exceptions.

Closes WAL-748

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon/routes.rs
+++ b/crates/walrus-service/src/client/daemon/routes.rs
@@ -1,7 +1,7 @@
 // Copyright (c) Walrus Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::HashSet, str::FromStr, sync::Arc};
+use std::{collections::HashSet, str::FromStr, sync::Arc, time::Duration};
 
 use anyhow::anyhow;
 use axum::{
@@ -16,18 +16,13 @@ use axum_extra::{
     TypedHeader,
 };
 use jsonwebtoken::{DecodingKey, Validation};
-use reqwest::header::{
-    ACCESS_CONTROL_ALLOW_HEADERS,
-    ACCESS_CONTROL_ALLOW_METHODS,
-    ACCESS_CONTROL_ALLOW_ORIGIN,
-    ACCESS_CONTROL_MAX_AGE,
-    CACHE_CONTROL,
-    CONTENT_TYPE,
-    ETAG,
-    X_CONTENT_TYPE_OPTIONS,
+use reqwest::{
+    header::{CACHE_CONTROL, CONTENT_TYPE, ETAG, X_CONTENT_TYPE_OPTIONS},
+    Method,
 };
 use serde::Deserialize;
 use sui_types::base_types::{ObjectID, SuiAddress};
+use tower_http::cors::{Any, CorsLayer};
 use tracing::Level;
 use utoipa::IntoParams;
 use walrus_core::{BlobId, EncodingType, EpochCount};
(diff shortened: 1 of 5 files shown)


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -392,14 +380,13 @@ impl From<ClientError> for StoreBlobError {
...
-        (ACCESS_CONTROL_ALLOW_ORIGIN, "*"),
-        (ACCESS_CONTROL_ALLOW_METHODS, "PUT, OPTIONS"),
-        (ACCESS_CONTROL_MAX_AGE, "86400"),
-        (ACCESS_CONTROL_ALLOW_HEADERS, "*"),
-    ]
+/// Returns a `CorsLayer` for the blob store endpoint.
+pub(super) fn store_blob_cors_layer() -> CorsLayer {
+    CorsLayer::new()
+        .allow_origin(Any)
+        .allow_methods([Method::GET, Method::PUT])
```
jpcsmith: Question: I see the OPTIONS was replaced with GET. Is this because OPTIONS was never used? But could this instead take the methods as an argument to the function, such that unsupported methods aren't advertised?

crates/walrus-service/src/node/server.rs
```diff
@@ -388,6 +393,15 @@ where
...
             .route(routes::SYNC_SHARD_ENDPOINT, post(routes::sync_shard))
     }
+
+    /// Returns the CORS leayer for the server.
+    // NOTE: The allowed methods should be kept in sync with the methods used in the routes.
+    fn cors_layer() -> CorsLayer {
+        CorsLayer::new()
+            .allow_origin(Any)
+            .allow_methods([Method::GET, Method::PUT, Method::POST, Method::HEAD])
+            .allow_headers(Any)
```
jpcsmith: Question: Same question as above, should we instead specify the `allow_methods` as an argument to `cors_layer` as an argument?

jpcsmith (approved):

crates/walrus-service/src/client/daemon/routes.rs
giac-mysten: good point, definitely GET should not be there. I'll also add OPTIONS to the storage node.

mlegner:
> Depends on https://github.com/MystenLabs/walrus/pull/1929 , because it resolves a RUSTSEC advisory in tokio that would have otherwise been added to the deny.toml exceptions.

@giac-mysten The `cargo-deny` CI job for PRs does not consider advisories (the command is `cargo deny check bans licenses sources`), so this should work without upgrading tokio.

giac-mysten:
> > Depends on #1929 , because it resolves a RUSTSEC advisory in tokio that would have otherwise been added to the deny.toml exceptions.
> 
> @giac-mysten The `cargo-deny` CI job for PRs does not consider advisories (the command is `cargo deny check bans licenses sources`), so this should work without upgrading tokio.

@mlegner just to be clear, i'm not upgrading tokio in this PR, i'm just based off of @wbbradley 's one . Are you saying i should point directly to main and merge without upgrading tokio?

jpcsmith (approved):

mlegner:
> Are you saying i should point directly to main and merge without upgrading tokio?

Yes. 🙂

PR #1958, opened 2025-04-11
chore: add mainnet contracts

karlwuest:
## Description

adds the mainnet contracts to a separate directory, s.t. `contracts` can be used for development
also adds a github action to add a warning comment for all PRs that modify the `mainnet-contracts` directory

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


mlegner (approved):
Thanks a lot! Ran a diff, it seems it is in fact simply a copy of the `contracts` directory without the wal_exchange 😁:

```
$ diff -r contracts mainnet-contracts
Only in contracts: .DS_Store
Only in contracts/subsidies: build
Only in contracts/wal: build
Only in contracts: wal_exchange
Only in contracts/walrus: build
```

Also 🤯:
image

PR #1959, opened 2025-04-11
test: test contract upgrades from previous contract version

karlwuest:
## Description

closes WAL-653
closes WAL-654

## Test plan

- existing tests
- PR modifies the upgrade test to actually test the upgrade from the previous contract version to the development version

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -11,7 +11,7 @@ use std::sync::{
 use std::{
     collections::{HashMap, HashSet},
     num::NonZeroU16,
-    path::{Path, PathBuf},
+    path::PathBuf,
     time::Duration,
 };
 
@@ -66,6 +66,8 @@ use walrus_sui::{
         SuiContractClient,
         UpgradeType,
     },
+    system_setup::copy_recursively,
+    test_utils::system_setup::{development_contract_dir, testnet_contract_dir},
     types::{
         move_errors::{MoveExecutionError, RawMoveError},
         move_structs::{BlobAttribute, SharedBlob, Subsidies},
@@ -1567,16 +1569,31 @@ async fn test_quorum_contract_upgrade() -> TestResult {
         test_cluster::E2eTestSetupBuilder::new()
             .with_deploy_directory(deploy_dir.path().to_path_buf())
             .with_delegate_governance_to_admin_wallet()
+            .with_contract_directory(testnet_contract_dir()?)
             .build()
             .await?;
 
-    // TODO(WAL-654): once mainnet upgrades follow testnet, upgrade from testnet-contracts instead
+    let previous_version = client
+        .as_ref()
+        .sui_client()
+        .read_client()
+        .system_object_version()
+        .await?;
+
+    // Copy new contracts to fresh directory
+    let upgrade_dir = TempDir::new()?;
+    copy_recursively(development_contract_dir()?, upgrade_dir.path())?;
+
+    // Copy Move.lock files of walrus contract and dependencies to new directory
+    for contract in ["wal", "walrus"] {
+        std::fs::copy(
+            deploy_dir.path().join(contract).join("Move.lock"),
+            upgrade_dir.path().join(contract).join("Move.lock"),
+        )?;
+    }
+
     // Change the version in the contracts
-    let walrus_package_path = deploy_dir.path().join("walrus");
-    let staking_path = walrus_package_path.join("sources/staking.move");
-    let system_path = walrus_package_path.join("sources/system.move");
-    replace_version(&staking_path)?;
-    replace_version(&system_path)?;
+    let walrus_package_path = upgrade_dir.path().join("walrus");
 
     // Vote for the upgrade
     // We can vote on behalf of all nodes from the client wallet since the client
@@ -1626,7 +1643,7 @@ async fn test_quorum_contract_upgrade() -> TestResult {
             .read_client()
             .system_object_version()
             .await?,
-        2
+        previous_version + 1
     );
 
     // Store a blob after the upgrade to check if everything works after the upgrade.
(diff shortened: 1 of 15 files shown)


karlwuest:
>after removing sui as explicit dependency, the simtests fail locally (CI is still running atm), do you have any idea why?

It seems that this is from initializing the static innitializers (with `SIMTEST_STATIC_INIT_MOVE`). I'm still not quite sure what this is used for, but the code that builds the contracts there (in `sui-proc-macros`) does not do the automatic dependency resolution.

It seems to work if we use an empty move package. I added a package `simtest_static_init_move` to the `scripts/simtest` directory for this now.

karlwuest:
@halfprice now the simtests fail in the CI for a different reason (I suspect that they take too long to run?) 😅 
Locally, they work, though. Can you help me figure out how to solve this?

karlwuest:
>@halfprice after removing sui as explicit dependency, the simtests fail locally (CI is still running atm), do you have any idea why? It seems that it doesn't do the automatic dependency resolution in simtests. It works outside of simtests, though.

I rolled back the change to remove sui as explicit dependency. I had initially only included this because I thought it would be a small change. However, it seems that there are too many pieces that break (including the move coverage) and require more changes, so we can revisit this separately (WAL-767).

karlwuest:
@halfprice disabling simtest for the upgrade test solved the issues. I suspect that this was because that test compiles the contracts multiple times (although it also did so before), but I'm still not exactly sure why this also affected the other tests.

Anyway... the PR is finally ready for review now 😅

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -1557,36 +1561,39 @@ async fn test_post_store_action(
...
-            ClientCommunicationConfig::default_for_test(),
-            false,
-            Some(contract_dir.path().to_path_buf()),
-            true,
-            None,
-        )
+        test_cluster::E2eTestSetupBuilder::new()
+            .with_deploy_directory(deploy_dir.path().to_path_buf())
+            .with_delegate_governance_to_admin_wallet()
+            .with_contract_directory(testnet_contract_dir()?)
```
halfprice, line 1572: Some more thoughts on the simtest failure here:

First of all, if this test does not really need to test concurrency bugs, I think removing it from the simtest list is ok.

Here are my thoughts on the simtest issue without trying to implement a fix myself:

I think the issue is the testnet_contract used here. In testnet contract, it uses sui testnet-v1.45.2, and in all the other contracts, they use testnet-v1.46.0. The simtest setup will use testnet-v1.46.0 to precompile the contract, which is outside of simtest runtime, and in all other tests, all the move libraries are already setup.

But here, it needs to compile using testnet-v1.45.2, which doesn't exist, so it needs to download the move library first, which is a blocking task on the main tokio thread. If this task runs longer than 5 seconds, the watchdog triggers. I saw in CI results that this test tries to clone testnet-v1.45.2 and the following test tries to clone testnet-v1.46.0. But if I remove this test, non of the tests need to clone the sui repo inside the test body.

Another solution other than updating testnet contract to use testnet-v1.46.0, is to move the contract compilation into a blocking thread wrapping by `nondeterministic!()` so that it does not block the main tokio thead.

crates/walrus-service/src/test_utils.rs
```diff
@@ -2237,333 +2237,387 @@ pub mod test_cluster {
...
+            let mut admin_wallet =
+                test_utils::new_wallet_on_sui_test_cluster(sui_cluster.clone()).await?;
 
-        for (i, wallet) in
-            test_utils::create_and_fund_wallets_on_cluster(sui_cluster.clone(), n_nodes)
-                .await?
-                .into_iter()
+            let test_nodes_config = self.test_nodes_config.unwrap_or_else(|| TestNodesConfig {
+                node_weights: vec![1, 2, 3, 3, 4],
+                use_legacy_event_processor: true,
```
halfprice, line 2331: question: should we set the default value to `false` here? I don't think we want to use the legacy event processor anywhere anymore, so I think we should also just remove this option. (I know this is not your change, so maybe just add a TODO for us to revisit here?).

halfprice (approved):
Thanks a lot for the clean up @karlwuest ! LGTM and enjoyed reading it!

crates/walrus-service/src/test_utils.rs
karlwuest: I'll try it with `false` as default and if nothing breaks, I'll add it to this PR. Otherwise, I'll add a TODO.

crates/walrus-e2e-tests/tests/test_client.rs
karlwuest: Thanks, that makes sense!

>if this test does not really need to test concurrency bugs, I think removing it from the simtest list is ok

Yes, I think that we don't need it for this test.

>Another solution other than updating testnet contract to use testnet-v1.46.0, is to move the contract compilation into a blocking thread wrapping by nondeterministic!() so that it does not block the main tokio thead.

Hmmm... the contract compilation should already happen in a blocking tokio task after b55b8c4efdbff79949fbdf20d331b9850f11ffc6, so it shouldn't block the main thread, but that did not help.

crates/walrus-service/src/test_utils.rs
karlwuest: Added a TODO. I tried running simtests and for some, I got this error (cc @sadhansood ):

````
2022-01-03T02:05:55.435737Z ERROR node{id=1 name="client"}:cluster-node{address=127.0.0.1:32820}: walrus_service::node::events::event_blob_writer: crates/walrus-service/src/node/events/event_blob_writer.rs:1235: Unexpected event blob certification error while attesting event blob error=Internal(client error (SendRequest)

Caused by:
    0: connection error
    1: peer 10.10.0.3:9016 hung up) blob_id=BlobId(I07I-UWA5O2oWHSbNcTm-znur0tYaum1Pnkcrvth-Y8)
2022-01-03T02:05:55.435737Z ERROR node{id=1 name="client"}:cluster-node{address=127.0.0.1:32820}: walrus_service::test_utils: crates/walrus-service/src/test_utils.rs:954: node stopped with an error error=No such file or directory (os error 2)
````

PR #1960, opened 2025-04-11
feat: allow looking up shared blobs by their object ID

karlwuest:
## Description

Allows reading a shared blob by object ID.

## Test plan

extended existing test

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


giac-mysten (approved):
Awesome, thanks!

karlwuest:
>@karlwuest Could we add some docs about this, emphasizing that the object ID can be the ID of either a Blob or a SharedBlob?

Sure --> #2005

PR #1962, opened 2025-04-11
fix: Recreate rpc client on extended failure and fallback to archive

sadhansood:
## Description

This PR does two fixes:
1. Recreate the rpc client on 10 minutes worth of continuous failure. Recreating the client is similar to restarting the node
2. Allows rpc to fallback to archive using a exponentially growing window size. This allows rpc to fallback to archive after 1 minute of no success  from any endpoint

## Test plan
Existing tests

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -9,11 +9,16 @@ use std::{
     collections::BTreeMap,
     fmt::Debug,
     future::Future,
+    mem,
     str::FromStr,
-    sync::{atomic::AtomicUsize, Arc},
-    time::Duration,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+    time::{Duration, Instant},
 };
 
+use atomic_time::AtomicInstant;
 use futures::{
     future::{self},
     stream,
@@ -68,7 +73,7 @@ use sui_types::{
     TypeTag,
 };
 use thiserror::Error;
-use tokio::time::Instant;
+use tokio::sync::RwLock;
 use tracing::Level;
 use url::ParseError;
 use walrus_core::ensure;
@@ -187,8 +192,8 @@ impl RetriableRpcError for RetriableClientError {
     fn is_retriable_rpc_error(&self) -> bool {
         match self {
             Self::RpcError(rpc_error) => rpc_error.is_retriable_rpc_error(),
-            Self::RetryableTimeoutError(_) => true,
-            Self::NonRetryableTimeoutError(_) => false,
+            Self::RetryableTimeoutError => true,
+            Self::NonRetryableTimeoutError => false,
             Self::FallbackError(fallback_error) => fallback_error.is_retriable_rpc_error(),
             Self::Other(_) => false,
         }
@@ -1175,11 +1180,11 @@ pub enum RetriableClientError {
 
     /// Timeout error from the primary client.
     #[error("primary RPC timeout")]
-    RetryableTimeoutError(tokio::time::error::Elapsed),
+    RetryableTimeoutError,
 
     /// Non-retryable timeout error from the primary client.
     #[error("primary RPC timeout")]
-    NonRetryableTimeoutError(tokio::time::error::Elapsed),
+    NonRetryableTimeoutError,
 
     /// Error from the fallback client.
     #[error("fallback error: {0}")]
@@ -1190,26 +1195,23 @@ pub enum RetriableClientError {
     Other(#[from] anyhow::Error),
 }
 
-impl From<tokio::time::error::Elapsed> for RetriableClientError {
-    fn from(error: tokio::time::error::Elapsed) -> Self {
-        Self::RetryableTimeoutError(error)
-    }
-}
-
 impl From<tonic::Status> for RetriableClientError {
     fn from(status: tonic::Status) -> Self {
-        Self::RpcError(CheckpointRpcError::from(status))
+        Self::RpcError(CheckpointRpcError::from((status, 0)))
     }
 }
 
 impl RetriableClientError {
-    fn is_eligible_for_fallback(&self, next_checkpoint: u64) -> bool {
+    fn is_eligible_for_fallback(&self, next_checkpoint: u64, window_size: u64) -> bool {
         match self {
             Self::RpcError(rpc_error) if rpc_error.status.code() == tonic::Code::NotFound => {
-                rpc_error
-                    .status
-                    .checkpoint_height()
-                    .is_some_and(|height| next_checkpoint < height)
+                rpc_error.status.checkpoint_height().is_some_and(|height| {
+                    if next_checkpoint <= height {
+                        true
+                    } else {
+                        height + window_size >= next_checkpoint
+                    }
+                })
             }
             _ => true,
         }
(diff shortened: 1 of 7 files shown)


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,25 +1256,104 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+            Ok(Ok(result)) => Ok(result),
+            Ok(Err(status)) => Err(status),
+            Err(_) => Err(tonic::Status::deadline_exceeded("request timed out")),
+        };
+
+        let Ok(inner_result) = result else {
+            self.num_failures.fetch_add(1, Ordering::Relaxed);
+            let last_success = self.last_success.load(Ordering::Relaxed);
+            if last_success.elapsed() > Self::FAILURE_WINDOW
+                && self.num_failures.load(Ordering::Relaxed) > Self::MAX_FAILURES
```
wbbradley, line 1345: It's not obvious to me why this `&&` isn't an `||`.

wbbradley: OK, I think I get it. It's because you don't want to bail during the `FAILURE_WINDOW`.

wbbradley: nit: Maybe `FAILURE_WINDOW` could be named `IGNORE_FAILURE_WINDOW`?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,25 +1256,104 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+    /// The underlying RPC client.
+    client: Arc<RwLock<Arc<RpcClient>>>,
+    /// The instant of the last successful RPC call.
+    last_success: AtomicInstant,
+    /// The number of failures since the last successful RPC call.
+    num_failures: AtomicUsize,
+}
+
+impl FallibleRpcClient {
+    /// The time window for which failures are counted.
```
wbbradley:
```suggestion
    /// The time window during which failures are counted up but we will not recreate the client.
```

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1190,26 +1195,23 @@ pub enum RetriableClientError {
...
             Self::RpcError(rpc_error) if rpc_error.status.code() == tonic::Code::NotFound => {
-                rpc_error
-                    .status
-                    .checkpoint_height()
-                    .is_some_and(|height| next_checkpoint < height)
+                rpc_error.status.checkpoint_height().is_some_and(|height| {
+                    if next_checkpoint <= height {
+                        true
+                    } else {
+                        height + window_size >= next_checkpoint
```
liquid-helium: There are several variables involved here: the local checkpoint sequence number, the sequence number from a specific fullnode, and the exponential window size. I'm considering that when multiple RPC nodes are present and they have varying heights, the reasoning may become complicated. 

If I read the code correctly, we do fallback when the checkpoint is pruned or when the full-node is not significantly behind the local sequence number.

Is it feasible to rely solely on the duration we have been stuck?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,25 +1256,104 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+    rpc_url: String,
+    /// The underlying RPC client.
+    client: Arc<RwLock<Arc<RpcClient>>>,
+    /// The instant of the last successful RPC call.
+    last_success: AtomicInstant,
+    /// The number of failures since the last successful RPC call.
+    num_failures: AtomicUsize,
+}
+
+impl FallibleRpcClient {
```
liquid-helium, line 1299: Question: the reason we were not able to recovering checkpoint downloading is because the RpcClient somehow could not recover internally, and the fix is to re-create a clean RpcClient.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,25 +1256,104 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+    /// The time window for which failures are counted.
+    const FAILURE_WINDOW: Duration = Duration::from_secs(600);
+    /// The maximum number of failures allowed.
+    const MAX_FAILURES: usize = 100;
```
halfprice, line 1303: suggestion: can you document why we need two criterias?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,25 +1256,104 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+            Err(_) => Err(tonic::Status::deadline_exceeded("request timed out")),
+        };
+
+        let Ok(inner_result) = result else {
+            self.num_failures.fetch_add(1, Ordering::Relaxed);
+            let last_success = self.last_success.load(Ordering::Relaxed);
+            if last_success.elapsed() > Self::FAILURE_WINDOW
+                && self.num_failures.load(Ordering::Relaxed) > Self::MAX_FAILURES
+            {
+                mem::drop(client);
```
halfprice: nit: drop() maybe sufficient?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1554,42 +1617,93 @@ impl RetriableRpcClient {
...
+        }
+
+        // After first minute, start exponential growth
+        // Subtract 60 to start counting from when exponential growth begins
+        let seconds_since_growth_start = seconds_since_last_height - 60;
+
+        // Base window size of 50 checkpoints (10 seconds worth at ~5 checkpoints/second)
+        let base_window = 50;
+
+        // Exponential growth after first minute (assuming 5 checkpoints/second), this gives us:
```
halfprice: Note: in general, we get about 4.2 checkpoints per seconds.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1554,42 +1617,93 @@ impl RetriableRpcClient {
...
+        // 50 checkpoints (10s worth) in 120 seconds
+        // and 500 checkpoints (100s worth) in 240 seconds and 5000 checkpoints (1000s worth) in 360 seconds
+        // which means we grow exponentially to catch up within 360 seconds ~ 6 minutes
+        let exponential_factor = 10u64.pow((seconds_since_growth_start / 60) as u32);
+
+        // Cap at 10 million checkpoints
+        (base_window * exponential_factor).min(10_000_000)
```
halfprice: note: if 10M checkpoints is roughly 1 month of checkpoints.

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,25 +1256,104 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+            self.num_failures.fetch_add(1, Ordering::Relaxed);
+            let last_success = self.last_success.load(Ordering::Relaxed);
+            if last_success.elapsed() > Self::FAILURE_WINDOW
+                && self.num_failures.load(Ordering::Relaxed) > Self::MAX_FAILURES
+            {
+                mem::drop(client);
+                let mut client = self.client.write().await;
+                *client = Arc::new(RpcClient::new(self.rpc_url.clone())?);
+            }
+            return Err(tonic::Status::deadline_exceeded("request timed out"));
```
halfprice: remark: should here return the error returned by `f()`?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,25 +1256,104 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+        self.num_failures.store(0, Ordering::Relaxed);
+        Ok(inner_result)
+    }
+}
+
+impl std::fmt::Debug for FallibleRpcClient {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("FallibleRpcClient")
+            .field("rpc_url", &self.rpc_url)
+            .field("num_failures", &self.num_failures.load(Ordering::Relaxed))
```
halfprice, line 1366: suggestion: also add `last_success`?

crates/walrus-sui/src/client/retry_client.rs
halfprice: nit: can you add some documentation to why we need to consider window_size here? The important aspect is that this is a changing window based on the last height

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,25 +1256,104 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+        let client = self.client.read().await;
+        let result = tokio::time::timeout(timeout, f(client.clone())).await;
+
+        let result = match result {
+            Ok(Ok(result)) => Ok(result),
+            Ok(Err(status)) => Err(status),
+            Err(_) => Err(tonic::Status::deadline_exceeded("request timed out")),
+        };
+
+        let Ok(inner_result) = result else {
```
halfprice, line 1336: Important: should we only start counting on certain error such as RPC error? For functioning error, e,g, the request objects/checkpoints do not exist, we probably shouldn't count them as failure that need client re-creation.

halfprice (commented):
Thank you so much @sadhansood for quick action on the issue! Amazing!

I have a few minor comments, but overall this LGTM!

crates/walrus-sui/src/client/retry_client.rs
sadhansood: Yes, that is correct.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: done, down below added this: ``` // The two conditions together mean that the client has been failing for more than 10
            // minutes and has had more than 100 failures. Given we request checkpoint data at 4.2
            // checkpoints per second, this means that we would have requested ~2400 checkpoints in
            // the last 10 minutes. So this is a good proxy for the client being stuck in a failed
            // state.```

crates/walrus-sui/src/client/retry_client.rs
sadhansood: I agree that would be ideal, but i think given we don't know how exactly the error manifests when the client gets stuck - it's probably safer for now to just do it for all errors. It's not going to be end of the world or performance impacting, given we re-create the client once every 10 minutes in some rare scenarios. lmkwyt

crates/walrus-sui/src/client/retry_client.rs
sadhansood: keeping `FAILURE_WINDOW` for now as it seems more intuitive. Hopefully this seems ok

crates/walrus-sui/src/client/retry_client.rs
sadhansood: done

crates/walrus-sui/src/client/retry_client.rs
sadhansood: done

crates/walrus-sui/src/client/retry_client.rs
sadhansood: I realized the whole exponential window thing is a bit over engineered, and not needed. Simplified it to something like `if all rpc requests to primary failed in last 5 minutes and at least 100 errors occured, then rpc request is eligible for archival fallback`.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: done

crates/walrus-sui/src/client/retry_client.rs
sadhansood: not needed anymore

crates/walrus-sui/src/client/retry_client.rs
sadhansood: not needed anymore

crates/walrus-sui/src/client/retry_client.rs
sadhansood: this is fixed

crates/walrus-sui/src/client/retry_client.rs
sadhansood: done

crates/walrus-sui/src/client/retry_client.rs
wbbradley: Sounds good.

crates/walrus-sui/src/client/retry_client.rs
wbbradley: Do we have a hunch as to where the RpcClient state is becoming corrupted, such that power cycling it becomes the best available option?

crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -1254,26 +1283,112 @@ impl From<tonic::Status> for CheckpointRpcError {
...
+            // minutes and has had more than 100 failures. Given we request checkpoint data at 4.2
+            // checkpoints per second, this means that we would have requested ~2400 checkpoints in
+            // the last 10 minutes. So this is a good proxy for the client being stuck in a failed
+            // state.
+            if last_success.elapsed() > Self::FAILURE_WINDOW
+                && self.num_failures.load(Ordering::Relaxed) > Self::MAX_FAILURES
+            {
+                drop(client);
+                let mut client = self.client.write().await;
+                *client = Arc::new(RpcClient::new(self.rpc_url.clone())?);
```
wbbradley, line 1350: Let's add a prometheus metric for when this happens.

wbbradley (changes requested):
Let's add a prometheus metric for whenever we recreate the RpcClient.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: There are some github issues for tonic client getting stuck. Other than that, it's mostly from observing clients getting stuck across server restarts. It'd be nice to reproduce it however.

crates/walrus-sui/src/client/retry_client.rs
sadhansood: Added a log line, can add metric in a follow up PR as that is a little bit more work.

PR #1963, opened 2025-04-11
chore: Improve rpc retry logic

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client.rs
+++ b/crates/walrus-sui/src/client/retry_client.rs
@@ -307,7 +307,7 @@ pub struct FailoverWrapper<T> {
 impl<T> FailoverWrapper<T> {
     /// The default maximum number of retries.
     const DEFAULT_MAX_RETRIES: usize = 3;
-    const DEFAULT_RETRY_DELAY: Duration = Duration::from_millis(500);
+    const DEFAULT_RETRY_DELAY: Duration = Duration::from_millis(100);
 
     /// Creates a new failover wrapper.
     pub fn new(instances: Vec<(T, String)>) -> anyhow::Result<Self> {
@@ -418,6 +418,7 @@ impl<T> FailoverWrapper<T> {
                             next_client = self.get_name(current_index + 1),
                             "Failed to execute operation on client, retrying with next client"
                         );
+                        // Sleep for a short duration to avoid aggressive retries.
                         tokio::time::sleep(Self::DEFAULT_RETRY_DELAY).await;
                         current_index += 1;
                     }


crates/walrus-sui/src/client/retry_client.rs
```diff
@@ -395,12 +409,16 @@ impl<T> FailoverWrapper<T> {
...
+                            // A custom target for filtering.
+                            target: "walrus_sui::client::retry_client::failover",
+                            tracing::Level::DEBUG,
                             ?last_error,
                             current_client = self.get_name(current_index),
                             next_client = self.get_name(current_index + 1),
-                            "Failed to execute operation on client, retrying with next client",
+                            "Failed to execute operation on client, retrying with next client"
                         );
+                        tokio::time::sleep(Self::DEFAULT_RETRY_DELAY).await;
```
sadhansood, line 422: @liquid-helium can you add a comment about why this sleep is important?

sadhansood (approved):
Thanks @liquid-helium , this pr lgtm!

crates/walrus-sui/src/client/retry_client.rs
sadhansood: @liquid-helium I was thinking about this a bit more and one thing which worries me a little bit is how frequently we will jump b/w endpoints in the failover wrapper because of the errors we can get while fetching checkpoints. Consider temporal errors like fetching full checkpoints for recent checkpoints from primary full node endpoint. At steady state, this will most definitely fail for recent checkpoints on primary full node A, but because of failover wrapper this will quickly failover to next endpoint B, and then C, etc. One suggestion I have (which I am also implementing for archival fallback and might also apply here) is to only failover once there was no success response in last X minutes from an endpoint. This allows fail over once every X minutes (~10 minutes) and not so quick jumps b/w endpoints.

liquid-helium: That's a good point, and I shared the same concern for these unnecessary failover. Your suggestion will effectively retry for more times on the same full-node for this type of temporal errors, before switching to another full-node, right?

I am thinking where we add this retry logic, currently, we have 2 timeout in the call chain, if there is an error or timeout, it seems that the error will kill the worker [task](https://github.com/MystenLabs/walrus/blob/34f40096c56f6df9eeef61b34124a9eb5e8a3ddd/crates/checkpoint-downloader/src/downloader.rs#L145-L163) and later rescheduled by the `pool_monitor`, if I read it correctly.

If we want to retry around the RetriableRpcClient, we might need to increase the timeout, what do you think?

Let me merge this one, and we can discuss more offline.

This is a good idea, thanks @sadhansood !

PR #1967, opened 2025-04-14
docs: Replace dead links of walrus.site with wal.app

teohaik:
## Description

Replaced some links in docs that were still directing to the old walrus.site 

## Test plan
-

### DIFF
--- a/docs/book/blog/03_whitepaper.md
+++ b/docs/book/blog/03_whitepaper.md
@@ -8,7 +8,7 @@ and broken links.
 ```
 
 In June, Mysten Labs announced Walrus, a new decentralized secure blob store design, and introduced
-a developer preview that currently stores over [12TiB](https://capacity.walrus.site/) of data.
+a developer preview that currently stores over [12TiB](https://capacity.wal.app/) of data.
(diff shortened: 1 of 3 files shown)


docs/book/blog/03_whitepaper.md
```diff
@@ -8,7 +8,7 @@ and broken links.
 ```
 
 In June, Mysten Labs announced Walrus, a new decentralized secure blob store design, and introduced
-a developer preview that currently stores over [12TiB](https://capacity.walrus.site/) of data.
+a developer preview that currently stores over [12TiB](https://capacity.wal.app/) of data.
```
teohaik, line 11: @mlegner  this is not working anymore? Shall we replace the link or remove it?

Himess:
Actually, since these were written as blog posts in an earlier period, there are many broken or outdated links inside. There are dead links in several blogs like 01-02-03-04.

If we don’t want any dead links in the project, fixing them makes sense — but if it's going to remain as an old archive post, fixing may not be very reasonable. I'm curious as well.

mlegner:
Thanks for the PR @teohaik. I'm closing this in favor of #2011.

PR #1969, opened 2025-04-14
chore: leverage RetriableSuiClient for reference gas price

wbbradley:
## Description

Reduce the usage of the WalletContext for access to Sui RPC nodes, since it lacks failover semantics. Note that RetriableSuiClient also lacks failover semantics, but we can add them in the future.

## Test plan

CI pipeline.


wbbradley:
> Thanks @wbbradley for the refactor! LGTM 💯, it seems that the read_client is initially created from the WalletContext, so there is no logic change, right?

That's how I read it, as well. Thanks!

PR #1971, opened 2025-04-14
chore: added logs for event db write options.

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/events/event_processor.rs
+++ b/crates/walrus-service/src/node/events/event_processor.rs
@@ -841,10 +841,16 @@ impl EventProcessor {
 
     /// Opens the stores for the event processor.
     pub fn open_stores(database: &Arc<RocksDB>) -> Result<EventProcessorStores, anyhow::Error> {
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -856,19 +862,24 @@ impl EventProcessor {
         let committee_store = DBMap::reopen(
             database,
             Some(constants::COMMITTEE_STORE),
-            &ReadWriteOptions::default(),
+            &default_rw_opts,
             false,
         )?;
+        let event_store_opts = ReadWriteOptions::default().set_ignore_range_deletions(true);
```
halfprice, line 868: Oh, interesting. Event store is the only cf that is in use and set ignore range deletions to true. (init_state is not used currently).

halfprice (approved):
LGTM! Thanks for the cleanup @liquid-helium !

PR #1972, opened 2025-04-14
chore: widen the compile-time interface of with_failover

wbbradley:
## Description

Allow with_failover to support all retriable error types.

## Test plan

CI Pipeline.


wbbradley:
> LGTM! This is to prepare with_failover also taking a function call from (existing) SuiContractClientInner, yes?

Exactly.

PR #1974, opened 2025-04-15
feat: add an option to request additional symbols from the beginning during sliver recovery

halfprice:
## Description

As we have seen, sliver recovery is currently bottlenecked by storage nodes that are overwhelmed by recovery symbols
requests, and may take long time to return 503. This currently is suspected to be the main bottleneck of blob recovery
after all the blob recovery improvements implemented by @jpcsmith. By allowing requesting additional symbols initially,
sliver recovery saves on latency to wait for a 503 to return and create another requests.

## Test plan

All simtest integration tests are randomly chosen a value to request additional symbols.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/committee/request_futures.rs
+++ b/crates/walrus-service/src/node/committee/request_futures.rs
@@ -241,8 +241,24 @@ where
         additional_symbols: usize,
     ) -> Option<Result<Sliver, InconsistencyProofEnum>> {
         let mut committee_listener = self.shared.subscribe_to_committee_changes();
+
+        // Total symbols required to decode the sliver.
+        let total_symbols_required = self.total_symbols_required(additional_symbols);
+
+        // Total symbols to request from other storage nodes initially.
+        let total_symbols_to_request = std::cmp::min(
+            total_symbols_required
+                + self
+                    .shared
+                    .config
+                    .experimental_sliver_recovery_additional_symbols,
+            self.metadata.n_shards().get().into(),
+        );
+
+        // Track the collection of recovery symbols.
         let mut symbol_tracker = SymbolTracker::new(
-            self.total_symbols_required(additional_symbols),
+            total_symbols_required,
+            total_symbols_to_request,
             self.target_index,
             self.target_sliver_type,
         );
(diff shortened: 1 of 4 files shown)


crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -547,8 +559,20 @@ where
...
     ) -> Option<Result<Sliver, InconsistencyProofEnum>> {
         let mut committee_listener = self.shared.subscribe_to_committee_changes();
+
+        // Total symbols required to decode the sliver.
+        let total_symbols_required = self.total_symbols_required(additional_symbols);
+
+        // Total symbols to request from other storage nodes initially.
+        let total_symbols_to_request = std::cmp::min(
+            total_symbols_required + self.shared.config.sliver_recovery_additional_symbols,
+            self.metadata.n_shards().get() as usize,
```
jpcsmith:
```suggestion
            self.metadata.n_shards().get().into()
```
Or try_into + expect as required. `as` will silently fail if we change integer types.

jpcsmith (approved):
Hey Zhe, nice work looking into this. And kudos on improving the node's recovery performance.

As I mentioned on Slack, while the redundant requests offers a solution, I am a bit hesitant to apply it to this problem as it means increasing the load on the remote storage nodes with requests that are later discarded. Since a lower timeout triages the issue for now, I would instead take the time to try to reduce the latency to a 404 or 503, which should also obviate the need for extra requests.

That being said, I will approve this in the event you feel you need this in to give additional levers in recovery situations while a full solution is being pursued. In which case, could you prefix the config variable with `experimental_*` to ensure that operators understand it's not a config variable they should eagerly set, and that it may be removed in a future version.

halfprice:
> Hey Zhe, nice work looking into this. And kudos on improving the node's recovery performance.
> 
> As I mentioned on Slack, while the redundant requests offers a solution, I am a bit hesitant to apply it to this problem as it means increasing the load on the remote storage nodes with requests that are later discarded. Since a lower timeout triages the issue for now, I would instead take the time to try to reduce the latency to a 404 or 503, which should also obviate the need for extra requests.
> 
> That being said, I will approve this in the event you feel you need this in to give additional levers in recovery situations while a full solution is being pursued. In which case, could you prefix the config variable with `experimental_*` to ensure that operators understand it's not a config variable they should eagerly set, and that it may be removed in a future version.

Thanks for the discussion @jpcsmith ! In addition to what we discussed offline, I want to add that my feeling is that even with reduced 404 or 503 latency, we will still constantly see high latency to some storage nodes, and this will hurt the blob recovery throughput (e.g. the tail of 202 response today is also >10s). Moreover, I'm worried that the current workaround of very short timeout may hurt the system availability if majority of nodes are under stress. So the additional request, which we control the granularity, is a better solution than shorter timeout. But of course if by re-architecting the symbol recovery service and achieve the same throughput, that'll be the best.

PR #1977, opened 2025-04-15
feat: add detailed metrics for recovering slivers

jpcsmith:
## Description

This adds the following metrics to the recover sliver flow:
- `walrus_recovery_future_backoff_total` - a simple counter of the total number of times that recoveries enter exponential backoff,
- `walrus_recovery_future_backoffs` -  a single histogram recording the number of backoffs that each sliver recovery entered before collecting enough symbols,
- `walrus_recovery_future_failed_requests` - a single histogram recording the number of request failures that each sliver recovery encountered before collecting enough symbols, and
- `walrus_recovery_future_state` - a vector of gauges, one for each of 5 states, that gives insight to the state that each future is in when the metrics are being scraped.

Depends on #1976 

## Test plan

Observed metrics being exported in the local testbed.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/committee/request_futures.rs
+++ b/crates/walrus-service/src/node/committee/request_futures.rs
@@ -4,6 +4,7 @@
 use std::{
     cmp,
     collections::{HashMap, VecDeque},
+    num::NonZero,
     pin::Pin,
     sync::{Arc, Mutex as SyncMutex, Weak},
     task::{Context, Poll, ready},
@@ -181,7 +182,8 @@ enum RecoveryStateLabel {
     Init,
     CollectingSymbols,
     Backoff,
-    TailRequest,
+    /// The tail number of remaining requests, min 1, max 5.
+    TailRequest(NonZero<u8>),
     BuildingSliver,
 }
 
@@ -191,7 +193,7 @@ impl AsRef<str> for RecoveryStateLabel {
             Self::Init => "init",
             Self::CollectingSymbols => "collecting-symbols",
             Self::Backoff => "backoff",
-            Self::TailRequest => "tail-request",
+            Self::TailRequest(_) => "tail-request",
             Self::BuildingSliver => "building-sliver",
         }
     }
@@ -215,7 +217,8 @@ impl RecoverSliverStats {
             RecoveryStateLabel::Init,
             OwnedGaugeGuard::acquire(walrus_utils::with_label!(
                 metrics.recovery_future_state,
-                RecoveryStateLabel::Init
+                RecoveryStateLabel::Init,
+                ""
             )),
         );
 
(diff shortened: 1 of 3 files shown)


halfprice:
Hi @jpcsmith , looks like this PR still include the change that removes the legacy sliver recovery code, which has been merged. Do you mind to rebase the code first?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -545,6 +629,14 @@ impl<'a, T: NodeService> CollectRecoverySymbols<'a, T> {
             new_request_count,
             "completed refilling pending requests with additional futures"
         );
+
+        match self.pending_requests.len() {
+            0 => (),
+            1 => self.stats.record_state(RecoveryStateLabel::TailRequest),
```
halfprice: suggestion: to make the result give a little bit more information, I would add maybe 1TailRequest, 2TailRequest.. maybe up to 5. I don't know if there is always just 1 that takes long time to return.

crates/walrus-service/src/node/metrics.rs
```diff
@@ -175,6 +175,23 @@ walrus_utils::metrics::define_metric_set! {
...
         #[help = "The number shards currently owned by this node"]
         shards_owned: U64Gauge[],
+
+        #[help = "The total number of times recovery futures entered exponential backoff."]
+        recovery_future_backoff_total: IntCounter[],
+
+        #[help = "The number of times a recovery futures entered exponential backoff before
+        completing."]
+        recovery_future_backoffs: Histogram {
+            buckets: [1.0, 2.0, 3.0, 4.0, 5.0, 6.0, 7.0, 8.0, 9.0, 10.0]
```
halfprice, line 185: Suggestion (not about this code particularly, but reminds me of another issue): when reading the http_client_...duration_second and http_server_..._duration_second. I think the max bucket is 10.0 second, and we see the tail is just 10.0. Can you maybe also visit there and add more buckets to maybe 5 min (since the timeout is 5min)?

crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -175,13 +176,78 @@ where
...
+}
+
+/// Tracks aggregated statistics and the state of a RecoverSliver future.
+struct RecoverSliverStats {
+    /// The total number of backoffs performed during the recovery.
+    total_backoffs: usize,
+    /// The total number of failed requests observed during the recovery.
+    total_failed_requests: usize,
+    /// The current "recovery state" of the request, as identified by the `RecoveryStateLabel`.
+    current_state: (RecoveryStateLabel, OwnedGaugeGuard),
```
halfprice, line 209: praise: the way how OwnedGaugeGuard works is smart! I won't be able to come up with a code design like this...

halfprice (approved):
Thanks @jpcsmith ! And this looks great!

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: I just copied it from mysten_metrics 😅

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: Added as a separate label, which takes values `"", "1", ..., "5"`.

crates/walrus-service/src/node/metrics.rs
jpcsmith: Created PR #2031 for this.

PR #1981, opened 2025-04-16
feat: introduce RPC failover mechanics to the RetriableSuiClient

wbbradley:
## Description

Failover the RetriableSuiClient to multiple RPC nodes as needed.

## Test plan

Manual testing, CI pipeline.

### DIFF
--- a/crates/walrus-service/src/common/config.rs
+++ b/crates/walrus-service/src/common/config.rs
@@ -3,8 +3,9 @@
 
 //! Common configuration module.
 
-use std::{sync::Arc, time::Duration};
+use std::{iter::once, sync::Arc, time::Duration};
 
+use indexmap::IndexSet;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, DurationMilliSeconds};
 use walrus_sui::{
@@ -57,11 +58,21 @@ pub struct SuiConfig {
     pub request_timeout: Option<Duration>,
 }
 
+/// Combines the main RPC URL with additional RPC endpoints, ensuring uniqueness of each URL string.
+pub fn combine_rpc_urls(rpc: &str, additional_rpc_endpoints: &[String]) -> Vec<String> {
+    once(rpc.to_string())
+        .chain(additional_rpc_endpoints.iter().cloned())
+        .collect::<IndexSet<String>>()
+        .into_iter()
+        .collect::<Vec<_>>()
+}
+
 impl SuiConfig {
     /// Creates a new [`SuiReadClient`] based on the configuration.
     pub async fn new_read_client(&self) -> Result<SuiReadClient, SuiClientError> {
-        SuiReadClient::new_for_rpc(
-            &self.rpc,
+        let combined_rpc_urls = combine_rpc_urls(&self.rpc, &self.additional_rpc_endpoints);
+        SuiReadClient::new_for_rpc_urls(
+            &combined_rpc_urls,
             &self.contract_config,
             self.backoff_config.clone(),
         )
@@ -108,7 +119,7 @@ impl From<&SuiConfig> for SuiReaderConfig {
     }
 }
 
-/// Backup-specific configuration for Sui.
+/// Reader-specific configuration for Sui.
 #[serde_with::serde_as]
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
 pub struct SuiReaderConfig {
@@ -132,7 +143,7 @@ pub struct SuiReaderConfig {
     /// The URL of the checkpoint download fallback endpoint.
     #[serde(default, skip_serializing_if = "defaults::is_none")]
     pub rpc_fallback_config: Option<RpcFallbackConfig>,
-    /// Additional RPC endpoints to use for the event processor.
+    /// Additional RPC endpoints to use for failover.
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub additional_rpc_endpoints: Vec<String>,
     /// The request timeout for communicating with Sui network.
@@ -143,8 +154,9 @@ pub struct SuiReaderConfig {
 impl SuiReaderConfig {
     /// Creates a new [`SuiReadClient`] based on the configuration.
     pub async fn new_read_client(&self) -> Result<SuiReadClient, SuiClientError> {
-        SuiReadClient::new_for_rpc(
-            &self.rpc,
+        let combined_rpc_urls = combine_rpc_urls(&self.rpc, &self.additional_rpc_endpoints);
+        SuiReadClient::new_for_rpc_urls(
+            &combined_rpc_urls,
             &self.contract_config,
             self.backoff_config.clone(),
         )
@@ -176,5 +188,27 @@ pub mod defaults {
         // The `cfg!(test)` check is there to allow serializing the full configuration, specifically
         // to generate the example configuration files.
         !cfg!(test) && t.is_none()
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn test_combine_rpc_urls() {
+        let rpc = "http://localhost:1".to_string();
+        let rpc_urls = vec![
+            "http://localhost:2".to_string(),
+            "http://localhost:2".to_string(),
+            "http://localhost:3".to_string(),
+            "http://localhost:1".to_string(),
+            "http://localhost:3".to_string(),
+        ];
+
+        // Check that the duplicates are removed and the order is preserved.
+        let combined = super::combine_rpc_urls(&rpc, &rpc_urls);
+        assert_eq!(combined.len(), 3);
+        assert_eq!(combined[0], "http://localhost:1");
+        assert_eq!(combined[1], "http://localhost:2");
+        assert_eq!(combined[2], "http://localhost:3");
     }
 }
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -72,9 +72,7 @@ use walrus_test_utils::WithTempDir;
 use walrus_utils::{backoff::ExponentialBackoffConfig, metrics::Registry};
 
 #[cfg(msim)]
-use crate::common::config::SuiConfig;
-#[cfg(msim)]
-use crate::node::ConfigLoader;
+use crate::common::config::{combine_rpc_urls, SuiConfig};
 use crate::node::{
     committee::{
         BeginCommitteeChangeError,
@@ -100,6 +98,8 @@ use crate::node::{
     Storage,
     StorageNode,
 };
+#[cfg(msim)]
+use crate::node::{events::event_processor::EventProcessorRuntimeConfig, ConfigLoader};
 
 /// Default buyer subsidy rate (5%)
 const DEFAULT_BUYER_SUBSIDY_RATE: u16 = 500;
@@ -260,7 +260,6 @@ impl StorageNodeHandleTrait for StorageNodeHandle {
     ) -> anyhow::Result<Self> {
         builder.build().await
     }
-
     // StorageNodeHandle is only used in integration test without crash and recovery. No need to
     // use distinct IP.
     fn use_distinct_ip() -> bool {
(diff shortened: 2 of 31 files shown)


crates/walrus-service/src/common/config.rs
```diff
@@ -104,14 +117,24 @@ impl From<&SuiConfig> for SuiReaderConfig {
...
+    /// This field can be omitted if `rpc_urls` is set.
+    #[serde(default, skip_serializing_if = "defaults::is_none")]
+    pub rpc: Option<String>,
+    /// HTTP URLs of the Sui full-node RPC endpoints (including scheme). These are used in the event
+    /// processor and in other read operations; for all write operations, the RPC URL from the
+    /// wallet is used.
+    ///
+    /// For backwards compatibility, if `rpc` is set, it will be added to this list.
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
+    pub rpc_urls: Vec<String>,
```
halfprice: Can we reuse the `additional_rpc_endpoints` below?

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -2,110 +2,217 @@
...
+/// this is only used by the `msim` feature, and is not used in production code.
+pub trait InjectableRetryError {
+    #[cfg(msim)]
+    fn make_injectable_error() -> Self;
+}
+
+impl InjectableRetryError for sui_sdk::error::Error {
+    #[cfg(msim)]
+    fn make_injectable_error() -> Self {
+        todo!()
```
wbbradley: TODO

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -2,110 +2,217 @@
...
+    }
+}
+
+impl InjectableRetryError for SuiClientError {
+    #[cfg(msim)]
+    fn make_injectable_error() -> Self {
+        todo!()
```
wbbradley: TODO

crates/walrus-service/src/common/config.rs
wbbradley: Done.

crates/walrus-service/src/common/config.rs
```diff
@@ -57,11 +58,22 @@ pub struct SuiConfig {
     pub request_timeout: Option<Duration>,
 }
 
+/// Combines the main RPC URL with additional RPC endpoints, ensuring uniqueness, and removing
+/// duplicates.
```
liquid-helium: nit: consider adding a comment noting that this deduplication relies on URL text, as opposed to the underlying endpoints.

crates/walrus-service/src/test_utils.rs
```diff
@@ -258,9 +258,11 @@ impl StorageNodeHandleTrait for StorageNodeHandle {
         _start_node: bool,
         _disable_event_blob_writer: bool,
     ) -> anyhow::Result<Self> {
+        // This process can take exorbitant amounts of stack memory due to the number of concurrent
+        // storage nodes. Let's move each storage node onto its own thread with a slightly larger
+        // stack.
```
liquid-helium: Nice catch.
Is this a TODO? can you create an issue to track it?

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -65,6 +68,86 @@ use crate::{
...
+            Self::Url {
+                rpc_url,
+                request_timeout: None,
+            } => Ok(Arc::new(
+                SuiClientBuilder::default()
+                    .build(rpc_url)
+                    .await
+                    .map_err(|e| FailoverError::FailedToGetClient(e.to_string()))?,
+            )),
+        }
```
liquid-helium: nit: 
match self {
    Self::Client(client) => Ok(client.clone()),
    Self::Url { rpc_url, request_timeout } => {
        let mut builder = SuiClientBuilder::default();
        
        if let Some(timeout) = request_timeout {
            builder = builder.request_timeout(*timeout);
        }
        
        Ok(Arc::new(
            builder.build(rpc_url)
                .await
                .map_err(|e| FailoverError::FailedToGetClient(e.to_string()))?
        ))
    }
}

crates/walrus-service/src/test_utils.rs
wbbradley: This is now leftover cruft as I figured out the optimization fix.

crates/walrus-service/src/common/config.rs
wbbradley: I'm not sure what "underlying endpoints" means in this context, but I'll make it explicit that we're talking about URL text here. Thanks!

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
wbbradley: Ok.

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -133,30 +244,162 @@ impl<T> FailoverWrapper<T> {
...
-                            tracing::Level::DEBUG,
-                            ?last_error,
-                            current_client = self.get_name(current_index),
-                            next_client = self.get_name(current_index + 1),
-                            "Failed to execute operation on client, retrying with next client"
-                        );
-                        // Sleep for a short duration to avoid aggressive retries.
-                        tokio::time::sleep(Self::DEFAULT_RETRY_DELAY).await;
-                        current_index += 1;
+                    let failed_rpc_url = self.get_current_rpc_url().await;
```
liquid-helium, line 272: Important: if there are multiple threads/tasks using the same FailoverWrapper:
1. task 1 gets client_1, and starts a call
2. task 2 gets client_1, and starts a call
3. client_1 fails in task_1, and fetch_next_client() is called, now the current client is client_2
4. task 2 failed with client_1, now the current client is already client_2, and task2 can use it directly.

We could attach an ID to each client, and when calling fetch_next_client, we pass in the current client ID.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -136,7 +214,7 @@ impl RetriableSuiClient {
             "get_client",
         )
         .await?;
-        Ok(Self::new(client, backoff_config))
+        Ok(Self::new(vec![client.into()], backoff_config).await?)
```
liquid-helium, line 223: nit: Can we add a TODO here to support adding more endpoints, even when called from new_from_wallet()?

wbbradley: https://linear.app/mysten-labs/issue/WAL-793/

crates/walrus-sui/src/client/retry_client/failover.rs
wbbradley: OK, I believe I've addressed this issue by tracking which client indices have been tried from each with_failover callsite.

sadhansood:
Thanks @wbbradley, this PR looks really good and I just have some general comments from scanning the PR which don't need to be addressed now but perhaps sometime later but food for thought:
1. I wonder if multiple instances of failover wrapper could cause cascading failures or thundering herd problem when client at the `current_index` goes down (if there are a lot of nodes at current_index, they all instantly jump to the next one at the same time). Would it make sense to add a jitter to the sleep before we move to the client at `current_index + 1`? Also, would it make sense to round robin with different increments i.e. `current_index + x` where `x` is variable?  
2. I worry a little about the state inconsistency for read failover for sui clients. This is not much of a problem with writes or checkpoint reads but may be for object (system, stake, etc) reads. The failover read might hit a node with older checkpoint state and hence may cause some subtle application level bugs which i don't know for sure about.
3. The system doesn't learn today if an endpoint is bad (always fails because its down). It is always a part of the set of endpoints that we round robin over and retry many times before moving on.
4. Dead entries in the list waste time during failover cycles and reduce the effective number of healthy failover targets. Typos in URLs, incorrect ports, or non-existent endpoints are included in the configuration. The LazyClientBuilder will fail when attempting to build the client for that endpoint index. with_failover treats this as a failure and rotates to the next index.

wbbradley:
> Thanks @wbbradley, this PR looks really good and I just have some general comments from scanning the PR which don't need to be addressed now but perhaps sometime later but food for thought:
> 
> 1. I wonder if multiple instances of failover wrapper could cause cascading failures or thundering herd problem when client at the `current_index` goes down (if there are a lot of nodes at current_index, they all instantly jump to the next one at the same time). Would it make sense to add a jitter to the sleep before we move to the client at `current_index + 1`? Also, would it make sense to round robin with different increments i.e. `current_index + x` where `x` is variable?

I think this is indeed possible, but then again, it's unclear to me that all the nodes would have the same exact configuration. One potential direction we could go here would be the `+ x` as you've alluded to, or to shuffle the rpc_urls on load, or some other time. Definitely open to further ideas here...

> 2. I worry a little about the state inconsistency for read failover for sui clients. This is not much of a problem with writes or checkpoint reads but may be for object (system, stake, etc) reads. The failover read might hit a node with older checkpoint state and hence may cause some subtle application level bugs which i don't know for sure about.

This was a primary concern I had early on as well. Certainly I feel like there could be a write-then-read inconsistency possible here, but @halfprice and I spoke about this last week, and I think we came to the conclusion that writes carry version information which would cause blocking on the server side, or cause a retry. But, we should bring this up again.

> 3. The system doesn't learn today if an endpoint is bad (always fails because its down). It is always a part of the set of endpoints that we round robin over and retry many times before moving on.

This one seems relatively tractable by way of maintaining state within the FailoverWrapper for each lazy_client_builder. Like, we could track a last failure time, and enforce a cool-down before retrying a particular node.

> 4. Dead entries in the list waste time during failover cycles and reduce the effective number of healthy failover targets. Typos in URLs, incorrect ports, or non-existent endpoints are included in the configuration. The LazyClientBuilder will fail when attempting to build the client for that endpoint index. with_failover treats this as a failure and rotates to the next index.

I agree on this, as well. Barring a more complex bookkeeping solution, I figure for now this solution _should_ still be more reliable than our current setup? And, as you mention, let's revisit these improvements moving forward.

sadhansood:
> This was a primary concern I had early on as well. Certainly I feel like there could be a write-then-read inconsistency possible here, but [@halfprice](https://github.com/halfprice) and I spoke about this last week, and I think we came to the conclusion that writes carry version information which would cause blocking on the server side, or cause a retry. But, we should bring this up again.
Thanks @wbbradley, I think there might be a confusion between write safety and read consistency (or I did not fully understand the statement). Sui's object versioning is primarily designed to ensure the safety and atomicity of write operations. When submitting a transaction, you specify the expected input object versions. The network checks these versions before execution. If a node receives a transaction with an incorrect input version (e.g., because another transaction modified the object already or the node itself is behind), it rejects the transaction (or waits for the new version of the input objects). This mechanism makes retrying the exact same write transaction on failover relatively safe – it will either execute based on the correct prerequisite state or fail cleanly due to a version mismatch (or wait if the node is lagging)

But I think the read problem remains: The core issue isn't about the safety of the write itself, but about the state visible to subsequent reads when failover occurs. Object versioning during write execution on Node A does not guarantee that Node B (a replica with potential replication lag) will have the updated state immediately available if a read request fails over to it.
Maybe we can think with some examples in two following scenarios. 
1. Write on A -> Failed-over Read on B (behind A): The write Tx1 (modifying Object O from version V to V+1) executes successfully on Node A because the version check (input V) passed on Node A. A subsequent read request for Object O fails over to Node B. If Node B hasn't yet replicated the results of Tx1, it still holds Object O at version V. Node B will successfully return Object O at version V to the client. The client does not read its own write.

2. Read on A -> Failed-over Read on B (behind A)): The client reads Object O (version V) from Node A. Later, the client makes another read request for Object O, which fails over to Node B. Node B still holds Object O at version V due to lag (even if Node A might now have V+1 due to some other transaction). Node B successfully returns Object O at version V.
The client reads a potentially stale version on the second attempt due to failover selecting a lagging replica.

If you ans @halfprice concluded that the above scenarios are not a cause of concern, than I am fine but just wanted to bring it up to make sure of it.

sadhansood:
> One more question on the write path, let's say we are doing a transaction on a owned-object, like changing the node config on full-node A, in the middle of the transaction, the connection failed (only the connection fails, node A is still running). then we switch to full-node B, and retry the same transaction, will the second transaction fail, could it create a dead lock?

I think in this case it should not be a problem since full node B will also submit the exact same transaction to the validators, it should not cause equivocation.

sadhansood (approved):
LGTM, thanks @wbbradley!

PR #1982, opened 2025-04-16
feat: Support Custom Subsidy Package

zihehuang:
## Description

Currently, the subsidy package only subsidizes the `reserve_space` call. To enable end-to-end walrus credits, we need to subsidize the `register_blob` call as well.

The PR enables the CLI to make use of custom subsidy package. It allows the calling of custom `register_blob` function in the subsidy package if the sender has the permission to call it. The custom `register_blob` could optionally subsidize the write fee. If the function is not present, the CLI would default to subsidize the `reserve_space` call only.

## Test plan

Simtest WIP

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Support Custom Subsidy Package. This allows the calling of custom `register_blob` function in the subsidy package, which enables walrus credits


mlegner:
@zihehuang Are you still planning to land this PR, or can it be closed?

zihehuang:
Im closing this

PR #1983, opened 2025-04-16
fix: make mainnet-contracts/*/Move.* correspond to deployed version

karlwuest:
## Description

This PR makes sure that the compiler `Move.toml` and `Move.lock` files in `mainnet-contracts` reflect the state at the actual time of deployment. For the `wal` contract, this was commit 16606e7e52e3f7573cd917665a71e7caf49b0807, and for the `walrus` and `subsidies` contracts, this was commit 10217c558c110b2f9dc7bbc35b8947628b65e5e3.

In particular, the goal is to make the used compiler version clear to ensure that running `sui client verify-source` succeeds if the sui version corresponding to the compiler version in the `Move.lock` files is used. Small changes from after the deployment to e.g. comments or formatting in the source code that do not affect this have been kept intact.

There are some caveats, namely, we cannot accurately reflect the dependency versions for `wal`, since the compilation of `walrus` and `subsidies` requires that the dependency versions are the same. However, this does not matter for source verification, so it does not affect the main goal.

## Test plan

- Verified the source of `walrus` and `subsidies` locally against the on-chain versions using sui version `mainnet-v1.44.3`

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/mainnet-contracts/subsidies/Move.lock
+++ b/mainnet-contracts/subsidies/Move.lock
@@ -2,7 +2,7 @@
 
 [move]
 version = 3
-manifest_digest = "EED1338DA187CE13DE19336996CBB716BAC5DB75A89666E9F95AD9F4E4863797"
+manifest_digest = "9A2082DC6F53FA70312454FB0E4E66F3442B2529705FE6BCB044AC8E71B81E04"
 deps_digest = "060AD7E57DFB13104F21BE5F5C3759D03F0553FC3229247D9A7A6B45F50D03A3"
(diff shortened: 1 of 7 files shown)


karlwuest:
>Are the corresponding files correct for the testnet-contracts?

Yes, you pushed those immediately after we redeployed. The issue with the mainnet-contracts was that we didn't have a separate directory when we deployed it.

PR #1987, opened 2025-04-16
feat: config sui client request timeout

halfprice:
## Description

Add a request timeout config option in ClientConfig (client side) and SuiConfig (server side) to config
the WalletContext and SuiClient request timeout communicating with Sui.

By default we are not setting it, which is the same behavior as today.

This is needed mostly for antithesis test at the moment.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/cli/runner.rs
+++ b/crates/walrus-service/src/client/cli/runner.rs
@@ -152,7 +152,7 @@ impl ClientCommandRunner {
                 .as_ref()
                 .ok()
                 .and_then(|config: &ClientConfig| config.wallet_config.clone()));
-        let wallet = WalletConfig::load_wallet_context_with_request_timeout(
+        let wallet = WalletConfig::load_wallet_context(
             wallet_config.as_ref(),
             config.as_ref().map_or(None, |config| {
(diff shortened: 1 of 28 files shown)


crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -152,7 +152,12 @@ impl ClientCommandRunner {
                 .as_ref()
                 .ok()
                 .and_then(|config: &ClientConfig| config.wallet_config.clone()));
-        let wallet = WalletConfig::load_wallet_context(wallet_config.as_ref());
+        let wallet = WalletConfig::load_wallet_context(
+            wallet_config.as_ref(),
+            config.as_ref().map_or(None, |config| {
```
wbbradley:
```suggestion
            config.as_ref().and_then(|config| {
```

wbbradley (approved):

crates/walrus-service/src/client/cli/runner.rs
halfprice: Applied

PR #1990, opened 2025-04-17
fix: epoch shard assignment should always use node `event_epoch` rather than the epoch from the event

halfprice:
## Description

Certified blob's event epoch is the epoch that the blob is originally certified, not the epoch of the event generated.
This epoch can be very old if certified blob event is generated from a blob extension.

## Test plan

Unittest, integration test, local testbed.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1783,7 +1783,6 @@ impl StorageNodeInner {
         }
 
         tracing::error!("unknown epoch {} when checking shard assignment", epoch);
-
         anyhow::bail!("unknown epoch {} when checking shard assignment", epoch);
     }
 
@@ -6009,6 +6008,7 @@ mod tests {
         Ok(())
     }
 
+    // Tests that blob extension is correctly handled after multiple epochs.
     #[tokio::test]
     async fn extend_blob_after_multiple_epochs() -> TestResult {
         let _ = tracing_subscriber::fmt::try_init();
@@ -6017,7 +6017,8 @@ mod tests {
             cluster_with_initial_epoch_and_certified_blob(&[&[0, 1, 2, 3]], &[], 1, None).await?;
 
         let blob_details = EncodedBlob::new(BLOB, cluster.encoding_config());
-        println!("blob_details: {:?}", blob_details.blob_id());
+
+        println!("blob to be extended: {:?}", blob_details.blob_id());
         let object_id = ObjectID::random();
         events.send(
             BlobRegistered {
@@ -6037,6 +6038,7 @@ mod tests {
             .into(),
         )?;
 
+        // Advance to multiple epochs before extending the blob.
         advance_cluster_to_epoch(&cluster, &[&events], 4).await?;
 
         events.send(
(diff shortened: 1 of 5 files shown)


crates/walrus-simtest/src/test_utils.rs
```diff
@@ -160,20 +163,49 @@ pub mod simtest_utils {
         Ok(())
     }
 
+    /// Probabilstically extend one of the blobs from blobs_written.
```
wbbradley:
```suggestion
    /// Probabilistically extend one of the blobs from blobs_written.
```

wbbradley (approved):
LGTM

crates/walrus-service/src/node.rs
```diff
@@ -1782,6 +1790,7 @@ impl StorageNodeInner {
                 .to_vec());
         }
 
+        tracing::error!("unknown epoch {} when checking shard assignment", epoch);
```
liquid-helium: nit: question: does this provide more info for us?

crates/walrus-service/src/node.rs
```diff
@@ -1139,10 +1144,13 @@ impl StorageNode {
         let histogram_set = self.inner.metrics.recover_blob_duration_seconds.clone();
 
         if !self.inner.is_blob_certified(&event.blob_id)?
+            // For blob extension events, the original blob certified event should already recover
```
liquid-helium, line 1147: nit: (maybe not in this pr)
consider adding a new Event type of Extension, it is confusing to use the same EventType. Can we check is_extension, and serialize it to a Extension type.

crates/walrus-service/src/node.rs
```diff
@@ -6006,4 +6015,51 @@ mod tests {
...
+    #[tokio::test]
+    async fn extend_blob_after_multiple_epochs() -> TestResult {
+        let _ = tracing_subscriber::fmt::try_init();
+
+        let (cluster, events, _blob_detail) =
+            cluster_with_initial_epoch_and_certified_blob(&[&[0, 1, 2, 3]], &[], 1, None).await?;
+
+        let blob_details = EncodedBlob::new(BLOB, cluster.encoding_config());
+
+        println!("blob to be extended: {:?}", blob_details.blob_id());
```
liquid-helium: nit: change to tracing?

crates/walrus-service/src/node.rs
halfprice: Good catch. Removed. Unit test does show the `bail` output. But I don't think we need this.

crates/walrus-service/src/node.rs
halfprice: Good idea. Agree that this is not only confusing, but also easy to miss.

I think adding the extension type requires quite some work. Added a TODO to track this.

crates/walrus-service/src/node.rs

halfprice:
Thanks for the quick review @wbbradley @liquid-helium !

PR #1994, opened 2025-04-17
chore: clean up around process_events function

halfprice:
## Description

It's getting long, and too many things going on. Doing a bit clean up to make it organized and easier to read.

Also added monitored scope in all event processing so we can expose some processing metrics.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -705,7 +705,10 @@ impl StorageNode {
             },
             result = self.process_events() => match result {
                 Ok(()) => unreachable!("process_events should never return successfully"),
-                Err(err) => return Err(err),
+                Err(err) => {
+                    tracing::error!("event processing terminated with an error: {:?}", err);
+                    return Err(err);
+                }
             },
             _ = cancel_token.cancelled() => {
                 self.inner.shut_down();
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -705,7 +705,10 @@ impl StorageNode {
             },
             result = self.process_events() => match result {
                 Ok(()) => unreachable!("process_events should never return successfully"),
-                Err(err) => return Err(err),
+                Err(err) => {
+                    tracing::error!("event processing terminated with an error: {:?}", err);
```
liquid-helium, line 709: nit: will this error fail the storage node, and get logged?

crates/walrus-service/src/node.rs
```diff
@@ -989,12 +918,119 @@ impl StorageNode {
...
+                let span = tracing::Span::current();
+                span.record("otel.status_code", "error");
+                span.record("otel.status_message", field::display(err));
+            })
+            .instrument(span)
+            .await
+    }
+
+    /// Checks if the node is severely lagging behind.
+    fn check_if_node_lagging(
```
liquid-helium: nit: change the name to include the possible action, since this is not a simple getter, it could change the node status.

crates/walrus-service/src/node.rs
halfprice: > will this error fail the storage node

yes

> get logged?

It'll be logged when the process terminated by the systemctl. However, in tests, this error may be preceded by other side effects error such as event handle being dropped, which masks the real error.

crates/walrus-service/src/node.rs
halfprice: Good idea. I changed the function name to include the potential node state change.

PR #1997, opened 2025-04-18
feat: blob sliver data existence check

halfprice:
## Description

This PR implements the basic sliver data existence check conducted along with blob info consistency check
at beginning of every epoch. It exposes a metric which represents the percentage of fully synced blobs that
are actually fully stored. Note that here it has already taken blobs that are being synced into consideration.

Contribute to WAL-701

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -148,3 +148,6 @@ balance_check:
 thread_pool:
   max_concurrent_tasks: null
   max_blocking_io_threads: 1024
+consistency_check:
+  enable_consistency_check: true
+  enable_sliver_data_existence_check: false
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -151,3 +151,4 @@ thread_pool:
 consistency_check:
   enable_consistency_check: true
   enable_sliver_data_existence_check: false
+  sliver_data_existence_check_sample_rate_percentage: 100
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -16,6 +16,7 @@ use std::{
 use anyhow::{Context, anyhow, bail};
 use blob_retirement_notifier::BlobRetirementNotifier;
 use committee::{BeginCommitteeChangeError, EndCommitteeChangeError};
+use consistency_check::StorageNodeConsistencyCheckConfig;
 use epoch_change_driver::EpochChangeDriver;
 use errors::{ListSymbolsError, Unavailable};
 use events::{CheckpointEventPosition, event_blob_writer::EventBlobWriter};
@@ -166,6 +167,7 @@ use crate::{
 
 pub mod committee;
 pub mod config;
+pub mod consistency_check;
 pub mod contract_service;
 pub mod dbtool;
 pub mod events;
@@ -176,7 +178,6 @@ pub(crate) mod metrics;
 
 mod blob_retirement_notifier;
 mod blob_sync;
-mod consistency_check;
 mod epoch_change_driver;
 mod node_recovery;
 mod recovery_symbol_service;
@@ -517,6 +518,7 @@ pub struct StorageNodeInner {
     thread_pool: BoundedThreadPool,
     registry: Registry,
     latest_event_epoch: AtomicU32, // The epoch of the latest event processed by the node.
+    consistency_check_config: StorageNodeConsistencyCheckConfig,
 }
 
 /// Parameters for configuring and initializing a node.
@@ -629,6 +631,7 @@ impl StorageNode {
             encoding_config,
             registry: registry.clone(),
             latest_event_epoch: AtomicU32::new(0),
+            consistency_check_config: config.consistency_check.clone(),
         });
 
         blocklist.start_refresh_task();
@@ -1329,17 +1332,20 @@ impl StorageNode {
             .wait_until_previous_task_done()
             .await;
 
-        if let Err(err) = consistency_check::schedule_background_consistency_check(
-            self.inner.clone(),
-            event.epoch,
-        )
-        .await
-        {
-            tracing::warn!(
-                ?err,
-                epoch = %event.epoch,
-                "failed to schedule background blob info consistency check"
-            );
+        if self.inner.consistency_check_config.enable_consistency_check {
+            if let Err(err) = consistency_check::schedule_background_consistency_check(
+                self.inner.clone(),
+                self.blob_sync_handler.clone(),
+                event.epoch,
+            )
+            .await
+            {
+                tracing::warn!(
+                    ?err,
+                    epoch = %event.epoch,
+                    "failed to schedule background blob info consistency check"
+                );
+            }
         }
 
         // During epoch change, we need to lock the read access to shard map until all the new
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1880,8 +1880,16 @@ impl StorageNodeInner {
             None => self.owned_shards_at_latest_epoch(),
         };
 
+        self.is_stored_at_specific_shards(blob_id, &shards).await
+    }
+
+    async fn is_stored_at_specific_shards(
+        &self,
+        blob_id: &BlobId,
+        shards: &[ShardIndex],
+    ) -> anyhow::Result<bool> {
         for shard in shards {
-            match self.storage.is_stored_at_shard(blob_id, shard).await {
+            match self.storage.is_stored_at_shard(blob_id, *shard).await {
                 Ok(false) => return Ok(false),
                 Ok(true) => continue,
                 Err(error) => {
@@ -1911,6 +1919,27 @@ impl StorageNodeInner {
     ) -> anyhow::Result<bool> {
         self.is_stored_at_all_shards_impl(blob_id, Some(epoch))
             .await
+    }
+
+    /// Returns true if the blob is stored at all shards that are in Active state.
+    pub(crate) async fn is_stored_at_all_active_shards(
+        &self,
+        blob_id: &BlobId,
+    ) -> anyhow::Result<bool> {
+        let shards = self
+            .storage
+            .existing_shard_storages()
+            .await
+            .iter()
+            .filter_map(|shard_storage| {
+                shard_storage
+                    .status()
+                    .is_ok_and(|status| status == ShardStatus::Active)
+                    .then_some(shard_storage.id())
+            })
+            .collect::<Vec<_>>();
+
+        self.is_stored_at_specific_shards(blob_id, &shards).await
     }
 
     pub(crate) fn storage(&self) -> &Storage {
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -167,7 +167,7 @@ use crate::{
 
 pub mod committee;
 pub mod config;
-pub mod consistency_check;
+pub(crate) mod consistency_check;
 pub mod contract_service;
 pub mod dbtool;
 pub mod events;
--- a/crates/walrus-service/src/node/consistency_check.rs
+++ b/crates/walrus-service/src/node/consistency_check.rs
@@ -191,7 +191,7 @@ fn compose_blob_list_digest_and_check_sliver_data_existence(
                             // column families, and perform sequential scan along with
                             // BlobInfoIterator to conduct more efficient existence check. This
                             // requires the SafeIterator to support seek() functionality first.
-                            rt.block_on(node.is_stored_at_all_shards_at_latest_epoch(&blob_info.0)),
+                            rt.block_on(node.is_stored_at_all_active_shards(&blob_info.0)),
                             &mut total_fully_stored,
                             &mut existence_check_error,
                             &blob_info.0,
@@ -206,7 +206,7 @@ fn compose_blob_list_digest_and_check_sliver_data_existence(
                     {
                         handle_existence_check_result(
                             futures::executor::block_on(
-                                node.is_stored_at_all_shards_at_latest_epoch(&blob_info.0),
+                                node.is_stored_at_all_active_shards(&blob_info.0),
                             ),
                             &mut total_fully_stored,
                             &mut existence_check_error,
--- a/crates/walrus-service/src/node/consistency_check.rs
+++ b/crates/walrus-service/src/node/consistency_check.rs
@@ -9,6 +9,7 @@ use std::{collections::HashSet, hash::Hasher, sync::Arc};
 
 use anyhow::{Context, Result};
 use prometheus::IntCounterVec;
+use rand::Rng;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 #[cfg(msim)]
@@ -33,6 +34,9 @@ pub struct StorageNodeConsistencyCheckConfig {
     pub enable_consistency_check: bool,
     /// Enable the sliver data existence check.
     pub enable_sliver_data_existence_check: bool,
+    /// The sample rate of the sliver data existence check. Value is between 0 and 100.
+    #[serde(deserialize_with = "deserialize_data_existence_check_sample_rate_percentage")]
+    pub sliver_data_existence_check_sample_rate_percentage: u64,
 }
 
 impl Default for StorageNodeConsistencyCheckConfig {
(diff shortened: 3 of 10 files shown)


crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -72,26 +117,96 @@ fn get_epoch_bucket(epoch: Epoch) -> String {
...
+                        handle_existence_check_result(
+                            // Note that we are running the async function in a new reusable Runtime
+                            // created at the beginning of the task. Note that performance of this
+                            // task is not a concern, and we should limit the resource this task
+                            // can use.
+                            //
+                            // TODO(zhewu): ideally, we should create iterators over the sliver
+                            // column families, and perform sequential scan along with
+                            // BlobInfoIterator to conduct more efficient existence check. This
+                            // requires the SafeIterator to support seek() functionality first.
```
halfprice, line 218: This is just unfortunate and I wish we can directly use raw RocksDb interface (which we might, but I didn't dig deeper).

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -72,26 +117,96 @@ fn get_epoch_bucket(epoch: Epoch) -> String {
...
+                            &mut total_fully_stored,
+                            &mut existence_check_error,
+                            &blob_info.0,
+                        );
+                    }
+
+                    // Unfortunately, msim does not support tokio::Runtime::block_on, so we need to
+                    // use less ideal method to execute the async function that checks blob
+                    // data existence. Note that simtest is mostly for correctness verification, so
+                    // the performance is not a concern.
```
halfprice, line 229: I spent on this issue quite a long time, and this is the best solution I can think of at the moment.

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -72,26 +117,96 @@ fn get_epoch_bucket(epoch: Epoch) -> String {
...
+                            // Note that we are running the async function in a new reusable Runtime
+                            // created at the beginning of the task. Note that performance of this
+                            // task is not a concern, and we should limit the resource this task
+                            // can use.
+                            //
+                            // TODO(zhewu): ideally, we should create iterators over the sliver
+                            // column families, and perform sequential scan along with
+                            // BlobInfoIterator to conduct more efficient existence check. This
+                            // requires the SafeIterator to support seek() functionality first.
+                            rt.block_on(node.is_stored_at_all_shards_at_latest_epoch(&blob_info.0)),
```
halfprice: Here, we have to use `block_on` is not ideal. Based on chatGPT, this is an anti-pattern, but it also explains that the reason it is bad is because it may repeatedly create multiple runtimes. However, here I only create one runtime, and use it in all the checks.

Because of this and unknown to the node performance, I added a configuration so that we can turn the blob existence check on and off (off by default in released binary, and on by default in tests). I'll do some benchmarks
and see the performance impact of this. But please let me know if there are better way to structure the consistency task.

BTW, I've tried to make the consistency task async, but I really don't want to do it due to the high IO occupying the main tokio thread.

crates/walrus-service/src/node/metrics.rs
```diff
@@ -152,6 +153,13 @@ walrus_utils::metrics::define_metric_set! {
         consistency check."]
         per_object_blob_info_consistency_check_certified_scanned: IntCounterVec["epoch"],
 
+        #[help = "The ratio of fully stored blobs during the blob info consistency check."]
+        blob_data_consistency_fully_stored_ratio: GaugeVec["epoch"],
+
+        #[help = "The number of errors occurred when checking the existence of the blobs during \
+        the blob info consistency check."]
+        blob_data_consistency_check_existence_error: IntCounterVec["epoch"],
+
```
jpcsmith: Suggestion: See the naming conventions for metrics: https://prometheus.io/docs/practices/naming/, which make it easier to understand what a metric is reporting when using them in the dashboards.

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -72,26 +120,101 @@ fn get_epoch_bucket(epoch: Epoch) -> String {
     (epoch % 100).to_string()
 }
 
-/// Compose the digest of the blob list returned by the iterator.
+fn handle_existence_check_result(
+    result: Result<bool, anyhow::Error>, // True means blobs are fully stored.
```
jpcsmith: Nit: rename to `is_fully_stored` or similar.

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -72,26 +120,101 @@ fn get_epoch_bucket(epoch: Epoch) -> String {
...
+    node_status: NodeStatus,
+    blobs_not_yet_fully_synced: &HashSet<BlobId>,
     scan_counter: &IntCounterVec,
-) -> Result<u64, TypedStoreError> {
+) -> Result<BlobConsistencyCheckResult, TypedStoreError> {
+    // Create a new tokio runtime for the async task to check blob existence.
+    #[cfg(not(msim))]
+    let rt = tokio::runtime::Builder::new_current_thread()
+        .build()
+        .expect("failed to create tokio runtime");
```
jpcsmith, line 176: Question: Why is it necessary to create a new runtime for this?

crates/walrus-service/src/node/consistency_check.rs
jpcsmith: Remark: We can directly use the raw RocksDB interface, as it's a public field on the DBMap IIRC. 

Currently, the existence check is a pain-point for both the recovery and the event processing. As it is, this may exacerbate the issue further if turned on and so your idea to use the underlying rocksDB iterator may be already worth investigating.

crates/walrus-service/src/node/consistency_check.rs
jpcsmith: Question: What exactly are you trying to avoid with the runtime and `blob_on`? The issue that `is_stored_at_all_shards_at_latest_epoch()` is async and you need to call it? Would using the underlying rocksdb iterator (as you suggested) help with this issue as well? Since you'd no longer need to rely on the async calls?

crates/walrus-service/src/node/consistency_check.rs
halfprice: This is exactly what I plan to do next. I took a stab at this earlier and the amount of changes to implement this is too much (1000+ line including the change in this PR), so therefore I wanted to do it in a 2 step. That PR itself, namely creating a seek-able iterator, worth its own attending IMO.

Also, given that @sadhansood is importing DBMap trait, I'm thinking now that after he finishes importing it, I can implement it as a function inside DBMap, which makes the seek a native function supported by our version of the DBMap.

crates/walrus-service/src/node/consistency_check.rs
halfprice: > The issue that is_stored_at_all_shards_at_latest_epoch() is async and you need to call it?

Correct. Please see my other reply that this is only temporary.

crates/walrus-service/src/node/metrics.rs
halfprice: Thanks for the pointer! Done, and will send out a PR for other metrics.

crates/walrus-service/src/node/consistency_check.rs
halfprice: Done.

crates/walrus-service/node_config_example.yaml
```diff
@@ -148,3 +148,6 @@ balance_check:
 thread_pool:
   max_concurrent_tasks: null
   max_blocking_io_threads: 1024
+consistency_check:
+  enable_consistency_check: true
+  enable_sliver_data_existence_check: false
```
sadhansood, line 153: could we also add a sampling config to check maybe `x%` of blobs. This is basically around resource consumption. For a system with millions or billions of blobs (and thus many more slivers), even a quick existence check per sliver can add up to significant I/O and CPU load over time i think. And maybe the determination of the blobs to check could use a hash function with epoch as a salt?

crates/walrus-service/src/node/metrics.rs
```diff
@@ -152,6 +153,13 @@ walrus_utils::metrics::define_metric_set! {
         consistency check."]
         per_object_blob_info_consistency_check_certified_scanned: IntCounterVec["epoch"],
 
+        #[help = "The ratio of fully stored blobs during the blob info consistency check."]
+        node_blob_data_fully_stored_ratio: GaugeVec["epoch"],
+
+        #[help = "The number of errors occurred when checking the existence of the blobs during \
+        the blob info consistency check."]
+        node_blob_data_consistency_check_existence_error: IntCounterVec["epoch"],
+
```
sadhansood, line 162: Should we also have a metric to track the number of non active shards that were not checked during a run?

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -72,26 +120,101 @@ fn get_epoch_bucket(epoch: Epoch) -> String {
...
+                            // Note that we are running the async function in a new reusable Runtime
+                            // created at the beginning of the task. Note that performance of this
+                            // task is not a concern, and we should limit the resource this task
+                            // can use.
+                            //
+                            // TODO(zhewu): ideally, we should create iterators over the sliver
+                            // column families, and perform sequential scan along with
+                            // BlobInfoIterator to conduct more efficient existence check. This
+                            // requires the SafeIterator to support seek() functionality first.
+                            rt.block_on(node.is_stored_at_all_active_shards(&blob_info.0)),
```
sadhansood, line 219: @halfprice - I did not fully understand, why does this need to run in a separate runtime. Could we not just call `existing_shard_storages` in the current runtime and pass it along to `is_stored_at_all_active_shards` in a spawn_blocking call? I am sure i am missing something obvious here.

crates/walrus-service/node_config_example.yaml
sadhansood, line 153: Also, would it make sense to rate limit the number of checks if not sample?

crates/walrus-service/src/node/consistency_check.rs
halfprice: The problem is that we need to call an async function inside a sync function.

spawn_blocking can only call sync function, but is_stored_at_all_active_shards is an async function. So we can't use it here.

The main idea from this pattern is from this blog post: https://tokio.rs/tokio/topics/bridging, where it suggests to create a mini runtime to call small async function in a sync function using block_on(). This is exactly what we are doing here.

halfprice: Another thing is that once Typed-store is migrated, and we have seek interface, we may not need this anymore, since all RocksDb access is sync. So I envision this part will become much clearer.

crates/walrus-service/src/node/metrics.rs
halfprice: I think this information can be found by the health command, as well as from metrics tracking shard status (basically the number of shards not in active status, if I understand the comment correctly)

crates/walrus-service/src/node/consistency_check.rs
halfprice: Same as above: to run async function in a sync function. I followed the recommendation from this post: https://tokio.rs/tokio/topics/bridging

crates/walrus-service/node_config_example.yaml
halfprice: That's a great suggestion! I added `sliver_data_existence_check_sample_rate_percentage` which samples blobs based on percentage. PTAL.

I made the change before seeing the suggestion to use epoch as salt. Is the idea to be deterministic or more random? I didn't do it yet but can update to it.

halfprice: Sampling added, which I feel is easier to use. For numbers, we need to know the number of total blobs before hand.

crates/walrus-service/src/node.rs
```diff
@@ -166,6 +167,7 @@ use crate::{
 
 pub mod committee;
 pub mod config;
+pub mod consistency_check;
```
liquid-helium: super nit: will pub(crate) be sufficient?

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -5,34 +5,90 @@
...
+        ))
+    } else {
+        Ok(value)
+    }
+}
+
+/// Helper struct to store the result of the blob consistency check.
+struct BlobConsistencyCheckResult {
+    /// The hash of the blob list.
+    blob_list_digest: u64,
```
liquid-helium, line 71: Question: Will `blob_list_digest` be used as a signature and compared among storage nodes?

crates/walrus-service/src/node.rs
```diff
@@ -1329,17 +1332,20 @@ impl StorageNode {
...
-        )
-        .await
-        {
-            tracing::warn!(
-                ?err,
-                epoch = %event.epoch,
-                "failed to schedule background blob info consistency check"
-            );
+        if self.inner.consistency_check_config.enable_consistency_check {
+            if let Err(err) = consistency_check::schedule_background_consistency_check(
```
liquid-helium, line 1336: Is the consistency check progress persisted? What happens if the storage node crashes after processing the epoch_change_start event?

crates/walrus-service/src/node/consistency_check.rs
halfprice: Yes. A hash value that all storage nodes should produce the same value.

crates/walrus-service/src/node.rs
halfprice: Good catch. Updated.

crates/walrus-service/src/node.rs
halfprice: > Is the consistency check progress persisted?

Not at the moment.

>What happens if the storage node crashes after processing the epoch_change_start event?

Currently, the terminated consistency check will not be re-initiated. However, I think that's fine because if there is consistency issue, the next run will discover it. On mainnet, this may be too long, so what we may need to add is the frequency to run consistency check (currently only during epoch change), to maybe twice per epoch.

crates/walrus-service/node_config_example.yaml
sadhansood: Yes basically to ensure that different sets of blobs are sampled in different epochs and we have wider coverage of blobs checked without checking all of them in any single epoch. Also, to to have consistency across nodes for the list of blobs that get picked up in an epoch (I am not sure how much important this is, so feel free to skip this)

sadhansood (approved):
I think this looks good to me, i hope @liquid-helium can also once take a look and ensure his comments are fully addressed.

crates/walrus-service/src/node.rs
liquid-helium: >so what we may need to add is the frequency to run consistency check (currently only during epoch change), to maybe twice per epoch.


Can we do the consistency check per 'initial_certified_epoch'? Also loop in @sadhansood since I saw a similar sampling with epoch as salt comment. So that we do the check only once, e.g., we only need to check the blobs certified in the last epoch, after the initial 'catchup', since this is a consistency check, not a tool to check if the operator is honest?

Also, I feel it might make things easier once we have a [checkpoint/backup](https://github.com/MystenLabs/walrus/pull/2020#issuecomment-2863803856), we can do it offline on a consistent checkpoint.

crates/walrus-service/src/node/consistency_check.rs
liquid-helium: That will require that all storage nodes produce it at the same event, right?

halfprice: Yes. The results are keyed by epoch. So for the same epoch, they should produce the same value. But they are not necessarily produce the result at the same time.

crates/walrus-service/node_config_example.yaml
halfprice: Yeah, the determinism part is a good idea. If it finds something, we can reproduce it. Let me tackle this in a followup PR.

crates/walrus-service/src/node.rs
halfprice: This tool is primarily there for checking software bugs, so I think exclusiveness is important. Also, we don't have blob deletion implemented yet, but once it is there, we also want to make sure that blobs are not accidentally deleted.

Offline consistency check is a good idea to offload the computation to somewhere else.

PR #1999, opened 2025-04-21
feat: add getters to SharedBlob

tamashi095:
## Description

Added getters that return references to the `Blob` and `Balance<WAL>` in the `SharedBlob` object. Right now, smart contracts onchain have no insight into `Blob` stored within a `SharedBlob`. For example, imagine a `Table<u256, ID>` that maps a Blob ID to an ID of a `SharedBlob`. Without access to the underlying `Blob`, it's not possible to validate that Blob ID is correct.

### DIFF
--- a/contracts/walrus/sources/system/shared_blob.move
+++ b/contracts/walrus/sources/system/shared_blob.move
@@ -50,3 +50,13 @@ public fun extend(
     system.extend_blob(&mut self.blob, extended_epochs, &mut coin);
     self.funds.join(coin.into_balance());
 }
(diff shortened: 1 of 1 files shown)


karlwuest (approved):
LGTM!

Just FYI: These changes will only become active once the contracts are upgraded, which is not planned immediately.

bwhli:
Thank you @karlwuest. Much appreciated!

tamashi095:
Thank you @karlwuest, much appreciated!

PR #2001, opened 2025-04-21
chore: add gas cost benchmark for contracts

karlwuest:
## Description

Adds a benchmark to evaluate the gas cost for calls to reserve & register blobs and to certify them.

## Test plan

ran benchmark

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/benches/gas_cost_bench.rs
+++ b/crates/walrus-sui/benches/gas_cost_bench.rs
@@ -1,0 +1,276 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+//! Gas cost benchmark for `walrus-sui`.
+
+use std::{io::Write, num::NonZeroU16, ops::Range, path::PathBuf, str::FromStr, time::Duration};
+
+use anyhow::{anyhow, bail};
+use clap::Parser;
+use sui_sdk::rpc_types::{
+    SuiTransactionBlockEffectsAPI,
+    SuiTransactionBlockResponseOptions,
+    SuiTransactionBlockResponseQuery,
+};
+use sui_types::gas::GasCostSummary;
+use walrus_core::{
+    encoding::{EncodingConfig, EncodingConfigTrait},
+    ensure,
+    merkle::Node,
+    BlobId,
+    EncodingType,
+};
+use walrus_sui::{
+    client::{BlobObjectMetadata, BlobPersistence, PostStoreAction, SuiContractClient},
+    test_utils::system_setup::initialize_contract_and_wallet_for_testing,
+};
+
+#[derive(Debug, Clone)]
+struct InputRange {
+    range: Range<u32>,
+    step: usize,
+    exp: bool,
+}
+
+impl FromStr for InputRange {
+    type Err = anyhow::Error;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut split = s.split(':');
+        let start = split.next().map(|s| s.parse()).transpose()?.unwrap_or(1);
+        let end = split
+            .next()
+            .ok_or_else(|| anyhow!("end of range must be provided"))?
+            .parse()?;
+        let step = split.next().map(|s| s.parse()).transpose()?.unwrap_or(1);
+        let exp = match split.next() {
+            None => false,
+            Some("exp") => true,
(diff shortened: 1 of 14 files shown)


crates/walrus-sui/benches/gas_cost_bench.rs
```diff
@@ -0,0 +1,276 @@
...
+            args.committee_size.get() as usize,
+        )
+        .await?;
+    // Make sure that we have enough sui to pay for gas.
+    sui_cluster_handle
+        .lock()
+        .await
+        .fund_addresses_with_sui(
+            vec![walrus_client.as_ref().address()],
+            Some(1_000_000_000_000_000),
```
jpcsmith: Nit: Use a constant for this.

crates/walrus-sui/benches/gas_cost_bench.rs
```diff
@@ -0,0 +1,276 @@
...
+        .encoded_blob_length(size)
+        .unwrap();
+
+    #[rustfmt::skip]
+        let root_hash = Node::from([
+            1, 2, 3, 4, 5, 6, 7, 8,
+            1, 2, 3, 4, 5, 6, 7, 8,
+            1, 2, 3, 4, 5, 6, 7, 8,
+            1, 2, 3, 4, 5, 6, 7, 8,
+        ]);
```
jpcsmith: Nit: Why is this indented?

scripts/plot_contract_cost.py
jpcsmith, line 1: Nit: Perhaps add a docstring at the top of the file to describe what it does.

crates/walrus-sui/src/test_utils/system_setup.rs
```diff
@@ -298,3 +310,137 @@ pub async fn end_epoch_zero(contract_client: &SuiContractClient) -> Result<()> {
...
+    // We only care about gas cost, so the actual subsidy rate can be zero to make it cheap to fund.
+    let subsidy_rate = 0;
+
+    if let Some(subsidies_pkg_id) = system_context.subsidies_pkg_id {
+        let (subsidies_object_id, _admin_cap_id) = contract_client
+            .create_and_fund_subsidies(
+                subsidies_pkg_id,
+                subsidy_rate,
+                subsidy_rate,
+                1_000_000_000_000_000, // 1M WAL
```
jpcsmith: Remark: We use a lot of flat values in the walrus-sui code to represent Sui, Mist, Wal, and Frost. It would be good if we could create some wrapper types for this (like `std::time::Duration`) and use that instead. It would generally make the code a lot more readable and remove some of the guess-work. This would then be `Wal::new(..)` or `Wal::from_frost(..)` as appropriate.

crates/walrus-sui/benches/gas_cost_bench.rs
```diff
@@ -0,0 +1,276 @@
...
+        transaction_type,
+        epochs_ahead,
+        blobs_per_call,
+        gas_cost_summary.gas_used(),
+        gas_cost_summary.net_gas_usage(),
+        gas_cost_summary.computation_cost,
+        gas_cost_summary.storage_cost,
+        gas_cost_summary.storage_rebate,
+    );
+    out_file.write(out.as_bytes())?;
```
jpcsmith: Remark: `write` may not write all of the data, I think there's a `write_all` method that could be used instead.

jpcsmith (approved):

crates/walrus-sui/benches/gas_cost_bench.rs
karlwuest: copy-paste in a refactor (+ `#[rustfmt::skip]`), fixing it

crates/walrus-sui/src/test_utils/system_setup.rs
karlwuest: I agree, created WAL-788 for this. We actually have types for this in the cli code, might make sense to move this.

crates/walrus-sui/benches/gas_cost_bench.rs
karlwuest: TIL, it seems that I didn't read the docs on `write` well enough, thanks!

PR #2004, opened 2025-04-22
feat: optimize call to reserve storage for multiple blobs

karlwuest:
## Description

Reduces the gas cost for calls to reserve storage when storing multiple blobs by buying one large storage object and splitting it instead of buying an individual storage resource for each blob.

The plots below show the computation cost before and after the optimization (note the different y-axis scales, 1e7 vs 1e6).

### Before

reserve_and_register_computation_cost_per_blob

### After

reserve_and_register_computation_cost_per_blob

## Test plan

- existing tests
- gas cost benchmarks (see #2001 )

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -1646,10 +1646,24 @@ impl SuiContractClientInner {
         );
 
         let mut pt_builder = self.transaction_builder()?;
-        for blob_metadata in blob_metadata_list.into_iter() {
-            let storage_arg =
-                reserve_space_fn(&mut pt_builder, blob_metadata.encoded_size, epochs_ahead).await?;
 
+        // Reserve enough space for all blobs
+        let mut main_storage_arg_size = blob_metadata_list
+            .iter()
+            .fold(0, |acc, metadata| acc + metadata.encoded_size);
+        let main_storage_arg =
+            reserve_space_fn(&mut pt_builder, main_storage_arg_size, epochs_ahead).await?;
+
+        for blob_metadata in blob_metadata_list.into_iter() {
+            // Split off a storage resource, unless the remainder is equal to the required size.
+            let storage_arg = if main_storage_arg_size != blob_metadata.encoded_size {
+                main_storage_arg_size -= blob_metadata.encoded_size;
+                pt_builder
+                    .split_storage_by_size(main_storage_arg.into(), main_storage_arg_size)
+                    .await?
+            } else {
+                main_storage_arg
+            };
             pt_builder
                 .register_blob(storage_arg.into(), blob_metadata, persistence)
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -1646,10 +1646,24 @@ impl SuiContractClientInner {
...
+        for blob_metadata in blob_metadata_list.into_iter() {
+            // Split off a storage resource, unless the remainder is equal to the required size.
+            let storage_arg = if main_storage_arg_size != blob_metadata.encoded_size {
+                main_storage_arg_size -= blob_metadata.encoded_size;
+                pt_builder
+                    .split_storage_by_size(main_storage_arg.into(), main_storage_arg_size)
+                    .await?
+            } else {
+                main_storage_arg
+            };
```
mlegner, line 1666: Nit: The bookkeeping with the `main_storage_arg_size` feels a bit unintuitive to me. How about instead using the `main_storage_arg` for the first blob (we know there is at least one because of the check above) and then just split off storage for all the rest?

mlegner (approved):
I love it, thanks so much for this simple but major improvement. Let's also cherry-pick this to the `v1.22` release branch.

mlegner:
@karlwuest One more question: Where does the increase of the per-blob cost above 64 blobs come from? Could we simply limit the number of blobs registered in a single PTB?

crates/walrus-sui/src/client.rs
karlwuest: I don't think that this would simplify the bookkeeping, because we still need to keep track of how large the remaining resource is and if we have multiple blobs, we always need to do a split before the first one (since the storage gets consumed by the call). We would then need to daisy-chain the calls and always change the argument that we use for splitting. 

This would look something like the following (of course, the names can be changed, e.g. `storage_remainder` instead of `main_storage_arg`). To me, this looks less clear because the sequence of operations (in the PTB) is less obvious. I initially thought about doing it this way, but it felt a bit cleaner to me to always split off the resources from the initial input. If you feel strongly about this, I'm happy to change it, but otherwise I prefer keeping it as is.

```suggestion
        for blob_metadata in blob_metadata_list.into_iter() {
            let storage_arg = main_storage_arg;
            // Split off a storage resource, unless the remainder is equal to the required size.
            if main_storage_arg_size != blob_metadata.encoded_size {
                main_storage_arg = pt_builder
                    .split_storage_by_size(main_storage_arg.into(), blob_metadata.encoded_size)
                    .await?;
                main_storage_arg_size -= blob_metadata.encoded_size;
            }
```

PR #2005, opened 2025-04-22
docs: mention that shared blob object ids can be used to retrieve blob

karlwuest:
## Description

Explicitly mention that shared blob object IDs can be used to retrieve blobs.

## Test plan

Existing tests

---

## Release notes

- [ ] Storage node:
- [x] Aggregator: Improved documentation for shared blob retrieval.
- [ ] Publisher:
- [ ] CLI:


mlegner (commented):
Thanks a lot for adding the documentation. Could you mention this [here](https://github.com/MystenLabs/walrus/blob/63faa1305dacc96a65b5d00c61ee48918013df83/docs/book/usage/web-api.md#L115) as well?

mlegner (approved):

PR #2006, opened 2025-04-22
fix(client): do not create transactions with more than 256 gas coins

giac-mysten:
## Description

Ensures that the coin selection process always returns 256 objects or less.
Since the coins are not returned in sorted order by balance, this change introduces a min-heap to keep track of the largest 256 coins encountered.

Closes WAL-429

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -2163,7 +2163,7 @@ pub async fn test_select_coins_max_objects() -> TestResult {
     let (sui_cluster_handle, _, _, _) = test_cluster::E2eTestSetupBuilder::new().build().await?;
 
     // Create a new wallet on the cluster.
-    let cluster_wallet = walrus_sui::config::load_wallet_context_from_path(
+    let mut cluster_wallet = walrus_sui::config::load_wallet_context_from_path(
         Some(
             sui_cluster_handle
                 .lock()
@@ -2181,13 +2181,17 @@ pub async fn test_select_coins_max_objects() -> TestResult {
 
     // Add 4 coins with 1 SUI each to the wallet.
     let address = wallet.as_mut().active_address()?;
-    for _ in 0..4 {
-        sui_cluster_handle
-            .lock()
-            .await
-            .fund_addresses_with_sui(vec![address], Some(sui(1)))
-            .await?;
-    }
+    walrus_sui::test_utils::fund_addresses(&mut cluster_wallet, vec![address; 4], Some(sui(1)))
+        .await?;
+
+    let balance = wallet
+        .as_mut()
+        .get_client()
+        .await?
+        .coin_read_api()
+        .get_balance(address, None)
+        .await?;
+    assert_eq!(balance.total_balance, u128::from(sui(4)));
 
     // Create a new client with the funded wallet.
     let retry_client =
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -65,9 +65,13 @@ use walrus_sui::{
         SuiClientError,
         SuiContractClient,
         UpgradeType,
+        retry_client::RetriableSuiClient,
     },
     system_setup::copy_recursively,
-    test_utils::system_setup::{development_contract_dir, testnet_contract_dir},
+    test_utils::{
+        self,
+        system_setup::{development_contract_dir, testnet_contract_dir},
+    },
     types::{
         Blob,
         BlobEvent,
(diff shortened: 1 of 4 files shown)


crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -183,22 +194,72 @@ impl RetriableSuiClient {
...
+                if enough_avail {
+                    tracing::error!(
+                        requested_amount = amount,
+                        max_available_with_256_coins = total_selected,
+                        total_available,
+                        "there is enough balance to cover the requested amount, but cannot be \
+                        achieved with less than 256 coins; consider merging the coins in the \
+                        wallet and retrying",
+                    );
+                }
```
giac-mysten: Question: would be nice to have another error here for `InsufficientFundsWithMaxCoins`, or something similar. Though this would mean changing the error type to `SuiClientError`. Thoughts?

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -183,22 +194,72 @@ impl RetriableSuiClient {
...
-                ready
-            })
-            .collect::<Vec<_>>()
-            .await;
+            .filter(|coin: &Coin| future::ready(!exclude.contains(&coin.coin_object_id))));
+
+        // Use a binary heap to keep track of the coins with the largest balances.
+        // `Reverse` turns the max heap into a min heap.
+        let mut selected_coins =
+            BinaryHeap::<Reverse<OrderedCoin>>::with_capacity(MAX_GAS_PAYMENT_OBJECTS);
```
giac-mysten: Question: another options would be to keep a vector until we reach 256 elements, and only if that threshold is exceeded we continue with the binary heap.
The performance gain is not too big as the push is O(1) on average, but may make a difference? Thoughts?

liquid-helium: I feel the difference could be minimal, maybe in the order of ms?
But on the other hand, it depends on which is the usual pattern, if in most of the cases, less than 256 coins will cover the cost, then the heap is an overkill in these cases.

Also since building a heap form a vector is O(n), hence your suggestion of using a vector to covert the common cases, and transit to a heap only when more than 256 coins are used is a good idea.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
liquid-helium: Adding a new error type with a clear error message you currently logged "there is enough balance to cover the requested amount, but cannot be achieved with less than 256 coins; consider merging the coins in the wallet and retrying" could be helpful.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -183,22 +194,73 @@ impl RetriableSuiClient {
...
+        let mut total_selected = 0u128;
+        let mut total_available = 0u128;
+
+        while let Some(coin) = coins_stream.as_mut().next().await {
+            let coin_balance = coin.balance as u128;
```
mlegner: Nit: Can you use `into` or `from` to emphasize that this conversion always succeeds? (Also below.)

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
mlegner: I don't think it matters a lot, the execution time will be dominated by the RPC requests anyway. But I like @liquid-helium's suggestion of starting with a vector and then converting it if necessary.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
mlegner: Agreed. Also, make sure we explicitly mention that this is about SUI (as opposed to WAL).

mlegner (commented):
Thanks, @giac-mysten, LGTM. 💯

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -2148,3 +2153,78 @@ async fn test_ptb_retriable_error() -> TestResult {
...
+            .lock()
+            .await
+            .fund_addresses_with_sui(vec![address], Some(sui(1)))
+            .await?;
+    }
+
+    // Create a new client with the funded wallet.
+    let retry_client =
+        RetriableSuiClient::new_from_wallet(&wallet.inner, ExponentialBackoffConfig::default())
+            .await?;
```
giac-mysten: Nit: not sure this is the best way to create a retriable client on the test cluster. Open to suggestions here.

PR #2007, opened 2025-04-22
chore: upgrade to rust edition 2024

wbbradley:
## Description

Upgrade to Rust Edition 2024.

## Test plan

CI Pipeline.

### DIFF
--- a/.rustfmt.toml
+++ b/.rustfmt.toml
@@ -1,0 +1,28 @@
+# TODO: WAL-795 Upgrade style_edition to 2024
+style_edition = "2021"
+
+# Note that the following flags are still in the nightly version of rustfmt, and must be manually
+# supplied to your IDE. Generally the simplest way to do this is to add them to your `rust-analyzer`
+# settings.
+#
+# group_imports = "StdExternalCrate"
+# imports_granularity = "Crate"
+# imports_layout = "HorizontalVertical"
+#
+# To do so in Neovim, for example, add the following to your `init.lua`:
+#
+# vim.g.rustaceanvim = {
+#   server = {
+#     default_settings = {
+#       ['rust-analyzer'] = {
+#         rustfmt = {
+#           extraArgs = {
+#             "--config",
+#             "group_imports=StdExternalCrate,imports_granularity=Crate,imports_layout=HorizontalVertical"
+#           }
+#         },
+#       }
+#     }
+#   }
+# }
+#
--- a/crates/walrus-sdk/src/config/committees_refresh_config.rs
+++ b/crates/walrus-sdk/src/config/committees_refresh_config.rs
@@ -41,30 +41,13 @@ pub struct CommitteesRefreshConfig {
 }
 
 impl CommitteesRefreshConfig {
-    /// Builds a new [`CommitteesRefresher`] and [`CommitteesRefresherHandle`].
-    pub async fn build_refresher_and_handle(
-        &self,
-        sui_client: impl ReadClient,
-    ) -> Result<(
-        CommitteesRefresher<impl ReadClient>,
-        CommitteesRefresherHandle,
-    )> {
-        let notify = Arc::new(Notify::new());
-        let (req_tx, req_rx) = mpsc::channel(self.refresher_channel_size);
-
-        let refresher =
-            CommitteesRefresher::new(self.clone(), sui_client, req_rx, notify.clone()).await?;
-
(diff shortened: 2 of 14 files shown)


crates/walrus-core/src/by_axis.rs
```diff
@@ -183,7 +183,7 @@ impl From<Axis> for ByAxis<(), ()> {
 /// and `Secondary` variants.
 #[macro_export]
 macro_rules! flat_map {
-    ($by_axis:expr, $($fn_once:tt)*) => {
+    ($by_axis:expr_2021, $($fn_once:tt)*) => {
```
mlegner: Suggestion: Keep the `expr` everywhere unless we really need the old matching behavior. See also the [migration recommendations](https://doc.rust-lang.org/edition-guide/rust-2024/macro-fragment-specifiers.html#migration).

.rustfmt.toml
```diff
@@ -0,0 +1,27 @@
+style_edition = "2024"
```
mlegner: AFAICT most of the changes in this PR are formatting changes. Could we do those separately by setting `style_edition = "2021"` here?

mlegner (commented):
Thanks for the housekeeping. Two initial comments from my side.

crates/walrus-core/src/by_axis.rs
wbbradley: Done.

.rustfmt.toml
wbbradley: Done.

.rustfmt.toml
```diff
@@ -0,0 +1,27 @@
+style_edition = "2021"
```
mlegner, line 2: Nit: Unless you intend to do the reformatting immediately, we should create a follow-up issue for changing this to `2024`.

crates/walrus-service/src/node.rs
```diff
@@ -644,7 +644,7 @@ impl StorageNode {
         let epoch_change_driver = EpochChangeDriver::new(
             system_parameters,
             contract_service.clone(),
-            StdRng::seed_from_u64(thread_rng().gen()),
+            StdRng::seed_from_u64(thread_rng().#gen()),
```
mlegner: Question: (Why) doesn't that have to be `r#gen()`?

crates/walrus-sdk/src/config/committees_refresh_config.rs
mlegner, line 1: Question: How is this related to the edition change? Is this the result of some new lint?

crates/walrus-sdk/src/utils.rs
```diff
@@ -205,23 +205,27 @@ where
...
-                self.being_executed.push(future);
-            } else {
-                break;
+            match self.futures.next() {
+                Some(future) => {
+                    self.being_executed.push(future);
+                }
+                _ => {
+                    break;
+                }
```
mlegner: Question: I see this change in a few places, is this a new lint that discourages the use of `if let Some(..) {} else {}`?

mlegner (approved):
Thanks a lot for the changes, @wbbradley. A few questions, otherwise LGTM. 👍

.rustfmt.toml
wbbradley: Good call.

crates/walrus-service/src/node.rs
wbbradley: Because `gen` was made a keyword in 2024. Since `rand` crate is not part of the std library, it got stuck with this.

crates/walrus-sdk/src/utils.rs
wbbradley: It's not a new lint _per se_ but it is part of the `cargo fix --edition 2024` changes. I've since backed those out as I agree with the sentiment expressed [here](https://codeandbitters.com/rust-2024-upgrade/#shorter-if-let-temporary-lifetimes).

crates/walrus-service/src/node.rs
mlegner: Yes, I understand that. But in other places it is `r#gen` and here just `#gen` (without the `r`).

wbbradley: Ah. Good catch. I think that was a typo. Let me double check.

wbbradley: Yes, in that particular revision, that was a typo. It's fixed in the current branch.

PR #2009, opened 2025-04-22
feat(publisher): allow blob sharing from put_blob

wbbradley:
## Description

Allow `put_blob` requests to share blobs.

## Test plan

TBD


giac-mysten (approved):
Thank you @wbbradley ! LGTM

PR #2010, opened 2025-04-23
feat: Optimize Storage Selection Algorithm for Batch Blob Operations

SatoshiIsHere:
## Description

Optimized storage selection algorithm for batch blob operations (WAL-363) by replacing the inefficient approach that made repeated RPC calls for each blob with a more efficient batch processing method. The new implementation fetches all storage resources in a single API call and processes them in memory, significantly reducing network round trips.

## Test plan

Tested the changes by:
Running full release build (cargo build --release)
Running complete test suite (cargo test --release)

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Improved performance for batch blob operations by optimizing storage selection. Operations with multiple blobs are now significantly faster with no change to user workflow.

### DIFF
--- a/crates/walrus-sdk/src/client/resource.rs
+++ b/crates/walrus-sdk/src/client/resource.rs
@@ -457,19 +457,14 @@ impl<'a> ResourceManager<'a> {
         let mut extended_blobs = Vec::with_capacity(max_len);
         let mut extended_blobs_noncertified = Vec::with_capacity(max_len);
 
-        // This keeps tracks of selected storage objects and exclude them from selecting again.
-        let mut excluded = Vec::with_capacity(max_len);
-
         // Gets the owned blobs once for all checks, to avoid multiple calls to the RPC.
         let owned_blobs = self
             .sui_client
             .owned_blobs(None, ExpirySelectionPolicy::Valid)
             .await?;
 
-        // For all the metadata, if the blob is registered in wallet, add it directly to results.
-        // Otherwise, check if there is existing storage resource selected for the encoded length,
-        // add it to reused_metadata_with_storage and its length to reused_encoded_lengths.
-        // Otherwise, add it to new_metadata_list and its length to new_encoded_lengths.
+        let mut blob_processing_items = Vec::with_capacity(max_len);
+
         for (metadata, encoded_length) in metadata_list.iter().zip(encoded_lengths) {
             if let Some(blob) = self
                 .find_blob_owned_by_wallet(
@@ -519,35 +514,53 @@ impl<'a> ResourceManager<'a> {
                         },
                     ));
                 }
-            } else if let Some(storage_resource) = self
-                .sui_client
-                .owned_storage_for_size_and_epoch(
-                    *encoded_length,
-                    epochs_ahead + self.write_committee_epoch,
-                    &excluded,
-                )
-                .await?
-            {
-                // TODO(joy): Currently select is done one at a time for each blob using `excluded`
-                // to filter, this might not be efficient if the list is too long, consider better
-                // storage selection strategy (WAL-363).
-                // TODO(giac): consider splitting the storage before reusing it (WAL-208).
-                tracing::debug!(
-                    blob_id=%metadata.blob_id(),
-                    storage_object=%storage_resource.id,
-                    "using an existing storage resource to register the blob"
-                );
-                excluded.push(storage_resource.id);
-                reused_metadata_with_storage.push(((*metadata).try_into()?, storage_resource));
-                reused_encoded_lengths.push(*encoded_length);
             } else {
-                tracing::debug!(
-                    blob_id=%metadata.blob_id(),
-                    "no storage resource found for the blob"
-                );
-                new_metadata_list.push(*metadata);
-                new_encoded_lengths.push(*encoded_length);
-            };
+                blob_processing_items.push((*metadata, *encoded_length));
+            }
+        }
+
+        if !blob_processing_items.is_empty() {
+            let all_storage_resources = self
+                .sui_client
+                .owned_storage(ExpirySelectionPolicy::Valid)
+                .await?;
+
+            let target_epoch = epochs_ahead + self.write_committee_epoch;
+            let mut available_resources: Vec<_> = all_storage_resources
+                .into_iter()
+                .filter(|storage| storage.end_epoch >= target_epoch)
+                .collect();
+
+            blob_processing_items.sort_by(|(_, size_a), (_, size_b)| size_b.cmp(size_a));
+
+            for (metadata, encoded_length) in blob_processing_items {
+                let best_resource_idx = available_resources
+                    .iter()
+                    .enumerate()
+                    .filter(|(_, storage)| storage.storage_size >= encoded_length)
+                    .min_by_key(|(_, storage)| storage.storage_size - encoded_length)
+                    .map(|(idx, _)| idx);
+
+                if let Some(idx) = best_resource_idx {
+                    let storage_resource = available_resources.remove(idx);
+
+                    tracing::debug!(
+                        blob_id=%metadata.blob_id(),
+                        storage_object=%storage_resource.id,
+                        "using an existing storage resource to register the blob"
+                    );
+
+                    reused_metadata_with_storage.push((metadata.try_into()?, storage_resource));
+                    reused_encoded_lengths.push(encoded_length);
+                } else {
+                    tracing::debug!(
+                        blob_id=%metadata.blob_id(),
+                        "no storage resource found for the blob"
+                    );
+                    new_metadata_list.push(metadata);
+                    new_encoded_lengths.push(encoded_length);
+                }
+            }
         }
 
         // Register all in reused_metadata_with_storage in one ptb.


crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,53 @@ impl<'a> ResourceManager<'a> {
                         },
                     ));
                 }
-            } else if let Some(storage_resource) = self
-                .sui_client
-                .owned_storage_for_size_and_epoch(
```
mlegner, line 524: Given this is no longer used, the corresponding function `owned_storage_for_size_and_epoch` can be removed.

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,53 @@ impl<'a> ResourceManager<'a> {
...
-                    *encoded_length,
-                    epochs_ahead + self.write_committee_epoch,
-                    &excluded,
-                )
-                .await?
-            {
-                // TODO(joy): Currently select is done one at a time for each blob using `excluded`
-                // to filter, this might not be efficient if the list is too long, consider better
-                // storage selection strategy (WAL-363).
-                // TODO(giac): consider splitting the storage before reusing it (WAL-208).
```
mlegner, line 534: Important: We need to move this TODO below as this is still relevant.

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,53 @@ impl<'a> ResourceManager<'a> {
...
+
+            blob_processing_items.sort_by(|(_, size_a), (_, size_b)| size_b.cmp(size_a));
+
+            for (metadata, encoded_length) in blob_processing_items {
+                let best_resource_idx = available_resources
+                    .iter()
+                    .enumerate()
+                    .filter(|(_, storage)| storage.storage_size >= encoded_length)
+                    .min_by_key(|(_, storage)| storage.storage_size - encoded_length)
+                    .map(|(idx, _)| idx);
```
mlegner: Remark: This is a bit inefficient; we could also sort the `available_resources` to find the best resource more quickly.

Also, suggestion: If there are multiple resources of the same size, we should select the one with the lowest `end_epoch`.

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,53 @@ impl<'a> ResourceManager<'a> {
...
+
+                if let Some(idx) = best_resource_idx {
+                    let storage_resource = available_resources.remove(idx);
```
mlegner: Suggestion: Removing items from a vector is not very efficient. Unless you decide to sort the `available_resources` to optimize something, I'd suggest to use `swap_remove` instead.

mlegner (changes requested):
Thanks a lot for the contribution, @SatoshiIsHere! This is definitely a major improvement over the existing mechanism.

I have a few more specific comments. Maybe @liquid-helium may want to take a brief look as well.

crates/walrus-sdk/src/client/resource.rs
0xKyungmin: removed

crates/walrus-sdk/src/client/resource.rs
0xKyungmin: restored the comment, but I think if need to keep the comment, the wording needs to be revised

crates/walrus-sdk/src/client/resource.rs
0xKyungmin: Applied the change, Thanks for helpful information

crates/walrus-sdk/src/client/resource.rs
0xKyungmin: Applied the change, Thanks for helpful information and review

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,60 @@ impl<'a> ResourceManager<'a> {
...
-                new_metadata_list.push(*metadata);
-                new_encoded_lengths.push(*encoded_length);
-            };
+                blob_processing_items.push((*metadata, *encoded_length));
+            }
+        }
+
+        // TODO(joy): Currently select is done one at a time for each blob using `excluded`
+        // to filter, this might not be efficient if the list is too long, consider better
+        // storage selection strategy (WAL-363).
```
mlegner: Sorry, I wasn't fully clear in my previous comment. I just wanted to keep the comment about splitting.
```suggestion
```

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,60 @@ impl<'a> ResourceManager<'a> {
...
+                }
+            });
+
+            for (metadata, encoded_length) in blob_processing_items {
+                let best_resource_idx = available_resources
+                    .iter()
+                    .position(|storage| storage.storage_size >= encoded_length);
+                
+                if let Some(idx) = best_resource_idx {
+                    let storage_resource = available_resources.swap_remove(idx);
```
mlegner, line 553: Important: As you now expect a sorted vector, you can't use `swap_remove` as it destroys the ordering.

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,60 @@ impl<'a> ResourceManager<'a> {
...
+                match a.storage_size.cmp(&b.storage_size) {
+                    std::cmp::Ordering::Equal => a.end_epoch.cmp(&b.end_epoch),
+                    other => other,
+                }
+            });
+
+            for (metadata, encoded_length) in blob_processing_items {
+                let best_resource_idx = available_resources
+                    .iter()
+                    .position(|storage| storage.storage_size >= encoded_length);
```
mlegner: Remark: The current implementation is still `O(m*n)` where `m` is the number of processing items and `n` is the number of resources. I think you should be able to get something like `O(m+n)` by storing the last index into the `available_resources` and moving from there. However, this would be a bit more complicated.

Maybe the easiest would be to skip the sorting of the `available_resources` after all, then you can keep the `swap_remove` below and the code is very simple.

mlegner (commented):
Thanks for the changes @SatoshiIsHere! A few more minor comments.

crates/walrus-sdk/src/client/resource.rs
0xKyungmin: applied, thanks you

crates/walrus-sdk/src/client/resource.rs
0xKyungmin: I missed it. Thanks for letting me know

crates/walrus-sdk/src/client/resource.rs
0xKyungmin: applied

mlegner: Suggestion: Change this back to something like the original version:
```suggestion
                let best_resource_idx = available_resources
                    .iter()
                    .enumerate()
                    .filter(|(_, storage)| storage.storage_size >= encoded_length)
                    .min_by(|(_, storage_a), (_, storage_b)| {
                        match storage_a.storage_size.cmp(&storage_b.storage_size) {
                            std::cmp::Ordering::Equal => {
                                storage_a.end_epoch.cmp(&storage_b.end_epoch)
                            }
                            ordering => ordering,
                        }
                    })
                    .map(|(idx, _)| idx);
```

0xKyungmin: I initially used .position() for efficiency (stops early) after sorting, but your explicit .min_by() approach is more robust since it doesn't rely on pre-sorting. I've updated the code with your suggestion.
Thanks for your review

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,62 @@ impl<'a> ResourceManager<'a> {
...
+                .filter(|storage| storage.end_epoch >= target_epoch)
+                .collect();
+
+            blob_processing_items.sort_by(|(_, size_a), (_, size_b)| size_b.cmp(size_a));
+
+            for (metadata, encoded_length) in blob_processing_items {
+                let best_resource_idx = available_resources
+                    .iter()
+                    .position(|storage| storage.storage_size >= encoded_length);
+
```
mlegner: Nit: Remove the left-over code here:
```suggestion
```

mlegner (approved):
Thanks for the patience, LGTM! 💯

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -519,35 +514,58 @@ impl<'a> ResourceManager<'a> {
...
+                    .min_by(|(_, storage_a), (_, storage_b)| {
+                        match storage_a.storage_size.cmp(&storage_b.storage_size) {
+                            std::cmp::Ordering::Equal => {
+                                storage_a.end_epoch.cmp(&storage_b.end_epoch)
+                            }
+                            ordering => ordering,
+                        }
+                    })
+                    .map(|(idx, _)| idx);
+
```
mlegner, line 551: This got accidentally removed:
```suggestion

                if let Some(idx) = best_resource_idx {
                    let storage_resource = available_resources.swap_remove(idx);
```

PR #2015, opened 2025-04-23
fix(publisher): add force and refactor query API

giac-mysten:
## Description

- Adds the ability to `force` a store on the publisher.
- Refactors the publisher's query parsing, to ensure `share` and `send_object_to` are mutually exclusive.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [x] Publisher: Adds the possiblity to `force` a store operation, like with the CLI.
- [ ] CLI:


wbbradley (approved):
LGTM

Issue #1920, opened 2025-04-08
Allow creating DB checkpoints while the storage node is running


mlegner:
### Description

This would allow creating backups without having to stop the node or risk corrupted states. See https://github.com/facebook/rocksdb/wiki/Checkpoints.

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct.

tamashi095:
Took a brief look at the docs for this feature, and I'm not sure how well it would work in practice. It seems to create a point-in-time consistent snapshot by hard-linking the live SST and MANIFEST files into a **separate directory**. This means node operators could need a ton of spare capacity just to perform the snapshot. I don't think mounting an S3 bucket to a directory would work for this because it requires hard links.

Personally, I think data backup for Walrus is out of scope of the node software. There are plenty of filesystem level tools for snapshotting (e.g. `zfs send` which can be directly streamed to offsite storage). What could be useful is some CLI command that temporarily suspends reads/writes for a few seconds to allow for a snapshot to happen.

poelzi:
I think the [BackupManager](https://github.com/facebook/rocksdb/wiki/How-to-backup-RocksDB) might be better. I used it in a ceph mon backup feature, but don't know how it will work with the blob feature.
To be honest, use a filesystem with snapshots like zfs or bcachefs (experimental). Since rocksdb usse a write ahead log (WAL), opening the snapshot will rollback to the last checkpoint. No corruption, just a transaction gets rolled back.

mario4tier:
@poelzi That is the big question to me: are we 100% sure that ZFS snapshots done "independently at any moment" always capture a valid DB state while walrus-node keeps writing?

I am wondering why the recent corruptions (on walrus-node kill) happen. In principle, should it have resumed on a last valid WAL writing? I don't know the answer, but may be there is another app level "atomicity expectation" (say between multiple DB or consecutive writes) and the snapshot/checkpoints would need to happen in between these "app level transaction". 

===

BTW, we have tested your ZFS suggestion with testnet, and it works... but still wondering if it might be "luck" so we keep multiple rolling snapshots for higher probability of holding a good one.

poelzi:
@mario4tier I can tell you from 15 years devops experience, it's always messy. I looked into the walrus storage code and could not find usage of rocksdb::Transaction. I'm not sure how the rocksdb WAL is used when not using transactions api.
What kind of corruption did you have ?
[...]

PR #2020, opened 2025-04-25
feat: checkpoint Walrus RocksDB database

liquid-helium:
## Description

* Added a  `CheckpointManager` component, which provides
    * Periodic checkpoint of the main database
    * Interface for manual checkpoint
* Introduced a  `node admin socket` so that operators can interact directly with a running node
    * Added Checkpoint commands in walrus-node command tool

contributes to WAL-842
Contributes to #1920 

## Test plan

---

## Release notes

For each box you, include information after the relevant heading that describes the impact of your changes that

- [x] Storage node:
* Added a  `CheckpointManager` component, which provides
    * Periodic checkpoint of the main database
    * Interface for manual checkpoint
* Introduced a  `node admin socket` so that operators can interact directly with a running node
    * Added Checkpoint commands in walrus-node command tool
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/db_checkpoint.rs
+++ b/crates/walrus-service/src/node/db_checkpoint.rs
@@ -27,6 +27,7 @@ pub struct DbCheckpointConfig {
     /// Note: It is strongly recommended to locate this directory on a separate physical disk
     /// from the main database to avoid potential I/O performance degradation.
     // TODO(WAL-843): Add a check to ensure the directory is on a separate physical disk.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     pub db_checkpoint_dir: Option<PathBuf>,
     /// Maximum number of db_checkpoints to keep.
     pub max_db_checkpoints: usize,
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -165,7 +165,6 @@ consistency_check:
   enable_sliver_data_existence_check: false
   sliver_data_existence_check_sample_rate_percentage: 100
 checkpoint_config:
-  db_checkpoint_dir: null
   max_db_checkpoints: 3
   db_checkpoint_interval:
     secs: 86400


crates/walrus-service/src/node/checkpoint.rs
```diff
@@ -0,0 +1,820 @@
...
+    const CHECKPOINT_CREATION_INITIAL_DELAY: time::Duration = time::Duration::from_secs(15 * 60);
+    /// Delay between checkpoint creation retries.
+    const CHECKPOINT_CREATION_RETRY_DELAY: time::Duration = time::Duration::from_secs(300);
+    /// Default timeout for tasks.
+    const DEFAULT_TASK_TIMEOUT: StdDuration = time::Duration::from_secs(60 * 60);
+
+    /// Create a new checkpoint manager for RocksDB.
+    pub async fn new(db: Arc<RocksDB>, config: CheckpointConfig) -> Result<Self, CheckpointError> {
+        if let Some(checkpoint_dir) = config.checkpoint_dir.as_ref() {
+            Self::create_checkpoint_dir_if_not_exists(checkpoint_dir)?;
```
wbbradley: What happens if `config.checkpoint_dir` is None? Will it fail later due to the absence of a checkpoint dir?

crates/walrus-service/src/node/checkpoint.rs
```diff
@@ -0,0 +1,820 @@
...
+        tracing::info!("restoring database from latest backup");
+        engine
+            .restore_from_latest_backup(db_path, wal_path, &restore_opts)
+            .map_err(|e| CheckpointError::Other(anyhow::anyhow!("restore error: {}", e)))?;
+
+        tracing::info!("database restored successfully");
+        Ok(())
+    }
+
+    fn create_backup_impl(
```
wbbradley: Consider using tracing::instrument here, instead of hand-rolling the duration tracking.

crates/walrus-service/src/node/checkpoint.rs
```diff
@@ -0,0 +1,820 @@
...
+                    "rocksDB backup failed",
+                );
+                Err(e)
+            }
+        }
+    }
+
+    /// Create the checkpoint directory if it doesn't exist.
+    fn create_checkpoint_dir_if_not_exists(checkpoint_dir: &Path) -> Result<(), CheckpointError> {
+        match checkpoint_dir.try_exists() {
```
wbbradley: You can just call create_dir_all, don't check whether it already exists. It does this already and succeeds if the dir already exists.

scripts/local-testbed.sh
```diff
@@ -182,7 +182,8 @@ if ! $use_existing_config; then
     --with-subsidies
 
   # Generate configs
-  generate_dry_run_args=( --working-dir "$working_dir" )
+  generate_dry_run_args=( --working-dir "$working_dir" --use-legacy-event-provider )
```
wbbradley: I don't think we want to check this in?

crates/walrus-service/bin/node.rs
```diff
@@ -152,6 +166,59 @@ enum Commands {
...
+
+    /// Admin commands for managing a running node.
+    Admin {
+        /// Admin subcommand to execute.
+        #[command(subcommand)]
+        command: AdminCommands,
+        /// Path to the admin socket.
+        #[arg(long)]
+        socket_path: PathBuf,
+    },
```
mlegner: A few remarks here:
1. We already have a `node-admin` subcommand for the `walrus` CLI. I guess the difference is that the `walrus node-admin` contains actions that can be executed on a separate host while the newly added ones are executed on the host itself. Still, it might make sense to either combine them or clearly document why they are separate. (cc @karlwuest)
2. There is a `walrus-node db-tool` subcommand that has some overlap as well. (cc @halfprice)
3. @sadhansood added a local HTTP endpoint to manage the log level of a running node in #1829. Maybe that feature could use the socket added here?

crates/walrus-service/bin/node.rs
```diff
@@ -152,6 +166,59 @@ enum Commands {
...
+}
+
+/// Admin subcommands for remote node management.
+#[derive(Subcommand, Debug, Clone, Serialize, Deserialize)]
+enum AdminCommands {
+    /// Checkpoint management.
+    Checkpoint {
```
mlegner, line 185: Important (here and in other places): We should make it clear in names (variables, files, modules, structs, enums) and docstrings that this is about DB checkpoints and not Sui checkpoints to prevent confusion. For example, call this `DbCheckpoint` or something like this.

mlegner (commented):
Thanks for implementing this, @liquid-helium! I have two initial high-level comments. As I'm not too familiar with RocksDB checkpoints, nor with UNIX sockets in Rust, I'd prefer if somebody else would take a closer look at the details there.

crates/walrus-service/bin/node.rs
liquid-helium: Thanks @mlegner for all the info, I agree we can consolidate or make the purposes more explicit.

>We already have a node-admin subcommand for the walrus CLI. I guess the difference is that the walrus node-admin contains actions that can be executed on a separate host while the newly added ones are executed on the host itself. Still, it might make sense to either combine them or clearly document why they are separate. (cc @karlwuest)

I just checked the node-admin commands, and it appears that all of them are on-chain operations (please correct me if I missed anything, @karlwuest). The new walrus-node Admin commands are only executed locally, as you pointed out. I will add more comments to the checkpoint command to make it clear why it is done this way.

>There is a walrus-node db-tool subcommand that has some overlap as well. (cc @halfprice)

I considered integrating the checkpoint command into the db-tool; however, the issue is that the operator would need to stop the walrus-node before using the checkpoint command, because they need to open the RocksDB in the exclusive write-mode in order to create a consistent backup. Creating a consistent backup in read-only mode would be complicated. And since restarting the walrus-node can be slow, it’s preferable to avoid that approach.

>@sadhansood added a local HTTP endpoint to manage the log level of a running node in https://github.com/MystenLabs/walrus/pull/1829. Maybe that feature could use the socket added here?

I wasn't aware of the feature @sadhansood is implementing. Currently, I'm using this Unix socket primarily for security reasons, but I'm open to discussing how we might combine them. I'm also willing to switch to the local HTTP endpoint if it can be ensured that this won't introduce a security risk, such as allowing others to trigger checkpoints maliciously.

sadhansood:
Thanks @liquid-helium for this change! I have one question - what happens to the checkpoints that we create periodically or on-demand? Wouldn't it increase the disk usage once created if not copied out to some archive/cold storage solution as time goes on and increase operational burden. One way we dealt with this in Sui is in `db_checkpoint_handler.rs` where we also manage the second part of moving the local db into a remote store once it is backed up.

crates/walrus-service/src/node/checkpoint.rs
liquid-helium: Yes, this is intentional, and the checkpoint_dir is None by default. This is to make sure users understand that they are going to create a checkpoint at a certain dir.

crates/walrus-service/src/node/checkpoint.rs
liquid-helium: Done. Thanks for the tip.

crates/walrus-service/src/node/checkpoint.rs
liquid-helium: Nice tip!

mario4tier:
> ... The disk usage is definitely something we need to consider, and storing checkpoint in the same storage is a counter-pattern.

In our case, creating fast and cheap hard link checkpoints on same storage would be preferred. A DB checkpoint would be followed by a ZFS snapshot of the whole storage (data+checkpoints).

The backup to a separate storage would then be the node operator responsibility (ZFS send/receive in our case).

My point being: consider keeping the feature of creating/recovering from hard links checkpoints ... and develop remote backup as a separate optional step/feature.

(seems you are already "there", just sharing my use case)

crates/walrus-service/bin/node.rs
liquid-helium: Good catch, it is interesting that we are actually using `Backup` feature from RocksDB, but in order to avoid collision with our backup, i chose Checkpoint, which led to the current collision.
Changed to DbCheckpoint for all the uses.

crates/walrus-service/bin/node.rs
karlwuest: Sorry for being a bit late to this thread.

Yes, the `walrus node-admin` commands are commands that can be used on different machines (depending on the delegation of the authorization in the contracts) and are commands that initiate on-chain operations. 

I agree with @mlegner that we should have some documentation about what makes these commands different and ideally, we should use different names (as `walrus-node admin` and `walrus node-admin` can be quite confusing 😅). Would it be possible to use a different name for the command in this PR?

mario4tier:
> Feel free to create a separate issue to track this and assign it to me. 

I felt that implementing DB checkpoints management (e.g. trigger, recovery...) is a common layer (for all use case) upon which the additional remote backup features would be built.

So I was hopeful that "relatively simpler" hard link checkpoints management could be exposed on the CLI as part (or even before) doing the "more complicated" remote backup :smile:

Not sure if I am the right person to create another PR (in addition of #1920). I prefer to leave this to you as you have more insight on how it is to be implemented.

mario4tier:
| The current implementation focuses on scenarios where the disk is damaged, using BackupEngine for recovery. 

Good, can't dispute the value of backups :smile:

The rest is my opinion about **priority** of implementing hard link checkpoints versus remote backup:

Worst-case scenario
====
- DB corruptions can have a bigger impact than drive failures. The most dramatic would be a walrus-node bug/upgrade that would affect multiple operators "at once".

- There were multiple instances of DB corruptions, likely due to walrus-node crash/kill, but not fully understand (AFAIK).

- Drive failures are already mitigated (to different degree) by many operators. It sucks when it happens, but it has a low probability of global impact.

Hard-links checkpoint is the "cheapest" solution against the "worst" scenario that we do not fully control/understand.

And "cheap" means also a better chance of being actually enabled by most operators on short term (if one did not bother already with RAID-like protection on live network, then lower chance it will invest into costly remote backup soon).

===

Just to be clear, I expect most responsible operator to want, and eventually use remote backups.

mario4tier:
Thanks for doing all this!

**Details**
In time, as you see fit, I can share measurements of daily data "turnover" (by design, this is easy to monitor with daily ZFS snapshots). That includes the cost of keeping a reference on "older versions" of the SST files.

In short, for our 21 mainnet shards, it currently costs 100GB to 300GB per day to hold to a snapshot of the whole DB... with the highest numbers when a compaction occurred (as you expected).

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+                                    TaskResult::Failed(DbCheckpointError::CheckpointInProgress)
+                                );
+                            } else {
+                                let db_clone = db.clone();
+                                let config_clone = config.clone();
+                                current_task = Some(Arc::new(DelayedTask::new(
+                                    time::Instant::now() + delay.unwrap_or_default(),
+                                    Self::DEFAULT_TASK_TIMEOUT,
+                                    move || {
+                                        let result = Self::create_backup_impl(
```
halfprice, line 394: Important: IIUC, this task currently runs on the tokio main event loop. If so, we should move it to a blocking thread (using tokio::task::spawn_blocking) to not blocking regular task processing.

crates/walrus-service/bin/node.rs
```diff
@@ -1218,12 +1345,142 @@ impl StorageNodeRuntime {
...
+                            message: error,
+                        },
+                    }
+                }
+                Err(e) => AdminCommandResponse {
+                    success: false,
+                    message: format!("Failed to parse command: {}", e),
+                },
+            }
+        } else {
```
halfprice: Nit: I feel we can re-order some of the logic here so that `handle_connection` can be used for other commands as well

- reading command
  - check command type
    - if checkpoint command, then try to get the checkpoint manager.
    - if other command, to other stuff....

Then, maybe we can remove

```
if admin_args.checkpoint_manager.is_none() {
            tracing::warn!("checkpoint manager is not initialized, skipping admin socket");
            return Ok(None);
        }
```

in start_admin_socket(), and return error here if checkpoint_manager does not exist.

crates/walrus-service/src/node/config.rs
```diff
@@ -171,6 +171,12 @@ pub struct StorageNodeConfig {
     /// Configuration for the blocking thread pool.
     #[serde(default, skip_serializing_if = "defaults::is_default")]
     pub thread_pool: ThreadPoolConfig,
+    /// Configuration for the checkpointing task.
+    #[serde(default, skip_serializing_if = "defaults::is_default")]
+    pub checkpoint_config: DbCheckpointConfig,
+    /// Admin socket path.
+    #[serde(default, skip_serializing_if = "defaults::is_none")]
+    pub admin_socket_path: Option<PathBuf>,
```
halfprice, line 187: Nit: can you add an example how to set this path? Not sure exactly how to use it.

crates/walrus-service/bin/node.rs
```diff
@@ -152,6 +166,59 @@ enum Commands {
...
+    /// Whether the command was successful.
+    success: bool,
+    /// A message describing the command.
+    message: String,
+}
+
+/// Commands for checkpoint management.
+#[derive(Subcommand, Debug, Clone, Serialize, Deserialize)]
+#[command(rename_all = "kebab-case")]
+enum CheckpointCommands {
```
halfprice, line 206: Nit: I think this checkpoint command currently only works on the main DB that stores the sliver data. Not event db or event writer DB. If so, can we add some document here about this?

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+    /// Directory where db_checkpoints will be stored.
+    pub db_checkpoint_dir: Option<PathBuf>,
+    /// Maximum number of db_checkpoints to keep.
+    pub max_db_checkpoints: usize,
+    /// How often to create db_checkpoints (in seconds).
+    pub db_checkpoint_interval_secs: u64,
+    /// Whether to sync files to disk before each db_checkpoint.
+    pub sync: bool,
+    /// Number of background operations for db_checkpoint/restore.
+    pub max_background_operations: i32,
```
halfprice, line 39: Nit: not used?

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+    const CHECKPOINT_CREATION_RETRY_DELAY: time::Duration = time::Duration::from_secs(300);
+    /// Default timeout for tasks.
+    const DEFAULT_TASK_TIMEOUT: StdDuration = time::Duration::from_secs(60 * 60);
+
+    /// Create a new db_checkpoint manager for RocksDB.
+    pub async fn new(
+        db: Arc<RocksDB>,
+        config: DbCheckpointConfig,
+    ) -> Result<Self, DbCheckpointError> {
+        if let Some(db_checkpoint_dir) = config.db_checkpoint_dir.as_ref() {
```
halfprice, line 246: Question: do we still want to create checkpoint manager if db_checkpoint_dir is not provided?

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+
+/// Configuration for RocksDB db_checkpoint management.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct DbCheckpointConfig {
+    /// Directory where db_checkpoints will be stored.
+    pub db_checkpoint_dir: Option<PathBuf>,
+    /// Maximum number of db_checkpoints to keep.
+    pub max_db_checkpoints: usize,
+    /// How often to create db_checkpoints (in seconds).
+    pub db_checkpoint_interval_secs: u64,
```
halfprice: Nit: internal configure, `Duration` type is prefered.

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+        engine
+            .create_new_backup_flush(db_ref, flush_before_backup)
+            .map_err(|e| DbCheckpointError::Other(anyhow::anyhow!("Backup error: {}", e)))?;
+
+        tracing::info!("rocksDB backup created successfully");
+        Ok(())
+    }
+
+    /// Cancel db_checkpoint manager.
+    pub fn cancel(&self) {
```
halfprice: Nit: not used? (Also looks redundant with `shutdown`)

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+                }
+
+                Some(cmd) = command_rx.recv() => {
+                    match cmd {
+                        DbCheckpointRequest::CreateDbCheckpoint {
+                            response,
+                            db_checkpoint_dir,
+                            delay,
+                        } => {
+                            if current_task.as_ref().is_some_and(|task| {
```
halfprice, line 381: Question: after a checkpoint is created, where do we clear `current_task`?

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+        if max_db_checkpoints == 0 {
+            return;
+        }
+
+        let Ok(mut engine) = Self::create_backup_engine(db_checkpoint_dir) else {
+            tracing::error!("Failed to create backup engine");
+            return;
+        };
+        let result = engine
+            .purge_old_backups(max_db_checkpoints)
```
halfprice: Important: for all these non-async RocksDb operation, we should execute them on blocking thread.

halfprice (commented):
Thanks a lot @liquid-helium and this looks amazing!

I primarily looked at the execution logic, and I only have some nits with one important comment that we should put all RocksDb operation on a blocking thread. Other than that, LGTM!

I don't really know much detail about RocksDb backup creation so I'll rely on @sadhansood to review that part.

crates/walrus-service/bin/node.rs
sadhansood: Alternatively, consider renaming the local commands to something like `walrus-node local-admin` or `walrus-node maintenance` to create a clearer distinction in the CLI interface itself.

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+            Ok(_) => tracing::info!(
+                "purged old db_checkpoints, keeping {} most recent",
+                max_db_checkpoints
+            ),
+            Err(e) => tracing::error!(?e, "Failed to purge old db_checkpoints"),
+        }
+    }
+
+    /// Restore from the most recent backup.
+    pub async fn restore_latest(
```
sadhansood, line 604: Could you add some comments or documentation clarifying that the `restore_latest` function is intended for **offline** use (i.e., the node must be stopped when restoring the database)? This is standard for RocksDB backups but worth making explicit. I only see it currently being used in a test.

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+            db_path = ?db.db_path
+        ),
+        err
+    )]
+    fn create_backup_impl(
+        db: &Arc<RocksDB>,
+        db_checkpoint_dir: &Path,
+        flush_before_backup: bool,
+    ) -> Result<(), DbCheckpointError> {
+        let mut engine = Self::create_backup_engine(db_checkpoint_dir)?;
```
sadhansood: Maybe consider adding metrics for checkpoint creation status (success/failure), duration, and perhaps the size of the created checkpoints? This would be very helpful for operational monitoring.

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+
+        tracing::info!("start creating RocksDB backup");
+
+        let db_ref = &db.underlying;
+        engine
+            .create_new_backup_flush(db_ref, flush_before_backup)
+            .map_err(|e| DbCheckpointError::Other(anyhow::anyhow!("Backup error: {}", e)))?;
+
```
sadhansood, line 648: The backup process likely involves reading the entire RocksDB database files and writing them to the checkpoint directory if the directory is an external disk. If the active RocksDB data directory and the db_checkpoint_dir are on the same physical storage device (disk), this will create severe I/O contention and disk usage over time (if opertor doesn;t move the backup out). negatively impacting the performance of the running Walrus node, potentially slowing down blob reads/writes significantly. I htink the documentation and configuration examples should strongly emphasize that the db_checkpoint_dir must be configured to point to a directory located on a physically separate disk or storage volume from the main RocksDB data directory. The application code itself cannot reliably enforce this, so it falls on the operator to configure the mount points correctly.

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+    }
+
+    /// Restore from the most recent backup.
+    pub async fn restore_latest(
+        db_checkpoint_dir: &Path,
+        db_path: &Path,
+        wal_dir: Option<&Path>,
+    ) -> Result<(), DbCheckpointError> {
+        // Create a fresh BackupEngine for this operation.
+        let mut engine = Self::create_backup_engine(db_checkpoint_dir)?;
```
sadhansood: Does RocksDB's BackupEngine have internal options related to throttling I/O or controlling background operations that could potentially be leveraged to limit the performance impact on the foreground operations? Though this comes at the cost of increasing backup duration. The current implementation uses BackupEngineOptions::new and BackupEngine::open, which likely use defaults unless specific options are set.

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+            return;
+        }
+
+        let Ok(mut engine) = Self::create_backup_engine(db_checkpoint_dir) else {
+            tracing::error!("Failed to create backup engine");
+            return;
+        };
+        let result = engine
+            .purge_old_backups(max_db_checkpoints)
+            .map_err(|e| DbCheckpointError::Other(anyhow::anyhow!("Purge error: {}", e)));
```
sadhansood, line 592: While purge_old_backups helps, if the backup disk runs out of space during a backup, the backup will fail. If this happens consistently or monitoring isn't in place, we might find yourself without recent valid backups. Should we add documentation or add monitoring for disk space usage of the backup disk to improve the reliability of the backup process?

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,814 @@
...
+                max_db_checkpoints
+            ),
+            Err(e) => tracing::error!(?e, "Failed to purge old db_checkpoints"),
+        }
+    }
+
+    /// Restore from the most recent backup.
+    pub async fn restore_latest(
+        db_checkpoint_dir: &Path,
+        db_path: &Path,
```
sadhansood, line 606: Is there a plan about testing restore procedures using actual backup files to ensure they are valid and can be used to recover data. Something like a ci job?

crates/walrus-service/src/node/db_checkpoint.rs
sadhansood: It might be possible to check if backup and main disk are the same using `stat` on linux and mac, however I am not completely sure about this.

sadhansood (commented):
Thanks @liquid-helium, this is a solid PR for backing up walrus data!

crates/walrus-service/bin/node.rs
liquid-helium: Thanks for all the suggestions, renamed to `LocalAdmin`.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: The timer task is a non-blocking tokio task, and the actual backup task is created as a blocking one on line 139:
`let worker_task = tokio::task::spawn_blocking(task_fn);`

crates/walrus-service/bin/node.rs
liquid-helium: good point, change applied!

crates/walrus-service/bin/node.rs
liquid-helium: comments added.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Removed.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Yes, it will be there for the manual checkpoint creation.

crates/walrus-service/src/node/config.rs
liquid-helium: example added.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Done.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Nice catch, removed.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: We don't clear it, we keep it as a record for the last task, that users can query.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Yes, these are executed in a blocking thread.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Yes, you are correct. I will add the instructions in the following LocalAdmin::Restore commands. The current restore is only for test.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: That's a very good point.
I added note on the doc string, and will include it in the future instructions.

I didn't find any existing platform independent library to check if the checkpoint dir is on the same disk as the main db. Added WAL-843 as a todo.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: The native RocksDB has this option to limit the io rate limit, as well as the c bindings, but not in the Rust RocksDB wrapper. Created WAL-844 to track this.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Sure, I am also thinking about it.
Let me add a TODO here, since the current PR is already large.
Created WAL-845 for this.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: That's a good point.
When there is not enough disk space, the creation will fail, and we log a Error locally.
In the future, with metrics, we can monitor how long a node hasn't created a backup or checkpoint.
Also open to other approaches.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Yes, a `restore` command/functionality will be added in a following pr, and tests will be added.

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -0,0 +1,823 @@
...
+use crate::node::errors::DbCheckpointError;
+
+/// Configuration for RocksDB db_checkpoint management.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct DbCheckpointConfig {
+    /// Directory where db_checkpoints will be stored.
+    /// Note: It is strongly recommended to locate this directory on a separate physical disk
+    /// from the main database to avoid potential I/O performance degradation.
+    // TODO(WAL-843): Add a check to ensure the directory is on a separate physical disk.
+    pub db_checkpoint_dir: Option<PathBuf>,
```
sadhansood, line 31: We should perhaps make it `#[serde(default, skip_serializing_if = "Option::is_none")]`

crates/walrus-service/src/node/db_checkpoint.rs
halfprice: Ah, I missed that. LGTM

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Change applied. Thanks for the tip.

PR #2027, opened 2025-04-25
refactor: rename some `Client` types to avoid confusion

wbbradley:
## Description

Rename `walrus_rest_client::client::Client` to `walrus_rest_client::WalrusRestClient`

## Test plan

CI Pipeline.


jpcsmith:
Hey @wbbradley, what's the rational for this change?

wbbradley:
> Hey @wbbradley, what's the rational for this change?

The primary rationale for this change is to avoid confusion when looking at a piece of code containing a thing called `Client`. Naming something `Client` causes the reader to rely on contextual clues to understand what's going on, and to get a sense of bearing. By giving these types more specific and unique names we can facilitate faster contextual awareness, as well as allowing developers to leverage tools such as `grep` and `ctags` to manage actions such as finding references or definitions. 

If Rust compilation times were instantaneous I'd be less inclined to make a change like this, but still would prefer it. Given the fact that it can take many seconds to get an editor environment loaded with `rust_analyzer` fully cached, _and_ this sort of operation is not possible from the command-line - barring more complex tools I'm not aware of - I would like to make the code easier to read and easier to manipulate using traditional methods which have stood the test of time. (`grep`, etc...)

Also, in the context of a downstream user application that will eventually use the Walrus SDK, I believe it is more polite to not pollute their namespace with `Client`. Yes, they can `use foo::bar::Client as FooBarClient` but that (1) adds an unnecessary decision to the programmers day (_should I disambiguate this token by giving it a more legible name? and what name should I give it?_) and (2) makes it less likely that folks reading code that leverages Walrus will be seeing the same token names across various codebases, again limiting the efficacy of the naming.

Further, in my experience namespaces are a great invention for programming languages, but they are not a panacea and I assert that most developers would prefer to not have to ingrain tacit knowledge for all of the names of entities within the systems that they utilize in order to expedite disambiguation of tokens while reading code - an activity that we spend most of our time doing.

Also, Rust has strong type inference capabilities, and modern editors have good autocomplete functionality, so this should not exacerbate anyone's carpal tunnel syndrome.

Also, I'll assert that longer more unique names for tokens result in better search results across all search providers. I don't have hard evidence for that, but it has borne out in my experience, and it just makes sense as a way of avoiding collisions.

Finally, if opinions on the team strongly lean away from this direction, then we can close this PR. No need for me to die on this hill. :-)

giac-mysten:
@jpcsmith what would the `AggregatorClient` be? and @wbbradley any plan to distinguish the `StorageNodeClient` in the rest api from the `Client` in the `walrus-sdk`?

wbbradley:
> I would suggest, perhaps, `StorageNodeClient` and `AggregatorClient`(?) as they provide a bit more information about which APIs are being connected to.

Thanks @jpcsmith, yeah, I'm down to change the actual name choices here.

Regarding changing `WalrusRestClient ` to `StorageNodeClient`, I think that makes sense. After sleeping on it some more, I'm not sure it makes sense to rename the `walrus_sdk::client::Client` since it is so all encompassing when it comes to low-level system functionality. And, it's not only used by the Aggregator, so I feel like the term AggregatorClient might be slightly misleading? I'll push an update of the approach I've just described shortly.

crates/walrus-service/src/node/server.rs
```diff
@@ -733,11 +737,13 @@ mod tests {
         (config, handle)
     }
 
-    fn default_client_builder() -> ClientBuilder {
-        Client::builder().no_proxy().tls_built_in_root_certs(false)
+    fn default_client_builder() -> StorageNodeClientBuilder {
```
halfprice: `default_storage_node_client_builder` for consistency?

halfprice (commented):
Thanks @wbbradley for the PR. Looks like there are quite a few CI failures, can you take a look? Also, the the amount of changes in Cargo.lock seems too large with a simple renaming PR, is this expected?

wbbradley:
> Thanks @wbbradley for the PR. Looks like there are quite a few CI failures, can you take a look? Also, the the amount of changes in Cargo.lock seems too large with a simple renaming PR, is this expected?

Hey @halfprice, thanks, I'll take a look. Sorry for the distractiong.

crates/walrus-service/src/node/server.rs
wbbradley: Good call.

mlegner (approved):
Thanks for the clarification work!

PR #2028, opened 2025-04-27
fix: e2e test race condition

liquid-helium:
## Description

Some of the e2e tests started failing, the symptom is the blob event is slow, e.g., after storing a blob successfully, read could fail because the certify event is not processed by majority of storage nodes.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -83,6 +83,7 @@ use walrus_sui::{
 use walrus_test_utils::{Result as TestResult, WithTempDir, assert_unordered_eq, async_param_test};
 use walrus_utils::backoff::ExponentialBackoffConfig;
 
+/// Read a blob with retries if the blob is registered but not certified.
 async fn read_blob_with_wait_for_certification(
     client: &Client<SuiContractClient>,
     blob_id: &BlobId,


Cargo.lock
```diff
@@ -2341,9 +2341,9 @@ dependencies = [
 
 [[package]]
 name = "crossbeam-channel"
-version = "0.5.15"
+version = "0.5.14"
```
giac-mysten: Remark: there is a version downgrade committed here. is this intentional?

crates/walrus-sdk/client_config_example.yaml
```diff
@@ -35,6 +35,7 @@ communication_config:
     max_backoff_millis: 5000
     max_retries: 5
   sui_client_request_timeout_millis: null
+  read_blob_certification_wait_time_secs: 30
```
giac-mysten: Important: in this example config the wait time is 30 seconds. However, the non-default config is `0` seconds. 
This is because the config is produced by a test case, but here it should really be `0` (otherwise, copying this config will lead to suboptimal wait times)

crates/walrus-sdk/src/client.rs
```diff
@@ -259,16 +259,39 @@ impl<T: ReadClient> Client<T> {
...
                         .await?
                 }
             };
-            blob_status
-                .initial_certified_epoch()
-                .ok_or_else(|| ClientError::from(ClientErrorKind::BlobIdDoesNotExist))?
+
+            while status.is_registered()
+                && status.initial_certified_epoch().is_none()
+                && start_time.elapsed() < timeout_duration
```
giac-mysten: Question: if a blob is only registered (and not certified, which is possible in practice), we have to wait for the full timeout duration. This is not a problem with the default timeout, which is `0` seconds. However, we do expose the possiblity to configure this timeout (see the problem above). 

Suggestion: I seems to me this code should only ever be used for tests. If this is correct, can we feature-gate it instead of using a `0` value in the config to disable it?

giac-mysten (commented):
Thanks @liquid-helium ! I'm wondering whether this "retry" feature should be feature flagged instead of just disabled through the config. Wdyt?

crates/walrus-sdk/client_config_example.yaml
mlegner: Ah this is an unfortunate side-effect of the way we generate and update this example configuration. I'm not sure we need to make this configurable for the release build. If we do, I'd suggest to set the default value to `0` and then use a different value in the `default_for_test` function.

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -270,4 +275,17 @@ pub(crate) mod default {
...
+        {
+            #[cfg(test)]
+            {
+                30
+            }
+            #[cfg(not(test))]
+            {
+                0
+            }
+        }
```
mlegner: Important: I'm not sure this behaves as you expect it to behave: The `cfg(test)` only affects unit tests of the current crate, not tests in other crates or E2E tests AFAIU. A better solution would be to set the default value to `0` and then use a different value in the `default_for_test` function as mentioned above.

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -62,6 +62,9 @@ pub struct ClientCommunicationConfig {
     #[serde(rename = "sui_client_request_timeout_millis")]
     #[serde_as(as = "Option<DurationMilliSeconds>")]
     pub sui_client_request_timeout: Option<Duration>,
+    /// The maximum time to wait for a blob to be certified.
+    #[serde(rename = "read_blob_certification_wait_time_secs")]
+    pub read_blob_certification_wait_time: u64,
```
mlegner: Suggestion: Use a `Duration` here similar to above with a `#[serde_as(as = "DurationSeconds")]` annotation.

Cargo.lock
liquid-helium: No, it is automatically generated, reverted.

crates/walrus-sdk/client_config_example.yaml
liquid-helium: Removed the config, using `test-utils` to enable it in test.

crates/walrus-sdk/src/config/communication_config.rs
liquid-helium: thanks for the tip! Removed this part, switched to feature flag.

crates/walrus-sdk/src/config/communication_config.rs
liquid-helium: ditto

crates/walrus-sdk/src/client.rs
liquid-helium: Nice idea, switched to test-only with the feature flag of  `test-utils`.

mlegner (approved):
Thanks @liquid-helium!

PR #2032, opened 2025-04-28
refactor: limit concurrent requests sent to remote nodes

jpcsmith:
## Description

Limits the number of concurrent requests that are sent to each individual storage node. In the event that the limit has been reached for an individual storage node, the node is skipped when collecting nodes to send requests to. When coupled with the exponential backoff that is already implemented, requests that fail to find sufficient ready nodes will enter exponential backoff.

## Test plan

Existing tests. The performance impact has not been measured, but will need to be observed on PTN.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/committee/node_service.rs
+++ b/crates/walrus-service/src/node/committee/node_service.rs
@@ -24,7 +24,7 @@ use std::{
 };
 
 use futures::{FutureExt, future::BoxFuture};
-use tower::Service;
+use tower::{Service, limit::ConcurrencyLimit};
 use walrus_core::{
     BlobId,
     Epoch,
@@ -50,6 +50,7 @@ use walrus_sui::types::StorageNode as SuiStorageNode;
 use walrus_utils::metrics::Registry;
 
 use super::{DefaultRecoverySymbol, NodeServiceFactory};
+use crate::node::config::defaults::REST_HTTP2_MAX_CONCURRENT_STREAMS;
 
 /// Requests used with a [`NodeService`].
 #[derive(Debug, Clone)]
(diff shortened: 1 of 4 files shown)


halfprice (commented):
Thanks @jpcsmith for the PR! I understand the motivation and I think this will be very helpful in terms of reducing tail latency for shard recovery. One high level comment I have is whether we can add some logs/metrics to differentiate backoff caused by client not ready vs. RPC failures. With the PR, I feel that under high load, it maybe hard to reason about what causes the high latency, so adding some signal to indicate different reasons for backoff, or the number of PRC failures caused by client concurrency limit is reached, maybe helpful.

jpcsmith:
> Thanks @jpcsmith for the PR! I understand the motivation and I think this will be very helpful in terms of reducing tail latency for shard recovery. One high level comment I have is whether we can add some logs/metrics to differentiate backoff caused by client not ready vs. RPC failures. With the PR, I feel that under high load, it maybe hard to reason about what causes the high latency, so adding some signal to indicate different reasons for backoff, or the number of PRC failures caused by client concurrency limit is reached, maybe helpful.

Any single backoff will likely be caused by a combination of both unreadiness and error responses, unless the the system is fully saturated. We could add a counter for every-time a skip is performed which would give some insight into the frequency of skips. At the moment, we should be able to infer this from the client metrics (number of active requests).

halfprice:
> > Thanks @jpcsmith for the PR! I understand the motivation and I think this will be very helpful in terms of reducing tail latency for shard recovery. One high level comment I have is whether we can add some logs/metrics to differentiate backoff caused by client not ready vs. RPC failures. With the PR, I feel that under high load, it maybe hard to reason about what causes the high latency, so adding some signal to indicate different reasons for backoff, or the number of PRC failures caused by client concurrency limit is reached, maybe helpful.
> 
> Any single backoff will likely be caused by a combination of both unreadiness and error responses, unless the the system is fully saturated (such as setting the max_concurrent_* values very high). We could add a counter for every-time a skip is performed which would give some insight into the frequency of skips. At the moment, we should be able to infer this from the client metrics (number of active requests).
> 
> Let me know if the counter would be sufficient.

Sorry for the late reply @jpcsmith ! For some reason, since my vacation, I stopped receiving a lot of notifications... For this PR, if you have a way to distinguish between them, then let's go with the current code, and see how it goes.

halfprice:
@jpcsmith looks like you are out for next 2 weeks. Will try to merge this PR for you.

halfprice (approved):

PR #2036, opened 2025-04-29
fix: fix deadlock when creating checkpoint downloader when lag is high

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/checkpoint-downloader/src/downloader.rs
+++ b/crates/checkpoint-downloader/src/downloader.rs
@@ -172,6 +172,19 @@ impl ParallelCheckpointDownloaderInner {
         let Ok(Some(current_checkpoint)) = checkpoint_store.get(&()) else {
             return Err(anyhow!("Failed to get current checkpoint"));
         };
+
+        #[cfg(msim)]
+        {
+            let mut checkpoint_lag_error = false;
+            sui_macros::fail_point_if!("fail_point_current_checkpoint_lag_error", || {
+                checkpoint_lag_error = true;
+            });
+
+            if checkpoint_lag_error {
+                return Err(anyhow!("fail point triggered checkpoint lag error"));
+            }
+        }
+
         let latest_checkpoint = client.get_latest_checkpoint_summary().await?;
         let current_lag =
             latest_checkpoint.sequence_number - current_checkpoint.inner().sequence_number;
@@ -294,9 +307,11 @@ impl ParallelCheckpointDownloaderInner {
                                 "checkpoint lag monitoring has failed repeatedly \
                                 - adjusting to average workers"
                             );
+                            let current_workers = *worker_count.read().await;
+                            tracing::info!("adjusting to average workers {:?}", current_workers);
                             Self::adjust_workers(
                                 &mut next_worker_id,
-                                *worker_count.read().await,
+                                current_workers,
                                 average_workers,
                                 &channels,
                                 &config,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1856,6 +1856,7 @@ dependencies = [
  "rocksdb",
  "serde",
  "serde_with",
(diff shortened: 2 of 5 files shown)


crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -294,9 +307,11 @@ impl ParallelCheckpointDownloaderInner {
                                 "checkpoint lag monitoring has failed repeatedly \
                                 - adjusting to average workers"
                             );
+                            let current_workers = *worker_count.read().await;
+                            tracing::info!("adjusting to average workers {:?}", current_workers);
                             Self::adjust_workers(
                                 &mut next_worker_id,
-                                *worker_count.read().await,
```
halfprice, line 299: @sadhansood here is the deadlock. The read lock is held inside the adjust_workers, where a write lock is also taken inside the function call.

sadhansood: @halfprice I believe you, but this is weird/unexpected behavior if true. Shouldn't the deference operator read the value and cause the lock guard to go out of scope immediately.

sadhansood (approved):

crates/checkpoint-downloader/src/downloader.rs
sadhansood: @halfprice so is the read lock is held by the caller for the entire duration of the function call?

sadhansood: I think you are correct @halfprice, I reproed the issue with a much simpler test:
```
async fn inner_func(_value_from_read: usize, lock: Arc<RwLock<usize>>) -> Result<(), ()> {
    tokio::time::sleep(Duration::from_millis(10)).await;

    println!("inner_func: Attempting to acquire write lock...");
    let mut writer = lock.write().await;
    println!("inner_func: Write lock acquired successfully.");
    *writer += 1;
    println!("inner_func: Releasing write lock.");
    Ok(())
}

#[tokio::test]
async fn test_read_lock_release_before_await() {
    let data = Arc::new(RwLock::new(10)); // Shared data protected by RwLock

    println!("Main test: Acquiring read lock for argument...");
    let value = *data.read().await;
    println!("Main test: Read lock released after getting value: {}", value);

    println!("Main test: Calling inner_func...");
    let result = timeout(
        Duration::from_secs(1), // Set a timeout to detect potential deadlocks
        inner_func(*data.read().await, data.clone()),
    )
    .await;

    println!("Main test: inner_func call completed.");

    // Assert that the timeout did NOT expire and inner_func returned Ok
    assert!(result.is_ok(), "Test timed out, indicating a likely deadlock (read lock was likely held).");
}
```

sadhansood: Ok, so i think now I know why this deadlock in pool monitor is deadlocking the entire checkpoint download process. So, the task which inserts new checkpoint sequence numbers in the queue also gets blocked trying to acquire the read lock:
```
  /// Starts downloading checkpoints from the given sequence number.
    pub fn start(
        mut self,
        sequence_number: CheckpointSequenceNumber,
    ) -> mpsc::Receiver<CheckpointEntry> {
        let (result_tx, result_rx) = mpsc::channel(self.config.result_queue_size());
        let message_sender = self.message_sender.clone();

        tokio::task::spawn(async move {
            let mut sequence_number = sequence_number;
            let mut in_flight = BTreeSet::new();
            let mut pending = BTreeMap::new();
            let mut next_expected = sequence_number;

            while !self.cancellation_token.is_cancelled() {
                let num_workers = *self.worker_count.read().await;
                let max_in_flight_requests =
                    num_workers * self.config.channel_config.work_queue_buffer_factor;
           .....
                
```
This is the line in start: `  let num_workers = *self.worker_count.read().await;`, so basically this blocks the entire checkpoint download flow.

PR #2037, opened 2025-04-30
feat: Improvements to checkpoint downloader

sadhansood:
## Description

Replace RWLock with Atomic
Add monitored scope
Increase error count threshold to trigger adjusting to avergae worker size

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -226,11 +224,7 @@ impl ParallelCheckpointDownloaderInner {
                         .message_sender
                         .send(WorkerMessage::Shutdown)
                         .await?;
-                    let new_count = {
-                        let mut count = worker_count.write().await;
-                        *count -= 1;
-                        *count
-                    };
+                    let new_count = worker_count.fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
```
wbbradley, line 227: Is there a reason to use Relaxed here? Might as well use SeqCst, no? I don't see a memory fence being a problem here.

wbbradley (approved):
LGTM. Unclear why we'd use Relaxed vs SeqCst for these sites. But, it doesn't seem super important either way.

crates/checkpoint-downloader/src/downloader.rs
sadhansood: I think Relaxed is just more performant (because no thread synchronization is needed) and the correctness isn't impacted from its usage.

wbbradley: Good point, thanks.

PR #2042, opened 2025-05-01
fix(aggregator): fail early for inexistent blobs

giac-mysten:
## Description

Closes WAL-673

Before this PR, the aggregator kept refreshing the committee and retrying to get the blob for a long time (anecdotally, up to ~50s) for non-existent blobs. This was because there is no explicit way to distinguish between the blob not being stored because (i) it does not exist, vs (ii) it exists but we are under expoch change and we are asking the wrong committee.

This PR forces a status check on the blobs being read. This adds a tiny bit of latency in the case the blob exists (i.e., 1 rtt to the nodes to get the status), but allows to fail much earlier if the blob does not (~0.1 s vs tens of seconds).

The PR also reduces the level of warnings related to retries to avoid noise in the common case.

## Test plan

Manual aggregator queries + existing tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -260,11 +260,17 @@ impl<T: ReadClient> Client<T> {
         let committees = self.get_committees().await?;
 
         let get_status_fn = |status: Option<BlobStatus>| async move {
-            if let Some(status) = status {
-                Ok(status)
+            let status = if let Some(status) = status {
+                status
             } else {
                 self.get_blob_status_with_retries(blob_id, &self.sui_client)
-                    .await
+                    .await?
+            };
+
+            if BlobStatus::Nonexistent == status {
+                Err(ClientError::from(ClientErrorKind::BlobIdDoesNotExist))
+            } else {
+                Ok(status)
             }
         };
 
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -257,21 +257,19 @@ impl<T: ReadClient> Client<T> {
         self.check_blob_id(blob_id)?;
         let committees = self.get_committees().await?;
 
-        let certified_epoch = if committees.is_change_in_progress() {
-            tracing::info!("epoch change in progress, reading from initial certified epoch");
-            let blob_status = match blob_status {
-                Some(status) => status,
-                None => {
-                    self.get_blob_status_with_retries(blob_id, &self.sui_client)
-                        .await?
-                }
-            };
-            blob_status
-                .initial_certified_epoch()
-                .ok_or_else(|| ClientError::from(ClientErrorKind::BlobIdDoesNotExist))?
-        } else {
-            // We are not during epoch change, we can read from the current epoch directly.
-            committees.epoch()
+        let status = match blob_status {
+            Some(status) => status,
+            None => {
+                // The status check will return `ClientErrorKind::BlobIdDoesNotExist` if the blob
+                // is not registered or expired.
+                self.get_blob_status_with_retries(blob_id, &self.sui_client)
+                    .await?
+            }
+        };
+
+        let Some(certified_epoch) = status.initial_certified_epoch() else {
+            // The blob ID has not been certified yet. No need to read it.
+            return Err(ClientError::from(ClientErrorKind::BlobIdDoesNotExist));
         };
 
         // Return early if the committee is behind.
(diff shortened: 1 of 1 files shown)


wbbradley (approved):
LGTM

crates/walrus-sdk/src/client.rs
```diff
@@ -257,21 +257,19 @@ impl<T: ReadClient> Client<T> {
...
-                        .await?
-                }
-            };
-            blob_status
-                .initial_certified_epoch()
-                .ok_or_else(|| ClientError::from(ClientErrorKind::BlobIdDoesNotExist))?
-        } else {
-            // We are not during epoch change, we can read from the current epoch directly.
-            committees.epoch()
+        let status = match blob_status {
```
liquid-helium: I am a bit concerned about the latency here, the previous behavior when there is no epoch change is: we read directly with the latest epoch, and if the blob is certified, we read back the slivers with 1 rt;
It seems that now we always read the status first, and for a certified blobs, there is an additional rt compared to the existing code.
Are there other considerations here that I missed?

mlegner: Could we maybe do the status check in parallel with an attempt to read and cancel the read if the status is `Nonexistent`? This could also be added as a follow-up though.

giac-mysten: > Could we maybe do the status check in parallel with an attempt to read and cancel the read if the status is Nonexistent? This could also be added as a follow-up though.

This is a good option, i'll stack this on top of this PR.

The other option I was thinking about is to interpret a quorum of "not found" metadata as the sign that the blob is not certified. This works similarly to the status check, because in the status check we also take a quorum of "status not found" replies as evidence that the blob does not exist.

However, this has the downside that if the blob does exist, the aggregator will be requesting much more data than necessary.

giac-mysten: hey @liquid-helium , this should be fixed now. Could you have another look? (sorry for the delay)

crates/walrus-sdk/src/client.rs
```diff
@@ -254,27 +254,44 @@ impl<T: ReadClient> Client<T> {
...
-                .ok_or_else(|| ClientError::from(ClientErrorKind::BlobIdDoesNotExist))?
+
+            if BlobStatus::Nonexistent == status {
+                Err(ClientError::from(ClientErrorKind::BlobIdDoesNotExist))
+            } else {
+                Ok(status)
+            }
+        };
+
+        let (certified_epoch, blob_status) = if committees.is_change_in_progress() {
```
mlegner: Nit: Can we rename the `certified_epoch`? In the `unwrap_or_else` of the `else` branch, this is set to some "dummy" value, so this isn't actually always the `certified_epoch`.

crates/walrus-sdk/src/client.rs
```diff
@@ -283,8 +300,15 @@ impl<T: ReadClient> Client<T> {
             }));
         }
 
-        self.read_metadata_and_slivers::<U>(certified_epoch, blob_id)
-            .await
+        // Execute the status request and the metadata/sliver request concurrently.
+        // If the status request fails, the metadata/sliver request will be cancelled. If the
+        // status request succeeds, the metadata/sliver request will be executed to completion. If
+        // we already have a blob status, the `get_status_fn` immediately returns Ok.
```
mlegner: Question: Is there a chance we wait for the status request after we already have the slivers?

mlegner (approved):
Really nice! 💯

crates/walrus-sdk/src/client.rs
giac-mysten: in principle yes, though i don't know how this could happen in practice: downloading all the data would need to be faster than the status checks.

Any suggestion on how to add a short-circuit here?

mlegner: Hmm... Not sure if there is a simple clean way of doing this. But I agree that it is most likely not too relevant in practice anyway, so maybe we can just keep it as it is.

giac-mysten: Ah found it, let me try something :)

giac-mysten: I think this should work :) ab752ad325423ef36591a8486d4e389cec56f466

mlegner: Very nice, I love it! 🥇

PR #2043, opened 2025-05-01
refactor: remove ClientErrorKind & minor simplification to a Client creation flows

wbbradley:
## Description

This PR describes a potential interface for the Walrus SDK.

## Test plan

CI Pipeline.

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -37,15 +37,12 @@ use walrus_core::{
 use walrus_proc_macros::walrus_simtest;
 use walrus_sdk::{
     client::{Blocklist, Client, WalrusStoreBlob, WalrusStoreBlobApi, responses::BlobStoreResult},
-    error::{
-        ClientError,
-        ClientErrorKind::{
-            self,
-            NoMetadataReceived,
-            NoValidStatusReceived,
-            NotEnoughConfirmations,
-            NotEnoughSlivers,
-        },
+    error::ClientError::{
+        self,
+        NoMetadataReceived,
+        NoValidStatusReceived,
+        NotEnoughConfirmations,
+        NotEnoughSlivers,
     },
     store_when::StoreWhen,
 };
@@ -191,7 +188,7 @@ async_param_test! {
 async fn test_store_and_read_blob_with_crash_failures(
     failed_shards_write: &[usize],
     failed_shards_read: &[usize],
-    expected_errors: &[ClientErrorKind],
+    expected_errors: &[ClientError],
 ) {
     telemetry_subscribers::init_for_testing();
     let result =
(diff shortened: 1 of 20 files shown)


crates/walrus-sdk/src/walrus.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// The blob ID.
+    BlobId(BlobId),
+    /// The blob's object ID.
+    ObjectId(ObjectID),
+}
+
+/// The main entrypoint for the Walrus SDK.
+pub struct Walrus {
+    /// The Sui wallet for the client.
+    _wallet_context: Arc<WalletContext>,
```
giac-mysten: Question: just curious, why the leading `_`? python syntax?

crates/walrus-sdk/src/walrus.rs
```diff
@@ -0,0 +1,219 @@
...
+#[derive(Debug)]
+pub enum BlobSpecifier {
+    /// The blob ID.
+    BlobId(BlobId),
+    /// The blob's object ID.
+    ObjectId(ObjectID),
+}
+
+/// The main entrypoint for the Walrus SDK.
+pub struct Walrus {
```
giac-mysten: Question: should this own a `client::Client` directly, instead of the wallet context?

crates/walrus-sdk/src/walrus.rs
```diff
@@ -0,0 +1,219 @@
...
+        _persistence: BlobPersistence,
+        _post_store: PostStoreAction,
+        _encoding_type: Option<EncodingType>,
+    ) -> Result<Vec<BlobId>> {
+        todo!()
+    }
+
+    /// Compute the details related to storing several blobs. This method does not actually store
+    /// the blobs, but returns the details of what would happen if they were stored.
+    pub async fn store_blobs_dry_run(
```
giac-mysten: Question: just thinking here -- i like the idea of separating the dry run command from the others.

would it make sense to go one step further make command building modular?
```rust
walrus.execute(
   WalrusCommand::store(files)
      .for_epochs(epochs)
      .persistence(..)
      .dry_run()
);
```

giac-mysten (commented):
thanks for starting on this @wbbradley ! a few thoughts

crates/walrus-sdk/src/walrus.rs
wbbradley: Because right now it's not compiling. Just did this to avoid unused argument errors. It's temporary.

crates/walrus-sdk/src/walrus.rs
wbbradley: A side quest I have in my back log is to remove network access from `WalletContext`, that is, to not use `WalletContext` directly for network access from Walrus but to prefer wrapping it in a nerfed type that does not allow network access - to only use it for signing and config. But yeah, for today, the inside of the `Walrus` object is not all thought through, since I haven't tried to pull the necessary bits in to the individual method implementations. Once we get going on implementing a chosen interface, I expect that the Walrus object will become more robust. 

And yeah, I think you're right that we'll end up wanting a client::Client, or some other variant.

crates/walrus-sdk/src/walrus.rs
wbbradley: That's a cool idea. I'm not certain it's necessary since the other commands don't quite compose this way, but let's play with that. I feel like that's a facade we can put on later if necessary, too. We'll have to see how well that plays with async/await, as well. Maybe it'll be nicer? Let's play with it.

PR #2046, opened 2025-05-02
feat: Enable recovery from local node forking of event blobs

sadhansood:
## Description

This PR does:
1. Read latest certified event blob upon startup
2. Updates local state with latest certified if local state is behind
3. Delete previous local state

## Test plan

Added a new sim test to test

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/common/event_blob_downloader.rs
+++ b/crates/walrus-service/src/common/event_blob_downloader.rs
@@ -6,15 +6,56 @@
 use std::path::Path;
 
 use anyhow::Result;
-use walrus_core::BlobId;
+use walrus_core::{BlobId, Epoch};
 use walrus_sdk::{client::Client as WalrusClient, error::ClientErrorKind};
 use walrus_storage_node_client::api::BlobStatus;
-use walrus_sui::client::{ReadClient, SuiReadClient};
+use walrus_sui::{
+    client::{ReadClient, SuiReadClient},
+    types::move_structs::EventBlob,
+};
 
 use crate::node::events::{
+    EventStreamCursor,
     event_blob::EventBlob as LocalEventBlob,
     event_processor::EventProcessorMetrics,
 };
+
+/// A struct that contains the metadata of an event blob.
+#[derive(Debug, Clone)]
+pub struct EventBlobWithMetadata {
+    /// The ID of the event blob.
+    pub blob_id: BlobId,
+    /// The cursor of the event stream.
+    pub event_stream_cursor: EventStreamCursor,
+    /// The epoch of the event blob.
+    pub epoch: Epoch,
+    /// The ending checkpoint sequence number of the event blob.
+    pub ending_checkpoint_sequence_number: u64,
+}
+
+impl EventBlobWithMetadata {
+    /// Returns the ending checkpoint sequence number of the event blob.
+    pub fn end(&self) -> u64 {
+        self.event_stream_cursor.element_index
+    }
+
+    /// Returns a reference to the event blob.
+    pub fn blob(&self) -> EventBlob {
+        EventBlob {
+            blob_id: self.blob_id,
+            ending_checkpoint_sequence_number: self.ending_checkpoint_sequence_number,
+        }
+    }
+}
+
+/// A struct that contains the metadata of an event blob.
+#[derive(Debug, Clone)]
+pub enum LastCertifiedEventBlob {
+    /// A struct that contains the metadata of an event blob.
+    EventBlobWithMetadata(EventBlobWithMetadata),
+    /// A reference to an event blob.
+    EventBlob(EventBlob),
+}
 
 /// Responsible for downloading and managing event blobs
 #[derive(Debug)]
@@ -29,6 +70,41 @@ impl EventBlobDownloader {
         Self {
             walrus_client,
             sui_read_client,
+        }
+    }
+
+    /// Returns the metadata of the last certified event blob.
+    pub async fn get_latest_certified_event_blob(&self) -> Result<Option<EventBlobWithMetadata>> {
+        let blob = self.sui_read_client.last_certified_event_blob().await?;
+        match blob {
+            Some(blob) => {
+                let blob_id = blob.blob_id;
+                let result = self
+                    .walrus_client
+                    .read_blob::<walrus_core::encoding::Primary>(&blob_id)
+                    .await?;
+                let blob = LocalEventBlob::new(&result)?;
+                let prev_blob_id = blob.prev_blob_id();
+                if prev_blob_id == BlobId::ZERO {
+                    return Ok(None);
+                }
+                let blob_epoch = blob.epoch();
+                let blob_ending_checkpoint_sequence_number = blob.end_checkpoint_sequence_number();
+                let blob_iter = blob.into_iter();
+                let blob_event = blob_iter
+                    .last()
+                    .ok_or(anyhow::anyhow!("No events in blob"))?;
+                Ok(Some(EventBlobWithMetadata {
+                    blob_id,
+                    event_stream_cursor: EventStreamCursor::new(
+                        blob_event.element.element.event_id(),
+                        blob_event.index + 1,
+                    ),
+                    epoch: blob_epoch,
+                    ending_checkpoint_sequence_number: blob_ending_checkpoint_sequence_number,
+                }))
+            }
+            None => Ok(None),
         }
     }
 
--- a/crates/walrus-service/src/common/event_blob_downloader.rs
+++ b/crates/walrus-service/src/common/event_blob_downloader.rs
@@ -48,7 +48,7 @@ impl EventBlobWithMetadata {
     }
 }
 
-/// A struct that contains the metadata of an event blob.
+/// This is used to store the last certified event blob in the event blob writer.
 #[derive(Debug, Clone)]
 pub enum LastCertifiedEventBlob {
     /// A struct that contains the metadata of an event blob.
@@ -74,38 +74,31 @@ impl EventBlobDownloader {
     }
 
     /// Returns the metadata of the last certified event blob.
-    pub async fn get_latest_certified_event_blob(&self) -> Result<Option<EventBlobWithMetadata>> {
-        let blob = self.sui_read_client.last_certified_event_blob().await?;
-        match blob {
-            Some(blob) => {
-                let blob_id = blob.blob_id;
-                let result = self
-                    .walrus_client
-                    .read_blob::<walrus_core::encoding::Primary>(&blob_id)
-                    .await?;
-                let blob = LocalEventBlob::new(&result)?;
-                let prev_blob_id = blob.prev_blob_id();
-                if prev_blob_id == BlobId::ZERO {
-                    return Ok(None);
-                }
-                let blob_epoch = blob.epoch();
-                let blob_ending_checkpoint_sequence_number = blob.end_checkpoint_sequence_number();
-                let blob_iter = blob.into_iter();
-                let blob_event = blob_iter
-                    .last()
-                    .ok_or(anyhow::anyhow!("No events in blob"))?;
-                Ok(Some(EventBlobWithMetadata {
-                    blob_id,
-                    event_stream_cursor: EventStreamCursor::new(
-                        blob_event.element.element.event_id(),
-                        blob_event.index + 1,
-                    ),
-                    epoch: blob_epoch,
-                    ending_checkpoint_sequence_number: blob_ending_checkpoint_sequence_number,
-                }))
-            }
-            None => Ok(None),
-        }
+    pub async fn get_last_certified_event_blob(&self) -> Result<Option<EventBlobWithMetadata>> {
+        let Some(blob) = self.sui_read_client.last_certified_event_blob().await? else {
+            return Ok(None);
+        };
+        let blob_id = blob.blob_id;
+        let result = self
+            .walrus_client
+            .read_blob::<walrus_core::encoding::Primary>(&blob_id)
+            .await?;
+        let blob = LocalEventBlob::new(&result)?;
+        let blob_epoch = blob.ending_epoch()?;
+        let blob_ending_checkpoint_sequence_number = blob.end_checkpoint_sequence_number();
+        let blob_iter = blob.into_iter();
+        let blob_event = blob_iter
+            .last()
+            .ok_or(anyhow::anyhow!("No events in blob"))?;
+        Ok(Some(EventBlobWithMetadata {
+            blob_id,
+            event_stream_cursor: EventStreamCursor::new(
+                blob_event.element.element.event_id(),
+                blob_event.index + 1,
+            ),
+            epoch: blob_epoch,
+            ending_checkpoint_sequence_number: blob_ending_checkpoint_sequence_number,
+        }))
     }
 
     /// Collects event blobs starting from the given blob ID and going backwards until reaching
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -158,8 +158,11 @@ use self::{
     system_events::{EventManager, SuiSystemEventProvider},
 };
 use crate::{
-    common::{config::SuiConfig, utils::should_reposition_cursor},
-    utils::ShardDiffCalculator,
+    common::{
+        config::SuiConfig,
+        event_blob_downloader::{EventBlobDownloader, LastCertifiedEventBlob},
+    },
+    utils::{ShardDiffCalculator, should_reposition_cursor},
 };
 
 pub(crate) mod db_checkpoint;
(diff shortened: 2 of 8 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -658,17 +658,54 @@ impl StorageNode {
...
+            .sui
+            .as_ref()
+            .expect("Sui config must be provided")
+            .new_read_client()
+            .await?;
+        let walrus_client = walrus_sdk::client::Client::new_read_client_with_refresher(
+            client_config,
+            sui_read_client.clone(),
+        )
+        .await?;
```
mlegner: Remark/suggestion: It seems that we have this pattern 3 times now, also in `crates/walrus-service/src/backup/service.rs` and `crates/walrus-service/src/common/utils.rs`. How about we define a separate function that takes the necessary inputs and builds a new client?

crates/walrus-service/src/node.rs
```diff
@@ -658,17 +658,54 @@ impl StorageNode {
...
+        let event_blob_downloader = EventBlobDownloader::new(walrus_client, sui_read_client);
         let last_certified_event_blob = contract_service.last_certified_event_blob().await?;
+        let last_certified_event_blob_metadata = event_blob_downloader
+            .get_latest_certified_event_blob()
+            .await
+            .ok()
+            .flatten();
+        let latest_certified_event_blob = last_certified_event_blob_metadata
+            .map(LastCertifiedEventBlob::EventBlobWithMetadata)
+            .or(last_certified_event_blob.map(LastCertifiedEventBlob::EventBlob));
```
mlegner: Remark: These variables are a bit confusing. What is the difference between `latest_certified_event_blob` and `last_certified_event_blob`?

crates/walrus-service/src/common/event_blob_downloader.rs
```diff
@@ -32,6 +73,41 @@ impl EventBlobDownloader {
         }
     }
 
+    /// Returns the metadata of the last certified event blob.
+    pub async fn get_latest_certified_event_blob(&self) -> Result<Option<EventBlobWithMetadata>> {
+        let blob = self.sui_read_client.last_certified_event_blob().await?;
+        match blob {
+            Some(blob) => {
```
mlegner: Suggestion (nit): How about the following to avoid some indentation?
```rust
        let Some(blob) = self.sui_read_client.last_certified_event_blob().await? else {
            return Ok(None);
        };
```

crates/walrus-service/src/common/event_blob_downloader.rs
```diff
@@ -32,6 +73,41 @@ impl EventBlobDownloader {
...
+                let blob_id = blob.blob_id;
+                let result = self
+                    .walrus_client
+                    .read_blob::<walrus_core::encoding::Primary>(&blob_id)
+                    .await?;
+                let blob = LocalEventBlob::new(&result)?;
+                let prev_blob_id = blob.prev_blob_id();
+                if prev_blob_id == BlobId::ZERO {
+                    return Ok(None);
+                }
```
mlegner: Question: What is the purpose of this check/return?

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -422,15 +424,63 @@ impl EventBlobWriterFactory {
             &failed_to_attest,
             &certified,
         )?;
+        let mut wb = pending.batch();
+        if let Some(LastCertifiedEventBlob::EventBlobWithMetadata(latest)) =
+            last_certified_event_blob
```
mlegner: Nit: A match statement with three arms might be more readable in this case.

Also, it's a bit unclear to me on which basis the different branches are distinguished and how this is related to the `LastCertifiedEventBlob` enum.

sadhansood:
Thank you @mlegner for your initial feed back. @halfprice would you be able to review it as well ? It'd be great for me to make changes to both your feedback at the same time. Thanks!

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -422,15 +424,63 @@ impl EventBlobWriterFactory {
...
+                        > current_certified.event_cursor.element_index
+                })
+                .is_some()
+            {
+                let certified_metadata = CertifiedEventBlobMetadata::new(
+                    latest.blob_id,
+                    latest.event_stream_cursor,
+                    latest.epoch,
+                );
+                wb.insert_batch(&certified, std::iter::once(((), certified_metadata)))?;
```
halfprice: question: am I reading this correct that if I haven't received the certified event blob event, I may throw away all my local pending blobs after the latest certified blobs read from chain, and rebuild from there?

Does it worth to save some work by comparing the next pending/attested blobs and see if it matches the on chain certified event blob? If so, we don't need to delete all the pendings, and continue.

crates/walrus-service/src/node.rs
```diff
@@ -658,17 +658,54 @@ impl StorageNode {
...
+            .contract_config
+            .clone();
+        let client_config = crate::client::ClientConfig {
+            contract_config,
+            exchange_objects: vec![],
+            wallet_config: None,
+            communication_config: Default::default(),
+            refresh_config: Default::default(),
+        };
+        let sui_read_client = config
```
halfprice: Suggestion: looks like `latest_certified_event_blob` taken in is an option in `EventBlobWriterFactory::new` and if we don't provide it, we continue with normal event processing. In the topic of not blocking node starting up when no fullnode is available, let's don't return error here, but instead use Non in `latest_certified_event_blob`.

crates/walrus-simtest/tests/simtest_event_blob.rs
```diff
@@ -65,6 +65,96 @@ mod tests {
...
+        loop {
+            let current_blob = get_last_certified_event_blob_must_succeed(client).await;
+
+            if current_blob.blob_id != last_certified_blob.blob_id {
+                tracing::info!("New event blob seen during fork wait: {:?}", current_blob);
+                last_certified_blob = current_blob;
+                tokio::time::sleep(Duration::from_secs(30)).await;
+
+                let node_blob_id = get_last_certified_event_blob_from_node(node).await?;
+                if node_blob_id != last_certified_blob.blob_id {
```
halfprice: Question: this can also just be that the node is slow and hasn't processed certified event blob?

halfprice (commented):
Thanks @sadhansood for this important improvements! And sorry for the delay reviewing it. Overall, the solution LGTM, and I have some high level questions in terms of process.

Besides the comment below, another question I have is whether it's worth adding a fork detection in the event writer to automatically detect fork and print some warning, or event restarts the node?

crates/walrus-service/src/common/event_blob_downloader.rs
sadhansood: Ah it's not really needed, left-over from a previous iteration - cleaned it up

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs
sadhansood: fixed the naming

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: Yes you are right @halfprice in that we throw away all uncertified blobs, I've updated the reasoning in the comments but basically I don't think we can know what the on chain certified event blob after last known certified event blob is, so we need to delete all pending blobs and continue.

crates/walrus-simtest/tests/simtest_event_blob.rs
sadhansood: Yeah indeed this is true, I tried to improve the test condition, ptal. But this is still just a proxy condition i.e. network is making progress but this node is stuck for 30s == node has forked. I agree that the node could be stuck could for other reasons too but that is somewhat unlikely in the test setup.

crates/walrus-service/src/common/event_blob_downloader.rs
sadhansood: done

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: done, and added comments to help. Please lmk if this is still unclear.

crates/walrus-service/src/common/event_blob_downloader.rs
```diff
@@ -6,16 +6,57 @@
...
+    /// Returns a reference to the event blob.
+    pub fn blob(&self) -> EventBlob {
+        EventBlob {
+            blob_id: self.blob_id,
+            ending_checkpoint_sequence_number: self.ending_checkpoint_sequence_number,
+        }
+    }
+}
+
+/// A struct that contains the metadata of an event blob.
```
mlegner: Nit: The docstring doesn't match the enum.

crates/walrus-service/src/node.rs
```diff
@@ -664,18 +667,45 @@ impl StorageNode {
...
+                    .flatten()
+                    .map(LastCertifiedEventBlob::EventBlob),
+            },
+            None => contract_service
+                .last_certified_event_blob()
+                .await
+                .ok()
+                .flatten()
+                .map(LastCertifiedEventBlob::EventBlob),
+        };
```
mlegner: Nit: Not sure if it's much better, but the following version avoids some code duplication and futures handling:
```suggestion
        let mut last_certified_event_blob = None;
        if let Some(downloader) = event_blob_downloader {
            last_certified_event_blob = downloader
                .get_last_certified_event_blob()
                .await
                .ok()
                .flatten()
                .map(LastCertifiedEventBlob::EventBlobWithMetadata);
        }
        if last_certified_event_blob.is_none() {
            last_certified_event_blob = contract_service
                .last_certified_event_blob()
                .await
                .ok()
                .flatten()
                .map(LastCertifiedEventBlob::EventBlob);
        }

```

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -422,15 +424,89 @@ impl EventBlobWriterFactory {
             &failed_to_attest,
             &certified,
         )?;
+        let mut wb = pending.batch();
+        match last_certified_event_blob {
```
mlegner, line 461: Nit: As cases 1 and 2 are pretty long, they could be moved to separate functions to increase readability (and decrease the indentation levels).

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -422,15 +424,89 @@ impl EventBlobWriterFactory {
...
+                    latest
+                );
+                if certified
+                    .get(&())?
+                    .filter(|current_certified| {
+                        latest.event_stream_cursor.element_index
+                            > current_certified.event_cursor.element_index
+                    })
+                    .is_some()
+                {
```
mlegner: Nit: The following variant feels more intuitive to me:
```suggestion
                if certified.get(&())?.is_some_and(|current_certified| {
                    latest.event_stream_cursor.element_index
                        > current_certified.event_cursor.element_index
                }) {
```

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -422,15 +424,89 @@ impl EventBlobWriterFactory {
...
+                        latest.blob_id,
+                        latest.event_stream_cursor,
+                        latest.epoch,
+                    );
+                    wb.insert_batch(&certified, std::iter::once(((), certified_metadata)))?;
+                    wb.delete_batch(&attested, std::iter::once(()))?;
+                    wb.delete_batch(&failed_to_attest, std::iter::once(()))?;
+                    for entry in pending.safe_iter() {
+                        wb.delete_batch(&pending, std::iter::once(entry?.0))?;
+                    }
```
mlegner: Nit/question: Is there any benefit of `std::iter::once` over `[]`?
```suggestion
                    wb.insert_batch(&certified, [((), certified_metadata)]);
                    wb.delete_batch(&attested, [()]);
                    wb.delete_batch(&failed_to_attest, [()]);
                    for entry in pending.safe_iter() {
                        wb.delete_batch(&pending, [entry?.0]);
                    }
```

Also, question: It seems we delete everything in `pending`; is there no simpler way of doing it than calling delete on each of the entries?

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -422,15 +424,89 @@ impl EventBlobWriterFactory {
...
+                } else {
+                    // The last certified event blob is same as the local certified event blob.
+                    // We reset the uncertified blobs to recover from potential global fork where
+                    // nodes cannot agree on the last certified event blob.
+                    Self::reset_uncertified_blobs(
```
mlegner: Question (not related to this PR): Why do we pass in all the DB tables instead of having `&self` as a first parameter?

crates/walrus-simtest/tests/simtest_event_blob.rs
halfprice: Sorry for not being clear. I understand that it's hard to differentiate slow vs. stuck. I just wonder if you verified in a test case where the node's event blob is indeed stuck.

crates/walrus-service/src/node/events/event_blob_writer.rs
halfprice: +1

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -422,15 +424,89 @@ impl EventBlobWriterFactory {
...
+                    .filter(|current_certified| {
+                        latest.event_stream_cursor.element_index
+                            > current_certified.event_cursor.element_index
+                    })
+                    .is_some()
+                {
+                    // The last certified event blob is ahead of the local certified event blob.
+                    // We skip past all the blobs until the last certified event blob and delete
+                    // all the uncertified blobs. If this was a local fork, the pending blobs
+                    // will also be in bad shape and discarding them is needed to recover.
```
halfprice: Nit: in your other comment, you mentioned that you were adding more comments about throwing away local pending/attested blob state, even though the node may not fork. This feels like the place, but looks like it hasn't been updated yet. Maybe you have a local comment hasn't pushed yet?

crates/walrus-simtest/tests/simtest_event_blob.rs
```diff
@@ -168,6 +285,49 @@ mod tests {
...
+        // Wait for event blob certification to get stuck
+        wait_for_event_blob_writer_to_fork(&mut walrus_cluster, &client, 0)
+            .await
+            .unwrap();
+
+        // Restart nodes with same checkpoint number to recover
+        restart_node_with_checkpoints(&mut walrus_cluster, 0, |_| 20).await;
+
+        // Verify recovery
+        wait_for_node_to_recover(&walrus_cluster.nodes[0])
```
halfprice: Nit: s/wait_for_node_to_recover/wait_for_event_blob_writer_to_recover

node to recover may be misleading since nodes also have a recovery mode to recover processing events.

crates/walrus-service/src/node/events/event_blob_writer.rs
```diff
@@ -422,15 +424,89 @@ impl EventBlobWriterFactory {
...
+                    let certified_metadata = CertifiedEventBlobMetadata::new(
+                        latest.blob_id,
+                        latest.event_stream_cursor,
+                        latest.epoch,
+                    );
+                    wb.insert_batch(&certified, std::iter::once(((), certified_metadata)))?;
+                    wb.delete_batch(&attested, std::iter::once(()))?;
+                    wb.delete_batch(&failed_to_attest, std::iter::once(()))?;
+                    for entry in pending.safe_iter() {
+                        wb.delete_batch(&pending, std::iter::once(entry?.0))?;
```
halfprice: Question: just want to make sure, the deleted attested and pending blob files will be deleted eventually, correct?

crates/walrus-simtest/tests/simtest_event_blob.rs
sadhansood: Yes ofcourse, I verified it to be forking!

crates/walrus-service/src/node/events/event_blob_writer.rs
halfprice, line 461: Should `cleanup_orphaned_blobs` be moved after this `match` statement, so that removed certified/attested/pending local blobs can be cleared?

halfprice (approved):
Thanks @sadhansood for address the comments, and appreciated the created simtest! The PR LGTM

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: I don't think that is possible to do via DBBatch, so it isn't guaranteed to be atomic with other operations.

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: These are the relevant comment line just above: 
```
// The last certified event blob is ahead of the local certified event blob.
// We skip past all the blobs until the last certified event blob and delete
// all the uncertified blobs. If this was a local fork, the pending blobs
// will also be in bad shape and discarding them is needed to recover.
```

crates/walrus-service/src/node/events/event_blob_writer.rs
sadhansood: Yes, they will be.

PR #2049, opened 2025-05-05
refactor: remove ClientErrorKind and some cleanup

wbbradley:
## Description

This is primarily cleanup. The ClientErrorKind type was unnecessary and caused bloat in various places. This PR removes it, and removes a couple extra methods placed on the walrus_sdk::client::Client, unifying them for both the read and write paths.

## Test plan

CI Pipeline.

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -93,7 +93,7 @@ async fn read_blob_with_wait_for_certification(
             Ok(data) => return Ok(data),
             Err(err) => {
                 // Check if the error is BlobIdDoesNotExist.
-                if matches!(err.kind(), ClientErrorKind::BlobIdDoesNotExist) {
+                if matches!(err, ClientError::BlobIdDoesNotExist) {
                     tokio::time::sleep(tokio::time::Duration::from_millis(500)).await;
                     continue;
                 }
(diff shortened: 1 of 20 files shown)


mlegner (commented):
AFAIR, having a simple outer error type with an inner `kind` that contains more information was intentional when this was originally set up in #339 (cc @jpcsmith @giac-mysten). This is similar to what some other crates do, including [`std::io`](https://doc.rust-lang.org/std/io/enum.ErrorKind.html) and [`std::num`](https://doc.rust-lang.org/std/num/enum.IntErrorKind.html). I'm not absolutely opposed to simplifying it, but would like to get input from @giac-mysten and/or @jpcsmith first.

What we *should* do anyway IMO is adding a [`non_exhaustive` attribute](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) to some of our public types.

wbbradley:
> What we _should_ do anyway IMO is adding a [`non_exhaustive` attribute](https://doc.rust-lang.org/reference/attributes/type_system.html#the-non_exhaustive-attribute) to some of our public types.

I can add that change, although my motivation for doing so would be to push people to avoid even trying to handle all the ways our system can fail. I think I would avoid rationalizing it based on source compatibility. That seems unnecessary considering we are not at the stage of the project where source compatibility is super important (long-term maintenance mode where new operators come along with zero knowledge of the system, they just want to recompile to keep their systems online).

giac-mysten:
hey @wbbradley as @mlegner mentioned this was set up to follow the practice of not exposing the full list of errors. This is to avoid developers expecting a stable API and matching against it. @jpcsmith was a strong proponent of this 😬 so i'd also like to hear from him about this.

wbbradley:
@jpcsmith What is the functional difference between having this layer of concrete indirection and not having it? I could understand if it was hiding an abstraction, or if it was hidden behind a link-time dependency that allowed for future compat. But this does neither, it simply requires the callsites to call `.kind()` to access the error, and `ClientError(...)` to wrap the kind in the error when constructing them.

wbbradley:
Closing this PR.

PR #2052, opened 2025-05-06
refactor: Wallet wrapper to isolate network traffic

wbbradley:
## Description

The intent here is to prevent network access via `WalletContext` which lacks all of the retry-failover-fallback semantics of retry_client.

## Test plan

CI Pipeline.


crates/walrus-sui/src/wallet.rs
```diff
@@ -0,0 +1,132 @@
...
+    pub fn active_address(&mut self) -> Result<SuiAddress, WalletError> {
+        self.wallet_context.active_address()
+    }
+
+    /// Passes through to the `WalletContext` to sign a transaction.
+    pub fn sign_transaction(&self, transaction_data: &TransactionData) -> Transaction {
+        self.wallet_context.sign_transaction(transaction_data)
+    }
+
+    // TODO: move callsites to the RetriableSuiClient.
```
mlegner: Nit: Can you create a follow-up issue for this (unless you address it already in this PR)?

crates/walrus-sui/src/wallet.rs
```diff
@@ -0,0 +1,132 @@
...
+    /// Get the rpc_url for the active environment.
+    #[deprecated(
+        note = "Avoid calling this method as it relies on the Wallet configuration for rpc urls. \
+        It's better to pass rpc urls directly to the location where they are needed, from Walrus \
+        configuration."
+    )]
+    pub fn get_rpc_url(&self) -> Result<String, WalletError> {
+        Ok(self.wallet_context.config.get_active_env()?.rpc.clone())
+    }
+    /// Get the path to the wallet configuration file.
```
mlegner, line 127: Nit: Can you add blank lines between functions?

(Side note: I'm surprised that rustfmt doesn't actually do this... 🤔)

mlegner (commented):
Thanks for the efforts to make the wallet operations more robust. I think I agree with the general direction, but @karlwuest is more familiar with the whole wallet and Sui code. So it would be better if he could take a closer look here in my stead.

crates/walrus-sui/src/wallet.rs
```diff
@@ -0,0 +1,132 @@
...
+        forbidden_gas_objects: BTreeSet<ObjectID>,
+    ) -> Result<(u64, SuiObjectData), WalletError> {
+        self.wallet_context
+            .gas_for_owner_budget(address, budget, forbidden_gas_objects)
+            .await
+    }
+
+    /// Update the `Move.lock` file with automated address management info. See
+    /// [`sui_package_management::update_lock_file`] for details.
+    pub async fn update_lock_file(
```
karlwuest, line 100: Remark: I'm wondering if we can get rid of this function, too. That would allow us to use the `Wallet` only for signing operations, which would give us more flexibility in the future for different ways of loading the wallet, e.g. directly from the walrus config instead of requiring a sui config etc.

I don't think this is concretely actionable at this point, but it may be worth to create an issue to revisit this? This would also require some changes to `sui_package_management::update_lock_file` unless we want to duplicate some of that code.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -289,26 +255,26 @@ impl RetriableSuiClient {
         &self,
         address: SuiAddress,
         coin_type: Option<String>,
-        amount: u128,
+        amount: Option<u128>,
```
karlwuest, line 271: Nit: I know that there was already a mismatch between the rustdoc comment and the function signature and functionality before this PR, but can you fix the comment? In particular, this is a new function and not a reimplementation of `sui_sdk::apis::CoinReadApi::select_coins`.

karlwuest (commented):
Thanks @wbbradley, this looks good! Please ping me once the PR is no longer a draft.
Are you planning to change the callsites as part of this PR or in a follow up?

crates/walrus-sui/src/wallet.rs
wbbradley: Done.

crates/walrus-sui/src/wallet.rs
wbbradley: Done.

wbbradley:
@karlwuest I've put up [this PR](https://github.com/MystenLabs/sui/pull/22112) in the Sui repo to decouple the `update_lock_file` behavior from the `WalletContext`.

crates/walrus-sui/src/wallet.rs
wbbradley: Added a TODO https://linear.app/mysten-labs/issue/WAL-821/refactor-update-lock-file-in-sui-to-allow-bypassing-walletcontext

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
wbbradley: Ok, I've updated the docs for this method.

wbbradley:
OK @karlwuest this should be ready for a more formal review. Thanks!

karlwuest (approved):
Thanks, LGTM!
Sorry for the reviewing delay, I missed the notification e-mail 😅

PR #2053, opened 2025-05-07
chore: add register_blob call to subsidies contract

zihehuang:
## Description

Add register_blob proxy call to subsidies contract. This would enable custom subsidies contract without modifying the client CLI significantly.

Also updated the subsidy calculation logic so that a buyer only needs to supply the post-subsidy portion of the WAL to pay for `extend_blob` and `reserve_space` calls. Before the update, a buyer has to supply the full amount of WAL of wall and then get refunded through the old `apply_subsidies` logic

## Test plan

Move test

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/subsidies/Move.lock
+++ b/contracts/subsidies/Move.lock
@@ -40,7 +40,7 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.47.0"
+compiler-version = "1.46.3"
 edition = "2024.beta"
 flavor = "sui"
 
--- a/contracts/subsidies/Move.lock
+++ b/contracts/subsidies/Move.lock
@@ -2,7 +2,7 @@
 
 [move]
 version = 3
-manifest_digest = "E796F3002DAB1A928B475FF3B73F42AF9036B570D87F6F8E8A09ECDF0A6A41B7"
+manifest_digest = "C8CDFE3465A9A9D4E9E9181FF2AA73860783105D612DE114A7C7B424B805878F"
 deps_digest = "060AD7E57DFB13104F21BE5F5C3759D03F0553FC3229247D9A7A6B45F50D03A3"
 dependencies = [
   { id = "Sui", name = "Sui" },
@@ -40,7 +40,7 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.46.3"
+compiler-version = "1.48.1"
 edition = "2024.beta"
 flavor = "sui"
 
--- a/contracts/subsidies/sources/subsidies.move
+++ b/contracts/subsidies/sources/subsidies.move
@@ -133,10 +133,81 @@ public fun set_system_subsidy_rate(self: &mut Subsidies, cap: &AdminCap, new_rat
     self.system_subsidy_rate = new_rate;
 }
 
+fun allocate_subsidies(self: &Subsidies, cost: u64, initial_pool_value: u64): (u64, u64) {
+    // Return early if the subsidy pool is empty.
+    if (initial_pool_value == 0) {
+        return (0, 0)
+    };
+    let buyer_subsidy = cost * (self.buyer_subsidy_rate as u64) / (MAX_SUBSIDY_RATE as u64);
+    let system_subsidy = cost * (self.system_subsidy_rate as u64) / (MAX_SUBSIDY_RATE as u64);
+    let total_subsidy = buyer_subsidy + system_subsidy;
+
+    // Apply subsidy up to the available amount in the pool.
+    if (initial_pool_value >= total_subsidy) {
+        (buyer_subsidy, system_subsidy)
+    } else {
+        // If we don't have enough in the pool to pay the full subsidies,
+        // split the remainder proportionally between the buyer and system subsidies.
+        let pool_value = initial_pool_value;
+        let total_subsidy_rate = self.buyer_subsidy_rate + self.system_subsidy_rate;
+        let buyer_subsidy =
+            pool_value * (self.buyer_subsidy_rate as u64) / (total_subsidy_rate as u64);
+        let system_subsidy =
+            pool_value * (self.system_subsidy_rate as u64) / (total_subsidy_rate as u64);
+        (buyer_subsidy, system_subsidy)
+    }
+}
+
+/// Combine payment and subsidy pool
+fun pre_system_call(
+    self: &mut Subsidies,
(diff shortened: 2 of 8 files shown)


contracts/subsidies/Move.lock
```diff
@@ -40,7 +40,7 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.47.0"
```
karlwuest, line 43: Nit: Can you roll back these changes to the lock files?

zihehuang: yeah im gonna update this once I have better internet connection

contracts/subsidies/sources/subsidies.move
```diff
@@ -133,48 +133,90 @@ public fun set_system_subsidy_rate(self: &mut Subsidies, cap: &AdminCap, new_rat
...
+#[allow(unused_function, unused_variable, unused_mut_parameter)]
+fun apply_subsidies(
+    self: &mut Subsidies,
+    cost: u64,
+    epochs_ahead: u32,
+    payment: &mut Coin<WAL>,
+    system: &mut System,
+    ctx: &mut TxContext,
+) {}
+
```
karlwuest: Suggestion: If this is no longer used, we can just delete it, since it's not a public function.

```suggestion
```

contracts/subsidies/sources/subsidies.move
```diff
@@ -133,48 +133,90 @@ public fun set_system_subsidy_rate(self: &mut Subsidies, cap: &AdminCap, new_rat
...
+    payment: &mut Coin<WAL>,
+    ctx: &mut TxContext,
+): (u64, u64) {
+    let initial_payment_value = payment.value();
+    let pool_coin = self.subsidy_pool.withdraw_all().into_coin(ctx);
+    let initial_pool_value = pool_coin.value();
+
+    payment.join(pool_coin);
+    (initial_payment_value, initial_pool_value)
+}
```
karlwuest: Important:
This function is quite dangerous, since it merges the whole subsidy pool into the payment coin which is owned by the caller, so this relies fully on the post-call handling to make sure that we don't empty the pool. I think there are two things that we can/should do to improve this:
1. Find a more descriptive name.
2. Instead of returning a tuple of two u64s and modifying the payment coin, return a new coin (i.e. merge the balance of `payment` into `pool_coin`. Then, this new coin and the `initial_payment_value, initial_pool_value` could be stored in a struct that does not have the `key` or `store` ability and from which we then use the coin as payment for the system calls + for the post-call handling.

contracts/subsidies/sources/subsidies.move
```diff
@@ -158,65 +158,64 @@ fun allocate_subsidies(self: &Subsidies, cost: u64, initial_pool_value: u64): (u
...
     ctx: &mut TxContext,
-): (u64, u64) {
+): CombinedPayment {
     let initial_payment_value = payment.value();
-    let pool_coin = self.subsidy_pool.withdraw_all().into_coin(ctx);
+    let mut pool_coin = self.subsidy_pool.withdraw_all().into_coin(ctx);
     let initial_pool_value = pool_coin.value();
+    let buyer_coin = payment.split(initial_payment_value, ctx);
+
+    pool_coin.join(buyer_coin);
```
karlwuest: Nit: I think we can avoid creating another coin here, which avoids the overhead of creating a new object ID:
```suggestion
    let buyer_balance = payment.balance_mut().split(initial_payment_value);

    pool_coin.balance_mut().join(buyer_balance);
```

contracts/subsidies/sources/subsidies.move
```diff
@@ -158,65 +158,64 @@ fun allocate_subsidies(self: &Subsidies, cost: u64, initial_pool_value: u64): (u
...
+    let cost = initial_combined_value - remaining_coin.value();
 
     // Refund buyer
-    self.subsidy_pool.join(payment.split(remaining_value, ctx).into_balance());
+    self.subsidy_pool.join(remaining_coin.into_balance());
     let (buyer_subsidy, system_subsidy) = self.allocate_subsidies(cost, initial_pool_value);
     let buyer_coin_value = initial_payment_value + buyer_subsidy - cost;
     let buyer_refund = self.subsidy_pool.split(buyer_coin_value).into_coin(ctx);
-    payment.join(buyer_refund);
+    buyer_payment.join(buyer_refund);
```
karlwuest: Nit: I think we can again avoid splitting off a coin, and just join the balance of `buyer_payment` with the  balance that is split from the subsidy pool (see comment above).

contracts/subsidies/sources/subsidies.move
```diff
@@ -233,16 +232,25 @@ public fun extend_blob(
...
 
-    system.extend_blob(blob, epochs_ahead, payment);
+    system.extend_blob(blob, epochs_ahead, &mut combined_coin);
 
-    handle_subsidies(
+    let combined_payment_after = CombinedPayment {
+        payment: combined_coin,
+        initial_payment_value,
+        initial_pool_value,
+    };
```
karlwuest: Suggestion: Instead of destructuring the CombinedPayment before the call and then creating a new one, you should be able to just access the coin while it's in the struct:

````
    let mut combined_payment = self.combine_payment_with_pool(payment, ctx);
    
    system.extend_blob(blob, epochs_ahead, &mut combined_payment.payment);
````

File: contracts/subsidies/sources/subsidies.move [hunk 7]
```diff
@@ -264,16 +272,25 @@ public fun reserve_space(
...
 
-    let storage = system.reserve_space(storage_amount, epochs_ahead, payment, ctx);
+    let storage = system.reserve_space(storage_amount, epochs_ahead, &mut combined_coin, ctx);
 
-    handle_subsidies(
+    let combined_payment_after = CombinedPayment {
+        payment: combined_coin,
+        initial_payment_value,
+        initial_pool_value,
+    };
```
karlwuest: Suggestion: Similar to above, there is no need to destructure the CombinedPayment.
````
    let combined_payment = self.combine_payment_with_pool(payment, ctx);
    let storage = system.reserve_space(storage_amount, epochs_ahead, &mut combined_payment.payment, ctx);
````

karlwuest (approved):
A few small last comments and then it's good to go!

PR #2057, opened 2025-05-07
chore: bump 1.24 version

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


ebmifa:
@halfprice we don't actually need this, since v1.24.2 already contains the latest fix

ebmifa:
> I see. So do we bump it if we need to create another patch version later? I saw that you bumped to 1.24.2 after creating the 1.24.1 release.

Yes, we only bump it if we know a fix is coming. In this case, I did not realize that Marcus has already bumped it to v1.24.1

PR #2060, opened 2025-05-07
test: retry storing metadata during cluster creation

mlegner:
## Description

Before this change, some unit tests, specifically `recovers_sliver_from_only_symbols_of_one_type` were failing from time to time with an error `the blob has not been registered or has already expired`. This error occurs when attempting to store the metadata before the node received the registration event. With this change, storing the metadata is retried, thus avoiding the issue.

## Test plan

Test suite + stress testing the corresponding test with [cargo-stress](https://github.com/danhhz/cargo-stress):

```sh
cargo stress recovers_sliver_from_only_symbols_of_one_type
```


halfprice (approved):
Thanks for the investigation and fix @mlegner !

PR #2061, opened 2025-05-07
chore: make ClientErrorKind non-exhaustive

wbbradley:
## Description

As per https://github.com/MystenLabs/walrus/pull/2049#pullrequestreview-2817348865, we would like to make `ClientErrorKind` non-exhaustive.

## Test plan

CI Pipeline.

### DIFF
--- a/crates/walrus-service/src/node/metrics.rs
+++ b/crates/walrus-service/src/node/metrics.rs
@@ -291,6 +291,8 @@ impl TelemetryLabel for ClientErrorKind {
             ClientErrorKind::FailedToLoadCerts(_) => "failed-to-load-certs",
             ClientErrorKind::Other(_) => "unknown",
             ClientErrorKind::StoreBlobInternal(_) => "store-blob-internal",
+            // Note: this is required due to the `#[non_exhaustive]` attribute on the enum.
+            _ => "telemetry-label-unhandled",
         }
     }
 }
--- a/crates/walrus-sdk/src/error.rs
+++ b/crates/walrus-sdk/src/error.rs
@@ -126,6 +126,7 @@ impl From<SuiClientError> for ClientError {
 /// Inner error type, raised when the client operation fails.
 #[derive(Debug, thiserror::Error)]
 #[error(transparent)]
+#[non_exhaustive]
 pub enum ClientErrorKind {
     /// The certification of the blob failed.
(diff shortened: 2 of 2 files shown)


crates/walrus-service/src/node/metrics.rs
```diff
@@ -291,6 +291,8 @@ impl TelemetryLabel for ClientErrorKind {
             ClientErrorKind::FailedToLoadCerts(_) => "failed-to-load-certs",
             ClientErrorKind::Other(_) => "unknown",
             ClientErrorKind::StoreBlobInternal(_) => "store-blob-internal",
+            // Note: this is required due to the `#[non_exhaustive]` attribute on the enum.
+            _ => "telemetry-label-unhandled",
         }
```
wbbradley, line 296: We are effectively creating work for ourselves later by doing this, as this compiler error/fix exhibits. The upsides seem negligible compared to the downsides.

`#[non_exhaustive]` seems to make it harder to augment sum-types with extra downstream functionality. 

This is a case where it is harmless to force exhaustive matching, yet this change makes it impossible to be exhaustive in the future without unnecessary toil or slippage in handling new cases, etc.

mlegner: Hmm... I guess the issue is that this implementation is outside the crate defining `ClientErrorKind` (as the `#[non_exhaustive]` doesn't have an effect in the defining crate). A way to work around this would be to define a function `label` or similar in `walrus-sdk` and then just call it here in the implementation of `TelemetryLabel`. WDYT?

But I do see the downside of the `non_exhaustive` attribute. I'm happy to keep it as it is now without adding it if you prefer.

mlegner (approved):
Approving the PR, but it's also fine for me to keep the behavior as before.

crates/walrus-service/src/node/metrics.rs
wbbradley: We could add a `.label()` method but this bleeds the requirements of the users into the implementation when they could just as well do it themselves if we avoid `non_exhaustive`. I guess I don't see the point of the `non_exhaustive` attribute except for source-compatibility, but I don't think that's high on our priority list? I'm going to go ahead and close this, but I don't have strong feelings about it. I just don't see the benefit. Feel free to reopen.

PR #2062, opened 2025-05-07
chore: bump version of all Sui dependencies to v1.48.0 testnet release

wbbradley:
## Description

Bump Sui and its necessary dependencies to v1.48.0.

## Test plan

CI Pipeline.

PR #2063, opened 2025-05-08
feat: Add typed store and derive crate to walrus repo

sadhansood:
## Description

Add typed store crate. The reasons for this change are as follows:
1. We can control Rocksdb version independent of Sui
2. The storage interface that Walrus needs is probably simpler than what typed-store provides and we also want to control this interface going forward (and likely simplify/change it according to our needs)
3. Using vanilla rust rocksdb is not enough since we lose out on metrics, perf context, etc. The plan is to first fork and then update rocksdb version.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,8 @@
 [workspace]
 default-members = [
   "crates/checkpoint-downloader",
+  "crates/typed-store",
+  "crates/typed-store-derive",
   "crates/walrus-core",
   "crates/walrus-e2e-tests",
   "crates/walrus-proc-macros",
@@ -40,17 +42,21 @@ checkpoint-downloader = { path = "crates/checkpoint-downloader" }
 chrono = "0.4"
 clap = { version = "4.5.37", features = ["derive"] }
 clap_complete = "4.5.48"
+collectable = "0.0.2"
 colored = "2.2.0"
 criterion = "0.5.1"
 diesel = { version = "2.2", features = ["chrono", "postgres", "uuid"] }
 diesel-async = { version = "0.5", features = ["postgres"] }
 diesel_migrations = { version = "2.2.0", features = ["postgres"] }
 enum_dispatch = "0.3"
+eyre = "0.6.12"
 fastcrypto = { git = "https://github.com/MystenLabs/fastcrypto", rev = "69d496c71fb37e3d22fe85e5bbfd4256d61422b9" }
+fdlimit = "0.3.0"
 futures = { version = "0.3.31", default-features = false, features = ["async-await", "std"] }
 futures-timer = "=3.0.3" # required for MSIM
 futures-util = "0.3.30"
 git-version = "0.3.9"
+hdrhistogram = "7.5.4"
 home = "0.5.11"
 hostname = "0.4.1"
 http-body-util = "0.1.1"
@@ -76,6 +82,7 @@ opentelemetry = { version = "=0.27.1", default-features = false, features = ["tr
 p256 = { version = "0.13.2", default-features = false }
 pin-project = "1.1.10"
 prettytable = "0.10.0"
+proc-macro2 = "1.0.95"
 prometheus = "0.13.4"
 quote = "1.0"
 rand = "0.8.5"
@@ -86,6 +93,7 @@ reed-solomon-simd = "3.0.1"
 regex = "1"
 reqwest = { version = "0.12.12", default-features = false, features = ["http2", "json", "rustls-tls"] }
 rocksdb = "0.21.0"
+rstest = "0.25.0"
 rustls = { version = "0.23.26", default-features = false, features = ["logging", "ring", "tls12"] }
 rustls-native-certs = "0.8.1"
 scoped-futures = "0.1.4"
@@ -110,6 +118,7 @@ sui-simulator = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.4
 sui-storage = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.47.0" }
 sui-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.47.0" }
 syn = "2.0"
+tap = "1.0.1"
 telemetry-subscribers = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.47.0" }
 tempfile = "3.19.1"
 test-cluster = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.47.0" }
@@ -124,8 +133,10 @@ tracing = "0.1.41"
 tracing-opentelemetry = { version = "=0.28.0", default-features = false }
 tracing-subscriber = { version = "0.3.19", default-features = false, features = ["fmt"] }
 twox-hash = "2.1.0"
+typed-store = { path = "crates/typed-store" }
+typed-store-derive = { path = "crates/typed-store-derive" }
+uint = "0.10.0"
 # TODO (WAL-790): update typed-store to >= 1.47 or remove it.
-typed-store = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.46.0" }
 url = "2.5.4"
 utoipa = { version = "5" }
 utoipa-redoc = { version = "6.0", features = ["axum"] }
--- a/crates/typed-store-derive/Cargo.toml
+++ b/crates/typed-store-derive/Cargo.toml
@@ -1,0 +1,20 @@
+[package]
+name = "typed-store-derive"
+publish = false
+authors.workspace = true
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[lib]
+proc-macro = true
+
+[dependencies]
+itertools.workspace = true
+proc-macro2.workspace = true
+quote.workspace = true
+syn = { workspace = true, features = ["full"] }
+
+
+[lints]
+workspace = true
--- a/crates/typed-store-derive/src/lib.rs
+++ b/crates/typed-store-derive/src/lib.rs
@@ -1,0 +1,625 @@
+// Copyright (c) Mysten Labs, Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+//! Derive macros for the typed-store crate.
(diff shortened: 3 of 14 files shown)


crates/typed-store-derive/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "typed-store-derive"
+publish = false
+authors.workspace = true
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
```
mlegner, line 8: Nit: Can we also add the following?

```suggestion

[lints]
workspace = true

```

crates/typed-store/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "typed-store"
+publish = false
+authors.workspace = true
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
```
mlegner, line 8: Nit: Can we also add the following?

```suggestion

[lints]
workspace = true

```

Cargo.toml
```diff
@@ -68,13 +74,16 @@ mime = "0.3.17"
 mockall = "0.12.1"
 move-core-types = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.47.0" }
 move-package = { git = "https://github.com/MystenLabs/sui", tag = "testnet-v1.47.0" }
+msim = { git = "https://github.com/MystenLabs/mysten-sim.git", rev = "9f175e3517812929ad6bdd8db443f1bbd8107965", package = "msim" }
```
mlegner: Question: (Why) do we need this explicitly?

crates/typed-store-derive/src/lib.rs
```diff
@@ -0,0 +1,618 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
```
mlegner: Remark: I'm not quite sure about the copyright headers given that this is copied from the Sui repo. Maybe we could keep the Mysten copyright or ML + WF and add the corresponding files to `.licensesnipignore`?

crates/typed-store/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+// Copyright (c) 2021, Facebook, Inc. and its affiliates
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
```
mlegner: Remark: The headers here and in other places need to be fixed as well.

crates/typed-store/src/rocks/mod.rs
mlegner, line 1: Nit: Can we name this `crates/typed-store/src/rocks.rs` following the "modern" naming convention?

crates/typed-store-derive/src/lib.rs
mlegner, line 1: Remark: I assume that the code is a copy from the Sui repo at `testnet-v1.46.X`. Is that correct? I'm not reviewing the code in detail then.

mlegner (commented):
Very nice, @sadhansood! A few nits from my side.

Can you also describe the rationale behind this move in the PR description for future reference?

crates/typed-store-derive/src/lib.rs
```diff
@@ -0,0 +1,618 @@
...
+
+struct ExtractedStructInfo {
+    field_names: Vec<Ident>,
+    cf_names: Vec<Ident>,
+    inner_types: Vec<AngleBracketedGenericArguments>,
+    derived_table_options: Vec<GeneralTableOptions>,
+    deprecated_cfs: Vec<Ident>,
+}
+
+#[proc_macro_derive(DBMapUtils, attributes(default_options_override_fn, rename))]
```
halfprice, line 191: For the entire `typed-store-derive` crate, I don't think we need it. The table derived from DBMapUtils is not used in Walrus and I don't think we will use it in the future. Let's just remove them.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -0,0 +1,2246 @@
...
+        Self {
+            ignore_range_deletions: true,
+            sync_to_disk: std::env::var("SUI_DB_SYNC_TO_DISK").is_ok_and(|v| v != "0"),
+        }
+    }
+}
+// TODO: refactor this into a builder pattern, where rocksdb::Options are
+// generated after a call to build().
+#[derive(Default, Clone)]
+pub struct DBOptions {
```
halfprice: I don't think we are using DBOptions in Walrus. Consider removing it?

crates/typed-store/src/rocks/mod.rs
```diff
@@ -0,0 +1,2246 @@
...
+    pub options: rocksdb::Options,
+    pub rw_options: ReadWriteOptions,
+}
+
+impl DBOptions {
+    // Optimize lookup perf for tables where no scans are performed.
+    // If non-trivial number of values can be > 512B in size, it is beneficial to also
+    // specify optimize_for_large_values_no_scan().
+    pub fn optimize_for_point_lookup(mut self, block_cache_size_mb: usize) -> DBOptions {
```
halfprice: I think we don't need these customized options created for Sui. Let's just keep the default_db_options.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -0,0 +1,2246 @@
...
+            rocksdb,
+            metric_conf,
+            PathBuf::from(path),
+        )))
+    })
+}
+
+/// Opens a database with options, and a number of column families with individual options that
+/// are created if they do not exist.
+pub fn open_cf_opts_secondary<P: AsRef<Path>>(
```
halfprice: Can be removed as well, seems specific to Sui

crates/typed-store/src/rocks/mod.rs
```diff
@@ -0,0 +1,2246 @@
...
+{
+    bincode::DefaultOptions::new()
+        .with_big_endian()
+        .with_fixint_encoding()
+        .serialize(t)
+        .map_err(typed_store_err_from_bincode_err)
+}
+
+#[derive(Clone)]
+pub struct DBMapTableConfigMap(BTreeMap<String, DBOptions>);
```
halfprice: If `typed-store-derived` is removed, this can be removed too.

crates/typed-store/src/rocks/mod.rs
```diff
@@ -0,0 +1,2246 @@
...
+    pub fn new(map: BTreeMap<String, DBOptions>) -> Self {
+        Self(map)
+    }
+
+    pub fn to_map(&self) -> BTreeMap<String, DBOptions> {
+        self.0.clone()
+    }
+}
+
+pub enum RocksDBAccessType {
```
halfprice: Not used

crates/typed-store/src/traits.rs
```diff
@@ -0,0 +1,128 @@
...
+
+pub struct TableSummary {
+    pub num_keys: u64,
+    pub key_bytes_total: usize,
+    pub value_bytes_total: usize,
+    pub key_hist: hdrhistogram::Histogram<u64>,
+    pub value_hist: hdrhistogram::Histogram<u64>,
+}
+
+pub trait TypedStoreDebug {
```
halfprice, line 119: Can also be removed if `typed-store-drive` is removed. This is not used in Walrus.

Cargo.toml
halfprice: I think this is because 

[target.'cfg(msim)'.dependencies]
msim.workspace = true

In typed-store/Cargo.toml


@sadhansood , use

[target.'cfg(msim)'.dependencies]
sui-simulator.workspace = true

and then you can remove msim here.

halfprice (changes requested):
Thanks a lot @sadhansood for quick action on migrating this!

Most of my comments are about removing stuff that we don't use in Walrus, and they were created too specific to suit Sui's need. I feel some of them are there to make test happy, and therefore, I'm ok with commenting out some of the tests, and see if we can bring them back without using those Sui specific functionality, or remove them.

halfprice (approved):
Synced with @sadhansood offline. Since this PR should be a no-op to the existing repo, and we want to unblock #2062 . We can address some of the refactoring in a followup PR.

crates/typed-store-derive/src/lib.rs
sadhansood: done

crates/typed-store/src/rocks/mod.rs
sadhansood: done

crates/typed-store/src/lib.rs
sadhansood: done

crates/typed-store-derive/src/lib.rs
sadhansood: as discussed offline, we will do this in a follow up PR

crates/typed-store/src/rocks/mod.rs
sadhansood: as discussed offline, we will do this in a follow up PR

crates/typed-store/src/rocks/mod.rs
sadhansood: as discussed offline, we will do this in a follow up PR

crates/typed-store/src/rocks/mod.rs
sadhansood: as discussed offline, we will do this in a follow up PR

crates/typed-store/src/rocks/mod.rs
sadhansood: will do in a follow up as discussed offline

crates/typed-store/src/traits.rs
sadhansood: will do in a follow up as discussed offline

crates/typed-store-derive/src/lib.rs
sadhansood: yes, that is correct

PR #2067, opened 2025-05-08
fix(node): reduce level of some logging events

mlegner:
## Description

This PR reduces the verbosity level of a few types of log events:
- During each epoch change, there is a `ShardsReceived` event emitted for every node, so ~100 on our public networks. These don't need to be logged at the INFO level.
- Problems with pushing metrics don't affect the node operation itself and should thus be reduced from ERROR to WARN.
- Similarly, problems with config sync should only be WARN, not ERROR.
- The log entry `we are in the genesis epoch, not creating any services for committee members` is confusing, in particular as it is also emitted in epoch 1 when attempting to (re-)create services for committee members in the previous epoch. This is reformulated and reduced to DEBUG.

## Test plan

None.

---

## Release notes

- [x] Storage node: Reduce verbosity levels of problems with metrics and config sync from ERROR to WARN.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1127,11 +1127,22 @@ impl StorageNode {
         epoch_change_event: EpochChangeEvent,
     ) -> anyhow::Result<()> {
         mysten_metrics::monitored_scope("ProcessEvent::EpochChangeEvent");
-        tracing::info!(
-            ?epoch_change_event,
-            "{} event received",
-            epoch_change_event.name()
-        );
+        match epoch_change_event {
+            EpochChangeEvent::ShardsReceived(_) => {
+                tracing::debug!(
+                    ?epoch_change_event,
+                    "{} event received",
+                    epoch_change_event.name()
+                );
+            }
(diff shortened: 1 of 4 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1127,11 +1127,22 @@ impl StorageNode {
...
+                );
+            }
+            _ => {
+                tracing::info!(
+                    ?epoch_change_event,
+                    "{} event received",
+                    epoch_change_event.name()
+                );
+            }
+        }
```
liquid-helium, line 1145: Question (may not related to this PR): do we have other metrics/logs that we can use to trace the shard assignment history?

mlegner: Good question. I'm not aware of anything related that we run ourselves. There is [this tool](https://walrus-stats.brightlystake.com/shard-owners) run by one of the node operators, which essentially tracks the results of the `walrus health` command.

mlegner: I created a separate issue for this.

PR #2071, opened 2025-05-09
fix(node): set optimized default params for blob recovery

mlegner:
## Description

The blob recovery seems to massively improve with these changes. Tested on our Mysten Labs nodes and by Natsai, they increase the sliver-recovery rate from ~10 to >200 per second. See the following metrics after making the change on Mysten Labs 0 (change applied and node restarted around 8:00):

image

## Test plan

Already tested on a few nodes on Mainnet.

---

## Release notes

- [x] Storage node: Optimize default parameters for blob recovery to increase the recovery performance.

### DIFF
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -545,7 +545,7 @@ pub struct BlobRecoveryConfig {
 impl Default for BlobRecoveryConfig {
     fn default() -> Self {
         Self {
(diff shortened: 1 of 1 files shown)


mlegner:
Thanks for the context, @halfprice. 

I'll close this PR, but it might still be worth tuning the default parameters, maybe @jpcsmith could take a look. We don't necessarily need to directly use the _optimal_ values, but if we are aware of _better_ values than the current ones, we should probably set them.

PR #2072, opened 2025-05-09
fix(node): increase limit for number of pending accept reset streams

mlegner:
## Description

Storage nodes frequently cancel pending requests to other storage nodes when they are no longer relevant. This triggers a limit within the `h2` crate and an associated warning that pollutes the logs.

Until we change the behavior of cancelling pending requests, we can increase the limit to a very high value to effectively disable the check.

Contributes to WAL-660.

## Test plan

Set this value in our Mysten Labs 1 node on Mainnet.

---

## Release notes

- [x] Storage node: Increase default value of `http2_max_pending_accept_reset_streams` parameter to disable warnings from the `h2` crate.

### DIFF
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -65,7 +65,7 @@ rest_server:
   http2_max_concurrent_streams: 1000
   http2_initial_stream_window_size: null
   http2_initial_connection_window_size: null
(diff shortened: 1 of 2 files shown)


halfprice (approved):
Thanks @mlegner ! I think there might be an implication for this change that now there might be unlimited number of cancellable streams maintained on the server side. We should pay attention to the server memory usage.

PR #2082, opened 2025-05-12
chore: include extensions in gas benchmark

karlwuest:
## Description

Adds blob extensions to the gas benchmark.
Closes WAL-802

## Test plan

Manually ran benchmark.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/benches/gas_cost_bench.rs
+++ b/crates/walrus-sui/benches/gas_cost_bench.rs
@@ -22,7 +22,13 @@ use walrus_core::{
     merkle::Node,
 };
 use walrus_sui::{
-    client::{BlobObjectMetadata, BlobPersistence, PostStoreAction, SuiContractClient},
+    client::{
+        BlobObjectMetadata,
+        BlobPersistence,
+        CertifyAndExtendBlobParams,
+        PostStoreAction,
+        SuiContractClient,
+    },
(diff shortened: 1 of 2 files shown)


crates/walrus-sui/benches/gas_cost_bench.rs
```diff
@@ -191,13 +209,38 @@ async fn gas_cost_for_contract_calls(args: Args) -> anyhow::Result<()> {
...
+                    .iter()
+                    .map(|blob| CertifyAndExtendBlobParams {
+                        blob,
+                        certificate: None,
+                        epochs_extended,
+                    })
+                    .collect();
+                walrus_client
+                    .as_ref()
+                    .certify_and_extend_blobs(&certify_and_extend_params, PostStoreAction::Keep)
```
giac-mysten, line 234: Question: why do we need the `certify_` part here, and not just extend?

giac-mysten (approved):
Thanks @karlwuest !

crates/walrus-sui/benches/gas_cost_bench.rs
karlwuest: That's because the walrus client uses `certify_and_extend_blobs` and we don't have an `extend_blobs` function (just this one or the singular `extend_blob`).

PR #2083, opened 2025-05-13
fix: RPC node failure resistance improvement

halfprice:
## Description

- For both RPC client and Sui client, we only create it when using it. 
- Before this change, the primary RPC node must be available when node is starting up. This PR removes this setup and finds an available RPC node to use when need.
- For checkpoint not produced error, before we constantly switching to other clients. Instead, we should treat the error as retriable error and retry shortly.

## Test plan

simtest

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client/failover.rs
+++ b/crates/walrus-sui/src/client/retry_client/failover.rs
@@ -7,12 +7,7 @@ use std::{collections::BTreeSet, future::Future, iter::once, sync::Arc, time::Du
 use sui_macros::fail_point_if;
 use tokio::sync::Mutex;
 
-use super::{
-    RetriableClientError,
-    RetriableRpcError,
-    ToErrorType,
-    retry_count_guard::RetryCountGuard,
-};
+use super::{RetriableClientError, ToErrorType, retry_count_guard::RetryCountGuard};
 use crate::client::{SuiClientError, SuiClientMetricSet};
 
 /// A trait that defines the implementation of a thunk to build (or re-use) a client lazily.
@@ -134,13 +129,13 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
 
     /// Returns the current client's RPC URL.
     // This is a helper function is mainly used for logging.
-    pub async fn get_current_rpc_url(&self) -> String {
+    pub async fn get_current_rpc_url(&self) -> anyhow::Result<String> {
         match self.state.lock().await.as_ref() {
             Some(state) => state
                 .rpc_url
                 .clone()
-                .unwrap_or_else(|| "unknown_url".to_string()),
-            None => "uninitialized_client".to_string(),
+                .ok_or_else(|| anyhow::anyhow!("no rpc url specified in failover state")),
+            None => Err(anyhow::anyhow!("no client initialized in failover wrapper")),
         }
     }
 
@@ -174,21 +169,20 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
         tried_client_indices: &mut BTreeSet<usize>,
         state: &mut Option<FailoverState<ClientT>>,
     ) -> Result<Arc<ClientT>, FailoverError> {
-        // Check if the state of the FailoverWrapper has already advanced to a client we haven't
-        // tried yet. If so, go ahead and use that one. Note that this condition is subtle as it
-        // mutates the `tried_client_indices`.
-        if let Some(state) = state.as_mut() {
+        // Compute the next wrapped index.
+        let mut next_index = if let Some(state) = state.as_ref() {
+            // Check if the state of the FailoverWrapper has already advanced to a client we haven't
+            // tried yet. If so, go ahead and use that one. Note that this condition is subtle as it
+            // mutates the `tried_client_indices`.
             if tried_client_indices.len() < self.max_tries
                 && tried_client_indices.insert(state.current_index)
             {
                 return Ok(state.client.clone());
             }
-        };
 
-        // Compute the next wrapped index.
-        let mut next_index = if let Some(state) = state.as_ref() {
             (state.current_index + 1) % self.client_count()
         } else {
+            // If the FailoverWrapper is not initialized, start from the first client.
             0
         };
 
@@ -198,7 +192,7 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
             // PLAN phase.
             if tried_client_indices.len() >= self.max_tries {
                 return Err(FailoverError::FailedToGetClient(format!(
-                    "max failovers {} exceeded",
+                    "max failovers exceeded [max_tries={}]",
                     self.max_tries
                 )));
             }
@@ -251,11 +245,7 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
     where
         F: FnMut(Arc<ClientT>, &'static str) -> Fut,
         Fut: Future<Output = Result<R, E>>,
-        E: MakeRetriableError
-            + RetriableRpcError
-            + ToErrorType
-            + std::fmt::Debug
-            + From<FailoverError>,
+        E: MakeRetriableError + ToErrorType + std::fmt::Debug + From<FailoverError>,
     {
         let mut retry_guard = metrics
             .as_ref()
@@ -317,20 +307,28 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
                     return Ok(result);
                 }
                 Err(error) => {
-                    // If the error is not eligible for failover, return it immediately.
-                    if !error.is_retriable_rpc_error() {
-                        tracing::debug!(
-                            ?error,
-                            "error is not eligible for failover, returning the last failed error",
-                        );
-                        return Err(error);
-                    }
+                    // Note that currently, for any kind of errors, we will failover to the next
+                    // client, event including application level errors. Although this is not
+                    // desirable, it is also hard to compose an extensive list of errors that we
+                    // should or should not failover on.
+                    // For any error that is not supposed to failover, we should return the error
+                    // here immediately. This process can be based on experience and add any error
+                    // that is not supposed to failover to the list.
+                    //
+                    // TODO(zhewu): we should add a new trait for this, and implement it for all
+                    // errors that are not supposed to failover.
 
-                    let failed_rpc_url = self.get_current_rpc_url().await;
+                    let failed_rpc_url = self
+                        .get_current_rpc_url()
+                        .await
+                        .unwrap_or_else(|error| error.to_string());
                     match self.fetch_next_client(&mut tried_client_indices).await {
                         Ok(next_client) => {
                             client = next_client;
-                            let next_rpc_url = self.get_current_rpc_url().await;
+                            let next_rpc_url = self
+                                .get_current_rpc_url()
+                                .await
+                                .unwrap_or_else(|error| error.to_string());
                             tracing::event!(
                                 // A custom target for filtering.
                                 target: "walrus_sui::client::retry_client::failover",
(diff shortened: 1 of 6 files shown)


crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -74,7 +74,7 @@ impl MakeRetriableError for RetriableClientError {
 #[derive(Clone)]
 pub struct FailoverWrapper<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug> {
     lazy_client_builders: Vec<BuilderT>,
-    state: Arc<Mutex<FailoverState<ClientT>>>,
+    state: Arc<Mutex<Option<FailoverState<ClientT>>>>,
```
liquid-helium, line 76: nit:  `std::sync::OnceCell::get_or_try_init()` would handle the synchronization and lazy initialization, it could replace both the LazyClientBuilder and FalioverState

wbbradley: I'm not sure how OnceCell could replace LazyClientBuilder or FailoverState. They contains app-specific data.

liquid-helium: Not to 'replace', but to implement LazyClientBuilder with OnceCell, which holds the Client after the first initialization, so that we don't need Arc<Mutex<Option<FailoverState<ClientT>>>>, to track the client and/or if they have been created.
Also it provides build-in synchronization.

halfprice: Thanks @liquid-helium ! It'll be nice to do this if I had read your comment earlier. I used a different approach to address the constant creating new client issue. I think we should not failover in such case, but simply retry the call in the same RPC.

Using the `sync::Once` is a good optimization, but now I think about it, there is another code path where the rpc client monitors the number of client failures and re-create clients, to account for GRPC proxy stuckness issue, and I don't know how complicated to incorporate that logic when using OnceCell though.

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -108,29 +112,36 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
...
 
         Ok(Self {
-            max_tries,
-            state: Arc::new(Mutex::new(FailoverState {
-                client: lazy_client_builders[0].lazy_build_client().await?,
-                rpc_url: lazy_client_builders[0].get_rpc_url().map(|s| s.to_string()),
-                current_index: 0,
-            })),
             lazy_client_builders,
+            state: Arc::new(Mutex::new(None)),
```
wbbradley, line 111: Just noting that this delays the call to `ensure_experimental_rest_endpoint_exists` that was otherwise happening eagerly. This might not be a problem, but just wanted to be sure it's been considered.

halfprice: Yes. I'm aware that and my personal opinion is that we don't really need that `ensure_experimental_rest_endpoint_exists`. Now, even if that fails, we just switch to the next client, which is also better than just failing the node directly.

crates/walrus-sui/src/client/retry_client/failover.rs
liquid-helium: Agreed, if we have to recreate the Client, OnceCell will make it complicated. The current approach looks good, thanks @halfprice.

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -269,6 +317,15 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
                     return Ok(result);
                 }
                 Err(error) => {
+                    // If the error is not eligible for failover, return it immediately.
+                    if !error.is_retriable_rpc_error() {
+                        tracing::debug!(
```
liquid-helium: This seems to be too strict for failover, the current retriable_rpc_error is designed for retry on the same rpc endpoint, but failover should be able to cover more errors.
I can see that when a full node is lagging in a reasonable range, we could retry more on the same fullnode, instead of trying another one, but the same error could also indicate a real lagging.
Perhaps we can increase the retry-timeout for a single node on 'retriable-rpc-errors', then if it times out, we know we need to switch to the next one.

crates/walrus-sui/src/client/retry_client/failover.rs
wbbradley: I agree. Maybe we should put a TODO in to remove ensure_experimental_rest...

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -108,29 +112,36 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
...
+                    .await?;
+                Ok(client)
+            }
+        }
     }
 
-    /// Returns the name of the current client.
+    /// Returns the current client's RPC URL.
+    // This is a helper function is mainly used for logging.
     pub async fn get_current_rpc_url(&self) -> String {
```
wbbradley: Now that the number of error cases for this routine has grown from 1 to 2, I think it is worth changing the return type to anyhow::Result\<String\>, and forcing the caller to understand that the returned value is not actually a URL. This would just bulletproof this method a bit more, so we don't have to read it and wonder whether the callers are trying to use this URL.

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -148,29 +159,48 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
...
+    ) -> Result<Arc<ClientT>, FailoverError> {
         // Check if the state of the FailoverWrapper has already advanced to a client we haven't
         // tried yet. If so, go ahead and use that one. Note that this condition is subtle as it
         // mutates the `tried_client_indices`.
-        if tried_client_indices.len() < self.max_tries
-            && tried_client_indices.insert(state.current_index)
-        {
-            return Ok(state.client.clone());
-        }
+        if let Some(state) = state.as_mut() {
```
wbbradley:
```suggestion
        if let Some(state) = state.as_ref() {
```
No need for `as_mut` here as far as I can tell?

wbbradley: Could also coalesce this `if let` with the next statement (assignment of `next_index`).

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -148,29 +159,48 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
...
         // let that stop us from trying the next one.
         loop {
             // PLAN phase.
             if tried_client_indices.len() >= self.max_tries {
-                return Err(FailoverError::FailedToGetClient(
-                    "max failovers exceeded".to_string(),
-                ));
+                return Err(FailoverError::FailedToGetClient(format!(
+                    "max failovers {} exceeded",
+                    self.max_tries
```
wbbradley:
```suggestion
                    "max failovers exceeded [max_tries={}]",
                    self.max_tries
```
nit: this will make it a hair easier to find in logs

crates/walrus-sui/src/client/retry_client/failover.rs
wbbradley: I agree that I don't think this change should live at this layer. I think the retry mechanics (via `retry_rpc_errors`) handle retry against a single url, and the failover mechanics are for when the retries are exhausted. We shouldn't leak retry logic into this layer until we are actually ready to handle all retry logic at this layer.

crates/walrus-sui/src/client/retry_client/failover.rs
```diff
@@ -281,7 +338,7 @@ impl<ClientT, BuilderT: LazyClientBuilder<ClientT> + std::fmt::Debug>
                                 last_error = ?error,
                                 failed_rpc_url,
                                 next_rpc_url,
-                                "Failed to execute operation on client, retrying with next client"
+                                "failed to execute operation on client, retrying with next client"
```
wbbradley, line 339: I wonder if there is a lint for this we can use.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -525,7 +561,7 @@ impl RetriableClientError {
     /// The time window during which failures are counted.
     const FAILURE_WINDOW: Duration = Duration::from_secs(300);
     /// The maximum number of failures allowed.
-    const MAX_FAILURES: usize = 100;
+    const MAX_FAILURES: usize = 20;
```
wbbradley: From above, it's unclear why we are moving retry logic into the failover logic.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -603,3 +639,13 @@ impl From<tonic::Status> for CheckpointRpcError {
...
     }
 }
+
+impl CheckpointRpcError {
+    pub(crate) fn is_checkpoint_not_produced(&self) -> bool {
+        self.status.code() == tonic::Code::NotFound
+            && self
+                .status
+                .checkpoint_height()
+                .is_some_and(|height| self.checkpoint_seq_num.unwrap_or(0) > height)
```
wbbradley, line 661: Should this be `>=`? Does `self.checkpoint_seq_num` represent the last loaded checkpoint, or the next checkpoint? If it's already loaded `self.checkpoint_seq_num`, then we should change `>` to `>=`, I think?

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -134,9 +134,36 @@ impl LazySuiClientBuilder {
 impl LazyClientBuilder<SuiClient> for LazySuiClientBuilder {
     // TODO: WAL-796 Out of concern for consistency, we are disabling the failover mechanism for
     // SuiClient for now.
-    const DEFAULT_MAX_TRIES: usize = 1;
+    const DEFAULT_MAX_TRIES: usize = 5;
```
wbbradley, line 137: I'm not sure we should bundle all these changes with the actual enabling of the SuiClient's lazy failover, especially since we are not addressing the previously identified consistency issues.

wbbradley: Maybe you could turn it on in `#[cfg(msim)]` though?

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
halfprice: `self.checkpoint_seq_num` is the checkpoint that the walrus node tries to read.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
halfprice: As I mentioned in slack, without this change, any one RPC node creation failure would prevent the `RetriableSuiClient` to fail initialize, also we should just enable Sui client failover as well.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
halfprice: Reverted this change. Changes related to this configure change is now handled in #2090

crates/walrus-sui/src/client/retry_client/failover.rs
halfprice: done

crates/walrus-sui/src/client/retry_client/failover.rs
halfprice: As discussed offline, I removed this filter and added a comment and a TODO about for any future noticed application error that we shouldn't failover, we need to add a filter here.

crates/walrus-sui/src/client/retry_client/failover.rs
halfprice: Done.

crates/walrus-sui/src/client/retry_client/failover.rs
halfprice: Done. Good catch!

halfprice:
Hi @wbbradley , I'm going to merge this so that I can integrate this PR into #2090 . If you have further comments, I'll address them in a follow up PR.

PR #2086, opened 2025-05-13
chore: retry connection establishment in FailoverWrapper::new

wbbradley:
## Description

Retry connection establishment in FailoverWrapper::new.

## Test plan

CI Pipeline.


wbbradley:
It looks like a more lazy approach is being sought out in #2083. I'll close this out.

PR #2087, opened 2025-05-13
ci: automate release creation

ebmifa:
## Description
automate Walrus release creation

## Test plan
Will test after this lands

### DIFF
--- a/.github/workflows/create-release-announce.yml
+++ b/.github/workflows/create-release-announce.yml
@@ -12,7 +12,7 @@ on:
         required: true
 
 env:
-  RELEASE_NOTES_FILE: "./release_notes.txt"        
+  RELEASE_NOTES_FILE: "./release_notes.txt"
 
 jobs:
   generate-tag-release:
@@ -58,7 +58,7 @@ jobs:
         run: |
           export walrus_commit=$(git rev-parse HEAD)
           echo "walrus_commit=${walrus_commit}" >> $GITHUB_ENV
-          git push --force origin ${walrus_commit}:${{ inputs.network }} 
+          git push --force origin ${walrus_commit}:${{ inputs.network }}
 
       - name: Create a release tag
         shell: bash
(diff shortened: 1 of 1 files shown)


johnjmartin (approved):

.github/workflows/create-release-announce.yml
```diff
@@ -0,0 +1,135 @@
...
+          fetch-depth: 0
+          ref: main
+
+      - name: Get current and previous release branches
+        shell: bash
+        run: |
+          export walrus_main_version=$(cat Cargo.toml | grep '^version =' | tr -d '\"' | awk '{ print $3 }')"
+          export major_version=$(echo $sui_main_version | cut -d. -f1)
+          export minor_version=$(echo $sui_main_version | cut -d. -f2)
+          export patch_version=$(echo $sui_main_version | cut -d. -f3)
```
mlegner, line 35: Question: Is there ever a reason to have something different from `0` here? If so, do we expect that to be also in the release branch?

.github/workflows/create-release-announce.yml
```diff
@@ -0,0 +1,135 @@
...
+        with:
+          fetch-depth: 0
+          ref: ${{ env.RELEASE_BRANCH }}
+
+      - name: Update ${{ inputs.network }} branch
+        shell: bash
+        run: |
+          export walrus_commit=$(git rev-parse HEAD)
+          echo "walrus_commit=${walrus_commit}" >> $GITHUB_ENV
+          git push --force origin ${walrus_commit}:${{ inputs.network }}
```
mlegner, line 61: Question: Do we need any special permissions for this?

mlegner (commented):
Thanks a lot, that's amazing! I just had two questions while looking through this.

.github/workflows/create-release-announce.yml
ebmifa: We don't, but this is just part of consistency.

.github/workflows/create-release-announce.yml
ebmifa: It's in the same repo and should be using GITHUB_TOKEN.

PR #2088, opened 2025-05-13
ci: walrus release create and announce work

ebmifa:
## Description
New walrus workflow to create and announce releases

## Test plan
https://github.com/MystenLabs/walrus/actions/runs/15028490490/job/42235064792


.github/workflows/create-release-announce.yml
```diff
@@ -8,16 +8,26 @@ on:
     inputs:
       network:
         description: Network to generate release for
-        type: string
+        required: true
+        type: choice
+        options:
+          - testnet
```
mlegner, line 14: Question: I assume `devnet` is different/manual, correct?

.github/workflows/create-release-announce.yml
```diff
@@ -133,3 +148,17 @@ jobs:
           body_path: ${{ env.RELEASE_NOTES_FILE }}
           draft: false
           prerelease: ${{ env.PRE_RELEASE }}
+
```
mlegner, line 151: Question: AFAIU, the `Create Release` step creates the GH release, which then triggers the `attach-binaries-to-release.yml` workflow. Can we make sure that only happens when the binaries/archives are already built as a result of the `trigger-artificts-builds.yml` workflow to be able to reuse these archives?

mlegner: Also: Could we mark the release here as `draft: true` and then set it to `false` in `attach-binaries-to-release.yml` after all binaries are attached?

.github/workflows/create-release-announce.yml
```diff
@@ -133,3 +148,17 @@ jobs:
...
+  announce-release:
+    name: Announce Release in Discord
+    if: ${{ inputs.announce == true }}
+    needs: create-release
```
mlegner, line 155: Question: Can we somehow make sure that this is only triggered after the binaries have been pushed to GCP and attached to the GH release and Docker images have been published?

mlegner (commented):
Thanks a lot for making this work, @ebmifa! I have a few questions about interdependencies of different steps. Let me know if they are already covered or whether it would be possible to implement them.

.github/workflows/create-release-announce.yml
ebmifa: Yes, we don't announce `devnet` externally

.github/workflows/create-release-announce.yml
ebmifa: > Can we make sure that only happens when the binaries/archives are already built as a result of the trigger-artificts-builds.yml

It should have been already called when the `testnet` and `mainnet` branches were updated.

> Also: Could we mark the release here as draft: true and then set it to false in attach-binaries-to-release.yml after all binaries are attached?

I don't think we need this, since it won't really gain us much

.github/workflows/create-release-announce.yml
ebmifa: Again, all of the artifacts should have already been built prior to this. Remember that we are basically building the same `commit hash` that should have already generated all the binaries we need in GCP, so when `testnet` and `mainnet` branches get updated, we are just renaming binaries and tar balls.

.github/workflows/create-release-announce.yml
mlegner: > It should have been already called when the `testnet` and `mainnet` branches were updated.

Yes, but it might still be running, right? So when the `attach-binaries-to-release.yml` workflow checks for existing archives, it may not find them yet.

ebmifa: See my comment below:
>Again, all of the artifacts should have already been built prior to this. Remember that we are basically building the same commit hash that should have already generated all the binaries we need in GCP, so when testnet and mainnet branches get updated, we are just renaming binaries and tar balls.

.github/workflows/create-release-announce.yml
mlegner: Maybe I'm missing something; I don't see where we actually wait for the builds to finish.

mlegner: Ah, I think now I see your point. We don't need to wait because it's anyway super-fast.

ebmifa: Yes, the 1st PTN release should have built all of this with the first deployment. If we have cherry picks, then we'd push them to `testnet` first which would create all the binaries and then they can be re-used going forward. So all of this should be lighting fast.

mlegner: > Yes, the 1st PTN release should have built all of this with the first deployment. If we have cherry picks, then we'd push them to `testnet` first which would create all the binaries and then they can be re-used going forward. So all of this should be lighting fast.

OK, I understand. But then we need to make sure to always build binaries for patch releases before doing the Testnet announcement (e.g., by bumping `devnet` first); otherwise the builds will only be triggered through this workflow.

ebmifa: Keep in mind that we always update `releases/walrus-v1.25.0-release` branch first before going to `testnet` and `mainnet` branches. And wait for the [release CI](https://github.com/MystenLabs/sui-operations/actions/workflows/walrus-staging-continuous.yaml) to pass before moving forward. So everything should have already been built.

mlegner: Thanks for the offline chat. The missing piece seems to be the artifact build being triggered with pushes to the release branches. If that is in place, then everything should work as described. 🙂

mlegner (approved):
Thanks @ebmifa! The only missing piece seems to be the artifact build being triggered by new commits to the release branches.

PR #2090, opened 2025-05-14
fix: skip fetching checkpoint from rpc nodes when encounter extended primary failures

halfprice:
## Description

When an RetriableRpcClient encounters failures for extended period of time when fetching checkpoints from RPC nodes, the node should just skip RPC for short period of time, and directly fetch checkpoints from archival. Otherwise, the RPC request + retry + failure takes 40s to 60s, which significantly limits the checkpoint downloading throughput.

## Test plan

Tested archival catchup in PTN
Screenshot 2025-05-14 at 1 17 00 AM

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/checkpoint-downloader/src/downloader.rs
+++ b/crates/checkpoint-downloader/src/downloader.rs
@@ -432,12 +432,22 @@ impl ParallelCheckpointDownloaderInner {
         config: &ParallelDownloaderConfig,
         rng: &mut StdRng,
     ) -> CheckpointEntry {
-        let mut backoff = ExponentialBackoff::new_with_seed(
-            config.initial_delay,
-            config.max_delay,
-            config.retries,
-            rng.next_u64(),
-        );
+        let mut backoff = if cfg!(test) {
+            ExponentialBackoff::new_with_seed(
+                config.initial_delay,
+                config.max_delay,
+                config.retries,
+                rng.next_u64(),
+            )
+        } else {
+            ExponentialBackoff::new_with_seed(
+                config.initial_delay,
+                config.max_delay,
+                None,
+                rng.next_u64(),
+            )
+        };
+
         loop {
             let result = client.get_full_checkpoint(sequence_number).await;
             let Ok(checkpoint) = result else {
@@ -449,6 +459,8 @@ impl ParallelCheckpointDownloaderInner {
                     continue;
                 }
 
+                // Note that we only return error in test mode.
+                assert!(cfg!(test));
                 return CheckpointEntry::new(sequence_number, Err(err.into()));
             };
 
--- a/crates/checkpoint-downloader/src/downloader.rs
+++ b/crates/checkpoint-downloader/src/downloader.rs
@@ -9,16 +9,14 @@ use std::{
 };
 
 use anyhow::{Result, anyhow};
-use rand::{SeedableRng, rngs::StdRng};
+use rand::{RngCore, SeedableRng, rngs::StdRng};
 use sui_rpc_api::client::ResponseExt;
 use sui_types::messages_checkpoint::{CheckpointSequenceNumber, TrustedCheckpoint};
 use tokio::{sync::mpsc, time::Instant};
 use tokio_util::sync::CancellationToken;
 use typed_store::{Map, rocks::DBMap};
 use walrus_sui::client::retry_client::{RetriableClientError, RetriableRpcClient};
-#[cfg(not(test))]
-use walrus_utils::backoff::ExponentialBackoff;
-use walrus_utils::{metrics::Registry, tracing_sampled};
+use walrus_utils::{backoff::ExponentialBackoff, metrics::Registry, tracing_sampled};
 
 use crate::{
     ParallelDownloaderConfig,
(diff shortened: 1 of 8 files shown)


crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -265,7 +263,6 @@ impl ParallelCheckpointDownloaderInner {
         let mut last_scale = Instant::now();
         let mut consecutive_failures = 0;
         let average_workers = (downloader_config.min_workers + downloader_config.max_workers) / 2;
-        const MAX_CONSECUTIVE_POOL_MONITOR_FAILURES: u32 = 100;
```
halfprice, line 270: @sadhansood , how was this count originally determined? In PTN, it took nodes more than an hour to reach to 100 failures to check checkpoint lag.

I convert it into a config setting (and a few other parameters), so that we can easily change it.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -554,16 +573,9 @@ impl RetriableClientError {
...
             }
-            _ => self.is_failure_window_exceeded(last_success, num_failures),
+            _ => false,
         }
     }
-
-    /// Returns `true` if the failure window has been exceeded. Failure window is exceeded if
-    /// the number of failures exceeds `MAX_FAILURES` and if the last successful RPC call was
-    /// more than `FAILURE_WINDOW` minutes ago.
-    fn is_failure_window_exceeded(&self, last_success: Instant, num_failures: usize) -> bool {
```
halfprice, line 612: Note: this part is now moved to FallbackClient so that it can be checked against config.

crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -426,75 +424,42 @@ impl ParallelCheckpointDownloaderInner {
...
 
-                let delay = backoff.next().expect("backoff should not be exhausted");
+                if let Some(delay) = backoff.next() {
+                    tokio::time::sleep(delay).await;
+                    continue;
+                }
 
-                tokio::time::sleep(delay).await;
-                continue;
+                return CheckpointEntry::new(sequence_number, Err(err.into()));
```
sadhansood: I don't think this logic should be changed to return an error as that would cause an error to be returned to the client (event processor in this case). The config `min_retries` should be removed imo since it is not being used.

crates/checkpoint-downloader/src/downloader.rs
sadhansood: I did not want it to start adjusting workers immediately after full nodes go down and ~ 1 hour seems reasonable

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -293,58 +301,80 @@ impl RetriableRpcClient {
...
-        if !error.is_eligible_for_fallback(
-            sequence_number,
-            self.last_success.load(Ordering::Relaxed),
-            self.num_failures.load(Ordering::Relaxed),
-        ) {
-            tracing::debug!(
-                "primary client error while fetching checkpoint is not eligible for fallback"
+            // RPC node failure is sustained, so we skip it for 5 minutes.
+            // Also, at this point, the fallback client must be set.
+            self.skip_rpc_for_checkpoint_until.store(
```
sadhansood, line 422: Does this also get set for something like missing events error? Since it is in the check for `is_elligible_for_fallback`

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -293,58 +301,80 @@ impl RetriableRpcClient {
...
         let _scope = monitored_scope("RetriableRpcClient::get_full_checkpoint");
         let start_time = Instant::now();
-        let error = match self.get_full_checkpoint_from_primary(sequence_number).await {
-            Ok(checkpoint) => {
+
+        // Check if we should directly skip RPC node due to previous failures, and use fallback.
+        let try_rpc_node_first = self.fallback_client.is_none()
+            || start_time >= self.skip_rpc_for_checkpoint_until.load(Ordering::Relaxed);
+
+        if try_rpc_node_first {
```
sadhansood, line 370: It'll be nice to refactor some of the logic in a separate function, it seems like this function is getting bigger and harder to follow. Also, could we update the docstring of this function?

sadhansood (commented):
Thanks @halfprice , the PR looks solid, I just have a few comments for your consideration.

crates/checkpoint-downloader/src/downloader.rs
halfprice: The code used to have two `download_with_retry` function, which was confusing sometimes. The change here just merged these two together, so that we don't have one that does retry and one don't.

I made some changes in the code to indicate that the error returned here is only possible in tests, which just to make `test_parallel_fetcher` work. PTAL.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
halfprice: Great catch! We should only skip calling RPC node when `is_failure_window_exceeded`. (one thing I don't feel to comfortable about this PR is that, for realistic scenario like this, it's very hard to test/discover such corner cases).

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
halfprice: I made a pass, and hopefully it's cleaner now. PTAL.

crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -426,75 +424,54 @@ impl ParallelCheckpointDownloaderInner {
...
     /// Downloads a checkpoint with retries.
-    #[cfg(not(test))]
     async fn download_with_retry(
         client: &RetriableRpcClient,
         sequence_number: CheckpointSequenceNumber,
         config: &ParallelDownloaderConfig,
         rng: &mut StdRng,
     ) -> CheckpointEntry {
-        let mut backoff = create_backoff(rng, config);
+        let mut backoff = if cfg!(test) {
```
sadhansood, line 435: nit: Maybe add a comment about the different behavior b/w prod and test for future. it's a small thing but I think we might scratch our heads one day later and wonder why this if/else branch is needed.

sadhansood (approved):
Thanks @halfprice, this PR is going to be really valuable during fallback scenarios!

crates/checkpoint-downloader/src/downloader.rs
halfprice: Done.

PR #2092, opened 2025-05-14
fix: Handle event processor bootstrap when latest checkpoint is unavailable

sadhansood:
## Description

This PR handles the scenario when `get_latest_checkpoint_summary` is unavailable because all full node endpoints are down. But `get_checkpoint_summary` still works because archival is working. We fallback to catchup using event blobs in such a scenario.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/events/event_processor.rs
+++ b/crates/walrus-service/src/node/events/event_processor.rs
@@ -245,6 +245,7 @@ pub struct EventProcessorRuntimeConfig {
 }
 
 /// Struct to group client-related parameters.
+#[derive(Clone)]
 pub struct SuiClientSet {
     /// Sui client.
     pub sui_client: RetriableSuiClient,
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -652,18 +652,36 @@ impl EventProcessor {
...
-                    localnet, testnet or devnet network."
+        let latest_checkpoint = match retry_client.get_latest_checkpoint_summary().await {
+            Ok(summary) => Some(summary),
+            Err(e) => {
+                tracing::warn!(
+                    error = ?e,
+                    "Failed to get latest checkpoint summary, proceeding without lag check"
+                );
+                None
+            }
```
liquid-helium: I want to make sure I understand the changes here: at a high level, when a storage node cannot get the latest_checkpoint_summary from full nodes, it will try to `catch up` from event blobs (but it could be that it is already up-to-date, only the fullnodes are down).
After the catchup phase, it will continue the work:
try fetch checkpoint from the full nodes, then fails, and eventually fallback to the archive?

sadhansood: yes, although I was just chatting with @halfprice and i realized that sui client and rpc client both now use mysten-rpc endpoint. So that basically means, that `catchup_up` is also going to fail if mysten-rpc is down. I am going to change it to directly fallback to downloadign checkpoints from rpc and fallback to archive.

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -652,18 +652,36 @@ impl EventProcessor {
...
 
-        let latest_checkpoint = retry_client.get_latest_checkpoint_summary().await?;
-        if current_checkpoint > latest_checkpoint.sequence_number {
-            tracing::error!(
-                current_checkpoint,
-                ?latest_checkpoint,
-                "Current store has a checkpoint that is greater than latest network checkpoint! \
-                    This is especially likely when a node is restarted running against a newer \
-                    localnet, testnet or devnet network."
+        let latest_checkpoint = match retry_client.get_latest_checkpoint_summary().await {
```
halfprice: Nit: this function is getting very long. Can you put `fetching lag + catch up` into a separate function?

halfprice (approved):
Thanks @sadhansood for jumping quickly on this! LGTM

crates/walrus-service/src/node/events/event_processor.rs
sadhansood: done

PR #2094, opened 2025-05-15
[docs] Update http api example

dantheman8300:
## Description

Update http api example to include usage of both `deletable` and `send_object_to` flags since most situations will require both. 

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/docs/book/usage/web-api.md
+++ b/docs/book/usage/web-api.md
@@ -32,7 +32,7 @@ You can interact with the daemon through simple HTTP PUT requests. For example, 
 curl -X PUT "$PUBLISHER/v1/blobs" -d "some string" # store the string `some string` for 1 storage epoch
 curl -X PUT "$PUBLISHER/v1/blobs?epochs=5" --upload-file "some/file" # store file `some/file` for 5 storage epochs
 curl -X PUT "$PUBLISHER/v1/blobs?send_object_to=$ADDRESS" --upload-file "some/file" # store file `some/file` and send the blob object to $ADDRESS
-curl -X PUT "$PUBLISHER/v1/blobs?deletable=true" --upload-file "some/file" # store file `some/file` as a deletable blob, instead of a permanent one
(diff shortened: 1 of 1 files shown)


dantheman8300:
I'm getting the lint error for line length, but I wonder if it should be ignored since it's in the code block? Let me know what you think is best for readability and style.

docs/book/usage/web-api.md
```diff
@@ -32,7 +32,7 @@ You can interact with the daemon through simple HTTP PUT requests. For example,
 curl -X PUT "$PUBLISHER/v1/blobs" -d "some string" # store the string `some string` for 1 storage epoch
 curl -X PUT "$PUBLISHER/v1/blobs?epochs=5" --upload-file "some/file" # store file `some/file` for 5 storage epochs
 curl -X PUT "$PUBLISHER/v1/blobs?send_object_to=$ADDRESS" --upload-file "some/file" # store file `some/file` and send the blob object to $ADDRESS
-curl -X PUT "$PUBLISHER/v1/blobs?deletable=true" --upload-file "some/file" # store file `some/file` as a deletable blob, instead of a permanent one
+curl -X PUT "$PUBLISHER/v1/blobs?deletable=true&send_object_to=$ADDRESS" --upload-file "some/file" # store file `some/file` as a deletable blob, instead of a permanent one and send the blob object to $ADDRESS
```
mlegner: Suggestion to fix the line length:
```suggestion
# store file `some/file` as a deletable blob, instead of a permanent one and send the blob object to $ADDRESS
curl -X PUT "$PUBLISHER/v1/blobs?deletable=true&send_object_to=$ADDRESS" --upload-file "some/file"
```
The ones above are fine, but we could also make the same change to be consistent (no strong opinion either way).

mlegner (approved):
Thanks for the change, @dantheman8300!

> I'm getting the lint error for line length, but I wonder if it should be ignored since it's in the code block? Let me know what you think is best for readability and style.

We already have an increased line limit for code blocks, see [here](https://github.com/MystenLabs/walrus/blob/831fc57c5ca7c8cf84ece2ebc2fa7f96eb88abbd/.markdownlint-cli2.yaml#L16). I think we should keep the limit and try to keep lines to reasonable lengths there.

PR #2100, opened 2025-05-19
chore: upgrade subsidies testnet contracts

karlwuest:
## Description

Updates `testnet-contracts/subsidies` to the newest deployed version.

## Test plan

Manually tested after deployment.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/testnet-contracts/subsidies/Move.lock
+++ b/testnet-contracts/subsidies/Move.lock
@@ -40,7 +40,7 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.45.2"
+compiler-version = "1.48.2"
 edition = "2024.beta"
 flavor = "sui"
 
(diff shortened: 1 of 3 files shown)


testnet-contracts/subsidies/Move.lock
mlegner, line 1: Question: Is it correct that you were still using `testnet-v1.45.2` for the dependencies (`move-stdlib` and `sui-framework`)?

mlegner (approved):
Thanks, @karlwuest!

testnet-contracts/subsidies/Move.lock
karlwuest: Ah, I should have updated this as well, but yes, I did not update the versions in the `Move.toml` before upgrading. This doesn't really matter for the deployment, though, since it only affects the interfaces that are available (the on-chain version always uses the latest version of `move-stdlib` and `sui-framework`).

karlwuest: Also, updating this would have required updating the dependencies in `walrus` as well, since otherwise the move compiler complains. So it's probably cleaner to keep it the same.

PR #2105, opened 2025-05-20
feat: Quilt encoding part ii

liquid-helium:
## Description

Added Iterator interface for QuiltV1 and QuiltDecoder, so that they can share they same decoding logic.

Added explicit blob header for each blob in a quilt, as well as identifier.

Contributes to WAL-858

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/encoding/quilt_encoding.rs
+++ b/crates/walrus-core/src/encoding/quilt_encoding.rs
@@ -84,6 +84,8 @@ pub trait QuiltApi<V: QuiltVersion> {
     ) -> Result<V::Quilt, QuiltError>;
 
     /// Gets a blob by its identifier from the quilt.
+    ///
+    /// TODO(WAL-862): Deduplicate the `get_blob*` functions.
     fn get_blob_by_identifier(&self, identifier: &str) -> Result<QuiltStoreBlobOwned, QuiltError>;
 
     /// Returns the quilt index.
@@ -343,8 +345,7 @@ impl QuiltVersionV1 {
         assert!(index_bytes.len() == index_size);
 
         // Decode the QuiltIndexV1.
-        let mut quilt_index: QuiltIndexV1 = bcs::from_bytes(&index_bytes)
-            .map_err(|e| QuiltError::QuiltIndexSerDerError(e.to_string()))?;
+        let mut quilt_index: QuiltIndexV1 = bcs::from_bytes(&index_bytes)?;
 
         let total_size = index_size + QUILT_INDEX_PREFIX_SIZE;
         let columns_needed = total_size.div_ceil(column_size);
@@ -592,7 +593,7 @@ impl QuiltApi<QuiltVersionV1> for QuiltV1 {
             symbol_size,
             quilt_index: None,
         };
-        quilt.quilt_index = Some(quilt.get_or_decode_quilt_index()?);
+        let _quilt_index = quilt.get_or_decode_quilt_index()?;
 
         Ok(quilt)
     }
@@ -678,9 +679,23 @@ impl QuiltColumnRangeReader for QuiltV1 {
 }
 
 impl QuiltV1 {
-    fn get_or_decode_quilt_index(&mut self) -> Result<QuiltIndexV1, QuiltError> {
+    /// Returns the quilt index.
+    ///
+    /// If the quilt index is not set, it will decode it from the quilt data, and set it.
+    fn get_or_decode_quilt_index(&mut self) -> Result<&QuiltIndexV1, QuiltError> {
+        if self.quilt_index.is_some() {
+            return Ok(self
+                .quilt_index
+                .as_ref()
+                .expect("quilt index should be set"));
+        }
+
         let columns_size = self.data.len() / self.row_size * self.symbol_size;
-        QuiltVersionV1::decode_quilt_index(self, columns_size)
+        self.quilt_index = Some(QuiltVersionV1::decode_quilt_index(self, columns_size)?);
+        Ok(self
+            .quilt_index
+            .as_ref()
+            .expect("quilt index should be set"))
     }
 }
 
@@ -1000,13 +1015,7 @@ impl QuiltEncoderApi<QuiltVersionV1> for QuiltEncoderV1<'_> {
         let mut quilt_index = QuiltIndexV1 { quilt_patches };
 
         // Get the serialized quilt index size.
-        let serialized_index_size =
-            u32::try_from(bcs::serialized_size(&quilt_index).map_err(|e| {
-                QuiltError::QuiltIndexSerDerError(format!(
-                    "failed to serialize quilt index: {:?}",
-                    e
-                ))
-            })?)
+        let serialized_index_size = u32::try_from(bcs::serialized_size(&quilt_index)?)
             .expect("serialized_index_size should fit in u32");
 
         // Calculate total size including the size prefix and the quilt type.
@@ -1133,6 +1142,7 @@ impl QuiltEncoderApi<QuiltVersionV1> for QuiltEncoderV1<'_> {
 pub struct QuiltDecoderV1<'a> {
     slivers: HashMap<SliverIndex, &'a SliverData<Secondary>>,
     quilt_index: Option<QuiltIndexV1>,
+    column_size: Option<usize>,
 }
 
 impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
@@ -1141,14 +1151,7 @@ impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
             return Ok(self.quilt_index.as_ref().expect("quilt index should exist"));
         }
         self.check_missing_slivers(0, 1)?;
-        let column_size = self
-            .slivers
-            .get(&SliverIndex(0))
-            .expect("first sliver should exist")
-            .symbols
-            .data()
-            .len();
-
+        let column_size = self.column_size.expect("column size should be set");
         self.quilt_index = Some(QuiltVersionV1::decode_quilt_index(self, column_size)?);
 
         Ok(self
@@ -1172,6 +1175,8 @@ impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
 
     fn add_slivers(&mut self, slivers: &'a [&'a SliverData<Secondary>]) {
         for sliver in slivers {
+            self.column_size
+                .get_or_insert_with(|| sliver.symbols.data().len());
             self.slivers.insert(sliver.index, sliver);
         }
     }
@@ -1186,13 +1191,7 @@ impl QuiltColumnRangeReader for QuiltDecoderV1<'_> {
         mut bytes_to_return: usize,
     ) -> Result<Vec<u8>, QuiltError> {
         self.check_missing_slivers(start_col, start_col + 1)?;
-        let column_size = self
-            .slivers
-            .get(&SliverIndex(start_col as u16))
-            .expect("sliver should exist")
-            .symbols
-            .data()
-            .len();
+        let column_size = self.column_size.expect("column size should be set");
         let end_col = start_col + (bytes_to_skip + bytes_to_return).div_ceil(column_size);
         self.check_missing_slivers(start_col, end_col)?;
 
@@ -1237,12 +1236,14 @@ impl QuiltColumnRangeReader for QuiltDecoderV1<'_> {
 impl<'a> QuiltDecoderV1<'a> {
     /// Creates a new QuiltDecoderV1 with the given slivers.
     pub fn new(slivers: &'a [&'a SliverData<Secondary>]) -> Self {
+        let column_size = slivers.first().map(|s| s.symbols.data().len());
         Self {
             slivers: slivers
                 .iter()
                 .map(|s| (s.index, *s))
                 .collect::<HashMap<_, _>>(),
             quilt_index: None,
+            column_size,
         }
     }
 
@@ -1251,9 +1252,11 @@ impl<'a> QuiltDecoderV1<'a> {
         slivers: &'a [&'a SliverData<Secondary>],
         quilt_index: QuiltIndexV1,
     ) -> Self {
+        let column_size = slivers.first().map(|s| s.symbols.data().len());
         Self {
             slivers: slivers.iter().map(|s| (s.index, *s)).collect(),
             quilt_index: Some(quilt_index),
+            column_size,
         }
     }
 
--- a/crates/walrus-core/src/encoding/errors.rs
+++ b/crates/walrus-core/src/encoding/errors.rs
@@ -166,7 +166,7 @@ pub enum QuiltError {
     FailedToExtractQuiltIndexSize,
     /// Failed to decode the quilt index.
     #[error("failed to decode the quilt index: {0}")]
-    QuiltIndexSerDerError(String),
+    QuiltIndexSerDerError(#[from] bcs::Error),
     /// Missing slivers.
     #[error("missing slivers: {0:?}")]
     MissingSlivers(Vec<SliverIndex>),


crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -516,13 +1122,105 @@ pub struct QuiltEncoderV1<'a> {
...
+            result_bytes.append(&mut inner_extension_vec);
+        }
+
+        Ok(result_bytes)
+    }
+
+    /// Adds a blob to the quilt as consecutive columns.
+    ///
+    /// The blob data layout is as follows:
+    // +------------------+-----------------------------+----------------------+------------------+
```
liquid-helium: change to doc string.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -32,19 +34,21 @@ use crate::{
...
-const QUILT_INDEX_SIZE_PREFIX_SIZE: usize = 4;
+const QUILT_INDEX_SIZE_BYTES_LENGTH: usize = 4;
+
+/// The number of bytes used to store the version of the quilt.
+const QUILT_VERSION_BYTES_LENGTH: usize = 1;
 
-/// The number of bytes used to store the type of the quilt.
-const QUILT_TYPE_SIZE: usize = 1;
+/// The number of bytes used to store the identifier of the blob.
+const BLOB_IDENTIFIER_BYTES_LENGTH: usize = 2;
```
mlegner: Question: Is this the length of the identifier or the length for length of the identifier?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -161,9 +373,23 @@ impl TryFrom<u8> for QuiltVersionEnum {
...
+impl fmt::Display for QuiltVersionEnum {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{:?}", self)
+    }
+}
+
+impl From<String> for QuiltVersionEnum {
+    fn from(value: String) -> Self {
+        match value.as_str() {
+            "V1" | "v1" | "1" => QuiltVersionEnum::V1,
```
mlegner, line 191: Remark: I wonder if it will lead to confusion that `V1` is encoded as `0`. Also, converting a number `1` to the version, would result in a future `V2` while the string `"1"` is parsed as `V1`. 🤔

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -127,14 +126,227 @@ pub trait QuiltDecoderApi<'a, V: QuiltVersion> {
...
+    row_size: usize,
+    symbol_size: usize,
+    n_rows: usize,
+
+    // Track the iter position.
+    current_col: usize,
+    end_col: usize,
+    current_row: usize,
+    bytes_to_skip: usize,
+    bytes_remaining: usize,
```
mlegner: Suggestion: Could you describe what these fields mean exactly? Especially the `bytes_to_skip` (also called `begin_position` or `start_byte` in other places) and `bytes_remaining` (also called `limit` or `limit_bytes` in other places).

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -265,6 +505,130 @@ pub struct QuiltVersionV1;
...
+
+        let mut header_bytes = Vec::new();
+        {
+            let iter =
+                data_source.iter_column_bytes(start_col, end_col, 0, Self::BLOB_HEADER_SIZE)?;
+            for symbol in iter {
+                header_bytes.extend_from_slice(symbol);
+            }
+        }
+        assert!(header_bytes.len() == Self::BLOB_HEADER_SIZE);
```
mlegner: Suggestion: We have almost exactly the same code in multiple places. How about we add a function to the `QuiltColumnDataIterator` trait?
```rust
    /// Returns a vector of bytes in the columns identified by the given start and end indices.
    ///
    /// Assuming a 0-indexed bytes, the vector will contain the bytes in the range
    /// `[start_byte, start_byte + limit_bytes)`.
    fn collect_to_vec(
        &self,
        start_col: usize,
        end_col: usize,
        start_byte: usize,
        limit_bytes: usize,
    ) -> Result<Vec<u8>, QuiltError> {
        let mut result = Vec::with_capacity(limit_bytes - start_byte);
        for symbol in self.iter_column_bytes(start_col, end_col, start_byte, limit_bytes)? {
            result.extend_from_slice(symbol);
        }
        Ok(result)
    }
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -127,14 +126,227 @@ pub trait QuiltDecoderApi<'a, V: QuiltVersion> {
...
+pub trait QuiltColumnDataIterator {
+    type RangeIter<'a>: Iterator<Item = &'a [u8]> + 'a
+    where
+        Self: 'a;
+
+    /// Returns an iterator over bytes in the columns identified by the given start and end indices.
+    ///
+    /// Assuming a 0-indexed bytes, the iterator will return the bytes in the range
+    /// `[start_byte, start_byte + limit_bytes)`.
+    fn iter_column_bytes(
```
mlegner: Question: Do we ever use this for anything besides combining it into an output vector? If not, we could probably remove and/or simplify some of the code. See also my comment about a `collect_to_vec` function above.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -265,6 +505,130 @@ pub struct QuiltVersionV1;
...
+        })?;
+
+        extension_size += identifier_size as usize + BLOB_IDENTIFIER_BYTES_LENGTH;
+
+        let total_size = extension_size + blob.data().len() + QuiltVersionV1::BLOB_HEADER_SIZE;
+        Ok(total_size)
+    }
+
+    /// Decodes a blob from a column data source.
+    ///
```
mlegner: Nit: Is there any additional info missing or can this line be removed?
```suggestion
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -265,6 +505,130 @@ pub struct QuiltVersionV1;
 
 impl QuiltVersionV1 {
     const QUILT_VERSION_BYTE: u8 = 0x00;
+    const BLOB_HEADER_SIZE: usize = 7;
+    const EXTENSION_SIZE_BYTES_LENGTH: usize = 4;
+
+    /// Returns the total size of the encoded blob.
+    pub fn compute_encoded_blob_size(blob: &QuiltStoreBlob) -> Result<usize, QuiltError> {
```
mlegner: Remark: The naming is a bit confusing. Can we call this `serialized_blob_size` or something to avoid the confusion with blob encoding?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -265,6 +505,130 @@ pub struct QuiltVersionV1;
...
+                "Identifier size exceeds maximum allowed value: {}",
+                e
+            ))
+        })?;
+
+        extension_size += identifier_size as usize + BLOB_IDENTIFIER_BYTES_LENGTH;
+
+        let total_size = extension_size + blob.data().len() + QuiltVersionV1::BLOB_HEADER_SIZE;
+        Ok(total_size)
+    }
```
mlegner: Nit: This feels overly complicated to me. How about the following?
```suggestion
    pub fn compute_encoded_blob_size(blob: &QuiltStoreBlob) -> Result<usize, QuiltError> {
            let identifier_size = bcs::serialized_size(&blob.identifier)
            .map_err(|e| QuiltError::Other(format!("Failed to compute identifier size: {}", e)))?;
        if identifier_size >= MAX_BLOB_IDENTIFIER_BYTES_LENGTH {
            return Err(QuiltError::Other(format!(
                "identifier size exceeds maximum allowed value: {}",
                MAX_BLOB_IDENTIFIER_BYTES_LENGTH
            )));
        }

        Ok(identifier_size
            + BLOB_IDENTIFIER_BYTES_LENGTH
            + blob.data().len()
            + QuiltVersionV1::BLOB_HEADER_SIZE)
```

This requires an additional constant:
```rust
/// The maximum number of bytes for the identifier of the blob.
const MAX_BLOB_IDENTIFIER_BYTES_LENGTH: usize = 1 << (8 * BLOB_IDENTIFIER_BYTES_LENGTH) - 1;
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -265,6 +505,130 @@ pub struct QuiltVersionV1;
 
 impl QuiltVersionV1 {
     const QUILT_VERSION_BYTE: u8 = 0x00;
+    const BLOB_HEADER_SIZE: usize = 7;
+    const EXTENSION_SIZE_BYTES_LENGTH: usize = 4;
```
mlegner: Question: What is this used for?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -280,14 +644,105 @@ impl QuiltVersion for QuiltVersionV1 {
     }
 }
 
+/// The header of a encoded blob in QuiltVersionV1.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct BlobHeaderV1 {
+    pub length: u64, // the lower 34 bits are used for the length.
+    pub mask: u16,   // the lower 14 bits are used for the mask.
+}
```
mlegner: As discussed offline, I feel the choice of using 34 bits for the length causes a lot of additional code complexity with questionable benefits: Any blob above 4GiB should not be stored as part of a quilt anyway. My suggestion would be to just use 32 bits for the length and specify that a quilt can only contain blobs of at most 4GiB. cc @karlwuest

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -503,11 +1052,68 @@ impl<'a> QuiltConfigApi<'a, QuiltVersionV1> for QuiltConfigV1 {
     }
 }
 
+struct MergeIterator<'a> {
```
mlegner: Question: Could you explain the purpose of this additional struct? (Why) do we need this somewhat complicated construction to just write some data to columns?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -516,13 +1122,105 @@ pub struct QuiltEncoderV1<'a> {
...
+        let mut extension_bytes = Vec::new();
+        let mut header = BlobHeaderV1::default();
+
+        let identifier_size =
+            u16::try_from(bcs::serialized_size(&blob.identifier).map_err(|e| {
+                QuiltError::Other(format!("Failed to serialize identifier: {}", e))
+            })?)
+            .map_err(|e| {
+                QuiltError::Other(format!("Failed to convert identifier size to u16: {}", e))
+            })?;
```
mlegner: Nit: We have almost the same code above, can we maybe reuse the `compute_encoded_blob_size` here? Also, can we have a separate error variant for invalid identifiers?

mlegner (commented):
Thanks so much for the changes, @liquid-helium. I made a pass, although I didn't look at everything in full detail yet. A few preliminary questions and suggestions.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice catch, renamed to `BLOB_IDENTIFIER_SIZE_BYTES_LENGTH`.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Ah, great catch, thanks!
I changed the encoding of V1 to `0x01` instead!

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Added comments.
changed to bytes_to_skip, and bytes_to_return.

Initially I was thinking the Api vs implementation, but it caused more confusion :(.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good idea! Change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice observation, removed this trait and replace it with a simpler `QuiltColumnRangeReader`

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Removed, originally I added the schema diagram here.

crates/walrus-core/src/encoding/quilt_encoding.rs

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice suggestion, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: I removed the implementation of the attributes for easier review, this is the leftover, I removed it.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good point, I removed MergeIterator and replaced it with` write_bytes_to_columns()`

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good point. changed to u32.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Added `InvalidIdentifier` error.
Agreed there are logic duplication, the  `serialized_blob_size`(previously `compute_encoded_blob_size`) goes through the same steps but only counts the serialized size (header + extension + blob), and here we get the serialized header + extension, we can dedupe it by serializing the header + extension first (at the current place of calling serialized_blob_size), and later concatenate it with the blob data.
I am keeping it as since the above version comes with a tradeoff on readability, but I don't have a strong preference here, and also open to other solutions.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -127,14 +129,228 @@ pub trait QuiltDecoderApi<'a, V: QuiltVersion> {
...
+        match self.next_n_bytes(self.bytes_to_return) {
+            Some(slice) => {
+                let yield_len = slice.len().min(self.bytes_to_return);
+                self.bytes_to_return -= yield_len;
+                Some(&slice[..yield_len])
+            }
+            None => None,
+        }
+    }
+}
```
mlegner: Question: Are these iterators actually needed?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -264,7 +508,112 @@ impl BlobWithIdentifierOwned {
...
+    const QUILT_VERSION_BYTE: u8 = 0x01;
+    const BLOB_HEADER_SIZE: usize = 6;
+
+    /// Returns the total size of the serialized blob.
+    pub fn serialized_blob_size(blob: &QuiltStoreBlob) -> Result<usize, QuiltError> {
+        let identifier_size = bcs::serialized_size(&blob.identifier)
+            .map_err(|e| QuiltError::Other(format!("Failed to compute identifier size: {}", e)))?;
+
+        if identifier_size >= MAX_BLOB_IDENTIFIER_BYTES_LENGTH {
+            return Err(QuiltError::Other(format!(
```
mlegner: Suggestion: Return an `InvalidIdentifier` error here.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -264,7 +508,112 @@ impl BlobWithIdentifierOwned {
...
+    {
+        let mut blob = QuiltStoreBlobOwned::default();
+
+        let header_bytes =
+            data_source.range_read_from_columns(start_col, end_col, 0, Self::BLOB_HEADER_SIZE)?;
+        assert!(header_bytes.len() == Self::BLOB_HEADER_SIZE);
+        let blob_header = BlobHeaderV1::from_bytes(
+            header_bytes
+                .try_into()
+                .expect("header_bytes should be 7 bytes"),
```
mlegner: Nit: I think it's now 6 bytes.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -264,7 +508,112 @@ impl BlobWithIdentifierOwned {
...
+        );
+
+        let mut offset = Self::BLOB_HEADER_SIZE;
+        let mut blob_bytes_size = usize::try_from(blob_header.length).map_err(|e| {
+            QuiltError::Other(format!("Blob length exceeds maximum allowed value: {}", e))
```
mlegner: Nit: I think we could use an `expect` here: We specify that Walrus requires at least a 32-bit architecture.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -264,7 +508,112 @@ impl BlobWithIdentifierOwned {
...
+    /// Decodes a blob from a column data source.
+    pub fn decode_blob<T>(
+        data_source: &T,
+        start_col: usize,
+        end_col: usize,
+    ) -> Result<QuiltStoreBlobOwned, QuiltError>
+    where
+        T: QuiltColumnRangeReader,
+    {
+        let mut blob = QuiltStoreBlobOwned::default();
```
mlegner: Nit: Instead of creating a default blob here and then updating it, why not just create the final version at the end of the function?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -280,14 +629,82 @@ impl QuiltVersion for QuiltVersionV1 {
     }
 }
 
+/// The header of a encoded blob in QuiltVersionV1.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct BlobHeaderV1 {
+    pub length: u32, // the lower 34 bits are used for the length.
+    pub mask: u8,    // the lower 14 bits are used for the mask.
```
mlegner: Nit: The comments are no longer correct (could be removed):
```suggestion
    pub length: u32,
    pub mask: u8,
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -280,14 +629,82 @@ impl QuiltVersion for QuiltVersionV1 {
...
+
+impl BlobHeaderV1 {
+    const LENGTH_BYTES: usize = 4;
+    const MASK_BYTES: usize = 1;
+    const METADATA_ENABLED: u8 = 1;
+
+    const LENGTH_MAX: u32 = u32::MAX;
+    const MASK_MAX: u8 = u8::MAX;
+
+    /// Creates a new header with length and mask initialized to 0.
```
mlegner: Nit: The docstring doesn't match the actual behavior.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -280,14 +629,82 @@ impl QuiltVersion for QuiltVersionV1 {
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default)]
+pub struct BlobHeaderV1 {
+    pub length: u32, // the lower 34 bits are used for the length.
+    pub mask: u8,    // the lower 14 bits are used for the mask.
+}
+
+impl BlobHeaderV1 {
+    const LENGTH_BYTES: usize = 4;
+    const MASK_BYTES: usize = 1;
+    const METADATA_ENABLED: u8 = 1;
```
mlegner, line 478: Nit: Can you describe the meaning of this constant?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -280,14 +629,82 @@ impl QuiltVersion for QuiltVersionV1 {
...
+        // Read 4 bytes for length (bytes 1-4)
+        let length = u32::from_le_bytes([bytes[1], bytes[2], bytes[3], bytes[4]]);
+
+        // Read 1 byte for mask (byte 5)
+        let mask = bytes[5];
+
+        Self { length, mask }
+    }
+
+    /// Converts the `BlobHeaderV1` to a 7-byte array.
```
mlegner: Nit: This is now 6 bytes.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -280,14 +629,82 @@ impl QuiltVersion for QuiltVersionV1 {
...
+        data[0] = QuiltVersionV1::QUILT_VERSION_BYTE;
+
+        // Set length bytes (1-4)
+        let length_bytes = self.length.to_le_bytes();
+        data[1..5].copy_from_slice(&length_bytes);
+
+        // Set mask byte (5)
+        data[5] = self.mask;
+
+        // byte 6 is unused, left as 0
```
mlegner: Nit:
```suggestion
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -358,29 +780,152 @@ impl QuiltApi<QuiltVersionV1> for QuiltV1 {
...
+        QuiltSymbolIter::new(
+            data,
+            row_size,
+            symbol_size,
+            start_col,
+            end_col,
+            begin_position,
+            limit,
+        )
+    }
```
mlegner: Question: Related to another question above, do we need an iterator here? It seems to me that something similar as `range_read_from_columns` would work here as well.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -516,13 +1060,153 @@ pub struct QuiltEncoderV1<'a> {
...
+    /// - BlobHeaderV1: Contains version byte, length , and mask feature flags.
+    /// - Identifier Size: 2-byte length of the serialized identifier.
+    /// - Serialized Identifier: BCS-encoded identifier string.
+    /// - Feature Data: Optional section for feature data, such as attributes,
+    ///   (when mask flag is set).
+    /// - Blob Data: The actual blob contents.
+    fn add_blob_to_quilt(
+        data: &mut [u8],
+        blob: &QuiltStoreBlob,
+        is_meta_blob: bool,
```
mlegner: Question: What is this parameter? Is this `true` for the index and `false` otherwise?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -516,13 +1060,153 @@ pub struct QuiltEncoderV1<'a> {
...
+    /// +------------------+----------------+------------+----------------------+------------------+
+    /// ```
+    ///
+    /// - BlobHeaderV1: Contains version byte, length , and mask feature flags.
+    /// - Identifier Size: 2-byte length of the serialized identifier.
+    /// - Serialized Identifier: BCS-encoded identifier string.
+    /// - Feature Data: Optional section for feature data, such as attributes,
+    ///   (when mask flag is set).
+    /// - Blob Data: The actual blob contents.
+    fn add_blob_to_quilt(
```
mlegner, line 910: Nit: Can you describe what this function returns? I assume the number of columns needed for the blob.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -590,54 +1277,45 @@ impl QuiltEncoderApi<QuiltVersionV1> for QuiltEncoderV1<'_> {
...
-                data[dest_idx..dest_idx + chunk.len()].copy_from_slice(chunk);
-                row = (row + 1) % n_rows;
-                if row == 0 {
-                    col += 1;
-                }
-                offset += chunk.len();
-            }
-        };
-
         // First pass: Fill data with actual blobs and populate quilt patches.
```
mlegner: Question: If this is the first pass, what is the second pass?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -698,43 +1376,41 @@ impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
...
-        let index_size = data_size - prefix_size;
-        let mut combined_data = Vec::with_capacity(index_size);
+        let index_size_bytes = self.range_read_from_columns(
+            0,
+            1,
+            QUILT_VERSION_BYTES_LENGTH,
+            QUILT_INDEX_SIZE_BYTES_LENGTH,
+        )?;
+        assert_eq!(index_size_bytes.len(), QUILT_INDEX_SIZE_BYTES_LENGTH);
+        assert_eq!(index_size_bytes.len(), QUILT_INDEX_SIZE_BYTES_LENGTH);
```
mlegner: Nit: Duplicated assertion:
```suggestion
        assert_eq!(index_size_bytes.len(), QUILT_INDEX_SIZE_BYTES_LENGTH);
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -772,6 +1448,57 @@ impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
...
+        self.check_missing_slivers(start_col, end_col)?;
+
+        let slivers: Vec<&SliverData<Secondary>> = (start_col..end_col)
+            .map(|col| {
+                *self
+                    .slivers
+                    .get(&SliverIndex::new(col as u16))
+                    .expect("Sliver exists")
+            })
+            .collect();
```
mlegner, line 1205: Remark: Depending on the `bytes_to_skip` and `bytes_to_return`, we wouldn't actually need all the slivers in the range. However, I expect that doesn't matter much for how it is used in practice.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -772,6 +1448,57 @@ impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
     }
 }
 
+// Implementation of QuiltColumnRangeReader for QuiltDecoderV1.
+impl QuiltColumnRangeReader for QuiltDecoderV1<'_> {
+    fn range_read_from_columns(
+        &self,
+        start_col: usize,
+        end_col: usize,
+        bytes_to_skip: usize,
```
mlegner: Nit: You can define this as mutable so you don't have to write `let mut bytes_to_skip = bytes_to_skip;` below:
```suggestion
        mut bytes_to_skip: usize,
```

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: My bad, I forgot to delete these after the refactoring. Deleted.

crates/walrus-core/src/encoding/quilt_encoding.rs

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Thanks for the catch, fixed.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good question! Change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Thanks for the catch, suggestion applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Fixed.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Comments added, and renamed the constants.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Fixed.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Suggestion applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Removed, thanks for the catch!

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Yes, it is. comments added.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Corrected the comment.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: suggestion applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Yes, you are right, the `end_col` is passed in to make it easy to check all `columns` are present, but not necessary for the function.

Keep it as, unless you have a strong preference here.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: thanks for the tip, applied changes.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Yes.
comments added.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Right, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -511,13 +311,57 @@ impl QuiltVersionV1 {
...
+        ))
+        .map_err(|_| QuiltError::FailedToExtractQuiltIndexSize)?;
+
+        let index_bytes =
+            data_source.range_read_from_columns(0, QUILT_INDEX_PREFIX_SIZE, index_size)?;
+        assert!(index_bytes.len() == index_size);
+
+        // Decode the QuiltIndexV1.
+        let mut quilt_index: QuiltIndexV1 = bcs::from_bytes(&index_bytes)
+            .map_err(|e| QuiltError::QuiltIndexSerDerError(e.to_string()))?;
```
mlegner: Nit: This particular error type could directly include the BCS error instead of its string representation. Also, the error message in the only other case is a bit confusing (`failed to decode the quilt index: failed to serialize quilt index: [...]`).

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -127,14 +129,28 @@ pub trait QuiltDecoderApi<'a, V: QuiltVersion> {
     fn get_or_decode_quilt_index(&mut self) -> Result<&V::QuiltIndex, QuiltError>;
```
mlegner, line 131: Suggestion: Can you update the docstring of this function? Also, given that this already takes a mutable reference, we could specify (and implement) that it updates `self` to store the index, then it wouldn't be necessary to do this explicitly in line 595.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -127,14 +129,28 @@ pub trait QuiltDecoderApi<'a, V: QuiltVersion> {
...
 
     /// Gets a blob by its identifier from the quilt.
-    fn get_blob_by_identifier(&self, identifier: &str) -> Result<Vec<u8>, QuiltError>;
+    fn get_blob_by_identifier(&self, identifier: &str) -> Result<QuiltStoreBlobOwned, QuiltError>;
```
mlegner, line 134: Remark: I just realized that we have the same function with an independent implementation in the `QuiltApi` trait. Is this intentional? Could we deduplicate this a bit (possibly in a follow-up)?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -1453,10 +1182,18 @@ impl QuiltColumnRangeReader for QuiltDecoderV1<'_> {
...
+        mut bytes_to_return: usize,
     ) -> Result<Vec<u8>, QuiltError> {
+        self.check_missing_slivers(start_col, start_col + 1)?;
+        let column_size = self
+            .slivers
+            .get(&SliverIndex(start_col as u16))
+            .expect("sliver should exist")
+            .symbols
+            .data()
+            .len();
```
mlegner: Question: Would it make sense to track the `column_size` somehow in `QuiltDecoderV1`?

mlegner (approved):
Thanks so much for the changes, @liquid-helium! LGTM, just a few final nits. 💯

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: good call! changed to `QuiltIndexSerDerError(#[from] bcs::Error),`.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good idea.
Change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice observation, currently they deduplication happens at `QuiltVersionV1::decode_blob(self, start_col)`, but we can add another trait to abstract `get_blob_by*` functions.
I added a TODO(WAL-862) here, and we can add this if there are more duplication in the future.

crates/walrus-core/src/encoding/quilt_encoding.rs

PR #2107, opened 2025-05-20
feat: storage node enters recovery when lagging behind during operation

halfprice:
## Description

Before this change, storage node can only enter recovery mode upon restarts.
However, a storage node can become lagging behind and slow processing events and eventually falling out of
the current committees' scope (2 epoch behind or more).

In this case, the node should directly enter recovery mode.

Resolve WAL-479

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1374,37 +1374,8 @@ impl StorageNode {
         // shards are created.
         let shard_map_lock = self.inner.storage.lock_shards().await;
 
-        if self.inner.storage.node_status()? == NodeStatus::RecoveryCatchUp {
-            self.process_epoch_change_start_while_catching_up(event_handle, event, shard_map_lock)
-                .await?;
-        } else {
-            let enter_recovery_mode = match self.begin_committee_change(event.epoch).await? {
-                BeginCommitteeChangeAction::ExecuteEpochChange => false,
-                BeginCommitteeChangeAction::SkipEpochChange => {
-                    event_handle.mark_as_complete();
-                    return Ok(());
-                }
-                BeginCommitteeChangeAction::EnterRecoveryMode => true,
-            };
-
-            if enter_recovery_mode {
-                tracing::info!("storage node entering recovery mode");
-                self.inner.set_node_status(NodeStatus::RecoveryCatchUp)?;
-                self.process_epoch_change_start_while_catching_up(
-                    event_handle,
-                    event,
-                    shard_map_lock,
-                )
-                .await?;
-            } else {
-                self.process_epoch_change_start_when_node_is_in_sync(
-                    event_handle,
-                    event,
-                    shard_map_lock,
-                )
-                .await?;
-            }
-        }
+        self.execute_epoch_change_start_event(event_handle, event, shard_map_lock)
+            .await?;
 
         // Update the latest event epoch to the new epoch. Now, blob syncs will use this epoch to
         // check for shard ownership.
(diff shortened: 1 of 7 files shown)


halfprice:
Hi @mlegner, this PR is ready for review~

crates/walrus-service/src/node/committee/committee_service.rs
```diff
@@ -574,35 +575,63 @@ where
...
-        } else if new_epoch < expected_next_epoch - 1 {
-            return Err(BeginCommitteeChangeError::EpochIsLess {
-                expected: expected_next_epoch,
-                actual: new_epoch,
-            });
         }
-        debug_assert_eq!(new_epoch, expected_next_epoch);
 
+        // Now, fetch the latest committee from the chain.
         let latest = self
```
mlegner: Nit (pre-existing): Can we call this `latest_committees` or something like that?

crates/walrus-service/src/node.rs
```diff
@@ -1384,9 +1386,50 @@ impl StorageNode {
         Ok(())
     }
 
+    /// Storage node execution of the epoch change start event.
+    async fn execute_epoch_change_start_event(
```
mlegner: Nit/question: It's a bit unclear what the relation is between "process" and "execute".

crates/walrus-service/src/node.rs
```diff
@@ -1600,26 +1639,45 @@ impl StorageNode {
...
+                debug_assert!(requested_epoch < latest_epoch);
+                // We are processing a backlog of events. Since the committee service has a
+                // more recent committee. In this situation, we have already lost the information
+                // and the shard assignment of the previous epoch relative to `event.epoch`, the
+                // node cannot execute the epoch change. Therefore, the node needs to enter recovery
+                // mode to catch up to the latest epoch as quickly as possible.
+                tracing::warn!(
+                    ?latest_epoch,
+                    ?requested_epoch,
+                    "epoch change requested for an older epoch than the latest epoch, this \
```
```suggestion
                    "epoch change requested for an older epoch than the latest epoch; this means \
```

crates/walrus-service/src/node.rs
```diff
@@ -1579,15 +1617,16 @@ impl StorageNode {
...
-    /// Initiates a committee transition to a new epoch.
+    /// Initiates a committee transition to a new epoch. Upon the return of this function, the
+    /// latest committee on chain is updated to the new node.
     ///
-    /// Returns `true` if epoch change event has started or was sufficiently recent such
-    /// that it should be handled.
+    /// Returns the action to execute epoch change based on the result of committee service,
+    /// including possible actions to enter recovery mode due to the node being severely lagging.
     #[tracing::instrument(skip_all)]
     async fn begin_committee_change(
```
mlegner, line 1629: Remark: AFAICT, the `CommitteeService::begin_committee_change` function is only used here (+ some tests). This results in effectively doing (almost) the same case distinction twice. Maybe this could be simplified (in a separate refactoring)?

crates/walrus-simtest/tests/simtest_failure.rs
```diff
@@ -557,4 +561,160 @@ mod tests {
...
+                    .find(|s| s.shard == shard.shard);
+                assert!(
+                    shard_i_status.is_none()
+                        || shard_i_status.unwrap().status != ShardStatus::ReadOnly
+                );
+            }
+        }
+
+        assert_eq!(
+            simtest_utils::get_nodes_health_info(&[&walrus_cluster.nodes[0]])
```
mlegner: Question: Why do we have to get the health info again instead of reusing the `node_health_info`?

crates/walrus-simtest/tests/simtest_failure.rs
```diff
@@ -557,4 +561,160 @@ mod tests {
...
+            .build_generic::<SimStorageNodeHandle>()
+            .await
+            .unwrap();
+
+        let blob_info_consistency_check = BlobInfoConsistencyCheck::new();
+
+        let client_arc = Arc::new(client);
+
+        // Starts a background workload that a client keeps writing and retrieving data.
+        // All requests should succeed even if a node crashes.
```
mlegner: Nit: There are several mentions of "crash" in this test. AFAICT, the node doesn't actually crash but just lag. Can you change some of the comments to avoid the confusion?

mlegner (approved):
Thanks a lot for taking care of this corner case! 🙏

crates/walrus-service/src/node.rs

crates/walrus-simtest/tests/simtest_failure.rs
halfprice: Good catch!

crates/walrus-simtest/tests/simtest_failure.rs
halfprice: Done.

crates/walrus-service/src/node/committee/committee_service.rs
halfprice: Done.

crates/walrus-service/src/node.rs
halfprice: what my original thoughts are:

process means process epoch change event, which includes everything the node needs to do upon receiving epoch change start event, including things like consistency check that is not really part of the epoch change execution logic.

execute means the node actually executes the epoch change logic to bring the node state to the next epoch.

So I replaced `execute_epoch_change_start` to `execute_epoch_change`, and added some more documentation to make it clear.

crates/walrus-service/src/node.rs
halfprice: Good suggestion. Will tackle it in a follow up PR.

PR #2110, opened 2025-05-21
fix: use config PRC endpoints when creating SuiContractClient from wallet

halfprice:
## Description

This was missed during the last round of GRPC proxy testing, and thanks @wbbradley for making the great
improvement to easily set RPC endpoints when creating SuiContractClient from a wallet!

Before this PR, if the endpoint specified in the wallet is unavailable, restarting the node won't work even other
RPC endpoints specified in the node config are still available. By only using the RPC endpoints from the config,
now SuiContractClient can correctly fall to the next client.

Changed back some simtest semantics from before #2052 since now the fall over should work.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/common/config.rs
+++ b/crates/walrus-service/src/common/config.rs
@@ -71,19 +71,19 @@ impl SuiConfig {
     }
 
     /// Creates a [`SuiContractClient`] based on the configuration.
+    /// Always use the RPC endpoints supplied in the configuration instead of the wallet's RPC URL
+    /// to allow for failover.
     pub async fn new_contract_client(
         &self,
         metrics: Option<Arc<SuiClientMetricSet>>,
     ) -> Result<SuiContractClient, SuiClientError> {
         let wallet = WalletConfig::load_wallet(Some(&self.wallet_config), self.request_timeout)?;
 
-        #[allow(deprecated)]
-        let rpc_urls = &[wallet.get_rpc_url()?];
-
+        let rpc_urls = combine_rpc_urls(&self.rpc, &self.additional_rpc_endpoints);
         if let Some(metrics) = metrics {
             SuiContractClient::new_with_metrics(
                 wallet,
-                rpc_urls,
+                &rpc_urls,
                 &self.contract_config,
                 self.backoff_config.clone(),
                 self.gas_budget,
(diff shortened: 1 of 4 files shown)


crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -103,14 +103,26 @@ pub enum LazyFallibleRpcClientBuilder {
...
-    async fn lazy_build_client(
-        &self,
-        _is_last_chance: bool,
-    ) -> Result<Arc<FallibleRpcClient>, FailoverError> {
+    async fn lazy_build_client(&self) -> Result<Arc<FallibleRpcClient>, FailoverError> {
         // Inject rpc client build failure for simtests.
         #[cfg(msim)]
         {
-            if !_is_last_chance {
+            let mut fail_client_creation = false;
```
halfprice, line 110: I would like to change the failure injection back to prior #2052 because since now ContractClient created from wallet should be able to leverage multiple RPCs, the old test semantic makes sure that we do fail the primary RPCs. This is also the way to test the change made in this PR.

crates/walrus-service/src/common/config.rs
```diff
@@ -71,19 +71,19 @@ impl SuiConfig {
     }
 
     /// Creates a [`SuiContractClient`] based on the configuration.
+    /// Always use the RPC endpoints supplied in the configuration instead of the wallet's RPC URL
```
wbbradley, line 74: It seems wrong to not include the wallet's rpc url, no?

halfprice: I thought that url from wallet is deprecated after #2052? It is marked as `deprecated` with notes that `Avoid calling this method as it relies on the Wallet configuration for rpc urls. It's better to pass rpc urls directly to the location where they are needed, from Walrus configuration.`

I can add it but want to make sure what's the plan for the url in wallet.

wbbradley: That's fair. If you are confident that the urls we're getting from the alternate source are better, then let's avoid calling `get_rpc_url`.

wbbradley (approved):
🚀

PR #2113, opened 2025-05-21
fix: avoid complex logic in select_coins_with_limit

wbbradley:
## Description

Fixes issue found in Antithesis run regarding polling a dead stream.

## Test plan

CI Pipeline.


karlwuest (approved):
Thanks for the fix, lgtm!

PR #2116, opened 2025-05-22
fix: simplify select_coins_with_limit

karlwuest:
## Description

Simplifies the `select_coins_with_limit` function to make it easier to understand.

## Test plan

existing tests

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
+++ b/crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
@@ -305,10 +305,7 @@ impl RetriableSuiClient {
                 .filter(|coin: &Coin| future::ready(!exclude.contains(&coin.coin_object_id)))
         );
 
-        // Limit the initial capacity to at most `MAX_GAS_PAYMENT_OBJECTS`, so that we don't try
-        // to allocate too much space if e.g. `usize::MAX` is passed as `max_num_coins`.
-        let mut selected_coins_heap: BinaryHeap<Reverse<OrderedCoin>> =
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -305,79 +305,49 @@ impl RetriableSuiClient {
                 .filter(|coin: &Coin| future::ready(!exclude.contains(&coin.coin_object_id)))
         );
 
-        let mut selected_coins = Vec::with_capacity(max_num_coins);
+        // Limit the initial capacity to at most `MAX_GAS_PAYMENT_OBJECTS`, so that we don't try
+        // to allocate too much space if e.g. `usize::MAX` is passed as `max_num_coins`.
+        let mut selected_coins_heap: BinaryHeap<Reverse<OrderedCoin>> =
+            BinaryHeap::with_capacity(max_num_coins.min(MAX_GAS_PAYMENT_OBJECTS));
```
mlegner: Suggestion: Don't set a capacity at all as we probably only need a single-digit number of coins in the vast majority of cases.

mlegner: Also, nit: Given we now *only* have the heap, we can just call it `selected_coins`.

mlegner (approved):
Thanks for improving the code clarity, @karlwuest!

PR #2122, opened 2025-05-22
feat: introduction of walrus upload relay

wbbradley:
## Description

Implement the Walrus Upload Relay.

## Test plan

Manual testing, CI.

Co-authored-by: giac-mysten <giacomo@mystenlabs.com>


scripts/merge_sui_coins.py
```diff
@@ -0,0 +1,132 @@
...
+
+def parse_args():
+    """Parses the command line arguments.
+
+    Only needed to show the help message.
+    """
+    parser = ArgumentParser(
+        description="Merge all SUI coins in the wallet into a single coin. \
+            Uses the address and the network that are set in the Sui CLI. \
+            As a result, the wallet will contain 2 SUI coins.",
```
mlegner: Question: Why are 2 coins left in the end? When using the gas-smashing, we should be able to end up with 1.

docker/walrus-upload-relay/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Build application
+#
+# Copy in all crates, Cargo.toml and Cargo.lock unmodified,
+# and build the application.
+FROM rust:1.88-bookworm AS builder
```
mlegner: Suggestion: Use the internal Docker registry to prevent issues with rate-limiting as we do in other Dockerfiles:
```suggestion
# This is the same as rust:1.88-bookworm but we use the GCP registry to avoid rate limiting.
FROM us-central1-docker.pkg.dev/cryptic-bolt-398315/base-images/rust:1.88-bookworm-amd64 AS builder
```

docker/walrus-upload-relay/Dockerfile
```diff
@@ -0,0 +1,30 @@
...
+RUN apt-get update && apt-get install -y cmake clang
+
+COPY Cargo.toml Cargo.lock ./
+COPY crates crates
+COPY contracts contracts
+
+RUN cargo build --profile $PROFILE --bin walrus-upload-relay
+
+# Production Image for walrus walrus-upload-relay
+FROM debian:bookworm-slim AS walrus-upload-relay
```
mlegner: Same as above:
```suggestion
# This is the same as debian:bookworm-slim but we use the GCP registry to avoid rate limiting.
FROM us-central1-docker.pkg.dev/cryptic-bolt-398315/base-images/debian:bookworm-slim-amd64 AS walrus-upload-relay
```

crates/walrus-upload-relay/.gitignore
mlegner, line 1: Nit: This should not be needed as it is covered by the `**/target` in the main `.gitignore`.

crates/walrus-upload-relay/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+publish = false
+authors.workspace = true
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+
+[[bin]]
+name = "walrus-upload-relay"
+path = "src/main.rs"
+
```
mlegner, line 12: Suggestion: Enable our standard lints:
```suggestion

[lints]
workspace = true
```

crates/walrus-upload-relay/Cargo.toml
```diff
@@ -0,0 +1,48 @@
...
+tower-http.workspace = true
+tracing.workspace = true
+tracing-subscriber.workspace = true
+utoipa.workspace = true
+utoipa-redoc.workspace = true
+walrus-sdk.workspace = true
+walrus-test-utils.workspace = true
+
+[features]
+default = []
```
mlegner: Question: Is this needed when there aren't actually any features?

crates/walrus-upload-relay/src/controller.rs
```diff
@@ -0,0 +1,419 @@
...
+
+/// The configuration for the Walrus Upload Relay.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub(crate) struct WalrusUploadRelayConfig {
+    /// The configuration for tipping.
+    tip_config: TipConfig,
+    /// The maximum time gap between the time the tip transaction is executed (i.e., the tip is
+    /// paid), and the request to store is made to the Walrus upload relay.
+    tx_freshness_threshold: Duration,
```
mlegner, line 73: Suggestion: Can we parse this from seconds as we normally do for configuration parameters (same below)?
```suggestion
    #[serde(rename = "tx_freshness_threshold_secs")]
    #[serde_as(as = "DurationSeconds")]
    tx_freshness_threshold: Duration,
```

crates/walrus-sdk/src/client.rs
```diff
@@ -1535,7 +1535,7 @@ impl<T> Client<T> {
         metadata: &VerifiedBlobMetadataWithId,
         pairs: &[SliverPair],
         blob_persistence_type: &BlobPersistenceType,
-        multi_pb: &MultiProgress,
+        multi_pb: Option<&MultiProgress>,
```
mlegner, line 1538: Praise: Very good to change that to an option! 👍

crates/walrus-sdk/src/client.rs
```diff
@@ -1559,29 +1559,41 @@ impl<T> Client<T> {
...
+                    metadata,
+                    pairs_per_node
+                        .remove(&n.node_index)
+                        .expect("there are shards for each node"),
+                    blob_persistence_type,
+                );
+
+                let progress_bar = progress_bar.clone();
+                async move {
+                    let progress_bar = progress_bar.clone();
```
mlegner: Nit: I think this clone is not necessary.
```suggestion
```

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
mlegner, line 1: Maybe @karlwuest could take a look at this.

crates/walrus-upload-relay/src/tip/config.rs
```diff
@@ -0,0 +1,76 @@
...
+        json!(TipKind::Const(31415)),
+        json!(TipKind::Linear{base: 101, encoded_size_mul_per_kb: 42})
+))]
+pub(crate) enum TipKind {
+    /// A constant tip.
+    Const(u64),
+    /// A tip that linearly depends on the encoded size of the blob.
+    Linear {
+        base: u64,
+        encoded_size_mul_per_kb: u64,
```
mlegner: Nit: Given we're actually using KiB, not kB, this could be
```suggestion
        encoded_size_mul_per_kib: u64,
```

crates/walrus-upload-relay/src/tip/config.rs
```diff
@@ -0,0 +1,76 @@
...
+    ) -> Option<u64> {
+        Some(match self {
+            TipKind::Const(constant) => *constant,
+            TipKind::Linear {
+                base,
+                encoded_size_mul_per_kb,
+            } => {
+                base + encoded_blob_length_for_n_shards(n_shards, unencoded_length, encoding_type)?
+                    / 1024
+                    * encoded_size_mul_per_kb
```
mlegner: Question: Is it intentional that you round down here? In other similar cases (e.g., when computing storage cost) we use `div_ceil`.

crates/walrus-upload-relay/README.md
mlegner, line 1: Remark: The terms "fan-out" and "proxy" are used a few times here. Can we change the hostname of the deployed instance and fix the URL path? And use "relay" consistently instead of "proxy".

mlegner: Also, can we maybe move (parts of?) this documentation to our public docs in `docs/book/` (can be done in a follow-up)?

crates/walrus-upload-relay/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+Once the transaction is executed, the client keeps the transaction ID `tx_id` , the `nonce`, and the
+`blob_id`, which are required for the next phase.
+
+Note: the proxy will enforce a freshness check on the transaction that paid the tip (currently 1h by
+default, but each proxy can independently configure this).
+
+### Sending data to the proxy
+
+See the OpenAPI spec for the fanout proxy here: https://fan-out.testnet.walrus.space/v1/api
```
mlegner: Suggestion: Generate the OpenAPI spec and add it to the repo, similar to the [`check_and_update_openapi_spec` test here](https://github.com/MystenLabs/walrus/blob/64bd3974e94e7ce53f16124a784ecb1771fac4da/crates/walrus-service/src/client/daemon/openapi.rs#L112). Then we don't depend on any public deployments that may or may not be available.

crates/walrus-upload-relay/README.md
```diff
@@ -0,0 +1,180 @@
...
+Then, it will send a response to the client, containing the `blob_id` of the stored blob, along with
+the `confirmation_certificate`. The client can then use this certificate to certify the blob on
+chain.
+
+## Installation
+
+### Download the Binary
+
+If you'd like to download a pre-built binary in order manually run `walrus-upload-relay`, you'll need to download it from
+[here](https://github.com/MystenLabs/walrus/releases). Note that `walrus-upload-relay` does not daemonize
```
mlegner, line 120: Question: Do we already include this in our build/release process?

crates/walrus-upload-relay/README.md
```diff
@@ -0,0 +1,180 @@
...
+
+### Configuration
+
+An example `walrus-upload-relay` configuration file can be found
+[here](./walrus_upload_relay_config_example.yaml). The
+various options are described
+[here](./src/controller.rs#L63) and
+[here](./src/tip/config.rs#L57).
+
+# Conclusion
```
mlegner: Nit:
```suggestion
## Conclusion
```

crates/walrus-upload-relay/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    about = "The Walrus Upload Relay",
+    long_about = None,
+    name = env!("CARGO_BIN_NAME"),
+    version = VERSION,
+rename_all = "kebab-case",
+)]
+struct Args {
+    #[arg(
+        long,
+        short,
```
mlegner: Suggestion: Remove the `short` option as it may be more confusing than helpful.

crates/walrus-upload-relay/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    metrics_address: std::net::SocketAddr,
+    /// The configuration context to use for the client, if omitted the default_context is used.
+    #[arg(long)]
+    context: Option<String>,
+    /// The file path to the Walrus read client configuration.
+    #[arg(long)]
+    walrus_config: PathBuf,
+    /// The address to listen on. Defaults to 0.0.0.0:57391.
+    #[arg(long)]
+    server_address: Option<SocketAddr>,
```
mlegner: Suggestion: Use `SocketAddr` here directly with a default value of `DEFAULT_SERVER_ADDRESS`.

crates/walrus-upload-relay/src/error.rs
```diff
@@ -0,0 +1,133 @@
...
+use walrus_sdk::{
+    core::{BlobId, BlobIdParseError, encoding::DataTooLargeError},
+    error::ClientError,
+    sui::client::SuiClientError,
+};
+
+use crate::tip::error::TipError;
+
+/// Walrus Upload Relay error type.
+#[derive(Debug, Error)]
```
mlegner, line 20: Suggestion: Also derive `RestApiError` here, which is defined [here](https://github.com/MystenLabs/walrus/blob/main/crates/walrus-proc-macros/src/lib.rsL116).

mlegner (commented):
Thanks so much, @giac-mysten and @wbbradley! Great work implementing this upload relay! 💯 

This all looks good to me, I mostly have minor comments. A few general points:
- I didn't review all the code in `crates/walrus-upload-relay` in detail as this was already peer-reviewed and is self-contained.
- It would be good if @karlwuest could take another look at the Sui-related code.

crates/walrus-upload-relay/README.md
wbbradley: Not yet, but I'm actively working with Eugene on that.

wbbradley: https://github.com/MystenLabs/sui-operations/pull/5811/files

crates/walrus-upload-relay/src/main.rs
wbbradley: I agree.

scripts/merge_sui_coins.py
giac-mysten: Correct, this is from an older impl that did not use gas smashing, thanks for catching it!

crates/walrus-upload-relay/Cargo.toml
giac-mysten: removed.

crates/walrus-upload-relay/src/tip/config.rs
giac-mysten: I had the same reaction 😅 then, I think it's a good idea to change it to `div_ceil` @wbbradley.

--> changed.

crates/walrus-upload-relay/README.md
giac-mysten: yes, i'm already planning a to open a PR for the `docs`. I will integrate this README, and the notion internal docs.
(WAL-922)

crates/walrus-upload-relay/src/error.rs
giac-mysten: We had tried, but could not manage easily because some of the required types are tangled up inside the service crate, IIRC. added an issue to track a followup ([WAL-923](https://linear.app/mysten-labs/issue/WAL-923/implement-restapierror-for-the-upload-relay))

crates/walrus-upload-relay/src/tip/config.rs
wbbradley: This will cause changes to the SDK and the docs. I'll try to address those today.

wbbradley: FWIW I chose floor intentionally because it can model the same total tips as using div_ceil when combined with `base` and I found it easier to code (SDK will have to code this as well) and generally easier to understand. Most people are accustomed to floor since it is the default behavior in integer division in most languages, not ceil.

crates/walrus-upload-relay/upload_relay_openapi.yaml
mlegner, line 1: Suggestion: Can we add this and the example config to the match lists [here](https://github.com/MystenLabs/walrus/blob/main/.github/actions/diffs/action.yml#L72-L77)? This will add warnings to future PRs when the interface is changed.

mlegner (approved):
Thanks so much for the changes. LGTM! 💯 🎉

crates/walrus-upload-relay/upload_relay_openapi.yaml
giac-mysten: ah this is great, changed!

wbbradley: nice i was wondering how those messages got created.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -1074,6 +1076,59 @@ impl RetriableSuiClient {
             .await
     }
 
+    /// Lower-level primitive, executes a transaction from its bytes representation.
+    ///
+    /// This function does not wait for the local execution on the full node. It is equivalent to
+    /// calling `execute_transaction_block` with
+    /// `ExecuteTransactionRequestType::WaitForEffectsCert`.
+    #[tracing::instrument(err, skip(self))]
+    pub async fn execute_transaction_from_bytes(
```
halfprice, line 1085: This function is never used?

crates/walrus-upload-relay/src/controller.rs
```diff
@@ -0,0 +1,446 @@
...
+    body: Bytes,
+) -> Result<impl IntoResponse, WalrusUploadRelayError> {
+    let start = Instant::now();
+    let blob_id = params.blob_id;
+    tracing::info!(?params, "starting to process a relay request");
+    let response = controller
+        .fan_out(body, params)
+        .await
+        .inspect_err(|error| tracing::debug!(?error, "responding to request with error"))?;
+    tracing::info!(
```
halfprice, line 356: Nit: time elapsed is going to be logged per request. Probably expose this in a metric is easier to observe.

crates/walrus-upload-relay/src/tip/check.rs
```diff
@@ -0,0 +1,122 @@
...
+
+                if owner_address == *address
+                    && change.coin_type
+                        == TypeTag::from_str(SUI_COIN_TYPE).expect("SUI is always valid")
+                {
+                    if change.amount >= expected_tip {
+                        return Ok(());
+                    } else {
+                        tracing::debug!("insufficient tip");
+                        return Err(TipError::InsufficientTip(change.amount, expected_tip));
```
halfprice, line 113: Question: say if the user does not tip enough, the tip will be lost forever? The next tip still needs to be at least `expected_tip`?

crates/walrus-upload-relay/src/params.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+}
+
+/// The subset of query parameters of the Walrus Upload Relay, necessary to check the tip.
+///
+/// Compared to `Params`, the `tx_id` and the `nonce` are not optional, and `blob_id` and
+/// `deletable_blob_object` are not necessary.
+#[derive(Debug, Clone)]
+pub(crate) struct PaidTipParams {
+    pub tx_id: TransactionDigest,
```
halfprice, line 106: Nit: tx_digest is more commonly used.

crates/walrus-upload-relay/src/tip/check.rs
```diff
@@ -0,0 +1,122 @@
...
+            tracing::debug!(?diff, "the transaction is too old");
+            return Err(TipError::TxTooOld);
+        }
+    }
+
+    Ok(())
+}
+
+/// Checks if the execution results are sufficient to cover the tip.
+pub(crate) fn check_response_tip(
```
halfprice, line 70: Question: does the tip have to be in the same PTB that register the blob or not? If I read the doc correctly, the tip can be any standalone transaction, is that correct? If so, if someone pay the service address once, but reuse that transaction as evidence of payment over and over, can this be detected?

crates/walrus-upload-relay/src/controller.rs
```diff
@@ -0,0 +1,446 @@
...
+        let blob_persistence = if let Some(object_id) = params.deletable_blob_object {
+            BlobPersistenceType::Deletable {
+                object_id: object_id.into(),
+            }
+        } else {
+            BlobPersistenceType::Permanent
+        };
+        let confirmation_certificate: ConfirmationCertificate = self
+            .client
+            .send_blob_data_and_get_certificate(&metadata, &sliver_pairs, &blob_persistence, None)
```
halfprice, line 172: Question: is there any functioning change in how the upload is done? Is it still that wait for 2/3 shards' response and wait for extra half of the previous time? As a service, if it can upload to more shards as it can (it can still return to the client right after 2/3 responses), this will going to help reducing blob recovery in storage node significantly!

halfprice (commented):
Sorry for the delay! Took a quick look and it is great! I appreciate the extensive documentation added!

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
wbbradley: Yeah, it's going to be used in the next PR, where we add an upload relay client to the CLI.

crates/walrus-upload-relay/src/controller.rs
wbbradley: Yeah, that was mostly for debugging. I can clean that up. Thanks.

crates/walrus-upload-relay/src/tip/check.rs
wbbradley: Yes, that's By Design. The idea is that tips should be small enough and the configuration process should be usable enough that that is a non-issue. Counterparties could discuss reimbursement as needed.

crates/walrus-upload-relay/src/tip/check.rs
wbbradley: No, the tip transaction just needs to be built with the correct convention (a pure input in the first input slot containing the hashed auth package. The auth package contains a nonce, the blob digest, and its unencoded length, so it cannot be used to spoof payment for other blobs.

crates/walrus-upload-relay/src/controller.rs
wbbradley: There is no functioning change to how upload is done. Yeah, once we get folks using this we should look into making it help with network data propagation, although, it's just like the client CLI, that would put more burden on the client egress costs, workload, memory load, etc. But, seems like it would be a good knob to add to configuration.

crates/walrus-upload-relay/src/tip/check.rs
halfprice: Ah, I see. Make sense. The pure input has to be the first input, or any pure input? I guess you will provide a "pay tip transaction" for users to use?

wbbradley: In the current design, the pure input has to be the first input and contain the auth package hash but if that becomes cumbersome we could allow the input slot to be varying and then provide it as a parameter to the HTTP endpoint so that it would know where to look in the PTB - though for now this seems unnecessary, I think. 

I'm not sure what you mean by 'provide a "pay tip transaction"' exactly. The functionality to hit this endpoint already exists in the TSSDK where the scaffolding there handles creating the tip transaction as part of the blob registration PTB, but like I mentioned, it's _possible_ (and not technically difficult) to pre-register the blob but then separately decide to tip the walrus upload relay operator and use their service for the actual upload.

PR #2125, opened 2025-05-23
chore: add simtest config option in vscode setting

halfprice:
## Description

I often use multiple local repos, and by setting this in user setting doesn't work for me.

Since we are tracking the vscode setting now, adding the rust analyzer simtest option so that we don't need to always look for the option somewhere else.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,10 @@
   "rust-analyzer.rustfmt.extraArgs": [
     "--config",
     "group_imports=StdExternalCrate,imports_granularity=Crate,imports_layout=HorizontalVertical"
-  ]
+  ],
+  // Uncomment this option to allow rust analyzer to work with simtests.
+  // "rust-analyzer.cargo.extraEnv": {
+  //   "RUSTFLAGS": "--cfg msim",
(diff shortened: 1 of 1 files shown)


mlegner (commented):
Thanks, @halfprice, this is very useful!

Question: AFAIK, JSON doesn't support comments. Does it work this way? If not, maybe instead mention this somewhere in the `CONTRIBUTING.md`? Also, is there a way to quickly switch between simtests en- and disabled?

halfprice:
> Thanks, @halfprice, this is very useful!
> 
> Question: AFAIK, JSON doesn't support comments. Does it work this way? If not, maybe instead mention this somewhere in the `CONTRIBUTING.md`? Also, is there a way to quickly switch between simtests en- and disabled?

Thanks @mlegner ! 

This is config is always what I had (with msim commented out) before `settings.json` is checked in the repo, and comments works.

I'm not sure if there is another way to quickly en- and disable the setting. Jumping to the config file and uncomment/comment is the fastest. I use this config fairly frequent, and in the past week, I find that I have to go to my notes to copy-paste the config here is quite inconvenient.

PR #2129, opened 2025-05-23
feat: instrument event stream polling with tokio metrics

halfprice:
## Description

This is to measure any tokio metrics related to event stream polling.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/node.rs
```diff
@@ -1974,6 +1989,7 @@ impl StorageNodeInner {
     }
 
     /// Returns true if the blob is stored at all shards at the latest epoch.
+    #[tracing::instrument(skip_all)]
```
mlegner: Nit/question: Given that we have the `instrument` at the `is_stored_at_all_shards_impl`, do we actually need it here?

crates/walrus-service/src/node.rs
```diff
@@ -1984,6 +2000,7 @@ impl StorageNodeInner {
     /// Returns true if the blob is stored at all shards at the given epoch.
     /// Note that since shard assignment is only available for the current and previous epoch,
     /// this function will return false if the given epoch is not the current or previous epoch.
+    #[tracing::instrument(skip_all)]
```
mlegner: Nit/question: Given that we have the `instrument` at the `is_stored_at_all_shards_impl`, do we actually need it here?

crates/walrus-service/src/node.rs
```diff
@@ -1942,6 +1955,7 @@ impl StorageNodeInner {
         anyhow::bail!("unknown epoch {} when checking shard assignment", epoch);
     }
 
+    #[tracing::instrument(skip_all)]
```
mlegner: Nit: It might be interesting to get include the epoch here.

crates/walrus-service/src/node.rs
halfprice: Thanks @mlegner !. Was interested to differentiate `is_stored_at_all_shards_at_latest_epoch` and `is_stored_at_all_shards_at_epoch`. Since now is_stored_at_all_shards_impl includes epoch, so I think this sufficient. Remove the instrument here.

PR #2136, opened 2025-05-27
chore: add github action to create transactions for multisig

karlwuest:
## Description

Adds a github action to create unsigned transactions for the walrus multisigs to upgrade/migrate the subsidies contract/object.

Contributes to WAL-675

## Test plan

- Manually ran script
- tested action locally with `act` (except for the uploading of the artifact)

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/.github/workflows/create-tx-for-multisig.yml
+++ b/.github/workflows/create-tx-for-multisig.yml
@@ -61,7 +61,7 @@ jobs:
           cat artifacts/tx-data.txt
 
       - name: Upload Transaction Artifact
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # pin@v4
         with:
           name: transaction-data
           path: artifacts/tx-data.txt
--- a/.github/workflows/create-tx-for-multisig.yml
+++ b/.github/workflows/create-tx-for-multisig.yml
@@ -1,0 +1,67 @@
+name: Build TX for Multisig Operation
+
+on:
+  workflow_dispatch:
+    inputs:
+      transaction_type:
+        description: "select transaction type to create"
+        type: choice
+        options:
+          - upgrade-subsidies
+          - migrate-subsidies
+      rpc:
+        description: "RPC url"
+        required: true
+        default: "https://fullnode.mainnet.sui.io:443"
+        type: string
+      gas_object_id:
+        description: "object id to get gas from for multisig transaction"
+        required: false
+        type: string
+      commit:
+        description: "commit hash of the walrus repo to create the transaction"
+        required: true
+        type: string
+
+jobs:
+  create-tx:
+    name: Create TX for multisig
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Selected transaction type
+        run: |
+          echo ${{ inputs.transaction_type }}
+
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 1
+          ref: ${{ inputs.commit }}
+
+      - name: Install Homebrew
+        run: |
+          /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+          echo "/home/linuxbrew/.linuxbrew/bin" >> $GITHUB_PATH
+
+      - name: Install Sui using Homebrew
+        run: brew install sui
+
+      - name: Setup sui client config
+        run: |
+          sui client --yes new-env --rpc ${{ inputs.rpc }} --alias mainnet
+          sui client switch --env mainnet
+
+      - name: Create unsigned transaction for multisig
+        run: |
+          mkdir -p artifacts
+          ./scripts/create_multisig_tx.sh -t '${{ inputs.transaction_type }}' -g '${{ inputs.gas_object_id }}' >artifacts/tx-data.txt
+
+      - name: Show Transaction Data (To sign)
+        run: |
+          cat artifacts/tx-data.txt
+
+      - name: Upload Transaction Artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: transaction-data
+          path: artifacts/tx-data.txt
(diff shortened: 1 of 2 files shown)


scripts/create_multisig_tx.sh
```diff
@@ -0,0 +1,89 @@
...
+  then
+    sui client gas $ADDRESS  --json | jq -r 'max_by(.mistBalance) | .gasCoinId'
+  else
+    echo $GAS_OBJECT_ID
+  fi
+}
+
+upgrade_subsidies() {
+  GAS=$(gas_obj_for_addr $WALRUS_ADMIN)
+  GAS_BUDGET=$(sui client object $GAS --json | jq -r '.content.fields.balance')
```
mlegner, line 40: Question: AFAIU, this means that if that gas object is used for any other transaction in the meantime, this one will fail, correct? Is this intentional? If not, can we reduce the gas budget?

scripts/create_multisig_tx.sh
mlegner, line 1: Suggestion: Can you add some minimal usage instructions? Can also be that it should only be used through the GH workflow.

scripts/create_multisig_tx.sh
```diff
@@ -0,0 +1,89 @@
...
+
+case "$TX_TYPE" in
+  upgrade-subsidies)
+    upgrade_subsidies
+    ;;
+  migrate-subsidies)
+    migrate_subsidies
+    ;;
+  *)
+    echo "Invalid transaction type \"$TX_TYPE\"" >&2
```
mlegner: Nit: Add a check for an empty tx type and print a different error in that case.

.github/workflows/create-tx-for-multisig.yml
```diff
@@ -0,0 +1,67 @@
...
+        description: "RPC url"
+        required: true
+        default: "https://fullnode.mainnet.sui.io:443"
+        type: string
+      gas_object_id:
+        description: "object id to get gas from for multisig transaction"
+        required: false
+        type: string
+      commit:
+        description: "commit hash of the walrus repo to create the transaction"
```
mlegner, line 22: Remark: This could also be a branch or tag, correct? But I guess it's intentional to force the user to specify a commit instead of using `main` as the default?

.github/workflows/create-tx-for-multisig.yml
```diff
@@ -0,0 +1,67 @@
...
+  create-tx:
+    name: Create TX for multisig
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Selected transaction type
+        run: |
+          echo ${{ inputs.transaction_type }}
+
+      - uses: actions/checkout@v4
```
mlegner: Nit: Can we pin the version?
```suggestion
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
```

.github/workflows/create-tx-for-multisig.yml
```diff
@@ -0,0 +1,67 @@
...
+          fetch-depth: 1
+          ref: ${{ inputs.commit }}
+
+      - name: Install Homebrew
+        run: |
+          /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
+          echo "/home/linuxbrew/.linuxbrew/bin" >> $GITHUB_PATH
+
+      - name: Install Sui using Homebrew
+        run: brew install sui
```
mlegner, line 47: Remark: Maybe a bit overkill to install homebrew just for that. You could also get the binary directly from the GH release (although that would require specifying the version). 🤷

.github/workflows/create-tx-for-multisig.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          mkdir -p artifacts
+          ./scripts/create_multisig_tx.sh -t '${{ inputs.transaction_type }}' -g '${{ inputs.gas_object_id }}' >artifacts/tx-data.txt
+
+      - name: Show Transaction Data (To sign)
+        run: |
+          cat artifacts/tx-data.txt
+
+      - name: Upload Transaction Artifact
+        uses: actions/upload-artifact@v4
```
mlegner: Nit: Can you pin that as well?
```suggestion
        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # pin@v4
```

mlegner (commented):
Thanks for adding this, @karlwuest. Generally looks good, just a few thoughts. 🙂

scripts/create_multisig_tx.sh
karlwuest: Yes, that's true, but that is anyway the case, because the transaction will contain the version of the gas object, so all transactions created in parallel using the same gas object will conflict with this transaction. 

We could set a lower gas budget, but we don't gain much from it and it's not clear how to set it. The problem is that automatic gas estimation does not work in the sui cli if a gas object is specified and it has less than the maximum gas (because for the estimation the maximum is set in a dry run).

.github/workflows/create-tx-for-multisig.yml
karlwuest: Yes, I specifically wrote the description to nudge the user to specify a commit hash, but a branch or tag could also be used.

.github/workflows/create-tx-for-multisig.yml
karlwuest: I agree, but this seemed like the easiest solution (and is used that way in the suins repo as well), because I didn't want to specify the version. I initially wanted to do it via `suiup`, but for some reason that did not work. 

Happy to change it to get the binary directly if you think that's preferable, then we just also have to bump it whenever we bump the version elsewhere 🤷‍♂️ Lmk

scripts/create_multisig_tx.sh
mlegner: Ah, makes sense. 👍

PR #2137, opened 2025-05-27
feat(fanout): add logic to check tips and registrations

giac-mysten:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/fan-out-proxy/src/tip/config.rs
```diff
@@ -0,0 +1,39 @@
...
+use sui_types::base_types::SuiAddress;
+
+/// The kinds of tip that the proxy can choose to configure.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub(crate) enum TipKind {
+    /// A constant tip.
+    Const(u64),
+    /// A tip that linearly depends on the encoded size of the blob.
+    // TODO: make this linear in the KB or MB size of the blob?
+    Linear { base: u64, encoded_size_mul: u64 },
```
wbbradley, line 16: Nice idea. Might need a denominator, in case they want to charge 1.5 SUI per byte, or something?

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -1076,6 +1078,59 @@ impl RetriableSuiClient {
...
+                    make_request(
+                        client.clone(),
+                        tx_bytes.clone(),
+                        signatures.clone(),
+                        options.clone(),
+                        request_type.clone(),
+                    )
+                },
+                self.metrics.clone(),
+                method,
```
wbbradley, line 1126: No need to parameterize "method" here, just hardcode "execute_transaction_from_bytes" here, I think.

crates/fan-out-proxy/src/tip/config.rs
giac-mysten: I think the values should all anyways be in MIST, which is the smalles unit we can charge. So 1.5 SUI would be `1_500_000_000 MIST`

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
giac-mysten: good point. I'll change this tomorrow, or otherwise please go ahead and change it directly 👍

PR #2138, opened 2025-05-27
chore: remove the old RaptorQ-based encoding

mlegner:
## Description

Closes WAL-846.
Closes WAL-610.
Contributes to WAL-611.

## Test plan

Adapted existing test suite.

---

## Release notes

I think we don't need release notes here, because the encoding is no longer used in any public Walrus network.


mlegner:
> Can we mark the variant as deprecated or does that [create warnings for derived functions](https://github.com/rust-lang/rust/issues/92313)?

Yes, indeed. But I can try implementing one of the workarounds.

crates/walrus-core/src/encoding/basic_encoding.rs
```diff
@@ -33,3 +37,305 @@ pub trait Decoder {
         T::IntoIter: Iterator<Item = DecodingSymbol<U>>,
         U: EncodingAxis;
 }
+
+/// Wrapper to perform a single encoding with Reed-Solomon for the provided parameters.
+pub struct ReedSolomonEncoder {
```
karlwuest, line 42: Just to confirm: The changes here are basically just moving the code from `basic_encoding/reed_solomon.rs` here, right? Or are there other changes?

mlegner: Yes, this is just an exact code move.

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -811,32 +794,24 @@ mod tests {
...
-        // 3. the metadata produced by `encode_with_metadata` is the same as
-        //    the metadata produced by `compute_metadata_only`.
-        // Takes long (O(1s)) to run.
+    /// A big test checking that:
+    /// 1. The sliver pairs produced by `encode_with_metadata` are the same as the ones produced by
+    ///    `encode`;
+    /// 2. the metadata produced by `encode_with_metadata` is the same as the metadata that can be
+    ///    computed from the sliver pairs directly.
+    /// 3. the metadata produced by `encode_with_metadata` is the same as the metadata produced by
+    ///    `compute_metadata_only`. Takes long (O(1s)) to run.
```
karlwuest: Question/Suggestion: Not directly related to this PR, but should we ignore the test by default?

crates/walrus-core/src/lib.rs
```diff
@@ -671,76 +647,86 @@ by_axis::derive_try_from_trait!(
...
-    RedStuffRaptorQ = 0,
-    /// RedStuff using the Reed-Solomon erasure code.
-    RS2 = 1,
-}
-
-impl From<EncodingType> for u8 {
-    fn from(value: EncodingType) -> Self {
-        value as u8
+// Defining the `EncodingType` in a separate module to be able to use `#[allow(deprecated)]` with
+// limited scope. Without this, the serde derivations would cause deprecation warnings.
```
karlwuest, line 651: Nice workaround!

karlwuest (approved):
Thanks for the cleanup, lgtm!

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: Good question. However, it turns out that the comment is not (maybe no longer?) correct 😁:
```
        PASS [   0.069s] walrus-core encoding::blob_encoding::tests::test_encode_with_metadata
```
I'll adjust the comment instead. 🙂

NodeGuy:
> Closes WAL-846.
> Closes WAL-610.
> Contributes to WAL-611.

Are these issues tracked elsewhere (e.g., in Linear)?  Is there public access to read them?

PR #2140, opened 2025-05-27
feat: update commands for contract upgrades

karlwuest:
## Description

- Adds a command to compute the package digest
- Separates the migrate command from the upgrade command
- Allows creating an unsigned transaction for the upgrade (to allow emergency upgrades from the multisig addr)
- Includes refactoring to build TransactionData in the WalrusPtbBuilder
- closes WAL-741

## Test plan

- Manual test of the new cli commands

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: Adds node-admin command to compute package digest

### DIFF
--- a/crates/walrus-service/bin/deploy.rs
+++ b/crates/walrus-service/bin/deploy.rs
@@ -13,7 +13,7 @@ use std::{
 
 use anyhow::Context;
 use clap::{Parser, Subcommand};
-use sui_types::base_types::ObjectID;
+use sui_types::base_types::{ObjectID, SuiAddress};
 use walrus_core::EpochCount;
 use walrus_service::{
     node::config::{
@@ -55,6 +55,8 @@ enum Commands {
     /// Upgrades the system contract with a quorum-based upgrade that has been voted for by
     /// a quorum of storage nodes by shard weight.
     Upgrade(UpgradeArgs),
+    /// Migrates the system and staking objects to a new package version.
+    Migrate(MigrateArgs),
 }
 
 #[derive(Debug, Clone, clap::Args)]
@@ -249,6 +251,34 @@ struct UpgradeArgs {
     /// The upgrade manager object ID.
     #[arg(long)]
     upgrade_manager_object_id: ObjectID,
+    /// Create and output a serialized unsigned transaction. Useful for an emergency upgrade
+    /// authorized by a multisig address.
+    #[arg(long)]
+    serialize_unsigned: bool,
+    /// The sender address to use to create the serialized unsigned transaction. If not specified,
+    /// the active address from the wallet is used.
+    #[arg(long, requires = "serialize_unsigned")]
+    sender: Option<SuiAddress>,
+    /// Also migrate the system and staking objects after upgrading the contract.
+    #[arg(long, conflicts_with = "serialize_unsigned")]
+    migrate: bool,
+}
+
+#[derive(Debug, Clone, clap::Args)]
+struct MigrateArgs {
+    /// The path to the wallet used to perform the upgrade. If not provided, the default
+    /// wallet path is used.
+    #[arg(long)]
+    wallet_path: Option<PathBuf>,
+    /// The staking object ID.
+    #[arg(long)]
+    staking_object_id: ObjectID,
+    /// The system object ID.
+    #[arg(long)]
+    system_object_id: ObjectID,
+    /// The package ID of the upgraded package.
+    #[arg(long)]
+    new_package_id: ObjectID,
 }
 
 fn main() -> anyhow::Result<()> {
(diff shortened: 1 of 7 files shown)


wbbradley:
LGTM, but I might wait for more feedback from Giacomo or Markus.

crates/walrus-service/bin/deploy.rs
```diff
@@ -578,13 +622,77 @@ mod commands {
...
+                .ok();
+            let (compiled_package, _build_config) =
+                compile_package(contract_dir, Default::default(), chain_id).await?;
+
+            let sender = sender.unwrap_or(contract_client.address());
+            let mut pt_builder = WalrusPtbBuilder::new(contract_client.read_client, sender);
+
+            pt_builder
+                .custom_walrus_upgrade(upgrade_manager_object_id, compiled_package, upgrade_type)
+                .await?;
```
mlegner, line 638: Nit/suggestion: This part has a lot of overlap with the [`SuiContractClient::upgrade`](https://github.com/MystenLabs/walrus/blob/3cb804cb54e6dd70ed58356273e8961ecf51be52/crates/walrus-sui/src/client.rs#L1983) function, just without sending the transaction and upgrading the lock file. Could we just add a flag to that function instead of duplicating the code here?

crates/walrus-service/bin/deploy.rs
```diff
@@ -578,13 +622,77 @@ mod commands {
...
+            new_package_id,
+        }: MigrateArgs,
+    ) -> anyhow::Result<()> {
+        utils::init_tracing_subscriber()?;
+
+        let wallet =
+            load_wallet_context_from_path(wallet_path, None).context("unable to load wallet")?;
+        let contract_config = ContractConfig::new(system_object_id, staking_object_id);
+
+        #[allow(deprecated)]
```
mlegner: Nit: This should no longer be necessary after https://github.com/MystenLabs/walrus/pull/2145.
```suggestion
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1383,6 +1374,54 @@ impl WalrusPtbBuilder {
...
+            self.authorize_upgrade(upgrade_manager, &digest).await?
+        };
+
+        // Execute the upgrade.
+        let modules = compiled_package.get_package_bytes(false);
+        let upgrade_receipt_arg = self.pt_builder.upgrade(
+            self.read_client.get_system_package_id(),
+            upgrade_ticket_arg,
+            compiled_package
+                .dependency_ids // TODO: do we need the dependencies as returned by `compile`?
```
mlegner: Question: What is the state of this TODO? Can this be removed or do we need some follow-up work?

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1392,14 +1431,67 @@ impl WalrusPtbBuilder {
...
+        } else {
+            let tx_kind =
+                TransactionKind::ProgrammableTransaction(programmable_transaction.clone());
+            self.read_client
+                .sui_client()
+                .estimate_gas_budget(self.sender_address, tx_kind, gas_price)
+                .await?
+        };
+
+        let min_gas_coin_balance = minimum_gas_coin_balance.max(gas_budget + self.tx_sui_cost);
```
mlegner: Nit: You could also name the new variable `minimum_gas_coin_balance` or define the `minimum_gas_coin_balance` as `mut` and overwrite it.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1392,14 +1431,67 @@ impl WalrusPtbBuilder {
...
+    pub(crate) async fn build_transaction_data_with_min_gas_balance(
+        mut self,
+        gas_budget: Option<u64>,
+        minimum_gas_coin_balance: u64,
+    ) -> SuiClientResult<TransactionData> {
+        self.transfer_remaining_outputs(None).await?;
+        let programmable_transaction = self.pt_builder.finish();
+
+        // Get the current gas price from the network
+        let gas_price = self.read_client.get_reference_gas_price().await?;
```
mlegner, line 1468: Question (unrelated to this PR): Do we do this for all transactions on the client? And is my understanding correct that this causes an additional roundtrip to the RPC node for each transaction? If so, it could be beneficial to cache this somehow.

crates/walrus-sui/src/client.rs
mlegner, line 1: Remark: AFAICT the changes to this file are essentially only because of the transaction finalization refactoring and functions moved elsewhere. Let me know if I missed anything important.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: Ah, I forgot to remove that. That was a comment for myself while I was refactoring this.

crates/walrus-sui/src/client.rs
karlwuest: Yes, that's correct.

crates/walrus-service/bin/deploy.rs
karlwuest: I thought about that, but that would require `upgrade` to take the sender as additional argument + the flag, and return an enum with either the result of the actual upgrade (ObjectId) or the serialized transaction. Adding these case distinctions there adds more complexity compared to duplicating a bit of code here, imho.

Lmk if you feel strongly about this.

crates/walrus-sui/src/client/transaction_builder.rs
karlwuest: Yes, we do, and yes, it would be good to cache this. We already have an issue for this (WAL-512), I'll add a TODO here with a reference to the issue.

PR #2141, opened 2025-05-27
chore: remove vscode settings because they conflict across editors

wbbradley:
## Description

.vscode/settings was breaking Neovim integration on certain plugins when comments and extraneous comments were added. Vscode itself didn't have that problem. Easiest solution is to just remove it to avoid conflicts across editors/plugins.

## Test plan

Manual testing.


halfprice (approved):

wbbradley:
> Thanks for the fix, @wbbradley! Maybe we can instead include it as a template/example and/or inline in the CONTRIBUTING.md?

Good call. I'll put that together.

PR #2143, opened 2025-05-28
feat: Make rpc client fallback on single checkpoint failures

sadhansood:
## Description

The current implementation doesn't track single checkpoint failures for archival fallback as long as other checkpoints are successfully getting downloaded from the full node. This PR tries to fix that by adding sequence-specific failure tracking that persists until the problematic checkpoint succeeds:
- Only start tracking a new checkpoint's failures if we're not already tracking another one
- Reset failure metrics only for the specific checkpoint that succeeded
- Added `last_failed` to fallback eligibility check to provide more context

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client/download_handler.rs
+++ b/crates/walrus-sui/src/client/retry_client/download_handler.rs
@@ -1,0 +1,471 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+//! Implements a checkpoint download handler for the RPC client.
+use std::{
+    collections::HashMap,
+    sync::{
+        Arc,
+        Mutex,
+        atomic::{AtomicUsize, Ordering},
+    },
+    time::Instant,
+};
+
+use atomic_time::AtomicInstant;
+
+use crate::client::{
+    SuiClientMetricSet,
+    retry_client::RetriableClientError,
+    rpc_config::RpcFallbackConfig,
+};
+
+/// The decision to make where to download a checkpoint.
+#[derive(Debug)]
+pub enum CheckpointDownloadDecision {
+    /// Attempt primary RPC first.
+    AttemptPrimary,
+    /// Direct fallback for the checkpoint.
+    DirectFallback,
+    /// Try fallback if primary fails.
+    FallbackUponPrimaryFailure(RetriableClientError),
+    /// Use primary error if fallback fails.
+    UsePrimaryError(RetriableClientError),
+}
+
+/// Updated failure stats for the current sequence number.
+#[derive(Debug)]
+pub struct CheckpointFailureStats {
+    /// Overall number of failures since last_success.
+    /// Note: This is not the number of failures for the current sequence number.
+    /// This is the number of failures since the last successful RPC call.
+    overall_failures_count: usize,
+    /// Number of consecutive failures for the current sequence number.
+    consecutive_failures_for_this_seq: usize,
+}
+
+/// Checkpoint download handler for the RPC client.
+pub struct CheckpointDownloadHandler {
+    // Configuration for fallback.
+    config: Option<RpcFallbackConfig>,
+    // Overall last success time from primary RPC.
+    last_success: Arc<AtomicInstant>,
+    // Overall number of failures since last_success.
+    num_failures: Arc<AtomicUsize>,
+    // When to skip primary due to overall failures.
+    skip_primary_until: Arc<AtomicInstant>,
+    // Map of checkpoint sequence numbers to their failure counts.
+    checkpoint_failures: Arc<Mutex<HashMap<u64, usize>>>,
+    // Metrics for the client.
+    metrics: Option<Arc<SuiClientMetricSet>>,
+}
+
+impl std::fmt::Debug for CheckpointDownloadHandler {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "CheckpointDownloadHandler {{ config: {:?}, last_success: {:?}, \
+            num_failures: {:?}, skip_primary_until: {:?}, \
+            checkpoint_failures: {:?} }}",
+            self.config,
+            self.last_success.load(Ordering::Relaxed),
+            self.num_failures.load(Ordering::Relaxed),
+            self.skip_primary_until.load(Ordering::Relaxed),
+            self.checkpoint_failures
+                .lock()
+                .expect("Mutex should not be poisoned")
+        )
+    }
+}
+
+impl CheckpointDownloadHandler {
+    /// Creates a new FallbackHandler with the given configuration.
+    pub fn new(
+        config: Option<RpcFallbackConfig>,
+        metrics: Option<Arc<SuiClientMetricSet>>,
+    ) -> Self {
+        Self {
+            config: config.clone(),
+            last_success: Arc::new(AtomicInstant::new(Instant::now())),
+            num_failures: Arc::new(AtomicUsize::new(0)),
+            skip_primary_until: Arc::new(AtomicInstant::new(Instant::now())),
+            checkpoint_failures: Arc::new(Mutex::new(HashMap::new())),
+            metrics,
+        }
+    }
+
+    /// Called on primary success for a specific sequence_number.
+    pub fn record_primary_success(&self, sequence_number: u64) {
+        self.last_success.store(Instant::now(), Ordering::Relaxed);
+        self.num_failures.store(0, Ordering::Relaxed);
+
+        // Remove this checkpoint from the failures map.
+        let mut failures = self
+            .checkpoint_failures
+            .lock()
+            .expect("Mutex should not be poisoned");
+        failures.remove(&sequence_number);
+    }
+
+    /// Called on fallback success for a specific sequence_number.
+    pub fn record_fallback_success_for_sequence(&self, sequence_number: u64) {
+        // Remove this checkpoint from the failures map.
+        let mut failures = self
+            .checkpoint_failures
+            .lock()
+            .expect("Mutex should not be poisoned");
+        failures.remove(&sequence_number);
+    }
+
+    /// Determines the initial action for fetching a checkpoint.
+    pub fn initial_download_decision(&self) -> CheckpointDownloadDecision {
+        if self.config.is_none() {
+            if let Some(metrics) = &self.metrics {
+                metrics.record_checkpoint_download_source("primary_only_no_fallback_config");
+            }
+            return CheckpointDownloadDecision::AttemptPrimary;
+        }
+        if Instant::now() >= self.skip_primary_until.load(Ordering::Relaxed) {
+            if let Some(metrics) = &self.metrics {
+                metrics.record_checkpoint_download_source("primary");
+            }
+            CheckpointDownloadDecision::AttemptPrimary
+        } else {
+            if let Some(metrics) = &self.metrics {
+                metrics.record_checkpoint_download_source("direct_fallback");
+            }
+            tracing::debug!(
+                "Skipping primary RPC (overall policy) until {:?}, attempting fallback directly.",
+                self.skip_primary_until.load(Ordering::Relaxed)
+            );
+            CheckpointDownloadDecision::DirectFallback
+        }
+    }
+
+    /// Updates all relevant failure counters based on the current failure
+    /// and returns a snapshot of the counts relevant to the decision.
+    /// Prioritizes tracking failures for smaller sequence numbers.
+    pub fn update_and_get_failure_stats(&self, sequence_number: u64) -> CheckpointFailureStats {
+        let overall_failures_count = self.num_failures.fetch_add(1, Ordering::Relaxed) + 1;
+
+        let mut failures = self
+            .checkpoint_failures
+            .lock()
+            .expect("Mutex should not be poisoned");
+        let consecutive_failures_for_this_seq = failures
+            .entry(sequence_number)
+            .and_modify(|count| *count += 1)
+            .or_insert(1);
+
+        CheckpointFailureStats {
+            overall_failures_count,
+            consecutive_failures_for_this_seq: *consecutive_failures_for_this_seq,
+        }
+    }
+
+    /// Assesses the primary checkpoint failure and returns a download decision
+    /// based on the failure stats.
+    /// This function is called when the primary RPC fails to fetch a checkpoint.
+    pub fn decision_upon_primary_failure(
+        &self,
+        sequence_number: u64,
+        primary_error: RetriableClientError,
+    ) -> CheckpointDownloadDecision {
+        let Some(ref cfg) = self.config else {
+            // If no fallback config is provided, we will return the primary error.
+            if let Some(metrics) = &self.metrics {
+                metrics.record_checkpoint_download_source("primary_error_no_fallback_config");
+            }
+            return CheckpointDownloadDecision::UsePrimaryError(primary_error);
+        };
+
+        // Update all failure counters and get their current relevant values.
+        let failure_stats = self.update_and_get_failure_stats(sequence_number);
+        let last_success_time = self.last_success.load(Ordering::Relaxed);
+
+        tracing::debug!(
+            seq = sequence_number, error = ?primary_error,
+            stats.overall_failures = failure_stats.overall_failures_count,
+            stats.consecutive_for_this_seq = failure_stats.consecutive_failures_for_this_seq,
+            stats.last_success_elapsed = ?last_success_time.elapsed(),
+            "Primary failure stats updated and assessed."
+        );
+
+        // Make fallback decisions based on the updated stats.
+        if primary_error.is_eligible_for_fallback_immediately(sequence_number) {
+            tracing::debug!(
+                seq = sequence_number,
+                "Error eligible for immediate fallback."
+            );
+            if let Some(metrics) = &self.metrics {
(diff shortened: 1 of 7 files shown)


crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -373,25 +378,52 @@ impl RetriableRpcClient {
...
+
+            let last_failed = self.last_failed_checkpoint.load(Ordering::Relaxed);
+            if last_failed == sequence_number {
+                self.consecutive_failures.fetch_add(1, Ordering::Relaxed);
+            } else if last_failed == 0 {
+                // Only start tracking consecutive failures for a sequence number if we are
+                // not already tracking it for another sequence number.
+                self.consecutive_failures.store(1, Ordering::Relaxed);
+                self.last_failed_checkpoint
+                    .store(sequence_number, Ordering::Relaxed);
```
halfprice: If there are multiple downloader experiencing errors, there is a race updating `last_failed_checkpoint`. We probably should store the lowest one.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -193,6 +193,8 @@ pub struct RetriableRpcClient {
     last_success: Arc<AtomicInstant>,
     num_failures: Arc<AtomicUsize>,
     skip_rpc_for_checkpoint_until: Arc<AtomicInstant>,
+    last_failed_checkpoint: Arc<AtomicU64>,
```
liquid-helium: Remark: Is RetriableRpcClient shared by different workers? It seems that each caller owns a RetriableRpcClient.
If there would multiple checkpoints being downloaded with the same RetriableRpcClient, would a hashMap<sequence_number, failure_count> make it easier?

sadhansood: Thanks @liquid-helium for your feedback. I am trying to do a slightly bigger refactor to address such issues. But we should be able to have some shareable state across all instances of the client. I will update this PR in next couple days.

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,599 @@
...
+
+/// The decision to make where to download a checkpoint.
+#[derive(Debug)]
+pub enum CheckpointDownloadDecision {
+    /// Attempt primary RPC first
+    AttemptPrimary,
+    /// Try fallback if primary fails
+    TryFallback(Option<RetriableClientError>),
+    /// Use primary error if fallback fails
+    UsePrimaryError(RetriableClientError),
```
wbbradley, line 33: It looks like this case always leads to `unreachable!`. If that is the case, it might be beneficial to simply `panic!` upstream at the point where we would be returning this value anyway.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -362,111 +348,67 @@ impl RetriableRpcClient {
...
-                    last_success.elapsed(),
-                    num_failures
-                );
-                return Err(error);
+        match self.download_handler.download_checkpoint_decision() {
+            CheckpointDownloadDecision::AttemptPrimary => {
+                self.handle_primary_attempt(sequence_number).await
+            }
+            CheckpointDownloadDecision::TryFallback(_) => {
+                self.handle_fallback_attempt(sequence_number).await
```
wbbradley, line 359: Should we be logging the error in this case? (Currently it looks like we are ignoring it via `_` in the pattern.)

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -362,111 +348,67 @@ impl RetriableRpcClient {
...
+                    .record_primary_success(sequence_number);
+                Ok(checkpoint)
+            }
+            Err(primary_error) => {
+                match self
+                    .download_handler
+                    .handle_primary_checkpoint_failure(sequence_number, primary_error)
+                {
+                    CheckpointDownloadDecision::TryFallback(Some(_)) => {
+                        self.handle_fallback_attempt(sequence_number).await
```
wbbradley, line 393: Similarly here I wonder if it would be helpful to log the error? If we decide to always ignore the error, then we could remove it from the `CheckpointDownloadDecision::TryFallback` enum value.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -362,111 +348,67 @@ impl RetriableRpcClient {
...
+                    }
+                    CheckpointDownloadDecision::UsePrimaryError(err) => Err(err),
+                    _ => unreachable!("Invalid decision after primary failure assessment"),
```
wbbradley: This is a nit, but my personal preference here would be to avoid irrefutable patterns `_` so that if/when we add another case, we don't accidentally skip handling it here. ie: it could be beneficial to be explicit as to which case we are marking as unreachable, that is `CheckpointDownloadDecision::AttemptPrimary`.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -362,111 +348,67 @@ impl RetriableRpcClient {
...
+        let result = fb_client
+            .get_full_checkpoint(sequence_number)
+            .await
+            .map_err(RetriableClientError::from);
+
+        if result.is_ok() {
+            self.download_handler
+                .record_fallback_success_for_sequence(sequence_number);
+        }
+        result
```
wbbradley:
```suggestion
        Ok(fb_client
            .get_full_checkpoint(sequence_number)
            .await
            .inspect(|_| {
                self.download_handler
                    .record_fallback_success_for_sequence(sequence_number);
            })?)
```
nit: just reflowing this code a bit to avoid extra control flow logic

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,599 @@
...
+                    .is_ok()
+                {
+                    // CAS succeeded and we're now tracking this sequence
+                    self.current_consecutive_failures_for_tracked_checkpoint
+                        .store(1, Ordering::Relaxed);
+                    break (1, true);
+                } else {
+                    // CAS failed as another thread started tracking a sequence.
+                    // We will loop to re-evaluate with the new state
+                    continue;
```
wbbradley: I'm not 100% certain, but it feels like you can just return `(0, false)` here? Unless maybe there are other places that are updating `last_tracked_failed_checkpoint`? Seems unlikely since we know that `last_tracked_failed_checkpoint` just changed from `0` to something.. seems unlikely that it'd go straight back to `0` after a few nanoseconds, since (I assume) there are network requests involved here...

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,599 @@
...
+                }
+            } else if sequence_number < current_tracked_seq {
```
wbbradley: Seems like the loop could be here instead of the outer location. I think this is the only case where we'd want to loop.

```
...
else if sequence_number > current_tracked_seq {
    return (0, false);
} else loop {
    ...
}
```
Then we could eliminate the outer loop?

wbbradley (commented):
Looks good! I left a few comments about minor cleanup stuff. 

Overall I don't know that this policy is the "best" or most optimal, but it seems to make sense in terms of what it's trying to do. One thing we could try in a future version would be maintaining a canary set that is - a small percentage of the workers - always attempting the primary just to get signal, which could then be used to switch back to the primary if it shows promise. But, that's just a random idea. 

Let me know your thoughts on this feedback. Thank you!

crates/walrus-sui/src/client/retry_client/download_handler.rs
sadhansood: It is not always the case, in `handle_primary_attempt` we do return the error contained in `UsePrimaryError`

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,606 @@
...
+
+use crate::client::{retry_client::RetriableClientError, rpc_config::RpcFallbackConfig};
+
+/// The decision to make where to download a checkpoint.
+#[derive(Debug)]
+pub enum CheckpointDownloadDecision {
+    /// Attempt primary RPC first
+    AttemptPrimary,
+    /// Try fallback if primary fails
+    TryFallback(Option<RetriableClientError>),
```
halfprice: Suggestion: since TryFallback(None) is only used in the initial decision, and TryFallback(err) is only used in the primary fallback, maybe we can separate them into two case: UseFallbackDirectly and TryFallbackAfterPrimaryFailure(err)

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,606 @@
...
+    }
+
+    /// Updates all relevant failure counters based on the current failure
+    /// and returns a snapshot of the counts relevant to the decision.
+    /// Prioritizes tracking failures for smaller sequence numbers.
+    pub fn update_and_get_failure_stats(&self, sequence_number: u64) -> CheckpointFailureStats {
+        let overall_failures_count = self.num_failures.fetch_add(1, Ordering::Relaxed) + 1;
+
+        let (consecutive_failures_for_this_seq, is_this_sequence_now_tracked) = loop {
+            let current_tracked_seq = self.last_tracked_failed_checkpoint.load(Ordering::Acquire);
```
halfprice: Question: multiple orderings are used in this function, and how do you decide which level to use in each operation?

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,606 @@
...
+        }
+    }
+
+    /// Called on primary success for a specific sequence_number
+    pub fn record_primary_success(&self, sequence_number: u64) {
+        self.last_success.store(Instant::now(), Ordering::Relaxed);
+        self.num_failures.store(0, Ordering::Relaxed);
+
+        // Reset consecutive failure tracking if this was the specific checkpoint being tracked
+        if self.last_tracked_failed_checkpoint.load(Ordering::Relaxed) == sequence_number {
```
halfprice: Question: I'm not too familiar with all the atomic op ordering sematics. One thing I don't know if possible is that if for some reason, `sequence_number` is the smallest sequence number across all the downloaders, but self.last_tracked_failed_checkpoint.load(Ordering::Relaxed) loads a different checkpoint number larger (again, this may not be possible, and I'm just not too sure what `Relaxed` ordering's effects), then we will never clear the current sequence number.

For example,

Download 2 failed, update last_tracked_failed_checkpoint to 2
Download 1 failed, update last_tracked_failed_checkpoint to 1
Download 1 success, reset last_tracked_failed_checkpoint, but for some reason, here we load 2.

Then we may never clear 1, and all future single checkpoint will never be tracked.

Let me know this is definitely not possible.

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,606 @@
...
+
+/// Updated failure stats for the current sequence number
+#[derive(Debug)]
+pub struct CheckpointFailureStats {
+    /// Overall number of failures since last_success
+    /// Note: This is not the number of failures for the current sequence number
+    /// This is the number of failures since the last successful RPC call
+    overall_failures_count: usize,
+    /// Number of consecutive failures for the current sequence number
+    consecutive_failures_for_this_seq: usize,
```
halfprice, line 44: Suggestion: IIUC, say upon 3 failures, a single checkpoint will fallback to archival, if there are consecutive failures, second may need to fail 6 times, and 9 may need to failre 9 times to fallback. This is a bit unintuitive.

What do you think of tracking BTreeMap<SequenceNumber, NumberOfFailure>? In this way, this also gives a more clear picture what's the lowest failure seq, and what's the failure count for each seq. Given that the number of checkpoint downloader is limited, the size of thie BTreeMap is also limited.

halfprice (commented):
Amazing! Thanks a lot @sadhansood and I enjoyed reading this PR a lot, and the logic now is much clearer. I have some initial comments and will take look at it later again and also the tests.

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,606 @@
...
+    time::Instant,
+};
+
+use atomic_time::AtomicInstant;
+
+use crate::client::{retry_client::RetriableClientError, rpc_config::RpcFallbackConfig};
+
+/// The decision to make where to download a checkpoint.
+#[derive(Debug)]
+pub enum CheckpointDownloadDecision {
```
liquid-helium, line 25: Praise: I like this struct, making code much cleaner!

crates/walrus-sui/src/client/retry_client/download_handler.rs
liquid-helium: +1 on keeping the individual failing sequence number to each worker. I still like more your previous approach of keeping that failing sequence number local to each worker :)

wbbradley: I agree in that I think it makes sense to use a structure like Zhe describes above. 

Maybe it could be worth it to wrap that in a Mutex and do the bookkeeping that way?

I totally might be misunderstanding this, so correct me if I'm wrong: I don't foresee lock contention being an issue here since the workers are going to be hitting the network between attempts to lock that structure to analyze what they should do next. 

I guess another way of thinking about it is that reasoning about a multiplicity of atomic ints is pretty hard, whereas putting state behind a single mutex is relatively trivial to reason about, and barring any unforeseen problems, should not result in high switching costs or latency, I think?

crates/walrus-sui/src/client/retry_client/download_handler.rs
wbbradley: Ah, I see that now, thanks for pointing that out. Cool!

crates/walrus-sui/src/client/retry_client/download_handler.rs
sadhansood: I'm fine with storing state per failing checkpoint sequence number since the total number of inflight checkpoints is never going to be too much. We definitely need to wrap the `BTreeMap` with a mutex since we will be concurrently reading/writing from it. I think that removes the need to do any atomics as well which simplifies things a bit as well.

crates/walrus-sui/src/client/retry_client/download_handler.rs
sadhansood: I think the scenario might be possible, and it does create a problem that you describe. Probably better to switch to `Ordering::SeqCst` or `Acquire/Release` kind of semantics b/w write and read

crates/walrus-sui/src/client/retry_client/download_handler.rs
sadhansood: I think this was to ensure that updates made to `last_tracked_failed_checkpoint` by other threads during CAS are visible during this load, otherwise with `Ordering::Relaxed` I don't think we get any such guarantee

sadhansood (commented):
Thanks @halfprice @liquid-helium @wbbradley for your review - looks like the consensus is to track failure state for all failing checkpoints and we can do away with complicated (but fun!) atomic orderings. I will go ahead and make the updates to the PR. Lmk if you have other suggestions.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -362,112 +349,86 @@ impl RetriableRpcClient {
...
+                {
+                    CheckpointDownloadDecision::FallbackUponPrimaryFailure(err) => {
+                        tracing_sampled::error!(
+                            "30s",
+                            ?err,
+                            "Rpc fallback for checkpoint {}",
+                            sequence_number
+                        );
+                        self.handle_fallback_attempt(sequence_number).await
+                    }
```
liquid-helium, line 394: Question: Do we already have metrics on the number of fallbacks? If not, consider adding one for easy debugging.

crates/walrus-sui/src/client/retry_client/download_handler.rs
```diff
@@ -0,0 +1,433 @@
...
+            );
+            return CheckpointDownloadDecision::FallbackUponPrimaryFailure(primary_error);
+        }
+
+        // Overall failure window exceeded
+        let is_overall_window_exceeded = failure_stats.overall_failures_count
+            >= cfg.min_failures_to_start_fallback
+            && last_success_time.elapsed() >= cfg.failure_window_to_start_fallback_duration;
+
+        // Consecutive failures for this specific checkpoint limit met
```
liquid-helium: nit: add a period at the end of each comment to match the codebase.

crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
sadhansood: Added

crates/walrus-sui/src/client/retry_client/download_handler.rs
sadhansood: done

PR #2146, opened 2025-05-29
feat: quilt encoding part iii

liquid-helium:
## Description

Added SliverIndexApi, QuiltPathApi, and QuiltPatchInternalIdApi.

Added QuiltPathId, which can be used in the future to read quilt patch from Walrus.

Allow the quilt index to occupy up to 10 slivers.

Contributes to WAL-858

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-core/src/encoding.rs
```diff
@@ -15,8 +15,8 @@ pub use basic_encoding::{
 mod blob_encoding;
 pub use blob_encoding::{BlobDecoder, BlobDecoderEnum, BlobEncoder};
 
-mod quilt_encoding;
-pub use quilt_encoding::{QuiltDecoderV1, QuiltEncoderV1, QuiltV1};
+/// Quilt encoding.
```
mlegner: Nit: We normally put module docstrings inside the respective module with `//!`.

crates/walrus-core/src/metadata.rs
```diff
@@ -103,6 +121,93 @@ pub enum QuiltIndex {
...
+/// Definition of the layout of the quilt patch internal id in QuiltVersionV1.
+///
+/// ```text
+///┌─────────────┬──────────────────┬──────────────────┐
+///│    Byte 0   │     Byte 1-2     │     Byte 3-4     │
+///├─────────────┼──────────────────┼──────────────────┤
+///│  Version    │   start_index    │    end_index     │
+///│    Byte     │   (16 bits LE)   │   (16 bits LE)   │
+///├─────────────┼──────────────────┼──────────────────┤
+///│     8 bits  │     16 bits      │     16 bits      │
```
mlegner: Nit: This doesn't really add any information and could be removed.
```suggestion
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -98,6 +112,52 @@ pub trait QuiltApi<V: QuiltVersion> {
     fn symbol_size(&self) -> usize;
 }
 
+/// API for QuiltIndex.
+pub trait QuiltIndexApi<V: QuiltVersion>: Clone + Into<QuiltIndex> {
+    /// Returns the quilt patch by its identifier.
```
mlegner, line 123: Question: What happens when there are duplicate identifiers? Same question for `get_sliver_indices_by_identifiers`.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -128,13 +194,13 @@ pub trait QuiltDecoderApi<'a, V: QuiltVersion> {
     ///
     /// The decoded quilt index is stored in the decoder and can be retrieved
     /// using the `get_quilt_index` method after this method returns.
-    fn get_or_decode_quilt_index(&mut self) -> Result<&V::QuiltIndex, QuiltError>;
+    fn get_or_decode_quilt_index(&mut self) -> Result<QuiltIndex, QuiltError>;
 
     /// Gets a blob by its identifier from the quilt.
     fn get_blob_by_identifier(&self, identifier: &str) -> Result<QuiltStoreBlobOwned, QuiltError>;
```
mlegner, line 211: Question: What happens here if there are duplicated identifiers?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -1146,9 +1238,14 @@ pub struct QuiltDecoderV1<'a> {
...
+    fn get_or_decode_quilt_index(&mut self) -> Result<QuiltIndex, QuiltError> {
         if self.quilt_index.is_some() {
-            return Ok(self.quilt_index.as_ref().expect("quilt index should exist"));
+            return Ok(self
+                .quilt_index
+                .as_ref()
+                .expect("quilt index should exist")
+                .clone()
+                .into());
         }
```
mlegner: Nit (pre-existing):
```rust
        if let Some(quilt_index) = self.quilt_index.as_ref() {
            return Ok(quilt_index.clone().into());
        }
```

And a few lines below:
```rust
        let quilt_index = QuiltVersionV1::decode_quilt_index(self, column_size)?;
        self.quilt_index = Some(quilt_index.clone());
        Ok(quilt_index.into())
```

(Similar case in `QuiltV1::get_or_decode_quilt_index` above.)

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -1146,9 +1238,14 @@ pub struct QuiltDecoderV1<'a> {
 }
 
 impl<'a> QuiltDecoderApi<'a, QuiltVersionV1> for QuiltDecoderV1<'a> {
-    fn get_or_decode_quilt_index(&mut self) -> Result<&QuiltIndexV1, QuiltError> {
+    fn get_or_decode_quilt_index(&mut self) -> Result<QuiltIndex, QuiltError> {
```
mlegner, line 1262: Nit: The whole function seems to be very similar to `QuiltV1::get_or_decode_quilt_index`.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -1670,6 +1773,8 @@ mod tests {
         max_blob_size: usize,
         n_shards: u16,
     ) {
+        let _guard = tracing_subscriber::fmt().try_init();
```
mlegner: Nit: This is not a guard but an error, so you can also assign to `_`.
```suggestion
        let _ = tracing_subscriber::fmt().try_init();
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -98,6 +112,52 @@ pub trait QuiltApi<V: QuiltVersion> {
...
+/// A quilt patch internal id is a unique identifier for a quilt patch.
+/// It can be used to identify the quilt patch in a quilt, and when combined with the quilt
+/// id, it can be used to identify the quilt patch in Walrus.
+pub trait QuiltPatchInternalIdApi: Clone {
+    /// The serialized bytes of the quilt patch internal id.
+    fn to_bytes(&self) -> Vec<u8>;
+
+    /// Deserializes the quilt patch internal id from bytes.
+    fn from_bytes(bytes: &[u8]) -> Result<Self, QuiltError>;
+}
```
mlegner, line 165: Question: Do we actually have to expose this publicly? Could we include this into the `QuiltPatchApi`, replacing the `quilt_patch_internal_id` function?

crates/walrus-core/src/metadata.rs
```diff
@@ -114,36 +219,55 @@ pub struct QuiltIndexV1 {
...
-    /// Returns an iterator over the identifiers of the blobs in the quilt.
-    pub fn identifiers(&self) -> impl Iterator<Item = &str> {
+    fn get_sliver_indices_by_identifiers(
+        &self,
+        identifiers: &[&str],
+    ) -> Result<Vec<SliverIndex>, QuiltError> {
+        let patches = identifiers
+            .iter()
+            .map(|identifier| self.get_quilt_patch_by_identifier(identifier))
+            .collect::<Result<Vec<_>, _>>()?;
```
mlegner: Suggestion: This doesn't feel very efficient to me: Effectively we go through the index once for each identifier. How about storing the identifiers in a `HashSet`, then going through the index once, checking each patch against the set of identifiers? If we want the same behavior where we only take the first in case of duplicate identifiers, we can remove the identifiers from the set after we found a patch.

mlegner (commented):
Thanks so much for the addition, and sorry (yet again) for my late review. A few mostly minor suggestions and questions.

crates/walrus-core/src/encoding.rs
liquid-helium: Thanks for the tip, change applied.

crates/walrus-core/src/metadata.rs
liquid-helium: True, removed.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good point!
Added logic to return an error if there are duplicated identifiers.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Now an check is added in the Encoder to ensure there are no duplicates.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Thanks, change applied, adjusted a little bit for QuiltV1::get_or_decode_quilt_index due to the borrow rules.

crates/walrus-core/src/encoding/quilt_encoding.rs

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: This is used in the `QuiltPathId`, which is independent of QuiltVersion, and is a concatenation of `BlobId + serialized QuiltPatchInternalId`.
A user can read a blob from Walrus with its `QuiltPathId`, by decoding the `QuiltPatchInternalId` part into internal indices in the Quilt, which identified by the `BlobId` part.

I added this standalone trait for the internal id for the above use case, but I am open to other approaches as well.

crates/walrus-core/src/metadata.rs
liquid-helium: Good idea! Change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice catch!

The differences are really subtle:
1. For Quilt, we assume that it has all the quilt data, so we don't really need to check the missing data, while for Decoder, we need to do some extra check.
2. QuiltDecoderApi returns QuiltIndex, so that in the read path, we don't need to worry about the QuiltVersion.

They reuse the same `QuiltVersionV1::decode_quilt_index` for the actual decoding.
I tried to dedupe it more, but haven't found any satisfying approaches, but open to any suggestions!

crates/walrus-core/src/encoding/quilt_encoding.rs
mlegner: Remark: This covers the likely scenario where the quilt is constructed with this library. However, theoretically, it's still possible to store quilts that have duplicated identifiers by constructing them "manually". So I think we should document what happens during decoding as well, see my separate comment.

crates/walrus-core/src/encoding/quilt_encoding.rs
mlegner: Hmm... I'm just slightly losing track of the large number of traits we have now. But I guess it makes sense, let's keep it like this.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -179,6 +181,10 @@ pub trait QuiltConfigApi<'a, V: QuiltVersion> {
 /// Encoder to construct a quilt and encode the blobs into slivers.
 pub trait QuiltEncoderApi<V: QuiltVersion> {
     /// Constructs a quilt by encoding the blobs.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the blobs have duplicate identifiers.
```
mlegner: Question: Can it also return errors in other cases (I guess so). Maybe mention this as well.

crates/walrus-core/src/metadata.rs
```diff
@@ -231,15 +230,18 @@ impl QuiltIndexApi<QuiltVersionV1> for QuiltIndexV1 {
...
+        let identifiers: HashSet<&str> = identifiers.iter().copied().collect();
+        let patches = self
+            .quilt_patches
             .iter()
-            .map(|identifier| self.get_quilt_patch_by_identifier(identifier))
-            .collect::<Result<Vec<_>, _>>()?;
-
-        Ok(patches
+            .filter(|patch| identifiers.contains(&patch.identifier.as_str()))
+            .collect::<Vec<_>>();
```
mlegner: Remark: Here we allow duplicate identifiers and just include all matching patches in that case. This makes sense to me, but I think we should document that behavior. For example:

> If the quilt contains duplicate identifiers, all matching patches are returned.

Similarly, for the `get_quilt_patch_by_identifier` function above, I'd add something like:

> If the quilt contains duplicate identifiers, the *first* matching patch is returned.

mlegner (approved):
Thanks for the changes and insights! 🙏 
LGTM.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: same here, hope the number won't increase 🤞

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good point, I added comments
```
/// If there are duplicate identifiers, the first one in the sort order will be returned.
/// Note that the sort could be unstable.
```

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: I reword it a little bit to avoid enumerate all possible errors.

crates/walrus-core/src/metadata.rs
liquid-helium: Nice catch, change applied.

PR #2147, opened 2025-05-29
feat: Quilt client part I

liquid-helium:
## Description

Added apis to retrieve slivers.

Contributes to WAL-858
## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-sdk/Cargo.toml
```diff
@@ -12,6 +12,7 @@ test-utils = [
 
 [dependencies]
 anyhow.workspace = true
+bimap = "0.6.3"
```
mlegner: Nit: Can we move the version to the global `Cargo.toml` file?

crates/walrus-sdk/src/client.rs
```diff
@@ -88,6 +92,55 @@ pub mod refresh;
 pub mod resource;
 pub mod responses;
 
+/// A set of slivers to be retrieved from Walrus.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub(crate) struct SliverSelector<E: EncodingAxis> {
+    indices_and_shards: BiMap<SliverIndex, ShardIndex>,
```
mlegner, line 102: Praise: I like this!

crates/walrus-sdk/src/client.rs
```diff
@@ -88,6 +92,55 @@ pub mod refresh;
...
+    _phantom: PhantomData<E>,
+}
+
+#[allow(unused)]
+impl<E: EncodingAxis> SliverSelector<E> {
+    /// Creates a new sliver selector.
+    pub fn new(sliver_indices: Vec<SliverIndex>, n_shards: NonZeroU16, blob_id: &BlobId) -> Self {
```
mlegner: Nit: Since we immediately turn the `sliver_indices` into an iterator, we could just accept `&[SliverIndex]` or `impl IntoIterator<Item = SliverIndex>`.

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
             result = future => result,
         }
     }
+
+    /// Retrieves slivers with retry logic, only requesting missing slivers in subsequent attempts.
+    ///
+    /// This function will keep retrying until all requested slivers are received or the maximum
+    /// number of retries is reached or the timeout is reached.
+    async fn retrieve_slivers_with_retry<E: EncodingAxis>(
```
mlegner, line 525: Remark: I'm wondering why Clippy doesn't complain about this private unused function... 🤔

crates/walrus-sdk/src/client.rs
```diff
@@ -88,6 +92,55 @@ pub mod refresh;
...
+    pub fn is_empty(&self) -> bool {
+        self.indices_and_shards.is_empty()
+    }
+
+    /// Returns true if the shard contains one of the slivers.
+    pub fn should_read_from_shard(&self, shard_index: &ShardIndex) -> bool {
+        self.indices_and_shards.contains_right(shard_index)
+    }
+
+    /// Sliver is complete.
```
mlegner: Nit: Can you document what this function returns?

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+        sliver_indices: &[SliverIndex],
+        certified_epoch: Epoch,
+        max_attempts: Option<usize>,
+        timeout_duration: Option<Duration>,
+    ) -> Result<Vec<SliverData<E>>, ClientError>
+    where
+        SliverData<E>: TryFrom<walrus_core::Sliver>,
+    {
+        let mut sliver_selector = SliverSelector::<E>::new(
+            sliver_indices.to_vec(),
```
mlegner: Nit: If you change the `SliverSelector::new` function, the `to_vec` should not be necessary.

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+            metadata.n_shards(),
+            metadata.blob_id(),
+        );
+        let mut attempts = 0;
+        let retry_delay_ms = 100;
```
mlegner: Nit: Can you either make this a constant or a config parameter? Also, this could be a `Duration` directly.

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+        let start_time = Instant::now();
+        let mut last_error: Option<ClientError> = None;
+
+        while !sliver_selector.is_empty() {
+            // Check if we've exceeded the timeout or the max retries.
+            if let Some(timeout_duration) = timeout_duration {
+                if start_time.elapsed() > timeout_duration {
+                    tracing::debug!(
+                        "Timeout reached after {:?} while retrieving slivers",
+                        timeout_duration
```
mlegner: Nit: You can format the timeout with `humantime` (after adding it as a dependency for `walrus-sdk` (it's already an indirect dependency through `walrus-utils`:
```suggestion
                        "timeout reached after {} while retrieving slivers",
                        humantime::Duration::from(timeout_duration)
```

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+                }
+            }
+            if let Some(max_attempts) = max_attempts {
+                if attempts > max_attempts {
+                    tracing::debug!("Max attempts ({}) reached", max_attempts);
+                    break;
+                }
+            }
+
+            tracing::debug!("Retrieving slivers: {:?}", sliver_selector);
```
mlegner: Nit: We normally prefer including additional data for log entries as fields, in particular when they use debug formatting.
```suggestion
            tracing::debug!(?sliver_selector, "retrieving slivers");
```

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+            {
+                Ok(new_slivers) => {
+                    // Track which indices we've successfully retrieved.
+                    for sliver in new_slivers {
+                        sliver_selector.complete_sliver(&sliver.index);
+                        all_slivers.insert(sliver.index, sliver);
+                    }
+                }
+                Err(e) => {
+                    // TODO(heliu): if the error is not retriable, return the error immediately.
```
mlegner: Can you add a tracking issue for this?

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+        SliverData<E>: TryFrom<walrus_core::Sliver>,
+    {
+        let mut sliver_selector = SliverSelector::<E>::new(
+            sliver_indices.to_vec(),
+            metadata.n_shards(),
+            metadata.blob_id(),
+        );
+        let mut attempts = 0;
+        let retry_delay_ms = 100;
+        let mut all_slivers: HashMap<SliverIndex, SliverData<E>> = HashMap::new();
```
mlegner: Question: Why is this a `HashMap`? AFAIU, this could just as well be a `Vec::with_capacity(sliver_selector.num_slivers())`.

crates/walrus-sdk/src/client.rs
```diff
@@ -88,6 +92,55 @@ pub mod refresh;
...
+            .collect::<BiMap<_, _>>();
+
+        Self {
+            indices_and_shards,
+            _phantom: PhantomData,
+        }
+    }
+
+    /// Returns the number of slivers.
+    pub fn num_slivers(&self) -> usize {
```
mlegner: Nit/question: Why not use `len` here?

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+                }
+            }
+
+            attempts += 1;
+            if !sliver_selector.is_empty() {
+                tokio::time::sleep(Duration::from_millis(retry_delay_ms)).await;
+            }
+        }
+
+        if all_slivers.len() != sliver_indices.len() {
```
mlegner: Important: This could also happen if `sliver_indices` contained some duplicates. We should document what happens in that case.

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+            "requesting {} slivers",
+            sliver_selector.num_slivers()
+        ));
+
+        let committees = self.get_committees().await?;
+        let comms = self
+            .communication_factory
+            .node_read_communications(&committees, certified_epoch)?;
+
+        // Create requests to get all slivers from all nodes.
```
mlegner: This comment is not correct:
```suggestion
        // Create requests to get all requested slivers.
```

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+                            }
+                        })
+                })
+        });
+
+        let mut requests = WeightedFutures::new(futures);
+
+        // Execute all requests with appropriate concurrency limits.
+        requests
+            .execute_weight(
```
mlegner: Important/question: Do we have/want a timeout here?

crates/walrus-sdk/src/client.rs
```diff
@@ -441,6 +510,172 @@ impl<T: ReadClient> Client<T> {
...
+                        &self.encoding_config,
+                        metadata.metadata().encoding_type(),
+                    ),
+            )
+            .await;
+
+        progress_bar.finish_with_message("slivers received");
+
+        let mut n_not_found = 0;
+        let mut n_forbidden = 0;
```
mlegner: Question: What is the point of these variables?

mlegner (commented):
Thanks a lot for the PR, @liquid-helium. I have a few questions and some minor suggestions.

crates/walrus-sdk/Cargo.toml
liquid-helium: Done.

crates/walrus-sdk/src/client.rs
liquid-helium: Good idea, changed to` &[SliverIndex]`

crates/walrus-sdk/src/client.rs
liquid-helium: Renamed the function to `remove_sliver`, and added comments.

crates/walrus-sdk/src/client.rs
liquid-helium: Change to const. Thanks for the tip.

crates/walrus-sdk/src/client.rs
liquid-helium: Nice tip, change applied.

crates/walrus-sdk/src/client.rs
liquid-helium: Thanks for the tip, change applied.

crates/walrus-sdk/src/client.rs
liquid-helium: Nice catch!
This was used to track the 'retrieved slivers' before I moved the function to the SliverSelector, fix applied.

crates/walrus-sdk/src/client.rs
liquid-helium: change applied.

crates/walrus-sdk/src/client.rs
liquid-helium: Good catch, I changed to explicitly retrieve and check the `unique` sliver indices, also added comments.

thanks.

crates/walrus-sdk/src/client.rs
liquid-helium: suggestion applied.

crates/walrus-sdk/src/client.rs
liquid-helium: Good call, added a timeout.

crates/walrus-sdk/src/client.rs
liquid-helium: Removed.

crates/walrus-sdk/src/client.rs
liquid-helium: Created WAL-685.

crates/walrus-sdk/src/client.rs
liquid-helium: Same, and after I added a constant which is used in this function, it complains that constant is unused 😕

crates/walrus-sdk/src/client.rs
```diff
@@ -513,79 +519,78 @@ impl<T: ReadClient> Client<T> {
...
-                        all_slivers.insert(sliver.index, sliver);
+                        sliver_selector.remove_sliver(&sliver.index);
+                        all_slivers.push(sliver);
                     }
                 }
                 Err(e) => {
-                    // TODO(heliu): if the error is not retriable, return the error immediately.
-                    tracing::warn!("Error retrieving slivers: {:?}", e);
+                    // TODO(WAL-685): if the error is not retriable, return the error immediately.
+                    tracing::warn!(?e, "error retrieving slivers");
```
mlegner: Nit: This way, the error will be recorded as the field `e`. Calling the variable `error` would be nicer.

crates/walrus-sdk/src/client.rs
```diff
@@ -513,79 +519,78 @@ impl<T: ReadClient> Client<T> {
...
             if !sliver_selector.is_empty() {
-                tokio::time::sleep(Duration::from_millis(retry_delay_ms)).await;
+                tokio::time::sleep(RETRIEVE_SLIVERS_RETRY_DELAY).await;
             }
         }
 
-        if all_slivers.len() != sliver_indices.len() {
+        if all_slivers.len() != num_unique_slivers {
             Err(ClientError::from(ClientErrorKind::Other(
                 format!("Failed to retrieve slivers: {:?}", last_error).into(),
```
mlegner: Nit: Can we include how many we are missing? Something like "failed to retrieve 
```suggestion
                format!("failed to retrieve some slivers ({}/{} successful): {:?}", all_slivers.len(), num_unique_slivers, last_error).into()
```

crates/walrus-sdk/src/client.rs
```diff
@@ -642,8 +645,9 @@ impl<T: ReadClient> Client<T> {
 
         // Execute all requests with appropriate concurrency limits.
         requests
-            .execute_weight(
-                &|_| false, // We want to execute all futures.
+            .execute_until(
+                &|_| false,       // We want to execute all futures.
+                timeout_duration, // Default 30s timeout if none provided
```
mlegner: Question: Where is that default defined?

crates/walrus-sdk/src/client.rs
```diff
@@ -655,23 +659,10 @@ impl<T: ReadClient> Client<T> {
...
-                        tracing::debug!(%node, %error, "retrieving sliver failed");
-                        if error.is_status_not_found() {
-                            n_not_found += 1;
-                        } else if error.is_blob_blocked() {
-                            n_forbidden += 1;
-                        }
-                    })
-                    .ok()
-            })
+            .filter_map(|NodeResult { result, .. }| result.ok())
```
mlegner: Nit: I think it makes sense to keep the debug log.

crates/walrus-sdk/src/client.rs
liquid-helium: Change applied.

crates/walrus-sdk/src/client.rs
liquid-helium: Thanks for the suggestion, change applied.

crates/walrus-sdk/src/client.rs
liquid-helium: Added back the debug log.

crates/walrus-sdk/src/client.rs
liquid-helium: Some leftover from previous version, comments removed, now the timeout_duration is required for the function.

PR #2151, opened 2025-05-30
feat: use subsidies contract to proxy register blob call

zihehuang:
## Description

This PR modifies the CLI to proxy `register_blob` to subsidies contract as well as `reserve_space`. By doing so, this enables Walrus partners to pay zero WAL when interacting with the network. One the new subsidies contract is deployed. The code path should still work since this feature is dependent on subsidies_object being set in the config

## Test plan

New sim test

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI: use subsidies contract to proxy register blob call

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -1599,21 +1599,18 @@ impl SuiContractClientInner {
             .iter()
             .fold(0, |acc, metadata| acc + metadata.encoded_size);
 
-        let main_storage_arg = match with_subsidies {
-            true => {
-                pt_builder
-                    .reserve_space_with_subsidies(
-                        main_storage_arg_size,
-                        epochs_ahead,
-                        subsidies_package_id,
-                    )
-                    .await?
-            }
-            false => {
-                pt_builder
-                    .reserve_space_without_subsidies(main_storage_arg_size, epochs_ahead)
-                    .await?
-            }
+        let main_storage_arg = if with_subsidies {
+            pt_builder
+                .reserve_space_with_subsidies(
+                    main_storage_arg_size,
+                    epochs_ahead,
+                    subsidies_package_id,
+                )
+                .await?
+        } else {
+            pt_builder
+                .reserve_space_without_subsidies(main_storage_arg_size, epochs_ahead)
+                .await?
         };
 
         for blob_metadata in blob_metadata_list.into_iter() {
@@ -1627,26 +1624,19 @@ impl SuiContractClientInner {
                 main_storage_arg
             };
 
-            match with_subsidies {
-                true => {
-                    pt_builder
-                        .register_blob_with_subsidies(
-                            storage_arg.into(),
-                            blob_metadata,
-                            persistence,
-                            subsidies_package_id,
-                        )
-                        .await?
-                }
-                false => {
-                    pt_builder
-                        .register_blob_without_subsidies(
-                            storage_arg.into(),
-                            blob_metadata,
-                            persistence,
-                        )
-                        .await?
-                }
+            if with_subsidies {
+                pt_builder
+                    .register_blob_with_subsidies(
+                        storage_arg.into(),
+                        blob_metadata,
+                        persistence,
+                        subsidies_package_id,
+                    )
+                    .await?
+            } else {
+                pt_builder
+                    .register_blob_without_subsidies(storage_arg.into(), blob_metadata, persistence)
+                    .await?
             };
         }
 
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -32,7 +32,7 @@ use sui_types::{
     TypeTag,
     base_types::SuiAddress,
     event::EventID,
-    transaction::{Argument, TransactionData},
+    transaction::{Argument, ProgrammableTransaction, TransactionData, TransactionKind},
 };
 use tokio::sync::Mutex;
 use tokio_stream::Stream;
@@ -1471,14 +1471,34 @@ impl SuiContractClientInner {
             return Ok(vec![]);
         }
 
+        let subsidies_package_id = self.read_client.get_subsidies_package_id();
+
         let expected_num_blobs = blob_metadata_and_storage.len();
         tracing::debug!(num_blobs = expected_num_blobs, "starting to register blobs");
         let mut pt_builder = self.transaction_builder()?;
         // Build a ptb to include all register blob commands for all blobs.
         for (blob_metadata, storage) in blob_metadata_and_storage.into_iter() {
-            pt_builder
-                .register_blob(storage.id.into(), blob_metadata, persistence)
-                .await?;
+            match subsidies_package_id {
+                Some(pkg_id) => {
+                    pt_builder
+                        .register_blob_with_subsidies(
+                            storage.id.into(),
+                            blob_metadata,
+                            persistence,
+                            pkg_id,
+                        )
+                        .await?;
+                }
+                None => {
+                    pt_builder
+                        .register_blob_without_subsidies(
+                            storage.id.into(),
+                            blob_metadata,
+                            persistence,
+                        )
+                        .await?;
+                }
+            };
         }
         let transaction = pt_builder.build_transaction_data(self.gas_budget).await?;
         let res = self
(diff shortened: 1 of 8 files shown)


crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1432,19 +1473,58 @@ impl WalrusPtbBuilder {
...
+        );
+        let buyer_pays = match self.read_client.get_subsidies_object_id() {
+            Some(subsidies_object_id) if with_subsidies => {
+                let subsidies_object = self
+                    .read_client
+                    .sui_client()
+                    .get_subsidies_object(subsidies_object_id)
+                    .await?;
+                full_price * (MAX_SUBSIDY_RATE - subsidies_object.buyer_subsidy_rate as u64)
+                    / MAX_SUBSIDY_RATE
```
karlwuest: Important: I think this calculation is not quite correct, because it rounds towards zero, so if the numerator is not divisible by `MAX_SUBSIDY_RATE`, one FROST may be missing in the calculated amount. Instead, we should always round up.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1432,19 +1473,58 @@ impl WalrusPtbBuilder {
...
+                    .read_client
+                    .sui_client()
+                    .get_subsidies_object(subsidies_object_id)
+                    .await?
+                    .buyer_subsidy_rate;
+                if buyer_subsidy_rate as u64 == MAX_SUBSIDY_RATE {
+                    0
+                } else {
+                    full_price
+                }
```
karlwuest: Remark: This is a bit of a hacky way to determine whether the subsidies object is used to also subsidize writes. But I also don't see a better way that does not introduce a lot of additional complexity 🤷‍♂️

crates/walrus-sui/src/utils.rs
```diff
@@ -45,6 +45,9 @@ use crate::{
     wallet::Wallet,
 };
 
+/// Max subsidy rate is 100%.
+pub const MAX_SUBSIDY_RATE: u64 = 10_000;
```
karlwuest: Nit/Remark: Let's call this `MAX_BUYER_SUBSIDY_RATE` instead of `MAX_SUBSIDY_RATE` just for clarity and future proofing. With the planned changes to the subsidies, we may want to have system subsidies above 100% so that will probably also change in the contracts.

crates/walrus-sui/src/client.rs
```diff
@@ -1553,20 +1573,66 @@ impl SuiContractClientInner {
...
-            |builder, encoded_size, epochs| {
-                Box::pin(async move {
-                    builder
-                        .reserve_space_with_subsidies(encoded_size, epochs, subsidies_package_id)
-                        .await
-                }) as BoxFuture<'_, SuiClientResult<Argument>>
-            },
-        )
-        .await
+        if blob_metadata_list.is_empty() {
```
karlwuest, line 1584: Suggestion: This now duplicates a lot of code with `reserve_and_register_blobs_without_subsidies`. Can we still keep an "inner" function, but instead of passing two closures for reserve and register, we just pass a flag `with_subsidies` (and then make a case distinction at the relevant points in the function)?

crates/walrus-sui/src/client/transaction_builder.rs
zihehuang: Yeah I think this is just a stopgap solution to deal with the subsidized write fees right now. When the new subsidy mechanism is live we can probably remove this if-else check completely. I will comment this section for posterity

crates/walrus-sui/src/client.rs
```diff
@@ -1645,14 +1626,33 @@ impl SuiContractClientInner {
             } else {
                 main_storage_arg
             };
-            pt_builder
-                .register_blob(storage_arg.into(), blob_metadata, persistence)
-                .await?;
+
+            match with_subsidies {
```
wbbradley: nit: maybe we should use`if/else` for switching on `bool` values.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1537,19 +1596,60 @@ impl WalrusPtbBuilder {
...
+        );
+        let buyer_pays = match self.read_client.get_subsidies_object_id() {
+            Some(subsidies_object_id) if with_subsidies => {
+                let buyer_subsidy_rate = self
+                    .read_client
+                    .sui_client()
+                    .get_subsidies_object(subsidies_object_id)
+                    .await?
+                    .buyer_subsidy_rate;
+                // Hack until new subsidy is integrated with system contract
```
wbbradley: Please add a TODO linking to a Linear task, like
```
        // TODO(WAL-007): short description of the hack/todo
```

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -332,15 +350,16 @@ impl WalrusPtbBuilder {
         Ok(result_arg)
     }
 
-    /// Adds a call to `register_blob` to the `pt_builder` and returns the result [`Argument`].
-    pub async fn register_blob(
+    /// Adds a call to `register_blob` to the `pt_builder` and returns the result [`Argument`]
+    /// without subsidies.
+    pub async fn register_blob_without_subsidies(
```
halfprice, line 340: Nit: can we remove `without_subsidies` from the function name, if using subsidies is a special case? Mentioning the special case in the general case may be confusing.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -319,8 +319,9 @@ impl SuiReadClient {
         let wal_type = sui_client.wal_type_from_package(walrus_package_id).await?;
         let subsidies = if let Some(subsidies_object_id) = contract_config.subsidies_object {
             let subsidies_package_id = sui_client
-                .get_subsidies_package_id_from_subsidies_object(subsidies_object_id)
-                .await?;
+                .get_subsidies_object(subsidies_object_id)
+                .await?
```
halfprice, line 323: Question: this makes that the subsidies contract is required to be published. Is this always true? Does it make sense to set subsidies to `None` and still use the normal contract, instead of returning the `not found` error?

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -167,47 +167,64 @@ impl WalrusPtbBuilder {
...
-        } else {
+
+        // If we don't have a wal_coin_arg yet, we need to select at least one coin
+        if self.wal_coin_arg.is_none() {
+            let mut coins = self
+                .read_client
+                .get_coins_with_total_balance(
+                    self.sender_address,
+                    CoinType::Wal,
+                    0, // Get at least a zero coin
```
halfprice: Question: why doing this two steps getting coins instead of using one call to get coins with at least `min_balance`?

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -167,47 +167,64 @@ impl WalrusPtbBuilder {
     /// Returns a [`SuiClientError::NoCompatibleWalCoins`] if no WAL coins with sufficient balance
     /// can be found.
     pub async fn fill_wal_balance(&mut self, min_balance: u64) -> SuiClientResult<()> {
-        if min_balance <= self.tx_wal_balance {
+        // If we already have a wal_coin_arg and sufficient balance, we're done
```
karlwuest, line 170: Question: Why are the changes to this function necessary? It should already get a zero coin if that is the only coin available and `min_balance` is 0.

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -1537,19 +1596,60 @@ impl WalrusPtbBuilder {
...
+        let buyer_pays = match self.read_client.get_subsidies_object_id() {
+            Some(subsidies_object_id) if with_subsidies => {
+                let subsidies_object = self
+                    .read_client
+                    .sui_client()
+                    .get_subsidies_object(subsidies_object_id)
+                    .await?;
+                (full_price
+                    * (MAX_BUYER_SUBSIDY_RATE - u64::from(subsidies_object.buyer_subsidy_rate)))
+                .div_ceil(MAX_BUYER_SUBSIDY_RATE)
```
karlwuest: Question/Suggestion: This should not depend on the maximum subsidy rate, right? Instead, what we want here is 100% - the buyer subsidy rate. I think it would be better to add a constant for 100% in basis points and use that here.

crates/walrus-sui/src/client/read_client.rs
zihehuang: I believe the else statement got retracted by github os it is already taking care of the `None` condition. It is doing exactly as you described. It should work even if the subsidies contract is no longer deployed or the subsidies object is no longer included in the config

crates/walrus-sui/src/client/transaction_builder.rs
zihehuang: When the `min_balance` is 0 this function will return ok but the `wal_coin_arg` will still be None. This will eventually be part of the PTB argument and error with a `NoCompatibleCoin` error. The update to this function is to handle this edge case but let me see if I can clean up the logic here a bit more. It is kind of messy I agree

PR #2152, opened 2025-05-30
feat: Refactor event processing code

sadhansood:
## Description

We are looking to move event processing from running on individual storage node to its own dedicated centralized service. This service will be used by storage nodes to read events (and authenticate them) from instead of locally doing checkpoint processing. With that goal in mind, this is the first PR which moves event processing code from `node` and also refactors it for easier usage. NOTE: there is no functionality that is getting added or removed in this PR.
 
## Test plan

Existing tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/lib.rs
```diff
@@ -16,6 +16,9 @@ pub use node::StorageNodeConfigLoader;
 #[cfg(feature = "node")]
 pub use node::errors::SyncNodeConfigError;
 
+#[cfg(feature = "node")]
+pub mod event;
```
mlegner, line 20: Question: (Why) does this have to be pub?

crates/walrus-service/src/test_utils.rs
```diff
@@ -487,29 +486,30 @@ impl SimStorageNodeHandle {
...
-                        .expect("temporary directory creation must succeed")
-                        .path()
-                        .to_path_buf()
-                ),
-                rpc_fallback_config: None,
-                db_config: DatabaseConfig::default(),
-            };
-            let system_config = crate::node::events::event_processor::SystemConfig {
+            let processor_config =
+                crate::event::event_processor::config::EventProcessorRuntimeConfig {
```
mlegner: Question (nit): Is there a reason to qualify this instead of importing as we do normally (also the `SystemConfig` below)?

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -0,0 +1,341 @@
...
+            .field("committee_store", &self.stores.committee_store)
+            .field("event_store", &self.stores.event_store)
+            .finish()
+    }
+}
+
+impl EventProcessor {
+    /// Creates a new event processor.
+    pub async fn new(
+        config: EventProcessorConfig,
```
mlegner: Question: Why was this changed from a reference to an owned parameter?

crates/walrus-service/src/node.rs
```diff
@@ -171,7 +176,7 @@ pub mod config;
 pub(crate) mod consistency_check;
 pub mod contract_service;
 pub mod dbtool;
-pub mod events;
+pub mod event_blob_writer;
```
mlegner, line 178: Question: What is the plan with the event-blob writer? Is this going to stay with the node?

crates/walrus-service/src/event.rs
mlegner, line 1: Question: Is the ultimate goal to move this to a separate crate or will this just remain a module within `walrus-service`?

mlegner (commented):
Thank you so much for the reorganization. This generally LGTM. I have a few more detailed questions.

Note that there is an E2E test and some simtests that are failing currently.

I didn't really look at the moved code; my assumption is that it is essentially unchanged. Let me know if I should review something in a bit more detail.

crates/walrus-service/src/node.rs
sadhansood: Yes, that is the plan since node still does event blob writing. Lmk what you think

crates/walrus-service/src/event.rs
sadhansood: i think this still stays within `walrus-service` as one of the services in walrus umbrella, but lmkwyt

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: ah my bad, will fix

crates/walrus-service/src/test_utils.rs
sadhansood: ok will fix

crates/walrus-service/src/lib.rs
sadhansood: ok it doesn't i think, will fix

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -1,1522 +0,0 @@
...
-            &ReadWriteOptions::default(),
-            false,
-        )?;
-        let committee_store = DBMap::reopen(
-            database,
-            Some(constants::COMMITTEE_STORE),
-            &default_rw_opts,
-            false,
-        )?;
-        let event_store_opts = ReadWriteOptions::default().set_ignore_range_deletions(true);
```
halfprice, line 904: We no longer need this option in event_store?

crates/walrus-service/src/node/events/event_processor.rs
```diff
@@ -1,1522 +0,0 @@
...
-#[async_trait]
-impl PackageStore for LocalDBPackageStore {
-    async fn fetch(&self, id: AccountAddress) -> sui_package_resolver::Result<Arc<Package>> {
-        let object = self.get(id).await?;
-        Ok(Arc::new(Package::read_from_object(&object)?))
-    }
-}
-
-#[cfg(test)]
-mod tests {
```
halfprice, line 1291: question: @sadhansood , it seems that these tests are removed?

PR #2153, opened 2025-05-31
docs(walrus-sites): add a note for browsing options of a ws on testnet

DionisisLougaris:
## Description

Added a note for browsing options of a walrus site on testnet because the current output depicted in the docs isn't entirely correct for testnet publishes.


mlegner (approved):
Thanks @DionisisLougaris. We can merge this after #2156, which fixes a broken link.

PR #2159, opened 2025-06-02
feat: rate limit concurrent blob recovery during node recovery

halfprice:
## Description

For full node recovery, the memory usage grows when the number of live blobs in the system, which doesn't scale and may
cause OOM.

## Test plan

Simtest
Each modified simtest passed 100 seeds.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/node_config_example.yaml
+++ b/crates/walrus-service/node_config_example.yaml
@@ -173,3 +173,5 @@ checkpoint_config:
   max_background_operations: 1
   periodic_db_checkpoints: false
 admin_socket_path: null
+node_recovery_config:
+  max_concurrent_blob_syncs_during_recovery: 1000
--- a/crates/walrus-service/src/node/node_recovery.rs
+++ b/crates/walrus-service/src/node/node_recovery.rs
@@ -3,12 +3,12 @@
 
 use std::sync::{Arc, Mutex};
 
-use futures::future::join_all;
+use futures::stream::{FuturesUnordered, StreamExt};
 use sui_macros::fail_point_async;
 use typed_store::TypedStoreError;
 use walrus_core::Epoch;
 
-use super::{StorageNodeInner, blob_sync::BlobSyncHandler};
+use super::{StorageNodeInner, blob_sync::BlobSyncHandler, config::NodeRecoveryConfig};
 use crate::node::{NodeStatus, storage::blob_info::CertifiedBlobInfoApi};
 
 #[derive(Debug, Clone)]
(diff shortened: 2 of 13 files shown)


crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -18,20 +18,29 @@ pub struct NodeRecoveryHandler {
...
             blob_sync_handler,
             task_handle: Arc::new(Mutex::new(None)),
+            config,
         }
     }
 
     /// Starts the node recovery process to recover blobs that are certified before the given epoch.
     /// For blobs that are certified after `certified_before_epoch`, the event processing is in
     /// charge of making sure the blob is stored at all shards.
+    // TODO(WAL-864): Refactor this function to make it readable.
```
halfprice, line 43: This function is getting long and I want to refactor it, but preferably in a separate PR since I want this PR to be easier to review by looking at the diff.

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -41,10 +50,23 @@ impl NodeRecoveryHandler {
...
         let blob_sync_handler = self.blob_sync_handler.clone();
+        let max_concurrent_blob_syncs_during_recovery =
+            self.config.max_concurrent_blob_syncs_during_recovery;
         let task_handle = tokio::spawn(async move {
             fail_point_async!("start_node_recovery_entry");
+
+            // Limit the number of concurrent blob syncs to avoid overwhelming the system.
+            let semaphore = Arc::new(tokio::sync::Semaphore::new(
+                max_concurrent_blob_syncs_during_recovery,
+            ));
```
liquid-helium, line 61: Remark: can we check FuturesUnordered::len() to determine is a new blob_sync task should be started, without using semaphore?

halfprice: Yeah, that is essentially implementing a semaphore (keep track of the number of running blob syncs and wait if there are more than limit)? By using a semaphore, the loop can be awaken up asynchronously, which seems more efficient.

liquid-helium: If I read it correctly, we are waiting on FuturesUnordered::next() after failing to obtain a permit, so there seems to be some duplication.

halfprice: Oh, I see , you mean the semaphore is not needed. Yeah, good point. let me send out a quick PR for this. sorry that missed the commemoration before merging

crates/walrus-service/node_config_example.yaml
```diff
@@ -173,3 +173,5 @@ checkpoint_config:
   max_background_operations: 1
   periodic_db_checkpoints: false
 admin_socket_path: null
+node_recovery_config:
+  max_concurrent_blob_syncs_during_recovery: 1000
```
mlegner, line 177: Question: How does this relate to the parameter `shard_sync_config. max_concurrent_blob_recovery_during_shard_recovery`?

mlegner (commented):
Thanks for fixing this @halfprice! One question about the added configuration parameter.

crates/walrus-service/node_config_example.yaml
halfprice: They are independent. During shard recovery, we don't use the blob sync handler to sync blob, and the recovery there is controled by max_concurrent_blob_recovery_during_shard_recovery

mlegner: Ah, I see. So we have 3 different parameters:
- During normal operation (`Active` state).
- During recovery (`RecoveryInProgress` state).
- During shard recovery (`Active` node state but `inRecovery` shard status).

Is that correct?

PR #2162, opened 2025-06-03
chore: remove redundant semaphore used in node recovery

halfprice:
## Description

The functionality of rate limiting concurrent blob sync can be achieved via monitoring ongoing_sync length.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/node_recovery.rs
+++ b/crates/walrus-service/src/node/node_recovery.rs
@@ -55,11 +55,6 @@ impl NodeRecoveryHandler {
         let task_handle = tokio::spawn(async move {
             fail_point_async!("start_node_recovery_entry");
 
-            // Limit the number of concurrent blob syncs to avoid overwhelming the system.
-            let semaphore = Arc::new(tokio::sync::Semaphore::new(
-                max_concurrent_blob_syncs_during_recovery,
-            ));
-
             loop {
                 // Keep track of ongoing blob syncs. Note that the memory usage of this list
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -123,18 +118,19 @@ impl NodeRecoveryHandler {
...
+                    if ongoing_syncs.len() >= max_concurrent_blob_syncs_during_recovery {
+                        tracing::debug!(
+                            walrus.blob_id = %blob_id,
+                            number_of_tasks = %ongoing_syncs.len(),
+                            "max concurrent blob syncs reached; wait for one to complete"
+                        );
+                        while ongoing_syncs.len() >= max_concurrent_blob_syncs_during_recovery {
+                            ongoing_syncs.next().await;
                         }
-                    };
```
liquid-helium, line 137: nit: a while loop should be sufficient
```
while ongoing_syncs.len() >= max_concurrent_blob_syncs_during_recovery {
    ongoing_syncs.next().await;
}
```

halfprice: Yes, the reason I did the if-then-while is to add that logging line. I can also do something like

```
if ongoing_syncs.len() >= .. {
  tracing::debug!();
}
while ongoing_syncs.len() >= max_concurrent_blob_syncs_during_recovery {
  ongoing_syncs.next().await;
}
```

but not sure if this is cleaner.

PR #2171, opened 2025-06-03
feat: Quilt client part ii

liquid-helium:
## Description

Added QuiltClient to handle storing/reading quilts.
Contributes to WAL-858
Close WAL-837

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client/quilt_client.rs
+++ b/crates/walrus-sdk/src/client/quilt_client.rs
@@ -50,12 +50,7 @@ pub fn read_blobs_from_paths<P: AsRef<Path>>(paths: &[P]) -> ClientResult<Vec<(P
             )));
         }
 
-        // Ignores non-file and non-directory paths.
-        if path.is_file() {
-            collected_files.insert(path.to_path_buf());
-        } else if path.is_dir() {
-            collected_files.extend(get_all_files_from_dir(path)?);
-        }
+        collected_files.extend(get_all_files_from_path(path)?);
     }
 
     let mut collected_files_with_content = Vec::with_capacity(collected_files.len());
@@ -69,16 +64,16 @@ pub fn read_blobs_from_paths<P: AsRef<Path>>(paths: &[P]) -> ClientResult<Vec<(P
 }
 
 /// Get all file paths from a directory recursively.
-fn get_all_files_from_dir<P: AsRef<Path>>(path: P) -> ClientResult<HashSet<PathBuf>> {
+fn get_all_files_from_path<P: AsRef<Path>>(path: P) -> ClientResult<HashSet<PathBuf>> {
     let path = path.as_ref();
-    let mut collected_files: HashSet<PathBuf> = HashSet::new();
-    let dir_entries = fs::read_dir(path).map_err(ClientError::other)?;
-    for entry in dir_entries {
-        let current_entry_path = entry.map_err(ClientError::other)?.path();
-        if current_entry_path.is_file() {
-            collected_files.insert(current_entry_path);
-        } else if current_entry_path.is_dir() {
-            collected_files.extend(get_all_files_from_dir(&current_entry_path)?);
+    let mut collected_files = HashSet::new();
+
+    if path.is_file() {
+        collected_files.insert(path.to_owned());
+    } else if path.is_dir() {
+        for entry in fs::read_dir(path).map_err(ClientError::other)? {
+            let current_entry_path = entry.map_err(ClientError::other)?.path();
+            collected_files.extend(get_all_files_from_path(&current_entry_path)?);
         }
     }
 
@@ -221,7 +216,7 @@ impl<T: ReadClient> QuiltClient<'_, T> {
         SliverData<V::SliverAxis>: TryFrom<Sliver>,
     {
         let mut all_slivers = Vec::new();
-        let mut refs = [first_sliver].to_vec();
+        let mut refs = vec![first_sliver];
         let first_sliver_refs = [first_sliver].to_vec();
         let mut decoder = V::QuiltConfig::get_decoder(&first_sliver_refs);
 
@@ -361,6 +356,8 @@ impl QuiltClient<'_, SuiContractClient> {
     ///
     /// The on-disk file names are used as identifiers for the quilt patches.
     /// If the file name is not valid UTF-8, it will be replaced with "unnamed-blob-<index>".
+    //
+    // TODO(WAL-887): Use relative paths to deduplicate the identifiers.
     fn assign_identifiers_with_paths(
         blobs_with_paths: &[(PathBuf, Vec<u8>)],
     ) -> Vec<QuiltStoreBlob> {


crates/walrus-sdk/src/client/responses.rs
```diff
@@ -55,7 +55,7 @@ pub struct BlobStoreResultWithPath {
 
 /// Result when attempting to store a blob.
 #[serde_as]
-#[derive(Debug, Clone, Serialize, ToSchema)]
+#[derive(Debug, Clone, Deserialize, Serialize, ToSchema)]
```
mlegner: Question: Why do we need `Deserialize` here and below?

crates/walrus-sdk/src/error.rs
```diff
@@ -108,6 +108,12 @@ impl ClientError {
     }
 }
 
+impl From<QuiltError> for ClientError {
+    fn from(value: QuiltError) -> Self {
+        ClientError::other(value)
```
mlegner: Suggestion: Define a `ClientErrorKind::QuiltError` variant below.

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+};
+use walrus_sui::client::{BlobPersistence, PostStoreAction, ReadClient, SuiContractClient};
+
+use crate::{
+    client::{Client, client_types::StoredQuiltPatch, responses::QuiltStoreResult},
+    error::{ClientError, ClientErrorKind, ClientResult},
+    store_when::StoreWhen,
+};
+
+// Reads all files recursively from a given path and returns them as path-content pairs.
```
mlegner: Nit:
```suggestion
/// Reads all files recursively from a given path and returns them as path-content pairs.
```

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+    }
+
+    let mut collected_files: HashSet<PathBuf> = HashSet::new();
+    for path in paths {
+        let path = path.as_ref();
+
+        // Validate path existence and accessibility.
+        if !path.exists() {
+            return Err(ClientError::from(ClientErrorKind::Other(
+                format!("Path does not exist: {:?}.", path).into(),
```
mlegner: Nit:
```suggestion
                format!("path '{}' does not exist", path.display()).into(),
```

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+            )));
+        }
+
+        if path.is_file() {
+            collected_files.insert(path.to_path_buf());
+        } else if path.is_dir() {
+            collected_files.extend(get_all_files_from_dir(path)?);
+        } else {
+            return Err(ClientError::from(ClientErrorKind::Other(
+                format!("Path is neither a file nor a directory: {:?}.", path).into(),
```
mlegner: Question: What else would it be?

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+            )));
+        }
+    }
+
+    let mut collected_files_with_content: Vec<(PathBuf, Vec<u8>)> = Vec::new();
+    for file_path in collected_files {
+        let content = fs::read(&file_path).map_err(ClientError::other)?;
```
mlegner: Nit: We have a `read_blob_from_file` function [here](https://github.com/MystenLabs/walrus/blob/2b1030b6a9fc751591793b9d37824cb94dcc23ad/crates/walrus-service/src/client/cli.rs#L395).

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+        } else if path.is_dir() {
+            collected_files.extend(get_all_files_from_dir(path)?);
+        } else {
+            return Err(ClientError::from(ClientErrorKind::Other(
+                format!("Path is neither a file nor a directory: {:?}.", path).into(),
+            )));
+        }
+    }
+
+    let mut collected_files_with_content: Vec<(PathBuf, Vec<u8>)> = Vec::new();
```
mlegner: Nit: You could use `Vec::with_capacity(collected_files.len())` here?

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+    for entry in dir_entries {
+        let current_entry_path = entry.map_err(ClientError::other)?.path();
+        if current_entry_path.is_file() {
+            collected_files.insert(current_entry_path);
+        } else if current_entry_path.is_dir() {
+            collected_files.extend(get_all_files_from_dir(&current_entry_path)?);
+        }
+    }
+
+    Ok(collected_files.into_iter().collect())
```
mlegner: Question: Why not just return the hash set?

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+        if path.is_file() {
+            collected_files.insert(path.to_path_buf());
+        } else if path.is_dir() {
+            collected_files.extend(get_all_files_from_dir(path)?);
+        } else {
+            return Err(ClientError::from(ClientErrorKind::Other(
+                format!("Path is neither a file nor a directory: {:?}.", path).into(),
+            )));
+        }
+    }
```
mlegner: Remark/suggestion: This is largely duplicated with `get_all_files_from_dir`, except for the error for paths that aren't files or directories (which are silently ignored in `get_all_files_from_dir`). Is this intentional? This could still be rewritten slightly as 
```rust
        if !path.is_file() && !path.is_dir() {
            return Err(ClientError::from(ClientErrorKind::Other(
                format!(
                    "path '{}' is neither a file nor a directory",
                    path.display()
                )
                .into(),
            )));
        }
        collected_files.extend(get_all_files_from_dir(path)?);
```

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+
+/// A facade for interacting with Walrus quilt.
+#[derive(Debug, Clone)]
+pub struct QuiltClient<'a, T> {
+    client: &'a Client<T>,
+}
+
+impl<'a, T> QuiltClient<'a, T> {
+    const MAX_RETRIEVE_SLIVERS_ATTEMPTS: usize = 2;
+    const TIMEOUT_DURATION: Duration = Duration::from_secs(10);
```
mlegner: Suggestion: Make these configurable by changing them to struct fields that need to be provided when instantiating the `QuiltClient`. They can then later be provided through command-line options or the `client_config.yaml` with these as defaults.

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+        };
+
+        Ok(quilt_index)
+    }
+
+    async fn retrieve_quilt_index_internal<V: QuiltVersion>(
+        &self,
+        metadata: &VerifiedBlobMetadataWithId,
+        certified_epoch: Epoch,
+        first_sliver: &[SliverData<V::SliverAxis>],
```
mlegner: Nit: This doesn't need to be a slice.
```suggestion
        first_sliver: &SliverData<V::SliverAxis>,
```

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+                    quilt
+                        .get_blob_by_identifier(identifier)
+                        .map_err(ClientError::other)
+                })
+                .collect::<Result<Vec<_>, _>>()
+        }
+    }
+
+    /// Retrieves the quilt from Walrus.
+    async fn get_quilt_enum(
```
mlegner: Nit: The function name is a bit confusing. How about something like `get_full_quilt`?

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+            } else {
+                // If the quilt index cannot be retrieved from the slivers, try to retrieve the
+                // quilt.
+                tracing::debug!(
+                    "failed to retrieve index slivers, trying to get quilt instead {}",
+                    quilt_id
+                );
+                self.get_quilt_enum(&metadata, certified_epoch)
+                    .await?
+                    .get_quilt_index()?
```
mlegner: Important: If we download the full quilt here, we should cache it and use it later to extract the patches.

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+                )
+                .await?
+            }
+        };
+
+        Ok(blobs)
+    }
+
+    /// Retrieves quilt patches from QuiltV1 by identifiers.
+    async fn get_blobs_by_identifiers_impl<V: QuiltVersion>(
```
mlegner: Question: Is this function specific to V1 as the docstring indicates or generic as the function signature?

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+        blobs_with_paths: &[(PathBuf, Vec<u8>)],
+    ) -> Vec<QuiltStoreBlob> {
+        blobs_with_paths
+            .iter()
+            .enumerate()
+            .map(|(i, (path, blob))| {
+                QuiltStoreBlob::new(
+                    blob,
+                    path.file_name()
+                        .and_then(|file_name| file_name.to_str())
```
mlegner, line 371: Important: We should have some way of handling duplicate filenames. 

One idea would be to use relative paths to some specified path (we have already deduplicated the file paths).

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -0,0 +1,583 @@
...
+            .await?;
+
+        Ok(result)
+    }
+
+    /// Encodes the blobs to a quilt and stores it to Walrus.
+    #[tracing::instrument(skip_all, fields(blob_id))]
+    pub async fn reserve_and_store_quilt<V: QuiltVersion>(
+        &self,
+        quilt_store_blobs: &[QuiltStoreBlob<'_>],
```
mlegner: Suggestion: Change this to `V::Quilt` (removing the quilt construction below), then change the `reserve_and_store_quilt_from_paths` to just call `construct_quilt_from_paths` and `reserve_and_store_quilt`.

mlegner (commented):
Thanks a lot for the PR, @liquid-helium, really nice! As usual, a few mostly minor comments and suggestions.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Suggestion applied.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Suggestion applied.

crates/walrus-sdk/src/client/responses.rs
liquid-helium: Good point, Removed `Deserialize`, did some search on why we can use only `Serialize ` as well

crates/walrus-sdk/src/error.rs
liquid-helium: Thanks for the tip, a new error kind added.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Some special files like `Unix Domain Sockets`, or other devices.
Giving it a second thought, it also makes sense to ignore these cases, since the function is clear that we only collect `files`.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Good idea, change applied.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Nice observation!

I moved `read_blob_from_file` to `walrus_utils` to share it for the two crates.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Nice catch, change applied.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: I removed the error, to silently ignore all special files in both functions.

Yes, there are some code duplicates. Originally I used a single function to go through the paths recursively, and read the files directly, but it has deeper nested loops.

I keep it as for the rest, but let me know if you have strong prefreferences.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Change applied.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Good point, added a TODO here.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: It is generic, corrected the doc string, thanks for the catch!

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Good catch!

I've given this some thought as well, and it involves a trade-off. Using the full path is likely the most stable option, but it can include a lot of unnecessary details or even sensitive information, like an account number embedded in the full path. Since we allow reading files from different directories, relying on relative paths could lead to duplicates as well.

That's why I opted to use only file names for now. Looking ahead, when we introduce support for per-blob attributes, I hope the input will be 'rich' by default. This will enable users to assign custom identifiers, providing more flexibility.

But I am open to other approaches, just keeping it as for now and see where we go.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Good idea! Change applied.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Good idea! Change applied.

crates/walrus-sdk/src/client/quilt_client.rs
mlegner: Thanks for the insights. I agree that this is a tricky point, and I also agree that using absolute paths leaks too much information.

My solution would be to have the user provide a path to which all other paths are taken relative. Something like `walrus store-quilt --paths-relative-to ~/my/base/dir ~/my/base/dir/some/dir/file1 ~/my/base/file2 ...`. This would then use `some/dir/file1` and `../file2` as identifiers. This should make sure that there are no duplicates, as we deduplicate paths. 

I think it's ok to use filenames in this PR, but I'd like to give the user a bit more control over this in a subsequent PR. For the time being, can you mention that filenames need to be unique?

crates/walrus-sdk/src/client/quilt_client.rs
mlegner: Suggestion: With the following changes, we can avoid having the same file/dir logic twice:
```rust
pub fn read_blobs_from_paths<P: AsRef<Path>>(paths: &[P]) -> ClientResult<Vec<(PathBuf, Vec<u8>)>> {

    // <snip>

    for path in paths {
        let path = path.as_ref();

        // Validate path existence and accessibility.
        if !path.exists() {
            return Err(ClientError::from(ClientErrorKind::Other(
                format!("Path '{}' does not exist.", path.display()).into(),
            )));
        }

        collected_files.extend(get_all_files_from_path(path)?);
    }

    // <snip>

}
```

```rust
fn get_all_files_from_path<P: AsRef<Path>>(path: P) -> ClientResult<HashSet<PathBuf>> {
    let path = path.as_ref();
    let mut collected_files: HashSet<PathBuf> = HashSet::new();

    if path.is_file() {
        collected_files.insert(path.to_owned());
    } else if path.is_dir() {
        for entry in fs::read_dir(path).map_err(ClientError::other)? {
            let current_entry_path = entry.map_err(ClientError::other)?.path();
            collected_files.extend(get_all_files_from_path(&current_entry_path)?);
        }
    }

    Ok(collected_files)
}
```

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -189,14 +215,14 @@ impl<T: ReadClient> QuiltClient<'_, T> {
...
+        first_sliver: &SliverData<V::SliverAxis>,
     ) -> ClientResult<QuiltIndex>
     where
         SliverData<V::SliverAxis>: TryFrom<Sliver>,
     {
         let mut all_slivers = Vec::new();
-        let mut refs = Vec::new();
-        let first_sliver_refs: Vec<_> = first_sliver.iter().collect();
+        let mut refs = [first_sliver].to_vec();
+        let first_sliver_refs = [first_sliver].to_vec();
```
mlegner: Question: (Why) do we need these three separate vectors?

Also, nit: You can simplify `[x].to_vec()` to `vec![x]`.

mlegner (approved):
Thanks a lot, @liquid-helium, LGTM! 💯

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: 👍 change applied.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Now I see what you were referring to by relative path, that's a good idea, I added WAL-887 to track it.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Those are to avoid the Rust borrowing rules, I didn't find a better solution.
Change applied.

PR #2175, opened 2025-06-05
feat(backfill): add initial backfill function

giac-mysten:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -76,7 +76,7 @@ pub(crate) use crate::utils::{CompletedReasonWeight, WeightedFutures};
 use crate::{
     active_committees::ActiveCommittees,
     config::CommunicationLimits,
-    error::{ClientError, ClientErrorKind, ClientResult},
+    error::{ClientError, ClientErrorKind, ClientResult, StoreError},
     store_when::StoreWhen,
     utils::{WeightedResult, styled_progress_bar, styled_spinner},
 };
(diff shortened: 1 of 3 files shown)


mlegner:
@giac-mysten @wbbradley We should probably try to get this merged into `main` in some form so it isn't lost for the future. WDYT?

crates/walrus-service/src/client/cli/backfill.rs
```diff
@@ -0,0 +1,444 @@
...
+        Err(error) => {
+            tracing::error!(
+                ?error,
+                ?blob_filename,
+                "error reading blob from disk. skipping..."
+            );
+        }
+    }
+    // Remove the blob file after processing, even if the backfill fails.
+    let _ = std::fs::remove_file(blob_filename);
```
halfprice, line 420: nit: this seems quite important because if we don't remove failed backfill, this may generate backpressure. can we document this here?

scripts/backfill.sh
```diff
@@ -0,0 +1,73 @@
...
+#
+# The script will spawn multiple processes in pairs. In each pair, one process pull the blobs from
+# GCS into a subdirectory based on the blob's prefix, and the other process encodes the blobs and
+# pushes the slivers to the node to backfill.
+#
+# Before running the script, set the node IDs for the storage nodes to be backfilled, and the name
+# of the GCS bucket in which the backup is stored.
+
+root_dir=/tmp/backfill
+all_blobs_file="$root_dir"/all-blobs.txt
```
halfprice, line 15: question: where/how do we get all-blobs.txt?

crates/walrus-service/src/client/cli/backfill.rs
```diff
@@ -0,0 +1,444 @@
...
+        tracing::trace!(?blob_id, "Blob ID does not match prefix, skipping");
+        return Ok(());
+    }
+    let blob_id: BlobId = blob_id.parse()?;
+
+    // Check if the blob has already been pulled.
+    if pulled_blobs.contains(&blob_id) {
+        tracing::info!(?blob_id, "Blob already exists, skipping download");
+        // Emit the blob ID to stdout for further downstream processing.
+        println!("{blob_id}");
```
halfprice: nit: is this still needed?

crates/walrus-service/src/client/cli/backfill.rs
```diff
@@ -0,0 +1,444 @@
...
+    match store.get(&blob_id.to_string().into()).await {
+        Ok(object) => {
+            // Write the blob to the specified backfill directory.
+            let blob_filename = PathBuf::from(&backfill_dir).join(blob_id.to_string());
+            let mut file = File::create(&blob_filename)?;
+            let bytes: Bytes = object.bytes().await?;
+            file.write_all(&bytes)?;
+            drop(file);
+
+            // Emit the blob ID to stdout for further downstream processing.
```
halfprice: same as above

halfprice (approved):
Thanks @giac-mysten and @wbbradley for implementing the backfill tool! This looks quite good!

One general nit comment is that the pushed_state and pulled_state can potentially become a bottleneck if the number of blobs are large, and I think we only need to store a watermark if the input blobs are sorted. Don't need to address this, can just document it if you agree.

crates/walrus-service/src/client/cli/backfill.rs
giac-mysten: excellent point!

scripts/backfill.sh
giac-mysten: In our previous usages, this list was compiled out of the blobs stored on the backup by @wbbradley

crates/walrus-service/src/client/cli/backfill.rs
giac-mysten: No, as we don't read the blobs to be backfilled from stdin anymore. I thought there was no harm in keeping this, but I'll remove it.

crates/walrus-service/src/client/cli/backfill.rs
giac-mysten: same as above :)

PR #2176, opened 2025-06-05
chore: bump sui version to v1.49.2

wbbradley:
## Description

Bump Sui version to v1.49.2. Also update cargo-simtest to be inline with recent changes to that script.

## Test plan

CI Pipeline.


mlegner:
Thanks for opening this, @wbbradley! `testnet-v1.50.1` should be released today, so we should be able to use that one.

PR #2177, opened 2025-06-06
feat: non-blocking event processing

halfprice:
## Description

This PR improves storage node event processing speed by moving event action execution to a set of background workers.

Currently, processing certified events is somewhat slow due to blob data existence check, and sequentially processing
events makes event catchup slow. This is particularly problematic for storage nodes with lagging highest persisted
event. After restart, such node needs to reprocess events from the highest persisted event, and may take long time
to reach to the latest event.

By moving all but blob info update to background workers, the node can bring the blob info table to the latest state
faster, and therefore, reduce the window of highest processed event catchup.

This PR also adds a fallback config as if setting NodeBlobEventProcessorConfig::num_workers to 0, it fallback
to the prior sequential event processing logic.

## Test plan

General simtest coverage
Added a simtest to test event ordering.
Added a dedicated integration test to test overlapping event processing.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -3,6 +3,8 @@
 
 use std::sync::Arc;
 
+use rand::Rng;
+use sui_macros::fail_point_async;
 use tokio::{
     sync::mpsc::{self, UnboundedReceiver, UnboundedSender},
     task::JoinHandle,
@@ -90,6 +92,7 @@ impl BackgroundEventProcessor {
         event: BlobCertified,
     ) -> anyhow::Result<()> {
         let start = tokio::time::Instant::now();
+
         let histogram_set = self.node.metrics.recover_blob_duration_seconds.clone();
 
         if !self.node.is_blob_certified(&event.blob_id)?
@@ -109,6 +112,8 @@ impl BackgroundEventProcessor {
 
             return Ok(());
         }
+
+        fail_point_async!("fail_point_process_blob_certified_event");
 
         // Slivers and (possibly) metadata are not stored, so initiate blob sync.
         self.blob_sync_handler
@@ -187,9 +192,11 @@ pub struct BlobEventProcessor {
 }
 
 impl BlobEventProcessor {
-    pub fn new(node: Arc<StorageNodeInner>, blob_sync_handler: Arc<BlobSyncHandler>) -> Self {
-        let num_workers = 4;
-
+    pub fn new(
+        node: Arc<StorageNodeInner>,
+        blob_sync_handler: Arc<BlobSyncHandler>,
+        num_workers: usize,
+    ) -> Self {
         let mut senders = Vec::with_capacity(num_workers);
         let mut workers = Vec::with_capacity(num_workers);
         for _ in 0..num_workers {
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -3,7 +3,6 @@
 
 use std::sync::Arc;
 
-use rand::Rng;
 use sui_macros::fail_point_async;
 use tokio::{
     sync::mpsc::{self, UnboundedReceiver, UnboundedSender},
@@ -24,6 +23,8 @@ use crate::node::{
     system_events::CompletableHandle,
 };
 
+/// Background event processor that processes blob events in the background. It processes events
+/// sequentially based on the order of events in the channel.
 #[derive(Debug)]
 struct BackgroundEventProcessor {
     node: Arc<StorageNodeInner>,
@@ -44,21 +45,29 @@ impl BackgroundEventProcessor {
         }
     }
 
-    async fn run(&mut self) {
-        // TODO: add metrics to measure events stay in the channel (queue length).
+    /// Runs the background event processor.
+    async fn run(&mut self, worker_index: usize) {
         while let Some((event_handle, blob_event)) = self.event_receiver.recv().await {
+            walrus_utils::with_label!(
+                self.node.metrics.pending_processing_blob_event_in_queue,
+                &worker_index.to_string()
+            )
+            .dec();
+
             if let Err(e) = self.process_event(event_handle, blob_event).await {
+                // TODO: to keep the same behavior as before BackgroundEventProcessor, we should
+                // propagate the error to the node and exit the process if necessary.
                 tracing::error!("error processing blob event: {}", e);
             }
         }
     }
 
+    /// Processes a blob event.
     async fn process_event(
         &self,
         event_handle: EventHandle,
         blob_event: BlobEvent,
     ) -> anyhow::Result<()> {
-        // TODO: what should happen to the processor if this returns error?
         match blob_event {
             BlobEvent::Certified(event) => {
                 monitored_scope::monitored_scope("ProcessEvent::BlobEvent::Certified");
@@ -85,6 +94,7 @@ impl BackgroundEventProcessor {
         Ok(())
     }
 
+    /// Processes a blob certified event.
     #[tracing::instrument(skip_all)]
     async fn process_blob_certified_event(
         &self,
@@ -126,6 +136,7 @@ impl BackgroundEventProcessor {
         Ok(())
     }
 
+    /// Processes a blob deleted event.
     #[tracing::instrument(skip_all)]
     async fn process_blob_deleted_event(
         &self,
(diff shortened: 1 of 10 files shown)


crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -0,0 +1,311 @@
...
+    async fn run(&mut self, worker_index: usize) {
+        while let Some((event_handle, blob_event)) = self.event_receiver.recv().await {
+            walrus_utils::with_label!(
+                self.node.metrics.pending_processing_blob_event_in_queue,
+                &worker_index.to_string()
+            )
+            .dec();
+
+            if let Err(e) = self.process_event(event_handle, blob_event).await {
+                // TODO: to keep the same behavior as before BackgroundEventProcessor, we should
```
halfprice: I'll create a linear issue if the current code looks reasonable.

crates/walrus-service/src/test_utils.rs
```diff
@@ -2802,12 +2803,17 @@ pub fn storage_node_config() -> WithTempDir<StorageNodeConfig> {
             // Turn on all consistency checks in integration tests.
             consistency_check: StorageNodeConsistencyCheckConfig {
                 enable_consistency_check: true,
-                enable_sliver_data_existence_check: true,
+                // TODO: re-enable blob data consistency check by tracking epoch in the event.
```
halfprice: I'll create a linear issue if the current code looks reasonable.

crates/walrus-service/src/node/config.rs
```diff
@@ -188,6 +188,9 @@ pub struct StorageNodeConfig {
     /// Configuration for node recovery.
     #[serde(default, skip_serializing_if = "defaults::is_default")]
     pub node_recovery_config: NodeRecoveryConfig,
+    /// Configuration for the blob event processor.
+    #[serde(default, skip_serializing_if = "defaults::is_default")]
+    pub node_blob_event_processor_config: NodeBlobEventProcessorConfig,
```
mlegner: Nit: I think the `node_` prefix here isn't necessary.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -0,0 +1,311 @@
...
+                    .expect(
+                        "sequential processor must be configured when no background \
+                            workers are configured",
+                    )
+                    .process_event(event_handle, blob_event)
+                    .await?;
+            } else {
+                // We send the event to one of the workers to process in parallel.
+                // Note that in order to remain sequential processing for the same BlobID, we always
+                // send events for the same BlobID to the same worker.
```
mlegner: Praise: I like this.

Remark (maybe to add to the comment): We need to be careful if we ever want to dynamically adjust the number of workers. (Not saying that we should do that, just to be sure.)

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -0,0 +1,311 @@
...
+            // If we want to do this in parallel, we shouldn't mix registered event processing with
+            // certified event processing, as certified events take longer and can block following
+            // registered events.
+            monitored_scope::monitored_scope("ProcessEvent::BlobEvent::Registered");
+            event_handle.mark_as_complete();
+        } else {
+            // If there are no background workers, we use a sequential processor to process events
+            // sequentially.
+            if self.background_processor_senders.is_empty() {
+                assert!(self.sequential_processor.is_some());
```
mlegner: Nit: Can we document this as an invariant of the struct?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -0,0 +1,311 @@
...
+        if let BlobEvent::Registered(_) = &blob_event {
+            // Registered event is marked as complete immediately. We need to process registered
+            // events as fast as possible to catch up to the latest event in order to not miss
+            // blob sliver uploads.
+            //
+            // If we want to do this in parallel, we shouldn't mix registered event processing with
+            // certified event processing, as certified events take longer and can block following
+            // registered events.
+            monitored_scope::monitored_scope("ProcessEvent::BlobEvent::Registered");
+            event_handle.mark_as_complete();
```
mlegner, line 278: Nit: We could avoid one level of indentation below if we add an `return Ok(());` here.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -0,0 +1,311 @@
...
+                rx,
+            )))
+        } else {
+            None
+        };
+
+        Self {
+            node,
+            background_processor_senders: senders,
+            _background_processors: workers,
```
mlegner, line 251: Question: How are these shut down when the node is stopped? What happens if any of them fail?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -0,0 +1,311 @@
...
+pub struct BlobEventProcessor {
+    node: Arc<StorageNodeInner>,
+
+    // Background processors that process events in parallel.
+    background_processor_senders: Vec<UnboundedSender<(EventHandle, BlobEvent)>>,
+    _background_processors: Vec<Arc<JoinHandle<()>>>,
+
+    // When there are no background workers, we use a sequential processor to process events using
+    // this processor. This is to keep the same behavior as before BackgroundEventProcessor.
+    sequential_processor: Option<Arc<BackgroundEventProcessor>>,
```
mlegner: Nit/remark/suggestion (maybe for a follow-up): Given that either one or the other is set, this would be a good match for an enum instead of multiple fields.
OTOH, we will probably remove the sequential option anyway (or use a single background process for it), so not worth investing many efforts in it.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -0,0 +1,311 @@
...
+                BackgroundEventProcessor::new(node.clone(), blob_sync_handler.clone(), rx);
+            workers.push(Arc::new(tokio::spawn(async move {
+                background_processor.run(worker_index).await;
+            })));
+        }
+
+        let sequential_processor = if num_workers == 0 {
+            // Create a sequential processor to process events sequentially if no background workers
+            // are configured.
+            let (_tx, rx) = mpsc::unbounded_channel();
```
mlegner, line 238: Nit: It feels a bit unclean to have this receiver that isn't actually used. I think it's fine though, as we probably will remove the sequential processing altogether soon, right?

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Good suggestion about using a enum! And also good point about removing it in the future, which seems that keeping in the current form is easier to do so.

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: My original thinking was that I don't want to have an Option<Receiver> in BackgroundEventProcessor, which makes user's responsibility to use BackgroundEventProcessor correctly. It's also a good point that we will remove it in the future so it seems better to keep Receiver in BackgroundEventProcessor required.

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Done

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Done

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Thanks! Good point and added comments.

crates/walrus-service/src/node/config.rs
halfprice: Removed

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: If any of them fails, currently the worker just stops and do not populate errors back to the node. WAL-874 is about addressing this issue. Since so far we have not seen any event processing error very often, and this adds complexity in this PR, I don't think we need to add it in 1.26

Regarding how are these shut down, these will be shut down if the runtime shuts down. I don't know if we need a graceful shutdown mechanism. I tried it in PTN and restarted the node a couple of times and don't see any error/panic. I added WAL-876 for this, but doesn't seem to be too urgent.

halfprice (commented):
Thanks a lot @mlegner for the detailed review on a Saturday! I've addressed all of your comments and added corresponding TODOs.

PR #2178, opened 2025-06-06
feat(backfill): implement backpressure

giac-mysten:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/cli/backfill.rs
+++ b/crates/walrus-service/src/client/cli/backfill.rs
@@ -288,10 +288,12 @@ pub(crate) async fn run_blob_backfill(
     loop {
         let entries = std::fs::read_dir(&backfill_dir)?
             .filter_map(|entry| {
-                entry.as_ref().ok().map(|entry| entry.path()).or_else(|| {
-                    tracing::error!(?entry, "error reading directory entry; skipping...");
-                    None
-                })
+                entry
+                    .inspect_err(|error| {
+                        tracing::error!(?error, "error reading directory entry; skipping...")
+                    })
+                    .ok()
+                    .map(|entry| entry.path())
             })
             .collect::<Vec<_>>();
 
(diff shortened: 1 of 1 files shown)


mlegner (commented):
Very nice and simple solution!

We should probably document somewhere that this requires the corresponding directory to be used by a single reader/writer pair.

crates/walrus-service/src/client/cli/backfill.rs
```diff
@@ -92,29 +97,88 @@ pub(crate) async fn pull_archive_blobs(
...
+///
+/// Checks if the number of files in the backfill directory is greater than the
+/// `MAX_IN_FLIGHT_BACKFILLS` limit. If so, it waits for `BACKPRESSURE_WAIT_TIME` and checks again,
+/// until the number of files is less than the limit.
+async fn check_and_apply_backpressure(backfill_dir: &str) -> Result<()> {
+    loop {
+        // We have processed a full batch. Now check if we need to backpressure.
+        let num_files = std::fs::read_dir(&backfill_dir)?.count();
+        if num_files > MAX_IN_FLIGHT_BACKFILLS {
+            tracing::info!(
```
mlegner, line 142: Nit: This could be a `debug` and the debug below could be a `trace`.

crates/walrus-service/src/client/cli/backfill.rs
mlegner, line 1: If we don't use the stdin on the writer job, we can remove `println!("{}", blob_id);` on line 206.

crates/walrus-service/src/client/cli/backfill.rs
```diff
@@ -213,49 +283,90 @@ pub(crate) async fn run_blob_backfill(
...
+                blob_filename,
+                &mut pushed_blobs,
+                &mut pushed_state,
+            )
+            .await;
+        }
+
+        // Small delay to avoid constant reads if the directory is empty.
+        if entries.is_empty() {
+            tracing::info!("no more blobs to backfill; sleeping for 500ms");
```
mlegner, line 313: What would be the stop signal for the write job? Assuming the read job is much faster, we could use an empty directory as the signal. Or have a timeout where it stops after the directory is empty for some time.

mlegner (approved):
Very nice, @giac-mysten, looks great!

PR #2179, opened 2025-06-06
fix(node): set latest_event_epoch correctly when restarting recovery

mlegner:
## Description

A node restarting in recovery mode reads the `latest_event_epoch` before this is set properly (initially set to `0`).

This is a simple fix. A more robust solution for the future could be to store that in the DB and load it on start.

## Test plan

TBD.


halfprice (approved):
Thanks a lot for the fix @mlegner !

PR #2180, opened 2025-06-07
test: add test coverage of node restart during recovery

halfprice:
## Description

Adds the missing tests for the fix in #2179.
 
## Test plan

Added in this PR.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1474,6 +1474,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "bimap"
+version = "0.6.3"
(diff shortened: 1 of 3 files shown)


Cargo.lock
mlegner, line 1: Remark: I think these changes are unrelated to this PR. I believe a previous PR somehow missed them.

mlegner (approved):
Thanks a lot for adding this test coverage! 💯 

Have you checked that this fails without #2179?

halfprice:
> Have you checked that this

Yes, I did. Without #2179, this test fails pretty quickly using seed search. After #2179 , 100 seeds passed.

PR #2187, opened 2025-06-10
feat: added restore

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/bin/node.rs
```diff
@@ -191,6 +191,27 @@ enum AdminCommands {
         #[command(subcommand)]
         command: CheckpointCommands,
     },
+    /// Restore the database from a checkpoint.
+    Restore {
+        /// The path where the checkpoint will be created. If not specified, the checkpoint will be
+        /// created in the `checkpoint_dir` specified in [`StorageNodeConfig::checkpoint_config`].
+        #[arg(long)]
+        #[serde(default)]
+        db_checkpoint_path: PathBuf,
```
halfprice, line 200: Question: can this field not be specified? Should it be `Option<PathBuf>`?

crates/walrus-service/src/node/errors.rs
```diff
@@ -411,6 +411,10 @@ pub enum DbCheckpointError {
     CheckpointCreationError(String),
     #[error("Backup already in progress")]
     CheckpointInProgress,
+    #[error("No checkpoint found")]
+    NoCheckpointFound,
+    #[error("Restore error: {0}")]
+    RestoreError(String),
```
halfprice: Nit: maybe we can do `RestoreError(#[from] rocksdb::Error),` to preserve the error struct instead of using a string.

halfprice (approved):
🚀🚀🚀🚀🚀

crates/walrus-service/bin/node.rs
liquid-helium: I thought about it, but I lean to force the users to specify the backup dir, since they could have moved the backup around, and also to make sure the users understand exactly what they are restoring.
Keeping it as, but let me know if you have a strong preference.

crates/walrus-service/src/node/errors.rs
liquid-helium: Good idea, change applied.

crates/walrus-service/bin/node.rs
halfprice: Sorry I was not clear. The doc says "If not specified, ...". Maybe just remove that part from the doc.

PR #2193, opened 2025-06-12
feat(node): enable node starting with incomplete event history

mlegner:
## Description

The current blob status tracking implemented in #764 assumes a complete history of blob events from the start of the Walrus system. However, at a steady state of Walrus, we will only have a limited history starting 2 years in the past.

This means that we may see "certified" events without related previous "registered" events, etc.

To make the status tracking compatible with this setting, this PR does the following:
- Add a new node status `RecoveryCatchUpWithIncompleteHistory`.
- This status is set when a new node is started when some event blobs are already expired.
- In this status, the node processes events differently.
  - Only update the status for events where the expiration epoch is in the future.
  - Treat blob extensions where the expiration epoch changes from a past epoch to an epoch in the future as a new blob (registered+certified).

This also contains some refactoring and test-setup improvements (mostly isolated in a separate commit).

## Test plan

Added a simtest. Open TODOs:
- [ ] The new test currently fails for some event blobs, this needs to be fixed.
- [ ] We should also test crashes during the recovery process.

---

## Release notes

- [x] Storage node: New storage nodes can start in epochs later than `MAX_EPOCHS_AHEAD` (53 on the current Testnet and Mainnet) without any additional action.


crates/walrus-service/src/node/events.rs
```diff
@@ -297,3 +299,13 @@ impl Ord for EventStreamCursor {
         self.element_index.cmp(&other.element_index)
     }
 }
+
+/// A stream of events with starting indices for event processing and the event-blob writer..
```
sadhansood: This statement is ending with two periods, i think this is probably un-intentional.

crates/walrus-service/src/node.rs
```diff
@@ -843,66 +846,144 @@ impl StorageNode {
...
+            panic!("recovery with incomplete event history is only possible for fresh nodes");
+        }
+
+        #[cfg(msim)]
+        sui_macros::fail_point!("fail_point_recovery_with_incomplete_history");
+
+        let current_epoch = self.inner.committee_service.get_epoch();
+        let first_complete_epoch =
+            current_epoch + 1 - self.inner.system_parameters.max_epochs_ahead;
+        if first_complete_epoch < init_state.epoch
```
sadhansood: nit: Consider adding a comment about `first_complete_epoch`. Perhaps: `first_complete_epoch` is the oldest epoch for which a node joining the network can be absolutely guaranteed that the historical data is still available and hasn't been pruned yet.

crates/walrus-service/src/node.rs
```diff
@@ -843,66 +846,144 @@ impl StorageNode {
...
+        // || first_complete_epoch == init_state.epoch
+        //     && todo!("first event is not an EpochChangeStart")
```
sadhansood: These commented lines are perhaps un-intentional.

sadhansood (commented):
@mlegner the change looks pretty good and well executed to me. I only have a few minor nits. One question which is un-answered for me - how do we handle seeing the blob certified event without seeing the corresponding register? I see that we handle the blob extension event by creating corresponding missing blob register and blob certified updates in blob_info table on the fly - but don't we need to do the same for a regular blob certified event (with missing corresponding register in blob_info table)?

crates/walrus-service/src/node.rs
mlegner: Actually, this is a missing check, although thinking about it, this may not be needed: I don't think it's possible for the first event in the incomplete history to the `EpochChangeStart` for the first complete epoch. 

Assume we have epoch `E` as the first complete epoch and that event blob `B` has the `EpochChangeStart` for epoch `E` as its first event. This means the previous event blob `B-1` cannot have been certified in epoch `E-1` as we would otherwise miss its registration and certification events. This means that `B-1` must have been certified in epoch `E` (or later). So we definitely have events for epoch `E-1` in our incomplete history.

I will describe this logic in a comment. Please let me know if I'm missing anything.

mlegner:
> One question which is un-answered for me - how do we handle seeing the blob certified event without seeing the corresponding register? I see that we handle the blob extension event by creating corresponding missing blob register and blob certified updates in blob_info table on the fly - but don't we need to do the same for a regular blob certified event (with missing corresponding register in blob_info table)?

Thanks so much for the review, @sadhansood!

The logic for "standard" certified events is the following: If a certified event is relevant for us, it means its end epoch is later than the `epoch_at_start`. This means that the blob must have been registered in `epoch_at_start - MAX_EPOCHS_AHEAD + 1 = first_complete_epoch` or later. This means we *have already processed* the corresponding registration event (which has the same end epoch).

If you agree with this logic, I'll add a comment explaining this to the code.

crates/walrus-service/src/node.rs
mlegner: Done in f10929cf8b5c3f407502214c5a23ddf7626b24d1.

crates/walrus-service/src/node.rs
mlegner: Changed the check and added a comment in f10929cf8b5c3f407502214c5a23ddf7626b24d1. WDYT?

crates/walrus-service/src/node.rs
```diff
@@ -843,66 +846,152 @@ impl StorageNode {
...
+            // `EpochChangeStart` for epoch `E` as its first event. This means the previous event
+            // blob `B-1` cannot have been certified in epoch `E-1` as we would otherwise miss its
+            // registration and certification events. This means that `B-1` must have been certified
+            // in epoch `E` (or later). So we definitely have events for epoch `E-1` in our
+            // incomplete history.
+            init_state.epoch < first_complete_epoch,
+            "inconsistent event-blob state: we do not have access to all events from epoch {}",
+            first_complete_epoch
+        );
+
```
sadhansood: Nice reasoning which makes sense, and it validates a critical safety check before the node repositions its state. It took me a little bit to convince myself but I agree with this. Lmk if i understood it properly: Assume current epoch is 11, and `MAX_EPOCHS_AHEAD` is 10. The first complete epoch would be epoch 2. Worst case, we created an event blob with first event as epoch start event for epoch 2 (`B`), and naturally the event blob `B-1` had only events from epoch 1. `B - 1` has to be registered and certified in epoch 2. So we do indeed see blob `B - 1` in our complete history and hence we must be able to read `B - 1` and thus we must see events from epoch 1. So init state epoch must be from one before the epoch with complete history.

crates/walrus-service/src/common/event_blob_downloader.rs
```diff
@@ -123,39 +122,43 @@ impl EventBlobDownloader {
...
-                    } else {
-                        return Err(err.into());
-                    }
+                Err(err) if matches!(err.kind(), ClientErrorKind::BlobIdDoesNotExist) => {
+                    BlobStatus::Nonexistent
                 }
+                Err(err) => return Err(err.into()),
             };
 
             if blob_status == BlobStatus::Nonexistent {
```
halfprice: Suggestion/thoughts: should we add a safety check here that when this happens, we must have downloaded at least 52 event blobs? The current read relies on Walrus read API and as we've seen in other cases reading after immediate write may return Nonexistent.

sadhansood (approved):
Thanks Markus, really appreciate the high quality of this contribution. Solving the problem of bootstrapping from an incomplete history is pretty huge. I want to highlight the clean design, smart on the fly creation of `register` and `certify` operations to handle extension of untracked blobs and detailed reasoning and safety checks. Great work!

crates/walrus-service/src/node.rs
```diff
@@ -843,66 +846,152 @@ impl StorageNode {
...
-                init_state
-                    .event_cursor
-                    .event_id
-                    .unwrap_or(EVENT_ID_FOR_CHECKPOINT_EVENTS),
-                actual_event_index,
-            )?;
-            storage_node_cursor_repositioned = true;
+        self.inner
+            .latest_event_epoch
+            .store(init_state.epoch, Ordering::SeqCst);
```
halfprice: Question: I think I understand the motivation for updating the `latest_event_epoch`(so that first event won't get filtered), but this is also not necessary? This is only used if the the node is in RecoveryCatchUpWithIncompleteHistory, and `init_state.epoch` is guaranteed to be smaller than `first_complete_epoch`(there is also an assertion below). So the uninitialized `latest_event_epoch` = 0 also serves the purpose of filtering events?

crates/walrus-service/src/node.rs
```diff
@@ -843,66 +846,152 @@ impl StorageNode {
...
+        next_event_index: u64,
+        init_state: InitState,
+    ) -> anyhow::Result<Option<EventStreamCursor>> {
+        let first_available_event_index = init_state.event_cursor.element_index;
+        if next_event_index >= first_available_event_index {
+            // Standard case: the node is not behind the first available event.
+            return Ok(None);
+        }
+        if next_event_index != 0 {
+            panic!("recovery with incomplete event history is only possible for fresh nodes");
```
halfprice: Nit: this is true for mainnet nodes, but may not be true for PTN nodes. I think PTN nodes can easily lag for more than 52 epochs. Maybe update the message to something like `if a node is lagging for more than MAX_EPOCHS_AHEAD, it should recovery from scratch`.

crates/walrus-service/src/node.rs
```diff
@@ -843,66 +846,152 @@ impl StorageNode {
...
+        // If we reach this point, we have to reposition at least one of the event cursors.
+        // Important: We need to make sure we start the event stream from the correct cursor.
+
+        // Reposition the node event cursor if it is behind the first available event.
+        let new_event_cursor = self
+            .reposition_event_cursor_if_starting_with_incomplete_history(
+                storage_node_cursor.element_index,
+                init_state.clone(),
+            )?
+            .unwrap_or(event_cursor);
```
halfprice: Question: is the new code take the case that `writer_starting_index is below first_available_event_index, but storage_cursor is above first_available_event_index` case? It seems that after `reposition_event_cursor_if_starting_with_incomplete_history `, it may also need to compare with the newly positioned event blob writer cursor?

This is to account for the case where storage cursor is not lagging, but event blob writer needs to re-write event blob from incomplete history.

cc @sadhansood to double check

crates/walrus-service/src/node.rs
mlegner: Yes, that is exactly right. 👍

mlegner:
Thanks so much for the review and the praise, @sadhansood! 😊 

I'm now going to fix the simtest and wait if @halfprice has additional comments.

crates/walrus-service/src/node.rs
```diff
@@ -1028,6 +1120,31 @@ impl StorageNode {
...
+            return Ok(true);
+        };
+        if self.inner.current_event_epoch() >= first_epoch {
+            return Ok(true);
+        }
+        if let EventStreamElement::ContractEvent(ContractEvent::EpochChangeEvent(
+            EpochChangeEvent::EpochChangeStart(EpochChangeStart { epoch, .. }),
+        )) = &stream_element.element
+        {
+            if *epoch >= first_epoch {
```
halfprice, line 1179: Nit: maybe add a comment here that processing EpochChangeStart event will update self.inner.current_event_epoch(), so that it's clear why we need the previous if statement.

crates/walrus-service/src/node.rs
```diff
@@ -1109,31 +1226,31 @@ impl StorageNode {
...
         };
 
-        // For blob extension events, the epoch is the event's original
-        // certified epoch, and not the current epoch. Skip node lagging check
-        // for blob extension events.
-        if event.is_blob_extension() {
+        let Some(event_epoch) = event.event_epoch() else {
+            // For blob extension events, the epoch is the event's original
+            // certified epoch, and not the current epoch. Skip node lagging check
+            // for blob extension events.
```
halfprice: Nit: I think this comment is no longer accurate if blob extension event does not contain an event_epoch()

crates/walrus-service/src/node/storage.rs
```diff
@@ -86,13 +91,16 @@ pub(crate) fn node_status_options(db_config: &DatabaseConfig) -> Options {
...
-//         \            /          ^       |
-//          \          /            \      |
-//           v        v              \     v
-//          RecoverMetadata  -------> Active
-//
+///
+/// ```text
+///    RecoveryCatchUpWithIncompleteHistory
+///      /                              \
+///     v                                v
```
halfprice: Nit: the left arrow should point up?

crates/walrus-service/src/node/storage.rs
```diff
@@ -105,8 +113,17 @@ pub enum NodeStatus {
...
+    /// The node is in recovery mode and recovering missing slivers. The included epoch is the
+    /// epoch in which the node started recovering.
     RecoveryInProgress(Epoch),
+    /// The node is in recovery mode and catching up with the chain, but the history is incomplete
+    /// due to expired event blobs.
+    RecoveryCatchUpWithIncompleteHistory {
+        /// The first epoch for which all events are available and relevant.
+        first_complete_epoch: Epoch,
+        /// The epoch at which the node started recovering.
+        epoch_at_start: Epoch,
```
halfprice: Nit: can we document the function of these two fields so that it's clear why we include them?

IIUC: first_complete_epoch is used to filter events that before this epoch, and epoch_at_start is for processing events that have no effect up to this epoch?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -149,6 +161,120 @@ impl BlobInfoTable {
...
+    /// Updates the blob info for a blob based on the [`BlobEvent`] when the node is in recovery
+    /// with incomplete history.
+    ///
+    /// Only updates the info if the provided `event_index` hasn't been processed yet.
+    #[tracing::instrument(skip(self))]
+    pub fn update_blob_info_during_recovery_with_incomplete_history(
+        &self,
+        event_index: u64,
+        event: &BlobEvent,
+        first_epoch: Epoch,
```
halfprice: Nit: not used?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -149,6 +161,120 @@ impl BlobInfoTable {
...
+                tracing::debug!(
+                    "skip updating blob info for event with end epoch before epoch at start"
+                );
+                return Ok(());
+            }
+            BlobEvent::Registered(_)
+            | BlobEvent::Certified(BlobCertified {
+                is_extension: false,
+                ..
+            })
```
halfprice: Nit: I think it maybe worth documenting that why if a non-expiring certified event must have the corresponding registered event available.

Here is my understanding: the Certified event's epoch is the certified epoch, which maybe register event epoch +1, but the end epoch must calculated from the registered event epoch, not certified event epoch.

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -406,6 +415,181 @@ mod tests {
         clear_fail_point("fail_point_direct_shard_sync_recovery");
     }
 
+    #[ignore = "ignore integration simtests by default"]
+    #[walrus_simtest]
+    async fn test_recovery_with_incomplete_history() {
```
halfprice, line 424: Love the test!

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -149,6 +161,120 @@ impl BlobInfoTable {
...
+            | BlobEvent::Deleted(_)
+            | BlobEvent::InvalidBlobID(_)
+            | BlobEvent::DenyListBlobDeleted(_) => {
+                tracing::debug!("performing standard blob-info update for event");
+                return self.update_blob_info(event_index, event);
+            }
+            BlobEvent::Certified(event) => {
+                // Extensions need special handling.
+                event.clone()
+            }
```
halfprice, line 209: Suggestion (in a follow up PR): I think it worth adding an explicit test the case where certified event with earlier end_epoch is in an expired event blob, but extension event with later end_epoch is not. I'm not sure if the simtest, where extension is issued immediately after certification will exercise this code path.

halfprice (commented):
Great work @mlegner and this PR is amazing! I like the high level approach and my comments below are mostly small nits. PTAL.

crates/walrus-service/src/common/event_blob_downloader.rs
mlegner: Good point. I think this would be an extremely unlikely event, but I added a check that we downloaded at least one event blob. I don't think we can be sure about the number of blobs, as it's not guaranteed that there is at least one per epoch.

crates/walrus-service/src/node.rs
mlegner: Good point. This was indeed what I meant, but I'll elaborate a bit more.

In principle, it should be possible to do recovery with incomplete history with a node that isn't fresh, but it's probably not very relevant in practice (I added a TODO anyway).

crates/walrus-service/src/node.rs
mlegner: If the storage curser doesn't have to be repositioned, the `reposition_event_cursor_if_starting_with_incomplete_history` returns `Ok(None)`, such that `new_event_cursor == event_cursor`, which was set to `storage_node_cursor.min(event_blob_writer_cursor)`. So I do think we cover the case correctly, but let me know if I misunderstood anything.

crates/walrus-service/src/node/storage.rs
mlegner: Ah, it should probably point in both directions.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Indeed. Weird that Clippy didn't complain about this... 🤔

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: I've added a comment.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Created a follow-up issue and added a comment.

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -406,6 +415,169 @@ mod tests {
...
+            .check_storage_node_consistency_from_epoch(new_node_starting_epoch);
+
+        assert!(
+            catchup_using_event_blobs_triggered.load(Ordering::SeqCst),
+            "catchup using event blobs should be triggered"
+        );
+        assert!(
+            recovery_with_incomplete_history_triggered.load(Ordering::SeqCst),
+            "recovery with incomplete history should be triggered"
+        );
```
mlegner: @halfprice I wonder if I could/should remove this check? It seems to be a bit tricky to consistently trigger this path because there may be gaps of several epochs in which no event blobs are created.

mlegner: Disabled this check in 51ab2973eeb22f6d47added7779131c2ed05cdb5, but I'll re-enable it when I find a better way to set up the test.

crates/walrus-service/src/node.rs
halfprice: When reposition_event_cursor_if_starting_with_incomplete_history returns Ok(None), the `event_cursor` is positioned at `writer_starting_index`, but `writer_starting_index` can be lower than the first available index. Should the `new_event_cursor` be positioned to first available index in this case?

crates/walrus-simtest/tests/simtest.rs
halfprice: My current feeling is that keeping `recovery_with_incomplete_history_triggered` will be very valuable. This just to make sure that the incomplete history will definitely be exercised. It'll be sad that if later change somehow make this test to not exercise the code path completely, then we may lose coverage.

But it's ok to disable it for now and we can try to re-enable it later.

mlegner: I've created a follow-up issue for this.

crates/walrus-service/src/node.rs
mlegner: I agree, this is not necessary. Removed.

crates/walrus-service/src/node.rs
mlegner: > Should `the new_event_cursor` be positioned to first available index in this case?

Yes, that's what happens a few lines below:

```rust
        // Reposition the event blob writer if it is behind the first available event.
        if let Some(writer) = event_blob_writer {
            if writer_starting_index < first_available_event_index {
                tracing::info!(
                    "repositioning event blob writer cursor from {} to {}",
                    writer_starting_index,
                    first_available_event_index
                );
                writer.update(init_state.clone()).await?;
            }
        }
```

halfprice: Yeah, that only updates writer's position, but not `new_event_cursor`. Should we add

`new_event_cursor = init_state.event_cursor` after `writer.update(init_state.clone()).await?;`

mlegner: Ah... 🤦

Sorry, I kept misunderstanding you questions. I believe you are right. I'll try to refactor this function a bit to make it correct and a bit clearer than it is now.

mlegner: OK, I believe this should be fixed in 84525d2d2b19629e24895461b62e10fa8e5c00ca.

halfprice: LGTM!

halfprice (approved):

mlegner:
Unfortunately this wasn't merged before I left for my vacation. Will resolve conflicts and merge it today.

mlegner:
@halfprice I finally managed to resolve all the conflicts. The simtest succeeds, and I'm running more tests on the PTN. 

I wanted to ask you for another quick check but unfortunately the auto-merge was still enabled. 😬 
Please, take another quick look, I'll fix any findings in a follow-up PR. 🙏

halfprice:
> @halfprice I finally managed to resolve all the conflicts. The simtest succeeds, and I'm running more tests on the PTN.
> 
> I wanted to ask you for another quick check but unfortunately the auto-merge was still enabled. 😬 Please, take another quick look, I'll fix any findings in a follow-up PR. 🙏

Hi @mlegner, great job finishing this important feature! Sorry that it was a long day for me and I didn't get into this. Since it is not been released yet, and you also created a simtest, let's wait and see any issue appears during weekends, and I'll take a look next Monday (and hopefully addressing WAL-896 in the meantime).

crates/walrus-service/src/node.rs
```diff
@@ -857,96 +861,167 @@ impl StorageNode {
...
-
-        let init_state = self.inner.event_manager.init_state(event_cursor).await?;
-        let Some(init_state) = init_state else {
-            tracing::info!(
-                ?event_cursor,
-                "no init state found, resuming from event cursor"
-            );
-            return self.continue_event_stream_from_cursor(event_cursor).await;
+    ) -> anyhow::Result<EventStreamWithStartingIndices> {
+        // TODO(mlegner): Check these changes again.
```
halfprice, line 872: Nit: consider removing this comment :)

halfprice (commented):
Hi @mlegner , I just went through this PR again, and with #2294 , everything looks great. I payed special attention to `continue_event_stream` and `update_blob_info_during_recovery_with_incomplete_history`, and I couldn't spot any issue. Great job and nicely done!

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: Took me quite sometime to understand this again...

PR #2197, opened 2025-06-13
feat: Quilt client: CLI

liquid-helium:
## Description

Added CLI commands for storing/retrieving quilts.
Contributes to WAL-858

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [x] CLI:
  - Added walrus store-quilt command, to store multiple blobs as a single Quilt to Walrus
  - Added walrus read-quilt command to read one or more blobs in a Quilt from Walrus
  - Added walrus list-patches-in-quilt to check the blobs in a quilt


crates/walrus-service/src/client/cli.rs
```diff
@@ -418,6 +424,14 @@ pub fn parse_blob_id(input: &str) -> Result<BlobId, BlobIdParseError> {
     })
 }
 
+/// Parses a quilt blob ID from a string.
+pub fn parse_quilt_patch_id(input: &str) -> Result<QuiltPatchId, QuiltError> {
```
mlegner: Nit: Can you check if your naming is fully consistent everywhere? (For example, here we have "quilt blob ID" vs. "quilt patch ID".)

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -246,6 +249,73 @@ pub enum CliCommands {
...
     },
+    /// Store files as a quilt.
+    #[command(alias("write-quilt"))]
+    StoreQuilt {
+        /// Paths to files to include in the quilt.
+        ///
+        /// If a path is a directory, all the files in the directory will be included
+        /// in the quilt, recursively.
+        /// If a path is a file, the file will be included in the quilt.
+        /// The filenames are used as the identifiers of the quilt patches.
```
mlegner, line 224: Suggestion: Emphasize that the operation will fail if there are duplicate filenames.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -246,6 +249,73 @@ pub enum CliCommands {
...
+        /// Custom identifiers and tags are NOT supported for quilt patches.
+        /// Use `--blob` to specify custom identifiers and tags.
+        #[arg(value_name = "PATH")]
+        #[serde(deserialize_with = "walrus_utils::config::resolve_home_dir_vec")]
+        #[arg(long = "path", action = clap::ArgAction::Append)]
```
mlegner: Suggestion: Use a similar setup as for the files for the `store` command instead of having to add `--path` for each path. The advantage is that you can then use glob patterns, which are impossible when using the `ArgAction::Append`.
```suggestion
        #[arg(long)]
        #[serde(deserialize_with = "walrus_utils::config::resolve_home_dir_vec")]
```

In general, I'm not a big fan of `ArgAction::Append` if it is not absolutely necessary.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -246,6 +249,73 @@ pub enum CliCommands {
...
+        /// Blobs to include in the quilt, each blob is specified as a JSON string.
+        ///
+        /// Example:
+        ///   walrus store-quilt --epochs 10
+        ///     --blob '{"path":"/path/to/food-locations.pdf","identifier":"paper-v2",\
+        ///     "tags":{"author":"Walrus","project":"food","status":"final-review"}}'
+        ///     --blob '{"path":"/path/to/water-locations.pdf","identifier":"water-v3",\
+        ///     "tags":{"author":"Walrus","project":"water","status":"draft"}}'
+        #[arg(value_name = "BLOB")]
+        #[arg(long = "blob", action = clap::ArgAction::Append)]
```
mlegner: Suggestion: Add a `conflicts_with = "paths"` here? I understand that this would prevent using `--path` and `--blob` together in one command, but I don't think this is a big loss: I'd prefer to have one simple command variant (with `--paths`) and one flexible one (with `--blob`).

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -246,6 +249,73 @@ pub enum CliCommands {
...
+        /// in the quilt, recursively.
+        /// If a path is a file, the file will be included in the quilt.
+        /// The filenames are used as the identifiers of the quilt patches.
+        /// Custom identifiers and tags are NOT supported for quilt patches.
+        /// Use `--blob` to specify custom identifiers and tags.
+        #[arg(value_name = "PATH")]
+        #[serde(deserialize_with = "walrus_utils::config::resolve_home_dir_vec")]
+        #[arg(long = "path", action = clap::ArgAction::Append)]
+        paths: Vec<PathBuf>,
+        /// Blobs to include in the quilt, each blob is specified as a JSON string.
```
mlegner, line 231: Suggestion: Emphasize that identifiers must be unique and that the filename will be used if none is specified.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -246,6 +249,73 @@ pub enum CliCommands {
...
+        #[serde(default)]
+        deletable: bool,
+        /// Whether to put the quilt into a shared quilt object.
+        #[arg(long)]
+        #[serde(default)]
+        share: bool,
+        /// The encoding type to use for encoding the files.
+        #[arg(long, hide = true)]
+        #[serde(default)]
+        encoding_type: Option<EncodingType>,
```
mlegner: AFAIU, this is duplicated with the corresponding options for the `store` command.

Suggestion: Define a new `struct CommonStoreOptions` and add them here and to the `store` command as
```rust
#[command(flatten)]
#[serde(flatten)]
common_store_options: CommonStoreOptions,
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -266,6 +336,38 @@ pub enum CliCommands {
         #[serde(flatten)]
         rpc_arg: RpcArg,
     },
+    /// Read quilt patches (blobs) from Walrus.
+    ReadQuilt {
+        /// The query parameters to specify the quilt patches to read.
+        #[command(flatten)]
+        #[serde(flatten)]
+        query: QuiltPatchQuery,
+        /// The directory path where to write the quilt patches.
```
mlegner, line 353: Suggestion: Mention what the resulting filenames will be.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1011,6 +1138,139 @@ impl std::fmt::Display for BlobIdentity {
...
+    ByTag(QuiltPatchByTag),
+    /// Patches by quilt_patch_id.
+    ByPatchId(QuiltPatchByPatchId),
+}
+
+/// Query parameters to read quilt patches.
+#[serde_as]
+#[derive(Debug, Clone, Args, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct QuiltPatchQuery {
```
mlegner: Question/suggestion: Could we use a similar construction as for `NodeSelection` with `#[group(required = true, multiple = false)]`? Alternatively, you could add some `conflicts_with` attributes to ensure that some of the options cannot be specified simultaneously.

Also, it would be good to allow just specifying the quilt ID (without identifiers or tags, in which case all files would be read).

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1011,6 +1138,139 @@ impl std::fmt::Display for BlobIdentity {
...
+    #[serde(default)]
+    pub identifiers: Vec<String>,
+
+    /// The tag key.
+    #[arg(long)]
+    pub tag: Option<String>,
+
+    /// The tag value.
+    #[arg(long)]
+    pub value: Option<String>,
```
mlegner: Question: Could we allow specifying multiple tags? (Can also be in a follow-up.)

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1011,6 +1138,139 @@ impl std::fmt::Display for BlobIdentity {
...
+    #[serde(default)]
+    pub quilt_patch_ids: Vec<QuiltPatchId>,
+}
+
+impl QuiltPatchQuery {
+    fn get_error(&self) -> anyhow::Error {
+        anyhow!(
+            "Exactly one query type must be specified. Valid query types are:\n\
+            - --quilt-id <ID> --identifier <NAME>...\n\
+            - --quilt-id <ID> --tag <KEY> --value <VALUE>\n\
```
mlegner: Question/suggestion: Could we use the same setup as for `SetBlobAttribute::attributes`?

crates/walrus-service/src/client/responses.rs
```diff
@@ -756,3 +766,25 @@ impl NodeHealthOutput {
...
+
+impl ReadQuiltOutput {
+    /// Creates a new [`ReadQuiltOutput`] object.
+    pub fn new(out: Option<PathBuf>, retrieved_blobs: Vec<QuiltStoreBlob<'static>>) -> Self {
+        Self {
+            out,
+            retrieved_blobs,
+        }
+    }
+}
```
mlegner, line 814: Suggestion: Don't include the raw blob data in the struct when `out.is_some()` (similar to the `ReadOutput` above).

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -1135,3 +1245,56 @@ impl CliOutput for GetBlobAttributeOutput {
...
+                    println!("   Tags: (none)");
+                } else {
+                    println!("   Tags:");
+                    for (key, value) in blob.tags() {
+                        println!("     {}: {}", key, value);
+                    }
+                }
+                println!();
+            }
+        }
```
mlegner: Question: What happens to the actual blob data in this case?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -289,6 +310,95 @@ impl CliOutput for DryRunOutput {
...
+            b->"Identifier"
+        ]);
+
+        for (i, quilt_patch) in self.patches().iter().enumerate() {
+            table.add_row(row![
+                bFc->format!("{i}"),
+                format!("{}..{}", quilt_patch.start_index, quilt_patch.end_index),
+                quilt_patch.identifier
+            ]);
+        }
```
mlegner: Question/suggestion: Could we just use (an adapted version of?) the `construct_stored_quilt_patch_table` function here?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -504,6 +558,75 @@ impl ClientCommandRunner {
...
+        rpc_url: Option<String>,
+    ) -> Result<()> {
+        let config = self.config?;
+        let sui_read_client =
+            get_sui_read_client_from_rpc_node_or_wallet(&config, rpc_url, self.wallet).await?;
+        let read_client = Client::new_read_client_with_refresher(config, sui_read_client).await?;
+
+        let quilt_read_client = read_client.quilt_client(QuiltClientConfig::default());
+
+        let retrieved_blobs = match query.get_selector()? {
```
mlegner: Nit: We could check at the very beginning if the selector is valid, to avoid creating read clients in case of failure.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -504,6 +558,75 @@ impl ClientCommandRunner {
...
+            QuiltPatchSelector::ByPatchId(QuiltPatchByPatchId { quilt_patch_ids }) => {
+                quilt_read_client.get_blobs_by_ids(&quilt_patch_ids).await?
+            }
+        };
+
+        let quilt_read_output = ReadQuiltOutput::new(out.clone(), retrieved_blobs);
+        if let Some(out) = out.as_ref() {
+            for blob in &quilt_read_output.retrieved_blobs {
+                let output_file_path = out.join(blob.identifier());
+                std::fs::write(output_file_path, blob.data())?;
```
mlegner: Important: What happens when there are duplicate identifiers?

mlegner (commented):
Thanks a lot for implementing all these quilt features. ❤️ 

A few suggestions for the details of the various CLI arguments.

crates/walrus-service/src/client/cli.rs
liquid-helium: Thanks for the catch!
I missed this one! Also found another two missed ones.
All changed to either quilt_patch_id or quilt_id.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Done.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Change applied!

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Nice tip! The idea was to allow only one of them, but I didn't know the conflicts_with = "paths" annotation. Change applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Done.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Nice suggestion, change applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: We definitely can support that, I added a WAL-899 to track it.
There are some extra complexity since we need to make rules about the relation between different tag matches, for example, color=red `AND` size=large, color=red `OR` size=large... more and more like a database

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Nice catch, comments added.

Created WAL-900 as a TODO.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Good idea! `conflict_with` and `required_unless_present` are used to define the rules, since both identifier and tag require `quilt-id`.

`action = clap::ArgAction::Append,` is used here 🥹, since the `quilt-patch-id` allows hyphen, and if it is not the last arg, it will try to parse the following ones like `--out ...`.

`All` mode is added, and get_all_blobs() is added to related structs.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: As response to another comment, the hyphen in QuiltPatchId would require that we put --quilt-patch-id at the end of the cli command manually.
Kept it as so that all three formats are consistent, unless you have strong preference.

crates/walrus-service/src/client/responses.rs
liquid-helium: The behavior is following ReadOutput.
The raw data is included in the ReadBlobOutput, but is only printed to stdout when `out.is_some()`.

The overall behavior is:
When --out is set, the blobs will be stored to the directory, regardless of `json`. And the metadata of the quilt patches are printed to stdout.
When --out is absent, in json mode, the json string is printed, otherwise, the raw data is printed to stdout.

Added the above comments to the ReadQuilt help output.

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: In short, this is printed only when --out is set, while in non-json mode.
See the above response for more details.

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: Thanks for the catch, I removed this function.

Some context in case you are curious why it is here:
For the QuiltIndex.print_cli_output(), it was used to print_cli_output for `Quilt`, when there was `construct_quilt` command. And when that happens, we don't know the quilt_id yet, (didn't encode the quilt when we only want to construct a quilt, otherwise, it will be the same as dry_run). That's why it didn't reuse construct_stored_quilt_patch_table since it has only the index info, but not the quiltPatchId.

Later I feel the store-quilt --dry-run might be sufficient for users, so I deleted 'construct_quilt'.

crates/walrus-service/src/client/cli/runner.rs
liquid-helium: Good point, although this should not happen if the user use the official binary, a safeguard is added `write_blobs_dedup()`, so that all files will be written out, with a `_i` attached to the filename if there are duplicates.

crates/walrus-service/src/client/cli/runner.rs
liquid-helium: nice catch, change applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: I removed this QuiltPatchQuery, and move all the options into ReadQuilt, so that these args can be displayed as 'Arguments' in the help message, otherwise, the help message is misleading, everything is optional [OPTIONS].

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -632,6 +656,28 @@ impl<T: ReadClient> QuiltClient<'_, T> {
             .await
     }
 
+    /// Retrieves all the blobs from the quilt.
+    pub async fn get_all_blobs(
+        &self,
+        quilt_id: &BlobId,
+    ) -> ClientResult<Vec<QuiltStoreBlob<'static>>> {
+        let (certified_epoch, _) = self
```
halfprice, line 664: Question: should we check the blob status here in case the quite blob is expired/deleted?

crates/walrus-service/src/client/cli.rs
```diff
@@ -418,6 +423,14 @@ pub fn parse_blob_id(input: &str) -> Result<BlobId, BlobIdParseError> {
     })
 }
 
+/// Parses a [`QuiltPatchId`] from a string.
+pub fn parse_quilt_patch_id(input: &str) -> Result<QuiltPatchId, QuiltError> {
+    if let Ok(quilt_id) = QuiltPatchId::from_str(input) {
+        return Ok(quilt_id);
+    }
+    Err(QuiltError::QuiltPatchIdParseError(input.to_string()))
```
halfprice: Nit/suggestion: also save the parsing error in QuiltPatchIdParseError. How about

```
 QuiltPatchId::from_str(input).map_err(|e| QuiltError::QuiltPatchIdParseError(e, input.to_string()))
```

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -289,6 +314,48 @@ impl CliOutput for DryRunOutput {
...
+    }
+}
+
+fn construct_stored_quilt_patch_table(quilt_patches: &[StoredQuiltPatch]) -> Table {
+    let mut table = Table::new();
+    table.set_format(default_table_format());
+    table.set_titles(row![
+        b->"Index",
+        b->"QuiltPatchId",
+        b->"Identifier"
```
halfprice, line 345: Question: what's the difference between a PatchId and a Identifier? Is one of these user defined?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -1134,3 +1201,61 @@ impl CliOutput for GetBlobAttributeOutput {
...
+        ))
+        .printstd();
+    }
+}
+
+impl CliOutput for ReadQuiltOutput {
+    fn print_cli_output(&self) {
+        if let Some(out) = &self.out {
+            println!(
+                "Retrieved {} Blobs and saved to directory: {}",
```
halfprice: question: is "B"lobs on purpose or should be "blobs"?

crates/walrus-service/src/client/cli/cli_output.rs
```diff
@@ -1134,3 +1201,61 @@ impl CliOutput for GetBlobAttributeOutput {
...
+                        println!("     {}: {}", key, value);
+                    }
+                }
+                println!();
+            }
+        } else {
+            for blob in &self.retrieved_blobs {
+                if let Err(e) = std::io::stdout().write_all(blob.data()) {
+                    eprintln!("Error writing {} to stdout: {e}", blob.identifier());
+                }
```
halfprice, line 1258: Nit: do we want to separate these blobs somehow?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -266,6 +266,122 @@ pub enum CliCommands {
...
+            conflicts_with = "identifiers",
+            conflicts_with = "tag",
+            help_heading = "Arguments",
+        )]
+        #[serde(default)]
+        quilt_patch_ids: Vec<QuiltPatchId>,
+
+        /// The directory path where to write the quilt patches.
+        /// The blobs are written to the directory with the same name as the identifier.
+        /// The metadata of the quilt patches, including identifiers and tags are printed to the
```
halfprice: Question: does the metadata mean `tag`? If not the same as the blob metadata storing the merkle tree, maybe use a different word to avoid confusion.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1015,6 +1201,89 @@ impl std::fmt::Display for BlobIdentity {
...
+        } else if tag.is_some() && value.is_some() {
+            Ok(QuiltPatchSelector::ByTag(QuiltPatchByTag {
+                quilt_id: quilt_id.expect("quilt_id should be present"),
+                tag: tag.as_ref().expect("tag should be present").clone(),
+                value: value.as_ref().expect("value should be present").clone(),
+            }))
+        } else if !quilt_patch_ids.is_empty() {
+            Ok(QuiltPatchSelector::ByPatchId(QuiltPatchByPatchId {
+                quilt_patch_ids: quilt_patch_ids.clone(),
+            }))
```
halfprice: Nit: feel all the as_ref() and clone() are not necessary.

Question: are the command line parser configured in a way that the expect() should not be triggered? Just wonder if should instead returning Err() here to not crash the process (probably not a big deal right now)

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -511,6 +572,71 @@ impl ClientCommandRunner {
         ReadOutput::new(out, blob_id, blob).print_output(self.json)
     }
 
+    #[allow(clippy::too_many_arguments)]
```
halfprice: Nit: this clippy suppress is not necessary.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -627,6 +753,157 @@ impl ClientCommandRunner {
...
+        persistence: BlobPersistence,
+        post_store: PostStoreAction,
+        encoding_type: Option<EncodingType>,
+    ) -> Result<()> {
+        epoch_arg.exactly_one_is_some()?;
+        if encoding_type.is_some_and(|encoding| !encoding.is_supported()) {
+            anyhow::bail!(ClientErrorKind::UnsupportedEncodingType(
+                encoding_type.expect("just checked that option is Some")
+            ));
+        }
```
halfprice, line 772: nit: in a follow up PR, I think we should just ignore this input option, or remove this option.

halfprice (approved):
🚀🚀🚀 Very nicely done PR! Clean and easy to read. I only have small nits.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: Yes, the `get_blob_status_and_certified_epoch` handles the `Noexistent` and `Invalid` status.

crates/walrus-service/src/client/cli.rs
liquid-helium: Good idea! Change applied.

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: `Identifier` is defined by the user, `QuiltPatchId ` is computed.

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: Thanks for the catch, fix applied.

crates/walrus-service/src/client/cli/cli_output.rs
liquid-helium: This is a tricky one, TBH, I don't know what is the best way to output multiple raw blobs to stdout.
Created WAL-858 as a followup.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Good point. replaced with `user-defined metadata`, to differentiate from the other metadata

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Thanks for the tip, change applied.
Yes, corresponding clap rules are already defined, all the `expect`s here are triggered only by bugs.

crates/walrus-service/src/client/cli/runner.rs
liquid-helium: Thanks for catching it, I refactored it after getting that warning, and forgot to remove it.

crates/walrus-service/src/client/cli/runner.rs
liquid-helium: It is basically a placeholder now, I will keep it as to be consistent with other functions, and I think we want to leave the room of support for multiple codecs.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -337,14 +267,98 @@ pub enum CliCommands {
...
+        #[serde_as(as = "Vec<DisplayFromStr>")]
+        #[arg(
+            long = "quilt-patch-id",
+            alias = "patch-id",
+            allow_hyphen_values = true,
+            value_parser = parse_quilt_patch_id,
+            action = clap::ArgAction::Append,
+            conflicts_with = "quilt_id",
+            conflicts_with = "identifiers",
+            conflicts_with = "tag",
```
mlegner, line 347: Remark (for future reference): Conflicts only have to be specified in one direction, see also [the docs](https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with).

crates/walrus-service/src/client/cli/args.rs
mlegner: Regarding how the arguments are displayed, we have the same issue with the `store-quilt --paths` option. Also, I believe using a separate struct *should* work when defining it as an argument group with `#[group(required = true)]`.

Regarding the issue with hyphens and `Append`: We could use a [`value_terminator`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_terminator) instead.

crates/walrus-service/src/client/cli/args.rs
mlegner: I was referring to the `--tag`/`--value` construction. In the `SetBlobAttribute` command, we have `--attr "key" "value"` instead of `--tag "key" --value "value"`. I don't have a strong opinion either way, but given the two have an almost identical use, we should use the same approach.

crates/walrus-service/src/client/responses.rs
mlegner: The problem I see is that the blob data is included in the JSON output irrespective of the `out` value. This is different from the `ReadOutput`.

crates/walrus-service/src/client/cli/args.rs
mlegner: Important: The rules are not checked when using the JSON input.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: - Recovered QuiltPatchQuery, with #[group(required = true)] and the conflict rules, works as charm.
- used -- as the terminator.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: refactored to --tag <KEY> <VALUE> style.

crates/walrus-service/src/client/responses.rs
liquid-helium: Good catch, I missed how the ReadOutput is constructed. Fix applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Nice catch, fix applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Fix applied.

PR #2202, opened 2025-06-15
feat: added tags in QuiltIndexV1.

liquid-helium:
## Description

Added Tags to quilt patches, so that a user can look up quilt patches by tags.
Contributes to WAL-858
Contributes to WAL-888

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/encoding/quilt_encoding.rs
+++ b/crates/walrus-core/src/encoding/quilt_encoding.rs
@@ -22,7 +22,7 @@ use alloc::{
     vec::Vec,
 };
 use core::{cmp, fmt};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use hex;
 use serde::{Deserialize, Serialize};
@@ -145,7 +145,9 @@ pub trait QuiltIndexApi<V: QuiltVersion>: Clone + Into<QuiltIndex> {
         self.patches()
             .iter()
             .find(|patch| patch.identifier() == identifier)
-            .ok_or(QuiltError::BlobNotFoundInQuilt(identifier.to_string()))
+            .ok_or(QuiltError::BlobsNotFoundInQuilt(vec![
+                identifier.to_string(),
+            ]))
     }
 
     /// Returns the quilt patches matching the given tag.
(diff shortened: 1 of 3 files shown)


crates/walrus-core/src/metadata.rs
```diff
@@ -69,6 +70,8 @@ pub struct QuiltPatchV1 {
     pub end_index: u16,
     /// The identifier of the blob, it can be used to locate the blob in the quilt.
     pub identifier: String,
+    /// The tags of the blob.
+    pub tags: BTreeMap<String, String>,
```
mlegner, line 75: Question/remark: I guess you use `BTreeMap` to ensure deterministic serialization? Can you mention this in a comment?

crates/walrus-core/src/metadata.rs
```diff
@@ -227,6 +246,32 @@ impl QuiltIndexApi<QuiltVersionV1> for QuiltIndexV1 {
...
+    }
+
+    fn get_sliver_indices_for_tag(
+        &self,
+        target_tag: &str,
+        target_value: &str,
+    ) -> Result<Vec<SliverIndex>, QuiltError> {
+        let patches = self.get_quilt_patches_by_tag(target_tag, target_value)?;
+        let sliver_indices = patches
+            .iter()
```
mlegner: Remark/suggestion: Here you collect an iterator into a vector, just to turn it into an iterator again. I'm not sure if the compiler is able to optimize away the temporary vector; it might be better to have an additional helper function `get_quilt_patches_iter_by_tag`, which is then used in both `get_quilt_patches_by_tag` and `get_sliver_indices_for_tag`. It might also be useful for `get_quilt_patches_by_tag`.

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -326,6 +358,8 @@ impl QuiltEnum {
 /// underscores, hyphens, and periods.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct QuiltStoreBlob<'a> {
```
mlegner, line 409: Suggestion (only partially related to this PR): Can we deduplicate the code with `QuiltStoreBlobOwned` by creating a generic `QuiltStoreBlob<B>`?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -1662,14 +1839,17 @@ mod tests {
                     QuiltStoreBlob {
                         blob: &[1, 3][..],
                         identifier: "test-blob-0".to_string(),
+                        tags: BTreeMap::new(),
                     },
                     QuiltStoreBlob {
                         blob: &[255u8; 1024][..],
                         identifier: "test-blob-1".to_string(),
+                        tags: BTreeMap::new(),
                     },
```
mlegner: Nit: Might be worth creating a helper function `QuiltStoreBlob::new_without_tags(blob, identifier)`.

crates/walrus-core/src/metadata.rs
liquid-helium: Yes, exactly for deterministic serialization. Comments added.

crates/walrus-core/src/metadata.rs
liquid-helium: Good idea thanks! Improvement applied.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Nice suggestion!
I gave it a try, and the refactor is minimal, I will include it in this PR to avoid further diverge.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Now in the refactored `QuiltStoreBlob`, we have new(), new_owned() and a separate .with_tags().

crates/walrus-core/src/metadata.rs
```diff
@@ -227,6 +248,28 @@ impl QuiltIndexApi<QuiltVersionV1> for QuiltIndexV1 {
...
+        &self,
+        target_tag: &str,
+        target_value: &str,
+    ) -> Result<Vec<SliverIndex>, QuiltError> {
+        let sliver_indices = self
+            .get_quilt_patch_iter_by_tag(target_tag, target_value)
+            .flat_map(|patch| (patch.start_index..patch.end_index).map(SliverIndex::new))
+            .collect();
+        Ok(sliver_indices)
+    }
```
mlegner: Question (nit): Could this be a default implementation on the trait itself?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -326,64 +365,39 @@ impl QuiltEnum {
 /// underscores, hyphens, and periods.
 #[derive(Debug, Clone, PartialEq, Eq)]
```
mlegner, line 408: Question (just to be sure): Do the `PartialEq` and `Eq` implementations simply compare the underlying data?

mlegner (approved):
Thanks a lot, @liquid-helium, LGTM! 💯

crates/walrus-core/src/metadata.rs
liquid-helium: Good idea.
I move these two and other to default.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Yes, a simple test is added to make sure of it.

PR #2205, opened 2025-06-16
fix(client): correct status checks when reading

giac-mysten:
## Description

Closes WAL-866

Ensures that `BlobIdDoesNotExist` is only returned when it actually does not exist, and continues the reads if the status check fails.

Removes the `read_blob_with_wait_for_certification` test function to ensure the reads work even if the storage nodes did not see the certification event.

## Test plan

Existing tests. No failures after 100k runs of `cargo stress test_blocklist`.

---

## Release notes

- [x] Aggregator: Fixes an issue where a blob was sometimes reported as non-existing immediately after certification.
- [x] CLI: Fixes an issue where a blob was sometimes reported as non-existing immediately after certification.


mlegner (approved):
Thanks for the fix, @giac-mysten!

I ran `cargo stress test_blocklist` and didn't find any failures in 100k runs.

PR #2206, opened 2025-06-17
feat: Quilt client part iii

liquid-helium:
## Description

Refactor: change several apis from accepting `&[&[T]]` to `impl IntoIterator<Item = &T>`

Added APIs: `QuiltClient::get_blobs_by_tag()`, `QuiltClient::get_blobs_by_ids()`

Contributes to WAL-858

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-sdk/src/error.rs
```diff
@@ -111,7 +111,7 @@ impl ClientError {
 impl From<QuiltError> for ClientError {
     fn from(value: QuiltError) -> Self {
         ClientError {
-            kind: Box::new(ClientErrorKind::QuiltError(value.to_string())),
+            kind: Box::new(ClientErrorKind::QuiltError(value)),
         }
```
mlegner: Nit: I think this could be simplified slightly:
```rust
        ClientErrorKind::from(value).into()
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -243,14 +252,20 @@ pub trait QuiltConfigApi<'a, V: QuiltVersion> {
         blobs: &'a [QuiltStoreBlob<'a>],
     ) -> V::QuiltEncoder<'a>;
 
-    /// Returns a new decoder for the given slivers.
-    fn get_decoder(slivers: &'a [&'a SliverData<V::SliverAxis>]) -> V::QuiltDecoder<'a>;
+    /// Returns a new decoder without slivers.
```
mlegner: Question: What does "without slivers" mean?

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -80,6 +82,207 @@ fn get_all_files_from_path<P: AsRef<Path>>(path: P) -> ClientResult<HashSet<Path
...
+    }
+
+    fn get_decoder(&self) -> V::QuiltDecoder<'_> {
+        match &self.quilt_index {
+            Some(quilt_index) => {
+                V::QuiltConfig::get_decoder_with_quilt_index(&self.slivers, quilt_index)
+            }
+            None => V::QuiltConfig::get_decoder(&self.slivers),
+        }
+    }
```
mlegner, line 136: Question: Is there any way to feed back slivers that are added to the decoder? Or is the idea that all necessary slivers are added to the `DecoderBasedCacheReader` at the beginning?

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -80,6 +82,207 @@ fn get_all_files_from_path<P: AsRef<Path>>(path: P) -> ClientResult<HashSet<Path
...
+
+    pub fn get_blobs_by_identifiers(
+        &self,
+        identifiers: &[&str],
+    ) -> Result<Vec<QuiltStoreBlob<'static>>, QuiltError> {
+        let decoder = self.get_decoder();
+        identifiers
+            .iter()
+            .map(|identifier| decoder.get_blob_by_identifier(identifier))
```
mlegner: Remark: This isn't very efficient, it (indirectly) calls `get_quilt_patch_by_identifier` once for each identifier, which in turn goes through the index to find the identifier.

Maybe it would be better to use `get_sliver_indices_for_identifiers`.

crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -80,6 +82,207 @@ fn get_all_files_from_path<P: AsRef<Path>>(path: P) -> ClientResult<HashSet<Path
...
+                "Reader not initialized".into(),
+            ))),
+            QuiltCacheReader::Decoder(cache) => cache
+                .get_blobs_by_identifiers(identifiers)
+                .map_err(ClientError::other),
+            QuiltCacheReader::FullQuilt(quilt) => identifiers
+                .iter()
+                .map(|identifier| {
+                    quilt
+                        .get_blob_by_identifier(identifier)
```
mlegner: Same remark as above about the efficiency.

mlegner (commented):
Thanks for the PR, @liquid-helium! A few minor comments from my side. 🙏

crates/walrus-sdk/src/error.rs
liquid-helium: Nice tip, change applied.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: The short answer is yes all necessary slivers are downloaded beforehand, this should be sufficient for the current use case.

The slivers are stored in DecoderBasedCacheReader, and can be reused if necessary, but some work is required to make it efficient.

crates/walrus-sdk/src/client/quilt_client.rs
liquid-helium: This is a really good catch!
I updated the get_*_by_identifiers() to plural by default.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Fixed the comments, it was a leftover for some refactor attempts.

mlegner (approved):
Thanks for the changes, @liquid-helium, LGTM. 👍

PR #2207, opened 2025-06-17
feat: Add metric to track number of blocklisted blobs

sadhansood:
## Description

Track number of blocklists per node.

### DIFF
--- a/crates/walrus-sdk/src/blocklist.rs
+++ b/crates/walrus-sdk/src/blocklist.rs
@@ -10,11 +10,13 @@ use std::{
 };
 
 use anyhow::{Context, Result};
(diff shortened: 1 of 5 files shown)


crates/walrus-sdk/src/blocklist.rs
```diff
@@ -29,6 +31,8 @@ pub struct Blocklist {
     blocked_blobs: Arc<RwLock<HashSet<BlobId>>>,
     deny_list_path: PathBuf,
     shutdown: CancellationToken,
+    /// Metric to track the number of blocklisted blobs
```
karlwuest: Nit:
```suggestion
    /// Metric to track the number of blocklisted blobs.
```

PR #2208, opened 2025-06-17
feat(system): adds protocol version event

damirka:
## Description

Closes WAL-668

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


contracts/walrus/tests/e2e_tests.move
```diff
@@ -634,7 +634,80 @@ fun test_register_invalid_pop_signer() {
...
+    let admin = @0xA11CE;
+    let mut nodes = test_node::test_nodes();
+    let mut runner = e2e_runner::prepare(admin)
+        .epoch_zero_duration(EPOCH_ZERO_DURATION)
+        .epoch_duration(EPOCH_DURATION)
+        .n_shards(N_SHARDS)
+        .build();
+
+    // === register candidates ===
+    let epoch = runner.epoch();
```
karlwuest: Nit: Use `setup_committee_for_epoch_one` on the runner instead of doing the setup manually.

crates/walrus-service/src/node.rs
```diff
@@ -1178,6 +1178,9 @@ impl StorageNode {
                 // TODO: Implement DenyListEvent handling (WAL-424)
                 event_handle.mark_as_complete();
             }
+            EventStreamElement::ContractEvent(ContractEvent::ProtocolEvent(_protocol_event)) => {
+                todo!("Handle protocol events");
```
karlwuest: Suggestion: we should not panic with a TODO here. Any node running this version of the software is not expecting a protocol version update, so we should panic explicitly, something like this:

```suggestion
                panic!("unexpected protocol version update: {_protocol_version}");
```

contracts/walrus/sources/system/events.move
```diff
@@ -120,6 +120,14 @@ public struct ContractUpgradeQuorumReached has copy, drop {
     package_digest: vector<u8>,
 }
 
+/// Signals that the protocol version has been updated.
+public struct ProtocolVersionUpdated has copy, drop {
+    epoch: u32,
+    node_id: ID,
```
karlwuest: Suggestion: You can remove the node_id here, I don't think we need this.

karlwuest (approved):
Thanks, lgtm! Just a few small comments.

karlwuest (approved):

PR #2209, opened 2025-06-17
ci: use rust-cache for slow test and simtest jobs

mlegner:
## Description

This uses GitHub's cache for the test and simtest jobs. It is not needed for the lint and build jobs, which are fast anyway.

## Test plan

CI.

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -76,9 +76,6 @@ jobs:
     runs-on: ubuntu-ghcloud
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
-      - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6 # pin@v2.7.8
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/code.yml
```diff
@@ -102,9 +99,6 @@ jobs:
     runs-on: ubuntu-ghcloud
     steps:
       - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
-      - uses: Swatinem/rust-cache@9d47c6ad4b02e050fd481d890b2ea34778fd09d6 # pin@v2.7.8
```
karlwuest, line 105: Question: Does this not affect incremental builds? Are we building from scratch anyway?

karlwuest (approved):

.github/workflows/code.yml
mlegner: This makes the build slower, yes. But the CI bottleneck is anyway the simtest job, so this doesn't matter.

The main reason to remove this here is that we frequently hit the 10GB limit for GitHub caches.

PR #2213, opened 2025-06-18
fix(client): minor improvements to the health command

mlegner:
## Description

* Improve error handling.
* Make number of concurrent requests configurable and increase the default.
* Some minor refactoring.

## Test plan

Run the command.

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -334,6 +334,10 @@ pub enum CliCommands {
         #[command(flatten)]
         #[serde(flatten)]
         sort: SortBy<HealthSortBy>,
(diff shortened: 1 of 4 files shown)


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1394,6 +1398,10 @@ pub(crate) mod default {
             "link".to_string(),
         ]
     }
+
+    pub(crate) fn concurrent_requests_for_health() -> usize {
+        60
```
liquid-helium, line 1403: Just curious, why 60 is chosen?

mlegner: It's a bit arbitrary. I first had 50 but then realized we have a bit more than 100 nodes, so they fit into two chunks with 60. 🤷

PR #2214, opened 2025-06-18
fix: use a guard to ensure active request gauge decrements

jpcsmith:
## Description

Switch to using a guard around the server active requests to ensure that the value always decrements.

## Test plan

Existing tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


halfprice (approved):
LGTM!

PR #2216, opened 2025-06-19
fix: race condition between node entering recovery mode and start blob sync

halfprice:
## Description

This PR fix a race condition between a node entering recovery mode due to lagging too far behind, and start of blob sync.

The specific sequence of the race condition is follows:

- Blob sync starts, checking the node is in Active state. Blob sync starts async task `A` to sync blob.
- Node enter recovery mode.
- task `A` starts. Because the node is lagging too far behind (the committee read by the node just now is too far ahead), the node does not know the shard assignment for the event epoch, therefore a panic is triggered.

The fix contains two parts:
- When a node enters recovery mode, it needs to cancel all ongoing blob syncs since once the node catches up to the latest epoch, it'll scan all the certified blobs and recovery any missing blobs
- However, the above is not enough because blob sync's read-start is not atomic, so at the beginning of task `A`, the blob sync task also needs to check if the node is in RecoveryCatchupMode and should skip the blob sync.

## Test plan

- With the new async parallel certified event processing, test_node_slow_process_events_entering_recovery has very high probability to manifest this issue. This PR passes 100% of this test with 100 test run.
- A specific unit test is added to test blob sync skip logic.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/blob_sync.rs
+++ b/crates/walrus-service/src/node/blob_sync.rs
@@ -43,7 +43,7 @@ use super::{
     storage::Storage,
     system_events::{CompletableHandle, EventHandle},
 };
-use crate::common::utils::FutureHelpers as _;
+use crate::{common::utils::FutureHelpers as _, node::NodeStatus};
 
 #[derive(Debug, Clone)]
 struct Permits {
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -228,6 +228,43 @@ impl BlobSyncHandler {
...
+        let count = join_handles.len();
+
+        try_join_all(join_handles)
+            .await?
+            .into_iter()
+            .for_each(CompletableHandle::mark_as_complete);
+
+        tracing::info!("cancelled {count} blob syncs");
+        Ok(count)
+    }
```
mlegner: Question: We have a separate function `cancel_all` and a large code overlap with `cancel_all_expired_syncs_and_mark_events_completed`. Could we consolidate these three functions somehow (maybe in a follow-up)?

mlegner (approved):
Thanks so much for the fix, @halfprice. This LGTM! 💯 

I only have one comment, which can be addressed in a follow-up. I suggest to get this merged as is to unblock other PRs that are currently blocked because of failing simtests. Any objections, @jpcsmith?

crates/walrus-service/src/node/blob_sync.rs
mlegner: I actually have a follow-up PR ready here: #2217.

PR #2217, opened 2025-06-19
refactor: deduplicate code related to blob sync cancellation

mlegner:
## Description

Deduplicates some code related to blob-sync cancellation.

## Test plan

CI.

### DIFF
--- a/crates/walrus-service/src/node/blob_sync.rs
+++ b/crates/walrus-service/src/node/blob_sync.rs
@@ -177,6 +177,55 @@ impl BlobSyncHandler {
         Ok(())
     }
 
+    /// Cancels all blob syncs using the provided closure and returns the number of cancelled syncs.
+    ///
+    /// If `should_mark_events_complete` is `true`, the corresponding events are marked as complete.
+    #[tracing::instrument(skip(self, filter_map_closure))]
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -177,19 +177,19 @@ impl BlobSyncHandler {
...
-    /// before any later events are being handled.
-    #[tracing::instrument(skip(self))]
-    pub async fn cancel_all_expired_syncs_and_mark_events_completed(
+    /// If `should_mark_events_complete` is `true`, the corresponding events are marked as complete.
+    #[tracing::instrument(skip(self, filter_map_closure))]
+    async fn cancel_syncs<F>(
         &self,
-    ) -> anyhow::Result<usize> {
-        tracing::info!("cancelling all blob syncs for expired blobs");
+        filter_map_closure: F,
```
jpcsmith, line 186: Remark: Is the `map` portion of the closure necessary? Could this just be replaced with a `fn(..) -> bool` closure?

mlegner: I think we need it to return event handles such that we can mark them as complete at the end.

PR #2220, opened 2025-06-19
refactor(sdk): replace `StoreWhen` enum by `StoreOptimizations` struct

mlegner:
## Description

The current `StoreWhen` enum combines two orthogonal optimizations (checking the current blob status and checking existing resources). As a result, it needs four variants, and the exact implications of each of them are a bit unclear.

This replaces this enum by a new `StoreOptimizations` struct, which contains boolean fields for all (currently 2) optimizations, which can be enabled independently.

No actual behavior is changed in the final binaries. There are some minor changes in tests, which shouldn't affect the outcome, however.

## Test plan

Existing tests.

### DIFF
--- a/crates/walrus-sdk/src/store_optimizations.rs
+++ b/crates/walrus-sdk/src/store_optimizations.rs
@@ -6,9 +6,15 @@
 /// Represents how the store operation should be carried out by the client.
 #[derive(Debug, Clone, Copy)]
 pub struct StoreOptimizations {
-    /// Check the status of the blob before storing it, and store it only if it is not already.
+    /// Check the status of the blob before storing it; store it only if it is not already stored as
+    /// a permanent blob on Walrus for a sufficient number of epochs.
     pub check_status: bool,
-    /// Reuse matching resources owned by the wallet if there are any.
+    /// Reuse matching `Storage` resources and already registered `Blob` objects owned by the wallet
+    /// if there are any.
+    ///
+    /// Specifically, check if there is an appropriate storage resource (with sufficient space and
+    /// for a sufficient duration) that can be used to register the blob or if there is an already
+    /// registered blob object that matches the blob ID and duration.
     pub reuse_resources: bool,
 }
 


crates/walrus-sdk/src/store_optimizations.rs
```diff
@@ -0,0 +1,64 @@
...
+
+//! Optimizations for storing blobs.
+
+/// Represents how the store operation should be carried out by the client.
+#[derive(Debug, Clone, Copy)]
+pub struct StoreOptimizations {
+    /// Check the status of the blob before storing it, and store it only if it is not already.
+    pub check_status: bool,
+    /// Reuse matching resources owned by the wallet if there are any.
+    pub reuse_resources: bool,
```
liquid-helium, line 18: nit: add more details about what resources are reused? Like 'registration (Sui) and storage (Walrus)', it might not be straightforward for non-developers.

I saw comments in the function, maybe copy that to the struct as well:
Tries to reuse existing blob registrations or storage resources if possible. Specifically:

crates/walrus-sdk/src/store_optimizations.rs
```diff
@@ -0,0 +1,64 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+//! Optimizations for storing blobs.
+
+/// Represents how the store operation should be carried out by the client.
+#[derive(Debug, Clone, Copy)]
+pub struct StoreOptimizations {
+    /// Check the status of the blob before storing it, and store it only if it is not already.
```
liquid-helium: nit: more details on how existing status (registered, certified) affect the behavior?

crates/walrus-sdk/src/store_optimizations.rs
mlegner: Done in 5ae67220ea322489ab7c98624dc635f4961d53f4.

crates/walrus-sdk/src/store_optimizations.rs
mlegner: Done in 5ae67220ea322489ab7c98624dc635f4961d53f4.

PR #2222, opened 2025-06-20
feat: Quilt aggregator

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [x] Aggregator:
  - Added `/v1/blobs/by-quilt-patch-id/{quilt_patch_id}` endpoint to read a blob from a quilt by its `QuiltPatchId`
  - Added `/v1/blobs/by-quilt-id/{quilt_id}/{identifier}` endpoint to read a blob from a quilt by its `quilt_id and identifier`
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -854,6 +854,12 @@ pub struct AggregatorArgs {
     #[arg(long, num_args = 1.., default_values_t = default::allowed_headers())]
     #[serde(default = "default::allowed_headers")]
     pub(crate) allowed_headers: Vec<String>,
+    /// Whether to allow quilt patch tags to be returned in the response headers.
+    /// If true, the tags will be returned in the response headers, regardless of the allowed
+    /// headers.
+    #[arg(long, default_value_t = false)]
+    #[serde(default)]
+    pub allow_quilt_patch_tags_in_response: bool,
```
halfprice, line 862: Question: why put this as a server side config?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -123,13 +113,48 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
     attribute: &BlobAttribute,
-    allowed_headers: &HashSet<String>,
+    allowed_headers: Option<&HashSet<String>>,
 ) {
     for (key, value) in attribute.iter() {
-        if allowed_headers.contains(key) {
+        tracing::info!(
+            attribute_key = key,
+            attribute_value = value,
+            "quilt blob attribute"
```
halfprice: Nit: maybe just "blob attribute", since this is used in both cases? Also should this log be in DEBUG level?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -373,6 +410,236 @@ pub(super) fn daemon_cors_layer() -> CorsLayer {
...
+        Ok(mut blobs) => {
+            if let Some(blob) = blobs.pop() {
+                build_quilt_patch_response(
+                    blob,
+                    &request_headers,
+                    &quilt_patch_id_str,
+                    &response_header_config,
+                )
+            } else {
+                tracing::info!("no blob returned for the requested quilt blob ID");
```
halfprice: Nit: maybe also log the patch id? Should this be in DEBUG level as well? Same below.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -854,6 +854,12 @@ pub struct AggregatorArgs {
     #[arg(long, num_args = 1.., default_values_t = default::allowed_headers())]
     #[serde(default = "default::allowed_headers")]
     pub(crate) allowed_headers: Vec<String>,
+    /// Whether to allow quilt patch tags to be returned in the response headers.
+    /// If true, the tags will be returned in the response headers, regardless of the allowed
```
halfprice, line 858: After reading the code, I think the `allowed_headers` are ignored if this config is true.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -373,6 +410,236 @@ pub(super) fn daemon_cors_layer() -> CorsLayer {
...
+) -> Response {
+    let identifier = blob.identifier().to_string();
+    let blob_attribute: BlobAttribute = blob.tags().clone().into();
+    let blob_data = blob.into_data();
+    let mut response = (StatusCode::OK, blob_data).into_response();
+    populate_response_headers_from_request(request_headers, etag, response.headers_mut());
+    populate_response_headers_from_attributes(
+        response.headers_mut(),
+        &blob_attribute,
+        if response_header_config.allow_quilt_patch_tags_in_response {
```
halfprice, line 521: Question: where do we add tags if `allow_quilt_patch_tags_in_response` is true?

halfprice (approved):
Also looks great! 💯

crates/walrus-service/src/client/cli/args.rs
liquid-helium: This is a related config as the above allowed_headers, it controls whether we allow arbitrary tags being used as response headers.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Thanks for the catch, I used this for debug, removed.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Change applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: The allowed_headers will be ignored for the `tags`, if this is true, basically we allow all tags to be used as headers.
However, allowed headers will be respected for `attributes`

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: when `allow_quilt_patch_tags_in_response ` is true, the allowed_headers is None, and we add all tags blindly (line 152) to the response headers.

PR #2223, opened 2025-06-20
feat: DB tuning

halfprice:
## Description

- Turn on bloom filter
- Increase block size
- Use large block cache in sliver CFs.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/typed-store/src/rocks.rs
+++ b/crates/typed-store/src/rocks.rs
@@ -1777,7 +1777,7 @@ pub fn default_db_options() -> DBOptions {
     // Increase block size to 16KiB.
     // https://github.com/EighteenZi/rocksdb_wiki/blob/master/
     // Memory-usage-in-RocksDB.md#indexes-and-filter-blocks
-    opt.set_block_based_table_factory(&get_block_options(128, 16 << 10));
+    opt.set_block_based_table_factory(&get_block_options(128 << 20, Some(16 << 10), Some(true)));
 
     // Set memtable bloomfilter.
     opt.set_memtable_prefix_bloom_ratio(0.02);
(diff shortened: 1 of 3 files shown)


sadhansood:
> Turn on bloom filter
Are we really turning them on in this PR (I didn't see it anywhere) but I also think filters are part of the design in Rocksdb, we can only tune their configuration (and we didn't explicitly disable them before).

crates/typed-store/src/rocks.rs
```diff
@@ -1788,22 +1788,32 @@ pub fn default_db_options() -> DBOptions {
...
-    // From https://github.com/EighteenZi/rocksdb_wiki/blob/master/
-    // Block-Cache.md#caching-index-and-filter-blocks
-    block_options.set_pin_l0_filter_and_index_blocks_in_cache(true);
+    if let Some(pin_l0_filter_and_index_blocks_in_block_cache) =
+        pin_l0_filter_and_index_blocks_in_block_cache
+    {
+        // From https://github.com/EighteenZi/rocksdb_wiki/blob/master/
+        // Block-Cache.md#caching-index-and-filter-blocks
+        block_options.set_pin_l0_filter_and_index_blocks_in_cache(
+            pin_l0_filter_and_index_blocks_in_block_cache,
```
sadhansood, line 1815: I think another thing to add here would be `set_cache_index_and_filter_blocks(true)` to cache all index and filter blocks (not just L0) and increase the block cache size by some amount to account for it.

sadhansood (approved):

halfprice:
> > Turn on bloom filter
> 
> Are we really turning them on in this PR (I didn't see it anywhere) but I also think filters are part of the design in Rocksdb, we can only tune their configuration (and we didn't explicitly disable them before).

I couldn't find reference anywhere on wiki, but looking at the source code, the filter policy is null by default: https://github.com/facebook/rocksdb/blob/fdc2970d37e7043836966d95bddfd415d8c98a23/include/rocksdb/table.h#L495

Let's deploy this ini PTN and see if it makes a difference.

crates/typed-store/src/rocks.rs
halfprice: Done. I added this option.

I didn't increase block cache size yet since not sure how much should I increase. The minimum cache size is 512 MB now so let's see if we need to further increase it.

PR #2237, opened 2025-06-26
fix: handling accessing shard db after shard removal

halfprice:
## Description

Fount in antithesis test. ShardStorage currently is held as Arc<ShardStorage>, meaning that after epoch change that removes a shard and deletes all the column families in the shard, there may still be background process that may access the shard db. This can cause node crash.

To fix the issue, we make DBMap.cf() call return optional column family handle so that if a shard is removed, an error will be returned instead of crashing the node. DBMap.cf() is called in all DB operations.

## Test plan

Enhanced simtest to manifest this crash, and verified the fix.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/typed-store/src/rocks.rs
+++ b/crates/typed-store/src/rocks.rs
@@ -586,22 +586,7 @@ impl<K, V> DBMap<K, V> {
         let from_buf = be_fix_int_ser(start)?;
         let to_buf = be_fix_int_ser(end)?;
         self.rocksdb
-            .compact_range_cf(&self.cf(), Some(from_buf), Some(to_buf));
-        Ok(())
-    }
-
-    /// Compact a range of keys in a specific column family
-    pub fn compact_range_raw(
-        &self,
-        cf_name: &str,
-        start: Vec<u8>,
-        end: Vec<u8>,
-    ) -> Result<(), TypedStoreError> {
-        let cf = self
-            .rocksdb
-            .cf_handle(cf_name)
-            .expect("compact range: column family does not exist");
-        self.rocksdb.compact_range_cf(&cf, Some(start), Some(end));
+            .compact_range_cf(&self.cf()?, Some(from_buf), Some(to_buf));
         Ok(())
     }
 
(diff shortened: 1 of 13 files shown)


crates/typed-store/src/rocks.rs
```diff
@@ -586,22 +586,7 @@ impl<K, V> DBMap<K, V> {
         let from_buf = be_fix_int_ser(start)?;
         let to_buf = be_fix_int_ser(end)?;
         self.rocksdb
-            .compact_range_cf(&self.cf(), Some(from_buf), Some(to_buf));
-        Ok(())
-    }
-
-    /// Compact a range of keys in a specific column family
-    pub fn compact_range_raw(
```
halfprice, line 594: Removing not used function.

crates/typed-store/src/rocks.rs
```diff
@@ -614,37 +599,24 @@ impl<K, V> DBMap<K, V> {
...
-            eprintln!("Full backtrace:\n{}", backtrace);
-            panic!(
-                "Map-keying column family {} should have been checked at DB creation",
-                &self.cf
-            )
-        })
+    pub fn cf(&self) -> Result<Arc<rocksdb::BoundColumnFamily<'_>>, TypedStoreError> {
+        self.rocksdb
+            .cf_handle(&self.cf)
+            .ok_or_else(|| TypedStoreError::UnregisteredColumn(self.cf.clone()))
```
halfprice, line 610: Key fix is here to return Result<> instead of crash

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -743,6 +648,147 @@ impl BlobSynchronizer {
...
+                                // It is possible that when the blob recovery is running, a shard
+                                // is removed from the storage, and therefore encounter
+                                // UnregisteredColumn error. In this case, we should not panic
+                                // and instead return an error to retry the blob sync.
+                                tracing::error!(
+                                    "database error during sliver sync: unregistered column {:?}",
+                                    column
+                                );
+                                return Err(RecoverSliverError::Database(err));
+                            }
```
halfprice: Here we need to special case CF not found error so that this error can be retried.

crates/walrus-simtest/tests/simtest_failure.rs
```diff
@@ -416,36 +417,26 @@ mod tests {
...
+            );
+            sui_macros::register_fail_point_async(
+                "fail_point_recover_sliver_before_put_sliver",
+                move || async move {
+                    if sui_simulator::current_simnode_id() == target_fail_node_id {
+                        tokio::time::sleep(Duration::from_secs(30)).await;
+                    }
+                },
+            );
+        }
```
halfprice, line 435: Change in this test can manifest the `access after remove` issue.

crates/walrus-service/src/node/blob_sync.rs
halfprice: For shard sync, we don't need to handle `cf does not exist` case, since the shard sync logic is inside ShardStorage, and ShardStorage object will be removed soon (shard sync will continue encounter error and retry, but I think that's ok).

sadhansood (approved):
Amazing work @halfprice and great way to repro with simtest. Thank you!

PR #2238, opened 2025-06-26
fix: move the creation of the recovery symbol to a blocking thread

jpcsmith:
## Description

This PR moves symbol creation to a blocking thread where it joins the proof creation. Tt was causing a lot of blocking on the tokio core thread. This was always the case, but the current amount of recovery on mainnet highlighted it as an issue.

## Test plan

Load tested using k6. The average slow poll duration for GET recoverySymbols is now lower, and it is now able to accept a large amount of requests and saturate the CPU before saturating the tokio core threads.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/recovery_symbol_service.rs
+++ b/crates/walrus-service/src/node/recovery_symbol_service.rs
@@ -7,8 +7,8 @@ use std::{
 };
 
 use fastcrypto::hash::Blake2b256;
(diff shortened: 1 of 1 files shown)


halfprice (approved):
LGTM!

PR #2239, opened 2025-06-26
feat: export the tls certificate expiration time as a metric

jpcsmith:
## Description

Report TLS expiration time in seconds since unix epoch.

Contributes to WAL-886.

## Test plan

Ran a local cluster and checked the expiration time exported.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/common/telemetry.rs
+++ b/crates/walrus-service/src/common/telemetry.rs
@@ -30,6 +30,7 @@ use axum::{
 };
 use opentelemetry::propagation::Extractor;
 use prometheus::{
+    Gauge,
     Histogram,
     HistogramVec,
     IntGaugeVec,
@@ -111,7 +112,10 @@ walrus_utils::metrics::define_metric_set! {
                 "http_response_status_code",
             ],
             buckets: walrus_utils::metrics::default_buckets_for_bytes()
-        }
+        },
+
+        #[help = "The seconds since UNIX epoch after which the TLS certificate is not valid."]
+        tls_not_after_seconds: Gauge[],
     }
 }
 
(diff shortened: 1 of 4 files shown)


crates/walrus-service/src/node/server.rs
```diff
@@ -439,6 +428,60 @@ where
...
+
+async fn configure_tls_from_pem(
+    certificate: &PathOrInPlace<Vec<u8>>,
+    key: &PathOrInPlace<Vec<u8>>,
+) -> Result<(RustlsConfig, Certificate), anyhow::Error> {
+    let (tls_config, certificate_pem) =
+        if let Some((certificate_path, key_path)) = certificate.path().zip(key.path()) {
+            let tls_config = RustlsConfig::from_pem_file(certificate_path, key_path)
+                .await
+                .context("failed to load certificate and key from provided paths")?;
```
mlegner: Question: Is there any benefit to use `from_pem_file` over `from_pem`? If not, could we skip the case distinction and just use the `else` branch in both cases?

crates/walrus-service/src/common/telemetry.rs
```diff
@@ -111,7 +112,10 @@ walrus_utils::metrics::define_metric_set! {
                 "http_response_status_code",
             ],
             buckets: walrus_utils::metrics::default_buckets_for_bytes()
-        }
+        },
+
+        #[help = "The seconds since UNIX epoch after which the TLS certificate is not valid."]
+        tls_not_after_seconds: Gauge[],
```
mlegner: Nit: Maybe explicitly add `certificate` here?
```suggestion
        tls_certificate_not_after_seconds: Gauge[],
```

Also, question: Do you know if there is a good way of using this to set up an alert? In a panel, one can use `From Now` as the unit, but I'm not sure how this can be done for an alert.

crates/walrus-service/src/node/server.rs
jpcsmith: Good point. I looked at the code, it just reads from the filesystem before calling `from_pem`, so I simplified the code there.

crates/walrus-service/src/common/telemetry.rs
jpcsmith: Added. 

There is also the [time](https://prometheus.io/docs/prometheus/latest/querying/functions/#time) function which gives the instant time as a unix timestamp. I will follow up once this lands on devnet.

PR #2244, opened 2025-06-27
feat: ensure that objects can only be migrated after epoch change

karlwuest:
## Description

- Make sure that staking and system objects can only be migrated after the next epoch starts after an upgrade
- Only allow migrating the subsidies object after the system object has been migrated

## Test plan

- Modified existing test

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


contracts/walrus/sources/staking.move
```diff
@@ -368,14 +371,31 @@ public(package) fun set_new_package_id(staking: &mut Staking, new_package_id: ID
...
 ///
 /// This function sets the new package id and version and can be modified in future versions
 /// to migrate changes in the `staking_inner` object if needed.
 public(package) fun migrate(staking: &mut Staking) {
     assert!(staking.version < VERSION, EInvalidMigration);
 
-    // Move the old system state inner to the new version.
+    // Check that the migration epoch is set and that the current epoch is greater than or equal to
+    // the migration epoch.
+    let mut migration_epoch = df::remove_if_exists(&mut staking.id, MIGRATION_EPOCH_KEY);
```
damirka: Nit: you can use `destroy_or!(abort EInvalidMigration)` here. The expression inside macro call won't be evaluated for `some` case.

contracts/walrus/sources/staking.move
```diff
@@ -368,14 +371,31 @@ public(package) fun set_new_package_id(staking: &mut Staking, new_package_id: ID
...
+    assert!(migration_epoch.is_some(), EInvalidMigration);
+    let migration_epoch = migration_epoch.extract();
+    assert!(staking.inner_without_version_check().epoch() >= migration_epoch, EInvalidMigration);
```
damirka: Should we make it a different error code? Something like `EMigratingTooSoon`?

contracts/walrus/sources/staking.move
```diff
@@ -368,14 +371,31 @@ public(package) fun set_new_package_id(staking: &mut Staking, new_package_id: ID
     staking.new_package_id = option::some(new_package_id);
 }
 
+/// Sets the epoch in which the staking and system objects can be migrated after an upgrade.
+entry fun set_migration_epoch(staking: &mut Staking) {
```
damirka, line 377: Important to note that due to visibility, you won't be able to call another command in a PTB that references Staking. Entry restricts "dirtied" values, disallowing multiple calls using the same object.

karlwuest: I was not aware of that, but for this case this should not be a problem. I don't see a necessity to use this in a PTB with other transactions. And I slightly regret setting the `migrate` function to public, so I'm now trying to keep the public interfaces for all migration related functions as tight as possible.

contracts/walrus/sources/staking.move
karlwuest: Added `EInvalidMigrationEpoch` and also emit that when the epoch is not set.

contracts/walrus/sources/staking.move
```diff
@@ -368,14 +373,34 @@ public(package) fun set_new_package_id(staking: &mut Staking, new_package_id: ID
...
+    assert!(staking.version < VERSION, EInvalidMigration);
+    if (df::exists_(&staking.id, MIGRATION_EPOCH_KEY)) {
+        return
```
damirka, line 380: Last nit: do you think it should abort here instead?

damirka (approved):
lgtm!

contracts/walrus/sources/staking.move
karlwuest: I didn't want to error here, because then an error implies that we cannot perform the migration in the next epoch. If the migration epoch is already set, we don't really care and can proceed.

PR #2245, opened 2025-06-27
feat: Add proto definitions for event streaming service

sadhansood:
## Description

This PR introduces a new event-streaming crate that provides gRPC-based event streaming functionality for Walrus with optional cryptographic proofs (of various types). It  includes automatic dependency management that downloads required protobuf definitions from GitHub when not available locally. 
Added two rpc endpoints to this service:
1. StreamEvents RPC for server-side streaming of verifiable events
2. GetNonInclusionProof RPC for cryptographic non-inclusion proofs

We chose gRPC for this service to stay consistent with Fullnode rpc as we want Walrus node (clients) to have seamless experience when switching between events from local sidecar or remote Fullnode.
## Test plan

Added basic unit tests covering all major message types and proof mechanisms

### DIFF
--- a/crates/event-streaming/build.rs
+++ b/crates/event-streaming/build.rs
@@ -1,0 +1,142 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{fs, path::PathBuf, process::Command};
+
+const SUI_REPO_URL: &str = "https://github.com/MystenLabs/sui.git";
+const SUI_TAG: &str = "testnet-v1.50.1";
+
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let sui_proto_source = ensure_sui_proto_source()?;
+
+    // Get the manifest directory (crate root) to resolve proto files relative to it
+    let manifest_dir = std::env::var("CARGO_MANIFEST_DIR")?;
+    let crate_root = PathBuf::from(manifest_dir);
+
+    let proto_file_names = vec!["proto/walrus/event/v1alpha/service.proto"];
+
+    // Convert to full paths and check that our proto files exist
+    let mut proto_files = Vec::new();
+    for proto_file in &proto_file_names {
+        let full_proto_path = crate_root.join(proto_file);
+        if !full_proto_path.exists() {
+            return Err(format!(
+                "Proto file not found: {} (resolved to: {})",
+                proto_file,
+                full_proto_path.display()
+            )
+            .into());
+        }
+        proto_files.push(full_proto_path.to_string_lossy().to_string());
+    }
+
+    // Use protox to compile proto files without requiring system protoc
+    let include_paths = vec![
+        crate_root.join("proto").to_string_lossy().to_string(),
+        sui_proto_source.to_string_lossy().to_string(),
+    ];
+
+    let file_descriptor_set = protox::compile(&proto_files, &include_paths)?;
+
+    // Generate Rust code using prost-build
+    let mut prost_config = prost_build::Config::new();
+    prost_config.include_file("mod.rs");
+
+    // Compile the file descriptor set
+    prost_config.compile_fds(file_descriptor_set)?;
+
+    // Tell cargo to rerun if proto files change
+    println!(
+        "cargo:rerun-if-changed={}",
+        crate_root.join("proto").display()
+    );
+
+    Ok(())
+}
+
+fn ensure_sui_proto_source() -> Result<PathBuf, Box<dyn std::error::Error>> {
+    // Always download Sui proto files from GitHub to ensure we have the exact version we need
+    // This eliminates potential version mismatches and simplifies the build process
+    download_sui_repo()
+}
+
+fn download_sui_repo() -> Result<PathBuf, Box<dyn std::error::Error>> {
+    let out_dir = std::env::var("OUT_DIR")?;
+    let sui_clone_dir = PathBuf::from(&out_dir).join(format!("sui_clone_{}", SUI_TAG));
+    let sui_proto_path = sui_clone_dir.join("crates/sui-rpc-api/proto");
+
+    cleanup_old_sui_caches(&out_dir)?;
+
+    // Check if already cloned and up to date
+    if sui_proto_path.exists() {
+        println!(
+            "cargo:warning=Using cached Sui proto files from: {} (tag: {})",
+            sui_proto_path.display(),
+            SUI_TAG
+        );
+        return Ok(sui_proto_path);
+    }
+
+    println!(
+        "cargo:warning=Downloading Sui repository from GitHub (tag: {})...",
+        SUI_TAG
+    );
+
+    // Remove existing clone if it exists but is incomplete
+    if sui_clone_dir.exists() {
+        fs::remove_dir_all(&sui_clone_dir)?;
+    }
+
+    // Clone the specific tag with shallow depth for faster download
+    let output = Command::new("git")
+        .args([
+            "clone",
+            "--depth=1",
+            "--branch",
+            SUI_TAG,
+            SUI_REPO_URL,
+            &sui_clone_dir.to_string_lossy(),
+        ])
+        .output()?;
+
+    if !output.status.success() {
+        return Err(format!(
+            "Failed to clone Sui repository: {}",
+            String::from_utf8_lossy(&output.stderr)
+        )
+        .into());
+    }
+
+    // Verify the proto files exist
+    if !sui_proto_path.exists() {
+        return Err("Cloned Sui repository doesn't contain expected proto files".into());
+    }
+
+    println!(
+        "cargo:warning=Successfully downloaded Sui proto files to: {} (tag: {})",
+        sui_proto_path.display(),
+        SUI_TAG
+    );
+    Ok(sui_proto_path)
+}
+
+fn cleanup_old_sui_caches(out_dir: &str) -> Result<(), Box<dyn std::error::Error>> {
+    let out_path = PathBuf::from(out_dir);
+    let current_cache_name = format!("sui_clone_{}", SUI_TAG);
+
+    // Read the directory and remove old sui_clone_* directories
+    if let Ok(entries) = fs::read_dir(&out_path) {
+        for entry in entries.flatten() {
+            if let Some(name) = entry.file_name().to_str() {
+                if name.starts_with("sui_clone_") && name != current_cache_name {
+                    if let Err(e) = fs::remove_dir_all(entry.path()) {
+                        println!("cargo:warning=Failed to clean up old cache {}: {}", name, e);
+                    } else {
+                        println!("cargo:warning=Cleaned up old Sui cache: {}", name);
+                    }
+                }
+            }
+        }
+    }
+    Ok(())
+}
--- a/crates/event-streaming/proto/walrus/event/v1alpha/service.proto
+++ b/crates/event-streaming/proto/walrus/event/v1alpha/service.proto
@@ -1,0 +1,164 @@
+syntax = "proto3";
+
+package walrus.event.v1alpha;
+
+import "google/protobuf/empty.proto";
+import "google/protobuf/timestamp.proto";
+import "sui/rpc/v2beta/checkpoint_contents.proto";
+import "sui/rpc/v2beta/checkpoint_summary.proto";
+import "sui/rpc/v2beta/event.proto";
+
+// EventStreamingService provides a verifiable stream of events from the Sui
+// blockchain for an event stream. It allows clients to receive events
+// along with cryptographic proofs that tie them to a specific checkpoint.
+// It is designed to be forward-compatible, allowing for proof mechanisms
+// to be upgraded over time without breaking clients.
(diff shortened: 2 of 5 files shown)


halfprice (commented):
Thanks @sadhansood! One general comment: can we make event streaming a feature in the codebase so that it won't get compiled when e.g. building a walrus node?

crates/event-streaming/build.rs
```diff
@@ -0,0 +1,199 @@
...
+            println!(
+                "cargo:warning=Using Sui proto files from relative path: {}",
+                sui_proto_path.display()
+            );
+            return Ok(sui_proto_path);
+        }
+    }
+
+    // If not found locally, clone from GitHub
+    download_sui_repo()
```
wbbradley, line 60: Might be nice to not automatically do this step, just to avoid confusion. Devs should manually opt in to this, I think.

```suggestion
    if ...some env var exists specifying that fallback to download Sui repo from internet is acceptable... {
        download_sui_repo()
    } else {
        fail build with explanation
    }
```

crates/event-streaming/build.rs
```diff
@@ -0,0 +1,199 @@
...
+fn ensure_sui_proto_source() -> Result<PathBuf, Box<dyn std::error::Error>> {
+    // Priority order for finding Sui proto files:
+    // 1. Environment variable SUI_REPO_PATH
+    // 2. Relative paths (assuming walrus and sui are siblings)
+    // 3. Download from GitHub to build cache
+
+    if let Ok(sui_path) = std::env::var("SUI_REPO_PATH") {
+        let sui_proto_path = PathBuf::from(sui_path).join("crates/sui-rpc-api/proto");
+        if sui_proto_path.exists() {
+            println!(
```
wbbradley: What guarantees that the sui repo in this case is checked out to the correct SHA?

crates/event-streaming/src/lib.rs
```diff
@@ -0,0 +1,228 @@
...
+                assert_eq!(merkle_proof.leaf_artifact, vec![0xde, 0xad, 0xbe, 0xef]);
+                assert_eq!(merkle_proof.sibling_hashes.len(), 2);
+            } else {
+                panic!("Expected MerkleProof proof type");
+            }
+        }
+    }
+
+    #[test]
+    fn test_merkle_inclusion_proof() {
```
wbbradley: This test is not testing anything that the compiler wouldn't already catch in regular usage, I think.

crates/event-streaming/proto/walrus/event/v1/service.proto
```diff
@@ -0,0 +1,175 @@
+syntax = "proto3";
+
+package walrus.event.v1;
```
bmwill: v1alpha

crates/event-streaming/proto/walrus/event/v1/service.proto
```diff
@@ -0,0 +1,175 @@
...
+// blockchain for an event stream. It allows clients to receive events
+// along with cryptographic proofs that tie them to a specific checkpoint.
+// It is designed to be forward-compatible, allowing for proof mechanisms
+// to be upgraded over time without breaking clients.
+service EventStreamingService {
+    // StreamEvents opens a server-side stream that sends sets of events
+    // as they are finalized in checkpoints. Each message in the stream contains
+    // all relevant events for a single checkpoint, plus the necessary data to verify them.
+    rpc StreamEvents(StreamEventsRequest) returns (stream VerifiableEvents) {
+        option (google.api.http) = {
```
bmwill: Nix the Google api annotations

crates/event-streaming/proto/walrus/event/v1/service.proto
```diff
@@ -0,0 +1,175 @@
...
+
+    // The V2 Merkle inclusion proof.
+    PROOF_TYPE_MERKLE_PROOF_V2 = 3;
+}
+
+message StreamEventsRequest {
+    // The checkpoint sequence number from which to start streaming.
+    // Events from checkpoints before this number will not be sent.
+    // Defaults to 0, starting from the first available checkpoint.
+    uint64 starting_checkpoint_sequence_number = 1;
```
bmwill: We're going to need to be very careful here when it comes to retention and how far back you can start the stream.

bmwill: Also everything that can needs to be prefixed with optional

crates/event-streaming/src/lib.rs
bmwill: Yeah this is sort of not testing anything

crates/event-streaming/build.rs
```diff
@@ -0,0 +1,199 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+use std::{fs, path::PathBuf, process::Command};
+
+const SUI_REPO_URL: &str = "https://github.com/MystenLabs/sui.git";
```
bmwill, line 6: As mentioned I'll be moving all the proto files to another repo soon.

crates/event-streaming/build.rs
sadhansood: Thanks @wbbradley, I think your other point about ensuring sui repo has the same GIT_TAG resonated with me due to which I made it the default option (we cannot ensure local sui repo has the same tag or it is complicated to deal with i think). The good thing with the current approach of downloading it from github is we cache the sui repo with tag in the file system path i.e. `target/sui_clone_testnet-v1.50.1/crates/sui-rpc-api/`, die to which we won't we doing the download every time we build (or until we do cargo clean).

crates/event-streaming/build.rs
sadhansood: handled now

crates/event-streaming/src/lib.rs
sadhansood: agree, removed tests

crates/event-streaming/proto/walrus/event/v1/service.proto
sadhansood: done

crates/event-streaming/proto/walrus/event/v1/service.proto
sadhansood: done

crates/event-streaming/proto/walrus/event/v1/service.proto
sadhansood: done (although it seems everything without optional is optional implicitly but i added it)

crates/event-streaming/build.rs
sadhansood: will update the repo once you do it, will make the git clone much cheaper.

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
```diff
@@ -0,0 +1,164 @@
...
+
+    // The method used to identify the event stream. The client MUST specify
+    // exactly one of these options. This makes the API explicit and avoids
+    // ambiguity between V1 (package-based) and V2 (object-based) streams.
+    oneof stream_selector {
+        // V1: Subscribe to all events emitted by a specific package.
+        PackageSubscription package_id = 2;
+
+        // V2: Subscribe to events managed by a dedicated stream head object.
+        StreamHead stream_head = 3;
```
mskd12, line 60: Suggest renaming to stream_id

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
```diff
@@ -0,0 +1,164 @@
...
+    repeated sui.rpc.v2beta.Event events = 2;
+
+    optional uint64 checkpoint_sequence_number = 3;
+
+    // The cryptographic proof that ties the events to the checkpoint summary.
+    // This is designed to be upgradeable.
+    optional Proof proof = 4;
+}
+
+message StreamHead {
```
mskd12, line 89: same as above: suggest rename to StreamID

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
```diff
@@ -0,0 +1,164 @@
...
+// Represents a proof that a specific event was not included in a checkpoint.
+// This is only possible with a V2 (Merkle tree) proofing system.
+message NonInclusionProof {
+    // The sequence number of the checkpoint this proof applies to.
+    optional uint64 checkpoint_sequence_number = 1;
+
+    // A valid non-inclusion proof consists of proving the existence of the
+    // two artifacts that are the immediate neighbors of the item being disproven
+    // in the sorted Merkle tree.
+    repeated MerkleInclusionProof adjacent_proofs = 2;
```
mskd12, line 158: Is it better to instead have two fields that are optional? One for the left leaf and another for the right leaf.

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
```diff
@@ -0,0 +1,164 @@
...
+    // supports for a given checkpoint. If empty, it uses its default behavior.
+    // Example: [MERKLE_PROOF_V2, CHECKPOINT_CONTENTS_V1] means client prefer V2,
+    // but will accept V1 if V2 is not available (would not accept response without
+    // a proof)
+    repeated ProofType desired_proof_types = 4;
+}
+
+// VerifiableEvents is the message type for the stream.
+// Each message represents a single checkpoint that contained Walrus events.
+message VerifiableEvents {
```
mskd12, line 74: Rename to CheckpointStreamEvents?

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
```diff
@@ -0,0 +1,164 @@
...
+    // The summary of the checkpoint containing these events. The client uses
+    // the `digest` field from this summary to verify the accompanying proof.
+    optional sui.rpc.v2beta.CheckpointSummary summary = 1;
```
mskd12, line 77: I suspect it makes more sense for the summary to be part of the proof, e.g., in v1 you would have a redundancy of summary (both here and in the contents).

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
```diff
@@ -0,0 +1,164 @@
...
+import "sui/rpc/v2beta/checkpoint_contents.proto";
+import "sui/rpc/v2beta/checkpoint_summary.proto";
+import "sui/rpc/v2beta/event.proto";
+
+// EventStreamingService provides a verifiable stream of events from the Sui
+// blockchain for an event stream. It allows clients to receive events
+// along with cryptographic proofs that tie them to a specific checkpoint.
+// It is designed to be forward-compatible, allowing for proof mechanisms
+// to be upgraded over time without breaking clients.
+service EventStreamingService {
```
mskd12, line 16: One API call that's missing is the one that fetches validator keys. But perhaps it belongs in a more generic `LightClientService`? If we do decide to create such a separate service, I feel like the `GetNonInclusionProof` call also belongs there.

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
```diff
@@ -0,0 +1,164 @@
...
+        MerkleInclusionProof merkle_proof = 2;
+
+        // No-Proof Option: Sent when the client trusts the server and opts out
+        // of verification to save bandwidth and CPU.
+        google.protobuf.Empty proof_none = 3;
+    }
+}
+
+// Represents a Merkle inclusion proof for a checkpoint artifact.
+message MerkleInclusionProof {
```
mskd12, line 122: Is it useful to go into this level of detail instead of just setting `MerkleInclusionProof` and `MerkleNonInclusionProof` as bytes? I'm mostly worried about keeping the defs flexible (until the Merkle PR merges). I don't know anything about gRPC and its best practices though, so perhaps there are good reasons to do so...

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
sadhansood: I like the two optional fields better, good idea!

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
sadhansood: sure, although I also like the word verifiable in there since it also has other stuff in there meant to verify the events.

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
sadhansood: contents don't have the summary (afaik)

crates/event-streaming/proto/walrus/event/v1alpha/service.proto
sadhansood: Validator keys are in end of epoch checkpoint (afaik). If in future we have a dedicated light client service, we can remove this endpoint from here, but right now we need this to be available in the local sidecar service, wdyt?

mlegner:
@sadhansood I've removed the review request for myself for now as multiple other people are already looking at this. Please feel free to re-request a review at any time though.

PR #2254, opened 2025-06-30
refactor: bump Rust to 1.88 and fix new Clippy warnings

mlegner:
## Description

Bump the Rust toolchain to the latest 1.88 release and fixes all new Clippy warnings with `cargo clippy --fix`. Additionally run `cargo fmt` again and fix a few too long lines manually.

## Test plan

Existing test suite.

### DIFF
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -605,16 +605,22 @@ impl ValidBlobInfoV1 {
     fn to_blob_status(&self, current_epoch: Epoch) -> BlobStatus {
         // TODO: The following should be adjusted/simplified when we have proper cleanup (WAL-473).
         let deletable_counts = DeletableCounts {
-            count_deletable_total: self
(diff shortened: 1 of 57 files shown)


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -605,16 +605,22 @@ impl ValidBlobInfoV1 {
     fn to_blob_status(&self, current_epoch: Epoch) -> BlobStatus {
         // TODO: The following should be adjusted/simplified when we have proper cleanup (WAL-473).
         let deletable_counts = DeletableCounts {
-            count_deletable_total: self
+            count_deletable_total: if self
```
karlwuest: Nit: I think the readability would be a bit better here if we assign `count_deletable_total` and `count_deletable_certified` outside of the construction of `deletable_counts`. This is just a matter of preference, though, so feel free to ignore.

karlwuest (approved):
Thanks, lgtm!

PR #2258, opened 2025-06-30
fix: Pass ending blob checkpoint instead of event index

sadhansood:
## Description

The PR fixes the call by passing ending checkpoint instead of ending index.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/event/event_processor/catchup.rs
+++ b/crates/walrus-service/src/event/event_processor/catchup.rs
@@ -314,7 +314,7 @@ impl EventBlobCatchupManager {
             }
 
             tracing::info!(
-                "Processed event blob {} with {} events, last event index: {}, \
+                "processed event blob {} with {} events, last event index: {}, \
                 start checkpoint: {}, end checkpoint: {}",
                 blob_id,
                 downloaded_blob.events.len(),
--- a/crates/walrus-service/src/event/event_processor/catchup.rs
+++ b/crates/walrus-service/src/event/event_processor/catchup.rs
@@ -35,6 +35,8 @@ struct DownloadedBlob {
     epoch: Epoch,
     first_event_index: Option<IndexedStreamEvent>,
     events: Vec<IndexedStreamEvent>,
+    start_checkpoint: u64,
+    end_checkpoint: u64,
 }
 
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/event/event_processor/catchup.rs
```diff
@@ -35,6 +35,8 @@ struct DownloadedBlob {
     epoch: Epoch,
     first_event_index: Option<IndexedStreamEvent>,
     events: Vec<IndexedStreamEvent>,
+    start_checkpoint: u64,
```
halfprice, line 38: Question: do we want to include `start_checkpoint` if it is not used?

crates/walrus-service/src/event/event_processor/catchup.rs
```diff
@@ -312,14 +314,17 @@ impl EventBlobCatchupManager {
             }
 
             tracing::info!(
-                "Processed event blob {} with {} events, last event index: {}",
+                "Processed event blob {} with {} events, last event index: {}, \
```
halfprice:
```suggestion
                "processed event blob {} with {} events, last event index: {}, \
```

halfprice (approved):
Thanks for the quick fix @sadhansood !

crates/walrus-service/src/event/event_processor/catchup.rs
sadhansood: Yeah, i think this is for logging purposes, it's probably better to have it.

mlegner:
Thanks a lot for the fix, @sadhansood! Given how easy it is to confuse various types of indices, IDs, sequence numbers, etc., it might make sense to use the newtype pattern in a few places to get help from the compiler making sure we pass the correct values. WDYT?

sadhansood:
> Thanks a lot for the fix, @sadhansood! Given how easy it is to confuse various types of indices, IDs, sequence numbers, etc., it might make sense to use the newtype pattern in a few places to get help from the compiler making sure we pass the correct values. WDYT?

I agree Markus, that would definitely help here. I can take it up.

PR #2259, opened 2025-06-30
test: add test coverage for event index consistency check

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/event/event_blob_downloader.rs
```diff
@@ -145,6 +147,10 @@ impl EventBlobDownloader {
 
             let blob_path = path.join(prev_event_blob.to_string());
             let blob = if blob_path.exists() {
+                metrics
+                    .event_catchup_manager_event_blob_fetched
```
halfprice, line 151: I feel this metrics is still good to have. So I added it back.

crates/walrus-service/src/event/event_processor/catchup.rs
```diff
@@ -493,10 +521,21 @@ impl EventBlobCatchupManager {
                     *state = Some(*expected_index + 1);
                     Some(event)
                 }
-                None => Some(event),
+                None => {
+                    // Ensure sequential event index stored in event blob.
+                    *state = Some(event.index + 1);
+                    Some(event)
+                }
```
halfprice, line 528: This change requires some attention. I tried to make the code comply with the function doc.

crates/walrus-service/src/test_utils.rs
```diff
@@ -2329,7 +2334,7 @@ pub mod test_cluster {
             Self {
                 epoch_duration: None,
                 test_nodes_config: None,
-                num_checkpoints_per_blob: Some(10),
+                num_checkpoints_per_blob: Some(100),
```
halfprice, line 2381: With smaller value, I'm not able to trigger event catching up using event blob due to that reading event blob always returns non-existence due to delayed certified event processing.

mlegner: I'm not quite sure I understand the root of the problem: Is it that the event blob is already expired? Or not certified yet?

crates/walrus-service/src/test_utils.rs
```diff
@@ -1897,6 +1932,15 @@ impl TestClusterBuilder {
         self
     }
 
+    /// Sets the event stream catchup min checkpoint lag for all storage nodes.
+    pub fn with_event_stream_catchup_min_checkpoint_lag(
+        mut self,
+        event_stream_catchup_min_checkpoint_lag: Option<u64>,
```
mlegner, line 1938: Nit: To be consistent with other functions like `with_num_checkpoints_per_blob`, this should take a `u64` instead of an `Option`.

crates/walrus-service/src/test_utils.rs
halfprice: The problem is that blob certification on chain finishes as soon as a quorum of votes received. However, that certified events also need to be processed by nodes in order for the event blob to be read. If num_checkpoints_per_blob is too low, we ran into the situation where the latest certified event blob is only within 10 events away, to short for all the storage nodes to finish processing it, and therefore almost always returns blob nonexist when being read.

crates/walrus-service/src/test_utils.rs
halfprice: Yeah, I'm aware of that. I want to put the default value in [the StorageNodeConfig creation](https://github.com/MystenLabs/walrus/blob/bb26978886d3e3829a5713e1c0072fc163fa2ef0/crates/walrus-service/src/test_utils.rs#L1078) instead of at the callsite so that we have an aggregated place to see all the default value.

`with_num_checkpoints_per_blob` is weird because the callsite is an option, and it passes in as u64, and updated to an option field. I feel it should also just take Option<> as an input.

crates/walrus-service/src/event/event_processor/catchup.rs
sadhansood: Change looks right to me, good to have this check even when starting with an empty db.

crates/walrus-service/src/test_utils.rs
sadhansood: ok this makes sense to me, with event blobs we don't fire register event until it has been certified. Which means to be able to read the event blob, we need some buffer to allow nodes to have received those events to facilitate blob read.

sadhansood (approved):
PR looks good to me, thanks @halfprice! Hopefully we can resolve comments from @mlegner before landing it.

PR #2261, opened 2025-06-30
chore: Quilt cli improvements

liquid-helium:
## Description

Addressed comments in #2197
Contributes to WAL-858

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,6 +1096,131 @@ impl FromStr for QuiltBlobInput {
...
+    )]
+    #[serde(default)]
+    identifiers: Vec<String>,
+
+    /// The tag key.
+    ///
+    /// It is required to be used with `--quilt-id` and `--value`.
+    ///
+    /// Example:
+    /// walrus read-quilt --quilt-id `<ID>` --tag `<KEY>` --value `<VALUE>`
```
mlegner: Important: The docstring isn't correct anymore. Also, I believe the example isn't needed as the format is already printed automatically (applies to other options as well).
```suggestion
    /// The tag key and value.
    ///
    /// It is required to be used with `--quilt-id`.
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,6 +1096,131 @@ impl FromStr for QuiltBlobInput {
...
+    /// It is required to be used with `--quilt-id`.
+    ///
+    /// Example:
+    /// walrus read-quilt --quilt-id `<ID>` --identifier `<IDENTIFIER>`...
+    #[arg(
+        long = "identifiers",
+        alias = "identifier",
+        conflicts_with = "tag",
+        conflicts_with = "quilt_patch_ids",
+        num_args = 0..,
```
mlegner: Nit: Can we make this `1..`?
```suggestion
        num_args = 1..,
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,6 +1096,131 @@ impl FromStr for QuiltBlobInput {
...
+
+impl QuiltPatchQuery {
+    /// Returns a QuiltPatchSelector from the command line arguments.
+    pub fn to_selector(&self) -> Result<QuiltPatchSelector> {
+        if !self.identifiers.is_empty() {
+            if !self.tag.is_empty() || !self.quilt_patch_ids.is_empty() || self.quilt_id.is_none() {
+                return Err(Self::invalid_query_error());
+            }
+            Ok(QuiltPatchSelector::ByIdentifier(QuiltPatchByIdentifier {
+                quilt_id: self.quilt_id.expect("quilt_id should be present"),
```
mlegner: Important: This is only checked when using the "normal" CLI, but not in JSON mode.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,6 +1096,131 @@ impl FromStr for QuiltBlobInput {
...
+                identifiers: self.identifiers.clone(),
+            }))
+        } else if !self.tag.is_empty() {
+            if self.tag.len() != 2 {
+                return Err(anyhow!("Only one tag is supported for now."));
+            }
+            if self.quilt_id.is_none() || self.quilt_patch_ids.is_empty() {
```
mlegner: Important:
```suggestion
            if self.quilt_id.is_none() || !self.quilt_patch_ids.is_empty() {
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,6 +1096,131 @@ impl FromStr for QuiltBlobInput {
...
+        help_heading = "Arguments",
+    )]
+    #[serde(default)]
+    quilt_patch_ids: Vec<QuiltPatchId>,
+}
+
+impl QuiltPatchQuery {
+    /// Returns a QuiltPatchSelector from the command line arguments.
+    pub fn to_selector(&self) -> Result<QuiltPatchSelector> {
+        if !self.identifiers.is_empty() {
```
mlegner: Remark: I believe the whole multi-level if/else if/else statement could be made a bit easier to read by using something like `match (self.quilt_id, self.tag.len(), self.identifiers.len(), self.quilt_patch_ids.len())` where we catch all the errors in the final `_` pattern.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,6 +1096,131 @@ impl FromStr for QuiltBlobInput {
...
+            Err(Self::invalid_query_error())
+        }
+    }
+
+    /// Returns an error message for an invalid query.
+    fn invalid_query_error() -> anyhow::Error {
+        anyhow!(
+            "Exactly one query type must be specified. Valid query types are:\n\
+            - --quilt-id <ID> --identifier <IDENTIFIER>...\n\
+            - --quilt-id <ID> --tag <KEY> --value <VALUE>\n\
```
mlegner: This is no longer correct:
```suggestion
            - --quilt-id <ID> --tag <KEY> <VALUE>\n\
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,6 +1096,131 @@ impl FromStr for QuiltBlobInput {
...
+            if self.quilt_id.is_none() || self.quilt_patch_ids.is_empty() {
+                return Err(Self::invalid_query_error());
+            }
+            Ok(QuiltPatchSelector::ByTag(QuiltPatchByTag {
+                quilt_id: self.quilt_id.expect("quilt_id should be present"),
+                tag: self.tag[0].clone(),
+                value: self.tag[1].clone(),
+            }))
+        } else if !self.quilt_patch_ids.is_empty() {
+            if self.quilt_id.is_none() {
```
mlegner: Important:
```suggestion
            if self.quilt_id.is_some() {
```

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,6 +1096,131 @@ impl FromStr for QuiltBlobInput {
...
+            Ok(QuiltPatchSelector::All(
+                self.quilt_id.expect("quilt_id should be present"),
+            ))
+        } else {
+            Err(Self::invalid_query_error())
+        }
+    }
+
+    /// Returns an error message for an invalid query.
+    fn invalid_query_error() -> anyhow::Error {
```
mlegner, line 1209: Remark: This will only ever be printed when using JSON mode, as the relevant checks are otherwise already performed by clap, resulting in more specific errors like the following:
```
error: the argument '--quilt-patch-ids [<QUILT_PATCH_IDS>...]' cannot be used with '--quilt-id <QUILT_ID>'
```

In JSON mode, the query looks a bit different, so the message may not be too useful.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -269,87 +269,10 @@ pub enum CliCommands {
     /// Read quilt patches (blobs) from Walrus.
     #[command(override_usage = "walrus read-quilt <Arguments> [OPTIONS]")]
```
mlegner: Nit: To be consistent with other clap formatting maybe something like the following:
```suggestion
    #[command(override_usage = "walrus read-quilt ARGUMENTS [OPTIONS]")]
```

crates/walrus-service/src/client/cli/args.rs
liquid-helium: ah! I noticed this this morning when I tested it manually :/ fixed!

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Yes, I think we can, I thought we couldn't, also made changes for identifiers.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Suggestion applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Good point, I updated the error message a bit, hopefully it can provide more help in case of error.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Good idea! change applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Switched to the match pattern.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,50 +1164,56 @@ pub struct QuiltPatchQuery {
...
-            if self.tag.len() != 2 {
-                return Err(anyhow!("Only one tag is supported for now."));
-            }
-            if self.quilt_id.is_none() || self.quilt_patch_ids.is_empty() {
-                return Err(Self::invalid_query_error());
+        match (
+            !self.identifiers.is_empty(),
+            !self.tag.is_empty(),
+            !self.quilt_patch_ids.is_empty(),
+            self.quilt_id.is_some(),
```
mlegner: Nit: If we match on `self.quilt_id`, we can match `Some(quilt_id)` instead of having to extract it with `.expect("quilt_id should be present")`.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1173,50 +1164,56 @@ pub struct QuiltPatchQuery {
 impl QuiltPatchQuery {
     /// Returns a QuiltPatchSelector from the command line arguments.
     pub fn to_selector(&self) -> Result<QuiltPatchSelector> {
```
mlegner: Suggestion: Change this to take an owned reference, then we can avoid the clones.
```suggestion
    pub fn into_selector(self) -> Result<QuiltPatchSelector> {
```
This may require doing a clone of the whole query, but it exposes the fact that we do in fact clone everything to the caller.

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -1445,7 +1451,7 @@ impl ClientCommandRunner {
                 out_dir.join(new_filename)
             };
 
-            std::fs::write(&output_file_path, blob.data())?;
+            std::fs::write(&output_file_path, blob.take_blob())?;
```
mlegner, line 1454: Praise: I like this approach to just remove the blob data when writing it!

mlegner (approved):
> > Thanks a lot for the changes, @liquid-helium! A few more comments about some of the details.
> 
> Thank you @mlegner for the review, and I apologize for these issues. I should have been more attentive.

Thanks a lot for the changes, LGTM. And no problem, I know that our CLI can be a bit finicky.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Good idea, change applied.

crates/walrus-service/src/client/cli/args.rs
liquid-helium: Nice idea, change applied.

PR #2265, opened 2025-07-01
fix(dbtool): fix drop column family command

mlegner:
## Description

Previously, the `walrus-node db-tool drop-column-family` command failed with the following error:
```
Error: Invalid argument: Column families not opened:
```

Apparently it is necessary to open all column families (https://github.com/rust-rocksdb/rust-rocksdb/issues/608). With this PR, we make sure to open all column families. Also, it allows passing multiple CF names to the command.

## Test plan

- [x] Manual testing.

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -463,6 +463,9 @@ impl SimStorageNodeHandle {
                                     10,
                                 )));
                             } else {
+                                // TODO(WAL-912): we need to alert the test if the node is stopped
+                                // unexpectedly. Currently, node crashing may not be noticed and
+                                // the test will hang.
                                 tracing::info!("node stopped with error: {e}");
                             }
                         }
--- a/crates/walrus-simtest/tests/simtest_event_blob.rs
+++ b/crates/walrus-simtest/tests/simtest_event_blob.rs
@@ -239,6 +239,12 @@ mod tests {
 
         kill_all_storage_nodes(&node_handles).await;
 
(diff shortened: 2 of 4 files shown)


crates/walrus-service/src/node/dbtool.rs
```diff
@@ -472,11 +473,31 @@ fn count_certified_blobs(db_path: PathBuf, epoch: Epoch) -> Result<()> {
 }
 
 /// Drop a column family from the RocksDB database.
-fn drop_column_family(db_path: PathBuf, column_family_name: String) -> Result<()> {
-    let db = DB::open(&RocksdbOptions::default(), db_path)?;
+fn drop_column_families(db_path: PathBuf, column_family_names: Vec<String>) -> Result<()> {
+    let db_column_family_names = DB::list_cf(&RocksdbOptions::default(), &db_path)?;
+    let db = DB::open_cf(
```
liquid-helium, line 478: shall we ignore the` DB already open ` error, before this improvement, the `drop family` is supposed to be used on a running storage node.

Another option is we only allow this operation when the storage node is not running, in that case, an explicit help message for this command would be helpful for the users.

mlegner: > the drop family is supposed to be used on a running storage node.

Ah, I wasn't aware of that. I was trying to use it with a stopped node.

Overall it seems safer to me to only allow it on a stopped node as I'm not really sure how the node behaves if suddenly some CFs are missing. I can add a more useful error message.

liquid-helium: The current implementation to require the DB closed makes more sense! 💯

PR #2266, opened 2025-07-01
fix: Update last certified event blob index upon startup

sadhansood:
## Description

Closes WAL-693.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


mlegner (approved):
Thanks, @sadhansood! 👍

PR #2273, opened 2025-07-02
fix(node): fix continuing event stream from init state

mlegner:
## Description

There are some problems with the `continue_event_stream` and the `EventProcessor::events` functions:

The event_stream returned by the continue_event_stream function always uses the lower of the provided event cursors, and does not necessarily match the returned `actual_event_index`. Then, in the `process_events` function, the returned event stream is zipped with an index stream which starts at the returned `actual_event_index`, so there can be a mismatch.

Furthermore, the `events` function doesn't work correctly when there aren't any events at the provided index: It doesn't check whether the events it polls from the event store actually match the expected index and updating the index doesn't work properly as a result.

What can happen in a node that has empty `events` and `event_blob_writer` DBs is that the event stream is started at index 0 (because that is the `event_blob_writer_cursor`), while the index stream is started at the "repositioned" index. The events function then keeps polling the events store with a low starting index (0, 1000, 2000, ...) such that always the same events (the first ones in the incomplete history) are returned and included in the stream.

This commit fixes the problems as follows:

- Modify the `continue_event_stream` function to return an event stream matching the returned event cursor.
- Add a consistency check in the `events` function which aborts when there is an index mismatch. (This should not actually be triggered, it is just a safeguard against further bugs.)

## Test plan

- [ ] Test on PTN.
- [ ] Add a simtest.

---

## Release notes

- [x] Storage node: Fixes an issue that can occur when a node starts with deleted `events` and `event_blob_writer` DBs.


mlegner:
@halfprice I was able to test the fix on PTN. The simtests are failing because of a timeout, but I don't think this is related to these changes: It has been [happening](https://github.com/MystenLabs/walrus/actions/runs/16018751331/job/45190545732) on `main` already since merging #2259.

crates/walrus-service/src/node/system_events.rs
```diff
@@ -292,9 +292,24 @@ impl SystemEventProvider for EventProcessor {
...
+                if n_events > 0 {
+                    let last_event_index = events
+                        .last()
+                        .expect("we just checked that events is not empty")
+                        .index;
+                    assert!(
+                        last_event_index == next_element_index - 1,
+                        "event index inconsistency in event store",
+                    );
+                };
```
liquid-helium, line 317: Question: So if the caller is requesting events that does not exist, this will check fail? Can we add this to the doc string of this function?

crates/walrus-service/src/node/system_events.rs
```diff
@@ -292,9 +292,24 @@ impl SystemEventProvider for EventProcessor {
                     .poll(element_index)
                     .inspect_err(|error| tracing::error!(?error, "failed to poll event stream"))
                     .ok()?;
+
                 // Update the index such that the next future continues the sequence.
                 let n_events = u64::try_from(events.len()).expect("number of events is within u64");
-                Some((stream::iter(events), (interval, element_index + n_events)))
+                let next_element_index = element_index + n_events;
```
halfprice, line 307: Nit/Suggestion: just to make this function more robust, should we do something like

`next_element_index = events.last().index+1`? And not rely on the `+ n_events` math?

And then the below check can become something like

```
assert_eq!(events.last().index - events.first().index == n_events);
```

to check the continuity of the event index.

Right now, the poll() function requires the input `cursor` must exist in the event table, which becomes the caller's responsibility. But if we can provide something like poll(index), and the stream returns events from the first events after index, maybe easier to use.

If the current code is preferred, I'd suggest we document that the `events(cursor)` call requires cursor must exist in the event db in the function doc.

crates/walrus-service/src/node.rs
```diff
@@ -976,11 +1007,11 @@ impl StorageNode {
             Some(factory) => Some(factory.create().await?),
             None => None,
         };
-        let (event_stream, next_event_index) = self
+        let (event_stream, event_cursor) = self
```
halfprice, line 1010: Nit: since we are returning the full event_cursor now, maybe add a log entry to print it out. Maybe helpful for future debugging.

halfprice (approved):
Amazing work @mlegner ! Thanks for the debugging and the fix! This LGTM and I'll try to create a test to manifest this issue and test the issue.

crates/walrus-service/src/node/system_events.rs
mlegner: Good point, I added some information to the docstring. 👍

crates/walrus-service/src/node.rs
mlegner: Sounds good. I have a few log entries inside the `continue_event_stream` function, but I added one more here.

crates/walrus-service/src/node/system_events.rs
mlegner: I already have a check that the range of events is contiguous:
```rust
                if n_events > 0 {
                    let last_event_index = events
                        .last()
                        .expect("we just checked that events is not empty")
                        .index;
                    assert!(
                        last_event_index == next_element_index - 1,
                        "event index inconsistency in event store",
                    );
                };
```
This is essentially what you suggested, right?

I also added the expectation and possible panic to the function's docstring as suggested by @liquid-helium [above](https://github.com/MystenLabs/walrus/pull/2273#discussion_r2181035896).

PR #2275, opened 2025-07-02
refactor: rename subsidies to credits, simplify code

karlwuest:
## Description

Renames subsidies to credits in the rust codebase and simplifies some parts of the code (including removing deployment of the old subsidies from the deploy command). This is in preparation of integrating the new walrus subsidies (to be called by storage nodes), to avoid naming conflicts and confusion.

For context: We will still need the client side calls in the future for a credit contract (with the same interfaces as the current subsidies) that DevRel is building.

## Test plan

Existing tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-sdk/client_config_example.yaml
mlegner, line 1: Question: Do we have an issue to modify the actual config files in [`setup/`](https://github.com/MystenLabs/walrus/tree/main/setup) after the contract upgrade?

mlegner (approved):
Thanks, @karlwuest, LGTM. Maybe @zihehuang can take a brief look as well.

crates/walrus-sdk/client_config_example.yaml
karlwuest: We can update them after the contract upgrade, but not before.

mlegner: Yes, I understand. We should just make sure to do it then. 🙂

karlwuest: Ahh, you meant a linear issue, I misunderstood "issue" as "is there a problem to do that" 😅. I'll just add it to the existing cleanup issue for the subsidies/credits.

crates/walrus-sui/src/test_utils/system_setup.rs
```diff
@@ -102,11 +102,12 @@ impl SystemContext {
...
-        ContractConfig::new_with_subsidies(
-            self.system_object,
-            self.staking_object,
-            self.subsidies_object,
-        )
+        ContractConfig {
+            system_object: self.system_object,
+            staking_object: self.staking_object,
+            credits_object: self.credits_object,
+            subsidies_object: None,
```
zihehuang, line 109: I assume this `subsidies_object` here is for backward compatibility?

zihehuang (approved):

crates/walrus-sui/src/test_utils/system_setup.rs
karlwuest: Yes, exactly. That's there so that the current configs are still correctly using the subsidies, but that we can remove the field once the subsidies contract is only used for credits.

PR #2279, opened 2025-07-03
feat: Quilt publisher

liquid-helium:
## Description

Added publisher endpoint to upload blobs as quilts to Walrus.

Close WAL-915.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [x] Publisher: Added "/v1/quilts" publisher endpoint to store quilts.
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon/routes.rs
+++ b/crates/walrus-service/src/client/daemon/routes.rs
@@ -950,24 +950,17 @@ async fn parse_multipart_quilt(
     }
 
     let mut res = Vec::with_capacity(blobs_with_identifiers.len());
+    for (identifier, data) in blobs_with_identifiers {
+        let tags = if let Some(meta) = metadata_map.get(&identifier) {
+            meta.tags
+                .iter()
+                .map(|(k, v)| (k.clone(), v.to_string()))
+                .collect()
+        } else {
+            BTreeMap::new()
+        };
 
-    if !metadata_map.is_empty() {
-        for (identifier, data) in blobs_with_identifiers {
-            let tags = if let Some(meta) = metadata_map.get(&identifier) {
-                meta.tags
-                    .iter()
-                    .map(|(k, v)| (k.clone(), v.to_string()))
-                    .collect()
-            } else {
-                BTreeMap::new()
-            };
-
-            res.push(QuiltStoreBlob::new_owned(data, identifier)?.with_tags(tags));
-        }
-    } else {
-        for (identifier, data) in blobs_with_identifiers {
-            res.push(QuiltStoreBlob::new_owned(data, identifier)?);
-        }
+        res.push(QuiltStoreBlob::new_owned(data, identifier)?.with_tags(tags));
     }
 
     Ok(res)


crates/walrus-service/daemon_openapi.yaml
```diff
@@ -312,6 +321,160 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/Status'
+  /v1/quilts:
+    put:
```
wbbradley, line 325: nit: I think publisher is using PUT where it should use POST. https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.3 PUT is most commonly used when you know the target URI for the resource and you are creating or updating it. In this case, we are not using a URI for the quilt (or blob), we're just posting data to this collection endpoint, and relying on the semantics of the endpoint to do what it will. (ie: after this PUT, we can't run GET on the same uri and get back our resource)

This is definitely a nit, but maybe in a future version of this API we can change these methods to use POST.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -911,4 +1129,37 @@ mod tests {
...
+                "tags": {
+                    "author": "alice",
+                    "version": "2.0"
+                }
+            },
+            {
+                "identifier": "logo-2024",
+                "tags": {
+                    "type": "logo",
+                    "format": "png"
```
wbbradley: What happens when an integer or float or array value is there on the right-hand side? Will that error out?

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -344,7 +344,7 @@ pub trait QuiltColumnRangeReader {
 }
 
 /// The version of the quilt.
-#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq)]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq, Eq, utoipa::ToSchema)]
```
jpcsmith: Remark: utoipa is optional in the walrus-core crate, so wherever this is used may need to be annotated with a `cfg` directive.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -702,6 +723,11 @@ pub struct PublisherQuery {
     /// number of epochs.
     #[serde(default)]
     pub force: bool,
+    /// The quilt version to use (for quilt endpoints only).
+    /// Valid values: "v1", "V1", or "1". Defaults to "v1" if not specified.
+    #[serde(default)]
+    #[serde(deserialize_with = "deserialize_quilt_version")]
+    pub quilt_version: Option<QuiltVersionEnum>,
```
jpcsmith, line 730: Suggestion: Create a separate query type for quilt instead of reusing PublisherQuery, as it pollutes the documentation. You can separate out the common fields and make use of serde's flatten for this.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -766,6 +793,195 @@ impl PublisherQuery {
     }
 }
 
+/// A helper structure to hold metadata for a quilt patch.
+#[derive(Debug, Deserialize)]
+pub struct QuiltPatchMetadata {
+    pub identifier: String,
+    pub tags: Option<BTreeMap<String, serde_json::Value>>,
```
jpcsmith: Question: Reason not to use `serde_json::Map` for this?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -766,6 +793,195 @@ impl PublisherQuery {
...
+}
+
+/// Store multiple blobs as a quilt using multipart/form-data.
+///
+/// Accepts a multipart form with blobs and optional per blob Walrus-native metadata.
+/// The form contains:
+/// - Blobs identified by their identifiers as field names (required)
+/// - A `{WALRUS_NATIVE_METADATA_FIELD_NAME}` field containing a JSON array with per blob metadata
```
jpcsmith: Question: Why doesn't this field follow json naming conventions (camelCase)? Or rather, if you intended for the variable to expand, it doesnt.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -766,6 +793,195 @@ impl PublisherQuery {
...
+        (status = 400, description = "The request is malformed"),
+        (status = 413, description = "The quilt is too large"),
+        StoreBlobError,
+    ),
+)]
+pub(super) async fn put_quilt<T: WalrusWriteClient>(
+    State(client): State<Arc<T>>,
+    Query(query): Query<PublisherQuery>,
+    bearer_header: Option<TypedHeader<Authorization<Bearer>>>,
+    multipart: Multipart,
```
jpcsmith, line 854: Question: Has the [size limit on the Multipart extension](https://docs.rs/axum/latest/axum/extract/struct.Multipart.html#large-files) been configured?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -766,6 +793,195 @@ impl PublisherQuery {
...
+
+    let blob_persistence = match query.blob_persistence() {
+        Ok(blob_persistence) => blob_persistence,
+        Err(error) => return error.into_response(),
+    };
+
+    // For now, we only support V1 quilts.
+    let result = match quilt_version {
+        QuiltVersionEnum::V1 => {
+            let quilt = match client
```
jpcsmith: Remark: Since only V1 exists and is supported, just make the quilt version check an assertion and save the indentation.
```rust
// runtime check
assert!(quilt_version == QuiltVersionEnum::V1); 
// or, compile time check that there are no other enum versions yet
let QuiltVersionEnum::V1 = quilt_version;`
```

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -766,6 +793,195 @@ impl PublisherQuery {
...
+
+/// Parse multipart form data and extract files with their metadata.
+async fn parse_multipart_quilt(
+    mut multipart: Multipart,
+) -> Result<Vec<QuiltStoreBlob<'static>>, anyhow::Error> {
+    let mut blobs_with_identifiers = Vec::new();
+    let mut metadata_map: BTreeMap<String, QuiltPatchMetadata> = BTreeMap::new();
+
+    while let Some(field) = multipart.next_field().await? {
+        let field_name = field.name().unwrap_or("").to_string();
```
jpcsmith, line 934: Question: Are we just silently overwriting fields which have the same name?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -766,6 +793,195 @@ impl PublisherQuery {
...
+        if field_name == WALRUS_NATIVE_METADATA_FIELD_NAME {
+            let metadata_json = field.text().await?;
+            for meta in serde_json::from_str::<Vec<QuiltPatchMetadata>>(&metadata_json)? {
+                let identifier = meta.identifier.clone();
+                metadata_map.insert(identifier, meta);
+            }
+        } else {
+            let data = field.bytes().await?.to_vec();
```
jpcsmith, line 947: Question: Why to_vec?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -766,6 +793,195 @@ impl PublisherQuery {
...
+            blobs_with_identifiers.push((field_name, data));
+        }
+    }
+
+    let mut res = Vec::with_capacity(blobs_with_identifiers.len());
+
+    if !metadata_map.is_empty() {
+        for (identifier, data) in blobs_with_identifiers {
```
jpcsmith: Question: Why is it necessary to condition on the map being empty? If there is only one metadata entry, then it's pretty much the same as if there are no metadata entries. Can't the cases just be handled with the first branch of the if statement?

crates/walrus-service/daemon_openapi.yaml
liquid-helium: Thanks for the tip! I'm not very familiar with these concepts and used the approach from put_blob(). Either option works for me, but since this also applies to put_blob()—where we don't have the URI when uploading the blob bytes—it seems logical to keep them consistent. Looping in @giac-mysten for input

wbbradley: Let's not worry about it for now. I don't want to hold things up for this nit.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: The parse function assumes the values are json::Value, although in the api string, we ask users to use string only, so we should be fine.
Also modified tests to verify this.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Thanks for the info, fix applied.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Good idea, change applied.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Switched to serde_json::Map

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Fixed.

crates/walrus-service/src/client/daemon.rs
```diff
@@ -206,6 +232,41 @@ impl WalrusWriteClient for Client<SuiContractClient> {
...
     }
 
+    async fn construct_quilt<V: QuiltVersion>(
+        &self,
+        blobs: &[QuiltStoreBlob<'_>],
+        encoding_type: Option<EncodingType>,
+    ) -> ClientResult<V::Quilt> {
+        let encoding_type = encoding_type.unwrap_or(DEFAULT_ENCODING);
+
+        self.quilt_client(QuiltClientConfig::default())
```
giac-mysten, line 243: Question: Are we sure the default is all we need? Or maybe should the `QuiltClientConfig` be part of the `ClientConfig`?

crates/walrus-service/src/client/daemon.rs
```diff
@@ -206,6 +232,41 @@ impl WalrusWriteClient for Client<SuiContractClient> {
...
+        quilt: V::Quilt,
+        encoding_type: Option<EncodingType>,
+        epochs_ahead: EpochCount,
+        store_optimizations: StoreOptimizations,
+        persistence: BlobPersistence,
+        post_store: PostStoreAction,
+    ) -> ClientResult<QuiltStoreResult> {
+        let encoding_type = encoding_type.unwrap_or(DEFAULT_ENCODING);
+
+        self.quilt_client(QuiltClientConfig::default())
```
giac-mysten, line 259: Same as above

crates/walrus-service/src/client/daemon/routes.rs
giac-mysten: The `DefaultBodyLimit` for the publisher is set [here](https://github.com/MystenLabs/walrus/blob/main/crates/walrus-service/src/client/daemon.rs#L391). IIUC the docs, this should be sufficient?

crates/walrus-service/src/client/daemon/routes.rs
giac-mysten, line 730: Question: Is this still needed? Or am I missing some changes?

crates/walrus-service/src/client/daemon/routes.rs
giac-mysten, line 854: Cool! I did not know of `Multipart`. Just curious, what is the advantage here over, e.g., a JSON array of "parts"?

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: I ended up creating a separate Query, but copied the fields over, if I make the current PublisherQuery a (flatten) field of the new QuiltPublisherQuery, it ended up with the following error, I tried to add a custom deserializer but still failed. Any suggestions will be great. Also, what documentation pollution are you worrying about?
```
curl -X PUT "http://127.0.0.1:31415/v1/quilts?epochs=1&quilt_version=V1" \
  -F "segment-1=@/Users/heliu/quilt/video-segments-2/segment_00313.ts" \
  -F "segment-2=@/Users/heliu/quilt/video-segments-2/segment_00313.ts" \
  -F "segment-3=@/Users/heliu/quilt/video-segments-2/segment_00313.ts" \
  -F "segment-4=@/Users/heliu/quilt/video-segments-2/segment_00313.ts" \
  -F "segment-5=@/Users/heliu/quilt/video-segments-2/segment_00313.ts" \
  -F "another_walrus.jpg=@/Users/heliu/quilt/walrus-46.jpg" \
>
Failed to deserialize query string: invalid type: string "1", expected u32%
```

liquid-helium: I reverted it back to the original version (with a new quilt_version in PublisherQuery, since PublisherQuery is also used in the `check_valid_upload` call, otherwise, we may need to create a separate path for the new query.
Let me know if you have a strong preference here.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Currently, it shares the same size limit of `put_blob()`. But it makes sense to use a dedicated one for quilt, thanks for the suggestion.

Added a new size limit for quilts.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: good idea, change applied.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Nice catch. For the _metadata part, I now added a check to avoid duplicated identifiers.
For the input file part, it will error out when constructing the quilt.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: To be used in the following: `res.push(QuiltStoreBlob::new_owned(data, identifier)?);`

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: I anticipate a large portion of quilt upload requests of a large number of blobs (around 100 maybe), without metadata, so use a separate branch to avoid calling` .get()` on the empty `metadata_map`, but I am fine either way.

crates/walrus-service/src/client/daemon.rs
liquid-helium: A configurable QuiltClientConfig will be better, added a TODO(WAL-927) here, since this affects not only Publisher.

crates/walrus-service/src/client/daemon.rs
liquid-helium: ditto

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: I kept this approach now due to the issues state in the above comment.
Let me know if there are other better options.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: I didn't know it either, and heard it from the customer, then I did some research, it seems that it is a popular format to transfer files, and there are some optimizations to transfer the bytes.

jpcsmith: @giac-mysten If you sent it as JSON, you'd need to send the binary payload as text (e.g. base64 encoded). Multipart supports sending the raw bytes.

crates/walrus-service/src/client/daemon/routes.rs
jpcsmith: Suggestion: Merging the branches would avoid the code diverging and generally reduce head-scratching. A hashmap's `get` is an O(1) operation. In the case of an empty hashmap it's even [special cased[(https://docs.rs/hashbrown/latest/src/hashbrown/map.rs.html#1351) to just immediately return None.

crates/walrus-service/src/client/daemon/routes.rs
giac-mysten: > @giac-mysten If you sent it as JSON, you'd need to send the binary payload as text (e.g. base64 encoded). Multipart supports sending the raw bytes.

ah, obviously JSON is text only 🤦 thanks. I guess some binary serialization would be more opaque, so this is nice!

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: change applied, thanks for the suggestion!

PR #2281, opened 2025-07-03
feat(client): add `permanent` flag to CLI and publisher

mlegner:
## Description

Currently, blobs are "permanent" as default and can be stored as "deletable" through the `--deletable` CLI option and a `deletable` query parameter on the publisher. We want to change this default behavior in the future, at which point we need a way to specify a blob as "permanent". Adding the `--permanent` flag in the same release where the default is changed is a bit dangerous as all users would have to change their invocations at exactly that point.

Adding the `--permanent` flag now allows users to already start calling `walrus store --permanent [...]` to make sure the behavior is unaffected by the later default change. Additionally, we log a warning about the upcoming default change from permanent to deletable if neither flag is specified.

The plan would then be to keep the `--deletable` around for some time longer and log a different warning about the default change if neither flag is specified.

Contributes to WAL-911.

## Test plan

Manual checks.

---

## Release notes

- [x] Publisher: Add a `permanent` query parameter. **Warning:** In the future, blobs will be stored as deletable by default; to keep the current behavior, you need to add the `permanent=true` query parameter.
- [x] CLI: Add a `--permanent` option to the `store` and `store-quilt` commands. **Warning:** In the future, blobs will be stored as deletable by default; to keep the current behavior, you need to add the `--permanent` option when storing blobs.


crates/walrus-sui/src/client.rs
```diff
@@ -339,15 +339,33 @@ impl BlobPersistence {
...
+            (false, false) => {
+                // TODO(WAL-911): Change the default behavior to return `Deletable` after a few
+                // releases.
+                tracing::warn!(
+                    "blob is marked as neither deletable nor permanent; blobs are currently \
+                    permanent by default, but this behavior will change in the future; use \
+                    `--deletable` or `--permanent` to explicitly specify the desired behavior"
+                );
+                Ok(Self::Permanent)
+            }
```
liquid-helium, line 359: Question: why do we need to have both `--deletable` and `--permanent`? Is it possible to have only the `--deletable`, if it is false, then it is permanent?

mlegner: Thanks for the review, @liquid-helium. 

> Question: why do we need to have both `--deletable` and `--permanent`? Is it possible to have only the `--deletable`, if it is false, then it is permanent?

This is the current state. However, as we want to make "deletable" the default in the future, we need a way to specify a blob as "permanent". Adding the `--permanent` flag in the same release where the default is a bit dangerous as all users would have to change their invocations at exactly that point.

Adding the `--permanent` flag now allows users to already start calling `walrus store --permanent [...]` to make sure the behavior is unaffected by the later default change.

The plan would then be to keep the `--deletable` around for some time longer and log a warning about the default change if neither flag is specified.

I'm adding this explanation to the PR description.

liquid-helium: Thanks for the explanation!

PR #2286, opened 2025-07-04
feat: Add Quilt patch identifier rules

liquid-helium:
## Description

Close WAL-914

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/encoding/quilt_encoding.rs
+++ b/crates/walrus-core/src/encoding/quilt_encoding.rs
@@ -480,17 +480,60 @@ impl<'a> QuiltStoreBlob<'a> {
     /// Returns if an identifier is valid.
     ///
     /// Note an empty identifier is not valid.
-    pub fn is_valid_identifier(identifier: impl AsRef<str>) -> bool {
+    ///
+    /// # Examples
+    ///
+    /// Valid identifiers:
+    /// ```
+    /// # use walrus_core::encoding::quilt_encoding::QuiltStoreBlob;
+    /// assert!(QuiltStoreBlob::check_identifier("test").is_ok());
+    /// assert!(QuiltStoreBlob::check_identifier("test123").is_ok());
+    /// assert!(QuiltStoreBlob::check_identifier("123test").is_ok());
+    /// assert!(QuiltStoreBlob::check_identifier("a").is_ok());
+    /// assert!(QuiltStoreBlob::check_identifier("A123").is_ok());
+    /// ```
+    ///
+    /// Invalid identifiers - doesn't start with alphanumeric:
+    /// ```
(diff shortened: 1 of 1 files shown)


crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -477,22 +477,40 @@ pub struct QuiltStoreBlob<'a> {
 }
 
 impl<'a> QuiltStoreBlob<'a> {
+    /// Returns if an identifier is valid.
+    ///
+    /// Note an empty identifier is not valid.
+    pub fn is_valid_identifier(identifier: impl AsRef<str>) -> bool {
+        let id = identifier.as_ref();
+        id.starts_with(|c: char| c.is_alphanumeric()) && id == id.trim_end()
+    }
```
mlegner: Nit (very optional): I generally like the pattern of having a function like the following:
```rust
    pub fn check_identifier(identifier: impl AsRef<str>) -> Result<(), QuiltError> {
        let id = identifier.as_ref();
        if id.starts_with(|c: char| c.is_alphanumeric()) && id == id.trim_end() {
            Ok(())
        } else {
            Err(QuiltError::InvalidIdentifier(identifier))
        }
    }
```

This can then be used simply as 
```rust
        Self::check_identifier(&identifier)?;
```

crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -2425,17 +2445,49 @@ mod tests {
...
+        assert!(!QuiltStoreBlob::is_valid_identifier("test\n"));
+        assert!(!QuiltStoreBlob::is_valid_identifier("test\r"));
+        assert!(!QuiltStoreBlob::is_valid_identifier("test   "));
+
+        // Valid identifiers with spaces/special chars in the middle
+        assert!(QuiltStoreBlob::is_valid_identifier("test blob"));
+        assert!(QuiltStoreBlob::is_valid_identifier("test-blob"));
+        assert!(QuiltStoreBlob::is_valid_identifier("test_blob"));
+        assert!(QuiltStoreBlob::is_valid_identifier("test.blob"));
     }
```
mlegner: Nit/suggestion: You could add these as doctests at the function as this is quite useful to understand the expectations.

mlegner (approved):
Thanks @liquid-helium! 🙏 

Just two optional nits.

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: That's a good one, change applied.

crates/walrus-core/src/encoding/quilt_encoding.rs

PR #2287, opened 2025-07-04
chore: upgrade walrus testnet contracts

karlwuest:
## Description

This PR prepares the walrus testnet contracts for the planned upgrade. The branch is used for voting on the upgrade and should only be merged once the upgrade is committed.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


testnet-contracts/subsidies/Move.lock
mlegner, line 1: Question: Are the changes to the `{subsidies,wal,wal_exchange}/Move.{lock,toml}` files intentional?

karlwuest: Yes, the versions of the sui dependency need to be the same in all dependencies.

PR #2291, opened 2025-07-07
feat: continue node recovery upon restart if the recovery takes longer than 1 epoch to finish

halfprice:
## Description

Previously, storage node assumed that a node recovery should finish within 1 epoch length, and if during node
restart, the node sees it is in RecoveryInProgress(epoch) state, and the epoch is lagging behind, it assumed that
the node is lagging again, and then directly enter RecoveryCatchUp mode.

However in practice, we have seen that it is not uncommon that the RecoveryInProgress takes more than 1 epoch.
Restarting node after epoch change would make the node goes back to RecoveryCatchup mode and therefore
not recovering any blobs until next epoch change, which can be 1 epoch length away (2 weeks on mainnet)

This PR addresses this issue by allowing in progress node recovery to continue after node restart. A list of changes
are made to achieve this goal:
- During epoch change, ongoing node recovery will be restarted to recovery to the new epoch, so that RecoveryInProgress should not lag.
- Previously, blob sync initiated during node recovery does not wait for any existing blob sync of the same blob to finish. Since now we need to restart node recovery during epoch change, we *have to* wait for any ongoing blob sync.
- Change latest event epoch using a tokio channel. Any functionality that requires it waits for it to become available and then proceed. There are mainly 2 places: (1) blob recovery, and (2) node recovery. 

Resolve WAL-917

## Test plan

Existing simtest. New simtest.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1208,6 +1208,7 @@ impl StorageNode {
         element_index: u64,
         maybe_epoch_at_start: &mut Option<Epoch>,
     ) -> anyhow::Result<()> {
+        sui_macros::fail_point!("fail_point_process_event");
         let _scope = monitored_scope::monitored_scope("ProcessEvent");
 
         let node_status = self.inner.storage.node_status()?;
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1301,12 +1301,6 @@ impl StorageNode {
             return Ok(());
         };
 
-        // Update initial latest event epoch. This is the first (non-extension) event the node
-        // processes.
-        self.inner
-            .latest_event_epoch_sender
-            .send(Some(event_epoch))?;
-
         tracing::debug!("checking the first contract event if we're severely lagging");
 
         // Clear the starting epoch, so that we won't make this check again in the current run.
@@ -1322,6 +1316,12 @@ impl StorageNode {
             );
             self.inner.set_node_status(NodeStatus::RecoveryCatchUp)?;
         }
+
+        // Update initial latest event epoch. This is the first (non-extension) event the node
+        // processes.
+        self.inner
+            .latest_event_epoch_sender
+            .send(Some(event_epoch))?;
 
         Ok(())
     }
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1176,25 +1176,23 @@ impl StorageNode {
             return Ok(false);
         };
 
-        let Some(current_event_epoch) = self.inner.try_get_current_event_epoch() else {
-            // When current event epoch is not set, we are processing events before the first epoch
-            // change start event, which should be excluded from processing.
-            return Ok(false);
-        };
-
-        if current_event_epoch >= first_complete_epoch {
-            return Ok(false);
-        }
-
-        if let EventStreamElement::ContractEvent(ContractEvent::EpochChangeEvent(
-            EpochChangeEvent::EpochChangeStart(EpochChangeStart { epoch, .. }),
-        )) = &stream_element.element
-        {
-            if *epoch >= first_complete_epoch {
-                // Processing the `EpochChangeStart` event for the first complete epoch will set the
-                // `current_event_epoch` to `epoch`, such that we will take the previous
-                // if-statement and return `false` for all future events.
+        // When current event epoch is not set, we are processing events before the first epoch
+        // change start event, which should be excluded from processing.
+        if let Some(current_event_epoch) = self.inner.try_get_current_event_epoch() {
+            if current_event_epoch >= first_complete_epoch {
                 return Ok(false);
+            }
+
+            if let EventStreamElement::ContractEvent(ContractEvent::EpochChangeEvent(
+                EpochChangeEvent::EpochChangeStart(EpochChangeStart { epoch, .. }),
+            )) = &stream_element.element
+            {
+                if *epoch >= first_complete_epoch {
+                    // Processing the `EpochChangeStart` event for the first complete epoch will set
+                    // the `current_event_epoch` to `epoch`, such that we will take the previous
+                    // if-statement and return `false` for all future events.
+                    return Ok(false);
+                }
             }
         }
 
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1280,6 +1280,8 @@ impl StorageNode {
             return Ok(());
         };
 
+        // Blob extensions do not contain their event emission epoch. So we use this to filter out
+        // blob extensions events.
         let Some(event_epoch) = event.event_epoch() else {
             return Ok(());
         };
@@ -1694,13 +1696,28 @@ impl StorageNode {
             );
         }
 
-        self.process_shard_changes_in_new_epoch(
-            event_handle,
-            event,
-            is_new_node_joining_committee,
-            shard_map_lock,
-        )
-        .await
+        if let NodeStatus::RecoveryInProgress(recovering_epoch) =
+            self.inner.storage.node_status()?
+        {
+            // If the node is already in recovery mode, we need to restart node recovery to recover
+            // to the latest epoch. This is to make sure that the node is always recovering to the
+            // latest epoch.
+            tracing::info!(
+                "node is currently recovering to epoch {recovering_epoch}, restarting \
+                node recovery to recover to the latest epoch {}",
+                event.epoch
+            );
+            self.start_node_recovery(event_handle, event, shard_map_lock)
+                .await
+        } else {
+            self.process_shard_changes_in_new_epoch(
+                event_handle,
+                event,
+                is_new_node_joining_committee,
+                shard_map_lock,
+            )
+            .await
+        }
     }
 
     /// Starts the node recovery process.
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -9,7 +9,7 @@ use std::{
     pin::Pin,
     sync::{
         Arc,
-        atomic::{AtomicBool, AtomicU32, Ordering},
+        atomic::{AtomicBool, Ordering},
     },
 };
 
@@ -534,7 +534,10 @@ pub struct StorageNodeInner {
     symbol_service: RecoverySymbolService,
     thread_pool: BoundedThreadPool,
     registry: Registry,
-    latest_event_epoch: AtomicU32, // The epoch of the latest event processed by the node.
+    // Receiver for watching the latest event epoch.
+    latest_event_epoch_watcher: watch::Receiver<Option<Epoch>>,
+    // Sender for updating the latest event epoch.
+    latest_event_epoch_sender: watch::Sender<Option<Epoch>>,
     consistency_check_config: StorageNodeConsistencyCheckConfig,
     checkpoint_manager: Option<Arc<DbCheckpointManager>>,
 }
@@ -649,6 +652,7 @@ impl StorageNode {
             }
         };
         let system_parameters = contract_service.fixed_system_parameters().await?;
+        let (latest_event_epoch_sender, latest_event_epoch_watcher) = watch::channel(None);
         let inner = Arc::new(StorageNodeInner {
             protocol_key_pair: config
                 .protocol_key_pair
@@ -676,7 +680,8 @@ impl StorageNode {
             thread_pool,
             encoding_config,
             registry: registry.clone(),
-            latest_event_epoch: AtomicU32::new(0),
+            latest_event_epoch_sender,
+            latest_event_epoch_watcher,
             consistency_check_config: config.consistency_check.clone(),
             checkpoint_manager,
         });
@@ -1288,9 +1293,7 @@ impl StorageNode {
 
         // Update initial latest event epoch. This is the first (non-extension) event the node
         // processes.
-        self.inner
-            .latest_event_epoch
-            .store(event_epoch, Ordering::SeqCst);
+        let _ = self.inner.latest_event_epoch_sender.send(Some(event_epoch));
 
         tracing::debug!("checking the first contract event if we're severely lagging");
 
@@ -1533,9 +1536,7 @@ impl StorageNode {
 
         // Update the latest event epoch to the new epoch. Now, blob syncs will use this epoch to
         // check for shard ownership.
-        self.inner
-            .latest_event_epoch
-            .store(event.epoch, Ordering::SeqCst);
+        let _ = self.inner.latest_event_epoch_sender.send(Some(event.epoch));
 
         Ok(())
     }
@@ -2240,12 +2241,22 @@ impl StorageNodeInner {
         Ok(metadata)
     }
 
+    /// Returns the latest event epoch.
+    /// The callsite must ensure that the first event is processed so that the
+    /// latest_event_epoch_watcher is initialized.
     fn current_event_epoch(&self) -> Epoch {
-        self.latest_event_epoch.load(Ordering::SeqCst)
+        self.latest_event_epoch_sender
+            .borrow()
+            .expect("latest event epoch is not set")
     }
 
     fn current_epoch(&self) -> Epoch {
         self.committee_service.get_epoch()
+    }
+
+    /// Returns a clone of the latest event epoch watcher receiver.
+    pub(crate) fn latest_event_epoch_watcher(&self) -> watch::Receiver<Option<Epoch>> {
+        self.latest_event_epoch_watcher.clone()
     }
 
     fn check_index<T>(&self, index: T) -> Result<(), IndexOutOfRange>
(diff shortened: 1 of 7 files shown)


crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -217,22 +229,11 @@ impl NodeRecoveryHandler {
...
-                    current_epoch = self.node.current_epoch(),
-                    "recovery epoch mismatch; skip recovery restart; next epoch change start event \
-                    will bring node to the latest state"
-                );
-                self.node.set_node_status(NodeStatus::RecoveryCatchUp)?;
-            }
+            // The `latest_event_epoch` is still set to `0` at this point.
+            self.node
+                .latest_event_epoch
+                .store(recovering_epoch, Ordering::SeqCst);
```
halfprice: This is definitely not set correctly yet. It should be the event epoch of the last processed events. Will see if we can set it correctly without storing the last event epoch in the DB.

mlegner: Is there a downside of storing the last event epoch in the DB?

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -217,22 +229,11 @@ impl NodeRecoveryHandler {
     /// Restarts any in progress recovery.
     pub async fn restart_recovery(&self) -> Result<(), TypedStoreError> {
```
mlegner: Remark: Given that we now keep updating the recovery epoch (IIUC the changes in `crates/walrus-service/src/node.rs`), the previous code here might actually be the correct behavior: In that case we *are in fact behind* on event processing and need to catch up. Also, it guarantees that we will see the latest "epoch change start" event and correctly switch back to `RecoveryInProgress`.

mlegner (commented):
Thanks for tackling this, @halfprice! I think the approach to just update the recovery epoch at every epoch change is a great way to solve this.

crates/walrus-service/src/node/node_recovery.rs
halfprice: The downside is mainly that (1) it stores a redundant information: we just want to know the first event's epoch, which will be available after processing it, and (2) adding complexity to the code.

I now uses a watch channel to store the latest_event_epoch and won't start the recovery process if the latest_event_epoch is not initialized. This feels easier to me, and any bug that around latest_event_epoch won't be persisted (and therefore easy to fix).

crates/walrus-service/src/node/node_recovery.rs
halfprice: Yes, but if we put the old code back, any existing node who has earlier RecoveryInProgress's epoch will enter RecoveryCatchUp again...

Essentially, the old code just does a event lagging checks, and we also check this at the beginning of the event processing, so I'll take a look tomorrow to see if that is sufficient. If not, we can consider adding the old code back after all the nodes are migrated away from old RecoveryInProgress (probably just after 1 release).

crates/walrus-service/src/node/node_recovery.rs
mlegner: > we just want to know the first event's epoch, which will be available after processing it

This is unfortunately not true for blob extensions.

> adding complexity to the code

I personally feel like the current approach is more complicated to reason about as one never knows if the `latest_event_epoch` is already initialized (I ran into this a few times in the past). However, I'm fine with keeping it this way.

crates/walrus-service/src/node.rs
```diff
@@ -2204,14 +2222,24 @@ impl StorageNodeInner {
         Ok(metadata)
     }
 
+    /// Returns the latest event epoch.
+    /// The callsite must ensure that the first event is processed so that the
+    /// latest_event_epoch_watcher is initialized.
     fn current_event_epoch(&self) -> Epoch {
-        self.latest_event_epoch.load(Ordering::SeqCst)
+        self.latest_event_epoch_sender
```
mlegner: Question: Don't we need the `latest_event_epoch_watcher` here?

crates/walrus-service/src/node/node_recovery.rs
mlegner: > we also check this at the beginning of the event processing

Ah, that's a good point. 👍

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -217,22 +230,26 @@ impl NodeRecoveryHandler {
...
+            tracing::info!("waiting for latest event epoch to be set to restart node recovery");
+
+            // Wait for the node.latest_event_epoch is initialized, so that the node recovery
+            // task can start.
+            while watcher.borrow().is_none() {
+                watcher
+                    .changed()
+                    .await
+                    .expect("watcher should not be closed");
             }
```
mlegner: Suggestion: Make the `current_event_epoch` function `async` and put this into it's body.

crates/walrus-service/src/node/node_recovery.rs
halfprice: > This is unfortunately not true for blob extensions.

Agreed, and I do wish that we can get blob extension's issuing epoch as well...

> as one never knows if the latest_event_epoch is already initialized

I share your concern, and I had this problem myself too.

The complexity of writing the latest_event_epoch comes from the node may reprocess any old events at any time, so the challenge is that when writing latest_event_epoch, we need to write the one that matches the first event when the node restarts. There is a decent chance that this value is not the epoch of the `highest_processed_event`'s epoch. One way may be to update the on disk `latest_event_epoch` when highest `persisted_event` changes, but not only now we have different in-memory and on disk value, this doesn't solve the problem because we may want to reprocess events before the latest `persisted_event`. So `reposition latest event epoch` also needs to be implemented. And at this point, I just don't want to think of other conner cases... and feel if just getting it from the first event is simpler to think about... It's possible that the current solution doesn't solve the problem and we may need to do it eventually. I think that change needs to be more careful and probably executed on a clear mind.

Another thing about persisted state is that, if they are wrong, it's very hard to fix them. But if an in memory state is wrong, it's easier to fix.

crates/walrus-service/src/node.rs
halfprice: Good catch! I think they refer to the same underlying value, but I should use `latest_event_epoch_watcher` for reads.

crates/walrus-service/src/node/node_recovery.rs
halfprice: Ok, I think I got a version that I feel good about.

Follow your suggestion @mlegner , I made a version of latest_event_epoch async and any places where it is required, calling it will wait until the first blob event processed and populated.

There is another version where we check if it is initialized and used only when recovery from an incomplete history.

Let's see how does this works and see if persisting latest event epoch is needed.

crates/walrus-service/src/node/node_recovery.rs
halfprice: Done.

crates/walrus-service/src/node.rs
```diff
@@ -1167,9 +1172,16 @@ impl StorageNode {
             return Ok(false);
         };
 
-        if self.inner.current_event_epoch() >= first_complete_epoch {
+        let Some(current_event_epoch) = self.inner.try_get_current_event_epoch() else {
```
halfprice: Previously, when processing events without epoch before first epoch change start (e.g., CheckpointBoundary), current_event_epoch() is set to 0, so this is ok. Now, the logic is changed to whether `current_event_epoch` is initialized in the tokio watch channel.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -333,16 +336,19 @@ impl BlobSyncHandler {
...
                 // try to increase the event cursor since it will only be advanced once that sync is
                 // finished or cancelled due to an invalid blob event.
                 event_handle.mark_as_complete();
-                finish_notify.notify_one();
+
+                existing_sync.get().finish_notify.clone()
             }
-        }
+        };
         Ok(finish_notify)
```
halfprice, line 352: Before, when there an existing blob sync task, start_sync returns a finish_notify that is notified immediately. Now, we return a notify that attaches to the existing blob sync tasks.

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -133,10 +145,25 @@ impl NodeRecoveryHandler {
                         while ongoing_syncs.len() >= max_concurrent_blob_syncs_during_recovery {
                             ongoing_syncs.next().await;
                         }
+
+                        // Since there is a wait, the blob might not be certified anymore. Check
```
halfprice, line 172: In addition to @mlegner 's change to watch for blob retirement inside BlobSyncHandler, adding another check here to avoid unnecessary tasks started.

crates/walrus-service/src/node/node_recovery.rs
```diff
@@ -215,25 +242,18 @@ impl NodeRecoveryHandler {
...
-                tracing::warn!(
-                    recovering_epoch,
-                    current_epoch = self.node.current_epoch(),
-                    "recovery epoch mismatch; skip recovery restart; next epoch change start event \
-                    will bring node to the latest state"
-                );
-                self.node.set_node_status(NodeStatus::RecoveryCatchUp)?;
-            }
+            tracing::info!("waiting for latest event epoch to be set to restart node recovery");
+            self.node.current_event_epoch().await?;
```
halfprice: I don't know we actually need to wait for the current_event_epoch() to be populated, since any caller of current_event_epoch() now waits for it. However, since this is the previous behavior, let's keep it to be safe. This is also an one of operation during node start up, so it doesn't have much overhead.

halfprice (commented):
Hi @liquid-helium @mlegner , I added some notes below so that you can pay more attention to the code in those areas.

crates/walrus-service/src/node/node_recovery.rs
halfprice: > >we just want to know the first event's epoch, which will be available after processing it

> This is unfortunately not true for blob extensions.

A nice thing about making latest_event_epoch() async and return only when it is available, is that it is actually eliminated the race condition that if the first few events are all blob extension events, blob sync may started without latest_event_epoch() populated. Now, the blob sync tasks just waits for latest_event_epoch() to be populated, which is much nicer I think.

crates/walrus-service/src/node.rs
```diff
@@ -1694,20 +1722,39 @@ impl StorageNode {
...
+                "node is currently recovering to epoch {recovering_epoch}, restarting \
+                node recovery to recover to the latest epoch {}",
+                event.epoch
+            );
+            self.process_shard_changes_in_new_epoch_and_start_node_recovery(
+                event_handle,
+                event,
+                shard_map_lock,
+            )
+            .await
```
liquid-helium, line 1739: Question: does this advance the persisted RecoveryInProgress(`recovering_epoch`)?
Is it possible that

The persisted `recovering_epoch` advances, while the processed events is lagging.
The server restarts and saw the old events, which is behind the persisted `recovering_epoch`.

Is it possible, and will it cause issues? I tried to figure it out, but won't finish reading all the code :/, just throwing the question here.

halfprice: Great question! This nuance is something that I pay most attention to when doing this PR. This situation is indeed possible.

Note that when a node is in `RecoveryInProgress(epoch)` status, it behaves the same as `Active`, except that it will start a background process to recover all the missing blobs certified up to `epoch`. Everything else will be the same as `Active`. This means that after RecoveryInProgress is moved forward a few epochs, say epoch 10, but persisted epoch is still 7, this will transition the node into RecoveryCatchUp mode. If highest persisted event epoch is 9, processing certified events may still initiate blob recovery task, however, for any blob ID, there can be only 1 task to recover the blob, so there won't be duplicated/wasted blob sync tasks.

crates/walrus-service/src/node/node_recovery.rs
halfprice: Checked that if the node is lagging and in RecoveryInProgress state, it'll make the node goes into RecoveryCatchUp mode, which is the same behavior as the removed code here.

crates/walrus-simtest/src/test_utils.rs
```diff
@@ -616,4 +616,49 @@ pub mod simtest_utils {
...
+
+    /// Simulates repeated node crash and restart with sim node id.
+    pub fn repeatedly_crash_target_node(
+        target_node_id: sui_simulator::task::NodeId,
+        next_fail_triggered_clone: Arc<Mutex<Instant>>,
+        crash_end_time: Instant,
+        min_crash_duration_secs: u64,
+        max_crash_duration_secs: u64,
+        min_live_duration_secs: u64,
+        max_live_duration_secs: u64,
```
mlegner: Nit: Some or all of these could be moved to a `NodeCrashConfig` struct, then the code at callsites would document itself (instead of having to add comments for the respective parameter names).

crates/walrus-service/src/node.rs
```diff
@@ -1167,9 +1176,16 @@ impl StorageNode {
             return Ok(false);
         };
 
-        if self.inner.current_event_epoch() >= first_complete_epoch {
+        let Some(current_event_epoch) = self.inner.try_get_current_event_epoch() else {
+            // When current event epoch is not set, we are processing events before the first epoch
+            // change start event, which should be excluded from processing.
+            return Ok(false);
```
mlegner: Important: This should be `true`.
```suggestion
            return Ok(true);
```

crates/walrus-service/src/node/node_recovery.rs
mlegner: Suggestion: Can we just remove the check above (line 101) then?

crates/walrus-service/src/node.rs
```diff
@@ -2223,8 +2270,20 @@ impl StorageNodeInner {
         Ok(metadata)
     }
 
-    fn current_event_epoch(&self) -> Epoch {
-        self.latest_event_epoch.load(Ordering::SeqCst)
+    /// Returns the latest event epoch.
+    /// If the storage node hasn't processed the first event yet, this function will block until
+    /// the first event is processed, so that the current event epoch is set.
+    async fn current_event_epoch(&self) -> Result<Epoch, watch::error::RecvError> {
```
mlegner, line 2274: Question: What could cause the `RecvError` here? Can this actually happen?

crates/walrus-service/src/node/blob_sync.rs
mlegner: Remark/question: I think this makes sense to me. But is there a concrete need to make this change?

mlegner (approved):
Thanks so much for these important improvements! One important comment, otherwise just a few questions.

crates/walrus-service/src/node/blob_sync.rs
halfprice: Yes, the issue without this change is that, say RecoveryInProgress(epoch=9) is happening, and started all the blob syncs, and then we transition into epoch 10. All the blob syncs initiated prior are still running. In the node recovery, we loop through all the certified blobs, find ones that needs recovery, and start the recovery. Without this change, the task handle returned by `start_sync` will finish immediately. This causes the loop to finish very quickly. However, when the loop runs again to see we indeed finished recovering all the blobs, they still see blobs that need recover. So we ended up running the recovery loop over and over. The motivation for this change is the observed behavior in simtest.

crates/walrus-service/src/node.rs
halfprice: Thanks @mlegner ! That's a great catch, and sorry for the careless mistake. I actually ran test_recovery_with_incomplete_history 100 time yesterday and didn't see any error.

Looking at the code logic, now I see that processing events before the first complete epoch is probably won't cause any wrong internal state, is that correct? `mark_event_complete()` is a no op for earlier events,`update_blob_info` will skip event before the first complete epoch, and `process_certified_event` won't start a blob sync since the blob id doesn't exist in the blob info table. Is that correct? I just wonder if we need to test more in situation like this.

crates/walrus-simtest/src/test_utils.rs
halfprice: Great suggestion! Done.

crates/walrus-service/src/node/node_recovery.rs
halfprice: Yes, and if we want to do the check right before starting blob sync, the sliver check also needs to be moved. Let me tackle this in a follow up PR as part of WAL-864

crates/walrus-service/src/node.rs
halfprice: RecvError is thrown when one side of the channel is closed. I think this theoretically can only happen when a node is turning down. However, I'd rather we populate the error instead of panicing the node.

crates/walrus-service/src/node.rs
mlegner: Ah, that's a very good observation. I believe you are right: Any events before the first complete epoch are anyway processed as no-ops. 

I'm not sure if it's better to remove this function completely to simplify the code or keep it as a safe-guard for any changes in the down-stream event processing. WDYT?

crates/walrus-service/src/node/blob_sync.rs
mlegner: Ah, that's a great catch! Thanks for the explanation! 🙏

PR #2294, opened 2025-07-07
fix(node): correctly update event cursor with incomplete history

mlegner:
## Description

When starting the node with incomplete history, the storage node cursor is repositioned to the `init_state`'s first event. However, the first events are for an earlier epoch than the first complete epoch and must be skipped. This was already implemented in #2193.

However, when skipping these events before the first complete epoch, we still need to update the "persisted" event progress, which was not done correctly before.

## Test plan

Modified an existing simtest and checked that it fails without this fix 
but succeeds with it.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1153,35 +1153,36 @@ impl StorageNode {
         bail!("event stream for blob events stopped")
     }
 
-    /// Returns `false` if the node is recovering with incomplete history and the event is not
-    /// relevant to the recovery.
-    ///
-    /// An event is relevant, if the current event epoch is greater than or equal to the first
-    /// complete epoch of the recovery.
-    fn should_process_event(&self, stream_element: &PositionedStreamEvent) -> anyhow::Result<bool> {
+    /// Returns `true` if the node is recovering with incomplete history and the event is not
+    /// relevant to the recovery because it is before the first complete epoch of the recovery.
+    fn should_skip_event_in_incomplete_history_before_first_complete_epoch(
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1243,6 +1241,15 @@ impl StorageNode {
             stream_element.element.event_id(),
             self.inner.clone(),
         );
+        if !self.should_process_event(&stream_element)? {
```
halfprice: Nit: can we change the function name to something that is more specific to the `incomplete history`? There are multiple places where events may be filtered for processing (e.g., node is recovery mode skips blob sync part of the event processing), so maybe something like `should_skip_events_not_part_of_event_history` or something similar to avoid confusion.

mlegner:
Thanks for the review, @halfprice!

> [...] when starting the node with incomplete history, the storage node cursor is repositioned to the first init_state's first event. However, that event must be before the epoch that we want to start processing events, and therefore, we must mark them as completed, is that correct?

Yes, that is correct. I'll add some more details to the PR description.

> If we can somehow reposition the storage node's event cursor to the first events the in the first complete epoch, this problem won't exist?

Yes, that is correct as well. I don't know of a simpler way of doing that than just going through the events. Theoretically we could add additional data to event blobs, but that would be a bigger change.

crates/walrus-service/src/node.rs
mlegner: Good idea, done in 2d68c2f2846d4c5eefe31d99455399b1894fd8bc.

PR #2296, opened 2025-07-07
chore: sending checkpoint to download checking if the channel is full

halfprice:
## Description

When the channel is full, sending may be blocked, and at the same time, we cannot consume any checkpoints, so the loop
may be stuck forever.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/checkpoint-downloader/src/config.rs
+++ b/crates/checkpoint-downloader/src/config.rs
@@ -134,11 +134,11 @@ impl Default for AdaptiveDownloaderConfig {
 
 impl AdaptiveDownloaderConfig {
     pub(crate) fn message_queue_size(&self) -> usize {
-        self.max_workers * self.channel_config.work_queue_buffer_factor
+        self.max_workers * self.channel_config.work_queue_buffer_factor * 2
     }
 
     pub(crate) fn checkpoint_queue_size(&self) -> usize {
-        self.max_workers * self.channel_config.work_queue_buffer_factor
+        self.max_workers * self.channel_config.work_queue_buffer_factor * 2
     }
 
     pub(crate) fn result_queue_size(&self) -> usize {
--- a/crates/checkpoint-downloader/src/downloader.rs
+++ b/crates/checkpoint-downloader/src/downloader.rs
@@ -382,82 +382,93 @@ impl ParallelCheckpointDownloaderInner {
 
     /// Starts downloading checkpoints from the given sequence number.
     pub fn start(
-        mut self,
+        self,
         sequence_number: CheckpointSequenceNumber,
     ) -> mpsc::Receiver<CheckpointEntry> {
         let (result_tx, result_rx) = mpsc::channel(self.config.result_queue_size());
         let message_sender = self.message_sender.clone();
 
         tokio::task::spawn(async move {
-            let mut sequence_number = sequence_number;
-            let mut in_flight = BTreeSet::new();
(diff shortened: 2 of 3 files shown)


crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -379,70 +382,93 @@ impl ParallelCheckpointDownloaderInner {
...
+        let mut sequence_number = sequence_number;
+        let mut in_flight = BTreeSet::new();
+        let mut pending = BTreeMap::new();
+        let mut next_expected = sequence_number;
+
+        while !self.cancellation_token.is_cancelled() {
+            let num_workers = self.worker_count.load(std::sync::atomic::Ordering::SeqCst);
+            let max_in_flight_requests =
+                num_workers * self.config.channel_config.work_queue_buffer_factor;
+            while !message_sender.is_full() && in_flight.len() < max_in_flight_requests {
```
halfprice, line 415: The key of the fix is adding `!message_sender.is_full()` here. I was planning to do some refactor around the cancellation token to make it more effective (as opposite to only cancel the task between checkpoints). But later found some difficulties to make it work. I preserved `run_download_loop`, but the while loop is almost identical to the old code (besides adding the metrics).

crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -231,7 +234,7 @@ impl ParallelCheckpointDownloaderInner {
                         .message_sender
                         .send(WorkerMessage::Shutdown)
                         .await?;
-                    let new_count = worker_count.fetch_sub(1, std::sync::atomic::Ordering::Relaxed);
+                    let new_count = worker_count.fetch_sub(1, std::sync::atomic::Ordering::SeqCst);
```
halfprice, line 237: I think the culprit of adding 9 (11 total workers) is the relaxed memory ordering. I changed this to the sequential consistency model for easy reasoning. We can revisit if this becomes a performance bottleneck (the work_count is not update very frequently).

crates/checkpoint-downloader/src/config.rs
```diff
@@ -134,11 +134,11 @@ impl Default for AdaptiveDownloaderConfig {
...
 impl AdaptiveDownloaderConfig {
     pub(crate) fn message_queue_size(&self) -> usize {
-        self.max_workers * self.channel_config.work_queue_buffer_factor
+        self.max_workers * self.channel_config.work_queue_buffer_factor * 2
     }
 
     pub(crate) fn checkpoint_queue_size(&self) -> usize {
-        self.max_workers * self.channel_config.work_queue_buffer_factor
+        self.max_workers * self.channel_config.work_queue_buffer_factor * 2
     }
```
mlegner, line 142: Question: What is the motivation for these changes?

mlegner (approved):
Thanks for the fix, @halfprice. I agree with your assessment and fix.

crates/checkpoint-downloader/src/config.rs
halfprice: Same as the fix: the channel should have some buffer and shouldn't ever be smaller than application logic limit, otherwise, the unexpected issue like this happens. So essentially, application level sets the soft limit, and underlying channel sets the upper limit, and lower limit < upper limit.

PR #2297, opened 2025-07-08
chore: move Restore command out of LocalAdmin.

liquid-helium:
## Description

Move `Restore` out of `LocalAdmin` since it doesn't require communication with a running storage node.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -1119,8 +1119,8 @@ mod commands {
 
         let target_checkpoint = checkpoint_id.map_or("latest".to_string(), |id| id.to_string());
         println!(
-            "Restore from {target_checkpoint} successfully. The node must be restarted for changes \
-            to take effect."
+            "Restored from {target_checkpoint} successfully. The node must be restarted for \
+            changes to take effect."
         );
 
         Ok(())
--- a/crates/walrus-service/src/node/db_checkpoint.rs
+++ b/crates/walrus-service/src/node/db_checkpoint.rs
@@ -284,7 +284,7 @@ impl DbCheckpointManager {
         db: Arc<RocksDB>,
         config: DbCheckpointConfig,
     ) -> Result<Self, DbCheckpointError> {
-        tracing::info!("DbCheckpointManager starting with config: {:?}", config);
+        tracing::info!(?config, "DbCheckpointManager starting...");
         if let Some(db_checkpoint_dir) = config.db_checkpoint_dir.as_ref() {
             create_dir_all(db_checkpoint_dir).map_err(|e| DbCheckpointError::Other(e.into()))?;
         }


crates/walrus-service/bin/node.rs
```diff
@@ -481,6 +463,25 @@ struct CatchupArgs {
     rpc_fallback_config_args: Option<RpcFallbackConfigArgs>,
 }
 
+#[derive(Debug, Clone, clap::Args)]
+struct RestoreArgs {
+    /// The path where the checkpoint is stored. Note, it will not be defaulted to the
```
liquid-helium, line 468: @halfprice I missed one of the comments in a previous PR, here I fixed it. Ty

crates/walrus-service/bin/node.rs
```diff
@@ -1097,6 +1100,32 @@ mod commands {
...
+            &db_checkpoint_path,
+            &db_path,
+            wal_path.as_deref(),
+            checkpoint_id,
+        )
+        .await?;
+
+        let target_checkpoint = checkpoint_id.map_or("latest".to_string(), |id| id.to_string());
+        println!(
+            "Restore from {target_checkpoint} successfully. The node must be restarted for changes \
```
wbbradley:
```suggestion
            "Restored from {target_checkpoint} successfully. The node must be restarted for changes \
```

crates/walrus-service/src/node/db_checkpoint.rs
```diff
@@ -284,6 +284,7 @@ impl DbCheckpointManager {
         db: Arc<RocksDB>,
         config: DbCheckpointConfig,
     ) -> Result<Self, DbCheckpointError> {
+        tracing::info!("DbCheckpointManager starting with config: {:?}", config);
```
wbbradley:
```suggestion
        tracing::info!(?config, "DbCheckpointManager starting...");
```

crates/walrus-service/bin/node.rs
```diff
@@ -557,6 +558,8 @@ fn main() -> anyhow::Result<()> {
 
         Commands::Catchup(catchup_args) => commands::catchup(catchup_args)?,
 
+        Commands::Restore(restore_args) => commands::restore(restore_args)?,
```
halfprice, line 561: Suggestion: how about we create a first level command similar to db-tool? Something like db-op and then put restore in it? I feel restore can also just go under db-tool

mlegner: Just want to mention that the `db-tool` command is hidden, so the `restore` command would be difficult to find. And I would avoid making the whole `db-tool` command visible as it has some slightly dangerous subcommands.

crates/walrus-service/bin/node.rs
liquid-helium: Done.

crates/walrus-service/src/node/db_checkpoint.rs
liquid-helium: Done.

crates/walrus-service/bin/node.rs
liquid-helium: Keep it as, we can move it around in the future if there is a better place.

PR #2300, opened 2025-07-08
feat: add upload relay to walrus to `walrus store` and `walrus store-quilt`

wbbradley:
## Description

Adds the possiblity to specify the `--upload-relay ...` to `walrus store` and `walrus store-quilt`.
The upload relay can be used to store multiple blobs in parallel (`walrus store`) and as quilts (`walrus store-quilt`).
This implementation allows to reuse all the nice storage optimizations in the Client SDK with the upload relay as well.

At a high level, there is, however, a design trade off in this implementation: To allow to multiple blob stores, and to avoid modifying the `walrus-sui` SDK too deeply, the PR _does not bundle_ the `tip` payements that are required for by the fanout with the registration. This simplifies the implementation and allows parallel stores, but has the side effect of paying for multiple gas fees.

This can maybe addressed in a followup, but a bit more design is required. The cleanest option would be to allow paying for multiple tips in a single ptb, and then implementing some more complex tracking on the update relay, and finally ripping out some code in the `walrus-sui` crate to allow more flexible certification flows.

At a low level, the following was implemented in this PR:

- Refactored the write path for the walrus client and compacted all arguments to a `StoreArg` object, that can be extended as needed;
- When the store operation is initiated, if the user specified an `--upload-relay` URL, the `StoreArg` gets augmented with a `UploadRelayClient`, that then is used down the stack when it comes the time to call `send_blob_data_and_get_certificate`.
- A lot of the old upload relay client code was removed, as it is now obsolete.

## Test plan

Manual testing, Pipeline.

```
cargo build --release; RUST_LOG=walrus_sdk=debug ./target/release/walrus --context testnet store--upload-relay https://upload-relay.testnet.walrus.space --epochs 1 --paths <UNSTORED_FILE>
```


crates/walrus-sdk/src/upload_relay.rs
```diff
@@ -13,10 +13,12 @@ use serde::{Deserialize, Serialize};
...
 use crate::core::{BlobId, messages::ConfirmationCertificate};
 
-/// The route for the Walrus Upload Relay API.
+/// The route to upload blobs using the relay.
 pub const BLOB_UPLOAD_RELAY_ROUTE: &str = "/v1/blob-upload-relay";
-/// The route for the Walrus Upload Relay Tip Config API.
+/// The route to fetch the update-relay's tip configuration.
 pub const TIP_CONFIG_ROUTE: &str = "/v1/tip-config";
+/// The route for the update-relay's OpenAPI docs.
+pub const API_DOCS: &str = "/v1/api";
```
wbbradley, line 23: nit: I had deliberately left this out because I didn't see that it would be needed anywhere in the sdk since I don't think the SDK is going to be hitting the API_DOCS.

crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -0,0 +1,196 @@
...
+    pub upload_relay_client: Option<UploadRelayClient>,
+}
+
+impl Default for StoreArgs {
+    fn default() -> Self {
+        Self {
+            encoding_type: DEFAULT_ENCODING,
+            epochs_ahead: 1,
+            store_optimizations: StoreOptimizations::all(),
+            persistence: BlobPersistence::Permanent,
```
liquid-helium: nit: add a TODO here, since we are planing to switch the default to Deletable.

crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -0,0 +1,196 @@
...
+    pub metrics: Option<Arc<ClientMetrics>>,
+    /// The optional upload relay client, that allows to store the blob via the relay.
+    pub upload_relay_client: Option<UploadRelayClient>,
+}
+
+impl Default for StoreArgs {
+    fn default() -> Self {
+        Self {
+            encoding_type: DEFAULT_ENCODING,
+            epochs_ahead: 1,
```
liquid-helium: nit: do we need this default? I think we are trying to be explicit on the epochs_ahead?

giac-mysten: yes, you are right. This struct is never externally constructed, and everywhere we initialize it from user input we change the epochs or use `new`. This default is mostly for tests, where there are many that just use `1` as the epoch.

I changed it a bit to make everything more explicit.

crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -0,0 +1,201 @@
...
+use super::{metrics::ClientMetrics, upload_relay_client::UploadRelayClient};
+use crate::{
+    client::upload_relay_client::UploadRelayClientError,
+    store_optimizations::StoreOptimizations,
+    upload_relay::tip_config::TipConfig,
+};
+
+/// Arguments for store operations that are frequently passed together.
+// NOTE: In the future, if the struct grows larger, we may need to consider using a builder.
+#[derive(Debug, Clone)]
```
liquid-helium, line 20: Praise, I like this StoreArgs, it is really clean!

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -639,6 +646,8 @@ impl ClientCommandRunner {
         persistence: BlobPersistence,
         post_store: PostStoreAction,
         encoding_type: Option<EncodingType>,
+        upload_relay: Option<Url>,
+        confirmation: UserConfirmation,
```
liquid-helium, line 650: nit: can we pass in a StoreArgs instead, so we can remove the #[allow(clippy::too_many_arguments)]

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1045,6 +1046,17 @@ pub struct CommonStoreOptions {
...
+    #[arg(long)]
+    #[serde(default)]
+    pub upload_relay: Option<Url>,
+    /// Whether to ask for user confirmations.
+    ///
+    /// If yes is specified, the client will not ask for confirmations and proceed with the
+    /// operation.
+    #[arg(long)]
+    #[serde(default)]
+    pub yes: bool,
```
liquid-helium: nit: I understand that 'yes' is a common practice to bypass confirmation, but if tip is the only one that requires confirmation, can we make it more explicit? e.g., --skip-tip-confirmation or --confirm-tip?

crates/walrus-service/src/client/cli/runner.rs
giac-mysten: mmm i would not do this for the moment. Unfortunately, resolving the `EpochArg` to a number of epochs ahead requires the sui client, and constructing the upload relay client as well. This means that passing the `StoreArgs` to `store` is not very clean (for example, a placeholder epoch must be set and then canged). Further, the `store_quilt` function would still have too many args. 

I think a bit more refactoring is needed for this to work.

crates/walrus-service/src/client/cli/args.rs
giac-mysten: good point!

PR #2302, opened 2025-07-09
fix(node): cancel blob syncs for expired blobs

mlegner:
## Description

Since the async event processing, it can happen that a new blob recovery is started after the epoch-change event for that blob's expiration epoch is processed. As a result, that blob sync is not cancelled by the task cancelling blob syncs for expired blobs. This blob sync will then remain until the next epoch change.

This commit adds expiration checks to the already existing blob sync task monitor, such that stale blob syncs are cancelled quickly.

It also makes the blob sync task monitor interval configurable, keeping the current value of 60 seconds as a default.

## Test plan

TBD.


crates/walrus-service/src/node.rs
```diff
@@ -812,7 +814,8 @@ impl StorageNode {
                 self.inner.shut_down();
                 self.blob_sync_handler.cancel_all().await?;
             },
-            blob_sync_result = self.blob_sync_handler.spawn_task_monitor() => {
+            blob_sync_result = self.blob_sync_handler.spawn_task_monitor(
+            ) => {
```
halfprice: Nit: this change feels weird

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -74,76 +75,86 @@ impl BlobSyncHandler {
...
     }
 
-    // Periodically checks the status of all in-progress blob syncs and return any that panicked.
+    /// Periodically checks the status of all in-progress blob syncs, cancelling those that are no
+    /// longer certified and returning any that panicked.
     pub fn spawn_task_monitor(&self) -> JoinHandle<()> {
         let blob_syncs = self.blob_syncs_in_progress.clone();
+        let node = self.node.clone();
+        let monitor_interval = if cfg!(test) {
+            Duration::from_secs(5)
```
halfprice: Suggestion: since monitor_interval is part of the node config, maybe set it here https://github.com/MystenLabs/walrus/blob/429d070f72d9fb20e7a8a20fc0dd757938f66bae/crates/walrus-service/src/test_utils.rs#L1090 so that all simtest related configs are in one place.

crates/walrus-service/src/node.rs
```diff
@@ -2474,6 +2477,11 @@ impl StorageNodeInner {
             .is_some_and(|blob_info| blob_info.is_certified(self.current_epoch())))
     }
 
+    /// Returns true if the blob is currently not certified.
+    fn is_blob_not_certified(&self, blob_id: &BlobId) -> bool {
+        !self.is_blob_certified(blob_id).is_ok_and(Not::not)
```
halfprice: This is a bit hard to parse :) there is a `!` and also `Not::not`.

I feel the `!` at the beginning should be removed?

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -74,76 +75,86 @@ impl BlobSyncHandler {
...
+                    let Err(error) = handle.await else {
+                        // Normal completion. This should not happen.
+                        // The blob sync handler should have removed the handle after the sync
+                        // finished. So technically the sync handle should not be in the map.
+                        tracing::warn!(
+                            walrus.blob_id = %blob_id,
+                            "blob sync finished with success, but still exists in \
+                            BlobSyncHandler"
+                        );
+                        continue;
```
halfprice, line 142: Question: will the cancelled blob sync handles appear here?

crates/walrus-service/src/node.rs
mlegner: Whoops, indeed. I confused myself there... I replaced this with an easier to understand if statement.

crates/walrus-service/src/node/blob_sync.rs
mlegner: Good question. To make reasoning easier, I now added a `continue` after the `handle.cancel()` above, so we don't actually process these syncs further in the same monitor run.

mlegner:
Rebased on #2334 to see if that fixes the simtest failure.

mlegner:
@halfprice This is now ready for another (final?) review. The simtests succeed after rebasing onto #2334.

crates/walrus-service/src/node.rs
```diff
@@ -2495,6 +2495,14 @@ impl StorageNodeInner {
             .is_some_and(|blob_info| blob_info.is_certified(self.current_epoch())))
     }
 
+    /// Returns true if the blob is currently not certified.
+    fn is_blob_not_certified(&self, blob_id: &BlobId) -> bool {
```
halfprice, line 2500: Just want to add a note that err reading is_blob_certified will cause this function to return false, which won't cancel blob sync.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -74,76 +75,83 @@ impl BlobSyncHandler {
...
                             }
                         }
+
+                        // Cancel the sync if the blob is no longer certified.
+                        if node.is_blob_not_certified(blob_id) {
+                            tracing::debug!(
+                                walrus.blob_id = %blob_id,
+                                "cancelling blob sync because blob is no longer certified"
+                            );
+                            handle.cancel();
```
halfprice, line 121: Remark: I think we also need to mark event complete in this case, if handle.cancel() returns an EventHandle.

For the prod issue we investigated that requires this fix, there is no event handle since the blob sync is initiated via node recovery, so I think this can be fixed in a follow up PR and maybe also cherry-pick to 1.29.

crates/walrus-service/src/node/blob_sync.rs
halfprice: LGTM!

crates/walrus-service/src/node.rs
halfprice: LGTM!

PR #2308, opened 2025-07-10
ci: ensure WFs involving Sui transactions are not concurrent

mlegner:
Otherwise, we could end up with conflicting transactions and locked objects.

### DIFF
--- a/.github/workflows/publish-docs.yaml
+++ b/.github/workflows/publish-docs.yaml
@@ -65,6 +65,9 @@ jobs:
       RUST_LOG: info
       EPOCHS: 10
       BUILD_DIR: docs/build/html
+    concurrency:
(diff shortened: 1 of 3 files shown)


mlegner:
Actually, I just realized, we need to be a bit careful when running this after merged PRs as well. Let me improve this.

giac-mysten (approved):

mlegner:
@giac-mysten Thanks for the review! I actually changed the approach: Instead of scheduling the workflows at different times, I define a concurrency group that definitely prevents these jobs running concurrently, see [these docs](https://docs.github.com/en/actions/how-tos/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs). Could you take another look?

PR #2314, opened 2025-07-11
fix: [1] epoch sync done should check latest epoch [2] bypassing consistency check if node is reprocessing events

halfprice:
## Description

This PR fixes two issues exposed by `simtest_test_repeated_node_crash`.

- When sending epoch_sync_done, it's possible that the epoch change has advanced. In this case, epoch sync done will retry indefinitely, blocking the process_epoch_change_start event from finishing, and therefore blocking the next epoch change event. The fix is that 
- When reprocessing events, the blob info table might have been affected by "future" events. Therefore, we should bypass consistency check if the node is reprocessing events. The first time processing events should always produce consistent blob info digest.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/contract_service.rs
+++ b/crates/walrus-service/src/node/contract_service.rs
@@ -22,7 +22,7 @@ use walrus_sui::{
         BlobObjectMetadata,
         CoinType,
         FixedSystemParameters,
-        ReadClient as _,
+        ReadClient,
         SuiClientError,
         SuiClientMetricSet,
         SuiContractClient,
@@ -31,6 +31,7 @@ use walrus_sui::{
     types::{
         StorageNodeCap,
         UpdatePublicKeyParams,
+        move_errors::{MoveExecutionError, StakingInnerError},
         move_structs::{EpochState, EventBlob},
     },
 };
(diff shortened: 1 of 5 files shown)


crates/walrus-service/src/node/contract_service.rs
```diff
@@ -513,6 +514,27 @@ impl SystemContractService for SuiSystemContractService {
                     tracing::debug!(walrus.epoch = epoch, "repeatedly submitted epoch_sync_done");
                     Some(())
                 }
+                Err(SuiClientError::TransactionExecutionError(
+                    MoveExecutionError::StakingInner(StakingInnerError::EInvalidSyncEpoch(error)),
+                )) => match self.read_client.current_epoch().await {
+                    Ok(latest_epoch_on_chain) => {
+                        if latest_epoch_on_chain > epoch {
+                            tracing::info!(
+                                ?error,
```
mlegner: Question: Do we need the error in this log entry? If so, maybe using the display instead of debug print would be better as it is a somewhat expected case.
```suggestion
                                %error,
```

halfprice: Good question. We probably don't need it right now as the error doesn't contain extra information. However, I'm leaning towards keeping it in case in the future, more information may be added in the error.

Changed to display.

PR #2318, opened 2025-07-11
fix: use original package ID to filter events

karlwuest:
## Description

Fixes a bug in the event processor that causes nodes to miss events if they restart after a package upgrade

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/event/event_processor/processor.rs
+++ b/crates/walrus-service/src/event/event_processor/processor.rs
@@ -43,8 +43,8 @@ pub struct EventProcessor {
     pub stores: EventProcessorStores,
     /// Event polling interval.
     pub event_polling_interval: Duration,
-    /// The address of the Walrus system package.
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -113,13 +113,13 @@ impl EventProcessor {
         let checkpoint_processor = CheckpointProcessor::new(
             stores.clone(),
             package_store.clone(),
-            system_config.system_pkg_id,
+            original_system_package_id,
```
karlwuest, line 116: This line is the actual fix.

mlegner (approved):
Thank you for the fix, @karlwuest! 🙏

halfprice (approved):
🙏🙏🙏🙏

PR #2328, opened 2025-07-15
feat: schedule call to process subsidies

karlwuest:
## Description

Schedules a call to process the subsidies on storage nodes, to be called shortly before epoch change.

## Test plan

Added test in e2e-tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/epoch_change_driver.rs
+++ b/crates/walrus-service/src/node/epoch_change_driver.rs
@@ -31,6 +31,8 @@ const MAX_SCHEDULE_JITTER: Duration = Duration::from_secs(180);
 const MIN_BACKOFF: Duration = Duration::from_secs(5);
 /// The maximum exponential backoff when performing retries.
 const MAX_BACKOFF: Duration = Duration::from_secs(300);
+/// The maximum amount of time before the epoch change that process subsidies is scheduled.
+const MAX_SUBSIDIES_TIME_BEFORE_EPOCH_CHANGE: Duration = Duration::from_secs(300);
 
 /// Function returning the current time in Utc.
 type UtcNowFn = Arc<dyn Fn() -> DateTime<Utc> + Send + Sync>;
@@ -74,6 +76,7 @@ impl EpochChangeDriver {
                 waker: None,
                 end_voting_future: None,
                 epoch_change_start_future: None,
+                subsidies_future: None,
             }))),
             system_parameters,
             contract_service,
@@ -121,7 +124,10 @@ impl EpochChangeDriver {
         match state {
             EpochState::EpochChangeSync(_) => (),
             EpochState::EpochChangeDone(_) => self.schedule_voting_end(next_epoch),
-            EpochState::NextParamsSelected(_) => self.schedule_initiate_epoch_change(next_epoch),
+            EpochState::NextParamsSelected(_) => {
+                self.schedule_initiate_epoch_change(next_epoch);
+                self.schedule_process_subsidies();
+            }
         }
 
         Ok(())
@@ -239,6 +245,45 @@ impl EpochChangeDriver {
         }
     }
 
+    /// Schedules a call to process subsidies for the next epoch that will be dispatched by
+    /// [`run()`][Self::run] if subsidies are configured in the contract service.
+    ///
+    /// Should be called after [`NextParamsSelected`] events. Subsequent calls to this
+    /// method cancel any earlier scheduled calls.
+    #[tracing::instrument(skip_all)]
+    pub fn schedule_process_subsidies(&self) {
+        if !self.contract_service.is_subsidies_object_configured() {
+            tracing::debug!("subsidies are not configured in the contract");
+            return;
+        }
+
+        let mut inner = self.inner.lock().expect("thread did not panic with lock");
+
+        if inner.subsidies_future.is_some() {
+            tracing::debug!("replacing subsidies future");
+            inner.subsidies_future = None;
+        }
+
+        let subsidies_future = ScheduledEpochOperation::new(
+            SubsidiesOperation {
+                time_before_epoch_change: (self.system_parameters.epoch_duration / 10)
+                    .min(MAX_SUBSIDIES_TIME_BEFORE_EPOCH_CHANGE),
+                system_params: self.system_parameters,
+                time_fn: self.utc_now_fn.clone(),
+            },
+            self.system_parameters.epoch_duration,
+            self.contract_service.clone(),
+            self.utc_now_fn.clone(),
+            &mut inner.rng,
+        )
+        .wait_then_call();
+
+        inner.subsidies_future = Some(subsidies_future.boxed());
+
+        tracing::debug!("scheduled process subsidies");
+        inner.wake();
+    }
+
     #[cfg(test)]
     /// Returns true if there is currently a scheduled call to end voting.
     pub fn is_voting_end_scheduled(&self) -> bool {
@@ -274,6 +319,8 @@ struct EpochChangeDriverInner<'pin> {
     end_voting_future: Option<(Epoch, BoxFuture<'pin, ()>)>,
     /// An optional future that will start epoch change for a given epoch.
     epoch_change_start_future: Option<(Epoch, BoxFuture<'pin, ()>)>,
+    /// An optional future that will process subsidies.
+    subsidies_future: Option<BoxFuture<'pin, ()>>,
 }
 
 impl EpochChangeDriverInner<'_> {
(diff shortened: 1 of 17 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -759,6 +769,14 @@ impl SuiContractClient {
         .await
     }
 
+    /// Call to initiate subsidy distribution.
+    ///
+    /// Requires the system-side walrus subsidy contract to be set.
+    pub async fn process_subsidies(&self) -> SuiClientResult<()> {
```
jpcsmith, line 784: Nit: Perhaps rename this to `distribute_subsidies`, as the current name is vague.

crates/walrus-sui/src/client.rs
```diff
@@ -904,6 +922,39 @@ impl SuiContractClient {
...
 
+    /// Creates a new walrus subsidies object (`walrus_subsidies::WalrusSubsidies`)
+    /// and returns the object ID and the admin cap ID.
+    pub async fn create_walrus_subsidies(
+        &self,
+        package_id: ObjectID,
+        system_subsidy_rate: u32,
+        base_subsidy: u64,
+        subsidy_per_shard: u64,
+    ) -> SuiClientResult<(ObjectID, ObjectID)> {
```
jpcsmith: Suggestion: Disambiguate what's what in the return type. You could wrap them in new types (e.g., `SubsidyAdminCap(ObjectId)` or a struct with two fields which name which is which.

crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -949,6 +950,76 @@ impl AssociatedContractStruct for Credits {
...
+    balance: u64,
+}
+
+/// A ring buffer holding the epoch balances for a continuous range of epochs.
+/// Used for deserialization of the WalrusSubsidiesInner.
+#[derive(Debug, PartialEq, Eq, Clone, Deserialize)]
+pub(crate) struct EpochBalanceRingBuffer {
+    current_index: u32,
+    ring_buffer: Vec<EpochBalance>,
+}
```
jpcsmith, line 1000: Question: What happens if the ring buffer isn't full? Don't you need both a head and tail?

crates/walrus-sui/src/types/move_structs.rs
```diff
@@ -949,6 +950,76 @@ impl AssociatedContractStruct for Credits {
...
+    /// The subsidy rate applied to the price paid to the system and added to the per-epoch rewards
+    /// pool of the system. Subsidy rates are expressed in basis points (1/100 of a percent). A
+    /// subsidy rate of 100 basis points means a 1% subsidy.
+    pub system_subsidy_rate: u32,
+    /// The balance of funds available in the subsidy pool.
+    pub subsidy_pool: u64,
+    /// The base subsidy (in FROST) paid directly per storage node per epoch.
+    pub base_subsidy: u64,
+    /// The additional subsidy (in FROST) paid to each storage node directly per shard.
+    pub subsidy_per_shard: u64,
```
jpcsmith: Nit: Would it be possible to strongly type the WAL, or at least rename the field so there's some indication of the unit?

crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -239,6 +245,45 @@ impl EpochChangeDriver {
         }
     }
 
+    /// Schedules a call to process subsidies for the next epoch that will be dispatched by
+    /// [`run()`][Self::run] if subsidies are configured in the contract service.
+    ///
+    /// Should be called after [`NextParamsSelected`] events. Subsequent calls to this
+    /// method cancel any earlier scheduled calls.
```
jpcsmith, line 252: Question: Do you need to already add the call to this method in node.rs?

crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -239,6 +245,45 @@ impl EpochChangeDriver {
...
+    #[tracing::instrument(skip_all)]
+    pub fn schedule_process_subsidies(&self) {
+        if !self.contract_service.is_subsidies_object_configured() {
+            tracing::debug!("subsidies are not configured in the contract");
```
jpcsmith:
```suggestion
            tracing::debug!("subsidies are not configured in the contract, skipping call to process subsidies");
```

crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -239,6 +245,45 @@ impl EpochChangeDriver {
...
+        let mut inner = self.inner.lock().expect("thread did not panic with lock");
+
+        if inner.subsidies_future.is_some() {
+            tracing::debug!("replacing subsidies future");
+            inner.subsidies_future = None;
+        }
+
+        let subsidies_future = ScheduledEpochOperation::new(
+            SubsidiesOperation {
+                time_before_epoch_change: (self.system_parameters.epoch_duration / 10)
```
jpcsmith: Question: Why the constant 10? Perhaps name it?

crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -588,6 +635,71 @@ impl EpochOperation for InitiateEpochChangeOperation {
...
+    fn time_until_ready(
+        &self,
+        utc_now: DateTime<Utc>,
+        current_epoch: Epoch,
+        state: EpochState,
+    ) -> Option<Duration> {
+        let current_epoch_started_at = match state {
+            EpochState::EpochChangeDone(_) | EpochState::EpochChangeSync(_) => {
+                todo!("How should we handle this? Panic? Or return time?")
+            }
```
jpcsmith: Question: We expect this to be called after the NextParamsSelected, so you could just log and return `None` to cancel the operation without a panic.

crates/walrus-service/src/node/epoch_change_driver.rs
```diff
@@ -588,6 +635,71 @@ impl EpochOperation for InitiateEpochChangeOperation {
...
+        Some(duration_until_subsidies_call)
+    }
+
+    async fn invoke(&self, contract: Arc<dyn SystemContractService>) -> Result<(), anyhow::Error> {
+        let last_subsidies_call = contract.last_walrus_subsidies_call().await?;
+        let current_time = (self.time_fn)();
+        // If the last subsidies call was less than time_before_epoch_change ago, return without
+        // calling them again. We use the same duration here to avoid having to configure too
+        // many parameters.
+        if current_time - self.time_before_epoch_change > last_subsidies_call {
```
jpcsmith: This seems like it should be
```suggestion
        if current_time - self.time_before_epoch_change < last_subsidies_call {
```
though in general I find the following easier to read
```suggestion
        if current_time < last_subsidies_call + self.time_before_epoch_change {
```

mlegner:
@halfprice @liquid-helium Also adding you as reviewers as discussed yesterday.

crates/walrus-service/src/node/epoch_change_driver.rs
karlwuest: That's an arbitrary choice similar to the jitter. It should only be relevant for tests. I can add a comment here, but I'm not sure if it makes sense to make it a named constant.

crates/walrus-service/src/node/epoch_change_driver.rs
karlwuest: 👍 I actually thought that I did that, but it seems I forgot.

crates/walrus-sui/src/client.rs
karlwuest: I used that initially but found that it sounded as if the caller distributes subsidies, so I opted for "process_subsidies", because that's also what the function is called in the contract. Happy to rename it if you feel strongly about it, though.

crates/walrus-sui/src/types/move_structs.rs
karlwuest: We actually only need this struct here for deserialization, s.t. we can read the `last_subsidized` timestamp. But I agree in general, but that should probably be part of a separate refactor (see [WAL-788](https://linear.app/mysten-labs/issue/WAL-788/use-wrapper-type-to-represent-walsui-throughout-the-rust-codebase)). I can add a comment here.

karlwuest: And regarding naming: I want to keep the field names the same as in the contract.

crates/walrus-sui/src/types/move_structs.rs
karlwuest: The ring buffer is always full and the size is fixed at creation time.

crates/walrus-service/src/node/epoch_change_driver.rs
karlwuest: This is called in the `schedule_relevant_calls_for_current_epoch` function, which I think should take care of that. But I'll confirm once I add a test.

crates/walrus-service/src/node/epoch_change_driver.rs
jpcsmith: It always makes sense to make it a named constant :)

PR #2334, opened 2025-07-15
fix: retry reading first event blob when encounter non-exist error

halfprice:
## Description

It's possible that 1 certified status from 1 storage node results in `read_blob_status()` returning certified status, but
event blob may still not be available for reads due to delay in propagating register/certify events to other storage nodes.
Similar to reading the blob status, also retrying reading first event blob if encountering non-existence error.

## Test plan

100 test_registered_node_update_protocol_key run passed, was failing with high probability before.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/event/event_blob_downloader.rs
+++ b/crates/walrus-service/src/event/event_blob_downloader.rs
@@ -130,11 +130,8 @@ impl EventBlobDownloader {
             event_blob_id
         );
 
-        // When reading the first certified event blob, it may be the case that the event blob is
-        // just certified, and the storage nodes may not know it yet. Therefore, upon encountering
-        // a nonexistent blob, we will retry and wait for the blob to be certified.
-        // For all the earlier blobs, they must have been seen by the storage nodes, since
-        // otherwise, no new certified event blobs would be created.
+        // Track if we are reading the first certified event blob. First event blob has special
+        // handling in `read_event_blob_status()`. See the comment there for more details.
         let mut reading_first_event_blob = true;
 
         loop {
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/event/event_blob_downloader.rs
```diff
@@ -259,4 +220,112 @@ impl EventBlobDownloader {
...
+                .await
+            {
+                Ok(blob_status) => blob_status,
+                Err(err) if matches!(err.kind(), ClientErrorKind::BlobIdDoesNotExist) => {
+                    BlobStatus::Nonexistent
+                }
+                Err(err) => return Err(err.into()),
+            };
+
+            if blob_status == BlobStatus::Nonexistent && reading_first_event_blob {
```
sadhansood, line 250: Should we try to consolidate the retry logic in the two functions, they seem almost similar. Alternatively, what if we skip reading the status and directly read the blob for all event blobs? One way to consolidate might be: 
```
async fn retry_and_read_blob<F, Fut, T>(
    &self,
    func: F,
    blob_id: BlobId,
) -> anyhow::Result<T>
where
    F: Fn() -> Fut,
    Fut: std::future::Future<Output = anyhow::Result<T>>,
{
```
and then we pass `read_blob` and `read_blob_status`. Or maybe we can use the generic `retry_with_backoff` we may already have in our utils.

crates/walrus-service/src/event/event_blob_downloader.rs
```diff
@@ -259,4 +220,112 @@ impl EventBlobDownloader {
...
+                tracing::debug!(
+                    "reading first certified event blob {} encountered a nonexistent \
+                blob, waiting for it to be certified",
+                    event_blob_id
+                );
+
+                if start_time.elapsed() > Duration::from_secs(30) {
+                    tracing::warn!(
```
sadhansood, line 258: It'd be nice to add a const for some of the magic numbers here: `FIRST_BLOB_CERTIFICATION_TIMEOUT`, etc

sadhansood (commented):
Great work identifying and fixing this race condition @halfprice!

crates/walrus-service/src/event/event_blob_downloader.rs
```diff
@@ -259,4 +220,112 @@ impl EventBlobDownloader {
...
+    /// a nonexistent blob, we will retry and wait for the blob to be certified.
+    /// For all the earlier blobs, they must have been seen by the storage nodes, since
+    /// otherwise, no new certified event blobs would be created.
+    async fn read_event_blob_status(
+        &self,
+        event_blob_id: BlobId,
+        reading_first_event_blob: bool,
+        start_time: Instant,
+    ) -> Result<BlobStatus> {
+        let blob_status = loop {
```
mlegner, line 237: Remark: Like @sadhansood, I find the double retries for (1) *getting the status* and (2) *reading* the first event blob a bit unsatisfying.

Besides skipping the status check as Sadhan suggested, there would be another option: We could remove or add a parameter to skip the [`verify_blob_status_event` in the `get_verified_blob_status` function](https://github.com/MystenLabs/walrus/blob/22481330ca235faef6351869e65e3d499c4c25fe/crates/walrus-sdk/src/client.rs#L2140). If we do that and get a certified status, we *know* that a quorum of nodes reported that status. Then we could remove the retries for reading the blob.

crates/walrus-service/src/event/event_blob_downloader.rs
mlegner: > what if we skip reading the status and directly read the blob for all event blobs

I think that might work, although this could cause a longer timeout on the first expired blob. Then again, that is only reached anyway for newly joining nodes *once*, so the performance there is pretty irrelevant.

mlegner (commented):
Thanks a lot for the investigation, analysis, and fix. I fully agree with your assessment, and AFAICT this solution is very effective at fixing the issue.

I think we should be able to simplify the logic a bit, though as discussed in some of the comments.

halfprice:
Thanks @mlegner @sadhansood ! Let's unblock others and I'll address your comments in a followup PR

PR #2335, opened 2025-07-16
feat: Move test_quorum_contract_upgrade test to simtest env

sadhansood:
## Description

This PR moves the test `test_quorum_contract_upgrade` to simtest environment, but I had to do a few changes to make it work. This is the main difference from tokio env about waiting for nodes to reach a particular epoch:
```
 // Wait for the nodes to reach the migration epoch.
let target_epoch = upgrade_epoch + 1;
loop {
    let mut all_nodes_ready = true;
    for node in &walrus_cluster.nodes {
        let current_epoch = simtest_utils::get_current_epoch_from_node(node).await;
        if current_epoch < target_epoch {
            all_nodes_ready = false;
            break;
        }
    }
    if all_nodes_ready {
        break;
    }
    tokio::time::sleep(Duration::from_secs(5)).await;
}
```
And then I also added restart of all nodes after package upgrade as that is what actually changes their view of the current package id in event processor. To ensure things work after restart, I then stored a blob to ensure it works.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-simtest/tests/simtest.rs
+++ b/crates/walrus-simtest/tests/simtest.rs
@@ -16,7 +16,7 @@ mod tests {
         time::{Duration, Instant},
     };
 
-    use rand::{Rng, SeedableRng};
+    use rand::{Rng, SeedableRng, seq::SliceRandom};
     use sui_macros::{
         clear_fail_point,
         register_fail_point,
@@ -25,9 +25,14 @@ mod tests {
     };
     use sui_protocol_config::ProtocolConfig;
     use sui_simulator::configs::{env_config, uniform_latency_ms};
+    use tempfile::TempDir;
     use tokio::sync::RwLock;
-    use walrus_core::EpochCount;
+    use walrus_core::{DEFAULT_ENCODING, EncodingType, EpochCount};
     use walrus_proc_macros::walrus_simtest;
+    use walrus_sdk::{
+        client::{Client, StoreArgs},
+        store_optimizations::StoreOptimizations,
+    };
     use walrus_service::{
         client::ClientCommunicationConfig,
         event::event_processor::config::EventProcessorConfig,
@@ -42,7 +47,13 @@ mod tests {
         repeatedly_crash_target_node,
     };
     use walrus_storage_node_client::api::ShardStatus;
-    use walrus_sui::client::ReadClient;
+    use walrus_sui::{
+        client::{BlobPersistence, PostStoreAction, ReadClient, SuiContractClient, UpgradeType},
+        system_setup::copy_recursively,
+        test_utils::system_setup::{development_contract_dir, testnet_contract_dir},
+        types::move_structs::EventBlob,
+    };
+    use walrus_test_utils::WithTempDir;
 
     /// Returns a simulator configuration that adds random network latency between nodes.
     ///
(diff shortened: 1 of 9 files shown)


crates/walrus-simtest/tests/simtest.rs
```diff
@@ -880,4 +886,182 @@ mod tests {
...
+                if current_epoch < target_epoch {
+                    all_nodes_ready = false;
+                    break;
+                }
+            }
+            if all_nodes_ready {
+                break;
+            }
+            tokio::time::sleep(Duration::from_secs(5)).await;
+        }
```
halfprice: Nit: there is a [wait_for_nodes_to_reach_epoch](https://github.com/MystenLabs/walrus/blob/ab5147f40c896f626989e19874c5fc5d46291f56/crates/walrus-simtest/src/test_utils.rs#L550) helper function, can we use it here?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -880,4 +886,182 @@ mod tests {
...
+                .read_client()
+                .system_object_version()
+                .await?,
+            previous_version + 1
+        );
+
+        // Restart all nodes
+        for i in 0..walrus_cluster.nodes.len() {
+            restart_node(&mut walrus_cluster, i).await;
+        }
```
halfprice: Nit: can we use [restart_nodes_with_checkpoints](https://github.com/MystenLabs/walrus/blob/ab5147f40c896f626989e19874c5fc5d46291f56/crates/walrus-simtest/src/test_utils.rs#L589) here? Maybe a bit more explicit that test_cluster::E2eTestSetupBuilder::new().with_num_checnpoints_per_blob(20)...

And then use `20` to call restart_nodes_with_checkpoints

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -880,4 +886,182 @@ mod tests {
...
+            std::fs::copy(
+                deploy_dir.path().join(contract).join("Move.lock"),
+                upgrade_dir.path().join(contract).join("Move.lock"),
+            )?;
+        }
+
+        // Change the version in the contracts
+        let walrus_package_path = upgrade_dir.path().join("walrus");
+
+        let upgrade_epoch = client.as_ref().sui_client().current_epoch().await?;
```
halfprice: Question: what is changed in the new contract? In the test, do we generate a new package id?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -880,4 +886,182 @@ mod tests {
...
+                    node_id,
+                    walrus_package_path.clone(),
+                )
+                .await?;
+        }
+
+        tracing::info!("voted for upgrade");
+
+        // Commit the upgrade
+        let new_package_id = loop {
```
karlwuest, line 995: Question/Remark: Why is this necessary? I assume that you saw this failing without the loop, but do you have any insight as to why? It also seems that after at most 4 iterations, this should always fail, because we are then progressing to the next epoch, and therefore all votes are no longer valid.

Suggestion: Add a comment to explain the loop here.

crates/walrus-simtest/tests/simtest.rs
karlwuest: The original test used that function @halfprice , so I assume that the change is necessary.

Question: @sadhansood can you explain a bit more why this is needed here? We actually don't need all nodes to be at the new epoch yet, it is only necessary that the new epoch has been started in the contract. Does that affect whether it's needed?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -880,4 +886,182 @@ mod tests {
...
+        assert_eq!(
+            client
+                .as_ref()
+                .sui_client()
+                .read_client()
+                .system_object_version()
+                .await?,
+            previous_version + 1
+        );
+
```
karlwuest, line 1057: Suggestion: Should we add a blob store operation before the restart as well, then we also cover the non-restart case?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -880,4 +886,182 @@ mod tests {
...
+        // Restart all nodes
+        for i in 0..walrus_cluster.nodes.len() {
+            restart_node(&mut walrus_cluster, i).await;
```
karlwuest: Suggestion: Only restart some of the nodes, so that if they go out of sync we see a problem with the quorum. The number of nodes could be chosen with the simtest pseudorandomness, so that we hopefully cover as many cases as possible.

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -880,4 +886,182 @@ mod tests {
...
+                &blobs,
+                DEFAULT_ENCODING,
+                1,
+                StoreOptimizations::none(),
+                BlobPersistence::Permanent,
+                PostStoreAction::Keep,
+                None,
+            )
+            .await?;
+
```
karlwuest, line 1101: Suggestion: We could additionally wait until an event blob has been produced to also cover that case. Especially combined with a random selection of restarting nodes (see comment above), this would probably be one of the best ways to spot diverging events.

crates/walrus-simtest/tests/simtest.rs
sadhansood: @karlwuest I replaced it because the previous non-simtest version of this function did not work here because it did not expect a simtest storage node handle, but @halfprice 's suggested `simtest_util::wait_for_nodes_to_reach_epoch` works just fine, so I am happy to replace it with this version.

crates/walrus-simtest/tests/simtest.rs
sadhansood: done

crates/walrus-simtest/tests/simtest.rs
sadhansood: yes, we do generate a new package digest and a new package id, i verified it.

crates/walrus-simtest/tests/simtest.rs
sadhansood: I see this failing with ENotEnoughVotes, i can just imagine that this is specific to simtest environment.

crates/walrus-simtest/tests/simtest.rs
sadhansood: done

crates/walrus-simtest/tests/simtest.rs
sadhansood: This is also done

sadhansood:
> Thanks a lot @sadhansood for adding the test! Did you verify that if the test would manifest the bug?
> 
> Also I'd request review from @karlwuest seems Karl used to have issue making the old test working with simtest.

Yes, replacing `original_package_id` with `system_ctx.package_id` will break this test. I;ve verified it.

crates/walrus-simtest/tests/simtest.rs
sadhansood: This is also done.

sadhansood: This is also done.

crates/walrus-simtest/tests/simtest.rs
karlwuest: Hmmm... that should not really happen at this point, because we await all the votes in the loop above and we wait for effects, so the transactions should all be finalized before we enter this loop. Not having enough votes at this point seems very strange and would indicate that the transactions get reordered on sui even though we have already received the effects.

@halfprice is it possible that the simtest environment introduces additional concurrency that could cause something like this?

karlwuest: @sadhansood don't block the PR on this, but let's maybe create an issue to investigate this.

crates/walrus-sui/src/client.rs
```diff
@@ -2023,12 +2046,34 @@ impl SuiContractClientInner {
...
     }
 
+    /// Vote as node `node_id` for upgrading the walrus package identified by
+    /// its pre-computed `package_digest`.
+    ///
+    /// This variant allows callers that already know the digest (for instance
+    /// when the package has been built elsewhere) to skip re-computing the
+    /// digest from the package path.
+    /// Returns the digest of the package (i.e. the same value supplied).
+    pub async fn vote_for_upgrade_with_digest(
```
karlwuest, line 2054: Nit: This now duplicates the voting code here and in the outer client. Could we instead just do a `vote_for_upgrade` where we pass an enum `PackagePathOrDigest`, or alternatively, keep the `vote_for_upgrade` and `vote_for_upgrade_with_digest` on the outer client, but then compute the digest there in the `vote_for_upgrade` and on the inner client, we can then just always pass the digest.

sadhansood: done.

PR #2338, opened 2025-07-16
docs: emphasize that the end epoch is exclusive

mlegner:
## Description

Also includes some other clarifications and improvements:
- Mention that the default blob persistence is going to change in the future.
- Reorganize the paragraphs describing how to define the blob lifetime.
- Emphasize that blobs can only be extended as long as they are not expired.
- Slightly reorganize the glossary.

## Test plan

Review and inspect preview.


docs/book/usage/client-cli.md
```diff
@@ -95,26 +95,36 @@ Storing one or multiple blobs on Walrus can be achieved through the following co
...
+   (e.g., "2024-03-20T15:00:00Z") or a more relaxed format (e.g., "2024-03-20 15:00:00") and ensures
+   the blob expires after that date if possible.
+1. The `--end-epoch <END_EPOCH>` option takes a specific end epoch for the blob.
+
+```admonish warning title="End epoch"
+A blob expires *at the beginning of its end epoch*. For example, a blob with end epoch `314` will
+become unavailable at the beginning of epoch `314`. One consequence of this is that when storing a
+blob with `--epochs 1` immediately before an epoch change, it will expire and become unavailable
+almost immediately. Note that blobs can only be [extended](#extending-the-lifetime-of-a-blob) as
+long as they are not expired.
```
liquid-helium: nit: Note that blobs can be [extended](#extending-the-lifetime-of-a-blob) only if they have not expired.

docs/book/usage/client-cli.md
```diff
@@ -95,26 +95,36 @@ Storing one or multiple blobs on Walrus can be achieved through the following co
...
 You can store a single file or multiple files, separated by spaces. Notably, this is compatible
 with glob patterns; for example, `walrus store *.png --epochs <EPOCHS>` will store all PNG files
 in the current directory.
 
-By default, the command will store the blob as a *permanent* blob. See the [section on deletable
-blobs](#reclaiming-space-via-deletable-blobs) for more details on deletable blobs. Also, by default
-an owned `Blob` object is created. It is possible to wrap this into a shared object, which can be
-funded and extended by anyone, see the [shared blobs section](#shared-blobs).
+By default, the command will store the blob as a *permanent* blob, although this is going to change
+in the near future. See the [section on deletable blobs](#reclaiming-space-via-deletable-blobs) for more
```
wbbradley, line 124: It would be nice to provide a specific date (or quarter, like Q4-2025) for when we'll switch this default.

wbbradley (approved):
Left a minor comment, feel free to address later or never.

docs/book/usage/client-cli.md
mlegner: Very good point. As just discussed offline, I'll leave this vague for now and make this more precise when we have more clarity about the timeline.

PR #2339, opened 2025-07-16
fix(client): bug in 404 computation

giac-mysten:
## Description

@mlegner found a [bug](https://github.com/MystenLabs/walrus/pull/866/files#r2210362724) in the way the client decides if a blob is non-existent.
This PR fixes it.

## Test plan

Local tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


mlegner (approved):
Thanks for the fix, @giac-mysten! 🙏

PR #2340, opened 2025-07-16
chore: Improve quilt test

liquid-helium:
## Description

Introduced `QuiltTestData` to generate random blobs with random identifiers and tags for testing quilts.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/test_utils.rs
+++ b/crates/walrus-core/src/test_utils.rs
@@ -57,22 +57,15 @@ pub struct QuiltTestData<'a> {
 }
 
 impl<'a> QuiltTestData<'a> {
-    /// Generates a new QuiltTestData from existing data.
-    pub fn new(
-        blob_data: &'a [&'a [u8]],
-        max_value_length: usize,
-        max_num_tags: usize,
-    ) -> Result<Self, QuiltError> {
-        let cow_data: Vec<Cow<'a, [u8]>> =
-            blob_data.iter().map(|&data| Cow::Borrowed(data)).collect();
-
-        let quilt_store_blobs_vec =
-            Self::generate_random_quilt_store_blobs(cow_data, max_value_length, max_num_tags)?;
-
-        Self::from_quilt_store_blobs(quilt_store_blobs_vec)
-    }
-
     /// Generates a new QuiltTestData with new owned random data.
+    ///
+    /// # Arguments
+    ///
+    /// * `num_blobs` - The number of blobs to be used to construct a quilt.
+    /// * `min_blob_size` - The minimum size of each blob.
+    /// * `max_blob_size` - The maximum size of each blob.
+    /// * `max_value_length` - The maximum length of each tag and identifier value.
+    /// * `max_num_tags` - The maximum number of tags per blob.
     pub fn new_owned(
         num_blobs: usize,
         min_blob_size: usize,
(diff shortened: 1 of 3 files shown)


crates/walrus-core/src/test_utils.rs
```diff
@@ -38,6 +47,135 @@ use crate::{
...
+pub struct QuiltTestData<'a> {
+    /// A map of QuiltStoreBlobs, keyed by their identifiers.
+    pub quilt_store_blobs: Vec<QuiltStoreBlob<'a>>,
+    /// A map of blob identifiers, keyed by tag keys and then tag values.
+    pub blob_identifiers_by_tag: HashMap<String, HashMap<String, HashSet<String>>>,
+}
+
+impl<'a> QuiltTestData<'a> {
+    /// Generates a new QuiltTestData from existing data.
+    pub fn new(
```
halfprice: Nit: this function does not seem to be used?

crates/walrus-core/src/test_utils.rs
```diff
@@ -38,6 +47,135 @@ use crate::{
...
+
+        Self::from_quilt_store_blobs(quilt_store_blobs_vec)
+    }
+
+    /// Generates a new QuiltTestData with new owned random data.
+    pub fn new_owned(
+        num_blobs: usize,
+        min_blob_size: usize,
+        max_blob_size: usize,
+        max_value_length: usize,
```
halfprice, line 73: Nit: maybe add doc to explain the input? IIUC, the blob is the entire walrus blob, not a quilt blob, correct?

crates/walrus-core/src/test_utils.rs
```diff
@@ -38,6 +47,135 @@ use crate::{
...
+    /// A random unique identifier is generated for each blob.
+    /// Random numbers of random tags are generated for each blob.
+    fn generate_random_quilt_store_blobs<'b>(
+        blob_data: Vec<Cow<'b, [u8]>>,
+        max_value_length: usize,
+        max_num_tags: usize,
+    ) -> Result<Vec<QuiltStoreBlob<'b>>, QuiltError> {
+        tracing::debug!("generating random quilt store blobs...");
+        let mut rng = rand::thread_rng();
+        let num_tags = rng.gen_range(0..=max_num_tags);
```
halfprice: Nit: I wonder if num_tags = 0 is more interesting and should have higher probability than other values.

Also, do we allow 0 byte blob inside a quilt?

halfprice (commented):
Very nice! LGTM and @mlegner probably should also take a look.

crates/walrus-test-utils/src/lib.rs
```diff
@@ -6,7 +6,7 @@
 use std::{fs, future::Future, path::Path};
 
 use anyhow::ensure;
-use rand::{Rng, RngCore, SeedableRng, rngs::StdRng, seq::SliceRandom};
+use rand::{Rng, RngCore, SeedableRng, rngs::StdRng, seq::SliceRandom, thread_rng};
```
mlegner: Nit: The convention is not to directly import functions, but their defining module. Here, we would import `rand::self` and then use the function as `rand::thread_rng()`.

crates/walrus-core/src/test_utils.rs
liquid-helium: I was planning to use that with some existing tests, so that we can directly work with &[u8] data.
Deleted for now, since it is not really used, will add it back when necessary.

crates/walrus-core/src/test_utils.rs
liquid-helium: Comments added.

crates/walrus-core/src/test_utils.rs
liquid-helium: Good point, change applied.

liquid-helium: 0 byte is allowed, I added an explicit test.

crates/walrus-test-utils/src/lib.rs
liquid-helium: Done.

PR #2341, opened 2025-07-16
chore: cleanup reading event blob codepath

halfprice:
## Description

Addressing comments in #2334 

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/event/event_blob_downloader.rs
+++ b/crates/walrus-service/src/event/event_blob_downloader.rs
@@ -10,8 +10,10 @@ use std::{
 
 use anyhow::Result;
 use walrus_core::{BlobId, Epoch};
-use walrus_sdk::{client::Client as WalrusClient, error::ClientErrorKind};
-use walrus_storage_node_client::api::BlobStatus;
+use walrus_sdk::{
+    client::Client as WalrusClient,
+    error::{ClientErrorKind, ClientResult},
+};
 use walrus_sui::{
     client::{ReadClient, SuiReadClient},
     types::move_structs::EventBlob,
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/event/event_blob_downloader.rs
```diff
@@ -140,33 +145,31 @@ impl EventBlobDownloader {
...
-                self.read_event_blob(
-                    event_blob_id,
-                    blob_status,
-                    reading_first_event_blob,
-                    start_time,
-                )
-                .await?
+                // Timer for waiting for the first certified event blob to be certified.
+                let result = self
+                    .read_event_blob(event_blob_id, reading_first_event_blob)
```
halfprice, line 153: @sadhansood @mlegner after reading the client code a bit more, I think just using the `read_event_blob()` is cleaner, as internally, it'll try to first get the status. So we do one retry here for both status and blob tracking. PTAL

crates/walrus-service/src/event/event_blob_downloader.rs
```diff
@@ -140,33 +145,31 @@ impl EventBlobDownloader {
...
             } else {
-                let start_time = Instant::now();
-                self.read_event_blob(
-                    event_blob_id,
-                    blob_status,
-                    reading_first_event_blob,
-                    start_time,
-                )
-                .await?
+                // Timer for waiting for the first certified event blob to be certified.
```
mlegner: Nit: This seems to be left over from the refactoring.
```suggestion
```
(Can optionally be moved inside the `read_event_blob` function.)

mlegner (approved):
Thanks for the refactoring, the code is very clean now. 🤩

crates/walrus-service/src/event/event_blob_downloader.rs
halfprice: Nice catch!

PR #2346, opened 2025-07-17
feat: Remove legacy event processor from the codebase

sadhansood:
## Description

This PR removes the lagacy event processor from the codebase, as the description says.

Closes WAL-405

## Test plan

Existing tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/test_utils.rs
+++ b/crates/walrus-service/src/test_utils.rs
@@ -208,8 +208,6 @@ pub trait StorageNodeHandleTrait {
 pub struct TestNodesConfig {
     /// The weights of the nodes in the cluster.
     pub node_weights: Vec<u16>,
-    /// Whether to use the legacy event processor.
-    pub use_legacy_event_processor: bool,
     /// Whether to disable the event blob writer.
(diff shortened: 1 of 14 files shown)


wbbradley (approved):
LGTM

crates/walrus-service/src/test_utils.rs
```diff
@@ -228,8 +226,6 @@ impl Default for TestNodesConfig {
     fn default() -> Self {
         Self {
             node_weights: vec![1, 2, 3, 3, 4],
-            // TODO(WAL-405): change default to checkpoint-based event processor
-            use_legacy_event_processor: true,
```
mlegner, line 228: Question: Is it possible the tests fail because of this change?

sadhansood: Yes, definitely - some of the tests have timeouts which work with lower values (small timeouts) for legacy event processor but for checkpoint based event processor those timeouts need to be higher because checkpoint processing takes longer to get to the right event.

PR #2350, opened 2025-07-18
test: [wip] randomized testing client

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-stress/src/main.rs
```diff
@@ -126,6 +132,8 @@ async fn main() -> anyhow::Result<()> {
     let metrics_address = SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), args.metrics_port);
     let registry_service = mysten_metrics::start_prometheus_server(metrics_address);
     let prometheus_registry = registry_service.default_registry();
+
+    // TODO(WAL-935): we should different metrics used in sdk::client and stress client.
```
liquid-helium: nit: TODO(WAL-935): we should use different metrics in sdk::client and stress client.

crates/walrus-stress/src/main.rs
```diff
@@ -327,3 +338,45 @@ async fn run_staking(config: ClientConfig, _metrics: Arc<ClientMetrics>) -> anyh
...
     }
 }
+
+async fn run_single_client_workload(
+    client_config: ClientConfig,
+    metrics: Arc<ClientMetrics>,
+    args: SingleClientWorkloadArgs,
+) -> anyhow::Result<()> {
+    tracing::info!("starting the single client stress runner");
+    tracing::info!("args: {:?}", args);
```
liquid-helium: nit: put them in a single `info!`?

crates/walrus-stress/src/single_client_workload.rs
```diff
@@ -0,0 +1,206 @@
...
+                store_length,
+            } => {
+                let now = Instant::now();
+                let store_result = self
+                    .client
+                    .reserve_and_store_blobs_retry_committees(
+                        &[blob.as_slice()],
+                        DEFAULT_ENCODING,
+                        *store_length,
+                        StoreOptimizations::none(),
```
liquid-helium: Question: why is this set as none() all the time, can we randomize it?

crates/walrus-stress/src/single_client_workload.rs
```diff
@@ -0,0 +1,206 @@
...
+                deletable,
+                store_length,
+            } => {
+                let now = Instant::now();
+                let store_result = self
+                    .client
+                    .reserve_and_store_blobs_retry_committees(
+                        &[blob.as_slice()],
+                        DEFAULT_ENCODING,
+                        *store_length,
```
liquid-helium: nit: rename to `epochs_ahead` to be consistent, it is a little misleading, I thought it was used as some data size until I reached this line

crates/walrus-stress/src/single_client_workload.rs
```diff
@@ -0,0 +1,206 @@
...
+                            client_op.clone(),
+                        );
+                    }
+                    _ => {
+                        anyhow::bail!(
+                            "client op {:?} received unexpected store result: {:?}",
+                            client_op,
+                            store_result[0]
+                        );
+                    }
```
liquid-helium, line 189: Shall we allow more results, like `AlreadyCertified` to the minimum?

crates/walrus-stress/src/single_client_workload/blob_pool.rs
```diff
@@ -0,0 +1,395 @@
...
+    blob_object_id: ObjectID,
+    /// Whether the blob is deletable.
+    deletable: bool,
+    /// The epoch at which the blob will be deleted.
+    end_epoch: Epoch,
+}
+
+/// Manages a pool of blobs that are live in the system.
+pub(crate) struct BlobPool {
+    blobs: HashMap<BlobId, BlobDataAndInfo>,
```
liquid-helium, line 30: Can we also test the case that the same blob is stored both as deletable and permanent, and even as a Quilt in the future, with a MultiMap? Or Keyed by ObjectID?

crates/walrus-stress/src/main.rs
halfprice: Done.

crates/walrus-stress/src/main.rs
halfprice: Good catch. Done

crates/walrus-stress/src/single_client_workload.rs
halfprice: Good question. I'm not sure what are the different optimizations and just learned about it. Seems there are quite a few combinations. I added a TODO for this.

crates/walrus-stress/src/single_client_workload.rs
halfprice: Done. Did the same change for Extend.

crates/walrus-stress/src/single_client_workload.rs
halfprice: Currently this is not possible. Therefore, if `AlreadyCertified` is returned, that should be an error.

However, I see you point that maybe we should also try to test upload the same blob multiple times. I added a TODO for this. (Mostly because I don't know what are the possible return values if we upload the same blob multiple times).

crates/walrus-stress/src/single_client_workload/blob_pool.rs
halfprice: Good ideal. I think this has the similar spirit of your other comments of testing uploading the same blob multiple times. I added a TODO for this. It may need some thinking to figure out what's the best structure to hold blob info.

halfprice:
Thanks a lot @liquid-helium for your initial review and feedback! I've addressed them (either by modifying the code or adding a TODO). PTAL.

crates/walrus-stress/src/single_client_workload/blob_pool.rs
```diff
@@ -0,0 +1,397 @@
...
+    deletable: bool,
+    /// The epoch at which the blob will be deleted.
+    end_epoch: Epoch,
+}
+
+/// Manages a pool of blobs that are live in the system.
+pub(crate) struct BlobPool {
+    // TODO(WAL-946): when writing the same blob but with different properties (e.g. deletable or
+    // permanent), we may want to keep the object id in the map as well.
+    blobs: HashMap<BlobId, BlobDataAndInfo>,
```
liquid-helium, line 30: Suggestion: add a max_num_blobs, to avoid OOM.

crates/walrus-stress/src/single_client_workload.rs
```diff
@@ -0,0 +1,208 @@
...
+                store_epoch_ahead,
+            } => {
+                let now = Instant::now();
+                // TODO(WAL-945): test more StoreOptimizations.
+                let mut store_args = StoreArgs::new(
+                    DEFAULT_ENCODING,
+                    *store_epoch_ahead,
+                    StoreOptimizations::none(),
+                    BlobPersistence::from_deletable_and_permanent(*deletable, !deletable)?,
+                    PostStoreAction::Keep,
```
liquid-helium, line 167: Suggestion: Test different PostStoreActoin. (Likely in a followup PR)

crates/walrus-stress/src/single_client_workload/single_client_workload_arg.rs
```diff
@@ -0,0 +1,178 @@
...
+        #[arg(long, default_value_t = 1024)]
+        min_size_bytes: usize,
+        /// Maximum size in bytes
+        #[arg(long, default_value_t = 20 * 1024 * 1024)]
+        max_size_bytes: usize,
+        /// Store length distribution configuration
+        #[command(subcommand)]
+        store_length_distribution: RequestStoreLengthDistributionArgs,
+    },
+    /// Configure workload with Poisson size distribution
```
liquid-helium: nit: add period at the end of comments, here and others.

crates/walrus-stress/src/single_client_workload/single_client_workload_config.rs
```diff
@@ -0,0 +1,394 @@
...
+        } else if random_value
+            < self.read_weight
+                + self.write_permanent_weight
+                + self.write_deletable_weight
+                + self.delete_weight
+        {
+            RequestType::Delete
+        } else {
+            RequestType::Extend
+        }
```
liquid-helium, line 78: nit: storing the thresholds, rather than the weights, like read_threshold, write_*_threshold, etc, so that we can avoid the sum operations.

crates/walrus-stress/src/single_client_workload/blob_pool.rs
halfprice: Great suggestion. Added max_blobs_in_pool parameter.

crates/walrus-stress/src/single_client_workload.rs
halfprice: TODO added

crates/walrus-stress/src/single_client_workload/single_client_workload_arg.rs
halfprice: Done.

crates/walrus-stress/src/single_client_workload/single_client_workload_config.rs
halfprice: Not sure if I understand the `threshold`, or may miss the point of the suggestion. The nice thing about current weight input is that say if read, write, delete have weight 3, 3, 4, we know that 30% of requests are reads, 30% of requests are writes, and 40% of requests are deletes. So it's easier for the user to set the input.

PR #2354, opened 2025-07-21
fix(node): mark events completed when cancelling sync for expired blobs

mlegner:
## Description

Follow-up on #2302.

## Test plan

Automatic tests.


halfprice (approved):
This looks great! Thanks for the quick fix @mlegner! Should we also include this in 1.29?

PR #2362, opened 2025-07-22
docs: updated quilt docs.

liquid-helium:
## Description

Contributes to WAL-943

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/docs/book/operator-guide/aggregator.md
+++ b/docs/book/operator-guide/aggregator.md
@@ -56,8 +56,8 @@ especially in a future Mainnet deployment.
 
 By default, [store blob](../usage/web-api.md#store) requests are limited to 10 MiB; you can
 increase this limit through the `--max-body-size` option.
-[Store quilt](../usage/web-api.md#storing-quilts) requests are limited to 100 MB by default, and can
-be increased using the `--max-quilt-body-size` option.
+[Store quilt](../usage/web-api.md#storing-quilts) requests are limited to 100 MiB by default, and
+can be increased using the `--max-quilt-body-size` option.
 
 ### Daemon metrics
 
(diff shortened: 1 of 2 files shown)


docs/book/operator-guide/aggregator.md
```diff
@@ -54,8 +54,10 @@ important to ensure only authorized parties may access it, or other measures to
 especially in a future Mainnet deployment.
 ```
 
-By default, PUT requests are limited to 10 MiB; you can increase this limit through the
-`--max-body-size` option.
+By default, [store blob](../usage/web-api.md#store) requests are limited to 10 MiB; you can
+increase this limit through the `--max-body-size` option.
+[Store quilt](../usage/web-api.md#storing-quilts) requests are limited to 100 MB by default, and can
```
mlegner: Question: Is this actually 100 MB or 100 MiB?

docs/book/usage/quilt.md
```diff
@@ -46,13 +46,21 @@ can be the target.
...
 
-The following table demonstrates shows the potential cost savings in WAL when storing 600 small
+- **Walrus storage and write fees**: By consolidating multiple small blobs into a single
+    quilt, storage costs can be reduced dramatically—more than 400x for files around 10KB—making
+    it an efficient solution for cost-sensitive applications.
+
+- **Sui computation and object storage fees**: Storing many blobs as a single quilt
+    significantly reduces Sui gas costs. In our test runs with 600 files stored in a quilt,
+    we observed **238x** savings in Sui fees compared to storing them as individual blobs.
+    Notably, Sui cost savings depend primarily on the number of files per quilt rather than
```
mlegner: Nit: I think the cost savings *only* depend on the number of files.
```suggestion
    Notably, Sui cost savings only depend on the number of files per quilt rather than
```

mlegner (approved):

docs/book/operator-guide/aggregator.md
liquid-helium: It should be MiB, thanks for the catch!

docs/book/usage/quilt.md
liquid-helium: Suggestion applied.

PR #2363, opened 2025-07-22
chore: cleanup database table options hierarchy

halfprice:
## Description

The existing DatabaseConfig is quite confusing. First of all, DatabaseTableOptions defines all the default CF options (standard, optmized_for_blobs), and then the DatabaseConfig can override any of these basic template, as well as any
individual CF options. What's more confusing is that when overriding the basic template, we take, e.g. DatabaseConfig::standard, only from DatabaseConfig completely, but when overriding individual CF options, it inherits the basic option first, and then override any fields that exist in the DatabaseConfig.

This PR attempts to make DB option setup clearer and not introduce any breaking change. It tries to clearly assign functionalities to 3 structures:

- DatabseTableOptions: this is where the options and default templates are defined.
- DatabaseConfig: this is the option config included in the StorageNodeConfig. Any fields that do not exist (or set to None) in the DatabaseConfig, the corresponding CF options will always inherit from their default template.
- DatabaseTableOptionsFactory: this is the object where each CF's options are generated by combining the template from DatabaseTableOptions and any overrides in DatabaseConfig.

To sum up, and how to set table option after this PR:
- each CF's option derives from one of the 4 basic templates: standard, optimized_for_blobs, metadata, and blob_info.
- Any fields added in DatabaseConfig only overrides that specific fields in the CF's options. All other fields are inherited from their default template.

The creation of DatabaseTableOptionsFactory is also where the shared blob cache pool going to be implemented.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage/database_config.rs
+++ b/crates/walrus-service/src/node/storage/database_config.rs
@@ -72,6 +72,13 @@ pub struct DatabaseTableOptions {
     hard_pending_compaction_bytes_limit: Option<usize>,
 }
 
+// DatabaseTableOptions specifies 4 generally used options for different tables to adopt as their
+// default options.
+//
+// 1. standard: used by all tables except for slivers, metadata, and blob_info column families.
+// 2. optimized_for_blobs: used by slivers column families.
+// 3. metadata: used by metadata column family.
+// 4. blob_info: used by blob_info and per_object_blob_info column families.
 impl DatabaseTableOptions {
     fn standard() -> Self {
         Self {
@@ -95,6 +102,8 @@ impl DatabaseTableOptions {
     }
 
     fn optimized_for_blobs() -> Self {
+        // - Use blob mode
+        // - Large block cache size
         Self {
             enable_blob_files: Some(true),
             min_blob_size: Some(1 << 20),
@@ -116,26 +125,21 @@ impl DatabaseTableOptions {
     }
 
     fn metadata() -> Self {
+        // Metadata column family by far is the most frequently accessed column family and also
+        // stores the most data.
         Self {
-            enable_blob_files: Some(false),
-            min_blob_size: None,
-            blob_file_size: None,
-            blob_compression_type: Some("zstd".to_string()),
-            enable_blob_garbage_collection: None,
-            blob_garbage_collection_age_cutoff: None,
-            blob_garbage_collection_force_threshold: None,
-            blob_compaction_read_ahead_size: None,
             write_buffer_size: Some(512 << 20),      // 512 MB,
             target_file_size_base: Some(512 << 20),  // 512 MB,
             max_bytes_for_level_base: Some(5 << 30), // 5 GB,
             block_cache_size: Some(512 << 20),       // 512 MB,
             block_size: Some(64 << 10),              // 64 KiB,
-            pin_l0_filter_and_index_blocks_in_block_cache: Some(true),
             soft_pending_compaction_bytes_limit: None,
             // TODO(WAL-840): decide whether we want to keep this option even after all the nodes
             // applied RocksDB 0.22.0, or apply it to all column families.
             hard_pending_compaction_bytes_limit: Some(0), // Disable write stall.
+            ..Default::default()
         }
+        .inherit_from(Self::standard())
     }
 
     fn blob_info() -> Self {
@@ -143,6 +147,7 @@ impl DatabaseTableOptions {
             block_cache_size: Some(512 << 20),
             ..Default::default()
         }
+        .inherit_from(Self::standard())
     }
 
     /// Inherit from the default options. If a field is None, use the value from the
@@ -320,16 +325,18 @@ impl From<&GlobalDatabaseOptions> for Options {
 /// Database configuration for Walrus storage nodes.
 ///
 /// The `standard` options are applied to all tables except for slivers and metadata. The
-/// `optimized_for_blobs` options are applied to sliver and metadata tables.
+/// `optimized_for_blobs` options are applied to sliver tables. The `metadata` options are applied
+/// to metadata tables.
 ///
 /// Options for all individual tables can be set as well through the `node_status`, `metadata`,
 /// `blob_info`, `per_object_blob_info`, `event_cursor`, `shard`, `shard_status`,
 /// `shard_sync_progress`, and `pending_recover_slivers` fields.
 ///
-/// **Warning:** Note that the configuration is currently not properly hierarchical. For example, if
-/// the `metadata` options are defined, they are *not* merged with the `optimized_for_blobs` or
-/// `standard` options. Any options that should not be `None` need to be set explicitly, even if
-/// they are equal to those from the `standard` or `optimized_for_blobs` options.
+/// For any `DatabaseTableOptions`, they have a default options to inherit from. For any partial
+/// options, unset fields will inherit from the default options.
+///
+/// Note that when changing `standard` or `optimized_for_blobs` options, the changes will affect any
+/// existing column families that inherit from them.
 #[serde_with::serde_as]
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq)]
 #[serde(default)]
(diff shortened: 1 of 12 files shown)


crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -317,19 +324,31 @@ impl From<&GlobalDatabaseOptions> for Options {
...
 }
 
-/// Database configuration for Walrus storage nodes.
+/// Database configuration for Walrus storage nodes. Note that any populated fields in this struct
+/// will only override the same field in the default options in `DatabaseTableOptionsFactory`.
 ///
 /// The `standard` options are applied to all tables except for slivers and metadata. The
-/// `optimized_for_blobs` options are applied to sliver and metadata tables.
+/// `optimized_for_blobs` options are applied to sliver tables. The `metadata` options are applied
+/// to metadata tables.
```
mlegner: Question: What about the `blob_info` tables?

crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -378,170 +397,266 @@ pub struct DatabaseConfig {
...
-        self.global.clone()
+impl Default for DatabaseConfig {
+    fn default() -> Self {
+        Self {
+            global: GlobalDatabaseOptions::default(),
+            standard: DatabaseTableOptions::standard(),
+            optimized_for_blobs: DatabaseTableOptions::optimized_for_blobs(),
+            metadata: DatabaseTableOptions::metadata(),
+            node_status: None,
+            blob_info: None,
```
mlegner: Important: It's unclear if the `metadata` and `blob_info` options here are relevant for the `DatabaseTableOptionsFactory` or the `DatabaseTableOptions` for the table with the same name, or both. AFAIU, it's both, but this is a bit confusing. In particular in case of `blob_info` as this is also inherited by the `per_object_blob_info`.

crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -378,170 +397,266 @@ pub struct DatabaseConfig {
...
-            .map(|options| options.inherit_from(self.optimized_for_blobs.clone()))
-            .unwrap_or_else(|| self.optimized_for_blobs.clone())
+    /// Returns the metadata database option inherited from the default
+    /// DatabaseTableOptions::metadata().
+    pub(crate) fn metadata(&self) -> DatabaseTableOptions {
+        Self::get_table_options_with_inheritance(
+            &self.database_config.metadata,
+            self.default_metadata_options.clone(),
+        )
     }
```
mlegner: Question: Aren't the two set by the same config parameters?

crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -378,170 +397,266 @@ pub struct DatabaseConfig {
...
+            self.default_optimized_for_blobs_options.clone(),
+        )
+    }
+
+    /// Returns the node status database option inherited from Self::standard().
+    pub(crate) fn node_status(&self) -> DatabaseTableOptions {
+        Self::get_optional_table_options_with_inheritance(
+            &self.database_config.node_status,
+            self.default_standard_options.clone(),
+        )
```
mlegner: Suggestion: Given we use inheritance from the `default_standard_options` many times, we could define a separate function for this.

crates/walrus-service/src/node/storage/database_config.rs
halfprice: good catch! done.

crates/walrus-service/src/node/storage/database_config.rs
halfprice: yeah, this is the major confusion I had too. These two option in DababaseConfig have two meanings and also have two behaviors. In my first attempt, I tried to make these two only relevant to  `DatabaseTableOptions`. That's why in your other comment, you also noticed that the metadata option is inherited from the same DatabaseConfig::metadata() twice (first time creating the default, second time getting the actual option used in the CF).

Thinking a bit more on this. I now completely separated option template with specific CF option override in DatabaseConfig. SInce we only add new fields, this shouldn't be a breaking change.

Now, we have standard, optimized_for_blobs, metadata_template, and blob_info_template, which are relevant to DatabaseConfig to set the template. Then we have individual CF option overrides. Here, again metadata and blob_info is inherited from the same option twice, which may also cause some confusing in the future. The goal here is to make DatabaseConfig clear to use

crates/walrus-service/src/node/storage/database_config.rs
halfprice: please see my reply to the other comment above. Now I separate metadata into metadata_template and metadata to make it clear that one is relevant to DatabaseTableOptions, and one is relevant to DatabaseTableOptionsFactory.

crates/walrus-service/src/node/storage/database_config.rs
halfprice: Great idea! Done.

halfprice:
> Thank you @halfprice for working on this! I could be wrong but after reading through the motivation behind this PR, we might be able to achieve the same thing with a simple few line changes:
> 
> 1. Add template functions so we can merge instead of replace:
> 
> ```
> impl DatabaseConfig {
>       pub fn standard(&self) -> DatabaseTableOptions {
>           self.standard.clone().inherit_from(DatabaseTableOptions::standard())
>       }
> 
>       pub fn optimized_for_blobs(&self) -> DatabaseTableOptions {
>           self.optimized_for_blobs.clone().inherit_from(DatabaseTableOptions::optimized_for_blobs())
>       }
>   }
> ```
> 
> 2. Ensure we use template functions instead of template variables in per table settings:
> 
> ```
> pub fn blob_info(&self) -> DatabaseTableOptions {
>         self.blob_info
>             .clone()
>             // notice we use self.standard() vs self.standard
>             .map(|options| options.inherit_from(self.standard().clone()))
>             .unwrap_or_else(|| self.standard().clone())
>     }
> ```
> 
> If this minimal change is doing the same thing, then I feel like it might be worth considering given it's simple and avoids adding another abstraction layer.

Hi @sadhansood , thanks for the review and the comment! Your understand is correct and what you suggested was exactly my first version of this PR. So what we eventually want to achieve is:

```
(DatabaseTableOptions_default_template + DatabaseConfig_default_override) --- generate ---> default template

default template + DatabaseConfig_individual_cf_override --- generate ----> cf specific options

cf specific options --- generate(combine with other setting such as shared block cache) ---> cf's RocksDb::Options
```

So the difference between your suggestion vs. this PR is just where the `default template` is stored. In this PR, the
default template is stored in the new DatabaseTableOptionsFactory, but in your proposal, it remains in DatabaseConfig. So fundamentally, they are the same.

In any case, we need a DatabaseTableOptionsFactory to hold the shared block cache anyway, so I think overall, the amount
of change between these two approaches is going to be similar.

Here https://github.com/MystenLabs/walrus/pull/2379, I made an attempt to implement your suggestion. If this is something
you prefer, I can start working in that PR. At least, in that PR, I don't need to create DatabaseTableOptionsFactory yet, but
the next PR introducing it may be large.

halfprice:
> > > Thank you @halfprice for working on this! I could be wrong but after reading through the motivation behind this PR, we might be able to achieve the same thing with a simple few line changes:
> > > 
> > > 1. Add template functions so we can merge instead of replace:
> > > 
> > > ```
> > > impl DatabaseConfig {
> > >       pub fn standard(&self) -> DatabaseTableOptions {
> > >           self.standard.clone().inherit_from(DatabaseTableOptions::standard())
> > >       }
> > > 
> > >       pub fn optimized_for_blobs(&self) -> DatabaseTableOptions {
> > >           self.optimized_for_blobs.clone().inherit_from(DatabaseTableOptions::optimized_for_blobs())
> > >       }
> > >   }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 2. Ensure we use template functions instead of template variables in per table settings:
> > > 
> > > ```
> > > pub fn blob_info(&self) -> DatabaseTableOptions {
> > >         self.blob_info
> > >             .clone()
> > >             // notice we use self.standard() vs self.standard
> > >             .map(|options| options.inherit_from(self.standard().clone()))
> > >             .unwrap_or_else(|| self.standard().clone())
> > >     }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > If this minimal change is doing the same thing, then I feel like it might be worth considering given it's simple and avoids adding another abstraction layer.
> > 
> > 
> > Hi @sadhansood , thanks for the review and the comment! Your understand is correct and what you suggested was exactly my first version of this PR. So what we eventually want to achieve is:
> > ```
> > (DatabaseTableOptions_default_template + DatabaseConfig_default_override) --- generate ---> default template
> > 
> > default template + DatabaseConfig_individual_cf_override --- generate ----> cf specific options
> > 
> > cf specific options --- generate(combine with other setting such as shared block cache) ---> cf's RocksDb::Options
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So the difference between your suggestion vs. this PR is just where the `default template` is stored. In this PR, the default template is stored in the new DatabaseTableOptionsFactory, but in your proposal, it remains in DatabaseConfig. So fundamentally, they are the same.
> > In any case, we need a DatabaseTableOptionsFactory to hold the shared block cache anyway, so I think overall, the amount of change between these two approaches is going to be similar.
> > Here #2379, I made an attempt to implement your suggestion. If this is something you prefer, I can start working in that PR. At least, in that PR, I don't need to create DatabaseTableOptionsFactory yet, but the next PR introducing it may be large.
> 
> Hey @halfprice! Thanks for the explanation and for putting together PR #2379 -I didn't realize that you've already tried the simpler approach and are forward thinking for handling the shared cache. I thought about this more and It seems like we're trying to solve two different problems at once in this current PR: 1> The inheritance issue (which is genuinely confusing right now) 2> Setting up for shared block cache which got me thinking, what if we could find a middle ground i.e. we could get that future-proofing you're looking for, but build it on top of the simpler, non-breaking fix (PR #2379). Consider this slightly alternative and simpler approach and lmk what you think - we could keep the factory idea but make it way simpler:
> 
> ```rust
>   struct DatabaseTableOptionsFactory {
>       config: DatabaseConfig,
>       shared_cache: Option<Arc<rocksdb::Cache>>,
>   }
> 
>   impl DatabaseTableOptionsFactory {
>       fn new(config: DatabaseConfig) -> Self {
>           let shared_cache = config.global.shared_block_cache_size
>               .map(|size| Arc::new(rocksdb::Cache::new_lru_cache(size)));
>           Self { config, shared_cache }
>       }
> 
>       pub fn metadata(&self) -> rocksdb::Options {
>           self.config.metadata().to_options(self.shared_cache.clone())
>       }
> 
>       // same for other column families
>   }
> ```
> 
> The factory literally just handles the cache - that's it. All the inheritance logic stays in DatabaseConfig where it already makes sense (especially with your #2379 fix). This way: 1> We get the inheritance fix right away with #2379 (like 10 lines changed) 2> The factory becomes super simple - just cache management and overall much less code change but still gets us to the same place we want to be.

thanks for the detailed comments @sadhansood ! If we move forward with #2379, what you proposed is exactly what we are going to do. (even with this PR, the way to support shared cache is also similar to what you suggested here). OK, let me clear that PR a bit while keep this one open.

halfprice:
This is replaced by #2379

PR #2372, opened 2025-07-24
fix(node): fix handling of invalid blobs and deletions with incomplete history

mlegner:
## Description

Previously, the node would not handle blob deletions and invalidations correctly while catching up with incomplete history. This makes the following changes:
- Allow handling `InvalidBlobId` events even when the corresponding blob is not yet tracked.
- Allow setting the `is_metadata_stored` flag for blobs that are not yet tracked.
- Do not log warnings when handling `BlobDeleted` events for untracked blobs while catching up with incomplete history.

Closes WAL-947
Also contributes to WAL-553 (fix skipping event blobs when starting with an incomplete history).

## Test plan

Started a new node on Testnet and had it catch up. No errors or warnings in the logs.

TODO: 
- [ ] Check why the `blob_info_consistency_check` differs.

### DIFF
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -159,7 +159,7 @@ impl BackgroundEventProcessor {
             // subsequent certify or delete events may update the `blob_info`; so we cannot remove
             // it even if it is no longer valid in the *current* epoch
             if !blob_info.is_registered(event.epoch) {
-                tracing::debug!("deleting data for deleted blob");
+                tracing::debug!(walrus.blob_id = %blob_id, "deleting data for deleted blob");
                 // TODO (WAL-201): Actually delete blob data.
             }
         } else if self
@@ -170,13 +170,13 @@ impl BackgroundEventProcessor {
         {
             tracing::debug!(
                 walrus.blob_id = %blob_id,
-                "skipping processing of `BlobDeleted` event while catching up with incomplete \
-                history"
+                "handling a `BlobDeleted` event for an untracked blob while catching up with \
+                incomplete history; not deleting blob data"
             );
         } else {
             tracing::warn!(
                 walrus.blob_id = %blob_id,
-                "handling `BlobDeleted` event for an untracked blob"
+                "handling a `BlobDeleted` event for an untracked blob"
             );
         }
 
@@ -198,7 +198,10 @@ impl BackgroundEventProcessor {
         self.blob_sync_handler
             .cancel_sync_and_mark_event_complete(&event.blob_id)
             .await?;
-        self.node.storage.delete_blob_data(&event.blob_id).await?;
+        self.node
+            .storage
+            .delete_blob_data(&event.blob_id, false)
+            .await?;
 
         event_handle.mark_as_complete();
         Ok(())
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -48,10 +48,10 @@ impl BackgroundEventProcessor {
             )
             .dec();
 
-            if let Err(e) = self.process_event(event_handle, blob_event).await {
+            if let Err(error) = self.process_event(event_handle, blob_event).await {
                 // TODO(WAL-874): to keep the same behavior as before BackgroundEventProcessor, we
                 // should propagate the error to the node and exit the process if necessary.
-                tracing::error!("error processing blob event: {}", e);
+                tracing::error!(?error, "error processing blob event");
             }
         }
     }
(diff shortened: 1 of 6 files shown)


crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -162,6 +162,17 @@ impl BackgroundEventProcessor {
...
             }
+        } else if self
+            .node
+            .storage
+            .node_status()?
+            .is_catching_up_with_incomplete_history()
+        {
+            tracing::debug!(
+                walrus.blob_id = %blob_id,
+                "skipping processing of `BlobDeleted` event while catching up with incomplete \
```
halfprice: Nit: it doesn't look like we are "skipping" anything here. Maybe just `handling `BlobDeleted` event for an untracked blob while catching up with incomplete history` or similar log?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1273,40 +1273,57 @@ impl Mergeable for BlobInfoV1 {
...
+                .into(),
+            ),
+            BlobInfoMergeOperand::MarkInvalid {
+                epoch,
+                status_event,
+            } => Some(BlobInfoV1::Invalid {
+                epoch,
+                event: status_event,
+            }),
+            BlobInfoMergeOperand::MarkMetadataStored(_) => Some(
```
halfprice: Question: why do we need this case? It seems that MarkMetadataStored should always be triggered when blob is registered (otherwise, the op that stores blob metadata won't be executed). Or do I miss something?

halfprice (commented):
Thanks @mlegner for the quick fix! The invalid blob handling LGTM! The PR title says it also fixes handling blob deletion, but I don't see it in the code. Is it still relevant? I don't feel we need to handle blob deletion specially.

crates/walrus-service/src/node/blob_event_processor.rs
mlegner: We do skip the deletion of data which has a placeholder above. But I can still reword this a bit. 👍

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: We update the metadata stored to `false` [when deleting the metadata](https://github.com/MystenLabs/walrus/blob/9b2aa5600c4d4a96e1d0f5c4723cacee3101f4ff/crates/walrus-service/src/node/storage.rs#L662), which in turn [is called when marking a blob as invalid](https://github.com/MystenLabs/walrus/blob/9b2aa5600c4d4a96e1d0f5c4723cacee3101f4ff/crates/walrus-service/src/node/blob_event_processor.rs#L201). 

However, now that I think about it, we can skip this blob-info update as it has no effect after having already set it to invalid.

mlegner: Not sure if we should allow this here though: There shouldn't be an issue with allowing it, but it's also not really needed assuming everything else is implemented correctly. WDYT?

crates/walrus-service/src/node/blob_event_processor.rs
mlegner: Done in 9dc2c1b7cd014f0d1e7c4955c21036a4b1edbcaf.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Change to the blob-info update is in 9dc2c1b7cd014f0d1e7c4955c21036a4b1edbcaf.

halfprice: > we can skip this blob-info update as it has no effect after having already set it to invalid.

Yeah, this was part of my original thoughts too. To me, it feels that if merge_new receiving MarkMetadataStored(), it indicates a bug, since this shouldn't happen. If my understanding is correct, I wonder if we should remove this branch to not overlook a potential issue. But I agree that this may not be super important based on the current code.

Also, it seems that we should set `is_metadata_stored` to the value passed in BlobInfoMergeOperand::MarkMetadataStored()?

halfprice (approved):
LGTM! And thanks for the revision Markus! I left a reply for you to think about, but it's a minor one so you can decide what you want to do there.

crates/walrus-service/src/node/storage/blob_info.rs
sadhansood: +1 to @halfprice 's comment about removing the `MarkMetadataStored` branch as it seems logically impossible.

crates/walrus-service/src/node/storage.rs
```diff
@@ -634,9 +642,13 @@ impl Storage {
 
     /// Deletes the metadata and slivers for the provided [`BlobId`] from the storage.
     #[tracing::instrument(skip_all)]
-    pub async fn delete_blob_data(&self, blob_id: &BlobId) -> Result<(), TypedStoreError> {
+    pub async fn delete_blob_data(
+        &self,
+        blob_id: &BlobId,
+        update_blob_info: bool,
```
sadhansood: Do we need this boolean flag? It seems like the only caller for this function is `process_blob_invalid_event`.

sadhansood (approved):
LGTM, thank you @mlegner!

crates/walrus-service/src/node/storage.rs
mlegner: Good point. I've removed the parameter here (documenting the behavior) but kept it in `delete_metadata` as a reminder to think about this in case we use it somewhere else in the future.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Ok, I've now changed the behavior for the `MarkMetadataStored` case to what we had before.

PR #2374, opened 2025-07-24
fix: use correct version in testnet-contracts

karlwuest:
## Description

I discovered that after deploying the walrus_subsidies contract on testnet, I accidentally bumped the version on the `testnet-contracts` instead of `contracts`. This fixes that. I checked that the correct version is deployed on testnet.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/subsidies/Move.lock
+++ b/contracts/subsidies/Move.lock
@@ -40,7 +40,7 @@ dependencies = [
 ]
 
 [move.toolchain-version]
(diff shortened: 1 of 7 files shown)


mlegner (approved):
Thanks for fixing this and sorry that I didn't catch this while reviewing #2316. 🙏

PR #2377, opened 2025-07-25
feat: Skip event blob attestation for non committee members

sadhansood:
## Description

This PR makes the non committee members not do any event blob attestations.

Contributes to WAL-553.

## Test plan

Added tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/event_blob_writer.rs
+++ b/crates/walrus-service/src/node/event_blob_writer.rs
@@ -33,6 +33,7 @@ use walrus_core::{
     ensure,
     metadata::VerifiedBlobMetadataWithId,
 };
+use walrus_sdk::active_committees::ActiveCommittees;
 use walrus_sui::{
     client::SuiClientError,
     types::{
@@ -1321,6 +1322,22 @@ impl EventBlobWriter {
             tracing::debug!("attestations are paused, skipping blob: {}", blob_id);
             return Ok(());
         }
+
+        let active_committees: ActiveCommittees = self.node.committee_service.active_committees();
+
+        // Nodes which are not in the latest committee should not try and attest the blob.
+        // They should just add the blob to the attested list and move on.
+        if !active_committees
+            .current_committee()
+            .contains(self.node.public_key())
+        {
+            tracing::debug!(
+                "node is not in the latest committee, skipping blob attestation: {}",
+                blob_id
+            );
+            return Ok(());
+        }
+
         tracing::info!(
             blob_id = %blob_id,
             epoch = self.current_epoch,
@@ -1821,7 +1838,7 @@ mod tests {
             DatabaseConfig,
             event_blob_writer::{EventBlobWriter, EventBlobWriterFactory},
         },
-        test_utils::StorageNodeHandle,
+        test_utils::{StorageNodeHandle, StubContractService},
     };
 
     #[tokio::test]
@@ -1831,7 +1848,8 @@ mod tests {
 
         let dir: PathBuf = tempfile::tempdir()?.keep();
         let registry = Registry::default();
-        let node = create_test_node().await?;
+        let stub_contract_service = std::sync::Arc::new(StubContractService::default());
+        let node = create_test_node_with_contract_service(stub_contract_service).await?;
         let blob_writer_factory = EventBlobWriterFactory::new(
             &dir,
             &DatabaseConfig::default(),
@@ -1883,7 +1901,8 @@ mod tests {
         const NUM_EVENTS_PER_CHECKPOINT: u64 = 1;
 
         let dir: PathBuf = tempfile::tempdir()?.keep();
-        let node = create_test_node().await?;
+        let stub_contract_service = std::sync::Arc::new(StubContractService::default());
+        let node = create_test_node_with_contract_service(stub_contract_service).await?;
         let registry = Registry::default();
 
         let blob_writer_factory = EventBlobWriterFactory::new(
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/contract_service.rs
mlegner, line 1: Suggestion: Move the new `AsAny` trait to `crates/walrus-service/src/test_utils.rs`. I also think you don't need to make it a supertrait of `SystemContractService` as you anyway implement it for all `Any + Send + Sync` types.

crates/walrus-service/src/node/event_blob_writer.rs
```diff
@@ -1332,14 +1348,15 @@ impl EventBlobWriter {
...
+        let node_capability_object_id = self.node.node_capability();
         match self
             .node
             .contract_service
             .certify_event_blob(
                 metadata.try_into()?,
                 checkpoint_sequence_number,
                 self.current_epoch,
-                self.node.node_capability(),
+                node_capability_object_id,
```
mlegner: Remark: This is an unrelated change, correct?

crates/walrus-service/src/node/event_blob_writer.rs
```diff
@@ -1439,7 +1456,11 @@ impl EventBlobWriter {
         let blob_id = *blob_metadata.blob_id();
 
         let mut batch = self.pending.batch();
-        match self.attest_blob(&blob_metadata, metadata.end).await {
+        let active_committees = self.node.committee_service.active_committees();
+        match self
+            .attest_blob(&blob_metadata, metadata.end, &active_committees)
+            .await
+        {
```
mlegner: Question/suggestion: Instead of getting the active committee first and passing it into the `attest_blob` function, wouldn't it be simpler to just get it inside the function?

mlegner (approved):
Thanks a lot for the fix, that was simpler than I expected. 🙂 

Thanks also for adding tests. Maybe @halfprice can take a brief look at those as well.

crates/walrus-service/src/node/event_blob_writer.rs
```diff
@@ -1332,14 +1348,15 @@ impl EventBlobWriter {
             .latest_attested_checkpoint_sequence_number
```
halfprice: nit: should we also update this metrics when skipping attestation?

crates/walrus-service/src/node/event_blob_writer.rs
```diff
@@ -1439,7 +1456,11 @@ impl EventBlobWriter {
         let blob_id = *blob_metadata.blob_id();
 
         let mut batch = self.pending.batch();
-        match self.attest_blob(&blob_metadata, metadata.end).await {
+        let active_committees = self.node.committee_service.active_committees();
+        match self
+            .attest_blob(&blob_metadata, metadata.end, &active_committees)
```
halfprice: question: it is possible that this node is lagging, and in the event blob's epoch, the node is not part of the committee, but self.node.committee_service.active_committees(), which returns the latest epoch (can be far ahead than event blob epoch), has the node in the committee. Is it ok if the node attests the blob in this case?

crates/walrus-service/src/test_utils.rs
```diff
@@ -1529,6 +1535,28 @@ impl CommitteeService for StubCommitteeService {
 pub struct StubContractService {
     pub(crate) system_parameters: FixedSystemParameters,
     pub(crate) node_capability_object: StorageNodeCap,
+    pub(crate) certify_event_blob_tx:
+        Arc<std::sync::Mutex<tokio::sync::mpsc::Sender<BlobObjectMetadata>>>,
+    #[allow(dead_code)]
+    pub(crate) certify_event_blob_rx:
+        Arc<std::sync::Mutex<tokio::sync::mpsc::Receiver<BlobObjectMetadata>>>,
```
halfprice, line 1561: Suggestion: first of all, this is a great idea to test whether the node has done a certain things or not. I just wonder if we can make the test more explicit.

Right now, the test assumes that a SubContractService will be used in the test, and then use as_ayn to downcast it and check the channel. Instead, can we make it more explicit that in the test body, we build storage node with `with_committee_service()` (https://github.com/MystenLabs/walrus/blob/3788d74f2c783d195622d79fdfc25c7feae1bb60/crates/walrus-service/src/test_utils.rs#L746) that passing in a StubCommitteeService explicitly? This way, if the default behavior of using StubCommitteeService is changed, the test won't get affected. Also, we no longer need the as_any trait, which seems to be only used in tests. WDYT?

halfprice (commented):
Thanks @sadhansood for implementing the improvement! The approach LGTM and I have one suggestion for making the test more explicit.

crates/walrus-service/src/node/event_blob_writer.rs
sadhansood: sure, will do

crates/walrus-service/src/node/event_blob_writer.rs
sadhansood: Yeah it is totally possible. For example - if there was an error in event blob writer which caused event blobs to not get certified for sometime, and then we fix the issue later in next epoch - the new committee will attest and certify events from previous epoch.

crates/walrus-service/src/test_utils.rs
sadhansood: cool, done

crates/walrus-service/src/node/event_blob_writer.rs
sadhansood: yes, its minor but i can put it back.

crates/walrus-service/src/node/event_blob_writer.rs
sadhansood: I don't think so, since we didn't actually attest.

PR #2378, opened 2025-07-25
chore: Make QuiltClientConfig part of ClientConfig.

liquid-helium:
## Description

Close WAL-927

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client/quilt_client.rs
+++ b/crates/walrus-sdk/src/client/quilt_client.rs
@@ -222,7 +222,7 @@ where
                 sliver_indices,
                 certified_epoch,
                 self.config.max_retrieve_slivers_attempts,
-                self.config.timeout_duration,
+                self.config.timeout,
             )
             .await;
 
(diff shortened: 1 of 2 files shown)


crates/walrus-sdk/src/client/quilt_client.rs
```diff
@@ -327,7 +328,7 @@ where
 }
 
 /// Configuration for the QuiltClient.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
```
mlegner, line 332: Suggestion: Use our "standard" way to deserialize the `Duration` with `DurationSeconds`:
```rust
    #[serde(rename = "timeout_secs")]
    pub timeout: Duration,
```

liquid-helium: Thanks for the suggestion, change applied.

PR #2379, opened 2025-07-26
chore: cleanup database table options hierarchy

halfprice:
## Description

(This is the second attempt to cleanup the database column family option hierarchy)

The existing DatabaseConfig is quite confusing. First of all, DatabaseTableOptions defines all the default CF options (standard, optmized_for_blobs), and then the DatabaseConfig can override any of these basic template, as well as any
individual CF options. What's more confusing is that when overriding the basic template, we take, e.g. DatabaseConfig::standard, only from DatabaseConfig completely, but when overriding individual CF options, it inherits the basic option first, and then override any fields that exist in the DatabaseConfig.

This PR attempts to make DB option setup clearer and not introduce any breaking change.  

First of all, DatabaseTableOptions now defines 4 template options: standard, optimized_for_blobs, metadata, and blob_info_template. Each individual column families are based on one of the templates.

Next, DatabaseConfig now used as overrides, and for any non present fields in DatabaseConfig, their default value will be used. This applies to both template options or specific column family options.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/storage/database_config.rs
+++ b/crates/walrus-service/src/node/storage/database_config.rs
@@ -420,12 +420,21 @@ impl DatabaseConfig {
             .inherit_from(DatabaseTableOptions::blob_info_template())
     }
 
+    /// Uses the template if the options are None, otherwise inherits from the template to set all
+    /// the None fields in `options`.
+    fn inherit_from_or_use_template(
+        options: &Option<DatabaseTableOptions>,
+        template: DatabaseTableOptions,
+    ) -> DatabaseTableOptions {
+        match options {
+            Some(options) => options.inherit_from(template),
+            None => template,
+        }
+    }
+
     /// Returns the node status database option.
     pub fn node_status(&self) -> DatabaseTableOptions {
-        self.node_status
-            .clone()
-            .map(|options| options.inherit_from(self.standard()))
-            .unwrap_or_else(|| self.standard())
+        Self::inherit_from_or_use_template(&self.node_status, self.standard())
     }
 
     /// Returns the metadata database option.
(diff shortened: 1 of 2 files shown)


halfprice:
Hi @sadhansood , please see https://github.com/MystenLabs/walrus/pull/2363#issuecomment-3121106753 and let me know if you prefer this version, and I can start clean this one up (adding tests, modifying comments, etc.)

sadhansood (approved):

halfprice:
Hi @sadhansood, I cleaned up this PR and it's ready for a full review.

Hi @mlegner, please see the [comment here](https://github.com/MystenLabs/walrus/pull/2363#issuecomment-3121390998) that lead to a new PR to tackle cf option cleanup.

crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -384,136 +402,165 @@ impl DatabaseConfig {
...
+    }
+
     /// Returns the node status database option.
     pub fn node_status(&self) -> DatabaseTableOptions {
         self.node_status
             .clone()
-            .map(|options| options.inherit_from(self.standard.clone()))
-            .unwrap_or_else(|| self.standard.clone())
+            .map(|options| options.inherit_from(self.standard()))
+            .unwrap_or_else(|| self.standard())
```
mlegner: Suggestion: Given we use this more than 10 times, how about we create a separate function for this? For example:
```rust
    fn inherit_from_or_use_template(
        options: &Option<DatabaseTableOptions>,
        template: DatabaseTableOptions,
    ) -> DatabaseTableOptions {
        match options {
            Some(options) => options.inherit_from(template),
            None => template,
        }
    }
```

halfprice: That is a very nice suggestion ! applied

PR #2389, opened 2025-07-31
feat: Track bloom filter performance using db statstics

sadhansood:
## Description

In this PR, we are adding tracking for bloom filter performance in rocksdb that will help us tune bloom filter settings which is very likely to help with point lookups and key existence checks.
For this to work, user will have to selectively enable statistics on the db of their choice through db options.

## Test plan

Added tests to ensure we can use this feature.

### DIFF
--- a/crates/typed-store/src/metrics.rs
+++ b/crates/typed-store/src/metrics.rs
@@ -372,6 +372,22 @@ pub struct OperationMetrics {
     pub rocksdb_very_slow_puts_count: IntCounterVec,
     /// Total duration of puts that took more than 1 second
     pub rocksdb_very_slow_puts_duration_ms: IntCounterVec,
+    /// Bloom filter ticker counters
+    /// Note: These are only available upon turning on statistics in the rocksdb options
+    /// using `opt.enable_statistics();` and with lowest setting
+    /// `opt.set_statistics_level(StatsLevel::ExceptHistogramOrTimers);`
+    /// Number of times bloom filter has avoided file reads (negatives)
+    pub rocksdb_bloom_filter_useful_total: IntGaugeVec,
+    /// Number of times bloom FullFilter has not avoided the reads
+    pub rocksdb_bloom_filter_full_positive_total: IntGaugeVec,
+    /// Number of times bloom FullFilter has not avoided the reads and data actually exist
+    pub rocksdb_bloom_filter_full_true_positive_total: IntGaugeVec,
+    /// Number of times prefix filter was queried
+    pub rocksdb_bloom_filter_prefix_checked_total: IntGaugeVec,
+    /// Number of times prefix filter returned false so prevented accessing data+index blocks
+    pub rocksdb_bloom_filter_prefix_useful_total: IntGaugeVec,
+    /// Number of times prefix filter found a key matching the point query
+    pub rocksdb_bloom_filter_prefix_true_positive_total: IntGaugeVec,
 }
 
 impl OperationMetrics {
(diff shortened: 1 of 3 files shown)


crates/typed-store/src/rocks/tests.rs
```diff
@@ -730,117 +730,154 @@ async fn test_multi_remove() {
...
-    tokio::time::advance(Duration::from_secs(1)).await;
-    tokio::task::yield_now().await;
-    assert!(sampling_interval.sample());
-}
-
 #[tokio::test]
-async fn test_iterator_seek() {
-    let db: DBMap<u32, String> = open_map(temp_dir(), None);
-
-    db.insert(&123, &"123".to_string())
```
halfprice: Question: are these tests removed? If so, why do we want to remove it?

crates/typed-store/src/metrics.rs
```diff
@@ -372,6 +372,19 @@ pub struct OperationMetrics {
...
+    /// Number of times bloom FullFilter has not avoided the reads
+    pub rocksdb_bloom_filter_full_positive: IntGaugeVec,
+    /// Number of times bloom FullFilter has not avoided the reads and data actually exist
+    pub rocksdb_bloom_filter_full_true_positive: IntGaugeVec,
+    /// Number of times prefix filter was queried
+    pub rocksdb_bloom_filter_prefix_checked: IntGaugeVec,
+    /// Number of times prefix filter returned false so prevented accessing data+index blocks
+    pub rocksdb_bloom_filter_prefix_useful: IntGaugeVec,
+    /// Number of times prefix filter found a key matching the point query
+    pub rocksdb_bloom_filter_prefix_true_positive: IntGaugeVec,
```
halfprice: suggestion: can you document that these metrics requires DB to enable statistics collection? And maybe a pointer to how to enable it.

jpcsmith: Remark: As an aside, the metrics exposed by typed-store do not follow prometheus's [best practices for metric names](https://prometheus.io/docs/practices/naming/#metric-names), and so hinders usability in Grafana as the metric type is often unclear. Grafana also checks for the naming convention when making suggestions and issuing warnings about the usage of variables in panels.

Since you're adding new metrics here, you could consider naming those metrics according to best practices.

liquid-helium: Adding that can we also document the proper StatsLevel.

sadhansood:
> Thanks @sadhansood ! This is going to be useful. QQ: what's the drawback if we enable statistics by default? It seems that the metrics currently are not collected.

@halfprice The biggest problem is I didn't find a way to disable statistics in the options i.e. there is no equivalent of `opt.enable_statistics()`, so it seems like opt-in is better than opt-out. I am planning to enable it in some of our db's in a follow-up PR however once this lands.

crates/typed-store/src/metrics.rs
sadhansood: done, i suffixed metrics with `_total` according to the linked best practices.

crates/typed-store/src/rocks/tests.rs
sadhansood: my mistake, i've now put them back.

crates/typed-store/src/rocks.rs
```diff
@@ -892,6 +901,60 @@ impl<K, V> DBMap<K, V> {
...
+            );
+        db_metrics
+            .op_metrics
+            .rocksdb_bloom_filter_prefix_true_positive_total
+            .with_label_values(&[&rocksdb.db_name()])
+            .set(
+                rocksdb
+                    .db_options
+                    .get_ticker_count(Ticker::BloomFilterPrefixTruePositive) as i64,
+            );
```
liquid-helium, line 957: nit: Shall we check if statistics is enabled before calling the get_ticker_*? Maybe put them in a separate function?

sadhansood: @liquid-helium I think we can check that statistic is enabled or not but it may still not be very helpful. Different tickers are updated by the db based on different stats level and tracking which is going to be enabled by which level is tricky and likely not worth it.

PR #2392, opened 2025-07-31
fix(deps): bump all Sui dependencies to `v1.53.1`

mlegner:
## Description

This includes some other related changes:
- Bump `tokio` to `1.46.1`.
- Bump `fastcrypto` to `16fa86d0dd943024a9088d46850a72ecd55b7f46`.
- Bump `msim` to `192bd9590f8552d5a1c5debf66c4ff2672af037e`.
- Add `allow(deprecated_usage)` to `subsidies_tests.move`.
- Minor required code changes.

## Test plan

Test suite.

### DIFF
--- a/contracts/wal/Move.lock
+++ b/contracts/wal/Move.lock
@@ -2,7 +2,7 @@
 
 [move]
 version = 3
-manifest_digest = "73B1256D8F5FEB88338C0E732008BD622B6BA688E9DD516C1A08FB4247389CF1"
+manifest_digest = "8D1D707A67C745B724B86B96B1CCECA1DE291ABC8FF2B9F8C7C500B6884F7BF6"
 deps_digest = "F8BBB0CCB2491CA29A3DF03D6F92277A4F3574266507ACD77214D37ECA3F3082"
 dependencies = [
   { id = "Sui", name = "Sui" },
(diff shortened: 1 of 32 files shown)


contracts/wal/Move.lock
```diff
@@ -2,26 +2,26 @@
...
+manifest_digest = "8D1D707A67C745B724B86B96B1CCECA1DE291ABC8FF2B9F8C7C500B6884F7BF6"
 deps_digest = "F8BBB0CCB2491CA29A3DF03D6F92277A4F3574266507ACD77214D37ECA3F3082"
 dependencies = [
   { id = "Sui", name = "Sui" },
 ]
 
 [[move.package]]
 id = "MoveStdlib"
-source = { git = "https://github.com/MystenLabs/sui.git", rev = "testnet-v1.52.1", subdir = "crates/sui-framework/packages/move-stdlib" }
+source = { git = "https://github.com/MystenLabs/sui.git", rev = "testnet-v1.53.1", subdir = "crates/sui-framework/packages/move-stdlib" }
```
karlwuest, line 13: Important: We now also need to bump the dependency versions in the `testnet-contracts` directory. Otherwise, the upgrade simtest will fail in CI.

karlwuest (approved):

contracts/wal/Move.lock
mlegner: Ah, thanks for the reminder. We should probably change some tooling such that this is done and checked automatically. Also, the warning about the `testnet-contracts` below should not be printed for changes to the `Move.*` files.

crates/walrus-simtest/src/test_utils.rs
mlegner, line 1: @halfprice Could you take a quick look at these changes?

halfprice: LGTM!

halfprice (approved):

PR #2395, opened 2025-07-31
feat: add blob attributes when certifying blobs.

liquid-helium:
## Description

Enable adding blob attributes when certifying a blob.

Adding this to certify_blobs, so that if a blob is registered but not certified, it can be re-used, without contamination.

Only Rust API is provided, and we will add support for publisher and CLI support (input format to be decided).

Currently, it is used by QuiltClient to add a ` _walrusBlobType: quilt` attribute.

Contributes to WAL-960
Contributes to WAL-789

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -960,51 +960,6 @@ impl Client<SuiContractClient> {
             .collect())
     }
 
-    /// Encodes multiple blobs into sliver pairs and metadata.
-    ///
-    /// Returns a list of sliver pairs and metadata for each blob.
-    ///
-    /// Failed blobs are filtered out.
-    pub fn encode_blobs_to_pairs_and_metadata(
-        &self,
-        blobs: &[&[u8]],
-        encoding_type: EncodingType,
-    ) -> ClientResult<Vec<(Vec<SliverPair>, VerifiedBlobMetadataWithId)>> {
-        let walrus_store_blobs =
-            WalrusStoreBlob::<String>::default_unencoded_blobs_from_slice(blobs, &[]);
-
-        let encoded_blobs = self.encode_blobs(walrus_store_blobs, encoding_type)?;
-
-        debug_assert_eq!(
-            encoded_blobs.len(),
-            blobs.len(),
-            "the number of encoded blobs and the number of blobs must be the same"
-        );
-
-        // Failed blobs are filtered out.
-        let result = encoded_blobs
-            .into_iter()
-            .filter_map(|encoded_blob| {
-                if encoded_blob.is_failed() {
-                    None
-                } else {
-                    Some((
-                        encoded_blob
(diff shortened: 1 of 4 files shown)


crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -186,11 +189,19 @@ impl<'a, T: Debug + Clone + Send + Sync> WalrusStoreBlob<'a, T> {
     /// Create a list of UnencodedBlobs with default identifiers in the form of "blob_{:06}".
     pub fn default_unencoded_blobs_from_slice(
         blobs: &'a [&[u8]],
+        attributes: &[BlobAttribute],
     ) -> Vec<WalrusStoreBlob<'a, String>> {
+        assert!(attributes.is_empty() || attributes.len() == blobs.len());
```
giac-mysten, line 195: Question: are we sure we want to panic here instead of returning an error?

crates/walrus-sdk/src/client.rs
```diff
@@ -854,10 +859,11 @@ impl Client<SuiContractClient> {
     pub async fn reserve_and_store_blobs_retry_committees(
```
giac-mysten, line 859: Question: We also have the function `reserve_and_store_blobs_retry_committees_with_path`. Why are the attributes not added to that one as well? Are you thinking that the attributes should not be set with `walrus store`?

crates/walrus-sdk/src/client/client_types.rs
wbbradley: Unless we are expecting a large number of blobs and attributes, the type of this function could be adapted to force the caller to construct a pre-emptively zipped data structure. This would allow this subprogram to not have to check this. Like, a &[(&[u8], BlobAttribute)]. The name of `BlobAttribute` is confusing in this context, as it implies that each blob has a single BlobAttribute, but typically attributes are a many-to-one property when applied to items (generically speaking).

wbbradley: I see `BlobAttribute` is a mirror of `BlobMetadata` in move. I wonder if we should consider renaming `BlobAttribute` to `BlobAttributes`, since it's not a singular attribute.

liquid-helium: Good question, there is a follow-up plan to unify the representations of blobs, (WAL-962), including, unencoded blob, QuiltStoreBlob, and now a blob with a mutable attribute.

For now, I will keep it as, and refactor it later.

crates/walrus-sdk/src/client.rs
liquid-helium: There is plan to support them in CLI commands, changes will be made to `reserve_and_store_blobs_retry_committees_with_path` together with that.

PR #2402, opened 2025-08-04
docs: update operator lists and improve cache inference script

mlegner:
## Description

- Also infer cache status based on absolute latency (in addition to speedup).
- Mark aggregators with errors with `functional: false` in operator lists.
- Add `remove-on-error` flag to cache inference script to completely remove aggregators with errors from operator lists.
- Add linting to cache inference script.
- Update dependencies.

## Test plan

Ran the script locally.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -24,7 +24,7 @@ indent_size = unset
 max_line_length = unset
 trim_trailing_whitespace = unset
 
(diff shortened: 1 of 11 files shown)


wbbradley:
This seems relatively easy to game, but I'm not sure whether anybody would be motivated to do so. The code change looks good to me, although I'm not a TS expert.

wbbradley (approved):

PR #2404, opened 2025-08-04
feat: Offline list checkpoints

liquid-helium:
## Description

Replaced the `local-admin checkpoint list` with walrus-node `list-dbcheckpoint`.
Close WAL-949
## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/db_checkpoint.rs
+++ b/crates/walrus-service/src/node/db_checkpoint.rs
@@ -398,25 +398,6 @@ impl DbCheckpointManager {
         Ok(result)
     }
 
-    /// List all db_checkpoints in the db_checkpoint directory.
-    ///
-    /// If no db_checkpoint_dir is provided, the directory configured in DbCheckpointConfig will be
-    /// used. If none of these are provided an error will be returned.
-    pub fn list_db_checkpoints(
-        &self,
-        db_checkpoint_dir: Option<&Path>,
-    ) -> Result<Vec<DisplayableDbCheckpointInfo>, DbCheckpointError> {
-        let db_checkpoint_dir = db_checkpoint_dir
-            .or(self.config.db_checkpoint_dir.as_deref())
-            .ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


sadhansood (commented):
@liquid-helium - Thanks for doing this! Does this also list the running backups? Also, what happens if I delete a particular backup from the checkpoint dir? Will it catch this? I remember you mentioned that files are linked across backups. Maybe, worth adding to the comments for this command.

crates/walrus-service/bin/node.rs
```diff
@@ -1130,6 +1133,54 @@ mod commands {
...
 
+    #[tokio::main]
+    pub(crate) async fn list_db_checkpoints(
+        ListDbCheckpointArgs { path }: ListDbCheckpointArgs,
+    ) -> anyhow::Result<()> {
+        use rocksdb::{
+            Env,
+            backup::{BackupEngine, BackupEngineOptions},
+        };
+        use walrus_service::DisplayableDbCheckpointInfo;
```
halfprice, line 1144: question: why put these here instead of at the beginning of the file?

crates/walrus-service/bin/node.rs
```diff
@@ -1541,26 +1592,6 @@ async fn handle_checkpoint_command(
                 },
             }
         }
-        CheckpointCommands::List { path } => {
-            let result = manager.list_db_checkpoints(path.as_deref());
```
halfprice, line 1545: Nit: the list_db_checkpoints in DbCheckpointManager can also be removed

crates/walrus-service/bin/node.rs
liquid-helium: Since` list_db_checkpoints` is only called once at a single place, it makes sense to put the imports local. Keep it as, but no strong preference.

crates/walrus-service/bin/node.rs
liquid-helium: Done.

PR #2405, opened 2025-08-04
feat: added list-patches-in-quilt aggregator api.

liquid-helium:
## Description

Added new endpoint in aggregator: ` /quilts/<quilt-id>/patches` to list all patches in a quilt.

Contributes to WAL-960

## Test plan

---

## Release notes

- [ ] Storage node:
- [x] Aggregator:
  Added new endpoint in aggregator:  /quilts/\<quilt-id\>/patches to list all patches in a quilt.
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -78,6 +78,8 @@ pub const QUILT_PATCH_BY_ID_GET_ENDPOINT: &str = "/v1/blobs/by-quilt-patch-id/{q
 /// The path to get blob from quilt by quilt ID and identifier.
 pub const QUILT_PATCH_BY_IDENTIFIER_GET_ENDPOINT: &str =
     "/v1/blobs/by-quilt-id/{quilt_id}/{identifier}";
+/// The path to list patches in a quilt.
+pub const LIST_PATCHES_IN_QUILT_ENDPOINT: &str = "/v1/quilts/{quilt_id}/patches";
```
giac-mysten, line 83: Question: I'm wondering if something like `/v1/quilts/patches-by-id/{quild_id}` would be closer to the structure of the other endpoints.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -672,6 +674,89 @@ pub(super) async fn get_blob_by_quilt_id_and_identifier<T: WalrusReadClient>(
...
+}
+
+/// List patches in a quilt.
+///
+/// Returns a list of identifiers and QuiltPatchIds for all patches contained in the specified
+/// quilt.
+///
+/// # Example
+/// ```bash
+/// curl -X GET "http://localhost:31415/v1/quilts/\
```
giac-mysten: Remark: keep the examples in sync with the result of the API discussion.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -672,6 +674,89 @@ pub(super) async fn get_blob_by_quilt_id_and_identifier<T: WalrusReadClient>(
...
+///   }
+/// ]
+/// ```
+#[tracing::instrument(level = Level::ERROR, skip_all)]
+#[utoipa::path(
+    get,
+    path = LIST_PATCHES_IN_QUILT_ENDPOINT,
+    params(
+        (
+            "quilt_id" = String, Path,
```
giac-mysten: Question: Should this be `BlobIdString`? If not, why `String, Path`?
```suggestion
            "quilt_id" = BlobIdString,
```
Related, are then the params in the GET endpoint for `get_blob` (line `95`) correct?
Should that also be changed?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -672,6 +674,89 @@ pub(super) async fn get_blob_by_quilt_id_and_identifier<T: WalrusReadClient>(
...
+            let error = GetBlobError::from(error);
+
+            match &error {
+                GetBlobError::BlobNotFound => {
+                    tracing::debug!(?quilt_id, "the requested quilt ID does not exist")
+                }
+                GetBlobError::Internal(error) => {
+                    tracing::error!(?error, "error retrieving quilt patches")
+                }
+                _ => (),
```
giac-mysten, line 747: Question: Should we have an `unreachable!` here?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -672,6 +674,89 @@ pub(super) async fn get_blob_by_quilt_id_and_identifier<T: WalrusReadClient>(
...
+                QuiltPatchIds. Each QuiltPatchId can be used with the \
+                /v1/blobs/by-quilt-patch-id endpoint to retrieve the actual blob data.",
+)]
+pub(super) async fn list_patches_in_quilt<T: WalrusReadClient>(
+    State(client): State<Arc<T>>,
+    Path(BlobIdString(quilt_id)): Path<BlobIdString>,
+) -> Response {
+    tracing::debug!("starting to list patches in quilt: {}", quilt_id);
+
+    match client.list_patches_in_quilt(&quilt_id).await {
```
giac-mysten, line 735: Question: Just want to check my understanding; IIUC, this call checks the blocklist, right?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -672,6 +674,89 @@ pub(super) async fn get_blob_by_quilt_id_and_identifier<T: WalrusReadClient>(
     }
 }
 
+/// Response item for a patch in a quilt.
+#[derive(Debug, Serialize, Deserialize, utoipa::ToSchema)]
+pub struct QuiltPatchItem {
+    /// The identifier of the patch (e.g., filename).
+    pub identifier: String,
+    /// The QuiltPatchId for this patch, encoded as URL-safe base64.
+    pub patch_id: String,
```
giac-mysten: Question: Why the string types? can't the serialization take care of converting to the correct string format?

giac-mysten (commented):
Thanks @liquid-helium ! A few comments from my side, but looks awesome overall.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Yeah, that's a good idea, change applied.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Done.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Nice catch, `String` is technically correct, but `BlobId` provides more accurate OpanApi docs.
BlobIdString doesn't work due to toSchema trait constraint, and `BlobId` is more accurate.
Changed here and other places.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: There can be `Blocked` error type I think. I keep it as.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Yes.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Changed to QuiltPatchId, nice catch!

crates/walrus-core/src/lib.rs
```diff
@@ -198,6 +198,7 @@ impl Debug for BlobId {
 /// A quilt patch is a individual blob within a quilt.
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Serialize, Deserialize, Hash)]
 #[serde(rename_all = "camelCase")]
+#[cfg_attr(feature = "utoipa", derive(utoipa::ToSchema))]
 pub struct QuiltPatchId {
```
mlegner, line 202: Important: We need to explicitly add the schema to the specs in `crates/walrus-service/src/client/daemon/openapi.rs`. In fact, I realized that we are already missing the `ObjectIdSchema` in the aggregator... 😬

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -78,6 +79,8 @@ pub const QUILT_PATCH_BY_ID_GET_ENDPOINT: &str = "/v1/blobs/by-quilt-patch-id/{q
 /// The path to get blob from quilt by quilt ID and identifier.
 pub const QUILT_PATCH_BY_IDENTIFIER_GET_ENDPOINT: &str =
     "/v1/blobs/by-quilt-id/{quilt_id}/{identifier}";
+/// The path to list patches in a quilt.
+pub const LIST_PATCHES_IN_QUILT_ENDPOINT: &str = "/v1/quilts/patches-by-id/{quilt_id}";
```
mlegner: Nit: From the path it's a bit unclear what ID is provided. Maybe `patches-by-quilt-id`?

mlegner (approved):
Thanks a lot for the addition, @liquid-helium. LGTM.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Alternative take: I'd say this endpoint should be `/v1/quilts/{quilt_id}/patches` to be more RESTful.

crates/walrus-core/src/lib.rs
liquid-helium: Done.

crates/walrus-service/src/client/daemon/routes.rs
liquid-helium: Done.

PR #2406, opened 2025-08-04
fix: Eliminate redundant recovery symbol computations

sadhansood:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


mlegner:
@jpcsmith I just looked at the encoding; unfortunately, it seems that the underlying library doesn't support computing a single repair symbol. However, what can be done relatively easily is skipping the encoding entirely if we only need some source symbols (including the case where we just need a single one). For individual symbols, this would be relevant in around half the cases (1/3 for secondary slivers, 2/3 for primary slivers). I can look into this today or tomorrow if you think it makes sense.

sadhansood:
@jpcsmith your proposal makes sense. If I understood it correctly, one way to do that would be to land this PR and @mlegner will optimize the current logic such that we pass in the full list of recovery symbols to the cache miss closure if the requested symbol is not one of the source symbols (since it would require us to compute the full list recovery symbols).

jpcsmith:
@sadhansood An approach that would avoid needing to revert changes, would be if this PR could already just check if the symbol would be one of the ones that require us to encode the sliver, and in that case, use the code you added, otherwise, use the previous code. Then, Markus' optimisation would not require further changes to this module.

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,179 @@ impl RecoverySymbolService {
...
-            .map_err(Arc::unwrap_or_clone)?;
 
-        let proof = merkle_tree
-            .get_proof(target_sliver_index.as_usize())
-            .expect("index is valid as it was valid to get the decoding symbol");
-
-        let recovery_symbol = match decoding_symbol {
-            ByAxis::Primary(symbol) => ByAxis::Primary(symbol.with_proof(proof)),
-            ByAxis::Secondary(symbol) => ByAxis::Secondary(symbol.with_proof(proof)),
+        let recovery_symbol = match req.source_sliver {
```
jpcsmith: Suggestion: Split the body into two methods: `handle_request_for_source_symbol` and `handle_request_for_recovery_symbol` and move the respective logic there.

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,179 @@ impl RecoverySymbolService {
...
+                    req.target_pair_index.to_sliver_index::<Secondary>(n_shards);
+
+                let is_source_target =
+                    Self::is_source_target_for_axis::<Secondary>(&config, target_sliver_index);
+
+                if is_source_target {
+                    let decoding_symbol = Self::make_source_decoding_symbol::<Primary>(
+                        sliver.index,
+                        sliver.symbols[target_sliver_index.as_usize()].to_vec(),
+                    );
```
jpcsmith: Suggestion: Use the `.decoding_symbol_for_sliver(...)` method and add the optimisation there, since you're already adding it.

jpcsmith (changes requested):

crates/walrus-service/src/node/recovery_symbol_service.rs
sadhansood: done

crates/walrus-service/src/node/recovery_symbol_service.rs
sadhansood: done

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,184 @@ impl RecoverySymbolService {
...
 
+    fn proof_from_cache_or_build<F>(
+        &self,
+        cache_key: CacheKey,
+        target_index: usize,
+        miss_total: &IntCounter,
+        build_tree: F,
+    ) -> Result<walrus_core::merkle::MerkleProof<Blake2b256>, RecoverySymbolError>
+    where
+        F: FnOnce() -> Result<Arc<MerkleTree<Blake2b256>>, RecoverySymbolError>,
```
jpcsmith:
```suggestion
        recovery_symbols: F,
    ) -> Result<walrus_core::merkle::MerkleProof<Blake2b256>, RecoverySymbolError>
    where
        F: FnOnce() -> Result<Symbols, RecoverySymbolError>,
```
- The variable `miss_total` does not need to be passed into the method, as it is not used in any of the callers. It can be directly fetched in this method with:
  ```rust
  let miss_total = &self.metrics.cache_miss_total;
  ```
- The `build_tree` function can be reduced in scope to simply a function that returns the recovery symbols (either directly or by creating them). The rest of the building logic for the merkle tree can be placed directly in this method (see below).

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,184 @@ impl RecoverySymbolService {
...
+            return Ok(tree
+                .get_proof(target_index)
+                .expect("bound already checked above"));
+        }
+
+        let tree = self
+            .cache
+            .try_get_with::<_, RecoverySymbolError>(cache_key.clone(), || {
+                miss_total.inc();
+                build_tree()
```
jpcsmith:
```suggestion
                let tree = MerkleTree::<Blake2b256>::build(recovery_symbols()?.to_symbols());
                Ok(Arc::new(tree))
```

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,184 @@ impl RecoverySymbolService {
...
+            })
+            .map_err(Arc::unwrap_or_clone)?;
+
+        Ok(tree
+            .get_proof(target_index)
+            .expect("bound already checked above"))
+    }
+
+    fn handle_request_for_source_symbol<SourceAxis: walrus_core::encoding::EncodingAxis>(
```
jpcsmith:
```suggestion
    fn handle_request_for_source_symbol<T: EncodingAxis>(
```
Suggestion: Directly import the needed types instead of using the full pathname, here and for the other arguments.

Suggestion: The naming convention for generic parameters is a single capital letter, unless it significantly helps with readability. Here I would say that a `T` would be sufficient, as it's clear it's the source axis as it's used only as `SliverData<T>` in the method signature.

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,184 @@ impl RecoverySymbolService {
...
+            DecodingSymbol::<SourceAxis::OrthogonalAxis>::new(sliver_index.get(), symbol_bytes);
+
+        let proof = self.proof_from_cache_or_build(
+            cache_key,
+            target_sliver_index.as_usize(),
+            miss_total,
+            move || {
+                let recovery_symbols = sliver.recovery_symbols(config)?;
+                let tree = MerkleTree::<Blake2b256>::build(recovery_symbols.to_symbols());
+                Ok(Arc::new(tree))
```
jpcsmith: Suggestion: This can be simplified to just:
```suggestion
                sliver.recovery_symbols(config)
```
with the replicated code directly being in `proof_from_cache_or_build` (see there for more details).

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,184 @@ impl RecoverySymbolService {
...
+    ) -> Result<GeneralRecoverySymbol, RecoverySymbolError>
+    where
+        walrus_core::encoding::DecodingSymbol<
+            <SourceAxis as walrus_core::encoding::EncodingAxis>::OrthogonalAxis,
+        >: Into<walrus_core::encoding::EitherDecodingSymbol>,
+    {
+        let symbol_bytes = sliver.symbols[target_sliver_index.as_usize()].to_vec();
+        let sliver_index = sliver.index;
+        let decoding_symbol =
+            DecodingSymbol::<SourceAxis::OrthogonalAxis>::new(sliver_index.get(), symbol_bytes);
```
jpcsmith: Suggestion: Please move the optimisation directly into the walrus-core crate in the `decoding_symbol_for_sliver` method, which could then be used directly here. This currently reimplements logic that should be in the walrus-core crate.

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,184 @@ impl RecoverySymbolService {
...
+            )
+            .expect("we have exactly `n_shards` symbols and the bound was checked");
+
+        let proof = self.proof_from_cache_or_build(
+            cache_key,
+            target_sliver_index.as_usize(),
+            miss_total,
+            move || {
+                let tree = MerkleTree::<Blake2b256>::build(recovery_symbols.to_symbols());
+                Ok(Arc::new(tree))
```
jpcsmith: Suggestion: Can be simplified to 
```suggestion
                recovery_symbols
```
with the suggestions in proof_from_cache_or_build

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,184 @@ impl RecoverySymbolService {
...
+        )?;
+
+        Ok(GeneralRecoverySymbol::from_recovery_symbol(
+            decoding_symbol.with_proof(proof),
+            target_sliver_index,
+        ))
+    }
+
     fn handle_request_and_cache(
         &mut self,
```
jpcsmith: Remark: method does not seem like it needs to be &mut.

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -113,53 +120,184 @@ impl RecoverySymbolService {
...
+                let target_sliver_index =
+                    req.target_pair_index.to_sliver_index::<Secondary>(n_shards);
+
+                let is_source_target =
+                    Self::is_source_target_for_axis::<Secondary>(&config, target_sliver_index);
+
+                if is_source_target {
+                    self.handle_request_for_source_symbol::<Primary>(
+                        sliver,
+                        &config,
```
jpcsmith: Suggestion: Deduplicate this code. You can create a generic method as follows

```rust
    fn get_recovery_symbol<T: EncodingAxis>(
        &self,
        cache_key: CacheKey,
        sliver: SliverData<T>,
        config: &EncodingConfigEnum,
    ) -> Result<GeneralRecoverySymbol, RecoverySymbolError>
    where
        DecodingSymbol<T::OrthogonalAxis>: Into<EitherDecodingSymbol>,
    {
        let n_shards = config.n_shards();

        let target_sliver_index = target_pair_index.to_sliver_index::<T::OrthogonalAxis>(n_shards);
        let is_source_target =
            Self::is_source_target_for_axis::<T::OrthogonalAxis>(&config, target_sliver_index);

        if is_source_target {
            self.handle_request_for_source_symbol::<T>(
                sliver,
                &config,
                cache_key,
                target_sliver_index,
            )
        } else {
            self.handle_request_for_recovery_symbol::<T>(
                sliver,
                &config,
                cache_key,
                target_sliver_index,
            )
        }
    }
```

Then the remainder of this method body can be simplified to dispatching on the generic type:
```rust
        match req.source_sliver {
            ByAxis::Primary(sliver) => {
                self.get_recovery_symbol(key, sliver, req.target_pair_index, &config)
            }
            ByAxis::Secondary(sliver) => {
                self.get_recovery_symbol(key, sliver, req.target_pair_index, &config)
            }
        }
```

crates/walrus-core/src/encoding/slivers.rs
jpcsmith, line 1: Important: There is no need to change the public API of the crate, we can make this optimisation as merely an internal change:

```rust
    /// Gets the decoding symbol for a specific target sliver starting from the current sliver.
    ///
    /// The `target_pair_index` is the index of the [`SliverPair`] to which the sliver to be
    /// recovered belongs.
    ///
    /// If the target symbol is one of the symbols present in the sliver, then it is directly
    /// returned, otherwise, the sliver is encoded to create the associated symbol.
    ///
    /// # Errors
    ///
    /// Returns a [`RecoverySymbolError::EncodeError`] if the sliver cannot be encoded. Returns a
    /// [`RecoverySymbolError::IndexTooLarge`] error if `target_pair_index >= n_shards`.
    pub fn decoding_symbol_for_sliver(
        &self,
        config: &EncodingConfigEnum,
    ) -> Result<DecodingSymbol<T::OrthogonalAxis>, RecoverySymbolError> {
        Self::check_index(target_pair_index.into(), config.n_shards())?;
        let target_sliver_index =
            target_pair_index.to_sliver_index::<T::OrthogonalAxis>(config.n_shards());

        let decoding_symbol = if target_sliver_index.get() < config.n_source_symbols::<T>().get() {
            DecodingSymbol::new(
                self.index.get(),
                self.symbols[target_sliver_index.as_usize()].to_vec(),
            )
        } else {
            self.recovery_symbols(config)?
                .decoding_symbol_at(target_sliver_index.as_usize(), self.index.into())
                .expect("we have exactly `n_shards` symbols and the bound was checked")
        };

        Ok(decoding_symbol)
    }
```

crates/walrus-core/src/encoding/symbols.rs
```diff
@@ -279,6 +279,33 @@ impl AsMut<[u8]> for Symbols {
...
+    }
+
+    /// Creates a new `DecodingSymbolEnvelope` with no recovery symbols.
+    pub fn new_without_symbols(symbol: DecodingSymbol<T>) -> Self {
+        Self {
+            symbol,
+            recovery_symbols: None,
+        }
+    }
+}
```
jpcsmith: Suggestion: This is a neat construct that simplifies the branching in the walrus-service crate, move this into the recovery_symbol_service module instead of changing the api of walrus-core. If it turns out to be something needed by others, we can make it back here, make it public, and commit to maintaining the interface; but for now, let's keep it internal.

crates/walrus-core/src/encoding/slivers.rs
sadhansood: Done, added this as another method and the current `decoding_symbol_for_sliver` one just uses this internally.

crates/walrus-service/src/node/recovery_symbol_service.rs
sadhansood: done

PR #2407, opened 2025-08-04
feat: Add metric to track negative point lookups and granular data usage

sadhansood:
## Description

This PR introduces some granular metrics for typed-store to provide better visibility into database operations, especially for point lookups:
1. Negative Lookup Metric to allows distinguishing between lookups for existing keys and non-existing keys (negative lookups). Added to `contains_key` and `get`
2. Metrics to track both `key` and `value` bytes separately which offers deeper insight into data usage pattern.
3. The logic for managing iterator-related metrics is refactored into a new IterContext struct.

## Test plan

Existing tests

### DIFF
--- a/crates/typed-store/src/metrics.rs
+++ b/crates/typed-store/src/metrics.rs
@@ -336,6 +336,10 @@ impl ColumnFamilyMetrics {
 pub struct OperationMetrics {
     /// Rocksdb iter latency in seconds
     pub rocksdb_iter_latency_seconds: HistogramVec,
+    /// Rocksdb iter key bytes
+    pub rocksdb_iter_key_bytes: HistogramVec,
+    /// Rocksdb iter value bytes
+    pub rocksdb_iter_value_bytes: HistogramVec,
     /// Rocksdb iter size in bytes
     pub rocksdb_iter_bytes: HistogramVec,
     /// Rocksdb iter num keys
@@ -344,6 +348,12 @@ pub struct OperationMetrics {
     pub rocksdb_get_latency_seconds: HistogramVec,
     /// Rocksdb get bytes
     pub rocksdb_get_bytes: HistogramVec,
+    /// Rocksdb get key bytes
+    pub rocksdb_get_key_bytes: HistogramVec,
+    /// Rocksdb get value bytes
+    pub rocksdb_get_value_bytes: HistogramVec,
+    /// Rocksdb contains_key latency in seconds
+    pub rocksdb_contains_key_latency_seconds: HistogramVec,
     /// Rocksdb multiget latency in seconds
     pub rocksdb_multiget_latency_seconds: HistogramVec,
     /// Rocksdb multiget bytes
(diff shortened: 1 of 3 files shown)


crates/typed-store/src/metrics.rs
```diff
@@ -336,24 +336,36 @@ impl ColumnFamilyMetrics {
...
+    /// Rocksdb contains_key latency in seconds
+    pub rocksdb_contains_key_latency_seconds: HistogramVec,
     /// Rocksdb multiget latency in seconds
     pub rocksdb_multiget_latency_seconds: HistogramVec,
     /// Rocksdb multiget bytes
     pub rocksdb_multiget_bytes: HistogramVec,
     /// Rocksdb put latency in seconds
     pub rocksdb_put_latency_seconds: HistogramVec,
-    /// Rocksdb put bytes
-    pub rocksdb_put_bytes: HistogramVec,
```
liquid-helium, line 354: Question do we need to wait for at least one release before removing these metrics, if some dashboards are using these metrics, it could break them when this one is released?

crates/typed-store/src/metrics.rs
```diff
@@ -406,20 +426,36 @@ impl OperationMetrics {
             .unwrap(),
             rocksdb_get_latency_seconds: register_histogram_vec_with_registry!(
                 "rocksdb_get_latency_seconds",
-                "Rocksdb get latency in seconds",
-                &["cf_name"],
+                "Rocksdb get latency in seconds. `found` label is true if key was found",
```
liquid-helium, line 459: Same here, will this break any dashboard?

crates/typed-store/src/rocks.rs
```diff
@@ -1415,9 +1430,14 @@ where
...
             .op_metrics
-            .rocksdb_put_bytes
+            .rocksdb_put_key_bytes
+            .with_label_values(&[&self.cf])
+            .observe(key_buf.len() as f64);
+        self.db_metrics
+            .op_metrics
+            .rocksdb_put_value_bytes
             .with_label_values(&[&self.cf])
-            .observe((key_buf.len() + value_buf.len()) as f64);
```
halfprice, line 1483: suggestion: I suspect that the old key+value metrics is still valuable, since this cannot be simply derived from key and value metrics. Can we keep `rocksdb_put_bytes`? In this way, we also don't break any existing dashboard.

crates/typed-store/src/rocks/safe_iter.rs
```diff
@@ -88,7 +89,8 @@ impl<K: DeserializeOwned, V: DeserializeOwned> Iterator for SafeIter<'_, K, V> {
                 .db_iter
                 .value()
                 .expect("valid iterator should be able to get value");
-            self.bytes_scanned_counter += raw_key.len() + raw_value.len();
```
halfprice, line 91: suggestion: similarly, keep the old bytes scanned.

sadhansood:
@liquid-helium @halfprice - thank you for your review. Now, i am only adding new metrics but not removing anything already existing.

PR #2411, opened 2025-08-06
fix(client): add HTML error codes to retriable RPC errors

giac-mysten:
## Description

The RPC can sometimes timeout if it is under load or during epoch change. It is important to retry within the retry client to ensure that the same transaction is repeated (and therefore no locked objects can occur).

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client/retry_client/retriable_rpc_error.rs
+++ b/crates/walrus-sui/src/client/retry_client/retriable_rpc_error.rs
@@ -8,7 +8,7 @@ use std::fmt::Debug;
 use super::{super::SuiClientError, CheckpointRpcError, FallbackError, RetriableClientError};
 
 /// The list of HTTP status codes that are retriable.
-const RETRIABLE_RPC_ERRORS: &[&str] = &["408", "429", "500", "502", "504"];
+const RETRIABLE_RPC_ERRORS: &[&str] = &["408", "429", "500", "502", "503", "504"];
 /// The list of gRPC status codes that are retriable.
 const RETRIABLE_GRPC_ERRORS: &[tonic::Code] = &[
     tonic::Code::ResourceExhausted,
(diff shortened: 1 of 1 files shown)


mlegner (commented):
Thanks for the PR, @giac-mysten! Let's see if we can test this somehow and then get this merged soon.

crates/walrus-sui/src/client/retry_client/retriable_rpc_error.rs
```diff
@@ -8,7 +8,7 @@ use std::fmt::Debug;
 use super::{super::SuiClientError, CheckpointRpcError, FallbackError, RetriableClientError};
 
 /// The list of HTTP status codes that are retriable.
-const RETRIABLE_RPC_ERRORS: &[&str] = &["429", "500", "502"];
+const RETRIABLE_RPC_ERRORS: &[&str] = &["408", "429", "500", "502", "504"];
```
mlegner: Maybe add 503 as well?
```suggestion
const RETRIABLE_RPC_ERRORS: &[&str] = &["408", "429", "500", "502", "503", "504"];
```

giac-mysten: have we seen 503?
"service unavailable" seems like quite a strong signal we should not retry?

mlegner: I haven't seen it. But according to the [Mozilla docs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/503), this should be used for a temporary outage, for example due to upgrades, maintenance, etc. So it makes sense to retry this IMO.

sadhansood (approved):

PR #2414, opened 2025-08-07
fix: test_store_blob_with_random_attributes.

liquid-helium:
## Description

The assert! didn't cover the empty attributes.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


halfprice (approved):
Thanks @liquid-helium for the quick fix!

PR #2416, opened 2025-08-09
chore: fix typo in ws-resources.json filename

daoauth:
## Description

The documentation incorrectly refers to `ws-resource.json`,  
but the correct filename is `ws-resources.json`.  
This PR fixes the typo.


mlegner (approved):
Thanks for the fix to avoid confusion. 🙏

PR #2419, opened 2025-08-11
feat(client): split storage to size and epoch before storing

giac-mysten:
## Description

Ensures that, when reusing existing storage resources, the client splits the storage to the exact size both in amount of storage and number of epochs.

Closes WAL-208
Contributes to  WAL-363

## Test plan

Added relevant tests to check the splits when reusing storage.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -423,14 +423,13 @@ async fn test_store_with_existing_blob_resource(
 
     let blob_data = walrus_test_utils::random_data_list(31415, 4);
     let blobs: Vec<&[u8]> = blob_data.iter().map(AsRef::as_ref).collect();
-    let encoding_type = DEFAULT_ENCODING;
     let metatdatum = blobs
         .iter()
         .map(|blob| {
             let (_, metadata) = client
                 .as_ref()
                 .encoding_config()
-                .get_for_type(encoding_type)
+                .get_for_type(DEFAULT_ENCODING)
                 .encode_with_metadata(blob)
                 .expect("blob encoding should not fail");
             let metadata = metadata.metadata().to_owned();
@@ -462,8 +461,7 @@ async fn test_store_with_existing_blob_resource(
         .collect::<HashMap<_, _>>();
 
     // Now ask the client to store again.
-    let store_args =
-        StoreArgs::default_with_epochs(epochs_ahead_required).with_encoding_type(encoding_type);
+    let store_args = StoreArgs::default_with_epochs(epochs_ahead_required);
     let blob_stores = client
         .inner
         .reserve_and_store_blobs(&blobs, &store_args)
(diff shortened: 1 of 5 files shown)


crates/walrus-sui/src/client.rs
```diff
@@ -600,12 +600,13 @@ impl SuiContractClient {
         &self,
         blob_metadata_and_storage: Vec<(BlobObjectMetadata, StorageResource)>,
         persistence: BlobPersistence,
+        target_epoch: Option<Epoch>,
     ) -> SuiClientResult<Vec<Blob>> {
```
liquid-helium, line 604: nit: will there be calls of register_blobs with target_epoch = None, expect in tests? Does it make sense to use Epoch instead of Option<Epoch>?

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -746,19 +743,34 @@ async fn test_store_with_existing_storage_resource(
...
         .collect::<Vec<_>>();
 
+    // The encoded sizes should all be the same.
+    assert!(encoded_sizes.iter().all(|&size| size == encoded_sizes[0]));
+    let encoded_size = encoded_sizes[0];
+
     // Reserve space for the blobs. Collect all original storage resource objects ids.
+    // Each storage object has `EXTRA_BYTES_PER_STORAGE` bytes reserved, such that we can later
+    // check they were correctly split off.
+    const EXTRA_BYTES_PER_STORAGE: u64 = 2718;
```
liquid-helium, line 753: Praise: Thanks for adding the tests!

I am wondering if we can test with spaces of different sizes pre-reserved: encoded_size - x, encoded_size, encoded_size + x. so that we can cover the case that some blobs reuse spaces, some registered new ones.

crates/walrus-sui/src/client.rs
```diff
@@ -1581,16 +1582,63 @@ impl SuiContractClientInner {
...
+
     /// Registers a blob with the specified [`BlobId`] using the provided [`StorageResource`],
     /// and returns the created blob object.
     ///
     /// `blob_size` is the size of the unencoded blob. The encoded size of the blob must be
     /// less than or equal to the size reserved in `storage`.
+    ///
+    /// Storage resources are automatically split to the correct size and epoch if needed.
     #[tracing::instrument(level = Level::DEBUG, skip_all)]
     pub async fn register_blobs(
```
karlwuest, line 1637: Remark/Suggestion:
I'm not sure if this is the main function that we should be concerned with. 
We use this function mainly for tests, but as I've noticed now, we also use it in `ResourceManager::get_existing_or_register_with_resources` as separate call in addition to `reserve_and_register`, which we should not do, because that means potentially sending two transactions instead of one. 

I think the main focus should be on `reserve_and_register` to ensure that existing storage is reused if possible (e.g. by passing a vector of storage resources that can then be merged & split, potentially combined with new storage), which should then also reduce the complexity in the `ResourceManager`. 

We can then also do the splitting in this function here, but then I'd prefer an approach where instead of passing pairs of `(BlobObjectMetadata, StorageResource)`, we pass two independent vectors (or slices) of `BlobObjectMetadata` and `StorageResource`, which then allows the storage resources to also be combined instead of just split. This would then also allow using the same logic here as in `reserve_and_register` for splitting & merging (modulo buying new storage, which would just cause an error here if that were needed).

mlegner:
@giac-mysten Should we resurrect this PR?

wbbradley:
Closing due to lack of action. Can be reopened if needed.

PR #2423, opened 2025-08-12
fix: add simple checks to avoid unnecessary computations where possible

mlegner:
## Description

This PR ensures that we only perform cryptographic validation of inconsistency proofs and recovery symbols after performing some checks on the size of them and their proofs.

Also adds some clarifying comments and docstrings and cleans up no-longer-used functions.

## Test plan

Existing/adapted tests.

### DIFF
--- a/crates/walrus-core/src/encoding/config.rs
+++ b/crates/walrus-core/src/encoding/config.rs
@@ -110,11 +110,9 @@ pub trait EncodingConfigTrait {
     }
 
     /// Returns the number of valid symbols required to recover a sliver of [`EncodingAxis`] `T`.
-    // Important: This is an *exact* value. If a future encoding requires a variable number of
-    // symbols, the interface here needs to be changed.
     #[inline]
-    fn n_symbols_for_recovery<T: EncodingAxis>(&self) -> usize {
-        self.n_source_symbols::<T::OrthogonalAxis>().get().into()
+    fn n_symbols_for_recovery<T: EncodingAxis>(&self) -> RequiredSymbolsCount {
+        RequiredSymbolsCount::Exact(self.n_source_symbols::<T::OrthogonalAxis>().get().into())
     }
 
     /// Returns the number of shards as a `usize`.
@@ -241,6 +239,20 @@ pub trait EncodingConfigTrait {
     fn max_sliver_size(&self) -> u64 {
         max_sliver_size_for_n_secondary(self.n_secondary_source_symbols(), self.encoding_type())
     }
+}
+
+/// The number of symbols required to recover a sliver.
+///
+/// For our current RS2 encoding, which is based on Reed-Solomon codes, the number of symbols
+/// required to recover a sliver is exactly the number of source symbols. This could be different
+/// for future encodings that use other types of erasure codes.
+// Important: Currently, quite some code assumes that the number of symbols required to recover a
+// sliver is an exact number. This code needs to be changed if this enum is extended in the future.
+// By explicitly matching on `Exact`, we ensure that the compiler will warn us in that case.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RequiredSymbolsCount {
+    /// An exact number of symbols required to recover a sliver.
+    Exact(usize),
 }
 
 /// Configuration parameters for the encoding.
--- a/crates/walrus-core/src/encoding/config.rs
+++ b/crates/walrus-core/src/encoding/config.rs
@@ -109,6 +109,14 @@ pub trait EncodingConfigTrait {
         }
     }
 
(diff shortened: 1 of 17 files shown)


crates/walrus-service/src/node/committee/request_futures.rs
```diff
@@ -295,9 +296,14 @@ where
...
-        // Since recovery currently consumes the symbols, rather than copy the symbols in every
-        // case to handle the rare cases when we fail to *decode* the sliver despite collecting the
+        // Since recovery currently consumes the symbols, rather than copy the symbols in every case
+        // to handle the rare cases when we fail to *decode* the sliver despite collecting the
         // required number of symbols, we instead retry the entire process with an increased amount.
+        //
+        // Remark: With the current RS2 encoding, we know the exact number of symbols required to
+        // decode the sliver, so this loop would not be necessary. However, future encodings may
+        // require a variable number of symbols to decode the sliver, in which case this loop is
+        // actually needed.
```
karlwuest: Suggestion: Should we just remove the loop and the `recover_with_additional_symbols` function, since we currently only use RS2 and we don't have concrete plans to add other encodings? Adding another encoding would anyway require additional changes in other locations (as also mentioned in a few other comments that you added in this PR) and I'm not sure if it makes sense to keep this in just in case we need it at some point in the future.

crates/walrus-core/src/merkle.rs
```diff
@@ -311,6 +342,14 @@ fn n_nodes(n_leaves: usize) -> usize {
...
+/// Returns the path length of a Merkle tree with `n_leaves` leaves.
+fn path_length(n_leaves: usize) -> usize {
+    if n_leaves <= 1 {
+        return 0;
+    }
+    usize::try_from((n_leaves - 1).ilog2()).expect("this is smaller than `n_leaves`") + 1
+}
+
 #[cfg(test)]
 mod test {
```
karlwuest, line 354: Nit: Add a test for `path_length` (e.g. for 4, 5, 8 leaves).

crates/walrus-service/src/node/committee/request_futures.rs
mlegner: Yes, good idea. Any objections, @jpcsmith?

crates/walrus-core/src/inconsistency.rs
```diff
@@ -114,6 +127,44 @@ impl<T: EncodingAxis, U: MerkleAuth> InconsistencyProof<T, U> {
         }
     }
 
+    /// Checks that the proof contains the correct number of recovery symbols and that they are
+    /// valid.
+    ///
+    /// By checking the recovery symbols, we also check that the metadata
```
liquid-helium: nit: is this an imcomplete sentense?

crates/walrus-core/src/encoding/config.rs
```diff
@@ -109,6 +109,14 @@ pub trait EncodingConfigTrait {
         }
     }
```
liquid-helium, line 111: Question: Does it make sense to add a function, `fn exact_n_symbols_for_recovery(&self) -> bool`, so that this function will be used only by those return true, or we can assert this condition inside the related functions?

crates/walrus-service/src/node/committee/request_futures.rs
jpcsmith: Sounds good.

crates/walrus-core/src/encoding/config.rs
mlegner: Good point. I actually went ahead and created a new `RequiredSymbolsCount` enum in cdcdbe22512ec22095b5751be7eb44e493348935, which currently just has a single variant `Exact`. This ensures that we revisit relevant code if we add an encoding type with variable symbols count. PTAL. 🙏

crates/walrus-service/src/node/committee/request_futures.rs
mlegner: OK, I removed the "additional symbols" parts but kept the loop because I didn't want to panic. Instead, I now retry after logging an error if the decoding fails.

crates/walrus-core/src/merkle.rs
mlegner: Done.

crates/walrus-core/src/encoding/config.rs
mlegner: @liquid-helium I'm going to merge this as is, please let me know if you want me to change anything, then I'll take it on in a follow-up.

liquid-helium: Sorry @mlegner for the slowness, yes, I think the way you addressed it is even better! LGTM!

PR #2432, opened 2025-08-14
chore: bump sui dep version to testnet-v1.54.0

wbbradley:
## Description

Bump sui dep to testnet-v1.54.0.

## Test plan

CI Pipeline.


mlegner (commented):
@wbbradley Thanks for the PR. I think the simtests should be fixed by also bumping the version of the `testnet-contracts`. The normal tests fail because some OpenAPI files would be changed; you can just run them locally and check in the modified files.

wbbradley:
Hmm.. simtests still failing locally for me.. let's see what happens in CI

mlegner (approved):
LGTM.

PR #2434, opened 2025-08-15
chore: bump all pre-commit versions and fix `move-tests-testnet` hook

mlegner:
## Description

* Bump all hooks to the latest version.
* Fix the argument of the `move-tests-testnet` hook.
* Make sure all Rust hooks are run on changes to the toolchain.

## Test plan

Ran pre-commit locally.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -69,7 +69,7 @@ repos:
           - "--config"
           - "group_imports=StdExternalCrate,imports_granularity=Crate,imports_layout=HorizontalVertical"
         language: rust
-        files: ^(crates/|Cargo\.(toml|lock)|rust-toolchain\.toml$)
+        files: &rust-files-pattern ^(crates/|Cargo\.(toml|lock)|rust-toolchain\.toml$)
         pass_filenames: false
       - id: cargo-test
         name: cargo-test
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -69,41 +69,41 @@ repos:
...
-        files: (^crates/|Cargo\.(toml|lock)$)
+        files: (^crates/|Cargo\.(toml|lock)|rust-toolchain\.toml$)
         pass_filenames: false
       - id: clippy-with-tests
         name: clippy-with-tests
         entry: cargo clippy
         args: ["--all-features", "--tests", "--", "-D", "warnings"]
         language: rust
-        files: ^(crates/|Cargo\.(toml|lock)$)
+        files: ^(crates/|Cargo\.(toml|lock)|rust-toolchain\.toml$)
```
karlwuest: Question/Nit: Is it possible to create a parameter for this pattern, s.t. it doesn't need to be repeated everywhere?

karlwuest: This should be possible [using yaml anchors and aliases](https://yaml.org/spec/1.2.2/#692-node-anchors) by naming the pattern in its first occurence (prepending with `&pattern_name`) and then it can be referenced later with `*pattern_name`.

mlegner: Good idea. Done in fdee98e4d25a2e39aad1d2b5be63b97110a12cee.

PR #2436, opened 2025-08-15
chore: bump Rust version in Dockerfiles to 1.89

mlegner:
Follow-up on #2435.


mlegner:
@pei-mysten Could you check that the relevant images exist in our internal registry?

PR #2440, opened 2025-08-18
fix(deps): bump all indirect dependencies

mlegner:
## Description

Bump all indirect dependencies by deleting and recreating the `Cargo.lock` file.

Also contains some minor code cleanup. 

## Test plan

CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,8 +207,8 @@ dependencies = [
  "rand 0.8.5",
  "rcgen",
  "ring",
- "rustls 0.23.31",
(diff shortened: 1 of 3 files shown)


mlegner:
@wbbradley I believe the previous errors in the tests were due to the fact that a dependency of the `walrus-orchestrator` crate enabled an additional `aws_lc_rs` feature for rustls, making it impossible for rustls to automatically determine the crypto provider. I reduced the set of enabled features for that dependency, which should fix the issue.

wbbradley (approved):

PR #2441, opened 2025-08-18
ci: add workflow to create sui testnet upgrade PR

ebmifa:
## Description
Add a workflow to create sui testnet upgrade PR

## Test plan
Will be testing and finalizing this PR once it lands

### DIFF
--- a/.github/workflows/gen-sui-upgrade-version-pr.yml
+++ b/.github/workflows/gen-sui-upgrade-version-pr.yml
@@ -65,9 +65,15 @@ jobs:
 
               # Use extended regex, in-place
               sed -E -i \
-                -e 's|^([[:space:]]*Sui[[:space:]]*=[[:space:]]*\{[^}]*git[[:space:]]*=[[:space:]]*"https://github.com/MystenLabs/sui\.git"[^}]*subdir[[:space:]]*=[[:space:]]*"crates/sui-framework/packages/sui-framework"[^}]*rev[[:space:]]*=[[:space:]]*")([^"]+)(")|\1'"$NEW_TAG"'\3|' \
-                -e 's|^([[:space:]]*[A-Za-z0-9_\-]+[[:space:]]*=[[:space:]]*\{[^}]*git[[:space:]]*=[[:space:]]*"https://github.com/MystenLabs/sui(\.git)?"[^}]*,?[[:space:]]*tag[[:space:]]*=[[:space:]]*")([^"]+)(")|\1'"$NEW_TAG"'\3|' \
-                -e 's|^([[:space:]]*[A-Za-z0-9_]+[[:space:]]*=[[:space:]]*")testnet-v[0-9]+\.[0-9]+\.[0-9]+(")|\1'"$NEW_TAG"'\2|' \
+                -e 's|^([[:space:]]*Sui[[:space:]]*=[[:space:]]*\{[^}]*\
+                  git[[:space:]]*=[[:space:]]*"https://github.com/MystenLabs/sui\.git"[^}]*\
+                  subdir[[:space:]]*=[[:space:]]*"crates/sui-framework/packages/sui-framework"\
+                  [^}]*rev[[:space:]]*=[[:space:]]*")([^"]+)(")|\1'"$NEW_TAG"'\3|' \
+                -e 's|^([[:space:]]*[A-Za-z0-9_\-]+[[:space:]]*=[[:space:]]*\
+                  \{[^}]*git[[:space:]]*=[[:space:]]*"https://github.com/MystenLabs/sui(\.git)?"\
+                  [^}]*,?[[:space:]]*tag[[:space:]]*=[[:space:]]*")([^"]+)(")|\1'"$NEW_TAG"'\3|' \
+                -e 's|^([[:space:]]*[A-Za-z0-9_]+[[:space:]]*=[[:space:]]*")\
+                  testnet-v[0-9]+\.[0-9]+\.[0-9]+(")|\1'"$NEW_TAG"'\2|' \
                 "$f"
 
               after="$(sha1sum "$f" | awk '{print $1}')"
@@ -99,4 +105,7 @@ jobs:
             --head "$BRANCH" \
             --title "chore: bump Sui to ${NEW_TAG}" \
             --reviewer "ebmifa,mlender" \
-            --body $'Automated Sui bump with single-pass updater:\n\n1) Move.toml Sui framework rev (repo+subdir strict)\n2) Cargo.toml Sui git deps tag (repo URL strict)\n3) docker/**/sui_version.toml\n\n`cargo build` run to refresh lockfiles.'
+            --body $'Automated Sui bump with single-pass updater:\n\n\
+                1) Move.toml Sui framework rev (repo+subdir strict)\n\
+                2) Cargo.toml Sui git deps tag (repo URL strict)\n\
+                3) docker/**/sui_version.toml\n\n`cargo build` run to refresh lockfiles.'
--- a/.github/workflows/gen-sui-upgrade-version-pr.yml
+++ b/.github/workflows/gen-sui-upgrade-version-pr.yml
@@ -1,0 +1,102 @@
+name: Generate Sui Version Upgrade PR
+run-name: Generate Sui Version Upgrade PR
+
+on:
+  workflow_dispatch:
+    inputs:
+      new_tag:
+        description: "New Sui tag (e.g., testnet-v1.55.0)"
+        required: true
+        type: string
+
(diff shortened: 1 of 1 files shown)


.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
...
+          # One list with all relevant TOML locations (globs ok)
+          FILES=(
+            "Move.toml"
+            "crates/**/Move.toml"
+            "contracts/**/Move.toml"
+            "apps/**/Move.toml"
+            "Cargo.toml"
+            "crates/**/Cargo.toml"
+            "apps/**/Cargo.toml"
+            "docker/**/sui_version.toml"
```
mlegner, line 47: The correct list is (currently) the following:
```suggestion
            "Cargo.toml"
            "contracts/*/Move.toml"
            "testnet-contracts/*/Move.toml"
            "docker/**/sui_version.toml"
```
In addition, we set a `SUI_TAG` in some workflows `.github/workflows/*.yml`.

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # pin@v4
+
+      - name: Update Sui refs (single loop), build, commit, push, PR
+        shell: bash
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
```
mlegner, line 32: Suggestion: Can we put this into a separate script `scripts/bump-sui-version.sh`? This makes it easier to run this manually if desired (and test it before merging this PR).

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
...
+          done
+
+          if [[ "$changed" -ne 1 ]]; then
+            echo "No matching lines changed; nothing to do."
+            exit 0
+          fi
+
+          # Build & lockfiles
+          cargo build --locked || cargo build
+          git add Cargo.lock **/Cargo.lock 2>/dev/null || true
```
mlegner, line 95: Nit: We only have the single `Cargo.lock` file in the root directory.

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
...
+              fi
+            done
+          done
+
+          if [[ "$changed" -ne 1 ]]; then
+            echo "No matching lines changed; nothing to do."
+            exit 0
+          fi
+
+          # Build & lockfiles
```
mlegner, line 93: Important: We also need to update the `Move.lock` files.

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
...
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git commit -m "chore: bump Sui to ${NEW_TAG}"
+          git push -u origin "$BRANCH"
+
+          gh pr create \
+            --base "$BASE" \
+            --head "$BRANCH" \
+            --title "chore: bump Sui to ${NEW_TAG}" \
+            --reviewer "ebmifa,mlender" \
```
mlegner, line 107: Suggestion: Add `@wbbradley` here. Also, my username is misspelled.
```suggestion
            --reviewer "ebmifa,mlegner,wbbradley" \
```

Also, remark: We use the https://github.com/peter-evans/create-pull-request action in some other workflows (e.g., [here](https://github.com/MystenLabs/walrus/blob/c932b0200137fb841932c380dcfbff5f2177a94d/.github/workflows/updates.yml#L49)), which is pretty convenient.

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
...
+            done
+          done
+
+          if [[ "$changed" -ne 1 ]]; then
+            echo "No matching lines changed; nothing to do."
+            exit 0
+          fi
+
+          # Build & lockfiles
+          cargo build --locked || cargo build
```
mlegner, line 94: Remark: My expectation is that in most cases (definitely more than half) this will fail, causing the workflow to fail as well. We should then either create an incomplete PR and have someone fix it manually or otherwise notify us of the failure.

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
+name: Generate Sui Version Upgrade PR
+run-name: Generate Sui Version Upgrade PR
+
+on:
+  workflow_dispatch:
+    inputs:
+      new_tag:
+        description: "New Sui tag (e.g., testnet-v1.55.0)"
+        required: true
+        type: string
```
mlegner, line 10: Question: Can we somehow trigger this on schedule and have it determine the latest Sui release automatically? Maybe on Wednesdays after the Sui release has been created?

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
...
+          # 2) Cargo.toml Sui git deps tag (strict: same inline table contains the Sui repo URL)
+          # 3) docker/*/sui_version.toml simple key="testnet-vX.Y.Z"
+          for pat in "${FILES[@]}"; do
+            for f in $pat; do
+              [[ -f "$f" ]] || continue
+              before="$(sha1sum "$f" | awk '{print $1}')"
+
+              # Use extended regex, in-place
+              sed -E -i \
+                -e 's|^([[:space:]]*Sui[[:space:]]*=[[:space:]]*\{[^}]*\
```
wbbradley, line 68: This regex seems like it may not be necessary from my experience doing these migrations. I've not experienced any problems to date with `s/testnet-v1.2.3/testnet-v1.3.0/`. @ebmifa did you see other patterns that indicate we need this multi-part sed statement?

.github/workflows/gen-sui-upgrade-version-pr.yml
wbbradley: My flow for this is to install the new required version of sui first, then run simtests - although I'd say I'm 60% sure about that flow.

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -0,0 +1,111 @@
...
+          # 1) Move.toml Sui framework rev (strict: repo + subdir must match)
+          # 2) Cargo.toml Sui git deps tag (strict: same inline table contains the Sui repo URL)
+          # 3) docker/*/sui_version.toml simple key="testnet-vX.Y.Z"
+          for pat in "${FILES[@]}"; do
+            for f in $pat; do
+              [[ -f "$f" ]] || continue
+              before="$(sha1sum "$f" | awk '{print $1}')"
+
+              # Use extended regex, in-place
+              sed -E -i \
```
wbbradley, line 67: @mlegner note that this script is not portable to BSD (macos). in-place edits in sed are incompatible between BSD and linux.

.github/workflows/gen-sui-upgrade-version-pr.yml
wbbradley: My intuition tells me (so take this with a grain of salt) it will be ultimately simpler to avoid the allowlist, but instead have a denylist consisting of the mainnet contracts toml files. My primary reasoning here is that it will clearly define the things that need to be locked, whereas an allowlist implicitly defines the set of things that need to be locked, but that includes a lot of things that are not related. ie: in this scheme the mainnet contracts would be excluded but so would any other toml files, as well as all the rust files, etc. In the denylist version, only the truly locked files would be excluded and everything else would be subject to moving forward (slightly simpler IMO). I know one is the dual of the other, but specifying a denylist also is a smaller specification, so it has that advantage in terms of reading comprehension.

.github/workflows/gen-sui-upgrade-version-pr.yml
wbbradley: Good idea! I would not want to block this PR for that feature add though.

wbbradley (commented):
Thanks for jumping on this @ebmifa!

.github/workflows/gen-sui-upgrade-version-pr.yml
wbbradley: Correction here: @mlegner informed me it's not the simtests we need to run, it's scripts/move_tests.sh (more specifically the move build part).

PR #2444, opened 2025-08-19
feat: store blobs as deletable by default unless specified as permanent

mlegner:
## Description

This implements the change to store new blobs as *deletable* by default instead of *permanent*.

**Important:** This should only be merged after the version bump to `v1.33.0` (#2488) but before the branch cut for `v1.33`.

## Test plan

Modified test suite + manual tests.

---

## Release notes

- [x] Publisher: **Warning:** Starting with this release (`v1.33`), blobs are stored as deletable by default; if you want to keep the previous behavior, you need to add the `permanent=true` query parameter.
- [x] CLI: **Warning:** Starting with this release (`v1.33`), blobs are stored as deletable by default; if you want to keep the previous behavior, you need to add the `--permanent` option when storing blobs.


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -709,13 +709,16 @@ pub struct PublisherQuery {
...
     pub epochs: EpochCount,
-    // TODO(WAL-911): Change the docstrings when the default behavior changes.
-    /// If true, the publisher creates a deletable blob instead of a permanent one. *This will
-    /// become the default behavior starting with v1.33.*
+    /// If true, the publisher creates a deletable blob.
+    ///
+    /// This flag is currently optional as new blobs are created as deletable by default since
+    /// v1.33; however, it is still recommended to explicitly use the flag when storing new
+    /// deletable blobs.
     #[serde(default)]
```
wbbradley, line 796: The `serde(default)` combined with the comment about the flag being optional and deletable by default has some cognitive dissonance. Should we mark the default_value as true now?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -709,13 +709,16 @@ pub struct PublisherQuery {
...
     pub deletable: bool,
-    /// If true, the publisher creates a permanent blob. This is currently the default behavior;
-    /// but *blobs will be deletable by default starting with v1.33*.
+    /// If true, the publisher creates a permanent blob instead of a deletable one.
+    ///
+    /// This was the default behavior before v1.33, but *blobs are now deletable by default*.
     #[serde(default)]
     pub permanent: bool,
```
wbbradley, line 802: We have two flags that mean the opposite of each other, both of which have default false values. Maybe we should reconsider this design?

crates/walrus-sui/src/client.rs
```diff
@@ -358,10 +358,11 @@ impl BlobPersistence {
...
 
-    /// Constructs [`Self`] based on the value of a `deletable` flag.
+    /// Constructs [`Self`] based on the value of a `deletable` and a `permanent` flag.
     ///
-    /// If `deletable` is true, returns [`Self::Deletable`], force otherwise returns
-    /// [`Self::Permanent`].
+    /// Returns the appropriate [`Self`] variant if exactly one of the flags is true. Returns an
+    /// error, if both flags are true. Returns [`Self::Deletable`] (new default behavior) if both
+    /// flags are false, but logs a warning in that case.
     pub fn from_deletable_and_permanent(
```
wbbradley, line 366: This function diverges by design. We should reconsider this API.

crates/walrus-service/src/client/daemon/routes.rs
mlegner: The reason to have two opposite flags is to facilitate a smooth transition of the default value: Users can already start using the `--permanent` flag before version 1.33 and stop using the `--deletable` flag later, they don't have to switch their invocations precisely with the new version.

crates/walrus-sui/src/client.rs
mlegner: Hmm... We could make this function private to not expose it?

Note that the possible error case here cannot happen when using the CLI, as the two flags are marked as conflicting with clap.

crates/walrus-sdk/src/client/resource.rs
```diff
@@ -197,9 +197,7 @@ impl StoreOp {
                     StoreOp::NoOp(BlobStoreResult::AlreadyCertified {
                         blob_id: blob.blob_id,
                         event_or_object: EventOrObjectId::Object(blob.id),
-                        end_epoch: blob
-                            .certified_epoch
-                            .expect("certified blob must have a certified epoch"),
+                        end_epoch: blob.storage.end_epoch,
```
wbbradley: Is the certified epoch no longer valid here?

mlegner: This was a pre-existing bug: the `certified_epoch` is the epoch in which the blob was certified and not that blob's end epoch. I created a separate PR to fix it (#2449) but also included the fix here as the tests fail otherwise.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: I see. Thanks for the explanation.

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1055,12 +1055,13 @@ pub struct CommonStoreOptions {
...
     ///
     /// Deletable blobs/quilts can be removed from Walrus before their expiration time.
     ///
-    /// *This will become the default behavior in v1.33.*
+    /// This flag is currently optional as new blobs are created as deletable by default since
+    /// v1.33; however, it is still recommended to explicitly use the flag when storing new
+    /// deletable blobs.
     #[arg(long, conflicts_with = "permanent")]
     #[serde(default)]
     pub deletable: bool,
```
wbbradley, line 1069: Maybe we should issue a warning to the console if `--deletable` is used informing the user that we are deprecating it.

crates/walrus-sui/src/client.rs
```diff
@@ -358,10 +358,11 @@ impl BlobPersistence {
...
         deletable: bool,
         permanent: bool,
```
wbbradley, line 368: The `(true, false)` case could have a deprecation notice.

crates/walrus-sui/src/client.rs
wbbradley: Now that I see that this is a temporary situation, I'm not so worried about it.

crates/walrus-service/src/client/cli/args.rs
mlegner: I'm honestly not sure if we should remove the option. It's always nice to be able to explicitly request the default value to be unaffected by changes to this default. 

We could maybe hide it in a future version though, when we also remove the warning about no specific persistence being set explicitly.

crates/walrus-sui/src/client.rs
mlegner: Hmm... We're currently printing a warning if neither flag is set to alert users of the default change. I think it would be a bit weird to print a warning in both cases.

I think we *can* deprecate the flag later, although I'm not convinced that we should.

PR #2445, opened 2025-08-19
chore: error relay early if rpc_url is not available

wbbradley:
## Description

Clarify how to ensure rpc_urls are available to walrus-upload-relay. Also, check that we have at least one rpc url prior to attempting to start the relay.

## Test plan

Manual testing, CI Pipeline.

### DIFF
--- a/crates/walrus-upload-relay/src/controller.rs
+++ b/crates/walrus-upload-relay/src/controller.rs
@@ -560,9 +560,11 @@ pub async fn get_client_with_config(
     tracing::debug!(?client_config, "loaded client config");
 
     if client_config.rpc_urls.is_empty() {
-        tracing::error!("No RPC URLs provided in the client configuration. Upload relay requires \
-            at least one RPC URL specified within your client config. (Run with RUST_LOG=debug for \
-            more information.)");
+        tracing::error!(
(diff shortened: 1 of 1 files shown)


crates/walrus-upload-relay/src/controller.rs
```diff
@@ -563,6 +559,15 @@ pub async fn get_client_with_config(
 ) -> Result<WalrusNodeClient<SuiReadClient>> {
     tracing::debug!(?client_config, "loaded client config");
 
+    if client_config.rpc_urls.is_empty() {
+        tracing::error!(
+            "No RPC URLs provided in the client configuration. Upload relay requires at least one \
+            RPC URL specified within your client config. (Run with RUST_LOG=debug for more \
+            information.)"
```
mlegner, line 566: Question: What would this additional information be?

mlegner (approved):
Thanks for the improvements here. 🙏

crates/walrus-upload-relay/src/controller.rs
wbbradley: > What would this additional information be?

This should print out the configuration that led to the failure.

PR #2447, opened 2025-08-20
chore: reduce log level for 404 when checkpoint is not found fetching from archival

halfprice:
## Description

Since this is expected as the checkpoint may not produced yet.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/checkpoint-downloader/src/downloader.rs
+++ b/crates/checkpoint-downloader/src/downloader.rs
@@ -15,7 +15,11 @@ use sui_types::messages_checkpoint::{CheckpointSequenceNumber, TrustedCheckpoint
 use tokio::{sync::mpsc, time::Instant};
 use tokio_util::sync::CancellationToken;
 use typed_store::{Map, rocks::DBMap};
-use walrus_sui::client::retry_client::{RetriableClientError, RetriableRpcClient};
+use walrus_sui::client::retry_client::{
+    RetriableClientError,
+    RetriableRpcClient,
+    retriable_rpc_client::FallbackError,
+};
 use walrus_utils::{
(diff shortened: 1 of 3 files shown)


crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -558,6 +562,21 @@ fn handle_checkpoint_error(err: &RetriableClientError, next_checkpoint: u64) {
             "failed to read next checkpoint, probably not produced yet",
         );
     }
+
+    if let RetriableClientError::FallbackError(fallback_error) = err
+        && let FallbackError::RequestFailed(error) = fallback_error
```
mlegner: Nit: These two could be merged:
```suggestion
    if let RetriableClientError::FallbackError(FallbackError::RequestFailed(error)) = err
```

crates/checkpoint-downloader/src/downloader.rs
```diff
@@ -558,6 +562,21 @@ fn handle_checkpoint_error(err: &RetriableClientError, next_checkpoint: u64) {
...
+        && error
+            .status()
+            .map(|status| status == axum::http::StatusCode::NOT_FOUND)
+            .unwrap_or(false)
```
mlegner: Nit: Maybe the following is more intuitive?
```suggestion
            .is_some_and(|status| status == axum::http::StatusCode::NOT_FOUND)
```

mlegner (approved):
Thanks for the verbosity fix, @halfprice!

halfprice:
THanks @mlegner ! Both are great suggestions!

Issue #1853, opened 2025-01-16
Add example to query system information


mlegner:
Our example `hello_walrus_sui_system.py` to query system information was outdated and temporarily removed in MystenLabs/walrus-docs#202. We should add some example code for this again.

PR #2448, opened 2025-08-20
Add example to query system information

lispking:
## Description

[Add example to query system information](https://github.com/MystenLabs/walrus/issues/1853)

## Test plan

1. Get the system object to obtain its version
2. Query the dynamic field with that version to get the actual system state

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:
- [x] Document:

### DIFF
--- a/docs/examples/python/README.md
+++ b/docs/examples/python/README.md
@@ -24,4 +24,5 @@
   client.
 - `hello_walrus_webapi.py` shows how to store and read blobs using the HTTP API of the Walrus
   client.
+- `hello_walrus_sui_system.py` shows how to query the Walrus system object on Sui.
 - `track_walrus_events.py` is a simple script to track all Walrus-related events on Sui.
--- a/docs/examples/python/hello_walrus_sui_system.py
+++ b/docs/examples/python/hello_walrus_sui_system.py
@@ -1,0 +1,193 @@
+# Author: Walrus Foundation
+# License: Apache-2.0
+
+# Example of querying the Walrus system object on Sui.
+
+# This script demonstrates how to retrieve and display information about the Walrus
+# decentralized storage system by querying the system object on the Sui blockchain.
+
+# The Walrus system object uses a new architecture where the actual system state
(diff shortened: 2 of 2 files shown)


mlegner:
@lispking Thanks for the contribution. I'll try to find time for this next week.

wbbradley:
This looks nice as a way of demonstrating basic Sui RPC access and navigating objects in general on Sui via Python with `requests`. However, I'm not sure that it's something that should live in the Walrus repo. @lispking was there a higher-level goal you are working towards that puts this in context?

lispking:
> This looks nice as a way of demonstrating basic Sui RPC access and navigating objects in general on Sui via Python with `requests`. However, I'm not sure that it's something that should live in the Walrus repo. @lispking was there a higher-level goal you are working towards that puts this in context?

Thanks for asking. this is because the previous issue stated that it needs to be placed here: https://github.com/MystenLabs/walrus/issues/1853

lispking:
> Thanks @lispking for looking into this. For now the linked issues has been closed due to lack of demand for that particular feature. The Python example code needs a higher-level treatment before we bring on more code. This is especially true about AI generated code without tests. Note to @mlegner lets chat about how to bring the most value to this area.

It's incorrect to say that this hasn't been tested. This Python script can run successfully. Moreover, other Python scripts follow a similar writing style. It's just that I've encapsulated it into a function here to make it easier for users to understand. One shouldn't assume that code is AI-generated just by looking at it.

```shell
$ python hello_walrus_sui_system.py
System object ID: 0x2134d52768ea07e8c43570ef975eb3e4c27a39fa6396bef985b5abc58d03ddd2
System object version: 2

=== Walrus System Information ===
Current epoch: 12
Committee members: 102
Number of shards: 1000
Storage price per unit: 11000 MIST
Write price per unit: 20000 MIST
Total capacity: 4,166,666,666,666,666 bytes
Used capacity: 573,115,192,953,200 bytes
Capacity usage: 13.75%
```

PR #2452, opened 2025-08-21
feat: Add transaction support in typed-store

sadhansood:
## Description

This PR adds the ability to do optimistic transactions with commit/rollback semantics and conflict handling with retry logic. Includes minor test helpers and fixes to existing tests.  The current functionality does not change.

## Test plan

Added some unit tests.

### DIFF
--- a/crates/typed-store/src/rocks.rs
+++ b/crates/typed-store/src/rocks.rs
@@ -1,10 +1,10 @@
 // Copyright (c) Walrus Foundation
 // SPDX-License-Identifier: Apache-2.0
 
-/// Error types and utilities for RocksDB operations
+/// Error types and utilities for RocksDB operations.
 pub mod errors;
 
-/// Safe iterator utilities for RocksDB
+/// Safe iterator utilities for RocksDB.
 pub(crate) mod safe_iter;
 
 use std::{
@@ -34,9 +34,14 @@ use rocksdb::{
     Error,
     LiveFile,
     MultiThreaded,
+    OptimisticTransactionDB,
+    OptimisticTransactionOptions,
     ReadOptions,
+    Transaction,
     WriteBatch,
+    WriteBatchWithTransaction,
     WriteOptions,
+    backup::BackupEngine,
     checkpoint::Checkpoint,
     properties::{self, num_files_at_level},
     statistics::Ticker,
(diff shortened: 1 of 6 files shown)


crates/typed-store/src/rocks.rs
```diff
@@ -78,8 +83,74 @@ const ROCKSDB_PROPERTY_TOTAL_BLOB_FILES_SIZE: &CStr =
 #[cfg(test)]
 mod tests;
 
+/// Repeatedly attempt an Optimistic Transaction until it succeeds.
+/// Since many callsites (e.g. the consensus handler) cannot proceed in the case of failed writes,
```
mlegner: Question: Is the remark about the consensus handler just due to copy/pasting this from Sui?

crates/typed-store/src/rocks.rs
```diff
@@ -120,13 +190,105 @@ impl RocksDB {
...
+pub struct OptimisticTransactionDBWrapper {
+    /// The underlying rocksdb database
+    pub underlying: rocksdb::OptimisticTransactionDB<MultiThreaded>,
+    /// The metric configuration
+    pub metric_conf: MetricConf,
+    /// The path of the database
+    pub db_path: PathBuf,
+    /// The database options
+    pub db_options: rocksdb::Options,
+}
```
mlegner: Question/suggestion: Any reason why we don't define a generic `DBWrapper<T>` struct?

crates/typed-store/src/rocks.rs
```diff
@@ -120,13 +190,105 @@ impl RocksDB {
...
+            (RocksDB::DB($d), RocksDBBatch::DB($b)) => $db_expr,
+            (RocksDB::OptimisticTransactionDB($td), RocksDBBatch::OptimisticTransactionDB($tb)) => {
+                $txn_expr
+            }
+            _ => Err(TypedStoreError::RocksDBError(
+                "using invalid batch type for the database".into(),
+            )),
+        }
+    };
+}
```
mlegner, line 301: Question: Do we actually need this macro? AFAICT this is only used a single time.

crates/typed-store/src/rocks.rs
```diff
@@ -120,13 +190,105 @@ impl RocksDB {
...
+
 impl RocksDB {
+    /// Returns the configured `rocksdb::Options` for this database instance.
+    pub fn db_options(&self) -> &rocksdb::Options {
+        delegate_call!(self.db_options)
+    }
+
     /// Get a value from the database
+    #[allow(dead_code)]
```
mlegner: Question: Why is this attribute needed, given that the function is public?

crates/typed-store/src/rocks.rs
```diff
@@ -249,27 +411,62 @@ impl RocksDB {
...
+            },
+            |d, b| {
+                d.underlying
+                    .write_opt(b, writeopts)
+                    .map_err(typed_store_err_from_rocks_err)
+            }
+        )
+    }
+
+    /// Create a new transaction without a snapshot
```
mlegner: Suggestion: Document that this can only be called on an `OptimisticTransactionDB`. Also, the `panic` statement below could contain a bit more information.

Same for the `transaction` method below.

crates/typed-store/src/rocks.rs
sadhansood: Yes, thanks for catching it! we ended up deleting the transaction support from `typed-store` when we ported it over to walrus, so there might be some out of date comments. I'll clean them up.

crates/typed-store/src/rocks.rs
sadhansood: No, not really. I can do this.

crates/typed-store/src/rocks.rs
sadhansood: I think this is mainly to keep the code clean in the DBMap. I (personally) don't like to see multiple `match` arms when browsing through the application logic, so abstracted it away.

crates/typed-store/src/rocks.rs
sadhansood: hmm, yeah good point. I can clean this sup.

crates/typed-store/src/rocks.rs
```diff
@@ -249,27 +391,75 @@ impl RocksDB {
...
+                    .write_opt(b, writeopts)
+                    .map_err(typed_store_err_from_rocks_err)
+            }
+        )
+    }
+
+    /// Create a new transaction without a snapshot.
+    /// This can only be called when the engine is an `OptimisticTransactionDB`.
+    /// Panics if called on a standard `RocksDB` engine.
+    pub fn transaction_without_snapshot(
```
liquid-helium: Can you add some comments about the how this could affect the consistency?

crates/typed-store/src/rocks.rs
```diff
@@ -301,63 +498,85 @@ impl RocksDB {
...
-            Checkpoint::new(&self.underlying).map_err(typed_store_err_from_rocks_err)?;
+        let checkpoint = self.new_checkpoint()?;
         checkpoint
             .create_checkpoint(path)
-            .map_err(|e| TypedStoreError::RocksDBError(e.to_string()))?;
-        Ok(())
+            .map_err(|e| TypedStoreError::RocksDBError(e.to_string()))
+    }
+
+    /// Create a new backup of the database
```
liquid-helium: nit: period at the end of a comment, and other places.

crates/typed-store/src/rocks.rs
```diff
@@ -78,10 +83,96 @@ const ROCKSDB_PROPERTY_TOTAL_BLOB_FILES_SIZE: &CStr =
 #[cfg(test)]
 mod tests;
 
-/// The rocksdb database
-pub struct RocksDB {
+/// Repeatedly attempt an Optimistic Transaction until it succeeds.
+/// This will loop forever until the transaction succeeds.
+#[macro_export]
+macro_rules! retry_transaction {
```
liquid-helium, line 89: Question: what is the advantage of a macro_rules! compared to a generic TransactionRunner(db)?

crates/typed-store/src/rocks.rs
```diff
@@ -78,10 +83,96 @@ const ROCKSDB_PROPERTY_TOTAL_BLOB_FILES_SIZE: &CStr =
...
+            rngs::ThreadRng,
+        };
+        use tokio::time::{Duration, sleep};
+
+        let mut retries = 0;
+        let max_retries = $max_retries;
+        loop {
+            let status = $transaction;
+            match status {
+                Err(TypedStoreError::RetryableTransactionError) => {
```
liquid-helium, line 111: Is it possible to filter RocksDB::Error? otherwise, the caller has to manually map the erorrs?

crates/typed-store/src/rocks.rs
mlegner: Given the relatively high complexity of the macro and the fact it's just used once, I'd personally prefer the non-macro version. But not blocking from my side.

crates/typed-store/src/rocks.rs
sadhansood: We could model it as a function, but this macro seems better because callers can just write the transactional expression inline and leverage the centralized error mapping to RetryableTransactionError, with no closure/lifetime worry and a nice defaulted form.

crates/typed-store/src/rocks.rs
sadhansood: done

crates/typed-store/src/rocks.rs
sadhansood: done

crates/typed-store/src/rocks.rs
sadhansood: sure, done

crates/typed-store/src/rocks.rs
```diff
@@ -78,34 +83,120 @@ const ROCKSDB_PROPERTY_TOTAL_BLOB_FILES_SIZE: &CStr =
...
 
-/// The rocksdb database
-pub struct RocksDB {
-    /// The underlying rocksdb database
-    pub underlying: rocksdb::DBWithThreadMode<MultiThreaded>,
-    /// The metric configuration
+/// Repeatedly attempt an Optimistic Transaction until it succeeds.
+/// This will loop forever until the transaction succeeds.
+#[macro_export]
+macro_rules! retry_transaction {
```
liquid-helium, line 89: Can we add metrics, perhaps in another PR to collect the retry rate?
In a mostly unlikely case, if the retry is frequent, we might want to fall back to the regular transaction mode.

crates/typed-store/src/rocks/tests.rs
```diff
@@ -981,3 +986,100 @@ async fn test_dbmap_ticker_statistics() {
         "Read histogram max should be a normal number"
     );
 }
+
+fn open_optimistic_rocksdb<P: AsRef<Path>>(path: P, opt_cfs: &[&str]) -> Arc<RocksDB> {
```
liquid-helium, line 994: Is it possible to enable it for our e2e tests, and other tests?

crates/typed-store/src/rocks.rs
sadhansood: yes sure, there are some more things that need to be added to the transactions but i am keeping it for a future PR to keep the size small. I can definitely add metrics to retries there, good point.

crates/typed-store/src/rocks/tests.rs
sadhansood: I intend to do it in a future PR (to keep the size  and scope of this PR small) as well where we switch our databases to optimistic transaction db's.

crates/typed-store/src/rocks.rs
```diff
@@ -374,28 +606,123 @@ impl RocksDB {
...
+        to: K,
+    ) -> Result<(), TypedStoreError> {
+        match self {
+            Self::DB(batch) => {
+                batch.delete_range_cf(cf, from, to);
+                Ok(())
+            }
+            Self::OptimisticTransactionDB(_) => {
+                panic!("delete_range_cf is not supported for OptimisticTransactionDB")
+            }
```
mlegner: Important: After looking at #2475 and reviewing this again, I'm a bit worried about the possible runtime panics of `delete_range_cf`, `transaction`, and `transaction_without_snapshot`.

I think we should remove these functions from the enums. Alternatively, we can modify this function such that it creates a transaction, and then iterates over the range, deleting all items.

sadhansood: Thanks @mlegner , i've removed panics and made `transaction` more explicit by adding `as_optimistic` and requiring users pass in a range delete capacity object in order to invoke it.

PR #2453, opened 2025-08-21
chore: add a script to poll for new sui dep version

wbbradley:
## Description

Created a script for use in monitoring for Sui dependency upgrades which should tell us to bump Walrus's Sui dep version.

## Test plan

Manually run.


halfprice:
Hi @wbbradley , echo what Markus pointed out, do we still need this PR?

PR #2454, opened 2025-08-21
ci: Creating sui testnet version update PR workflow

ebmifa:
## Description
Creating sui testnet version update PR

## Test plan
Workflow run: https://github.com/MystenLabs/walrus/actions/runs/17162351338

PR generated: https://github.com/MystenLabs/walrus/pull/2460


scripts/bump_sui_testnet_version.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+# Copyright (c) Walrus Foundation
+# SPDX-License-Identifier: Apache-2.0
+#
+# This script creates a PR branch and updates Sui testnet versions in
+# selected files.
+
+set -Eeuo pipefail
```
mlegner, line 8: Question: Can we make the script a bit more robust to failures? For example, I expect that in many cases the `cargo build` will fail because of some breaking changes.

scripts/bump_sui_testnet_version.sh
```diff
@@ -0,0 +1,106 @@
...
+
```
mlegner, line 9: Suggestion: Can we check at the beginning that relevant binaries are installed (mainly `gh` and `sui`)?

scripts/bump_sui_testnet_version.sh
```diff
@@ -0,0 +1,106 @@
...
+# Set up branch for changes.
+STAMP="$(date +%Y%m%d%H%M%S)"
+BRANCH="${GITHUB_ACTOR}/bump-sui-${NEW_TAG}-${STAMP}"
+git checkout -b "$BRANCH"
+
+# Allow recursive globs.
+shopt -s globstar nullglob
+
+# List of relevant TOML locations (globs allowed).
+FILES=(
```
mlegner, line 45: We also need to update the `SUI_TAG` in some of our GH workflows.

.github/workflows/gen-sui-upgrade-version-pr.yml
mlegner, line 1: Suggestion (potentially for a follow-up): Can we somehow trigger this whenever a new Sui release with a major or minor version bump is created?

scripts/bump_sui_testnet_version.sh
```diff
@@ -0,0 +1,106 @@
...
+
+  for f in "${TARGETS[@]}"; do
+    sed -i -E \
+      "s/( = \")testnet-v[0-9]+\.[0-9]+\.[0-9]+/\1${NEW_TAG}/g" "$f"
+  done
+fi
+
+# Update Cargo.lock
+echo "Running cargo build --release ..."
+cargo build --release
```
mlegner: Question: Any reason to do a release build here? Probably a simple `cargo check` would be sufficient here.

scripts/bump_sui_testnet_version.sh
wbbradley: I'd still prefer this to be a denylist, not an allowlist. My thoughts were in the prior review, happy to recap them as needed.

We could do something like 

```
git ls-files | while read -r file; do
    if [[ ! "$file" =~ ^mainnet-contracts/ ]] && [[ ! "$file" =~ \.lock$ ]]; then
        sed -i ... "$file"
    fi
done
```

scripts/bump_sui_testnet_version.sh
wbbradley: Avoid failing the script if this fails.

```suggestion
cargo build --release ||:
```

scripts/bump_sui_testnet_version.sh
ebmifa: If there are any build issues, it will exit `1` and the workflow will fail as well

scripts/bump_sui_testnet_version.sh
ebmifa: It's actually better to be explicit about what files/paths we are updating. It reads better and makes it easier to add more paths/files if needed. The tl;dr is that being explicit scales better.

ebmifa: > We also need to update the `SUI_TAG` in some of our GH workflows.

@mlegner, which workflows are you referring to?

.github/workflows/gen-sui-upgrade-version-pr.yml
ebmifa: @mlegner do you mean when we cut the branch? or everytime we bump a version in the current release branch?

scripts/bump_sui_testnet_version.sh
ebmifa: Does `cargo check` generate the duplicate lock files correctly?

ebmifa: cargo failures exit with `1` so this will always get caught

scripts/bump_sui_testnet_version.sh
ebmifa: @mlegner this is what `set -Eeuo pipefail` is for actually

scripts/bump_sui_testnet_version.sh
mlegner: * https://github.com/MystenLabs/walrus/blob/main/.github/workflows/code.yml
* https://github.com/MystenLabs/walrus/blob/main/.github/workflows/updates.yml
* https://github.com/MystenLabs/walrus/blob/main/.github/workflows/scheduled_reports.yml

.github/workflows/gen-sui-upgrade-version-pr.yml
mlegner: I'd suggest whenever a new Testnet release is created in the Sui repo (with a minor version bump).

scripts/bump_sui_testnet_version.sh
mlegner: I *think* so... 😬

scripts/bump_sui_testnet_version.sh
mlegner: I think we do not want the workflow to fail in that case but rather create a partial PR, which we can then manually fix.

mlegner: > @mlegner this is what `set -Eeuo pipefail` is for actually

Yes, I understand. My point is that we shouldn't completely fail if there are build issues.

ebmifa: >  but rather create a partial PR, which we can then manually fix.

So if the `cargo check` fails, it's ok for us to proceed with creating a PR?

.github/workflows/gen-sui-upgrade-version-pr.yml
ebmifa: @mlegner what about when a new sui release gets deployed to `testnet` for the 1st time?

scripts/bump_sui_testnet_version.sh
ebmifa: @mlegner Why are we hardcoding tags in those workflows? This sounds like a bad design.
Can't we get the tag programmatically? Or at least have it accept an input via automation or even manually.

scripts/bump_sui_testnet_version.sh
mlegner: Yes, although ideally with some warning that the lockfile generation failed. The CI workflows will anyway block merging the PR.

scripts/bump_sui_testnet_version.sh
mlegner: The main reason is that we don't want PRs to be blocked because there was a new Sui release. Of course we want *some* workflow to fail and alert us of this fact, but that's why we have the main branch CI.

.github/workflows/gen-sui-upgrade-version-pr.yml
mlegner: That works, assuming the corresponding tag already exists then. 👍

scripts/bump_sui_testnet_version.sh
ebmifa: @mlegner I am actually going to handle these workflows in a separate PR to make them read the testnet-vX.Y.Z tag programmatically from Cargo.toml files, since these would be identical, right?

mlegner: Yes, that makes sense.

.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -5,8 +5,8 @@ on:
   workflow_dispatch:
```
mlegner, line 5: Question: Can we already add a "repository dispatch" or a schedule as well?

mlegner (approved):
Thanks for all the work, @ebmifa, LGTM.

.github/workflows/gen-sui-upgrade-version-pr.yml
ebmifa: Yes, I was thinking of doing it on the schedule. Will be doing it in the next PR

PR #2458, opened 2025-08-22
feat: Allow event blob catchup at runtime

sadhansood:
## Description

This PR adds runtime catchup support to the walrus-service event processor so a node that falls behind at runtime can catch up using event blobs, while the checkpoint downloader continues to run safely.

1. Runtime lag monitoring and catchup trigger periodically computes lag between local and network checkpoints which triggers EventBlobCatchupManager when lag exceeds threshold at runtime (not only at startup).
2. New CatchupCoordinationState handles Stop/Restart/CatchupFailed with an unbounded channel.
3. Tailing uses start_tailing_checkpoints_with_restart_support to listen for coordination messages.
4. Processing phase sends a stop message, waits for acks, then restarts tailing afterwards.
5. Added an ack using Notify to signal when checkpoint tailing has fully stopped.
6. Tailing task uses a child CancellationToken

Process:
Monitor task checks lag at runtime_lag_check_interval. If lag > runtime_catchup_lag_threshold, it triggers catchup:
Phase 1: Download blobs (parallel to normal tailing).
Phase 2: Send Stop to tailing; wait for ack with timeout.
Phase 3: Process blobs; update stores and metrics.
Phase 4: Send Restart to tailing.

## Test plan

Added simtest

### DIFF
--- a/crates/walrus-service/src/event/event_processor/catchup.rs
+++ b/crates/walrus-service/src/event/event_processor/catchup.rs
@@ -3,7 +3,7 @@
 
 //! Catchup module for catching up the event processor with the network.
 
-use std::{fs, path::PathBuf};
+use std::{fs, path::PathBuf, sync::Arc, time::Duration};
 
 use sui_types::{
     committee::Committee,
@@ -22,6 +22,7 @@ use crate::event::{
     event_blob::EventBlob,
     event_processor::{
         config::{SuiClientSet, SystemConfig},
+        coordination::CatchupCoordinationState,
         db::EventProcessorStores,
     },
     events::{IndexedStreamEvent, InitState},
@@ -57,6 +58,28 @@ impl std::fmt::Debug for DownloadedBlob {
     }
 }
 
+/// Configuration for runtime catchup coordination.
+#[derive(Clone, Debug)]
+pub struct CatchupRuntimeConfig {
+    /// The coordination state for the catchup process.
+    pub coordination_state: Arc<CatchupCoordinationState>,
+    /// The timeout for the coordination process.
+    pub coordination_timeout: Duration,
+    /// The timeout for the processing process.
+    pub processing_timeout: Duration,
+}
+
+/// Structured error type for catchup.
+#[derive(thiserror::Error, Debug)]
+pub enum CatchupError {
+    /// A recoverable error that can be retried.
+    #[error("recoverable catchup error: {0}")]
+    Recoverable(#[from] anyhow::Error),
+    /// A non-recoverable error that cannot be retried.
+    #[error("non-recoverable catchup error: {0}")]
+    NonRecoverable(anyhow::Error),
+}
+
 /// Manages the catchup process for events in the event processor using event blobs.
 ///
 /// This manager handles the process of catching up the local event store with the network state.
@@ -71,6 +94,7 @@ pub struct EventBlobCatchupManager {
     system_config: SystemConfig,
     recovery_path: PathBuf,
     metrics: EventCatchupManagerMetrics,
+    catchup_runtime_config: CatchupRuntimeConfig,
 }
 
 impl std::fmt::Debug for EventBlobCatchupManager {
@@ -91,6 +115,7 @@ impl EventBlobCatchupManager {
         system_config: SystemConfig,
         recovery_path: PathBuf,
         registry: &Registry,
+        catchup_runtime_config: CatchupRuntimeConfig,
     ) -> Self {
         let metrics = EventCatchupManagerMetrics::new(registry);
         Self {
@@ -99,31 +124,27 @@ impl EventBlobCatchupManager {
             system_config,
             recovery_path,
             metrics,
+            catchup_runtime_config,
         }
     }
 
     /// Checks if the event processor is lagging behind the network and performs catchup if needed.
     pub async fn catchup(&self, lag_threshold: u64) -> anyhow::Result<()> {
-        let current_checkpoint = self.get_current_checkpoint()?;
-        let latest_checkpoint = self.get_latest_network_checkpoint().await;
-
-        let current_lag = self.calculate_lag(current_checkpoint, latest_checkpoint)?;
+        let current_lag = self.get_current_lag().await?;
 
         if current_lag > lag_threshold {
-            tracing::info!(
-                current_checkpoint,
-                latest_checkpoint,
-                lag_threshold,
-                "performing catchup - lag is above threshold"
-            );
-            self.perform_catchup().await?;
+            tracing::info!(lag_threshold, "performing catchup - lag is above threshold");
+            match self.perform_catchup().await {
+                Ok(()) => {}
+                Err(CatchupError::Recoverable(error)) => {
+                    tracing::warn!(?error, "recoverable error during catchup");
+                }
+                Err(CatchupError::NonRecoverable(error)) => {
+                    return Err(error);
+                }
+            }
         } else {
-            tracing::info!(
-                current_checkpoint,
-                latest_checkpoint,
-                lag_threshold,
-                "skipping catchup - lag is below threshold"
-            );
+            tracing::info!(lag_threshold, "skipping catchup - lag is below threshold");
         }
 
         Ok(())
@@ -184,16 +205,39 @@ impl EventBlobCatchupManager {
         Ok(lag)
     }
 
+    /// Gets the current lag between the local store and the network
+    pub async fn get_current_lag(&self) -> anyhow::Result<u64> {
+        let current_checkpoint = self.get_current_checkpoint()?;
+        let latest_checkpoint = self.get_latest_network_checkpoint().await;
+        self.calculate_lag(current_checkpoint, latest_checkpoint)
+    }
+
     /// Performs the catchup operation using event blobs
-    pub async fn perform_catchup(&self) -> anyhow::Result<()> {
-        match self.catchup_using_event_blobs().await {
+    pub async fn perform_catchup(&self) -> Result<(), CatchupError> {
+        let coordination_state = self.catchup_runtime_config.coordination_state.clone();
+        if !coordination_state.try_start_catchup().await {
+            tracing::info!("runtime catchup already active, skipping new run");
+            return Ok(());
+        }
+
+        let result = self.catchup_using_event_blobs().await;
+        coordination_state.mark_catchup_inactive();
+
+        match result {
             Ok(()) => {
                 tracing::info!("successfully caught up using event blobs");
                 Ok(())
             }
-            Err(error) => {
-                tracing::error!(?error, "failed to catch up using event blobs");
-                Err(error)
+            Err(CatchupError::Recoverable(error)) => {
+                tracing::warn!(?error, "recoverable error during catchup using event blobs");
+                Err(CatchupError::Recoverable(error))
+            }
+            Err(CatchupError::NonRecoverable(error)) => {
+                tracing::error!(
+                    ?error,
+                    "non-recoverable error during catchup using event blobs"
+                );
+                Err(CatchupError::NonRecoverable(error))
             }
         }
     }
@@ -218,27 +262,99 @@ impl EventBlobCatchupManager {
     /// starting from `N+1`). If however, the local store is empty, the catch-up will store all
     /// events from the earliest available event blob (in which case the first stored event index
     /// could be greater than `0`).
-    async fn catchup_using_event_blobs(&self) -> anyhow::Result<()> {
+    async fn catchup_using_event_blobs(&self) -> Result<(), CatchupError> {
         #[cfg(msim)]
         sui_macros::fail_point!("fail_point_catchup_using_event_blobs_start");
+        let coordination_state = self.catchup_runtime_config.coordination_state.clone();
 
-        let next_checkpoint = self.get_next_checkpoint()?;
+        let next_checkpoint = self
+            .get_next_checkpoint()
+            .map_err(|error| CatchupError::Recoverable(anyhow::Error::from(error)))?;
         tracing::info!(next_checkpoint, "starting event catchup using event blobs");
-        self.ensure_recovery_directory()?;
+        self.ensure_recovery_directory()
+            .map_err(CatchupError::Recoverable)?;
 
-        let blobs = self
-            .collect_event_blobs_for_catchup(next_checkpoint)
-            .await?;
-        let next_event_index = self.get_next_event_index()?;
+        let blobs = match self.collect_event_blobs_for_catchup(next_checkpoint).await {
+            Ok(blobs) => blobs,
+            Err(error) => {
+                tracing::error!(error = ?error, "failed to collect event blobs for catchup");
+                return Err(CatchupError::Recoverable(anyhow::anyhow!(
+                    "failed to collect event blobs for catchup"
+                )));
+            }
+        };
 
-        self.process_event_blobs(blobs, next_event_index).await?;
+        if blobs.is_empty() {
+            tracing::info!(
+                "no event blobs collected for catchup; skipping stop of checkpoint tailing"
+            );
+            return Ok(());
+        }
 
-        Ok(())
+        if let Err(e) = coordination_state.start_catchup_processing_phase().await {
+            tracing::error!(
+                error = ?e,
+                "failed to send stop message to checkpoint tailing"
+            );
+            return Err(CatchupError::NonRecoverable(anyhow::anyhow!(
+                "failed to send stop message to checkpoint tailing"
+            )));
+        }
+
+        tracing::info!("waiting for tailing to stop before processing blobs");
+        if !coordination_state
+            .wait_for_tailing_stopped(self.catchup_runtime_config.coordination_timeout)
+            .await
+        {
+            tracing::error!(
+                timeout_secs = self.catchup_runtime_config.coordination_timeout.as_secs(),
+                "timed out waiting for checkpoint tailing to stop"
+            );
+            return Err(CatchupError::NonRecoverable(anyhow::anyhow!(
+                "timed out waiting for checkpoint tailing to stop"
+            )));
+        }
+
+        let next_event_index = self
+            .get_next_event_index()
+            .map_err(|error| CatchupError::NonRecoverable(anyhow::Error::from(error)))?;
+        let processing_result = tokio::time::timeout(
+            self.catchup_runtime_config.processing_timeout,
+            self.process_event_blobs(blobs, next_event_index),
+        )
+        .await;
+
+        let mut recoverable_error: Option<anyhow::Error> = None;
+        match processing_result {
+            Ok(Ok(num_events_recovered)) => {
+                tracing::info!("successfully processed {} events", num_events_recovered);
+            }
+            Ok(Err(error)) => {
+                tracing::error!(error = ?error, "failed to process event blobs");
+                recoverable_error = Some(error);
+            }
+            Err(_) => {
+                let timeout_error = anyhow::anyhow!("processing event blobs timed out");
+                tracing::error!(error = ?timeout_error, "processing event blobs timed out");
+                recoverable_error = Some(timeout_error);
+            }
+        }
+
+        if let Err(e) = coordination_state.complete_catchup().await {
+            tracing::error!(error = ?e, "failed to send restart message to checkpoint tailing");
+            Err(CatchupError::NonRecoverable(anyhow::anyhow!(
+                "failed to send restart message to checkpoint tailing"
+            )))
+        } else if let Some(error) = recoverable_error {
+            Err(CatchupError::Recoverable(error))
+        } else {
+            Ok(())
+        }
     }
 
     /// Gets the next checkpoint sequence number that is after the latest checkpoint in the
     /// checkpoint store.
-    fn get_next_checkpoint(&self) -> Result<Option<u64>, TypedStoreError> {
+    pub fn get_next_checkpoint(&self) -> Result<Option<u64>, TypedStoreError> {
         Ok(self
             .stores
             .checkpoint_store
@@ -249,7 +365,7 @@ impl EventBlobCatchupManager {
     }
 
     /// Gets the next event index that is after the latest event index in the event store.
-    fn get_next_event_index(&self) -> Result<Option<u64>, TypedStoreError> {
+    pub fn get_next_event_index(&self) -> Result<Option<u64>, TypedStoreError> {
         Ok(self
             .stores
             .event_store
@@ -260,7 +376,7 @@ impl EventBlobCatchupManager {
     }
 
     /// Ensures the recovery directory exists
-    fn ensure_recovery_directory(&self) -> anyhow::Result<()> {
+    pub fn ensure_recovery_directory(&self) -> anyhow::Result<()> {
         if !self.recovery_path.exists() {
             fs::create_dir_all(&self.recovery_path)?;
         }
@@ -281,7 +397,7 @@ impl EventBlobCatchupManager {
     /// This function creates a client to download event blobs up to a specified
     /// checkpoint. The blobs are stored in the provided recovery path.
     #[cfg(feature = "client")]
-    async fn collect_event_blobs_for_catchup(
+    pub async fn collect_event_blobs_for_catchup(
         &self,
         starting_checkpoint_to_process: Option<u64>,
     ) -> anyhow::Result<Vec<BlobId>> {
@@ -313,11 +429,18 @@ impl EventBlobCatchupManager {
         Ok(blob_ids)
     }
 
-    async fn process_event_blobs(
+    /// Processes event blobs and stores them in the event store.
+    ///
+    /// This function performs the following steps:
+    /// 1. Iterates through the event blobs in reverse order (oldest to newest).
+    /// 2. Processes each blob:
+    ///    - Collects relevant events that maintain a continuous sequence with the local store.
+    ///    - Stores the events in the event store.
+    pub async fn process_event_blobs(
         &self,
         blobs: Vec<BlobId>,
         next_event_index: Option<u64>,
-    ) -> anyhow::Result<()> {
+    ) -> anyhow::Result<usize> {
         tracing::info!("starting to process event blobs");
 
         let mut num_events_recovered = 0;
@@ -372,7 +495,7 @@ impl EventBlobCatchupManager {
         }
 
         tracing::info!("recovered {} events from event blobs", num_events_recovered);
-        Ok(())
+        Ok(num_events_recovered)
     }
 
     async fn process_single_blob(
--- a/crates/walrus-service/src/event/event_processor/config.rs
+++ b/crates/walrus-service/src/event/event_processor/config.rs
@@ -106,6 +106,25 @@ pub struct EventProcessorConfig {
     #[serde_as(as = "DurationSeconds")]
     #[serde(rename = "sampled_tracing_interval_secs")]
     pub sampled_tracing_interval: Duration,
+    /// Enable runtime catchup functionality.
+    pub enable_runtime_catchup: bool,
+    /// Runtime catchup lag threshold for triggering catchup at runtime.
+    /// When the lag exceeds this threshold during runtime, the catchup manager
+    /// will be invoked in parallel with the checkpoint downloader to accelerate
+    /// recovery. Should typically be the same as event_stream_catchup_min_checkpoint_lag.
+    pub runtime_catchup_lag_threshold: u64,
+    /// Interval for checking lag during runtime to trigger catchup if needed.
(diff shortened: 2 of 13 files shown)


crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -12,27 +12,35 @@ use sui_types::{
...
+        coordination::CoordinationMessage,
         db::EventProcessorStores,
     },
     events::{IndexedStreamEvent, InitState, StreamEventWithInitState},
 };
 
 /// The maximum number of events to poll per poll.
 const MAX_EVENTS_PER_POLL: usize = 1000;
+/// Startup catchup timeout (seconds) used during initial catchup in EventProcessor::new
+const STARTUP_CATCHUP_TIMEOUT_SECS: u64 = 300 * 5;
```
mlegner: Question: Is there anything special about the values 300 and 5? If not, I'd just put 1500 there or `25 * 60`.

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -244,17 +277,52 @@ impl EventProcessor {
 
     /// Starts the event processor. This method will run until the cancellation token is cancelled.
     pub async fn start(&self, cancellation_token: CancellationToken) -> Result<(), anyhow::Error> {
-        tracing::info!("starting event processor");
+        tracing::info!(
+            runtime_catchup_enabled = self.config.enable_runtime_catchup,
+            "starting event processor with runtime catchup support"
```
mlegner: Remark: The message here is a bit misleading if one doesn't look at the additional `runtime_catchup_enabled` field.

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -244,17 +277,52 @@ impl EventProcessor {
...
+            coordination_state.clone(),
+        );
+        let catchup_monitoring_task = if self.config.enable_runtime_catchup {
+            Some(self.start_runtime_catchup_monitoring(
+                cancellation_token.clone(),
+                coordination_state.clone(),
+            ))
+        } else {
+            None
+        };
```
mlegner: Nit: I think this could be moved into the then clause of the following if statement.

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -244,17 +277,52 @@ impl EventProcessor {
...
+        } else {
+            select! {
+                pruning_result = pruning_task => {
+                    cancellation_token.cancel();
+                    pruning_result
+                }
+                tailing_result = tailing_task => {
+                    cancellation_token.cancel();
+                    tailing_result
+                }
```
mlegner, line 328: Question: Couldn't we just run the `self.start_tailing_task` here?

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -327,6 +398,64 @@ impl EventProcessor {
...
+                    }
+
+                }
+                join_res = async {
+                    tailing_task
+                    .take()
+                    .expect("tailing_task present").await }, if tailing_task.is_some() => {
+                    match join_res {
+                        Ok(Ok(_)) => tracing::info!("tailing task exited"),
+                        Ok(Err(e)) => tracing::error!(?e, "tailing task error"),
```
mlegner: Nit: Can we either rename `e` to `error` or use the field `error = ?e`? (Also below.)

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -327,6 +398,64 @@ impl EventProcessor {
...
+    async fn start_tailing_checkpoints_with_restart_support(
+        &self,
+        cancel_token: CancellationToken,
+        coordination_state: Arc<CatchupCoordinationState>,
+    ) -> Result<()> {
+        let mut coordination_rx = coordination_state.get_coordination_receiver().await;
+        let mut child_cancel_token = cancel_token.child_token();
+
+        let mut tailing_task: Option<JoinHandle<Result<()>>> =
+            Some(self.start_tailing_task(child_cancel_token.clone(), coordination_state.clone()));
```
mlegner, line 431: Question: Why does this have to be an `Option`?

crates/walrus-service/src/event/event_processor/config.rs
```diff
@@ -116,6 +136,11 @@ impl Default for EventProcessorConfig {
             adaptive_downloader_config: Default::default(),
             event_stream_catchup_min_checkpoint_lag: 20_000,
             sampled_tracing_interval: Duration::from_secs(3600),
+            runtime_catchup_lag_threshold: 20_000,
+            runtime_lag_check_interval: Duration::from_secs(30),
```
mlegner: Remark: Given the lag threshold of 20k corresponds to more than 1h of checkpoints, we could increase the check interval a bit.

crates/walrus-service/src/event/event_processor/catchup.rs
```diff
@@ -221,24 +243,84 @@ impl EventBlobCatchupManager {
...
 
-        let blobs = self
-            .collect_event_blobs_for_catchup(next_checkpoint)
-            .await?;
+        let blobs = match self.collect_event_blobs_for_catchup(next_checkpoint).await {
+            Ok(blobs) => blobs,
+            Err(e) => {
+                tracing::error!(error = ?e, "failed to collect event blobs for catchup");
+                // Since we are not able to collect event blobs, we can't proceed with catchup.
+                // We return Ok(()) because it is still safe to continue with the event processor.
```
mlegner: Question: How is this different from other errors above? I think it would be better overall to return a more structured error (e.g., splitting into recoverable and non-recoverable errors) and let the caller handle it.

crates/walrus-service/src/event/event_processor/catchup.rs
```diff
@@ -221,24 +243,84 @@ impl EventBlobCatchupManager {
     async fn catchup_using_event_blobs(&self) -> anyhow::Result<()> {
         #[cfg(msim)]
         sui_macros::fail_point!("fail_point_catchup_using_event_blobs_start");
+        let coordination_state_clone = self.catchup_runtime_config.coordination_state.clone();
```
mlegner: Nit: Can we just call this `coordination_state`?

crates/walrus-service/src/event/event_processor/coordination.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Get a receiver for coordination messages (for checkpoint tailing task)
+    pub async fn get_coordination_receiver(&self) -> mpsc::UnboundedReceiver<CoordinationMessage> {
+        // Take the receiver out of the Arc<Mutex<>> - this should only be called once
+        let mut rx_guard = self.coordination_rx.lock().await;
+        let (_new_tx, new_rx) = mpsc::unbounded_channel();
+
+        // Replace the old receiver with a new one (though it won't be used)
+        std::mem::replace(&mut *rx_guard, new_rx)
+    }
```
mlegner: Question: So we take out the existing receiver and just replace it with a dummy one? I find this very unintuitive and also a bit dangerous if this is called again later. Instead of storing this in the `CatchupCoordinationState`, couldn't we simply provide the receiver to the `start_tailing_checkpoints_with_restart_support` directly?

crates/walrus-service/src/event/event_processor/coordination.rs
```diff
@@ -0,0 +1,254 @@
...
+
+    /// Mark that checkpoint tailing has started.
+    pub fn mark_tailing_started(&self) {
+        self.is_tailing_stopped
+            .store(false, std::sync::atomic::Ordering::Release);
```
mlegner, line 86: Question: Any particular reason not to use the strongest consistency guarantee with `SeqCst`? (Also applies to other atomic operations in this file.)

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: done

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: fixed

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: done

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: i don't see how (and why), we call `start_tailing_task` inside `start_tailing_checkpoints_with_restart_support` because we need to keep track of the task handle.

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: done

crates/walrus-service/src/event/event_processor/config.rs
sadhansood: done

crates/walrus-service/src/event/event_processor/catchup.rs
sadhansood: done, returning proper errors now

crates/walrus-service/src/event/event_processor/catchup.rs
sadhansood: done

crates/walrus-service/src/event/event_processor/coordination.rs
sadhansood: fixed

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: i think compiler will complain, because after we call `.await` on the handle, it is moved so in the next iteration it does not have a placeholder which option provides (as `None`)

crates/walrus-service/src/event/event_processor/coordination.rs
sadhansood: i think the two atomic variables are used as simple flags and are completely indepdent with their own publish/observe pairs which means there is no complicated interleaving of writes and reads and simple acquire/release semantic should suffice. The main reason to not use SeqCst would be performance but it will obviously also work but seems unnecessary (although if you still have concerns i am happy to change it)

mlegner: I generally try to err on the side of caution and easier reasoning unless it's performance-critical (which it probably isn't in this case). But I agree that the simpler flags are probably fine in this case.

crates/walrus-service/src/event/event_processor/processor.rs
mlegner: Hmm, I see. I was mainly confused that we run `start_tailing_checkpoints_with_restart_support` even if we don't support the runtime catchup.

crates/walrus-service/src/event/event_processor/catchup.rs
```diff
@@ -184,16 +218,30 @@ impl EventBlobCatchupManager {
...
+                tracing::warn!(?error, "recoverable error during catchup using event blobs");
+                Err(CatchupError::Recoverable(error))
+            }
+            Err(CatchupError::NonRecoverable(error)) => {
+                tracing::error!(
+                    ?error,
+                    "non-recoverable error during catchup using event blobs"
+                );
+                Err(CatchupError::NonRecoverable(error))
             }
```
mlegner, line 241: AFAICT, we do the same match at the callsites of this function. So we could remove the log in this place.

crates/walrus-service/src/event/event_processor/catchup.rs
```diff
@@ -184,16 +218,30 @@ impl EventBlobCatchupManager {
         Ok(lag)
     }
 
+    /// Gets the current lag between the local store and the network
+    pub async fn get_current_lag(&self) -> anyhow::Result<u64> {
```
halfprice, line 209: Nit: the other call site of calculate_lag can also use `get_current_lag()`

crates/walrus-service/src/event/event_processor/coordination.rs
```diff
@@ -0,0 +1,243 @@
...
+    /// Notify that checkpoint tailing has fully stopped and wake any waiters.
+    pub fn notify_tailing_stopped(&self) {
+        self.is_tailing_stopped
+            .store(true, std::sync::atomic::Ordering::Release);
+        self.tailing_stopped_notify.notify_one();
+    }
+
+    /// Wait until checkpoint tailing is stopped, with timeout.
+    /// Returns true if stopped, false on timeout.
+    pub async fn wait_for_tailing_stopped(&self, timeout: std::time::Duration) -> bool {
```
halfprice, line 98: Nit: can we return result<()> here to be more explicit?

Also, can we combine this function with start_catchup_processing_phase? So that the message and wait are done in the same function?

crates/walrus-service/src/event/event_processor/config.rs
```diff
@@ -116,6 +136,11 @@ impl Default for EventProcessorConfig {
             adaptive_downloader_config: Default::default(),
             event_stream_catchup_min_checkpoint_lag: 20_000,
             sampled_tracing_interval: Duration::from_secs(3600),
+            runtime_catchup_lag_threshold: 20_000,
```
halfprice, line 138: Remark: should this be close to NUM_CHECKPOINTS_PER_BLOB, which is 216K at the moment? The latest event blob may not event being created when the lag is 20K

crates/walrus-service/src/event/event_processor/coordination.rs
```diff
@@ -0,0 +1,243 @@
...
+        }
+        match tokio::time::timeout(timeout, self.tailing_stopped_notify.notified()).await {
+            Ok(()) => true,
+            Err(_) => false,
+        }
+    }
+
+    /// Start the download phase of catchup.
+    /// During this phase, event blobs are downloaded in parallel with checkpoint tailing.
+    pub async fn start_catchup_download_phase(&self) {
```
halfprice, line 114: I think this is not called anywhere other than in tests.

Looks like the purpose of setting catchup_active to true is to prevent multiple catch up tasks are initiated, is that the case? Right now, catchup_active is also never set.

crates/walrus-service/src/event/event_processor/metrics.rs
```diff
@@ -13,6 +13,20 @@ walrus_utils::metrics::define_metric_set! {
...
+        #[help = "Whether runtime catchup is currently in progress (1) or not (0)"]
+        runtime_catchup_in_progress: IntGauge[],
+        #[help = "Current lag in checkpoints detected during runtime monitoring"]
+        runtime_lag_current: IntGauge[],
+        #[help = "Number of events recovered during runtime catchup operations"]
+        runtime_catchup_events_recovered: IntCounter[],
+        #[help = "Number of coordination conflicts between catchup and checkpoint downloader"]
+        coordination_conflicts_total: IntCounter[],
+        #[help = "Number of checkpoints deferred due to catchup coordination"]
+        deferred_checkpoints_total: IntCounter[],
```
halfprice: I think except `runtime_lag_current`, all other metrics are not used currently.

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -147,13 +176,23 @@ impl EventProcessor {
...
             system_config,
             runtime_config.db_path.join("recovery"),
             metrics_registry,
+            catchup::CatchupRuntimeConfig {
+                coordination_state: coordination_state.clone(),
+                coordination_timeout: Duration::from_secs(STARTUP_CATCHUP_TIMEOUT_SECS),
+                processing_timeout: Duration::from_secs(STARTUP_CATCHUP_TIMEOUT_SECS),
+            },
         );
+        coordination_state.notify_tailing_stopped();
```
halfprice: question: do we need this call here? Tailing is not started at this point. Also, IIUC, this coordination state is a short lived coordination state, so can we document that this is actually not needed/used during initial catchup?

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -147,13 +176,23 @@ impl EventProcessor {
...
+
         if let Err(e) = catchup_manager
             .catchup(config.event_stream_catchup_min_checkpoint_lag)
```
halfprice: Suggestion: can we create EventBlobCatchupManager only if we need to perform catchup using event blob?

crates/walrus-service/src/event/event_processor/processor.rs
halfprice: Oh, I think I know why this is needed. In catchup_using_event_blobs, it is waiting for tailing to stop.

I feel it's cleaner if we initialize coordination_state with `is_tailing_stopped` set to true. But not too strongly.

Ideally, if there is only one stance of CoordinationState, the one that used in the long running monitor, it's easier to read the code.

In general, I feel the interaction between EventBlobCatchupManager and CoordinationState requires someone to know the code very well to reason about.

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -327,6 +408,70 @@ impl EventProcessor {
...
+
+        let mut tailing_task: Option<JoinHandle<Result<()>>> =
+            Some(self.start_tailing_task(child_cancel_token.clone(), coordination_state.clone()));
+        loop {
+            tokio::select! {
+                msg = coordination_rx.recv() => {
+                    match msg {
+                        Some(CoordinationMessage::StopCheckpointTailing) => {
+                            child_cancel_token.cancel();
+                            match tailing_task.take().expect("tailing task not running").await {
```
halfprice: Suggestion: instead of panic, can we log an error if tailing_task is None and continue? At least this can be true today since catchup_active is not set.

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -327,6 +408,70 @@ impl EventProcessor {
...
+                                Ok(Ok(_)) => tracing::info!("tailing task exited"),
+                                Ok(Err(error)) => tracing::error!(?error, "tailing task error"),
+                                Err(error) => tracing::error!(?error, "tailing task panicked"),
```
halfprice: Suggestion: propagate panic to upper layer to terminate the process. This may cause the node to silently stop downloading checkpoint. This can be done in a followup PR.

halfprice (commented):
Thanks a lot @sadhansood for the PR! The main logic LGTM.

To make this feature more useful, I wonder if we can track the last certified event's latest index, along with checkpoint lag. The motivation is that we create 1 event blob every 12hr or so. So if a node is lagging for 3hrs, event blob may or may not be helpful to catchup. This depends on when the last event blob was created.

crates/walrus-service/src/event/event_processor/config.rs
halfprice: Now I think about it. We probably don't want runtime_catchup_lag_threshold to be close to NUM_CHECKPOINTS_PER_BLOB, but as small as possible. If there is no recent event blob created, we just don't use event blob, and continue tailing checkpoint. If there happens to be a recent event blob, then we can use it.

This also means that we probably need a special logic in catchup_using_event_blobs, so that when there is no event blob downloaded, let's not stop checkpoint tailing, but return sooner.

As of today, I think what happens is that we stop tailing, wait for tailing to stop, and then process event blob, event there is no event blob downloaded.

crates/walrus-service/src/event/event_processor/catchup.rs
sadhansood: done

crates/walrus-service/src/event/event_processor/coordination.rs
sadhansood: I think this function returning boolean is probably fine as callers don't really need error details here (it is just timeout). Also, combining them together in a single function would be ok but also keeping them separate is good for clarity. I don't feel strongly enough about this, so keeping it as is.

crates/walrus-service/src/event/event_processor/config.rs
sadhansood: Yes, thanks this is a good point. I've added logic to not invoke stopping of checkpoint tailing if zero blobs were downloaded.

crates/walrus-service/src/event/event_processor/coordination.rs
sadhansood: Yes, this is fixed now and is being invoked as expected, thanks!

crates/walrus-service/src/event/event_processor/metrics.rs
sadhansood: fixed

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: done, there is only one instance of coordination now.

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: This should be not needed now

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: This is handled now.

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: Handled now.

crates/walrus-service/src/event/event_processor/catchup.rs
sadhansood: done

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -244,17 +237,94 @@ impl EventProcessor {
 
     /// Starts the event processor. This method will run until the cancellation token is cancelled.
     pub async fn start(&self, cancellation_token: CancellationToken) -> Result<(), anyhow::Error> {
-        tracing::info!("starting event processor");
+        if self.config.enable_runtime_catchup {
+            tracing::info!(
+                runtime_catchup_enabled = true,
```
halfprice: Nit: this field is redundant

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -244,17 +237,94 @@ impl EventProcessor {
...
+            self.system_config.clone(),
+            self.recovery_path.clone(),
+            &self.metrics_registry,
+            catchup::CatchupRuntimeConfig {
+                coordination_state: coordination_state.clone(),
+                coordination_timeout: Duration::from_secs(STARTUP_CATCHUP_TIMEOUT_SECS),
+                processing_timeout: Duration::from_secs(STARTUP_CATCHUP_TIMEOUT_SECS),
+            },
+        );
+        coordination_state.notify_tailing_stopped();
```
halfprice, line 268: Suggestion: can we add a comment here why we need to call `notify_tailing_stopped()` (allow event blob catchup to start IIUC)? I think it may not be obvious why we need this here.

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -244,17 +237,94 @@ impl EventProcessor {
...
+        } else {
+            select! {
+                pruning_result = pruning_task => {
+                    cancellation_token.cancel();
+                    pruning_result
+                }
+                tailing_result = tailing_task => {
+                    cancellation_token.cancel();
+                    tailing_result
+                }
```
halfprice, line 328: Nit: I think we just use one select

```
        select! {
            pruning_result = pruning_task => {
                cancellation_token.cancel();
                pruning_result
            }
            tailing_result = tailing_task => {
                cancellation_token.cancel();
                tailing_result
            }
            catchup_result = self.start_runtime_catchup_monitoring(
                cancellation_token.clone(),
                coordination_state.clone(),
            ), if self.config.enable_runtime_catchup => {
                cancellation_token.cancel();
                catchup_result
            }
        }
```

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -364,4 +506,86 @@ impl EventProcessor {
...
+            self.recovery_path.clone(),
+            &self.metrics_registry,
+            catchup::CatchupRuntimeConfig {
+                coordination_state: coordination_state_clone,
+                coordination_timeout: self.config.catchup_coordination_timeout,
+                processing_timeout: self.config.catchup_processing_timeout,
+            },
+        );
+
+        loop {
```
halfprice, line 544: Suggestion: can you add a monitored_scope in the loop so that we know how much time is spent on each iteration. It also helps debugging loop stuck

crates/walrus-service/src/event/event_processor/metrics.rs
```diff
@@ -13,6 +13,14 @@ walrus_utils::metrics::define_metric_set! {
         event_processor_latest_downloaded_checkpoint: IntGauge[],
         #[help = "The number of checkpoints downloaded. Useful for computing the download rate"]
         event_processor_total_downloaded_checkpoints: IntCounter[],
+        #[help = "Number of runtime catchup operations triggered"]
+        runtime_catchup_triggered_total: IntCounter[],
+        #[help = "Duration of runtime catchup operations in seconds"]
+        runtime_catchup_duration_seconds: Histogram[],
+        #[help = "Whether runtime catchup is currently in progress (1) or not (0)"]
+        runtime_catchup_in_progress: IntGauge[],
```
halfprice: These three are still not used.

crates/walrus-service/src/event/event_processor/coordination.rs
halfprice: Hmm, looks like this is still only used in test. If so, can we annotate it with #[cfg(test)]?

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -12,27 +12,40 @@ use sui_types::{
...
+        coordination::CoordinationMessage,
         db::EventProcessorStores,
     },
     events::{IndexedStreamEvent, InitState, StreamEventWithInitState},
 };
 
 /// The maximum number of events to poll per poll.
 const MAX_EVENTS_PER_POLL: usize = 1000;
+/// Startup catchup timeout (seconds) used during initial catchup in EventProcessor::new
+const STARTUP_CATCHUP_TIMEOUT_SECS: u64 = 25 * 60;
```
halfprice: Nit: I feel 25min is too long for waiting for tailing to stop, and too short for processing event blobs (once we have 2 years of events to process). Can we separate them and use a large value, say 2 days, for processing event blob? Not sure if this is long enough. If not, full recovery will never be able to catchup using event blob.

Another option is to change this in a config.

crates/walrus-simtest/tests/simtest_event_blob.rs
```diff
@@ -291,4 +294,120 @@ mod tests {
             .await
             .unwrap();
     }
+
+    /// This integration test simulates pausing checkpoint tailing to trigger runtime catchup.
+    #[ignore = "ignore integration simtests by default"]
+    #[walrus_simtest]
+    async fn test_runtime_catchup_triggers_on_tailing_pause() {
```
halfprice, line 298: Praise: thanks for creating a comprehensive test!

crates/walrus-service/src/event/event_processor/config.rs
```diff
@@ -106,6 +106,26 @@ pub struct EventProcessorConfig {
...
+    /// Enable runtime catchup functionality.
+    pub enable_runtime_catchup: bool,
+    /// Timeout for catchup operations to prevent indefinite blocking.
+    #[serde_as(as = "DurationSeconds")]
+    #[serde(rename = "catchup_coordination_timeout_secs")]
+    pub catchup_coordination_timeout: Duration,
+    /// Timeout for catchup processing to prevent indefinite blocking.
+    #[serde_as(as = "DurationSeconds")]
+    #[serde(rename = "catchup_processing_timeout_secs")]
+    pub catchup_processing_timeout: Duration,
```
halfprice, line 127: Nit: put enable_runtime_catchup before other runtime catchup configs, and clearly add a section for runtime catchup.

sadhansood: done

crates/walrus-service/src/event/event_processor/processor.rs
sadhansood: So, processing event blob timeout is just for reading the blobs stored locally and inserting events into the db and not downloading it. I think 2 days is a stretch but i can separate the two out and make `STARTUP_CATCHUP_TIMEOUT_SECS` 5 minutes and other one 5 hours.

crates/walrus-service/src/event/event_processor/metrics.rs
sadhansood: fixed, thanks!

PR #2459, opened 2025-08-22
fix(sui): optimize RPC requests in the `SuiReadClient`

mlegner:
## Description

This improves the `SuiReadClient` in two ways:
1. Whenever possible, perform RPC requests in parallel (using
   `tokio::try_join!`).
1. Cache the Walrus system and staking objects for a short period of
   time to avoid repeated RPC requests.

The system and staking objects are set to expire within a few seconds or
when the next epoch state can occur, whatever is first. Within the
node's committee service, the cache is explicitly flushed. As a result,
there is little risk of stale data. Despite this, the performance improvement is significant, see below.

The feature can be turned off completely by setting the TTL to 0.

Closes WAL-512.
Closes WAL-335.

## Evaluation
Performed with a release build on my MacBook Pro at home in central Europe with a good Internet connection.

### Store
Storing 100 random minimal-sized blobs on Testnet consecutively with `walrus store --epochs 1 $TESTBLOB` using a non-rate-limited Testnet RPC node.

* **Before**:
Output of `time`: `143.30s user 45.32s system 17% cpu 18:16.42 total`
--> 1096 seconds total, **~11.0 seconds per blob**

* **After**:
Output of `time`: `138.61s user 43.89s system 18% cpu 16:15.64 total`
--> 976 seconds total, **~9.8 seconds per blob**

--> **Latency reduction of ~1.2 seconds per blob (~11%).**

### Info
Issuing 100 `walrus info all` commands using `fullnode.testnet.sui.io`.

* **Before**:
Output of `time`: `10.07s user 4.50s system 5% cpu 4:27.08 total`
--> 267 seconds total, **~2.7 seconds per command**

* **After**:
Output of `time`: `6.90s user 3.13s system 6% cpu 2:45.92 total`
--> 166 seconds total, **~1.7 seconds per command**

--> **Latency reduction of ~1.0 seconds per command (~37%).**

## Test plan

Existing test suite.


mlegner:
I'm currently investigating a bunch of issues with staleness of data. AFAICT, all of those could already be triggered before in case the RPC node was lagging, but they are now visible in tests.

PR #2462, opened 2025-08-25
fix: skip setting latest event epoch when the first event is an epoch change event after node restart

halfprice:
## Description

Previously in #2326 , we found that when the first event after a node restart is an epoch change event, we need to
set the latest event epoch to the previous epoch, since the epoch change hasn't been executed yet. What can happen
now is that after setting the epoch to the previous epoch, the `latest event epoch` may be 2 epochs earlier than the
latest epoch on chain. This can cause blob sync initiated by node recovery task to fail since we try to recover
blobs that are from 2 epochs ago.

The fix for this issue is that instead of setting the epoch to the previous epoch, we don't set `latest event epoch`
when the first event is an epoch change event. This is because after the epoch change event is executed, it will
set it to the latest one. In this case, blob recovery will always uses the epoch of the current event the node is processing.

## Test plan

Originally manifested in test_repeated_node_crash. 100 tests passed.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1309,7 +1309,12 @@ impl StorageNode {
             return Ok(());
         };
 
-        tracing::debug!("checking the first contract event if we're severely lagging");
+        tracing::debug!(
+            "checking the first contract event if we're severely lagging, \
+            first new event epoch: {}, epoch at start: {}",
+            first_new_event_epoch,
+            epoch_at_start
+        );
 
         // Clear the starting epoch, so that we won't make this check again in the current run.
         *maybe_epoch_at_start = None;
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1309,7 +1309,12 @@ impl StorageNode {
             return Ok(());
         };
 
-        tracing::debug!("checking the first contract event if we're severely lagging");
+        tracing::debug!(
+            "checking the first contract event if we're severely lagging, \
+            first new event epoch: {}, epoch at start: {}",
+            first_new_event_epoch,
+            epoch_at_start
```
mlegner: Nit: Can we record the variables as fields?
```suggestion
            first_new_event_epoch,
            epoch_at_start,
            "checking the first contract event if we're severely lagging",
```

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -583,12 +583,16 @@ impl BlobSynchronizer {
...
             .node
             .owned_shards_at_epoch(latest_event_epoch)
-            .unwrap_or_else(|_| {
+            .unwrap_or_else(|error| {
                 tracing::error!(
-                    "shard assignment must be found at the certified epoch {}",
-                    latest_event_epoch
+                    "shard assignment must be found at the certified epoch {}, error: {:?}",
+                    latest_event_epoch,
+                    error
```
mlegner: Nit: Can we record the variables as fields?
```suggestion
                    certified_epoch = latest_event_epoch,
                    ?error,
                    "shard assignment must be found at the certified epoch",
```

PR #2469, opened 2025-08-26
feat: Enable Rocksdb statistics at runtime through config

sadhansood:
## Description

This PR allows us to enable RocksDB statistics at runtime through database config change. Also adds some db compaction related statistics  to be tracked.

### DIFF
--- a/crates/typed-store/src/rocks.rs
+++ b/crates/typed-store/src/rocks.rs
@@ -954,6 +954,34 @@ impl<K, V> DBMap<K, V> {
                     .db_options
                     .get_ticker_count(Ticker::BloomFilterPrefixTruePositive) as i64,
             );
+        db_metrics
+            .op_metrics
+            .rocksdb_compaction_num_bytes_written
+            .with_label_values(&[&rocksdb.db_name()])
+            .set(
+                rocksdb
+                    .db_options
+                    .get_ticker_count(Ticker::CompactWriteBytes) as i64,
+            );
+        db_metrics
+            .op_metrics
+            .rocksdb_compaction_num_bytes_read
+            .with_label_values(&[&rocksdb.db_name()])
+            .set(
+                rocksdb
+                    .db_options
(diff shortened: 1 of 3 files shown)


crates/typed-store/src/rocks.rs
```diff
@@ -954,6 +954,34 @@ impl<K, V> DBMap<K, V> {
                     .db_options
                     .get_ticker_count(Ticker::BloomFilterPrefixTruePositive) as i64,
             );
+        db_metrics
+            .op_metrics
+            .rocksdb_compaction_num_bytes_written
+            .with_label_values(&[&rocksdb.db_name()])
```
halfprice, line 960: question: should we use cf_name instead here? compaction is on cf level, correct?

crates/typed-store/src/rocks.rs
```diff
@@ -954,6 +954,34 @@ impl<K, V> DBMap<K, V> {
...
+        db_metrics
+            .op_metrics
+            .rocksdb_compaction_num_bytes_read
+            .with_label_values(&[&rocksdb.db_name()])
+            .set(
+                rocksdb
+                    .db_options
+                    .get_ticker_count(Ticker::CompactReadBytes) as i64,
+            );
+        db_metrics
```
halfprice, line 975: question: are these just aggregate value of the <iter|get> bytes? If so, I wonder whether we need this aggregated metrics

halfprice (approved):
Thanks @sadhansood for adding more metrics! The PR LGTM. Just have two questions.

crates/typed-store/src/rocks.rs
sadhansood: Statistics is per db in rocksdb unfortunately (per cf is not supported)

crates/typed-store/src/rocks.rs
sadhansood: It potentially is (not sure if it includes read amplification or user read bytes), however I am not sure given the documentation is not really good.

PR #2471, opened 2025-08-26
fix(client): fix JSON parsing of storeQuilt options

mlegner:
## Description

The missing `serde(default)` meant that the `paths` argument always had to be included when using the JSON mode.

## Test plan

Manual test.


nikos-terzo (approved):
Thank you!

PR #2472, opened 2025-08-26
fix: minor improvements to config files

mlegner:
## Description

- Remove the `Option` for the `max_data_in_flight` parameter.
- Include the `additional_rpc_endpoints` in the autogenerated example config.
- Print a more useful error if the client configuration cannot be parsed.
- Improve documentation around the client configuration.

## Test plan

Modified unit test, manual inspection.

### DIFF
--- a/crates/walrus-sdk/src/config.rs
+++ b/crates/walrus-sdk/src/config.rs
@@ -127,22 +127,22 @@ impl ClientConfig {
         context: Option<&str>,
     ) -> anyhow::Result<(Self, Option<String>)> {
         let path = path.as_ref();
-        match walrus_utils::load_from_yaml(path)? {
-            MultiClientConfig::SingletonConfig(config) => {
+        match walrus_utils::load_from_yaml(path) {
+            Ok(MultiClientConfig::SingletonConfig(config)) => {
                 if let Some(context) = context {
                     bail!(
                         "cannot specify context when using a single-context configuration file \
-                    [config_filename='{}', specified_context='{}']",
+                        [config_filename='{}', specified_context='{}']",
                         path.display(),
                         context,
                     )
                 }
                 Ok((config, None))
             }
-            MultiClientConfig::MultiConfig {
+            Ok(MultiClientConfig::MultiConfig {
                 contexts,
                 default_context,
-            } => {
+            }) => {
                 let target_context = context.unwrap_or(&default_context);
                 Ok((
                     contexts.get(target_context).cloned().ok_or_else(|| {
@@ -157,6 +157,12 @@ impl ClientConfig {
                     Some(target_context.to_string()),
                 ))
             }
+            Err(e) => bail!(
+                "unable to parse the client config file: [config_filename='{}', error='{}']\n\
+                see https://docs.wal.app/usage/setup.html#configuration for the correct format",
+                path.display(),
+                e
+            ),
         }
     }
 
@@ -389,7 +395,7 @@ mod tests {
             wallet_config: path/to/wallet
             communication_config:
                 max_concurrent_writes: 42
-                max_data_in_flight: 1000
+                max_data_in_flight: null
                 reqwest_config:
                     total_timeout_millis: 30000
                     http2_keep_alive_while_idle: false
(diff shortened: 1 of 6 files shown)


crates/walrus-sdk/src/config.rs
```diff
@@ -127,22 +127,22 @@ impl ClientConfig {
         context: Option<&str>,
     ) -> anyhow::Result<(Self, Option<String>)> {
         let path = path.as_ref();
-        match walrus_utils::load_from_yaml(path)? {
-            MultiClientConfig::SingletonConfig(config) => {
+        match walrus_utils::load_from_yaml(path) {
```
wbbradley, line 130: I'd prefer that we pull out the return value of `load_from_yaml(path).with_context(...)?` into a local variable, then match on that. I have the feeling the code is easier to read if the error handling is closer to the callsite. This would also keep the nicety of the `?`.

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -37,7 +37,9 @@ pub struct ClientCommunicationConfig {
     /// The maximum number of nodes the client contacts to get a blob status in parallel.
     pub max_concurrent_status_reads: Option<usize>,
     /// The maximum amount of data (in bytes) associated with concurrent requests.
-    pub max_data_in_flight: Option<usize>,
+    // The `serde_as` attribute is added for backwards compatibility.
```
wbbradley, line 40: Do we have a plan to deprecate unintentional `null` here? It might be best to add a new overriding field and issue a deprecation notice, then eventually remove the field.

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -266,7 +265,7 @@ pub(crate) mod default {
         3
     }
 
-    /// This corresponds to 100Mb, i.e., 1 second on a 100 Mbps connection.
+    // This corresponds to 100Mb, i.e., 1 second on a 100 Mbps connection.
```
wbbradley, line 268: Why is this `pub`? Could be `pub(crate)`?

mlegner: This is already `pub(crate)` as the surrounding module is `pub(crate) mod default`.

crates/walrus-sdk/src/config.rs
mlegner: So something like the following?
```rust 
let config = walrus_utils::load_from_yaml(path).context("new error message")?;
match config { 
    // existing code
}
```

crates/walrus-sdk/src/config/communication_config.rs
mlegner: The behavior of an explicit `null` here is the same as before, so I don't think we need any explicit warning or message.

Or is your goal to eventually get rid of the `#[serde_as(deserialize_as = "DefaultOnNull")]` attribute?

crates/walrus-sdk/src/config.rs
wbbradley: Yep!

wbbradley: You can use `with_context` to add the parameters to the string for better error context, as I'm sure you know, but just calling it out.

crates/walrus-sdk/src/config/communication_config.rs
wbbradley: Gotcha. I read it backwards. Ok, so if they had an implicit null value, then their behavior will now change? That seems fine. Explicit null will stay null?

I guess my intuition was that because `crates/walrus-sdk/client_config_example.yaml` has the null explicitly, it would be quite common for folks to have `null` explicitly in their config, which would undermine the goal of this change to some degree.

wbbradley: Oh... nevermind. I see the behavior of DefaultOnNull now.

wbbradley (approved):

crates/walrus-sdk/src/config/communication_config.rs
wbbradley: Ah, that's confusing. I guess I thought that the linter would pick that up as an inconsistency. No worries. Thanks!

PR #2475, opened 2025-08-27
feat: Enable transactions in storage db

sadhansood:
## Description

Enable optimistic transaction through a config. Rollout plan would be to enable it on selective (mysten) nodes and monitor performance before rolling out widely.

## Test plan

Existing tests

### DIFF
--- a/crates/typed-store/src/rocks.rs
+++ b/crates/typed-store/src/rocks.rs
@@ -104,6 +104,8 @@ macro_rules! retry_transaction {
             match status {
                 Err(TypedStoreError::RetryableTransactionError) => {
                     retries += 1;
+                    let metrics = $crate::metrics::DBMetrics::get();
+                    metrics.op_metrics.rocksdb_optimistic_tx_retries_total.inc();
                     // Randomized delay to help racing transactions get out of each other's way.
                     let delay = {
                         let mut rng = ThreadRng::default();
@@ -127,7 +129,15 @@ macro_rules! retry_transaction {
                     }
                     sleep(delay).await;
                 }
-                _ => break status,
+                _ => {
+                    // Observe per-transaction retries in a histogram (0 if no retries).
+                    let metrics = $crate::metrics::DBMetrics::get();
+                    metrics
+                        .op_metrics
+                        .rocksdb_optimistic_tx_retries_per_tx
+                        .observe(f64::from(retries));
+                    break status;
+                }
             }
         }
     }};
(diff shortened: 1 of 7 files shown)


crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -282,6 +282,10 @@ pub struct GlobalDatabaseOptions {
     pub wal_ttl_seconds: Option<u64>,
     /// The size limit for the WAL in MB.
     pub wal_size_limit_mb: Option<u64>,
+    /// If true, open databases using OptimisticTransactionDB instead of the standard DB.
+    /// This enables optimistic transactions for write paths that need transactional semantics.
+    /// Defaults to false.
+    pub use_optimistic_transaction_db: bool,
```
mlegner: AFAIU, we will at some point enable transactions in all cases and remove the configuration parameter. Otherwise, we can never actually implement logic that relies on transactions, correct?

In this case, I'd probably call this something like `experimental_use_optimistic_transaction_db` and mention in the comment that this is just a temporary parameter.

crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -292,6 +296,16 @@ impl Default for GlobalDatabaseOptions {
...
             wal_size_limit_mb: Some(10 * 1024),      // 10 GB,
+            use_optimistic_transaction_db: {
+                #[cfg(any(test, feature = "test-utils"))]
+                {
+                    true
+                }
+                #[cfg(not(any(test, feature = "test-utils")))]
+                {
+                    false
+                }
```
mlegner: Important: We should not use a different value just because the `test-utils` feature is enabled: This can be enabled for various reasons even outside test environments. 

If we want to have a particular set of options for tests, I'd rather create a `default_for_test` function or similar. See the [corresponding function for the `ClientCommunicationConfig`](https://github.com/MystenLabs/walrus/blob/9dc6842e0496f5637ba4f17bf9d496539f1f7f2e/crates/walrus-sdk/src/config/communication_config.rs#L96).

mlegner (commented):
Thanks for making this happen, @sadhansood. One important comment and one nit from my side.

crates/typed-store/src/rocks/errors.rs
```diff
@@ -91,7 +91,12 @@ pub fn typed_store_err_from_bcs_err(err: bcs::Error) -> TypedStoreError {
 
 /// Convert the rocksdb error to the typed store error
 pub fn typed_store_err_from_rocks_err(err: RocksError) -> TypedStoreError {
-    TypedStoreError::RocksDBError(format!("{err}"))
+    match err.kind() {
+        rocksdb::ErrorKind::Busy | rocksdb::ErrorKind::TryAgain => {
+            TypedStoreError::RetryableTransactionError
+        }
```
halfprice, line 104: Suggestion: it is not obvious that these two errors indicate transaction conflicts. Is there any documentation you can point to?

crates/typed-store/src/rocks.rs
```diff
@@ -78,34 +83,126 @@ const ROCKSDB_PROPERTY_TOTAL_BLOB_FILES_SIZE: &CStr =
...
+            let status = $transaction;
+            match status {
+                Err(TypedStoreError::RetryableTransactionError) => {
+                    retries += 1;
+                    let metrics = $crate::metrics::DBMetrics::get();
+                    metrics
+                        .op_metrics
+                        .rocksdb_optimistic_tx_retries_total
+                        .with_label_values(&["typed_store"])
+                        .inc();
```
halfprice: Suggestion: I think this metrics is not very helpful that it just counts the total number of retries. Can we instead store a histogram of per transaction retry? Which also include the total number of retries in the metrics.

the label also seems to only set to `typred_store` at the moment. So we don't really need it.

crates/typed-store/src/rocks.rs
```diff
@@ -1587,14 +1997,14 @@ where
...
 
-    /// Writes a range delete tombstone to delete all entries in the db map
-    /// If the DBMap is configured with ignore_range_deletions set to false,
-    /// the effect of this write will be visible immediately i.e. you won't
-    /// see old values when you do a lookup or scan. But if it is configured
-    /// with ignore_range_deletions set to true, the old value are visible until
-    /// compaction actually deletes them which will happen sometime after. By
-    /// default ignore_range_deletions is set to true on a DBMap (unless it is
-    /// overridden in the config), so please use this function with caution
+    /// Writes a range delete tombstone to delete all entries in the db map.
```
halfprice, line 1833: Important: many of these `.` added to the end of the line are incorrect. Can you go through them and update accordingly? There are many such cases in this PR.

crates/walrus-service/src/node/storage/database_config.rs
sadhansood: done

PR #2476, opened 2025-08-28
feat: re-enable sliver data consistency check

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -229,15 +295,22 @@ impl BlobEventProcessor {
...
         let mut senders = Vec::with_capacity(num_workers);
         let mut workers = Vec::with_capacity(num_workers);
+        let mut background_per_processor_pending_event_count = Vec::with_capacity(num_workers);
         for worker_index in 0..num_workers {
             let (tx, rx) = mpsc::unbounded_channel();
             senders.push(tx);
-            let mut background_processor =
-                BackgroundEventProcessor::new(node.clone(), blob_sync_handler.clone(), rx);
+            let pending_event_count = Arc::new(AtomicU64::new(0));
+            background_per_processor_pending_event_count.push(pending_event_count.clone());
```
mlegner: Nit: I don't think we need the `clone` here.
```suggestion
            background_per_processor_pending_event_count.push(pending_event_count);
```

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -259,6 +333,7 @@ impl BlobEventProcessor {
             background_processor_senders: senders,
             _background_processors: workers,
             sequential_processor,
+            background_per_processor_pending_event_count,
```
mlegner, line 339: Question: Just to be sure: If we have a sequential processor, we don't need to keep track of the pending event counts?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -316,12 +391,58 @@ impl BlobEventProcessor {
             )
             .inc();
 
+            // Increment the counter and create a guard that will decrement it when dropped
+            let current_pending_event_count = self.background_per_processor_pending_event_count
```
mlegner: Nit: As we use `self.background_per_processor_pending_event_count[processor_index]` multiple times, it might be nicer to store it in a local variable.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -316,12 +391,58 @@ impl BlobEventProcessor {
...
+            return;
+        }
+
+        // Check if any of the background processors still have pending events.
+        while self
+            .background_per_processor_pending_event_count
+            .iter()
+            .any(|c| c.load(Ordering::SeqCst) > 0)
+        {
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
mlegner: Suggestion: Decrease this sleep duration to something like 5-10ms. The check is pretty fast, so it doesn't hurt to check more often. Also, this could be an explicit constant or a configuration parameter.

mlegner (approved):
Thanks so much, @halfprice, this is great.

Just a few minor comments from my side. Maybe @liquid-helium or @sadhansood could do another brief review after any changes.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -316,12 +391,58 @@ impl BlobEventProcessor {
...
+            // sequentially in the same thread, and no events are processed in the background.
+            // Therefore, we don't need to wait, and can return immediately.
+            return;
+        }
+
+        // Check if any of the background processors still have pending events.
+        while self
+            .background_per_processor_pending_event_count
+            .iter()
+            .any(|c| c.load(Ordering::SeqCst) > 0)
```
wbbradley, line 447: I don't think we need sequential consistency here if I'm understanding that the count will always be descending.

crates/walrus-service/src/node/blob_event_processor.rs
wbbradley: Or 1ms. This check is going to run in probably less than 1 microsecond. How long are these background processes running generally?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -316,12 +391,58 @@ impl BlobEventProcessor {
...
+                event_handle,
+                blob_event,
+                _guard: guard,
+            };
+
             self.background_processor_senders[processor_index]
-                .send((event_handle, blob_event))
+                .send(tracked_event)
                 .map_err(|e| {
                     anyhow::anyhow!("failed to send event to background processor: {}", e)
```
sadhansood, line 428: If this send fails (due to queue full), do we end up with higher metric counts because we are incrementing them before the send? Should we increment them after send

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: Could this counter briefly become 0 before the arrival of a new event but with no pending events at the moment?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -212,13 +271,20 @@ pub struct BlobEventProcessor {
...
     // this processor. This is to keep the same behavior as before BackgroundEventProcessor.
     // INVARIANT: sequential_processor must be Some if background_processor_senders is empty.
     sequential_processor: Option<Arc<BackgroundEventProcessor>>,
+
+    // The number of events that are pending to be processed in each background processor.
+    // Each counter is shared with individual BackgroundEventProcessor.
+    //
+    // We use per background processor count to avoid high contention on the Atomic variable when
+    // tracking the total number of pending events to be processed.
+    background_per_processor_pending_event_count: Vec<Arc<AtomicU64>>,
```
sadhansood: nit: idk for sure, but would `AtomicU32` be enough, wondering if it saves any cache bandwidth.

sadhansood: oh i also wonder if there will be false sharing because all the atomics are packed tightly back to back and end up in the same cache line. Might also be worth testing if adding padding helps with performance.

sadhansood: i wonder if it might be useful to have this as a util, something like `ShardedCounter` with api's like `inc`, `dec` and `total`

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Correct, because the pending event count will always be 0 since the observer, BlobEventProcessor::process_event will always process them online.

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Done.

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: A single task can run ranging before sub million seconds, to 100s of millionseconds.

Thanks for the suggestion! I moved the value into an explicit const.

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: It's a single producer, multiple consumer workload.

So far, I've been preferring simplicity. I'm not sure other consistency model will suffice or not. What I read is that if the contention is low, the overhead of SeqCst is also low, so I think this is probably fine for now.

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: The send cannot fail unless the receiver side quit. We are using an unbounded queue.

Also, send failure should terminate the process.

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Great suggestion! let me change this to u32 in this PR, and work on the util in a follow up PR.

PR #2480, opened 2025-08-28
fix: various optimizations and cleanup related to RS encoding

mlegner:
## Description

This contains several optimizations that are possible after the switch from RaptorQ to Reed-Solomon codes. In particular, as Reed-Solomon guarantees decoding given a fixed number of symbols, certain code paths are no longer necessary, such as the `decode_sliver_by_sliver` function.

In addition, this prevents copying data in some cases.

This also contains some refactoring and cleanup: 
- Improve the return types for decoding.
- Convert some panics to errors.
- Remove the no-longer-needed `BlobDecoderEnum`.
- Rename the `EncodingConfigTrait` to `EncodingFactory`.
- Improve various docstrings.

This should not contain any functional change as the removed codepaths are no longer taken with Reed-Solomon encoding anyway.

Closes WAL-611
Closes WAL-621

## Test plan

Existing test suite (slightly adapted).


crates/walrus-core/src/encoding/basic_encoding.rs
```diff
@@ -126,38 +121,79 @@ impl ReedSolomonEncoder {
...
     #[inline]
     pub fn symbol_size(&self) -> NonZeroU16 {
         self.symbol_size
     }
 
-    /// Returns an iterator over all `n_shards` source and repair symbols.
-    // TODO (WAL-611): Return an iterator over all symbols.
-    pub fn encode_all(&mut self) -> Vec<Vec<u8>> {
+    /// Returns a vector of all `n_shards` source and repair symbols.
+    pub fn encode_all(mut self) -> Vec<Vec<u8>> {
```
giac-mysten, line 150: Question: since i don't recall, why do we want to strictly enforce the single-use of the encoders/decoders?

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -451,28 +451,6 @@ impl<'a> ExpandedMessageMatrix<'a> {
     }
 }
 
-/// A wrapper around the blob decoder for different encoding types.
```
giac-mysten, line 454: Question: I see the YAGNI motivation behind all the red here, but it is foreseeable that we may add other encodings in the future. Would you opt for another way of selecting them? Is this because we have the one-shot `decode` on the `EncodingFactory`?

crates/walrus-core/src/encoding/basic_encoding.rs
```diff
@@ -37,34 +46,34 @@ pub trait Decoder {
...
 /// Wrapper to perform a single encoding with Reed-Solomon for the provided parameters.
-pub struct ReedSolomonEncoder {
+pub struct ReedSolomonEncoder<'a> {
     encoder: reed_solomon_simd::ReedSolomonEncoder,
     n_source_symbols: NonZeroU16,
     // INV: n_source_symbols <= n_shards.
     n_shards: NonZeroU16,
     symbol_size: NonZeroU16,
-    source_symbols: Vec<Vec<u8>>,
+    source_symbols: Vec<&'a [u8]>,
```
giac-mysten, line 63: Nice call!

crates/walrus-core/src/encoding/config.rs
```diff
@@ -109,10 +160,17 @@ pub trait EncodingConfigTrait {
...
     }
 
-    /// Returns the number of valid symbols required to recover a sliver of [`EncodingAxis`] `T`.
+    /// Returns the number of valid symbols required to recover a sliver of [`EncodingAxis`] `A`.
+    #[inline]
+    fn n_symbols_for_recovery<A: EncodingAxis>(&self) -> RequiredCount {
+        self.n_slivers_for_reconstruction::<A::OrthogonalAxis>()
+    }
+
+    /// Returns the number of source symbols required to reconstruct a blob using slivers of
```
giac-mysten: Suggestion: or am I getting this wrong?
```suggestion
    /// Returns the number of slivers required to reconstruct a blob using slivers of
```

crates/walrus-sdk/src/client.rs
```diff
@@ -1817,25 +1817,18 @@ impl<T> WalrusNodeClient<T> {
...
-                >= self
-                    .encoding_config
-                    .get_for_type(metadata.metadata().encoding_type())
-                    .n_source_symbols::<U>()
-                    .get()
-                    .into()
-        };
-        requests
+
+        // Note: The following code may have to be changed if we add encodings that require a
```
giac-mysten, line 1823: Nice safeguard 👍

crates/walrus-sdk/src/client.rs
```diff
@@ -1844,118 +1837,65 @@ impl<T> WalrusNodeClient<T> {
...
+                        metadata.blob_id(),
+                        metadata.metadata().unencoded_length(),
+                        slivers,
+                    )
+                    .map_err(|e| {
+                        ClientErrorKind::Other(
+                            anyhow::anyhow!(
+                                "unable to decode blob from a sufficient number of slivers: {e}; \
+                                this should not happen"
+                            )
```
giac-mysten: Question: Should we panic here instead? This is a denotes a mistake in the sliver verification logic?

crates/walrus-sdk/src/utils.rs
```diff
@@ -270,7 +270,8 @@ where
 }
 
 /// Represents the reason why the `WeightedFutures::execute_weight` completed.
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[must_use]
```
giac-mysten, line 274: Question: interesting addition, what's the reason?

crates/walrus-sdk/src/client.rs
```diff
@@ -1844,118 +1837,65 @@ impl<T> WalrusNodeClient<T> {
...
-            .await
-        }
-    }
+                    },
+                );
 
-    /// Decodes the blob of given blob ID by requesting slivers and trying to decode at each new
-    /// sliver it receives.
-    #[tracing::instrument(level = Level::ERROR, skip_all)]
-    async fn decode_sliver_by_sliver<'a, I, Fut, U>(
```
giac-mysten, line 1903: happy to see this go 😬

giac-mysten (approved):
Thanks @mlegner , i really like the refactor, especially the simplification of `request_slivers_and_decode`! I just have a few questions for my own understanding.

crates/walrus-sdk/src/utils.rs
mlegner: I think anyone using the `WeightedFutures` should in some way be interested in why they completed. At least it should be an active decision to not look at the completion reason.

giac-mysten: hehe very nice to have.

crates/walrus-sdk/src/client.rs
mlegner: Good point. This indeed indicates a severe issue in our encoding.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: TBH, I wasn't really happy with that construct to begin with. So even if we would need something like this again, I'd probably implement it differently.

crates/walrus-core/src/encoding/config.rs
mlegner: Whoops, copy-🍝...

crates/walrus-core/src/encoding/basic_encoding.rs
mlegner: Good question. I think it's not really necessary, but it helps enforce more efficient code patterns. I don't really see a reason why you want to perform the same encoding multiple times. Instead, you should just encode it once and reuse the result.

But I don't feel very strongly about this and am happy to change it.

giac-mysten: no, i tend to agree. lgtm.

PR #2482, opened 2025-08-28
chore: set version bump pr to auto-merge

ebmifa:
## Description
set version bump pr to auto-merge

## Test plan
👀


mlegner:
I don't think we should have auto-merge on these PRs as they often introduce issues, for example #2481.

mlegner:
@ebmifa Can we revert this?

PR #2487, opened 2025-08-28
ci: add option to generate walrus version bump PR

ebmifa:
## Description
Add option to generate walrus version bump PR

## Test plan
👀

### DIFF
--- a/.github/workflows/gen-sui-upgrade-version-pr.yml
+++ b/.github/workflows/gen-sui-upgrade-version-pr.yml
@@ -4,25 +4,30 @@ run-name: Generate Sui Version Upgrade PR
 on:
   workflow_dispatch:
     inputs:
-      new_tag:
+      new_sui_tag:
         description: "New Sui tag (e.g., testnet-v1.55.0), if empty will default to current sui testnet version"
         required: false
         type: string
+      bump_walrus_version:
+        description: "Create Walrus version bump PR"
+        required: false
+        type: boolean
+        default: false
 
   repository_dispatch:
     types: [update-walrus-sui-testnet-tag]
 
-concurrency: ${{ github.workflow }}-${{ inputs.new_tag }}
+concurrency: ${{ github.workflow }}-${{ inputs.new_sui_tag }}
 
 permissions:
   contents: write
   pull-requests: write
 
 env:
-  NEW_TAG: ${{ github.event.client_payload.new_tag || inputs.new_tag || '' }}
+  NEW_SUI_TAG: ${{ github.event.client_payload.new_tag || inputs.new_sui_tag || '' }}
 
 jobs:
-  bump:
+  bump-sui-testnet-version:
     name: Bump Sui Testnet Version
     runs-on: ubuntu-ghcloud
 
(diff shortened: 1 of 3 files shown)


.github/workflows/gen-sui-upgrade-version-pr.yml
```diff
@@ -4,25 +4,30 @@ run-name: Generate Sui Version Upgrade PR
...
-      new_tag:
+      new_sui_tag:
         description: "New Sui tag (e.g., testnet-v1.55.0), if empty will default to current sui testnet version"
         required: false
         type: string
+      bump_walrus_version:
+        description: "Create Walrus version bump PR"
+        required: false
+        type: boolean
+        default: false
```
mlegner: In my opinion, these two PRs should be created separately: The Sui bump should happen after a Sui testnet release, the version bump after the branch cut. These are typically a few days apart.

scripts/bump_sui_testnet_version.sh
```diff
@@ -105,12 +105,15 @@ This PR updates the Sui testnet version to ${NEW_TAG}
 EOF
 )
 
+# Create PR
 PR_URL=$(gh pr create \
   --base main \
   --head "$BRANCH" \
   --title "chore: bump Sui version to ${NEW_TAG}" \
-  --reviewer "ebmifa,mlegner,wbbradley" \
+  --reviewer "MystenLabs/walrus-pe" \
```
mlegner: Given that there are normally breaking changes during a version bump, I think this needs to be handled by someone from the Walrus core team.

scripts/bump_sui_testnet_version.sh
```diff
@@ -105,12 +105,15 @@ This PR updates the Sui testnet version to ${NEW_TAG}
...
   --body "$BODY" \
   2>&1 | grep -Eo 'https://github.com/[^ ]+')
 
+# Setting the PR to auto merge
+gh pr merge --auto --squash --delete-branch "$BRANCH"
```
mlegner, line 118: I don't fully understand the semantics here. Enabling "automerge" is fine, as long as all mandatory checks (approval, CI) are enforced.

scripts/bump_walrus_version.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# Parse "X.Y.Z" from $WALRUS_VERSION into major/minor/patch, then set NEW_WALRUS_VERSION to "X.(Y+1).Z".
+IFS=. read -r major minor patch <<<"$WALRUS_VERSION"; NEW_WALRUS_VERSION="$major.$((minor+1)).$patch"
+
+# Setup new branch for staging
+STAMP="$(date +%Y%m%d%H%M%S)"
+BRANCH="${GITHUB_ACTOR}/walrus-v${NEW_WALRUS_VERSION}-version-bump-${STAMP}"
+git checkout -b "$BRANCH"
+
+# Update the version in Cargo.toml and openrpc.json
```
mlegner: Nit: There is no `openrpc.json` file.

.github/workflows/gen-sui-upgrade-version-pr.yml
ebmifa: @mlegner Since the dispatch will not trigger it https://github.com/MystenLabs/sui/blob/main/.github/workflows/create-release-announce.yml#L175
and it will be mostly triggered manually, and just for version bump for walrus. But I can separate this into a  sepaate workflow, just to cleaner

scripts/bump_sui_testnet_version.sh
ebmifa: @mlegner which handles exactly? we can add them to the walrus-pe https://github.com/orgs/MystenLabs/teams/walrus-pe group. We can also rename the group if it's confusing

scripts/bump_sui_testnet_version.sh
ebmifa: Yep, it will auto merge only when all the required tests pass and it has been approved

scripts/bump_sui_testnet_version.sh
mlegner: Hmm... In the past this was mainly my task, but this might become @wbbradley or @halfprice in the future. I think some kind of "walrus-maintenance" group would be useful for this.

scripts/bump_walrus_version.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# push branch
+git commit -m "$BODY"
+git push -u origin "$BRANCH"
+
+# Create PR
+PR_URL=$(gh pr create \
+  --base main \
+  --head "$BRANCH" \
+  --title "$BODY" \
```
wbbradley:
```suggestion
  --title "chore: $BODY" \
```

PR #2492, opened 2025-08-29
feat: simtest cluster tear down

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/.github/workflows/code.yml
+++ b/.github/workflows/code.yml
@@ -155,9 +155,7 @@ jobs:
     if: ${{ needs.diff.outputs.isRelevantForRustTests == 'true' }}
     runs-on: ubuntu-ghcloud
     env:
-      # TODO(zhewu): set this back to 5 after proper walrus cluster cleanup is implemented after
-      # each test run.
-      MSIM_TEST_NUM: 1
+      MSIM_TEST_NUM: 5
       # Turn off Sui logging since it can be very distractive in CI
       RUST_LOG: simtest=info,error,sui_=off,consensus_core=off
     steps:


crates/walrus-sui/src/client/retry_client/retriable_rpc_client.rs
```diff
@@ -136,27 +144,31 @@ impl LazyClientBuilder<FallibleRpcClient> for LazyFallibleRpcClientBuilder {
...
-                        ))
-                    })?
-                    .map_err(|e| {
-                        FailoverError::FailedToGetClient(format!(
-                            "Client validation failed for {rpc_url:?}: {e:?}"
-                        ))
-                    })?;
-                }
-                Ok(client)
+                    .get_strategy(StdRng::from_entropy().next_u64()),
```
liquid-helium, line 153: nit: is it possible to capsulate this retry logic into a new_with_retry() for the FallibelRpcClient? If this is not the only caller for the FallibleRpcClient, it can be also beneficial for other cases.

halfprice: Thanks for the review @liquid-helium ! This retry contains two parts: retry error in FallibleRpcClient::new, and also retry error when calling ensure_experimental_rest_endpoint_exists. So we also need the retry logic for ensure_experimental_rest_endpoint_exists as well.

Also just check that this is the only place where FallibelRpcClient::new is called.

PR #2494, opened 2025-08-31
docs: add section on avoiding duplicate-content SEO in custom portals

DionisisLougaris:
## Description

This pull request adds new documentation to help users avoid duplicate-content SEO issues when deploying custom Walrus Sites portals. The main focus is on providing guidance for canonicalization and best practices to ensure search engines do not penalize sites for duplicate content.

Documentation updates:

* Added a new section to the table of contents in `SUMMARY.md` for avoiding duplicate-content SEO in custom portals.
* Created a new page, `avoid-duplicate-content-seo.md`, with detailed recommendations on canonical host policies, handling of base36 and BYOD domains, implementation examples for redirects and headers, and advice for managing multiple SuiNS names for a single site.

### DIFF
--- a/docs/book/walrus-sites/avoid-duplicate-content-seo.md
+++ b/docs/book/walrus-sites/avoid-duplicate-content-seo.md
@@ -43,8 +43,8 @@ In the HTML `<head>`:
 
 ## Recommended Practices
 
-- **Canonical host policy**  
-  Choose one hostname type per site (custom domain > SuiNS > base36 is a good priority).  
+- **Canonical host policy**
+  Choose one hostname type per site (custom domain > SuiNS > base36 is a good priority).
   All other hosts should include canonical hints pointing to that choice.
 
 - **Base36 subdomains**
@@ -52,6 +52,7 @@ In the HTML `<head>`:
   - Disable if not needed: `B36_DOMAIN_RESOLUTION_SUPPORT=false`.
   - If enabled, add canonical hints (`<link rel="canonical">` or `Link: rel="canonical"`).
   - Optionally, block them from indexing with:
+
     ```http
     X-Robots-Tag: noindex
     ```
@@ -63,6 +64,6 @@ In the HTML `<head>`:
   - It’s always a good practice to add canonical hints in your HTML or headers pointing to your chosen custom domain,
     to avoid duplicate indexing across different hosts.
 
-- **Multiple SuiNS names → one site**  
+- **Multiple SuiNS names → one site**
   This can happen permissionlessly. If you own the extra names, add canonical hints in your site HTML
   to point to your preferred domain.
--- a/docs/book/walrus-sites/avoid-duplicate-content-seo.md
+++ b/docs/book/walrus-sites/avoid-duplicate-content-seo.md
@@ -1,14 +1,14 @@
 # Avoiding Duplicate-Content SEO in Custom Portals
 
-When you deploy your own Walrus Sites portal, you may expose the same site under multiple hostnames:
-
 > **Note:** This guidance does not require any changes to the Walrus Sites portal itself.
 > It is focused on **site-level fixes** (HTML tags or HTTP headers) that developers can
 > add when hosting their own portals.
 
+When you deploy your own Walrus Sites portal, you may expose the same site under multiple hostnames:
+
 - **[SuiNS](https://suins.io/)** – human-readable names registered on-chain, e.g. `snowreads.wal.app`.
 - **Base36** – subdomains derived directly from the site object ID, e.g. `1myb…xd.portal.com`.
-- **[BYOD](https://docs.wal.app/walrus-sites/bring-your-own-domain.html)** – Bring Your Own Domain,
+- **[Custom domains](https://docs.wal.app/walrus-sites/bring-your-own-domain.html)** – Bring Your Own Domain,
   e.g. `example.com`.
 
 Search engines treat the same content served at multiple hosts as duplicates,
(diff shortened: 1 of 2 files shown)


docs/book/walrus-sites/avoid-duplicate-content-seo.md
```diff
@@ -0,0 +1,53 @@
...
+which can dilute ranking signals.
+
+The official `wal.app` portal avoids this by serving only **SuiNS domains** with base36 and BYOD disabled.
+If you enable base36 or BYOD on your own portal, you should enforce canonicalization.
+
+## Recommended Practices
+
+- **Canonical host policy**
+
+  - Choose one canonical type per site (e.g., BYOD > SuiNS > base36).
```
Tzal3x: question: does this mean that base36 is the canonical and BYOD and SuiNS will redirect to base36?

docs/book/walrus-sites/avoid-duplicate-content-seo.md
```diff
@@ -0,0 +1,53 @@
...
+  - Redirect all other hosts to the canonical one.
+
+- **Base36 subdomains**
+
+  - Disable if not needed: `B36_DOMAIN_RESOLUTION_SUPPORT=false`.
+  - If enabled, redirect (`301`) base36 → canonical host, or add `X-Robots-Tag: noindex` to base36 responses.
```
Tzal3x: question: have you tried doing this? We don't currently support redirects in the portal unless you mean defining a custom `Route`. But custom routes redirect to other routes, not other domains.  Also the status code wouldn't change to 301 with the current implementation. Can you elaborate on your reasoning here?

docs/book/walrus-sites/avoid-duplicate-content-seo.md
```diff
@@ -0,0 +1,53 @@
+# Avoiding Duplicate-Content SEO in Custom Portals
+
+When you deploy your own Walrus Sites portal, you may expose the same site under multiple hostnames
+(e.g., **SuiNS**, **base36**, or **BYOD**). Search engines treat this as duplicate content,
```
Tzal3x: suggestion: Let's add a link or footnote to each SuiNS, base36, BYOD which would explain what each one of this is.

docs/book/walrus-sites/avoid-duplicate-content-seo.md
```diff
@@ -0,0 +1,53 @@
...
+
+  - Add redirects or headers in `portal/server/src/main.ts` after host parsing.
+  - Example redirect:
+
+    ```ts
+    return new Response(null, {
+      status: 301,
+      headers: { Location: canonicalUrl },
+    });
+    ```
```
Tzal3x: Blocking: I'm not confident that this is clear enough. Do you mean that every response in your portal would return this response? How would the `canonicalUrl` be constructed? Plus, what would happen to the other portals that are out of your control but still serve your content?

docs/book/walrus-sites/avoid-duplicate-content-seo.md
DionisisLougaris: No. It’s just a priority order suggestion. If BYOD is enabled, that’s canonical. If not, then SuiNS. If neither exists and base36 is supported, would base36 be the fallback.

docs/book/walrus-sites/avoid-duplicate-content-seo.md
DionisisLougaris: No, I haven’t tried it. You’re correct, the portal today doesn’t have a way to do permanent redirects (`301`) from one domain to another. As I understand, the only redirect feature we got is for moving between paths on the same domain, using a temporary redirect (`302`).

**My point was** that if someone runs their own portal and chopses to expose base36/BYOD, they would need to add their own redirect logic to point everything to their chosen canonical host. For wal.app specifically, if my assumptions [here](https://www.notion.so/mystenlabs/Walrus-Sites-1d16d9dcb4e9804badceea082d69ffb2?p=25b6d9dcb4e980b0996bf31418ca3ca9&pm=c) are correct, **we don’t need to do anything because base36 and BYOD are disabled.**

docs/book/walrus-sites/avoid-duplicate-content-seo.md
DionisisLougaris: No, this wouldn’t mean every response turns into a `301`. The idea is only to send a `301 `when a request comes in on a non-canonical host. If the request already hits the canonical host, the portal just serves the content as normal.

The canonicalUrl would be built with a simple priority rule:

- If BYOD is configured, that’s the canonical domain.
- If not, and a SuiNS is set, use the SuiNS hostname.
- If neither exist but base36 is exposed, fallback to base36.

When constructing the redirect target, you keep the same path and query from the incoming request. For example, if the incoming request is https://snowreads.wal.app/math and the canonical is BYOD snowreads.com, then the 301 Location would be https://snowreads.com/math

For base36 → SuiNS specifically, you’d need an operator-supplied mapping of objectId → preferred SuiNS, **since the portal can’t infer which SuiNS is “the right one”** if multiple names point to the same site. If you don’t have that mapping, **the safer option is to either skip the redirect for base36 or disable base36 for public traffic.**

As for other portals we don’t control, we can’t force them to redirect. The best we can do is add canonical signals on our side (Link: rel="canonical" headers or <link rel="canonical"> in **HTML**) and provide guidance so operators who expose base36 or BYOD know they should implement redirects or noindex. As mentioned above, for wal.app specifically, none of this is needed, since base36 and BYOD are disabled.

docs/book/walrus-sites/avoid-duplicate-content-seo.md
Tzal3x: > You’re correct, the portal today doesn’t have a way to do permanent redirects (301) from one domain to another. 

Let's remove this suggestion then if it's not feasible atm. 👌 

> we don’t need to do anything because base36 and BYOD are disabled

_We_ as walrus-sites "core" developers, don't have to do anything, correct. But the community devs that deploy their websites as walrus sites do have to do some work, maybe by using canonical hints. 

Nevertheless, that's still not clear to the current version of the doc. 

Let's include more details and in the end please add an example use case with:
1. A scenario where duplicate SEO content is parsed from a web crawler, meaning that the developer must take action.
2. And then, how would the developer resolve the issue? Show them the actual steps with an example.

DionisisLougaris: Sure thing, will do!

docs/book/walrus-sites/avoid-duplicate-content-seo.md
Tzal3x: > For example, if the incoming request is https://snowreads.wal.app/math and the canonical is BYOD snowreads.com, then the 301 Location would be https://snowreads.com/math

What do you mean by 301 Location? As mentioned above, we don't support redirects at the moment. I think a catch-all solution is using just canonical hints.

> As for other portals we don’t control, we can’t force them to redirect.

Why would we have a different approach for the portal we control and third-party portals then? I think the whole doc should be about resolving the duplicate-content SEO on the frontend's code rather than messing with the portal's functionality.

docs/book/walrus-sites/avoid-duplicate-content-seo.md
```diff
@@ -0,0 +1,75 @@
...
+  - Enable with care: `BRING_YOUR_OWN_DOMAIN=true`.
+  - Decide if BYOD or SuiNS is canonical, and ensure non-canonical hosts send canonical hints or `noindex`.
+
+- **Multiple SuiNS names → one site**
+  This can happen permissionlessly. If you own the extra names, add canonical hints to consolidate them.
+  Otherwise, add a `<link rel="canonical">` in the HTML to indicate your preferred host.
+
+---
+
+## Example Scenario
```
Tzal3x, line 23: Nit: let's add this section before the `## Recommended Practices` section.

docs/book/walrus-sites/avoid-duplicate-content-seo.md
```diff
@@ -0,0 +1,75 @@
...
+  - Alternatively, block them from indexing with:
+
+    ```http
+    X-Robots-Tag: noindex
+    ```
+
+- **BYOD domains**
+
+  - Enable with care: `BRING_YOUR_OWN_DOMAIN=true`.
+  - Decide if BYOD or SuiNS is canonical, and ensure non-canonical hosts send canonical hints or `noindex`.
```
Tzal3x: >  - Enable with care: `BRING_YOUR_OWN_DOMAIN=true`.

question: What do you mean by enabling with care the BYOD? In the portal side enabling this would mean that the portal is served by the site developer and also it only serves one site: the owner's. I'm not sure that I understand the tip here. Can you elaborate?

Tzal3x (approved):
Thank you @DionisisLougaris! Left two more comments. When you update that part, feel free to merge the PR. 🙇‍♂️

PR #2499, opened 2025-09-02
feat: Make shard sync directly ingest data in sst files

sadhansood:
## Description

This PR introduces an optional, high-throughput write path for shard sync that ingests sliver data via RocksDB SST files instead of per-key batched writes. It also adds typed-store utilities to build and ingest SSTs safely. The feature is disabled by default and can be toggled per node via config.

The primary motivation for doing this work is during shard sync, the node downloading the shard does writes to shard column family at a high rate which causes compaction to back up and slow down (or potentially stall) *writes to the entire db* which is not great. Writing data through sst files is also the recommended way of bulk ingesting data according to RocksDB wiki.

## Test plan

Added unit tests, and the rollout will be done initially only on testnet.

### DIFF
--- a/crates/typed-store/src/rocks.rs
+++ b/crates/typed-store/src/rocks.rs
@@ -32,11 +32,13 @@ use rocksdb::{
     DBPinnableSlice,
     DBWithThreadMode,
     Error,
+    IngestExternalFileOptions,
     LiveFile,
     MultiThreaded,
     OptimisticTransactionDB,
     OptimisticTransactionOptions,
     ReadOptions,
+    SstFileWriter,
     Transaction,
     WriteBatch,
     WriteBatchWithTransaction,
@@ -521,7 +523,24 @@ impl RocksDB {
         delegate_call!(self.compact_range_cf_opt(cf, start, end, opt))
     }
 
-    /// Flush the database.
+    /// Ingest external SST files into a specific column family. Ingesting into an existing
+    /// column family is allowed. If any ingested keys already exist in the target CF,
+    /// RocksDB will, by default, assign a global sequence number to the ingested entries so
+    /// that they are treated as the latest versions. Reads will then observe the ingested
+    /// values for those conflicting keys. Behavior can be tuned via `IngestExternalFileOptions`
+    /// (e.g., file move vs copy, and other ingestion modes provided by RocksDB).
+    /// Note: Files must reside on the same filesystem for move/link optimizations.
+    pub fn ingest_external_file_cf<P: AsRef<Path>>(
+        &self,
+        cf: &impl AsColumnFamilyRef,
+        paths: &[P],
+        opts: &IngestExternalFileOptions,
+    ) -> Result<(), rocksdb::Error> {
+        let v: Vec<&Path> = paths.iter().map(|p| p.as_ref()).collect();
+        delegate_call!(self.ingest_external_file_cf_opts(cf, opts, v))
+    }
+
+    /// Flush the database
     #[allow(dead_code)]
     pub fn flush(&self) -> Result<(), TypedStoreError> {
         delegate_call!(self.flush()).map_err(|e| TypedStoreError::RocksDBError(e.into_string()))
@@ -966,6 +985,23 @@ impl<K, V> DBMap<K, V> {
         self.rocksdb
             .flush_cf(&self.cf()?)
             .map_err(|e| TypedStoreError::RocksDBError(e.into_string()))
+    }
+
+    /// Ingest external SST files into this map's column family.
+    /// Ingesting into an existing column family is allowed. If any ingested keys
+    /// already exist in the target CF, RocksDB will, by default, assign a global
+    /// sequence number to the ingested entries so that they are treated as the
+    /// latest versions. Reads will then observe the ingested values for those
+    /// conflicting keys. Behavior can be tuned via `IngestExternalFileOptions`
+    /// (e.g., file move vs copy, and other ingestion modes provided by RocksDB).
+    pub fn ingest_external_file<P: AsRef<Path>>(
+        &self,
+        paths: &[P],
+        opts: &IngestExternalFileOptions,
+    ) -> Result<(), TypedStoreError> {
+        self.rocksdb
+            .ingest_external_file_cf(&self.cf()?, paths, opts)
+            .map_err(typed_store_err_from_rocks_err)
     }
 
     fn get_int_property(
@@ -2514,7 +2550,7 @@ pub fn safe_drop_db(path: PathBuf) -> Result<(), rocksdb::Error> {
     rocksdb::DB::destroy(&rocksdb::Options::default(), path)
 }
 
-/// Populate missing column families.
+/// Populate missing column families
 fn populate_missing_cfs(
     input_cfs: &[(&str, rocksdb::Options)],
     path: &Path,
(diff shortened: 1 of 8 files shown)


crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -954,9 +986,68 @@ impl ShardStorage {
...
+            .map_err(|e| SyncShardClientError::Internal(anyhow::anyhow!(e)))?;
+        let sst_path = sst_dir.join(format!("ingest-{}.sst", Self::current_nanos()));
+        build_sst_from_typed(&sst_path, entries.iter().cloned())
+            .map_err(|e| SyncShardClientError::Internal(anyhow::anyhow!(e)))?;
+        let mut opts = IngestExternalFileOptions::default();
+        opts.set_move_files(true);
+        dbmap
+            .ingest_external_file(&[sst_path], &opts)
+            .map_err(SyncShardClientError::from)
+    }
```
halfprice: question: currently we are fetching 1000 slivers at a time. Does this mean that we are creating a 1000 row SST file and inject them into the main DB for each fetch sliver request?

crates/walrus-service/src/node/config.rs
```diff
@@ -689,6 +693,8 @@ impl Default for ShardSyncConfig {
             shard_sync_concurrency: 10,
             shard_sync_retry_switch_to_recovery_interval: Duration::from_secs(12 * 60 * 60), // 12hr
             restart_shard_sync_always_retry_transfer_first: true,
+            use_sst_ingest_for_shard_sync: false,
```
halfprice: Suggestion: can you verify the correctness of this PR by setting this to true and run CI once? If it passes, we can then turn it off before merging.

crates/typed-store/src/rocks.rs
```diff
@@ -304,6 +306,18 @@ impl RocksDB {
         self.underlying.compact_range_cf_opt(cf, start, end, opt)
     }
 
+    /// Ingest external SST files into a specific column family.
+    /// Note: Files must reside on the same filesystem for move/link optimizations.
```
liquid-helium, line 532: Can you add some guideline here about the implication on data consistency, e.g., should we always create a new-empty cf for ingestion? If this is ingested into an existing cf, what happens for key conflicts?

crates/typed-store/src/rocks.rs
```diff
@@ -2022,6 +2047,52 @@ pub fn safe_drop_db(path: PathBuf) -> Result<(), rocksdb::Error> {
     rocksdb::DB::destroy(&rocksdb::Options::default(), path)
 }
 
+/// Build an SST file from raw key value bytes.
+/// NOTE: Keys need to be pre-sorted and no duplicates are allowed or else ingester will
+/// reject the file.
+pub fn build_sst_from_kv<P, K, V, I>(sst_path: P, entries: I) -> Result<(), TypedStoreError>
+where
```
liquid-helium, line 2604: I remember that when we ingest a external sst file, we have to make sure the key/values are sorted, can you help confirm/disapprove this, if this is true, we should leave a comments to make sure keys are ordered.

crates/typed-store/src/rocks.rs
```diff
@@ -2098,3 +2169,69 @@ fn test_helpers() {
...
+        writer.open(&sst_path).unwrap();
+        for (k, v) in &entries {
+            let kb = be_fix_int_ser(k).unwrap();
+            let vb = bcs::to_bytes(v).unwrap();
+            writer.put(&kb, &vb).unwrap();
+        }
+        writer.finish().unwrap();
+
+        let mut ingest_opts = IngestExternalFileOptions::default();
+        ingest_opts.set_move_files(true);
```
liquid-helium, line 2926: can we make sure we always set `verify_checksums_before_ingest` true, not sure if the same option in Rust.

sadhansood:
> > > @sadhansood I thought this does not work with the DB in blob mode?
> > 
> > 
> > @jpcsmith - absolutely, this is only going to create blob files upon compaction. One thing we could do is trigger compaction after we done ingesting.
> 
> @sadhansood I don't understand. When we get an SST file from the remote storage node, it does not have the blob data, just pointers to the blob file, so how can that be ingested by the local storage node?
> 
> Remark: If this is addressed in the code, my apologies, feel free to just point that out.

@jpcsmith we are still downloading slivers from the remote storage nodes but instead of writing them to the db directly using `put`, we are writing those slivers into an sst file on the local file system. We then ingest this sst file into the db. And in the end we can optionally trigger a compaction to force move the large values from sst into the blob file (if we don't trigger manual compaction, the values will still be moved but during auto compaction slowly)

crates/walrus-service/src/node/storage/shard.rs
sadhansood: Yeah this would have been the case, now we don't do that anymore but flush once the file has enough entries across retry boundaries.

crates/walrus-service/src/node/config.rs
sadhansood: yes, done

crates/typed-store/src/rocks.rs
sadhansood: Added comments.

crates/typed-store/src/rocks.rs
sadhansood: Yes that is correct, we let user tell us whether keys are already sorted or not, and sort ourselves if not.

crates/typed-store/src/rocks.rs
sadhansood: I don't see that option in the rust api unfortunately.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -864,12 +895,21 @@ impl ShardStorage {
 
                 // Record sync progress.
                 last_synced_blob_id = fetched_slivers.last().map(|(id, _)| *id);
-                if let Some(last_synced_blob_id) = last_synced_blob_id {
-                    batch.insert_batch(
-                        &self.shard_sync_progress,
-                        [((), ShardSyncProgress::new(last_synced_blob_id, sliver_type))],
+                let use_sst = config.use_sst_ingest_for_shard_sync;
+                if use_sst {
+                    let sst_file_threshold = config.sst_max_entries.unwrap_or(100_000);
```
halfprice: Remark: in the current setup, all tests are injecting to 1 sst files. Can you add test cases in scenario where sst_max_entries == 3 (or some other small value)? I think you can potentially use tests that are setup using `setup_cluster_for_shard_sync_tests`. Maybe just add another parameter in the test to use sst injection and small sst_max_entries value.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -902,6 +942,72 @@ impl ShardStorage {
...
+        };
+        if flushed && let Some(id) = last_pushed {
+            self.record_last_synced_blob_id(batch, sliver_type, id)?;
+        }
+        Ok(())
+    }
+
+    fn flush_sst<V>(
+        &self,
+        buf_opt: Option<&std::sync::Mutex<typed_store::rocks::SstIngestBuffer<BlobId, V>>>,
```
halfprice: Remark: I think this buf_opt cannot be None if this is called, is that correct? Can we check it's non-null outside and pass in the inner type here?

crates/walrus-service/src/node/config.rs
```diff
@@ -675,6 +675,12 @@ pub struct ShardSyncConfig {
     /// transfer. This is the preferred option since it's always cheaper to scan the blob info
     /// table without transferring the shards.
     pub restart_shard_sync_always_retry_transfer_first: bool,
+    /// Use SST ingestion for sliver writes during shard sync.
+    pub use_sst_ingest_for_shard_sync: bool,
+    /// SST flush thresholds for shard sync.
+    pub sst_max_entries: Option<usize>,
+    /// Compact SST after shard sync.
+    pub compact_sst_after_shard_sync: bool,
```
halfprice: Nit: can we put these options in an optional sst_ingestion_config struct? So that they are either all set, or None (turn this feature off).

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -929,20 +1037,57 @@ impl ShardStorage {
...
-                    assert_eq!(sliver_type, SliverType::Primary);
-                    batch.insert_batch(
-                        &self.primary_slivers,
-                        [(blob_id, &PrimarySliverData::from(primary.clone()))],
-                    )?;
+            if use_sst {
+                let max_entries = config.sst_max_entries.unwrap_or(100_000);
+                let options = SstIngestOptions {
+                    max_entries,
+                    dir_name: Some(self.cf_names.slivers(sliver_type).to_string()),
```
halfprice: Question: currently, this is set to the cf_name instead of `sst_ingest/cf_name`. Should this just set to None with a comment that it is `db_dir/sst_ingest/cf_name`?

crates/typed-store/src/rocks.rs
```diff
@@ -2022,6 +2058,184 @@ pub fn safe_drop_db(path: PathBuf) -> Result<(), rocksdb::Error> {
...
+        if self.options.assume_sorted {
+            build_sst_from_typed(&sst_path, entries)?;
+        } else {
+            let mut kv_bytes: Vec<(Vec<u8>, Vec<u8>)> = Vec::with_capacity(entries.len());
+            for (k, v) in entries.into_iter() {
+                let key = be_fix_int_ser(&k)?;
+                let val = bcs::to_bytes(&v).map_err(typed_store_err_from_bcs_err)?;
+                kv_bytes.push((key, val));
+            }
+            kv_bytes.sort_by(|a, b| a.0.cmp(&b.0));
```
halfprice, line 2763: Question: this only sorts the current buffer, is that correct? What happens if this keys overlap with existing keys in SSTs?

crates/typed-store/src/rocks.rs
```diff
@@ -2022,6 +2058,184 @@ pub fn safe_drop_db(path: PathBuf) -> Result<(), rocksdb::Error> {
...
+    /// Get the number of entries in the buffer.
+    pub fn size(&self) -> usize {
+        self.buffer.len()
+    }
+
+    /// Clear any buffered entries without flushing them.
+    /// Intended for abort/restart scenarios where the producer will re-fetch entries.
+    pub fn clear(&mut self) {
+        self.buffer.clear();
+    }
```
halfprice: Remark: should we also clear the sst dir?

crates/typed-store/src/rocks.rs
```diff
@@ -2022,6 +2058,184 @@ pub fn safe_drop_db(path: PathBuf) -> Result<(), rocksdb::Error> {
...
+
+    /// Force a flush of the current buffer. No-op if buffer is empty.
+    pub fn flush(&mut self) -> Result<(), TypedStoreError> {
+        if self.buffer.is_empty() {
+            return Ok(());
+        }
+
+        let entries: Vec<(K, V)> = std::mem::take(&mut self.buffer);
+
+        let base_dir = if let Some(dir) = &self.options.dir_name {
```
halfprice, line 2743: Question: what happens if the dir already exist?

crates/typed-store/src/rocks.rs
```diff
@@ -2098,3 +2312,173 @@ fn test_helpers() {
...
+        dbmap
+            .compact_range_to_bottom(&0u32, &10u32)
+            .expect("compact should succeed");
+
+        assert_eq!(dbmap.get(&1).unwrap(), Some(10));
+        assert_eq!(dbmap.get(&2).unwrap(), Some(20));
+        assert_eq!(dbmap.get(&3).unwrap(), Some(30));
+        assert_eq!(dbmap.get(&4).unwrap(), Some(40));
+        assert_eq!(dbmap.get(&5).unwrap(), Some(50));
+    }
```
halfprice, line 3046: Maybe add a test case for two flushes with overlaping key ranges? Have asked in other comment that not sure if this is allowed.

crates/typed-store/src/rocks.rs
```diff
@@ -2022,6 +2058,184 @@ pub fn safe_drop_db(path: PathBuf) -> Result<(), rocksdb::Error> {
...
+        kv_bytes.push((key, val));
+    }
+    build_sst_from_kv(sst_path, kv_bytes)
+}
+
+/// Options to control buffered SST ingestion.
+#[derive(Debug, Clone)]
+pub struct SstIngestOptions {
+    /// Flush when the number of buffered entries reaches this threshold.
+    pub max_entries: usize,
```
halfprice, line 2650: suggestion: probabilistically set this to a low value in simtest (maybe [here](https://github.com/MystenLabs/walrus/blob/8a3880f9a34b653c7b9932b4c79f78b57db5555a/crates/walrus-service/src/test_utils.rs#L1117) so that it probabilistically use a low value in all tests).

crates/walrus-service/src/node/storage/shard.rs
sadhansood: @halfprice in `sync_shard_repeated_failures` test - we set `sst_max_entries` to 2 which should cover this usecase.

crates/typed-store/src/rocks.rs
sadhansood: The keys are overwritten with a higher sequence number in the db, so newly written key versions override the older ones.

crates/typed-store/src/rocks.rs
sadhansood: I've set it to 1 in all tests.

crates/typed-store/src/rocks.rs
sadhansood: We now delete the dir upon creation of the buffer.

crates/typed-store/src/rocks.rs
halfprice: What I wanted to ask is, here we only sort the input in each flush.

Say in the first flush, we have key {1, 3, 5}, and second flush, we have key {2, 4, 6}. Across flushes, they are not sorted. Is this ok?

sadhansood: yeah, each flush creates a new sst file which is inserted in the bottom most level where it doesnt overlap with any files in that layer and layers above it. So in your example if first file  {1, 3, 5} goes into L5 (for example), then {2,4,6} cannot go into L5 but has to go to L4 or above (where it doesn't overlap with existing files).

crates/typed-store/src/rocks.rs
sadhansood: done below

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -252,10 +252,6 @@ mod tests {
     #[ignore = "ignore integration simtests by default"]
     #[walrus_simtest]
     async fn test_new_node_joining_cluster() {
-        register_fail_point("fail_point_direct_shard_sync_recovery", move || {
```
sadhansood, line 255: Shard sync assumes the source node will authorize the destination (node which wants to download) because the destination is (or will be) in the relevant committee. In practice, that assumption is brittle around epoch boundaries, and shard sync can’t make progress due to committee view skew between the two nodes. Assuming current epoch in which new node (dst) joins the cluster is 5 and trying to do shard sync:
1. Source is behind: The source (src) is still on the previous epoch 4 and hasn’t populated `next_committee` yet, so it does not recognize the destination (dst) as a committee member and rejects sync with 403.
2. Rapid epoch change: Destination node advances two epochs in quick succession. So destination reaches epoch 7.
3. Destination lost the committee: Dst doesn’t find the committee for epoch 5 (as previous committee is for epoch 6) anymore in its local committee tracker and fails sync with `NoSyncClient` during the retry attempt. And shard sync falls back to recovery.
So, timing around committee updates creates a temporary authorization gap where src doesn’t recognize dst, causing sync to stall and flip to recovery.

crates/walrus-service/src/test_utils.rs
```diff
@@ -1126,6 +1128,14 @@ impl StorageNodeHandleBuilder {
             "At least one Sui RPC URLs must be provided"
         );
 
+        let mut shard_sync_config = ShardSyncConfig::default();
+        shard_sync_config.sst_ingestion_config = Some(SstIngestionConfig::default());
+        shard_sync_config
+            .sst_ingestion_config
+            .as_mut()
+            .unwrap()
+            .max_entries = Some(1);
```
halfprice: suggestion: can we set sst_ingestion_config 50% of the time, and also add a log saying whether the test uses ingestion or not? before sst_ingestion is turned on by default, there is still value to have test coverage of the old behavior.

crates/walrus-service/src/node/storage/shard.rs
sadhansood: yes, done

crates/walrus-service/src/node/storage/shard.rs
sadhansood: done

crates/walrus-simtest/tests/simtest.rs
halfprice: Thanks a lot for the debugging @sadhansood ! Kudos for looking deeply into the cause!

I think the issue is not this check (this check is quite important actually), but in simtest, you use ShardSyncConfig::default() to create ShardSyncConfig, which has 60s min retry delay.

The reason it was ok before was that in test cluster, by default we use 1s min and 3s max retry delay: https://github.com/MystenLabs/walrus/blob/b10e62b87825b5f155d1a99132c15c84f152105b/crates/walrus-service/src/test_utils.rs#L2983

Just to be safe, I'd say you can update the new sst_ingestion_config after StorageNodeConfig is firstly created (so that we don't override any test default), and later specifically set sst_ingestion_config in the constructed StorageNodeConfig (probabilistically, as suggested in the other comment).

halfprice (approved):
Thanks a lot @sadhansood for the revision! It looks good to me with 2 minor comments regarding tests.

PR #2500, opened 2025-09-03
chore: create PendingEventCounter to track pending events in BlobEventProcessor

halfprice:
## Description

Follow up in #2476 

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -43,12 +43,22 @@ impl PendingEventCounter {
 
     /// Increments the counter and returns the new value.
     fn inc(&self) -> u32 {
+        // We must ensure the increment happens-before the event is observable by the worker via
+        // the channel. Without a strong ordering, the send could be observed before the increment,
+        // allowing the receiver to drop the guard and do fetch_sub first, which would underflow
+        // the counter.
         self.inner.fetch_add(1, Ordering::SeqCst) + 1
     }
 
     /// Decrements the counter and returns the new value.
     fn dec(&self) -> u32 {
+        // Same reasoning as inc() applies hereto use SeqCst ordering.
         let prev = self.inner.fetch_sub(1, Ordering::SeqCst);
+        if prev == 0 {
+            // This should never happen, since the counter is incremented before the event is
+            // observable by the worker via the channel.
+            panic!("pending event counter underflow");
+        }
         prev.saturating_sub(1)
     }
 


crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -26,18 +26,50 @@ use crate::node::{
 // Poll interval for checking pending background events.
 const PENDING_EVENTS_POLL_INTERVAL: Duration = Duration::from_millis(10);
 
+/// A utility struct that wraps an Arc<AtomicU32> for tracking pending events.
+/// Provides convenient inc() and dec() methods that return the new value.
+#[derive(Debug, Clone)]
+struct PendingEventCounter {
+    inner: Arc<AtomicU32>,
```
sadhansood, line 33: It'd be even cool if we can make this `Vec<Arc<AtomicU32>>` and have apis like `inc(index: u32)` and `dec(index: u32)` or at least that was the original suggestion.

halfprice: Interesting. I read the comments completely differently.

I feel its better to make the vector explicit in BlobEventProcessor because both `_background_processors`, and `background_processor_senders` are all vectors and have 1 to 1 mapping with the counter in `background_per_processor_pending_event_count`. Currently it is easy to visualize and read this relationship, whereas if we hide the index inside the new struct, it takes extra hop for readers to build this relationship.

The current layout helps in terms of abstract atomic op complexity.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -26,18 +26,50 @@ use crate::node::{
...
+    /// Creates a new counter initialized to 0.
+    fn new() -> Self {
+        Self {
+            inner: Arc::new(AtomicU32::new(0)),
+        }
+    }
+
+    /// Increments the counter and returns the new value.
+    fn inc(&self) -> u32 {
+        self.inner.fetch_add(1, Ordering::SeqCst) + 1
```
sadhansood, line 50: Maybe worth mentioning in the comments why we need `Ordering::SeqCst` or why a cheaper ordering is not enough.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -26,18 +26,50 @@ use crate::node::{
...
+    }
+
+    /// Decrements the counter and returns the new value.
+    fn dec(&self) -> u32 {
+        let prev = self.inner.fetch_sub(1, Ordering::SeqCst);
+        prev.saturating_sub(1)
```
sadhansood, line 62: If a user decrements first after creating the counter initialized with 0, `prev` will underflow to u32::Max. We should probably add a check and panic in such scenario.

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: Ah ok, sounds good.

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Done. PTAL

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Good point! Done. This is also why we need SeqCst ordering

PR #2506, opened 2025-09-08
feat: add k6 load test scripts

jpcsmith:
## Description

Adds load testing scripts that use k6 to the repository.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/scripts/k6/package.json
+++ b/scripts/k6/package.json
@@ -4,7 +4,6 @@
   "engines": {
     "node": "16 || 18"
   },
-  "license": "MIT",
   "devDependencies": {
     "@types/k6": "~0.48.0",
     "typescript": "5.3.3"


tests/k6/src/config/environment.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export const PUBLISHER_URL: string = __ENV.PUBLISHER_URL
+    || ENVIRONMENT_DEFAULTS[ENVIRONMENT].publisherUrl;
+
+/**
+ * The path to a file that is used as the source of all the blobs.
+ * Each blob is taken as a slice from the raw bytes of the file.
+ */
+export const PAYLOAD_SOURCE_FILE: string = __ENV.PAYLOAD_SOURCE_FILE
+    || ENVIRONMENT_DEFAULTS[ENVIRONMENT].payloadSourceFile;
```
halfprice: Suggestion: can we set the default environment to testnet so that it requires less parameter input when trying out the script? Similar for PAYLOAD_SOURCE_FILE that we don't need to set it when just trying out the tests.

halfprice (approved):
Thanks @jpcsmith for the testing scripts! I tried a few and it produced interesting information.

Can we move the testing script under `scripts/performance_tests/` instead of `tests`?

PR #2507, opened 2025-09-08
chore: reduce test failures of upload relay due to port conflicts

wbbradley:
## Description

There were some instances where ports were not being released in time for retries. This change unifies the unused_socket_address code paths and leverages them in the upload relay tests.

## Test plan

CI Pipeline.

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -2458,8 +2458,10 @@ async fn test_store_with_upload_relay_no_tip() {
         ],
         ..cluster_config.clone()
     };
-    let server_address = unused_socket_address(UnusedSocketAddressIp::AllInterfaces, false)
-        .expect("get unused socket address");
+    let server_address = unused_socket_address(UnusedSocketAddressIp::AllInterfaces {
+        force_time_wait: false,
+    })
+    .expect("get unused socket address");
 
     let registry = Registry::default();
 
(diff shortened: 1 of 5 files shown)


crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -2429,9 +2428,10 @@ async fn test_store_with_upload_relay_no_tip() {
         ],
         ..cluster_config.clone()
     };
-    let server_address: SocketAddr = DEFAULT_SERVER_ADDRESS
-        .parse()
-        .expect("valid server address");
+    let server_address = SocketAddr::new(
```
halfprice: Nit: there is an existing `unused_socket_address` util. Can we use that here?

wbbradley: That function seems like it does the opposite of what we want. 

The problem is that after dropping the listener and connections, the OS keeps the port in TIME_WAIT state (typically for 2-4 minutes) to handle delayed packets. Another process trying to bind to that same port will fail with "Address already in use" unless it uses `SO_REUSEADDR`.

I've removed the random approach, and taken a much simpler approach, which is to just bind to "0.0.0.0:0" and then query for the port, then drop the binding. This ensures that the port is available (besides a race condition with other processes, which is highly unlikely.)

halfprice: Ah, make sense. The problem is across tests, not within the test.

wbbradley: Ok, I've unified the `unused_socket_address` function so that it can be called with different purposes in mind.

PR #2511, opened 2025-09-09
chore: fix pr creation

ebmifa:
## Description
 fix pr creation

## Test plan
👀

### DIFF
--- a/scripts/bump_sui_testnet_version.sh
+++ b/scripts/bump_sui_testnet_version.sh
@@ -96,24 +96,34 @@ git config user.email \
   "41898282+github-actions[bot]@users.noreply.github.com"
 
 # Push branch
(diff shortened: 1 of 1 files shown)


scripts/bump_sui_testnet_version.sh
```diff
@@ -96,24 +96,34 @@ git config user.email \
   "41898282+github-actions[bot]@users.noreply.github.com"
 
 # Push branch
-git commit -m "chore: bump Sui version to ${NEW_TAG}"
+git commit -m "ci: bump Sui testnet version to ${NEW_TAG}"
```
wbbradley, line 99: Why change this? I think "semantic commit messages" specify "ci" as "Changes to CI configuration files or scripts". this doesn't seem like that, this is more of a "chore".

PR #2521, opened 2025-09-12
chore: clean up "Getting started" link in book

wbbradley:
## Description

Rename the Setup page to the Getting started page.

## Test plan

Look at book preview.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ and more extensive tests:
 ## Using the Walrus client
 
 If you just want to use the Walrus Mainnet or Testnet, please follow the [setup instructions in our
-documentation](https://docs.wal.app/usage/setup.html). Further information about the usage is
+documentation](https://docs.wal.app/usage/started.html). Further information about the usage is
 also available in the [public documentation](https://docs.wal.app/usage/interacting.html).
 
 If you want to build the binary from source or run local tests, see the instructions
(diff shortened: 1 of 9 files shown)


docs/book/design/overview.md
```diff
@@ -5,3 +5,5 @@ mechanisms](./encoding.md) of the Walrus system.
 
 Use the [glossary](../glossary.md) as a reference for many of the bolded terms used in this
 documentation.
+
+Learn more about the [science](https://snowreads.wal.app) behind Walrus.
```
mlegner, line 9: I'm not sure I understand this reference. AFAIU, the papers on Snowreads aren't specifically related to Walrus.

docs/book/usage/started.md
```diff
@@ -1,8 +1,247 @@
-# Getting started
+# Getting Started
```
mlegner, line 1: Nit: We use sentence casing for all titles in this documentation (except in the blog part AFAIK).

mlegner (commented):
Thanks for the PR, @wbbradley. No general objections against this change.

However, can we make sure that clients using old links (that may still be around externally) are properly [redirected](https://rust-lang.github.io/mdBook/format/configuration/renderers.html#outputhtmlredirect)? For example, https://setup.wal.app redirects to the setup chapter, which would no longer work without the extra redirects.

PR #2529, opened 2025-09-17
fix: fix walrus-backup garbage collection statistics

wbbradley:
## Description

Fix a regression in garbage collection statistics in walrus-backup. The regression was introduced recently in 7321bdf9431841f75afc97e6927be35c548cbbb3 wherein I delayed the garbage collection mechanism until 2 epochs later than expiry, but failed to update the related stats collection.

## Test plan

CI Pipeline.

### DIFF
--- a/crates/walrus-service/src/backup/garbage_collector.rs
+++ b/crates/walrus-service/src/backup/garbage_collector.rs
@@ -65,6 +65,30 @@ async fn collect_garbage(
         .await
         .expect("failed to connect to postgres for collect_garbage");
 
+    let garbage_query = format!(
+        "
+            WITH expired_blob_ids AS (
+                SELECT blob_id
+                FROM blob_state
+                WHERE
+                    state = 'archived' AND
+                    (initiate_gc_after IS NULL OR initiate_gc_after < NOW()) AND
+                    COALESCE(
+                        end_epoch + {} <= (
+                            SELECT MAX(epoch) FROM epoch_change_start_event),
+                        FALSE)
+                LIMIT 15
+            ),
+            _updated_count AS (
+                UPDATE blob_state
+                SET initiate_gc_after = NOW() + INTERVAL '1 day'
+                WHERE blob_id IN (SELECT blob_id FROM expired_blob_ids)
+            )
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/backup/garbage_collector.rs
```diff
@@ -65,6 +65,31 @@ async fn collect_garbage(
...
+    let garbage_query = format!(
+        "
+            WITH expired_blob_ids AS (
+                SELECT blob_id
+                FROM blob_state
+                WHERE
+                    state = 'archived' AND
+                    (initiate_gc_after IS NULL OR initiate_gc_after < NOW()) AND
+                    COALESCE(
+                        end_epoch + {} <= (
```
halfprice, line 77: Is the `+ {}` the only part changed in the query, is that right?

crates/walrus-service/src/backup/service.rs
```diff
@@ -364,63 +364,62 @@ fn start_db_metrics_loop(metrics_runtime: &MetricsAndLoggingRuntime, config: &Ba
...
+            GROUP BY 1
+            UNION
+            SELECT
+                state AS name,
+                COUNT(*)::bigint AS value
+            FROM blob_state
+            WHERE
+                state = 'deleted' OR
+                (state = 'waiting' AND
+                    COALESCE(end_epoch > (SELECT MAX(epoch) FROM epoch_change_start_event), TRUE))
```
halfprice, line 386: Is this for blobs explicitly deleted? Do we have a grace period for those blobs explicitly deleted?

halfprice (approved):
Have 2 clarification questions below. The PR LGTM 💯

crates/walrus-service/src/backup/garbage_collector.rs
wbbradley: Yes, that's correct.

crates/walrus-service/src/backup/service.rs
wbbradley: We do not handle explicitly deleted blobs in the backup system. It only tracks MAX(end_epoch) across the various incoming objects that point to the same blob id.

PR #2530, opened 2025-09-17
fix: race condition in shard sync and shard movement

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/shard_sync.rs
+++ b/crates/walrus-service/src/node/shard_sync.rs
@@ -135,7 +135,7 @@ impl ShardSyncHandler {
         let blob_infos = self
             .node
             .storage
-            .certified_blob_info_iter_before_epoch(self.node.current_epoch());
+            .certified_blob_info_iter_before_epoch(self.node.current_committee_epoch());
 
         #[cfg(msim)]
         {
@@ -250,10 +250,10 @@ impl ShardSyncHandler {
                 tracing::error!(
                     "{shard_index} is not assigned to this node; cannot start shard sync"
                 );
-                ShardNotAssigned(shard_index, self.node.current_epoch())
+                ShardNotAssigned(shard_index, self.node.current_committee_epoch())
             })?;
 
-        let shard_status = shard_storage.status()?;
+        let shard_status = shard_storage.status().await?;
 
         // Skip if shard is already active
         if shard_status == ShardStatus::Active {
@@ -266,7 +266,7 @@ impl ShardSyncHandler {
 
         // Update status and start sync. After this function returns, we can always restart
         // the sync upon node restart.
-        shard_storage.record_start_shard_sync()?;
+        shard_storage.record_start_shard_sync().await?;
         self.start_shard_sync_impl(shard_storage.clone()).await;
         Ok(())
     }
@@ -298,7 +298,7 @@ impl ShardSyncHandler {
             for shard_storage in self.node.storage.existing_shard_storages().await {
                 // Restart the syncing task for shards that were previously syncing (in ActiveSync
                 // status).
-                let shard_status = shard_storage.status()?;
+                let shard_status = shard_storage.status().await?;
                 match shard_status {
                     ShardStatus::ActiveSync => {
                         self.start_shard_sync_impl(shard_storage.clone()).await;
@@ -306,7 +306,7 @@ impl ShardSyncHandler {
                     ShardStatus::ActiveRecover => {
                         if self.config.restart_shard_sync_always_retry_transfer_first {
                             let shard_last_sync_status =
-                                shard_storage.resume_active_shard_sync()?;
+                                shard_storage.resume_active_shard_sync().await?;
                             tracing::info!(
                                 walrus.shard_index = %shard_storage.id(),
                                 ?shard_last_sync_status,
@@ -325,7 +325,7 @@ impl ShardSyncHandler {
     async fn start_shard_sync_impl(&self, shard_storage: Arc<ShardStorage>) {
         // This epoch must be the same as the epoch in the committee we refreshed when processing
         // epoch start event, or when the node starts up.
-        let current_epoch = self.node.current_epoch();
+        let current_epoch = self.node.current_committee_epoch();
 
         tracing::info!(
             walrus.shard_index = %shard_storage.id(),
@@ -697,6 +697,7 @@ mod tests {
                 .await
                 .expect("Failed to get shard storage")
                 .update_status_in_test(ShardStatus::ActiveSync)
+                .await
                 .expect("Failed to update shard status");
         }
         let shard_sync_handler = ShardSyncHandler::new(
@@ -740,6 +741,7 @@ mod tests {
             .await
             .expect("Failed to get shard storage")
             .update_status_in_test(ShardStatus::None)
+            .await
             .expect("Failed to update shard status");
 
         assert_eq!(shard_sync_handler.current_sync_task_count().await, 0);
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -461,10 +461,11 @@ impl Storage {
     pub async fn list_shard_status(&self) -> HashMap<ShardIndex, Option<ShardStatus>> {
         let shards = self.shards.read().await;
 
-        shards
-            .iter()
-            .map(|(shard, storage)| (*shard, storage.status().ok()))
-            .collect()
+        let mut result = HashMap::new();
+        for (shard, storage) in shards.iter() {
+            result.insert(*shard, storage.status().await.ok());
+        }
+        result
     }
 
     /// Store the verified metadata without updating blob info. This is only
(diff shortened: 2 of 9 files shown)


crates/walrus-service/src/node/storage.rs
```diff
@@ -461,10 +461,11 @@ impl Storage {
...
 
-        shards
-            .iter()
-            .map(|(shard, storage)| (*shard, storage.status().ok()))
-            .collect()
+        let mut result = HashMap::new();
+        for (shard, storage) in shards.iter() {
+            result.insert(*shard, storage.status().await.ok());
+        }
+        result
```
mlegner: Remark: I felt this should also work with an iterator-like approach, but because we want a hash map, it turned out to be a bit tedious 🤷:
```rust
        futures::future::join_all(
            shards
                .iter()
                .map(|(shard, storage)| async move { (*shard, storage.status().await.ok()) }),
        )
        .await
        .into_iter()
        .collect()
```

crates/walrus-service/src/node/storage.rs
```diff
@@ -820,17 +821,18 @@ impl Storage {
...
+
+        for shard_storage in shards.values() {
+            if let Ok(status) = shard_storage.status().await
+                && status.is_owned_by_node()
+            {
+                result.push(shard_storage.id());
+            }
+        }
+
+        result
```
mlegner: Just in case you want to keep the iterator-like syntax:
```rust
        futures::future::join_all(shards.values().map(|shard_storage| async move {
            shard_storage
                .status()
                .await
                .is_ok_and(|status| status.is_owned_by_node())
                .then_some(shard_storage.id())
        }))
        .await
        .into_iter()
        .flatten()
        .collect()
```

crates/walrus-service/src/node/shard_sync.rs
```diff
@@ -72,16 +72,19 @@ impl ShardSyncHandler {
...
             sync_handler
                 .sync_shards_task(shards, recover_metadata)
                 .await
         });
 
-        // Abort any existing task before replacing it
-        if let Some(old_task) = task_handle.replace(new_task) {
-            old_task.abort();
-        }
+        *task_handle = Some(new_task);
```
mlegner, line 87: Question: Is this change actually necessary or does it just make reasoning about it easier?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -137,6 +138,31 @@ impl ShardStatus {
...
+        self.shard_status.insert(&(), &new_status)?;
+        Ok(())
+    }
+
+    pub fn get_status(&self) -> Result<Option<ShardStatus>, TypedStoreError> {
+        self.shard_status.get(&())
+    }
+
+    /// Executes a function with mutable access to the shard_status DBMap.
+    /// This allows atomic operations on the shard status within a critical section.
```
mlegner, line 184: Question: This assumes that the column family wasn't opened multiple times, correct?

mlegner: Also, question: When we enable transactions on the DB, we could use those instead, right?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -740,10 +824,23 @@ impl ShardStorage {
...
+                let mut batch = self.shard_sync_progress.batch();
+                let current_status = shard_status.get(&())?;
+                // If the shard is in active sync or active recover, we need to update the status to
+                // active. Otherwise, the shard status is interrupted by another process and are in
+                // unexpected state and we should not update the status.
+                if current_status == Some(ShardStatus::ActiveSync)
+                    || current_status == Some(ShardStatus::ActiveRecover)
+                {
+                    batch.insert_batch(shard_status, [((), ShardStatus::Active)])?;
+                }
```
mlegner, line 867: Question: Just to make sure I fully understand the fix: Is this the core of the solution as mentioned in the PR description?

> The fix for this issue is to not allow shard status to be transition to other status if it is in LockedToMove, unless, the node starts a new shard sync that explicitly requests a new shard sync task. This prevents the shard mistakenly move from LockedToMove to Active status.

mlegner (commented):
Thanks so much for fixing this complicated issue, @halfprice. This LGTM, I just have a few questions to make sure I fully understand the solution.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -1020,7 +1117,15 @@ impl ShardStorage {
         sui_macros::fail_point!("fail_point_shard_sync_recovery");
 
         tracing::info!("shard sync is done; still has missing blobs; shard enters recovery mode");
-        self.shard_status.insert(&(), &ShardStatus::ActiveRecover)?;
+        self.shard_status.write().await.with_shard_status_mut(
+            |shard_status| -> Result<(), TypedStoreError> {
+                let existing_status = shard_status.get(&())?;
+                if existing_status == Some(ShardStatus::ActiveSync) {
+                    shard_status.insert(&(), &ShardStatus::ActiveRecover)?;
+                }
```
sadhansood, line 1154: Should we have thrown an error and aborted recovery here if shard is not in ActiveSync and trying to move to ActiveRecover? It seems like the best effort status update could lead to inconsistent status and shard state.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -537,34 +570,89 @@ impl ShardStorage {
...
     }
 
-    pub(crate) fn set_active_status(&self) -> Result<(), TypedStoreError> {
-        self.shard_status.insert(&(), &ShardStatus::Active)
+    pub(crate) async fn set_active_status(&self) -> Result<(), TypedStoreError> {
+        self.shard_status
+            .write()
+            .await
+            .update_status(ShardStatus::Active)?;
+        Ok(())
```
sadhansood, line 633: Should we make this a conditional update based on whether the current state is ActiveSync or ActiveRecover and return an error otherwise?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -682,12 +770,8 @@ impl ShardStorage {
                 sui_macros::fail_point!("fail_point_direct_shard_sync_recovery");
             }
         }
+        let shard_status = self.status().await?;
```
sadhansood, line 802: What is the purpose of reading shard_status here?

sadhansood (commented):
Solid change @halfprice , just a few comments. One other suggestion would be to maybe consider a small centralized transition function or table to avoid transition rules in scattered locations and improve code understanding.

crates/walrus-service/src/node/shard_sync.rs
halfprice: Exactly that this makes the reasoning a bit easier. But currently it does not causing any issue.

If both the old task and new task are running, there may be multiple shard syncs on the same shard running, which causes confusing during debugging, even though one of them will be aborted shortly.

crates/walrus-service/src/node/storage/shard.rs
halfprice: If you are talking about opening same column family multiple times from RocksDb, I don't think RocksDb currently allows that. However, if you are talking about opening the column family once, but have a reference to the DBMap in multiple places, and not all through the new ShardStatusState, yes, here the code assumes that.

> When we enable transactions on the DB, we could use those instead, right?

Yes! The need here is to do a read-modify-write transaction. If not implemented in the application logic, another way to do this is to use merge operator, which will further complicate the code. Since the logic of the shard status is simple, I just guarded the access to it using a mutex to achieve transaction.

But I agree that this can be replaced by transaction if using transaction DB.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Yes, this is part of the core to address the test failure that is manifested. The real fix is all the places where with_shard_status_mut() is used: shard status transition needs to be transactional and we cannot blindly apply new status. E.g., previously we transition shards from ActiveSync to ActiveRecover, but when doing this transition, the shard may not in ActiveSync anymore.

halfprice: Thinking more about the core fix, it is to prevent shard in LockedToMove to transition to other state, unless moved out shard moves back in again.

crates/walrus-service/src/node/storage.rs
halfprice: Applied. Thanks!

crates/walrus-service/src/node/storage.rs
halfprice: Applied. Thanks!

crates/walrus-service/src/node/storage/shard.rs
halfprice: This is related to WAL-903 that what should shard sync do if the shard is already moved out. At that time, we decide to let the shard keep syncing, because ultimately, it is this shard's responsibility to transfer all the blobs to the new node. That's why we just want to keep moving without returning an error. I added a TODO pointing to WAL-903 here as well.

Regarding inconsistent state, when a node is in LockedToMove status, the next epoch, the shard will be removed from the storage nodes. When this shard moves back in again, it'll current shard sync will be abort and new shard sync will be created.

crates/walrus-service/src/node/storage/shard.rs
halfprice: This set_active_status is a bit different. It is only called in one place where the node was in recovery mode and just entered the latest epoch. In this case, the epoch change logic will read all the owned shards, set them to Active state, and start shard recovery task.

So this is not meant to be a general purpose set_active_status function. I changed it to force_set_active_status() and add documentation to make it clear.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Few lines below, we call get_last_sync_status(), which resumes the shard sync from the last shard sync task. There, it needs to know what the shard status is, and will only read the checkpoint of shard sync progress if the node is in ActiveSync status.

crates/walrus-service/src/node/storage/shard.rs
sadhansood: got it, makes sense

halfprice:
Hi @mlegner , will merge the PR for now to unblock other PRs. Please let me know if you have more comments!

PR #2533, opened 2025-09-18
tests: add aggregator latency and breakpoint tests

jpcsmith:
## Description

Adds latency and breakpoint tests for the aggregator, which requests blob IDs cached in a redis cluster.

## Test plan

Ran locally.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


halfprice (approved):
LGTM! Thanks for creating all the tests @jpcsmith !

Before you leave, you need to educate us how to use these scripts, and hopefully by that time, there are workflows that can start these tests easily.

Issue #2534, opened 2025-09-19
Potential Unhandled Empty Quilt Blob


QiLOL:
### What happened?

There is a potential denial-of-service panic when constructing a `QuiltV1` directly from an empty quilt blob. Previously, `QuiltV1::new_from_quilt_blob` accepted an empty `Vec<u8>` and later triggered a division-by-zero panic during index decoding. 

```rust
    fn new_from_quilt_blob(
        quilt_blob: Vec<u8>,
        encoding_config: &EncodingConfigEnum<'_>,
    ) -> Result<QuiltV1, QuiltError> {
        ...
        let symbol_size = quilt_blob.len() / n_source_symbols;  // @note return 0 if quilt_blob.len() is 0
        let row_size = symbol_size * n_secondary_source_symbols; // @note  return 0
        let mut quilt = QuiltV1 {
            data: quilt_blob,
            row_size,
            symbol_size,
            quilt_index: None,
        };
        let _ = quilt.get_or_decode_quilt_index()?;

        Ok(quilt)
    }

    fn get_or_decode_quilt_index(&mut self) -> Result<&QuiltIndexV1, QuiltError> {
        ...

        let columns_size = self.data.len() / self.row_size * self.symbol_size;  // @note panic since row_size is 0 
        let quilt_index = QuiltVersionV1::decode_quilt_index(self, columns_size)?; // @note columns_size should also not be 0 since div_ceil inside
        self.quilt_index = Some(quilt_index);

        Ok(self
            .quilt_index
            .as_ref()
            .expect("quilt index should be set"))
    }
```

I am pretty new to the space. Currently, there may be a defence in `QuiltEnum::new` that rejects empty input via `new_from_sliver`, but a more direct defence would be great. Thanks for checking!

Added a PR: https://github.com/MystenLabs/walrus/pull/2535

### Search terms

- new_from_quilt_blob
- empty quilt blob

### Steps to reproduce

The following test could demonstrate the issue:

```rust
    #[test]
    fn test_new_from_quilt_blob_panics_on_empty_input() {
        use core::num::NonZeroU16;
        let config = ReedSolomonEncodingConfig::new(NonZeroU16::new(7).unwrap());
        let _ = QuiltV1::new_from_quilt_blob(
            Vec::new(),
            &EncodingConfigEnum::ReedSolomon(&config),
        );
    }
```

### Relevant log output

```shell

```

### Version

- [x] My `walrus` binary and configuration are up to date.

### Code of Conduct

[...]

PR #2535, opened 2025-09-19
chore: empty quilt blob protection

QiLOL:
## Description

According to the https://github.com/MystenLabs/walrus/issues/2534

This PR fixes a denial-of-service panic when constructing a QuiltV1 directly from an empty quilt blob. This change adds explicit input validation and defensive guards to return a proper error instead of panicking.

## Test plan

unit test

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/encoding/quilt_encoding.rs
+++ b/crates/walrus-core/src/encoding/quilt_encoding.rs
@@ -1096,7 +1096,7 @@ impl QuiltV1 {
                 .as_ref()
                 .expect("quilt index should be set"));
         }
-        
+
         let columns_size = self.data.len() / self.row_size * self.symbol_size;
         let quilt_index = QuiltVersionV1::decode_quilt_index(self, columns_size)?;
         self.quilt_index = Some(quilt_index);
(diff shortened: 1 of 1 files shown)


crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -1092,6 +1096,10 @@ impl QuiltV1 {
                 .as_ref()
                 .expect("quilt index should be set"));
         }
+
```
liquid-helium: consider moving this to the constructor, however, this should never happen, a blob is either 0 bytes, otherwise, it will be larger than the n_symbol, so the check of ` if quilt_blob.is_empty() ` should be sufficient.

QiLOL: Hi @liquid-helium,  thanks for checking, I agree that the `quilt_blob.is_empty()` is enough here, so I just removed the validation from `get_or_decode_quilt_index`.

Please feel free to let me know if you need any other modifications.

liquid-helium: Thanks @QiLOL for the improvements, LGTM.

QiLOL: Thank you @liquid-helium ! And I just fixed the lint, sorry about the inconvenience.

PR #2536, opened 2025-09-19
feat: Add update mode to walrus client cli

sadhansood:
## Description

This PR introduces an overlay semantics for upload presets so config file values are respected by default. Adding a `--force-upload-mode` CLI flag to override config file communication settings when desired.

## Test plan

Ran locally and noticed big performance improvements.

### DIFF
--- a/crates/walrus-sdk/src/config.rs
+++ b/crates/walrus-sdk/src/config.rs
@@ -32,11 +32,13 @@ mod committees_refresh_config;
 mod communication_config;
 mod reqwest_config;
 mod sliver_write_extra_time;
+mod upload_mode;
 
 pub use self::{
     committees_refresh_config::CommitteesRefreshConfig,
     communication_config::{ClientCommunicationConfig, CommunicationLimits},
     reqwest_config::RequestRateConfig,
+    upload_mode::UploadMode,
 };
 
 /// Returns the default paths for the Walrus configuration file.
--- a/crates/walrus-sdk/src/config/upload_mode.rs
+++ b/crates/walrus-sdk/src/config/upload_mode.rs
@@ -1,0 +1,72 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+//! Preset upload modes that tune client communication for throughput/latency.
+
+use serde::{Deserialize, Serialize};
+
+use crate::config::communication_config::ClientCommunicationConfig;
+
+/// Upload preset modes for tuning client concurrency and network usage.
+///
+/// These presets only adjust in-memory ClientCommunicationConfig and are intended
+/// to be applied per-run by the CLI (or other callers). They do not mutate the
+/// on-disk configuration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "camelCase")]
+pub enum UploadMode {
+    /// Targets ~10–100 Mbps links with low memory footprint and modest per-node pressure.
+    /// Sets bytes-in-flight near 100 Mb (~12.5 MB) and a conservative per-node cap.
+    Conservative,
+    /// Targets ~1 Gbps links with balanced throughput and stability.
+    /// Sets bytes-in-flight near 1 Gb (~125 MB) and moderate per-node cap.
+    #[default]
+    Balanced,
+    /// Targets ~10 Gbps environments where higher memory and per-node parallelism are acceptable.
+    /// Sets bytes-in-flight near 10 Gb (~1.25 GB) and higher per-node cap.
+    Aggressive,
+}
+
+struct CommunicationParameters {
+    max_concurrent_writes: usize,
+    max_data_in_flight: usize,
+}
+
+impl UploadMode {
+    fn to_communication_parameters(self) -> CommunicationParameters {
+        match self {
+            UploadMode::Conservative => CommunicationParameters {
+                max_concurrent_writes: 1000,
+                max_data_in_flight: 12_500_000, // 12.5 MB
+            },
+            UploadMode::Balanced => CommunicationParameters {
+                max_concurrent_writes: 1000,
+                max_data_in_flight: 125_000_000, // 125 MB
+            },
+            UploadMode::Aggressive => CommunicationParameters {
+                max_concurrent_writes: 1280,
+                max_data_in_flight: 1_250_000_000, // 1.25 GB
+            },
+        }
+    }
+
+    /// Applies this preset to a given communication config and returns the updated config.
+    /// If `force` is true, overrides values unconditionally; otherwise only fills defaults.
+    pub fn apply_to(self, mut config: ClientCommunicationConfig) -> ClientCommunicationConfig {
+        let default_bytes = crate::config::communication_config::default::max_data_in_flight();
+        let CommunicationParameters {
+            max_concurrent_writes,
+            max_data_in_flight,
+        } = self.to_communication_parameters();
+
+        if config.max_concurrent_writes.is_none() {
+            config.max_concurrent_writes = Some(max_concurrent_writes);
+        }
+
+        if config.max_data_in_flight == default_bytes {
+            config.max_data_in_flight = max_data_in_flight;
+        }
+
+        config
+    }
+}
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -28,6 +28,7 @@ use walrus_core::{
     encoding::{EncodingConfig, EncodingFactory},
     ensure,
 };
+use walrus_sdk::config::UploadMode;
 use walrus_sui::{
     client::{ExpirySelectionPolicy, ReadClient, SuiContractClient},
     types::{StorageNode, move_structs::Authorized},
@@ -1095,6 +1096,10 @@ pub struct CommonStoreOptions {
     #[arg(long)]
     #[serde(default)]
     pub skip_tip_confirmation: bool,
+    /// Preset upload mode to tune network concurrency and bytes-in-flight.
+    #[arg(long, value_enum)]
+    #[serde(default)]
+    pub upload_mode: Option<UploadModeCli>,
 }
 
 #[serde_as]
@@ -1111,6 +1116,31 @@ pub struct FileOrBlobId {
     #[serde_as(as = "Option<DisplayFromStr>")]
     #[serde(default)]
     pub(crate) blob_id: Option<BlobId>,
+}
+
+/// CLI enum for selecting upload presets. Converted to SDK UploadMode at runtime.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, clap::ValueEnum)]
+#[serde(rename_all = "camelCase")]
+pub enum UploadModeCli {
+    Conservative,
+    Balanced,
+    Aggressive,
+}
+
+impl Default for UploadModeCli {
+    fn default() -> Self {
+        Self::Balanced
+    }
+}
+
+impl From<UploadModeCli> for UploadMode {
+    fn from(value: UploadModeCli) -> Self {
+        match value {
+            UploadModeCli::Conservative => UploadMode::Conservative,
+            UploadModeCli::Balanced => UploadMode::Balanced,
+            UploadModeCli::Aggressive => UploadMode::Aggressive,
+        }
+    }
 }
 
 impl FileOrBlobId {
@@ -1803,6 +1833,7 @@ mod tests {
                 encoding_type: Default::default(),
                 upload_relay: None,
                 skip_tip_confirmation: false,
+                upload_mode: None,
             },
         })
     }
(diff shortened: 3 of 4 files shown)


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1113,6 +1123,31 @@ pub struct FileOrBlobId {
...
+
+impl From<UploadModeCli> for UploadMode {
+    fn from(value: UploadModeCli) -> Self {
+        match value {
+            UploadModeCli::Conservative => UploadMode::Conservative,
+            UploadModeCli::Balanced => UploadMode::Balanced,
+            UploadModeCli::Aggressive => UploadMode::Aggressive,
+        }
+    }
+}
```
mlegner, line 1144: Question: Why do we need a separate struct here? Couldn't we use `UploadMode` directly?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1095,6 +1096,15 @@ pub struct CommonStoreOptions {
     #[arg(long)]
     #[serde(default)]
     pub skip_tip_confirmation: bool,
+    /// Preset upload mode to tune network concurrency and bytes-in-flight.
+    #[arg(long, value_enum, default_value_t = default::upload_mode())]
+    #[serde(default = "default::upload_mode")]
+    pub upload_mode: UploadModeCli,
```
mlegner: Important: AFAIU, this currently overwrites the value from the configuration file in all cases.

Suggestion: Use an `upload_mode: Option<UploadMode>` flag, similar to what we do with similar flags like the [RPC URL](https://github.com/MystenLabs/walrus/blob/fae992e0d52888dfc50cb524dda7b0461109bca9/crates/walrus-service/src/client/cli/args.rs#L1002-L1009).

The semantics would then be to use the explicit CLI flag if it is set, use the value from the config file, or the default, in that order.

crates/walrus-sdk/src/config.rs
```diff
@@ -32,11 +32,13 @@ mod committees_refresh_config;
 mod communication_config;
 mod reqwest_config;
 mod sliver_write_extra_time;
+mod upload_mode;
 
 pub use self::{
     committees_refresh_config::CommitteesRefreshConfig,
     communication_config::{ClientCommunicationConfig, CommunicationLimits},
     reqwest_config::RequestRateConfig,
+    upload_mode::UploadMode,
```
mlegner, line 41: Question/Suggestion: Could we also have a force parameter here?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -676,7 +682,16 @@ impl ClientCommandRunner {
...
-        let client = get_contract_client(self.config?, self.wallet, self.gas_budget, &None).await?;
+        // Apply CLI upload preset to the in-memory config before building the client.
+        // Honor explicit config-file values unless --force-upload-mode is set.
+        let mut config = self.config?;
+        let preset: UploadMode = upload_mode;
+        if force_upload_mode {
+            config.communication_config = preset.apply_force(config.communication_config.clone());
+        } else {
+            config.communication_config = preset.apply_to(config.communication_config.clone());
+        }
```
mlegner: Suggestion: This could be extracted into a separate function.

crates/walrus-sdk/src/config/upload_mode.rs
```diff
@@ -0,0 +1,113 @@
...
+            }
+            UploadMode::Aggressive => {
+                c.max_concurrent_writes = Some(1000);
+                c.max_data_in_flight = 1_250_000_000; // 1250 MB (1.25 GB)
+                c.request_rate_config.max_node_connections = 100;
+            }
+        }
+        c
+    }
+}
```
mlegner, line 72: I'm not quite sure whether it's intentional that the parameters are different in the "normal" and "force" cases. If not (which is my current understanding), how about something like the following?
```rust
struct CommunicationParameters {
    max_concurrent_writes: usize,
    max_data_in_flight: usize,
    max_node_connections: usize,
}

impl UploadMode {
    /// Applies this preset to a given communication config and returns the updated config.
    pub fn apply_to(
        self,
        mut config: ClientCommunicationConfig,
        force: bool,
    ) -> ClientCommunicationConfig {
        let default_bytes = crate::config::communication_config::default::max_data_in_flight();
        let default_node_conns = RequestRateConfig::default().max_node_connections;

        let CommunicationParameters {
            max_concurrent_writes,
            max_data_in_flight,
            max_node_connections,
        } = self.to_communication_parameters();

        if force {
            config.max_concurrent_writes = Some(max_concurrent_writes);
            config.max_data_in_flight = max_data_in_flight;
            config.request_rate_config.max_node_connections = max_node_connections;
        } else {
            if config.max_concurrent_writes.is_none() {
                config.max_concurrent_writes = Some(max_concurrent_writes);
            }
            if config.max_data_in_flight == default_bytes {
                config.max_data_in_flight = max_data_in_flight;
            }
            if config.request_rate_config.max_node_connections == default_node_conns {
                config.request_rate_config.max_node_connections = max_node_connections;
            }
        }

        config
    }

    fn to_communication_parameters(self) -> CommunicationParameters {
        match self {
            UploadMode::Conservative => CommunicationParameters {
                max_concurrent_writes: 1000,
                max_data_in_flight: 12_500_000, // 12.5 MB
                max_node_connections: 100,
            },
            UploadMode::Balanced => CommunicationParameters {
                max_concurrent_writes: 1000,
                max_data_in_flight: 125_000_000, // 125 MB
                max_node_connections: 100,
            },
            UploadMode::Aggressive => CommunicationParameters {
                max_concurrent_writes: 1280,
                max_data_in_flight: 1_250_000_000, // 1.25 GB
                max_node_connections: 100,
            },
        }
    }
}
```

mlegner (commented):
Thanks a lot, this is a very nice addition, @sadhansood.

I have a few questions and minor suggestions.

crates/walrus-sdk/src/config/upload_mode.rs
```diff
@@ -0,0 +1,113 @@
...
+                // Global cap (max_concurrent_writes) is set high so bytes‑in‑flight limit is the
+                // primary governor.
+                if c.max_concurrent_writes.is_none() {
+                    c.max_concurrent_writes = Some(1000);
+                }
+                if c.max_data_in_flight == default_bytes {
+                    c.max_data_in_flight = 50_000_000; // 50 MB
+                }
+                if c.request_rate_config.max_node_connections == default_node_conns {
+                    c.request_rate_config.max_node_connections = 100;
```
halfprice: question: all 3 modes have the same max_node_connections. Do we still need it here?

halfprice (commented):
Thanks @sadhansood for adding this new mode in CLI!

In addition to all the good comments that @mlegner asked, I also wonder the mode can be put inside the sdk, so that it can be applied in more general scenario.

crates/walrus-sdk/src/config/upload_mode.rs
```diff
@@ -0,0 +1,113 @@
...
+    reqwest_config::RequestRateConfig,
+};
+
+/// Upload preset modes for tuning client concurrency and network usage.
+///
+/// These presets only adjust in-memory ClientCommunicationConfig and are intended
+/// to be applied per-run by the CLI (or other callers). They do not mutate the
+/// on-disk configuration.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "kebab-case")]
```
liquid-helium: nit: I don't think we are using kebab-case in our client config file?

crates/walrus-sdk/src/config/upload_mode.rs
```diff
@@ -0,0 +1,113 @@
...
+    #[default]
+    Balanced,
+    /// Targets ~10 Gbps environments where higher memory and per-node parallelism are acceptable.
+    /// Sets bytes-in-flight near 10 Gb (~1.25 GB) and higher per-node cap.
+    Aggressive,
+}
+
+impl UploadMode {
+    /// Applies this preset to a given communication config and returns the updated config.
+    pub fn apply_to(self, mut c: ClientCommunicationConfig) -> ClientCommunicationConfig {
```
liquid-helium: nit: s/c/config

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1095,6 +1096,15 @@ pub struct CommonStoreOptions {
...
+    /// If set, the preset overrides config-file `communication_config` values.
+    /// By default, the preset only fills unset/defaulted fields.
+    #[arg(long)]
```
liquid-helium: with the above Option<UploadModeCli>, do we still need this `force_upload_mode` option?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1113,6 +1123,31 @@ pub struct FileOrBlobId {
     pub(crate) blob_id: Option<BlobId>,
 }
 
+/// CLI enum for selecting upload presets. Converted to SDK UploadMode at runtime.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, clap::ValueEnum)]
+#[serde(rename_all = "kebab-case")]
```
liquid-helium: nit: I think we are using camelCase for the args in this file.

crates/walrus-service/src/client/cli/args.rs
sadhansood: Using UploadMode directly from the CLI would drag a clap dependency into the SDK, so I kept a thin UploadModeCli wrapper and convert to UploadMode at runtime.

crates/walrus-service/src/client/cli/args.rs
sadhansood: done

sadhansood: The CLI now uses an Option<UploadMode>-like semantics: CLI flag overrides config if set; otherwise, the config file is respected.

crates/walrus-service/src/client/cli/runner.rs
sadhansood: done

crates/walrus-sdk/src/config/upload_mode.rs
sadhansood: done

crates/walrus-sdk/src/config/upload_mode.rs
sadhansood: good point, removed

crates/walrus-sdk/src/config/upload_mode.rs
sadhansood: fixed, thanks

crates/walrus-service/src/client/cli/args.rs
sadhansood: good call, no it isnt needed - removed it.

crates/walrus-sdk/src/config.rs
sadhansood: I strongly feel that "force" is an execution-time knob. Persisting it risks surprising overrides and complicates precedence imo.

sadhansood:
Thank you @mlegner , @liquid-helium and @halfprice for all your great comments. I feel like they've all been addressed.

crates/walrus-service/src/client/cli/args.rs
mlegner: Hmm... I may have misunderstood the semantics of the `force` option. I thought it would be used to control whether or not to overwrite parameters that are explicitly set in the communication config. 

IIUC, the new version always overwrites these. As there are only two parameters, this seems reasonable to me, but we may have to revisit it if our presets control more parameters.

crates/walrus-service/src/client/cli/args.rs
mlegner: That makes sense. It still feels a bit suboptimal to duplicate the code, but I didn't find a cleaner way... 😢

crates/walrus-sdk/src/config/upload_mode.rs
```diff
@@ -0,0 +1,83 @@
...
+            },
+        }
+    }
+
+    /// Applies this preset to a given communication config and returns the updated config.
+    /// If `force` is true, overrides values unconditionally; otherwise only fills defaults.
+    pub fn apply_to(
+        self,
+        mut config: ClientCommunicationConfig,
+        force: bool,
```
mlegner: AFAICT, this is now always `true`. I think we should either "bring back" the `force` parameter or remove this function parameter.

mlegner (approved):
Thanks a lot, @sadhansood!

crates/walrus-sdk/src/config/upload_mode.rs
halfprice: +1

I think maybe we can just call this force_apply_to() and remove this parameter. Also, the semantic of `force` is clearer: completely update the two parameters to the one in the config. Whereas the non `force` version, it may still update it when the config is not set.

halfprice (approved):
LGTM! Thanks for the improvement @sadhansood !

PR #2538, opened 2025-09-22
fix: extending blob at the latest epoch

halfprice:
## Description

test_repeated_node_crash exposes a bug where the first blob certified event may not trigger blob sync due to that
the blob may have been expired based on the current committee epoch. However, in the same epoch where
the blob was expired based the first certified event, the blob may be extended. Currently, blob extension does not
trigger blob sync, which may leave the node to miss the data.

To fix this issue, we need to allow blob extension event to also trigger blob sync if the blob misses some shard data.
There is another complication associated with this method because `current_event_epoch()` may not be available
to use to check blob data integrity. In this case, we just need to blindly start the blob sync: if the blob is fully stored,
blob sync still won't do any work.

## Test plan

200 test_repeated_node_crash
new unit test added to test blob extension triggered blob sync.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -204,9 +204,6 @@ impl BackgroundEventProcessor {
         let histogram_set = self.node.metrics.recover_blob_duration_seconds.clone();
 
         if !self.node.is_blob_certified(&event.blob_id)?
-            // For blob extension events, the original blob certified event should already recover
-            // the entire blob, and we can skip the recovery.
-            || event.is_extension
             || self.node.storage.node_status()?.is_catching_up()
             || self
                 .node
@@ -217,6 +214,12 @@ impl BackgroundEventProcessor {
                 .await?
         {
             event_handle.mark_as_complete();
+
+            tracing::debug!(
+                "skipping blob certified event for blob id: {}, epoch: {}",
+                event.blob_id,
+                event.epoch
+            );
 
             walrus_utils::with_label!(histogram_set, metrics::STATUS_SKIPPED)
                 .observe(start.elapsed().as_secs_f64());
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -203,22 +203,41 @@ impl BackgroundEventProcessor {
 
         let histogram_set = self.node.metrics.recover_blob_duration_seconds.clone();
 
-        if !self.node.is_blob_certified(&event.blob_id)?
+        // Get the current event epoch to check if the blob is fully stored up to the shard
+        // assignment in this epoch. If the current event epoch is not set (the node may just
+        // started), we always start the blob sync.
+        let current_event_epoch = self.node.try_get_current_event_epoch();
+
+        #[allow(unused_mut)]
+        let mut skip_blob_sync_in_test = false;
+        // No op when not in simtest mode.
+        sui_macros::fail_point_if!(
+            "skip_non_extension_certified_event_triggered_blob_sync",
+            || {
+                skip_blob_sync_in_test = !event.is_extension;
+            }
+        );
+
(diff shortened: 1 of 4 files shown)


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -285,6 +285,10 @@ impl BlobSyncHandler {
     }
 
     /// Starts a blob sync for the given `blob_id` and `certified_epoch`.
+    /// The `certified_epoch` is only used to find the proper committee to read the slivers, and
+    /// may not be the same as the current epoch.
+    /// To find the shard assignment the blob is stored at, we uses `current_event_epoch()` instead.
+    /// See `recover_blob_slivers()` for more details.
```
mlegner: Nits:
```suggestion
    ///
    /// The `certified_epoch` is only used to find the proper committee to read the metadata and
    /// recovery symbols during epoch change, and may differ from the current epoch.
    ///
    /// To find the shard assignment the blob is stored at, we use
    /// [`StorageNodeInner::current_event_epoch`] instead. See
    /// [`BlobSynchronizer::recover_blob_slivers`] for more details.
```

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -203,21 +203,43 @@ impl BackgroundEventProcessor {
...
         let histogram_set = self.node.metrics.recover_blob_duration_seconds.clone();
 
-        if !self.node.is_blob_certified(&event.blob_id)?
-            // For blob extension events, the original blob certified event should already recover
-            // the entire blob, and we can skip the recovery.
-            || event.is_extension
+        // Get the current event epoch to check if the blob is fully stored up to the shard
+        // assignment in this epoch. If the current event epoch is not set (the node may just
+        // started), we always start the blob sync.
+        let current_event_epoch = self.node.try_get_current_event_epoch();
```
mlegner, line 209: Question: (Why) can't we just wait for this, i.e., call `self.node.current_event_epoch().await`?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -203,21 +203,43 @@ impl BackgroundEventProcessor {
...
         {
             event_handle.mark_as_complete();
 
+            tracing::debug!(
+                "skipping syncing blob during certified event processing for blob id: {}, \
+                epoch: {}, is_extension: {}",
+                event.blob_id,
+                event.epoch,
+                event.is_extension
+            );
```
mlegner, line 240: Nit: How about this?
```suggestion
            tracing::debug!(
                %event.blob_id,
                %event.epoch,
                %event.is_extension,
                "skipping syncing blob during certified event processing",
            );
```

mlegner (approved):
Thanks for the fix, @halfprice! 🙏

crates/walrus-service/src/node/blob_event_processor.rs
liquid-helium: +1, if loading the current epoch is async, would this create unnecessary sync?

I scanned the code, it seems to be async, but I could miss something.

halfprice: The problem is following, say after restart, first event is blob extension event, and the second event is epoch change start event. Blob extension event won't set the `current_event_epoch()`, therefore this will block. However, before processing epoch change start event, we need to [wait until all prior events are processed](https://github.com/MystenLabs/walrus/blob/27ef5a3922e5206f160e7ffd41b2fe22e1486c05/crates/walrus-service/src/node.rs#L1554). Because of this, there is a deadlock.

I've thought for quite sometime how to address this, and by initiating the blob sync is the best simple solution I can come up right now.

mlegner: Ah, this is a very good point, I didn't think about this case. 😬 
Good that you thought about it!

PR #2540, opened 2025-09-22
test: make unit tests work with `cargo test`

mlegner:
## Description

We normally run tests with `cargo nextest`, which has some additional isolation mechanisms compared to `cargo test`.
This resulted in some tests not running successfully with `cargo test`.

This adds locks for some unit tests to prevent them from running simultaneously in the same process. It also unifies how tracing subscribers are initialized in tests.

Finally, this runs unit tests in the standard CI workflow to ensure that we don't accidentally break them again.

Unfortunately, integration tests are still a bit flaky with `cargo test` and need to be fixed separately.

## Test plan

Added `cargo test` to the CI workflow + ran it locally.


halfprice (approved):
Thanks @mlegner for improving tests!

PR #2542, opened 2025-09-22
feat(node): update blob info when deletable blobs expire

mlegner:
## Description

For deletable blobs, we store object counts in the aggregate blob info. These need to be updated when deletable blobs expire.

This adds a new (background) task running at the beginning of each epoch that cleans up expired entries in the per-object blob info and updates the aggregate blob info for expired deletable blobs.

This also improves some test utils and fixes several E2E tests that reused object IDs in an incorrect way.

## Test plan

A newly added E2E test (`test_storage_nodes_do_not_serve_data_for_expired_deletable_blobs`) verifies that this works correctly. I verified that the test fails without the added `process_expired_blob_objects` task and succeeds with it.

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -918,6 +918,88 @@ async fn test_storage_nodes_do_not_serve_data_for_deleted_blobs() -> TestResult 
     assert_eq!(count_deleted, 1, "should have deleted one blob");
     tokio::time::sleep(Duration::from_millis(50)).await;
 
+    check_that_blob_is_not_available(client, blob_id).await
+}
+
+/// Tests that nodes correctly update the blob info for expired deletable blobs and no longer serve
+/// data for them.
+#[ignore = "ignore E2E tests by default"]
+#[walrus_simtest]
+async fn test_storage_nodes_do_not_serve_data_for_expired_deletable_blobs() -> TestResult {
+    telemetry_subscribers::init_for_testing();
+    let epoch_duration = Duration::from_secs(15);
+    let (_sui_cluster_handle, cluster, client, _) = test_cluster::E2eTestSetupBuilder::new()
+        .with_epoch_duration(epoch_duration)
+        .build()
+        .await?;
+    let client = client.as_ref();
+    let blob = walrus_test_utils::random_data(314);
+    let blobs = vec![blob.as_slice()];
+
+    // Store the blob for 42 epochs.
+    let [
+        BlobStoreResult::NewlyCreated {
+            blob_object:
+                Blob {
+                    blob_id,
+                    id: object_id,
+                    ..
+                },
+            ..
+        },
+    ] = client
+        .reserve_and_store_blobs(
+            &blobs,
+            &StoreArgs::default_with_epochs(42)
+                .no_store_optimizations()
+                .deletable(),
+        )
+        .await?[..]
+    else {
+        panic!("should have exactly one newly created blob");
+    };
+
+    // Store the blob again for a single epoch.
+    let [
+        BlobStoreResult::NewlyCreated {
+            ref blob_object, ..
+        },
+    ] = client
+        .reserve_and_store_blobs(
+            &blobs,
+            &StoreArgs::default_with_epochs(1)
+                .no_store_optimizations()
+                .deletable(),
+        )
+        .await?[..]
+    else {
+        panic!("should have exactly one newly created blob");
+    };
+    assert_eq!(blob_object.blob_id, blob_id);
+    let early_end_epoch = blob_object.storage.end_epoch;
+
+    // Make sure we can read the blob.
+    assert_eq!(client.read_blob::<Primary>(&blob_id).await?, blob);
+
+    // Delete the blob with the later end epoch.
+    client.delete_owned_blob_by_object(object_id).await?;
+
+    // Wait until the end epoch of the early blob is reached, at which point it should no longer be
+    // available.
+    tokio::time::timeout(
+        epoch_duration * 3,
+        cluster.wait_for_nodes_to_reach_epoch(early_end_epoch),
+    )
+    .await
+    .expect("timed out waiting for nodes to reach end epoch of first blob");
+
+    check_that_blob_is_not_available(client, blob_id).await
+}
+
+async fn check_that_blob_is_not_available(
+    client: &WalrusNodeClient<SuiContractClient>,
+    blob_id: BlobId,
+) -> TestResult {
     let status_result = client
         .get_verified_blob_status(
             &blob_id,
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -926,7 +926,7 @@ async fn test_storage_nodes_do_not_serve_data_for_deleted_blobs() -> TestResult 
 #[ignore = "ignore E2E tests by default"]
 #[walrus_simtest]
 async fn test_storage_nodes_do_not_serve_data_for_expired_deletable_blobs() -> TestResult {
-    telemetry_subscribers::init_for_testing();
+    walrus_test_utils::init_tracing();
     let epoch_duration = Duration::from_secs(15);
     let (_sui_cluster_handle, cluster, client, _) = test_cluster::E2eTestSetupBuilder::new()
         .with_epoch_duration(epoch_duration)
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1584,6 +1584,12 @@ impl StorageNode {
         // shards are created.
         let shard_map_lock = self.inner.storage.lock_shards().await;
 
+        // TODO(WAL-473): Move this to a background task.
+        self.inner
+            .storage
+            .process_expired_blob_objects(event.epoch)
+            .await?;
+
         // Now the general tasks around epoch change are done. Next, entering epoch change logic
         // to bring the node state to the next epoch.
         self.execute_epoch_change(event_handle, event, shard_map_lock)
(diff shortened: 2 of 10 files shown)


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -358,6 +358,73 @@ impl BlobInfoTable {
...
             .unwrap_or(0))
     }
+
+    /// Processes blobs that have expired in the given epoch.
+    ///
+    /// This function iterates over the per-object blob info table, deleting any entries that have
+    /// an end epoch equal to or less than the current epoch, and updating the aggregate blob info
+    /// table in case of deletable blobs to reflect the new status of the blob objects.
+    #[tracing::instrument(skip_all)]
+    pub(crate) async fn process_expired_blob_objects(
```
liquid-helium, line 368: Question: is the idea that this is called when a new epoch has already started, so there won't be any conflicts? I feel it is likely the upcoming indexer will rely on the same mechanism to GC the expired entries, very glad to see this happening for the blob data :)

crates/walrus-service/src/node/storage.rs
```diff
@@ -585,6 +585,25 @@ impl Storage {
...
+    ///
+    /// This function is called during epoch change to clean up the blob info for blob objects that
+    /// are no longer valid.
+    pub(crate) async fn process_expired_blob_objects(
+        &self,
+        current_epoch: Epoch,
+    ) -> anyhow::Result<()> {
+        tracing::info!(epoch = %current_epoch, "processing expired deletable blobs");
+
+        self.blob_info
```
halfprice, line 598: Suggestion: can we add an option in the config to optionally run this cleanup logic, in case it has significant prod impact? In this way, we can first test this out in our node and then turn it on by default in all nodes.

halfprice: Also, I think if we add this option in this PR, we can implement moving this function to the background in a follow up PR. I think running this function in the background requires that there are no two instances of this function running at the same time, so it's not just a simple spawn

crates/walrus-service/src/node.rs
```diff
@@ -1584,6 +1584,12 @@ impl StorageNode {
         // shards are created.
         let shard_map_lock = self.inner.storage.lock_shards().await;
 
+        // TODO(WAL-473): Move this to a background task.
+        self.inner
+            .storage
+            .process_expired_blob_objects(event.epoch)
+            .await?;
```
halfprice: remark: this is a bit tricky, so @mlegner if you have already thought through this, I trust you. To me, I need to think more about whether this should be here or put it after latest_event_epoch_sender.send(). Theoretically, the node is still operating at event.epoch-1 at this moment (shards are still in old epoch). I cannot think of a case where this may cause a problem, but on the other hand, the blob info table potentially may have reflected that the node is already in the new epoch.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -358,6 +358,73 @@ impl BlobInfoTable {
...
+        &self,
+        current_epoch: Epoch,
+    ) -> anyhow::Result<()> {
+        tracing::info!(epoch = %current_epoch, "processing expired deletable blobs");
```
halfprice: Nit: this is not just for deletable blob, correct? It also clears permanent blob's per object blob info

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -801,21 +879,26 @@ impl ValidBlobInfoV1 {
...
         let exists_certified_permanent_blob = self
             .permanent_certified
             .as_ref()
             .is_some_and(|p| p.end_epoch > current_epoch);
-        let maybe_exists_certified_deletable_blob = self.count_deletable_certified > 0
+        let probably_exists_certified_deletable_blob = self.count_deletable_certified > 0
             && self
                 .latest_seen_deletable_certified_epoch
-                .is_some_and(|l| l > current_epoch);
+                .is_some_and(|e| e > current_epoch);
```
halfprice, line 899: question: not sure if I fully understand the logic here. If count_deletable_certified > 0, but latest_seen_deletable_certified_epoch < current_epoch, isn't that also indicate that blob is certified? I think I may miss something here.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Yes, exactly.

crates/walrus-service/src/node/storage.rs
mlegner: Done. I've created a follow-up issue to move the task to the background and enable it by default.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Ah, this might be due to a suboptimal naming of `latest_seen_deletable_certified_epoch`. This should maybe rather be `latest_seen_deletable_certified_**end**_epoch`.

The idea is to keep track of the latest end epoch of any deletable blob. This allows us to conclude the blob is expired even before cleaning up the counts. I can clarify this in the docstrings.

crates/walrus-service/src/node.rs
mlegner: The whole task only operates on the blob-info table and never uses the `current_event_epoch`, so I think the exact position in the code does not matter. Also, as the blob info is updated in the sequential part of event processing, it should be fully up-to-date there.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: I've renamed these fields in 63d2a9ef0a5124590a8855cdbcc63d2fcccabfa4 and also added a minor improvement there, namely unsetting these fields when the deletable counters reach 0. I'm happy to create a separate PR for these changes (as they are unrelated to the rest of this PR) or keep them here.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -918,6 +918,88 @@ async fn test_storage_nodes_do_not_serve_data_for_deleted_blobs() -> TestResult
...
+    // Wait until the end epoch of the early blob is reached, at which point it should no longer be
+    // available.
+    tokio::time::timeout(
+        epoch_duration * 3,
+        cluster.wait_for_nodes_to_reach_epoch(early_end_epoch),
+    )
+    .await
+    .expect("timed out waiting for nodes to reach end epoch of first blob");
+
+    check_that_blob_is_not_available(client, blob_id).await
```
halfprice, line 996: Question: will the check fail without this PR?

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: Thanks for the clarification! This makes sense now!

crates/walrus-service/src/node.rs
```diff
@@ -1584,6 +1586,18 @@ impl StorageNode {
...
 
+        // TODO(WAL-1040): Move this to a background task.
+        if self
+            .inner
+            .garbage_collection_config
+            .enable_blob_info_cleanup
+        {
+            self.inner
+                .storage
+                .process_expired_blob_objects(event.epoch)
```
halfprice, line 1597: question: right now, this runs if the node is in recovery, which is fine during development. In the final version, I think maybe we can only run this if the node is in Active status, WDYT?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -913,33 +1010,40 @@ impl ValidBlobInfoV1 {
...
+        latest_seen_end_epoch: &mut Option<Epoch>,
+    ) {
         debug_assert!(*counter > 0);
         *counter = counter.checked_sub(1).unwrap_or_else(|| {
             tracing::error!("attempt to delete blob when count was already 0");
             0
         });
+        if *counter == 0 {
+            *latest_seen_end_epoch = None;
+        }
```
halfprice, line 1046: question: we don't cleanup the BlobInfo entry yet if deletable == 0 and there is no permanent blob, is that correct?

halfprice (approved):
Great work @mlegner ! Looks great! I only have some clarification questions.

crates/walrus-service/src/node/storage/blob_info.rs
liquid-helium: I'm wondering if it would be feasible to maintain a list of all expiration epochs rather than just tracking the most recent one? I'm encountering similar challenges with blob mapping tracking, so I'm interested to see if there are any lessons or patterns here that could be applicable to the indexer.

halfprice: I think Markus's design there is to have a constant access time for every row in blob info table. Otherwise, it may become an attack vector. This is also the main idea IIRC behind all the counters in the BlobInfo struct.

halfprice (approved):

crates/walrus-e2e-tests/tests/test_client.rs
mlegner: Yes. Without this PR, we would still have one deletable blob at this point and the `latest_certified_end_epoch` would be in the future, so we would conclude that it is still valid.

crates/walrus-service/src/node.rs
mlegner: Good question. Ideally it runs before we start the actual recovery such that we don't unnecessarily recover expired deleted blobs. WDYT?

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Yes, that's correct. We don't actually clean up the aggregate blob info at all. This will be part of the next step, that is the actual data cleanup. Until then, we need to keep track of all blobs for which we potentially store metadata or slivers.

mlegner:
@halfprice I've enabled auto-merge. PTAL at my answers above and resolve the conversations if you agree. 🙂

crates/walrus-service/src/node.rs
halfprice: Yeah, that sounds good!

PR #2548, opened 2025-09-24
feat: add walrus-indexer storage skeleton for architectural review

liquid-helium:
## Description

This PR introduces a minimal skeleton of the walrus-indexer storage module extracted from #2503, providing just enough structure to review the storage schema without implementation details.

Contributes to WAL-1052

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-indexer/Cargo.toml
+++ b/crates/walrus-indexer/Cargo.toml
@@ -30,7 +30,7 @@ walrus-core.workspace = true
 walrus-sdk.workspace = true
 walrus-service.workspace = true
 walrus-sui.workspace = true
-walrus-utils.workspace = true```
+walrus-utils.workspace = true
 
 # Sui dependencies
 sui-types = { workspace = true }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12748,7 +12748,7 @@ dependencies = [
 
 [[package]]
 name = "walrus-indexer"
-version = "0.1.0"
+version = "1.35.0"
 dependencies = [
  "anyhow",
  "async-trait",


crates/walrus-indexer/Cargo.toml
```diff
@@ -0,0 +1,38 @@
+[package]
+name = "walrus-indexer"
+version = "0.1.0"
+edition = "2021"
+authors = ["Walrus Foundation"]
+license = "Apache-2.0"
```
mlegner: Suggestion: Reuse the workspace parameters here.
```suggestion
publish = false
authors.workspace = true
version.workspace = true
edition.workspace = true
license.workspace = true
```

crates/walrus-indexer/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+tokio-util = { workspace = true }
+tracing = { workspace = true }
+typed-store = { workspace = true }
+
+# Walrus dependencies
+walrus-core = { path = "../walrus-core" }
+walrus-sdk = { path = "../walrus-sdk" }
+walrus-service = { path = "../walrus-service" }
+walrus-sui = { path = "../walrus-sui" }
+walrus-utils = { path = "../walrus-utils" }
```
mlegner: Suggestion: Also use the workspace here.
```suggestion
walrus-core.workspace = true
walrus-sdk.workspace = true
walrus-service.workspace = true
walrus-sui.workspace = true
walrus-utils.workspace = true```

File: crates/walrus-indexer/src/storage.rs [hunk 3]
```diff
@@ -0,0 +1,471 @@
...
+
+/// Key structure for primary index: blob_manager/identifier/sequence_number.
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct PrimaryIndexKey {
+    /// The blob manager ID.
+    blob_manager: ObjectID,
+    /// The identifier.
+    identifier: String,
+    /// The sequence number.
+    sequence_number: u64,
```
halfprice, line 49: question: what is sequence number? IIUC, blob_manager is the namespace, identifier is the user defined file name, is that correct?

crates/walrus-indexer/src/storage.rs
```diff
@@ -0,0 +1,471 @@
...
+    fn new(blob_manager: ObjectID, identifier: String, sequence_number: u64) -> Self {
+        Self {
+            blob_manager,
+            identifier,
+            sequence_number,
+        }
+    }
+
+    /// Custom serialization to preserve alphanumeric order.
+    /// Format: [blob_manager_bytes(32 bytes)] / [identifier_utf8] / [sequence_be_bytes(8 bytes)].
```
halfprice, line 64: question: do we want to include some kind of versioning in this scheme?

crates/walrus-indexer/src/storage.rs
```diff
@@ -0,0 +1,471 @@
...
+
+impl std::fmt::Display for ObjectIndexKey {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.object_id)
+    }
+}
+
+/// Key structure for quilt patch index.
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct QuiltPatchIndexKey {
```
halfprice, line 180: Question: what's the relationship between QuiltPatchIndexKey and PrimaryIndexKey?

deny.toml
```diff
@@ -4,6 +4,8 @@
 [advisories]
 version = 2
 ignore = [
+  # `fxhash` is unmaintained, but is a transitive dependency via starlark_map from Sui.
+  { id = "RUSTSEC-2025-0057", reason = "transitive dependency from Sui via starlark_map" },
```
halfprice: Nit: I think this is no longer needed.

crates/walrus-indexer/src/storage.rs
```diff
@@ -0,0 +1,471 @@
...
+    db: Arc<typed_store::rocks::RocksDB>,
+    // Primary index: blob_manager/identifier -> IndexTarget.
+    primary_index: DBMap<String, IndexTarget>,
+    // Object index: object_id -> blob_manager/identifier.
+    object_index: DBMap<String, ObjectIndexValue>,
+    // Sequence store: stores the last processed sequence number for resumption.
+    sequence_store: DBMap<String, u64>,
+    // Index for quilt patches, so that we can look up quilt patches by the corresponding
+    // files' blob IDs.
+    quilt_patch_index: DBMap<String, QuiltPatchId>,
```
halfprice, line 378: A couple of questions:
- primary_index: makes a lot of sense! (not really a question)
- what is object_index will be used for? Lifetime management?
- Unclear what is `sequence`
- for quilt_patch_index, based on the comment, the key is `files' blob IDs`, but the key of the DBMap is String. Is it supposed to be BlobID or the comment is out dated?

crates/walrus-indexer/src/storage.rs
```diff
@@ -0,0 +1,471 @@
...
+
+        Ok(Self {
+            patch_blob_id,
+            object_id,
+        })
+    }
+}
+
+// For scanning all entries with a specific blob_manager and identifier.
+fn get_primary_index_bounds(blob_manager: &ObjectID, identifier: &str) -> (Vec<u8>, Vec<u8>) {
```
halfprice, line 248: Nit (can be created in a followup PR): these helper functions can use some tests.

halfprice (approved):
The PR in general LGTM. I just have some clarification question, and since this currently does not have much impact in prod code, I think we can move a bit fast.

crates/walrus-indexer/src/storage.rs
liquid-helium: Great question.
Sequence number is the event sequence number technically, but semantically, it is the version number for the index entry.

crates/walrus-indexer/src/storage.rs
liquid-helium: Do you mean a version number for the entire schema?

That's a good catch! I thought about this, and my current stand is: 
since the schema is not syntactically defined, rather represented by these low-level key encoding, and there exist this coherent relation between different keys on different column families, we can encode a version number into the keys and/or values, and the final code will be complex :/.
Instead, the current strategy is that for all new 'versions' of schemas, we use a new set of column families.

For new types of 'blobs', e.g., large blobs, we can add new variant to the IndexTarget enum without changing the existing schema.

WDYT?

crates/walrus-indexer/src/storage.rs
liquid-helium: PrimaryIndexKey is for ALL `files` stored in Walrus, here `file` can be a regular blob, or a quilt patch, the key is the user defined identifiers.

QuiltPatchIndex is keyed by the nominal `BlobId` of the quilt patches, it indexes only patches.

crates/walrus-indexer/src/storage.rs
liquid-helium: Good catch! Sure, there are some tests in the prototype, but since I haven't added any read functions in this PR, due to some changes with the typed store, hence I didn't include any tests in the current PR, (we can write, but read will require reading directly to the db). 

The tests will be added once the read functions are ready.

crates/walrus-indexer/src/storage.rs
liquid-helium: The `object_index` is mapping from the Sui object id to the primary index entry, so that if a blob is deleted via the Sui object, which emits an event containing the Sui object ID, we know which entries to clean up.

Also, for expiration, similar to the current `by_object` table in the Walrus main db, we check the expiration by the object id, which is compatible with the current approach, if eventually we change how lifetime is managed, we can deprecate this index. I hope that makes sense to you.

I explained sequence in an above comment, but in short, it is the version of the index entry.

Long story short, see this [PR](https://github.com/MystenLabs/walrus/pull/2525/files), we cannot get alphanumerical order with the bincode, that's why the keys in this files are Strings, to make it easy to construct key ranges.

deny.toml
liquid-helium: ACK.

crates/walrus-indexer/src/storage.rs
halfprice: I see. I think I didn't connect the String with the to_key() function earlier. SGTM

halfprice (approved):
Thanks @liquid-helium for addressing the comments! LGTM!

PR #2549, opened 2025-09-24
fix(core): optimize memory allocations during blob decoding

mlegner:
## Description

In the past, when we were using RaptorQ for the 1D encoding, it was necessary to keep around intermediate decoding state as the exact number of slivers needed for decoding was unknown and we would potentially have to reuse the decoder for a new decoding attempt.

This is no longer the case with Reed-Solomon, as we know the exact number of slivers needed for decoding. As a result, we can allocate a single 1D decoder and reuse it for all sub-decodings during blob decoding.

This commit also makes some other changes to reduce the number of allocations during blob decoding, specifically by using flat arrays instead of nested vectors where possible.

Both together result in both a **major reduction in memory usage** and a **significant performance improvement**:

- The peak memory usage while reading a 1GiB sliver from the public Testnet (before the blob re-encoding) was reduced from more than 10GiB to ~2.5GiB.
- The decoding time (without re-encoding) for 1GiB blobs was reduced by more than 60% (corresponding to a ~160% throughput improvement).

<details><summary>Benchmark details</summary>

```console
$ cargo bench -p walrus-core blob_decoding/decode/ -- --baseline main
     Running benches/basic_encoding.rs (target/release/deps/basic_encoding-669d9b131c7330c5)
     Running benches/blob_encoding.rs (target/release/deps/blob_encoding-6ae446cf9853bda3)
blob_decoding/decode/1B time:   [237.42 ms 239.74 ms 242.12 ms]
                        thrpt:  [4.1302   B/s 4.1712   B/s 4.2119   B/s]
                 change:
                        time:   [-17.007% -16.021% -14.985%] (p = 0.00 < 0.05)
                        thrpt:  [+17.626% +19.077% +20.492%]
                        Performance has improved.
blob_decoding/decode/1KiB
                        time:   [236.83 ms 238.76 ms 241.03 ms]
                        thrpt:  [4.1488 KiB/s 4.1883 KiB/s 4.2224 KiB/s]
                 change:
                        time:   [-20.303% -18.427% -16.803%] (p = 0.00 < 0.05)
                        thrpt:  [+20.196% +22.590% +25.475%]
                        Performance has improved.
blob_decoding/decode/1MiB
                        time:   [238.91 ms 240.91 ms 243.52 ms]
                        thrpt:  [4.1065 MiB/s 4.1510 MiB/s 4.1856 MiB/s]
                 change:
                        time:   [-18.806% -17.945% -16.974%] (p = 0.00 < 0.05)
                        thrpt:  [+20.444% +21.869% +23.162%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high mild
blob_decoding/decode/16MiB
                        time:   [285.12 ms 286.72 ms 289.04 ms]
                        thrpt:  [55.356 MiB/s 55.804 MiB/s 56.117 MiB/s]
                 change:
                        time:   [-39.476% -38.440% -37.370%] (p = 0.00 < 0.05)
                        thrpt:  [+59.667% +62.443% +65.223%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
Benchmarking blob_decoding/decode/256MiB: Warming up for 10.000 ms
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 11.4s.
blob_decoding/decode/256MiB
                        time:   [1.0962 s 1.1041 s 1.1144 s]
                        thrpt:  [229.72 MiB/s 231.86 MiB/s 233.54 MiB/s]
                 change:
                        time:   [-60.495% -60.093% -59.635%] (p = 0.00 < 0.05)
                        thrpt:  [+147.74% +150.58% +153.13%]
                        Performance has improved.
Found 1 outliers among 10 measurements (10.00%)
  1 (10.00%) high severe
Benchmarking blob_decoding/decode/1GiB: Warming up for 10.000 ms
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 40.2s.
blob_decoding/decode/1GiB
                        time:   [3.9828 s 4.0173 s 4.0607 s]
                        thrpt:  [252.18 MiB/s 254.90 MiB/s 257.10 MiB/s]
                 change:
                        time:   [-62.436% -61.965% -61.439%] (p = 0.00 < 0.05)
                        thrpt:  [+159.33% +162.91% +166.21%]
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) high mild
  1 (10.00%) high severe

```

</details> 

## Test plan

Existing test suite.

Additionally checked memory usage with `cargo instruments` and performance through benchmarks (see above).

---

## Release notes

- [x] Aggregator: Major reduction in memory usage for large blobs.
- [x] CLI: Major reduction in memory usage when reading large blobs.

### DIFF
--- a/crates/walrus-core/src/encoding/basic_encoding.rs
+++ b/crates/walrus-core/src/encoding/basic_encoding.rs
@@ -49,16 +49,16 @@ pub trait Decoder: Sized {
     /// Attempts to decode the source data from the provided iterator over
     /// [`DecodingSymbol`s][DecodingSymbol].
     ///
-    /// Returns the source data as a byte vector if decoding succeeds or `None` if decoding fails.
+    /// Returns the source data as a byte vector if decoding succeeds or a [`DecodeError`] if
+    /// decoding fails.
     ///
-    /// If decoding failed due to an insufficient number of provided symbols, it can be continued
-    /// by additional calls to [`decode`][Self::decode] providing more symbols.
+    /// Symbols of incorrect size are dropped with a warning.
+    ///
+    /// If decoding failed due to an insufficient number of provided symbols
+    /// ([`DecodeError::DecodingUnsuccessful`]), it can be continued by additional calls to
+    /// [`decode`][Self::decode] providing more symbols.
     ///
     /// After the decoding is complete, the decoder can be reused for a new decoding.
-    ///
-    /// # Panics
-    ///
-    /// Can panic if the provided symbols have an incorrect size.
     fn decode<T, U>(&mut self, symbols: T) -> Result<Vec<u8>, DecodeError>
     where
         T: IntoIterator,
@@ -262,6 +262,10 @@ impl Decoder for ReedSolomonDecoder {
         let decoder = &mut self.decoder;
         let symbol_size = self.source_symbols.symbol_size();
         for symbol in symbols.into_iter() {
+            if symbol.data.len() != usize::from(symbol_size.get()) {
+                tracing::warn!("dropping symbol of incorrect size");
+                continue;
+            }
             if symbol.index < self.n_source_symbols.get() {
                 self.source_symbols[usize::from(symbol.index)].copy_from_slice(&symbol.data);
                 let _ = decoder.add_original_shard(symbol.index.into(), symbol.data);
--- a/crates/walrus-core/src/encoding/basic_encoding.rs
+++ b/crates/walrus-core/src/encoding/basic_encoding.rs
@@ -2,7 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use alloc::vec::Vec;
-use core::{fmt, num::NonZeroU16};
+use core::{fmt, mem, num::NonZeroU16};
 
 use reed_solomon_simd;
 use tracing::Level;
@@ -10,7 +10,14 @@ use tracing::Level;
 use super::{DecodingSymbol, EncodingAxis, EncodingFactory};
 use crate::{
     EncodingType,
-    encoding::{DecodeError, EncodeError, InvalidDataSizeError, ReedSolomonEncodingConfig, utils},
+    encoding::{
+        DecodeError,
+        EncodeError,
+        InvalidDataSizeError,
+        ReedSolomonEncodingConfig,
+        Symbols,
+        utils,
+    },
 };
 
 /// The key of blob attribute, used to identify the type of the blob.
@@ -46,6 +53,12 @@ pub trait Decoder: Sized {
     ///
     /// If decoding failed due to an insufficient number of provided symbols, it can be continued
     /// by additional calls to [`decode`][Self::decode] providing more symbols.
+    ///
+    /// After the decoding is complete, the decoder can be reused for a new decoding.
+    ///
+    /// # Panics
+    ///
+    /// Can panic if the provided symbols have an incorrect size.
     fn decode<T, U>(&mut self, symbols: T) -> Result<Vec<u8>, DecodeError>
     where
         T: IntoIterator,
@@ -201,14 +214,15 @@ impl<'a> ReedSolomonEncoder<'a> {
 pub struct ReedSolomonDecoder {
     decoder: reed_solomon_simd::ReedSolomonDecoder,
     n_source_symbols: NonZeroU16,
-    symbol_size: NonZeroU16,
-    source_symbols: Vec<Vec<u8>>,
+    n_shards: NonZeroU16,
+    source_symbols: Symbols,
 }
 
 impl fmt::Debug for ReedSolomonDecoder {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        f.debug_struct("RSDecoder")
-            .field("symbol_size", &self.symbol_size)
+        f.debug_struct("ReedSolomonDecoder")
+            .field("n_source_symbols", &self.n_source_symbols)
+            .field("n_shards", &self.n_shards)
             .finish()
     }
 }
@@ -223,6 +237,7 @@ impl Decoder for ReedSolomonDecoder {
         symbol_size: NonZeroU16,
     ) -> Result<Self, DecodeError> {
         tracing::trace!("creating a new Decoder");
+        let source_symbols = Symbols::zeros(n_source_symbols.get().into(), symbol_size);
         Ok(Self {
             decoder: reed_solomon_simd::ReedSolomonDecoder::new(
                 n_source_symbols.get().into(),
@@ -231,8 +246,8 @@ impl Decoder for ReedSolomonDecoder {
             )
             .map_err(DecodeError::IncompatibleParameters)?,
             n_source_symbols,
-            symbol_size,
-            source_symbols: alloc::vec![alloc::vec![]; n_source_symbols.get().into()],
+            n_shards,
+            source_symbols,
         })
     }
 
@@ -243,9 +258,10 @@ impl Decoder for ReedSolomonDecoder {
         U: EncodingAxis,
     {
         let decoder = &mut self.decoder;
+        let symbol_size = self.source_symbols.symbol_size();
         for symbol in symbols.into_iter() {
             if symbol.index < self.n_source_symbols.get() {
-                self.source_symbols[usize::from(symbol.index)] = symbol.data.clone();
+                self.source_symbols[usize::from(symbol.index)].copy_from_slice(&symbol.data);
                 let _ = decoder.add_original_shard(symbol.index.into(), symbol.data);
             } else {
                 let _ = decoder.add_recovery_shard(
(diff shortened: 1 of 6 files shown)


crates/walrus-core/src/encoding/basic_encoding.rs
```diff
@@ -201,14 +214,15 @@ impl<'a> ReedSolomonEncoder<'a> {
 pub struct ReedSolomonDecoder {
     decoder: reed_solomon_simd::ReedSolomonDecoder,
     n_source_symbols: NonZeroU16,
-    symbol_size: NonZeroU16,
-    source_symbols: Vec<Vec<u8>>,
```
giac-mysten, line 205: Remark: this is great, i'm wondering why it wasn't the case already 🤦

crates/walrus-core/src/encoding/basic_encoding.rs
```diff
@@ -201,14 +214,15 @@ impl<'a> ReedSolomonEncoder<'a> {
 pub struct ReedSolomonDecoder {
```
giac-mysten, line 216: Nit: the docs above say "Wrapper to perform a **single** decoding with...". I guess this could be changed to "any nymber of"?

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -456,12 +457,18 @@ impl<'a> ExpandedMessageMatrix<'a> {
 #[derive(Debug)]
 pub struct BlobDecoder<'a, D: Decoder, E: EncodingAxis = Primary> {
     _decoding_axis: PhantomData<E>,
-    decoders: Vec<D>,
+    decoder: D,
     blob_size: usize,
     symbol_size: NonZeroU16,
+    sliver_length: usize,
+    n_columns: usize,
```
giac-mysten, line 465: Nit: maybe add a comment saying that this is the number of columns in the message matrix, i.e. the number of secondary slivers.

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -547,55 +577,66 @@ impl<'a, D: Decoder, E: EncodingAxis> BlobDecoder<'a, D, E> {
...
+                                .to_vec(),
+                        )
+                    });
+            let decoded_data = self.decoder.decode(symbols)?;
+            // Overwrite the decoding symbols in the workspace with the decoded data.
+            for (row_index, symbol) in decoded_data.chunks(self.symbol_usize()).enumerate() {
+                self.workspace[workspace_symbol_index(row_index, decoder_index)]
+                    .copy_from_slice(symbol);
+            }
+        }
```
giac-mysten, line 644: Praise: this part is great. The idea that you can pack the symbols in the decoding "workspace", while remembering their index, and then overwriting them because they are not needed anymore is brilliant!

Question: is there a big advantage to this in terms of memory footprint, compared to just keeping the symbols in their own structure? AFAIU, we still need to double the amount of memory when copying from the iterators to the workspace?

crates/walrus-core/src/encoding/basic_encoding.rs
giac-mysten, line 214: Nit: The comment above mentions
"/// Wrapper to perform a **single** decoding with Reed-Solomon for the provided parameters."
AFAIU, this is not the case anymore. Shoudl this be changed?

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: > AFAIU, we still need to double the amount of memory when copying from the iterators to the workspace?

Yes, this is true and a bit unfortunate. There might be a way to write slivers to the workspace directly when getting them from the storage nodes. However, that probably becomes pretty complicated and I'm not sure if/how much this actually improves the situation.

> is there a big advantage to this in terms of memory footprint, compared to just keeping the symbols in their own structure?

This is a good question. I needed some structure where I could collect the slivers (after checking them) as they are only supplied as an iterator. However, you are right that I could probably just collect them in a vector instead of writing them to the workspace. I'll investigate this a bit more.

giac-mysten: Very interesting, thanks for the details! I'm not sure there is a need to investigate this further; I feel the current implementation is pretty much optimal (barring the case you mentioned where the slivers are directly copied from the network to the workspace, which would be very complex).

I asked the question because it's a neat solution, but I was left wondering if I had missed something given the extra complication compared to just having a side vector :)

crates/walrus-core/src/encoding/basic_encoding.rs
mlegner: Done in 38a3af700a25cd06b86e79f5a4c03f5ad66e50e9.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: Done in 38a3af700a25cd06b86e79f5a4c03f5ad66e50e9.

crates/walrus-core/src/encoding/basic_encoding.rs
```diff
@@ -46,6 +53,12 @@ pub trait Decoder: Sized {
     ///
     /// If decoding failed due to an insufficient number of provided symbols, it can be continued
     /// by additional calls to [`decode`][Self::decode] providing more symbols.
+    ///
+    /// After the decoding is complete, the decoder can be reused for a new decoding.
+    ///
+    /// # Panics
+    ///
+    /// Can panic if the provided symbols have an incorrect size.
```
liquid-helium: Is this panic intentional, can we return an Error instead?

mlegner: We could instead log an error message if symbols of incorrect size are encountered and drop them.

liquid-helium: Sounds great, I am thinking a crash on a server when it is decoding some fabricated data would be unnecessary.

mlegner: Agreed. Note, however, that we already check all data we use for decoding before using it. So this potential panic is never triggered in any of our code.

Still, we can make this a bit more robust in case we change any verification code in the future.

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -513,27 +531,41 @@ impl<'a, D: Decoder, E: EncodingAxis> BlobDecoder<'a, D, E> {
...
+
+        let mut slivers_count = 0;
         for sliver in slivers {
-            let expected_len = self.decoders.len();
+            if slivers_count == expected_sliver_count {
+                tracing::info!("dropping surplus slivers during blob decoding");
+                break;
+            }
+
+            let expected_len = self.sliver_length;
```
sadhansood, line 574: Should we also dedupe sliver index here? `sliver_indices` is a vec, so it'll happily consume duplicate slivers.

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -547,55 +579,66 @@ impl<'a, D: Decoder, E: EncodingAxis> BlobDecoder<'a, D, E> {
...
+        for decoder_index in 0..self.sliver_length {
+            let symbols =
+                self.sliver_indices
+                    .iter()
+                    .enumerate()
+                    .map(|(row_index, sliver_index)| {
+                        DecodingSymbol::<E>::new(
+                            sliver_index.0,
+                            self.workspace[workspace_symbol_index(row_index, decoder_index)]
+                                .to_vec(),
```
sadhansood: Could we have borrowed the data ref here instead of allocating to vec?

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -547,55 +579,66 @@ impl<'a, D: Decoder, E: EncodingAxis> BlobDecoder<'a, D, E> {
...
+        });
+    }
+
+    fn perform_decoding(&mut self) -> Result<(), DecodeError> {
+        let workspace_symbol_index = |sliver_index: usize, decoder_index: usize| {
+            if E::IS_PRIMARY {
+                sliver_index * self.n_columns + decoder_index
+            } else {
+                decoder_index * self.n_columns + sliver_index
+            }
```
sadhansood: Nit: This logic might be nice to put in a function.

crates/walrus-core/src/encoding/basic_encoding.rs
mlegner: Done in 5e0004e44727a038a06f737449cebbf0280dfdc3.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: Unfortunately, the `DecodingSymbol` takes an owned vector. We could try changing this, but I'd prefer doing this separately as it's a larger refactoring.

In any case, I expect the compiler to be able to optimize this away. At least I don't actually see these allocations in the trace.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: Good point. I added deduplication in 2787c3f33a841fdcbf5a83c2169dbabbe5a38834.

crates/walrus-core/src/encoding/blob_encoding.rs
mlegner: I agree. The problem I faced was multiple borrows of `self`, one of which is mutable. That's why I went for a closure here. I've now slightly changed the code in 2787c3f33a841fdcbf5a83c2169dbabbe5a38834 to improve the overwriting for secondary slivers and removed this closure in the process.

crates/walrus-core/src/encoding/blob_encoding.rs
sadhansood: Thank you @mlegner !

PR #2554, opened 2025-09-25
feat(aggregator): add CLI argument to set size limit

mlegner:
## Description

Currently, aggregators attempt to serve any blob that is requested. Unfortunately, this can be abused easily to request very large blobs to cause a massive amount of required memory, which can cause the aggregator to run out of memory and get killed.

This adds a CLI argument to limit the size of blobs that can be requested.

## Test plan

Manually ran an aggregator and checked that the limit works as expected.

---

## Release notes

- [x] Aggregator: Adds a `--max-blob-size` option to limit the size of blobs that can be requested through the aggregator.


mlegner:
> in the long term, do we want to put the limit to RAM or other metrics, while keep the coming request in queue, and relaxing the limit on blob size a bit?

Yes, that's definitely a good idea, but probably a bit more involved than this quick fix.

PR #2559, opened 2025-09-25
chore: update install script to ensure a testnet configuration is available

wbbradley:
## Description

Note that this will also require a change to the relate Github Action var. I have that change queued up out-of-band.

## Test plan

Plan is to test a separate site-builder invocation and test it manually. And CI Pipeline.


docs/book/usage/setup.md
```diff
@@ -194,7 +194,7 @@ CLI configuration refers to the standard location for Sui configuration
 The easiest way to obtain the latest configuration is by downloading it directly from Walrus:
```
mlegner, line 199: Suggestion: Can we add something like the following after the first paragraph of this section?
```md
If you installed Walrus through the [install script](#nix-install), this configuration file was already created. If not, ...
```

mlegner (approved):
Very nice, thanks a lot, @wbbradley! Just one suggestion for the docs.

Regarding the failing CI job, we may need to add `npmjs\.com` to the [linkcheck include list](https://github.com/MystenLabs/walrus/blob/6cf2925098096bd4cc6d74327bf7c57e49601e4e/book.toml#L17).

wbbradley:
This is out of date. Closing.

PR #2565, opened 2025-09-28
fix: Use the correct db options in db-tools

liquid-helium:
## Description

Extract the db cf options, so that we can use the same options in db-tool.

The current one is kept until we are sure the new one is 100% correct.

Contributes to WAL-940

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/db_options.rs
+++ b/crates/walrus-service/src/node/db_options.rs
@@ -74,6 +74,87 @@ pub enum DbColumnFamily {
 }
 
 impl DbColumnFamily {
+    /// Parse the base column family name to extract shard ID.
+    ///
+    /// Expected format: "shard-{id}"
+    /// Returns the shard ID if the format matches, None otherwise.
+    fn parse_base_column_family_name(name: &str) -> Option<u16> {
+        name.strip_prefix("shard-")?.parse::<u16>().ok()
+    }
+
+    /// Parse a primary slivers column family name.
+    ///
+    /// Expected format: "shard-{id}/primary-slivers"
+    fn parse_primary_slivers(name: &str) -> Option<Self> {
+        let (base, suffix) = name.split_once('/')?;
+        if suffix != "primary-slivers" {
+            return None;
+        }
+        let shard_id = Self::parse_base_column_family_name(base)?;
+        Some(Self::PrimarySliver(shard_id))
+    }
+
+    /// Parse a secondary slivers column family name.
+    ///
+    /// Expected format: "shard-{id}/secondary-slivers"
+    fn parse_secondary_slivers(name: &str) -> Option<Self> {
+        let (base, suffix) = name.split_once('/')?;
+        if suffix != "secondary-slivers" {
+            return None;
+        }
+        let shard_id = Self::parse_base_column_family_name(base)?;
+        Some(Self::SecondarySliver(shard_id))
+    }
+
+    /// Parse a shard status column family name.
+    ///
+    /// Expected format: "shard-{id}/status"
+    fn parse_shard_status(name: &str) -> Option<Self> {
+        let (base, suffix) = name.split_once('/')?;
+        if suffix != "status" {
+            return None;
+        }
+        let shard_id = Self::parse_base_column_family_name(base)?;
+        Some(Self::ShardStatus(shard_id))
+    }
+
+    /// Parse a shard sync progress column family name.
+    ///
+    /// Expected format: "shard-{id}/sync-progress"
+    fn parse_shard_sync_progress(name: &str) -> Option<Self> {
+        let (base, suffix) = name.split_once('/')?;
+        if suffix != "sync-progress" {
+            return None;
+        }
+        let shard_id = Self::parse_base_column_family_name(base)?;
+        Some(Self::ShardSyncProgress(shard_id))
+    }
+
+    /// Parse a pending recover slivers column family name.
+    ///
+    /// Expected format: "shard-{id}/pending-recover-slivers"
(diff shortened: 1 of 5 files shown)


crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+        } else if name == pending_cf_name() {
+            Self::Pending
+        } else if name == failed_to_attest_cf_name() {
+            Self::FailedToAttest
+        } else if let Some(shard_str) = name.strip_prefix("primary_sliver_") {
+            if let Ok(shard) = shard_str.parse::<u16>() {
+                Self::PrimarySliver(shard)
+            } else {
+                Self::Custom(name.to_string())
+            }
```
sadhansood: I think walrus uses `"shard-<id>/primary-slivers"` names for primary slivers.

crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+        } else if let Some(shard_str) = name.strip_prefix("secondary_sliver_") {
+            if let Ok(shard) = shard_str.parse::<u16>() {
+                Self::SecondarySliver(shard)
+            } else {
+                Self::Custom(name.to_string())
+            }
+        } else if let Some(shard_str) = name.strip_prefix("shard_status_") {
```
sadhansood: Similarly, "shard-<id>/secondary-slivers"` for secondary nodes

crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+            if let Ok(shard) = shard_str.parse::<u16>() {
```
sadhansood: Also, if we need to parse the shard index we can reuse `id_from_column_family_name` if we needed to.

crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+            Self::AggregateBlobInfo => blob_info_cf_options(db_config),
+            Self::PerObjectBlobInfo => per_object_blob_info_cf_options(db_config),
+            Self::EventCursor => {
+                // Use the EventCursorTable::options which includes the proper merge operator.
+                let (_, opts) = event_cursor_table_options(db_config);
+                opts
+            }
+            Self::PrimarySliver(_) | Self::SecondarySliver(_) | Self::PendingRecoverSlivers(_) => {
+                // Sliver column families use the same shard options.
+                db_config.shard().to_options()
```
sadhansood, line 285: There is a different `db_config.pending_recover_slivers` in the file, why not use it?

crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+            }
+            Self::ShardStatus(_) | Self::ShardSyncProgress(_) => {
+                // These don't have special options defined.
+                let mut opts = Options::default();
+                opts.set_allow_mmap_reads(true);
+                opts
+            }
```
sadhansood, line 292: There is also different `db_config.shard_status()`

crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+            Self::EventStore | Self::InitState => {
```
sadhansood, line 293: oh, and `db_config.shard_sync_progress()`

crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+                // Event processor CFs don't have special merge operators.
+                let mut opts = Options::default();
+                opts.set_allow_mmap_reads(true);
+                opts
+            }
+            Self::Certified | Self::Attested | Self::Pending | Self::FailedToAttest => {
+                // Event blob writer CFs don't have special merge operators.
+                let mut opts = Options::default();
+                opts.set_allow_mmap_reads(true);
+                opts
```
sadhansood, line 303: `db_config.certified()/attested()/pending()/failed_to_attest()` exists too

crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+            }
+            Self::Custom(_) => {
+                // Default options for unknown column families.
+                let mut opts = Options::default();
+                opts.set_allow_mmap_reads(true);
+                opts
+            }
+        }
```
sadhansood, line 311: There is existing db options for bunch of other column families, i think we should use those for consistency

crates/walrus-service/src/node/db_options.rs
```diff
@@ -0,0 +1,402 @@
...
+                let (_, opts) = event_cursor_table_options(db_config);
+                opts
+            }
+            Self::PrimarySliver(_) | Self::SecondarySliver(_) | Self::PendingRecoverSlivers(_) => {
+                // Sliver column families use the same shard options.
+                db_config.shard().to_options()
+            }
+            Self::ShardStatus(_) | Self::ShardSyncProgress(_) => {
+                // These don't have special options defined.
+                let mut opts = Options::default();
```
sadhansood, line 289: Also, instead of `Options::default()` i wonder if we need to use `Options::from(&db_config.global())` everywhere

sadhansood (commented):
Thanks He, I have some comments. Also, should we move `db_column_family.rs` and `db_options.rs`  into `/node/storage`. I also feel like we should have both `db_tool.rs` and `dbtool_v2.rs` under `/node/tools` or something similar.

PR #2567, opened 2025-09-29
fix: cancel blob sync scheduled before node entering recovery mode

halfprice:
## Description

Fix two simtest issues.

First, in #2538 , we allow blob sync to be started when extending a blob. However, what can happen is that after restart, first event is an blob extension event, second event is start epoch change event for epoch X, but on chain, the epoch has move forward to X+2. In this case, the node will enter recovery mode after blob sync is initiated, and once the `event epoch` is set after epoch change start event, it will unblock the blob sync, which will sync blob to epoch X that is 2 epoch earlier than the latest epoch. The fix for this case is that we enter recovery mode, we should always cancel any ongoing blob sync. This is already the case for node entering recovery in the middle of operation. In this PR. we add that the node needs to cancel blob sync if they enter recovery mode at the beginning of a node restart.

The second issue is about sliver data consistency check. Say a node is processing epoch change to epoch 5, but on chain, the committee has moved to epoch 6 or later. In this case shard sync actually won't sync any blobs that is not certified at epoch 6. The blob can be certified at epoch 5, which cause data consistency check failure. In this case, we don't know if the blob will be uncertified in the future epoch, therefore, the data sliver consistency check monitors
committee epoch, and stop sliver data consistency check if the committee epoch is moved forward than the epoch of the data consistency check performed.

This PR also adds a script that I use commonly to run all simtest locally X number of times locally.

## Test plan

1000 repeated_node_crash
100 test_long_node_recovery
100 recovery_in_progress_with_node_restart
100 test_lagging_node_recovery
new unit test added

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-service/src/node.rs
```diff
@@ -1328,7 +1332,8 @@ impl StorageNode {
                 epoch_at_start,
                 first_new_event_epoch,
             );
-            self.inner.set_node_status(NodeStatus::RecoveryCatchUp)?;
+
+            self.enter_recovery_mode().await?;
```
halfprice, line 1336: Fix 1: we need to cancel any ongoing blob sync even if this the first event we use to check node lagging. This is because there may exist prior blob extension events that skipped node lag checking and started blob sync.

crates/walrus-service/src/node/consistency_check.rs
```diff
@@ -200,6 +200,21 @@ fn compose_blob_list_digest_and_check_sliver_data_existence(
...
+                    enable_sliver_data_existence_check = false;
+                    total_synced_scanned = 0;
+                    total_fully_stored = 0;
+                    tracing::info!(
+                        event_epoch = epoch,
+                        committee_epoch = node.current_committee_epoch(),
+                        "current committee epoch is not the same as the event epoch, skipping \
+                        sliver data existence check"
+                    );
+                }
```
halfprice, line 216: Fix 2: what could happen is that during data existence check, the epoch may move to the next epoch, which makes is_blob_certified to return based on next epoch status. In this case, it's possible that certified blob that will be expired may not stored locally, and therefore we skip sliver data existence check.

crates/walrus-service/src/node.rs
mlegner: Ah, makes sense. This should no longer be an issue if we persist the event epoch, correct?

crates/walrus-service/src/node.rs
```diff
@@ -2164,6 +2164,24 @@ impl StorageNode {
...
+    /// This function should only be called when the node is lagging behind.
+    async fn enter_recovery_mode(&self) -> anyhow::Result<()> {
+        self.inner.set_node_status(NodeStatus::RecoveryCatchUp)?;
+
+        // Now the node is entering recovery mode, we need to cancel all the blob syncs
+        // that are in progress, since the node is lagging behind, and we don't have
+        // any information about the shards that the node should own.
+        //
+        // The node now will try to only process blob info upon receiving a blob event
+        // and blob recovery will be triggered when the node is in the lasted epoch.
```
mlegner: Typo:
```suggestion
        // and blob recovery will be triggered when the node is in the latest epoch.
```

mlegner (approved):
Thanks for these fixes, @halfprice. These LGTM.

Could you create a separate PR for the simtest scripts?

crates/walrus-service/src/node.rs
halfprice: Done!

PR #2571, opened 2025-09-29
chore: remove unnecessary reference in EncodingConfigEnum

wbbradley:
## Description

Remove unnecessary references around EncodingConfigEnum causing unnecessary bloat in impls of EncodingFactory

## Test plan

No functional changes. CI Pipeline.

### DIFF
--- a/crates/walrus-core/src/encoding/config.rs
+++ b/crates/walrus-core/src/encoding/config.rs
@@ -363,7 +363,7 @@ impl EncodingConfig {
     /// [`EncodingConfigEnum`].
     pub fn get_for_type(&self, encoding_type: EncodingType) -> EncodingConfigEnum {
         match encoding_type {
-            EncodingType::RS2 => EncodingConfigEnum::ReedSolomon(self.reed_solomon),
+            EncodingType::RS2 => self.reed_solomon.into(),
             _ => panic!("unsupported encoding type: {:?}", encoding_type),
         }
     }
@@ -510,7 +510,7 @@ impl ReedSolomonEncodingConfig {
         &self,
         blob: &'a [u8],
     ) -> Result<BlobEncoder<'a>, DataTooLargeError> {
-        BlobEncoder::new(EncodingConfigEnum::ReedSolomon(*self), blob)
+        BlobEncoder::new((*self).into(), blob)
     }
 
     /// Returns a [`BlobDecoder`] for the given `blob_size`.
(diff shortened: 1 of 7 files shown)


crates/walrus-core/src/encoding/config.rs
```diff
@@ -507,10 +507,10 @@ impl ReedSolomonEncodingConfig {
 
     /// Returns a [`BlobEncoder`] for the given blob.
     pub fn get_blob_encoder<'a>(
-        &'a self,
+        &self,
         blob: &'a [u8],
     ) -> Result<BlobEncoder<'a>, DataTooLargeError> {
-        BlobEncoder::new(self.into(), blob)
+        BlobEncoder::new(EncodingConfigEnum::ReedSolomon(*self), blob)
```
mlegner: Question/nit: Why not `(*self).into()`?

crates/walrus-core/src/encoding/mapping.rs
```diff
@@ -140,15 +140,18 @@ mod tests {
...
+    };
 
     // Fixture
     fn sliver_pairs(num: u16) -> Vec<SliverPair> {
         let encoding_config = ReedSolomonEncodingConfig::new_for_test(1, 1, num);
         (0..num)
             .map(|n| {
                 SliverPair::new_empty(
-                    &(&encoding_config).into(),
+                    EncodingConfigEnum::ReedSolomon(encoding_config),
```
mlegner: As above, why not `encoding_config.into()`?

crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -599,7 +599,7 @@ mod tests {
             recovery_symbols,
             SliverIndex(0),
             symbol_size.try_into().unwrap(),
-            &(&config).into(),
+            EncodingConfigEnum::ReedSolomon(config),
```
mlegner: As above, why not `config.into()`?

mlegner (approved):
Thanks for the refactoring, @wbbradley. I don't have any objections, but just some historical context for future reference:

When using RaptorQ, the encoding config was much larger as it contained some pre-computed data to speed up the encoding. That's why it was not copyable and we used references everywhere. It's possible that we'll have a similar case again in the future if we use a different erasure code or a different library for Reed-Solomon.

crates/walrus-core/src/encoding/config.rs
wbbradley: Good call.

PR #2576, opened 2025-09-30
refactor: reduce server_address labels on client duration metrics

jpcsmith:
## Description

This removes the server_address and server_port labels from http_client_request_duration_* dependent on the request result, and should help to reduce the number of series being exported.

## Test plan

---

## Release notes

- [x] Storage node: The `http_client_request_duration_*` metrics now omit the `server_address` and `server_port` labels for fast, successful responses.
- [x] Aggregator:  The `http_client_request_duration_*` metrics now omit the `server_address` and `server_port` labels for fast, successful responses.
- [x] Publisher:  The `http_client_request_duration_*` metrics now omit the `server_address` and `server_port` labels for fast, successful responses.
- [ ] CLI:

### DIFF
--- a/crates/walrus-storage-node-client/src/client/middleware.rs
+++ b/crates/walrus-storage-node-client/src/client/middleware.rs
@@ -27,6 +27,7 @@ use tower::Service;
 use tracing::{Instrument as _, Span, field, instrument::Instrumented};
 use tracing_opentelemetry::OpenTelemetrySpanExt as _;
 use walrus_utils::{
+    backoff::SuccessOrFailure,
     http::{BodyVisitor, VisitBody, http_body},
     metrics::{self as metric_utils, OwnedGaugeGuard},
 };
@@ -38,6 +39,7 @@ pub(crate) struct UrlTemplate(pub &'static str); // Helps with incorrect lifetim
 
 const HTTP_RESPONSE_PART_HEADERS: &str = "headers";
 const HTTP_RESPONSE_PART_PAYLOAD: &str = "payload";
+const HTTP_SERVER_ADDRESS_DURATION_THRESHOLD: Duration = Duration::from_secs(1);
 
 metric_utils::define_metric_set! {
     #[namespace = "http_client"]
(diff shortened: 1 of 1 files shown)


crates/walrus-storage-node-client/src/client/middleware.rs
```diff
@@ -38,6 +39,7 @@ pub(crate) struct UrlTemplate(pub &'static str); // Helps with incorrect lifetim
 
 const HTTP_RESPONSE_PART_HEADERS: &str = "headers";
 const HTTP_RESPONSE_PART_PAYLOAD: &str = "payload";
+const HTTP_SERVER_ADDRESS_DURATION_THRESHOLD: Duration = Duration::from_secs(1);
```
halfprice, line 42: question: how is this threshold being picked? Feel like a const value for all requests may generate quite some false positive, if the data size being requested is large.

crates/walrus-storage-node-client/src/client/middleware.rs
```diff
@@ -86,46 +88,187 @@ impl HttpClientMetrics {
...
+    pub fn with_http_response_part<'b>(mut self, http_response_part: &'b str) -> Self
+    where
+        'b: 'a,
+    {
+        assert!(
+            N > HttpLabels::LENGTH,
+            "`N` must be more than `Labels::LENGTH`"
+        );
+        self.0[HttpLabels::LENGTH] = http_response_part;
+        self
```
halfprice, line 183: question: why do we want to bind the implementation with const N? I feel this may make extension of this array hard, especially, the self.0[HttpLabels::LENGTH] may be confusing because we first need to crate HttpLabelArray<LENGTH+1>. Can we simply remove N and make it a variable sized array?

crates/walrus-storage-node-client/src/client/middleware.rs
jpcsmith: Remark: It's arbitrary, but the default is that the server address is just never added. Therefore, this would only serve that, if for some reason we do want some export from a single node, any requests that are taking less than a second are excluded.

If you prefer this can be made variable.

crates/walrus-storage-node-client/src/client/middleware.rs
jpcsmith: It's to avoid heap allocations both for strings and the resulting slice that's needed to get the metric from the prometheus crate. 

Each time we want to log a metric, we need a `&[&str]`. We don't want to use a `Vec<_>` as that would require an allocation each time we want to write the metrics. Furthermore, since it's not possible to store all the labels in a slice (not everything is a `&'static str`, some are `String`, and some things like the StatusCode and Method are used at multiple places which would be a pain to to have to parse them back from a `&str` each time), the usual approach would be to construct the slice `&[&string1, &string2, ...]` each time. The prior code had this logic in a function, this moves it into a struct. 

An alternative is to remove the const parameter N and always just make an array of `HttpLabels::LENGTH + 1`, then have an `as_ref_without_http_response_part()` method which returns `array[..array.len()-1]` that excludes the last element.

```rust
struct HttpLabels {
    http_request_method: Method,
    server_address: String,
    server_port: String,
    url_scheme: Cow<'static, str>,
    url_template: &'static str,
    network_protocol_version: Cow<'static, str>,
    error_type: Option<ErrorType>,
    http_response_status_code: Option<StatusCode>,
}
```

halfprice (approved):
LGTM ! Thanks @jpcsmith !

PR #2577, opened 2025-09-30
test: create a simtest script that use seed search to run all simtests

halfprice:
## Description

This is useful to run all large number of simtests in parallel.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


wbbradley:
Why do we need all the process group gymnastics? Do the child processes kill the parent process group by default?

PR #2578, opened 2025-09-30
fix: use blob ID as the ETag for quilt patches.

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon/routes.rs
+++ b/crates/walrus-service/src/client/daemon/routes.rs
@@ -502,13 +502,15 @@ pub(super) async fn get_patch_by_quilt_patch_id<T: WalrusReadClient>(
     let quilt_patch_id_str = quilt_patch_id.to_string();
     tracing::debug!("starting to read quilt patch: {}", quilt_patch_id_str);
 
+    let quilt_id = quilt_patch_id.quilt_id;
+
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -502,13 +502,15 @@ pub(super) async fn get_patch_by_quilt_patch_id<T: WalrusReadClient>(
...
+    let quilt_id = quilt_patch_id.quilt_id;
+
     match client.get_blobs_by_quilt_patch_ids(&[quilt_patch_id]).await {
         Ok(mut blobs) => {
             if let Some(blob) = blobs.pop() {
                 build_quilt_patch_response(
                     blob,
                     &request_headers,
-                    &quilt_patch_id_str,
+                    &quilt_id.to_string(),
```
wbbradley, line 513: Do we support If-Not-Match somewhere? Wouldn't that also need to be updated?

wbbradley: Ah, nevermind, I see the Cache-control setting max-age.

PR #2579, opened 2025-09-30
feat: Improve quilt decoder, to accept both borrowed and owned sliver data

liquid-helium:
## Description

Use Cow to hold the data for a quilt decoder, this will make it easier to use a decoder as a 'cache'.

Contributes to WAL-685

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-core/src/encoding/quilt_encoding.rs
+++ b/crates/walrus-core/src/encoding/quilt_encoding.rs
@@ -2569,4 +2569,132 @@ mod tests {
         let config = ReedSolomonEncodingConfig::new(NonZeroU16::new(7).unwrap());
         let _ = QuiltV1::new_from_quilt_blob(Vec::new(), &EncodingConfigEnum::ReedSolomon(&config));
     }
+
+    #[test]
+    fn test_column_size_mismatch_errors() {
+        use core::num::NonZeroU16;
+
+        use crate::encoding::errors::QuiltError;
+
+        // Create slivers with different sizes.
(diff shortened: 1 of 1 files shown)


crates/walrus-core/src/encoding/quilt_encoding.rs
```diff
@@ -1749,6 +1749,30 @@ impl<'a> QuiltDecoderV1<'a> {
...
+
+    /// Adds owned slivers to the decoder.
+    pub fn add_owned_slivers(&mut self, slivers: impl IntoIterator<Item = SliverData<Secondary>>) {
+        for sliver in slivers {
+            self.column_size
+                .get_or_insert_with(|| sliver.symbols.data().len());
+            let index = sliver.index;
+            self.slivers.insert(index, Cow::Owned(sliver));
+        }
+    }
```
mlegner, line 1813: Question/suggestion: Can we document that we require all slivers to be of equal size and that this is not checked? Alternatively, we could check all sliver sizes.

mlegner (approved):
Thanks, @liquid-helium!

crates/walrus-core/src/encoding/quilt_encoding.rs
liquid-helium: Good idea. I added check to make sure all sizes are equal.

PR #2582, opened 2025-10-01
feat: add RawKeyDBMap for custom key serialization in typed-store

liquid-helium:
## Description

Introduced RawKeyDBMap, a wrapper around typed-store's DBMap that enables custom key
serialization/deserialization. 
This is particularly useful for scenarios requiring custom key encoding, such as preserving alphanumeric order or implementing custom sorting schemes.

Close WAL-1053
## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1134,6 +1134,10 @@ impl StorageNode {
             })
             .await
         {
+            self.inner
+                .metrics
+                .started_processing_event(stream_element.checkpoint_event_position);
+
             let event_label: &'static str = stream_element.element.label();
             let monitor = task_monitors.get_or_insert_with_task_name(&event_label, || {
                 format!("process_event {event_label}")
@@ -1168,6 +1172,10 @@ impl StorageNode {
             };
 
             TaskMonitor::instrument(&monitor, task).await?;
+
+            self.inner
+                .metrics
+                .completed_processing_event(stream_element.checkpoint_event_position);
         }
 
         bail!("event stream for blob events stopped")
--- a/crates/walrus-service/src/node/metrics.rs
+++ b/crates/walrus-service/src/node/metrics.rs
@@ -24,7 +24,7 @@ use walrus_sui::types::{
 
 use crate::{
     common::telemetry::{CurrentEpochMetric, CurrentEpochStateMetric},
-    event::events::EventStreamElement,
+    event::events::{CheckpointEventPosition, EventStreamElement},
 };
 
 pub(crate) const STATUS_FAILURE: &str = "failure";
@@ -37,6 +37,7 @@ pub(crate) const STATUS_QUEUED: &str = "queued";
 pub(crate) const STATUS_PENDING: &str = "pending";
 pub(crate) const STATUS_PERSISTED: &str = "persisted";
 pub(crate) const STATUS_IN_PROGRESS: &str = "in-progress";
+pub(crate) const STATUS_COMPLETED: &str = "completed";
 pub(crate) const STATUS_HIGHEST_FINISHED: &str = "highest_finished";
 
 type U64GaugeVec = GenericGaugeVec<AtomicU64>;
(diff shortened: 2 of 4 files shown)


crates/typed-store/src/raw_key_db_map.rs
```diff
@@ -0,0 +1,498 @@
...
+            }
+            None => Ok(None),
+        }
+    }
+
+    /// Reads a range of key-value pairs within a transaction using efficient range bounds.
+    /// Returns key-value pairs in the range [begin, end).
+    ///
+    /// This is more efficient than prefix-based iteration for range queries.
+    pub fn read_range(
```
halfprice: question: DBMap already has iterator support https://github.com/MystenLabs/walrus/blob/c4db4518ac8bd8876446a290d23be1ffe2dfcfb5/crates/typed-store/src/rocks.rs#L2189-L2218, can we use it here?

crates/typed-store/src/raw_key_db_map.rs
```diff
@@ -0,0 +1,498 @@
...
+                    }
+                }
+            }
+
+            if results.len() >= row_limit {
+                break;
+            }
+        }
+
+        Ok(results)
```
halfprice, line 226: suggestion: can we return an iterator here so that we don't need to hold all the results in memory before returning?

crates/typed-store/src/raw_key_db_map.rs
```diff
@@ -0,0 +1,498 @@
...
+/// This struct provides a minimal interface for using DBMap with raw
+/// key serialization/deserialization while keeping the default BCS serialization for values.
+pub struct RawKeyDBMap<K, V>
+where
+    K: KeyCodec,
+    V: Serialize + for<'de> Deserialize<'de> + Clone,
+{
+    inner: DBMap<Vec<u8>, V>,
+    _phantom: PhantomData<K>,
+}
```
halfprice, line 37: question: is the future plan to implement the full [Map trait](https://github.com/MystenLabs/walrus/blob/ea3735eb08f63abc44f3e4a6e8d8bfc34f08c3fe/crates/typed-store/src/traits.rs#L11) in RawKeyDBMap as well? Right now, although inner is DBMap, the implementation only uses inner to access the underlying RocksDB column family, and directly operate on that. If the usage remain like this, I wonder if the `inner `can just be the column family handle instead of a DBMap. It doesn't really need to use DBMap.

Also, I feel it makes things a bit more unintuitive to "untype" the key on top of DBMap (whose original design is just type the raw byte key and value, and help with serialization and deserialization).

crates/typed-store/src/raw_key_db_map.rs
```diff
@@ -0,0 +1,498 @@
...
+        Ok(RawKeyDBMap {
+            inner,
+            _phantom: PhantomData,
+        })
+    }
+
+    /// Puts a key-value pair within a transaction.
+    pub fn put_cf_with_txn(
+        &self,
+        txn: &rocksdb::Transaction<'_, OptimisticTransactionDB>,
```
halfprice, line 101: question: where will the txn be created? Inside or outside of RawKeyDBMap?

halfprice (commented):
Thanks @liquid-helium for the PR! The implementation makes sense, but I wonder if we can untight it with DBMap, because RawKeyDBMap doesn't really need to use DBMap IIUC.

crates/typed-store/src/raw_key_db_map.rs
liquid-helium: The short answer is no, since these are not using transactions.

crates/typed-store/src/raw_key_db_map.rs
liquid-helium: That's a good question. The use case here is each time we can only handle a small number of the items we read, like async tasks. So it is efficient to read it once, release the iterator, do the tasks, and come back. WDYT?

crates/typed-store/src/raw_key_db_map.rs
liquid-helium: The transaction will be created outside the RawKeyDBMap, and can include different column families.

crates/typed-store/src/raw_key_db_map.rs
halfprice: Yeah, we can re-evaluate if this ended up concuming a lot memory.

crates/typed-store/src/raw_key_db_map.rs
halfprice: Ah, I missed the transaction part. Maybe mention this in the function name like other functions?

crates/typed-store/src/raw_key_db_map.rs
```diff
@@ -0,0 +1,564 @@
...
+impl<K, V> ColumnFamilyHandle<K, V>
+where
+    K: KeyCodec,
+    V: Serialize + for<'de> Deserialize<'de> + Clone,
+{
+    /// Creates a new ColumnFamilyHandle with direct RocksDB and ColumnFamily access.
+    pub fn new(
+        db: Arc<RocksDB>,
+        cf_name: &str,
+        is_deprecated: bool,
```
halfprice, line 61: Nit: do we need this? May make `clone` a bit tricky.

halfprice (approved):
Thanks for the revision! LGTM!

crates/typed-store/src/raw_key_db_map.rs
liquid-helium: are you referring to the background db-metrics-reporter? If so, maybe let's keep it for now, it is useful to understand some basic metrics with transactions, since this is the first time we have this type of workload?

crates/typed-store/src/raw_key_db_map.rs
liquid-helium: Done.

PR #2585, opened 2025-10-01
feat: export the Sui checkpoint of the last processed Walrus event

jpcsmith:
## Description

Exports the Sui checkpoint from which the last processed walrus event was taken. This allows us to

- determine storage node lag independently of other storage nodes, and without a global view of the network (currently we measure this by how far behind a storage node is relative to other nodes; and
- determine the delay between a Sui checkpoint and a node actually processing the related walrus events, which is an input to write latency.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1168,6 +1168,10 @@ impl StorageNode {
             };
 
             TaskMonitor::instrument(&monitor, task).await?;
+
+            self.inner
+                .metrics
+                .set_last_processed_event_position(stream_element.checkpoint_event_position);
         }
 
         bail!("event stream for blob events stopped")
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1134,6 +1134,10 @@ impl StorageNode {
             })
             .await
         {
+            self.inner
+                .metrics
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -1168,6 +1168,10 @@ impl StorageNode {
             };
 
             TaskMonitor::instrument(&monitor, task).await?;
+
+            self.inner
+                .metrics
+                .set_last_processed_event_position(stream_element.checkpoint_event_position);
```
halfprice: Suggestion: I suspect that it'll be more useful to put this at the beginning of the loop, so that we know which checkpoint the current event is stuck at. Knowing the last successfully processed event may not necessarily lead us to the next event easily.

halfprice (approved):
Thanks @jpcsmith for adding this!

If the motivation of adding this is from trying to find the stuck event, you can also use db-tool to query the event table to directly go to that event. Just FYI.

sadhansood (approved):

crates/walrus-service/src/node.rs
jpcsmith: Renamed the metric and track both the in-progress and last processed.

PR #2586, opened 2025-10-01
fix(client): clean up some CLI options and publisher query parameters

mlegner:
## Description

- Remove the warning if deletable/permanent is not specified.
- Hide `--deletable` CLI option.
- Mark `deletable` query parameter for publisher as deprecated.
- Clarify some docstrings.
- Minor refactoring in `crates/walrus-service/src/client/cli/runner.rs`.

## Test plan

Inspect generated OpenAPI specs and `walrus store --help` output.


crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -1632,6 +1598,53 @@ impl ClientCommandRunner {
     }
 }
 
+pub struct StoreOptions {
```
wbbradley: Looks like this doesn't need to be `pub`?

mlegner: Done in 0073b0725c36822f58471d8ddc9371ce64711781.

PR #2592, opened 2025-10-02
feat: add CLI option and aggregator query parameter for strict integrity checks

mlegner:
## Description

In preparation for making the default integrity check more performant while slightly loosening the resulting guarantee, this PR adds a CLI option `--strict-integrity-check` and an aggregator query parameter `strict-integrity-check` to enable the current strict integrity check.

It also extends the related documentation.

## Test plan

TBD.

---

## Release notes

- [ ] Aggregator:
- [ ] CLI:


docs/book/design/encoding.md
```diff
@@ -37,18 +45,74 @@ Each blob is also associated with some metadata including a **blob ID** to allow
...
-  to check that the blob matches. This prevents a malformed blob (incorrectly erasure coded) from
-  ever being read as a valid blob at any correct recipient.
+- Inconsistencies of the last type (incorrect blob-ID computation) are detected immediately when the
+  client attempts to upload metadata to storage nodes. Such blobs are never even certified.
 
-- A set of slivers equal to the reconstruction threshold belonging to a blob ID that are either
-  inconsistent or lead to the reconstruction of a different ID represent an incorrect encoding. This
-  happens only if the user that encoded the blob was faulty or malicious and encoded it incorrectly.
+- Inconsistencies of the second type (incorrect sliver-hash computation) are detected either before
+  or immediately after certification.
```
karlwuest: Nit: This is only true if the node that should store the sliver is correct. I don't think we need to get into the details of this, but maybe softening the claim a bit is better:

```suggestion
- Inconsistencies of the second type (incorrect sliver-hash computation) are generally detected
  either before or shortly after certification.
```

docs/book/design/encoding.md
```diff
@@ -37,18 +45,74 @@ Each blob is also associated with some metadata including a **blob ID** to allow
...
+The encoding used by Walrus has the property that *the first 334 primary slivers contain the
+(potentially padded) unencoded data*. This means that the (cryptographic) hashes of these slivers
+together with the blob length uniquely determine the content of the blob. Therefore, by recomputing
+the sliver hashes of the first 334 slivers and checking them against the metadata, the client can
+verify that the decoded data is correct. If this check succeeds, the data is provided to the user,
+otherwise an error is returned. This check provides the following guarantee:
+
+```admonish tip title="Data integrity property"
+Any correct client attempting to read a blob will either read one specific value authenticated by
+the writer or nothing.
```
karlwuest: Nit: I would rephrase this slightly. "Reading nothing" is clear to us, but it may not be so clear to a layperson, so it might be better to say the reader returns an error. And I would say "the" instead of "one" to make it clearer that it's the same for all clients.
```suggestion
Any correct client attempting to read a blob will either read the specific value authenticated by
the writer or return an error.
```

docs/book/design/encoding.md
```diff
@@ -37,18 +45,74 @@ Each blob is also associated with some metadata including a **blob ID** to allow
...
+```
+
+### Strict integrity check
+
+If the blob was encoded correctly, any set of 334 primary slivers decode to the same data. However,
+if the writer of a blob encoded the data incorrectly, sets of slivers may decode to different data.
```
karlwuest: Nit:
```suggestion
if the writer of a blob encoded the data incorrectly, different sets of slivers may decode to different data.
```

docs/book/design/encoding.md
```diff
@@ -37,18 +45,74 @@ Each blob is also associated with some metadata including a **blob ID** to allow
...
+Even in that case, there is only one correct value that can be read by an honest client (guarantee
+stated above), but some read attempts may result in a failure. Furthermore, the blob may be detected
```
karlwuest: Nit:
```suggestion
stated above), but some read attempts may result in a failure while others succeed. Furthermore, the blob may be detected
```

docs/book/design/encoding.md
```diff
@@ -37,18 +45,74 @@ Each blob is also associated with some metadata including a **blob ID** to allow
...
+For this reason, Walrus optionally offers a *strict integrity check*: After decoding the blob, the
+client fully *re-encodes* it and recomputes all hashes and the blob ID. Only if that computation
+results in the blob ID they are trying to read is the read considered successful.
+
+This process *ensures that the original writer encoded the blob correctly*. This implies the
+following guarantee:
+
+```admonish tip title="Strict integrity property"
+Any correct client attempting to read a blob during its lifetime will always succeed and read the
+same data.
```
karlwuest: Suggestion: The clients may still fail, so we should say that either all read the same or all fail.
```suggestion
All correct clients attempting to read a blob during its lifetime will either always succeed and
read the same data or all will return an error.
```

mlegner: Good point. My intention was to describe the guarantee after a read with the strict integrity check succeeded. Will reword the sentence above.

mlegner: Done in 724cb4a024665a5f32910f6b30d84c2b6d26e14c.

docs/book/design/encoding.md
```diff
@@ -37,18 +45,81 @@ Each blob is also associated with some metadata including a **blob ID** to allow
...
+1. Incorrectly compute slivers.
+1. Incorrectly compute the sliver hashes based on the slivers.
+1. Incorrectly compute the blob ID based on the sliver hashes and other metadata.
+
+Walrus has several mechanisms to detect each of these:
 
-  This process involves re-encoding the blob using the erasure code, and deriving the blob ID again
-  to check that the blob matches. This prevents a malformed blob (incorrectly erasure coded) from
-  ever being read as a valid blob at any correct recipient.
+- Inconsistencies of the last type (incorrect blob-ID computation) are detected immediately when the
```
liquid-helium: nit: can we be explicit about where the inconsistency is detected, e.g., either by the client code, or by the storage nodes?

docs/book/design/encoding.md
```diff
@@ -37,18 +45,81 @@ Each blob is also associated with some metadata including a **blob ID** to allow
...
+  client attempts to upload metadata to storage nodes. Such blobs are never even certified.
 
-- A set of slivers equal to the reconstruction threshold belonging to a blob ID that are either
-  inconsistent or lead to the reconstruction of a different ID represent an incorrect encoding. This
-  happens only if the user that encoded the blob was faulty or malicious and encoded it incorrectly.
+- Inconsistencies of the second type (incorrect sliver-hash computation) are generally detected
+  either before or shortly after certification.
 
-  Walrus can extract one symbol per sliver to form an inconsistency proof. Storage nodes can delete
+  In the latter case, storage nodes can extract one symbol per sliver to form an *inconsistency
```
liquid-helium: >In the latter case, storage nodes can extract one symbol per sliver to form an *inconsistency

Is this guaranteed to happen, or is this a side effect of recovery?

mlegner: Essentially there are only two cases:
1. Recovery succeeds.
2. The node can create an inconsistency proof.

So we are guaranteed to either recover or create the proof. I'll change the statement there slightly.

mlegner: Actually, strike that. This is actually guaranteed to happen if the corresponding slivers are handled by honest storage nodes as the nodes attempt recovery and recover a sliver from authentic symbols that is inconsistent with its hash.

mlegner: Added some clarification in 2fe8812e8877cf66b05fc9d424e12d4718af9d13.

PR #2599, opened 2025-10-03
feat(node): delete metadata and slivers for expired/deleted blobs

mlegner:
## Description

This PR adds a cleanup job deleting blob data for expired blobs at the beginning of an epoch. The new process is gated through a configuration parameter and is disabled by default.

Depends on #2475 as data deletion is implemented using transactions. 

Open tasks:
- [x] Add metrics.
- [x] Temporarily disable DB compactions.

## Test plan

- [x] Add a simtest for data deletion.
- [x] Enable deletion in a local testbed.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -1626,7 +1626,8 @@ impl StorageNode {
         if !garbage_collection_config.enable_blob_info_cleanup {
             if garbage_collection_config.enable_data_deletion {
                 tracing::warn!(
-                    "data deletion is enabled, but requires blob info cleanup to be enabled"
+                    "data deletion is enabled, but requires blob info cleanup to be enabled; \
+                    skipping data deletion",
                 );
             } else {
                 tracing::info!("garbage collection is disabled, skipping cleanup");
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -619,7 +619,8 @@ impl Storage {
             .await
     }
 
-    /// Deletes the metadata and slivers for blobs that are expired in the given epoch.
+    /// Deletes the aggregate blob info, metadata, and slivers for blobs that are expired in the
+    /// `current_epoch`.
     #[tracing::instrument(skip_all, fields(walrus.epoch = %current_epoch))]
     pub(crate) async fn delete_expired_blob_data(
         &self,
@@ -677,7 +678,7 @@ impl Storage {
 
             // At this point we know that the blob is no longer registered, and we can attempt to
             // delete the related data.
-            if self
+            match self
                 .attempt_to_delete_blob_data_inner(
                     &optimistic_handle,
                     &blob_id,
@@ -685,9 +686,19 @@ impl Storage {
                     &shards,
                     node_metrics,
                 )
-                .await?
+                .await
             {
-                cleaned_up_blob_id_count += 1;
+                Ok(true) => {
+                    cleaned_up_blob_id_count += 1;
+                }
+                Ok(false) => (),
+                Err(error) => {
+                    tracing::error!(
+                        ?error,
+                        %blob_id,
+                        "encountered an error while attempting to delete blob data"
+                    );
+                }
             }
         }
 


crates/walrus-service/src/node.rs
```diff
@@ -1617,6 +1609,31 @@ impl StorageNode {
         Ok(())
     }
 
+    /// Performs database cleanup operations including blob info cleanup and data deletion.
+    #[tracing::instrument(skip_all)]
+    async fn perform_db_cleanup(&self, epoch: Epoch) -> anyhow::Result<()> {
+        // TODO(mlegner): Disable DB compactions during cleanup.
```
liquid-helium: Question: Do we perform db cleanup only when the node is in Active state?

crates/walrus-service/src/node/storage.rs
```diff
@@ -595,6 +596,123 @@ impl Storage {
...
+            // delete the related data.
+            self.attempt_to_delete_blob_data(&optimistic_handle, &blob_id, current_epoch, &shards)
+                .await?;
+        }
+
+        Ok(())
+    }
+
+    async fn attempt_to_delete_blob_data(
+        &self,
```
liquid-helium: Question: when a blob expires, are all the recovery tasks for this blob cancelled, (if any), so we don't need to worry about any race condition here?

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -304,6 +304,45 @@ impl BlobInfoTable {
...
+            .cf()
+            .expect("we know that this CF exists")
+    }
+
+    pub fn get_for_update_in_transaction(
+        &self,
+        transaction: &Transaction<'_, rocksdb::OptimisticTransactionDB>,
+        blob_id: &BlobId,
+    ) -> Result<Option<BlobInfo>, rocksdb::Error> {
+        // Question(mlegner): Should `exclusive` parameter be true or false?
```
liquid-helium: My understanding is that with optimistic transactions, this setting doesn't actually matter either way—but worth double-checking!

crates/walrus-service/src/node.rs
```diff
@@ -1617,6 +1609,31 @@ impl StorageNode {
...
+            .enable_blob_info_cleanup
+        {
+            self.inner
+                .storage
+                .process_expired_blob_objects(epoch)
+                .await?;
+        }
+
+        if self.inner.garbage_collection_config.enable_data_deletion {
+            self.inner.storage.delete_expired_blob_data(epoch).await?;
```
halfprice: question: if enable_blob_info_cleanup is not enabled, do we still need to delete blob data (anything to delete)?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -593,6 +593,26 @@ impl ShardStorage {
...
+    /// Deletes the sliver pair for the given [`BlobId`] within a DB transaction.
+    pub(crate) fn delete_sliver_pair_in_transaction(
+        &self,
+        transaction: &Transaction<'_, rocksdb::OptimisticTransactionDB>,
+        blob_id: &BlobId,
+    ) -> Result<(), rocksdb::Error> {
+        let column_families = [
+            self.primary_slivers
+                .cf()
+                .expect("primary slivers CF must always exist"),
```
halfprice: remark: this may not be true if the gc job running longer than a epoch, where a shard can move out and then being deleted. I think we can simply filter out none exist cf here.

halfprice: The main issue is that after we delete the cf, ShardStorage could still exist due to some of the callsite taking an Arc<ShardStorage>, even though we have remove the shard from ShardStorage.

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: +1. I agree that using both do not affect correctness, since if modification is needed, aggregate_blob_info table will also need to be updated, which essentially serialize concurrent transactions. However, I suspect that setting to true is better to have less aborts, and we probably don't expect much concurrent updates on the same key very often.

I don't know if there is a performance difference in different settings though.

halfprice (commented):
Thanks @mlegner for implementing this important feature 🚀🚀🚀! I took a quick look and only have some minor comments. Will take closer look in the next round!

crates/walrus-service/src/node/storage/blob_info.rs
liquid-helium: > My understanding is that with optimistic transactions, this setting doesn't actually matter either way—but worth double-checking!

Expand on this: since this is a `optimistic` transaction, and the `exclusive` setting determines how lock is acquired in a `pessimistic` transaction, hence it should be a no-op here.

crates/walrus-service/src/node.rs
mlegner: I think the first part of the cleanup (update blob info) makes sense in any state. This just updates the counts of deletable blob objects.

The actual garbage collection probably only makes sense in the active state. I'll add some checks there.

crates/walrus-service/src/node.rs
mlegner: This is a very good question. In principle, the two are somewhat complementary. However, I just realized that when enabling data deletion and not blob-info cleanup, but later enabling the blob-info cleanup as well, we could get into an inconsistent state. I'll make some changes there and a check that data deletion requires the blob-info cleanup.

crates/walrus-service/src/node/storage/shard.rs
mlegner: Hmm... Maybe one option would be to simply stop and re-start the cleanup at every epoch change. WDYT?

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: makes great sense!

crates/walrus-service/src/node/storage.rs
```diff
@@ -595,6 +596,123 @@ impl Storage {
...
+        current_epoch: Epoch,
+    ) -> anyhow::Result<()> {
+        let Some(optimistic_handle) = self.database.as_optimistic() else {
+            tracing::warn!("data deletion is only possible when the DB supports transactions");
+            return Ok(());
+        };
+
+        tracing::info!(epoch = %current_epoch, "deleting expired blob data");
+        let shards = futures::future::try_join_all(self.shards.read().await.values().map(
```
sadhansood: Would this not deadlock here because we already hold the write lock in node.rs when we call `perform_db_cleanup`

crates/walrus-service/src/node/storage.rs
```diff
@@ -595,6 +596,123 @@ impl Storage {
...
+        &self,
+        optimistic_handle: &OptimisticHandle<'_>,
+        blob_id: &BlobId,
+        current_epoch: Epoch,
+        shards: &[Arc<ShardStorage>],
+    ) -> Result<(), rocksdb::Error> {
+        let transaction = optimistic_handle.transaction();
+        // Question(mlegner): Should `conclusive` be true or false?
+        // Question(mlegner): Does this even work in conjunction with merge operators?
+        let Some(blob_info) = self
```
sadhansood: I think the answer to this one is yes.

crates/walrus-service/src/node/storage/shard.rs
halfprice: Yeah, I think that may work. I think right now, the purge happens synchronously inside EpochChangeStart. Let's see how is simtest once moving these task to the background.

crates/walrus-service/src/node/storage.rs
mlegner: Just for context, the reason I added this question is a slightly unclear sentence in the [RocksDB docs](https://docs.rs/rocksdb/latest/rocksdb/struct.Transaction.html#method.get_for_update_cf_opt):

> Currently, this function will return an error of the [MergeInProgress](https://docs.rs/rocksdb/latest/rocksdb/enum.ErrorKind.html#variant.MergeInProgress) if the most recent write to the queried key in this batch is a Merge.

crates/walrus-service/src/node.rs
mlegner: Done in 16f2ca2dacee0f3019c2df17042627e6c7d9b0cf.

crates/walrus-service/src/node.rs
mlegner: Done in 16f2ca2dacee0f3019c2df17042627e6c7d9b0cf.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Replaced the question with a comment explaining this in eaeaf82004080642ea6f378857f1169f87bdce5f.

crates/walrus-service/src/node/storage.rs
mlegner: Great catch. I now moved the invocation of the `perform_db_cleanup` to the end of the `process_epoch_change_start_event` function in 16f2ca2dacee0f3019c2df17042627e6c7d9b0cf. The write lock is transferred through some steps to a background task that performs the shard changes. After that completes, we can read the shards here.

Currently, this is problematic in the sense that we block here but we'll move the DB cleanup to the background anyway.

crates/walrus-service/src/node/storage/shard.rs
mlegner: In any case, I've changed the `expect`s here to a `?` in 259b5da5e1c28b8acce75f8030801b09100a4098.

crates/walrus-service/src/node/storage.rs
mlegner: Great point. After thinking about this for some time, I realized that we have to be very careful to avoid creating "storage leaks", i.e., metadata and slivers where we don't have a corresponding entry in the blob-info table. I see a few possible ways to do that, but they may come with performance penalties:  

1. Always use a DB transaction to store slivers and metadata, in which we check that the blob is registered using a `get_for_update`.
2. Always include some update to the blob-info table in the batch where we write slivers or metadata. There we need to be careful that the update works even for untracked blobs (e.g., creating a kind of placeholder entry).

WDYT? Do you see a different/better way?

mlegner:
@sadhansood What are your plans for landing #2475? That would be pretty important for testing this PR properly.

crates/walrus-service/src/node/storage.rs
liquid-helium: Both sounds feasible, and the penalty `seems` to be similar, in the sense that one requires a read, and the other requires a write to the blob-info-table.
Option 2 seems to be more aligned with other existing operations (except for the new deletion workflow).

>Always use a DB transaction to store slivers and metadata, in which we check that the blob is registered using a get_for_update.

This approach is organic, and I hope the penalty is low enough if we don't have many conflicts.

>Always include some update to the blob-info table in the batch where we write slivers or metadata. There we need to be careful that the update works even for untracked blobs (e.g., creating a kind of placeholder entry).

Do you mean we introduce a new 'invalid' state for the blob, so that a later cleanup can delete it? Will this be done via a merge operator?

mlegner: > > Always include some update to the blob-info table in the batch where we write slivers or metadata. There we need to be careful that the update works even for untracked blobs (e.g., creating a kind of placeholder entry).
> 
> Do you mean we introduce a new 'invalid' state for the blob, so that a later cleanup can delete it? Will this be done via a merge operator?

Yes, something like a `Touch` merge operand, which is a noop on a valid entry and creates a `ToBeCleanedUp` value if no entry exists. That value would then be handled the same as a non-existing entry for all other operands but treated as "expired" during the garbage collection.

mlegner: I created a follow-up issue for this (WAL-1066). I may address it as part of this PR or in a follow-up.

crates/walrus-service/src/node.rs
```diff
@@ -1623,6 +1611,50 @@ impl StorageNode {
...
+
+    /// Performs database cleanup operations including blob info cleanup and data deletion.
+    #[tracing::instrument(skip_all)]
+    async fn perform_db_cleanup(&self, epoch: Epoch) -> anyhow::Result<()> {
+        let garbage_collection_config = self.inner.garbage_collection_config;
+
+        if !garbage_collection_config.enable_blob_info_cleanup {
+            if garbage_collection_config.enable_data_deletion {
+                tracing::warn!(
+                    "data deletion is enabled, but requires blob info cleanup to be enabled"
```
halfprice: nit: maybe add that `data deletion is not performed`

crates/walrus-service/src/node.rs
```diff
@@ -1623,6 +1611,50 @@ impl StorageNode {
...
+                );
+            } else {
+                tracing::info!("garbage collection is disabled, skipping cleanup");
+            }
+            return Ok(());
+        }
+
+        // Disable DB compactions during cleanup to improve performance.
+        self.inner.storage.enable_auto_compactions(false)?;
```
halfprice: suggestion: maybe wrap this disable/enable functionality in a guard, so that early return will re-enable DB compaction as well?

crates/walrus-service/src/node/storage.rs
```diff
@@ -592,18 +612,189 @@ impl Storage {
...
-        tracing::info!(epoch = %current_epoch, "processing expired blob objects");
-
         self.blob_info
-            .process_expired_blob_objects(current_epoch)
-            .await?;
+            .process_expired_blob_objects(current_epoch, node_metrics)
+            .await
+    }
+
+    /// Deletes the metadata and slivers for blobs that are expired in the given epoch.
```
halfprice: nit: can we mention `blob info` table in the comment as well to be more explicit?

crates/walrus-service/src/node/storage.rs
```diff
@@ -592,18 +612,189 @@ impl Storage {
...
+            // delete the related data.
+            if self
+                .attempt_to_delete_blob_data_inner(
+                    &optimistic_handle,
+                    &blob_id,
+                    current_epoch,
+                    &shards,
+                    node_metrics,
+                )
+                .await?
```
halfprice: question: should we continue when failure happens? Right now, I think this will terminate the server, which also may not be the intended behavior.

halfprice (approved):
Thanks @mlegner and this looks great! I only have a few nits/small comments.

crates/walrus-service/src/node.rs
mlegner: Ah, that's a great catch. 👍

I've added such a guard in ae251f64e9180ebbc512acd0e1622f2469cb9f9b.

crates/walrus-service/src/node/storage.rs
mlegner: Good point. I changed this to log an error instead in 0fd04520fbe2122baff4d427c2d05960647fbd46.

halfprice (approved):
Thanks a lot @mlegner ! The PR and the newly added tests look great 💯!

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -282,9 +283,14 @@ impl BackgroundEventProcessor {
             // We use the event's epoch for this check (as opposed to the current epoch) as
             // subsequent certify or delete events may update the `blob_info`; so we cannot remove
             // it even if it is no longer valid in the *current* epoch
-            if !blob_info.is_registered(event.epoch) {
+            if !blob_info.is_registered(current_epoch)
```
halfprice, line 286: Sorry that I missed this while doing the review. Here is a change of behavior: we were using the event epoch, but now changed to using the latest committee epoch.

mlegner: Oh... Don't know why I changed this. The comment above it even explains why the previous version is correct... 🤦

Will fix this as part of #2640.

PR #2603, opened 2025-10-06
test: add latency configs for the performance-main-baseline experimen…

jpcsmith:
…t stack

## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


halfprice (approved):
Amazing! Thanks a lot @jpcsmith!

I didn't read the full PR (nor I could reasoning about it), but I look forward to use it and providing some feedback! 

I have a few questions in general about the framework:
- can we trigger the performance test via a workflow, so that starting test can be async?
- Say if I have a PR and I want to test the performance impact, what's the best workflow? If I orchestrate the test locally, how to read the metrics?
- There is/will be a nightly run of this, is that correct?

PR #2604, opened 2025-10-06
feat: added AsyncTask, AsyncTaskStore, AsyncTaskSorter traits/types.

liquid-helium:
## Description

Traits and basic types for a asynchronous event processing framework for Walrus indexer.

Contributes to WAL-1058

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+    UpToDate,
+}
+
+/// In-memory buffer for async tasks.
+/// It maintains two queues:
+/// - catchup_queue: for oldest tasks, loaded from disk during catch-up.
+/// - latest_queue: for latest tasks, where the oldest tasks are purged when at capacity.
+struct TaskBuffer<T: AsyncTask> {
+    /// Lock for the buffer.
+    lock: std::sync::Mutex<()>,
```
halfprice: suggestion: consider using tokio::sync::mutex so that pending task does not occupy event loop.

crates/walrus-indexer/src/lib.rs
```diff
@@ -3,6 +3,80 @@
...
+/// Persistent storage for async tasks.
+/// It provides the APIs for a task queue and a retry queue, where each queue is a sorted list of
+/// tasks by their task IDs.
+#[async_trait]
+pub trait AsyncTaskStore<T>: Send + Sync
+where
+    T: AsyncTask,
+{
+    /// Stores a task.
+    async fn store_task(&self, task: &T) -> Result<(), TypedStoreError>;
```
halfprice: Question: the TypedStoreError seems a bit too specific to TypedStore. Can it be a more generic type and more flexible?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+/// - Accepts real-time tasks, and persists them to disk.
+/// - Guarantees that tasks are processed in order of their task IDs.
+pub struct AsyncTaskSorter<T, S>
+where
+    T: AsyncTask,
+    S: AsyncTaskStore<T>,
+{
+    config: AsyncTaskSorterConfig,
+    store: Arc<S>,
+    buffer: Arc<RwLock<TaskBuffer<T>>>,
```
halfprice: question: all accesses in TaskBuffer are already guarded by the internal lock. Is having a outer RwLock and inner Mutex is redundant? Maybe either remove the inner lock, or change TaskBuffer {} to TaskBuffer { inner: Mutex<TaskBufferInner>}

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+
+    /// Enqueue a task to the queue, the task will be persisted to storage.
+    pub async fn enqueue_task(&self, task: T) -> Result<()> {
+        self.store.store_task(&task).await?;
+
+        // Add to latest queue.
+        self.buffer.write().await.push_to_latest(task);
+
+        // Notify waiting pullers.
+        self.item_available.notify_waiters();
```
halfprice: remark: should we use `notify_one` here, so that the next call of `item_available.notified()` will be notified?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+    T: AsyncTask,
+    S: AsyncTaskStore<T>,
+{
+    config: AsyncTaskSorterConfig,
+    store: Arc<S>,
+    buffer: Arc<RwLock<TaskBuffer<T>>>,
+    /// Notifier for when new tasks are available to the puller.
+    item_available: Arc<Notify>,
+    /// Background task handle for loading tasks from storage.
+    background_task_handle: Arc<RwLock<Option<tokio::task::JoinHandle<()>>>>,
```
halfprice: question: why all the RwLock used in this struct? Making AsyncTaskSorter Sync?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+
+                    // Wait for notification that tasks are available.
+                    self.item_available.notified().await;
+                }
+            }
+        }
+    }
+
+    /// Spawn a background task to load tasks from the store.
+    fn spawn_background_load_task(
```
halfprice: suggestion: after reading this part, it feels that the `store` can be an implementation detail of `buffer`, and the AsyncTaskSorter can only operate on `buffer` to enqueue and fetch tasks. Will there be more implementation that will benefit from `store` being exposed in the AsyncTaskSorter?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+            catchup_queue: VecDeque::new(),
+            latest_queue: VecDeque::new(),
+            catch_up_low_watermark: None,
+            grace_period_start: Some(Instant::now()),
+            state: QueueState::CatchingUp,
+            config,
+        }
+    }
+
+    /// Push a task to the latest queue, potentially dropping old tasks if at capacity.
```
halfprice: nit: consider adding a comment that the dropped task will still be stored in `store`? Otherwise, it might be confusing why can we drop task. This is another place I few the `store` can be inside TaskBuffer.

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+    /// Lock for the buffer.
+    lock: std::sync::Mutex<()>,
+    /// Queue for oldest tasks, loaded from disk during catch-up.
+    catchup_queue: VecDeque<T>,
+    /// Queue for latest tasks, where the oldest tasks are purged when at capacity.
+    latest_queue: VecDeque<T>,
+    /// Highest task_id seen in catchup queue.
+    catch_up_low_watermark: Option<T::TaskId>,
+    /// Grace period start time, the latest tasks are not purged during the grace period.
+    grace_period_start: Option<Instant>,
```
halfprice: question: why do we need grace period?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+        // Check if we're in grace period.
+        let in_grace_period = self
+            .grace_period_start
+            .map(|start| start.elapsed() < self.config.grace_period)
+            .unwrap_or(false);
+
+        // Only drop items if not in grace period and at capacity.
+        while !in_grace_period
+            && self.latest_queue.len() >= self.config.max_num_latest_tasks
+            && !self.catchup_queue.is_empty()
```
halfprice: question: hmm, I looked at this condition for a bit, and I wonder should it be `self.catchup_queue.is_empty()`? If catchup_queue is not empty, should the popped task from latest_queue being added to catchup_queue?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,512 @@
...
+                    tracing::error!("Background load task failed: {}", e);
+                }
+            }
+        });
+
+        // Store the task handle for proper cleanup.
+        if let Ok(mut handle_guard) = self.background_task_handle.try_write() {
+            *handle_guard = Some(handle);
+        } else {
+            tracing::warn!("Failed to store background task handle - unable to acquire write lock");
```
halfprice: question: it seems that the handle is dropped here, is this expected?

mlegner:
@liquid-helium As @halfprice already made a pass, I'm removing my review request for now. Please re-request a review if you want me to take a look as well.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: The lock is currently used in synchronous functions only, and all operations are in-memory, in this case, is it still a concern?

crates/walrus-indexer/src/lib.rs
liquid-helium: Good idea, added a new error type!

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: That's a good point.
The original design is that it could be called by multiple consumers, but in reality for us there is always one consumer. I will switch to notify_one() to avoid any complicated synchronization to make this right for multiple consumers.
Also added comments to be clear this is supposed to be called by a single consumer.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Thanks for the catch, this is a leftover of a previous impl, removed.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: For background_task_handle, the RwLock is to provide mutability, switched to Cell.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Yes, store will be used by other methods as well, that's why I keep the TaskBuffer a complete in-memory struct, and don't actively call store, but instead relies on AsyncTaskSort to load the task from store and call add_loaded_task() to do the backfill.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: I've added comments to clarify the expected behavior.
The TaskBuffer is intentionally designed as a simple caching utility rather than a self-contained struct. It makes more sense when viewed as a basic component within AsyncTaskSorter rather than a standalone entity.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: To avoid 'oscillating', i.e., it catches up, and immediately fall behind, and then repeat this.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Thanks for the catch, this is actually supposed to be a safety check `!inner.latest_queue.is_empty()` , fixed it.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Switched to Cell, no concern now. Thanks for the catch.

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,637 @@
...
+            let should_spawn = match (&from_id, &to_id) {
+                (Some(from), Some(to)) if from >= to => {
+                    // Already caught up!
+                    tracing::debug!(?from, ?to, "already caught up, no fetch needed");
+                    false
+                }
+                _ => true,
+            };
+
+            (from_id, to_id, should_spawn)
```
halfprice: Nit: can we document the reasoning here a bit more? Mostly because if either from_id or to_id is None, should_spawn will be set to true, and it's not obvious why that's the case.

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,637 @@
...
+        from_task_id: Option<T::TaskId>,
+        to_task_id: Option<T::TaskId>,
+    ) {
+        let mut inner = self.inner.lock().expect("failed to lock buffer");
+
+        // Clean up finished task or skip if already running.
+        if let Some(handle) = &inner.background_task_handle
+            && !handle.is_finished()
+        {
+            tracing::debug!("background fetch already in progress");
```
halfprice: Question: what if the background task loads different range? Is it ok to skip this one?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,637 @@
...
+                .catchup_queue
+                .back()
+                .map(|t| t.task_id())
+                .or(inner.catch_up_low_watermark.as_ref().cloned());
+
+            let to_id = inner.latest_queue.front().map(|t| t.task_id());
+
+            // Check if range is valid.
+            let should_spawn = match (&from_id, &to_id) {
+                (Some(from), Some(to)) if from >= to => {
```
halfprice: Remark: should the condition be `from +1 >= to` since we don't need to fetch both `from` and `to`?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,637 @@
...
+                    );
+                    inner_guard.add_loaded_tasks(tasks);
+                }
+                Err(e) => {
+                    tracing::error!(error = ?e, "background load task failed");
+                }
+            }
+
+            // Notify waiting pullers.
+            inner_guard.fetch_completed.notify_one();
```
halfprice, line 381: Nit: can also set `inner_guard.background_task_handle = None` here

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,637 @@
...
+                    // Caught up! Transition to UpToDate.
+                    let task_id = task.task_id();
+                    self.transition_to_up_to_date();
+                    tracing::debug!(
+                        ?task_id,
+                        ?first_latest_id,
+                        "caught up to latest queue, transitioning to UpToDate"
+                    );
+                    return;
+                }
```
halfprice: Nit: can we check `if tasks.back().task_id() >= first_lastest.task_id()` at the beginning of this function and return quickly there?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,637 @@
...
+    async fn pull(&self) -> Option<T> {
+        loop {
+            // Try to get a task from queues and determine if we need to wait.
+            let wait_state = {
+                let mut inner = self.inner.lock().expect("failed to lock buffer");
+
+                if inner.state == QueueState::UpToDate {
+                    // UpToDate: prioritize catchup queue, then latest queue.
+                    if let Some(task) = inner.catchup_queue.pop_front() {
+                        inner.catch_up_low_watermark = Some(task.task_id().clone());
```
halfprice: Question: updating catch_up_low_watermark is a bit confusing. Sometimes it is set to the highest task in the catch up quue, and here it is set to the lowest tasks in the catch up queued. Can you add a bit more information how is it used?

crates/walrus-indexer/src/async_task_sorter.rs
```diff
@@ -0,0 +1,637 @@
...
+            return;
+        }
+
+        // Handle overflow when at capacity.
+        if inner.latest_queue.len() >= self.config.latest_queue_max {
+            inner.drop_tasks(self.config.drop_batch_size, self.config.catchup_queue_max);
+        }
+
+        tracing::debug!(?task, "added task to latest queue");
+        inner.latest_queue.push_back(task);
```
halfprice, line 251: nit: can you add some doc why we need to add to the lasted queue, if the node is in catch up mode? It's not very obvious.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Done.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Yes, since eventually what matters is the tasks inserted into the catchup queue, if some tasks are missing, another fetch will be requested.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: That's right, however, the generic type AsyncTask doesn't implement "+", it might be a overkill, keeping it as.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Sure, added.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Refactored a bit, now it is always the 'highest task' we have ever added to catchup queue.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Refactored a bit to truncate the vector before adding them to the queue.

crates/walrus-indexer/src/async_task_sorter.rs
liquid-helium: Done.

PR #2608, opened 2025-10-07
[1/n][tail uploads] feat: Remove lifetime from node handles

sadhansood:
## Description

This PR refactorS storage-node communication handles to own their data so async uploader work can hold them 'static. Converted the communication layer to use Arc<StorageNode> and Arc<EncodingConfig> and removed lifetime-heavy APIs making node handles cloneable across tasks.

## Test plan

cargo clippy
---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client/communication/node.rs
+++ b/crates/walrus-sdk/src/client/communication/node.rs
@@ -91,11 +91,12 @@ impl<T, E> WeightedResult for NodeResult<T, E> {
     }
 }
 
-pub(crate) struct NodeCommunication<'a, W = ()> {
+#[derive(Debug)]
+pub struct NodeCommunication<W = ()> {
     pub node_index: NodeIndex,
     pub committee_epoch: Epoch,
-    pub node: &'a StorageNode,
-    pub encoding_config: &'a EncodingConfig,
+    pub node: Arc<StorageNode>,
+    pub encoding_config: Arc<EncodingConfig>,
     pub span: Span,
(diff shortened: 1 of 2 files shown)


crates/walrus-sdk/src/client/communication/node.rs
```diff
@@ -91,31 +91,32 @@ impl<T, E> WeightedResult for NodeResult<T, E> {
...
 }
 
-pub(crate) struct NodeCommunication<'a, W = ()> {
+#[derive(Debug)]
+pub struct NodeCommunication<W = ()> {
     pub node_index: NodeIndex,
     pub committee_epoch: Epoch,
-    pub node: &'a StorageNode,
-    pub encoding_config: &'a EncodingConfig,
+    pub node: Arc<StorageNode>,
```
mlegner: Question: Is there any benefit of using an `Arc` here over just an owned `StorageNode`?

mlegner (approved):
LGTM.

crates/walrus-sdk/src/client/communication/node.rs
halfprice: +1, similar question to the `encoding_config` field as well.

It seems that `node` currently is not cloned anywhere, so the Arc doesn't save much memory.

sadhansood: done

PR #2609, opened 2025-10-07
feat: [2/n][tail uploads] Introduce shared uploader and detached tail handling

sadhansood:
## Description

1. This PR introduces `uplaoder.rs` as the shared uploader engine which does multiple things  - tailHandling, event emission, etc so multiple callers can orchestrate background uploads consistently. 
2. Also reworked WalrusNodeClient::send_blob_data_and_get_certificate to stream uploader events, forward optional quorum notifications, and manage spawned tail tasks before fetching confirmations
3. Added `distributed_upload_without_confirmation` and new `StoreArgs` knobs (tail_handling, quorum channel, tail-handle collector) to let callers opt into detached tails (This will be used for background child processing)

## Test plan

Enabled detached tail uplaods in e2e test blob r/w


crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -161,6 +167,14 @@ where
         .reserve_and_store_blobs_retry_committees_with_path(&blobs_with_paths, &store_args)
         .await?;
 
+    // Wait for the tail uploads to complete.
+    // This is going to ensure that we wait for tail uploads to complete before reading the blobs.
+    // This is similar to blocking on tail uploads but happens after forming the certificate.
+    let mut handles = tail_handle_collector.lock().await;
```
wbbradley, line 175: Are we sure that at this point the tail_handle_collector has been populated? Is it a safe assumption that the population of these tail handles is done before `reserve_and_store_blobs_retry_committees_with_path` returns?

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -161,6 +167,14 @@ where
...
+    while let Some(handle) = handles.pop() {
```
wbbradley, line 176: Thinking out loud: There may be a way to use futures::join! here, I'm not sure if it's super important if these are just completion handles, but, it could be.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -161,6 +167,14 @@ where
...
+        handle.await.unwrap();
```
wbbradley, line 177: I guess the purpose of unwrapping here is just to ensure that it is unwrappable?

crates/walrus-sdk/src/client.rs
```diff
@@ -1610,18 +1615,19 @@ impl<T> WalrusNodeClient<T> {
...
     #[tracing::instrument(skip_all)]
+    #[allow(clippy::too_many_arguments)]
     pub async fn send_blob_data_and_get_certificate(
         &self,
         metadata: &VerifiedBlobMetadataWithId,
         pairs: &[SliverPair],
         blob_persistence_type: &BlobPersistenceType,
         multi_pb: Option<&MultiProgress>,
+        tail_handling: TailHandling,
+        quorum_forwarder: Option<tokio::sync::mpsc::Sender<UploaderEvent>>,
```
wbbradley, line 1663: Not sure what our policies on this our, but it might be nice to `use tokio::sync::mpsc::Sender`, Mutex, and JoinHandle here to simplify reading this code.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+    Detached,
+}
+
+/// Outcome returned by the uploader run.
+#[derive(Debug)]
+pub struct RunOutput<R, E> {
+    /// The results of the upload.
+    pub results: Vec<NodeResult<R, E>>,
+    /// The handle to the detached tail upload.
+    pub tail_handle: Option<tokio::task::JoinHandle<()>>,
```
wbbradley: nit: `use tokio::task::JoinHandle` should be fine, I think. It's pretty common.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/// A trait for types that have a `BlobId`.
+pub trait HasBlobId {
```
wbbradley: This should probably not be `pub` if it's only used in this file as a convenience for generalization.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+impl<T> HasBlobId for (BlobId, T) {
+    fn blob_id(&self) -> &BlobId {
+        &self.0
+    }
+}
+
+/// A work item for the uploader, representing a set of sliver pairs for a single blob
+/// that need to be sent to a specific node.
+#[derive(Debug, Clone)]
+pub struct UploadWorkItem {
```
wbbradley, line 50: Probably should not be `pub` unless we want it exposed to Rust SDK users which I think we don't if I'm understanding it.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+impl UploadWorkItem {
+    /// The blob ID.
+    pub fn blob_id(&self) -> &BlobId {
+        self.metadata.blob_id()
+    }
+}
+
+/// Tracks the upload progress for a single blob.
+#[derive(Debug, Clone, Default)]
+pub struct BlobUploadProgress {
```
wbbradley: Probably should not be `pub`.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+            let mut pairs_per_node: HashMap<usize, Vec<SliverPair>> = HashMap::new();
+            for pair in pairs {
+                let shard_index = pair.index().to_shard_index(committees.n_shards(), &blob_id);
+                for (node_index, node) in committees.write_committee().members().iter().enumerate()
+                {
+                    if node.shard_ids.contains(&shard_index) {
+                        pairs_per_node
+                            .entry(node_index)
+                            .or_default()
+                            .push(pair.clone());
```
wbbradley: This looks like it's copying all the data in the blob. This will duplicate the memory of the slivers, no?

wbbradley: We might want to move this data entirely by ownership, or pass it by reference.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+            committees,
+            progress,
+            comms,
+            sliver_write_limit,
+            sliver_write_extra_time,
+        }
+    }
+
+    /// Runs the distributed upload process.
+    pub async fn run<F, Fut, R, T, E>(
```
wbbradley: Can we give this a more greppable name? `run` is hard to search for. Can we name it `run_distributed_upload` to make it more unique?

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+        &mut self,
+        upload_action: F,
+        event_sender: tokio::sync::mpsc::Sender<UploaderEvent>,
```
wbbradley, line 170: `use tokio::sync::mpsc::Sender` (nit/preference)

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+                (n, work)
+            })
+            .collect();
+
+        let upload_action = Arc::new(upload_action);
+        let futures = scheduled.into_iter().map({
+            let upload_action = Arc::clone(&upload_action);
+            let semaphore = Arc::clone(&semaphore);
+            move |(n, work)| {
+                let upload_action = Arc::clone(&upload_action);
```
wbbradley: Why can't we just do `upload_action.clone()` here?

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+                let sem_clone = Arc::clone(&semaphore);
```
wbbradley: Same here.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+                        .acquire()
+                        .await
+                        .expect("semaphore acquire should not fail");
+                    (*upload_action)(n, work).await
+                }
+            }
+        });
+
+        let mut requests = WeightedFutures::new(futures);
+        let start = std::time::Instant::now();
```
wbbradley: `use std::time::Instant`

wbbradley: Or could use `tokio::time::Instant` ... I tend to use the latter when inside of async fns just to keep things aligned with tokio, but I don't have strong preferences about that

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+                        } else {
+                            tracing::debug!(blob_id = %blob_id, "sent blob quorum reached event");
+                        }
+                    }
+                }
+            }
+
+            results.push(node_result);
+            let done = blobs_at_quorum == self.progress.len();
+            if done {
```
wbbradley:
```suggestion
            if blobs_at_quorum == self.progress.len()  {
```
Maybe omit extra variable since it's not really needed?

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+                break;
+            }
+        }
+
+        let extra_time = self.sliver_write_extra_time.extra_time(start.elapsed());
+        let n_shards: usize = self.committees.n_shards().get().into();
+
+        let tail_handle = if matches!(tail_handling, TailHandling::Detached)
```
wbbradley: `TailHandling` supports `Eq` so you can just compare `tail_handling == TailHandling::Detached` in this case.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+            && extra_time > Duration::from_millis(0)
+        {
+            tracing::debug!("uploader: spawning detached tail handle");
+            Some(tokio::spawn(async move {
+                let mut requests = requests;
+                let reason = requests.execute_time(extra_time, n_shards).await;
+                tracing::debug!(
+                    "uploader: detached tail handle completed with reason: {:?}",
+                    reason
+                );
```
wbbradley:
```suggestion
                tracing::debug!(
                    reason,
                    "uploader: detached tail handle completed"
                );
```
Could do above.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,314 @@
...
+                let results = requests.into_results();
+                tracing::debug!(
+                    "uploader: detached tail handle results: {:?}",
+                    results.len()
+                );
+            }))
+        } else {
+            if extra_time > Duration::from_millis(0) {
```
wbbradley:
```suggestion
            if !extra_time.is_zero() {
```

crates/walrus-sdk/src/uploader.rs
wbbradley: Looking at this tail_handle, I think you would want to pump all the futures together with `join!` or something like that, otherwise, they won't make progress unless `await`-ed. I could be misreading this.

crates/walrus-e2e-tests/tests/test_client.rs
wbbradley: Coming back to this after reading the rest of the PR, I do think this is either serial by design or accident, in terms of which tail_handles are pumped or not. But, I couldn't tell if it was on purpose. Sorry if I'm misreading it. So, yes, I think you want to use a multi-future pumper, like `join!`.

crates/walrus-e2e-tests/tests/test_client.rs
sadhansood: Yes, that is correct. `send_blob_data_and_get_certificate` is what is updating the collector with tail handles which is called from `reserve_and_store_blobs_retry_committees_with_path`.

crates/walrus-e2e-tests/tests/test_client.rs
sadhansood: This vector contains one tokio::task handle (not futures) per blob, and it seems like just awaiting in a while loop is enough if we need to wait out all of the handles.

crates/walrus-e2e-tests/tests/test_client.rs
sadhansood: Yes, just to be sure there were no errors.

crates/walrus-sdk/src/client.rs
sadhansood: I only used the fully qualified name for `Sender` as there could be other type of senders like `Oneshot`, etc

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: all done.

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: This needs to be pub as it is in the signature of `run` function

crates/walrus-sdk/src/uploader.rs
sadhansood: This needs to be pub as it is in the signature of `run` function

crates/walrus-sdk/src/uploader.rs
sadhansood: Just for preventing confusion with other kinds of tokio channel senders, I'll keep it.

crates/walrus-sdk/src/uploader.rs
sadhansood: good catch! I've now fixed it by wrapping `Vec<SliverPair>` with an `Arc` and keeping a list of index into this vec in `UploadWorkItem` per blob.

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: i think this comment is in the context of where we are await all the tail handles in a while loop (which I've already responded to that these are tokio task handles and not futures)

crates/walrus-e2e-tests/tests/test_client.rs
wbbradley: Gotcha, ok sweet. I was worried it was the actual futures. Thanks!

crates/walrus-sdk/src/client.rs
wbbradley: Gotcha. I wasn't aware of that.

crates/walrus-sdk/src/client.rs
```diff
@@ -1635,122 +1641,213 @@ impl<T> WalrusNodeClient<T> {
...
-                    metadata,
-                    pairs_per_node
-                        .remove(&n.node_index)
-                        .expect("there are shards for each node"),
-                    blob_persistence_type,
-                );
+        let blobs = vec![(metadata.clone(), Arc::new(pairs.to_vec()))];
+        let (event_tx, mut event_rx) = tokio::sync::mpsc::channel(blobs.len().max(1));
+
+        let mut upload_fut = Box::pin(self.distributed_upload_without_confirmation(
```
halfprice, line 1690: question: will we ever upload more than 1 blob? The function takes a vector of blobs as input, but the only use case inputs one blob.

crates/walrus-sdk/src/client.rs
```diff
@@ -1635,122 +1641,213 @@ impl<T> WalrusNodeClient<T> {
...
+                            }
+                            UploaderEvent::BlobQuorumReached { .. } => {
+                                tracing::debug!(
+                                    blob_id = %metadata.blob_id(),
+                                    "received blob quorum reached event");
+                                if let Some(pb) = &progress_bar && !pb.is_finished() {
+                                    pb.finish_with_message(
+                                        format!("slivers sent ({})", metadata.blob_id()));
+                                }
+                            }
```
halfprice, line 1735: question: do we need to quit the loop in this case?

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,316 @@
...
+            }
+        });
+
+        let mut requests = WeightedFutures::new(futures);
+        let start = Instant::now();
+
+        let mut blobs_at_quorum = 0;
+        let mut results: Vec<NodeResult<R, E>> = Vec::new();
+
+        while let Some(node_result) = requests.next(self.committees.n_shards().get().into()).await {
```
halfprice, line 212: Correct me if I'm wrong here: this is a change of implementation that we no longer use WeightedFutures to wait for a quorum. Instead, the while loop itself does weight counting? If so, why do we need this change?

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,316 @@
...
+            move |(n, work)| {
+                let upload_action = upload_action.clone();
+                let sem_clone = semaphore.clone();
+                async move {
+                    let _permit = sem_clone
+                        .acquire()
+                        .await
+                        .expect("semaphore acquire should not fail");
+                    (*upload_action)(n, work).await
+                }
```
halfprice, line 202: Question: I think this code is inherited from the old one, but do you know why we are using semaphore here for concurrency control but not using the n_concurrent in the WeightedFuture? Also, the node communication is also constructed by taking sliver_write_limit as an input. The limit seems can be in effect in multiple places is a bit confusing.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,316 @@
...
+        blob_id: BlobId,
+        /// The elapsed time since the upload started.
+        elapsed: Duration,
+    },
+}
+
+/// The `DistributedUploader` component.
+#[derive(Debug)]
+pub struct DistributedUploader {
+    /// The blobs to be uploaded.
```
halfprice:
```suggestion
    /// Keyed by node index, the vector contains sliver pairs need to be uploaded to the corresponding storage node.
```

Or something similar. The current doc is too vague, and especially the meaning of the key needs more clarity.

crates/walrus-sdk/src/client.rs
```diff
@@ -1635,122 +1641,213 @@ impl<T> WalrusNodeClient<T> {
...
+
+                        match response.result {
+                            Ok(()) => stored.push(*item.blob_id()),
+                            Err(err) => {
+                                return NodeResult::new(
+                                    response.committee_epoch,
+                                    response.weight,
+                                    response.node,
+                                    Err(err),
+                                );
```
halfprice, line 1870: Remark: this seems to be exactly the same as `response`.

crates/walrus-sdk/src/client.rs
```diff
@@ -1635,122 +1641,213 @@ impl<T> WalrusNodeClient<T> {
...
+                        }
+                    }
 
-        let results = requests.into_results();
+                    NodeResult::new(
+                        node.committee_epoch,
+                        node.n_owned_shards().get().into(),
+                        node.node_index,
+                        Ok(stored),
+                    )
```
halfprice, line 1880: Remark: it seems that if this function takes multiple blobs, one blob's sliver upload failure to one storage node also affects other blobs? It looks like the for loop returns error here if encounter an error at earlier `work`. IIUC, the work contains the same number of item as the number of blobs called in this function. Is that correct? Like mentioned in an earlier comment, I feel the logic can be simplified if the function only uploads 1 blob.

crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -78,6 +95,27 @@ impl StoreArgs {
...
+    }
+
+    /// Sets the uploader event channel used to forward quorum notifications.
+    pub fn with_quorum_event_tx(mut self, tx: MpscSender<UploaderEvent>) -> Self {
+        self.quorum_event_tx = Some(tx);
+        self
+    }
+
+    /// Sets the collector for detached tail handles.
+    pub fn with_tail_handle_collector(
```
halfprice, line 116: Question: is this function only created for testing purpose? If so, can we make it only visible in tests?

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -147,12 +150,15 @@ where
...
 
+    let tail_handle_collector = Arc::new(Mutex::new(vec![]));
     let store_args = {
         let store_args = StoreArgs::default_with_epochs(1).no_store_optimizations();
         if let Some(upload_relay_client) = upload_relay_client {
             store_args.with_upload_relay_client(upload_relay_client)
         } else {
             store_args
+                .with_tail_handling(TailHandling::Detached)
+                .with_tail_handle_collector(tail_handle_collector.clone())
```
halfprice, line 161: Question: is this required? What happens if we don't add it?

crates/walrus-sdk/src/client.rs
```diff
@@ -1635,122 +1641,213 @@ impl<T> WalrusNodeClient<T> {
...
+            if let Some(collector) = tail_handle_collector {
+                collector.lock().await.push(handle);
+                tracing::debug!(blob_id = %metadata.blob_id(), "queued tail handle for collector");
+            } else if matches!(tail_handling, TailHandling::Detached) {
+                tracing::debug!(blob_id = %metadata.blob_id(), "spawned detached tail handler");
+                tokio::spawn(async move {
+                    if let Err(err) = handle.await {
+                        tracing::warn!(?err, "tail upload task failed");
+                    }
+                });
```
halfprice, line 1778: question: why do we need to spawn another task? Isn't the task in `handle` already running in another thread?

crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -33,6 +38,12 @@ pub struct StoreArgs {
     pub metrics: Option<Arc<ClientMetrics>>,
     /// The optional upload relay client, that allows to store the blob via the relay.
     pub upload_relay_client: Option<UploadRelayClient>,
+    /// Tail handling preference for sliver uploads.
+    pub tail_handling: TailHandling,
+    /// Optional channel to forward uploader events to.
+    pub quorum_event_tx: Option<MpscSender<UploaderEvent>>,
+    /// Optional collector that receives detached tail handles for later awaiting.
+    pub tail_handle_collector: Option<Arc<Mutex<Vec<JoinHandle<()>>>>>,
```
halfprice, line 50: Remark: these two option, `tail_handling` and `tail_handle_collector` 's usage is a bit unclear. 

Is this the intended behavior that if tail_handling is detached, tail_handle_collector is not None, put all tail handling task in the collector, but tail_handle_collector is None, running tail handling in a background thread?

I'm also not sure how the tail_handle_collector will be used. Right now, it is only used in one test, so also not sure if this is only for testing purpose. 

I wonder if we change TailHandling to

enum TailHandling {
    Blocking,
    Detached,
    Collected(Arc<Mutex<Vec<JoinHandle<()>>>>)
}

is cleaner

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,316 @@
...
+#[derive(Debug)]
+pub struct RunOutput<R, E> {
+    /// The results of the upload.
+    pub results: Vec<NodeResult<R, E>>,
+    /// The handle to the detached tail upload.
+    pub tail_handle: Option<JoinHandle<()>>,
+}
+
+/// A trait for types that have a `BlobId`.
+pub trait HasBlobId {
```
halfprice: Question: is this trait needed? Right now, it seems that we can replace where this trait object to BlobID type? I don't see where `impl<T> HasBlobId for (BlobId, T) ` is used.

crates/walrus-sdk/src/uploader.rs
```diff
@@ -0,0 +1,316 @@
...
+    /// The extra time to wait for tail-end writes.
+    sliver_write_extra_time: SliverWriteExtraTime,
+}
+
+impl DistributedUploader {
+    /// Creates a new `DistributedUploader`.
+    pub fn new(
+        blobs: &[(VerifiedBlobMetadataWithId, Arc<Vec<SliverPair>>)],
+        committees: Arc<ActiveCommittees>,
+        comms: Vec<NodeWriteCommunication>,
```
halfprice, line 122: Nit: I think it is critical that the `comms` must be created based on `committees`. Can you add a doc that these tow must be consistent? Otherwise, the node index in comms may be pointing to a different node.

crates/walrus-sdk/src/client.rs
sadhansood: We can’t break the loop on `BlobQuorumReached`. we need to drain the channel until `run_distributed_upload` resolves so we obtain the tail handle and surface terminal errors. That holds even once the child process starts managing tails.

crates/walrus-sdk/src/client.rs
sadhansood: This is just to log any errors that might happen while awaiting the tail upload handle.

crates/walrus-sdk/src/client.rs
sadhansood: The closure returns NodeResult<Vec<BlobId>, E>, but response is NodeResult<(), E>. That's why we needed the rewrap.

crates/walrus-sdk/src/client.rs
sadhansood: You’re right that once we start batching blobs we’ll want to isolate failures per blob rather than bailing the whole node. I can follow up by collecting successes into `stored` but continuing through the remaining work items, then wrapping the error information so the orchestrator can report which blob(s) failed on that node without discarding the others.

crates/walrus-sdk/src/client/store_args.rs
sadhansood: I’ll expand the docs around tail_handling and tail_handle_collector to explain the expected combinations and mention that the collector is how callers keep explicit ownership of detached tasks. Folding the collector into the enum would cost `Copy` and complicate the blocking path, so I’d prefer to keep the current shape for now. Btw, in the following PR #2612 the usage of collector and handle becomes much clear (it is needed beyond tests).

crates/walrus-sdk/src/client/store_args.rs
sadhansood: No, it is indeed needed for the client the usage is visible in #2612

crates/walrus-sdk/src/uploader.rs
sadhansood: good point, this was useful in a previous iteration but not anymore. Removed the trait.

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: The explicit semaphore caps how many sliver uploads we drive concurrently but `WeightedFutures` still polls up to `n_shards` so it can satisfy the quorum predicate quickly. But, the behavior here mimics the current implementation very closely.

crates/walrus-sdk/src/uploader.rs
sadhansood: Switching away from `execute_until` is intentional because the while loop lets us count weight per blob, emit progress and quorum events, and bail once every tracked blob hits quorum behavior which the child-process follow-up also depends on.

crates/walrus-sdk/src/client.rs
sadhansood: We still hand a single blob to `distributed_upload_without_confirmation` today, but keeping the vector matches the uploader API the new background-worker PR (#2612) already relies on for multi-blob batches.

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-e2e-tests/tests/test_client.rs
sadhansood: It'll absolutely still work in blocking mode. I just added it so we can test the new code path.

crates/walrus-sdk/src/client.rs
halfprice: ah, make sense

crates/walrus-e2e-tests/tests/test_client.rs
halfprice: I think this testing client is used only in limited tests.

Can you add a an option here as well https://github.com/MystenLabs/walrus/blob/5fccc498670ead72c56926e0d936cd1c81382b82/crates/walrus-simtest/src/test_utils.rs#L138-L142 ? So that all simtest can exercise the new code path.

Also, only select Detached mode probabilstically so that we still have coverage for the existing code path.

sadhansood: done

sadhansood:
Thank you @halfprice and @wbbradley for your review. I have addressed all your comments.

crates/walrus-sdk/src/client.rs
```diff
@@ -1672,122 +1678,213 @@ impl<T> WalrusNodeClient<T> {
...
-
-            move |n| {
-                let fut = n.store_metadata_and_pairs(
-                    metadata,
-                    pairs_per_node
-                        .remove(&n.node_index)
-                        .expect("there are shards for each node"),
-                    blob_persistence_type,
-                );
+        let blobs = vec![(metadata.clone(), Arc::new(pairs.to_vec()))];
```
mlegner, line 1687: Hey @sadhansood! I know I'm a bit late to the party... 😅 

While working on improvements to the memory usage of the client, I noticed duplicated allocations for all slivers. My investigation led me to this change, where we effectively duplicate all slivers, which causes an additional ~4.5GiB of memory usage when storing a 1GiB blob. 

Could you take a look and investigate whether we could avoid these clones (e.g., by creating `Arc`s earlier in the call tree)?

sadhansood: Thank you for catching this @mlegner , could you please review https://github.com/MystenLabs/walrus/pull/2659. Thanks

PR #2612, opened 2025-10-08
feat: [3/n][tail uploads] Add child process to offload sliver uploads

sadhansood:
## Description

This PR adds a config option on the client to optionally perform sliver uploads through a background process in the cli. The main changes are in `internal_run.rs` and `runner.rs`. 
In `internal_run.rs` - we add all the scaffolding to spawn a new child process and do event passing b/w uploader and parent.
In `runner.rs` - we spawn child process if it is enabled in the config for both `store` and `store_quilt` commands

## Test plan

Running locally several times yield expected results.

### DIFF
--- a/crates/walrus-sdk/src/config/communication_config.rs
+++ b/crates/walrus-sdk/src/config/communication_config.rs
@@ -33,6 +33,24 @@ fn default_sliver_status_check_threshold() -> usize {
     DEFAULT_SLIVER_STATUS_CHECK_THRESHOLD
 }
 
+/// Upload mode for controlling concurrency and aggressiveness of uploads.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum UploadMode {
+    /// Conservative mode: lower concurrency, slower but more reliable.
+    Conservative,
+    /// Balanced mode: moderate concurrency (default).
+    Balanced,
+    /// Aggressive mode: higher concurrency, faster but more resource-intensive.
+    Aggressive,
+}
+
+impl Default for UploadMode {
+    fn default() -> Self {
+        Self::Balanced
+    }
+}
+
 /// Configuration for the communication parameters of the client
 #[serde_as]
 #[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
@@ -65,10 +83,11 @@ pub struct ClientCommunicationConfig {
     /// The extra time allowed for sliver writes.
     pub sliver_write_extra_time: SliverWriteExtraTime,
     /// Limit under which the client skips the sliver status check before uploads.
-    ///
     /// Defaults to 5_560 bytes.
     #[serde(default = "default_sliver_status_check_threshold")]
     pub sliver_status_check_threshold: usize,
+    /// Enable uploading slivers via a detached child process that continues tail writes.
+    pub child_process_uploads_enabled: bool,
     /// The delay for which the client waits before storing data to ensure that storage nodes have
     /// seen the registration event.
     #[serde(rename = "registration_delay_millis")]
@@ -99,6 +118,7 @@ impl Default for ClientCommunicationConfig {
             request_rate_config: Default::default(),
             disable_proxy: Default::default(),
             sliver_write_extra_time: Default::default(),
+            child_process_uploads_enabled: false,
             registration_delay: Duration::from_millis(200),
             max_total_blob_size: 1024 * 1024 * 1024, // 1GiB
             sliver_status_check_threshold: DEFAULT_SLIVER_STATUS_CHECK_THRESHOLD,
--- a/crates/walrus-service/bin/client.rs
+++ b/crates/walrus-service/bin/client.rs
@@ -16,6 +16,7 @@ use walrus_service::{
     utils::{self, MetricsAndLoggingRuntime},
 };
 use walrus_sui::client::retry_client::RetriableRpcError;
+use walrus_utils::is_internal_run;
 
 // Define the `GIT_REVISION` and `VERSION` consts for the Walrus client.
 walrus_utils::bin_version!();
@@ -61,7 +62,9 @@ fn client() -> Result<()> {
     let mut app = ClientArgs::parse().inner;
     app.extract_json_command()?;
 
-    tracing::info!("client version: {VERSION}");
+    if !is_internal_run() {
+        tracing::info!("client version: {VERSION}");
+    }
     let runner = ClientCommandRunner::new(
         &app.config,
         app.context.as_deref(),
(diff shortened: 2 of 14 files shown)


crates/walrus-sui/src/config.rs
```diff
@@ -109,7 +110,12 @@ impl WalletConfig {
 
         let path = path_or_defaults_if_exist(wallet_config.and_then(|c| c.path()), &default_paths)
             .ok_or(anyhow!("could not find a valid wallet config file"))?;
-        tracing::info!("using Sui wallet configuration from '{}'", path.display());
+        if !env::var("INTERNAL_RUN")
```
halfprice: question: why using ENV to control this behavior instead of a client option?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -761,12 +807,124 @@ impl ClientCommandRunner {
...
+        if !internal_run {
+            tracing::info!(
+                duration = ?start_timer.elapsed(),
+                "{} out of {} blobs stored",
+                results.len(),
+                blobs_len
+            );
+        }
+
+        if internal_run {
```
halfprice, line 822: Nit: this can be combined with previous if statement

crates/walrus-service/src/client/cli/internal_run.rs
```diff
@@ -0,0 +1,766 @@
...
+    internal_run: bool,
+    command_name: &str,
+    add_command_args: F,
+    num_blobs: usize,
+) -> Result<Option<()>>
+where
+    F: FnOnce(&mut TokioCommand),
+{
+    let child_uploads_enabled = client.config().communication_config.child_uploads_enabled;
+    if !(child_uploads_enabled && upload_relay.is_none() && !internal_run) {
```
halfprice, line 474: Question: this condition is a bit hard to follow.. Can you add a doc to explain? I think what is not obvious here is that `internal_run` means the execution is inside child process already.

crates/walrus-service/src/client/cli/internal_run.rs
```diff
@@ -0,0 +1,766 @@
...
+    let config_filename = format!(
+        "walrus_child_config_{}_{}.yaml",
+        std::process::id(),
+        Utc::now().timestamp_millis()
+    );
+    let config_yaml_path = tmp_dir.join(config_filename);
+
+    if let Err(e) = stdfs::write(
+        &config_yaml_path,
+        serde_yaml::to_string(client.config()).context("serialize ClientConfig")?,
```
halfprice, line 490: remark: should we also update the extra wait time here to a longer time? Right now, it is still the same (50% extra time) even in detached mode.

crates/walrus-service/src/client/cli/internal_run.rs
```diff
@@ -0,0 +1,766 @@
...
+        mut store_args: StoreArgs,
+    ) -> StoreArgs {
+        if let Some(collector) = self.tail_handle_collector.as_ref() {
+            store_args = store_args.with_tail_handle_collector(collector.clone());
+        }
+        if let Some(tx) = self.uploader_event_tx.as_ref() {
+            store_args = store_args.with_quorum_event_tx(tx.clone());
+        }
+        if internal_run {
+            store_args = store_args.with_tail_handling(TailHandling::Detached);
```
halfprice, line 652: question: just want to confirm that this waits for all to finish, is that correct?

halfprice (commented):
Thanks @sadhansood ! The PR LGTM mostly! I think there are a lot duplication with the previous 2 PRs, so it'll be great if we can merge the first 2, and then take another look at this with clean diff.

crates/walrus-sui/src/config.rs
sadhansood: We read `INTERNAL_RUN` because the wallet and config loaders sit in lower layer crates that do not know about CLI options. The parent flips that env var when it forks the background uploader, so the child process can suppress user facing logging without plumbing an extra flag through every call site. A client facing option would not reach these helpers early enough I think to suppress logs, so the env flag is our way to do it.

crates/walrus-service/src/client/cli/runner.rs
sadhansood: Good catch! will do

crates/walrus-service/src/client/cli/internal_run.rs
sadhansood: That guard is preventing us from recursively spawning children. We only fork when (a) child uploads are enabled in config, (b) we are not using an upload relay and (c) we’re still on the original parent process (internal_run == false). I’ll add a short comment about this.

crates/walrus-service/src/client/cli/internal_run.rs
sadhansood: I think we could add a multiplier for detached mode, but also users can adjust the wait time through config. But right now, it is purely driven through config. Maybe, when we update the example config, we can also update the wait time multiplier to 75% along with tail upload to be detached.

crates/walrus-service/src/client/cli/internal_run.rs
sadhansood: Yes, every detached tail handle is pushed into the collector, and `InternalRunContext::await_tail_handles` drains and awaits them, so we don’t return to the user until all background tails finish.

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -65,10 +83,11 @@ pub struct ClientCommunicationConfig {
     /// The extra time allowed for sliver writes.
     pub sliver_write_extra_time: SliverWriteExtraTime,
     /// Limit under which the client skips the sliver status check before uploads.
-    ///
     /// Defaults to 5_560 bytes.
     #[serde(default = "default_sliver_status_check_threshold")]
     pub sliver_status_check_threshold: usize,
+    /// Enable uploading slivers via a detached child process that continues tail writes.
+    pub child_uploads_enabled: bool,
```
halfprice: nit: I think `child_uploads` is a bit confusing. Maybe `child_process_uploads_enabled`?

crates/walrus-service/bin/client.rs
```diff
@@ -61,7 +61,12 @@ fn client() -> Result<()> {
     let mut app = ClientArgs::parse().inner;
     app.extract_json_command()?;
 
-    tracing::info!("client version: {VERSION}");
+    if !env::var("INTERNAL_RUN")
+        .map(|v| v == "true")
+        .unwrap_or(false)
```
halfprice: Nit: can we extract this to a global function? This condition has been used in multiple places.

crates/walrus-sdk/src/config/communication_config.rs
sadhansood: done

crates/walrus-service/bin/client.rs
sadhansood: done

sadhansood:
> Thanks @sadhansood ! This PR LGTM and great work! 💯
> 
> Question: this PR is mainly focus on the CLI Store operation. For SDK Store operation, will we do similar thing there as well?

Thank you for in depth review @halfprice! I definitely plan on adding this and some of the other optimizations too sdk as well. Once we are done with cli, we will start porting these to sdk side as well.

PR #2613, opened 2025-10-08
feat: Make sliver existence check threshold configurable

sadhansood:
## Description

This PR makes the sliver existence check in `NodeCommunication` configurable through config. This gives us another lever to tune when trying to optimize upload performance.

## Test plan

Running locally.

### DIFF
--- a/crates/walrus-sdk/src/config/communication_config.rs
+++ b/crates/walrus-sdk/src/config/communication_config.rs
@@ -19,6 +19,20 @@ use crate::config::{
     sliver_write_extra_time::SliverWriteExtraTime,
 };
 
+/// Below this threshold, the `NodeCommunication` client will not check if the sliver is present on
+/// the node, but directly try to store it.
+///
+/// The threshold is chosen in a somewhat arbitrary way, but with the guiding principle that the
+/// direct sliver store should only take 1 RTT, therefore having similar latency to the sliver
+/// status check. To ensure this is the case, we take compute the threshold as follows: take the TCP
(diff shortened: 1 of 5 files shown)


crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -19,6 +19,17 @@ use crate::config::{
     sliver_write_extra_time::SliverWriteExtraTime,
 };
 
+/// Below this threshold, the `NodeCommunication` client will not check if the sliver is present on
+/// the node, but directly try to store it.
+///
+/// The threshold is chosend in a somewhat arbitrary way, but with the guiding principle that the
```
halfprice:
```suggestion
/// The threshold is chosen in a somewhat arbitrary way, but with the guiding principle that the
```

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -50,6 +61,10 @@ pub struct ClientCommunicationConfig {
     pub disable_native_certs: bool,
     /// The extra time allowed for sliver writes.
     pub sliver_write_extra_time: SliverWriteExtraTime,
+    /// Optional limit under which the client skips the sliver status check before uploads.
+    ///
+    /// `None` disables the check entirely.
+    pub sliver_status_check_threshold: Option<usize>,
```
halfprice: Nit: can we use the default value, 5560, as the default instead of setting this to an Option<usize>? It's not obvious that if not setting this field indicates always skipping or not skipping status check. So better to be explicit.

PR #2614, opened 2025-10-08
feat: Support deferring blob recovery by client during tail uploads

sadhansood:
## Description

This PR adds support for delaying blob recovery during event processing by a configurable amount configured via the rest endpoint by the client. This is needed because with tail upload optimization we are going to create certificate before waiting for tail uploads to finish in which case client is going to send deferral requests to all clients with the max_wait duration.
NOTE: This PR only includes server side change,  and client side change and enabling this in tests will happen in a follow up PR.


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -537,6 +537,37 @@ impl BlobSynchronizer {
         let this = Arc::new(self);
         let histograms = &this.metrics().recover_blob_part_duration_seconds;
 
+        match this
+            .node
+            .wait_until_recovery_deferral_expires(&this.blob_id)
```
halfprice: question: ideally, this wait should happen before creating blob_sync task. I guess there, it needs to manage waiting tasks, which adds more complexity?

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -180,6 +187,50 @@ pub async fn put_metadata<S: SyncServiceState>(
...
+    let mut duration = std::time::Duration::from_millis(ms);
+    let maybe_size_bound = state
+        .service
+        .retrieve_metadata(&blob_id)
+        .ok()
+        .and_then(|md| {
+            let size = md.as_ref().unencoded_length();
+            state.config.live_upload_deferral.deferral_for_size(size)
+        });
+    let bound = maybe_size_bound.unwrap_or(state.config.live_upload_deferral.max_total_defer);
```
halfprice: question: this indicates that `max_total_defer` must be the largest deferral time possible, correct?

crates/walrus-service/src/node/config.rs
```diff
@@ -783,6 +791,96 @@ impl GarbageCollectionConfig {
     }
 }
 
+/// Configuration for pre-certify notice handling and recovery deferral.
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(default)]
+pub struct PrecertifyNoticeConfig {
```
halfprice: question: what is this struct used for?

crates/walrus-sdk/src/config.rs
mlegner, line 1: Question: (Why) is this required?

crates/walrus-service/src/node/config.rs
```diff
@@ -783,6 +787,74 @@ impl GarbageCollectionConfig {
...
+
+/// Configuration that controls deferring recovery when a live client upload is likely.
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(default)]
+pub struct LiveUploadDeferralConfig {
+    /// Enables applying a deferral window based on blob size.
+    pub enabled: bool,
+    /// Lookup table from size upper-bounds to deferral durations. The first matching entry applies.
+    pub table: Vec<SizeDeferralEntry>,
```
mlegner: Suggestion: Can we change the variable name to better convey what this "table" is?

crates/walrus-service/src/node/config.rs
```diff
@@ -783,6 +787,74 @@ impl GarbageCollectionConfig {
...
+        for entry in &self.table {
+            if size_bytes <= entry.max_unencoded_bytes {
+                candidate = match candidate {
+                    Some(best) if best.max_unencoded_bytes <= entry.max_unencoded_bytes => {
+                        Some(best)
+                    }
+                    _ => Some(entry),
+                };
+            }
+        }
```
mlegner: Important: Unless I misunderstand something, this implementation is not consistent with the docstring "The first matching entry applies." above.

crates/walrus-service/src/node/errors.rs
```diff
@@ -105,19 +105,45 @@ pub struct IndexOutOfRange {
 )]
 pub struct Unavailable;
 
+#[derive(Debug, thiserror::Error, RestApiError)]
+#[rest_api_error(domain = ERROR_DOMAIN)]
```
mlegner: Question: Is this ever returned in an API response? If not, we could skip the `RestApiError` derivation.

crates/walrus-service/src/test_utils.rs
```diff
@@ -3079,6 +3102,15 @@ pub fn storage_node_config() -> WithTempDir<StorageNodeConfig> {
...
             garbage_collection: GarbageCollectionConfig::default_for_test(),
+            live_upload_deferral: LiveUploadDeferralConfig {
+                enabled: true,
+                table: vec![SizeDeferralEntry {
+                    max_unencoded_bytes: u64::MAX,
+                    defer: Duration::from_millis(100),
+                }],
+                max_total_defer: Duration::from_millis(100),
+                max_checkpoint_lag: 1,
+            },
```
mlegner: Suggestion: Define a `LiveUploadDeferralConfig::default_for_test` function, and use it here and in `crates/walrus-service/src/testbed.rs`.

crates/walrus-service/src/node/storage/shard.rs
mlegner: Nit: Can you document the changed return value?

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -61,6 +61,11 @@ use crate::{
 
 type ShardMetrics = CommonDatabaseMetrics;
 
+// Type aliases to simplify complex return types flagged by clippy::type_complexity.
+type NextBlobInfo = Option<(BlobId, BlobInfo)>;
+type ClearedBlobIds = Vec<BlobId>;
+type BatchFetchedSliversOutcome = Result<(NextBlobInfo, ClearedBlobIds), SyncShardClientError>;
```
mlegner: Question/nit: How about defining a struct instead?
```rust
struct BatchFetchedSliversOutcome {
    next_blob: Option<(BlobId, BlobInfo)>,
    cleared_blob_ids: Vec<BlobId>,
}
```

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -1053,6 +1060,10 @@ impl ShardStorage {
 
                 batch.write()?;
 
+                for blob_id in cleared_blob_ids {
+                    node.clear_recovery_deferral(&blob_id).await;
+                }
```
mlegner: Question: Can you explain why this is necessary? I don't fully understand the mechanism here.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -146,15 +150,23 @@ impl BackgroundEventProcessor {
...
             if let Err(error) = self
-                .process_event(tracked_event.event_handle, tracked_event.blob_event)
+                .process_event(event_handle, blob_event, checkpoint_position)
                 .await
             {
                 // TODO(WAL-874): to keep the same behavior as before BackgroundEventProcessor, we
                 // should propagate the error to the node and exit the process if necessary.
                 tracing::error!(?error, "error processing blob event");
             }
+            drop(guard);
```
mlegner: Nit: This should not be necessary as the guard automatically goes out of scope here.

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -537,6 +537,37 @@ impl BlobSynchronizer {
...
+            .await
+        {
+            Ok(token) => {
+                token.cancelled().await;
```
mlegner: Question: Why is this necessary? We already have this inside the `wait_until_recovery_deferral_expires` function.

crates/walrus-service/src/node.rs
```diff
@@ -2204,10 +2241,170 @@ impl StorageNode {
...
+    /// Returns the latest checkpoint sequence number known to the node.
+    pub fn latest_checkpoint_sequence_number(&self) -> Option<u64> {
+        self.event_manager.latest_checkpoint_sequence_number()
+    }
+
+    /// Applies a recovery deferral if the node determines the blob is part of an in-flight upload.
+    pub async fn maybe_apply_live_upload_deferral(
+        &self,
+        blob_id: BlobId,
+        checkpoint_sequence: u64,
```
mlegner, line 2287: Question: What is the `checkpoint_sequence` here?

crates/walrus-service/src/node.rs
```diff
@@ -2204,10 +2241,170 @@ impl StorageNode {
...
+
+        if checkpoint_sequence > latest_checkpoint {
+            tracing::trace!(
+                %blob_id,
+                checkpoint_sequence,
+                latest_checkpoint,
+                "checkpoint sequence ahead of latest; skipping live-upload deferral"
+            );
+            return;
+        }
```
mlegner, line 2309: Question: Can this even happen?

crates/walrus-service/src/node/config.rs
```diff
@@ -783,6 +787,74 @@ impl GarbageCollectionConfig {
...
+    #[serde_as(as = "DurationSeconds<u64>")]
+    #[serde(rename = "max_total_defer_secs")]
+    pub max_total_defer: Duration,
+    /// Maximum Sui checkpoint lag (inclusive) that still qualifies for a live-upload deferral.
+    #[serde(default = "default_max_checkpoint_lag")]
+    pub max_checkpoint_lag: u64,
+}
+
+const fn default_max_checkpoint_lag() -> u64 {
+    1
```
mlegner: Suggestion: Increase this a bit. Given we have 4-5 checkpoints per second, a lag of 10-50 should be fine.

crates/walrus-service/src/node.rs
```diff
@@ -2204,10 +2241,170 @@ impl StorageNode {
...
+            tracing::trace!(
+                %blob_id,
+                lag,
+                threshold = self.live_upload_deferral_config.max_checkpoint_lag,
+                "checkpoint lag too high; skipping live-upload deferral"
+            );
+            return;
+        }
+
+        let metadata = match self.storage.get_metadata(&blob_id) {
```
mlegner: Remark: It's a bit unfortunate that we don't have the unencoded size in the blob info...

Suggestion: Given that we only need this to compute the deferral duration, we could simply use a default value in case we don't have the metadata.

crates/walrus-service/src/node.rs
```diff
@@ -2204,10 +2241,170 @@ impl StorageNode {
...
+            .deferral_for_size(size_bytes)
+    }
+
+    /// Returns the latest checkpoint sequence number known to the node.
+    pub fn latest_checkpoint_sequence_number(&self) -> Option<u64> {
+        self.event_manager.latest_checkpoint_sequence_number()
+    }
+
+    /// Applies a recovery deferral if the node determines the blob is part of an in-flight upload.
+    pub async fn maybe_apply_live_upload_deferral(
```
mlegner, line 2284: Suggestion: Add a `tracing::instrument` here:

```suggestion
    #[tracing::instrument(skip(self))]
    pub async fn maybe_apply_live_upload_deferral(
```

Then you also don't need to explicitly record the `blob_id` in all the log events below.

crates/walrus-service/src/node.rs
```diff
@@ -2657,6 +2861,137 @@ impl StorageNodeInner {
...
+
+        Ok(())
+    }
+
+    fn set_recovery_deferral(
+        &self,
+        blob_id: BlobId,
+        defer_for: std::time::Duration,
+    ) -> impl Future<Output = Result<(), SetRecoveryDeferralError>> + Send {
+        self.set_recovery_deferral_inner(blob_id, defer_for)
```
mlegner: Question: Why is the inner function necessary here?

crates/walrus-service/src/node.rs
```diff
@@ -2657,6 +2861,137 @@ impl StorageNodeInner {
...
+    async fn recovery_deferral_cleanup_task(self: Arc<Self>) {
+        loop {
+            if self.recovery_deferral_cleanup_token.is_cancelled() {
+                break;
+            }
+
+            self.remove_expired_recovery_deferrals().await;
+
+            let sleep_duration = match self.next_recovery_deferral_deadline().await {
+                Some(deadline) => deadline.saturating_duration_since(std::time::Instant::now()),
```
mlegner, line 3159: Question: What happens if a deferral with a shorter deadline is added afterwards?

Suggestion: Use the minimum of the existing deadlines and the shortest possible deadline for new blobs.

sadhansood: done

crates/walrus-sdk/src/config.rs
sadhansood: fixed, it isn't anymore.

crates/walrus-service/src/node.rs
sadhansood: done

crates/walrus-service/src/node.rs
sadhansood: added comment down below.

crates/walrus-service/src/node/blob_sync.rs
sadhansood: done, moved before acquiring permit

crates/walrus-service/src/node/errors.rs
sadhansood: good catch, removed

crates/walrus-service/src/node/storage/shard.rs
sadhansood: This is mainly meant to clear any recovery deferral that was set up for this blob once we know that blob has been recovered.

crates/walrus-service/src/node.rs
sadhansood: yes, I added a comment below.

crates/walrus-service/src/test_utils.rs
```diff
@@ -410,6 +413,25 @@ impl SimStorageNodeHandle {
...
+                    defer: Duration::from_millis(100),
+                });
+            }
+            if cfg.live_upload_deferral.max_total_defer.is_zero() {
+                cfg.live_upload_deferral.max_total_defer = Duration::from_millis(100);
+            }
+            if cfg.live_upload_deferral.max_checkpoint_lag == 0 {
+                cfg.live_upload_deferral.max_checkpoint_lag = 32;
+            }
+        }
```
mlegner, line 417: Question: Any reason not to simply use the `default_for_test` here?

crates/walrus-service/src/node/config.rs
```diff
@@ -788,6 +792,92 @@ impl GarbageCollectionConfig {
...
+        if !self.enabled {
+            return None;
+        }
+
+        let capped = self.max_total_defer;
+        (!capped.is_zero()).then_some(capped)
+    }
+
+    /// Returns a configuration that is convenient for tests.
+    pub fn default_for_test() -> Self {
```
mlegner, line 871: Nit: This is only needed for tests:

```suggestion
    #[cfg(any(test, feature = "test-utils"))]
    pub fn default_for_test() -> Self {
```

crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -531,12 +533,44 @@ impl BlobSynchronizer {
...
         let this = Arc::new(self);
         let histograms = &this.metrics().recover_blob_part_duration_seconds;
 
+        // After deferral elapses, re-check existence at shards
+        // owned by this node in the current epoch.
+        let current_epoch = this
+            .node
+            .current_event_epoch()
+            .await
+            .expect("current event epoch should be set");
```
mlegner, line 559: Nit (pre-existing): We could maybe replace the `?` by an `expect` [here](https://github.com/MystenLabs/walrus/blob/c63b8d16f38d3c51f5af5b0ea7cfd5dce2df25a9/crates/walrus-service/src/node.rs#L2618) and change the return type of `current_event_epoch` to `Epoch`. The error should never happen as the sender of the channel is never dropped. Then we don't need the `expect`s everywhere this function is called. cc @halfprice

crates/walrus-service/src/node.rs
```diff
@@ -2235,10 +2272,157 @@ impl StorageNode {
...
+            Ok(None) => {
+                tracing::trace!(
+                    "metadata not yet stored, falling back to default live-upload deferral"
+                );
+                None
+            }
+            Err(error) => {
+                tracing::warn!(
+                    ?error,
+                    "failed to load metadata while evaluating live-upload deferral"
```
mlegner: Nit: Can we clarify that this is a DB error and not just an inexistent metadata?

crates/walrus-service/src/node.rs
```diff
@@ -2235,10 +2272,157 @@ impl StorageNode {
...
+            Err(SetRecoveryDeferralError::AlreadyStored)
+            | Err(SetRecoveryDeferralError::NotRegistered) => {
+                tracing::trace!("blob state changed before applying live-upload deferral");
+            }
+            Err(SetRecoveryDeferralError::Unsupported) => {
+                tracing::debug!("storage node does not support recovery deferrals");
+            }
+            Err(SetRecoveryDeferralError::Internal(error)) => {
+                tracing::warn!(?error, "failed to apply live-upload recovery deferral");
+            }
```
mlegner: Remark: It seems to me that this is pretty much the same case distinction made inside the `set_recovery_deferral` function. Would it make sense to move the log statements inside the function to avoid doing the same case distinction multiple times?

crates/walrus-service/src/node.rs
```diff
@@ -2717,6 +2902,155 @@ impl StorageNodeInner {
...
+
+        let until = std::time::Instant::now() + defer_duration;
+        let mut map = self.recovery_deferrals.write().await;
+        if map.contains_key(&blob_id) {
+            tracing::trace!(
+                walrus.blob_id = %blob_id,
+                "recovery deferral already exists for blob"
+            );
+            return Err(SetRecoveryDeferralError::AlreadyExists);
+        }
```
halfprice: Suggestion: update RecoveryDeferralEntry with the latest expiration time.

crates/walrus-service/src/node.rs
```diff
@@ -2717,6 +2902,155 @@ impl StorageNodeInner {
...
+        map.insert(
+            blob_id,
+            (
+                until,
+                std::sync::Arc::new(tokio_util::sync::CancellationToken::new()),
+            ),
+        );
+        let current_size = map.len();
+        self.update_recovery_deferral_size_metric(current_size);
+        drop(map);
```
halfprice: nit: instead of calling drop(map), put everything in the guard in a `{...}` block from `let mut map = self.recovery_deferrals.write().await;`, which is easier to read.

crates/walrus-service/src/node/config.rs
```diff
@@ -788,6 +792,93 @@ impl GarbageCollectionConfig {
...
+const fn default_max_checkpoint_lag() -> u64 {
+    32
+}
+
+impl Default for LiveUploadDeferralConfig {
+    fn default() -> Self {
+        Self {
+            enabled: false,
+            buckets: Vec::new(),
+            max_total_defer: Duration::from_secs(30),
```
halfprice: suggestion: 30s is a bit low. I'd say maybe 5min consider uploading a large files takes quite some time (10+ minutes)

crates/walrus-service/src/node/config.rs
```diff
@@ -788,6 +792,93 @@ impl GarbageCollectionConfig {
...
+/// Configuration that controls deferring recovery when a live client upload is likely.
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(default)]
+pub struct LiveUploadDeferralConfig {
+    /// Enables applying a deferral window based on blob size.
+    pub enabled: bool,
+    /// Lookup table from size upper-bounds to deferral durations. Earlier entries take precedence.
+    #[serde(alias = "table")]
+    pub buckets: Vec<SizeDeferralEntry>,
```
halfprice, line 818: Suggestion: document that when this is empty, max_total_defer is applied to all sizes.

crates/walrus-service/src/node/storage/shard.rs
```diff
@@ -1066,6 +1080,11 @@ impl ShardStorage {
 
                 batch.write()?;
 
+                // Clear the recovery deferral now that we have definitively synchronized the blob.
+                for blob_id in cleared_blob_ids {
+                    node.clear_recovery_deferral(&blob_id).await;
```
halfprice: Important: I don't think we should clear deferral here. Blob sync is for all shards, but here this only indicate that one shard is synced. There can be other shards that are missing.

crates/walrus-service/src/node.rs
```diff
@@ -2717,6 +2902,155 @@ impl StorageNodeInner {
...
+            // Any newly scheduled deferral notifies the waiters below, forcing us to recompute the
+            // minimum deadline so we never oversleep a shorter timeout.
+            tokio::select! {
+                _ = self.recovery_deferral_cleanup_token.cancelled() => break,
+                _ = self.recovery_deferral_notify.notified() => continue,
+                _ = tokio::time::sleep(sleep_duration) => {}
+            }
+        }
+
+        self.cancel_all_recovery_deferrals().await;
```
halfprice, line 3172: Suggestion: also disable recovery deferral in the node. Existing this loop won't cause node to stop, and future blob sync may still use deferral to start blob sync, and will never be started.

crates/walrus-service/src/node.rs
```diff
@@ -2597,6 +2777,9 @@ impl StorageNodeInner {
             .await
             .context("unable to store sliver")?;
 
+        // Cancel any pending recovery deferral for this blob now that a sliver arrived.
+        self.clear_recovery_deferral(metadata.blob_id()).await;
```
halfprice: Important: we should only cancel this if all shards are stored.

halfprice (commented):
Thanks @sadhansood ! I like the current approach a lot!

There are a few important comments, PTAL. Given that this changes blob sync behavior, can you also add a integration test? One specific case in mind is to have a node with 2 shards, that need to run recovery. There are a few test examples in src/node.rs where we setup a cluster and write data, but do not populate one storage node when client doing the upload. The invariant is that all blobs need to eventually be fully populated in all shards.

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs

crates/walrus-service/src/node/config.rs
sadhansood: done

crates/walrus-service/src/node/config.rs
sadhansood: done

crates/walrus-service/src/node/storage/shard.rs
sadhansood: You are right, removed it.

crates/walrus-service/src/node.rs
sadhansood: I am not sure I completely follow. The cleanup loop only terminates when `recovery_deferral_cleanup_token` is cancelled, and the only place we do that is during shutdown. Once `StorageNodeInner::shut_down` runs, we mark the node as shutting down, cancel the deferral task, and stop doing everything, so new blob-sync requests won’t be accepted in that state.

crates/walrus-service/src/node.rs
sadhansood: Given that sliver existence check across all shards is expensive - I decide to remove `clear_recovery_deferral` from here. I think it is not very important to do it here and deferral expiry is probably enough.

crates/walrus-service/src/node/config.rs
sadhansood: done

crates/walrus-service/src/node.rs
halfprice: Ah, make sense. If `StorageNodeInner::shut_down` is called, do we still need cancel all recovery deferrals? Node will stop in this case, correct?

crates/walrus-service/src/node.rs

crates/walrus-service/src/node/blob_sync.rs
sadhansood: Thanks @mlegner, I agree with your observation but let's do it in a different PR - since it is unrelated and this PR is already quite large.

crates/walrus-service/src/test_utils.rs
sadhansood: done

crates/walrus-service/src/node.rs
sadhansood: Yes, but i think it is better to call `cancel` on the token since we hold a reference to it.

PR #2623, opened 2025-10-13
feat(client): implement new default consistency check

mlegner:
## Description

This adds a new default consistency check that only verifies the hashes of the systematic primary slivers instead of doing a full re-encoding. This is significantly faster and only slightly reduces the security guarantees (see documentation added in
#2592).

At this point, the CLI and aggregator still perform the strict consistency check. The new default consistency check will be enabled in v1.37 through a follow-up PR.

## Test plan

Extended an existing unit test.

## Evaluation

Also ran benchmarks showing that the new consistency check is significantly faster than the strict consistency check: Decoding+verification time is reduced by more than 50% for all blob sizes between 1B and 1GiB.

lines

In addition, peak memory usage when reading a 2GiB blob is reduced from ~12.5GiB (strict consistency check) to ~4GiB (new default consistency check). (NB: The memory usage of the strict consistency check will be optimized separately.)

### DIFF
--- a/crates/walrus-core/src/encoding/blob_encoding.rs
+++ b/crates/walrus-core/src/encoding/blob_encoding.rs
@@ -220,15 +220,17 @@ impl<'a> BlobEncoder<'a> {
         self.blob.chunks(self.n_columns * self.symbol_usize())
     }
 
-    fn empty_slivers<T: EncodingAxis>(&self) -> Vec<SliverData<T>> {
+    fn empty_sliver<E: EncodingAxis>(&self, index: SliverIndex) -> SliverData<E> {
+        SliverData::<E>::new_empty(
+            self.config.n_source_symbols::<E::OrthogonalAxis>().get(),
+            self.symbol_size,
+            index,
+        )
+    }
+
+    fn empty_slivers<E: EncodingAxis>(&self) -> Vec<SliverData<E>> {
         (0..self.config.n_shards().get())
-            .map(|i| {
-                SliverData::<T>::new_empty(
-                    self.config.n_source_symbols::<T::OrthogonalAxis>().get(),
-                    self.symbol_size,
-                    SliverIndex(i),
-                )
-            })
+            .map(|i| self.empty_sliver::<E>(SliverIndex(i)))
             .collect()
     }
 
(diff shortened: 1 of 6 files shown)


crates/walrus-core/src/encoding/slivers.rs
```diff
@@ -115,17 +115,28 @@ impl<T: EncodingAxis> SliverData<T> {
             Ok(self.symbols.symbol_size()) == metadata.symbol_size(encoding_config),
             SliverVerificationError::SymbolSizeMismatch
         );
+        self.check_hash(&encoding_config_for_type, metadata)
+    }
+
+    /// Checks that the hash of the sliver matches the hash in the metadata.
+    ///
+    /// This assumes that all relevant size checks have already been performed.
+    pub fn check_hash(
```
giac-mysten, line 124: Nit: since the function has strict-ish assumptions on the inputs, and it is only used withing `walrus-core`, maybe it's worth restricting access?

```suggestion
    pub(crate) fn check_hash(
```

crates/walrus-core/src/encoding/blob_encoding.rs
```diff
@@ -247,6 +249,63 @@ impl<'a> BlobEncoder<'a> {
...
+        let n_systematic_primary_slivers = self.config.n_systematic_slivers::<Primary>().get();
+        tracing::debug!(
+            n_systematic_primary_slivers,
+            "performing default consistency check"
+        );
+        if let Some(already_verified_slivers) = already_verified_slivers {
+            assert_eq!(
+                already_verified_slivers.len(),
+                usize::from(n_systematic_primary_slivers)
+            );
```
giac-mysten, line 290: Nit: the possible panic is not documented

crates/walrus-core/src/encoding/config.rs
```diff
@@ -590,13 +603,34 @@ impl EncodingFactory for ReedSolomonEncodingConfig {
...
+                let mut already_verified_slivers = vec![false; n_systematic_slivers];
+                (
+                    self.decode(
+                        metadata.metadata().unencoded_length(),
+                        slivers.into_iter().inspect(|sliver| {
+                            let index = sliver.index.as_usize();
+                            if index < n_systematic_slivers {
+                                already_verified_slivers[index] = true;
+                            };
+                        }),
```
giac-mysten, line 618: Very nice. Made me think for a while about how to implement this 😬

mlegner:
--> #2636

PR #2625, opened 2025-10-14
feat: Add an auto tuner for optimizing upload

sadhansood:
## Description

This commit introduces a new upload `auto-tuner` to dynamically optimize concurrency, providing more consistent and efficient performance than the static `--upload-mode` settings. The tuner implements a simple algorithm: it first aggressively probes for peak network capacity by multiplicatively increasing concurrent uploads, confirms the peak with a verification step to ensure stability, and then locks concurrency at 90% of the optimal level for the remainder of the upload. This strategy allows the client to quickly saturate a connection and maintain high, stable throughput, intelligently operating within the global ceiling set by configurations like `max_data_in_flight`.

## Test plan

Numerous uploads running locally and checking upload performance.

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -71,7 +71,10 @@ use walrus_sui::{
 };
 use walrus_utils::{backoff::BackoffStrategy, metrics::Registry};
 
+mod auto_tune;
+
 use self::{
+    auto_tune::AutoTuneHandle,
     communication::NodeResult,
     refresh::{CommitteesRefresherHandle, RequestKind, are_current_previous_different},
     resource::{PriceComputation, RegisterBlobOp, ResourceManager, StoreOp},
@@ -865,16 +868,14 @@ impl<T: ReadClient> WalrusNodeClient<T> {
             .pairs_per_node(metadata.blob_id(), &pairs, &committees)
             .await;
 
-        let sliver_write_limit = self
-            .communication_limits
-            .max_concurrent_sliver_writes_for_blob_size(
-                metadata.metadata().unencoded_length(),
-                &self.encoding_config,
-                metadata.metadata().encoding_type(),
-            );
+        let (sliver_write_semaphore, auto_tune_handle) = self.build_sliver_write_throttle(
+            metadata.metadata().unencoded_length(),
+            metadata.metadata().encoding_type(),
+        );
         let comms = self.communication_factory.node_write_communications_by_id(
             &committees,
-            Arc::new(Semaphore::new(sliver_write_limit)),
+            sliver_write_semaphore,
+            auto_tune_handle,
             node_ids,
         )?;
 
@@ -1651,6 +1652,27 @@ impl<T> WalrusNodeClient<T> {
         QuiltClient::new(self, self.config.quilt_client_config.clone())
     }
 
+    fn build_sliver_write_throttle(
+        &self,
+        blob_size: u64,
+        encoding_type: EncodingType,
+    ) -> (Arc<Semaphore>, Option<AutoTuneHandle>) {
+        let initial_permits = self
+            .communication_limits
+            .max_concurrent_sliver_writes_for_blob_size(
+                blob_size,
+                &self.encoding_config,
+                encoding_type,
+            );
+        let auto_tune_handle =
+            AutoTuneHandle::new(&self.communication_limits.auto_tune, initial_permits);
+        if let Some(handle) = &auto_tune_handle {
+            (handle.semaphore(), auto_tune_handle)
+        } else {
+            (Arc::new(Semaphore::new(initial_permits)), None)
+        }
+    }
+
     /// Stores the already-encoded metadata and sliver pairs for a blob into Walrus, by sending
     /// sliver pairs to at least 2f+1 shards.
     ///
@@ -1669,18 +1691,6 @@ impl<T> WalrusNodeClient<T> {
     ) -> ClientResult<ConfirmationCertificate> {
         tracing::info!(blob_id = %metadata.blob_id(), "starting to send data to storage nodes");
         let committees = self.get_committees().await?;
-        let sliver_write_limit = self
-            .communication_limits
-            .max_concurrent_sliver_writes_for_blob_size(
-                metadata.metadata().unencoded_length(),
-                &self.encoding_config,
-                metadata.metadata().encoding_type(),
-            );
-        tracing::debug!(
-            blob_id = %metadata.blob_id(),
-            communication_limits = sliver_write_limit,
-            "establishing node communications"
-        );
 
         let progress_bar = multi_pb.map(|multi_pb| {
             let pb = styled_progress_bar(bft::min_n_correct(committees.n_shards()).get().into());
@@ -1820,22 +1830,26 @@ impl<T> WalrusNodeClient<T> {
             .map(|(metadata, _)| metadata.metadata().unencoded_length())
             .max()
             .unwrap_or(0);
+
         let encoding_type = blobs
             .first()
             .map(|(metadata, _)| metadata.metadata().encoding_type())
             .unwrap_or(DEFAULT_ENCODING);
 
-        let sliver_write_limit = self
-            .communication_limits
-            .max_concurrent_sliver_writes_for_blob_size(
-                max_unencoded,
-                &self.encoding_config,
-                encoding_type,
-            );
+        let (sliver_write_semaphore, auto_tune_handle) =
+            self.build_sliver_write_throttle(max_unencoded, encoding_type);
 
-        let comms = self
-            .communication_factory
-            .node_write_communications(&committees, Arc::new(Semaphore::new(sliver_write_limit)))?;
+        if auto_tune_handle.is_some() {
+            tracing::info!("auto tune is enabled");
+        } else {
+            tracing::info!("auto tune is disabled");
+        }
+
+        let comms = self.communication_factory.node_write_communications(
+            &committees,
+            sliver_write_semaphore,
+            auto_tune_handle,
+        )?;
 
         let sliver_write_extra_time = self
             .config
--- a/crates/walrus-sdk/src/client/auto_tune.rs
+++ b/crates/walrus-sdk/src/client/auto_tune.rs
@@ -1,0 +1,366 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+/// A simple auto-tune mechanism for the data-in-flight limit.
+///
+/// The auto-tune mechanism is based on a an algorithm which works in three phases which are:
+/// 1. Searching: The auto-tune mechanism is searching for the peak throughput. It'll increase
+///    the permits until the peak throughput is found.
+/// 2. Verifying: The auto-tune mechanism is verifying if the peak throughput is real. It'll hold
+///    the permits at the peak throughput until the throughput drops below the peak throughput.
+/// 3. Locked: The auto-tune mechanism is locked in the peak throughput. It'll hold the permits
+///    at the peak throughput.
+///
+use std::{
+    sync::{Arc, Mutex},
+    time::{Duration, Instant},
+};
+
+use tokio::sync::{
+    Semaphore,
+    mpsc::{UnboundedSender, unbounded_channel},
+};
+use walrus_core::SliverType;
+
+use crate::config::communication_config::DataInFlightAutoTuneConfig;
+
+/// Handle used by write tasks to report progress to the auto tuner.
+#[derive(Clone, Debug)]
+pub(crate) struct AutoTuneHandle {
+    inner: Arc<AutoTuneInner>,
+    semaphore: Arc<Semaphore>,
+    primary_weight: f64,
+    secondary_weight: f64,
+}
+
+impl AutoTuneHandle {
+    pub fn new(config: &DataInFlightAutoTuneConfig, allowed_permits: usize) -> Option<Self> {
+        if !config.enabled {
+            return None;
+        }
+
+        let allowed_permits = allowed_permits.max(1);
+
+        let (config_min, config_max) = config.permit_bounds();
+        let max_permits = config_max.min(allowed_permits).max(1);
+        let min_permits = config_min.min(max_permits).max(1);
+
+        if max_permits <= 1 {
+            tracing::warn!(
+                target: "walrus::auto_tune",
+                max_permits,
+                "auto tune disabled because at most one concurrent write is allowed"
+            );
+            return None;
+        }
+
+        debug_assert!(
+            min_permits < max_permits,
+            "min_permits should be < max_permits"
+        );
+
+        let sample_target = config.sample_target();
+        let window_timeout = config.timeout();
+        let increase_factor = config.increase_factor();
+        let lock_factor = config.lock_factor();
+        let primary_weight = 1.0_f64;
+        let secondary_weight = config.secondary_sliver_weight().min(primary_weight);
+
+        let initial_permits = min_permits;
+        let state = AutoTuneState::new(
+            initial_permits,
+            min_permits,
+            max_permits,
+            sample_target,
+            window_timeout,
+            increase_factor,
+            lock_factor,
+        );
+        let semaphore = Arc::new(Semaphore::new(initial_permits));
+
+        Some(Self {
+            inner: Arc::new(AutoTuneInner::new(semaphore.clone(), state)),
+            semaphore,
+            primary_weight,
+            secondary_weight,
+        })
+    }
+
+    pub fn semaphore(&self) -> Arc<Semaphore> {
+        self.semaphore.clone()
+    }
+
+    pub fn record_success(&self, bytes: usize, sliver_type: SliverType, completed_at: Instant) {
+        let weight = self.weight_for(sliver_type);
+        let mut state = self
+            .inner
+            .state
+            .lock()
+            .expect("mutex should not be poisoned");
+
+        state.observe(bytes, weight, completed_at);
+        let adjustment = self.inner.compute_adjustment(&mut state);
+
+        drop(state);
+        self.inner.apply_adjustment(adjustment);
+    }
+
+    fn weight_for(&self, sliver_type: SliverType) -> f64 {
+        match sliver_type {
+            SliverType::Primary => self.primary_weight,
+            SliverType::Secondary => self.secondary_weight,
+        }
+    }
+}
+
+#[derive(Debug)]
+struct AutoTuneInner {
+    state: Mutex<AutoTuneState>,
+    adjust_tx: UnboundedSender<AdjustmentKind>,
+}
+
+impl AutoTuneInner {
+    fn new(semaphore: Arc<Semaphore>, state: AutoTuneState) -> Self {
+        let (adjust_tx, mut adjust_rx) = unbounded_channel::<AdjustmentKind>();
+        tokio::spawn(async move {
+            while let Some(kind) = adjust_rx.recv().await {
+                match kind {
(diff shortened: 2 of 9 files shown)


crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -50,6 +61,12 @@ pub struct ClientCommunicationConfig {
     pub disable_native_certs: bool,
     /// The extra time allowed for sliver writes.
     pub sliver_write_extra_time: SliverWriteExtraTime,
+    /// Optional limit under which the client skips the sliver status check before uploads.
+    ///
+    /// `None` disables the check entirely.
+    pub sliver_status_check_threshold: Option<usize>,
```
mlegner: Question: Is this a size in bytes of the sliver?

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -209,10 +233,30 @@ impl CommunicationLimits {
...
-            self.sliver_size_for_blob(blob_size, encoding_config, encoding_type),
-            self.max_concurrent_writes,
-        )
+        let sliver_size = self.sliver_size_for_blob(blob_size, encoding_config, encoding_type);
+        let effective_max_data_in_flight = self.max_data_in_flight.max(sliver_size.get());
+        let allowed_by_bytes = effective_max_data_in_flight / sliver_size.get();
+        let limited_by_connections = allowed_by_bytes >= self.max_concurrent_writes;
+        let concurrency = allowed_by_bytes.min(self.max_concurrent_writes).max(1);
+
+        tracing::info!(
```
mlegner: Nit: This is probably not relevant for the standard user and could be a `debug` log instead.

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,284 @@
...
+};
+
+use crate::config::communication_config::DataInFlightAutoTuneConfig;
+
+/// The duration of each measurement window.
+const WINDOW_DURATION: Duration = Duration::from_secs(10);
+/// The multiplicative factor for increasing permits during the search phase.
+const INCREASE_FACTOR: f64 = 1.5;
+/// The factor to apply to the best-performing permit count to find the stable, locked-in value.
+const LOCK_FACTOR: f64 = 0.9;
```
mlegner: Question/suggestion: Could these be configuration parameters in the `DataInFlightAutoTuneConfig`?

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,284 @@
...
+        min_permits: usize,
+        max_permits: usize,
+    ) -> Option<Self> {
+        if !config.enabled {
+            return None;
+        }
+
+        let min_permits = min_permits.max(1);
+        if max_permits <= min_permits {
+            return None;
```
mlegner: Nit: This seems like a configuration error and could result in a panic or at least a warning log message.

crates/walrus-sdk/src/client/communication/node.rs
```diff
@@ -28,22 +32,16 @@ use walrus_sui::types::StorageNode;
...
+/// directly. The default value still follows the TCP congestion-window heuristic documented here,
+/// but operators can override it through `ClientCommunicationConfig`.
 ///
-/// The threshold is chosend in a somewhat arbitrary way, but with the guiding principle that the
-/// direct sliver store should only take 1 RTT, therefore having similar latency to the sliver
-/// status check. To ensure this is the case, we take compute the threshold as follows: take the TCP
-/// payload size (1440 B); multiply it for an initial congestion window of 4 packets (although in
-/// modern systems this is usually 10, there may be other data being sent in this window); and
-/// conservatively subtract 200 B to account for HTTP headers and other overheads.
-const SLIVER_CHECK_THRESHOLD: usize = 5560;
```
mlegner: Remark: I think the remaining docstring is now associated with the `NodeIndex`. Also, can you copy the explanation to `crates/walrus-sdk/src/config/communication_config.rs`?

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,284 @@
...
+const INCREASE_FACTOR: f64 = 1.5;
+/// The factor to apply to the best-performing permit count to find the stable, locked-in value.
+const LOCK_FACTOR: f64 = 0.9;
+
+/// Handle used by write tasks to report progress to the auto tuner.
+#[derive(Clone)]
+pub(crate) struct AutoTuneHandle {
+    inner: Arc<AutoTuneInner>,
+    semaphore: Arc<Semaphore>,
+    initial_permits: usize,
```
mlegner: Question: Is my understanding correct that the permits apply to both primary and secondary slivers in the same way? This may slightly decrease the effectiveness of the algorithm as primary slivers are twice as large as secondary ones.

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,284 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+/// A simple auto-tune mechanism for the data-in-flight limit.
+///
+/// The auto-tune mechanism is based on a an algorithm which works in three phases which are:
+/// 1. Searching: The auto-tune mechanism is searching for the peak throughput. It'll increase
+///    the permits until the peak throughput is found.
+/// 2. Verifying: The auto-tune mechanism is verifying if the peak throughput is real. It'll hold
+///    the permits at the peak throughput until the throughput drops below the peak throughput.
```
mlegner, line 10: Question: What happens if the initial permits are set too high? My current understanding is that the algorithm doesn't really work in that case. If this is true, we should either set the initial number of permits so low that it's never too high. Or modify the algorithm to handle that case.

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+        self.window_start_time = Instant::now();
+        self.window_bytes = 0;
+        self.last_throughput = current_throughput;
+    }
+
+    fn search(&mut self, current_throughput: f64) {
+        if current_throughput >= self.last_throughput {
+            if current_throughput > self.max_observed_throughput {
```
mlegner, line 310: Important: AFAIU, this will always be true after the first measurement. So even if the initial permits are set too high, we will always further increase them.

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,284 @@
...
+
+use tokio::sync::{
+    Semaphore,
+    mpsc::{UnboundedSender, unbounded_channel},
+};
+
+use crate::config::communication_config::DataInFlightAutoTuneConfig;
+
+/// The duration of each measurement window.
+const WINDOW_DURATION: Duration = Duration::from_secs(10);
```
mlegner: AFAIU, the ideal window duration depends on the blob size: It must be chosen large enough that at least a few slivers are actually stored successfully, but as small as possible such that the auto-tuning can work effectively. Not sure if it's worth investing resources into optimizing this though.

crates/walrus-sdk/src/client.rs
```diff
@@ -100,6 +103,8 @@ pub mod upload_relay_client;
 /// The delay between retries when retrieving slivers.
 #[allow(unused)]
 const RETRIEVE_SLIVERS_RETRY_DELAY: Duration = Duration::from_millis(100);
+const MIN_AUTO_TUNE_SLIVER_WRITE_PERMITS: usize = 100;
+const MAX_AUTO_TUNE_SLIVER_WRITE_PERMITS: usize = 1000;
```
mlegner: Remark: For very large blobs or very bad connections, the 100 permits may still be too large.

mlegner (commented):
Really nice addition, @sadhansood.

I have a few questions and some suggestions.

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: Absolutely agree! I was thinking about using time window based on time to observe at least `x` sliver uploads or a timeout. I have added something along those lines with `window_effective_slivers` which makes observation either on a timeout or once we've observed certain number of sliver uploads.

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: yes good callout, i will make this change.

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: You’re right—the semaphore just limits the number of in-flight sliver uploads, but the auto-tuner decides how to adjust permits based on `self.window_bytes`. Because we aggregate raw bytes, larger primary slivers contribute proportionally more to the measured throughput, so the algorithm already weights them implicitly. Also, we have `window_effective_slivers` now, which closes the observation window based on number of weighted slivers where secondary sliver is 0.5 * primary_weight.

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: Yes, that is correct. I've taken your input and now we start from `min_permits` and go up.

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: Done, added warn log.

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: Yes, that is correct. Which is why now we set `initial_permits` to `min_permits` and always go up.

crates/walrus-sdk/src/config/communication_config.rs
sadhansood: Ah, this is outdated now as this was from another change.

crates/walrus-sdk/src/config/communication_config.rs
sadhansood: done

crates/walrus-sdk/src/client/communication/node.rs
sadhansood: This was from another change, i think it is not needed here.

crates/walrus-storage-node-client/src/client.rs
```diff
@@ -763,8 +763,11 @@ impl StorageNodeClient {
         let (url, template) = self.endpoints.sliver::<A>(blob_id, pair_index);
         let request = self.create_request_with_payload(Method::PUT, url, &sliver);
         self.send_and_parse_service_response::<String>(request, template)
-            .await?;
-
+            .await
+            .map_err(|error| {
+                tracing::debug!("error storing sliver: {:?}", error);
```
mlegner: Nit: I think an `inspect_err` would work here as well.

crates/walrus-sdk/src/client.rs
```diff
@@ -1642,6 +1643,31 @@ impl<T> WalrusNodeClient<T> {
...
+            AutoTuneHandle::new(&self.communication_limits.auto_tune, initial_permits);
+        let semaphore = auto_tune_handle
+            .as_ref()
+            .map(|handle| handle.semaphore())
+            .unwrap_or(Arc::new(Semaphore::new(initial_permits)));
+        let initial_permits = auto_tune_handle
+            .as_ref()
+            .map(|handle| handle.initial_permits())
+            .unwrap_or(initial_permits);
+        (semaphore, initial_permits, auto_tune_handle)
```
mlegner: Nit: IMO, it would be easier to understand with an `if let Some(auto_tune_handle) = auto_tune_handle {...} else {...}` construct.

crates/walrus-sdk/src/client/communication/node.rs
```diff
@@ -486,21 +510,72 @@ impl NodeWriteCommunication {
...
         }
 
-        self.store_sliver(blob_id, sliver, pair_index).await
+        let start = Instant::now();
+        match self.store_sliver(blob_id, sliver, pair_index).await {
+            Ok(()) => {
+                let completed_at = Instant::now();
+                if let Some(handle) = &self.auto_tune_handle {
+                    handle.record_success(sliver.len(), A::sliver_type(), completed_at);
+                }
```
mlegner: Question: Is there a reason to have this outside of `self.record_throughput_success`?

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -33,6 +35,134 @@ fn default_sliver_status_check_threshold() -> usize {
...
+    DEFAULT_AUTO_TUNE_MIN_PERMITS
+}
+
+fn default_auto_tune_max_permits() -> usize {
+    DEFAULT_AUTO_TUNE_MAX_PERMITS
+}
+
+fn default_auto_tune_secondary_weight() -> f64 {
+    DEFAULT_AUTO_TUNE_SECONDARY_WEIGHT
+}
```
mlegner: Nit: These could be moved to the `mod default` block starting in line 418.

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -33,6 +35,134 @@ fn default_sliver_status_check_threshold() -> usize {
...
+
+/// Configuration for runtime auto-tuning of the data-in-flight limit.
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(default)]
+pub struct DataInFlightAutoTuneConfig {
+    /// Enables the additive-increase / multiplicative-decrease controller.
+    pub enabled: bool,
+    /// Number of completed sliver uploads required to close a measurement window.
+    #[serde(default = "default_auto_tune_window_sample_target")]
```
mlegner: Nit: AFAIU, these individual `serde(default)` annotations are not needed if the whole struct is annotated with `serde(default)`.

crates/walrus-sdk/src/config/communication_config.rs
```diff
@@ -33,6 +35,134 @@ fn default_sliver_status_check_threshold() -> usize {
...
+    DEFAULT_AUTO_TUNE_SECONDARY_WEIGHT
+}
+
+/// Configuration for runtime auto-tuning of the data-in-flight limit.
+#[serde_as]
+#[derive(Debug, Serialize, Deserialize, Clone, PartialEq)]
+#[serde(default)]
+pub struct DataInFlightAutoTuneConfig {
+    /// Enables the additive-increase / multiplicative-decrease controller.
+    pub enabled: bool,
```
mlegner, line 79: Question/suggestion: Instead of a blanked enable/disable switch, could we define a lower limit for the blob size where the auto tune is enabled? AFAIU, it is only relevant for blobs that need more than ~10s to be uploaded, so setting a ~50MiB lower bound could be useful.

crates/walrus-sdk/src/client/auto_tune.rs
mlegner: Very nice! 🤩

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,373 @@
...
+pub(crate) struct AutoTuneHandle {
+    inner: Arc<AutoTuneInner>,
+    semaphore: Arc<Semaphore>,
+    initial_permits: usize,
+    primary_weight: f64,
+    secondary_weight: f64,
+}
+
+impl AutoTuneHandle {
+    pub fn new(config: &DataInFlightAutoTuneConfig, allowed_permits: usize) -> Option<Self> {
```
mlegner, line 37: Question: What are the intended semantics of the `allowed_permits` parameter?

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,373 @@
...
+
+        if min_permits >= max_permits {
+            tracing::warn!(
+                target: "walrus::auto_tune",
+                min_permits,
+                max_permits,
+                "auto tune disabled due to invalid permit bounds"
+            );
+            return None;
+        }
```
mlegner: Nit: With the checks above, this case should not be possible.

crates/walrus-sdk/src/client/auto_tune.rs
```diff
@@ -0,0 +1,373 @@
...
+        self.semaphore.clone()
+    }
+
+    pub fn initial_permits(&self) -> usize {
+        self.initial_permits
+    }
+
+    pub fn record_success(&self, bytes: usize, sliver_type: SliverType, completed_at: Instant) {
+        let weight = self.weight_for(sliver_type);
+        let mut state = self.inner.state.lock().expect("auto tune mutex poisoned");
```
mlegner: Nit: Normally we use a formulation describing why the expect should succeed.

```suggestion
        let mut state = self.inner.state.lock().expect("mutex should not be poisoned");
```

crates/walrus-sdk/src/config/communication_config.rs
sadhansood: done

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: `allowed_permits` is the caller imposed hard cap on concurrency for this upload context. And is usually computed from `max_data_in_flight` and `max_concurrent_writes` for the blob size.

crates/walrus-sdk/src/client.rs

crates/walrus-sdk/src/client/communication/node.rs
sadhansood: no, not really. Fixed

crates/walrus-sdk/src/config/communication_config.rs
sadhansood: done

crates/walrus-sdk/src/config/communication_config.rs
sadhansood: good call, fixed!

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: done

crates/walrus-sdk/src/client/auto_tune.rs
sadhansood: fixed

sadhansood:
Thank you @mlegner for in depth review! I've addressed all your comments and answered remaining questions. Please feel free to lmk if something still seems off (and i'll take care of it in a follow up).

halfprice:
Thanks @sadhansood for this great improvement!

I just have a question: for CLI users, will the turned parameters get reset for every new CLI invocation?

sadhansood:
> Thanks @sadhansood for this great improvement!
> 
> I just have a question: for CLI users, will the turned parameters get reset for every new CLI invocation?

@halfprice iiuc, then answer is yes - the parameters are used up during the upload and config values are used for every time `store` is called on the cli.

halfprice:
> > Thanks @sadhansood for this great improvement!
> > I just have a question: for CLI users, will the turned parameters get reset for every new CLI invocation?
> 
> @halfprice iiuc, then answer is yes - the parameters are used up during the upload and config values are used for every time `store` is called on the cli.

Does that mean that the auto tuner will be more effective if the user uploads more blob in a single `store` (comparing to multiple `store`, each upload single blob)?

crates/walrus-sdk/src/client/auto_tune.rs
mlegner: OK, this means that we should tell users to set these relatively high if they enable auto-tuning. Can you add some documentation around this in a follow-up PR?

PR #2628, opened 2025-10-15
feat: basic load shedding for aggregators

halfprice:
## Description

This PR implements the basic load shedding based on number of ongoing requests being handled by the aggregator. The load shedding only applies to aggregator specific logic.

The current concurrency limit is selected based on production numbers.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon.rs
+++ b/crates/walrus-service/src/client/daemon.rs
@@ -613,10 +613,7 @@ async fn handle_service_error(error: BoxError, service_name: &str) -> Response {
     if error.is::<Overloaded>() {
         (
             StatusCode::TOO_MANY_REQUESTS,
-            format!(
-                "the {} is receiving too many requests; please try again later",
-                service_name
-            ),
+            format!("the {service_name} is receiving too many requests; please try again later"),
         )
             .into_response()
     } else {
@@ -671,7 +668,11 @@ mod tests {
     }
 
     impl WalrusReadClient for MockSlowClient {
-        async fn read_blob(&self, _blob_id: &BlobId) -> ClientResult<Vec<u8>> {
+        async fn read_blob(
+            &self,
+            _blob_id: &BlobId,
+            _consistency_check: ConsistencyCheckType,
+        ) -> ClientResult<Vec<u8>> {
             // Increment active request counter and track max
             let current = self.active_requests.fetch_add(1, Ordering::SeqCst) + 1;
             self.max_concurrent.fetch_max(current, Ordering::SeqCst);
@@ -727,8 +728,8 @@ mod tests {
         // Get the router (without global middleware for simpler testing)
         let app = daemon.router.with_state(daemon.client);
 
-        // Create a fake blob ID for testing
-        let blob_id = "8h2jF5xHGH8sS4vVUOgbW2VcwxoK0wJPXlJviMtGHTU";
+        // Create a random blob ID for testing
+        let blob_id = walrus_core::test_utils::random_blob_id();
 
         // Launch concurrent requests
         let mut handles = vec![];


mlegner:
> > One more question: Is there a straight-forward way to include any size information in this load shedding? That is, instead of counting requests equally, count requests for large blobs multiple times?
> 
> It's hard to get that. That information only available after we fetch the metadata, which already inside the application logic. I think eventually, there needs to be resource based back pressure. This account based won't work too well: it's just a measure to prevent complete unavailable.

That totally makes sense. @giac-mysten Can you make a pass and approve the PR?

crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1875,6 +1888,14 @@ pub(crate) mod default {
     pub(crate) fn concurrent_requests_for_health() -> usize {
         60
     }
+
+    pub(crate) fn max_aggregator_concurrent_requests() -> usize {
+        256
```
giac-mysten, line 1900: Remark: these defaults look ok, if a bit high. Maybe @jpcsmith 's load tests can inform the choice?

crates/walrus-service/src/client/daemon.rs
```diff
@@ -566,18 +599,198 @@ pub(crate) async fn auth_layer(
...
+/// Returns HTTP 429 for overload errors, and HTTP 500 with error details for other errors.
+async fn handle_service_error(error: BoxError, service_name: &str) -> Response {
     if error.is::<Overloaded>() {
         (
             StatusCode::TOO_MANY_REQUESTS,
-            "the publisher is receiving too many requests; please try again later",
+            format!(
+                "the {} is receiving too many requests; please try again later",
+                service_name
+            ),
```
giac-mysten: Nit: can you use the inline formatter syntax here too?
```suggestion
            format!(
                "the {service_name} is receiving too many requests; please try again later"
            ),
```

crates/walrus-service/src/client/daemon.rs
```diff
@@ -566,18 +599,198 @@ pub(crate) async fn auth_layer(
...
+            "127.0.0.1:0".parse().unwrap(),
+            &registry,
+            &args,
+        );
+
+        // Get the router (without global middleware for simpler testing)
+        let app = daemon.router.with_state(daemon.client);
+
+        // Create a fake blob ID for testing
+        let blob_id = "8h2jF5xHGH8sS4vVUOgbW2VcwxoK0wJPXlJviMtGHTU";
```
giac-mysten: Nit:  `random_blob_id` from `test_utils` could be used here.

crates/walrus-service/src/client/daemon.rs
```diff
@@ -566,18 +599,198 @@ pub(crate) async fn auth_layer(
...
+            "Total responses should equal number of requests"
+        );
+
+        // The number of successful requests should not exceed max_concurrent + max_buffer
+        assert!(
+            success_count <= max_concurrent + max_buffer,
+            "Success count {} should not exceed max_concurrent + max_buffer = {}",
+            success_count,
+            max_concurrent + max_buffer
+        );
```
giac-mysten, line 786: Remark: awesome tests! really like this.

Question: why would this not be `==`? What could cause the flakyness?

giac-mysten (approved):
This looks excellent, thanks @halfprice! Only a couple of comments. 
I agree with @mlegner that a blob-size-accounting approach would be preferrable to the current request-count one. However, this would need to reach much deeper into the client logic.

crates/walrus-service/src/client/daemon.rs
halfprice: Cool! Thanks for the pointer!

crates/walrus-service/src/client/daemon.rs
halfprice: Good question. Half of the time, the success count cannot match max_concurrent and max_buffer, and I didn't spend too much time digging, as long as rate limiting is happening.

crates/walrus-service/src/client/cli/args.rs
halfprice: This number is picked by looking at historical data. I think typically, the concurrency is less than 5. It has been jumped to 120 in a few spots, and the service is running fine. During the incident, it went up to 40K, so I just picked a random between 100 and 40K. We can adjust this after watching some production number.

PR #2629, opened 2025-10-15
feat: SingleClientWorkload allow repeatedly upload the same data

halfprice:
## Description

This is to test ref counting in BlobInfo (mix of permanent and deletable objects) .

Resolve WAL-946

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-stress/src/single_client_workload.rs
+++ b/crates/walrus-stress/src/single_client_workload.rs
@@ -57,7 +57,7 @@ pub struct SingleClientWorkload {
     /// Whether to check the read result matches the record of writes.
     check_read_result: bool,
     /// The percentage of write operations that will write the same data.
-    write_same_data_percentage: u32,
+    write_same_data_ratio: f64,
     /// The maximum number of blobs to keep in the blob pool.
     max_blobs_in_pool: usize,
     /// The initial number of blobs to pre-create in the blob pool.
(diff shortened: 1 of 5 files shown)


crates/walrus-stress/src/single_client_workload/client_op_generator.rs
```diff
@@ -45,20 +46,23 @@ pub(crate) struct ClientOpGenerator {
     request_type_distribution: RequestTypeDistributionConfig,
     blob_generator: BlobGenerator,
     epoch_length_generator: EpochLengthGenerator,
+    write_same_data_percentage: u32,
```
mlegner: Question/nit/suggestion: AFAICT, this is only ever used as an `f64`, so we could store it already as that.

crates/walrus-stress/src/single_client_workload.rs
```diff
@@ -56,6 +56,8 @@ pub struct SingleClientWorkload {
     target_requests_per_minute: u64,
     /// Whether to check the read result matches the record of writes.
     check_read_result: bool,
+    /// The percentage of write operations that will write the same data.
+    write_same_data_percentage: u32,
```
mlegner: Nit: A `u8` would be sufficient here.

crates/walrus-stress/src/single_client_workload/client_op_generator.rs
halfprice: Done.

crates/walrus-stress/src/single_client_workload.rs
halfprice: Changed to f64 per other comment.

halfprice:
Thanks @mlegner for the quick review! Should be a simple one.

PR #2630, opened 2025-10-15
feat: Add pending metadata and sliver cache on nodes

sadhansood:
## Description

This PR adds a cache for storing slivers for small blobs which haven't been registered yet. This allows performance improvement during uploads of small blobs.

## Test plan

Tests will added in the following PR which makes changes to the client.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -4704,6 +4704,8 @@ mod tests {
             .await?;
         assert_eq!(buffered_status, StoredOnNodeStatus::Buffered);
 
+        // This unit test bypasses the event processor, so flush manually. The registration-driven
+        // path is covered in `flush_after_registration`.
         storage_node
             .as_ref()
             .inner
@@ -8047,7 +8049,7 @@ mod tests {
             tokio::time::timeout(Duration::from_secs(5), flush_blocked.notified())
                 .await
                 .expect("cache flush should hit failpoint before certification runs");
-            tokio::time::timeout(Duration::from_millis(100), certify_started.notified())
+            tokio::time::timeout(Duration::from_secs(1), certify_started.notified())
                 .await
                 .expect_err(
                     "certified event processing should not start before pending caches are flushed",
@@ -8070,12 +8072,14 @@ mod tests {
                 .await
                 .expect("shard storage should exist");
 
-            assert!(
-                shard_storage
-                    .get_sliver(&blob_id, SliverType::Primary)
-                    .expect("sliver lookup should succeed")
-                    .is_some(),
-                "pending sliver should be flushed before certified handling"
+            let stored_sliver = shard_storage
+                .get_sliver(&blob_id, SliverType::Primary)
+                .expect("sliver lookup should succeed")
+                .expect("pending sliver should be flushed before certified handling");
+            assert_eq!(
+                stored_sliver,
+                Sliver::Primary(sliver_pair.primary.clone()),
+                "sliver contents should match the uploaded data"
             );
             assert!(
                 storage_node
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -183,6 +183,8 @@ impl BackgroundEventProcessor {
                     .await?
             }
             BackgroundTask::FlushPendingCaches { blob_id } => {
+                // Keep the worker busy until the flush completes so later events for this blob
+                // (which are routed to the same worker) never observe stale pending data.
                 self.node.flush_pending_caches_with_logging(blob_id).await
             }
         }
@@ -377,11 +379,17 @@ impl BackgroundEventProcessor {
 
 #[derive(Debug, Clone)]
 enum BlobEventProcessorImpl {
+    // Background processors that process events in parallel.
     BackgroundProcessors {
         background_processor_senders: Vec<UnboundedSender<TrackedBackgroundTask>>,
         _background_processors: Vec<Arc<JoinHandle<()>>>,
+        // The number of events pending in each background processor, shared with the corresponding
+        // `BackgroundEventProcessor`. Per-worker counters avoid contention on a single atomic when
+        // tracking pending events.
         background_per_processor_pending_event_count: Vec<PendingEventCounter>,
     },
+    // When there are no background workers, we process events sequentially to retain the previous
+    // behavior instead of parallelizing.
     SequentialProcessor {
         background_event_processor: Arc<BackgroundEventProcessor>,
     },
@@ -400,6 +408,12 @@ impl BlobEventProcessorImpl {
                 background_per_processor_pending_event_count,
                 ..
             } => {
+                // We send the event to one of the workers to process in parallel.
+                // Note that in order to remain sequential processing for the same BlobID, we always
+                // send events for the same BlobID to the same worker.
+                // Currently the number of workers is fixed through the lifetime of the node. But in
+                // case the node want to dynamically adjust the worker, we need to be careful to not
+                // break this requirement.
                 let processor_index =
                     blob_id.first_two_bytes() as usize % background_processor_senders.len();
 
@@ -411,6 +425,7 @@ impl BlobEventProcessorImpl {
 
                 let current_processor_pending_event_count =
                     background_per_processor_pending_event_count[processor_index].clone();
+                // Increment the counter and create a guard that will decrement it when dropped
                 let current_pending_event_count = current_processor_pending_event_count.inc();
                 walrus_utils::with_label!(
                     node.metrics
@@ -419,12 +434,15 @@ impl BlobEventProcessorImpl {
                 )
                 .set(<i64 as From<u32>>::from(current_pending_event_count));
 
+                // Create the guard that will automatically decrement the counter when dropped
                 let guard = PendingEventGuard::new(
                     current_processor_pending_event_count,
                     processor_index,
                     node.metrics.clone(),
                 );
 
+                // Wrap the task with the guard to track the pending event count
+                // and send it to the corresponding worker.
                 let tracked_task = TrackedBackgroundTask {
                     task,
                     _guard: guard,
@@ -577,7 +595,13 @@ impl BlobEventProcessor {
         event_handle: EventHandle,
         event: &BlobRegistered,
     ) -> anyhow::Result<()> {
-        // Registered events are marked as done immediately to keep up with registration lag.
+        // Registered event is marked as complete immediately. We need to process registered events
+        // as fast as possible to catch up to the latest event in order to not miss blob sliver
+        // uploads.
+        //
+        // If we want to do this in parallel, we shouldn't mix registered event processing with
+        // certified event processing, as certified events take longer and can block following
+        // registered events.
         let _scope = monitored_scope::monitored_scope("ProcessEvent::BlobEvent::Registered");
 
         if let Err(error) = self
@@ -615,5 +639,7 @@ impl BlobEventProcessor {
                 tokio::time::sleep(PENDING_EVENTS_POLL_INTERVAL).await;
             }
         }
+        // When using sequential processor, all events are processed in the same thread,
+        // so there are no pending events to wait for.
     }
 }
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -969,12 +969,8 @@ pub mod defaults {
     pub const PENDING_SLIVER_CACHE_MAX_BYTES: usize = 512 * 1024 * 1024;
     /// Default maximum size for a single sliver buffered in the pending sliver cache (~4 MiB).
     pub const PENDING_SLIVER_CACHE_MAX_SLIVER_BYTES: usize = 4 * 1024 * 1024;
-    /// Default time-to-live for pending uploads before eviction.
-    pub const PENDING_SLIVER_CACHE_TTL: Duration = Duration::from_secs(10);
     /// Default capacity for the pending metadata cache (number of entries).
     pub const PENDING_METADATA_CACHE_MAX_ENTRIES: usize = 512;
-    /// Default time-to-live for pending metadata before eviction.
-    pub const PENDING_METADATA_CACHE_TTL: Duration = PENDING_SLIVER_CACHE_TTL;
 
     /// Returns the default metrics port.
     pub fn metrics_port() -> u16 {
@@ -1011,9 +1007,14 @@ pub mod defaults {
         PENDING_SLIVER_CACHE_MAX_SLIVER_BYTES
     }
 
-    /// Returns the default time-to-live for pending uploads.
-    pub const fn pending_sliver_cache_ttl() -> Duration {
-        PENDING_SLIVER_CACHE_TTL
+    /// Returns the default time-to-live for pending uploads. Defaults to 0 to keep cache rollout
+    /// gated; tests keep caching enabled to exercise the code paths.
+    pub fn pending_sliver_cache_ttl() -> Duration {
+        if cfg!(any(test, feature = "test-utils")) {
+            Duration::from_secs(10)
+        } else {
+            Duration::from_secs(0)
+        }
     }
 
     /// Returns the default maximum number of metadata entries retained in the cache.
@@ -1021,9 +1022,9 @@ pub mod defaults {
         PENDING_METADATA_CACHE_MAX_ENTRIES
     }
 
-    /// Returns the default time-to-live for pending metadata uploads.
-    pub const fn pending_metadata_cache_ttl() -> Duration {
-        PENDING_METADATA_CACHE_TTL
+    /// Returns the default time-to-live for pending metadata uploads (matches sliver cache TTL).
+    pub fn pending_metadata_cache_ttl() -> Duration {
+        pending_sliver_cache_ttl()
     }
 
     /// Returns the default network ([`SuiNetwork::Devnet`])


crates/walrus-service/storage_openapi.yaml
```diff
@@ -218,7 +224,7 @@ paths:
               schema:
                 $ref: '#/components/schemas/ApiSuccess_String'
         '400':
-          description: May be returned when (1) Storing the metadata cannot be completed because the blob has been marked as invalid by the system. (2) The blob has not been registered or has already expired. (3) The provided metadata is not valid for the blob.
+          description: May be returned when (1) Storing the metadata cannot be completed because the blob has been marked as invalid by the system. (2) The blob has not been registered or has already expired. (3) The node refused the metadata upload because the pending cache is full. (4) The provided metadata is not valid for the blob.
```
mlegner, line 241: Nit (preexisting): We should probably differentiate the description of 400 errors for the different endpoints.

crates/walrus-service/src/node/config.rs
```diff
@@ -806,6 +837,14 @@ pub mod defaults {
...
     /// The default number of max concurrent streams for the rest API.
     pub const REST_HTTP2_MAX_CONCURRENT_STREAMS: u32 = 1000;
+    /// Default capacity for the pending sliver cache (number of slivers).
+    pub const PENDING_SLIVER_CACHE_MAX_SLIVERS: usize = 4_096;
+    /// Default byte capacity for the pending sliver cache (~512 MiB).
+    pub const PENDING_SLIVER_CACHE_MAX_BYTES: usize = 512 * 1024 * 1024;
+    /// Default maximum size for a single sliver buffered in the pending sliver cache (~4 MiB).
+    pub const PENDING_SLIVER_CACHE_MAX_SLIVER_BYTES: usize = 4 * 1024 * 1024;
+    /// Default time-to-live for pending uploads before eviction.
+    pub const PENDING_SLIVER_CACHE_TTL: Duration = Duration::from_secs(5 * 60);
```
mlegner: Given that blob registration takes on the order of 1 second, I'd suggest a much shorter default TTL for the cache, on the order of 10 seconds.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -424,6 +424,20 @@ impl BlobEventProcessor {
             // certified event processing, as certified events take longer and can block following
             // registered events.
             let _scope = monitored_scope::monitored_scope("ProcessEvent::BlobEvent::Registered");
+            if let Err(error) = self.node.flush_pending_metadata(&event.blob_id).await {
+                tracing::error!(
```
mlegner: Question: Is my understanding correct that this isn't a major problem just an "inconvenience"? If so, I'd change this to a warning (also below).

crates/walrus-service/src/node/errors.rs
```diff
@@ -161,6 +161,11 @@ pub enum ComputeStorageConfirmationError {
     #[rest_api_error(reason = "NOT_REGISTERED", status = ApiStatusCode::FailedPrecondition)]
     NotCurrentlyRegistered,
 
+    /// The pending upload cache is full and cannot accept additional metadata.
+    #[error("pending upload cache is saturated; retry once the node catches up")]
+    #[rest_api_error(reason = "CACHE_SATURATED", status = ApiStatusCode::FailedPrecondition)]
+    CacheSaturated,
+
```
mlegner: Question: Why is this relevant for the storage confirmation?

crates/walrus-service/src/node/config.rs
```diff
@@ -594,6 +598,33 @@ impl BlobRecoveryConfig {
...
 
+/// Configuration for the pending sliver cache.
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(default)]
+pub struct PendingSliverCacheConfig {
+    /// Maximum number of slivers retained in memory while the blob registration is pending.
+    pub max_cached_slivers: usize,
+    /// Maximum number of bytes retained in memory while the blob registration is pending.
+    pub max_cached_bytes: usize,
```
mlegner: Question: Is this only for the sliver cache or for both slivers and metadata?

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -104,7 +117,14 @@ impl<T: ServiceState + Send + Sync + 'static> SyncServiceState for T {}
 #[utoipa::path(
     get,
     path = METADATA_ENDPOINT,
-    params(("blob_id" = BlobId,)),
+    params(
+        ("blob_id" = BlobId, ),
+        (
```
mlegner: Suggestion: Add `derive(utoipa::ToSchema)` and `#[into_params(parameter_in = Query, style = Form)]`, then you can use this here simply as `UploadIntentQuery`. See the `ReadOptions` in `crates/walrus-service/src/client/daemon/routes.rs` for reference.

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -263,13 +299,23 @@ pub async fn put_sliver<S: SyncServiceState>(
...
+        .store_sliver(blob_id, sliver_pair_index, sliver, intent)
         .await?;
 
-    // TODO(WAL-253): Change to CREATED.
-    Ok(ApiSuccess::ok("sliver stored successfully"))
+    let (code, message) = if stored {
+        if intent.is_pending() {
+            (StatusCode::ACCEPTED, "sliver buffered pending registration")
+        } else {
+            (StatusCode::OK, "sliver stored successfully")
```
mlegner: Remark: We should either keep the TODO or change this to `CREATED`.

crates/walrus-service/src/node/config.rs
```diff
@@ -594,6 +598,33 @@ impl BlobRecoveryConfig {
     }
 }
 
+/// Configuration for the pending sliver cache.
+#[serde_as]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[serde(default)]
+pub struct PendingSliverCacheConfig {
+    /// Maximum number of slivers retained in memory while the blob registration is pending.
+    pub max_cached_slivers: usize,
```
mlegner, line 615: AFAICT, this is also used to limit the entries in the metadata cache, which I find not very intuitive. 

Suggestion: Add a separate `max_metadata_cache_entries` (and use a default value which is smaller by a factor of ~10).

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,473 @@
...
+        Self {
+            sliver_pair_index,
+            sliver_type,
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+struct CachedSliverEntry {
+    inserted_at: Instant,
```
mlegner: Suggestion: Put this on the `PendingBlobEntry` instead. I don't think it makes sense to evict only *some* slivers for a particular blob.

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,473 @@
...
+
+    pub async fn insert_many(&self, blob_id: BlobId, slivers: Vec<CachedSliver>) {
+        if slivers.is_empty() {
+            return;
+        }
+
+        let mut inner = self.inner.lock().await;
+        if inner.insert_many(blob_id, slivers).is_err() {
+            tracing::warn!(
+                ?blob_id,
```
mlegner: Nit: Use `%blob_id` here.

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,150 @@
...
+            .map(|cached| cached.metadata.clone())
+    }
+
+    fn insert(
+        &mut self,
+        blob_id: BlobId,
+        metadata: Arc<VerifiedBlobMetadataWithId>,
+    ) -> Result<(), ()> {
+        let now = Instant::now();
+        self.evict_expired(now);
```
mlegner: Suggestion: To avoid unnecessary calls, only call this if the entries limit is reached when attempting to insert a new entry. (Same suggestion for the sliver cache.)

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,150 @@
...
+                blob_id,
+                CachedMetadata {
+                    inserted_at: now,
+                    metadata,
+                },
+            );
+            return Ok(());
+        }
+
+        if self.max_entries == 0 || self.entries.len() >= self.max_entries {
```
mlegner: Nit: The first condition is implied by the second.

crates/walrus-service/src/node/pending_metadata_cache.rs
mlegner, line 1: Question: Have you considered using [moka](https://github.com/moka-rs/moka) for the cache? We already use that for the recovery symbols cache (in `crates/walrus-service/src/node/recovery_symbol_service.rs`).

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,150 @@
...
+        if self.entries.contains_key(&blob_id) {
+            self.entries.insert(
+                blob_id,
+                CachedMetadata {
+                    inserted_at: now,
+                    metadata,
+                },
+            );
+            return Ok(());
+        }
```
mlegner: Nit: The following seems more idiomatic to me:

```suggestion
        if let Entry::Occupied(mut entry) = self.entries.entry(blob_id) {
            entry.insert(CachedMetadata {
                inserted_at: now,
                metadata,
            });
            return Ok(());
        }
```

Potentially, this could also be a match statement where the code below is added to the `Vacant` case.

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,150 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! In-memory cache for metadata awaiting blob registration.
+
+use std::{
+    sync::Arc,
+    time::{Duration, Instant},
+};
+
+use indexmap::IndexMap;
```
mlegner: Question: Why do we need an `IndexMap` here? This causes some inefficiencies with the used `shift_remove`.

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,473 @@
...
+            removed_bytes += entry_bytes;
+            !entry.slivers.is_empty()
+        });
+
+        if removed_slivers > 0 {
+            self.sliver_count = self.sliver_count.saturating_sub(removed_slivers);
+        }
+        if removed_bytes > 0 {
+            self.total_bytes = self.total_bytes.saturating_sub(removed_bytes);
+        }
```
mlegner: Question: Why do we need the separate `removed_*` variables? Couldn't we do the subtraction directly in the loop?

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,150 @@
...
+            .retain(|_, cached| now.saturating_duration_since(cached.inserted_at) < ttl);
+    }
+
+    fn get(&self, blob_id: &BlobId) -> Option<Arc<VerifiedBlobMetadataWithId>> {
+        self.entries
+            .get(blob_id)
+            .map(|cached| cached.metadata.clone())
+    }
+
+    fn insert(
```
mlegner, line 62: Suggestion: Can you document in what cases an error is returned here and for similar functions?

crates/walrus-service/src/node.rs
```diff
@@ -2835,53 +2973,100 @@ impl ServiceState for StorageNodeInner {
...
-            .get_blob_info(metadata.blob_id())
-            .context("could not retrieve blob info")?
-        else {
-            return Err(StoreMetadataError::NotCurrentlyRegistered);
-        };
+            .has_metadata(&blob_id)
+            .context("could not check metadata existence")?
+        {
+            return Ok(false);
+        }
```
mlegner: Suggestion: As the function `has_metadata` internally also gets the blob info, I'd suggest to keep the code to attempt getting the blob info here (instead of a few lines below) and perform further checks on it.

crates/walrus-service/src/node.rs
```diff
@@ -2835,53 +2973,100 @@ impl ServiceState for StorageNodeInner {
     async fn store_metadata(
```
mlegner, line 3654: Nit (preexisting): I think it would be good to document the semantics of the return value.

crates/walrus-service/src/node.rs
```diff
@@ -2835,53 +2973,100 @@ impl ServiceState for StorageNodeInner {
...
+        );
 
-        self.storage
-            .put_verified_metadata(&verified_metadata_with_id)
-            .await
-            .context("unable to store metadata")?;
+        let blob_info = self
+            .storage
+            .get_blob_info(&blob_id)
+            .context("could not retrieve blob info")?;
```
mlegner: Question: Why do we get the blob info again?

crates/walrus-service/src/node.rs
mlegner, line 3654: Remark: I have some trouble following the logic of this function now. It would be helpful if it were split into some dedicated smaller functions or structured with additional comments.

crates/walrus-service/src/node.rs
```diff
@@ -2835,53 +2973,100 @@ impl ServiceState for StorageNodeInner {
...
+            .context("could not check metadata existence")?
+        {
+            return Ok(false);
+        }
 
-        if let Some(event) = blob_info.invalidation_event() {
-            return Err(StoreMetadataError::InvalidBlob(event));
+        if self.pending_metadata_cache.get(&blob_id).await.is_some() && intent.is_pending() {
+            return Ok(false);
         }
```
mlegner: Can you explain the reasoning here? Could we not simply attempt to (over-)write the metadata to the cache?

crates/walrus-service/src/node.rs
```diff
@@ -2835,53 +2973,100 @@ impl ServiceState for StorageNodeInner {
...
+                return Err(StoreMetadataError::NotCurrentlyRegistered);
+            }
+
+            self.pending_metadata_cache
+                .insert(blob_id, verified)
+                .await
+                .map_err(|_| StoreMetadataError::CacheSaturated)?;
+
+            if self.is_blob_registered(&blob_id)?
+                && let Err(error) = self.flush_pending_metadata(&blob_id).await
```
mlegner: Question: Why is this needed? We already checked if the blob is registered a few lines above.

Isn't it sufficient to flush the cache in the `compute_storage_confirmation` function?

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3113,119 @@ impl ServiceState for StorageNodeInner {
...
+                    .await
+                    .map(|arc| (*arc).clone())
+                    .ok_or(StoreSliverError::MissingMetadata)?
+            }
+            (None, false, true) => self
+                .pending_metadata_cache
+                .get(&blob_id)
+                .await
+                .map(|arc| (*arc).clone())
+                .ok_or(StoreSliverError::MissingMetadata)?,
```
mlegner: Suggestion (nit): Combine these two cases.

```suggestion
            (_, true, _) | (_, _, true) => self
                .pending_metadata_cache
                .get(&blob_id)
                .await
                .map(|arc| (*arc).clone())
                .ok_or(StoreSliverError::MissingMetadata)?,
```

crates/walrus-service/src/node.rs
mlegner, line 3761: Similar to the `store_metadata` function above, this function is now pretty complex and would benefit from some substructure and/or comments.

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3113,119 @@ impl ServiceState for StorageNodeInner {
...
+        let max_sliver_bytes = self.pending_sliver_cache.max_sliver_bytes();
+        if max_sliver_bytes == 0 || sliver.len() > max_sliver_bytes {
+            tracing::debug!(
+                blob_id = %blob_id,
+                sliver_len = sliver.len(),
+                max_sliver_bytes,
+                "pending sliver too large for cache; require registration first"
+            );
+            return Err(StoreSliverError::NotCurrentlyRegistered);
+        }
```
mlegner: Question: Isn't this already checked by the cache internally? So we could just try adding it to the cache and map the error here.

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3113,119 @@ impl ServiceState for StorageNodeInner {
...
+            .context("database error when checking sliver existence")?
+        {
+            return Ok(false);
+        }
+
+        drop(shard_storage);
+
+        let (_, verified_sliver) = self
+            .verify_sliver_against_metadata(metadata, sliver)
+            .await?;
```
mlegner: Suggestion: Put this into a separate function as this is duplicated 1:1 with the `store_sliver_unchecked` function.

crates/walrus-service/src/node.rs
```diff
@@ -2963,6 +3238,27 @@ impl ServiceState for StorageNodeInner {
             ComputeStorageConfirmationError::NotCurrentlyRegistered,
         );
 
+        if let Err(error) = self.flush_pending_metadata(blob_id).await {
+            return Err(match error {
+                StoreMetadataError::NotCurrentlyRegistered => {
+                    ComputeStorageConfirmationError::NotCurrentlyRegistered
+                }
```
mlegner: Remark: AFAICT, this error is never returned by the `flush_pending_metadata` function.

crates/walrus-service/src/node.rs
```diff
@@ -2963,6 +3238,27 @@ impl ServiceState for StorageNodeInner {
...
+        }
+
+        if let Err(error) = self.flush_pending_slivers(blob_id).await {
+            return Err(match error {
+                StoreSliverError::NotCurrentlyRegistered | StoreSliverError::MissingMetadata => {
+                    ComputeStorageConfirmationError::NotCurrentlyRegistered
+                }
+                StoreSliverError::ShardNotAssigned(inner) => {
+                    ComputeStorageConfirmationError::Internal(inner.into())
+                }
```
mlegner: Remark: AFAICT, these errors are never returned by the `flush_pending_slivers` function.

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3113,119 @@ impl ServiceState for StorageNodeInner {
...
+
+        if self.is_blob_registered(&blob_id)?
+            && let Err(error) = self.flush_pending_slivers(&blob_id).await
+            && !matches!(
+                error,
+                StoreSliverError::NotCurrentlyRegistered | StoreSliverError::MissingMetadata
+            )
+        {
+            return Err(error);
+        }
```
mlegner: As above: Why is this code needed?

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,473 @@
...
+                let base_bytes = self.total_bytes.checked_sub(previous_len).ok_or(())?;
+                let new_total_bytes = base_bytes.checked_add(new_len).ok_or(())?;
+                if new_total_bytes > self.max_bytes {
+                    Err(())
+                } else {
+                    previous.sliver = sliver;
+                    previous.inserted_at = now;
+                    self.total_bytes = new_total_bytes;
+                    Ok(false)
+                }
```
mlegner: Remark: IIUC, we only ever write verified slivers to the cache, in which case their length should always be identical.

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,473 @@
...
+                        self.sliver_count += 1;
+                        self.total_bytes = new_total_bytes;
+                        Ok(true)
+                    }
+                }
+            }
+        };
+
+        // Always put the potentially modified entry back.
+        self.blobs.insert(blob_id, entry);
```
mlegner: Question: Why is this needed? AFAIU, the main point of using the `.entry` function is to be able to modify an entry in place (without having to explicitly insert it).

mlegner (commented):
Thank you very much for implementing this cache, this will be very useful to reduce the upload latency. I have a number of detailed questions and small suggestions.

Also, it might be useful for @liquid-helium to do a review as he already had a similar functionality drafted in #2446.

crates/walrus-service/storage_openapi.yaml
sadhansood: fixed

crates/walrus-service/src/node/config.rs
sadhansood: done, idk why i set it to 5 minutes but thanks for spotting it!

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: yes, that is correct - switched to warn!

crates/walrus-service/src/node/errors.rs
sadhansood: it is not, removed.

crates/walrus-service/src/node/config.rs
sadhansood: This is only for sliver cache, there is no size bound for metadata cache (only number of entries)

crates/walrus-service/src/node/server/routes.rs
sadhansood: done, thanks

crates/walrus-service/src/node/server/routes.rs
sadhansood: fixed

crates/walrus-service/src/node/config.rs
sadhansood: make sense, to keep power of 2 - i now use `PENDING_SLIVER_CACHE_MAX_SLIVERS/8`

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: make sense, done

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: done

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: done, good call

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: fixed

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: I looked at `Moka` but decided against using it. The pending caches need blob-scoped batching i.e. we store hashmap of `blobid -> hash_map of (sliverkey, sliver)` which is mutable. In moka, it would not be possible to do so i.e. cache entries have to be immutable. And then eviction would not cleanly evict all slivers of a blob. Also, returning cache saturation error instead of evicting older entries seems harder to do cleanly in `moka`. I think our simple solution is probably better overall and exactly does what we want it to do.

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: done!

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: yeah good point, i agree. Removed it.

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: done

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: done

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs
sadhansood: done

crates/walrus-service/src/node.rs
sadhansood: I added a comment about the race condition 
```
if self.is_blob_registered(&blob_id)? {
            // Read the blob info again because registration can race between the initial blob_info
            // check and adding to the cache. Flush immediately in that case so the pending metadata
            // doesn’t linger indefinitely.
            if let Err(error) = self.flush_pending_caches(&blob_id).await {
                return Err(match error {
                    PendingCacheError::Metadata(inner) => inner,
                    PendingCacheError::Sliver(inner) => map_sliver_error_to_metadata(inner),
                });
            }
        }
``` 
Lmk if you disagree.

crates/walrus-service/src/node.rs
sadhansood: refactored

crates/walrus-service/src/node.rs
sadhansood: good point, done

crates/walrus-service/src/node.rs
sadhansood: I think this is mostly to avoid the scenario when `compute_storage_confirmation` doesn't arrive on a particular node in which case even though the node has the data, it could end up throwing it away from pending caches eventually because of the race condition which I'll try to summarize in simple terms: 
```
1. A client uploads metadata or slivers with `pending=true`. The node checks on-chain state and sees the blob isn’t registered yet, so it queues the payload in the in-memory cache.
2. Right after that check, but before we finish queueing the data in pending caches, the chain’s “blob registered” event arrives. The event handler flushes anything already in the cache, but our new payload hasn’t been inserted yet.
3. We finish inserting the payload, registration is now true, and—if we didn’t check again—it would just sit in the pending cache forever because the event handler already ran and there might never be a confirmation call to trigger another flush.
```

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs
sadhansood: done

crates/walrus-service/src/node.rs
sadhansood: good point, done

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs
sadhansood: fixed it

crates/walrus-service/src/node.rs
sadhansood: fixed it, we just map the error into `ComputeStorageConfirmationError::Internal` now.

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: fixed

crates/walrus-service/src/node.rs
sadhansood: I believe, it was the same reason about the race condition i mentioned above.

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+                Ok(())
+            }
+            Entry::Vacant(entry) => {
+                entry.insert(CachedMetadata {
+                    inserted_at: now,
+                    metadata,
+                });
+                Ok(())
+            }
+        }
```
halfprice: question: why does this read and modify? Seems can directly insert into the HashMap?

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,153 @@
...
+
+    fn len(&self) -> usize {
+        self.entries.len()
+    }
+}
+
+/// Metadata cached prior to registration.
+#[derive(Debug)]
+pub(crate) struct PendingMetadataCache {
+    inner: Mutex<PendingMetadataCacheInner>,
```
halfprice: remark: the Mutex here will serialize all the reads and writes, which I'm worried going to have bad performance impact, even across blobs. At least in MetadataCache, you can use Moka cache, since the entries are not muted?

crates/walrus-service/src/node.rs
```diff
@@ -2832,54 +3043,71 @@ impl ServiceState for StorageNodeInner {
...
+        self.pending_metadata_cache
+            .insert(blob_id, verified)
             .await
-            .context("unable to store metadata")?;
-
-        self.metrics
-            .uploaded_metadata_unencoded_blob_bytes
-            .observe(verified_metadata_with_id.as_ref().unencoded_length() as f64);
-        self.metrics.metadata_stored_total.inc();
+            .map_err(|_| StoreMetadataError::CacheSaturated)?;
```
halfprice, line 3711: question: this means that someone can send large number of cache requests without registering the blob, and cause other metadata upload to fail?

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,532 @@
...
+        }
+        Ok(())
+    }
+}
+
+/// Slivers waiting for a blob registration to be observed.
+#[derive(Debug)]
+pub(crate) struct PendingSliverCache {
+    max_sliver_bytes: usize,
+    inner: Mutex<PendingSliverCacheInner>,
```
halfprice, line 320: remark: similarly, I feel this global lock could be problematic.

What about using MokaCache<BlobId, Arc<Mutex< PendingBlobEntry >>>. I think in this case, PendingBlobEntry is mutable, and also have parallelism across blobs?

crates/walrus-service/src/node.rs
```diff
@@ -2542,6 +2603,153 @@ impl StorageNodeInner {
...
+            return Ok(false);
+        }
+
+        self.pending_metadata_cache.remove(blob_id).await;
+
+        self.storage
+            .put_verified_metadata(&verified)
+            .await
+            .context("unable to store metadata")
+            .map_err(StoreMetadataError::Internal)?;
```
halfprice: suggestion: write to DB first before removing from the cache, in case db write is not successful.

crates/walrus-service/src/node.rs
```diff
@@ -2542,6 +2603,153 @@ impl StorageNodeInner {
...
+
+        self.metrics
+            .uploaded_metadata_unencoded_blob_bytes
+            .observe(verified.metadata().unencoded_length() as f64);
+        self.metrics.metadata_stored_total.inc();
+
+        Ok(true)
+    }
+
+    async fn resolve_metadata_for_sliver(
```
halfprice, line 2892: nit: can you document the return value? It's unclear what `bool` is for by just looking at this function.

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -93,6 +95,21 @@ pub const BLOB_STATUS_ENDPOINT: &str = "/v1/blobs/{blob_id}/status";
 pub const HEALTH_ENDPOINT: &str = "/v1/health";
 pub const SYNC_SHARD_ENDPOINT: &str = "/v1/migrate/sync_shard";
 
+#[derive(Debug, Default, Deserialize, IntoParams, utoipa::ToSchema)]
+#[into_params(parameter_in = Query, style = Form)]
+#[serde(default, deny_unknown_fields)]
+pub(crate) struct UploadIntentQuery {
+    /// Set to true to buffer the payload until the blob is registered.
+    #[serde(default)]
+    pending: Option<bool>,
```
halfprice, line 98: question: because `pending` is Option, so this is not a breaking change?

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3156,80 @@ impl ServiceState for StorageNodeInner {
         blob_id: BlobId,
         sliver_pair_index: SliverPairIndex,
         sliver: Sliver,
+        intent: UploadIntent,
     ) -> Result<bool, StoreSliverError> {
         self.check_index(sliver_pair_index)?;
+        let (metadata, persisted) = self
+            .resolve_metadata_for_sliver(&blob_id, intent.is_pending())
```
halfprice, line 3770: Suggestion: instead of return error directly when no metadata is returned, we can also do a little bit of wait here. In this way, we can also parallelize metadata and sliver uploads.

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3156,80 @@ impl ServiceState for StorageNodeInner {
...
+        }
+
+        if shard_storage
+            .is_sliver_type_stored(&blob_id, sliver.r#type())
+            .context("database error when checking sliver existence")?
+        {
+            return Ok(false);
+        }
+
+        drop(shard_storage);
```
halfprice: nit: this drop seems unnecessary since shard_storage is just an Arc

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3156,80 @@ impl ServiceState for StorageNodeInner {
...
-            .ok_or(StoreSliverError::MissingMetadata)?;
-
-        // Check if encoding type is supported
         let encoding_type = metadata.metadata().encoding_type();
         if !encoding_type.is_supported() {
             return Err(StoreSliverError::UnsupportedEncodingType(encoding_type));
         }
 
-        self.store_sliver_unchecked(metadata, sliver_pair_index, sliver)
+        if persisted {
```
halfprice, line 3778: important: shouldn't the condition here be that blob is registered?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -424,6 +424,13 @@ impl BlobEventProcessor {
             // certified event processing, as certified events take longer and can block following
             // registered events.
             let _scope = monitored_scope::monitored_scope("ProcessEvent::BlobEvent::Registered");
+            if let Err(error) = self.node.flush_pending_caches(&event.blob_id).await {
```
halfprice: Important: with the locking, this for sure will affect event processing throughput. This should be done in the worker above.

halfprice (commented):
Thanks a lot @sadhansood for this important improvement! It's a quite large PR and I haven't gone through all the details.

I also have some general questions if you have thought about these:
- How should the client interact with the server? Upon server restarts, it's gonna lose all the cache data, whereas the client may think that they have successfully uploaded those data.
- When the cache is full, right now the server returns error immediately. The client may need to do a retry based on the returned status? I'm wondering if we can make this caching layer somehow transparent to the client implementation, so that you don't need to add complicated logic to deal with different errors. If all the client side needs to do is to parallelize the txn execution and upload, and the rest remain the same, that will be great. In this way, for different SDKs, we don't need to implemented the complicated error handling multiple times.

crates/walrus-service/src/node/pending_metadata_cache.rs
mlegner: OK, sounds good. 👍

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: fixed

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: replaced mutex with rwlock

crates/walrus-service/src/node.rs
sadhansood: The worst case (many unique blob IDs with pending=true) is bounded by the configured entry limit + TTL. Because each key can overwrite only its own slot, an attacker has to keep the node flooded continuously with different blobs. We log and export saturation metrics so operators can spot this, and the API keeps the same auth path as ordinary uploads. If we see it in prod we can revisit additional rate limits, but for now the combination of bounded cache + explicit back-pressure seems safe to start with. wdyt?

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: We still rely on a single guard for the inner map because we maintain global byte and sliver counters and enforce per insert invariants atomically. Each operation holds the lock just long enough to tweak bookkeeping. And then flushes drop it before touching storage, so event throughput should not get impacted much. `Moka` would complicate the size accounting quite a bit but happy to prototype that separately if contention shows up in profiling.

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs
sadhansood: Thanks for the suggestion! I’m aligned with the end goal but I’d rather keep this PR focused on landing the new pending caches. Today the sliver path either finds persisted metadata or the in-memory pending copy and succeeds immediately. Otherwise it returns `MissingMetadata`, `NotCurrentlyRegistered` and the client retries with its own back‑off. Adding a sleep loop here would change that contract: the request would tie up its executor slot waiting, we’d need to decide how long to wait, and we’d have to make sure that still plays nicely with
our timeout and/or error reporting. I started sketching a tiny retry when I read your note, but it became big quickly. Let's do it in a dedicated change later.

crates/walrus-service/src/node.rs
sadhansood: fixed

crates/walrus-service/src/node.rs
sadhansood: `persisted` comes from hitting durable storage, which only happens once we’ve seen registration. Using that flag keeps the fast path narrow, so I’ve left it as-is.

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: Keeping the flush inline gives us the minimal latency from registration to durable storage. The method drains the caches while holding their locks and releases them before any database writes, so the impact on the registered event loop should be small. Moving it to the background worker would only delay durability without reducing contention. We can revisit if contention shows up in any profile (I imagine there are other places in the event handling code we take locks for brief periods).

crates/walrus-service/src/node.rs
sadhansood: done

crates/walrus-service/src/node/server/routes.rs
sadhansood: yes

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Sorry for not being clear. Locking is just part of the problem, `flush_pending_caches ` could take long time to run is the actual problem: this could heavily limit the event processing rate.

I think not doing this inline is very important, and this comes from past experience.

Before we have event blob workers, initiating blob sync is inline in event processing, and all it does is checking if all slivers are stored. Just by doing this, we have seen that event arriving rate is higher than processing rate, and backlog starts to accumulating.

Looking at the implementation of flush_pending_caches, not only it does sliver existence check (in store_sliver_unchecked), it does a few other things too: multiple DB writes, store_sliver_unchecked uses a rate limited thread pool to write slivers, etc. I have a strong feeling that this should not be inline, unless there are experiment data showing that this is fine.

sadhansood: Cool, good point @halfprice! I've added a worker to do flushing and not block event processing in any way.

crates/walrus-service/src/node/config.rs
mlegner: Can we still make this a separate config parameter?

crates/walrus-service/src/node/config.rs
mlegner: OK, makes sense. This is not really clear from the parameter name or the docstring though.

crates/walrus-service/src/node.rs
```diff
@@ -2542,6 +2604,179 @@ impl StorageNodeInner {
...
+                StoreSliverError::NotCurrentlyRegistered | StoreSliverError::MissingMetadata
+            )
+        {
+            return Err(PendingCacheError::Sliver(error));
+        }
+
+        Ok(())
+    }
+
+    pub(crate) async fn flush_pending_caches_with_logging(&self, blob_id: BlobId) {
```
halfprice, line 3013: nit: the `_with_logging` suffix of this function is a bit weird. Why calling it out specifically?

crates/walrus-service/src/node.rs
```diff
@@ -2542,6 +2604,179 @@ impl StorageNodeInner {
...
+    ) -> Result<(Arc<VerifiedBlobMetadataWithId>, bool), StoreSliverError> {
+        let persisted = self
+            .storage
+            .get_metadata(blob_id)
+            .context("database error when storing sliver")?;
+        if let Some(metadata) = persisted {
+            return Ok((Arc::new(metadata), true));
+        }
+
+        let pending = self.pending_metadata_cache.get(blob_id).await;
```
halfprice, line 2905: nit: move this line before `let persisted =...`. It's possible that cached item is moving to the storage. So it's possible that both persisted and pending are false.

crates/walrus-service/src/node.rs
halfprice: can you document this then? I think it's important to note that blob is registered is the main condition to check here.

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3183,78 @@ impl ServiceState for StorageNodeInner {
         blob_id: BlobId,
         sliver_pair_index: SliverPairIndex,
         sliver: Sliver,
+        intent: UploadIntent,
     ) -> Result<bool, StoreSliverError> {
         self.check_index(sliver_pair_index)?;
+        let (metadata, persisted) = self
```
halfprice: nit: can we be a bit more explicit here and maybe call the variable `metadata_persisted`

crates/walrus-service/src/node.rs
```diff
@@ -3185,13 +3493,23 @@ impl ServiceState for StorageNodeInner {
...
+        match shard_storage.is_sliver_stored::<A>(blob_id) {
             Ok(true) => Ok(StoredOnNodeStatus::Stored),
-            Ok(false) => Ok(StoredOnNodeStatus::Nonexistent),
+            Ok(false) => {
+                if self
+                    .pending_sliver_cache
+                    .contains(blob_id, sliver_pair_index, A::sliver_type())
+                    .await
+                {
+                    Ok(StoredOnNodeStatus::Stored)
```
halfprice: remark: I'm not sure if we can call the sliver as stored if it is in cache, since it can be lost at any time. Why is this change needed?

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,145 @@
...
+    fn new(max_entries: usize, ttl: Duration) -> Self {
+        Self {
+            max_entries,
+            ttl,
+            entries: HashMap::new(),
+        }
+    }
+
+    fn evict_expired(&mut self, now: Instant) {
+        if self.ttl.is_zero() {
```
halfprice, line 40: Question: why not evicting if ttl is zero? Isn't ttl == 0 indicate that the cache is disabled?

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    /// Inserts metadata into the cache.
+    /// Returns `Err(())` when the cache is at capacity and the entry could not be stored after
+    /// evicting expired items.
+    fn insert(
+        &mut self,
+        blob_id: BlobId,
+        metadata: Arc<VerifiedBlobMetadataWithId>,
+    ) -> Result<(), ()> {
```
halfprice: Suggestion/nit: how about returning bool to indicating whether the metadata is added to the cache? It's a bit more explicit this way, also similar to PendingSliverCache.insert()

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,532 @@
...
+            max_bytes,
+            ttl,
+            blobs: HashMap::new(),
+            sliver_count: 0,
+            total_bytes: 0,
+        }
+    }
+
+    fn evict_expired(&mut self, now: Instant) {
+        if self.ttl.is_zero() {
```
halfprice, line 94: same question: why not evition if ttl is zero?

crates/walrus-service/src/node/pending_metadata_cache.rs
```diff
@@ -0,0 +1,145 @@
...
+            CachedMetadata {
+                inserted_at: now,
+                metadata,
+            },
+        );
+        Ok(())
+    }
+
+    fn remove(&mut self, blob_id: &BlobId) -> Option<Arc<VerifiedBlobMetadataWithId>> {
+        self.evict_expired(Instant::now());
```
halfprice: suggestion: calling eviction is a bit inconsistent (e.g., insert does not call eviction). Should we call this at the beginning of all functions?

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,532 @@
...
+    /// Returns `Err(PendingSliverCacheError::SliverTooLarge)` if the sliver exceeds the configured
+    /// per-sliver limit, or `Err(PendingSliverCacheError::Saturated)` if adding the sliver would
+    /// exceed the total capacity of the cache.
+    pub async fn insert(
+        &self,
+        blob_id: BlobId,
+        sliver_pair_index: SliverPairIndex,
+        sliver: Sliver,
+    ) -> Result<bool, PendingSliverCacheError> {
+        if self.max_sliver_bytes == 0 || sliver.len() > self.max_sliver_bytes {
```
halfprice: nit: seems self.max_sliver_bytes == 0 is unnecessary

crates/walrus-service/src/node.rs
```diff
@@ -2435,6 +2483,10 @@ impl StorageNodeInner {
 
         walrus_utils::with_label!(self.metrics.event_cursor_progress, "persisted").set(persisted);
         self.metrics.current_node_status.set(node_status.to_i64());
+        self.metrics.pending_sliver_cache_slivers.set(0);
+        self.metrics.pending_sliver_cache_blobs.set(0);
+        self.metrics.pending_sliver_cache_bytes.set(0);
+        self.metrics.pending_metadata_cache_entries.set(0);
```
halfprice: nit: this is probably not necessary.

crates/walrus-service/src/node.rs
```diff
@@ -2832,54 +3070,71 @@ impl ServiceState for StorageNodeInner {
...
+            .get_blob_info(&blob_id)
+            .context("could not retrieve blob info")?;
 
-        if let Some(event) = blob_info.invalidation_event() {
+        if let Some(event) = blob_info
+            .as_ref()
+            .and_then(|info| info.invalidation_event())
+        {
             return Err(StoreMetadataError::InvalidBlob(event));
         }
```
halfprice, line 3672: remark: return early here if metadata is already stored in this node.

crates/walrus-service/src/node.rs
```diff
@@ -2928,29 +3183,78 @@ impl ServiceState for StorageNodeInner {
...
+            .context("database error when checking sliver existence")?
+        {
+            return Ok(false);
+        }
+
+        let metadata_for_verification = metadata.as_ref().clone();
+        let (_, verified_sliver) = self
+            .verify_sliver_against_metadata(metadata_for_verification, sliver)
+            .await?;
+
```
halfprice: suggestion: this part is duplicated from `store_sliver_unchecked`. Can we extract them into a common function?

crates/walrus-service/src/node.rs
```diff
@@ -4622,14 +4944,128 @@ mod tests {
...
-            .store_metadata(blob.metadata.into_unverified())
+            .store_metadata(blob.metadata.into_unverified(), UploadIntent::Immediate)
             .await?;
 
         assert!(!is_newly_stored);
 
         Ok(())
     }
 
+    #[tokio::test]
```
halfprice: suggestion: in general, I think we should try to use simtest as much as we can (just put #[walrus_simtest] here, because tokio::test will mask any flacky failures.

halfprice (commented):
Hi @sadhansood . I've done with my full review. Most of them are minor comments. PTAL

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -153,10 +170,14 @@ pub async fn get_metadata_status<S: SyncServiceState>(
 #[utoipa::path(
     put,
     path = METADATA_ENDPOINT,
-    params(("blob_id" = BlobId,)),
-    request_body(content = [u8], description = "BCS-encoded metadata octet-stream"),
```
mlegner, line 151: Question: Why was this removed?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -430,6 +445,27 @@ impl BlobEventProcessor {
...
+                {
+                    tracing::error!(
+                        blob_id = %event.blob_id,
+                        "failed to enqueue cache flush after registration, flushing inline",
+                    );
+                    self.node
+                        .flush_pending_caches_with_logging(event.blob_id)
+                        .await;
+                }
+            }
```
mlegner: Suggestion: Can we move this into a separate function?

crates/walrus-service/src/node/pending_metadata_cache.rs
mlegner: Question: Why do we call `evict_expired` here?

crates/walrus-service/src/node/pending_sliver_cache.rs
```diff
@@ -0,0 +1,532 @@
...
+        entry.total_bytes = entry
+            .total_bytes
+            .checked_add(new_len)
+            .ok_or(PendingSliverCacheError::Saturated)?;
+        entry.slivers.insert(key, sliver);
+        Ok(true)
+    }
+
+    fn contains(&mut self, blob_id: &BlobId, key: SliverCacheKey) -> bool {
+        self.evict_expired(Instant::now());
```
mlegner: Question: Why do we call `evict_expired` here instead of just checking if the corresponding entry is non-expired? This would allow taking a non-mutable reference and make this constant-time.

Similar for the `has_blob` function below.

crates/walrus-service/src/node.rs
```diff
@@ -2602,6 +2604,179 @@ impl StorageNodeInner {
...
+            self.metrics
+                .uploaded_metadata_unencoded_blob_bytes
+                .observe(metadata.metadata().unencoded_length() as f64);
+            self.metrics.metadata_stored_total.inc();
+        }
+
+        Ok(())
+    }
+
+    pub(crate) async fn flush_pending_slivers(
```
mlegner, line 2943: Suggestion: Accept the metadata as another parameter. This should anyway only be called after the metadata is available.

crates/walrus-service/src/node.rs
```diff
@@ -2602,6 +2604,179 @@ impl StorageNodeInner {
...
+
+        Ok(())
+    }
+
+    pub(crate) async fn flush_pending_caches_with_logging(&self, blob_id: BlobId) {
+        if let Err(error) = self.flush_pending_caches(&blob_id).await {
+            match error {
+                PendingCacheError::Metadata(inner) => {
+                    tracing::warn!(
+                        ?inner,
```
mlegner: Nit: The name `inner` is a bit unclear. Can we also just call that `error`? Also below.

crates/walrus-service/src/node.rs
```diff
@@ -2602,6 +2604,179 @@ impl StorageNodeInner {
...
+            return Ok((Arc::new(metadata), true));
+        }
+
+        let pending = self.pending_metadata_cache.get(blob_id).await;
+        let is_registered = self.is_blob_registered(blob_id)?;
+
+        match (pending, is_registered, allow_pending) {
+            (Some(metadata), _, _) => Ok((metadata, false)),
+            (None, true, _) => Err(StoreSliverError::MissingMetadata),
+            (None, false, true) => Err(StoreSliverError::MissingMetadata),
```

```suggestion
            (None, true, _) | (None, _, true) => Err(StoreSliverError::MissingMetadata),
```

crates/walrus-service/src/node.rs
```diff
@@ -2988,29 +3183,78 @@ impl ServiceState for StorageNodeInner {
...
-            .ok_or(StoreSliverError::MissingMetadata)?;
-
-        // Check if encoding type is supported
         let encoding_type = metadata.metadata().encoding_type();
         if !encoding_type.is_supported() {
             return Err(StoreSliverError::UnsupportedEncodingType(encoding_type));
         }
 
-        self.store_sliver_unchecked(metadata, sliver_pair_index, sliver)
+        if persisted {
```
mlegner, line 3778: Nit: Can you add a comment explaining the argument here? AFAIU, it's because if we have persisted the metadata, we know that the blob is registered and we can also persist the sliver.

crates/walrus-service/src/node.rs
```diff
@@ -2551,6 +2543,23 @@ impl StorageNodeInner {
         (summary, detail)
     }
 
+    async fn verify_sliver_against_metadata(
+        &self,
+        metadata: VerifiedBlobMetadataWithId,
```
mlegner: Question: Couldn't this take a reference or an `Arc`? Then we wouldn't have to clone it when calling this function.

crates/walrus-service/src/node/pending_sliver_cache.rs
mlegner: This remark/nit is still open.

crates/walrus-service/src/node/config.rs
sadhansood: This is resolved now, as we have a separate config for metadata and sliver now.

crates/walrus-service/src/node/config.rs
sadhansood: ok metadata has `max_cached_entries` now, resolved

crates/walrus-service/src/node.rs
sadhansood: done

crates/walrus-service/src/node.rs
sadhansood: In `flush_pending_metadata`, we remove the cache entry before writing it to storage, so there is always a window where neither store nor cache contains the metadata once a flush starts. That window exists regardless of whether `resolve_metadata_for_sliver ` looks at storage or cache first. I think moving the cache lookup ahead of the storage lookup just shifts when the race can happen.

crates/walrus-service/src/node.rs

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: done

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: done

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: done

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: done, all `PendingMetadataCache` functions now call `evict_expired`.

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: good point, removed

crates/walrus-service/src/node.rs
sadhansood: removed

crates/walrus-service/src/node.rs
sadhansood: done, moved to `prepare_sliver_for_storage`.

crates/walrus-service/src/node/server/routes.rs
sadhansood: sorry, put it back

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: flush_senders is removed

crates/walrus-service/src/node/pending_metadata_cache.rs
sadhansood: ah good point, it isnt needed - removed.

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: This is just to make sure we are returning correct responses. Even though a blob is in the cache, if we don't evict entries, we could return true for a blob expired by ttl (but could be evicted anytime).

crates/walrus-service/src/node.rs
sadhansood: done

crates/walrus-service/src/node.rs
sadhansood: fixed

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs
sadhansood: done

crates/walrus-service/src/node.rs
sadhansood: now using an Arc.

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: fixed now

crates/walrus-service/src/node/pending_sliver_cache.rs
mlegner: My suggestion would be to check if the entry is not expired and only return `true` in that case.

crates/walrus-service/src/node.rs
```diff
@@ -4591,6 +4665,77 @@ mod tests {
...
+        storage_node
+            .as_ref()
+            .inner
+            .storage
+            .update_blob_info(0, &BlobRegistered::for_testing(blob_id).into())?;
+        storage_node
+            .as_ref()
+            .inner
+            .flush_pending_metadata(&blob_id)
+            .await?;
```
halfprice, line 4718: question: why do we need to call this? Shouldn't the registered event trigger this? Even the flush is async, it'll be nice to make sure that the registration triggered flush works.

crates/walrus-service/src/node.rs
```diff
@@ -7837,6 +7992,107 @@ mod tests {
...
+                )
+                .await?;
+
+            events.send(BlobRegistered::for_testing(blob_id).into())?;
+            events.send(BlobCertified::for_testing(blob_id).into())?;
+
+            tokio::time::timeout(Duration::from_secs(5), flush_blocked.notified())
+                .await
+                .expect("cache flush should hit failpoint before certification runs");
+            tokio::time::timeout(Duration::from_millis(100), certify_started.notified())
```
halfprice: nit: 100ms seem short. Given the test uses simulated time, maybe increase this to 1s?

crates/walrus-service/src/node.rs
```diff
@@ -7837,6 +7992,107 @@ mod tests {
...
+                .storage
+                .shard_storage(SHARD_INDEX)
+                .await
+                .expect("shard storage should exist");
+
+            assert!(
+                shard_storage
+                    .get_sliver(&blob_id, SliverType::Primary)
+                    .expect("sliver lookup should succeed")
+                    .is_some(),
```
halfprice: Nit: we can also assert that the sliver data matches what the test originally uploaded.

crates/walrus-service/src/node.rs
```diff
@@ -7837,6 +7992,107 @@ mod tests {
             Ok(())
         }
 
+        #[walrus_simtest]
+        async fn flush_after_registration() -> TestResult {
```
halfprice, line 7998: Praise! Nicely done!

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -143,33 +155,41 @@ impl BackgroundEventProcessor {
...
+            BackgroundTask::ProcessEvent {
+                event_handle,
+                blob_event,
+                checkpoint_position,
+            } => {
+                self.process_event(event_handle, blob_event, checkpoint_position)
+                    .await?
+            }
+            BackgroundTask::FlushPendingCaches { blob_id } => {
+                self.node.flush_pending_caches_with_logging(blob_id).await
```
halfprice, line 188: suggestion: can you document why it is important to block future event processing in the same worker here?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -357,48 +377,31 @@ impl BackgroundEventProcessor {
...
-                _background_processors,
                 background_per_processor_pending_event_count,
+                ..
             } => {
-                // We send the event to one of the workers to process in parallel.
-                // Note that in order to remain sequential processing for the same BlobID, we always
-                // send events for the same BlobID to the same worker.
-                // Currently the number of workers is fixed through the lifetime of the node. But in
-                // case the node want to dynamically adjust the worker, we need to be careful to not
-                // break this requirement.
```
halfprice, line 398: remark: these comments are still relevant and important.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -357,48 +377,31 @@ impl BackgroundEventProcessor {
...
-        // when tracking the total number of pending events to be processed.
         background_per_processor_pending_event_count: Vec<PendingEventCounter>,
     },
-
-    // When there are no background workers, we use a sequential processor to process events using
-    // this processor. This is to keep the same behavior as before BackgroundEventProcessor.
-    // INVARIANT: sequential_processor must be Some if background_processor_senders is empty.
     SequentialProcessor {
         background_event_processor: Arc<BackgroundEventProcessor>,
     },
```
halfprice, line 395: remark: can we bring all the docs back?

crates/walrus-service/src/node/blob_event_processor.rs
halfprice: Looks like a lot of comments are removed in the latest comment in this file. Can you go through them and add them back?

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -548,36 +563,7 @@ impl BlobEventProcessor {
             .update_blob_info(event_handle.index(), &blob_event)?;
 
         if let BlobEvent::Registered(event) = &blob_event {
-            // Registered event is marked as complete immediately. We need to process registered
-            // events as fast as possible to catch up to the latest event in order to not miss
-            // blob sliver uploads.
-            //
-            // If we want to do this in parallel, we shouldn't mix registered event processing with
-            // certified event processing, as certified events take longer and can block following
-            // registered events.
```
halfprice, line 537: Remark: these comments are important. Can you move it to `handle_registered_event`?

crates/walrus-service/src/node/config.rs
```diff
@@ -904,6 +963,18 @@ pub mod defaults {
...
     /// The default number of max concurrent streams for the rest API.
     pub const REST_HTTP2_MAX_CONCURRENT_STREAMS: u32 = 1000;
+    /// Default capacity for the pending sliver cache (number of slivers).
+    pub const PENDING_SLIVER_CACHE_MAX_SLIVERS: usize = 4_096;
+    /// Default byte capacity for the pending sliver cache (~512 MiB).
+    pub const PENDING_SLIVER_CACHE_MAX_BYTES: usize = 512 * 1024 * 1024;
+    /// Default maximum size for a single sliver buffered in the pending sliver cache (~4 MiB).
+    pub const PENDING_SLIVER_CACHE_MAX_SLIVER_BYTES: usize = 4 * 1024 * 1024;
+    /// Default time-to-live for pending uploads before eviction.
+    pub const PENDING_SLIVER_CACHE_TTL: Duration = Duration::from_secs(10);
```
halfprice: suggestion: just to be safe, can we set this to 0 to begin with, and turn it on in a few nodes and see if it has any issues, before making this as a default?

crates/walrus-service/src/node/pending_sliver_cache.rs
sadhansood: Yes, we could definitely also just check whether an entry is expired or not (instead of invoking cache eviction of expired entries), and i thought about doing it but ultimately just went ahead with eviction. I think it seemed like a trade-off b/w optimizing read vs write path and it wasn't clear to me which one to optimize for. If we find this to be taking too many cpu cycles, I can make that change in future.

crates/walrus-service/src/node.rs
sadhansood: fixed this with BUFFERED state

crates/walrus-service/src/node.rs
sadhansood: This test doesn't register the blob, but instead uses manual flush to test. The registration driven path is simtested in `flush_after_registration` however.

crates/walrus-service/src/node.rs
sadhansood: fixed

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: done

crates/walrus-service/src/node/config.rs
sadhansood: done, disabled by default other than tests

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: fixed

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: done

crates/walrus-service/src/node/blob_event_processor.rs
sadhansood: done

PR #2639, opened 2025-10-21
chore: bump Rust to 1.90 and fix new lint warnings

mlegner:
## Description

Bumps Rust to the latest stable toolchain and fixes all new lint warnings (some with `cargo clippy --fix`, some manually). 

## Test plan

CI.

### DIFF
--- a/docker/walrus-orchestrator/Dockerfile
+++ b/docker/walrus-orchestrator/Dockerfile
@@ -3,8 +3,7 @@
 # Copy in all crates, Cargo.toml and Cargo.lock unmodified,
 # and build the application.
 
-# This is the same as rust:1.90-bookworm but we use the GCP registry to avoid rate limiting.
-FROM us-central1-docker.pkg.dev/cryptic-bolt-398315/base-images/rust:1.90-bookworm-amd64 AS builder
+FROM rust:1.90-bookworm AS builder
 
 ARG PROFILE=release
 ARG GIT_REVISION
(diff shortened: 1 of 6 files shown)


crates/typed-store/src/rocks/errors.rs
mlegner, line 1: @sadhansood Clippy identified the `RocksErrorDef` and `BincodeErrorDef` as unused. As a fix, I annotated them with `allow(unused)`, but if they are not needed it would be better to remove them completely. WDYT?

docker/walrus-orchestrator/Dockerfile
```diff
@@ -3,8 +3,8 @@
 # Copy in all crates, Cargo.toml and Cargo.lock unmodified,
 # and build the application.
 
-# This is the same as rust:1.89-bookworm but we use the GCP registry to avoid rate limiting.
-FROM us-central1-docker.pkg.dev/cryptic-bolt-398315/base-images/rust:1.89-bookworm-amd64 AS builder
+# This is the same as rust:1.90-bookworm but we use the GCP registry to avoid rate limiting.
+FROM us-central1-docker.pkg.dev/cryptic-bolt-398315/base-images/rust:1.90-bookworm-amd64 AS builder
```
mlegner: @pei-mysten Does this image exist already?

halfprice (approved):
LGTM!

docker/walrus-orchestrator/Dockerfile
pei-mysten: let's just use back the original docker hub link - now the image builder is able to log into our own docker account so no rate limits is expected

mlegner: OK, then I'll use the normal images everywhere. 👍

crates/typed-store/src/rocks/errors.rs
mlegner: OK, I kept the code for now, we can always clean it up later.

docker/walrus-orchestrator/Dockerfile
mlegner: I switched back to the public images everywhere and cleaned up related comments. I also noticed that #2421 changed some images from `debian:bookworm-slim` to `rust:X-bookworm`, which I reverted now.

PR #2640, opened 2025-10-21
fix(node): ensure aggregate blob info is not deleted prematurely

mlegner:
## Description

There was a bug where in a particular corner case, an aggregate blob info was deleted while some per-object blob info entries still existed (see details below). If then the garbage collection attempted to update the aggregate blob info for the expired object, it caused a panic in the merge operator.

This fixes the bug by adding a check to delete the aggregate blob info only if no corresponding per-object blob info entries still exist.

The bug was triggered in the `simtest_test_lagging_node_recovery` test as follows:
- A blob was registered twice as deletable in epoch 2 with an end epoch of 4.
- One of the corresponding objects was deleted in epoch 3.
- The deletion event was processed at the beginning of epoch 4. At this point, the blob was already expired, meaning that blob data could be deleted.
- However, the second deletable object was still stored in the per-object blob info table as the clean-up was not yet complete.

Besides fixing the bug, this PR also adds a consistency check for invariants of the blob info tables (turned off by default in the config) and improves some logging.

## Test plan

The bug first occurred in the [simtests of an unrelated PR](https://github.com/MystenLabs/walrus/actions/runs/18676507157/job/53247531595?pr=2639). The corresponding test passed more than 100 seeds with this fix.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -6503,35 +6503,48 @@ mod tests {
                 move || skip_blob_certification_at_recovery_beginning,
             );
 
-            let skip_stored_blob_index: [usize; 12] = [3, 4, 5, 9, 10, 11, 15, 18, 19, 20, 21, 22];
-            // Blob 9 is a deletable blob.
-            let deletable_blob_index: [usize; 1] = [9];
+            let skip_stored_blob_index = [3, 4, 5, 9, 10, 11, 15, 18, 19, 20, 21, 22];
+            let expired_blob_index = [3];
+            let deletable_blob_index = [9];
+            let invalid_blob_index = [19];
+            let non_synced_blob_index =
+                [expired_blob_index, deletable_blob_index, invalid_blob_index].concat();
 
             let (cluster, blob_details, event_senders) = setup_shard_recovery_test_cluster(
                 |blob_index| !skip_stored_blob_index.contains(&blob_index),
                 // Blob 3 expires at epoch 2, which is the current epoch when
                 // `setup_shard_recovery_test_cluster` returns.
-                |blob_index| if blob_index == 3 { 2 } else { 42 },
+                |blob_index| {
+                    if expired_blob_index.contains(&blob_index) {
+                        2
+                    } else {
+                        42
+                    }
+                },
                 |blob_index| deletable_blob_index.contains(&blob_index),
             )
             .await?;
 
             // Delete blob 9 and invalidate blob 19.
-            event_senders
-                .all_other_node_events
-                .send(BlobDeleted::for_testing(*blob_details[9].blob_id()).into())?;
-
-            event_senders
-                .all_other_node_events
-                .send(InvalidBlobId::for_testing(*blob_details[19].blob_id()).into())?;
+            for i in invalid_blob_index {
+                event_senders
+                    .all_other_node_events
+                    .send(InvalidBlobId::for_testing(*blob_details[i].blob_id()).into())?;
+            }
+            for i in deletable_blob_index {
+                event_senders
+                    .all_other_node_events
+                    .send(BlobDeleted::for_testing(*blob_details[i].blob_id()).into())?;
+            }
 
             // Make sure that blobs in `skip_stored_blob_index` are not certified in node 0.
             for i in skip_stored_blob_index {
+                let blob_id = blob_details[i].blob_id();
                 let blob_info = cluster.nodes[0]
                     .storage_node
                     .inner
                     .storage
-                    .get_blob_info(blob_details[i].blob_id());
+                    .get_blob_info(blob_id);
                 if deletable_blob_index.contains(&i) {
                     assert!(matches!(
                         blob_info.unwrap().unwrap().to_blob_status(1),
@@ -6544,13 +6557,19 @@ mod tests {
                         }
                     ));
                 } else {
-                    assert!(matches!(
-                        blob_info.unwrap().unwrap().to_blob_status(1),
-                        BlobStatus::Permanent {
-                            is_certified: false,
-                            ..
-                        }
-                    ));
+                    let blob_status = blob_info.unwrap().unwrap().to_blob_status(1);
+                    if expired_blob_index.contains(&i) {
+                        assert!(
+                            matches!(blob_status, BlobStatus::Nonexistent),
+                            "unexpected blob status for expired blob {i} ({blob_id}): \
+                            {blob_status:?}",
+                        );
+                    } else {
+                        assert!(
+                            matches!(blob_status, BlobStatus::Permanent { .. }),
+                            "unexpected blob status for blob {i} ({blob_id}): {blob_status:?}",
+                        );
+                    }
                 }
             }
 
@@ -6582,7 +6601,7 @@ mod tests {
                 &blob_details,
                 &node_inner.storage,
                 shard_storage_dst.as_ref(),
-                &[3, 9, 19],
+                &non_synced_blob_index,
             )?;
 
             // Checks that the shard sync progress is reset.


crates/walrus-service/src/node/storage.rs
```diff
@@ -766,10 +766,14 @@ impl Storage {
         }
 
         // At this point we are sure that the blob is no longer registered and can actually delete
-        // the blob info and data. If the blob is reregistered outside this transaction, the
-        // transaction will fail.
-        self.blob_info
-            .delete_in_transaction(&transaction, blob_id)?;
+        // the data. If the blob is reregistered outside this transaction, the transaction will
+        // fail.
+        if blob_info.can_blob_info_be_deleted(current_epoch) {
```
mlegner, line 769: This if clause is the essence of the fix.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -486,6 +487,82 @@ impl BlobInfoTable {
...
+    /// The checks are not exhaustive yet.
+    pub fn check_invariants(&self) -> Result<(), anyhow::Error> {
+        for result in self.per_object_blob_info.safe_iter()? {
+            let Ok((object_id, PerObjectBlobInfo::V1(per_object_blob_info))) = result else {
+                return Err(anyhow::anyhow!(
+                    "error encountered while iterating over per-object blob info: {result:?}"
+                ));
+            };
+            let blob_id = per_object_blob_info.blob_id();
+            let Some(blob_info) = self.aggregate_blob_info.get(&blob_id)? else {
```
halfprice, line 506: question: the read to aggregate_blob_info and the per_object_blob_info may not be performed on the same DB snapshot, is that possible (if so, the data may not be consistent)? I wonder if we should create a db snapshot at the beginning of this function, and performs all the reads on the same snapshot.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -486,6 +487,82 @@ impl BlobInfoTable {
 
         Ok(())
     }
+
+    /// Checks some internal invariants of the blob info table.
+    ///
+    /// The checks are not exhaustive yet.
+    pub fn check_invariants(&self) -> Result<(), anyhow::Error> {
```
halfprice, line 498: praise: really like these checks!

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1384,6 +1482,19 @@ impl BlobInfoApi for BlobInfoV1 {
...
+    fn can_blob_info_be_deleted(&self, current_epoch: Epoch) -> bool {
+        match self {
+            Self::Invalid { .. } => {
+                // We don't know whether there are any deletable blob objects for this blob ID.
+                false
+            }
+            Self::Valid(ValidBlobInfoV1 {
+                count_deletable_total,
+                ..
+            }) => *count_deletable_total == 0 && self.can_data_be_deleted(current_epoch),
```
halfprice: > Nit: I think this condition is specific tailored to the bug and also the current implementation. I think we can make it more robust by also making sure that permanent blob per object blob info is empty.

I was about to suggest above, but after reading more code, I realized that for permanent blob expiration, we only remove the per object blob info entry, but not updating the counter in aggregated blob info. This different behavior between deletable and permanent is quite subtle and easy to miss I feel. Also, I saw [this code](https://github.com/MystenLabs/walrus/blob/c4dfdf60e5f08f46b73c135294140a0b38a2fe2a/crates/walrus-service/src/node/storage/blob_info.rs#L1019) that we don't really need? If so, I think we probably need to add some doc here as well as in `process_expired_blob_objects` about this subtly.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -486,6 +487,82 @@ impl BlobInfoTable {
...
+                            .is_some_and(|e| e >= per_object_end_epoch),
+                        "latest_seen_deletable_certified_end_epoch for blob ID {blob_id} is \
+                        {latest_seen_deletable_certified_end_epoch:?}, which is inconsistent with \
+                        the end epoch of a deletable certified blob object: {per_object_end_epoch} \
+                        (object ID: {object_id})"
+                    );
+                }
+            }
+        }
+
```
halfprice, line 554: suggestion (can be done in a follow up PR): another check maybe worth adding is to assert that for expired blobs, there is no per object blob info exists.

crates/walrus-service/src/node/storage.rs
halfprice: Remark: I agree that this fix should fix the correctness, but I think the real problem is this: https://github.com/MystenLabs/walrus/pull/2599#discussion_r2453997287. It should use the event epoch as opposite to the latest epoch. The doc above the original code (created in [this PR](https://github.com/MystenLabs/walrus/pull/1132/files)) also explains why we need to use the event epoch: the event needs to be operate in the scope of the current event, because there will be future events that is depending on the effects of this events. Following this logic, I wonder if there are other scenario where this fix may not cover, but I cannot think of any at the moment.

If we are using event epoch here, I think when processing the first blob deletion event, the blob won't be considered as deleted. That event cannot be at epoch 4 because deletion at epoch 4 (blob expiration epoch) is not allowed, so it must be an epoch < 4.

halfprice (commented):
Great work @mlegner tackling this important and tricky problem! Really like the invariant check added! Below, I have a suggestion of another way to fix the (possibly more fundamental) issue. PTAL and let me know your thoughts!

crates/walrus-service/src/node/storage.rs
mlegner: Thanks for noticing this. I don't know why I made that change, given I made the original fix myself in #1132 and even added a comment... 🤦 

I also realized that the corresponding test added in that PR didn't catch this because I added some events to the `cluster_at_epoch1_without_blobs` function which meant that the assertion succeeded anyway. I fixed this and made the tests more robust in dfbf6cb43d2b9d745a44b7bb9d36e5ecb54cd940, also adding a test case for a permanent blob. That added case showed that we do in fact have to use the correct epoch when processing the deleted event (test fails for that commit), so I reverted that change in 19d46d6bc641d5777380280b20341e9506d96e7e.

I believe that we do in fact need both fixes: The `can_blob_info_be_deleted` is needed for invalid blobs and for when the garbage collection will run in the background.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Hmm... This is a great question. I believe you are right and we should perform this on a snapshot. AFAICT, we don't have this implemented in `typed-store` (cc @sadhansood), so this might require a few additional changes. How about we do this in a follow-up?

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Hmm... This seems like a pretty slow check as we would have to search the per-object table. 🤔

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: Good catch. In fact, it's probably best to also update the aggregate blob info for expired permanent blobs. I added this in 2c3d3432b58ac591b187bea62a3741aa75b14a73.

Note that I used a new `PermanentExpired` enum variant to avoid changing existing variants (even though they are not yet used in production).

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: I think this check only needs to be run in tests that testing correctness (simtest + antithesis), where performance isn't a concern.

crates/walrus-service/src/node/storage/blob_info.rs
sadhansood: Ah yeah, we don't have snapshot implemented yet. @mlegner lmk if you want me to do it (and totally awesome if you also just want to do it yourself!)

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -259,14 +259,18 @@ impl BackgroundEventProcessor {
...
+        skip_all,
+        fields(walrus.blob_id = %event.blob_id, walrus.epoch = tracing::field::Empty),
+    )]
     async fn process_blob_deleted_event(
         &self,
         event_handle: EventHandle,
         event: BlobDeleted,
     ) -> anyhow::Result<()> {
         let blob_id = event.blob_id;
         let current_epoch = self.node.current_committee_epoch();
```
halfprice, line 272: Nit: can we change the name to current_committee_epoch? concurrent_epoch may be confusing.

crates/walrus-service/src/node/storage.rs
```diff
@@ -668,7 +668,7 @@ impl Storage {
                 }
             };
 
-            if blob_info.is_registered(current_epoch) {
+            if !blob_info.can_data_be_deleted(current_epoch) {
                 tracing::trace!(
                     %blob_id,
                     "skipping blob that is still registered",
```
halfprice, line 674: nit: also update the doc to mention about `deletion`, and disconnect it wit `register`?

crates/walrus-service/src/node/storage.rs
halfprice: Thanks for the thorough updates!

> The can_blob_info_be_deleted is needed for invalid blobs and for when the garbage collection will run in the background.

question: for invalid blobs, it makes sense. Why `can_blob_info_be_deleted` is needed when GC run in background? I think epoch triggered GC in the background probably probably don't need this? I can see that GC triggered by BlobDeletion event, and we run that blob deletion in the background, may need this. Sorry for asking this detailed question on already working code. I want to make sure that I don't miss anything, and have the correct understanding of the code.

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: I think we could just call `create_snapshot` on the RocksDb instance, and use the snapshot in the read option to create the iterator? I think this is needed here otherwise the check may fail. Created WAL-1077 to track this.

PR #2645, opened 2025-10-22
fix: set default aggregator request buffer size correctly

halfprice:
## Description

I dig into one of @giac-mysten's [comment](https://github.com/MystenLabs/walrus/pull/2628#discussion_r2444865178), and realized that I understood `buffer` completely wrong. I thought it meant `additional` requests, which is also implied from the BufferLayer doc, but it actually also includes the current being service requests.

This PR fix the default (needs to be greater than the max concurrent limit), and update the test to use the correct assertion. Thanks for catching this @giac-mysten !

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -895,7 +895,7 @@ pub struct AggregatorArgs {
     #[serde(default)]
     pub max_blob_size: Option<u64>,
     /// The maximum number of requests that can be buffered before the server starts rejecting new
-    /// ones.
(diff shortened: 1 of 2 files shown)


mlegner (approved):
LGTM, although @giac-mysten's opinion is probably more relevant.

giac-mysten:
Exellent catch @halfprice, i am now wondering if we have the same problem with the publisher 😬 will take a look

giac-mysten (approved):

PR #2648, opened 2025-10-23
feat: Add signal forwarder for child process

sadhansood:
## Description

This PR adds a signal forwarder for child process. This is needed for when a user ctrl+c the process and expect to see both parent and child to terminate (which it won't do today). I'll however admit, the signal handling part of the code is mostly written by claude but i tested it to be working as expected locally.

## Test plan

Running locally.


crates/walrus-service/src/client/cli/internal_run.rs
```diff
@@ -447,6 +474,84 @@ fn apply_common_store_child_args(
...
+
+    for (kind, signo, name) in signals {
+        let handle_clone = handle.clone();
+        match signal(kind) {
+            Ok(stream) => {
+                let mut stream = stream;
+                tokio::spawn(async move {
+                    if stream.recv().await.is_some() {
+                        forward_unix_signal_to_child(handle_clone, signo, name).await;
+                        let exit_code = 128 + signo;
```
halfprice: question: what does 128 do here? Also, std::process::exit(exit_code); is how the main process gets terminated now?

crates/walrus-service/src/client/cli/internal_run.rs
```diff
@@ -447,6 +474,84 @@ fn apply_common_store_child_args(
...
+        (SignalKind::terminate(), libc::SIGTERM, "SIGTERM"),
+        (SignalKind::interrupt(), libc::SIGINT, "SIGINT"),
+    ];
+
+    for (kind, signo, name) in signals {
+        let handle_clone = handle.clone();
+        match signal(kind) {
+            Ok(stream) => {
+                let mut stream = stream;
+                tokio::spawn(async move {
```
halfprice, line 494: question: say the child process finishes, but the main process is a long running process, this thread will always stay until the main process quits, is that correct? I wonder if there is a way to abort this thread if the child process finishes.

sadhansood: Yes, good call. I think even though it'd be rare for parent to hang around longer than child, i think we should handle the scenario by not forwarding the signal to another process. I added a cancellation token which gets cancelled when child exits, this cancellation token gets set to true in a tokio task which is waiting on the child handle to exit.

crates/walrus-service/src/client/cli/internal_run.rs
sadhansood: Added a comment, this is because when we catch a signal - we want to exit with `128 + signo`  which is the exit code shell shows when we do not catch the signal through a signal handler. So essentially we are just preserving the normal behavior.

crates/walrus-service/src/client/cli/internal_run.rs
halfprice: I see. I wasn't sure if this is a CLI only behavior or a sdk behavior. Looks like it's a CLI only behavior, which is fine. If in SDK (e.g. may affect publisher), we need to make sure that we won't constantly bringing up the monitoring threads and not cleaning them up.

PR #2655, opened 2025-10-24
docs: remove quick wallet setup admonishment - set testnet as the default in client_config.yaml

wbbradley:
## Description

Remove "Quick Wallet Setup" admonishment that leads people into a non-functional setup and set testnet as the default.

## Test plan

CI Pipeline.

### DIFF
--- a/docs/book/usage/networks.md
+++ b/docs/book/usage/networks.md
@@ -63,8 +63,7 @@ available.
 Interacting with Walrus requires a valid Sui Testnet wallet with some amount of SUI tokens. The
 normal way to set this up is via the Sui CLI; see the [installation
 instructions](https://docs.sui.io/guides/developer/getting-started/sui-install) in the Sui
-documentation. If you do not want to install the Sui CLI, you can also generate a new Sui wallet for
-Testnet using `walrus generate-sui-wallet --network testnet`.
+documentation.
 
 After installing the Sui CLI, you need to set up a Testnet wallet by running `sui client`, which
 prompts you to set up a new configuration. Make sure to point it to Sui Testnet, you can use the


docs/book/usage/setup.md
```diff
@@ -15,12 +15,6 @@ for an overview over all Walrus networks.
 
 ## Prerequisites: Sui wallet, SUI and WAL {#prerequisites}
 
-```admonish tip title="Quick wallet setup"
-If you just want to set up a new Sui wallet for Walrus, you can generate one using the
-`walrus generate-sui-wallet --sui-network mainnet` command after [installing Walrus](#installation).
-You still need to obtain some SUI and WAL tokens, but you do not have to install the Sui CLI.
-```
-
```
mlegner: Can you elaborate on what is "non-functional" about the resulting setup? Also, we should probably change or remove the corresponding CLI command.

wbbradley: There has been feedback from UX research that while this tip is helpful to folks that need that particular route, it leaves naive users in a less than ideal position because the reality is that to work with accounts, one generally needs `sui`. And, in the long run, it's best that folks using the CLI grow accustomed to the `sui` CLI. Also, asking a brand-new user "If you just want to set up a new Sui wallet for Walrus" is a distraction from learning since the user has no idea at this point whether or to what extent they will or won't need just this, or something more. In other words, we can't expect new users to be informed enough at this juncture to understand this question and its potential outcomes. It's possible that we could rephrase it such that it explicitly filters out new users, but that is also just noise to a new user. 

Yes, I agree that Walrus users can entirely live without awareness of (or usage of) the `walrus generate-sui-wallet` command. I'll log a Linear task to investigate removal of that command.

setup/client_config.yaml
```diff
@@ -29,4 +29,4 @@ contexts:
       # active_address: 0x0000000000000000000000000000000000000000000000000000000000000000
     rpc_urls:
       - https://fullnode.testnet.sui.io:443
-default_context: mainnet
+default_context: testnet
```
wbbradley, line 32: The context here is also based on feedback from usability tests. People do not want to be accidentally thrust into mainnet when they are learning.

docs/book/usage/setup.md
wbbradley: I'm trying to find the "non-functional" bit in internal conversations with UX researchers (I'm pulling that from memory at the moment), but I think it amounts to the fact that once you've generated a sui wallet, in order to manipulate it at all, you will generally still need the `sui` CLI. So, in that sense, the `generate-sui-wallet` command is only a partial optimization for power users that already have the `sui` CLI installed.

setup/client_config.yaml
mlegner: Makes sense. Then I'd suggest to also remove/change [this admonition](https://github.com/MystenLabs/walrus/blob/main/docs/book/usage/setup.md?plain=1#L11-L14).

docs/book/usage/setup.md
mlegner: OK, happy to remove this admonition.

PR #2668, opened 2025-10-28
feat: Enable tail uploads in the daemon

sadhansood:
## Description

This PR is needed to enable tail uploads in the publisher. Tail uploads are set to detached mode based on the configs. 

1. We added a new config `tail_handling` which can be set to `blocked` or `detached` and handles tail upload.
2. `child_process` is removed from the config and added as a cli param to the store commands (as this is a cli construct) and remove ambiguity from tail handling.

### DIFF
--- a/crates/walrus-service/src/client/cli/args.rs
+++ b/crates/walrus-service/src/client/cli/args.rs
@@ -1190,6 +1190,11 @@ pub struct CommonStoreOptions {
     #[arg(long, value_enum)]
     #[serde(default)]
     pub upload_mode: Option<UploadModeCli>,
+    /// Spawn a helper process that continues detached tail uploads after quorum is reached.
+    /// This is only effective when tail handling is configured as `detached`.
+    #[arg(long)]
+    #[serde(default)]
+    pub child_process_uploads: bool,
     /// Internal flag to signal the process is running as a child for background uploads.
     #[arg(long, hide = true)]
     #[serde(default)]
@@ -1961,6 +1966,7 @@ mod tests {
                 upload_relay: None,
                 skip_tip_confirmation: false,
                 upload_mode: None,
+                child_process_uploads: false,
                 internal_run: false,
             },
         })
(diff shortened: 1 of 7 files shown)


crates/walrus-service/src/client/cli/args.rs
```diff
@@ -1190,6 +1190,11 @@ pub struct CommonStoreOptions {
     #[arg(long, value_enum)]
     #[serde(default)]
     pub upload_mode: Option<UploadModeCli>,
+    /// Spawn a helper process that continues detached tail uploads after quorum is reached.
+    /// This is only effective when tail handling is configured as `detached`.
+    #[arg(long)]
+    #[serde(default)]
+    pub child_process_uploads: bool,
```
halfprice, line 1197: question: does this mean that we will not support child process uploads in SDK?

crates/walrus-service/src/client/cli/runner.rs
```diff
@@ -1123,6 +1169,18 @@ impl ClientCommandRunner {
...
+            let mut handles_guard = collector.lock().await;
+            let mut handles = Vec::new();
+            std::mem::swap(&mut *handles_guard, &mut handles);
+            drop(handles_guard);
+            for handle in handles {
+                if let Err(err) = handle.await {
+                    tracing::warn!(?err, "tail upload task failed");
+                }
+            }
+        }
```
halfprice, line 1194: Question: does this mean that the tail handle is blocking?

If so, I think we should either print an warning message in CLI saying that detached mode is not allowed in CLI, and the behavior is blocking mode, and please use child process for better performance. Or return an error when detached mode is selected in CLI.

halfprice (approved):
Thanks Sadhan for the updates! I have one question below about using detached mode in CLI. Can you please take a look? I approve the PR since it's a minor behavior issue.

crates/walrus-service/src/client/cli/args.rs
sadhansood: SDK doesn't need spawning a child but it can just do tail uploads in the detached mode. It can still do that with the `tail_upload_mode` config param.

crates/walrus-service/src/client/cli/runner.rs
sadhansood: This is blocking only when detached mode is selected with `child_process_enabled` set to false. I can add a warning message in this particular scenario, so user doesn't have unrealistic expectations.

PR #2676, opened 2025-10-30
feat: add a blob concatenation endpoint to aggregator

wbbradley:
## Description

Add an endpoint to concatenate blobs by blob_id or object_id. Includes support in local-testbed.sh to instantiate the aggregator.

## Test plan

Manual testing with local-testbed.sh. CI Pipeline.

---

## Release notes

- [ ] Storage node:
- [x] Aggregator: Includes a new endpoint `/v1alpha/blobs/concat` handling `GET` and `POST` methods. This endpoint streams the given blobs back as one concatenated response. This is an alpha endpoint and is not yet stable. We're looking for feedback!
- [ ] Publisher:
- [ ] CLI:


scripts/walrus-store-sliced.sh
```diff
@@ -0,0 +1,154 @@
...
+
+# Display the chunks
+for chunk in "${chunk_files[@]}"; do
+  note "  - $(basename "$chunk")"
+done
+
+# Call walrus store with all chunks
+note "running: walrus store ${walrus_args[*]} ${chunk_files[*]}"
+
+if walrus store "${walrus_args[@]}" "${chunk_files[@]}"; then
```
mlegner: Important: We have [a parameter](https://github.com/MystenLabs/walrus/blob/c4422e4b08a5495bf23b2f51f2cd21877608a7ef/crates/walrus-sdk/src/config/communication_config.rs#L202) that limits the total size of blobs when multiple ones are stored simultaneously. This is currently [set to 1 GiB by default](https://github.com/MystenLabs/walrus/blob/c4422e4b08a5495bf23b2f51f2cd21877608a7ef/crates/walrus-sdk/src/config/communication_config.rs#L229).

If we want to recommend this chunked upload, we should probably revisit this parameter and its default and/or add a CLI option to overwrite it.

docs/examples/shell/walrus-store-sliced.sh
mlegner, line 1: Suggestion: So far, the `scripts/` directory was used mainly for internal scripts. Can we put user-facing scripts somewhere else? For example in `docs/` or `docs/examples`?

docs/book/operator-guide/aggregator.md
```diff
@@ -144,3 +144,18 @@ configuration:
...
 
 The setup and use of an "authenticated publisher" is covered in a [separate section](./auth-publisher.md).
+
+### Advanced aggregator uses
+
+#### Large file support
+
+The aggregator can perform concatenation of multiple blobs via the `/v1alpha/blobs/concat` endpoint.
+This endpoint enables delivery of very large files - which would otherwise be unsupported due to
+individual blob size restrictions. See `./scripts/walrus-store-sliced.sh` for an example of how a
```
mlegner: Nit: Can we provide the full URL to the GitHub repo here? Alternatively, we can create a symlink to some location under `docs/book` and use a relative link here to be self-contained (similar to the symlinks in `docs/book/setup`).

docs/book/operator-guide/aggregator.md
```diff
@@ -144,3 +144,18 @@ configuration:
 ### Advanced publisher uses
 
 The setup and use of an "authenticated publisher" is covered in a [separate section](./auth-publisher.md).
+
+### Advanced aggregator uses
+
+#### Large file support
+
+The aggregator can perform concatenation of multiple blobs via the `/v1alpha/blobs/concat` endpoint.
```
mlegner: Suggestion: Mention the version requirements (probably 1.37 or 1.38).

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -111,14 +117,82 @@ pub struct ReadOptions {
...
+#[serde(deny_unknown_fields)]
+pub struct ConcatQueryParams {
+    /// Comma-separated list of blob IDs or object IDs to concatenate.
+    pub ids: String,
+    /// Whether to perform a strict consistency check.
+    #[serde(default)]
+    pub strict_consistency_check: bool,
+    /// Whether to skip consistency checks entirely.
+    #[serde(default)]
+    pub skip_consistency_check: bool,
```
mlegner: Suggestion: Could we instead use the flattened `ReadOptions` here?

```suggestion
    /// General read options.
    #[serde(flatten)]
    pub read_options: ReadOptions,
```

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -351,6 +425,225 @@ impl From<ClientError> for GetBlobError {
...
+pub(super) async fn get_blobs_concat<T: WalrusReadClient + Send + Sync + 'static>(
+    State((client, response_header_config)): State<(Arc<T>, Arc<AggregatorResponseHeaderConfig>)>,
+    request_headers: HeaderMap,
+    Query(params): Query<ConcatQueryParams>,
+) -> Response {
+    // Parse consistency check options
+    let consistency_check = match params.consistency_check() {
+        Ok(check) => check,
+        Err(error) => return error.into_response(),
+    };
```
mlegner: Suggestion: Put this into the `concat_blobs_impl`.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -351,6 +425,225 @@ impl From<ClientError> for GetBlobError {
...
+    let body = Body::from_stream(stream);
+
+    // Build response with appropriate headers
+    let mut response = Response::builder()
+        .status(StatusCode::OK)
+        .header(CONTENT_TYPE, "application/octet-stream")
+        .header(CACHE_CONTROL, "max-age=86400, stale-while-revalidate=3600")
+        .header(X_CONTENT_TYPE_OPTIONS, "nosniff")
+        .body(body)
+        .expect("failed to build concatenated blob response");
```
mlegner: Question: Can this actually happen? If so, can we return an error and log a warning instead of panicking?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -351,6 +425,225 @@ impl From<ClientError> for GetBlobError {
...
+    let mut blob_ids = Vec::new();
+    let mut first_attribute: Option<BlobAttribute> = None;
+    for (index, id_str) in id_strings.iter().enumerate() {
+        match parse_and_resolve_id(id_str.as_str(), client.as_ref()).await {
+            Ok((blob_id, attribute)) => {
+                blob_ids.push(blob_id);
+                // Only capture the first blob's attribute
+                if index == 0 {
+                    first_attribute = attribute;
+                }
```
mlegner, line 532: Suggestion: If we only need the first blob's attribute, we could skip getting the other ones to avoid unnecessary requests.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -351,6 +425,225 @@ impl From<ClientError> for GetBlobError {
...
+                        },
+                        ClientErrorKind::BlobTooLarge(size) => ConcatBlobError::BlobTooLarge {
+                            blob_id: blob_id.to_string(),
+                            size: *size,
+                        },
+                        _ => ConcatBlobError::BlobReadFailed {
+                            blob_id: blob_id.to_string(),
+                            message: error.to_string(),
+                        },
+                    };
```
mlegner: Suggestion: Define a separate function for this or implement `From<ClientError> for ConcatBlobError` to avoid cluttering this function.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -351,6 +425,225 @@ impl From<ClientError> for GetBlobError {
...
+
+    // Populate response headers with priority:
+    // 1. Request header override (if Content-Type in request)
+    // 2. First blob's attributes (if available)
+    // 3. Default (already set above)
+    if let Some(content_type) = request_headers.get(CONTENT_TYPE) {
+        response
+            .headers_mut()
+            .insert(CONTENT_TYPE, content_type.clone());
+    } else if let Some(attribute) = first_attribute {
```
mlegner: Question: Is it intentional that this behavior is different from other endpoints like the `get_blob_by_object_id`? Also, do we need any standard headers as set by `populate_response_headers_from_request`? 

Remark: Already now the behavior seems to be a bit inconsistent...

scripts/walrus-store-sliced.sh
wbbradley: Yeah, I actually noticed this the other day and logged [this issue](https://linear.app/mysten-labs/issue/WAL-1089/allow-unlimited-walrus-store-aggregate-blob-sizes) to track serializing uploads to avoid running into this limit.

wbbradley: For now I'll make the uploads run serially in this script.

docs/examples/shell/walrus-store-sliced.sh
wbbradley: Done. (Moved into `docs/examples/shell`)

docs/book/operator-guide/aggregator.md
wbbradley: Going with symlink approach.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Great observation. Thanks.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Good point. We should be able to fail more gracefully here.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -351,6 +425,225 @@ impl From<ClientError> for GetBlobError {
...
+                    first_attribute = attribute;
+                }
+            }
+            Err(error) => return error.into_response(),
+        }
+    }
+
+    // Create streaming response
+    let client_clone = client.clone();
+    let stream = stream::iter(blob_ids).then(move |blob_id| {
```
halfprice, line 540: question: what's the behavior of this stream iterator? When will the second blob read start in respect first blob read operation?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -351,6 +425,225 @@ impl From<ClientError> for GetBlobError {
...
+                            blob_id: blob_id.to_string(),
+                        },
+                        ClientErrorKind::BlobTooLarge(size) => ConcatBlobError::BlobTooLarge {
+                            blob_id: blob_id.to_string(),
+                            size: *size,
+                        },
+                        _ => ConcatBlobError::BlobReadFailed {
+                            blob_id: blob_id.to_string(),
+                            message: error.to_string(),
+                        },
```
halfprice: Question: if blobs read in the middle encounter one of these errors, what happens to the following blobs?

docs/examples/shell/walrus-store-sliced.sh
```diff
@@ -0,0 +1,154 @@
+#!/bin/bash
```
halfprice, line 1: question: Is this a utility script just for testing purpose?

halfprice (approved):
Thanks @wbbradley for the PR! This looks great to me and I just have a few questions.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: We actually need to scan the list and do this conversion because the interface allows specification by blob_id, or object_id, so this loop will convert all ids to blob_ids. The `attribute` has minimal cost and comes along with `get_blob_by_object_id`.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Done.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Unfortunately From won't work because the errors can really benefit from the exterior context of the blob_id. I'll wrap this in a separate fn. Thanks.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Ok, I've DRY'd up the treatment of headers, and extended it to support the `Accept` header when it doesn't contain `*`.

wbbradley: We may want to deprecate the `Content-type` api since it is non-standard in HTTP.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: The second blob read waits until the first blob read is complete. I believe it also waits until the send to the client has completed, as well. I don't think there is any pipelining at the moment. This is not ideal, but I believe it should be sufficient as a v1 since managing the memory usage of multiple blobs loaded simultaneously will become more complex quickly.

docs/examples/shell/walrus-store-sliced.sh
wbbradley: Yes, it's primarily for testing and just as an example illustrating a basic splitting flow. It does not currently handle gathering the blob IDs in an elegant way. https://github.com/MystenLabs/walrus/pull/2637 is intended to be a more elegant solution to that, but still requires some work.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Any failure will cause the entire request to fail. How exactly it fails is not entirely clear. The interface here in axum for `Body::from_stream` is fairly opaque. Given that this is a public endpoint, I'm not too worried about terrible things happening (security/privacy/data loss). So, I think we can live with it and test it out once its live.

PR #2677, opened 2025-10-31
refactor: reduce database impact on recovery symbol calls

jpcsmith:
## Description

This commit improves the serving of recovery symbols in several ways.

First, it makes the fetching of the associated sliver an asynchronous operation so that it does not block the core tokio thread. Second, fetching metadata is currently unnecessary since there is only one supported encoding type, which can be known in advance. Third, requests to the blob-info table which were performed once per symbol to check blob access is now performed once per request (get-symbol or list-symbol).

Finally, whereas before it was possible that a single list-symbols request could have some symbols fail to be fetched due to the symbol service being unavailable. Now, if the symbol service is ready at the beginning of the request an attempt will be made to process all of the symbols.

## Test plan

Existing tests.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -49,6 +49,7 @@ use tracing::{Instrument as _, Span, field};
 use typed_store::{TypedStoreError, rocks::MetricConf};
 use walrus_core::{
     BlobId,
+    EncodingType,
     Epoch,
     InconsistencyProof,
     PublicKey,
@@ -57,6 +58,7 @@ use walrus_core::{
     SliverPairIndex,
     SliverType,
     SymbolId,
+    by_axis::Axis,
     encoding::{
         EncodingAxis,
         EncodingConfig,
@@ -2661,48 +2663,6 @@ impl StorageNodeInner {
         self.is_shutting_down.load(Ordering::SeqCst)
     }
 
-    async fn try_retrieve_recovery_symbol(
-        &self,
-        blob_id: &BlobId,
-        sliver_pair_index: SliverPairIndex,
-        target_sliver_type: SliverType,
-        target_pair_index: SliverPairIndex,
-    ) -> Result<GeneralRecoverySymbol, RetrieveSymbolError> {
-        // Claim a worker for performing the expansion necessary to get the symbol.
-        let mut worker = match self.symbol_service.clone().ready_oneshot().now_or_never() {
-            Some(result) => result.expect("polling the symbol service is infallible"),
-            None => return Err(Unavailable.into()),
-        };
-
-        let sliver = self
-            .retrieve_sliver(blob_id, sliver_pair_index, target_sliver_type.orthogonal())
-            .await?;
-        let convert_error = |error| match error {
-            RecoverySymbolError::IndexTooLarge => {
-                panic!("index validity must be checked above")
-            }
-            RecoverySymbolError::EncodeError(error) => {
-                RetrieveSymbolError::Internal(anyhow!(error))
-            }
-        };
-        let metadata = self
-            .storage
-            .get_metadata(blob_id)
-            .context("could not retrieve blob metadata")?
-            .ok_or_else(|| {
-                RetrieveSymbolError::Internal(anyhow!("metadata not found for blob {:?}", blob_id))
-            })?;
-
-        let request = RecoverySymbolRequest {
-            blob_id: *metadata.blob_id(),
-            source_sliver: sliver,
-            target_pair_index,
-            encoding_type: metadata.metadata().encoding_type(),
-        };
-
-        worker.call(request).map_err(convert_error).await
-    }
-
     /// Common validation for blob operations
     fn validate_blob_access<E>(
         &self,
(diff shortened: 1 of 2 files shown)


jpcsmith:
@halfprice This should fix the issues you were encountering on not being able to read while uploading (there were recoveries at the time you tried). The blocking work that was being done on the core threads during recovery were for getting the metadata (specifically fetching and decoding it for the encoding type), blob_info, and sliver. Getting the sliver is now done on a blocking thread. Getting the metadata is now avoided since there is only one encoding type. Getting the blob info was being done for each symbol in the list-symbols request, which was unnecessary (the same was true of getting the metadata). Getting the blob info is now only done once per request (but is still a blocking call).

crates/walrus-service/src/node.rs
```diff
@@ -2717,6 +2677,195 @@ impl StorageNodeInner {
...
+        }
+    }
+
+    /// Retrieve the recovery symbol without calling `validate_blob_access`.
+    #[tracing::instrument(level = "debug", skip_all)]
+    async fn retrieve_recovery_symbol_unchecked(
+        &self,
+        blob_id: &BlobId,
+        symbol_id: SymbolId,
+        sliver_type: Option<SliverType>,
```
halfprice, line 2724: question: when `sliver_type` can be none?

crates/walrus-service/src/node.rs
```diff
@@ -3148,46 +3256,55 @@ impl ServiceState for StorageNodeInner {
         blob_id: &BlobId,
         filter: RecoverySymbolsFilter,
     ) -> Result<Vec<GeneralRecoverySymbol>, ListSymbolsError> {
-        let n_shards = self.n_shards();
+        // Begin by fetching a ready worker, as this gates all database reads based
+        // on whether we have capacity to even serve the request.
+        let mut worker = Some(self.get_ready_symbol_service()?);
```
halfprice, line 3261: question: IIUC, if the request is retrieving multiple recovery symbols, and if the node currently does not have capacity, it will reject the request. But if the node has 1 capacity, but the request will retrieve 10 symbols, this will go though? Shouldn't we check if the number of workers available matches the number of requested symbols (or some threshold based on the number of requested symbols?)

crates/walrus-service/src/node.rs
```diff
@@ -2717,6 +2677,195 @@ impl StorageNodeInner {
...
+            let wait_for_worker = async {
+                match worker.take() {
+                    Some(worker) => worker,
+                    None => self.wait_for_ready_symbol_service().await,
+                }
+            };
+
+            let (mut worker, sliver_result) = tokio::join!(
+                wait_for_worker,
+                self.retrieve_sliver_unchecked(
```
halfprice, line 2773: nit: should this be part of the RecoverySymbolService so that the overhead of reading DB is counted inside of serving the request?

crates/walrus-service/src/node.rs
```diff
@@ -2717,6 +2677,195 @@ impl StorageNodeInner {
...
+    ) -> Result<Option<EncodingType>, TypedStoreError> {
+        match walrus_core::SUPPORTED_ENCODING_TYPES {
+            &[EncodingType::RS2] => Ok(Some(EncodingType::RS2)),
+            // If additional encoding types are used, then it becomes necessary to fetch the
+            // encoding type from the database. If the metadata is otherwise available, then the
+            // encoding type is available there and this function should not be used, otherwise, a
+            // database call may be required.
+            _ => self
+                .storage
+                .get_metadata(blob_id)
```
halfprice, line 2821: Nit: should we put this in blocking thread as well?

halfprice (approved):
Thanks a lot @jpcsmith for the great improvement! The PR looks great and I only have a few small comments. Look forward to seeing its effects in prod!

crates/walrus-service/src/node.rs
jpcsmith: When the caller doesn't care whether the symbol comes from the primary or the secondary sliver. Ideally, the argument should be completely removed, and the storage node would just return the recovery symbol and proof for the symbol from whichever sliver it has stored. However, because the proofs of inconsistency currently require all the supporting proofs for the symbols to be of the same type, the argument is kept for compatability.

crates/walrus-service/src/node.rs
jpcsmith: That would change the semantics of that service, which currently is solely responsible for extraction of the symbol and construction of the merkle tree, and the benefit is not clear to me. The overhead of reading the DB is currently counted inside the time for serving the overall GET request, but if we want something that encapsulates and reports metrics for each individual request, I would recommend a wrapper around the RecoverySymbolService that takes the database and delegates, at which point, however, it may make sense to have that service handle all the methods related to the recovery flow.

crates/walrus-service/src/node.rs
jpcsmith: Ideally, yes, this simply delays that change, since this should currently never need to be called when getting recovery symbols.

The options for this that I considered were (1) refactor self.storage to make it cheaply cloneable so that it can be moved to a background thread for each op, which seemed a lot of refactoring, (2) make get_metadata async, which seemed like even more refactoring, (3) add an additional variant of get_metadata that's async, by cloning the inner DBMap and serving from that (likely the easiest). In any-case, since the above changes means that get_metadata no longer affects the recovery symbol perf which is addressed in PR, I did not undertake the refactoring.

crates/walrus-service/src/node.rs
jpcsmith: Your understanding is correct. The request will begin processing. The first worker will be assigned to the first of the sub-requests, and the other sub-requests will then each also try to acquire a worker (later in the code), but instead of failing if no worker is ready, they will simply wait for a worker to become available. This results in an implicit additional queue. 

However, 
- once at the point where at least 1 request has at least 1 sub-request waiting then no further requests would be accepted; and
- since we've already gotten 1 permit, then once it's used for a sub-request it should become available for another sub-request and the requests could just be processed sequentially using that 1 permit, which may be better than failing when we were able to get (M >= 1)-out-of-N permits.

The worst case occurs when are a total of X permits (e.g., around 3000), all available, and X requests come in each for a symbol from the N shards (i.e., N subrequests). If the coroutines interleave such that each of the X requests gets 1 of the X permits, then the implicit queue becomes `(X - 1)N`. In practice, I would expect that this would be a small multiple of N, however, it's to quantify.

If we would like more certaintly, we can also just do an `acquire_n_permits_with_timeout_if_at_least_1_is_available(N, 50ms)` and fail the request if the timeout elapses before acquiring the N permits (N <= nShards), then the MAX time for a successful request increases by 50ms and the MAX time for a request failing due to to not getting a permit similar increases by 50 ms, whereas the average times will probably remain the same.

halfprice: What I'm worried about is the live lock issue. Permits may constantly be available, but the earliest task may not get it. Let's go with this for now, and see how it goes though. We probably already have pending metrics about requests.

PR #2678, opened 2025-10-31
docs(walrus-sites): update docs for quilts

Tzal3x:
## Description

- Updates the Walrus Sites docs in order to depict the new state of the `site-builder`. 
- Removes references to the publish/update commands since we use `deploy` from now on.
- Mentions the suiup tool as an alternative way of installing the site-builder.
- Fixes typos here and there. 

## Test plan

Proofread.

### DIFF
--- a/docs/book/walrus-sites/tutorial-publish.md
+++ b/docs/book/walrus-sites/tutorial-publish.md
@@ -140,8 +140,9 @@ resource corresponds to a quilt patch. This happens because the site-builder sto
 using quilts. This approach offers significant benefits: faster upload speeds and lower
 storage costs, especially when uploading many small files. The only disadvantage is that you cannot
 update a single file within a quilt - if even a tiny file changes, the entire quilt must be
-re-uploaded. Upcoming versions of the site-builder will introduce optimizations in the storage usage between updates.
-For a detailed overview of Quilts, visit this [section](https://docs.wal.app/usage/quilt.html) of the documentation.
+re-uploaded. Upcoming versions of the site-builder will introduce optimizations in the storage
+usage between updates. For a detailed overview of Quilts, visit this
+[section](https://docs.wal.app/usage/quilt.html) of the documentation.
 
 ```admonish note
 The wallet you are using must be the *owner* of the Walrus Site object to be able to update it.


docs/book/walrus-sites/tutorial-publish.md
```diff
@@ -35,46 +35,50 @@ duration of an epoch is one day. On Mainnet, the duration of an epoch is two wee
...
 
 ```admonish note
 Keep in mind that option 2 is only available on `mainnet`.
 For testnet sites, use option 1 - see the [portal setup guide](./portal.md) for instructions.
 ```
 
-This output tells you that, for each file in the folder, a new Walrus blob was created, and the
-respective blob ID. Further, it prints the object ID of the Walrus Site object on Sui (so you can
+This output tells you that, for each file in the folder, a new Walrus quilt was created, and the
+respective quilt ID. Further, it prints the object ID of the Walrus Site object on Sui (so you can
```
dalxds, line 78: is there a notion of a quilt ID? isn't it blob ID for the whole quilt or quilt patch id for the individual files?

docs/book/walrus-sites/tutorial-publish.md
```diff
@@ -91,45 +95,54 @@ Let's say now you want to update the content of the site, for example by changin
...
+            or find third-party hosted testnet portals.
 ```
 
-Compared to the when the site was first published, we can see that now the only actions performed
-were to delete the old `index.html`, and update it with the newer one.
-
 Browsing to the provided URL should reflect the change. You've updated the site!
 
+Notice that all site resources are deleted (because they all belong to the same quilt), even those
+that haven't been modified. They then get re-uploaded together as a single quilt, where each
```
dalxds, line 138: it could be more than one quilts if the files are >666

docs/book/walrus-sites/tutorial-publish.md
```diff
@@ -91,45 +95,54 @@ Let's say now you want to update the content of the site, for example by changin
...
+resource corresponds to a quilt patch. This happens because the site-builder stores files on Walrus
+using quilts by default. This approach offers significant benefits: faster upload speeds and lower
+storage costs, especially when uploading many small files. The only disadvantage is that you cannot
+update a single file within a quilt—if even a tiny file changes, the entire quilt must be
+re-uploaded. If you want to tweak with the updates mechanism, you should check out the `--strategy`
```
dalxds: should we remove the strategy section?

docs/book/walrus-sites/tutorial-publish.md
```diff
@@ -91,45 +95,54 @@ Let's say now you want to update the content of the site, for example by changin
...
 
-Compared to the when the site was first published, we can see that now the only actions performed
-were to delete the old `index.html`, and update it with the newer one.
-
 Browsing to the provided URL should reflect the change. You've updated the site!
 
+Notice that all site resources are deleted (because they all belong to the same quilt), even those
+that haven't been modified. They then get re-uploaded together as a single quilt, where each
+resource corresponds to a quilt patch. This happens because the site-builder stores files on Walrus
+using quilts by default. This approach offers significant benefits: faster upload speeds and lower
```
dalxds: might be a good opportunity to link to Quilts docs. 

> by default

could be ommited

docs/book/walrus-sites/tutorial-publish.md
nikos-terzo: I think Quilt ID == Blob ID, but yes, blob-id might make more sense. Quoting:

> `Quilt`: A collection of files that has been encoded using the Quilt encoding scheme
`Blob`: A file/quilt after it has been encoded using RedStuff - this becomes the fundamental storage unit in Walrus

docs/book/walrus-sites/tutorial-publish.md
nikos-terzo: I think this refers to the specific example demonstrated above, so I think it is fine.

docs/book/walrus-sites/tutorial-publish.md
nikos-terzo: Let's instead add that improvements for tiny-site-updates are planned?

nikos-terzo (approved):
As @dalxds mentioned, "by default" ~~could~~ should be omitted.
Other than that minor suggestions.

docs/book/walrus-sites/tutorial-publish.md
nikos-kitmeridis: I think quilt id makes more sense here.
The terminology used in the documentation is the following:

* A quilt is consisted of many blobs and is assigned of a QuiltId
* Each blob within a quilt has it's own id which is derived by the overall QuiltId and is called QuiltPatchId

Eg of quilt-id used as a flag:
> To read blobs by their identifiers, use the --identifiers flag:
>
> walrus read-quilt --out <download dir> \
  --quilt-id 057MX9PAaUIQLliItM_khR_cp5jPHzJWf-CuJr1z1ik --identifiers walrus.jpg another-walrus.jpg

PR #2681, opened 2025-10-31
chore: do not try write same blob when blob pool is empty; log error when getting blob status failed

halfprice:
## Description

- Do not try write same blob when blob pool is empty, this will cause client panic.
- Log error when getting blob status failed
- Fix request generation distribution to closely follow the configuration. The caveat before was that if the blob pool
is full, no writes to the same blob will occur. Now, if pool is full or empty and the corresponding picked operation is
not possible, we simply generate a None op.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -2157,16 +2157,21 @@ impl<T> WalrusNodeClient<T> {
                 {
                     return Err(client_error);
                 }
-                Err(error) => {
-                    tracing::debug!("fetching blob status failed: {:?}", error);
-                }
+                Err(_) => (),
             };
 
             if peekable.peek().is_some() {
-                tracing::debug!(?delay, "fetching blob status failed; retrying after delay");
+                tracing::debug!(
+                    ?delay,
+                    latest_status = ?maybe_status,
+                    "fetching blob status failed; retrying after delay",
+                );
                 tokio::time::sleep(delay).await;
             } else {
-                tracing::warn!("fetching blob status failed; no more retries");
+                tracing::warn!(
+                    latest_status = ?maybe_status,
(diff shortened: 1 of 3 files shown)


crates/walrus-sdk/src/client.rs
```diff
@@ -2157,7 +2157,9 @@ impl<T> WalrusNodeClient<T> {
                 {
                     return Err(client_error);
                 }
-                Err(_) => (),
+                Err(error) => {
+                    tracing::debug!("fetching blob status failed: {:?}", error);
```
mlegner: Nit: This change is not really necessary as we have a similar debug statement a few lines below.

crates/walrus-stress/src/single_client_workload/blob_pool.rs
```diff
@@ -210,15 +210,21 @@ impl BlobPool {
         }
     }
 
-    /// Returns true if the blob pool is empty.
-    pub fn is_empty(&self) -> bool {
-        self.blobs.is_empty()
+    /// Returns the number of blobs in the pool.
+    pub fn size(&self) -> usize {
```
mlegner: Nit: A more canonical name would be `len`.

crates/walrus-sdk/src/client.rs
halfprice: Thanks @mlegner for the quick review! The problem for the log line below is that they don't log the error. So I reverted this change, and added error details in the log lines below.

crates/walrus-stress/src/single_client_workload/blob_pool.rs
halfprice: Done

PR #2684, opened 2025-11-02
feat: blob manager prototype

liquid-helium:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


contracts/blob_manager/sources/blob_manager.move
```diff
@@ -0,0 +1,243 @@
...
+public struct BlobManager has key, store {
+    id: UID,
+    /// Storage management strategy.
+    storage: BlobStorage,
+    /// Blob stash strategy - handles both individual blobs and managed blobs.
+    blob_stash: BlobStash,
+}
+
+/// A capability which represents the authority to manage blobs in the BlobManager.
+public struct BlobManagerCap has key, store {
```
poelzi, line 48: How about adding a permission bitfield.
WRITE, DELETE, METADATA, ADMIN
Only Admin can create new ManagerCaps. This would allow more fine granular permissions. Give buildbots for example only permission to create new artifacts.

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+  - `delete()` - Removes deletable blobs and emits event (package visibility)
+  - `burn()` - Destroys managed blob without storage release (public)
+  - `derive_blob_id(root_hash, encoding_type, size)` - Derives blob ID (public)
+
+### 1.2 `blob_stash.move`
+**Purpose**: Set of ManagedBlob objects within a BlobManager.
+
+**Key Components**:
+- **BlobStashByBlobId struct** (simplified single-table design):
+  - Single `Table<u256, ManagedBlob>` keyed by blob_id
```
karlwuest, line 52: Suggestion/Remark: (Much of this applies to other parts of the design, but it fits here as context)
I assume that the main reason why the `ManagedBlob` has an object ID is for backwards compatibility to allow the same tracking on storage nodes for deletable blobs and to allow the same behaviour for blob attributes on the aggregator etc. If this is not needed and we can have another solution (e.g. we replace the Object ID for tracking deletable blobs on storage nodes with a hash of BlobManager ID & Blob ID, and enforce uniqueness per BlobManager, which seems to be the case anyway), then the objects do not need the `key` ability since they will always be stored in a BlobManager. This would save a significant amount of space in the actual sui objects.

On the other hand, without this ability, it is also not possible to look up the objects by Object ID from the object store, e.g., to fetch attributes. This means, it really depends on the goals of this design whether they should have `key` or not. However, if they do have `key`, then they should not be stored in a `Table` because that will prevent the lookup as well. Instead, they should be stored in an `ObjectTable` which preserves this.

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+  - Tracks `total_unencoded_size`
+  - Only one blob per blob_id (either permanent or deletable, not both)
```
karlwuest, line 54: Remark: I think this is fine in most cases, but there are some special cases where it isn't. Namely, if I have stored a blob as either deletable or permanent, and now I want to store it as the opposite. This would require deleting a blob or letting it expire and reuploading to change the flag. These scenarios can probably be handled separately, but I would guess that at some point, someone wants to do this and will get frustrated if we don't have a solution.

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+  - Permanency conflict error if attempting to register with different deletable flag
+
+- **BlobStash enum**:
+  - `BlobIdBased(BlobStashByBlobId)` - Current variant
```
karlwuest, line 58: Question: What does this enum do?

Remark: If this is intended for versioning, be aware that it is not possible to add variants to existing enums in move with contract upgrades.

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+
+- **ManagedBlobInfo struct**:
+  - `object_id: ID` - The ManagedBlob's object ID
+  - `is_certified: bool` - Certification status
+
+- **Dispatch Functions** (dispatch to variant):
+  - `find_blob_in_stash(blob_id, deletable)` - Find blob and check certification, returns `Option<ManagedBlobInfo>`
```
karlwuest, line 65: Question: Why have a separate `ManagedBlobInfo` struct instead of returning a shared reference to the `ManagedBlob`?

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+### 1.4 `blob_manager.move`
+**Purpose**: Main BlobManager implementation coordinating all components.
+
+**Key Components**:
+- **BlobManager struct**:
+  ```move
+  struct BlobManager {
+      id: UID,
+      storage: BlobStorage,
+      blob_stash: BlobStash,
```
karlwuest: Question/Remark: Why are the `BlobStorage` and `BlobStash` separate? This seems a bit dangerous to me because it makes it harder to ensure that invariants are enforced, in particular that adding a blob decreases the available storage and deleting a blob increases the available storage. It would seem more natural to me if the `BlobStash` was owned by the `BlobStorage`.

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+  - `delete_blob()` - Remove deletable blobs
+
+- **Blob Attribute Operations**:
+  - `set_blob_attribute(cap, blob_id, key, value)` - Set attribute
+  - `remove_blob_attribute(cap, blob_id, key)` - Remove attribute
+  - `clear_blob_attributes(cap, blob_id)` - Clear all attributes
+
+- **Community Funding**:
+  - `deposit_wal_to_coin_stash()` - Add WAL tokens
+  - `deposit_sui_to_coin_stash()` - Add SUI tokens
```
karlwuest, line 157: Remark: This is something where George (and maybe Alonso) may have some inputs on. Since the plan is to enable pricing in USD, it may be desirable to allow adding funds in a stablecoin to the coin stash to then exchange it for WAL at the moment of purchase. This would require taking a dependency on e.g. deepbook, in which case it should not be part of the core Walrus package.

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+  }
+  ```
+
+- **Capability Management**:
+  - `create_cap(is_admin, fund_manager)` - Returns new cap (PTB handles transfer)
+  - `cap_manager_id()` / `is_admin_cap()` / `is_fund_manager_cap()` - Accessors
+  - Permission rules:
+    - Admin caps can create new caps
+    - Only fund_manager caps can create new fund_manager caps
+    - Any cap can write blobs
```
karlwuest: Question/Remark: Is the intention that there can also be caps that are neither admin nor fund manager caps? I.e. that only allow writing? 

I think this would be useful in cases where we may want to burn the admin and fund manager caps to ensure the public that funds can only be used to extend the storage, but perhaps still allow writing new blobs. Otherwise people may generally be hesitant to donate to the coin stash if someone can just pull out the money.

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+
+**Key Components**:
+- **CoinStashByBalance struct**:
+  - Holds WAL and SUI balances
+  - No contribution tracking (simplified)
+
+- **Key Functions**:
+  - `deposit_wal()` / `deposit_sui()` - Accept contributions
+  - `withdraw_wal_for_storage()` - Use funds for storage
+  - `withdraw_sui_for_gas()` - Use funds for gas
```
karlwuest, line 187: Question: What is the idea here? There is no way to pay for gas from a shared object. Is the idea that a user can call this to get refunded for gas? If so, how do you ensure that only the correct amount is paid out?

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+
+---
+
+## 2. Modified Existing Modules
+
+### 2.1 `system/events.move`
+**New Event Types Added**:
+- `ManagedBlobRegistered` - Blob registered with BlobManager
+- `ManagedBlobCertified` - Blob certified
+- `ManagedBlobDeleted` - Blob deleted from BlobManager
```
karlwuest, line 198: Question: Is it necessary to have new event types here or can the old ones be used as well? If it is necessary, can we have new event types that work for both types of blobs and deprecate the old ones?

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+  - `withdraw_wal_for_storage()` - Use funds for storage
+  - `withdraw_sui_for_gas()` - Use funds for gas
+  - Balance query functions
+
+---
+
+## 2. Modified Existing Modules
+
+### 2.1 `system/events.move`
+**New Event Types Added**:
```
karlwuest, line 195: Question: I see that there is no new event to extend all Blobs from a BlobManager at once. How are extensions handled? Do we loop through all Blobs to send events? How do you ensure that we actually do this for all of them (since it won't be possible in a single transaction). 

Question/Suggestion: Do the storage nodes keep a mapping from `BlobManager` (or `BlobStash`) to Blobs? If so, could we just have a single event to extend all of them at once?

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+      FundManagerOnly,
+      Constrained {
+          expiry_threshold_epochs: u32,
+          max_extension_epochs: u32,
+      },
+  }
+  ```
+
+- **Policy Variants**:
+  - `Disabled` - No one can extend storage (blocks everyone including fund managers)
```
karlwuest: Question: What is the purpose for this?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,669 @@
...
+            system_max_epochs_ahead,
+        );
+
+    // Execute the extension with the effective epochs.
+    self.execute_extension(system, effective_extension, ctx)
+}
+
+/// Internal helper to execute the storage extension.
+/// Returns the effective extension epochs, or 0 if no extension possible.
+fun execute_extension(
```
karlwuest: Important: This doesn't seem to emit any events to inform storage nodes about the extensions.

contracts/blob_manager/sources/coin_stash.move
```diff
@@ -0,0 +1,151 @@
...
+public fun withdraw_sui_for_gas(
+    self: &mut BlobManagerCoinStash,
+    amount: u64,
+    ctx: &mut TxContext,
+): Coin<SUI> {
+    assert!(amount > 0, EInvalidWithdrawalAmount);
+    assert!(self.sui_balance.value() >= amount, EInsufficientSuiBalance);
+
+    coin::from_balance(self.sui_balance.split(amount), ctx)
+}
```
karlwuest, line 141: Important: These two functions just allow anyone to withdraw an arbitrary amount from the coin stash.

contracts/walrus/sources/system/messages.move
```diff
@@ -108,6 +108,11 @@ public struct CertifiedMessage has drop {
 public enum BlobPersistenceType has copy, drop {
     Permanent,
     Deletable { object_id: ID },
+    Managed {
+        blob_manager_id: ID,
+        deletable: bool,
+        blob_object_id: ID,
+    },
```
karlwuest: Question/Remark: Did you try to run the contract upgrade test (in rust)? Extending an enum by adding a variant should not be possible afaik.

karlwuest (commented):
Thanks for your effort on this @liquid-helium ! 
I added a few comments and questions, we can discuss them in our call. I focused mostly on the summary doc, but skimmed some of the contract changes a little bit.

contracts/walrus/blob_manager_contract_changes_summary.md
gdanezis: Lets start simple for the moment with just WAL and SUI indeed.

contracts/walrus/blob_manager_contract_changes_summary.md
```diff
@@ -0,0 +1,258 @@
...
+### 3.1 `extension_policy.move`
+**Purpose**: Controls when and how storage can be extended by anyone using the coin stash.
+
+**Key Components**:
+- **ExtensionPolicy enum**:
+  ```move
+  public enum ExtensionPolicy has store, copy, drop {
+      Disabled,
+      FundManagerOnly,
+      Constrained {
```
gdanezis, line 220: I like this constrained policy - what I like is to allow the FundManager for most of the time, but then open to public for the last X epochs - is this what you foresee?

contracts/walrus/blob_manager_contract_changes_summary.md
gdanezis: This is in-line with my feelings as well. Would it be simpler to say: manager can always extend blobs. But the policy allows the public to extend for `max_extension_epochs` within the `expiry_threshold_epochs` trailing epochs. Setting either of these to 0 means no public extension.

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+public struct BlobManager has key, store {
+    id: UID,
+    /// Unified storage that handles both capacity management and blob storage.
+    storage: BlobStorage,
+    /// Coin stash for community funding - embedded directly.
+    coin_stash: BlobManagerCoinStash,
+    /// Extension policy controlling who can extend storage and under what conditions.
+    extension_policy: ExtensionPolicy,
+    /// Tip policy controlling how much to tip transaction senders for helping with operations.
+    tip_policy: TipPolicy,
```
gdanezis: Should tip policy be integrated with the extension policy, as I suggested it is simplified above - namely a min tip, only for non managers when the params are both >0?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+/// A capability which represents the authority to manage blobs in the BlobManager.
+/// Admin can create new capabilities and perform all operations.
+/// Fund manager can withdraw funds from the coin stash.
+/// revocation for caps.
+public struct BlobManagerCap has key, store {
+    id: UID,
+    /// The BlobManager this capability is for.
+    manager_id: ID,
+    /// Whether this capability has admin permissions (can create new caps).
+    is_admin: bool,
```
gdanezis: should we call these `can_delegate`, and `can_withdraw_funds`?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    /// Whether this capability has fund manager permissions (can withdraw funds).
+    fund_manager: bool,
+}
+
+// === Constructors ===
+
+/// Creates a new shared BlobManager with deterministic grace period and returns its capability.
+/// The BlobManager is automatically shared in the same transaction.
+/// Requires minimum initial capacity of 100MB.
```
gdanezis: Is this 100MB or 500MB?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    system::{Self, System},
+    tip_policy::{Self, TipPolicy}
+};
+
+// === Constants ===
+
+/// Minimum initial capacity for BlobManager (500MB in bytes).
+const MIN_INITIAL_CAPACITY: u64 = 500_000_000; // 500 MB.
+
+// Grace period calculation thresholds (based on storage duration in epochs)
```
gdanezis: In all honesty if it makes things much simpler, I am not sure we need grace periods given the new abstraction that makes it easy to extend, and store stashes to ensure things are auto-extended.

contracts/blob_manager/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+/// If the creating cap has fund_manager = true, they can create new caps with fund_manager = true.
+/// Returns the newly created capability (caller/PTB handles transfer).
+public fun create_cap(
+    self: &BlobManager,
+    cap: &BlobManagerCap,
+    is_admin: bool,
+    fund_manager: bool,
+    ctx: &mut TxContext,
+): BlobManagerCap {
+    // Verify the capability matches this BlobManager.
```
gdanezis, line 149: I am not sure `self` is needed here, and the checks that ensure the cap correspond to the blob manager are not clearly needed. You can just copy the ID to the new cap from the old cap, and I think this is sufficient @karlwuest ? The checks ensure the blob manager maybe still exists, but I think this is it.

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    let manager_id = cap_manager_id(cap);
+    assert!(object::id(self) == manager_id, EInvalidBlobManagerCap);
+}
+
+/// Ensures the capability is an Admin capability.
+fun ensure_admin(cap: &BlobManagerCap) {
+    assert!(is_admin_cap(cap), ERequiresAdminCap);
+}
+
+/// Calculates grace period based on duration.
```
gdanezis: As before, not sure that grace periods are a must, so reducing code in this large PR may be preferable.

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    cap: &BlobManagerCap,
+    system: &System,
+    blob_id: u256,
+    deletable: bool,
+    signature: vector<u8>,
+    signers_bitmap: vector<u8>,
+    message: vector<u8>,
+) {
+    // Verify the capability.
+    check_cap(self, cap);
```
gdanezis: Do we strictly need the cap here to certify a blob? Why not allow anyone to certify a registered blob for the blob manager? There are checks below that the blob is registered. Now that the blob manager is shared by default allowing 3rd parties to finish the certification may give us more flexibility to implement publishers that are less trusted.

contracts/blob_manager/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+
+    // Delete the managed blob (this emits the ManagedBlobDeleted event).
+    managed_blob.delete(epoch);
+}
+
+// === Coin Stash Operations ===
+
+/// Deposits WAL coins to the BlobManager's coin stash.
+/// Anyone can deposit funds to support storage operations.
+public fun deposit_wal_to_coin_stash(self: &mut BlobManager, payment: Coin<WAL>) {
```
gdanezis, line 321: Lets not call this `payment` as this suggests it might transmit funds. Here we are only depositing funds for the extension of blobs. Maybe `coins` or `funds` is a better name?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+
+/// Deposits SUI coins to the BlobManager's coin stash.
+/// Anyone can deposit funds to support gas operations.
+public fun deposit_sui_to_coin_stash(self: &mut BlobManager, payment: Coin<SUI>) {
+    // Simply deposit to coin stash.
+    self.coin_stash.deposit_sui(payment);
+}
+
+/// Buys additional storage using funds from the coin stash.
+/// Anyone can call this to extend the BlobManager's capacity.
```
gdanezis: Here is says "anyone" but I see a cap check?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    let new_storage = system::reserve_space(
+        system,
+        storage_amount,
+        epochs_ahead,
+        &mut payment,
+        ctx,
+    );
+
+    // Return any unused funds to the stash.
+    if (payment.value() > 0) {
```
gdanezis: Do we need the check & destroy, or depositing zero just works in all cases?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+        system,
+        total_capacity,
+        extension_start_epoch,
+        new_end_epoch,
+        &mut payment,
+        ctx,
+    );
+
+    // Return any unused funds to the stash.
+    if (payment.value() > 0) {
```
gdanezis: Make deposit work for zero values?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+        self.coin_stash.deposit_wal(payment);
+    } else {
+        coin::destroy_zero(payment);
+    };
+
+    // Apply the extension to the BlobManager's storage.
+    self.storage.extend_storage(extension_storage);
+
+    // Tip the transaction sender for helping execute the extension.
```
gdanezis: Should we always tip or only tip the public extension, and not the manager extension?

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    // Ensure the caller has fund_manager permission.
+    assert!(cap.fund_manager, ERequiresFundManager);
+
+    // Withdraw all WAL.
+    self.coin_stash.withdraw_all_wal(ctx)
+}
+
+/// Withdraws all SUI funds from the coin stash.
+/// Requires fund_manager permission.
+public fun withdraw_all_sui(
```
gdanezis: Here since we have no way to pay SUI gas fees what is the function of stored SUI that can be withdrawn?

contracts/blob_manager/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    // Ensure the caller has fund_manager permission.
+    assert!(cap.fund_manager, ERequiresFundManager);
+
+    // Withdraw the specified amount of WAL.
+    self.coin_stash.withdraw_wal(amount, ctx)
+}
+
+/// Withdraws a specific amount of SUI funds from the coin stash.
+/// Requires fund_manager permission.
+public fun withdraw_sui(
```
gdanezis, line 378: Let's have a discussion about withdraw, and whether it is needed (at least initially, or in this generality).

contracts/blob_manager/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    // Ensure the caller has fund_manager permission.
+    assert!(cap.fund_manager, ERequiresFundManager);
+
+    // Withdraw all SUI.
+    self.coin_stash.withdraw_all_sui(ctx)
+}
+
+/// Withdraws a specific amount of WAL funds from the coin stash.
+/// Requires fund_manager permission.
+public fun withdraw_wal(
```
gdanezis, line 364: Let's have a discussion about withdraw, and whether it is needed (at least initially, or in this generality).

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+    // Verify the capability.
+    check_cap(self, cap);
+    // Ensure the caller has fund_manager permission.
+    assert!(cap.fund_manager, ERequiresFundManager);
+
+    // Withdraw the specified amount of SUI.
+    self.coin_stash.withdraw_sui(amount, ctx)
+}
+
+/// Sets the extension policy to fund_manager only.
```
gdanezis: See suggestions above that might simplify the following 3 function, ie setting a default admin can always extend + just parameter.

contracts/walrus/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+/// Sets the tip policy to a fixed amount.
+/// Requires fund_manager permission.
+public fun set_tip_policy_fixed_amount(
+    self: &mut BlobManager,
+    cap: &BlobManagerCap,
+    tip_amount: u64,
+) {
+    check_cap(self, cap);
+    assert!(cap.fund_manager, ERequiresFundManager);
+    self.tip_policy = tip_policy::new_fixed_amount(tip_amount);
```
gdanezis: Here, to help those who are looking for opportunities to collect tips: when we create a blob manager lets emit an event with its ID and the tip policy if it allows for public. And then upon updating the policy lets also emit an event. Those hunting for tips can follow such events and identify opportunities.

contracts/blob_manager/sources/blob_manager.move
```diff
@@ -0,0 +1,833 @@
...
+}
+
+// === Blob Attribute Operations ===
+
+/// Sets an attribute on a managed blob.
+///
+/// If the key already exists, the value is updated.
+/// Aborts if the blob is not found or if attribute limits are exceeded.
+/// Requires a valid BlobManagerCap to prove write access.
+public fun set_blob_attribute(
```
gdanezis, line 469: Should we allow for attributes to be set (optionally) when a blob is registered? This seems to be a dominant use-case, and might save folks the need to call this separate function in most cases.

contracts/walrus/sources/blob_storage.move
```diff
@@ -0,0 +1,254 @@
...
+    end_epoch: u32,
+    // === Blob Storage Fields ===
+    /// Maps blob_id directly to the ManagedBlob object (one blob per blob_id).
+    blobs: Table<u256, ManagedBlob>,
+    /// Total unencoded size of all blobs.
+    total_unencoded_size: u64,
+}
+
+/// Result of finding a managed blob with its certification status.
+public struct ManagedBlobInfo has drop {
```
gdanezis: Lets make this copy, store too, and also add the full blob info such as if it is deletable and how many epochs it is certified for? This way you can use it elsewhere are a proof of availability of the blob?

contracts/walrus/sources/blob_storage.move
```diff
@@ -0,0 +1,254 @@
...
+    /// The object ID of the managed blob.
+    object_id: ID,
+    /// Whether the blob is certified.
+    is_certified: bool,
+}
+
+/// Capacity information for the storage.
+public struct CapacityInfo has copy, drop {
```
gdanezis: Why not also store?

contracts/walrus/sources/blob_storage.move
```diff
@@ -0,0 +1,254 @@
...
+    // Destroy the storage object (we only use it for the end_epoch)
+    extension_storage.destroy();
+}
+
+// === Blob Management Functions ===
+
+/// Finds a matching blob that is already stored in the managed table and checks if it's certified.
+/// Returns ManagedBlobInfo if found, otherwise Option::none().
+/// This function combines finding and certification check to avoid multiple table lookups.
+public fun find_blob(self: &BlobStorage, blob_id: u256, deletable: bool): Option<ManagedBlobInfo> {
```
gdanezis: Here adding more info about duration & deletable status would allow the returned type to act as a proof of availability elsewhere.

contracts/walrus/sources/coin_stash.move
```diff
@@ -0,0 +1,192 @@
...
+    assert!(amount > 0, EInvalidWithdrawalAmount);
+    assert!(self.wal_balance.value() >= amount, EInsufficientWalBalance);
+
+    coin::from_balance(self.wal_balance.split(amount), ctx)
+}
+
+/// Tips the transaction sender with SUI from the stash.
+/// Used to reward users who help execute storage operations like extensions.
+/// Returns the tip coin if successful, or an empty coin if insufficient funds.
+public(package) fun tip_sender(
```
gdanezis: Why not tip in WAL and forego the SUI stash.

contracts/walrus/sources/blob_manager.move
gdanezis: Lets discuss an adaptive tip policy.

contracts/walrus/sources/extension_policy.move
```diff
@@ -0,0 +1,257 @@
+// Copyright (c) Walrus Foundation
+// SPDX-License-Identifier: Apache-2.0
+
+/// Extension policy for BlobManager storage extensions.
+/// Controls when and how storage can be extended by anyone using the coin stash.
+module walrus::extension_policy;
```
gdanezis: See comments above to simplify if possible.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: Thanks @karlwuest for the in-depth analysis!
As we discussed offline: we can remove the ObjectId for the managed blobs, beside the above reasons, another reason to keep the ObjectID around is due to the storage confirmation enum, adding a new BlobPersistenceType variant would require much rewrite to keep the backward compatibility.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: +1, this is mostly a product design question, whether there are use cases that a blob exists both as permanent and deletable in the same blob manager, and I really have no answer. In the future, with the addition of blob index, we can support this deletable alias concept, and I hope that can cover the use cases.

I added an API to allow converting a deletable blob in a blob manager into a permanent one, which was also suggested by @mlegner. 

Converting permanent to deletable is not allowed though.

With that said, if we really feel this can be a future blocker, we can use a vector to store the managed blobs in the blob_id map, so that we have the capability to support it if we want in the future.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: Thanks for the catch, it was intended for versioning, before the chat with you.
I replaced the enum with a plain struct.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: Initially, it is a workaround, since returning a Option<&ManagedBlob> caused errors in the MOVE contract.
I refactored this part, and removed the `ManagedBlobInfo`

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: That's a great idea!
This part is refactored as you suggested, PTAL!

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: Done, only SUI and WAL are present in the coin stash.
The SUI is used for tip, and the WAL is used to pay the write fees, to purchase more storage, and to extend the storage end_epoch.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: Good idea on the burning. 🔥 
Yes, there are `fund manager`, `admin`,  `writer` caps

liquid-helium: As discussed offline, a revoke_blob_manager_cap() API is added, and an admin can call it to revoke any caps.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: Replaced it with the tip, the idea is the same as giving a tip.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: We can do that, and per our discussion, keeping them separate is an simple/safe approach.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: The extension workflow aligns with what you described: 

1. The storage nodes keep track of all the blob managers.
2. Each blob in the storage nodes keeps track of all the blob managers that contains it, and its end_epoch is the max of all blob_managers'
3. When a blob manager is extended, a single event is sent out, which will update the corresponding blob manager's end_epoch in the storage nodes.
4. When GC, each blob is checked against the blob managers, and is deleted only when all blob managers have expired, (of course, the final decision is a combination with the regular blob references).

contracts/walrus/sources/blob_manager.move
liquid-helium: It is implemented now!

contracts/blob_manager/sources/coin_stash.move
liquid-helium: 😢 Fixed!

contracts/walrus/sources/system/messages.move
liquid-helium: I reverted this change, per our discussion offline.

contracts/walrus/sources/blob_manager.move
liquid-helium: Done.

contracts/walrus/sources/blob_manager.move
liquid-helium: 500MB. Sorry for the outdated comments!
In storage nodes, all blob managers associated with a blob are recorded in a vector within the blob's info table. We want to make it economically prohibitive to fabricate a blobInfo value with an excessive number of blob managers.

contracts/walrus/sources/blob_manager.move
liquid-helium: grace-period removed.

contracts/walrus/sources/blob_manager.move
liquid-helium: grace-period removed

contracts/walrus/sources/blob_manager.move
liquid-helium: That's a good point! Thanks!
Change applied as suggested.

contracts/walrus/sources/blob_manager.move
poelzi: I think having grace periode later is something definitely desirable. My thoughts some time ago were that, the blob becomes unavailable and you have to pay 2 times the storage cost of the epoch to resurrect it.

liquid-helium: > I think having grace periode later is something definitely desirable. My thoughts some time ago were that, the blob becomes unavailable and you have to pay 2 times the storage cost of the epoch to resurrect it.

Yes, making it as a potential future work makes more sense!
We can monitor the actual usage after the initial launch and decide if this is necessary.
Another consideration is: The changes required for a grace period are more complicated than I initially thought. When paying for the grace period, the payment should go to the committee for those past epochs, which requires substantial changes to the current contract.

contracts/blob_manager/sources/blob_manager.move
liquid-helium: Thanks for the suggestion, a more finer grained permission control is implemented, currently with admin, fund manager and writer, we can future improve it if necessary!
The metadata permission is really interesting, I will add it later!

contracts/walrus/sources/system/events.move
```diff
@@ -45,12 +45,88 @@ public struct BlobDeleted has copy, drop {
     was_certified: bool,
 }
 
+/// Signals that a managed blob has been registered.
+public struct ManagedBlobRegistered has copy, drop {
```
gdanezis: Question on the structure of events: are these state updates? Namely, is it the case that I can take the latest for a blob and use it to update the blob to the latest state on the storage node?

contracts/walrus/sources/system/blob_v2.move
```diff
@@ -0,0 +1,439 @@
...
+const MAX_ATTRIBUTES: u64 = 100;
+/// Maximum length of an attribute key in bytes.
+const MAX_ATTRIBUTE_KEY_LENGTH: u64 = 1024;
+/// Maximum length of an attribute value in bytes.
+const MAX_ATTRIBUTE_VALUE_LENGTH: u64 = 1024;
+
+// === Blob Type ===
+
+/// Type of blob: Regular or Quilt (composite blob).
+public enum BlobType has copy, drop, store {
```
gdanezis, line 65: Since we are redefining things here: is it an opportunity to only have quilt, and regular being a quilt with a single patch?

contracts/walrus/sources/system/blob_v2.move
```diff
@@ -0,0 +1,439 @@
...
+    self: &mut ManagedBlob,
+    current_epoch: u32,
+    end_epoch_at_certify: u32,
+    message: CertifiedBlobMessage,
+) {
+    // Check that the blob is registered in the system
+    assert!(self.blob_id == message.certified_blob_id(), EInvalidBlobId);
+
+    // Check that the blob is not already certified
+    assert!(!self.certified_epoch.is_some(), EAlreadyCertified);
```
gdanezis, line 295: Should we abort here, or just return?

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: > Would it be simpler to say: manager can always extend blobs. But the policy allows the public to extend for max_extension_epochs within the expiry_threshold_epochs trailing epochs. Setting either of these to 0 means no public extension.

Thanks folks for the suggestions: I revised this as the above suggestion.

contracts/walrus/blob_manager_contract_changes_summary.md
liquid-helium: Changed the policy as suggested, now FundManager can always extend the storage, while the community follows the above rules.

contracts/walrus/sources/blob_manager.move
liquid-helium: That's a great point.
As discussed in the meeting, the incentive could be very low with a small fixed amount, since it is a single transaction regardless of the size of the blob manager.

We can implement the time based adaptive algorithm, to reward more closer to the end of the storage.
Also, I will try to make the reward relevant to the used_capacity (the total encoded blob size). WDYT?

contracts/blob_manager/sources/blob_manager.move
liquid-helium: I updated this part, now the blob manager internally stores all the caps, so that the admin can revoke any of them if necessary.

contracts/walrus/sources/blob_manager/blob_manager_inner_v1.move
```diff
@@ -0,0 +1,760 @@
...
+    /// Coin stash for community funding - embedded directly.
+    coin_stash: BlobManagerCoinStash,
+    /// Storage purchase policy controlling both capacity purchases and time extensions.
+    storage_purchase_policy: StoragePurchasePolicy,
+    /// Maps capability IDs to their info, including revocation status.
+    caps_info: Table<ID, CapInfo>,
+}
+
+/// Information about a capability, including its revocation status.
+public struct CapInfo has drop, store {
```
liquid-helium, line 73: This cannot be public.

contracts/walrus/sources/blob_manager/blob_manager_inner_v1.move
```diff
@@ -0,0 +1,760 @@
...
+/// Stored as a dynamic field of BlobManager, keyed by version number.
+public struct BlobManagerInnerV1 has store {
+    /// Unified storage that handles both capacity management and blob storage.
+    storage: BlobStorage,
+    /// Coin stash for community funding - embedded directly.
+    coin_stash: BlobManagerCoinStash,
+    /// Storage purchase policy controlling both capacity purchases and time extensions.
+    storage_purchase_policy: StoragePurchasePolicy,
+    /// Maps capability IDs to their info, including revocation status.
+    caps_info: Table<ID, CapInfo>,
```
liquid-helium, line 69: revoked_ids:

contracts/walrus/sources/blob_manager/blob_manager_inner_v1.move
karlwuest: All structs in move are public

wbbradley:
Closing because this PR now overlaps with (and has drifted from) work ongoing in other PRs.

PR #2686, opened 2025-11-03
refactor(sdk): use type-state pattern for blob store state machine

mlegner:
## Description

This commit changes how we encode the state of to-be-stored blobs in the client. Instead of using a single enum with a variant for each state, we now use a type-state pattern with a separate type for each state.

This ensures that we always operate on blobs in the correct state, thus simplifying some code and avoiding runtime panics. We currently keep an owned reference to the blob object in the `UnencodedBlob` type, but this will probably be changed to a `Bytes` object in a follow-up commit.

The majority of changes are in `crates/walrus-sdk/src/client.rs`, `crates/walrus-sdk/src/client/client_types.rs`, and `crates/walrus-sdk/src/client/resource.rs`, but there are related changes in other locations as well.

All changes do not change any functionality (except logs), and are backwards-compatible.

Closes WAL-755
Closes WAL-962

## Test plan

- [x] Existing (slightly modified) test suite.
- [x] Manual tests of the CLI, aggregator, publisher.

### DIFF
--- a/crates/walrus-sdk/src/client/client_types.rs
+++ b/crates/walrus-sdk/src/client/client_types.rs
@@ -374,11 +374,11 @@ impl<S> WalrusStoreBlobState<S> {
 impl<S: WalrusStoreBlobStateApi> WalrusStoreBlobState<S> {
     /// Maps the blob state to a new blob state with the given function.
     ///
-    /// If the blob state is already in a success or failure state, it is returned unchanged. If it
-    /// is [`WalrusStoreBlobState::BlobInState`], the function is applied to the state. If the
-    /// function returns an error, the blob state is set to [`WalrusStoreBlobState::Failure`],
-    /// unless the error indicates that the operation should fail immediately, in which case the
-    /// error is returned.
+    /// If the blob state is already in a [WalrusStoreBlobState::Finished], it is returned
+    /// unchanged. If it is [WalrusStoreBlobState::Unfinished], the function is applied to the
+    /// state. If the function returns an error, the blob state is set to
+    /// [WalrusStoreBlobState::Finished] with an error, unless the error indicates that the
+    /// operation should fail immediately, in which case the error is returned.
     pub(crate) fn map<T, F>(
         self,
         f: F,
@@ -434,7 +434,9 @@ impl WalrusStoreBlobMaybeFinished {
                 unencoded_length: blob.len(),
                 encoding_config,
             },
-            state: WalrusStoreBlobState::new(UnencodedBlob { blob }),
+            state: WalrusStoreBlobState::new(UnencodedBlob {
+                unencoded_data: blob,
+            }),
         }
     }
 
@@ -478,16 +480,16 @@ impl<S: WalrusStoreBlobStateApi> WalrusStoreBlobMaybeFinished<S> {
 pub struct UnencodedBlob {
     /// The raw blob data to be stored.
     // TODO(WAL-1008): Consider using a `Bytes` object here instead.
-    pub blob: Vec<u8>,
+    pub unencoded_data: Vec<u8>,
 }
 
 impl Debug for UnencodedBlob {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("UnencodedBlob")
-            .field("blob_len", &self.blob.len())
+            .field("blob_len", &self.unencoded_data.len())
             .field(
-                "blob_prefix",
-                &walrus_core::utils::data_prefix_string(&self.blob, 5),
+                "data_prefix",
+                &walrus_core::utils::data_prefix_string(&self.unencoded_data, 5),
             )
             .finish()
     }
@@ -517,15 +519,15 @@ impl UnencodedBlob {
 
         let (metadata, data) = if let Some(upload_relay_client) = upload_relay_client {
             let metadata = encoding_config
-                .compute_metadata(&self.blob)
+                .compute_metadata(&self.unencoded_data)
                 .map_err(ClientError::other)?;
             (
                 metadata,
-                BlobData::BlobForUploadRelay(Arc::new(self.blob), upload_relay_client),
+                BlobData::BlobForUploadRelay(Arc::new(self.unencoded_data), upload_relay_client),
             )
         } else {
             let (pairs, metadata) = encoding_config
-                .encode_with_metadata(&self.blob)
+                .encode_with_metadata(&self.unencoded_data)
                 .map_err(ClientError::other)?;
             (metadata, BlobData::SliverPairs(Arc::new(pairs)))
         };
@@ -718,11 +720,11 @@ impl WalrusStoreEncodedBlobApi for RegisteredBlob {
 impl RegisteredBlob {
     /// Classifies the registered blob into either a blob that needs to be certified or a blob that
     /// is ready to be extended.
-    pub fn classify(self) -> Either<BlobToBeCertified, BlobReadyForCertifyAndExtend> {
+    pub fn classify(self) -> Either<BlobAwaitingUpload, BlobPendingCertifyAndExtend> {
         match self.operation {
             RegisterBlobOp::ReuseAndExtend {
                 epochs_extended, ..
-            } => Either::Right(BlobReadyForCertifyAndExtend {
+            } => Either::Right(BlobPendingCertifyAndExtend {
                 blob_object: self.blob_object,
                 operation: self.operation,
                 certificate: None,
@@ -738,7 +740,7 @@ impl RegisteredBlob {
                 } else {
                     None
                 };
-                Either::Left(BlobToBeCertified {
+                Either::Left(BlobAwaitingUpload {
                     encoded_blob: self.encoded_blob,
                     status: self.status,
                     blob_object: self.blob_object,
@@ -754,7 +756,7 @@ impl RegisteredBlob {
 // INV: The operation is NOT `ReuseAndExtend`.
 // INV: The epochs_extended is `Some` iff the operation is `ReuseAndExtendNonCertified`.
 #[derive(Debug, Clone, PartialEq)]
-pub struct BlobToBeCertified {
+pub struct BlobAwaitingUpload {
     /// The encoded blob.
     pub encoded_blob: EncodedBlob,
     /// The current status of the blob in the system.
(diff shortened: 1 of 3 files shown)


crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -37,7 +37,7 @@ pub struct StoreArgs {
     /// The metrics to use for the blob.
     pub metrics: Option<Arc<ClientMetrics>>,
     /// The optional upload relay client, that allows to store the blob via the relay.
-    pub upload_relay_client: Option<UploadRelayClient>,
+    pub upload_relay_client: Option<Arc<UploadRelayClient>>,
```
wbbradley, line 40: I think `UploadRelayClient` is already Send + Sync, no? It should be safe to clone. Maybe I'm thinking about it wrong.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -392,7 +392,8 @@ pub(super) async fn put_blob<T: WalrusWriteClient>(
     tracing::debug!("starting to store received blob");
     match client
         .write_blob(
-            &blob[..],
+            // TODO(WAL-1008): Can we avoid the clone here?
```
wbbradley, line 395: Bytes has 
```
impl From<Bytes> for Vec<u8> {
```
which does no copying.

mlegner: This depends on what the underlying data structure of the `Bytes` object is. The trait is implemented [here](https://github.com/tokio-rs/bytes/blob/master/src/bytes.rs#L1057-L1062) and (AFAIU) calls [`shared_to_vec`](https://github.com/tokio-rs/bytes/blob/master/src/bytes.rs#L1392-L1419) for a shared underlying buffer, which *does* copy the data unless it's the last reference. That said, I'm currently looking into whether I can generally use `Bytes` throughout the write path and encoding to avoid copying wherever possible.

crates/walrus-sdk/src/client/store_args.rs
mlegner: I was more trying to avoid copying too much data, but looking at it again, this seems rather small and the `reqwest::Client` is already an `Arc` internally, so should be fine to clone. 👍

mlegner: However, there is also no real downside of using the `Arc` AFAICT and it's still smaller than storing and cloning the `UploadRelayClient`. So I'd suggest to keep the `Arc` unless you feel strongly about it.

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -5,1826 +5,892 @@
...
+    fn from(value: WalrusStoreBlobUnfinished<S>) -> Self {
+        value.into_maybe_finished()
     }
+}
 
-    fn get_metadata(&self) -> Option<&VerifiedBlobMetadataWithId> {
-        None
-    }
+impl<S: WalrusStoreBlobStateApi> WalrusStoreBlobUnfinished<S> {
+    pub(crate) fn map_infallible<T: WalrusStoreBlobStateApi>(
```
liquid-helium, line 157: Nit: add a commment about the function, something like: "A state transition helper for WalrusStoreBlob that transforms the blob state using a function that cannot fail", if I understand it correctly

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -5,1826 +5,892 @@
...
+pub type WalrusStoreBlobUnfinished<S = UnencodedBlob> = WalrusStoreBlob<S>;
 
-    fn get_identifier(&self) -> &T {
-        self.input_blob.get_identifier()
-    }
+/// A blob that is being stored in Walrus, which can be in a particular state or finished
+/// (successfully or with an error).
+///
+/// See also [`WalrusStoreBlobState`].
+pub type WalrusStoreBlobMaybeFinished<S = UnencodedBlob> = WalrusStoreBlob<WalrusStoreBlobState<S>>;
```
liquid-helium, line 139: I haven't gone through all the changes, but is it possible to separate the UnFinished and Finished decisively? The state transition is deterministic, or did I miss something?

There is partition_unfinished_finished, and try_finish, in theory it is possible to get rid of the MaybeFinished?

liquid-helium (commented):
Really appreciate you taking this on, @mlegner — the refactor looks excellent! Just wrapped up my initial review. Thanks for introducing this cleaner pattern, I definitely learned something new from this approach.

crates/walrus-sdk/src/client/client_types.rs
mlegner: Thanks. I added some missing docstrings in df3f227d461f90a41c06cb03fb0aca9c5c241aeb.

crates/walrus-sdk/src/client/client_types.rs
mlegner: There are a few cases where a state transition can fail or lead to a finished state. For example [when encoding blobs](https://github.com/MystenLabs/walrus/blob/5bec4acbdb89afa014b7f2a3d56ec54290890348/crates/walrus-sdk/src/client.rs#L1107), when [adding the status](https://github.com/MystenLabs/walrus/blob/5bec4acbdb89afa014b7f2a3d56ec54290890348/crates/walrus-sdk/src/client.rs#L1293), or [including the certificate result](https://github.com/MystenLabs/walrus/blob/df3f227d461f90a41c06cb03fb0aca9c5c241aeb/crates/walrus-sdk/src/client/client_types.rs#L778).

It would be possible to handle these cases differently, but I believe it would be more complicated. Let me know if you feel strongly about this, then I can investigate the possibilities further.

liquid-helium: Thanks for the pointers! IIUIC, the complexity comes from the fact that some transitions produces 3 possible outcomes: {finished, unfinished, Err}, and the MaybeFinished is an superposition of {finished | unfinished}, and the callers can observe it with try_finish(). Is it possible to user a Either::Left/right to represent it, and we always get a decisive results, without the need to call try_finish?

I don't have a strong preference on this, the current refactor has already simplified the code significantly!

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -5,1826 +5,908 @@
...
-        ))
+}
+impl<S: WalrusStoreBlobStateApi> WalrusStoreBlobState<S> {
+    /// Maps the blob state to a new blob state with the given function.
+    ///
+    /// If the blob state is already in a success or failure state, it is returned unchanged. If it
+    /// is [`WalrusStoreBlobState::BlobInState`], the function is applied to the state. If the
+    /// function returns an error, the blob state is set to [`WalrusStoreBlobState::Failure`],
+    /// unless the error indicates that the operation should fail immediately, in which case the
+    /// error is returned.
```
liquid-helium: nit: are the referenced types outdated?

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -5,1826 +5,908 @@
...
-    pub input_blob: UnencodedBlob<'a, T>,
-    /// The blob ID.
-    pub blob_id: Option<BlobId>,
-    /// The phase where the error occurred.
-    pub failure_phase: String,
-    /// The error.
-    pub error: ClientError,
+/// A blob before it is encoded.
+#[derive(Clone, PartialEq)]
+pub struct UnencodedBlob {
```
liquid-helium, line 480: Do we want to add Attributes to Unencoded blobs, there has been several discussions about generalize the representation of a input blob.

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -5,1826 +5,908 @@
...
+    /// The raw blob data to be stored.
+    // TODO(WAL-1008): Consider using a `Bytes` object here instead.
+    pub blob: Vec<u8>,
```
liquid-helium: super nit: call this file?

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -5,1826 +5,908 @@
...
-        result: Result<(Vec<SliverPair>, VerifiedBlobMetadataWithId), ClientError>,
-    ) -> ClientResult<WalrusStoreBlob<'a, T>> {
-        Err(invalid_operation_for_blob(
-            &self,
-            format!("with_encode_result: {result:?}"),
-        ))
-    }
+        blob_object: Blob,
+        operation: RegisterBlobOp,
+    ) -> WalrusStoreBlobState<RegisteredBlob> {
```
liquid-helium, line 627: nit: this transfer a BlobWithStatus into a  WalrusStoreBlobState<RegisteredBlob>, I believe it is correct, but do we want to let the transition happen in the WalrusStoreBlob layer?

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -5,1826 +5,908 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub struct RegisteredBlob {
+    /// The encoded blob.
+    pub encoded_blob: EncodedBlob,
+    /// The current status of the blob in the system.
+    pub status: BlobStatus,
+    /// The blob object that was registered.
+    pub blob_object: Blob,
+    /// The operation that is performed to store the blob.
+    pub operation: RegisterBlobOp,
```
liquid-helium, line 709: Praise: thanks for simplify the operation type!

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -5,1826 +5,908 @@
...
-            Ok(WalrusStoreBlob::Error(FailedBlob {
-                input_blob: self.input_blob,
-                blob_id,
-                failure_phase: "with_certificate".to_string(),
-                error,
-            }))
+impl RegisteredBlob {
+    /// Classifies the registered blob into either a blob that needs to be certified or a blob that
+    /// is ready to be extended.
+    pub fn classify(self) -> Either<BlobToBeCertified, BlobReadyForCertifyAndExtend> {
```
liquid-helium: nit: The name BlobToBeCertified is a bit unclear (likely from my earlier refactor). Consider renaming to something more descriptive like BlobPendingCertification or BlobAwaitingUpload to better reflect its state.

liquid-helium (approved):
Really amazing work, thanks @mlegner!

LGTM. I left only some minor comments.

crates/walrus-sdk/src/client/client_types.rs
mlegner: > Is it possible to user a Either::Left/right to represent it, and we always get a decisive results, without the need to call try_finish?

I think this is possible, yes, but I'm not sure it actually simplifies matters much. I'll leave it as it is, but we can revisit this in the future if needed.

crates/walrus-sdk/src/client/client_types.rs
mlegner: Indeed, thank you. Fixed. 

I'll also add `--document-private-items` to our docs check in a separate PR to catch such issues in the future.

crates/walrus-sdk/src/client/client_types.rs
mlegner: We have the attribute in [the `CommonBlobParameters`](https://github.com/MystenLabs/walrus/blob/a1aa388ae2ce89c6141b0b721ba69d036a546013/crates/walrus-sdk/src/client/client_types.rs#L96). Let me know if this is not sufficient, then we can change this in a follow-up.

crates/walrus-sdk/src/client/client_types.rs
mlegner: Good point. I changed this to `unencoded_data`.

crates/walrus-sdk/src/client/client_types.rs
mlegner: This is lifted to the `WalrusStoreBlob` level with a `map_infallible` [here](https://github.com/MystenLabs/walrus/blob/a8ffcb43ed89720939a2d1fc037f38d810a88a01/crates/walrus-sdk/src/client/resource.rs#L444).

crates/walrus-sdk/src/client/client_types.rs
mlegner: Good idea. I also slightly changed `BlobReadyForCertifyAndExtend` to `BlobPendingCertifyAndExtend`. I also changed corresponding variable names.

PR #2697, opened 2025-11-07
feat: Move upload mode to client config

sadhansood:
## Description

Added `communication_config.upload_mode ` so a preset can be selected per context while still yielding to explicit concurrency limits, and wired it into the load path so presets are applied as the config is parsed.
Removed the `--upload-mode` CLI knob and the `UploadModeCli` plumbing. Child upload processes now automatically reuse the serialized config without needing extra flags.

## Test plan

Running locally


halfprice (approved):
LGTM!

PR #2702, opened 2025-11-10
fix(client): Prevent repeated copying of blob bytes when uploading to relay

sadhansood:
## Description

This PR adds a simple fix to prevent calling `blob.to_vec()` repeatedly on every retry attempt to upload the blob.

## Test plan

Local run

### DIFF
--- a/crates/walrus-sdk/src/client/upload_relay_client.rs
+++ b/crates/walrus-sdk/src/client/upload_relay_client.rs
@@ -11,6 +11,7 @@ use std::{
 };
 
 use anyhow::Result;
+use bytes::Bytes;
 use rand::{Rng, RngCore, rngs::ThreadRng};
 use reqwest::{Response, Url};
 use serde::Serialize;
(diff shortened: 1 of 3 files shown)


crates/walrus-sdk/src/client/upload_relay_client.rs
```diff
@@ -313,23 +314,18 @@ impl UploadRelayClient {
...
-                .send()
-                .await
-        };
+        let post_fn = |body: Bytes| async { client.post(post_url.clone()).body(body).send().await };
 
         while let Some(delay) = backoff.next_delay() {
-            let response = post_fn(blob.to_vec()).await;
-
-            match response {
+            match post_fn(payload.clone()).await {
```
wbbradley, line 328: nit for logging: Let's move the `attempts += 1` line up to just before the `tracing::debug!` line, so that we can see that "attempts" is the count of attempts so far, instead of the 0-based number.

PR #2703, opened 2025-11-11
feat(client): Unblock blob encoding to run without downloading committee and shard assignments

sadhansood:
## Description

Today, we cannot start blob encoding until `WalrusNodeClient` when called from `get_contract_client` downloads the full committee and shard assignment before becoming usable. But blob encoding doesn't really need any of that. All it needs is `n_shards`.
In this PR, we make `WalrusNodeClient` no longer blocks on full committee hydration before becoming usable. When we build a contract client we now read `n_shards` directly from the staking object and return the client immediately. The committees refresher still runs in the background and takes over once it finishes loading the full committee data.

This change cut ~1.5 s off CLI startup - the encoding and registration phases can begin as soon as the shard count is known, while the committee fan-out now happen concurrently in the refresher task.

## Test plan

Ran locally and observed speed-up.

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -193,32 +193,35 @@ pub struct WalrusNodeClient<T> {
 
 impl WalrusNodeClient<()> {
     /// Creates a new Walrus client without a Sui client.
+    #[allow(clippy::unused_async)]
     pub async fn new(
         config: ClientConfig,
         committees_handle: CommitteesRefresherHandle,
     ) -> ClientResult<Self> {
-        Self::new_inner(config, committees_handle, None, None).await
+        Self::new_inner(config, committees_handle, None, None)
     }
     /// Creates a new Walrus client without a Sui client.
+    #[allow(clippy::unused_async)]
     pub async fn new_with_max_blob_size(
         config: ClientConfig,
         committees_handle: CommitteesRefresherHandle,
         max_blob_size: Option<u64>,
     ) -> ClientResult<Self> {
-        Self::new_inner(config, committees_handle, None, max_blob_size).await
+        Self::new_inner(config, committees_handle, None, max_blob_size)
     }
 
     /// Creates a new Walrus client without a Sui client, that records metrics to the provided
     /// registry.
+    #[allow(clippy::unused_async)]
     pub async fn new_with_metrics(
         config: ClientConfig,
         committees_handle: CommitteesRefresherHandle,
         metrics_registry: Registry,
     ) -> ClientResult<Self> {
-        Self::new_inner(config, committees_handle, Some(metrics_registry), None).await
+        Self::new_inner(config, committees_handle, Some(metrics_registry), None)
     }
 
-    async fn new_inner(
+    fn new_inner(
         config: ClientConfig,
         committees_handle: CommitteesRefresherHandle,
         metrics_registry: Option<Registry>,
@@ -226,17 +229,9 @@ impl WalrusNodeClient<()> {
     ) -> ClientResult<Self> {
         tracing::debug!(?config, "running client");
 
-        // Request the committees and price computation from the cache.
-        let (committees, _) = committees_handle
-            .send_committees_and_price_request(RequestKind::Get)
-            .await
-            .map_err(ClientError::other)?;
-
-        let encoding_config = EncodingConfig::new(committees.n_shards());
+        let encoding_config = Arc::new(EncodingConfig::new(committees_handle.n_shards()));
         let communication_limits =
             CommunicationLimits::new(&config.communication_config, encoding_config.n_shards());
-
-        let encoding_config = Arc::new(encoding_config);
 
         Ok(Self {
             sui_client: (),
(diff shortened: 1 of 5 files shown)


crates/walrus-sdk/src/config/committees_refresh_config.rs
```diff
@@ -45,40 +46,36 @@ impl CommitteesRefreshConfig {
...
+        let n_shards = sui_client
+            .clone()
+            .n_shards()
+            .await
+            .context("failed to determine n_shards before starting refresher")?;
+
+        let notify = Arc::new(Notify::new());
+        let (req_tx, req_rx) = mpsc::channel(self.refresher_channel_size);
+        let handle = CommitteesRefresherHandle::new(notify.clone(), req_tx, n_shards);
+        let config = self.clone();
```
sadhansood, line 59: Here, is the main change. We no longer await the `build_refresher_and_handle()` anymore, but instead create and run refresher in the background task.

crates/walrus-sui/src/client/read_client.rs
```diff
@@ -1208,6 +1211,10 @@ impl ReadClient for SuiReadClient {
         ))
     }
 
+    async fn n_shards(&self) -> SuiClientResult<NonZeroU16> {
+        Ok(self.get_staking_object().await?.inner.n_shards)
+    }
```
mlegner, line 1215: Suggestion: How about we add the number of shards to the client configuration file (see WAL-969)? Then we wouldn't even need to get the staking object.

This can be done as a follow-up as well.

crates/walrus-sdk/src/config/committees_refresh_config.rs
```diff
@@ -45,40 +46,36 @@ impl CommitteesRefreshConfig {
     /// returns the [`CommitteesRefresherHandle`].
     pub async fn build_refresher_and_run(
         &self,
-        sui_client: impl ReadClient + 'static,
+        sui_client: impl ReadClient + Clone + 'static,
     ) -> Result<CommitteesRefresherHandle> {
-        let (mut refresher, handle) = build_refresher_and_handle(sui_client, self.clone()).await?;
+        let n_shards = sui_client
+            .clone()
```
mlegner: Question: Why do we need the clone here?

crates/walrus-sdk/src/config/committees_refresh_config.rs
mlegner: Very nice! 💯

crates/walrus-sdk/src/client.rs
```diff
@@ -180,15 +180,35 @@ impl WalrusNodeClient<()> {
...
+            Self::build_encoding_artifacts(&config, committees_handle.n_shards());
+        Self::new_inner(
+            config,
+            committees_handle,
+            encoding_config,
+            communication_limits,
+            None,
+            max_blob_size,
+        )
+        .await
```
mlegner: Question: Why do we need the separate `build_encoding_artifacts` function? Could we not do this in the `new_inner` function?

mlegner (commented):
Thank you so much for this PR, @sadhansood, this is a simple but huge improvement. I have two questions and one suggestion to speed this up even more (maybe in a follow-up).

crates/walrus-sui/src/client/read_client.rs
mlegner: Actually, I just went ahead and did this in #2705. PTAL. 🙏

crates/walrus-sdk/src/config/committees_refresh_config.rs
mlegner: Removed in #2705.

crates/walrus-sui/src/client/read_client.rs
sadhansood: Appreciate the idea @mlegner! But i think reading `n_shards` from the staking object until `WAL‑969` fully lands is fine, lmk otherwise. It’s the canonical value and already cached via `SuiReadClient`, and keeping it out of the config prevents skew if the network reconfigures (devnet, etc).

crates/walrus-sdk/src/config/committees_refresh_config.rs
sadhansood: Good catch - `build_refresher_and_run` now uses the owned instance we receive, so the inner clone() is gone.

crates/walrus-sdk/src/client.rs
sadhansood: Dropped the helper - `new_inner` now builds the encoding config and communication limits directly so all constructors share the same path.

crates/walrus-sdk/src/config/committees_refresh_config.rs
```diff
@@ -45,40 +46,35 @@ impl CommitteesRefreshConfig {
     /// returns the [`CommitteesRefresherHandle`].
     pub async fn build_refresher_and_run(
         &self,
-        sui_client: impl ReadClient + 'static,
+        sui_client: impl ReadClient + Clone + 'static,
```
mlegner: Nit: I think this can be reverted.

```suggestion
        sui_client: impl ReadClient + 'static,
```

crates/walrus-sui/src/client/read_client.rs
mlegner: Here's the new more holistic PR: #2710

PR #2705, opened 2025-11-11
feat(client): allow specifying number of shards in client config

mlegner:
## Description

As a result, we can start the encoding even without reading the staking object.

Also removes some unnecessary `async`s.

Contributes to WAL-969.

This PR can be merged into #2703 or into main after #2703.

## Test plan

Existing test suite + manual tests.

---

## Release notes

- [x] CLI: Add a parameter to the `client_config.yaml` to specify the number of shards. You can obtain the latest version of the config file at http://docs.wal.app/setup/client_config.yaml.


sadhansood:
> @sadhansood I'm closing this in favor of a more holistic PR. Still working on that one though as it doesn't deliver the expected improvements yet.

Make sense @mlegner, lmk once it is ready!

PR #2708, opened 2025-11-14
feat: create byte range read client API

halfprice:
## Description

This PR adds a new byte range read API in the SDK crate. Given a byte range and a BlobID, the client reads corresponding
slivers and extracts the byte range from the sliver.

Currently, if the sliver is unavailable, the client returns error.

Next step:
- enable single sliver recovery on the client
- aggregator supports byte range reads
- CLI support byte range reads

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sdk/src/client/byte_range_read_client.rs
+++ b/crates/walrus-sdk/src/client/byte_range_read_client.rs
@@ -9,6 +9,7 @@ use serde::{Deserialize, Serialize};
 use tokio::time::Instant;
 use walrus_core::{
     BlobId,
+    EncodingType,
     Epoch,
     SliverIndex,
     encoding::{EncodingFactory, Primary, SliverData},
@@ -103,6 +104,16 @@ impl<T: ReadClient> ByteRangeReadClient<'_, T> {
             .client
             .retrieve_metadata(certified_epoch, blob_id)
             .await?;
+
+        if metadata.metadata().encoding_type() != EncodingType::RS2 {
+            return Err(ClientError::from(ClientErrorKind::ByteRangeReadError(
+                format!(
+                    "byte range read client only supports RS2 encoding, got {}",
+                    metadata.metadata().encoding_type()
+                ),
+            )));
+        }
+
         let blob_size = usize::try_from(metadata.metadata().unencoded_length()).map_err(|_| {
             ClientError::from(ClientErrorKind::ByteRangeReadError(format!(
                 "invalid blob size from metadata: {}",


halfprice:
Thanks @mlegner for the feedback! I agree with them and this PR should be compatible with your concerns. Both QuiltClient and ByteRangeReaderClient is a wrapper around [WalrusNodeClient::retrieve_slivers_retry_committee](https://github.com/MystenLabs/walrus/blob/11e8104d3812c8721438454bfb89e09eb430ac33/crates/walrus-sdk/src/client.rs#L654). The client itself implements its own logic of figuring out which slivers to read, which is client specific. So there is no duplicated logic in these client.

> Thanks for the PR, @halfprice. This will be very useful. I still have to look at all the code in `crates/walrus-sdk/src/client/byte_range_read_client.rs` in detail, but have one overall question/concern:
> 
> It seems that there is some overlap between this new client and what the `QuiltClient` does. For example, the corresponding configurations (`ByteRangeReadClientConfig` and `QuiltClientConfig`) are identical. The `QuiltReader::download_data` and `ByteRangeReadClient::retrieve_slivers_for_range` do almost the same.

Yes, I agree. I thought about combining them into one `RetrieveSliverConfig` and use in both clients. These config is just
used to call retrieve_slivers_retry_committee. If @liquid-helium is ok with that, I can make the change.

> 
> This duplication means that if/when we want to implement sliver recovery at the client, we need to do this in multiple places as well. It also increases the danger of introducing inconsistent behavior.

Valid concern, but this won't be the case. The single sliver recovery will be implemented inside retrieve_slivers_retry_committee, which will be used in both QuiltClient and ByteRangeReadClient

> 
> My feeling is that it would be better to have general sub-blob read functionality in the `WalrusNodeClient` (or possibly a specialized client like the `ByteRangeReadClient`) and then use that for reading quilts as well. WDYT? Also cc @liquid-helium.

`retrieve_slivers_retry_committee` is the one that already exist in WalrusNodeClient for the purpose of `sub-blob` read. We can later add one for reading byte ranges inside a sliver, but that can be an incremental optimization.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+}
+
+/// Calculates the sliver indices that contain the byte range, and the start byte position of
+/// the first sliver that we need to read from.
+///
+/// `byte_length` must be greater than 0.
+fn calculate_sliver_indices(
+    primary_sliver_size: usize,
+    start_byte_position: usize,
+    byte_length: usize,
```
mlegner: Question/suggestion: Can we use a `NonZeroUsize` here and in related places?

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+        // After the first sliver, we don't need to skip any more bytes.
+        bytes_to_skip = 0;
+
+        // Early exit if we've collected all needed bytes
+        if bytes_remaining == 0 {
+            break;
+        }
+    }
+
+    Ok(result)
```
sadhansood, line 331: Important: I think `bytes_remaining` only goes to zero if the concatenated sliver.symbols.data().len() values cover the requested window. If the overall sliver list is partial, or a sliver is misconfigured I think we end up returning an Ok() but a wrong result.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+            .await?;
+
+        // Retrieve metadata
+        let metadata = self
+            .client
+            .retrieve_metadata(certified_epoch, blob_id)
+            .await?;
+        let blob_size = usize::try_from(metadata.metadata().unencoded_length())
+            .expect("blob size should be valid");
+
```
sadhansood, line 123: Instead of "blob size should be valid", should we instead say "blob too large for this platform" ?

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -2811,3 +2811,60 @@ async fn test_store_with_upload_relay_with_tip() {
...
+        node_weights: vec![7, 7, 7, 7, 7],
+        ..Default::default()
+    };
+    let test_cluster_builder =
+        test_cluster::E2eTestSetupBuilder::new().with_test_nodes_config(test_nodes_config);
+    let (_sui_cluster_handle, _cluster, client, _) = test_cluster_builder.build().await?;
+    let client = client.as_ref();
+
+    // Generate a random blob.
+    let blob_size = random::<usize>() % 1000000;
```
sadhansood: Could `random::<usize>() % 1000000` this not be 0 as well?

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+fn calculate_sliver_indices(
+    primary_sliver_size: usize,
+    start_byte_position: usize,
+    byte_length: usize,
+) -> ClientResult<(usize, Vec<SliverIndex>)> {
+    tracing::debug!("primary_sliver_size: {:?}", primary_sliver_size);
+
+    assert!(byte_length > 0, "byte_length must be greater than 0");
+
+    // Calculate which slivers contain the byte range
```
sadhansood: Why not return a `ClientError`? What is the benefit of panicking here?

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+#[serde_with::serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq)]
+pub struct ByteRangeReadClientConfig {
+    /// The maximum number of attempts to retrieve slivers.
+    pub max_retrieve_slivers_attempts: usize,
+    /// The timeout duration for retrieving slivers.
+    #[serde_as(as = "serde_with::DurationSeconds")]
+    #[serde(rename = "timeout_secs")]
+    pub timeout: Duration,
+}
```
mlegner, line 35: As discussed [here](https://github.com/MystenLabs/walrus/pull/2708#issuecomment-3533582022), this could be combined with the `QuiltClientConfig`.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+impl<T: ReadClient> ByteRangeReadClient<'_, T> {
+    /// Reads a specific byte range from a blob.
+    pub async fn read_byte_range(
+        &self,
+        blob_id: &BlobId,
+        start_byte_position: usize,
+        byte_length: usize,
+    ) -> ClientResult<Vec<u8>> {
+        tracing::debug!(
+            ?blob_id,
```
mlegner: Nit:

```suggestion
            %blob_id,
```

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+            return Ok(vec![]);
+        }
+
+        // Validate blob ID
+        self.client.check_blob_id(blob_id)?;
+
+        // Get blob status and certified epoch
+        let (certified_epoch, _) = self
+            .client
+            .get_blob_status_and_certified_epoch(blob_id, None)
```
mlegner, line 99: Nit (preexisting): We should probably rename that function because it only gets the status under special circumstances.

crates/walrus-sdk/src/client/byte_range_read_client.rs
mlegner: In fact we should probably return an error instead of panicking.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+            calculate_sliver_indices(primary_sliver_size, start_byte_position, byte_length)?;
+
+        // Retrieve the necessary slivers. Note that the returned slivers may not be in the same
+        // order as the sliver_indices.
+        let mut slivers = self
+            .retrieve_slivers_for_range(&metadata, &sliver_indices, certified_epoch)
+            .await?;
+
+        // Sort slivers by index to process them in order
+        slivers.sort_by_key(|s| s.index.get());
```
mlegner: Suggestion: Move the sorting into either the `retrieve_slivers_for_range` or `construct_requested_data_from_slivers` function.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,415 @@
...
+                ),
+            )));
+        }
+
+        // Copy the needed bytes directly into result
+        let start_in_sliver_inclusive = bytes_to_skip;
+        let end_in_sliver_exclusive = (start_in_sliver_inclusive + bytes_remaining).min(data.len());
+
+        // Copy only the needed bytes directly into result
+        result.extend_from_slice(&data[start_in_sliver_inclusive..end_in_sliver_exclusive]);
```
mlegner, line 303: Remark: When implementing some optimizations for the encoding, I noticed lower performance of the `extend` functions compared to something like `copy_from_slice`. That said, this might not be practically relevant here.

crates/walrus-e2e-tests/tests/test_client.rs
halfprice: Good call! Fixed.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Done

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Changed to returning an error

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Moved to `retrieve_slivers_for_range`

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Done

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: I followed Markus's suggestion to change this to NonZeroUsize. In terms of why using assert, this is an invariant of the function, so I use assert. But I agree that returning an error is probably better to not abruptly crash the process if something goes wrong.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Good catch! Checked that bytes_remaining is 0 after the loop, and returns an error if it is. Also added a test for this.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Yeah, I'd still like to hear @liquid-helium 's opinion if it's ok to merge them. Even we decide merge them, we probably should still keep the quilt one for a while since removing it could break existing clients who are using it (not sure if there exists such client though).

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Ok, let me tackle this in a separate PR. I need to understand the nuance a bit better.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: I tried to do only one copy here, which is to move from the returned sliver to the final location. Not sure if there are better ways to avoid copying.

crates/walrus-sdk/src/client/byte_range_read_client.rs
liquid-helium: Sorry for the delay on reviewing this PR – I got pulled into the new blob_manager prototype and it ended up taking more time than expected. I'll get to the review today.
On the technical side, I'm thinking we could extract the common parts into a client-side sliver retrieval config struct and reference that in QuiltConfig. WDYT?

liquid-helium: After discussing with @halfprice offline, I realized I had overlooked the compatibility concerns. I agree it makes sense to keep things as-is—breaking compatibility probably isn't worth it.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -85,6 +85,8 @@ impl<T: ReadClient> ByteRangeReadClient<'_, T> {
             return Ok(vec![]);
         }
 
+        let byte_length = NonZeroUsize::new(byte_length).expect("byte length must be non zero");
```
mlegner: Nit: These could be combined:
```rust
        let Some(byte_length) = NonZeroUsize::new(byte_length) else {
            return Ok(vec![]);
        };
```

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -221,7 +225,7 @@ fn calculate_sliver_indices(
     // Each systematic (primary) sliver contains primary_sliver_size bytes of the original blob
     // Note that we subtract 1 from the end byte position because we want to include the last
     // byte in the last sliver.
-    let end_byte_position = start_byte_position + byte_length - 1;
+    let end_byte_position = start_byte_position + byte_length.get() - 1;
 
     // Last sliver that contains end_byte_position
     let end_sliver_index = end_byte_position / primary_sliver_size;
```
mlegner, line 249: Important: We need to validate that these are valid or document that this function must be called *after* `calculate_and_validate_end_byte_position`.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Good suggestion!

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Added a comment.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,475 @@
...
+
+impl<'a, T> ByteRangeReadClient<'a, T> {
+    /// Creates a new ByteRangeReadClient.
+    pub fn new(client: &'a WalrusNodeClient<T>, config: ByteRangeReadClientConfig) -> Self {
+        Self { client, config }
+    }
+}
+
+impl<T: ReadClient> ByteRangeReadClient<'_, T> {
+    /// Reads a specific byte range from a blob.
```
liquid-helium, line 71: nit: Document the expected layout of the underlying encoding, i.e., now the first unencoded_size bytes of the concatenation of all the primary systematic slivers in PrimaryIndex order is the original file.

Maybe also consider adding the encoding type as an input.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,475 @@
...
+        self.client.check_blob_id(blob_id)?;
+
+        // Get blob status and certified epoch
+        let (certified_epoch, _) = self
+            .client
+            .get_blob_status_and_certified_epoch(blob_id, None)
+            .await?;
+
+        // Retrieve metadata
+        let metadata = self
```
liquid-helium, line 103: Do we need retry logic, if it fails to retrieve the slivers?

liquid-helium (approved):
Thanks very much @halfprice for adding this, since @mlegner has already reviewed this, I mainly checked the byte_range_read_client file, just two minor comments.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Added some documentation to explain the implementation detail of how to extract requested bytes from slivers.

Regarding encoding type, that is determined by the blob itself, which we will get from blob metadata?

liquid-helium: >Regarding encoding type, that is determined by the blob itself, which we will get from blob metadata?

Yes, you are right, we can get the encoding type from the blobMetadata.
What I was thinking is to check the encoding type is expected, e.g., if encoding == RS2, then this is how to read the ranges of the original file.

crates/walrus-sdk/src/client/byte_range_read_client.rs
```diff
@@ -0,0 +1,479 @@
...
+            byte_length,
+            "start reading byte range"
+        );
+
+        let Some(byte_length) = NonZeroUsize::new(byte_length) else {
+            return Ok(vec![]);
+        };
+
+        // Validate blob ID
+        self.client.check_blob_id(blob_id)?;
```
sadhansood: I think `check_blob_id` should be done before we do byte length check. For a byte length of 0, we will return Ok(vec![]) but if it is for a blob in blocklist, it should ideally see an error with correct error code. This also keeps the behavior consistent with full blob read.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Good question! As a client library, ideally we can retry some transient error here such as networking error. On the other hand, this can also be addressed in the application level. Leaning towards to skip it for now to avoid adding retry config to the client. Can revisit in the future when there is a need.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: Good call! I overlooked what `check_blob_id` actually did. Your suggestion is important. Done.

crates/walrus-sdk/src/client/byte_range_read_client.rs
halfprice: That's a very good suggestion! Added a check that returns error if the blob is not encoded using RS2.

PR #2710, opened 2025-11-14
feat(client): increase CLI start-up performance

mlegner:
## Description

This reduces the latency for storing blobs by
- building the `WalrusNodeClient` in a background task and
- adding static network parameters `n_shards` and `max_epochs_ahead` to
  the client config.

This allows us to start the encoding immediately, even without reading
the staking object.

Blocking tasks like encoding blobs and reading native certificates are
now performed in a blocking thread with `tokio::task::block_in_place`.

Also add some tracing instrumentation to improve debugging with Jaeger.

Open tasks:
- [x] Rebase this after #2703 is merged.
- [x] Maybe some code cleanup?
- [x] Enable the optimizations also for the `store-quilt` command.

Closes WAL-969
Contributes to WAL-968

## Test plan

Run `TRACE_FILTER=walrus=debug walrus store $BLOB --epochs 1 --ignore-resources --trace-cli otlp` with blobs of different sizes and check the duration of the `encode_blobs` function and the start of the `get_blob_statuses` function on Jaeger (see https://github.com/MystenLabs/walrus/blob/main/CONTRIBUTING.md#tracing). In all cases, encoding started immediately. For larger blobs, there is no longer any delay due to the client initialization.

| Blob size | Encoding duration | Start getting status |
|----------:|------------------:|---------------------:|
|       1MB |             0.48s |                1.64s |
|     100MB |             1.06s |                1.59s |
|     500MB |             3.88s |                3.94s |

---

## Release notes

- [x] CLI: Add optional static parameters `n_shards` and `max_epochs_ahead` to the `client_config.yaml` to reduce CLI start-up time. You can obtain the latest version of the config file at http://docs.wal.app/setup/client_config.yaml.

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -2173,24 +2173,25 @@ impl<T> WalrusNodeClient<T> {
 // INV: either client is initialized xor the join handle is Some.
 // INV: if the encoding_config is not initialized, the join handle is Some.
 #[derive(Debug)]
-pub struct WalrusNodeClientCreatedInBackground<T, E> {
+pub struct WalrusNodeClientCreatedInBackground<T> {
     encoding_config: OnceLock<EncodingConfig>,
-    client: OnceLock<WalrusNodeClient<T>>,
+    /// Contains the result of the background task if was already completed. If it returned an
+    /// error, the error is stored as a string.
+    client: OnceLock<Result<WalrusNodeClient<T>, String>>,
     #[allow(clippy::type_complexity)]
-    join_handle: Mutex<Option<JoinHandle<Result<WalrusNodeClient<T>, E>>>>,
+    join_handle: Mutex<Option<JoinHandle<ClientResult<WalrusNodeClient<T>>>>>,
     start_time: Instant,
 }
 
-impl<T, E> WalrusNodeClientCreatedInBackground<T, E>
+impl<T> WalrusNodeClientCreatedInBackground<T>
 where
     T: Debug + Send + 'static,
-    E: Send + 'static,
 {
     /// Creates a new [`WalrusNodeClientCreatedInBackground`] that will create the
     /// [`WalrusNodeClient`] in a background task using the provided future.
     pub fn new<F>(create_future: F, maybe_n_shards: Option<NonZeroU16>) -> Self
     where
-        F: Future<Output = Result<WalrusNodeClient<T>, E>> + Send + 'static,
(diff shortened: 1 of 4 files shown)


crates/walrus-sui/src/client/read_client.rs
```diff
@@ -323,20 +326,24 @@ const EVENT_CHANNEL_CAPACITY: usize = 1024;
 
 impl SuiReadClient {
     /// Constructor for `SuiReadClient`.
+    #[tracing::instrument(name = "SuiReadClient::new", skip_all, level = Level::DEBUG)]
     pub async fn new(
```
mlegner, line 330: Remark: This function is changed significantly in a follow-up PR: #2717

sadhansood: got it, thanks - looking at it now

crates/walrus-sdk/src/client.rs
```diff
@@ -2399,6 +2160,410 @@ impl<T> WalrusNodeClient<T> {
...
+    pub async fn client(&self) -> Result<&WalrusNodeClient<T>, E> {
+        if let Some(client) = self.client.get() {
+            return Ok(client);
+        }
+        let start = Instant::now();
+        let result = if let Some(join_handle) = self.join_handle.lock().await.take() {
+            let client = join_handle
+                .await
+                .expect("creating the WalrusNodeClient must not panic")?;
+            Ok(self.client.get_or_init(|| client))
```
sadhansood: If the background task returns with an error, we probably want to re-spawn the background task or cache the error. But now, it looks like we will once the join_handle is consumed, the client stays unset when background task fails to create the client. Future calls go straight to `cell must have been initialized if the join handle is no longer set`. I think we should return an error instead of panic.

mlegner: Ah, that's a good catch! AFAICT, in all current cases, we propagate the error if the `client()` call fails, so we probably don't actually reach the panic. But it's still good to prevent the panic in all cases.

It turns out that both storing the error and re-spawning the task is a bit tedious as `ClientError` does not implement `Clone` and we also cannot just clone the future. I'll try to find a way to make it work though.

mlegner: I've changed this behavior in 36dd3251155a808a7f7429e779879b35d256c039. It's not super clean but still nicer than risking panics in additional calls to `client` or `into_client` if the background task fails. Let me know if you see issues with this. 🙂

sadhansood: Yes, looks good!

PR #2687, opened 2025-11-04
feat(core): improve memory management in blob encoding

mlegner:
## Description

This PR improves the memory management in the blob encoding in multiple ways as described below.

The previous implementation is kept for tests and comparison, but is marked as deprecated and can be removed in the future.

A downside of these changes is a reduction in code reuse, as we now have more specialized functions.

#### changes
- Instead of performing the full encoding and storing all encoded data (in addition to the unencoded blob data), we only keep the unencoded blob data and non-systematic secondary slivers.
- Allocate buffers once and reuse them whenever possible.

#### impact
- Memory usage is reduced from ~5.5x to ~1.5x the unencoded blob data. A measurement of running `walrus blob-id` with a 1 GB blob shows a reduction of peak memory usage from ~5.5 GiB to ~1.6 GiB.
- Performance is increased by ~29% for 32 MiB blobs and ~14% for 1 GiB blobs.
- This is useful for the `walrus blob-id` command, when storing blobs with the upload relay, but also when reading blobs and using the strict consistency check.

<details><summary>Benchmark details</summary>

```console
$ cargo bench -p walrus-core blob_encoding/compute_metadata -- --baseline main
    Finished `bench` profile [optimized] target(s) in 0.81s
     Running benches/basic_encoding.rs (target/release/deps/basic_encoding-2961e301b26c80b5)
     Running benches/blob_encoding.rs (target/release/deps/blob_encoding-990712f4dd7efcde)
Benchmarking blob_encoding/compute_metadata/1B: Collecting 10 samples in estimated 9.2561 s
blob_encoding/compute_metadata/1B
                        time:   [444.86 ms 450.40 ms 456.23 ms]
                        thrpt:  [2.1919   B/s 2.2202   B/s 2.2479   B/s]
                 change:
                        time:   [-8.3665% -7.1340% -5.7694%] (p = 0.00 < 0.05)
                        thrpt:  [+6.1227% +7.6820% +9.1304%]
                        Performance has improved.
Benchmarking blob_encoding/compute_metadata/1KiB: Collecting 10 samples in estimated 9.0555
blob_encoding/compute_metadata/1KiB
                        time:   [445.75 ms 452.83 ms 462.12 ms]
                        thrpt:  [2.1640 KiB/s 2.2083 KiB/s 2.2434 KiB/s]
                 change:
                        time:   [-8.2556% -6.7615% -4.7785%] (p = 0.00 < 0.05)
                        thrpt:  [+5.0183% +7.2518% +8.9984%]
[...]

PR #2713, opened 2025-11-16
[WIP] Add SymbolStore / Arena allocator

poelzi:
## Description

Currently the memory usage of walrus is beyond acceptable. Encoding a 1.4 GB file requires 16GB of memory.

This patch introduces a mmap file backed store which is used when the file exceeds a threshold.

The current design is only half the way, the current design would result in to many temp files. I'm in the process of adding an arena allocator, but I'm very unsure which API directions to go. The current design requires quite some refactoring.The minimal surface would be the add an allocator to encode_pairs_and_metadata and pass it down.

But I tend to create a EncodingSession struct, that bundles the allocator, encoding config etc. This would also make instrumentation, metrics and parallelization easier.

Before I spend more time on this I rather ask.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -25,9 +25,9 @@ dependencies = [
 
 [[package]]
 name = "addr2line"
-version = "0.24.2"
+version = "0.25.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dfbe277e56a376000877090da837660b4427aad530e3028d44e0bffe4f89a1c1"
+checksum = "1b5d307320b3181d6d7954e663bd7c774a838b8220fe0593c86d9fb09f498b4b"
 dependencies = [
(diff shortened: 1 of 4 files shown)


mlegner:
Thanks a lot for your contribution, @poelzi. As mentioned by @halfprice, we've been working on several improvements to the memory usage. In particular, https://github.com/MystenLabs/walrus/pull/2687, which will be included in the 1.38 release will reduce peak memory usage when storing blobs by almost 50%.

We've also been looking into writing temporary data to disk if the blob is too large to be handled in memory. This is ongoing work; so far, we've done initial benchmarks. We'll be able to provide an update this week. cc @jpcsmith

PR #2716, opened 2025-11-18
ci: update GH action to set up Walrus and site-builder

nikos-kitmeridis:
* Supports both v1 and v2 versions of the site-builder binary

## Description

Modifies the set-up-walrus action in order to support the selection of site-builder version. 
Defaults to v2. 
If v1 or v2 input is passed it parses the walrus-sites releases to fetch the latest corresponding.
Eg: v1 => mainnet-1.3.1, v2 => mainnet-2.0.0

## Test plan
~~(Pending, hence the Draft PR)~~
Deployment of the example in the new walrus-sites actions [repo](https://github.com/MystenLabs/walrus-sites-github-actions/tree/main/example)

Update: this is tested in mainnet and ready for review (and merged if there's anything else to get fixed) 
cc. @mlegner - per our discussion earlier.

### DIFF
--- a/.github/actions/set-up-walrus/action.yaml
+++ b/.github/actions/set-up-walrus/action.yaml
@@ -124,7 +124,8 @@ runs:
       run: |
         echo "Installing site-builder ${VERSION_TAG} for ${INPUT_SUI_NETWORK}..."
 
-        DOWNLOAD_URL="https://github.com/MystenLabs/walrus-sites/releases/download/${INPUT_SUI_NETWORK}-${VERSION_TAG}/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64.tgz"
+        BASE_URL="https://github.com/MystenLabs/walrus-sites/releases/download"
+        DOWNLOAD_URL="$BASE_URL/${INPUT_SUI_NETWORK}-${VERSION_TAG}/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64.tgz"
         echo "Downloading from: $DOWNLOAD_URL"
 
         curl -L "$DOWNLOAD_URL" -o /tmp/site-builder.tgz
@@ -135,8 +136,6 @@ runs:
         # Find and move the extracted binary
         if [[ -f /tmp/site-builder-extract/site-builder ]]; then
           mv /tmp/site-builder-extract/site-builder bin/site-builder
-        elif [[ -f "/tmp/site-builder-extract/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64" ]]; then
-          mv "/tmp/site-builder-extract/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64" bin/site-builder
         else
           echo "❌ ERROR: Could not find site-builder binary in archive"
           echo "Contents of extraction directory:"


.github/actions/set-up-walrus/action.yaml
```diff
@@ -14,6 +14,14 @@ inputs:
...
     description: The Sui network to use
     default: mainnet
+  SITE_BUILDER_VERSION:
+    description: |
+      The major version of site-builder to use. Choose between 'v1' or 'v2'.
+      - v1: Automatically fetches the latest v1.x release from GitHub
+      - v2: Automatically fetches the latest v2.x release from GitHub
+      If not provided, uses the latest version from Google Cloud Storage.
+    required: false
+    default: ""
```
nikos-terzo, line 24:
```suggestion
    default: "v2"
```

.github/actions/set-up-walrus/action.yaml
```diff
@@ -14,6 +14,14 @@ inputs:
   SUI_NETWORK:
     description: The Sui network to use
     default: mainnet
+  SITE_BUILDER_VERSION:
+    description: |
+      The major version of site-builder to use. Choose between 'v1' or 'v2'.
+      - v1: Automatically fetches the latest v1.x release from GitHub
+      - v2: Automatically fetches the latest v2.x release from GitHub
+      If not provided, uses the latest version from Google Cloud Storage.
+    required: false
```
nikos-terzo, line 23: We set `required` to false, but in the first step we validate against 'v1' and 'v2'.
Either mark it as `required: true` or set a `default` as suggestion.

.github/actions/set-up-walrus/action.yaml
```diff
@@ -60,8 +68,70 @@ runs:
...
+
+          # Fetch all releases from GitHub API
+          # Use -f to fail on HTTP errors and save to a file to avoid control character issues
+          if ! curl -sf https://api.github.com/repos/MystenLabs/walrus-sites/releases -o /tmp/releases.json; then
+            echo "❌ ERROR: Failed to fetch releases from GitHub API"
+            exit 1
+          fi
+
+          # Find the latest release tag matching the network and major version
+          VERSION_TAG=$(jq -r --arg network "${{ inputs.SUI_NETWORK }}" --arg version "$VERSION" \
```
nikos-terzo: It is recommended to first sanitise the inputs by using the `env:` section.

.github/actions/set-up-walrus/action.yaml
```diff
@@ -60,8 +68,70 @@ runs:
...
+            echo "❌ ERROR: Could not find any ${VERSION}.x release for network ${{ inputs.SUI_NETWORK }}"
+            echo "Available releases:"
+            jq -r '.[].tag_name' /tmp/releases.json | grep "^${{ inputs.SUI_NETWORK }}-" | head -10 || echo "No releases found"
+            exit 1
+          fi
+
+          echo "Installing site-builder ${VERSION_TAG} for ${{ inputs.SUI_NETWORK }}..."
+
+          # Construct the download URL for GitHub release
+          DOWNLOAD_URL="https://github.com/MystenLabs/walrus-sites/releases/download/${{ inputs.SUI_NETWORK }}-${VERSION_TAG}/site-builder-${{ inputs.SUI_NETWORK }}-${VERSION_TAG}-ubuntu-x86_64.tgz"
```
nikos-terzo: We do not have a _site-builder-**testnet**-v..._ release. I am wondering whether it makes sense to remove the network prefix all-together, or if we are going to add this in the future.
For now it mostly makes sense to just hardcode `mainnet`

.github/actions/set-up-walrus/action.yaml
nikos-terzo: Actually, I just read "If not provided, uses the latest version from Google Cloud Storage."
If this should hold true, ignore these suggestions. The issue instead is in the validate version step.

nikos-kitmeridis: Right Nick, if it is not provided it falls back to the previous logic, which is hosted in gcs

```
        else
          # Use latest version from Google Cloud Storage
          echo "Installing latest site-builder for ${{ inputs.SUI_NETWORK }}..."
          curl https://storage.googleapis.com/mysten-walrus-binaries/site-builder-${{ inputs.SUI_NETWORK }}-latest-ubuntu-x86_64 -o bin/site-builder
          chmod +x bin/site-builder
        fi
```

.github/actions/set-up-walrus/action.yaml
nikos-kitmeridis: Yeah it makes senses, but it was hard coded once and then there was a commit making it env var. 
Not sure what's the reason behind this but I don't want to make unecessary changes in the context of this PR.
We can have a follow up ticket for this if you want.

.github/actions/set-up-walrus/action.yaml
nikos-kitmeridis: Thanks I think I have fixed this now, can you confirm that this is what you meant?

.github/actions/set-up-walrus/action.yaml
nikos-terzo: I didn't see that. I was confused by the multiple indents. Do you think there could be a simpler version of this?
Maybe a step of determining the version takes nothing or v1 or v2 and outputs "latest" or "mainnet-v1.x.x" or "mainnet-v2.x.x". And the next step using the output of the previous?

.github/actions/set-up-walrus/action.yaml
nikos-terzo: Almost. I think adding the `INPUT_` prefix is somewhat noisy, but ok, this is subjective I guess.
Another think that needs fixing is always putting quotes (`"`) when using these variables. I will add a comment on an example too.

.github/actions/set-up-walrus/action.yaml
```diff
@@ -24,44 +32,120 @@ runs:
...
+
+          # Extract the binary
+          mkdir -p /tmp/site-builder-extract
+          tar -xzf /tmp/site-builder.tgz -C /tmp/site-builder-extract
+
+          # Find and move the extracted binary
+          if [[ -f /tmp/site-builder-extract/site-builder ]]; then
+            mv /tmp/site-builder-extract/site-builder bin/site-builder
+          elif [[ -f /tmp/site-builder-extract/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64 ]]; then
+            mv /tmp/site-builder-extract/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64 bin/site-builder
```
nikos-terzo: Here input to `mv` command should be wrapped in quotes (`"`).

.github/actions/set-up-walrus/action.yaml
nikos-kitmeridis: I split this to multiple conditional steps, per your recommendation. 
Still a bit complex but it read a bit easier now I hope.

.github/actions/set-up-walrus/action.yaml
nikos-kitmeridis: No I prefer the `INPUT_` so we can distinct from `SECRET_` or `VAR_` if there's any.

Also quoted the mv commands where paths where using those unresolved variables, nice catch!

.github/workflows/publish-docs.yaml
mlegner, line 1: We have two other workflows that probably need this change:
- https://github.com/MystenLabs/walrus/blob/main/.github/workflows/update-ws-binaries.yaml
- https://github.com/MystenLabs/walrus/blob/main/.github/workflows/update-ws-install.yaml

.github/actions/set-up-walrus/action.yaml
mlegner: > It is recommended to first sanitise the inputs by using the `env:` section.

Can you explain what the benefits are of doing that?

.github/actions/set-up-walrus/action.yaml
```diff
@@ -14,6 +14,14 @@ inputs:
   SUI_NETWORK:
     description: The Sui network to use
     default: mainnet
+  SITE_BUILDER_VERSION:
```
mlegner, line 17: Suggestion (maybe for a follow-up): Allow also specifying a particular release like `testnet-v1.4.2` and the same for Walrus.

.github/actions/set-up-walrus/action.yaml
```diff
@@ -24,44 +32,135 @@ runs:
...
+
+    - name: Install site-builder from GitHub releases
+      if: steps.version.outputs.use_github == 'true'
+      env:
+        INPUT_SUI_NETWORK: ${{ inputs.SUI_NETWORK }}
+        VERSION_TAG: ${{ steps.version.outputs.version_tag }}
+      run: |
+        echo "Installing site-builder ${VERSION_TAG} for ${INPUT_SUI_NETWORK}..."
+
+        DOWNLOAD_URL="https://github.com/MystenLabs/walrus-sites/releases/download/${INPUT_SUI_NETWORK}-${VERSION_TAG}/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64.tgz"
```
mlegner: Nit: This line is too long. Can you split this, e.g., by first defining a `BASE_URL`?

```suggestion
        BASE_URL="https://github.com/MystenLabs/walrus-sites/releases/download"
        DOWNLOAD_URL="$BASE_URL/${INPUT_SUI_NETWORK}-${VERSION_TAG}/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64.tgz"
```

.github/actions/set-up-walrus/action.yaml
```diff
@@ -24,44 +32,135 @@ runs:
...
+        curl -L "$DOWNLOAD_URL" -o /tmp/site-builder.tgz
+
+        mkdir -p /tmp/site-builder-extract
+        tar -xzf /tmp/site-builder.tgz -C /tmp/site-builder-extract
+
+        # Find and move the extracted binary
+        if [[ -f /tmp/site-builder-extract/site-builder ]]; then
+          mv /tmp/site-builder-extract/site-builder bin/site-builder
+        elif [[ -f "/tmp/site-builder-extract/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64" ]]; then
+          mv "/tmp/site-builder-extract/site-builder-${INPUT_SUI_NETWORK}-${VERSION_TAG}-ubuntu-x86_64" bin/site-builder
```
mlegner: Question: Why do we need this case distinction?

.github/actions/set-up-walrus/action.yaml
nikos-kitmeridis: I believe it's more defensive because the values are being resolved as variable values and treated as pure data subsequently, while with the previous approach, by resolving directly in the bash command there's an opportunity for malicious command interpolation.

.github/workflows/publish-docs.yaml
nikos-kitmeridis: Right Markus! I have applied the same fix for these 2 jobs now, i.e. to use v1 of the site builder, so they are not breaking. We also have a ticket about fixing these jobs, which means we need to make them use v2.
We'll follow up with a new PR for that after testing that everything works fine for these jobs in v2.

.github/actions/set-up-walrus/action.yaml
nikos-kitmeridis: Thanks, created a follow up ticket for that: https://linear.app/mysten-labs/issue/SEW-493/in-the-deploy-github-action-allow-for-selecting-a-specific-site

.github/actions/set-up-walrus/action.yaml
nikos-kitmeridis: done!

.github/actions/set-up-walrus/action.yaml
nikos-kitmeridis: Actually there's no need, I just wanted to ensure that it will work no matter how the executable is named. But I checked and the name is plain "site-builder" so I removed the "else if" part. Thanks for pointing this out.

.github/actions/set-up-walrus/action.yaml
mlegner: I still don't understand the benefits of this (see also https://github.com/MystenLabs/walrus/pull/2716#issuecomment-3556277971) or who would inject malicious data, but I also don't see a downside besides additional boilerplate. 🤷

mlegner (approved):
Thanks a lot for the work and explanations, @nikos-kitmeridis. LGTM.

PR #2717, opened 2025-11-18
feat(sui): reduce, combine, and parallelize RPC requests

mlegner:
## Description

This significantly improves the performance of the Sui client and thus reduces the latency for client write operations through the following changes:
- Reduce and parallelize RPC requests when creating a new `SuiReadClient`.
- Store the initial shared versions of the system and staking objects and fixed system parameters when creating the `SuiReadClient`.
- Get the gas price and budget in parallel.
- Add a new `ArgumentOrOwnedObject::ObjectArg`. This is then used to get object args once and reuse them in concurrent transactions.
- When certifying/extending multiple blobs, get the object args concurrently.

This commit also contains some refactoring of the codebase:
- Remove the `async` keyword from many functions that no longer need it.
- Rename some functions to make it clear that they do not get any data from the network.
- Move some functions from the `SuiReadClient` to the `RetriableSuiClient`.
- Remove some no-longer-used functions.

Depends on #2710.

## Test plan

Test suite + manual performance tests using tracing and Jaeger (see [here](https://github.com/MystenLabs/walrus/blob/main/CONTRIBUTING.md#tracing)): 

Running `walrus store --epochs 1 $BLOB --ignore-resources` against Testnet using a release build and a 1MB blob takes ~7.4s compared to ~8.3s before (i.e., a **latency reduction of ~900ms**). 

Latency reduction for uploading through a publisher is less pronounced with a reduction of ~300ms (relatively high variance).

---

## Release notes

- [x] Publisher: Reduce number of Sui RPC requests and slightly decrease upload latency.
- [x] CLI: Reduce number of Sui RPC requests and decrease latency when storing blobs.

### DIFF
--- a/crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
+++ b/crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
@@ -232,6 +232,15 @@ impl LazyClientBuilder<SuiClient> for LazySuiClientBuilder {
     }
 }
 
+/// A struct that contains the gas price and gas budget.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct GasBudgetAndPrice {
+    /// The gas budget.
+    pub gas_budget: u64,
+    /// The gas price.
+    pub gas_price: u64,
+}
+
 /// A [`SuiClient`] that retries RPC calls with backoff in case of network errors.
 ///
 /// This retriable client wraps functions from the [`CoinReadApi`][sui_sdk::apis::CoinReadApi] and
@@ -685,14 +694,17 @@ impl RetriableSuiClient {
         object_ids: &[ObjectID],
         options: SuiObjectDataOptions,
     ) -> SuiClientResult<Vec<SuiObjectResponse>> {
-        let mut responses = Vec::with_capacity(object_ids.len());
-        for obj_id_batch in object_ids.chunks(MULTI_GET_OBJ_LIMIT) {
-            responses.extend(
-                self.multi_get_object_with_options(obj_id_batch, options.clone())
-                    .await?,
-            );
-        }
-        Ok(responses)
+        let results = futures::future::try_join_all(object_ids.chunks(MULTI_GET_OBJ_LIMIT).map(
+            move |obj_id_batch| {
+                let options = options.clone();
+                async move {
+                    self.multi_get_object_with_options(obj_id_batch, options)
+                        .await
+                }
+            },
+        ))
+        .await?;
+        Ok(results.into_iter().flatten().collect())
     }
 
     /// Returns a map consisting of the move package name and the normalized module.
@@ -1108,9 +1120,12 @@ impl RetriableSuiClient {
         gas_budget: Option<u64>,
         signer: SuiAddress,
         transaction_kind: TransactionKind,
-    ) -> SuiClientResult<(u64, u64)> {
+    ) -> SuiClientResult<GasBudgetAndPrice> {
         if let Some(gas_budget) = gas_budget {
-            Ok((gas_budget, self.get_reference_gas_price().await?))
+            Ok(GasBudgetAndPrice {
+                gas_budget,
+                gas_price: self.get_reference_gas_price().await?,
+            })
         } else {
             self.estimate_gas_budget(signer, transaction_kind).await
         }
@@ -1126,11 +1141,11 @@ impl RetriableSuiClient {
     /// - requests the gas price concurrently, and
     /// - scales the computation cost of the dry run accordingly.
     #[tracing::instrument(skip_all, level = Level::DEBUG)]
-    pub(crate) async fn estimate_gas_budget(
+    async fn estimate_gas_budget(
         &self,
         signer: SuiAddress,
         kind: TransactionKind,
-    ) -> SuiClientResult<(u64, u64)> {
+    ) -> SuiClientResult<GasBudgetAndPrice> {
         let dry_run_tx_data = TransactionData::new_with_gas_coins_allow_sponsor(
             kind,
             signer,
@@ -1145,18 +1160,17 @@ impl RetriableSuiClient {
             tokio::try_join!(self.get_reference_gas_price(), dry_run_future)?;
         let gas_cost_summary = dry_run_result.effects.gas_cost_summary();
 
-        let computation_cost = if gas_price == DUMMY_GAS_PRICE {
-            gas_cost_summary.computation_cost
-        } else {
-            (gas_cost_summary.computation_cost * gas_price).div_ceil(DUMMY_GAS_PRICE)
-        };
-
+        let computation_cost =
+            (gas_cost_summary.computation_cost * gas_price).div_ceil(DUMMY_GAS_PRICE);
         let safe_overhead = GAS_SAFE_OVERHEAD * gas_price;
         let net_gas_usage_non_negative = (computation_cost + gas_cost_summary.storage_cost)
             .saturating_sub(gas_cost_summary.storage_rebate);
         let gas_budget = computation_cost.max(net_gas_usage_non_negative) + safe_overhead;
 
-        Ok((gas_budget, gas_price))
+        Ok(GasBudgetAndPrice {
+            gas_budget,
+            gas_price,
+        })
     }
 
     /// Executes a transaction.
--- a/crates/walrus-sui/src/client/transaction_builder.rs
+++ b/crates/walrus-sui/src/client/transaction_builder.rs
@@ -51,6 +51,7 @@ use super::{
     UpgradeType,
 };
 use crate::{
+    client::retry_client::retriable_sui_client::GasBudgetAndPrice,
     contracts::{self, FunctionTag},
     types::{
         NetworkAddress,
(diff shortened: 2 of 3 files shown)


crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -1045,42 +1069,68 @@ impl RetriableSuiClient {
...
+        let dry_run_future = self.dry_run_transaction_block(dry_run_tx_data);
+
+        let (gas_price, dry_run_result) =
+            tokio::try_join!(self.get_reference_gas_price(), dry_run_future)?;
+        let gas_cost_summary = dry_run_result.effects.gas_cost_summary();
+        let computation_cost = if gas_price == DUMMY_GAS_PRICE {
+            gas_cost_summary.computation_cost
+        } else {
+            (gas_cost_summary.computation_cost * gas_price).div_ceil(DUMMY_GAS_PRICE)
+        };
```
mlegner: @karlwuest PTAL and let me know if this is correct.

crates/walrus-sui/src/client.rs
```diff
@@ -2724,33 +2718,49 @@ impl SuiContractClientInner {
             return Ok(vec![]);
         }
 
+        let object_args = futures::future::try_join_all(
+            certify_and_extend_parameters
+                .iter()
+                .map(|blob_params| self.read_client.object_arg_for_object(blob_params.blob.id)),
+        )
+        .await?;
```
sadhansood: Nit: Would a `multi_get_object_with_options` be better here?

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -1045,42 +1069,68 @@ impl RetriableSuiClient {
...
-                tracing::debug!(%error, "transaction dry run failed");
-            })?
-            .effects;
-        let gas_cost_summary = effects.gas_cost_summary();
+        let dry_run_future = self.dry_run_transaction_block(dry_run_tx_data);
+
+        let (gas_price, dry_run_result) =
+            tokio::try_join!(self.get_reference_gas_price(), dry_run_future)?;
+        let gas_cost_summary = dry_run_result.effects.gas_cost_summary();
+        let computation_cost = if gas_price == DUMMY_GAS_PRICE {
```
sadhansood: Do we not need to adjust `gas_cost_summary` as well? So `gas_cost_summary * (gas_price/DUMMY_GAS_PRICE)`. I could be wrong, however.

crates/walrus-sui/src/client.rs
mlegner: Ah, that's a good idea, will do. 🙏

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
mlegner: Ah, good catch. I don't think we need to scale all values, just the computation fee. But that means we should probably compute the `net_gas_usage` ourselves as `scaled_computation_cost + storage_cost - storage_rebate` (instead of calling the `net_gas_usage()` method). WDYT?

sadhansood: Yes, that would be certainly better than just scaling it!

mlegner: OK, I now have the following code:
```rust 
        let computation_cost = if gas_price == DUMMY_GAS_PRICE {
            gas_cost_summary.computation_cost
        } else {
            (gas_cost_summary.computation_cost * gas_price).div_ceil(DUMMY_GAS_PRICE)
        };

        let safe_overhead = GAS_SAFE_OVERHEAD * gas_price;
        let net_gas_usage_non_negative = (computation_cost + gas_cost_summary.storage_cost)
            .saturating_sub(gas_cost_summary.storage_rebate);
        let gas_budget = computation_cost.max(net_gas_usage_non_negative) + safe_overhead;
```

cc @karlwuest

crates/walrus-sui/src/client.rs
mlegner: Done in 9ea192e3da5329109aa9a5dbad716d96dd78f16f.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -1045,43 +1100,63 @@ impl RetriableSuiClient {
...
+    /// - requests the gas price concurrently, and
+    /// - scales the computation cost of the dry run accordingly.
     #[tracing::instrument(skip_all, level = Level::DEBUG)]
     pub(crate) async fn estimate_gas_budget(
         &self,
         signer: SuiAddress,
         kind: TransactionKind,
-        gas_price: u64,
-    ) -> SuiClientResult<u64> {
+    ) -> SuiClientResult<(u64, u64)> {
```
karlwuest: Nit: Maybe returning a new type here (and in `gas_budget_and_price`) would make sense, just to make it easier to use the function without having to look up which return value is the budget and which is the price.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -1045,43 +1100,63 @@ impl RetriableSuiClient {
...
+
+        let (gas_price, dry_run_result) =
+            tokio::try_join!(self.get_reference_gas_price(), dry_run_future)?;
+        let gas_cost_summary = dry_run_result.effects.gas_cost_summary();
+
+        let computation_cost = if gas_price == DUMMY_GAS_PRICE {
+            gas_cost_summary.computation_cost
+        } else {
+            (gas_cost_summary.computation_cost * gas_price).div_ceil(DUMMY_GAS_PRICE)
+        };
```
karlwuest: Nit: This could just be replaced with the `else` clause. It would be an unnecessary multiplication and division, but then the branching wouldn't be needed. Feel free to keep it if you prefer the readability of the current version, of course.

Praise: Nice idea to parallelize the two requests by using a dummy value :)

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
karlwuest: Yes, this looks correct.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
karlwuest: Yes, this is correct!

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -662,6 +676,25 @@ impl RetriableSuiClient {
...
+        object_ids: &[ObjectID],
+        options: SuiObjectDataOptions,
+    ) -> SuiClientResult<Vec<SuiObjectResponse>> {
+        let mut responses = Vec::with_capacity(object_ids.len());
+        for obj_id_batch in object_ids.chunks(MULTI_GET_OBJ_LIMIT) {
+            responses.extend(
+                self.multi_get_object_with_options(obj_id_batch, options.clone())
+                    .await?,
+            );
+        }
```
karlwuest: Nit/Remark: In principle these requests could also be sent in parallel, although I'm not sure if that might cause issues with the rate limiting of the full nodes (for a large number of objects).

crates/walrus-sui/src/client/transaction_builder.rs
```diff
@@ -79,6 +79,8 @@ pub const MAX_BURNS_PER_PTB: usize = 1000;
 pub enum ArgumentOrOwnedObject {
     /// An [`Argument`].
     Argument(Argument),
+    /// An [`ObjectArg`].
+    ObjectArg(ObjectArg),
```
karlwuest, line 84: Remark: This is a good idea! We could also consider removing the `Object` variant to make it less convenient to just call functions with `ObjectId::into()`, which for a PTB with multiple chained calls adds quite a few RPCs. I don't know how many callsites use this though, so this doesn't necessarily need to be in this PR.

karlwuest (approved):
Thanks, this looks great!

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
mlegner: Ah, that's a good idea. 🙏

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
mlegner: Done in 8b5edbfcb0af355b05d3b680f7c396bcc13ca439.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
mlegner: Done in 8b5edbfcb0af355b05d3b680f7c396bcc13ca439.

crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
mlegner: Agreed. I removed the branching in 8b5edbfcb0af355b05d3b680f7c396bcc13ca439.

PR #2721, opened 2025-11-19
feat: client recover sliver

halfprice:
## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:


crates/walrus-sdk/src/client.rs
```diff
@@ -834,6 +838,160 @@ impl<T: ReadClient> WalrusNodeClient<T> {
...
+    ) -> ClientResult<Vec<SliverData<A>>>
+    where
+        SliverData<A>: TryFrom<Sliver>,
+        RecoverySymbol<MerkleProof>: TryInto<RecoverySymbolData<A, MerkleProof>>,
+    {
+        // Call recover_sliver in parallel.
+        let futures = sliver_indices
+            .iter()
+            .map(|s| self.recover_sliver(metadata, *s, certified_epoch, timeout_duration));
+        let results = futures::future::join_all(futures).await;
```
jpcsmith, line 899: Remark: This is not in parallel, only concurrent.

crates/walrus-sdk/src/client.rs
```diff
@@ -834,6 +838,160 @@ impl<T: ReadClient> WalrusNodeClient<T> {
...
+        // Returns error is any of the recover_sliver calls returns an error.
+        if results.iter().any(|result| result.is_err()) {
```
jpcsmith, line 901: Remark: If the plan is to retry failures somewhere, you probably should not discard the successful slivers.

crates/walrus-sdk/src/client.rs
```diff
@@ -834,6 +838,160 @@ impl<T: ReadClient> WalrusNodeClient<T> {
...
+    async fn recover_sliver<A: EncodingAxis>(
+        &self,
+        metadata: &VerifiedBlobMetadataWithId,
+        sliver_index: SliverIndex,
+        certified_epoch: Epoch,
+        timeout_duration: Duration,
+    ) -> ClientResult<SliverData<A>>
+    where
+        SliverData<A>: TryFrom<Sliver>,
+        RecoverySymbol<MerkleProof>: TryInto<RecoverySymbolData<A, MerkleProof>>,
```
jpcsmith: Important: The client must not request symbols with proofs, as this will be a huge burden on the storage nodes.

crates/walrus-sdk/src/client.rs
```diff
@@ -834,6 +838,160 @@ impl<T: ReadClient> WalrusNodeClient<T> {
...
+                    .inspect({
+                        let value = progress_bar.clone();
+                        move |result| {
+                            if result.is_ok() {
+                                value.inc(1)
+                            }
+                        }
+                    })
+            })
+        });
```
jpcsmith, line 949: Important: The list recovery symbols endpoint allows for a single request to say "I am interested in recovering sliver-X, give me all the symbols from all the shards you have that can be used to recovery X". That approach should be used to avoid 1 request per symbol.

crates/walrus-sdk/src/client.rs
```diff
@@ -834,6 +838,160 @@ impl<T: ReadClient> WalrusNodeClient<T> {
...
+
+        let comms = self
+            .communication_factory
+            .node_read_communications(&committees, certified_epoch)?;
+        // Create requests to get all slivers from all nodes.
+        let verified_slivers_futures = comms.iter().flat_map(|n| {
+            // NOTE: the cloned here is needed because otherwise the compiler complains about the
+            // lifetimes of `s`.
+            n.node.shard_ids.iter().cloned().map(|s| {
+                n.retrieve_recovery_symbols::<A>(Arc::new(metadata.clone()), s)
```
jpcsmith: Remark: Arc the metadata before calling `recover_slivers`. Accepting it as an shared reference only to clone it is unnecessary. The caller can clone the metadata, move the metadata, or clone their Arc before passing it to the method.

crates/walrus-sdk/src/client.rs
```diff
@@ -834,6 +838,160 @@ impl<T: ReadClient> WalrusNodeClient<T> {
...
+        progress_bar.set_message("requesting recovery symbols");
+
+        let comms = self
+            .communication_factory
+            .node_read_communications(&committees, certified_epoch)?;
+        // Create requests to get all slivers from all nodes.
+        let verified_slivers_futures = comms.iter().flat_map(|n| {
+            // NOTE: the cloned here is needed because otherwise the compiler complains about the
+            // lifetimes of `s`.
+            n.node.shard_ids.iter().cloned().map(|s| {
```
jpcsmith: Hint: Since shard IDs are copy, you can use copied() instead of cloned() on the iterator.

crates/walrus-core/src/merkle.rs
```diff
@@ -120,6 +120,14 @@ where
             path: path.into(),
         }
     }
+
+    /// Creates a new empty proof.
+    pub fn new_empty_proof() -> Self {
```
halfprice, line 125: I tried and spent a long time trying to make GeneralRecoverySymbol<U> work with something like GeneralRecoverySymbol<EmptyProof>, but failed...

We can also make the new clientListRecoverySymbol to return a new type `GeneralRecoverySymbolWithoutProof`. But I don't know if using MerkleProof with empty fields is ok. Doing this here is mostly for prototyping convenience

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -2859,7 +2859,7 @@ async fn test_byte_range_read_client() -> TestResult {
 
     // Test reading byte range from the blob. We do 20 randomly generated byte range reads and
     // verify that the data is correct.
-    for _ in 0..20 {
+    for _ in 0..100 {
```
halfprice, line 2862: This test is how I test that the code is working. It is hard coded to always uses sliver recovery to read byte range currently.

crates/walrus-sdk/src/client.rs
```diff
@@ -709,27 +724,56 @@ impl<T: ReadClient> WalrusNodeClient<T> {
                 break;
             }
```
halfprice, line 726: Don't need to look at the change in this function yet. I haven't put too much thoughts into it. This is closer to the client and easier to change.

crates/walrus-sdk/src/client.rs
halfprice: I agree, but probably won't add across retry cache in the first PR.

crates/walrus-service/src/node.rs
```diff
@@ -3215,6 +3216,7 @@ impl StorageNodeInner {
         sliver_type: Option<SliverType>,
         encoding_type: EncodingType,
         worker: RecoverySymbolService,
+        client_request: bool,
```
halfprice, line 3219: This server side function is shared between the old ListRecoverySymbol endpoint and the new ClientListRecoverySymbol. It is the main logic to retrieve the recovery symbol. The client_request indicates that this is a client side request and should 
- Only recovery symbols targeting at source sliver
- No proof

This currently uses a single bad named variable for convenience, let me know if separating them into behavior flags (return_proof: booL, only_retrieve_source_symbol: bool) is cleaner.

crates/walrus-service/src/node.rs
```diff
@@ -3236,7 +3238,13 @@ impl StorageNodeInner {
 
         for target_sliver_type in [SliverType::Secondary, SliverType::Primary]
             .into_iter()
-            .filter(|target_type| sliver_type.is_none() || sliver_type == Some(*target_type))
+            .filter(|target_type| {
+                if client_request {
+                    *target_type == SliverType::Primary
```
halfprice, line 3243: Here I hard coded that if it is a client request, it can only try to recover primary sliver. May not be the best place to do this check, but this is where it is now.

crates/walrus-service/src/node/recovery_symbol_service.rs
```diff
@@ -187,17 +191,27 @@ impl RecoverySymbolService {
             target_pair_index.to_sliver_index::<T::OrthogonalAxis>(config.n_shards());
         let is_source_target = usize::from(target_sliver_index.get()) < sliver.symbols.len();
 
+        if client_request && !is_source_target {
+            return Err(RecoverySymbolError::ClientRequestIndexTooLarge(
+                target_sliver_index,
+            ));
+        }
```
halfprice, line 198: Here it requires that client_request must recover source target symbol. No no encoding should run.

crates/walrus-core/src/lib.rs
```diff
@@ -425,6 +425,9 @@ macro_rules! index_type {
 
 index_type!(
     /// Represents the index of a (primary or secondary) sliver.
+    /// It indicates the position of the sliver from the beginning of the blob.
+    /// Note that the first secondary sliver index is paired with the last primary sliver index,
+    /// and is in the last SliverPairIndex.
```
jpcsmith, line 430: Perhaps make this a regular comment, that pairing is not strictly required, and there may be a need to support other pairings in the future. putting this a docstring implies that we guarantee this in the api.

crates/walrus-core/src/merkle.rs
jpcsmith: Suggestion: The recovery symbol is specifically for use with proofs, return decoding symbols instead.

crates/walrus-service/src/node.rs
jpcsmith: Suggestion: Create a different function to handle these reads. Since you've structured it such that expansion should never be necessary and proofs are never returned, having them share the same code path means that they share the same bottleneck. Namely, if the service used to do the encode/create proofs is heavily utilised, then direct reads for symbols will be throttled. Since this is essentially just a subread for a sliver, it can be implemented as such: read the sliver and then take and return the symbol.

PR #2722, opened 2025-11-19
chore: provide for bidirectional shutdown coordination between background processors and the storage node

wbbradley:
### Error Flow (WAL-874 Resolved)

  1. Error occurs during event processing in background processor
  2. Processor triggers `node_cancel_token.cancel()`
  3. Processor breaks out of event loop
  4. Node's `select!` detects cancellation
  5. Node executes graceful shutdown sequence
  6. All processors are awaited and finish cleanly
  7. Node returns error to caller

### Normal Shutdown Flow (WAL-876 Resolved)

  1. External cancellation signal received
  2. Node's `run()` hits cancellation branch
  3. Calls `blob_event_processor.shutdown()`
  4. All background processors see cancellation and exit
  5. Node waits for all processors to finish
  6. Continues with rest of shutdown cleanup

## Test plan

CI Pipeline.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -824,7 +824,27 @@ impl StorageNode {
     }
 
     /// Run the walrus-node logic until cancelled using the provided cancellation token.
-    pub async fn run(&self, cancel_token: CancellationToken) -> anyhow::Result<()> {
+    pub async fn run_storage_node(&self, cancel_token: CancellationToken) -> anyhow::Result<()> {
+        let result = self.run_storage_node_core(cancel_token).await;
+
+        if let Err(ref error) = result {
+            tracing::error!(?error, "storage node shutting down due to error");
+        } else {
+            tracing::warn!("storage node shutting down");
+        }
+
+        if let Some(checkpoint_manager) = self.checkpoint_manager() {
+            checkpoint_manager.shutdown();
+        }
+        self.inner.shut_down();
+        self.blob_event_processor.shutdown();
+        self.blob_sync_handler.cancel_all().await?;
+        self.blob_event_processor.wait_for_shutdown().await;
+
+        result
+    }
+
+    async fn run_storage_node_core(&self, cancel_token: CancellationToken) -> anyhow::Result<()> {
         if let Err(error) = self
             .epoch_change_driver
             .schedule_relevant_calls_for_current_epoch()
@@ -847,11 +867,9 @@ impl StorageNode {
             },
             _ = cancel_token.cancelled() => {
                 tracing::info!("external cancellation received, shutting down node");
-                self.shut_down().await?;
             },
             _ = self.blob_processor_cancel_token.cancelled() => {
                 tracing::error!("blob event processor triggered shutdown due to error");
-                self.shut_down().await?;
                 return Err(anyhow::anyhow!("blob event processor encountered fatal error"));
             },
             blob_sync_result = BlobSyncHandler::spawn_task_monitor(
@@ -883,17 +901,6 @@ impl StorageNode {
             }
         }
 
-        Ok(())
-    }
-
-    async fn shut_down(&self) -> anyhow::Result<()> {
-        if let Some(checkpoint_manager) = self.checkpoint_manager() {
-            checkpoint_manager.shutdown();
-        }
-        self.inner.shut_down();
-        self.blob_event_processor.shutdown();
-        self.blob_sync_handler.cancel_all().await?;
-        self.blob_event_processor.wait_for_shutdown().await;
         Ok(())
     }
 
--- a/crates/walrus-service/src/node/blob_event_processor.rs
+++ b/crates/walrus-service/src/node/blob_event_processor.rs
@@ -190,7 +190,6 @@ impl BackgroundEventProcessor {
                                     worker_index = self.worker_index,
                                     "error processing blob event, triggering node shutdown"
                                 );
-                                self.node_cancel_token.cancel();
                                 break;
                             }
                         }
@@ -206,6 +205,9 @@ impl BackgroundEventProcessor {
                 }
             }
         }
+        // If any background event processor exits its loop, we should trigger cancellation to the
+        // entire node.
+        self.node_cancel_token.cancel();
     }
 
     /// Processes a blob event.
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -1409,7 +1409,9 @@ impl StorageNodeRuntime {
         let walrus_node_cancel_token = cancel_token.child_token();
         let walrus_node_handle = tokio::spawn(async move {
             let cancel_token = walrus_node_cancel_token.clone();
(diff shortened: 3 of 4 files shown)


sadhansood (dismissed):

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -132,39 +138,68 @@ impl BackgroundEventProcessor {
...
+                            ?error,
+                            worker_index = self.worker_index,
+                            "error processing blob event, triggering node shutdown"
+                        );
+                        self.node_cancel_token.cancel();
+                        break;
+                    }
+                }
+                else => {
+                    // Channel closed, exit gracefully
```
sadhansood: This else branch will not fire if `self.cancel_token.cancelled() => {` is still active. I think we should just match recv inside one single branch with Some/None

crates/walrus-service/src/node.rs
```diff
@@ -838,11 +846,13 @@ impl StorageNode {
                 }
             },
             _ = cancel_token.cancelled() => {
```
sadhansood, line 868: Should we also call `shut_down()` before returning Err() on process_events?

crates/walrus-service/src/node.rs
sadhansood, line 884: Also here

sadhansood, line 900: Not sure, but maybe here as well

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -132,39 +138,68 @@ impl BackgroundEventProcessor {
...
+                    tracing::info!(
+                        worker_index = self.worker_index,
+                        "background event processor channel closed, shutting down"
+                    );
+                    break;
```
sadhansood: I wonder if in None branch (i.e. event channel closed), we should also trip the `node_cancel_token` to trigger shutdown.

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -566,4 +615,57 @@ impl BlobEventProcessor {
...
+
+    /// Waits for all background event processors to finish after shutdown is triggered.
+    /// This method should be called after `shutdown()` to ensure all processors have exited.
+    pub async fn wait_for_shutdown(&self) {
+        match &self.blob_event_processor_impl {
+            BlobEventProcessorImpl::BackgroundProcessors {
+                background_processors,
+                ..
+            } => {
+                // Take the join handles out of the mutex
```
liquid-helium: nit: period at the end of comments.

crates/walrus-service/src/node.rs
```diff
@@ -531,6 +531,9 @@ pub struct StorageNode {
     start_epoch_change_finisher: StartEpochChangeFinisher,
     node_recovery_handler: NodeRecoveryHandler,
     blob_event_processor: BlobEventProcessor,
+    /// Cancellation token for blob event processors. When this is cancelled (either by external
+    /// shutdown or internal error), the node should shut down.
+    blob_processor_cancel_token: CancellationToken,
```
halfprice, line 536: suggestion: this cancellation token is currently used like a notify. This is used for BackgroundEventProcessor to notify the node to shutdown. Can we use Notify instead to be more explicit?

crates/walrus-service/src/node.rs
```diff
@@ -876,6 +886,17 @@ impl StorageNode {
...
     }
 
+    async fn shut_down(&self) -> anyhow::Result<()> {
+        if let Some(checkpoint_manager) = self.checkpoint_manager() {
+            checkpoint_manager.shutdown();
+        }
+        self.inner.shut_down();
+        self.blob_event_processor.shutdown();
+        self.blob_sync_handler.cancel_all().await?;
+        self.blob_event_processor.wait_for_shutdown().await;
```
halfprice: question: where signal for blob event processor to stop?

halfprice (commented):
Thanks @wbbradley to cleanup the leftovers! Took a quick look and had two questions for now. The use of multiple cancellation tokens needs refresh mind to look at them :)

crates/walrus-service/src/node/blob_event_processor.rs
wbbradley: Great catch @sadhansood. My understanding of `select!` was incorrect prior to this.

crates/walrus-service/src/node.rs
wbbradley: Yeah, I'll look at all these sites. Thank you.

crates/walrus-service/src/node.rs
wbbradley: It's in `StorageNode::shut_down()`.

crates/walrus-service/src/node/blob_event_processor.rs
wbbradley: Yeah, I think any exit from this loop will trigger it. Adjusting accordingly.

crates/walrus-service/src/node.rs
halfprice: Yeah, I think the correct code is correct.

I understand the BackgroundEvnetProcessor shutdown is triggered here. I was more thinking about BlobEventProcessor shutdown. But as you pointed out in the code doc, BlobEventProcessor::process_event() is in the same `select!` as the shut_down() is called, so BlobEventProcessor::process_event() will not be pulled anymore if shut_down is called. I was mostly not connecting the dots between StorageNode::shut_down() and BlobEventProcessor::process_event()

crates/walrus-service/src/node/blob_event_processor.rs
```diff
@@ -478,28 +525,34 @@ impl BlobEventProcessor {
...
+                    cancel_token.clone(),
+                    node_cancel_token.clone(),
                 );
-                // TODO(WAL-876): gracefully shut down the background processor when the node is
-                // shutting down.
-                background_processors.push(Arc::new(tokio::spawn(async move {
+                background_processors.push(tokio::spawn(async move {
                     background_processor.run_background_event_processor().await;
-                })));
+                }));
```
halfprice, line 535: remark: in addition to propagating error back and stop the node, another thing we want to catch is run_background_event_processor panics. Although this can cause the node to stop when the next event sent to the worker is not successful, but it'll be nice if we can catch this earlier.

crates/walrus-service/src/node/blob_event_processor.rs
wbbradley: Ok, I think I've addressed this in the latest commit.

crates/walrus-service/src/node.rs
wbbradley: Yep, done in latest commit by way of wrapper function.

crates/walrus-service/src/node.rs
wbbradley: Yep, done in latest commit by way of wrapper function.

crates/walrus-service/src/node/blob_event_processor.rs
wbbradley: Yep, makes sense. This now happens any time the loop breaks.

PR #2724, opened 2025-11-20
feat: aggregator supports byte range read

halfprice:
## Description

Add a new aggregator API that supports reading byte ranges from a blob.

If any sliver is missing, currently an error will be returned. Follow up PR will add single sliver recovery functionality in the client to address this issue.

## Test plan

---

## Release notes

- [ ] Storage node:
- [x] Aggregator: introduce a new aggregator API to read byte ranges of a blob
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -2854,7 +2854,7 @@ async fn test_byte_range_read_client() -> TestResult {
     for _ in 0..20 {
         // Generate a random byte range.
         let start_byte = random::<u64>() % blob_size;
-        let byte_length = random::<u64>() % (blob_size - start_byte);
+        let byte_length = random::<u64>() % (blob_size - start_byte) + 1;
         tracing::info!(
             "reading byte range {start_byte}-{end_byte} of blob {blob_id}",
             start_byte = start_byte,
--- a/crates/walrus-service/src/client/daemon/routes.rs
+++ b/crates/walrus-service/src/client/daemon/routes.rs
@@ -544,7 +544,7 @@ impl From<ClientError> for ByteRangeReadError {
             ClientErrorKind::BlobIdDoesNotExist => Self::BlobNotFound,
             ClientErrorKind::BlobIdBlocked(_) => Self::Blocked,
             ClientErrorKind::BlobTooLarge(max_blob_size) => Self::BlobTooLarge(*max_blob_size),
-            ClientErrorKind::ByteRangeReadError(msg) => Self::InvalidByteRange {
+            ClientErrorKind::ByteRangeReadInputError(msg) => Self::InvalidByteRange {
                 message: msg.to_string(),
             },
             _ => anyhow::anyhow!(error).into(),


crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -384,6 +397,149 @@ pub(super) async fn get_blob_by_object_id<T: WalrusReadClient>(
...
+
+    /// The blob size exceeds the maximum allowed size that was configured for this service.
+    #[error("the blob size exceeds the maximum allowed size: {0}")]
+    #[rest_api_error(reason = "BLOB_TOO_LARGE", status = ApiStatusCode::SizeExceeded)]
+    BlobTooLarge(u64),
+
+    #[error(transparent)]
+    #[rest_api_error(delegate)]
+    Internal(#[from] anyhow::Error),
+}
```
halfprice, line 539: I followed other APIs' error definition. Is the idea here that the status code returned will match the error got from the sdk client?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -384,6 +397,149 @@ pub(super) async fn get_blob_by_object_id<T: WalrusReadClient>(
...
+    params(
+        ("blob_id" = BlobId,),
+        ByteRangeQueryParams,
+    ),
+    responses(
+        (status = 200, description = "The byte range was read successfully", body = [u8]),
+        ByteRangeReadError,
+    ),
+)]
+pub(super) async fn get_blob_byte_range<T: WalrusReadClient>(
```
liquid-helium, line 599: Add `get_blob_byte_range` to `openapi.rs`, so that it shows up in the help doc

liquid-helium (approved):
Thanks @halfprice for adding this, LGTM.

wbbradley (changes requested):
HTTP already has a `Range` header. I think it might be better to add support for it to existing endpoints. 

1. Video streaming from the web automatically uses this as far as I can tell.
2. I think axum has built-in support for it which we may need to analyze.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -384,6 +397,149 @@ pub(super) async fn get_blob_by_object_id<T: WalrusReadClient>(
...
+#[tracing::instrument(
+    level = Level::ERROR,
+    skip_all,
+    fields(
+        %blob_id,
+        start=%params.start,
+        length=%params.length,
+    ),
+)]
+#[utoipa::path(
```
sadhansood, line 587: I would also document the consistency guarantees of this endpoint and mention that it is weaker than the full blob read (with strict consistency check and user should default to that if they want strict consistency check).

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -384,6 +397,149 @@ pub(super) async fn get_blob_by_object_id<T: WalrusReadClient>(
...
+    let etag = format!("{}-{}-{}", blob_id, params.start, params.length);
+
+    // Perform the byte range read using the trait method
+    match client
+        .read_byte_range(&blob_id, params.start, params.length)
+        .await
+    {
+        Ok(data) => {
+            tracing::debug!("successfully retrieved byte range of {} bytes", data.len());
+            let mut response = (StatusCode::OK, data).into_response();
```
sadhansood: Just checked mozilla's website, I think the response code range read is 206 https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -384,6 +397,149 @@ pub(super) async fn get_blob_by_object_id<T: WalrusReadClient>(
...
+    tracing::debug!(
+        "starting to read byte range from blob: start={}, length={}",
+        params.start,
+        params.length
+    );
+
+    // Validate parameters upfront
+    if params.length == 0 {
+        return ByteRangeReadError::InvalidByteRange {
+            message: "length must be greater than 0".to_string(),
```
sadhansood, line 615: Also mozilla's website says out of bound response should get 416 https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests

sadhansood (commented):
Thanks for doing this @halfprice, i think getting the response codes right would be really useful for the end clients. Please take a look.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: These have been removed.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: This endpoint is removed. The update in the new get_blob endpoint is applied in openapi doc.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Documented this in the updated get_blob endpoint

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Actually, I added this endpoint back, and add it in openapi.rs. Thanks for the tip!

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Use 206 in range request from the `get_blob` endpoint. Still use 200 here so that the result can be cached.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Thanks for the pointer! Use 416 in the new `get_blob` implementation with range request.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -210,33 +299,42 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
-        Err(error) => {
-            let error = GetBlobError::from(error);
 
-            match &error {
-                GetBlobError::BlobNotFound => {
-                    tracing::debug!(?blob_id, "the requested blob ID does not exist")
-                }
-                GetBlobError::Internal(error) => tracing::error!(?error, "error retrieving blob"),
-                _ => (),
+    let mut response = if let Some(range_header) = request_headers.get(RANGE) {
```
halfprice, line 310: @wbbradley @sadhansood here is the implementation for the Range support of the get blob api.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -155,6 +158,69 @@ pub struct ConcatRequestBody {
...
+/// Returns (start_byte_position, byte_length) if valid, otherwise returns an error.
+fn parse_range_header(range_value: &HeaderValue) -> Result<(usize, usize), GetBlobError> {
+    let range_value = range_value
+        .to_str()
+        .map_err(|e| GetBlobError::InvalidByteRange {
+            message: format!("invalid range header format: {}", e),
+        })?
+        .trim();
+
+    let parsed_range = http_range_header::parse_range_header(range_value).map_err(|e| {
```
halfprice, line 191: I used `http_range_header` crate to parse the Range header, but limit it to (1) only 1 range can be provided, (2) both start and end must be provided.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -187,6 +253,29 @@ async fn parse_and_resolve_id<T: WalrusReadClient>(
...
+///
+/// This endpoint also accepts the HTTP `Range` header to request only a portion of the blob.
+/// When a `Range` header is provided with a specific byte range (e.g., `bytes=100-199`), the
+/// endpoint returns HTTP 206 Partial Content with the requested byte range.
+///
+/// **Important**: When using byte range requests, only the consistency of the individual fetched
+/// slivers is validated. The entire blob's consistency is NOT validated, which provides better
+/// performance for partial reads but with reduced integrity guarantees compared to full blob reads.
+///
+/// # Range Header Format
```
wbbradley, line 272: Maybe we should avoid attempting to specify the Range Header Format unless we are not implementing it by the spec. It's unclear whether that is the case by explicitly giving it a partial specification here.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley, line 310: Might need to check that `axum` is not already handling some sort of slicing here. I'm not sure.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: This is the subset of the standard format. Calling this out here because I don't want to add the complication to support the full semantics yet. The following is not supported:
bytes=start-
bytes=-end

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -155,6 +158,69 @@ pub struct ConcatRequestBody {
...
+    /// The starting byte position (0-indexed, inclusive).
+    pub start: usize,
+    /// The number of bytes to read.
+    pub length: usize,
+}
+
+/// Parses an HTTP Range header value.
+/// Supports the format "bytes=start-end" for range to end.
+/// Returns (start_byte_position, byte_length) if valid, otherwise returns an error.
+fn parse_range_header(range_value: &HeaderValue) -> Result<(usize, usize), GetBlobError> {
```
wbbradley: we should probably support  open-ended ranges wherein the total size is not yet known.

```
Range: bytes=900-
```

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -187,6 +253,29 @@ async fn parse_and_resolve_id<T: WalrusReadClient>(
...
+///
+/// # HTTP Range Support
+///
+/// This endpoint also accepts the HTTP `Range` header to request only a portion of the blob.
+/// When a `Range` header is provided with a specific byte range (e.g., `bytes=100-199`), the
+/// endpoint returns HTTP 206 Partial Content with the requested byte range.
+///
+/// **Important**: When using byte range requests, only the consistency of the individual fetched
+/// slivers is validated. The entire blob's consistency is NOT validated, which provides better
+/// performance for partial reads but with reduced integrity guarantees compared to full blob reads.
```
wbbradley, line 270: Why are the integrity guarantees reduced, actually? Are the bytes not correct?

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: It looks like it doesn't, but should we consider using `axum-range` https://crates.io/crates/axum-range?

halfprice: My search results recommended [http-range-header](https://crates.io/crates/http-range-header) or [http-range](https://crates.io/crates/http-range)

Picked http-range-header given that the download is way higher than the other two.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Sure, but probably not in this PR though. The SDK does not support this yet, and I don't think we want to add changes to support this in this PR.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: You may potentially read an inconsistent blob's internal slivers, which appears to be consistent.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: I was playing with having a `<video>` tag in an html file and then investigating what happens in terms of the network requests. i recall seeing some unintuitive behavior. i don't recall the exact semantics. it might be worth trying this out with a large .mp4 file or something. IIRC the browser automatically did Range reads

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: fair. it might be worth calling out that difference, in the sense of saying that the entire spec is not (yet) implemented.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Which SDK? I'm thinking more about `<video>` tags, or other web-based scenarios where browsers may automatically use `Range` headers.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: I thought we had the hash of the sliver in the merkle tree?

wbbradley: Oh. I'm realizing we don't have the merkle tree in this context.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Sounds good!

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Underlying, this endpoint uses the crates/walrus-sdk to calculate which slivers to fetch. There, it must input valid (start position, length) pair, and the start+length cannot exceed end position.

Until the walrus-sdk supports optional length argument, to implement open end, here it needs to get the full blob length, which needs to make some calls to storage nodes to get metadata, etc.. All of these should be tackled separately. Ideally this should be supported by the walrus-sdk crate since there blob metadata is already fetched.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -210,33 +299,42 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+        let (start, length) = match parse_range_header(range_header) {
+            Ok(v) => v,
+            Err(error) => {
+                tracing::debug!(?error, "invalid byte range header");
+                return error.to_response();
             }
+        };
 
-            error.to_response()
+        match client.read_byte_range(&blob_id, start, length).await {
```
wbbradley: Another thing to consider here is whether it's worthwhile to attempt to stream the slivers as they are decoded. This might not be too difficult, depending on the internals. See `futures::stream::iter`.

halfprice: That would be nice. This still requires the SDK to support an iter to get data sliver by sliver. Created [WAL-1107](https://linear.app/mysten-labs/issue/WAL-1107/when-reading-data-from-aggregator-use-streaming-reads-instead-of)

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Added more comments. Also created [WAL-1108](https://linear.app/mysten-labs/issue/WAL-1108/aggregator-get-blob-range-request-support-open-end-range) to support open ended range request.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: That's a client side behavior? Or you are saying that the restriction on the Range header here may break currently working browser workflow?

Another thing I'm not sure is that, when requesting via cloudflare with Range header, will cloudflare respect it or remove it and fetch the whole thing from the backend. Different sources say different things and we may just have to try this out. But if the whole file is cached in cloudflare, Range header definitely work nicely (and now, serving Range queries does not interact with aggregator at all).

crates/walrus-service/src/client/daemon/routes.rs
sadhansood: Thanks Zhe. I have a general question - why do we need two different endpoint i.e get_blob and /byte_range. Isn't get_blob with a range header enough? If we want to keep both, we end up having different semantics i.e. 200/400 and 206/416. Should we just have get_blob?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -210,33 +301,43 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+        };
 
-            error.to_response()
+        // TODO(WAL-1107): support streaming the response sliver by sliver.
+        match client.read_byte_range(&blob_id, start, length).await {
+            Ok(blob) => (StatusCode::PARTIAL_CONTENT, blob).into_response(),
+            Err(error) => {
+                tracing::debug!(?error, "failed to read byte range");
+                return GetBlobError::from(error).to_response();
+            }
```
sadhansood, line 325: `client.read_byte_range` could still return an error if start_index+length are not in the blob range. Right now, it seems like an error from `read_byte_range` i.e. `InvalidByteRange` is getting mapped back to internal error (default branch in the GetBlobError::from enum mapping) when it should be 400/416.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -210,33 +301,43 @@ pub(super) async fn get_blob<T: WalrusReadClient>(
...
+            Err(error) => {
+                tracing::debug!(?error, "invalid byte range header");
+                return error.to_response();
             }
+        };
 
-            error.to_response()
+        // TODO(WAL-1107): support streaming the response sliver by sliver.
+        match client.read_byte_range(&blob_id, start, length).await {
+            Ok(blob) => (StatusCode::PARTIAL_CONTENT, blob).into_response(),
```
sadhansood: @halfprice - also check out `content-range` response header meant to show the relative position and total length of the blob. Claude tells me it is important to have this response header in range requests for browsers/clients to process it correctly. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Range

crates/walrus-service/src/client/daemon/routes.rs
sadhansood: Also, is it documented somewhere that CDNs cannot cache 206 responses (if so, it would be nice to mention or link it in the comments for future reference)

halfprice: >  why do we need two different endpoint i.e get_blob and /byte_range. Isn't get_blob with a range header enough?

The main problem is that Range requests with 206 response from the origin won't be cached. Moreover, I'm worried that cloudflare may remove the range header when cache miss, which fetch the entire blob from aggregator (this is from reading some of the community post as well). In the context of the checkpoint app, this won't work because aggregator currently cannot handle large blob read yet. That's why we specifically create a `/byte-range` api with 200/400 response. The intention is that this is not partial data, the data fetched is all the user cares (as opposite to using range request to run multiple partial fetch, but the goal is to read the entire file).

The official documentation from cloudflare is a bit fussy and the pointers are just [here](https://developers.cloudflare.com/workers/runtime-apis/cache/#invalid-parameters), where it mentions that 206 is an invalid status passed to the cache (206 response is not cached). The conclusion is supported by this [community post](https://community.cloudflare.com/t/range-requests-and-the-cache-api/263252) (this is kinda old though). Given this, I'm not sure what to add to the code. But I'll leave them here as future reference.

wbbradley: This makes sense. Thanks for the background.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Thanks for the pointer @sadhansood ! I added Content-Range header in the success response in the latest commit (along with Content-Length header which I read should be included as well). I'm not super familiar with these so hopefully I'm not creating tech debt here.

Regarding Content-Range in 416, this is a hard one because when the read_byte_range returns error, we may not know the full blob length yet (e.g., reading blob metadata returned error). Therefore, I haven't added Content-Range in 416 response and not sure exactly what to do there.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Good point. I separated ByteRangeError returned from `read_byte_range` into two: ByteRangeReadError which maps the internal error, and ByteRangeReadInputError, which will map to 416.

>Right now, it seems like an error from read_byte_range i.e. InvalidByteRange is getting mapped back to internal error

This is only the case in `/v1/blob/{blob_id}/byte-range/...` api, and this is intentional, since this API is not meant for reading partial data of a large request, as discussed [here](https://github.com/MystenLabs/walrus/pull/2724#discussion_r2557890960).

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -384,6 +509,158 @@ pub(super) async fn get_blob_by_object_id<T: WalrusReadClient>(
...
+    fn from(error: ClientError) -> Self {
+        match error.kind() {
+            ClientErrorKind::BlobIdDoesNotExist => Self::BlobNotFound,
+            ClientErrorKind::BlobIdBlocked(_) => Self::Blocked,
+            ClientErrorKind::BlobTooLarge(max_blob_size) => Self::BlobTooLarge(*max_blob_size),
+            ClientErrorKind::ByteRangeReadError(msg) => Self::InvalidByteRange {
+                message: msg.to_string(),
+            },
+            _ => anyhow::anyhow!(error).into(),
+        }
```
sadhansood, line 551: Shouldn't we also map `ByteRangeReadInputError` here? Otherwise in the byte range endpoint we end up returning 500 for invalid range.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -2850,8 +2853,8 @@ async fn test_byte_range_read_client() -> TestResult {
     // verify that the data is correct.
     for _ in 0..20 {
         // Generate a random byte range.
-        let start_byte = random::<usize>() % blob_size;
-        let byte_length = random::<usize>() % (blob_size - start_byte);
+        let start_byte = random::<u64>() % blob_size;
```
sadhansood, line 2856: If this lands at the last byte, i think we will end up with byte length of 0 and read_byte_range request will fail.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Gotcha. Makes sense.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: Good question. After adding ByteRangeReadInputError, we should replace ByteRangeReadError with ByteRangeReadInputError to return InvalidArgument. ByteRangeReadError should now map to 500 since that is about the internal execution error.

crates/walrus-e2e-tests/tests/test_client.rs
halfprice: Good catch! Fixed. This is overlooked since byte_range_read can take 0 as length before, but not anymore.

PR #2725, opened 2025-11-20
feat(node): move garbage collection to a background task

mlegner:
## Description

Moves the garbage collection added in #2599 and #2640 to a background task. 

This background task is (by default) started at a pseudorandom time during the first half of an epoch, which is deterministically set based on the node's public key and the epoch. This is to make sure that not all nodes perform garbage collection at the same time.

Garbage collection is done in batches within blocking threads to make sure the normal node operation is not affected and to make sure that we can still interrupt the whole process and shut the node down cleanly even if GC is running.

The task is restarted upon node restart if it is not finished.

## Test plan

- [x] Existing unit and E2E tests.
- [x] Existing (fixed) simtests.

### In a follow-up
- [ ] Add additional consistency checks for the DB that is performed in tests.
- [ ] Add simtest testing node restart during garbage collection.

### DIFF
--- a/crates/walrus-service/src/node/storage/constants.rs
+++ b/crates/walrus-service/src/node/storage/constants.rs
@@ -12,8 +12,8 @@ const EVENT_INDEX_COLUMN_FAMILY_NAME: &str = "latest_handled_event_index";
 const EVENT_CURSOR_COLUMN_FAMILY_NAME: &str = "event_cursor";
 const EVENT_CURSOR_KEY: [u8; 6] = *b"cursor";
 const GARBAGE_COLLECTOR_TABLE_COLUMN_FAMILY_NAME: &str = "garbage_collector_last_completed_epoch";
-const GARBAGE_COLLECTOR_LAST_STARTED_EPOCH_KEY: [u8; 9] = *b"started\0\0";
-const GARBAGE_COLLECTOR_LAST_COMPLETED_EPOCH_KEY: [u8; 9] = *b"completed";
+const GARBAGE_COLLECTOR_LAST_STARTED_EPOCH_KEY: &str = "started";
+const GARBAGE_COLLECTOR_LAST_COMPLETED_EPOCH_KEY: &str = "completed";
 
 // Base name for shard-related column families
 const SHARD_BASE_COLUMN_FAMILY_NAME: &str = "shard";
@@ -67,12 +67,12 @@ pub fn garbage_collector_table_cf_name() -> &'static str {
     GARBAGE_COLLECTOR_TABLE_COLUMN_FAMILY_NAME
 }
 
-pub fn garbage_collector_last_started_epoch_key() -> &'static [u8; 9] {
-    &GARBAGE_COLLECTOR_LAST_STARTED_EPOCH_KEY
+pub fn garbage_collector_last_started_epoch_key() -> String {
+    GARBAGE_COLLECTOR_LAST_STARTED_EPOCH_KEY.to_string()
 }
 
-pub fn garbage_collector_last_completed_epoch_key() -> &'static [u8; 9] {
-    &GARBAGE_COLLECTOR_LAST_COMPLETED_EPOCH_KEY
+pub fn garbage_collector_last_completed_epoch_key() -> String {
+    GARBAGE_COLLECTOR_LAST_COMPLETED_EPOCH_KEY.to_string()
 }
 
 /// Returns the column family name for primary slivers of a shard.
--- a/crates/walrus-service/src/node/storage.rs
+++ b/crates/walrus-service/src/node/storage.rs
@@ -184,7 +184,7 @@ pub struct Storage {
     metadata: DBMap<BlobId, BlobMetadata>,
     blob_info: BlobInfoTable,
     event_cursor: EventCursorTable,
-    garbage_collector_table: DBMap<[u8; 9], Epoch>,
+    garbage_collector_table: DBMap<String, Epoch>,
     shards: Arc<RwLock<HashMap<ShardIndex, Arc<ShardStorage>>>>,
     db_table_opts_factory: DatabaseTableOptionsFactory,
     metrics: Arc<CommonDatabaseMetrics>,
@@ -315,18 +315,20 @@ impl Storage {
             false,
         )?;
         if garbage_collector_table
-            .get(garbage_collector_last_started_epoch_key())?
+            .get(&garbage_collector_last_started_epoch_key())?
             .is_none()
         {
             garbage_collector_table
-                .insert(garbage_collector_last_started_epoch_key(), &GENESIS_EPOCH)?;
+                .insert(&garbage_collector_last_started_epoch_key(), &GENESIS_EPOCH)?;
         }
         if garbage_collector_table
-            .get(garbage_collector_last_completed_epoch_key())?
+            .get(&garbage_collector_last_completed_epoch_key())?
             .is_none()
         {
-            garbage_collector_table
-                .insert(garbage_collector_last_completed_epoch_key(), &GENESIS_EPOCH)?;
+            garbage_collector_table.insert(
+                &garbage_collector_last_completed_epoch_key(),
+                &GENESIS_EPOCH,
+            )?;
         }
 
         let metadata = DBMap::reopen(
@@ -398,7 +400,7 @@ impl Storage {
         &self,
     ) -> anyhow::Result<(Epoch, Epoch)> {
         let &[last_started_epoch, last_completed_epoch] =
-            &self.garbage_collector_table.multi_get([
+            &self.garbage_collector_table.multi_get(&[
                 garbage_collector_last_started_epoch_key(),
                 garbage_collector_last_completed_epoch_key(),
             ])?[..]
@@ -416,7 +418,7 @@ impl Storage {
     pub(crate) fn garbage_collector_last_started_epoch(&self) -> Result<Epoch, TypedStoreError> {
         Ok(self
             .garbage_collector_table
-            .get(garbage_collector_last_started_epoch_key())?
+            .get(&garbage_collector_last_started_epoch_key())?
             .unwrap_or(GENESIS_EPOCH))
     }
 
@@ -426,14 +428,14 @@ impl Storage {
         epoch: Epoch,
     ) -> Result<(), TypedStoreError> {
         self.garbage_collector_table
-            .insert(garbage_collector_last_started_epoch_key(), &epoch)
+            .insert(&garbage_collector_last_started_epoch_key(), &epoch)
     }
 
     /// Returns the highest epoch for which garbage collection was completed.
     pub(crate) fn garbage_collector_last_completed_epoch(&self) -> Result<Epoch, TypedStoreError> {
         Ok(self
             .garbage_collector_table
-            .get(garbage_collector_last_completed_epoch_key())?
+            .get(&garbage_collector_last_completed_epoch_key())?
             .unwrap_or(GENESIS_EPOCH))
     }
 
@@ -443,7 +445,7 @@ impl Storage {
         epoch: Epoch,
     ) -> Result<(), TypedStoreError> {
         self.garbage_collector_table
-            .insert(garbage_collector_last_completed_epoch_key(), &epoch)
+            .insert(&garbage_collector_last_completed_epoch_key(), &epoch)
     }
 
     pub(crate) fn clear_blob_info_table(&self) -> Result<(), TypedStoreError> {


crates/walrus-service/src/node.rs
```diff
@@ -1646,49 +1660,82 @@ impl StorageNode {
...
-        Ok(())
+        tracing::info!(
+            last_started_epoch,
+            last_completed_epoch,
+            current_epoch,
+            "restarting unfinished garbage-collection task on startup"
+        );
+        self.garbage_collector
+            .start_garbage_collection_task(current_epoch, epoch_start)
+            .await
```
halfprice: question: should we also check node status here (e.g. if it is in catch up mode)? If it's just `normal`, or `data sync`, this function can proceed.

crates/walrus-service/src/node/garbage_collector.rs
```diff
@@ -0,0 +1,244 @@
...
+    /// - The epoch is the genesis epoch.
+    /// - The garbage-collection task cannot be started.
+    #[tracing::instrument(skip_all)]
+    pub async fn start_garbage_collection_task(
+        &self,
+        epoch: Epoch,
+        epoch_start: DateTime<Utc>,
+    ) -> anyhow::Result<()> {
+        if epoch == GENESIS_EPOCH {
+            anyhow::bail!("cannot start garbage collection for genesis epoch");
```
halfprice: question: should we return Ok(()) here instead an error? By definition, genesis epoch has nothing to garbage collect

crates/walrus-service/src/node/garbage_collector.rs
```diff
@@ -0,0 +1,244 @@
...
+            .storage
+            .set_garbage_collector_last_started_epoch(epoch)?;
+
+        // Calculate target time and update metric before spawning the background task.
+        let target_time = self.cleanup_target_time(epoch, epoch_start);
+        self.metrics
+            .cleanup_task_start_time
+            .set(target_time.timestamp().try_into().unwrap_or_default());
+
+        let new_task = tokio::spawn(async move {
```
halfprice, line 169: question: what happens if the internal task fails in the middle? Currently do we just let it go through?

halfprice (commented):
Thanks @mlegner for this important PR! This is super critical!

I made a quick pass and left some minor comments. Overall, the code makes sense and I'll do a more indepth review when the remaining tasks are done.

crates/walrus-service/src/node.rs
```diff
@@ -1646,49 +1660,82 @@ impl StorageNode {
...
+            .garbage_collector
+            .start_garbage_collection_task(event.epoch, epoch_start)
+            .await
+        {
+            tracing::error!(
+                ?error,
+                epoch = event.epoch,
+                "failed to start garbage-collection task"
+            );
+        }
```
halfprice, line 1739: question: IIUC, this is the main change in this PR, in terms of functionality?

crates/walrus-service/src/node/garbage_collector.rs
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        let epoch_duration = self.node.system_parameters.epoch_duration;
+        let max_delay = epoch_duration / 2;
+        let public_key = self.node.protocol_key_pair.public();
+
+        // Create a deterministic seed from the public key bytes and epoch.
+        // Use a hash function to combine them into a u64 seed.
+        let public_key_bytes = bcs::to_bytes(public_key).unwrap_or_default();
+        let mut hasher = DefaultHasher::new();
```
halfprice: nit: not very important, but I think DefaultHasher is not stable. xxhash seems to be our go to hasher in various places.

crates/walrus-service/src/node/garbage_collector.rs
mlegner: Currently this should not be called in the Genesis epoch. But I still removed the error:
```rust
        if epoch == GENESIS_EPOCH {
            tracing::info!("garbage collection is not relevant in the genesis epoch");
            return Ok(());
        }
```

crates/walrus-service/src/node.rs
mlegner: I believe this is not needed here: If we started a GC task in epoch `E` while the node was active, and the latest epoch is still `E`, there is no way we went to recovery in the meantime.

However, we *should* actually check the node status when starting the GC task in the `process_epoch_change_start_event` function.

I will add a corresponding comment here and add the missing checks to `process_epoch_change_start_event`.

crates/walrus-service/src/node/garbage_collector.rs
mlegner: Yes, currently we don't really handle any errors there besides logging them. I believe this is appropriate because the GC task is not mission-critical. That is, if it fails, we can just retry after a restart or in the next epoch. Let me know if you disagree, then we can add more sophisticated error handling.

crates/walrus-service/src/node.rs
mlegner: Yes. Plus the restart and the random start time.

crates/walrus-service/src/node/garbage_collector.rs
mlegner: Ah, good point. I changed this accordingly.

crates/walrus-service/src/node.rs
mlegner: Done in 27dc099bbd62f9d706ed33dcbfcba5ec2f2b7751, but there is an open TODO regarding what should be the condition for the blob-info cleanup.

crates/walrus-service/src/common/utils.rs
```diff
@@ -697,6 +697,75 @@ pub(crate) fn unwrap_or_resume_unwind<T, E: std::convert::From<tokio::task::Join
...
+            break;
+        }
+
+        if batch_items.is_empty() {
+            // Skip this iteration if all items in the batch had errors.
+            continue;
+        }
+
+        let processed_count = spawn_blocking({
+            let batch_items = batch_items;
```
halfprice: Nit: I think this is redundant.

crates/walrus-service/src/common/utils.rs
```diff
@@ -697,6 +697,75 @@ pub(crate) fn unwrap_or_resume_unwind<T, E: std::convert::From<tokio::task::Join
...
+            .take(batch_size)
+            .inspect(|_| items_processed = true)
+            .filter_map(|result| {
+                result
+                    .inspect_err(|error| {
+                        tracing::warn!(?error, "{}", error_message);
+                    })
+                    .ok()
+            })
+            .collect();
```
halfprice: suggestion: should we move iterating values into the blocking thread as well? Reading from RocksDb is an blocking operation

crates/walrus-service/src/node.rs
```diff
@@ -1692,50 +1707,115 @@ impl StorageNode {
...
+    ///
+    /// This method is intended to be run at startup to check if a garbage-collection task was
+    /// started but not completed. If this is the case and we are still in the same epoch, it
+    /// restarts the garbage-collection task.
+    async fn check_and_start_garbage_collection_on_startup(&self) -> anyhow::Result<()> {
+        let last_started_epoch = self.inner.storage.garbage_collector_last_started_epoch()?;
+        let last_completed_epoch = self
+            .inner
+            .storage
+            .garbage_collector_last_completed_epoch()?;
```
halfprice: nit: we can use a multi_get to read these two values atomically.

crates/walrus-service/src/node.rs
```diff
@@ -1692,50 +1707,115 @@ impl StorageNode {
...
         }
 
-        if self.inner.garbage_collection_config.enable_data_deletion {
-            self.inner
-                .storage
-                .delete_expired_blob_data(epoch, &self.inner.metrics)
-                .await?;
+        let (current_epoch, state) = self.inner.contract_service.get_epoch_and_state().await?;
+
+        if current_epoch != last_started_epoch {
```
halfprice, line 1777: question: have you thought about the case where the node is processing event at the latest epoch, however, there was an event from previous epoch that is pending and not marked as persisted. Upon restart, last_started_epoch will be the latest epoch, but the node is re-processing events from previous epoch.

It feels that the [current event epoch](https://github.com/MystenLabs/walrus/blob/bfb135298448ac531ddc4c99408a5f493f51d4b1/crates/walrus-service/src/node.rs#L2737) should be used here instead of the current_epoch from chain. However, simply replacing `current_epoch` with `current_event_epoch().await` won't work since getting current event epoch requires seeing the first event, which is after this function completes.

Maybe just keep this as is. If there is any issue, we can do similar things as restarting node recovery, where we start the work in a thread and [wait for the current_event_epoch](https://github.com/MystenLabs/walrus/blob/3a13312b6815bd36b7f77528d4cdcb96ee8f8db8/crates/walrus-service/src/node/node_recovery.rs#L70) to become available.

crates/walrus-service/src/node/garbage_collector.rs
```diff
@@ -0,0 +1,271 @@
...
+    ///
+    /// The random delay is uniformly distributed between 0 and half the epoch duration. If random
+    /// delay is disabled, returns the epoch start time immediately.
+    fn cleanup_target_time(&self, epoch: Epoch, epoch_start: DateTime<Utc>) -> DateTime<Utc> {
+        if !self.config.enable_random_delay {
+            return epoch_start;
+        }
+
+        let epoch_duration = self.node.system_parameters.epoch_duration;
+        let max_delay = epoch_duration / 2;
```
halfprice: suggestion: 7 days (14 days / 2) is a long time. Can we make this as a config to easily change the start time if we need GC sooner?

crates/walrus-service/src/node/garbage_collector.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// Panics if blob info cleanup is not enabled.
+    async fn perform_db_cleanup_task(&self, epoch: Epoch) -> anyhow::Result<()> {
+        assert!(
+            self.config.enable_blob_info_cleanup,
+            "garbage-collection task must only be run if blob info cleanup is enabled"
+        );
+
+        // Disable DB compactions during cleanup to improve performance. DB compactions are
+        // automatically re-enabled when the guard is dropped.
+        let _guard = self.node.storage.temporarily_disable_auto_compactions()?;
```
halfprice, line 257: question: just want to double check that this option is not persisted in DB and reloaded upon restart? It looks like [we always use the default RocksDb option upon restart](https://github.com/MystenLabs/walrus/blob/1ecc90d80ab52e2da935ac2e62292d1c1b6ba84c/crates/walrus-service/src/node/storage/database_config.rs#L268). cc @sadhansood .

crates/walrus-service/src/test_utils.rs
```diff
@@ -1268,7 +1268,6 @@ impl Default for StorageNodeHandleBuilder {
     }
 }
 
-#[cfg(any(test, feature = "test-utils"))]
```
halfprice, line 1271: nit: is this intentional?

crates/walrus-service/src/node/storage/constants.rs
```diff
@@ -11,6 +11,9 @@ const METADATA_COLUMN_FAMILY_NAME: &str = "metadata";
 const EVENT_INDEX_COLUMN_FAMILY_NAME: &str = "latest_handled_event_index";
 const EVENT_CURSOR_COLUMN_FAMILY_NAME: &str = "event_cursor";
 const EVENT_CURSOR_KEY: [u8; 6] = *b"cursor";
+const GARBAGE_COLLECTOR_TABLE_COLUMN_FAMILY_NAME: &str = "garbage_collector_last_completed_epoch";
+const GARBAGE_COLLECTOR_LAST_STARTED_EPOCH_KEY: [u8; 9] = *b"started\0\0";
```
halfprice: question: why there is `\0\0` at the end?

crates/walrus-service/src/node.rs
```diff
@@ -1692,50 +1707,115 @@ impl StorageNode {
...
+        // available (e.g., in tests), use the current time as the epoch start.
+        let epoch_start = self
+            .inner
+            .contract_service
+            .get_epoch_and_state()
+            .await
+            .ok()
+            .and_then(|(current_epoch, state)| {
+                if current_epoch == epoch {
+                    state.start_of_current_epoch()
```
halfprice: question: the epoch start time will not be available in the common case since this is called right after EpochChangeStart event, is that correct?

crates/walrus-service/src/node.rs
```diff
@@ -1692,50 +1707,115 @@ impl StorageNode {
...
+        // in the same epoch, the node should still be active.
+        assert!(
+            self.inner.storage.node_status()?.is_active(),
+            "node status should not have changed since the last garbage-collection task was \
+            started if there was no epoch change"
+        );
+
+        let epoch_start = state
+            .start_of_current_epoch()
+            .expect("no garbage-collection task should have been started during epoch change");
```
halfprice: remark: I think this expect is not expected? GC can start during epoch change (right after EpochChangeStart event)? If the node restart before EpochChangeDone, this assertion may hit?

crates/walrus-service/src/node.rs
halfprice: Yeah, make sense. I think if this is invariant needs to be hold in garbage_collector. start_garbage_collection_task, maybe we can move the node status check from StorageNode::start_garbage_collection_task into it to prevent future garbage_collector. start_garbage_collection_task misuse?

halfprice (approved):
Thanks a lot @mlegner for implementing this important feature!

I have one relatively important comment (start with `remark`), and the rest are all pretty minor! As discussed, given this feature is flag guarded, I can approve it while you are tightening the tests.

Great job!

crates/walrus-service/src/common/utils.rs
mlegner: That's what I wanted to do initially, but unfortunately I did not find a good way to move the iterator into the thread. I'll try again, but let me know if you see a good way to do that.

crates/walrus-service/src/node.rs
mlegner: Luckily the blob info is updated immediately during event processing and in order. This is what we care most about, and I don't see any case where this could lead to incorrect behavior. 🤔

crates/walrus-service/src/node/garbage_collector.rs
mlegner: There is already an option to remove the randomness completely. But I can add another one to control the randomness period.

crates/walrus-service/src/node/garbage_collector.rs
mlegner: Good point. Do you know how RocksDB behaves there, @sadhansood?

To be sure, I can otherwise just add a call to enable compactions during startup.

crates/walrus-service/src/test_utils.rs
mlegner: Yes, because we have this attribute [on the whole module](https://github.com/MystenLabs/walrus/blob/65bd761db56ab8e099e04c1f653276727f5407fa/crates/walrus-service/src/lib.rs#L32).

crates/walrus-service/src/node/storage/constants.rs
mlegner: This is a filler to get to 9 bytes because AFAICT you cannot use a byte slice as key for typed store. I could of course use strings but wanted to keep this as simple as possible.

@sadhansood Do you have a better solution for this case?

crates/walrus-service/src/node.rs
mlegner: Indeed, I think I managed to confuse myself there. This also affects the following [comment](https://github.com/MystenLabs/walrus/pull/2725#discussion_r2561926713). I will fix these.

crates/walrus-service/src/node.rs
mlegner: Good catch. This demonstrates that we need a simtest that tests a restart during garbage collection. I added this to the PR description.

crates/walrus-service/src/node/garbage_collector.rs
mlegner: Added an optional parameter `randomization_time_window_secs`.

crates/walrus-service/src/node.rs
mlegner: Ah, just now this was actually detected in a simtest: https://github.com/MystenLabs/walrus/actions/runs/19706930375/job/56456869495?pr=2725

crates/walrus-service/src/node/garbage_collector.rs
mlegner: I added a call to the `Storage::open` function with a TODO to revisit this.

crates/walrus-service/src/common/utils.rs
mlegner: OK, I can just create the iterator within the blocking thread with a lower bound.

mlegner: Improved the batch processing in 9cd9913ce1be51ef521924401c9ba8bb1535dd67.

crates/walrus-service/src/node.rs
mlegner: Done in 6cc927526360ffbf58744a2690ddf1b1342c8dee.

crates/walrus-service/src/node.rs
mlegner: Fixed the `expect` in 6cc927526360ffbf58744a2690ddf1b1342c8dee.

crates/walrus-service/src/node.rs
mlegner: Moved it back into the `delete_expired_blob_data` in 02c7ba477218c5554338d187e3a2343d068fdae2.

crates/walrus-service/src/node/storage/constants.rs
halfprice: Oh, I see. I think this can just be a String type. Using 9 bytes is a bit restrictive since we can't create another key longer than 9 bytes.

mlegner: Done in 4afdf06126738fb1cac40ae69f5864ca35db27d7.

halfprice (approved):
LGTM! Thanks a lot for addressing the comments quickly @mlegner! Just want to make sure, you will continue adding more (sim)tests to test the feature, is that correct?

PR #2726, opened 2025-11-20
feat: reading individual sliver should respect aggregtor max blob size

halfprice:
## Description

The max_blob_size client config is only enforced when reading the entire blob. This PR makes it in effect when reading
individual slivers.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -2931,10 +2931,9 @@ async fn test_byte_range_read_client() -> TestResult {
 async fn test_byte_range_read_size_too_large() -> TestResult {
     walrus_test_utils::init_tracing();
 
-    let test_nodes_config: TestNodesConfig = TestNodesConfig {
-        node_weights: vec![7, 7, 7, 7, 7],
-        ..Default::default()
-    };
+    let test_nodes_config = TestNodesConfig::builder()
+        .with_node_weights(&[7, 7, 7, 7, 7])
+        .build();
     let test_cluster_builder =
         test_cluster::E2eTestSetupBuilder::new().with_test_nodes_config(test_nodes_config);
     let (_sui_cluster_handle, _cluster, mut client, _) = test_cluster_builder.build().await?;
@@ -2959,7 +2958,7 @@ async fn test_byte_range_read_size_too_large() -> TestResult {
 
     // Generate a random byte range. Pick a start position randomly choosing from 0 to 3000, so that
     // the read byte range can be in the middle of the blob.
-    let start_byte = random::<usize>() % 3000;
+    let start_byte = random::<u64>() % 3000;
     {
         // Note that when `byte_length` is smaller, but close the max read size, it may still
         // return BlobTooLarge, due to the config is cap the total data read from storage, rather
@@ -2982,7 +2981,7 @@ async fn test_byte_range_read_size_too_large() -> TestResult {
     }
 
     {
-        let byte_length = 500;
+        let byte_length = 300;
         tracing::info!(
             "reading byte range {start_byte}-{end_byte} of blob {blob_id}",
             start_byte = start_byte,
@@ -2994,8 +2993,10 @@ async fn test_byte_range_read_size_too_large() -> TestResult {
             .read_byte_range(&blob_id, start_byte, byte_length)
             .await?;
         assert_eq!(
-            byte_range_data,
-            blobs[0][start_byte..start_byte + byte_length].to_vec()
+            byte_range_data.data,
+            blobs[0][usize::try_from(start_byte).expect("start byte should be valid")
+                ..usize::try_from(start_byte + byte_length).expect("end byte should be valid")]
+                .to_vec()
         );
     }
     Ok(())


crates/walrus-sdk/src/client.rs
```diff
@@ -690,6 +689,23 @@ impl<T: ReadClient> WalrusNodeClient<T> {
...
+        let sliver_size = u64::from(
+            self.encoding_config()
+                .get_for_type(metadata.metadata().encoding_type())
+                .sliver_size_for_blob::<E>(metadata.metadata().unencoded_length())
+                .map_err(|e| ClientError::from(ClientErrorKind::Other(e.into())))?
+                .get(),
+        );
+        let read_total_sliver_size = sliver_size
+            .checked_mul(sliver_indices.len() as u64)
+            .ok_or(ClientError::from(ClientErrorKind::BlobTooLarge(u64::MAX)))?;
```
halfprice, line 705: The semantics here is that the config is to restrict the total data size read from the storage nodes, which can be higher than what the user requested to read. This means that it is possible for the request to be rejected if the request bytes is close to the limit due to that returned slivers may contain additional data than what the user wants to read. This can be weird from user experience perspective. So comments are welcome.

crates/walrus-sdk/src/client.rs
```diff
@@ -690,6 +689,23 @@ impl<T: ReadClient> WalrusNodeClient<T> {
     where
         SliverData<E>: TryFrom<walrus_core::Sliver>,
     {
+        if metadata.metadata().unencoded_length() == 0 {
```
liquid-helium, line 691: question: does this work with 0-byte files?

crates/walrus-sdk/src/client.rs
liquid-helium: Question: Is the goal to limit the in-flight data for read-range?
If so, shall we add it at the `bytes_read_range_client`, and limit only the in-flight data size. For large reads, as long as we can break it down into smaller reads, we proceed.

E.g., say we want to read 500K in total, each sliver size is 100K, and the size limit is set at 200K, in this case, we read 200K, 200K and 100K, so that the in-flight data size is always < 200K.

halfprice: The goal is to limit the overall data read from storage nodes to serve a single request.

The inflight data rate is already configured using the [communication config](https://github.com/MystenLabs/walrus/blob/0b3601504d45def5bdd7d76b35ba85c4138e5984/crates/walrus-sdk/src/config/communication_config.rs#L176), and works in the way that you described (reading data in batches to not exceed in flight data size). The idea here is to prevent client (mostly aggregators) from overload of serving large requests, by preventing reading large blobs. This is an aggregator specific setting, so each aggregator operator itself can decide what's the maximum data it can be served from a single request.

crates/walrus-sdk/src/client.rs
halfprice: Are you asking if we set the max blob size to 0, should 0-byte file request being rejected or return success? I think it should return success since `max_blob_size` is inclusive.

crates/walrus-sdk/src/client.rs
```diff
@@ -322,6 +322,11 @@ impl<T: ReadClient> WalrusNodeClient<T> {
             .with_client(sui_read_client))
     }
 
+    /// Sets the maximum blob size for the client for testing purposes.
```
mlegner, line 323: If this is just for testing, can we add an attribute `#[cfg(any(test, feature = "test-utils"))]`?

```suggestion
    /// Sets the maximum blob size for the client for testing purposes.
    #[cfg(any(test, feature = "test-utils"))]
```

crates/walrus-sdk/src/client.rs
mlegner, line 705: Remark: I don't think this can ever happen, we can never have a blob with encoded size above `u64::MAX`, so any subset of slivers can never be larger as well.

crates/walrus-sdk/src/client.rs
```diff
@@ -690,6 +689,23 @@ impl<T: ReadClient> WalrusNodeClient<T> {
...
+        // Get the sliver size for the given encoding type.
+        let sliver_size = u64::from(
+            self.encoding_config()
+                .get_for_type(metadata.metadata().encoding_type())
+                .sliver_size_for_blob::<E>(metadata.metadata().unencoded_length())
+                .map_err(|e| ClientError::from(ClientErrorKind::Other(e.into())))?
+                .get(),
+        );
+        let read_total_sliver_size = sliver_size
+            .checked_mul(sliver_indices.len() as u64)
```
mlegner: Nit: Can we use a `try_from` instead of the `as`? (I'm also surprised that Clippy doesn't complain about this... 🤔)

crates/walrus-sdk/src/client.rs
halfprice: Yes, I agree. Here I want to use safe operation, so I need to deal with the Error it returns. I don't think we want to crash the node, so instead, return an error here.

crates/walrus-sdk/src/client.rs
halfprice: Done.

crates/walrus-sdk/src/client.rs
halfprice: Done. Weird that clippy typically throw a warning on this. Not sure why it isn't in this case.

PR #2730, opened 2025-11-26
chore: remove use_legacy_event_provider option

wbbradley:
## Description

Usage of the legacy event provider is no longer supported. Let's remove it.

## Test plan

---

## Release notes

- [x] Storage node: Storage nodes will need to remove the `use_legacy_event_provider` flag from their configuration if it is present.
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -939,9 +939,10 @@ async fn test_storage_nodes_do_not_serve_data_for_deleted_blobs() -> TestResult 
 
     let count_deleted = client.delete_owned_blob(&blob_id).await?;
     assert_eq!(count_deleted, 1, "should have deleted one blob");
-    tokio::time::sleep(Duration::from_millis(50)).await;
 
-    check_that_blob_is_not_available(client, blob_id).await
+    // Wait for the deletion event to be processed by storage nodes.
+    // The checkpoint-based event processor needs time to process the deletion.
+    wait_for_blob_to_be_unavailable(client, blob_id, Duration::from_secs(30)).await
 }
 
 /// Tests that nodes correctly update the blob info for expired deletable blobs and no longer serve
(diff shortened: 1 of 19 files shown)


crates/walrus-service/src/event/event_processor/runtime.rs
```diff
@@ -89,37 +85,27 @@ impl EventProcessorRuntime {
...
-        let (event_manager, event_processor_handle): (Box<dyn EventManager>, _) =
-            if use_legacy_event_provider {
-                let read_client = runtime.block_on(async { sui_config.new_read_client().await })?;
-                (
-                    Box::new(SuiSystemEventProvider::new(
-                        read_client,
-                        sui_config.event_polling_interval,
-                    )),
-                    tokio::spawn(async { std::future::pending().await }),
+        let (event_manager, event_processor_handle): (Box<dyn EventManager>, _) = {
```
mlegner, line 88: Question: Do we still need to use a trait object here? As we now only support one  event provider, this should no longer be necessary. Or do you want to keep it to support the planned more efficient event processor?

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -1027,6 +1035,35 @@ async fn test_storage_nodes_do_not_serve_data_for_expired_deletable_blobs() -> T
     check_that_blob_is_not_available(client, blob_id).await
 }
 
+/// Polls until the blob status becomes `Nonexistent` and the blob cannot be read.
+async fn wait_for_blob_to_be_unavailable(
```
mlegner, line 1032: Nit/hint: We have a function `retry_until_success_or_timeout` in `crates/walrus-service/src/node.rs` that could be useful here (would probably have to be moved to a public module).

crates/walrus-service/src/test_utils.rs
```diff
@@ -201,35 +201,161 @@ pub trait StorageNodeHandleTrait {
...
-            disable_event_blob_writer: false,
+            disable_event_blob_writer: true, // Disabled by default to avoid test interference
             blocklist_dir: None,
             enable_node_config_synchronizer: false,
             node_recovery_config: None,
         }
     }
+
+    /// Sets the weights of the nodes in the cluster.
+    pub fn with_node_weights(mut self, weights: Vec<u16>) -> Self {
```
mlegner: Nit (absolutely optional): This could be changed to `&[u16]` to avoid the `vec!` at callsites.

crates/walrus-service/src/event/event_processor/runtime.rs
wbbradley: According to some comments in https://github.com/MystenLabs/walrus/pull/2346/files @sadhansood wanted to keep this abstraction around for use in the future.

crates/walrus-e2e-tests/tests/test_client.rs
wbbradley: Done.

crates/walrus-service/src/test_utils.rs
wbbradley: Good idea.

PR #2735, opened 2025-12-01
fix: address race condition and deadlock in dropping BlobRetirementNotify

halfprice:
## Description

The problem is manifested as when a storage node is recovery multiple shards during shard sync (not in recovery mode),
there maybe multiple tasks syncing different shards of the same blob. The dropping and acquiring a `blob retirement notify`
may cause deadlock (will explain in the code comment below).

This PR fixes the issue, clean up the `BlobRetirementNotifier`, and add more tests to tighten the invariant of BlobRetirementNotifier.

Note that the bug is not possible to manifest in simtest because the bug requires multi-thread interaction.

Fix WAL-1095

## Test plan

Additional tests added.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node/blob_retirement_notifier.rs
+++ b/crates/walrus-service/src/node/blob_retirement_notifier.rs
@@ -224,6 +224,16 @@ impl Drop for BlobRetirementNotify {
                 self.blob_id,
                 *ref_count
             );
+
+            if *ref_count == 0 {
+                // This should not happen since the ref count should be at least 1 when there is
+                // BlobRetirementNotify instance.
+                tracing::error!(
+                    %self.blob_id,
+                    "BlobRetirementNotifier drop NotifyWrapper: ref count is 0",
+                );
+            }
+
             *ref_count = ref_count.checked_sub(1).unwrap_or(0);
             *ref_count
         };


crates/walrus-service/src/node/blob_retirement_notifier.rs
```diff
@@ -185,30 +228,31 @@ impl Drop for BlobRetirementNotify {
...
         if new_ref_count == 1 {
             self.node_wide_blob_retirement_notifier
-                .registered_blobs
-                .lock()
-                .expect("mutex should not be poisoned")
-                .remove(&self.blob_id);
+                .cleanup_blob_retirement_notify(&self.blob_id);
         }
     }
 }
```
halfprice, line 248: Remark: essentially the sequence of events that triggers the deadlock is as follows:

Thread1: acquire_blob_retirement_notify
Thread1: drop notify, get new_ref_count == 1, pause
Thread2: acquire_blob_retirement_notify, update ref count to 2
Thread1: continue, since `if new_ref_count == 1` is true, it removes the blob from `registered_blobs`. This triggers `another DROP of a BlobRetirementNotify referenced by the copy inside the HashMap`. Note that now, registered_blobs lock is still being hold. This DROP starts executing the `drop()` again, trying to acquire registered_blobs lock, which will never succeed.

The fix is quite simple: before removing the blob from registered_blob, after acquiring the registered_blob, it needs to check if the reference remains to be 1. This is essentially what cleanup_blob_retirement_notify does.

crates/walrus-service/src/node/blob_retirement_notifier.rs
```diff
@@ -384,4 +422,222 @@ mod tests {
...
+
+    /// Test that BlobRetirementNotifier can handle a large number of concurrent acquire and drop
+    /// operations without deadlocking or panicking.
+    ///
+    /// This test uses a thread-based timeout mechanism to detect deadlocks. The test spawns
+    /// multiple concurrent tasks in a separate thread, and if they don't complete within the
+    /// timeout, the main test thread will panic and terminate the test, even if worker threads
+    /// are deadlocked on mutexes.
+    #[test]
+    fn test_concurrent_acquire_and_notify() {
```
halfprice, line 444: Note: this test can manifest the deadlock with very high probability, so retry the test won't mask the test failure.

crates/walrus-service/src/node/blob_retirement_notifier.rs
```diff
@@ -384,4 +422,222 @@ mod tests {
...
+    ///
+    /// This test randomly interleaves three operations:
+    /// - acquire_blob_retirement_notify (and drop)
+    /// - notify_blob_retirement
+    /// - Both operations combined
+    ///
+    /// The randomization helps expose race conditions and deadlocks that might not
+    /// appear in deterministic tests.
+    #[test]
+    fn test_randomized_concurrent_blob_retirement_operations() {
```
halfprice, line 522: Note: this is randomized test that tries to simulate the usage of BlobRetirementNotifier with very high frequency and check invariants (1) no stuckness (2) no reference leak.

crates/walrus-service/src/node/blob_retirement_notifier.rs
```diff
@@ -29,41 +30,47 @@ pub enum ExecutionResultWithRetirementCheck<T> {
...
     }
 
     /// Notify all BlobRetirementNotify for a blob.
     pub fn notify_blob_retirement(&self, blob_id: &BlobId) {
-        let notify = {
-            let mut registered_blobs = self
-                .registered_blobs
-                .lock()
-                .expect("mutex should not be poisoned");
-            registered_blobs.remove(blob_id)
```
halfprice, line 61: Note (less important comment, better to read after reading other comments first): previously, we remove the blob id from `registered_blobs` since theoretically, notify_waiters() should eventually cause the blob to be removed. However, I find this is not very clean and it's better to use the reference count as the ultimate metrics to decide whether a blob should be removed from BlobRetirementNotifier or not. Therefore, I removed this `remove` which theoretically should be a no-op. This is also the motivation for me to add test_randomized_concurrent_blob_retirement_operations test.

crates/walrus-service/src/node/metrics.rs
```diff
@@ -216,6 +216,10 @@ walrus_utils::metrics::define_metric_set! {
 
         #[help = "The last epoch for which garbage collection was started or finished"]
         garbage_collection_last_epoch: U64GaugeVec["status"],
+
+        #[help = "The number of blobs registered to be notified when the blob expires/gets \
+        deleted/gets invalidated"]
+        blob_retirement_notifier_registered_blobs: IntGauge[],
```
halfprice, line 222: This metrics can tell us in prod whether there will be any notifier leak in storage node, as another measure for checking issue.

crates/walrus-service/src/node/blob_retirement_notifier.rs
```diff
@@ -115,6 +126,38 @@ impl BlobRetirementNotifier {
...
+            // acquire_blob_retirement_notify (every acquire_blob_retirement_notify call creates
+            // a new BlobRetirementNotify reference). This is the original reference created when
+            // we first register the blob in BlobRetirementNotifier. We can safely remove it from
+            // the HashMap.
+            if current_count == 1 {
+                registered_blobs.remove(blob_id);
+                self.metrics
+                    .blob_retirement_notifier_registered_blobs
+                    .set(i64::try_from(registered_blobs.len()).unwrap_or(i64::MAX));
+            }
```
sadhansood, line 153: nit: wonder if it makes sense to add a warn! log if `current_count` ever becomes <= 0.

sadhansood (approved):
Very cool @halfprice! It took me some time to understand the deadlock but i get it now, and really appreciate your fix. It must not have been easy to track this down, great work!

crates/walrus-service/src/node/blob_retirement_notifier.rs
halfprice: Thanks for the suggestion! Added a warning in DROP if the ref count is 0 before substraction.

PR #2737, opened 2025-12-01
chore: address move package build lint warning

halfprice:
## Description

Mainly suppress

```
   ┌─ /home/ubuntu/walrus_simtest_tmp/tmp.IAr1cNunyD/.tmpUSdDMk/wal/sources/wal.move:35:37
   │
35 │     let (mut cap, metadata) = coin::create_currency(
   │                                     ^^^^^^^^^^^^^^^ The function 'sui::coin::create_currency' is deprecated: Use `coin_registry::new_currency_with_otw` instead
   │
   = This warning can be suppressed with '#[allow(deprecated_usage)]' applied to the 'module' or module member ('const', 'fun', or 'struct')
```

by allow deprecated usage, since we won't publish another WAL token again.

Contributes to WAL-1113.

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/contracts/wal/sources/wal.move
+++ b/contracts/wal/sources/wal.move
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /// The WAL token is the native token for the Walrus Protocol.
+#[allow(deprecated_usage)]
 module wal::wal;
 
 use sui::{coin::{Self, TreasuryCap, Coin}, dynamic_object_field as dof, url};
@@ -31,7 +32,7 @@ public struct TreasuryCapKey has copy, drop, store {}
 ///
 /// After publishing this, the `UpgradeCap` must be burned to ensure that the supply
 /// of minted WAL cannot change.
-#[allow(lint(share_owned), deprecated_usage)]
+#[allow(lint(share_owned))]
 fun init(otw: WAL, ctx: &mut TxContext) {
     let (mut cap, metadata) = coin::create_currency(
         otw,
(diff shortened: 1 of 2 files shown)


mlegner (commented):
Thanks for addressing this. AFAIU, however, if we want to suppress this in tests, we need to (also) make these changes in the `testnet-contracts` directory.

contracts/wal/sources/wal.move
```diff
@@ -31,7 +31,7 @@ public struct TreasuryCapKey has copy, drop, store {}
 ///
 /// After publishing this, the `UpgradeCap` must be burned to ensure that the supply
 /// of minted WAL cannot change.
-#[allow(lint(share_owned))]
+#[allow(lint(share_owned), deprecated_usage)]
```
karlwuest: Suggestion/Nit: We might as well directly add `#[allow(deprecated_usage)]` to the whole module to also catch future changes.

karlwuest (approved):
LGTM! As Markus said, we probably also need to apply this to `testnet-contracts`.

contracts/wal/sources/wal.move
halfprice: Thanks @karlwuest @mlegner for the suggestion. Applied to mode wide and also added to testnet-contracts. Do we need to add to `mainnet-contracts` as well? I originally thought that the change applied to `contracts/` will be copied over to `testnet-contracts` and `mainnet-contracts` during the next contract upgrade. That's not the case?

karlwuest: We use `testnet-contracts` in some tests (namely where we test the upgrade process). That's why we need it there as well to supress the linter. We don't need it in `mainnet-contracts` because those are not used in any tests. The next contract upgrade will not affect the `wal` package (since it cannot be upgraded), so that process doesn't matter for this.

halfprice: Thanks for the context!

PR #2738, opened 2025-12-01
ci: Updates jobs with dependency to site-builder to its latest version

nikos-kitmeridis:
## Description

~Updates jobs with dependency to site-builder to its latest version (v2 that uses quilts)~
Only docs action uses site-builder v2 (quilts)

## Test plan

@mlegner do you think it would be ok to manually trigger these workflows and verify that nothing breaks with the quilt-ed uploads?

### DIFF
--- a/.github/workflows/update-ws-binaries.yaml
+++ b/.github/workflows/update-ws-binaries.yaml
@@ -37,7 +37,7 @@ jobs:
           SUI_KEYSTORE: "${{ secrets.SUI_KEYSTORE }}"
           WALRUS_CONFIG: "${{ vars.WALRUS_CONFIG }}"
           SUI_NETWORK: mainnet
-          SITE_BUILDER_VERSION: v1
+          SITE_BUILDER_VERSION: v1 # keep uploading this as single blobs instead of quilts for now (SEW-509)
 
       - name: Create temporary directory
         run: "mkdir -p ${{ env.BUILD_DIR }}"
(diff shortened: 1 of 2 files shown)


.github/workflows/update-ws-binaries.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
           SUI_KEYSTORE: "${{ secrets.SUI_KEYSTORE }}"
           WALRUS_CONFIG: "${{ vars.WALRUS_CONFIG }}"
           SUI_NETWORK: mainnet
-          SITE_BUILDER_VERSION: v1 # use v1 temporarily
+          SITE_BUILDER_VERSION: v1
```
mlegner: Nit: Can we mention why we use v1 here and maybe create a backlog issue to revisit this in the future?

.github/workflows/publish-docs.yaml
```diff
@@ -79,11 +79,10 @@ jobs:
...
           WALRUS_CONFIG: "${{ vars.WALRUS_CONFIG }}"
           SUI_NETWORK: mainnet
-          SITE_BUILDER_VERSION: v1 # use v1 temporarily
+          SITE_BUILDER_VERSION: v2
 
       - name: Update Walrus Site
         run: >
           site-builder
           update ${{ env.BUILD_DIR }} ${{ vars.WALRUS_SITE_OBJECT_DOCS }}
           --epochs ${{ env.EPOCHS }}
```
mlegner, line 88: Question: Does the new site-builder delete previous versions automatically? If not, maybe it would be good to reduce the number of epochs a bit (currently 10).

nikos-kitmeridis: It does delete previous blobs and sui objects.

.github/workflows/update-ws-binaries.yaml
nikos-kitmeridis: Ticket: https://linear.app/mysten-labs/issue/SEW-509/consider-updating-walrus-jobs-to-site-builder-v2

PR #2739, opened 2025-12-01
feat(typed-store): add snapshot functionality

mlegner:
## Description

Also use it in blob-info consistency check.

## Test plan

Existing tests.

### DIFF
--- a/crates/typed-store/src/rocks.rs
+++ b/crates/typed-store/src/rocks.rs
@@ -27,6 +27,7 @@ use rocksdb::{
     Cache,
     ColumnFamilyDescriptor,
     CompactOptions,
+    DBAccess,
     DBPinnableSlice,
     DBWithThreadMode,
     Error,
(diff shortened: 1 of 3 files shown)


crates/typed-store/src/rocks.rs
```diff
@@ -1691,48 +1836,13 @@ where
...
-                .read_perf_ctx_metrics
-                .report_metrics(&self.cf);
-        }
-        match res {
-            Some(data) => Ok(Some(
-                bcs::from_bytes(&data).map_err(typed_store_err_from_bcs_err)?,
-            )),
-            None => Ok(None),
-        }
+        self.read_value_with_metrics(key_buf, true, result, start)
```
sadhansood: `read_value_with_metrics` builds RocksDBPerfContext after the read has already completed. This means `read_perf_ctx_metrics` will now report zeros and the actual RocksDB get is never measured. Perf metrics worked previously because the context was created before the DB call.

crates/typed-store/src/rocks.rs
```diff
@@ -1032,6 +1043,85 @@ impl<K, V> DBMap<K, V> {
...
+    pub fn get_with_snapshot(
+        &self,
+        snapshot: &RocksDBSnapshot<'_>,
+        key: &K,
+    ) -> Result<Option<V>, TypedStoreError>
+    where
+        K: Serialize,
+        V: DeserializeOwned,
+    {
+        let start = std::time::Instant::now();
```
sadhansood: i think we should also do a crossdb check here to prevent user passing in a snapshot of different db. We do similar checks in batch apis.

sadhansood (commented):
Thanks @mlegner for making this super useful addition! I have a couple comments (and especially addressing the perf ctx one is important as it is breaking existing behavior)

crates/typed-store/src/rocks.rs
halfprice: Thanks @sadhansood for pointing this out! This is very subtle and I also don't know about this. Can we also document this behavior?

@sadhansood I also have a question, RocksDBPerfContext turns on and off perf stats per thread. Is this because RocksDB is blocking call, so overlapping calls must happen in different threads and won't affect this operation's stats?

sadhansood: @halfprice yes that is my understanding (that perf ctx counters are thread local). Turning the perf context on/off only affects the current thread, and rocksdb ops are all blocking calls so nothing else can run on the thread concurrently to impact the counter value.

mlegner: Oh, I completely missed/misunderstood the importance of the performance context. Thanks for catching this, I will fix it.

mlegner: Fixed in cbae1e4880e40c9a5bf811275b0a71a148c4c9ab.

crates/typed-store/src/rocks.rs
mlegner: Done in 9a62c0fbb2c9c4165b07d51fbe5d14d667983a0e. Let me know what you think.

sadhansood (approved):
Looks good, thank you @mlegner!

PR #2742, opened 2025-12-02
feat(client): Support targeted uploads and cooperative cancellation

sadhansood:
## Description

This PR changes are all no-op currently but are needed for allowing client to opportunistically use pending cache for small blobs. The changes are as follows:
1. Allow distributed uploads to target specific node indices via `node_write_communications_by_index` enabling retries without rebuilding full committee comms.
2. Added `failed_node_indices` helper so callers can retry only failed nodes.
3. Add optional `CancellationToken` to distributed upload paths, so uploader now returns partial results and skips tail work on cancellation while still emitting progress and quorum events.

## Test plan

Existing tests

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -1604,12 +1604,10 @@ impl<T> WalrusNodeClient<T> {
         if let Some((target_epoch, _)) = target_nodes
             && committees.epoch() != target_epoch
         {
-            let message = format!(
-                "target nodes belong to epoch {target_epoch},
-                current committee epoch is {}; refresh committees before retrying targeted upload",
-                committees.epoch()
-            );
-            return Err(ClientError::other(std::io::Error::other(message)));
+            return Err(ClientError::other(StoreError::TargetEpochMismatch {
+                target_epoch,
+                current_epoch: committees.epoch(),
+            }));
         }
 
         let max_unencoded = blobs
--- a/crates/walrus-sdk/src/error.rs
+++ b/crates/walrus-sdk/src/error.rs
@@ -16,6 +16,17 @@ pub enum StoreError {
     /// The metadata could not be stored on the node.
     #[error("the metadata could not be stored")]
     Metadata(NodeError),
+    /// The target nodes belong to a different committee epoch than the current committees.
+    #[error(
+        "target nodes belong to epoch {target_epoch}, current committee epoch is {current_epoch}; \
+        refresh committees before retrying targeted upload"
+    )]
+    TargetEpochMismatch {
+        /// The epoch associated with the targeted nodes.
+        target_epoch: Epoch,
+        /// The epoch of the currently active committees.
+        current_epoch: Epoch,
+    },
     /// One or more slivers could not be stored on the node.
     #[error(transparent)]
     SliverStore(#[from] SliverStoreError),
@@ -106,6 +117,13 @@ impl ClientError {
                 | ClientErrorKind::NotEnoughSlivers
                 // The client was notified that the committee has changed.
                 | ClientErrorKind::CommitteeChangeNotified
+        ) || matches!(
+            self.kind.as_ref(),
+            ClientErrorKind::Other(error)
+                if matches!(
+                    error.downcast_ref::<StoreError>(),
+                    Some(StoreError::TargetEpochMismatch { .. })
+                )
         )
     }
 


crates/walrus-sdk/src/client/communication/factory.rs
```diff
@@ -195,6 +195,61 @@ impl NodeCommunicationFactory {
...
+        let mut seen = HashSet::new();
+        let mut comms = Vec::new();
+
+        for index in node_indices {
+            if !seen.insert(index) {
+                continue;
+            }
+
+            if index >= write_committee.members().len() {
+                return Err(ClientError::store_blob_internal(format!(
```
wbbradley, line 232: Is this expected to ever happen? Seems like a `panic!` situation, no?

crates/walrus-sdk/src/uploader.rs
```diff
@@ -194,91 +210,109 @@ impl DistributedUploader {
 
         let mut requests = WeightedFutures::new(futures);
         let start = Instant::now();
+        let n_shards: usize = self.committees.n_shards().get().into();
+        let cancel_token = cancellation.unwrap_or_default();
 
         let mut blobs_at_quorum = 0;
         let mut results: Vec<NodeResult<R, E>> = Vec::new();
+        let mut cancelled = false;
```
wbbradley: Can we avoid this variable by using `.is_cancelled`? https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html#method.is_cancelled

crates/walrus-sdk/src/client.rs
```diff
@@ -1549,6 +1551,46 @@ impl<T> WalrusNodeClient<T> {
...
+        .await
+    }
+
+    async fn distributed_upload_without_confirmation_inner(
+        &self,
+        blobs: &[(VerifiedBlobMetadataWithId, Arc<Vec<SliverPair>>)],
+        event_sender: tokio::sync::mpsc::Sender<UploaderEvent>,
+        tail_handling: TailHandling,
+        cancellation: Option<CancellationToken>,
+        target_nodes: Option<&[NodeIndex]>,
```
halfprice: question: this index is specific to one committee in one epoch, is that correct? What if the committee got inside this call is changed from what it was returned? Say a new epoch just started?

crates/walrus-sdk/src/client/communication/factory.rs
```diff
@@ -195,6 +195,61 @@ impl NodeCommunicationFactory {
...
+            }
+
+            match self.create_write_communication(
+                write_committee,
+                index,
+                sliver_write_limit.clone(),
+                auto_tune_handle.clone(),
+            ) {
+                Ok(Some(comm)) => comms.push(comm),
+                Ok(None) => continue,
```
halfprice: Question: when will this happen, and should this be a warning event?

crates/walrus-sdk/src/uploader.rs
```diff
@@ -46,6 +47,20 @@ pub struct RunOutput<R, E> {
     pub tail_handle: Option<JoinHandle<()>>,
 }
 
+/// Returns the unique set of node indices that returned an error.
+pub fn failed_node_indices<R, E>(results: &[NodeResult<R, E>]) -> Vec<NodeIndex> {
```
halfprice: nit: this is not used currectly

crates/walrus-sdk/src/uploader.rs
```diff
@@ -194,91 +210,109 @@ impl DistributedUploader {
...
+                tracing::debug!(
+                    ?reason,
+                    "uploader: detached tail handle completed with reason",
+                );
+                let results = requests.into_results();
+                tracing::debug!(
+                    "uploader: detached tail handle results: {:?}",
+                    results.len()
+                );
+            }))
```
halfprice, line 298: question: should cancellation have any effect in tail uploads?

crates/walrus-sdk/src/uploader.rs
sadhansood: done

crates/walrus-sdk/src/client/communication/factory.rs
sadhansood: ah, i'd still like to keep it as an error to avoid panic and lean toward surfacing a recoverable error since callers may hand us stale indices.

crates/walrus-sdk/src/client.rs
sadhansood: yes node index is tied to a specific epoch, i've now ensured that we pass that particular epoch that is used to derive the index and verify that committee epoch hasn't moved forward and surface an error which callers can handle by refreshing the committee.

crates/walrus-sdk/src/client/communication/factory.rs
sadhansood: It should not, but i've added a warn level log instead of silent skip.

crates/walrus-sdk/src/uploader.rs
sadhansood: removed.

crates/walrus-sdk/src/uploader.rs
sadhansood: The idea was that caller can just abort the tail handle but i've now added cancellation token in the tail handles as well. So anytime we trigger cancellation, it will also stop tail uploads and now user doesn't have to worry about explicitly invoking abort on the tail handle.

crates/walrus-sdk/src/client.rs
```diff
@@ -1558,6 +1601,16 @@ impl<T> WalrusNodeClient<T> {
...
         let committees = self.get_committees().await?;
+        if let Some((target_epoch, _)) = target_nodes
+            && committees.epoch() != target_epoch
+        {
+            let message = format!(
+                "target nodes belong to epoch {target_epoch},
+                current committee epoch is {}; refresh committees before retrying targeted upload",
+                committees.epoch()
+            );
+            return Err(ClientError::other(std::io::Error::other(message)));
```
halfprice: suggestion: can we add another error in StoreError? io::Error seems an odd error to use here.

halfprice (approved):
LGTM!

crates/walrus-sdk/src/client.rs

PR #2744, opened 2025-12-02
feat: Wire upload intent into client code

sadhansood:
## Description

This PR adds upload intent into the client code base. The intent is currently `Immediate` everywhere, but in the next PR we will start using `Pending` intent for blobs to be buffered in the storage node caches.

### DIFF
--- a/crates/walrus-sdk/src/client/communication/node.rs
+++ b/crates/walrus-sdk/src/client/communication/node.rs
@@ -26,6 +26,7 @@ use walrus_core::{
 use walrus_storage_node_client::{
     NodeError,
     StorageNodeClient,
+    UploadIntent,
     api::{BlobStatus, StoredOnNodeStatus},
 };
 use walrus_sui::types::StorageNode;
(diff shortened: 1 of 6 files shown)


crates/walrus-sdk/src/client/communication/node.rs
```diff
@@ -362,11 +363,28 @@ impl NodeWriteCommunication {
         &self,
         metadata: &VerifiedBlobMetadataWithId,
         pairs: impl IntoIterator<Item = &SliverPair>,
+    ) -> NodeResult<(), StoreError> {
+        self.store_metadata_and_pairs_without_confirmation_with_intent(
+            metadata,
+            pairs,
+            UploadIntent::Immediate,
```
halfprice: suggestion: instead of hard code UploadIntent::Immediate, should we add this as an input to store_metadata_and_pairs_without_confirmation to be a bit more flexible?

crates/walrus-sdk/src/client/communication/node.rs
```diff
@@ -412,13 +431,24 @@ impl NodeWriteCommunication {
...
+                self.retry_with_limits_and_backoff(|| {
+                    self.client.store_metadata_with_intent(metadata, intent)
+                })
+                .await?;
             }
         }
         Ok(metadata_status)
     }
 
+    #[allow(dead_code)]
```
halfprice: question: do we still need this function?

crates/walrus-sdk/src/client/communication/node.rs
```diff
@@ -593,19 +630,35 @@ impl NodeWriteCommunication {
     }
 
     /// Stores a sliver on a node.
+    #[allow(dead_code)]
```
halfprice: question: do we still need this function?

halfprice:
Hi @sadhansood , is this PR ready for another look? The diff seems larger than before. Maybe rebase to the main first?

PR #2746, opened 2025-12-03
fix(node): delete per-object blob info when blob is deleted

mlegner:
## Description

Previously, when processing a `BlobDeleted` event, we updated the per-object blob-info table using a `PerObjectBlobInfoMergeOperand` containing a `BlobStatusChangeType::Delete`. This is unnecessary, as we can simply delete the entry instead.

The previous behavior was not just inefficient, it could also cause a state where we still had a "deleted" per-object blob-info entry after having already deleted the aggregate blob-info entry.

Note that this behavior does *not* occur during the background garbage collection but was only possible when processing deletion events and if the garbage collection was enabled. Essentially, we update the aggregate and per-object blob info and at some point afterwards call [`process_blob_deleted_event`](https://github.com/MystenLabs/walrus/blob/a0fe2650979b49a4bbf8106792f8377aed0a61c8/crates/walrus-service/src/node/blob_event_processor.rs#L309), in which we [attempt to delete blob data](https://github.com/MystenLabs/walrus/blob/a0fe2650979b49a4bbf8106792f8377aed0a61c8/crates/walrus-service/src/node/blob_event_processor.rs#L339) (if data deletion is enabled). That function essentially assumes that the blob-info tables are consistent, and only looks at the aggregate blob info.

Note that the temporary state with a per-object blob-info entry and without aggregate blob-info entry did not affect the correctness as it is semantically still correct: The per-object entry is marked as "deleted" and would be actually removed from the DB in the next background garbage collection. However, the behavior is still unexpected, violates some formulated consistency properties, and is inefficient (as noted above).

---

This PR also contains some test improvements:
* Improve the blob-info internal consistency check and fail in tests if the consistency check fails.
* Improve logging and error outputs in tests.

Contributes to WAL-1111.

## Test plan

CI with extended tests.

### DIFF
--- a/crates/walrus-service/src/node/storage/blob_info.rs
+++ b/crates/walrus-service/src/node/storage/blob_info.rs
@@ -4,6 +4,7 @@
 //! Keeping track of the status of blob IDs and on-chain `Blob` objects.
 
 use std::{
+    collections::HashSet,
     fmt::Debug,
     num::NonZeroU32,
     ops::Bound::{self, Unbounded},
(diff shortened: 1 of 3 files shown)


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -162,20 +163,42 @@ impl BlobInfoTable {
...
-        if let Some(object_id) = event.object_id() {
-            let per_object_operation =
-                PerObjectBlobInfoMergeOperand::from_blob_info_merge_operand(operation)
-                    .expect("we know this is a registered, certified, or deleted event");
-            batch.partial_merge_batch(
-                &self.per_object_blob_info,
-                [(object_id, per_object_operation.to_bytes())],
-            )?;
-        }
+        self.update_per_object_blob_info(&mut batch, event)?;
```
halfprice, line 166: question: do we need to update [here](https://github.com/MystenLabs/walrus/blob/ef27570b54ad9b53fda068e2054464897d44bca8/crates/walrus-service/src/node/storage/blob_info.rs#L282) to use the new `update_per_object_blob_info` as well?

halfprice (commented):
Thanks @mlegner for the PR! Can you describe what is the issue this PR tries to fix?  To me, it seems just moving some of the per object blob info update from the merge operation to when processing the event. So keeping all the per object blob info update in one place seems cleaner.

Also, if this test is needed, do we still need the logic in the GC task where we go through per object blob info and find the expired ones?

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: No, because we only have extensions there. However, I can make this a bit clearer.

halfprice (approved):
Thanks @mlegner for adding detailed explanation on the PR description! This PR LGTM!

I understand it now. It basically all about data deletion during delete event processing. But this is very subtle and TBO I'm not sure if I can still remember this subtlety in the future. From maintenance perspective, one thing that I feel will confuse people is the existence of per object blob info processing still takes DELETED event as an input, but after this PR, it should not happen anymore. Can we maybe remove that code path or mark it as unreachable, in a follow up PR?

Also, how did you discover the issue? Was this triggered by some invariant check or based on your knowledge?

PR #2747, opened 2025-12-03
fix(node): avoid errors when storing metadata for untracked blobs

mlegner:
## Description

There is an extremely unlikely race condition when storing the blob metadata where a blob-info entry is deleted between checking it in the [`StorageNodeInner::store_metadata` method](https://github.com/MystenLabs/walrus/blob/6aba4f7b87359f3d8be3ec57e09fbec7465e1d6e/crates/walrus-service/src/node.rs#L3758) and actually persisting the metadata in the [`Storage::put_metadata` method](https://github.com/MystenLabs/walrus/blob/ef27570b54ad9b53fda068e2054464897d44bca8/crates/walrus-service/src/node/storage.rs#L644). This can result in a `BlobInfoMergeOperand::MarkMetadataStored` merge operand being applied to an untracked blob ID, which would previously result in a panic.

With this change, we allow the merge operation to proceed. The resulting blob info will then be removed during the next garbage collection.

Theoretically, we could also use a DB transaction to avoid this, but that adds complexity for an extremely rare case, and I don't see a relevant downside of this solution.

This commit also adds a new simtest where we repeatedly store and delete blobs with a relatively short epoch duration.

Contributes to WAL-1111.

## Test plan

CI with added simtest. Manual inspection.

### DIFF
--- a/crates/walrus-simtest/tests/simtest.rs
+++ b/crates/walrus-simtest/tests/simtest.rs
@@ -30,7 +30,13 @@ mod tests {
     use walrus_core::{Epoch, EpochCount};
     use walrus_proc_macros::walrus_simtest;
     use walrus_sdk::{
-        client::{StoreArgs, StoreBlobsApi as _, WalrusNodeClient, metrics::ClientMetrics},
+        client::{
+            StoreArgs,
+            StoreBlobsApi as _,
+            WalrusNodeClient,
+            metrics::ClientMetrics,
+            responses::BlobStoreResult,
+        },
         config::SliverWriteExtraTime,
     };
     use walrus_service::{
@@ -59,7 +65,7 @@ mod tests {
         client::{BlobPersistence, PostStoreAction, ReadClient, SuiContractClient, UpgradeType},
         system_setup::copy_recursively,
         test_utils::system_setup::{development_contract_dir, testnet_contract_dir},
-        types::move_structs::EventBlob,
+        types::{Blob, move_structs::EventBlob},
     };
     use walrus_test_utils::{WithTempDir, random_data_from_rng};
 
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1683,8 +1683,21 @@ impl Mergeable for BlobInfoV1 {
...
+                    during the next garbage collection"
+                );
+                Some(
+                    ValidBlobInfoV1 {
+                        is_metadata_stored,
+                        ..Default::default()
+                    }
+                    .into(),
+                )
+            }
```
halfprice, line 1699: question: does this mean that we may have orphaned metadata? Since the blob info is gone, but blob metadata may be written in the same write batch.

This also reminds me that in general, we don't check orphaned data in consistency check. Let me see if I can add this easily.

crates/walrus-service/src/node/storage/blob_info.rs
```diff
@@ -1683,8 +1683,21 @@ impl Mergeable for BlobInfoV1 {
                 epoch,
                 event: status_event,
             }),
+            BlobInfoMergeOperand::MarkMetadataStored(is_metadata_stored) => {
+                tracing::info!(
+                    is_metadata_stored,
+                    "marking metadata stored for an untracked blob ID; blob info will be removed \
```
halfprice, line 1689: question: why blob info will be removed? Isn't that the blob info has already be gone, which triggers the issue?

crates/walrus-simtest/tests/simtest.rs
```diff
@@ -1143,6 +1141,118 @@ mod tests {
...
+                                });
+                        }
+                    }
+                    Err(error) => {
+                        tracing::warn!(?error, "error reserving and storing blobs");
+                        continue;
+                    }
+                }
+            }
+        });
```
halfprice, line 1260: nit: I think this workload can reuse [start_background_workload](https://github.com/MystenLabs/walrus/blob/d535e5402d35ce14292f83b637eebc9b0651c92e/crates/walrus-simtest/src/test_utils.rs#L310) by providing a deletable probability and deleting probability.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: > does this mean that we may have orphaned metadata? Since the blob info is gone, but blob metadata may be written in the same write batch.

No, the write batch writes both the blob info and the metadata simultaneously. The blob info will only be deleted together with the metadata (in a transaction).

This does not prevent orphaned slivers however, which we would have to handle separately.

crates/walrus-service/src/node/storage/blob_info.rs
mlegner: The blob info had been deleted previously, but with this merge operand it will be recreated. During GC, we notice that this refers to a blob with no non-expired objects, so we will remove both the blob info and the related metadata.

crates/walrus-simtest/tests/simtest.rs
mlegner: I wanted to store multiple blobs simultaneously, but I can see if I can modify the existing workload. Also, I just noticed that I should also *delete* multiple blobs instead of just one, will change that.

crates/walrus-service/src/node/storage/blob_info.rs
halfprice: Thanks for the explanation! I got it now. I thought this change basically allows no-op `MarkMetadataStored`. I missed that, here, we are actually re-creating the BlobInfo. All my questions are due to missing this point. LGTM!

PR #2748, opened 2025-12-03
feat(client): Add oppportunistic upload in parallel to registration

sadhansood:
## Description

This PR adds opportunistic uploads for small blobs. The new workflow is as follows:
1. Kick off sliver uploads with pending intent in parallel with blob registration.
2. tokio::select! races registration against the uplaod task with pending intent. When registration wins, cancel the pending task.
3. Remember this from the pending upload task: which nodes succeeded (and their total weight) and which committee nodes didn’t get slivers.
4. Run the immediate upload seeded with that state: only target the remaining nodes and start with the previously accumulated weight when checking quorum.
 5. Collect confirmations. If quorum is still short (e.g., buffered slivers were evicted), do one more immediate upload pass to the nodes missing confirmations, then try to build the certificate again.

## Test plan

All existing tests are enabled to use this feature. Disabled in prod by default.

### DIFF
--- a/crates/walrus-sdk/src/client.rs
+++ b/crates/walrus-sdk/src/client.rs
@@ -99,7 +99,7 @@ use crate::{
     },
     config::CommunicationLimits,
     error::{ClientError, ClientErrorKind, ClientResult, StoreError},
-    uploader::{DistributedUploader, RunOutput, TailHandling, UploaderEvent, failed_node_indices},
+    uploader::{DistributedUploader, RunOutput, TailHandling, UploaderEvent},
     utils::{
         self,
         CompletedReasonWeight,
@@ -215,8 +215,9 @@ struct SendBlobOptions<'a> {
 /// Pending upload state carried through a single reserve+store execution.
 #[derive(Debug, Default, Clone)]
 struct PendingUploadContext {
-    initial_upload_weight: Option<HashMap<BlobId, usize>>,
-    initial_failed_nodes: Option<HashMap<BlobId, Vec<NodeIndex>>>,
+    /// Nodes that reported successful pending uploads per blob.
+    /// Nodes that did not report success (errors or missing results) are treated as missing.
+    initial_success_nodes: Option<HashMap<BlobId, Vec<NodeIndex>>>,
 }
 
 struct PendingUploadHandle<'a> {
@@ -1082,8 +1083,7 @@ impl WalrusNodeClient<SuiContractClient> {
         tracing::debug!(?registered_blobs);
         store_args.maybe_observe_store_operation(store_op_duration);
 
-        let pending_context =
-            self.apply_pending_upload_outcome(pending_upload_result, &pending_blobs, &committees);
+        let pending_context = self.apply_pending_upload_outcome(pending_upload_result);
 
         let (mut final_result, blobs_awaiting_upload, mut blobs_pending_certify_and_extend) =
             Self::partition_registered_blobs(registered_blobs, blobs_count);
@@ -1236,16 +1236,22 @@ impl WalrusNodeClient<SuiContractClient> {
             let mut pending_upload_result = None;
             let registered_blobs = tokio::select! {
                 reg = &mut registration_fut => {
+                    // Cancel pending uploads once registration completes; if they already finished
+                    // we'll still await their output below.
                     pending.cancel.cancel();
                     reg
                 }
                 pending_res = pending.future.as_mut() => {
+                    // Pending uploads may finish first; record their output before awaiting
+                    // registration.
                     pending_upload_result = Some(pending_res?);
                     registration_fut.await
                 }
             }?;
 
             if pending_upload_result.is_none() {
+                // Registration finished first; still await the pending task to capture any
+                // completed work before or during cancellation.
                 pending.cancel.cancel();
                 match pending.future.as_mut().await {
                     Ok(res) => pending_upload_result = Some(res),
@@ -1262,8 +1268,6 @@ impl WalrusNodeClient<SuiContractClient> {
     fn apply_pending_upload_outcome(
         &self,
         mut pending_upload_result: Option<RunOutput<Vec<BlobId>, StoreError>>,
-        pending_blobs: &[(VerifiedBlobMetadataWithId, Arc<Vec<SliverPair>>)],
-        committees: &ActiveCommittees,
     ) -> PendingUploadContext {
         let mut context = PendingUploadContext::default();
         // Cancel any in-flight pending tail. Tail handle is returned during the upload with
@@ -1275,29 +1279,9 @@ impl WalrusNodeClient<SuiContractClient> {
         }
 
         if let Some(pending_output) = pending_upload_result {
-            let weight_map = Self::success_weight_by_blob(&pending_output.results);
-            if !weight_map.is_empty() {
-                context.initial_upload_weight = Some(weight_map);
-            }
-            let mut failed_nodes = failed_node_indices(&pending_output.results);
-            // Add nodes that were not seen in the pending upload results to the failed nodes.
-            let committee_size = committees.write_committee().members().len();
-            let seen: HashSet<NodeIndex> = pending_output.results.iter().map(|r| r.node).collect();
-            for idx in 0..committee_size {
-                let node_idx = NodeIndex::from(idx);
-                if !seen.contains(&node_idx) {
-                    failed_nodes.push(node_idx);
-                }
-            }
-            failed_nodes.sort_unstable();
-            failed_nodes.dedup();
-            if !failed_nodes.is_empty() {
-                let failed_map = HashMap::from_iter(
-                    pending_blobs
-                        .iter()
-                        .map(|(meta, _)| (*meta.blob_id(), failed_nodes.clone())),
-                );
-                context.initial_failed_nodes = Some(failed_map);
+            let success_nodes = Self::successful_nodes_by_blob(&pending_output.results);
+            if !success_nodes.is_empty() {
+                context.initial_success_nodes = Some(success_nodes);
             }
         }
 
@@ -1754,6 +1738,9 @@ impl<T> WalrusNodeClient<T> {
         certificate
     }
 
+    /// Uploads slivers (optionally to a target node subset) and then collects confirmations from
+    /// the full write committee; missing confirmations from non-targeted nodes still count as
+    /// missing and can trigger retries.
     async fn send_blob_data_and_collect_confirmations(
         &self,
         metadata: &VerifiedBlobMetadataWithId,
@@ -2050,16 +2037,36 @@ impl<T> WalrusNodeClient<T> {
         pending_context: &PendingUploadContext,
     ) -> ClientResult<ConfirmationCertificate> {
         let blobs = vec![(metadata.clone(), pairs.clone())];
-        let target_nodes = if let Some(failed) = pending_context
-            .initial_failed_nodes
+        let committees = self.get_committees().await?;
+        let mut missing_nodes = None;
+        let mut initial_completed_weight = None;
+        if let Some(success_nodes) = pending_context
+            .initial_success_nodes
             .as_ref()
             .and_then(|m| m.get(metadata.blob_id()))
         {
-            let target_epoch = self.get_committees().await?.epoch();
-            Some((target_epoch, failed.as_slice()))
-        } else {
-            None
-        };
+            // Nodes that did not report success are treated as missing.
+            let success_set: HashSet<NodeIndex> = success_nodes.iter().copied().collect();
+            let all_nodes = committees.write_committee().members().len();
+            let missing: Vec<NodeIndex> = (0..all_nodes)
+                .filter(|idx| !success_set.contains(idx))
+                .collect();
+            if !missing.is_empty() {
+                missing_nodes = Some(missing);
+            }
+
+            let members = committees.write_committee().members();
+            let weight: usize = success_nodes
+                .iter()
+                .map(|idx| members[*idx].shard_ids.len())
+                .sum();
+            if weight > 0 {
+                initial_completed_weight = Some(HashMap::from([(*metadata.blob_id(), weight)]));
+            }
+        }
+        let target_nodes = missing_nodes
+            .as_ref()
+            .map(|nodes| (committees.epoch(), nodes.as_slice()));
 
         let send_options = SendBlobOptions {
             blob_persistence_type,
@@ -2068,7 +2075,7 @@ impl<T> WalrusNodeClient<T> {
             quorum_forwarder: store_args.quorum_event_tx.clone(),
             tail_handle_collector: store_args.tail_handle_collector.clone(),
             target_nodes,
-            initial_completed_weight: pending_context.initial_upload_weight.as_ref(),
+            initial_completed_weight: initial_completed_weight.as_ref(),
         };
 
         let (confirmation_results, upload_result) = self
@@ -2083,9 +2090,10 @@ impl<T> WalrusNodeClient<T> {
                     return Err(err);
                 }
 
-                let committees = self.get_committees().await?;
                 let certified_epoch = committees.write_committee().epoch;
 
+                // Pending slivers can be cache-evicted, so retry uploads to nodes that never
+                // confirmed.
                 self.retry_missing_nodes(
                     metadata,
                     blob_persistence_type,
@@ -2101,8 +2109,6 @@ impl<T> WalrusNodeClient<T> {
     }
 
     /// Retries the upload with immediate intent to the missing nodes until quorum is reached.
-    /// It is possible that we fail to get storage confirmation from quorum because buffered slivers
-    /// were cache evicted.
     #[allow(clippy::too_many_arguments)]
     async fn retry_missing_nodes(
         &self,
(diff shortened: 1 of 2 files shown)


crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -95,6 +104,8 @@ impl StoreArgs {
             quorum_event_tx: None,
             tail_handle_collector: None,
             encoding_event_tx: None,
+            initial_upload_weight: None,
+            initial_failed_nodes: None,
         }
```
mlegner, line 93: Nit: Can we reuse the `default_inner` here?

```rust
        Self {
            encoding_type,
            epochs_ahead,
            store_optimizations,
            persistence,
            post_store,
            ..Self::default_inner()
        }
```

crates/walrus-sdk/src/client.rs
```diff
@@ -946,6 +947,14 @@ impl WalrusNodeClient<SuiContractClient> {
             return Ok(vec![]);
         }
 
+        // Allow config to enable optimistic/pending uploads even if the caller
+        // used default store args.
+        let mut store_args = store_args.clone();
+        if self.config.communication_config.pending_uploads_enabled {
+            store_args.store_optimizations =
+                store_args.store_optimizations.with_optimistic_uploads(true);
```
mlegner: Nit: This seems a bit unnecessarily complicated.

```suggestion
            store_args.store_optimizations.optimistic_uploads = true;
```

crates/walrus-sdk/src/client.rs
mlegner, line 1023: Important: This PR adds more than 100 LOC to an already long and complex function. Can you take it as an occasion to refactor the function? Maybe something along the lines of your PR description?

> 1. Kick off sliver uploads with pending intent in parallel with blob registration.
> 1. tokio::select! races registration against the uplaod task with pending intent. When registration wins, cancel the pending task.
> 1. Remember this from the pending upload task: which nodes succeeded (and their total weight) and which committee nodes didn’t get slivers.
> 1. Run the immediate upload seeded with that state: only target the remaining nodes and start with the previously accumulated weight when checking quorum.
> 1. Collect confirmations. If quorum is still short (e.g., buffered slivers were evicted), do one more immediate upload pass to the nodes missing confirmations, then try to build the certificate again.

crates/walrus-sdk/src/client.rs
```diff
@@ -1387,6 +1500,8 @@ impl<T> WalrusNodeClient<T> {
         ByteRangeReadClient::new(self, self.config.byte_range_read_client_config.clone())
     }
 
+    const OPTIMISTIC_MAX_BLOB_BYTES: u64 = 4 * 1024 * 1024;
```
mlegner: Question: How is this constant chosen and could it be configurable?

crates/walrus-sdk/src/client.rs
```diff
@@ -1546,51 +1691,16 @@ impl<T> WalrusNodeClient<T> {
...
-            event_sender,
-            tail_handling,
-            Some(node_indices),
-            cancellation,
-        )
-        .await
-    }
-
-    async fn distributed_upload_without_confirmation_inner(
+    #[allow(clippy::too_many_arguments)]
```
mlegner: Suggestion: Instead of disabling the lint, could you refactor the code a bit?

crates/walrus-sdk/src/client.rs
```diff
@@ -972,17 +981,86 @@ impl WalrusNodeClient<SuiContractClient> {
...
+                    }
+                    let encoded = &blob_with_status.encoded_blob;
+                    let BlobData::SliverPairs(sliver_pairs) = &encoded.data else {
+                        return None;
+                    };
+                    let unencoded_len = encoded.metadata.metadata().unencoded_length();
+                    if !self.is_blob_small(unencoded_len) {
+                        return None;
+                    }
+                    Some((encoded.metadata.as_ref().clone(), sliver_pairs.clone()))
```
mlegner: Suggestion: This could be factored out into a separate method, e.g., on `WalrusStoreBlob<WalrusStoreBlobState<BlobWithStatus>>`.

crates/walrus-sdk/src/client.rs
```diff
@@ -1678,6 +1794,151 @@ impl<T> WalrusNodeClient<T> {
...
+                let committees = self.get_committees().await?;
+                let certified_epoch = committees.write_committee().epoch;
+                let confirmation_results = self
+                    .collect_confirmations(
+                        metadata.blob_id(),
+                        certified_epoch,
+                        blob_persistence_type,
+                        &committees,
+                    )
+                    .await?;
```
mlegner: Important: AFAIU, this means sending one additional request to each node, even though we already sent the request. Can we instead return the relevant information in the `send_blob_data_and_get_certificate` function?

crates/walrus-sdk/src/client/store_args.rs
sadhansood: done

crates/walrus-sdk/src/client.rs
sadhansood: done

crates/walrus-sdk/src/client.rs

crates/walrus-sdk/src/client.rs
sadhansood: It is now configurable (and defaults to server thresholds)

crates/walrus-sdk/src/client.rs

crates/walrus-sdk/src/client.rs

crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -72,6 +76,11 @@ pub struct StoreArgs {
     pub tail_handle_collector: Option<Arc<Mutex<Vec<JoinHandle<()>>>>>,
     /// Optional channel to forward encoding progress events to.
     pub encoding_event_tx: Option<UnboundedSender<EncodingProgressEvent>>,
+    /// Initial upload weight accumulated before certification (e.g., from pending uploads).
+    pub initial_upload_weight: Option<std::collections::HashMap<walrus_core::BlobId, usize>>,
```
halfprice: nit: import `std::collections::HashMap` in the file?

crates/walrus-sdk/src/client.rs
```diff
@@ -1002,15 +1047,15 @@ impl WalrusNodeClient<SuiContractClient> {
         );
         store_args.maybe_observe_checking_blob_status(status_timer_duration);
 
+        let pending_blobs = self.pending_upload_candidates(&encoded_blobs_with_status, &store_args);
```
halfprice, line 1064: question: how important is it to get `encoded_blobs_with_status`? This is not optional ATM, but does this needed for every upload?

crates/walrus-sdk/src/client/client_types.rs
```diff
@@ -694,6 +697,34 @@ impl WalrusStoreBlobMaybeFinished<BlobWithStatus> {
...
+        pending_uploads_enabled: bool,
+        store_args: &StoreArgs,
+        optimistic_upload_max_blob_bytes: u64,
+    ) -> Option<(VerifiedBlobMetadataWithId, Arc<Vec<SliverPair>>)> {
+        let WalrusStoreBlobState::Unfinished(blob_with_status) = &self.state else {
+            return None;
+        };
+        if blob_with_status.status.is_registered()
+            || !pending_uploads_enabled
+            || !store_args.store_optimizations.optimistic_uploads_enabled()
```
halfprice: Nit: pending_upload and optimistic_upload mean the same thing, is that correct? If so, can we converge them into one name?

crates/walrus-sdk/src/client.rs
```diff
@@ -1092,13 +1121,220 @@ impl WalrusNodeClient<SuiContractClient> {
...
+            store_args.persistence,
+            store_args.store_optimizations,
+        );
+
+        if let Some(ref mut pending) = pending_upload_handle {
+            tokio::pin!(registration_fut);
+            let mut pending_upload_result = None;
+            let registered_blobs = tokio::select! {
+                reg = &mut registration_fut => {
+                    pending.cancel.cancel();
```
halfprice, line 1241: question: does this mean that we are throwing away any uploads done?

crates/walrus-sdk/src/client.rs
```diff
@@ -1092,13 +1121,220 @@ impl WalrusNodeClient<SuiContractClient> {
...
+        mut pending_upload_result: Option<RunOutput<Vec<BlobId>, StoreError>>,
+        pending_blobs: &[(VerifiedBlobMetadataWithId, Arc<Vec<SliverPair>>)],
+        committees: &ActiveCommittees,
+    ) -> StoreArgs {
+        // Cancel any in-flight pending tail. Tail handle is returned during the upload with
+        // immediate intent later.
+        if let Some(ref mut pending_output) = pending_upload_result
+            && let Some(handle) = pending_output.tail_handle.take()
+        {
+            handle.abort();
```
halfprice, line 1278: Question: why do we want to cancel tail upload here?

crates/walrus-sdk/src/client.rs
```diff
@@ -1827,6 +2338,30 @@ impl<T> WalrusNodeClient<T> {
         ClientErrorKind::NotEnoughConfirmations(weight, committees.min_n_correct()).into()
     }
 
+    fn success_weight_by_blob(
+        results: &[NodeResult<Vec<BlobId>, StoreError>],
+        excluded_nodes: &HashSet<NodeIndex>,
```
halfprice: remark: excluded_node is always set to empty

crates/walrus-sdk/src/uploader.rs
```diff
@@ -122,13 +136,23 @@ impl DistributedUploader {
...
     ) -> Self {
         let mut work_items: HashMap<usize, Vec<UploadWorkItem>> = HashMap::new();
         let mut progress: HashMap<BlobId, BlobUploadProgress> = HashMap::new();
 
         for (metadata, pairs) in blobs {
             let blob_id = *metadata.blob_id();
-            progress.entry(blob_id).or_default();
+            let entry = progress.entry(blob_id).or_default();
+            if let Some(initial_weight) = initial_completed_weight.and_then(|m| m.get(&blob_id)) {
+                entry.completed_weight = *initial_weight;
```
halfprice, line 148: question: do we need to remember the nodes finished from the `initial_weight`?

crates/walrus-sdk/src/client/store_args.rs
```diff
@@ -72,6 +76,11 @@ pub struct StoreArgs {
...
+    /// Nodes that failed during pending uploads, keyed by blob ID, to scope immediate retries.
+    pub initial_failed_nodes:
+        Option<std::collections::HashMap<walrus_core::BlobId, Vec<NodeIndex>>>,
```
halfprice: remark: I find that passing these two HashMap, which are the implementation detail of the client execution, in StoreArgs is quite confusing. Can we not place them here, but use a dedicated data structure in `reserve_and_store_encoded_blobs`? The lifetime of these HashMap should be within `reserve_and_store_encoded_blobs`, is that correct?

crates/walrus-sdk/src/client.rs
```diff
@@ -1718,6 +2019,175 @@ impl<T> WalrusNodeClient<T> {
...
+            {
+                if result.is_ok() && confirmed_nodes.insert(*node) {
+                    confirmed_weight += weight;
+                }
+            }
+
+            let all_nodes: HashSet<NodeIndex> =
+                (0..committees.write_committee().members().len()).collect();
+            let missing_nodes: Vec<NodeIndex> =
+                all_nodes.difference(&confirmed_nodes).copied().collect();
```
halfprice, line 2155: question: IIUC, `confirmed_nodes` here are the nodes that performed `send_blob_data_and_collect_confirmations`, which are the nodes that did not finish pending upload successfully. Will the missing_nodes here include successfully uploaded nodes from pending upload?

crates/walrus-sdk/src/client.rs
```diff
@@ -1718,6 +2019,175 @@ impl<T> WalrusNodeClient<T> {
...
+            // Drain uploader progress events so bounded channels don't deadlock the retry path.
+            let buffer = committees
+                .write_committee()
+                .members()
+                .len()
+                .saturating_mul(blobs.len().max(1))
+                .saturating_add(1)
+                .max(1);
+            let (tx, mut rx) = tokio::sync::mpsc::channel(buffer);
+            tokio::spawn(async move { while rx.recv().await.is_some() {} });
```
halfprice, line 2190: question: is this just to make progress sending work? Looks a bit hacky

crates/walrus-sdk/src/client.rs
sadhansood: We need `encoded_blobs_with_status` for every upload because the store flow depends on current status. It tells us whether each blob is already registered permanent or invalid so we can choose the correct registration op and also decide eligibility for pending uploads.

crates/walrus-sdk/src/client.rs
sadhansood: Not necessarily. We cancel the pending task once registration finishes to stop extra background work, but we still await the pending future and capture any results that already completed (Check `apply_pending_upload_outcome` on how we use the results to capture how many nodes returned and how many failed in the committee set)

crates/walrus-sdk/src/client.rs
sadhansood: Since we are about to trigger immediate upload, and if there is a tail upload running - we can just abort it.

crates/walrus-sdk/src/client.rs
sadhansood: So, `confirmed_nodes` is the nodes that returned storage confirmation i.e. have data persisted (it could be from either pending or immediate intent but it doesnt matter). `missing_nodes` include nodes which have not persisted data yet (could be from ttl expiring on the pending cache because registration event arrived too late)

crates/walrus-sdk/src/client.rs
sadhansood: Yes, without draining this queue, sends can block and deadlock because the channel is bounded. I can make it an unbounded channel and then we wouldn;t need to drain it. Or we can make progress reporting optional but it make code more complicated.

crates/walrus-sdk/src/uploader.rs
sadhansood: Yes, we remember the failed nodes in the initial run. So, we don't retry uploads on those. Check

 ```/// Pending upload state carried through a single reserve+store execution.
#[derive(Debug, Default, Clone)]
struct PendingUploadContext {
    initial_upload_weight: Option<HashMap<BlobId, usize>>,
    initial_failed_nodes: Option<HashMap<BlobId, Vec<NodeIndex>>>,
}```

crates/walrus-sdk/src/client/store_args.rs
sadhansood: fixed

crates/walrus-sdk/src/client/client_types.rs
sadhansood: done

crates/walrus-sdk/src/client.rs
sadhansood: fixed

crates/walrus-sdk/src/client/store_args.rs
sadhansood: Created a new struct called `PendingContext` to store thins information.

crates/walrus-sdk/src/client.rs
halfprice: Can we document here and also line 1243 that even we cancel the pending future here, any completed work will still be retrieved at line 1244?

crates/walrus-sdk/src/client.rs
```diff
@@ -1718,6 +2032,176 @@ impl<T> WalrusNodeClient<T> {
...
+                    confirmation_results,
+                )
+                .await
+            }
+        }
+    }
+
+    /// Retries the upload with immediate intent to the missing nodes until quorum is reached.
+    /// It is possible that we fail to get storage confirmation from quorum because buffered slivers
+    /// were cache evicted.
```
halfprice: Nit: I think this doc is better to be moved to the callsite of `retry_missing_nodes`. It's important to provide the context there that this additional uploads are meant for addressing cache eviction.

crates/walrus-sdk/src/client.rs
halfprice: Yeah, and this requirement is because of the current implementation. Let revisit this later to make this cleaner.

crates/walrus-sdk/src/client.rs
halfprice: Ok. The confusing part is that `send_blob_data_and_collect_confirmations` is only uploading to target_nodes, but tries to collect confirmation from `all` nodes. Can you document this at the function doc? This is quite important, and given the length of `send_blob_data_and_collect_confirmations`, this is also not obvious.

crates/walrus-sdk/src/client.rs
```diff
@@ -1827,6 +2352,26 @@ impl<T> WalrusNodeClient<T> {
         ClientErrorKind::NotEnoughConfirmations(weight, committees.min_n_correct()).into()
     }
 
+    fn success_weight_by_blob(
```
halfprice: Suggestion: I think we should combine this function with `failed_node_indices`.

After `apply_pending_upload_outcome` returns, and when going into multiple layer where the next patch of uploads starts, it is not obvious that failed_nodes and success_weights are mutually exclusive.

crates/walrus-sdk/src/uploader.rs
halfprice: Do we need to keep both? Can we just keep `initial_failed_nodes`? Or rather initial_succeeded_nodes? Everything needed here should be able to be derived from `initial_succeeded_nodes`

crates/walrus-simtest/tests/simtest_pending_uploads.rs
```diff
@@ -0,0 +1,161 @@
...
+            .with_persistence(BlobPersistence::Deletable)
+            .with_tail_handling(TailHandling::Blocking);
+
+        let store_results = client
+            .inner
+            .reserve_and_store_blobs_retry_committees(vec![blob.clone()], vec![], &store_args)
+            .await
+            .expect("client should recover when pending uploads cannot reach quorum");
+
+        apply_restarts(&mut walrus_cluster, restart_handles).await;
```
halfprice, line 88: question: why do we need to wait for restart here? If the upload was successful, it means that at least one of the nodes has been restarted?

crates/walrus-simtest/tests/simtest_pending_uploads.rs
```diff
@@ -0,0 +1,161 @@
...
+
+        let blob = walrus_test_utils::random_data(512);
+        let store_args = StoreArgs::default_with_epochs(1)
+            .with_store_optimizations(StoreOptimizations::all().with_optimistic_uploads(true))
+            .with_persistence(BlobPersistence::Deletable)
+            .with_tail_handling(TailHandling::Blocking);
+
+        let store_results = client
+            .inner
+            .reserve_and_store_blobs_retry_committees(vec![blob.clone()], vec![], &store_args)
```
halfprice, line 83: question: what happens if the retry happens before the two nodes starts? Will error be returned in that case?

sadhansood: If the retry happens before the two nodes restart, the confirmation collection will still see them as missing. After `MAX_MISSING_CONFIRMATION_RETRY_ROUNDS` is exhausted the client returns `NotEnoughConfirmations`.

crates/walrus-sdk/src/client.rs
sadhansood: alright, addressed.

crates/walrus-simtest/tests/simtest_pending_uploads.rs
sadhansood: We wait so the cluster has valid node IDs and the restarted nodes are fully bootstrapped before the read.

crates/walrus-sdk/src/client.rs

crates/walrus-sdk/src/client.rs
sadhansood: done

crates/walrus-sdk/src/uploader.rs
sadhansood: done.

PR #2754, opened 2025-12-05
fix: WAL-874 and WAL-876 - shutdown coordination

wbbradley:
## Description

Clean up BlobEventProcessor and related shutdown coordination code.

## Test plan

CI Pipeline.

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -6,7 +6,7 @@
 use std::{
     collections::hash_map::Entry,
     future::Future,
-    num::{NonZero, NonZeroU16},
+    num::{NonZero, NonZeroU16, NonZeroUsize},
     pin::Pin,
     sync::{
         Arc,
@@ -179,6 +179,7 @@ use crate::{
         },
     },
     node::{
+        blob_event_processor::pending_events::PendingEventCounter,
         config::LiveUploadDeferralConfig,
         event_blob_writer::EventBlobWriter,
         garbage_collector::GarbageCollector,
@@ -194,12 +195,12 @@ pub mod event_blob_writer;
 pub mod server;
 pub mod system_events;
 
+pub(crate) mod blob_event_processor;
 pub(crate) mod consistency_check;
 pub(crate) mod db_checkpoint;
 pub(crate) mod errors;
 pub(crate) mod metrics;
 
-mod blob_event_processor;
 mod blob_retirement_notifier;
 mod blob_sync;
 mod config_synchronizer;
@@ -556,8 +557,9 @@ pub struct StorageNode {
     shard_sync_handler: ShardSyncHandler,
     epoch_change_driver: EpochChangeDriver,
     start_epoch_change_finisher: StartEpochChangeFinisher,
+    num_blob_event_processors: NonZeroUsize,
+    pending_event_counter: PendingEventCounter,
     node_recovery_handler: NodeRecoveryHandler,
-    blob_event_processor: BlobEventProcessor,
     garbage_collector: GarbageCollector,
     event_blob_writer_factory: Option<EventBlobWriterFactory>,
     config_synchronizer: Option<Arc<ConfigSynchronizer>>,
@@ -798,12 +800,6 @@ impl StorageNode {
         );
         node_recovery_handler.restart_recovery().await?;
 
-        let blob_event_processor = BlobEventProcessor::new(
-            inner.clone(),
-            blob_sync_handler.clone(),
-            config.blob_event_processor_config.num_workers,
-        );
-
         tracing::debug!(
             "num_checkpoints_per_blob for event blobs: {:?}",
             node_params.num_checkpoints_per_blob
@@ -850,7 +846,8 @@ impl StorageNode {
             epoch_change_driver,
             start_epoch_change_finisher,
             node_recovery_handler,
-            blob_event_processor,
+            num_blob_event_processors: config.blob_event_processor_config.num_workers,
+            pending_event_counter: PendingEventCounter::default(),
             garbage_collector,
             event_blob_writer_factory,
             config_synchronizer,
@@ -942,6 +939,11 @@ impl StorageNode {
     /// of the local shard storage.
     pub async fn existing_shards(&self) -> Vec<ShardIndex> {
         self.inner.storage.existing_shards().await
+    }
+
+    /// Returns the pending event counter for blob events.
+    pub fn get_pending_event_counter(&self) -> PendingEventCounter {
+        self.pending_event_counter.clone()
     }
 
     /// Wait for the storage node to be in at least the provided epoch.
@@ -1179,6 +1181,13 @@ impl StorageNode {
     }
 
     async fn process_events(&self) -> anyhow::Result<()> {
+        let blob_event_processor = BlobEventProcessor::new(
+            self.inner.clone(),
+            self.blob_sync_handler.clone(),
+            self.num_blob_event_processors,
+            self.pending_event_counter.clone(),
+        );
+
         let writer_cursor = match self.event_blob_writer_factory {
             Some(ref factory) => factory.event_cursor().unwrap_or_default(),
             None => EventStreamCursor::new(None, u64::MAX),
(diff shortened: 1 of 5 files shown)


wbbradley:
I've gotten the mainline code working, next I need to go back and retrofit the TestCluster code so that it exposes the `blob_event_processor`. This is unfortunate.

crates/walrus-service/src/node.rs
```diff
@@ -1662,9 +1691,11 @@ impl StorageNode {
         // the current epoch to be processed (note that this does not include waiting for all
         // pending blob syncs to finish). This is to make sure that the node is in a consistent
         // state before processing the epoch change start event.
-        self.blob_event_processor
+        blob_event_processor
+            .get_pending_event_counter()
             .wait_for_all_events_to_be_processed()
             .await;
+        tracing::info!("AAAAAAAAAAAAAAA");
```
mlegner: Nit: This placeholder should probably be replaced.

crates/walrus-service/bin/node.rs
mlegner, line 1: Question: Why change these log events to warnings? Seems to me that those cover expected behavior, so info seems appropriate to me.

crates/walrus-service/src/test_utils.rs
mlegner, line 1: Same question as above: Why use `warn` and `error` here?

wbbradley: Debugging.

crates/walrus-service/bin/node.rs
wbbradley: Debugging.

crates/walrus-service/src/node.rs
wbbradley: Debugging.

crates/walrus-service/src/node.rs
```diff
@@ -1186,6 +1188,13 @@ impl StorageNode {
     }
 
     async fn process_events(&self) -> anyhow::Result<()> {
+        let blob_event_processor = BlobEventProcessor::new(
```
halfprice: nit: can you document why we are creating the BlobEventProcessor on stack?

crates/walrus-service/src/node/blob_event_processor/pending_events.rs
```diff
@@ -0,0 +1,100 @@
...
+
+// Poll interval for checking pending background events.
+pub(crate) const PENDING_EVENTS_POLL_INTERVAL: Duration = Duration::from_millis(10);
+
+/// A utility struct that wraps an `Arc<AtomicU32>` for tracking pending events.
+/// Provides convenient `inc()` and `dec()` methods that return the new value.
+#[derive(Default, Debug, Clone)]
+pub struct PendingEventCounter {
+    inner: Arc<AtomicU32>,
+    total_enqueued: Arc<AtomicU64>,
```
halfprice: question: what is `total_enqueued` used for?

crates/walrus-service/src/node/blob_event_processor/pending_events.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// Waits for all events to be processed in the background processors.
+    pub async fn wait_for_all_events_to_be_processed(&self) {
+        while self.get() > 0 {
+            tokio::time::sleep(PENDING_EVENTS_POLL_INTERVAL).await;
+        }
+        tracing::error!(
+            total = self.total_enqueued.load(Ordering::SeqCst),
+            "All pending blob events have been processed in background processors."
+        );
```
halfprice: question: is this supposed to be an `error!`?

crates/walrus-service/src/node.rs
```diff
@@ -563,8 +564,9 @@ pub struct StorageNode {
     shard_sync_handler: ShardSyncHandler,
     epoch_change_driver: EpochChangeDriver,
     start_epoch_change_finisher: StartEpochChangeFinisher,
+    num_blob_event_processors: NonZeroUsize,
+    pending_event_counter: PendingEventCounter,
     node_recovery_handler: NodeRecoveryHandler,
-    blob_event_processor: BlobEventProcessor,
```
halfprice, line 560: suggestion: I feel that storing the internal state and config of BlobEventProcessor as StorageNode field is a bit unintuitive. Can we keep BlobEventProcessor as a struct member and put the other two inside? We can still pass self.blob_event_processor into StorageNode::process_event() if that is still prefered.

crates/walrus-service/src/node.rs
```diff
@@ -1186,6 +1188,13 @@ impl StorageNode {
     }
 
     async fn process_events(&self) -> anyhow::Result<()> {
+        let blob_event_processor = BlobEventProcessor::new(
+            self.inner.clone(),
+            self.blob_sync_handler.clone(),
+            self.num_blob_event_processors,
+            self.pending_event_counter.clone(),
+        );
```
halfprice, line 1189: remark: please see my other comments about keep BlobEventProcessor as a member of StorageNode, and we can pass in self.blob_event_processor. Also can we document why passing it is better?

crates/walrus-service/src/node/blob_event_processor/pending_events.rs
wbbradley: Just for logging.

wbbradley: Ah, yeah, this was for debugging. I can pull it out.

crates/walrus-service/src/node/blob_event_processor/pending_events.rs
wbbradley: good catch. that was for debugging.

crates/walrus-service/src/node.rs
wbbradley: The way to think about it is that it's not the internal state of the `BlobEventProcessor`. It's a synchronization primitive that is important to the `StorageNode`, and the `BlobEventProcessor` is allowed to increment/decrement it, but it doesn't solely own it, and it's not internal. 

This is in line with its actual usage which is as a semantic for synchronization within/around the storage node.

halfprice: For, `num_blob_event_processors`, I think your code is fine, because at some point, we will inevitably store the entire config in the StorageNode.

For `pending_event_counter`, I find it a bit unintuitive to use StorageNode field as a communication mechanism to communicate BlobEventProcessor internal state. But I'm not too strong about this, and if you find it easier to reason about, let's go with your solution.

wbbradley: The `process_events` method leverages the `BlobEventProcessor` to modify the `pending_event_counter`, and to track events, and process them, etc. Looking at the code, the `BlobEventProcessor` does not care what the actual value of the `pending_event_counter` is at any time. It's just - effectively - sending a message to the `StorageNode` saying "I'm adding a pending event" or "I'm removing a pending event." So, in this sense, the counter is actually not internal state of the `BlobEventProcessor` it is a side-effect of the algorithm that is tracked by the `StorageNode`. The processor does not use this state, the `StorageNode`'s clients (and its tests) do. The `StorageNode` itself only needs to be able to - at times - look at the number of outstanding pending events. In terms of ownership, the `StorageNode` object does not own the `BlobEventProcessor`, the method `StorageNode::process_events` owns it (it's on the stack.) There's a difference there that is important because one is bound by an algorithm, the other is bound by an object with a lifetime on purpose.

A good [rule of thumb](https://stackoverflow.com/questions/102009/when-is-it-best-to-use-the-stack-instead-of-the-heap-and-vice-versa) for when to use the stack and when to use the heap is:

> Use the stack when your variable will not be used after the current function returns. Use the heap when the data in the variable is needed beyond the lifetime of the current function.

In this case the counter is something that only the StorageNode and its clients care about. The BlobEventProcessor just signs up to do the work of mutating it, but it does not depend on it directly.

> For pending_event_counter, I find it a bit unintuitive to use StorageNode field as a communication mechanism to communicate BlobEventProcessor internal state

The unintuitiveness, I think, comes from the explicit assumption that we are communicating BlobEventProcessor state. This is not what we are communicating. We are communicating the state of the program `StorageNode::process_events`, which lives on the stack, but not for as long as `StorageNode`.

crates/walrus-service/src/node.rs
wbbradley: Resolving as per in-person communication.

PR #2758, opened 2025-12-09
feat: storage node list raw recovery symbols endpoint

halfprice:
## Description

This PR adds a new endpoint in storage nodes to return raw recovery symbol, which is meant to be used by the client
to reconstruct slivers that the client is interested. This endpoint can be used to spot reconstruct missing slivers
in byte range API or read quilt patch API.

Some design decision related to this PR:
- The server side operation is designed to be lightweight: it should only access DB and avoid any heavy computation.
- Since the symbols are used to reconstruct the source slivers, the server should not need to expand to calculate the requested symbols: there should always be a in place intersection between the target sliver and source sliver
- It does not return symbol proof: the client relies on sliver verification against metadata to check data integrity
- allow batching multiple source slivers to return symbols for multiple target slivers.

This PR only adds the new endpoint and server side logic. Client side logic will be added in follow up PR.

Related to WAL-1120

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/node.rs
+++ b/crates/walrus-service/src/node.rs
@@ -66,7 +66,7 @@ use walrus_core::{
     SliverPairIndex,
     SliverType,
     SymbolId,
-    by_axis::{Axis, ByAxis},
+    by_axis::{self, Axis},
     encoding::{
         DecodingSymbol,
         EitherDecodingSymbol,
@@ -327,11 +327,11 @@ pub trait ServiceState {
         filter: RecoverySymbolsFilter,
     ) -> impl Future<Output = Result<Vec<GeneralRecoverySymbol>, ListSymbolsError>> + Send;
 
-    /// Retrieves multiple raw recovery symbols.
+    /// Retrieves multiple decoding symbols.
     ///
-    /// Attempts to retrieve multiple raw recovery symbols, skipping any failures that occur.
+    /// Attempts to retrieve multiple decoding symbols, skipping any failures that occur.
     /// Returns an error if none of the requested symbols can be retrieved.
-    fn retrieve_multiple_raw_recovery_symbols(
+    fn retrieve_multiple_decoding_symbols(
         &self,
         blob_id: &BlobId,
         target_slivers: Vec<SliverIndex>,
@@ -3323,37 +3323,32 @@ impl StorageNodeInner {
     }
 
     /// Extracts decoding symbols from a source sliver for multiple target slivers.
-    /// Populate `output` with decoding symbols, keyed by target sliver index.
+    /// Returns a map of target sliver indexes to decoding symbols.
     fn extract_decoding_symbols_for_target_sliver_into_output<T: EncodingAxis>(
         &self,
         sliver: SliverData<T>,
         target_sliver_indexes: &[SliverIndex],
-        output: &mut BTreeMap<SliverIndex, Vec<EitherDecodingSymbol>>,
-    ) -> Result<(), ListSymbolsError>
+    ) -> Result<BTreeMap<SliverIndex, EitherDecodingSymbol>, ListSymbolsError>
     where
         EitherDecodingSymbol: From<DecodingSymbol<<T as EncodingAxis>::OrthogonalAxis>>,
     {
+        let mut output = BTreeMap::new();
         for target_sliver_index in target_sliver_indexes {
             if usize::from(target_sliver_index.get()) >= sliver.symbols.len() {
-                return Err(ListSymbolsError::RetrieveRawRecoverySymbolOutOfRange(
-                    format!(
-                        "target sliver index out of range: {} < {}",
-                        target_sliver_index.get(),
-                        sliver.symbols.len(),
-                    ),
-                ));
+                return Err(ListSymbolsError::RetrieveDecodingSymbolOutOfRange(format!(
+                    "target sliver index out of range: {} < {}",
+                    target_sliver_index.get(),
+                    sliver.symbols.len(),
+                )));
             }
 
             let symbol_bytes = sliver.symbols[target_sliver_index.as_usize()].to_vec();
             let sliver_index = sliver.index;
             let decoding_symbol: EitherDecodingSymbol =
                 DecodingSymbol::<T::OrthogonalAxis>::new(sliver_index.get(), symbol_bytes).into();
-            output
-                .entry(*target_sliver_index)
-                .or_default()
-                .push(decoding_symbol);
+            output.insert(*target_sliver_index, decoding_symbol);
         }
-        Ok(())
+        Ok(output)
     }
 
     fn get_encoding_type_for_blob(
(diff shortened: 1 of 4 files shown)


crates/walrus-service/src/node.rs
```diff
@@ -332,6 +337,19 @@ pub trait ServiceState {
         filter: RecoverySymbolsFilter,
     ) -> impl Future<Output = Result<Vec<GeneralRecoverySymbol>, ListSymbolsError>> + Send;
 
+    /// Retrieves multiple raw recovery symbols.
+    ///
+    /// Attempts to retrieve multiple raw recovery symbols, skipping any failures that occur.
+    /// Returns an error if none of the requested symbols can be retrieved.
+    fn retrieve_multiple_raw_recovery_symbols(
```
jpcsmith: Nit: We already use decoding symbols and recovery symbols in the code, perhaps just use decoding symbols as opposed to introducing the alias "raw recovery symbols" for decoding symbols

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -80,7 +87,9 @@ pub const DELETABLE_BLOB_CONFIRMATION_ENDPOINT: &str =
 /// The path to get recovery symbols.
 pub const RECOVERY_SYMBOL_ENDPOINT: &str = "/v1/blobs/{blob_id}/recoverySymbols/{symbol_id}";
 /// The path to get multiple recovery symbols.
-pub const RECOVERY_SYMBOL_LIST_ENDPOINT: &str = "/v1/blobs/{blob_id}/recoverySymbols";
+pub const LIST_RECOVERY_SYMBOL_ENDPOINT: &str = "/v1/blobs/{blob_id}/recoverySymbols";
+/// The path to get raw recovery symbols without proof.
+pub const LIST_RAW_RECOVERY_SYMBOL_ENDPOINT: &str = "/v1/blobs/{blob_id}/rawRecoverySymbols";
```
jpcsmith: Nit: rename decodingSymbols

crates/walrus-service/src/node.rs
```diff
@@ -3290,6 +3308,40 @@ impl StorageNodeInner {
...
+    /// Populate `output` with decoding symbols, keyed by target sliver index.
+    fn extract_decoding_symbols_for_target_sliver_into_output<T: EncodingAxis>(
+        &self,
+        sliver: SliverData<T>,
+        target_sliver_indexes: &[SliverIndex],
+        output: &mut BTreeMap<SliverIndex, Vec<EitherDecodingSymbol>>,
+    ) -> Result<(), ListSymbolsError>
+    where
+        EitherDecodingSymbol: From<DecodingSymbol<<T as EncodingAxis>::OrthogonalAxis>>,
+    {
```
jpcsmith, line 3334: Nit: After verifying the sliver indices, this could return an `impl Iterator<>` instead of using an output parameter.

crates/walrus-service/src/node.rs
```diff
@@ -4050,6 +4114,91 @@ impl ServiceState for StorageNodeInner {
...
+        )
+        .map_err(RetrieveSymbolError::RetrieveSliver)?;
+
+        let mut output: BTreeMap<SliverIndex, Vec<EitherDecodingSymbol>> = BTreeMap::new();
+        let n_shards = self.n_shards();
+        let owned_shards = self.owned_shards_at_latest_epoch();
+
+        // Makes sure that the target sliver indexes are all targeting source slivers, and not
+        // derived slivers. This endpoint is meant to only serve symbols used to recover
+        // source slivers.
```
jpcsmith, line 4107: Remark: Given this check that ensures the indices are targeting source slivers, it does not seem necessary to repeat the check in `extract_decoding_symbols_for_target_sliver_into_output`.

crates/walrus-service/src/node.rs
```diff
@@ -4075,6 +4114,91 @@ impl ServiceState for StorageNodeInner {
...
+        for source_pair_index in owned_shards
+            .iter()
+            .map(|shard_index| shard_index.to_pair_index(n_shards, blob_id))
+        {
+            // TODO(WAL-1120): make fetching slivers parallel.
+            let sliver_result = self
+                .retrieve_sliver_unchecked(blob_id, source_pair_index, target_type.orthogonal())
+                .await
+                .map_err(RetrieveSymbolError::RetrieveSliver)?;
+
```
sadhansood: Do we need to bail while failing to retrieve the sliver? The docstring above says that this endpoint is best effort which also makes sense, so I guess we can continue to process other shards?

crates/walrus-service/src/node/server/routes.rs
sadhansood: Should we also make nodes return proof along with the decoding symbol? Otherwise, how can we know which node(s) are returning garbage?

crates/walrus-service/src/node/server/routes.rs
```diff
@@ -577,6 +586,45 @@ fn limit_symbol_recovery_requests(
...
+pub async fn list_raw_recovery_symbols<S: SyncServiceState>(
+    State(state): State<RestApiState<S>>,
+    Path(BlobIdString(blob_id)): Path<BlobIdString>,
+    ExtraQuery(query): ExtraQuery<ListRawRecoverySymbolsQuery>,
+) -> Result<Bcs<BTreeMap<SliverIndex, Vec<EitherDecodingSymbol>>>, ListSymbolsError> {
+    // TODO(WAL-1120): add a configuration to throttle this endpoint.
+    let symbols = state
+        .service
+        .retrieve_multiple_raw_recovery_symbols(&blob_id, query.target_slivers, query.target_type)
+        .await?;
```
sadhansood, line 592: Should we validate that `target_slivers` is non empty?

halfprice: Good call! Doing this check avoid reading slivers from storage if input is empty.

crates/walrus-service/src/node/server/routes.rs
halfprice: Not returning proof is by design because calculating proof in storage node is expensive. You ask a great question about how to detect abnormally. The way I'm thinking is that on the client side, we can always try to read the entire blob and extract the "authentic" shard. With this, the client can then run encoding, and detect corrupt symbols. Reading the entire blob is still cheaper than requesting thousands of symbols with proof.

crates/walrus-service/src/node.rs
halfprice: Good point. Log a warning and continue.

crates/walrus-service/src/node/server/routes.rs
halfprice: This is done inside `retrieve_multiple_decoding_symbols`

crates/walrus-service/src/node.rs

crates/walrus-service/src/node.rs
halfprice: Done

crates/walrus-service/src/node.rs
halfprice: This one here is somewhat redundant. We want the one in `retrieve_multiple_decoding_symbols` so that the call does not do any unnecessary work if the later index is invalid.

For the check here, I wanted to make this method robust and prevent it from misuse, especially want to avoid unnecessary storage node crash. So I prefer to keep it here, even though in the current logic, this is redundant.

PR #2760, opened 2025-12-09
fix: Add a metric to track recovery without deferral

sadhansood:
## Description

This metric tracks recovery duration without deferral window.

### DIFF
--- a/crates/walrus-service/src/node/blob_sync.rs
+++ b/crates/walrus-service/src/node/blob_sync.rs
@@ -19,6 +19,7 @@ use rayon::prelude::*;
 use tokio::{
     sync::{Notify, Semaphore, watch},
     task::{JoinHandle, JoinSet},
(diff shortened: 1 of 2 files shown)


crates/walrus-service/src/node/blob_sync.rs
```diff
@@ -415,9 +416,11 @@ impl BlobSyncHandler {
                         (metrics::STATUS_CANCELLED, None)
                     },
 
-                    guard = async {
+                    (guard, start) = async {
                         synchronizer.wait_for_recovery_window().await;
 
+                        let active_start = tokio::time::Instant::now();
```
wbbradley, line 423: nit: Can we `use tokio::time::Instant` to avoid its fqn being used multiple times?

sadhansood: done

PR #2766, opened 2025-12-10
fix: runtime catchup using event blob and checkpoint tailing should never run at the same time

halfprice:
## Description

When trying to use event blob to catchup on events, the catchup manager always waits for checkpoint tailing to stop by
waiting on a notify. However, when starting the node, it always calls `notify_tailing_stopped`, which stores a notify
permit inside it. This causes `wait_for_tailing_stopped` in `catchup_using_event_blobs()` always return immediately,
which results in checkpoint downloading task and event blob catchup run at the same time.

This PR fix the issue by using the combination of notify_waiters and notify to avoid storing the permit. Also, we add
several invariant checks to guard the correctness of the interaction between catchup manager and checkpoint downloading
process.

Fix WAL-1115

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/event/event_processor/catchup.rs
+++ b/crates/walrus-service/src/event/event_processor/catchup.rs
@@ -133,7 +133,11 @@ impl EventBlobCatchupManager {
         let current_lag = self.get_current_lag().await?;
 
         if current_lag > lag_threshold {
-            tracing::info!(lag_threshold, "performing catchup - lag is above threshold");
+            tracing::info!(
+                lag_threshold,
+                current_lag,
+                "performing catchup - lag is above threshold"
+            );
             match self.perform_catchup().await {
                 Ok(()) => {}
                 Err(CatchupError::Recoverable(error)) => {
@@ -144,7 +148,11 @@ impl EventBlobCatchupManager {
                 }
             }
         } else {
-            tracing::info!(lag_threshold, "skipping catchup - lag is below threshold");
+            tracing::info!(
+                lag_threshold,
+                current_lag,
+                "skipping catchup - lag is below threshold"
+            );
         }
 
         Ok(())
@@ -318,6 +326,14 @@ impl EventBlobCatchupManager {
         let next_event_index = self
             .get_next_event_index()
             .map_err(|error| CatchupError::NonRecoverable(anyhow::Error::from(error)))?;
+
+        // Inv: when process_event_blob is running, checkpoint tailing should definitely be stopped.
+        if !coordination_state.is_tailing_stopped() {
+            return Err(CatchupError::Recoverable(anyhow::anyhow!(
+                "checkpoint tailing should be stopped when process_event_blob is running"
+            )));
+        }
+
         let processing_result = tokio::time::timeout(
             self.catchup_runtime_config.processing_timeout,
             self.process_event_blobs(blobs, next_event_index),
@@ -582,10 +598,8 @@ impl EventBlobCatchupManager {
             .await?;
         let verified_checkpoint = VerifiedCheckpoint::new_unchecked(checkpoint_summary.clone());
 
-        batch.insert_batch(
-            &self.stores.checkpoint_store,
-            [((), verified_checkpoint.serializable_ref())],
-        )?;
+        self.stores
+            .insert_checkpoint_in_batch(batch, &verified_checkpoint)?;
 
         let next_committee = self.get_next_committee(&checkpoint_summary).await?;
         batch.insert_batch(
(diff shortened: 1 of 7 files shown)


crates/walrus-service/src/event/event_processor/db.rs
```diff
@@ -149,4 +149,34 @@ impl EventProcessorStores {
...
+
+    /// Inserts a checkpoint into the checkpoint store in a batch.
+    /// Inv: it should never insert smaller sequence number than the current latest checkpoint
+    /// sequence number.
+    pub fn insert_checkpoint_in_batch(
+        &self,
+        batch: &mut DBBatch,
+        checkpoint: &VerifiedCheckpoint,
+    ) -> Result<(), TypedStoreError> {
+        #[cfg(debug_assertions)]
```
halfprice, line 161: it'll be nice to have this invariant check enabled in prod. But we are doing a DB read on every checkpoint, which I tried to avoid.

crates/walrus-service/src/event/event_processor/coordination.rs
```diff
@@ -82,27 +82,39 @@ impl CatchupCoordinationState {
...
             .is_tailing_stopped
             .load(std::sync::atomic::Ordering::Acquire)
         {
             return true;
         }
-        match tokio::time::timeout(timeout, self.tailing_stopped_notify.notified()).await {
+        match tokio::time::timeout(timeout, waiting_for_tailing_stopped_notify_future).await {
             Ok(()) => true,
             Err(_) => false,
         }
```
halfprice, line 120: Here is the second part of the fix.

crates/walrus-service/src/event/event_processor/coordination.rs
```diff
@@ -82,27 +82,39 @@ impl CatchupCoordinationState {
...
             .store(false, std::sync::atomic::Ordering::Release);
     }
 
     /// Notify that checkpoint tailing has fully stopped and wake any waiters.
     pub fn notify_tailing_stopped(&self) {
+        tracing::debug!("notifying tailing stopped");
         self.is_tailing_stopped
             .store(true, std::sync::atomic::Ordering::Release);
-        self.tailing_stopped_notify.notify_one();
+        self.tailing_stopped_notify.notify_waiters();
```
halfprice, line 95: Here is the first part of the fix.

crates/walrus-simtest/tests/simtest_event_blob.rs
```diff
@@ -306,7 +306,9 @@ mod tests {
             test_cluster::E2eTestSetupBuilder::new()
                 .with_epoch_duration(Duration::from_secs(15))
                 .with_num_checkpoints_per_blob(20)
-                //.with_event_stream_catchup_min_checkpoint_lag(Some(u64::MAX))
```
halfprice, line 309: I'm curious why this was disabled before.

crates/walrus-service/src/event/event_processor/processor.rs
```diff
@@ -364,6 +368,11 @@ impl EventProcessor {
         sui_macros::fail_point_async!("pause_checkpoint_tailing_entry");
 
         while let Some(entry) = rx.recv().await {
+            ensure!(
+                !coordination_state.is_tailing_stopped(),
+                "catchup using event blob is active, this is unexpected when tailing checkpoints \
+                is also active"
+            );
```
halfprice, line 375: This invariant check to make sure that catchup is not running if checkpoint downloading task is running

crates/walrus-service/src/event/event_processor/catchup.rs
```diff
@@ -318,6 +326,14 @@ impl EventBlobCatchupManager {
         let next_event_index = self
             .get_next_event_index()
             .map_err(|error| CatchupError::NonRecoverable(anyhow::Error::from(error)))?;
+
+        // Inv: when process_event_blob is running, checkpoint tailing should definitely be stopped.
+        if !coordination_state.is_tailing_stopped() {
+            return Err(CatchupError::Recoverable(anyhow::anyhow!(
+                "checkpoint tailing should be stopped when process_event_blob is running"
+            )));
+        }
```
halfprice, line 335: This invariant check is to make sure that when catchup is running, no checkpoint downloading is running. The signal here is still indirect since it just checks the bool. it'll be nice to have some direct check to make sure that the downloading task is indeed not running.

halfprice:
This is replaced by #2767

PR #2767, opened 2025-12-10
fix: Fix racing notification b/w catchup and checkpoint tailing

sadhansood:
## Description

This PR fixes the following race condition:
1.The catchup and tailing coordination uses a Notify to wait for “tailing has stopped” before catchup writes checkpoints. 
2.On startup we call `notify_tailing_stopped()` to mark the initial idle state. If that notification is sent when no one is waiting, it remains queued. Later, when runtime catchup starts and calls `wait_for_tailing_stopped()`, it consumes that old notification immediately and assumes tailing is stopped even though tailing is actually running. 
3.As a result, catchup proceeds in parallel with tailing: the catchup path writes new checkpoints, but the still-running tailing loop quickly overwrites checkpoint_store with its older value, so the restart sees a stale checkpoint.
4. **The fix is quite simple**: Only use `is_tailing_stopped` as the source of truth and avoid spurious notification wake ups. Just wait until `is_tailing_stopped` is set to true. Startup `notify_tailing_stopped()` still sets `is_tailing_stopped = true` so the first catchup returns immediately but a stale stored permit just causes one extra wake, then we keep waiting until the flag flips back to true.
5. 
## Test plan

Re-run the simtest

### DIFF
--- a/crates/walrus-service/src/event/event_processor/coordination.rs
+++ b/crates/walrus-service/src/event/event_processor/coordination.rs
@@ -3,7 +3,10 @@
 
 //! Coordination module for managing operation between checkpoint downloader and catchup manager.
 
-use std::{sync::Arc, time::Instant};
+use std::{
+    sync::{Arc, atomic::Ordering},
+    time::Instant,
+};
 
 use tokio::sync::{Mutex, Notify, mpsc};
 
@@ -35,9 +38,9 @@ pub struct CatchupCoordinationState {
     /// Timestamp when catchup was last started
     pub last_catchup_start: Arc<Mutex<Option<Instant>>>,
     /// Notifier signaled when checkpoint tailing has fully stopped
-    tailing_stopped_notify: Arc<Notify>,
+    pub tailing_stopped_notify: Arc<Notify>,
     /// Flag indicating whether checkpoint tailing is currently stopped
-    is_tailing_stopped: Arc<std::sync::atomic::AtomicBool>,
+    pub is_tailing_stopped: Arc<std::sync::atomic::AtomicBool>,
 }
 
 impl CatchupCoordinationState {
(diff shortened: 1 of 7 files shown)


crates/walrus-service/src/event/event_processor/coordination.rs
```diff
@@ -82,30 +85,38 @@ impl CatchupCoordinationState {
...
-        }
+        let fut = async {
+            loop {
+                if self.is_tailing_stopped.load(Ordering::Acquire) {
+                    return true;
+                }
+                self.tailing_stopped_notify.notified().await;
+            }
+        };
+        tokio::time::timeout(timeout, fut).await.is_ok()
```
sadhansood, line 124: This is the fix.

crates/walrus-service/src/event/event_processor/coordination.rs
```diff
@@ -82,30 +85,38 @@ impl CatchupCoordinationState {
...
-        match tokio::time::timeout(timeout, self.tailing_stopped_notify.notified()).await {
-            Ok(()) => true,
-            Err(_) => false,
-        }
+        let fut = async {
+            loop {
+                if self.is_tailing_stopped.load(Ordering::Acquire) {
+                    return true;
+                }
+                self.tailing_stopped_notify.notified().await;
```
halfprice: if `is_tailing_stopped == false`, and there is already a `tailing_stopped_notify` stored from the node start up, isn't the case that `self.tailing_stopped_notify.notified().await;` will return immediately, but is_tailing_stopped is still false?

sadhansood: Yes, exactly - that was the bug! The authoritative state is `is_tailing_stopped`. A stale permit from startup can wake the waiter, but we don’t exit until the flag is true. Startup ordering still works: (1) `notify_tailing_stopped` sets` is_tailing_stopped = true` (2) `mark_tailing_started` flips it back to false (the old permit remains) (3) `wait_for_tailing_stopped` may wake immediately on that old permit, but it re-checks and loops until `is_tailing_stopped` becomes true again when the tail task calls `notify_tailing_stopped`. This way a stale notification can’t let us proceed while tailing is running. I think this fix ensure the two properties of:
 a.) Safety: a stale permit can wake us, but we only return once the source of truth `is_tailing_stopped` is true again (which is only set by the tailing task when it ends), so catchup never proceeds while tailing is running.
b.) Liveness: if tailing is already stopped when we start waiting, the flag is already true and we return immediately, otherwise we keep waiting until the real stop flag is set to true.

halfprice: Ah, it is inside a loop. This makes sense.

Nit: I think we can still make `wait_for_tailing_stopped` a bit more robust by using notify_waiters instead of notify_one. The problem of using notify_one is that if there are multiple `wait_for_tailing_stopped` calls, only one will be waken up.

sadhansood: ok sure, i think we can make it `notify_waiters` now, ill make the change

sadhansood: done

halfprice: Thanks Sadhan!

Just one last suggestion: create `future = self.tailing_stopped_notify.notified()` before checking `is_tailing_stopped`, and poll the future if `is_tailing_stopped` is false.

In the current code, it's possible that `notify_waiters()` is called in between checking `is_tailing_stopped` and `self.tailing_stopped_notify.notified().await;`. When that happens, `self.tailing_stopped_notify.notified().await` will never get wake up.

This is very very unlikely, but I'd prefer let's not leave opportunities for race conditions.

sadhansood: Great suggestion @halfprice!

PR #2772, opened 2025-12-11
chore(node): enable DB transactions and garbage collection by default

mlegner:
## Description

This enables blob-info cleanup and data deletion, which were implemented in previous PRs (mainly https://github.com/MystenLabs/walrus/pull/2542, https://github.com/MystenLabs/walrus/pull/2599, and https://github.com/MystenLabs/walrus/pull/2725), by default. Also enables DB transactions, which were implemented in #2475 by default.

Also modifies the default batch sizes following test results on PTN and Testnet. The new values correspond to slightly less than 1s expected processing time per batch.

Smaller batch sizes are used in tests to ensure we actually test the batching logic.

## Test plan

Features were previously tested on PTN and Testnet.

---

## Release notes

- [x] Storage node: Enables DB transactions (implemented in https://github.com/MystenLabs/walrus/pull/2475) and garbage collection (implemented in https://github.com/MystenLabs/walrus/pull/2542, https://github.com/MystenLabs/walrus/pull/2599, and https://github.com/MystenLabs/walrus/pull/2725) by default on Testnet. The features can be disabled by adding the following to your node configuration:
  ```yaml
  db_config:
    global:
      use_optimistic_transaction_db: false
  garbage_collection:
    enable_blob_info_cleanup: false
    enable_data_deletion: false
  ```
  If you have `experimental_use_optimistic_transaction_db: true` in your configuration, you should remove that.

### DIFF
--- a/crates/walrus-service/bin/node.rs
+++ b/crates/walrus-service/bin/node.rs
@@ -64,7 +64,7 @@ use walrus_sui::{
     types::move_structs::VotingParams,
     utils::SuiNetwork,
 };
-use walrus_utils::load_from_yaml;
+use walrus_utils::config::Config as _;
 
 // Define the `GIT_REVISION` and `VERSION` consts
 walrus_utils::bin_version!();
@@ -505,8 +505,10 @@ fn main() -> anyhow::Result<()> {
             cleanup_storage,
             ignore_sync_failures: _,
         } => loop {
+            tracing::info!("loading node configuration from {}", config_path.display());
+            let config = StorageNodeConfig::load_and_validate(&config_path)?;
             let result = commands::run(
-                load_from_yaml(&config_path)?,
+                config,
                 cleanup_storage,
                 Arc::new(StorageNodeConfigLoader::new(config_path.clone())),
             );
@@ -870,7 +872,7 @@ mod commands {
     /// the new config file may not contain it after adding the storage node capability object ID.
     #[tokio::main]
     pub(crate) async fn register_node(config_path: PathBuf, force: bool) -> anyhow::Result<()> {
-        let mut config: StorageNodeConfig = load_from_yaml(&config_path)?;
+        let mut config: StorageNodeConfig = StorageNodeConfig::load_and_validate(&config_path)?;
         let contract_client = get_contract_client_from_node_config(&config).await?;
 
         if !force
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -251,7 +251,23 @@ impl Default for StorageNodeConfig {
     }
 }
 
+impl walrus_utils::config::Config for StorageNodeConfig {
+    fn validate(&self) -> anyhow::Result<()> {
+        if !self.db_config.use_optimistic_transaction_db()
+            && self.garbage_collection.enable_data_deletion
+        {
+            anyhow::bail!(
+                "data deletion is only supported when DB transactions are enabled; \
+                either set `db_config.global.use_optimistic_transaction_db` to `true` or \
+                `garbage_collection.enable_data_deletion` to `false`"
+            );
+        }
+        Ok(())
+    }
+}
+
 impl StorageNodeConfig {
+    /// Loads the config from a file.
     /// Rotates the protocol key pair.
     pub fn rotate_protocol_key_pair(&mut self) {
         if let Some(next_key_pair) = self.next_protocol_key_pair.clone() {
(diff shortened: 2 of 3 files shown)


crates/walrus-service/src/node/storage/database_config.rs
```diff
@@ -230,12 +230,10 @@ pub struct GlobalDatabaseOptions {
     pub wal_size_limit_mb: Option<u64>,
     /// Whether to enable statistics.
     pub enable_statistics: bool,
-    /// Temporary/experimental: if true, open databases using OptimisticTransactionDB instead of
-    /// the standard DB. This enables optimistic transactions for write paths that need
-    /// transactional semantics. Once transactions are fully adopted, this flag should be removed
-    /// and the behavior made unconditional.
-    #[serde(alias = "use_optimistic_transaction_db")]
-    pub experimental_use_optimistic_transaction_db: bool,
```
sadhansood, line 238: Is it possible any of the node operators are using `experimental_use_optimistic_transaction_db` in their  config? In that case, this might create serde issues - wondering if we should deprecate and add new config.

crates/walrus-service/node_config_example.yaml
```diff
@@ -228,9 +228,9 @@ node_recovery_config:
 blob_event_processor_config:
   num_workers: 10
 garbage_collection:
-  enable_blob_info_cleanup: false
-  enable_data_deletion: false
+  enable_blob_info_cleanup: true
+  enable_data_deletion: true
```
sadhansood, line 232: Should we do a validation is `enable_data_deletion` is true then `use_optimistic_transaction_db` must be true as well?

crates/walrus-service/node_config_example.yaml
```diff
@@ -228,9 +228,9 @@ node_recovery_config:
...
   enable_random_delay: true
   randomization_time_window_secs: null
-  blob_objects_batch_size: 10000
-  data_deletion_batch_size: 1000
+  blob_objects_batch_size: 100000
```
sadhansood: Just want to make sure these defaults are safe for nodes since this seem to be a 10x increase.

mlegner: This should actually make the process more efficient. The only downside is that if the node is shut down, they stay around longer. 

The parameters were chosen such that each batch should be finished within ~1s, but we can keep them a bit lower or even revert the change (as previous tests used the previous default values).

crates/walrus-service/node_config_example.yaml
mlegner: There is a check at runtime: If data deletion is enabled but transactions are not, data deletion will be skipped with a [warning log entry](https://github.com/MystenLabs/walrus/blob/165b05123d5403306e1cad3765b2149727ebe932/crates/walrus-service/src/node/storage.rs#L736).

crates/walrus-service/src/node/storage/database_config.rs
mlegner: I don't think we have promoted this flag in any way. But even if, we don't fail on unknown config parameters (just confirmed this by adding an extra parameter); so if some operator still has this in their config, this would just be ignored.

crates/walrus-service/node_config_example.yaml
sadhansood: Ah yeah I knew about the warning log entry, my point was this might create unexpected expectation for the user if they were to mistakenly flip `use_optimistic_transaction_db` to false but keep `enable_data_deletion` to true. And perhaps it'll be better if we validate it upfront and crash the binary to let user loudly know about this.

crates/walrus-service/src/node/storage/database_config.rs
sadhansood: Thanks for checking this @mlegner!

crates/walrus-service/node_config_example.yaml
sadhansood: ok ~1s is not bad, thanks for confirming.

mlegner: I've decreased the batch sizes a bit in b16ffc20a5e2edafe466a0ca652a6f4e8d3361e9 to be safe.

crates/walrus-service/node_config_example.yaml
mlegner: I've added config validation mechanisms in 3e4fb815892de013ac29bb47d1052839e0971374. This can be extended in the future and/or used for other configs as well.

sadhansood (approved):
Thank you @mlegner!

PR #2774, opened 2025-12-11
fix(ci): Use 1 SUI for gas-budget for docs-site update

nikos-terzo:
## Description

Set a gas-budget of 1 SUI for docs-site update.


jessiemongeon1:
Merging because of reported errors that are already fixed but since the site has not deployed, they are still present on the current site.

PR #2801, opened 2025-12-18
feat: implement blob streaming in the aggregator by slivers

wbbradley:
## Description

Implement streaming blobs by slivers via a new endpoint `/v1/blobs/{blob_id}/stream`.

## Test plan

Manually tested. CI Pipeline.

---

## Release notes

- [ ] Storage node:
- [x] Aggregator: There is a new alpha endpoint `/v1alpha/blobs/{blob_id}/stream` which is intended to reduce latency to first byte, and is intended to be used to stream blob data.
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-e2e-tests/Cargo.toml
+++ b/crates/walrus-e2e-tests/Cargo.toml
@@ -11,6 +11,7 @@ anyhow.workspace = true
 prometheus.workspace = true
 rand.workspace = true
 # explicitly import reqwest in test to disable its system proxy cache. It causes indeterminism in simtest.
+bytes.workspace = true
 futures.workspace = true
 hex.workspace = true
 indicatif.workspace = true
@@ -34,7 +35,6 @@ walrus-sui.workspace = true
 walrus-test-utils.workspace = true
 walrus-upload-relay = { workspace = true }
 walrus-utils.workspace = true
-bytes.workspace = true
 
 [lints]
 workspace = true


crates/walrus-sdk/src/client/streaming_read_client.rs
```diff
@@ -0,0 +1,634 @@
...
+    blob_size: u64,
+    /// Total slivers in the blob.
+    total_slivers: u16,
+}
+
+#[derive(Clone)]
+struct StreamingState {
+    /// Prefetch buffer managing sliver states.
+    prefetch_buffer: Arc<Mutex<SliverPrefetchBuffer>>,
+    /// Notifier for when slivers become ready.
```
wbbradley: I think this won't work because of Notify idiosyncrasies around calling `notify_one` more than once. I may rework this into a semaphore, or just go back to regular old polling.

wbbradley: I double checked this. It should be fine, actually.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -661,6 +663,114 @@ pub(super) async fn get_blob_byte_range<T: WalrusReadClient>(
...
+                headers,
+            );
+            // Add streaming-specific headers
+            headers.insert(
+                HeaderName::from_static("x-walrus-streaming"),
+                HeaderValue::from_static("true"),
+            );
+            // Add blob size header for progress tracking
+            if let Ok(size_value) = HeaderValue::from_str(&blob_size.to_string()) {
+                headers.insert(HeaderName::from_static("x-walrus-blob-size"), size_value);
```
halfprice: question: will the ContentRange header automatically set by the HTTP stream response? I think that is what is used to track progress.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -661,6 +663,114 @@ pub(super) async fn get_blob_byte_range<T: WalrusReadClient>(
...
+            let headers = response.headers_mut();
+            populate_response_headers_from_request(
+                request_method,
+                &request_headers,
+                &blob_id.to_string(),
+                headers,
+            );
+            // Add streaming-specific headers
+            headers.insert(
+                HeaderName::from_static("x-walrus-streaming"),
```
halfprice, line 772: Question: how will this header be used?

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -661,6 +663,114 @@ pub(super) async fn get_blob_byte_range<T: WalrusReadClient>(
...
+)]
+pub(super) async fn stream_blob<T: WalrusReadClient + Send + Sync + 'static>(
+    request_method: Method,
+    request_headers: HeaderMap,
+    State(client): State<Arc<T>>,
+    Path(BlobIdString(blob_id)): Path<BlobIdString>,
+) -> Response {
+    tracing::debug!("starting to stream blob");
+
+    match Arc::clone(&client).stream_blob(&blob_id).await {
```
halfprice: Nit: the clone here does seem to be necessary, since client is already an Arc.

crates/walrus-service/src/client/daemon/routes.rs
```diff
@@ -661,6 +663,114 @@ pub(super) async fn get_blob_byte_range<T: WalrusReadClient>(
...
+    SliverRetrievalFailed { message: String },
+
+    /// The blob size exceeds the maximum allowed size.
+    #[error("the blob size exceeds the maximum allowed size: {0}")]
+    #[rest_api_error(reason = "BLOB_TOO_LARGE", status = ApiStatusCode::SizeExceeded)]
+    BlobTooLarge(u64),
+
+    #[error(transparent)]
+    #[rest_api_error(delegate)]
+    Internal(#[from] anyhow::Error),
```
halfprice: question: do we need this case? Looks like below all other unspecified error type will be converted to SliverRetrievalFailed

crates/walrus-sdk/src/client/streaming.rs
```diff
@@ -0,0 +1,620 @@
...
+        return Ok((futures::stream::empty().left_stream(), 0));
+    }
+
+    let primary_sliver_size = get_primary_sliver_size(&client, blob_size, &metadata)?;
+    let total_slivers = u16::try_from(blob_size.div_ceil(u64::from(primary_sliver_size.get())))
+        .map_err(|_| {
+            ClientError::from(ClientErrorKind::Other(
+                "primary sliver count exceeds u16 max".into(),
+            ))
+        })?;
```
halfprice: remark: it should just call [this](https://github.com/MystenLabs/walrus/blob/c480fd8056356d212bf628a069609dec0683e239/crates/walrus-core/src/encoding/config.rs#L196) instead of calculating the number of the slivers.

crates/walrus-sdk/src/client/streaming.rs
```diff
@@ -0,0 +1,620 @@
...
+    client: Arc<WalrusNodeClient<T>>,
+    config: StreamingConfig,
+    state: StreamingState,
+) -> impl Stream<Item = Result<Bytes, ClientError>> + Send {
+    futures::stream::unfold(
+        (state, client, config),
+        |(state, client, config)| async move {
+            loop {
+                if state.is_aborted() {
+                    return None;
```
halfprice, line 355: Question: should this case return error?

crates/walrus-sdk/src/client/streaming.rs
```diff
@@ -0,0 +1,620 @@
...
+                }
+
+                spawn_prefetch_tasks(client.clone(), config.clone(), state.clone()).await;
+
+                if state.is_complete().await {
+                    return None;
+                }
```
halfprice: nit: move this before spawn_prefetch_tasks?

crates/walrus-sdk/src/client/streaming.rs
```diff
@@ -0,0 +1,620 @@
...
+                    sliver_index.get(),
+                    attempts
+                )
+                .into(),
+            )));
+        }
+
+        // Exponential backoff with cap
+        tokio::time::sleep(backoff).await;
+        backoff = std::cmp::min(backoff * 2, config.retry_max_backoff);
```
halfprice: Queston: why reimplementing the retry logic here instead of using the one in retrieve_slivers_retry_committees?

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: I'll take a look. IIRC, there was a borrowing issue and this was a way to work around it.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: I'll check. This is useful for calling anyhow::Error methods, and using `?`.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: The HTTP Content-Range [response header](https://developer.mozilla.org/en-US/docs/Glossary/Response_header) is used in [range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Range_requests) to indicate where the content of a response body belongs in relation to a complete resource.

This is not a range request.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: This is a hint to the client response handler that the data is in fact being streamed. It may or may not be useful, but I don't think it hurts and could aid in debugging.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Good catch. It wasn't necessary.

crates/walrus-service/src/client/daemon/routes.rs
wbbradley: Indeed it wasn't used. We can get rid of it.

crates/walrus-sdk/src/client/streaming.rs
wbbradley: In this case, the error has already been returned to the caller, but the maintenance of the aborted flag is just to bulletproof the state machine of that streaming `futures::stream::unfold` and make sure that it ends iterating on the loop in case it is awaited again.

crates/walrus-sdk/src/client/streaming.rs
wbbradley: Because there is no retry logic in there for lower level errors (like network errors, etc.) See https://linear.app/mysten-labs/issue/WAL-865/retry-retrieving-slivers-only-on-retriable-errors.

wbbradley: Ah, maybe I read that error case wrong, actually. It looks like there is retry logic. I'll remove my retry logic then.

wbbradley:
@halfprice I believe I've addressed your comments. I've also implemented a new e2e test called `test_streaming_blob`.

crates/walrus-service/src/client/daemon/routes.rs
halfprice: TIL. After looking it up, my assumption that streaming reads are implemented by range reads is incorrect. It's good that cloudflare will cache streaming reads (but not range reads).

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -2990,3 +2992,79 @@ async fn test_byte_range_read_size_too_large() -> TestResult {
     }
     Ok(())
 }
+
+/// Tests that streaming a blob returns the correct data.
+#[ignore = "ignore E2E tests by default"]
+#[walrus_simtest]
+async fn test_streaming_blob() -> TestResult {
```
halfprice, line 2999: Thanks for adding the test!

PR #2802, opened 2025-12-19
feat: adding client interface to fetch decoding symbols

halfprice:
## Description

This PR add client side interface to call ListDecodingSymbols endpoint to fetch symbols to reconstruct slivers on client side.

Contribute to WAL-1120

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-storage-node-client/src/client.rs
+++ b/crates/walrus-storage-node-client/src/client.rs
@@ -345,7 +345,7 @@ where
 /// Filter for [`StorageNodeClient::list_decoding_symbols()`] endpoint.
 #[derive(Debug, Clone)]
 pub struct DecodingSymbolsFilter {
-    /// The sliver indexes of the target sliver being recovered.
+    /// The sliver indexes of the target slivers being recovered.
     pub target_slivers: Vec<SliverIndex>,
     /// The type of the sliver being recovered.
     pub target_type: SliverType,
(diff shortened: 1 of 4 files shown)


crates/walrus-storage-node-client/src/client.rs
```diff
@@ -333,6 +342,36 @@ where
...
+
+        let mut map = serializer.serialize_map(Some(self.target_slivers.len() + 1))?;
+        for sliver in &self.target_slivers {
+            // Serialize the inner u16 value directly, not the SliverIndex wrapper
+            map.serialize_entry(TARGET_SLIVERS, &sliver.0)?;
+        }
+        map.serialize_entry(TARGET_TYPE, &self.target_type)?;
+        map.end()
+    }
+}
```
halfprice, line 373: Failed to find any more "native" way in serde to conduct the serialization. Nor did claude know....

jpcsmith: Remark: `SymbolIdFilter` seems to solve the same problem using `serialize_with` on the vector field. The examples in the test cases seem to match what you expect.

halfprice: Yeah, that was the first thing I tried/followed. But the serialization always returns `source: Custom("unsupported value")`, which I could not figure out how to resolve.

The way I implemented is something like
```
fn serialize_index_in_query<S>(
    sliver_indexes: &[SliverIndex],
    serializer: S,
) -> Result<S::Ok, S::Error>
where
    S: Serializer,
{
    serializer.collect_map(
        sliver_indexes
            .iter()
            .map(|index| ("targetSlivers", index.0)),
    )
}
```

Very similar to `serialize_ids_in_query`and this doesn't work.

jpcsmith: You're right, that can't work: the serialiser expects the fields to be basic values not vecs or maps. The enum + flatten is likely what avoids the issue in the `SymbolIdFilter`. You're approach here is best.

PR #2809, opened 2025-12-29
feat: allow configuration of max_body_size_limit in upload relay

wbbradley:
## Description

Allow configuration of max_body_size_limit in upload relay.

## Test plan

Manual testing. CI Pipeline.

### DIFF
--- a/crates/walrus-e2e-tests/tests/test_client.rs
+++ b/crates/walrus-e2e-tests/tests/test_client.rs
@@ -2612,12 +2612,14 @@ async fn test_store_with_upload_relay_no_tip() {
     let upload_relay_sui_client = get_client_with_config(walrus_client_config, &registry)
         .await
         .expect("create upload relay sui client");
+    const BLOB_SIZE: usize = 40000;
     let upload_relay_handle: UploadRelayHandle = walrus_upload_relay::start_upload_relay(
         upload_relay_sui_client,
         WalrusUploadRelayConfig {
             tip_config: TipConfig::NoTip,
             tx_freshness_threshold: Duration::from_mins(5),
             tx_max_future_threshold: Duration::from_secs(10),
+            max_body_size_limit: BLOB_SIZE,
         },
         server_address,
         registry,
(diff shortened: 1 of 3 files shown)


crates/walrus-simtest/tests/simtest_failure.rs
```diff
@@ -608,6 +608,7 @@ mod tests {
             .with_test_nodes_config(
                 TestNodesConfig::builder()
                     .with_node_weights(&[1, 2, 3, 3, 4])
+                    .with_node_recovery_config(node_recovery_config)
```
wbbradley, line 611: This was causing a simtest build warning and was a regression caused by an earlier refactoring.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -2646,9 +2648,13 @@ async fn test_store_with_upload_relay_no_tip() {
...
-    match basic_store_and_read(&cluster_client, 1, 40000, Some(upload_relay_client), || {
-        Ok(())
-    })
+    match basic_store_and_read(
+        &cluster_client,
+        1,
+        BLOB_SIZE,
+        Some(upload_relay_client),
+        || Ok(()),
+    )
```
halfprice, line 2651: suggestion: looks like you can add a simple test case where the test tries to store BLOB_SIZE +1, and expect to get rejected.

crates/walrus-e2e-tests/tests/test_client.rs
```diff
@@ -2615,12 +2615,14 @@ async fn test_store_with_upload_relay_no_tip() {
...
         .await
         .expect("create upload relay sui client");
+    const BLOB_SIZE: usize = 40000;
     let upload_relay_handle: UploadRelayHandle = walrus_upload_relay::start_upload_relay(
         upload_relay_sui_client,
         WalrusUploadRelayConfig {
             tip_config: TipConfig::NoTip,
             tx_freshness_threshold: Duration::from_mins(5),
             tx_max_future_threshold: Duration::from_secs(10),
+            max_body_size_limit: BLOB_SIZE,
```
halfprice, line 2622: question: this could be set to anything that is larger than BLOB_SIZE, is that correct? Any particular reason to use BLOB_SIZE instead of the `default_max_body_size_limit()` here?

halfprice (approved):
LGTM! It'll be great if we can test that this works as expected, since a simple test case that stores limit size +1 would do it.

PR #2810, opened 2025-12-29
chore: refactor Wallet to avoid unnecessary Result return types

wbbradley:
## Description

This PR should **not** result in any noticeable functional change at runtime. Let me know if you see something that is likely to change any runtime behaviors. This is a predominantly compiler-driven refactoring at the type/trait level.

Improve the interface for `walrus_sui::wallet::Wallet` by caching some data on initialization to avoid later failures and clean up most related callsites.

## Test plan

CI Pipeline.

### DIFF
--- a/crates/walrus-sui/src/wallet.rs
+++ b/crates/walrus-sui/src/wallet.rs
@@ -6,7 +6,6 @@
 use std::{
     collections::BTreeSet,
     path::{Path, PathBuf},
-    sync::Arc,
 };
 
 use sui_package_management::LockCommand;
@@ -24,12 +23,11 @@ use sui_types::{
 
 /// The `Wallet` struct wraps the `WalletContext` from the Sui SDK. This allows us to
 /// reduce the scope of the `WalletContext` to only the methods we need.
-#[derive(Clone)]
 pub struct Wallet {
     active_address: SuiAddress,
     active_env: SuiEnv,
     config_path: PathBuf,
-    wallet_context: Arc<WalletContext>,
+    wallet_context: WalletContext,
 }
 
 impl std::fmt::Debug for Wallet {
@@ -52,7 +50,7 @@ impl Wallet {
             active_address: wallet_context.active_address()?,
             active_env: wallet_context.config.get_active_env()?.clone(),
             config_path: wallet_context.config.path().to_path_buf(),
-            wallet_context: Arc::new(wallet_context),
+            wallet_context,
         })
     }
 
@@ -118,7 +116,7 @@ impl Wallet {
         response: &SuiTransactionBlockResponse,
     ) -> Result<(), WalletError> {
         sui_package_management::update_lock_file(
-            self.wallet_context.as_ref(),
+            &self.wallet_context,
             lock_command,
             install_dir,
             lock_file,


crates/walrus-service/src/testbed.rs
```diff
@@ -884,7 +882,7 @@ async fn create_storage_node_wallets(
 fn print_wallet_addresses(wallets: &mut [Wallet]) -> anyhow::Result<()> {
```
halfprice: remark: this function no longer needs to return `Result<()>`

crates/walrus-sui/src/client.rs
```diff
@@ -2340,7 +2340,7 @@ impl SuiContractClientInner {
     pub fn transaction_builder(&mut self) -> SuiClientResult<WalrusPtbBuilder> {
```
halfprice: Similarly, this now can just return `WalrusPtbBuilder`

crates/walrus-sui/src/wallet.rs
```diff
@@ -20,11 +21,16 @@ use sui_types::{
...
+#[derive(Clone)]
 pub struct Wallet {
-    wallet_context: WalletContext,
+    active_address: SuiAddress,
+    active_env: SuiEnv,
+    config_path: PathBuf,
+    wallet_context: Arc<Mutex<WalletContext>>,
 }
 
 impl std::fmt::Debug for Wallet {
```
halfprice, line 33: suggestion: print other fields such as `active_address`, `active_env`, etc. as well, since they are now pretty easy to get.

crates/walrus-sui/src/wallet.rs
```diff
@@ -39,32 +45,49 @@ pub type WalletError = anyhow::Error;
...
+            .await
     }
 
     // TODO: WAL-820 move callsites to the RetriableSuiClient.
     /// Passes through to the `WalletContext` to get an [`sui_types::base_types::ObjectRef`].
     #[deprecated(note = "Avoid this method. Use the RetriableSuiClient instead.")]
     pub async fn get_object_ref(&self, object_id: ObjectID) -> Result<ObjectRef, WalletError> {
-        self.wallet_context.get_object_ref(object_id).await
+        self.wallet_context
+            .lock()
```
halfprice: remark: I think there may be performance implication of this locking here. I recall that @mlegner tried to make some of the Sui RPC parallel to improve client performance, and this essentially removed that benefit. One work around is to use RwLock instead: `sign_transaction` and `execute_transaction_may_fail` acquire write lock, and all other accesses acquire read lock.

crates/walrus-sui/src/wallet.rs
```diff
@@ -20,11 +21,16 @@ use sui_types::{
     message_envelope::Envelope,
     transaction::{SenderSignedData, Transaction, TransactionData},
 };
+use tokio::sync::Mutex;
 
 /// The `Wallet` struct wraps the `WalletContext` from the Sui SDK. This allows us to
 /// reduce the scope of the `WalletContext` to only the methods we need.
+#[derive(Clone)]
```
halfprice: question: do we clone the `Wallet` anywhere in the code? If not, I think we can still keep `wallet_context: WalletContext` for the performance (see my other comments about rpc performance affected by locking).

halfprice (commented):
Thanks @wbbradley for the cleanup! I like the change that reduce the usage of Result. I just have one comment about performance below.

crates/walrus-sui/src/wallet.rs
mlegner: Yes, this would be indeed problematic and affect all read operations (including `get_reference_gas_price` above).

However, both of these methods here are marked as deprecated and are only used in tests and system setup where it doesn't matter. So I don't think we need to worry about it.

wbbradley: I don't think we call any of these functions enough to make this a performance bottleneck. Also, this change does not break any existing concurrency because prior to this change `Wallet` was not Cloneable therefore it wasn't being shared anyway. This doesn't add any prominent instances of cloning `Wallet`, so I believe it's fine. 

I'll go ahead and resolve this based on @mlegner's feedback, and my intuition above. @halfprice please let me know if you do not concur.

wbbradley: I will go ahead and switch it over to an RwLock.

wbbradley: I'm realizing now that it may not need a lock at all. We only call const-ref methods on it... I might be able to simply use an Arc by itself.

wbbradley: Ah. I remember why I need the Mutex, because WalletContext is not Sync.

wbbradley: Actually, it is Sync. I'm going to remove the Mutex altogether. I actually don't see a need for it.

crates/walrus-sui/src/wallet.rs
wbbradley: Fair point that we might not need to clone Wallet yet. Right now we don't need it. In later refactors we will probably need it. I'll rename this PR and make it just about removing these unnecessary Result types.

PR #2813, opened 2026-01-02
fix: implement quilt patch methods for ClientMultiplexer

liquid-helium:
The daemon's ClientMultiplexer was missing implementations for quilt patch reading methods, causing "quilt functionality not supported by this client" errors when calling quilt patch endpoints.

## Description

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-service/src/client/daemon.rs
+++ b/crates/walrus-service/src/client/daemon.rs
@@ -118,18 +118,7 @@ pub trait WalrusReadClient {
     fn get_blobs_by_quilt_patch_ids(
         &self,
         _quilt_patch_ids: &[QuiltPatchId],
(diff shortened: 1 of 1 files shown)


crates/walrus-service/src/client/multiplexer.rs
```diff
@@ -190,6 +191,29 @@ impl WalrusReadClient for ClientMultiplexer {
     ) -> ClientResult<BlobWithAttribute> {
         self.read_client.get_blob_by_object_id(blob_object_id).await
     }
+
```
wbbradley, line 194: Can we remove the default impl of these methods now? (The ones in the trait definition.)

wbbradley (approved):
I've added the cleanup of the default impls. LGTM

PR #2816, opened 2026-01-05
feat: allow expired certificates if the key matches the pinned key

jpcsmith:
## Description

Accepts certificates if they are expired but the pinned public key matches the expected public key.

Closes WAL-760.

## Test plan

Added a unit test.

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-storage-node-client/src/client/builder.rs
+++ b/crates/walrus-storage-node-client/src/client/builder.rs
@@ -80,7 +80,7 @@ impl StorageNodeClientBuilder {
     /// Passing true to this constructor disables that behaviour, resulting in expired certificates
     /// failing even if the public key is pinned.
     ///
-    /// This setting does not if not authenticating with the public key via
+    /// This setting has no effect if not authenticating with the public key via
     /// `authenticate_with_public_key` or `new_with_pinned_public_key`.
     pub fn disallow_expired_pinned_keys(mut self, is_disallowed: bool) -> Self {
         self.disallow_expired_pinned_keys = is_disallowed;


crates/walrus-storage-node-client/src/tls.rs
```diff
@@ -198,8 +195,9 @@ impl TlsCertificateVerifier {
 
     fn has_pinned_public_key(&self, end_entity: &CertificateDer<'_>) -> bool {
         let Some(ref expected_key) = self.public_key else {
-            return true;
+            return false;
```
halfprice, line 208: suggestion: presumably this was not a bug before (since you created `has_pinned_public_key_or_unset` which still treat key not set as a happy case). Can you document what exactly the returned bool indicates?

crates/walrus-storage-node-client/src/tls.rs
```diff
@@ -224,23 +226,28 @@ impl ServerCertVerifier for TlsCertificateVerifier {
...
-        // point, the only thing left to check is whether the public key is as expected.
-        if self.has_pinned_public_key(end_entity) {
-            Ok(verified)
-        } else {
-            Err(rustls::Error::General(
-                "the certificate presented by the server does not match the pinned public key"
-                    .to_owned(),
-            ))
+        match result {
+            Ok(verified) if !self.has_pinned_public_key_or_unset(end_entity) => {
```
halfprice, line 255: question: why not setting public key is ok?

crates/walrus-storage-node-client/src/tls.rs
jpcsmith: Correct, i changed the semantics of the function here. I added docstrings.

crates/walrus-storage-node-client/src/tls.rs
jpcsmith: Because it's not necessary to get the key from the chain and pin it in the client as expected. If you got the DNS name from the chain, then assuming the Web's PKI is operating correctly, your client will only trust the returned certificate if it was issued by a trusted Issuer for the name you're expecting.

crates/walrus-storage-node-client/src/client/builder.rs
```diff
@@ -70,6 +71,22 @@ impl StorageNodeClientBuilder {
...
+    /// Disallow expired certificates even if the public key is pinned.
+    ///
+    /// By default, pinning a public key allows connections to be established with a server using
+    /// that public key even if its certificate has expired. This is because the source of truth is
+    /// the public key and the server has the associated private key.
+    ///
+    /// Passing true to this constructor disables that behaviour, resulting in expired certificates
+    /// failing even if the public key is pinned.
+    ///
+    /// This setting does not if not authenticating with the public key via
```
mlegner: Nit: I guess a "matter" or "have an effect" is missing here?

PR #2822, opened 2026-01-07
chore: explicitly set localnet environment in all walrus packages in test to work with new package management system in Sui 1.63

halfprice:
## Description

This PR does two things:

1. resolve conflicts when updating sui to 1.63 in Walrus codebase

2. This is an effort to integrate the new package management system introduced in sui 1.63. Since the old package
management system is no longer available, we have to make this change in order to integrate Sui 1.63

However, there are some issue related to the new package management system that cause walrus contracts publising
not working in test setup. This is mostly due to that the ephemeral package publishing does not work well with
dependencies that are also an ephemeral published published package.

The work around in this PR is to manually set all the contracts to use a localnet environment, and update its chain ID
during each setup. We create a new Move.test.toml in all the contracts and will use its updated version to replace the
Move.toml in system setup.

Note that this is a temporary solution and should be resolved once the new package management system can work
with ephemeral packages.

Resolve WAL-821
Related to WAL-1125 WAL-1126 WAL-1127

## Test plan

---

## Release notes

- [ ] Storage node:
- [ ] Aggregator:
- [ ] Publisher:
- [ ] CLI:

### DIFF
--- a/crates/walrus-sui/src/client.rs
+++ b/crates/walrus-sui/src/client.rs
@@ -871,23 +871,11 @@ impl SuiContractClient {
     /// Returns the digest of the package at `package_path` for the active network identified by
     /// the enclosed wallet.
     pub async fn compute_package_digest(&self, package_path: PathBuf) -> SuiClientResult<[u8; 32]> {
-        // Compile package to get the digest.
-        let chain_id = self
-            .retriable_sui_client()
-            .get_chain_identifier()
-            .await
-            .ok();
-        tracing::info!(?chain_id, "chain identifier");
-        let (compiled_package, _build_config, _root_package) = system_setup::compile_package(
-            package_path,
-            Default::default(),
-            chain_id,
-            &self.inner.lock().await.wallet,
-        )
-        .await?;
+        let (compiled_package, _build_config, _root_package) = self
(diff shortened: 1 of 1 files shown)


contracts/subsidies/Move.test.toml
```diff
@@ -0,0 +1,15 @@
...
+implicit-dependencies = false
+
+[dependencies]
+std = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-framework/packages/move-stdlib", rev = "testnet-v1.63.1" }
+sui = { git = "https://github.com/MystenLabs/sui.git", subdir = "crates/sui-framework/packages/sui-framework", rev = "testnet-v1.63.1" }
+wal = { local = "../wal" }
+walrus = { local = "../walrus" }
+
+[environments]
+localnet = "ReplaceChainId"
```
halfprice, line 15: The way the hack works is to replace the existing Move.toml with this Move.test.toml, and update `ReplaceChainId` with the actual chain ID.

Why can't we just update the existing Move.toml to this? Because we want to maintain compatibility for any contracts that depend on Walrus contract.

wbbradley (approved):

PR #2829, opened 2026-01-08
feat: Add a way to override node config per network

sadhansood:
## Description

This PR adds per‑network config overrides baked into the binary. The final flow for config generation is:
1. Parse the user yaml to a Value and detect the network kind (system/staking IDs or cfg!(test)/cfg!(msim) for Tests and Simtest respectively).
2. Load the corresponding `overrides/<network>.yaml` (compiled in the binary).
3. Deep‑merge user yaml over the overrides (user values win over network overrides).
4. Deserialize into `StorageNodeConfig` so serde defaults fill missing fields, then validate the config.

## Test plan

Added unit tests covering merge semantics and network‑kind detection (including Tests overrides).

### DIFF
--- a/crates/walrus-service/src/node/config.rs
+++ b/crates/walrus-service/src/node/config.rs
@@ -1813,6 +1813,7 @@ mod tests {
         assert_eq!(
             config.blob_recovery.monitor_interval,
             Duration::from_secs(5),
+            // In unit tests, default_network_kind() resolves to Tests, which uses a 5s interval.
             "network defaults apply when user omits a field"
         );
         assert_eq!(
@@ -1835,7 +1836,7 @@ mod tests {
         let cases = vec![
             (network_overrides::NetworkKind::Mainnet, mainnet_ids),
             (network_overrides::NetworkKind::Testnet, testnet_ids),
-            (network_overrides::NetworkKind::Default, default_ids),
+            (network_overrides::default_network_kind(), default_ids),
         ];
 
         let dir = TempDir::new()?;
@@ -1850,7 +1851,14 @@ mod tests {
             "},
             );
             let config_path = dir.path().join(format!("node-{expected_kind:?}.yaml"));
-            std::fs::write(&config_path, yaml)?;
+            std::fs::write(&config_path, yaml.clone())?;
+
+            let raw_value: serde_yaml::Value = serde_yaml::from_str(&yaml)?;
+            assert_eq!(
+                network_overrides::detect_network_kind(&raw_value),
+                expected_kind,
+                "network kind is detected from contract ids"
+            );
 
             let config = StorageNodeConfig::load_config(&config_path)?;
 
@@ -1863,7 +1871,7 @@ mod tests {
 
             let expected_defaults = network_overrides::defaults_for(expected_kind);
             assert_eq!(
-                config.pending_metadata_cache, expected_defaults.pending_metadata_cache,
+                config.live_upload_deferral.enabled, expected_defaults.live_upload_deferral.enabled,
                 "default-only values are preserved"
             );
         }
--- a/crates/walrus-service/src/node/network_overrides.rs
+++ b/crates/walrus-service/src/node/network_overrides.rs
@@ -159,6 +159,43 @@ mod tests {
     }
 
     #[test]
+    fn known_network_ids_match_expected() {
+        const MAINNET_SYSTEM_OBJECT: &str =
+            "0x2134d52768ea07e8c43570ef975eb3e4c27a39fa6396bef985b5abc58d03ddd2";
+        const MAINNET_STAKING_OBJECT: &str =
+            "0x10b9d30c28448939ce6c4d6c6e0ffce4a7f8a4ada8248bdad09ef8b70e4a3904";
+        const TESTNET_SYSTEM_OBJECT: &str =
+            "0x6c2547cbbc38025cf3adac45f63cb0a8d12ecf777cdc75a4971612bf97fdf6af";
+        const TESTNET_STAKING_OBJECT: &str =
+            "0xbe46180321c30aab2f8b3501e24048377287fa708018a5b7c2792b35fe339ee3";
+
+        let known = known_network_ids();
+        let mainnet = known.mainnet.as_ref().expect("mainnet ids available");
+        let testnet = known.testnet.as_ref().expect("testnet ids available");
+
+        assert_eq!(
+            mainnet.system_object,
+            ObjectID::from_hex_literal(MAINNET_SYSTEM_OBJECT)
+                .expect("valid mainnet system object id")
+        );
+        assert_eq!(
+            mainnet.staking_object,
+            ObjectID::from_hex_literal(MAINNET_STAKING_OBJECT)
+                .expect("valid mainnet staking object id")
+        );
+        assert_eq!(
+            testnet.system_object,
+            ObjectID::from_hex_literal(TESTNET_SYSTEM_OBJECT)
+                .expect("valid testnet system object id")
+        );
+        assert_eq!(
+            testnet.staking_object,
+            ObjectID::from_hex_literal(TESTNET_STAKING_OBJECT)
+                .expect("valid testnet staking object id")
+        );
+    }
+
+    #[test]
     fn test_detects_network_kind_from_ids() -> Result<()> {
         let known = known_network_ids();
         let mainnet = known.mainnet.as_ref().expect("mainnet ids available");
--- a/crates/walrus-utils/src/config.rs
+++ b/crates/walrus-utils/src/config.rs
@@ -84,15 +84,4 @@ pub trait Config: DeserializeOwned {
     fn validate(&self) -> anyhow::Result<()> {
         Ok(())
     }
-
-    /// Loads the configuration from a file and validates it.
-    ///
-    /// # Errors
-    ///
-    /// Returns an error if the configuration cannot be read or parsed, or if the config is invalid.
-    fn load_and_validate(path: impl AsRef<Path>) -> anyhow::Result<Self> {
-        let config: Self = super::load_from_yaml(path)?;
-        config.validate()?;
-        Ok(config)
-    }
 }
--- a/setup/client_config_mainnet.yaml
+++ b/setup/client_config_mainnet.yaml
@@ -1,3 +1,5 @@
+# NOTE: walrus-service uses these IDs to detect network defaults. Changing them changes node
+# behavior and must be coordinated.
 system_object: 0x2134d52768ea07e8c43570ef975eb3e4c27a39fa6396bef985b5abc58d03ddd2
 staking_object: 0x10b9d30c28448939ce6c4d6c6e0ffce4a7f8a4ada8248bdad09ef8b70e4a3904
 n_shards: 1000
--- a/setup/client_config_testnet.yaml
+++ b/setup/client_config_testnet.yaml
@@ -1,3 +1,5 @@
+# NOTE: walrus-service uses these IDs to detect network defaults. Changing them changes node
+# behavior and must be coordinated.
 system_object: 0x6c2547cbbc38025cf3adac45f63cb0a8d12ecf777cdc75a4971612bf97fdf6af
 staking_object: 0xbe46180321c30aab2f8b3501e24048377287fa708018a5b7c2792b35fe339ee3
 exchange_objects:


crates/walrus-service/node_config_example.yaml
```diff
@@ -93,7 +93,7 @@ db_config:
 live_upload_deferral:
   enabled: false
   buckets: []
-  max_total_defer_secs: 300
+  max_total_defer_millis: 300000
```
halfprice: remark: is this a breaking change? I saw that at least our testnet node is using the old name.

crates/walrus-service/bin/node.rs
```diff
@@ -506,7 +504,7 @@ fn main() -> anyhow::Result<()> {
             ignore_sync_failures: _,
         } => loop {
             tracing::info!("loading node configuration from {}", config_path.display());
-            let config = StorageNodeConfig::load_and_validate(&config_path)?;
```
halfprice, line 509: nit: seems load_and_validate function can be removed.

crates/walrus-service/src/node/config.rs
```diff
@@ -267,6 +268,25 @@ impl walrus_utils::config::Config for StorageNodeConfig {
 }
 
 impl StorageNodeConfig {
+    /// Loads the config from a file, applying per-network overrides before deserialization.
+    pub fn load_config(path: impl AsRef<Path>) -> anyhow::Result<Self> {
+        let path = path.as_ref();
+        let config_str = std::fs::read_to_string(path)
+            .with_context(|| format!("unable to load config from {}", path.display()))?;
+        let raw_value: serde_yaml::Value = serde_yaml::from_str(&config_str)
```
halfprice, line 380: question: I guess the reason we are reading the raw_value instead of loading the config into a StorageNodeConfig struct is to differentiate default values vs. user explicitly set values. Is that correct? If so, I think it worth adding a comment here.

crates/walrus-service/src/node/network_overrides.rs
```diff
@@ -0,0 +1,258 @@
...
+    testnet: Option<ContractIds>,
+}
+
+fn known_network_ids() -> &'static KnownNetworkIds {
+    static KNOWN: OnceLock<KnownNetworkIds> = OnceLock::new();
+    KNOWN.get_or_init(|| KnownNetworkIds {
+        mainnet: load_contract_ids("mainnet", MAINNET_CLIENT_CONFIG_YAML),
+        testnet: load_contract_ids("testnet", TESTNET_CLIENT_CONFIG_YAML),
+    })
+}
```
halfprice, line 92: remark: can we simply hard code the system object ID and staking object ID in the code? This makes the node not need to rely on the external example config.

crates/walrus-service/node_config_example.yaml
halfprice: I saw that your custom serde function deals with the two versions. But why the name change?

crates/walrus-service/src/node/config.rs
```diff
@@ -832,42 +870,148 @@ impl Default for BlobEventProcessorConfig {
...
+    buckets: Vec<SizeDeferralEntry>,
+    #[serde(default)]
+    max_total_defer_millis: Option<u64>,
+    #[serde(default)]
+    max_total_defer_secs: Option<u64>,
+    #[serde(default = "default_max_checkpoint_lag")]
+    max_checkpoint_lag: u64,
+}
+
+impl<'de> Deserialize<'de> for LiveUploadDeferralConfig {
```
halfprice: remark: why this is needed and can we not do this? This looks fragile to me... (same for SizeDeferralEntryYaml)

crates/walrus-service/src/node/network_overrides.rs
```diff
@@ -0,0 +1,258 @@
...
+    "/../../overrides/testnet.yaml"
+));
+const TESTS_OVERRIDES_YAML: &str = include_str!(concat!(
+    env!("CARGO_MANIFEST_DIR"),
+    "/../../overrides/tests.yaml"
+));
+const SIMTEST_OVERRIDES_YAML: &str = include_str!(concat!(
+    env!("CARGO_MANIFEST_DIR"),
+    "/../../overrides/simtest.yaml"
+));
```
halfprice: suggestion: instead of defining the defaults in a yaml file, I wonder if we explicitly set it in the code is more robust. For example, we could create

```
impl StorageNodeConfig {
   default() -> Self
   default_mainnet() -> Self
   default_testnet() -> Self
   default_simtest() -> Self
   default_test() -> Self
   default_testbed() -> Self
}
```

The reasons this may be better are
1. if we define the default in the code, it's easier to change and don't need to worry about the format is correct or not when updating the default.
2. simtest/integration tests do not load from yaml, but directly create StorageNodeConfig ([test config creation](https://github.com/MystenLabs/walrus/blob/2750f167255dfbf7e0956d3ffd241df98d363096/crates/walrus-service/src/test_utils.rs#L1091), [simtest config creation](https://github.com/MystenLabs/walrus/blob/2750f167255dfbf7e0956d3ffd241df98d363096/crates/walrus-service/src/test_utils.rs#L1255), [another test config creation](https://github.com/MystenLabs/walrus/blob/2750f167255dfbf7e0956d3ffd241df98d363096/crates/walrus-service/src/test_utils.rs#L3156), [testbed config creation](https://github.com/MystenLabs/walrus/blob/2750f167255dfbf7e0956d3ffd241df98d363096/crates/walrus-service/src/testbed.rs#L707)

crates/walrus-service/src/node/config.rs
```diff
@@ -1613,6 +1757,64 @@ mod tests {
         Ok(())
     }
 
+    #[test]
+    fn merged_config_resolves_precedence_and_defaults() -> TestResult {
```
halfprice, line 1762: Thanks for adding this test to test merge_yaml!

Just to make things very reliable, which is important in this PR, can we add a test to test StorageNodeConfig::load_config e2e? What I imagine this can do is to (1) create a example yaml config in code (2) set the system object and staking object to mainnet, testnet, random value respectively, and (3) assert that after loading, mainnet should have some non-set value using mainnet default (same for testnet, default), and the value set in yaml is present and match despite the network (basically testing override).

This also gives us the confidence that mainnet config will definitely be identified as mainnet config, which is an important aspect to test.

crates/walrus-service/src/node/network_overrides.rs
sadhansood: Hmm, we’re including `setup/client_config_{mainnet,testnet}.yaml` at build time, not runtime. This avoids duplicating canonical IDs in two places and reduce the risk of drifting, if we really want the hardcode I'd prefer a single const source but given we already have this yaml which we test and verify to be correct - I am not seeing a strong reason to create another hardcoded value. In summary, I feel like this yaml is already doing the job well!

crates/walrus-service/bin/node.rs
sadhansood: correct, `load_and_validate ` doesn't exist anymore.

crates/walrus-service/src/node/config.rs
sadhansood: Added a comment explaining why we parse `serde_yaml::Value` first i.e. to detect network and preserve user‑set fields before defaults are applied.

crates/walrus-service/src/node/config.rs
sadhansood: Done, added `load_config_e2e_detects_network_kind_from_ids` and `load_config_applies_network_defaults_and_user_values` to test this.

crates/walrus-service/node_config_example.yaml
sadhansood: this is not being used anymore

crates/walrus-service/src/node/config.rs
sadhansood: removed it

crates/walrus-service/src/node/network_overrides.rs
sadhansood: done

crates/walrus-service/src/node/config.rs
```diff
@@ -1613,6 +1758,194 @@ mod tests {
...
+            assert_eq!(config.commission_rate, 9, "user values override defaults");
+            assert_eq!(
+                config.blob_recovery.monitor_interval,
+                Duration::from_secs(2),
+                "user values override defaults"
+            );
+
+            let expected_defaults = network_overrides::defaults_for(expected_kind);
+            assert_eq!(
+                config.pending_metadata_cache, expected_defaults.pending_metadata_cache,
```
halfprice: remark: this doesn't really test that the corresponding defaults are used, since `pending_metadata_cache` is the same for all network. How about let's set the default `StorageNodeConfig::rest_graceful_shutdown_period_secs` field in each default config such as "mainnet_node", "testnet_node", "default_node", etc. 

```
pub fn default_mainnet() -> Self {
        let mut config = Self::default();

        // Explicitly setting a default rest_graceful_shutdown_period_secs to test per network config defaults. Updating here should also update XXX test correspondingly.
        config.rest_graceful_shutdown_period_secs = Some(Some(REST_GRACEFUL_SHUTDOWN_PERIOD_SECS));
        config
    }
```

And set rest_graceful_shutdown_period_secs to REST_GRACEFUL_SHUTDOWN_PERIOD_SECS + 1 in testnet, etc.

And in the test here, we can use this to test if the corresponding network's config is applied.

This also gives an example of setting default value in each network config.

crates/walrus-service/bin/node.rs
halfprice: I don't see that it is removed in this PR.

crates/walrus-service/src/node/config.rs
```diff
@@ -912,9 +1056,9 @@ impl LiveUploadDeferralConfig {
             enabled: true,
             buckets: vec![SizeDeferralEntry {
                 max_unencoded_bytes: u64::MAX,
-                defer: Duration::from_millis(100),
+                defer: Duration::from_secs(1),
             }],
-            max_total_defer: Duration::from_millis(100),
+            max_total_defer: Duration::from_secs(1),
```
halfprice, line 1061: question: why do we need to change this value in test?

crates/walrus-service/src/node/network_overrides.rs
halfprice: I agree that duplicating is not ideal, but at its current form, it's not obvious that the setup/client_<network>_config.yaml is critical for storage node operation. If anyone changes setup/client_<network>_config.yaml for whatever reason that looks legit, and they shift away from the real contract ids used in prod, we may not notice that. This will change all the network default, and there is no warning about it. The downside of this can be catastrophic.

But if in the code, we clearly define const mainnet_contract_id, testnet_contract_id, with clearly comments that they should never be modified. Or explicitly create a config file under something like prod/<network>_contract_id.yaml in git with rules that anyone who changes these file needs more strict permission, it's less likely to make mistakes.

crates/walrus-service/src/node/config.rs
```diff
@@ -1613,6 +1758,194 @@ mod tests {
...
+        let config = StorageNodeConfig::load_config(&config_path)?;
+
+        assert!(
+            !config.live_upload_deferral.enabled,
+            "user values override network defaults"
+        );
+        assert_eq!(
+            config.blob_recovery.monitor_interval,
+            Duration::from_secs(5),
+            "network defaults apply when user omits a field"
```
halfprice, line 1817: This uses `default_test` default, is that correct? Worth documenting it since the simtest_default also use this value.

halfprice (approved):
Thanks a lot @sadhansood for the revision and it LGTM mostly! I only have a few minor comments now and feel free to merge it afterwards.

crates/walrus-service/src/node/config.rs
sadhansood: Good call on the test signal. I don't think we need to mutate mainnet/testnet defaults just to create a test delta. Instead I updated the e2e test to (1) assert ID‑based detection explicitly for mainnet and testnet and (2) validate default‑only values via a field that actually differs in the test default (`live_upload_deferral.enabled`). That exercises the load‑config merge path while keeping production defaults intact.

crates/walrus-service/src/node/config.rs
sadhansood: This change is only to preserve a non‑zero deferral after the new defaults merge path. DurationSeconds can’t represent 100ms, so it would round‑trip to 0 and effectively disable deferral in tests and simtests.  
```
pub struct SizeDeferralEntry {
    /// Maximum unencoded blob size (inclusive) in bytes for this bucket.
    pub max_unencoded_bytes: u64,
    /// Deferral duration to apply for this bucket, in seconds.
    #[serde(rename = "defer_secs")]
    pub defer: Duration,
}
```

crates/walrus-service/bin/node.rs
sadhansood: removed now, thanks for catching it!

crates/walrus-service/src/node/config.rs
sadhansood: yes, added comment.

crates/walrus-service/src/node/network_overrides.rs
sadhansood: Added a test `known_network_ids_match_expected` to detect any change in the network constants and also added comments on the top of these files.

PR #2830, opened 2026-01-08
chore: remove variant from LazySuiClientBuilder (enum -> struct)

wbbradley:
This is a refactoring change which eliminates several dead branches in the code, which will make it simpler to integrate a dual-read feature path for the eventual gRPC client integration.

### DIFF
--- a/crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
+++ b/crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
@@ -139,13 +139,8 @@ impl LazyClientBuilder<SuiClient> for LazySuiClientBuilder {
             sui_macros::fail_point_arg!(
                 "failpoint_sui_client_build_client",
                 |url_to_fail: String| {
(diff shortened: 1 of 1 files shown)


crates/walrus-sui/src/client/retry_client/retriable_sui_client.rs
```diff
@@ -109,45 +108,18 @@ const RPC_MAX_TRIES: u32 = 3;
...
-                .debug_tuple("Url")
-                .field(rpc_url)
-                .field(request_timeout)
-                .finish(),
-            Self::Client(_) => f.debug_tuple("Client").finish(),
-        }
-    }
-}
-
-impl From<SuiClient> for LazySuiClientBuilder {
```
halfprice, line 141: Hmm, this is never used?

wbbradley: Nope! We are going entirely from URLs now.

